Issue #1200, opened 2019-08-20
Scenes, external events, external layouts, extentions all can be without name


Bouh:
## Describe the bug
Scenes, external events, external layouts, extentions all can be without name.
One thing is strange extension have a different behavior when you try to rename it.
He warning about forbidden characters. 

## To Reproduce
1. Open an project
2. Add a new scene and rename it but delete just all letters and valid.
4. See scene without name

![image](https://user-images.githubusercontent.com/1670670/63391206-f2b3ad00-c3b1-11e9-8603-aadba7bb147e.png)

## Other details
* W10 b79 and lower

4ian:
Right, so scenes/external events/external layouts should be able to have any name (with special characters). In theory, an empty name is also working (for the game engine, it's just a scene called "") but as it's strange and breaking the UI, better add something to forbid it

For extensions, the name has more constraints - as it will be used for code generation. An empty name should be forbidden (because this might or might not actually lead to issue with code generation). There is already a check that is calling `gd.Project.validateObjectName` - a function that was made at the beginning for object but that should be renamed because it's used to validate other things now like parameters, objects and extension names (in short, anything that can be used for code generation). This is a C++ method, I'll fix it to reject empty names.

For the scenes/external layouts/events, would be great if you or someone else has time to add a warning. There should be one already in case you try to use a name that is already taken by another  scene/external layout/event.

Bouh:
> Pour les scènes / mises en page externes / événements, serait bien si vous ou quelqu'un d'autre a le temps d'ajouter un avertissement. Il devrait y en avoir déjà un si vous essayez d'utiliser un nom déjà pris par une autre scène / mise en page / événement externe.

Done in #1203

PR #1203, opened 2019-08-21
Avoid empty and special name in project manager

Bouh:
#1200


4ian:
You went a bit too far, actually you just have to check if the new name is not empty ;)

Scenes/external events/layouts can be named with any name as long as it's not already used and not empty. Special characters are fine. Only the extensions can't be using special characters (it's already taken care of).

Bouh:
I've simplify code, with warning.

4ian:
Looks great! 😄

PR #1204, opened 2019-08-21
Improve context menu in project manager

Bouh:
This add button in context menu is for adding scene, external layout, external events, extensions because when users have lot of scenes (+60) he need scroll down list for add scene.

This this PR you can add scene just under scene clicked.

Related to 
https://forum.gdevelop-app.com/t/add-a-scene-button/20201

### DIFF
--- a/newIDE/app/src/ProjectManager/index.js
+++ b/newIDE/app/src/ProjectManager/index.js
@@ -164,23 +164,7 @@ class Item extends React.Component<ItemProps, {||}> {
             click: () => this.props.onDelete(),
           },
           {
-            label: 'Add new scene',
-            visible: this.props.typeAddedable === 'layout',
-            click: () => this.props.onAdd(),
-          },
-          {
-            label: 'Add external events',
-            visible: this.props.typeAddedable === 'external-events',
-            click: () => this.props.onAdd(),
-          },
-          {
-            label: 'Add external layout',
-            visible: this.props.typeAddedable === 'external-layout',
-            click: () => this.props.onAdd(),
-          },
-          {
-            label: 'Add external layout',
-            visible: this.props.typeAddedable === 'events-functions-extension',
+            label: 'Add a new one',
             click: () => this.props.onAdd(),
           },
           { type: 'separator' },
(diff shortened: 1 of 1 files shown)


blurymind:
can you put a screenshot :)

Bouh:
Right click on Scene1
Add scene => NewScene is created just under Scene1
Same for  external layout, external events, extensions
image

newIDE/app/src/ProjectManager/index.js
```diff
@@ -705,6 +794,8 @@ export default class ProjectManager extends React.Component<Props, State> {
                     }
                     onEdit={() => this.props.onOpenLayout(name)}
                     onDelete={() => this.props.onDeleteLayout(layout)}
+                    onAdd={() => this._onAdd('layout', name, i)}
```
4ian: Prefer calling directly `this._addLayout` (and same for the rest).
This removes a level of indirection/abstraction which is not really useful and less type safe (here it's not changing anything, but we could imagine that we pass at some point something that is typed like a `gd.Layout` to the function).

newIDE/app/src/ProjectManager/index.js
```diff
@@ -161,6 +163,26 @@ class Item extends React.Component<ItemProps, {||}> {
             label: 'Delete',
             click: () => this.props.onDelete(),
           },
+          {
```
4ian: I see the idea with `typeAddedable`, but I think it would be simpler to pass a "addLabel" that is the label to be used for the menu item to add something. If addLabel is null/undefined/empty string, we'll hide it :)

Bouh: I understand, what you mean with `addLabel` ?
Or
I can simplify and delete visible options, i keep just one label named `Add a new one` linked to 
`this.props.onAdd()`

```
{
     label: 'Add a new one',
     click: () => this.props.onAdd(),
 },
```

4ian: Consider the label as a prop, and only show the item if there is onAdd defined:

```
{
     label: this.props.addLabel,
     visible: !!this.props.onAdd, 
     click: () => this.props.onAdd(),
 },
```

newIDE/app/src/ProjectManager/index.js
```diff
@@ -760,6 +812,8 @@ export default class ProjectManager extends React.Component<Props, State> {
                     onDelete={() =>
                       this.props.onDeleteExternalEvents(externalEvents)
                     }
+                    typeAddedable={'external-events'}
```
4ian: onAdd is fine, but typeAddedable is now useless. Instead, pass the label that should be shown for the item:

```
addLabel={"Add a new external event"}
```

4ian:
For the label, see my comment: I'm thinking of passing, in addition to "onAdd", the label to display "addLabel". You don't need typeAddedable prop in this case, as you keep Item generic without having to pass it a type (the responsibility of knowing what to display is on the parent).

newIDE/app/src/ProjectManager/index.js
Bouh: Ok `!!` convert value in boolean and invert boolean, very usefull thank you !

Bouh:
Oops i've delete and refork Gdevelop repo, so my previous branch is now deleted i can't pull my new code here.

newIDE/app/src/ProjectManager/index.js
4ian: Yes it's a way to convert ("cast") an object to a boolean (so that null/undefined => false, any object or array => true).

Issue #1202, opened 2019-08-21
javascript API gdjs.evtTools.object.twoListsTest not working as expected


andineck:
## Describe the bug
I would like to write a custom condition function that checks two lists and found the `twoListTest` function I want to make use of in my function. I provide two lists, but the comparison does not take place.

## To Reproduce

my function:
```js
const objectsA = eventsFunctionContext.getObjects('objectsA');
const objectsB = eventsFunctionContext.getObjects('objectsB');
eventsFunctionContext.returnValue = gdjs.evtTools.object.twoListsTest(gdjs.RuntimeObject.collisionTest, objectsA, objectsB, false);
```
Note: I used gdjs.RuntimeObject.collisionTest as predicate just for simplicity.

`objectsA` and `objectsB` are at runtime both Arrays with `SpriteRuntimeObject`s.

When I debug the `twoListsTest` function, I see the Arrays in `objectsLists1`, but not in the converted `objects1Lists` (length === 0), and therefore no comparison takes place.

http://4ian.github.io/GD-Documentation/GDJS%20Runtime%20Documentation/events-tools_objecttools.js.html#line57

## Other details
I might do something wrong here, but I can't figure it out. 
In order to limit the selected objects in the further conditions and actions, do I have to set the `pick` attribute to true on the selected ones?

4ian:
This is not working because `gdjs.evtTools.object.twoListsTest` is expecting its arguments to be list**s** of objects, while for each one you provided a list of objects. See the difference? No, it's normal ^^

This is subtle and not greatly documented for now I must admit - ideally I'll add some JSDoc so that you have type information.

These "lists of objects" are stored using `Hashtable`, a helper to store a map of key/values. This is done because the events can actually be applied not only on a single list of objects, but on multiple lists of objects (like when you're running events on a group of objects).

```
const objectsA = eventsFunctionContext.getObjects('objectsA');
const objectsALists = Hashtable.newFrom({ObjectA: objectsA});
const objectsB = eventsFunctionContext.getObjects('objectsB');
const objectsBLists = Hashtable.newFrom({ObjectB: objectsB});
eventsFunctionContext.returnValue = gdjs.evtTools.object.twoListsTest(gdjs.RuntimeObject.collisionTest, objectsALists, objectsBLists, false);

// objectsALists and objectsBLists will be filtered to contain only the objects satisfying the predicate
```

[...]

PR #1205, opened 2019-08-21
add objecttools argument types

andineck:
As discussed in #1202 , documenting the parameters Types could be helpful.

### DIFF
--- a/GDJS/Runtime/events-tools/objecttools.js
+++ b/GDJS/Runtime/events-tools/objecttools.js
@@ -34,6 +34,15 @@ gdjs.evtTools.object.pickOnly = function(objectsLists, runtimeObject) {
 };
 
 /**
+ * A predicate to be passed to `gdjs.evtTools.object.twoListsTest`.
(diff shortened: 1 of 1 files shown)


4ian:
That was fast! I've added a commit to use `*` instead of `any` (I think in JSDoc, you have to use *). And added the typing for the predicate (could be done for other predicates too).

4ian:
Note that I've seen you've made your PR from your master branch. Avoid this and prefer creating a branch, because when I merge your PR, I "squash" the commits into a single one, so that your master will diverge from mine. Better work on a branch, then go back to master and pull my changes :)

andineck:
cool, thank you for merging, you are right `*` is correct, was confused with typescript. 
will do a branch next time :-)

4ian:
Yeah JSDoc is a bit different than typescript - but the goal is to keep using it and at some point add Typescript to do typechecking (as Typescript understand JSDoc annotation). So these kind of improvements to the types are very welcome!! :D

PR #1206, opened 2019-08-22
Add more layer effects

Wend1go:
- [x] Add filters present in PIXI.filters
- [ ] Add filters present in external repo https://github.com/pixijs/pixi-filters
- [x] Add Action/Condition to enable/disable filters

Some effects like the displacement filter need a sprite to work, so I left it out for now. But it would be very cool to use an existing image ressource for that.

The pixi-filters has a bunch of nice effects but it is located in a different repository. Since it needs to be ES5 complient, how would I have to add the external JS file without using the `import` command? 

The pixi filters all have a property `enabled` that could theoretically be set the same way as the other filter parameters but I think it would be cleaner to create a separate action and condition for that.
I just don't know yet in which file I need to add that condition/action to appear in the events list.

### DIFF
--- a/GDJS/Runtime/events-tools/cameratools.js
+++ b/GDJS/Runtime/events-tools/cameratools.js
@@ -129,24 +129,24 @@ gdjs.evtTools.camera.setLayerEffectParameter = function(runtimeScene, layer, eff
 }
 
 /**
- * Enable an effect.
- * @param {gdjs.RuntimeScene} runtimeScene The current scene
+ * Enable, or disable, an effect of a layer.
+ * @param {gdjs.RuntimeScene} runtimeScene The scene
  * @param {string} layer The name of the layer
  * @param {string} effect The name of the effect
- * @param {number} value The new value
+ * @param {boolean} enabled true to enable, false to disable.
  */
-gdjs.evtTools.camera.enableLayerEffect = function(runtimeScene, layer, effect, value) {
+gdjs.evtTools.camera.enableLayerEffect = function(runtimeScene, layer, effect, enabled) {
     if ( !runtimeScene.hasLayer(layer) ) { return; }
 
-    runtimeScene.getLayer(layer).enableEffect(effect, value);
+    runtimeScene.getLayer(layer).enableEffect(effect, enabled);
 }
 
 /**
- * Enable an effect.
- * @param {gdjs.RuntimeScene} runtimeScene The current scene
+ * Check if an effect is enabled.
+ * @param {gdjs.RuntimeScene} runtimeScene The scene
  * @param {string} layer The name of the layer
- * @param {string} filter The name of the effect
- * @return {boolean} Filter is enabled
+ * @param {string} effect The name of the effect
+ * @return {boolean} true if the effect is enabled, false otherwise.
  */
 gdjs.evtTools.camera.layerEffectEnabled = function(runtimeScene, layer, effect) {
     if ( !runtimeScene.hasLayer(layer) ) { return true; }
--- a/GDJS/Runtime/events-tools/cameratools.js
+++ b/GDJS/Runtime/events-tools/cameratools.js
@@ -128,6 +128,12 @@ gdjs.evtTools.camera.setLayerEffectParameter = function(runtimeScene, layer, eff
     return runtimeScene.getLayer(layer).setEffectParameter(effect, parameter, value);
 }
 
+gdjs.evtTools.camera.enableLayerEffect = function(runtimeScene, layer, effect, value) {
+    if ( !runtimeScene.hasLayer(layer) ) { return; }
+
+    return runtimeScene.getLayer(layer).enableEffect(effect, value);
+}
+
 gdjs.evtTools.camera.setLayerTimeScale = function(runtimeScene, layer, timeScale) {
     if ( !runtimeScene.hasLayer(layer) ) { return; }
     return runtimeScene.getLayer(layer).setTimeScale(timeScale);
--- a/GDJS/Runtime/events-tools/cameratools.js
+++ b/GDJS/Runtime/events-tools/cameratools.js
@@ -128,10 +128,30 @@ gdjs.evtTools.camera.setLayerEffectParameter = function(runtimeScene, layer, eff
     return runtimeScene.getLayer(layer).setEffectParameter(effect, parameter, value);
 }
 
+/**
+ * Enable an effect.
+ * @param {gdjs.RuntimeScene} runtimeScene The current scene
+ * @param {string} layer The name of the layer
+ * @param {string} effect The name of the effect
+ * @param {number} value The new value
+ */
 gdjs.evtTools.camera.enableLayerEffect = function(runtimeScene, layer, effect, value) {
     if ( !runtimeScene.hasLayer(layer) ) { return; }
 
-    return runtimeScene.getLayer(layer).enableEffect(effect, value);
+    runtimeScene.getLayer(layer).enableEffect(effect, value);
+}
+
+/**
+ * Enable an effect.
+ * @param {gdjs.RuntimeScene} runtimeScene The current scene
+ * @param {string} layer The name of the layer
+ * @param {string} filter The name of the effect
+ * @return {boolean} Filter is enabled
+ */
+gdjs.evtTools.camera.layerEffectEnabled = function(runtimeScene, layer, effect) {
+    if ( !runtimeScene.hasLayer(layer) ) { return 1; }
+
+    return runtimeScene.getLayer(layer).isEffectEnabled(effect);
 }
 
 gdjs.evtTools.camera.setLayerTimeScale = function(runtimeScene, layer, timeScale) {
--- a/GDJS/Runtime/events-tools/cameratools.js
+++ b/GDJS/Runtime/events-tools/cameratools.js
@@ -149,7 +149,7 @@ gdjs.evtTools.camera.enableLayerEffect = function(runtimeScene, layer, effect, v
  * @return {boolean} Filter is enabled
  */
 gdjs.evtTools.camera.layerEffectEnabled = function(runtimeScene, layer, effect) {
-    if ( !runtimeScene.hasLayer(layer) ) { return 1; }
+    if ( !runtimeScene.hasLayer(layer) ) { return true; }
 
     return runtimeScene.getLayer(layer).isEffectEnabled(effect);
 }
--- a/GDJS/Runtime/layer.js
+++ b/GDJS/Runtime/layer.js
@@ -216,6 +216,10 @@ gdjs.Layer.prototype.setEffectParameter = function(name, parameterIndex, value) 
     return this._renderer.setEffectParameter(name, parameterIndex, value);
 };
 
+gdjs.Layer.prototype.enableEffect = function(name, value) {
+    return this._renderer.enableEffect(name, value);
+};
+
 gdjs.Layer.prototype.setEffectsDefaultParameters = function() {
     for (var i = 0; i < this._effects.length; ++i) {
         var effect = this._effects[i];
--- a/GDJS/Runtime/layer.js
+++ b/GDJS/Runtime/layer.js
@@ -218,17 +218,17 @@ gdjs.Layer.prototype.setEffectParameter = function(name, parameterIndex, value) 
 
 /**
  * Enable or disable an effect.
- * @param {string} filter The name of the effect to enable or disable.
- * @param {boolean} value true to enable, false to disable
+ * @param {string} effect The name of the effect to enable or disable.
+ * @param {boolean} enable true to enable, false to disable
  */
-gdjs.Layer.prototype.enableEffect = function(name, value) {
-    this._renderer.enableEffect(name, value);
+gdjs.Layer.prototype.enableEffect = function(name, enable) {
+    this._renderer.enableEffect(name, enable);
 };
 
 /**
  * Check if an effect is enabled
- * @param {string} filter The name of the filter effect
- * @return {boolean} Filter is enabled
+ * @param {string} effect The name of the effect
+ * @return {boolean} true if the effect is enabled, false otherwise.
  */
 gdjs.Layer.prototype.isEffectEnabled = function(name) {
     return this._renderer.isEffectEnabled(name);
--- a/GDJS/Runtime/layer.js
+++ b/GDJS/Runtime/layer.js
@@ -217,8 +217,8 @@ gdjs.Layer.prototype.setEffectParameter = function(name, parameterIndex, value) 
 };
 
 /**
- * Enable an effect.
- * @param {string} filter The name of the filter effect
+ * Enable or disable an effect.
+ * @param {string} filter The name of the effect to enable or disable.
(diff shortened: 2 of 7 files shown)


4ian:
> The pixi filters all have a property enabled that could theoretically be set the same way as the other filter parameters but I think it would be cleaner to create a separate action and condition for that.

Yes I think a separate action would be more interesting.

> The pixi-filters has a bunch of nice effects but it is located in a different repository. Since it needs to be ES5 complient, how would I have to add the external JS file without using the import command?

We're starting to reach a point where we would have a lot of potential effects, which could end up taking a "large" amount of space on disk. To keep exported games as lean as possible, I would rather think about a system to have extensions to be able to declare new effects :) 
Right now as you've seen they are all hardcoded in the game engine and hardcoded in the IDE.
The idea would be to:
* Add an API to Extension (in C++ and JS) to declare effects and the JS file to include.
* Allow the IDE to read these effects declared in extensions.
* Allow the code generation to only include the effects used in layers (or in the future, objects :))
* Allow the JS game engine to use these effects (probably like done for objects and behaviors, by scanning things inside the `gdjs` "namespace", searching for a special property indicating the presence of an effect).

I can see that all filters from Pixi filters (https://github.com/pixijs/pixi-filters/tree/master/filters) are in separate files, so that would make sense to have them in an extension but with multiple include files - so that only the filters actually used are included in the exported game :)

I'm fine for now going ahead with the effects bundled with Pixi.js - because they won't be impacting the size of the game files as already included.
To add more, like the one from pixi-filters, I think it would be interesting to see if we can add them using this extension system like behaviors/objects.

GDJS/Runtime/pixi-renderers/pixi-filters-tools.js
```diff
@@ -96,6 +96,43 @@ gdjs.PixiFiltersTools._filters = {
...
+            filter.noise = value;
+        },
+    },
+    Blur: {
+        makeFilter: function() {
+            var blur = new PIXI.filters.BlurFilter();
+            return blur;
+        },
+        updateParameter: function(filter, parameterName, value) {
+            if (parameterName !== 'blur' &&
```
4ian: While it's not done for existing effects, I think it would be worth adding a bit of extra check to avoid invalid values to be set for the parameters. What do you think? For example, return if the kernelSize is not a valid value, or value is out of range, etc.

4ian:
Almost forgot to say it: can't wait for these sweet shiny new shaders to be in GDevelop 🤗Will unlock a tons of cool visual effects.

GDJS/Runtime/pixi-renderers/pixi-filters-tools.js
Wend1go: Yep makes sense.
I have already implemented a check for the kernelSize in [EffectDescription.js](https://github.com/4ian/GDevelop/pull/1206/commits/c1da506b419e862c4fcbfdca0693b9d3629df41e#diff-6c71ccf608de8a3035773caafdc2230aR15) 
Moved the checks to `pixi-filters-tools.js` so that we don't have to declare the same check functions twice.

Wend1go:
Distributing filters in a modular way via Extension sounds good. Since my hollidays are approaching, I won't be able to work on this during the upcoming weeks. So take you time.
I'll try to add the enable/disable functionality at the weekend, feel free to merge this after that if you intend to do a new release. I can open a new PR for the additional effects later.

newIDE/app/src/EffectsList/EffectDescription.js
```diff
@@ -11,8 +11,6 @@ export type EffectDescription = {|
   parameterDefaultValues: Array<EffectDefaultValue>,
 |};
 
-const clampValue = value => Math.max(0, Math.min(1, value));
```
4ian: Actually, both the editor and the game engine carry the responsibility of validating the user input, but for two different reasons:
- The game engine because we don't want to pass invalid values to the filters - that could make strange results or even things like crashs (in other words, no bugs).
- The editor because it's a bad user experience for the user to be able to enter invalid values - if we know about the valid values we should as possible avoid to store them.

It's of course entirely possible for the user to enter invalid values using events, but that's why we have the checks in the game engine. The second checks are the editor best effort to avoid you entering non sense things :)

GDJS/Runtime/pixi-renderers/pixi-filters-tools.js
```diff
@@ -1,5 +1,8 @@
 gdjs.PixiFiltersTools = function() {};
 
+const clampValue = (value, min, max) => Math.max(min, Math.min(max, value));
```
4ian: We're in the game engine, so ES5 only ;)
`var clampValue = function(value, min, max) { return Math.max(min, Math.min(max, value)); }`

Also, remember that classical JavaScript is a pain when it comes to global variable and variable scope. Using var to declare a variable outside of any scope like this will create a global variable, both polluting the global namespace and risking our function to be erased by something else. We don't want to run this risk, so instead be sure to always declare things inside functions (in which case they are function scoped) or attach them as "static" members:

`gdjs.PixiFiltersTools.clampValue = function(value, min, max) { return Math.max(min, Math.min(max, value)); }`;

then use it as `filter.uniforms[parameterName] = gdjs.PixiFiltersTools.clampValue(value, 0, 1);`

More verbose, but that's the price of a nice "self contained" code that won't break even if someone mess with the global namespace.

GDJS/Runtime/pixi-renderers/pixi-filters-tools.js
```diff
@@ -1,5 +1,8 @@
...
+const clampKernelSize = value => ([5, 7, 9, 11, 13, 15].includes(value)) ? value : 5;
```
4ian: Same for this function. 
`includes` is also fairly recent and could create issue with old browser that we still support (read: Android 4.x). Use indexOf instead: `[5, 7, 9, 11, 13, 15].indexOf(value) !== -1`

4ian: Note that in newIDE, we're running with Babel, so all the latest features of JS can be used, and using webpack to bundle JS, so declare function/variables anywhere is fine because they will be scoped to the JS file - no risk of collision.

GDJS/Runtime/pixi-renderers/pixi-filters-tools.js
4ian: (note that the "someone mess with the global namespace" could be actually another extension declaring a `clampValue` with a different implementation - imagine the strange bug and how hard to debug it is "my game effects are broken when I use this unrelated extension")

4ian:
> I'll try to add the enable/disable functionality at the weekend, feel free to merge this after that if you intend to do a new release. I can open a new PR for the additional effects later.

Yep, both sounds good :)
- Check my comments, but apart from this I'm fine merging this as it's adding new filters without any price in terms of JS size, as they are already embedded in Pixi and so the exported games.
- Enable/disable will be a "bonus" if you have time, otherwise again it's already ok to go with this. You can even start another PR to avoid polluting this one and be sure I can merge it.
- I'll work on enabling extensions to provide effects next week or so.

GDJS/Runtime/pixi-renderers/pixi-filters-tools.js
4ian: Wrote about this here: https://github.com/4ian/GDevelop/commit/5c648e3f2b8c444bd99221cea56b92c00cdddddd

newIDE/app/src/EffectsList/EffectDescription.js
Wend1go: So should we keep a second copy of the validation functions for the editor?

4ian: Yes let's keep a copy of it, one in the game engine and one in the editor. Will generally in programming, it's best to factor things, here it's better to have both of them because there serves a slightly different purpose and it's not harmful. Better have a copy rather than a bad abstraction :)

> Is it possible to check for valid values while entering them in the textbox?

Not really, the text fields are made so that changes are applied only when they are losing the focus. Doing validation during the writing is trickier and can lead to strange results (like, you enter an invalid value like 4, and it's getting correct immediately => people can think of a bug).

Wend1go:
Ready for a final review @4ian

GDJS/Runtime/events-tools/cameratools.js
```diff
@@ -128,6 +128,32 @@ gdjs.evtTools.camera.setLayerEffectParameter = function(runtimeScene, layer, eff
...
+
+/**
+ * Enable an effect.
+ * @param {gdjs.RuntimeScene} runtimeScene The current scene
+ * @param {string} layer The name of the layer
+ * @param {string} filter The name of the effect
+ * @return {boolean} Filter is enabled
+ */
+gdjs.evtTools.camera.layerEffectEnabled = function(runtimeScene, layer, effect) {
+    if ( !runtimeScene.hasLayer(layer) ) { return 1; }
```
4ian: `return true`

GDJS/Runtime/layer.js
```diff
@@ -216,6 +216,24 @@ gdjs.Layer.prototype.setEffectParameter = function(name, parameterIndex, value)
...
+ * Enable an effect.
+ * @param {string} filter The name of the filter effect
+ * @param {boolean} value true to enable, false to disable
+ */
+gdjs.Layer.prototype.enableEffect = function(name, value) {
+    this._renderer.enableEffect(name, value);
+};
+
+/**
+ * Enable an effect.
```
4ian: "Check if an effect is enabled"

GDJS/Runtime/layer.js
```diff
@@ -216,6 +216,24 @@ gdjs.Layer.prototype.setEffectParameter = function(name, parameterIndex, value)
     return this._renderer.setEffectParameter(name, parameterIndex, value);
 };
 
+/**
+ * Enable an effect.
```
4ian: "Enable or disable an effect."

GDJS/Runtime/layer.js
```diff
@@ -216,6 +216,24 @@ gdjs.Layer.prototype.setEffectParameter = function(name, parameterIndex, value)
...
+ * @param {string} filter The name of the filter effect
```
4ian: "The name of the effect to enable or disable."

The term "filter" is Pixi.js specific, in GDevelop I think it's better to keep the metaphor of the "effect" everywhere.

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
```diff
@@ -110,3 +110,17 @@ gdjs.LayerPixiRenderer.prototype.setEffectParameter = function (name, parameterN
...
+
+gdjs.LayerPixiRenderer.prototype.enableEffect = function (name, value) {
+    if (!this._filters.hasOwnProperty(name)) return;
+
+    var theFilter = this._filters[name];
+    gdjs.PixiFiltersTools.enableEffect(theFilter.filter, value);
+};
+
+gdjs.LayerPixiRenderer.prototype.isEffectEnabled = function (name) {
+    if (!this._filters.hasOwnProperty(name)) return 1;
```
4ian: `return false` - an effect that does not exists must be considered as not enabled.

GDJS/Runtime/pixi-renderers/pixi-filters-tools.js
```diff
@@ -93,11 +96,70 @@ gdjs.PixiFiltersTools._filters = {
...
+            }
+            
+            filter[parameterName] = value;
         },
     },
 };
 
+/**
+ * Enable an effect.
+ * @param {string} filter The name of the filter effect
```
4ian: This annotation is not right, it should ba a `{PIXI.Filter}` I think and the description should also be updated (`filter The filter to enable or disabled`)

GDJS/Runtime/pixi-renderers/pixi-filters-tools.js
```diff
@@ -93,11 +96,70 @@ gdjs.PixiFiltersTools._filters = {
...
+ * @param {boolean} value Set to true to enable, false to disable
+ */
+gdjs.PixiFiltersTools.enableEffect = function(filter, value) {
+    filter.enabled = value;
+}
+
+/**
+ * Enable an effect.
```
4ian: Same comments about description, filter parameter typing and description

GDJS/Runtime/pixi-renderers/pixi-filters-tools.js
```diff
@@ -93,11 +96,70 @@ gdjs.PixiFiltersTools._filters = {
...
+ * @param {string} filter The name of the filter effect
+ * @return {boolean} Filter is enabled
```
4ian: I'm usually more explicit about the values: `true` if the filter is enabled

Bouh:
Field value can be empty and crash preview
image

Wend1go:
Ups, I forgot to update the function calls to use the min/max parameters when reverting the changes. 🤦‍♂ 
Will correct this tonight.

Wend1go:
Ok, fixed now.
I added a brightness filter as a little present, to make amends. ;)
This one should be useful to crate screen transitions with a fade to/from black effect.

Since all the filters only work on the WebGL renderer, I think we should add a condition to check if WebGL is available so one can react to this and add a workaround if filters can't be used.
This could be achieved by checking the type of renderer (canvas / webgl)  that was automatically created when the game started.

4ian:
Sweet! Code looks good, I'll take another look and merge this tonight :)

> (Btw: What is the reason for only allowing a single effect per layer?)

If I remember properly in GD4 when starting to implement this, the native game engine based on SFML was only supporting a single shader on layers - so I decided to limit it for HTML5 game engines and find a solution for SFML before allowing more effects. Turns out that this kind of concern is not relevant anymore - maybe in the future we'll have a new native game engine or what not - but that's not a reason to keep the limitation.
I'll enable multiple effects after merging this (still need a bit of rework in the game engine)

> Field value can be empty and crash preview

I'll take a look

> At the bottom of the layer list we have just a icon +, in other list we have a little text like "Click to add a objet"

Right I started changes to have proper buttons in most lists, but still missing this one and variables. I'll take a look.

For the scrolling issues, it's true that it's not ideal. Would need to use ScrollView component I think to solve this (feel free to take a look, it's less of a priority for me right now).

4ian:
I've made a few last fixes. Merging now! Thanks a lot for working on this 👍👍

4ian:
I've not forgotten about adding effects from extensions - feel free to ping me about this again if I'm not starting this :) I've been busy working on upgrading the internal UI components to a new version (with denser lists, native selects, etc).

Bouh:
Indeed texture is needed for Displacement effect.
I think we can put a default texture in base64.

With the paramters torepeat (wrapmode) the texture.
It would also be great to be able to animate the positions and rotation of the texture. To have a moving effect on the screen.
There is little missing for this addition.

I managed to put the effect with the JSevent but I can't get it to move.

i think we can use file opener for import custom texture in this effect.
In any case texture should be a parameters too.

[displacementFilter.zip](https://github.com/4ian/GDevelop/files/3721596/displacementFilter.zip)

PR #1207, opened 2019-08-23
Add continuous deployment to a AWS S3 bucket

4ian:


4ian:
@Bouh You should be interested by this ;)

PR #1208, opened 2019-08-23
Improve context menu in manager

Bouh:
Main PR #1204 

Sorry for this but i've deleted my fork and all my PR are now broken with 
Bouh wants to merge 2 commits into 4ian:master from **unknown repository**
**unknown repository** arggg, i can't reclaim i guess it so i reopen this here...

### DIFF
--- a/newIDE/app/src/ProjectManager/index.js
+++ b/newIDE/app/src/ProjectManager/index.js
@@ -164,23 +164,7 @@ class Item extends React.Component<ItemProps, {||}> {
             click: () => this.props.onDelete(),
           },
           {
-            label: 'Add new scene',
-            visible: this.props.typeAddedable === 'layout',
-            click: () => this.props.onAdd(),
-          },
-          {
-            label: 'Add external events',
-            visible: this.props.typeAddedable === 'external-events',
-            click: () => this.props.onAdd(),
-          },
-          {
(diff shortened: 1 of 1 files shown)


newIDE/app/src/ProjectManager/index.js
```diff
@@ -882,6 +939,8 @@ export default class ProjectManager extends React.Component<Props, State> {
                         eventsFunctionsExtension
                       )
                     }
+                    addLabel={'Add a New Extension'}
+                    onAdd={() => this._addEventsFunctionsExtension(name, i)}
```
4ian: Just thinking about it now, `name` arguments are actually not used at all right? Can you remove them? :)

newIDE/app/src/ProjectManager/index.js
```diff
@@ -345,6 +352,50 @@ export default class ProjectManager extends React.Component<Props, State> {
     this.forceUpdate();
   };
 
+  _addLayout = (name: string, index: number) => {
```
4ian: name is not used?

newIDE/app/src/ProjectManager/index.js
Bouh: yes

4ian:
Looks good!

Issue #1212, opened 2019-08-25
PixiJS v4.8.6 outdated, Google chrome and webview cannot display video anymore


Bouh:
## Describe the bug
PixiJS 4.8.6 dosen't display video on Chrome because WebGL texture in Chromium project have been updated.

Google chrome and webview cannot display video anymore with old pixi lib.

## To Reproduce
1. Open video example
2. Run preview over wifi
3. Open game on Chrome or webview
4. Video have sound but no image
5. See warnings in console

I'am wondering if we need update PixiJS with the lastest build.

You can see at left PixiJS v4.8.6 used in game, warning come from line highlighted.
![image](https://user-images.githubusercontent.com/1670670/63650871-a4a2fe80-c74f-11e9-8a2a-e590aada011d.png)

[Error line 5487 from Chronium
](https://chromium.googlesource.com/chromium/src/third_party/+/master/blink/renderer/modules/webgl/webgl_rendering_context_base.cc?autodive=0%2F%2F%2F#5487)

## Other details
Since Chrome v76

4ian:
> I'am wondering if we need update PixiJS with the lastest build.

If that is the case, I guess we could find some issues reported on Pixi github?

Bouh:
Maybe  ?
https://github.com/pixijs/pixi.js/issues/5996

Bouh:
They say that sleep brings advices, it works on Chronium ecosystem and on mobile.

PR #1216, opened 2019-08-29
Fix video object for exports

Bouh:
#1212


4ian:
Sweet, can you add a comment, before the 2 lines, explaining why these lines are needed? You can put a link to a GitHub issue for example.

Thanks for investigating!

Bouh:
Voilà :)

PR #1220, opened 2019-09-06
Disable _invalidTexture assignment

Fermain:
This line is responsible for a 'bunny.png' 404 error in exported projects.

### DIFF
--- a/GDJS/Runtime/pixi-renderers/pixi-image-manager.js
+++ b/GDJS/Runtime/pixi-renderers/pixi-image-manager.js
@@ -14,7 +14,7 @@
 gdjs.PixiImageManager = function(resources)
 {
     this._resources = resources;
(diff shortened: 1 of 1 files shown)


4ian:
Thanks for the PR :)
This variable is still used at various other places to return a PIXI.Texture when a resource is not found. See for example:

https://github.com/4ian/GDevelop/blob/db93a25b331c207d2572744602a7034268e3571d/GDJS/Runtime/pixi-renderers/pixi-image-manager.js#L33

It would be better to see if it's possible to create an empty PIXI.Texture instead?

Fermain:
Hey, I did notice that it's used elsewhere.

An empty texture would definitely be better than a resource that may or may not be present in the project.

4ian:
I'll close this but any help to remove the default non existing texture resulting in a warning is welcome :)

PR #1223, opened 2019-09-08
improve and fix isEnded for Video object

Bouh:
- Fix boolean for isEnded
- Add an expression for isEnded

### DIFF
--- a/Extensions/Video/JsExtension.js
+++ b/Extensions/Video/JsExtension.js
@@ -378,18 +378,6 @@ module.exports = {
       .setFunctionName("isEnded");
 
     object
-      .addExpression(
-        "IsEnded",
-        _("Is ended"),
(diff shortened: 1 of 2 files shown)


4ian:
isEnded returns a boolean => this should not be an expression, expressions are number only :) (I know that JS will "tolerate" number and booleans, but this will lead to confusion and potential bugs down the road to leak this loose typing into GDevelop).
Could you remove it? :) There is already the condition, which should be the standard way of exposing booleans to the events sheet.

Bouh:
It's  useful for debugging, and compare expressions.
Nevermind, it's not really asked by users, i thought it was i good idea.
It's removed :)

4ian:
I do understand it's practical, but I would rather avoid this and add real support for booleans later :)
Thanks for the fix!

Issue #1228, opened 2019-09-15
New Layer Effects and freeze on Android webview


KinkGD:
## Describe the bug
The new Layer effects aren't working on android : 
I used at start a layer a effect (blur) and modified the blur parameter in a scene : the effect works perfectly smooth on desktop version of the game, but not android : it just makes the game freeze.

So i can report the bug occures in two scenarios on android :
1) Layer effect "blur) + effect modifiation through events : the scene freeze
2) Just the "blur" effect, without event modification of the the effect parameters
Both scenario => freeze. It seems the webGL effect just cannot be rendered  on android (webview/canva). 

## To Reproduce
Here a source file of the platformer example with scenario 1 events :
[LayerEffectTrouble.zip](https://github.com/4ian/GDevelop/files/3613584/LayerEffectTrouble.zip)
+ APK of scenario 1: [APK-scenario1.zip](https://github.com/4ian/GDevelop/files/3613586/APK-scenario1.zip)
+ APK of scenario 2: [APK-Scenario2.zip](https://github.com/4ian/GDevelop/files/3613587/APK-Scenario2.zip)

## Solutions ? 

Disable everything related to these effects on Android ? At least the possibility of disabling the effects on Mobile through event would be great for people like me who develop a single project for both desktop and mobile (right now, i have to remove the effect totally because of this android issue).

## Other details
GD5b80 - tested on android 6/7/8

4ian:
If the game freeze, might be either because of a crash related to the Pixi Filter - but I assumed that Pixi was just ignoring the filter when it's in canvas mode - or in the GD game engine because we assume a filter exists while it's not the case or something like this.

In any case, it should not freeze/crash the game. Would be great to see if you can have the webview console logs - it would show immediately the problem. Not sure exactly how to do it though, especially if the webview is not based on Chrome and is old.

Another option for debugging can be to force canvas rendering on desktop and see if it crash.

Bouh:
What I noticed is that the problem is only in the webview.
By testing with the wifi preview, we have the effect, but it's ultra slow.
The device seems to be poorly optimized for shaders.

4ian:
There might be a difference according to the browser. Android 4.x devices webviews are based on the slow and non webgl capable old internal browser, even if they have Chrome. If you're running a wifi preview with Chrome, it might be slow because the device won't have a GPU fast enough for rendering the shader, but at least Chrome will render with WebGL and not canvas.

KinkGD:
[...]

PR #1230, opened 2019-09-16
Add contition to check for WebGL on target device

Wend1go:
Might also help to find the issue in #1228


Extensions/SystemInfo/systeminfotools.js
```diff
@@ -28,3 +28,17 @@ gdjs.evtTools.systemInfo.isMobile = function() {
 
 	return false;
 };
+
+/**
+ * The device supports WebGL
```
4ian: Check if the the device supports WebGL.

Extensions/SystemInfo/systeminfotools.js
```diff
@@ -28,3 +28,17 @@ gdjs.evtTools.systemInfo.isMobile = function() {
...
+ * @param {gdjs.RuntimeScene} runtimeScene
+ * @returns {Boolean} WebGL is supported
```
4ian: Boolean (with a capital B) is said to be a "boxed primitive". Basically, never use it - always use the type "boolean" in JSdoc (more information here: http://adripofjavascript.com/blog/drips/the-difference-between-boolean-objects-and-boolean-primitives-in-javascript.html).

```suggestion
 * @returns {boolean} true if WebGL is supported
```

Extensions/SystemInfo/systeminfotools.js
```diff
@@ -28,3 +28,17 @@ gdjs.evtTools.systemInfo.isMobile = function() {
...
+ */
+gdjs.evtTools.systemInfo.isWebGLSupported = function(runtimeScene) {
+	var pixiRenderer = runtimeScene.getGame().getRenderer()._pixiRenderer;
```
4ian: It's a good start, but we can do without having to dig into a private property of the renderer.
For this, let's add a new method into the PixiRenderer (and CocosRenderer), that is called isWebGLSupported.
From here, you just call it:
```suggestion
	return runtimeScene.getGame().getRenderer().isWebGLSupported();
```

And then you implement the method in both the Pixi renderer and Cocos2d-JS.
Here for Pixi: https://github.com/4ian/GDevelop/blob/master/GDJS/Runtime/pixi-renderers/runtimegame-pixi-renderer.js#L428
Here for Cocos2d-JS: https://github.com/4ian/GDevelop/blob/master/GDJS/Runtime/cocos-renderers/runtimegame-cocos-renderer.js#L121

You can see that `getCanvas` is following the same pattern: abstract the details in the renderer, so that you have one implementation for Pixi:
https://github.com/4ian/GDevelop/blob/33487fb0f69ab0f16fa3c8c0eefac3919484e05d/GDJS/Runtime/pixi-renderers/runtimegame-pixi-renderer.js#L425-L427
and one for Cocos2d-JS:
https://github.com/4ian/GDevelop/blob/33487fb0f69ab0f16fa3c8c0eefac3919484e05d/GDJS/Runtime/cocos-renderers/runtimegame-cocos-renderer.js#L118-L120

By doing this, we keep the game engine and events relatively agnostic of whatever is the game using to render itself on screen :) When implementing a new renderer, it's also easier: copy/paste an existing renderer and implement everything that is needed in this file - rather than having to poke around in the codebase, searching for "pixiRenderer" and add a new if/else branch for your new renderer.

Last question: how to implement this for Cocos2D-JS?? :D 
No obvious answer in the doc, the Cocos2d-JS documentation is quite lacking information and it's often hard to find what is the proper way to do things :/
There is a documented constant called cc.game.RENDER_TYPE_WEBGL but we have to compare it to the current render type which is stored in a undocumented `cc._renderType`. Not a big fan of doing it, but by looking at the [source](https://docs.cocos2d-x.org/api-ref/js/V3.13/symbols/src/_Users_jtsm_Chukong-Inc_cocos2d-x_web_CCBoot.js.html), I can't see anything better.

So for cocos2d-JS, it's: `return cc._renderType === cc.game.RENDER_TYPE_WEBGL;`. 
We could also choose not to implement this, but as it seems relatively ok, let's use this. If it breaks later (let's say that _renderType is renamed), then the condition will return false - not too bad.

Extensions/TweenBehavior/tweenruntimebehavior.js
```diff
@@ -7,8 +7,8 @@
  */
 gdjs.TweenRuntimeBehavior = function (runtimeScene, behaviorData, owner) {
   gdjs.RuntimeBehavior.call(this, runtimeScene, behaviorData, owner);
-
   /** @type Object.<string, gdjs.TweenRuntimeBehavior.TweenInstance > */
+
```
4ian: See if you can revert this file to the state that is in master :)

4ian: If you struggle with this, let it alone and I'll revert the file when merging the PR.

Wend1go: I guess I'm indeed struggling. ^^

Wend1go:
I moved the logic into the renderers. 
Pixi is working fine. 
I couldn't test cocos export, looks like it doesn't run without a web server.

Wend1go:
Tested the cocos export as well and it is working too.

4ian:
Great! Will merge this after this weekend :)

PR #1185, opened 2019-08-06
Embed yarn editor in GD

blurymind:
This bundles Yarn - dialogue tree editor with gdevelop - making it possible to create and edit dialogue trees directly inside gdevelop.
yarnGD

It's worth noting that Yarn's devs gave me full access to do commits to it. If there is anything required to fix it- I can fix it :) 

This design bundles a new version of the electron branch of yarn, which I did some refactoring to in order to make it also work as a standalone web app. I am attaching it as a zip here, for those who want to test. Here is a build of
https://github.com/InfiniteAmmoInc/Yarn/tree/electron 

updated yarn package:
https://github.com/YarnSpinnerTool/YarnEditor/commit/a5fb76b771481c92b2ec4af47d4da460c4b56aa8
[yarn-editor.zip](https://github.com/4ian/GDevelop/files/3627667/yarn-editor.zip) (a5fb76b )

This pr also gives the path editor a new feature- a [>] button to toggle it, as well as css cleanup - to make the style easier to read and carry over to other editors.

Not sure if I should do the bbs parser next or ability to read mod music files :P 

updated the example:
[dialogueExample.zip](https://github.com/4ian/GDevelop/files/3513105/dialogueExample.zip)
Here is a screenshot example of how to set up an entire npc system with the extension:

One with animated avatars and conditional branches

trello
https://trello.com/c/vcRqzoGk/74-bundle-yarn-a-dialogue-editor-for-interactive-stories

PR #1236, opened 2019-09-25
Embed yarn editor

blurymind:
this is a continuation of 
https://github.com/4ian/GDevelop/pull/1185

the example demo game needed some cleaning up


blurymind:
Cleaned up the project file and made it clearer to press Z to progress dialogue :)

4ian:
This works well! Easier to find now what the controls are ;) No more warning/unused behaviors too, so looks good. I've tested with both the JSON resource preloaded or not, and it works in both case (the contrary would have been surprising, as if we handled the not preloaded case, the preloaded case should be working as well).

I'll merge this tomorrow morning and update the example to have it on the web-app.

Thanks a lot for working on this! This seems a great way to create dialogues - I can see this being super useful for tons of games!
Might be a good idea to work to expose these external editors somewhere in the resources editor interface - so that they are a bit less hidden :)

Keep up the good work 💪

blurymind:
Thank you @4ian
For patiently reviewing this evergrowing Pr and for helping get these
wonderful open source web apps together with GD.
I really hope to see more storytellers start using GD over renpy, rpg
maker, ags and other similar engines. If you look at games made with them,
alot of times it's by non programmers who are more invested in telling
stories. GD is a great match for that type of a dev imo.

I had a few ideas on how to make the editors more discoverable.. will do
some brainstorming on the issue tracker or discord  😊 I agree that the
resource manager is a good place to access them to create and edit
resources. Getting the music tracker would be huge too. After making the
dialogue tree extension I now know how to expose external libraries to the
game engine. Adding support for mod files doesn't seem so hard now.. there
is also the desire to add a rich text label object to compliment yarn.. so
many things that would are super fun to work on :p either way will do a few
small prs for a while before tackling another huge one

4ian:
Merged :) Yeah resources manager might be a good place - as we have now 3 external editors it's a good thing if we can make them more obvious to reach :)
Other nice features waiting to be done yeah! Thanks again for working on this!

4ian:
@blurymind There is still a "TODO…" in the http://wiki.compilgames.net/doku.php/gdevelop5/all-features/dialogue-tree help page - and the example was linking to nowhere. I added the example - could you take a look at the TODO?
I'm releasing a new version, better have finished help page, especially as people or press will have a look at it :)

blurymind:
@4ian sorry I forgot to remove that todo after adding the api commands to the wiki. Removed now :)

PR #1244, opened 2019-09-30
Add helpPagePath Video object

Bouh:
I have questions about the video object that I hope could be resolved if users use a help link that directs them to the wiki.

When I wrote the video extension the help path was added to the extension, but it seems that this path is not the same as the button for objects.

I saw that there is a button for the Skeleton object, I reproduced for the Video object but there are some things missing, it doesn't work.
I would need to be enlightened:)

### DIFF
--- a/Extensions/ExampleJsExtension/JsExtension.js
+++ b/Extensions/ExampleJsExtension/JsExtension.js
@@ -14,59 +14,59 @@ module.exports = {
   createExtension: function(_, gd) {
     const extension = new gd.PlatformExtension();
     extension.setExtensionInformation(
-      "MyDummyExtension",
-      "My Dummy Extension",
-      "An example of a declaration of an extension",
-      "Florian Rival",
-      "MIT"
+      'MyDummyExtension',
+      'My Dummy Extension',
+      'An example of a declaration of an extension',
+      'Florian Rival',
+      'MIT'
     );
 
     // Declare conditions, actions or expressions:
     extension
       .addCondition(
-        "MyNewCondition",
-        _("Dummy condition example"),
+        'MyNewCondition',
+        _('Dummy condition example'),
         _(
-          "This is an example of a condition displayed in the events sheet. Will return true if the number is less than 10 and the length of the text is less than 5."
+          'This is an example of a condition displayed in the events sheet. Will return true if the number is less than 10 and the length of the text is less than 5.'
         ),
-        _("Call the example condition with _PARAM0_ and _PARAM1_"),
-        _("Dummy Extension"),
-        "res/conditions/camera24.png",
-        "res/conditions/camera.png"
+        _('Call the example condition with _PARAM0_ and _PARAM1_'),
+        _('Dummy Extension'),
+        'res/conditions/camera24.png',
+        'res/conditions/camera.png'
       )
-      .addParameter("expression", _("Number 1"), "", false)
-      .addParameter("string", _("Text 1"), "", false)
+      .addParameter('expression', _('Number 1'), '', false)
+      .addParameter('string', _('Text 1'), '', false)
       .getCodeExtraInformation()
       .setIncludeFile(
-        "Extensions/ExampleJsExtension/examplejsextensiontools.js"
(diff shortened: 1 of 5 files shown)


Bouh:
I found by mimicry, it's seem working well :)

Bouh:
I found by mimicry, it's seem working well :)

4ian:
This works :) I'm thinking of an alternative where you can pass the helpPagePath as an argument: `getDefaultObjectJsImplementationPropertiesEditor({ helpPagePath:  "/test" })`.

This would require you to add an argument to getDefaultObjectJsImplementationPropertiesEditor and to change the 3/4 mentions to getDefaultObjectJsImplementationPropertiesEditor in the codebase: https://github.com/4ian/GDevelop/search?q=getDefaultObjectJsImplementationPropertiesEditor&unscoped_q=getDefaultObjectJsImplementationPropertiesEditor

That would be I think a bit better, because we would keep a simple method `getDefaultObjectJsImplementationPropertiesEditor` that we can easily search for, instead of having it existing but not used by any object.

4ian:
Basically:

```
  getDefaultObjectJsImplementationPropertiesEditor() {
    return {
      component: ObjectPropertiesEditor,
      createNewObject: object =>
        gd
          .asObjectJsImplementation(object)
          .clone()
          .release(),
      castToObjectType: object => gd.asObjectJsImplementation(object),
    };
  },
```

would be:

```
  getDefaultObjectJsImplementationPropertiesEditor(options) {
    return {
      component: ObjectPropertiesEditor,
      createNewObject: object =>
        gd
          .asObjectJsImplementation(object)
          .clone()
          .release(),
      castToObjectType: object => gd.asObjectJsImplementation(object),
      helpPagePath: options.helpPagePath,
    };
  },
```

Then make sure to replace all calls to `getDefaultObjectJsImplementationPropertiesEditor` in the codebase by passing options: `getDefaultObjectJsImplementationPropertiesEditor({helpPagePath: '/the/page'})`. Also update the mentions to it in the doc, and you're good! :)

Bouh:
Why not use the helpPagePath from here 
https://github.com/4ian/GDevelop/blob/d1c6347f3d8456ea752ee0ff15c02387f2bc45df/Extensions/Video/JsExtension.js#L24

And this is just for JS extension, the other extensions in cpp have `SetExtensionHelpPath()` too.
Why set this path again when we have already.

This seems link the declarations made in an extension to the object editor in GDevelop.
Why make a bridge between both files?
I mean in an extension we can say directly which editor we need.
That way you don't need this file  ObjectsEditorService.js.
But i understand the entire role of ObjectsEditorService.js maybe :)

4ian:
> Why not use the helpPagePath from here

It would be the best solution, but that means that in ObjectEditorDialog you have to fetch the object metadata using `gd.MetadataProvider.getObjectMetadata`. I would also have to add `GetHelpUrl` in ObjectMetadata in Bindings.idl for libGD.js. Then we could use it to get the help url.

> I mean in an extension we can say directly which editor we need.
> That way you don't need this file ObjectsEditorService.js.

Extensions don't know anything about the IDE Javascript files, that's why this file is in the newIDE sources. Also this is necessary for extensions written in C++: you can't set in a C++ extension a JavaScript object to be used for the editor :)
Though I agree it would be better if we could remove this "bridge" entirely.

Bouh:
> It would be the best solution, but that means that in ObjectEditorDialog you have to fetch the object metadata using gd.MetadataProvider.getObjectMetadata. I would also have to add GetHelpUrl in ObjectMetadata in Bindings.idl for libGD.js. Then we could use it to get the help url.

That adds code for little things, so let's start with the basic idea!
Existing file have been edited like you wish.
I've also added link to [wiki page ](http://wiki.compilgames.net/doku.php/gdevelop5/extensions/extend-gdevelop)about the built-in extension for the object in `ExampleJsExtension`

4ian:
Sounds good, I've re-run Prettier with single quotes for the JsExtension.js file (looks like I need to add a config file somewhere for these files), merging now!

4ian:
Thanks for adding this 👍

PR #1249, opened 2019-10-05
Use center point as flip center

4ian:
Flipping of sprites was previously done using the center of the image. This is fine in the case where the center point is the default one (i.e: at the center of the image, which is the default setting).
But if center is moved, then rotation is properly done around this point. However, flipping is still done centered around the image center. This create strange "jumps" when flipping the object, especially when it is rotated (see `GDJS/tests/games/rotate-flip-around-center-point/ship-rotate-flip.json` test game).

This PR fixes the behavior by having sprites flipped around the center point. This make sense because we can consider in 2D the flipping as a way to "rotate the object" around X and Y axes.

While this make more sense, this would be a breaking change for games having sprites that don't have a default center point and that are flipped. 
I was thinking of keeping backward compatibility with an option to activate to use the new behavior, but I'm now wondering if this use case is not small enough to just go ahead with the new behavior in the next version.

👋 **To help me, try this new testing version of GDevelop** and **let me know** if there is anything in your game that changes drastically:

* [Windows](https://gdevelop-releases.s3.amazonaws.com/fix/use-center-point-as-flip-center/latest/GDevelop+5+Setup+5.0.0-beta80.exe) (extract with 7zip to avoid replacing your existing installation of GD): 
* [Linux](https://gdevelop-releases.s3.amazonaws.com/fix/use-center-point-as-flip-center/latest/gdevelop-5.0.0-beta80.tar.gz)
* [macOS](https://gdevelop-releases.s3.amazonaws.com/fix/use-center-point-as-flip-center/latest/GDevelop+5-5.0.0-beta80-mac.zip)

**Make a backup of your game** before testing. If you see that some sprites position are changing (should be the case only in case of flipping with a non default center), check the center point of your sprites. Let me know if you have doubts :)

TODO:
- [ ] Add explanations about center and origin on http://wiki.compilgames.net/doku.php/gdevelop5/objects/sprite/edit-points?do=edit


Jose-Moreno:
While I understand what this PR does, what I don't understand is that currently all the sprite object rotation center usually come with an automatic rotation center placed roughly on SpriteObject.Width()/2 & Height()/2 right?

When you say this will effect other sprites that don't have a center you mean like Tiled Sprites and Panel Sprites? Or what exactly is the breaking change here? :thinking:

4ian:
This is affecting sprites that have the center point that is not the default one. You can modify the position of the center: 
image

If you do this, and flip the sprite, its rendered position on screen won't be valid in the current version.

zatsme:
Looks good to me, tested most of my games and demos 😄

Jose-Moreno:
Ah ok, so you're saying that if the center point is not set to `(auto)` it will not work in the newer version? Now I understand, thanks for the explanation! :smile:

4ian:
Any more feedback from people using this? 
I would like to merge this, hoping that this "breaking change" won't actually break too much games - if any?

PR #1254, opened 2019-10-12
Extract events into new group

Bouh:
qsd

Add the new button in submenu with right click, just under "Extract events to a function"
Move the events selected in a new group, this group is closed by default.

[Roadmap card](https://trello.com/c/JykQTO5L/364-move-selected-events-in-a-new-group-with-right-click)

----------

Thank for help @4ian , it's difficult to understand all relations between GDCore, GDJS, GDevelop.js.

[GDcore c++ Documentation](http://4ian.github.io/GD-Documentation/GDCore%20Documentation/namespacegd.html)

### DIFF
--- a/newIDE/app/src/EventsSheet/index.js
+++ b/newIDE/app/src/EventsSheet/index.js
@@ -757,7 +757,7 @@ export default class EventsSheet extends React.Component<Props, State> {
       eventsList.insertEvent(event, eventsList.getEventsCount())
     );
 
-    this._replaceSelectionByGroupsOfEvents(eventsList);
+    this._replaceSelectionByGroupOfEvents(eventsList);
     eventsList.delete();
   };
 
@@ -788,7 +788,7 @@ export default class EventsSheet extends React.Component<Props, State> {
     this.deleteSelection({ deleteInstructions: false });
   };
 
-  _replaceSelectionByGroupsOfEvents = (eventsList: gdEventsList) => {
+  _replaceSelectionByGroupOfEvents = (eventsList: gdEventsList) => {
     const contexts = getSelectedEventContexts(this.state.selection);
     if (!contexts.length) return;
 
--- a/newIDE/app/src/EventsSheet/index.js
+++ b/newIDE/app/src/EventsSheet/index.js
@@ -803,7 +803,7 @@ export default class EventsSheet extends React.Component<Props, State> {
 
     const groupEvent = gd.asGroupEvent(newEvents[0]);
 
-    groupEvent.setName('Grouped event');
+    groupEvent.setName('Grouped events');
     groupEvent.setFolded(true);
     groupEvent
       .getSubEvents()
--- a/newIDE/app/src/EventsSheet/index.js
+++ b/newIDE/app/src/EventsSheet/index.js
@@ -807,12 +807,7 @@ export default class EventsSheet extends React.Component<Props, State> {
     groupEvent.setFolded(true);
     groupEvent
       .getSubEvents()
-      .insertEvents(
-        eventsList,
-        0,
-        eventsList.getEventsCount(),
-        groupEvent.indexInList
-      );
+      .insertEvents(eventsList, 0, eventsList.getEventsCount(), 0);
 
     this.deleteSelection({ deleteInstructions: false });
   };
--- a/newIDE/app/src/EventsSheet/index.js
+++ b/newIDE/app/src/EventsSheet/index.js
@@ -805,7 +805,9 @@ export default class EventsSheet extends React.Component<Props, State> {
 
     groupEvent.setName('Grouped event');
     groupEvent.setFolded(true);
-    groupEvent.getSubEvents().insertEvents(
+    groupEvent
+      .getSubEvents()
+      .insertEvents(
         eventsList,
         0,
         eventsList.getEventsCount(),
--- a/newIDE/app/src/EventsSheet/index.js
+++ b/newIDE/app/src/EventsSheet/index.js
@@ -750,6 +750,17 @@ export default class EventsSheet extends React.Component<Props, State> {
     eventsList.delete();
   };
 
+  moveEventsIntoNewGroup = () => {
+    const eventsList = new gd.EventsList();
+
+    getSelectedEvents(this.state.selection).forEach(event =>
+      eventsList.insertEvent(event, eventsList.getEventsCount())
+    );
+
+    this._replaceSelectionByGroupsOfEvents(eventsList);
+    eventsList.delete();
+  };
+
   _replaceSelectionByEventsFunction = (
     extensionName: string,
     eventsFunction: gdEventsFunction
@@ -773,6 +784,33 @@ export default class EventsSheet extends React.Component<Props, State> {
     const standardEvt = gd.asStandardEvent(newEvents[0]);
     standardEvt.getActions().push_back(action);
     action.delete();
+
+    this.deleteSelection({ deleteInstructions: false });
+  };
+
+  _replaceSelectionByGroupsOfEvents = (eventsList: gdEventsList) => {
+    const contexts = getSelectedEventContexts(this.state.selection);
+    if (!contexts.length) return;
+
+    const newEvents = this.addNewEvent(
+      'BuiltinCommonInstructions::Group',
+      contexts[0]
+    );
+    if (!newEvents.length) {
+      console.error('A new event should have been created');
+      return;
+    }
+
+    const groupEvent = gd.asGroupEvent(newEvents[0]);
+
+    groupEvent.setName('Grouped event');
+    groupEvent.setFolded(true);
+    groupEvent.getSubEvents().insertEvents(
+        eventsList,
+        0,
+        eventsList.getEventsCount(),
+        groupEvent.indexInList
+      );
 
     this.deleteSelection({ deleteInstructions: false });
   };
@@ -1092,6 +1130,10 @@ export default class EventsSheet extends React.Component<Props, State> {
                             {
                               label: 'Extract Events to a Function',
                               click: () => this.extractEventsToFunction(),
+                            },
+                            {
+                              label: 'Move Events into a Group',
+                              click: () => this.moveEventsIntoNewGroup(),
                             },
                             {
                               label: 'Analyze Objects Used in this Event',


4ian:
This looks great! 

> Thank for help @4ian , it's difficult to understand all relations between GDCore, GDJS, GDevelop.js.

Let's try to make this easier, do you think I could add some explanations somewhere? In which case, where would it be easy to find/read? 
Basically, the idea is:
GDCore (C++): classes for what is a game, events and tools to make a game from it
GDJS: the special classes for generating JavaScript games (C++) and the game engine (JavaScript)
GDevelop.js: It's the C++ classes in GDCore and GDJS made available to JavaScript.

Maybe a note that you can read `Bindings.idl` file to see everything that you can use, and then read the documentation of GDCore where you'll find explanations/descriptions of methods?

newIDE/app/src/EventsSheet/index.js
```diff
@@ -777,6 +788,35 @@ export default class EventsSheet extends React.Component<Props, State> {
...
+
+    groupEvent.setName('Grouped events');
+    groupEvent.setFolded(true);
+    groupEvent
+      .getSubEvents()
+      .insertEvents(
+        eventsList,
+        0,
+        eventsList.getEventsCount(),
+        groupEvent.indexInList
```
4ian: groupEvent is a `gd.GroupEvent`: https://github.com/4ian/GDevelop/blob/master/GDevelop.js/Bindings/Bindings.idl#L1386-L1416

It does not have a indexInList property - which is something that is only on objects created to represent a selection. This will evaluate to undefined, which, luckily, will be converted to 0 in C++ ;)

insertEvents means: "insert this list called `eventsList`, from event number `0` to number `eventsList.getEventsCount()` (so the whole list) in `groupEvent` subevents, at the position indicated by the last parameter".

I'll modify this to be simply 0 ;) (whatever you indicate, the group event subevents are empty, so any number will work.

4ian:
That was almost perfect ;) Fixed a few typos and changed the insertion position (you were passing `undefined`, which was luckily converted to 0).

Bouh:
- `GDevelop/GDJS/Runtime/ `
 Is the content available in JSevent and used by JS extension in` GDevelop/Extensions/ `

- I understand what are `GDevelop/GDJS/GDJS/` and this content in sub folders, how it's used by GDevelop and where i can use it, or where GD use it.

- Same for `GDevelop/GDCpp/` yes it's for native games, so it's used by GD5 for export on Windows/Linux ? I see nothing about these OS, i see just Android, but android use JS in webview so idk why C++ is used here. 
These file come from GD4 ? I'm guessing that this `GDevelop/GDCpp/` is unused in GD5?


-  `GDevelop/Core/GDCore/Events`
One Event can be composed of multiple instructions and instruction can be an condition or action.
I've used these for move events in an other event, so `GDCore` is for manupulate events and structure of games project, but i see `Extensions`, `IDE`, `tinyXML`, `Tools` folders.

- I understand why multiples Extensions folder exist.
`GDevelop/Core/GDCore/Extensions` ?
`GDevelop/GDJS/GDJS/Extensions/` ?
`GDevelop/Extensions/ `This is for add content like objects, or instructions. I see it like mods files for GD.

-  `GDevelop/GDevelop.js/Bindings/Bindings.idl`
I wondering how is choosen what part of code from `Core` and `GDJS` are available somewhere else.
For example in JSevent we set string value on variable with `setString()` 
https://github.com/4ian/GDevelop/blob/master/GDevelop.js/Bindings/Bindings.idl#L160

   We cant use `SerializeTo()`, `renameChild()` and some others.
In any case, it is not visible in the auto-completion of a JSevent

   And why bind these method from C++ when `GDJS `have these method in pure JS here :
https://github.com/4ian/GDevelop/blob/master/GDJS/Runtime/variable.js#L169


- I don't understand the other files around `Bindings.idl`
`XXXImplementation.cpp`, maybe these file should be placed somewhere in `GDCore `?
Anyway i understand not at all these files. Since they're in this folder i guess it's used for make a bridge betweens C++/JS.

I see that the structure of the GD project is clean, but we only have short descriptions of the main folders.

arthuro555 have technical questions too on GD codebase.
https://forum.gdevelop-app.com/t/request-some-infos-about-the-code-of-gdevelop/20984

4ian:
> - `GDevelop/GDJS/Runtime/ `
 Is the content available in JSevent and used by JS extension in` GDevelop/Extensions/ `

Correct. Think of GDJS/Runtime as the game engine. JS events are running in your game (at "runtime"). JS Extension have "runtime" objects, behaviors, etc... that are all js files that are executed at "runtime" (runtime = during the game).

> - I understand what are `GDevelop/GDJS/GDJS/` and this content in sub folders, how it's used by GDevelop and where i can use it, or where GD use it.

While GDJS/Runtime is the game engine that is executed inside a game, `GDJS/GDJS` is the "editor" part, which are the C++ classes describing to the IDE various things (like, how [do you do an export](https://github.com/4ian/GDevelop/tree/master/GDJS/GDJS/IDE), what are [the default extensions](https://github.com/4ian/GDevelop/tree/master/GDJS/GDJS/Extensions/Builtin), how do you [generate JS code from events](https://github.com/4ian/GDevelop/tree/master/GDJS/GDJS/Events/CodeGeneration) 


> - Same for `GDevelop/GDCpp/` yes it's for native games, so it's used by GD5 for export on Windows/Linux ? 

It's not used by GD5. This is like GDJS, both the source code for the native game engine and the part exposed to the editor. The difference can be seen because the game engine is in `GDCpp/GDCpp/Runtime`, and the rest of the files are for the editor. 

It's technically dead code and most could be removed and GD5 would still work. I wanted to keep it to try to revive it but it would need quite a few adaptations to have the same extensions as GDJS and maybe find an alternative to SFML for rendering games.

>  These file come from GD4 ? I'm guessing that this `GDevelop/GDCpp/` is unused in GD5?

That's mostly correct yeah as I just explained.

>  I see nothing about these OS, i see just Android

There was an experimental export with the native game engine for Android in GD4. Never worked properly because SFML support for Android was experimental and unstable (at this time at least).

> One Event can be composed of multiple instructions and instruction can be an condition or action.

Yeah almost correct. An event is by default something that [is mostly empty](http://4ian.github.io/GD-Documentation/GDCore Documentation/classgd_1_1_base_event.html). Then in GDCore, some [default events are defined](https://github.com/4ian/GDevelop/tree/master/Core/GDCore/Events/Builtin). In particular, there is StandardEvent, which has conditions and actions. Conditions and actions are both a list of [gd::Instruction](https://4ian.github.io/GD-Documentation/GDCore%20Documentation/classgd_1_1_instruction.html). A gd::Instruction is either a condition or an action (it depends only on the context where they are used). A gd::Instruction is "just" a type (the name of the action or condition), and some parameters. You think of it as a function in a programming language (`add(2, 3)` is calling a function called "add" with parameter "2" and "3"). Conditions have the special thing that they are functions returning true or false.

(side note: a "program" is usually defined as a "set of instructions" - hence the name that was chosen to describe a condition or action).

GDJS (and GDCpp), in their default extensions ([see here for example](https://github.com/4ian/GDevelop/blob/master/GDJS/GDJS/Extensions/Builtin/CommonInstructionsExtension.cpp#L76-L96)) have functions to transform these events into real code.

> I've used these for move events in an other event, so `GDCore` is for manupulate events and structure of games project, but i see `Extensions`, `IDE`, `tinyXML`, `Tools` folders.

Yes, GDCore is basically = everything to manipulate the structure of a game. This includes events, scenes, objects, behaviors, etc...
`Extensions` folder contains default extensions. TinyXML, don't care it's for XML writing - now unused. `IDE` is for c++ classes allowing to do operations on the structure of a game. For example, [WholeProjectRefactorer](https://github.com/4ian/GDevelop/blob/master/Core/GDCore/IDE/WholeProjectRefactorer.cpp) is very powerful, it is used to rename all objects in a game or update events after an object is deleted, etc.. The directory contains other "tool" functions to manipulate [resources of projects](https://github.com/4ian/GDevelop/tree/master/Core/GDCore/IDE/Project) or do [search in events](https://github.com/4ian/GDevelop/tree/master/Core/GDCore/IDE/Events)

> - I understand why multiples Extensions folder exist.
`GDevelop/Core/GDCore/Extensions` ?
`GDevelop/GDJS/GDJS/Extensions/` ?
`GDevelop/Extensions/ `This is for add content like objects, or instructions. I see it like mods files for GD.

* `GDevelop/Core/GDCore/Extensions` is the **declaration** of default extensions, available for any platform (native, JS, etc...) and any game.
* `GDevelop/GDJS/GDJS/Extensions/` and `GDevelop/GDCpp/GDCpp/Extensions/` are reusing these **declarations** and **adding** their own declaration. Mainly, they are setting the name of the functions to be called (either in JS or in C++) for each action, condition or expression.
* `GDevelop/Extensions/` is indeed like "mods", "plugins" for GDevelop. I chose to call them "Extensions". The difference is that they are not part of GDCore - they work on their own.

In theory, all extensions could be moved to `GDevelop/Extensions/`. In practice, it's good to have a "core" of default extensions. It's like a "standard library" in a programming language.

> -  `GDevelop/GDevelop.js/Bindings/Bindings.idl`
I wondering how is choosen what part of code from `Core` and `GDJS` are available somewhere else.

The goal of GDevelop.js is to make available all the code written in C++ to JavaScript, so that we can write an editor in JavaScript and have it working in a web browser.
* We're using Emscripten which is compiling C++, but instead of writing a "native binary", it's writing a file that works in a browser (basically, JavaScript!).
* All the required C++ files are imported into this huge list: https://github.com/4ian/GDevelop/blob/master/GDevelop.js/Bindings/Wrapper.cpp#L1-L79. C++ compilation is done with a "build system" called CMake, which is using [this file](https://github.com/4ian/GDevelop/blob/master/GDevelop.js/CMakeLists.txt#L82-L101) to see what to compile.
* Finally, `Bindings.idl` is used by the compiler (Emscripten) to generate a bridge between fake JS classes (`gd.Project`, `gd.Object`, `gd.StandardEvent`) and the C++ compiled. `libGD.js` is the sum of this C++ compiled code and the "fake JS classes" generated from Bindings.idl.

> For example in JSevent we set string value on variable with `setString()` 
> https://github.com/4ian/GDevelop/blob/master/GDevelop.js/Bindings/Bindings.idl#L160

Here you have two different things. 
* The **editor** `gd::Variable` that is part of the structure of the game, so living in [GDCore in Variable.h](https://4ian.github.io/GD-Documentation/GDCore%20Documentation/classgd_1_1_variable.html)
* The **game engine** variable, called `gdjs.Variable` (this is a *bad decision from mine*. I should have called it `gdjs.RuntimeVariable`, like `gdjs.RuntimeObject` and like every class of the game engine). [Documentation is in the GDJS **game engine**](http://4ian.github.io/GD-Documentation/GDJS%20Runtime%20Documentation/gdjs.Variable.html).

In a JavaScript event code block, you're in the game, so the variables you're dealing with are `gdjs.Variable`.

>   We cant use `SerializeTo()`, `renameChild()` and some others.

Exact, these things are part of GDCore: `gd::Variable`.

> - I don't understand the other files around `Bindings.idl`
`XXXImplementation.cpp`, maybe these file should be placed somewhere in `GDCore `?
Anyway i understand not at all these files. Since they're in this folder i guess it's used for make a bridge betweens C++/JS.

Correct, they are used to make a bridge between the editor, writtin in JavaScript, and C++.
These things are mostly "ugly" things doing weird stuff to translate from JS to C++ or vice versa. It requires a bit of knowledge about how the ["bridge", made by Emscripten, called WebIDL](https://emscripten.org/docs/porting/connecting_cpp_and_javascript/WebIDL-Binder.html), is working. 
You don't need to work with these unless you want to expose something that is written in C++ to the editor, and writing the interface in Bindings.idl is not sufficient.
In 90% of the time, you can just write the class in Bindings.idl, add the header file in Wrapper.cpp and your C++ class is "automagically" compiled and available in JavaScript.

In an idea world, there would be something to do this automatically, so GDevelop.js folder would not even exist :)

> I see that the structure of the GD project is clean, but we only have short descriptions of the main folders.

I'll try to add some comments or `README.md` files in folders. Please feel free to suggest some `README.md` files (you can even send a pull request, I'm happy to complete/correct whatever you write).

4ian:
I tried to write a useful doc about the architecture: https://github.com/4ian/GDevelop/blob/master/Core/GDevelop-Architecture-Overview.md

Let me know if you have more questions and I'll try to fill the missing pieces :)

Bouh:
Thank you very much for these explanations! The roles of each part of the code are now much clearer :)

PR #1256, opened 2019-10-14
Add new shapes on Shape Painter

Bouh:
This PR will add more shapes like Ellipse, RoundedRect, Star, arc, arcTo.
I start with Ellipse.

Before starting, this extension is a mix between C++ and JS, and because extension declaration are in C++ i need compile GDevelop.js
It's still not possible to do it from my PC, I still have many errors. So I can't test it.

I need your review on actual code. Something are missing in a file for my ellipse ?
Then I would add the other shapes in the same way.

If it's needed i can convert this extension in pure JS, after that no longer need to launch the compilation of GDevelop.js.

I work a blindly without being able to test it, but if you tell me it will work, it will avoid to me to rewriting the extension in JS.


Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -79,6 +79,23 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+                _("Draw at _PARAM1_;_PARAM2_ a ellipse of width _PARAM3_ and height _PARAM4_ "
+                "with _PARAM0_"),
+                _("Drawing"),
+                "res/actions/line24.png",
+                "res/actions/line.png")
+
+      .AddParameter("object", _("Shape Painter object"), "Drawer")
+      .AddParameter("expression", _("X Position of start point"))
+      .AddParameter("expression", _("Y Position of start point"))
+      .AddParameter("expression", _("Width of ellipse"))
```
4ian: According to Pixi docs, this is "The half width of the ellipse"

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -79,6 +79,23 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+      .AddParameter("expression", _("Height of ellipse"))
```
4ian: "The half height of the ellipse"

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -79,6 +79,23 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+                _("Ellipse"),
+                _("Draw a ellipse on screen"),
+                _("Draw at _PARAM1_;_PARAM2_ a ellipse of width _PARAM3_ and height _PARAM4_ "
+                "with _PARAM0_"),
+                _("Drawing"),
+                "res/actions/line24.png",
+                "res/actions/line.png")
+
+      .AddParameter("object", _("Shape Painter object"), "Drawer")
+      .AddParameter("expression", _("X Position of start point"))
```
4ian: "X coordinate of the center of the ellipse"

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -79,6 +79,23 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+      .AddParameter("expression", _("Y Position of start point"))
```
4ian: "Y coordinate of the center of the ellipse"

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -79,6 +79,23 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
       .SetFunctionName("DrawLine")
       .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
 
+  obj.AddAction("Ellipse",
+                _("Ellipse"),
+                _("Draw a ellipse on screen"),
```
4ian: Draw an ellipse

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -79,6 +79,23 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+                _("Draw at _PARAM1_;_PARAM2_ a ellipse of width _PARAM3_ and height _PARAM4_ "
```
4ian: "an ellipse"
"half width"
"half height"

Extensions/PrimitiveDrawing/shapepainterruntimeobject.js
```diff
@@ -60,6 +60,10 @@ gdjs.ShapePainterRuntimeObject.prototype.drawLine = function(x1, y1, x2, y2, thi
     this._renderer.drawLine(x1, y1, x2, y2, thickness);
 };
 
+gdjs.ShapePainterRuntimeObject.prototype.drawEllipse = function(x1, y1, width, height) {
```
4ian: as x1 and y1 are the only points, and are the center coordinates, let's name them "centerX", "centerY".

4ian:
> It's still not possible to do it from my PC, I still have many errors. So I can't test it.

I got a computer with Windows 10 so I should be able to update the build and instructions to get this working on Windows soon.

Note that you can see that all the checks are passing, including Travis which is doing a recompilation of GDevelop.js for each of your commit. So this seems fine:

image

If there is an error at some point, open the "Details", then click on the GCC or Clang build, and look what is the last errors reported in these Travis logs. (currently, there is no issue: https://travis-ci.org/4ian/GDevelop/jobs/597894698)

Looks good according to what I see in the code! 👍 Added a few comments about variable/parameters naming.

Bouh:
Yes Travis shows me errors during compilation, I know. Can we get the files compiled by Travis?

If these can help here all my log files and a [video about how i run the build.](https://www.youtube.com/watch?v=3vPJcuo5EyY) 
At the beginning i start with cmake-GUI, and after with console.

Command executed in cmd as admin.
https://pastebin.com/raw/UC5ZGWh8

Log from npm
https://pastebin.com/raw/cijMagQ0

CMakeError.log
https://pastebin.com/raw/ygxXYXh6

MinGw install with 
https://sourceforge.net/projects/mingw/

Cmake 3.5 install with 
https://cmake.org/files/v3.5/
cmake-3.5.0-win32-x86.msi

4ian:
> Yes Travis shows me errors during compilation, I know. Can we get the files compiled by Travis?

No there are no errors :) There is a green check, your code is correct:
image

> If these can help here all my log files and a video about how i run the build.

Cool, thanks for the video and the logs, very useful.
The build commands are outdated and I'll update them tonight (emconfigure should be used on WIndows too - while MinGW is needed for `mingw32-make`, `gcc` should NOT be called. Instead, `emcc`, the emscripten compiler should. This is because we're missing the call to emconfigure like done on mac/Linux).

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -99,20 +99,40 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+                "res/actions/star24.png",
+                "res/actions/star.png")
+
+      .AddParameter("object", _("Shape Painter object"), "Drawer")
+      .AddParameter("expression", _("X coordinate of the center of the star"))
+      .AddParameter("expression", _("Y coordinate of the center of the star"))
+      .AddParameter("expression", _("Points of the star"))
+      .AddParameter("expression", _("The radius (in pixels)"))
+      .AddParameter("expression", _("The inner radius (in pixels, half radius by default)"))
+      .SetDefaultValue("1")
```
Bouh: I don't know here, because the[ documentation say ](https://pixijs.download/v4.8.6/docs/PIXI.Graphics.html#drawStar)inner radius is by default half of radius.
In pixi renderer i have added ternary conditions, but i don't know how manage this, in extension.cpp or in pixi renderer.

4ian: Default value is not supported very well. I would suggest to:
- Write "(in pixels, recommended default value: the radius divided by 2)" so that users know what they should use
- in Pixi renderer, if the value is 0 (which will be what GDevelop will be passing if the expression is not fulfilled), then use the radius/2 instead.

for the 2nd point, I'm following the hypothesis that 0 is not really a valid value (and is what GDevelop is passing for empty expression), so we can detect it.

4ian:
Managed to compile on Windows :)

Screenshot 2019-10-15 23 16 06

There are changes that I need to do in the build file, will do it tomorrow

Bouh:
📌 **Note:** [This commit that fix the compilation for Windows 10
](https://github.com/4ian/GDevelop/commit/5b5f213bd551616aa3c6f95827d799e5683e1647#commitcomment-35520454) 

-  `arcTo` is a bit weird to use, maybe i will teak the parameters of this function, because it's depending of previous graphics or we need use `moveTo()` but we can't use same coord x;y for `moveTo()` and set same x;y in x1;y1 for `arcTo()`...

   The documentation ([link](http://www.dbp-consulting.com/tutorials/canvas/CanvasArcTo.html), [link2](https://www.scriptol.fr/html5/canvas/arcto.php)) is very clear but i don't see how user can need this features Oo.
   [Jsfiffle sandbox](https://jsfiddle.net/Babar31/y9051pu8/42/)

- `star`, `roundedRectangle` and `ellipse` working very well.

Todo : 
- [ ]  `arc` need an option for close path, i need use parameters
- [ ] Take a look on `bezierCurveTo` and `quadraticCurveTo` curve are better, control is more implicit/friendly than `arcTo()`
- [ ] Think how to implement `polygon`.
- [ ] Think how to implement custom shape, chaining functions like `moveTo()`, `lineTo()`, `arcTo()`, etc...

4ian:
Great progress :) 
You might need to moveTo before arcTo, though not sure why you could not use the same coords moveTo and arcTo?
In case of doubt, feel free to move to other shapes - we can always add arcTo later or in another PR :)

Let me know how it goes!

4ian:
Also I would like to make a new release more or less soon. Let's try to get a bunch of changes that are working and add other things like arcTo if another PR?

Bouh:
Yes you can merge this PR now.
I will work on next shapes on an new PR

4ian:
Looks great! Made a few fixes and merged 👍

PR #1258, opened 2019-10-19
More advanced shapes

Bouh:
More shape : 

- LineV2 (because actual drawLine() is wrong, we have a difference of x2 between drawLine() and lineV2() when line() use drawPolygon)
- Arc
- BezierCurve
- QuadraticCurve

Use event like we can chainning method in pixijs:
- beginFillPath
- endFillPath
- PathMoveTo
- PathLineTo
- PathBezierCurveTo
- PathArc
- PathQuadraticCurveTo
- closePath

ArcTo, PathArcTo are wreid i don't know how simplify these action for user.
I working on the icons on my side.

image


Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -116,47 +134,47 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
       .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
 
-  // These actions are not exposed yet as the way they work is unsure. See https://github.com/4ian/GDevelop/pull/1256
-/*
+
   obj.AddAction("Arc",
                 _("Arc"),
                 _("Draw an arc on screen"),
-                _("Draw at _PARAM1_;_PARAM2_ an arc with _PARAM3_ of radius, start angle: _PARAM4_, end angle: _PARAM5_ (anticlockwise: _PARAM6_) "
+                _("Draw at _PARAM1_;_PARAM2_ an arc with _PARAM3_ of radius, start angle: _PARAM4_, end angle: _PARAM5_ (anticlockwise: _PARAM6_, close path:_PARAM7_) "
```
4ian: Space after :

4ian:
Looks awesome 🔥

> LineV2 (because actual drawLine() is wrong, we have a difference of x2 between drawLine() and lineV2() when line() use drawPolygon)

Yeah your implementation is much better. I think we can hide the old version actually to "deprecate" it and encourage people to use your new, better and simpler, version.

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -168,9 +187,194 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
       .AddParameter("expression", _("Y coordinate end arc"))
       .AddParameter("expression", _("Radius (in pixels)"))
+      .AddParameter("yesorno", _("Close path"))
       .SetFunctionName("DrawArcTo")
       .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
 */
+  obj.AddAction("BezierCurve",
+                _("Bezier curve"),
+                _("Draw a bezier curve on screen"),
+                _("Draw from _PARAM1_;_PARAM2_ to _PARAM7_;_PARAM5_ a bezier curve (first control point: _PARAM3_;_PARAM4_, second control point: _PARAM5_;_PARAM6_) "
```
4ian: PARAM5 should be PARAM8 I guess?

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -116,47 +135,47 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
       .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
 
-  // These actions are not exposed yet as the way they work is unsure. See https://github.com/4ian/GDevelop/pull/1256
-/*
+
   obj.AddAction("Arc",
                 _("Arc"),
                 _("Draw an arc on screen"),
-                _("Draw at _PARAM1_;_PARAM2_ an arc with _PARAM3_ of radius, start angle: _PARAM4_, end angle: _PARAM5_ (anticlockwise: _PARAM6_) "
+                _("Draw at _PARAM1_;_PARAM2_ an arc with _PARAM3_ of radius, start angle: _PARAM4_, end angle: _PARAM5_ (anticlockwise: _PARAM6_, close path: _PARAM7_) "
```
4ian: I think "x of radius" is not proper english, let's use instead "[...] an arc with radius: _PARAM3_, start angle: [...]"

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -57,42 +57,61 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
       .AddParameter("expression", _("Y position of start point"))
       .AddParameter("expression", _("X position of end point"))
       .AddParameter("expression", _("Y position of end point"))
-      .AddParameter("expression", _("Thickness ( in pixels )"))
+      .AddParameter("expression", _("Thickness (in pixels)"))
       .SetFunctionName("DrawLine")
       .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
+  
+   obj.AddAction("LineV2",
+                _("Line V2"),
```
4ian: On screen, we can display "Line" to the user, as this is now the recommended "Line" action :)

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -57,42 +57,61 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+                _("Draw a line V2 on screen"),
```
4ian: Same, you can remove V2 from descriptions.

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -57,42 +57,61 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+      .AddParameter("expression", _("Y position of end point"))
+      .AddParameter("expression", _("Thickness (in pixels)"))
+      .SetFunctionName("drawLineV2")
+      .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
 
   obj.AddAction("Ellipse",
                 _("Ellipse"),
                 _("Draw an ellipse on screen"),
-                _("Draw at _PARAM1_;_PARAM2_ an ellipse of half width _PARAM3_ and half height _PARAM4_ "
+                _("Draw at _PARAM1_;_PARAM2_ an ellipse of width _PARAM3_ and height _PARAM4_ "
```
4ian: There is a useless extra space at the end of this sentence and some others too.

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -116,47 +135,47 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
-      .AddParameter("expression", _("Y coordinate of the center of the arc"))
+      .AddParameter("expression", _("X position of center"))
+      .AddParameter("expression", _("Y position of center"))
       .AddParameter("expression", _("Radius (in pixels)"))
-      .AddParameter("expression", _("Start angle of the arc"))
-      .AddParameter("expression", _("End angle of the arc"))
+      .AddParameter("expression", _("Start angle of the arc (in degrees)"))
+      .AddParameter("expression", _("End angle of the arc (in degrees)"))
       .AddParameter("yesorno", _("Anticlockwise"))
+      .AddParameter("yesorno", _("Close path"))
```
4ian: I wonder if we should be a bit more explicit: "Close the path with a line between start and end"?

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -168,9 +187,194 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+      .AddParameter("expression", _("Control point x"))
+      .AddParameter("expression", _("Control point y"))
+      .AddParameter("expression", _("Destination point x"))
+      .AddParameter("expression", _("Destination point y"))
+      .SetFunctionName("drawQuadraticCurve")
+      .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
+
+  obj.AddAction("BeginFillPath",
+                _("Begin fill path"),
+                _("Begins drawing filling of an advanced path"),
```
4ian: Begin to draw a simple one-color fill. Subsequent actions, such as "Path line" (in the Advanced category) can be used to draw. Be sure to use "End fill path" action when you're done drawing the shape.

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -168,9 +187,194 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+                "res/actions/delete24.png",
+                "res/actions/delete.png")
+
+      .AddParameter("object", _("Shape Painter object"), "Drawer")
+      .SetFunctionName("endFillPath")
+      .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
+
+  obj.AddAction("MovePathTo",
+                _("Move path to"),
+                _("Move the origin of a path."),
```
4ian: "Move the drawing position for the current path"?

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -168,9 +187,194 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+                _("Move the origin of a path to _PARAM1_;_PARAM2_ "
```
4ian: "Move the drawing position of the path to [xxx]"?

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -168,9 +187,194 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+                _("Advanced"),
+                "res/actions/delete24.png",
+                "res/actions/delete.png")
+
+      .AddParameter("object", _("Shape Painter object"), "Drawer")
+      .SetFunctionName("endFillPath")
+      .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
+
+  obj.AddAction("MovePathTo",
+                _("Move path to"),
```
4ian: "Move path drawing position"?

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -168,9 +187,194 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+                "res/actions/position.png")
+
+      .AddParameter("object", _("Shape Painter object"), "Drawer")
+      .AddParameter("expression", _("X position of start point"))
+      .AddParameter("expression", _("Y position of start point"))
+      .SetFunctionName("drawPathMoveTo")
+      .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
+
+  obj.AddAction("PathLineTo",
+                _("Path line to"),
```
4ian: I would remove from the name shown to the user the "to": "Path line to" => "Path line"

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -168,9 +187,194 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+                _("Add to a path a line to a position, the origin come from the previous action or from BeginFillPath or PathMoveTo, by default if nothing exist, start position will be the object's position."),
```
4ian: Better use shorter sentences. Also, use the full, "user friendly" name of the actions (i.e: the one shown in the UI) in the descriptions:

`Add to a path a line to a position. The origin comes from the previous action or from "Begin fill path" or "Move path drawing position". By default if nothing exist, start position will be the object's position.`

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -168,9 +187,194 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+
+      .AddParameter("object", _("Shape Painter object"), "Drawer")
+      .AddParameter("expression", _("X position of start point"))
+      .AddParameter("expression", _("Y position of start point"))
+      .SetFunctionName("drawPathLineTo")
+      .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
+
+  obj.AddAction("PathBezierCurveTo",
+                _("Path bezier curve to"),
+                _("Add to a path a bezier curve to a position, the origin come from the previous action or from BeginFillPath or PathMoveTo, by default if nothing exist, start position will be the object's position."),
```
4ian: Would be great if you can update this and later descriptions as per my comment below :)

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -168,9 +187,194 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+      .AddParameter("expression", _("Y coordinate end arc"))
+      .AddParameter("expression", _("Radius (in pixels)"))
+      .SetFunctionName("drawPathArcTo")
+      .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
+*/
+
+  obj.AddAction("PathArc",
+                _("Path arc to"),
+                _("Add to a path an arc to a position, the origin come from the previous action or from BeginFillPath or PathMoveTo, by default if nothing exist, start position will be the object's position."),
+                _("Add to a path an arc at the position _PARAM1_;_PARAM2_ (radius: _PARAM3_,start angle: _PARAM4_, end angles: _PARAM5_,anticlockwise: _PARAM6_ "
```
4ian: space before "start angle"
space before anticlockwise
no s to "end angles"
no space at the end and close the parenthesis

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -168,9 +187,194 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+      .AddParameter("expression", _("X position of start point"))
+      .AddParameter("expression", _("Y position of start point"))
+      .SetFunctionName("drawPathMoveTo")
+      .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
+
+  obj.AddAction("PathLineTo",
+                _("Path line to"),
+                _("Add to a path a line to a position, the origin come from the previous action or from BeginFillPath or PathMoveTo, by default if nothing exist, start position will be the object's position."),
+                _("Add to a path a line to the position _PARAM1_;_PARAM2_ "
+                  " with _PARAM0_"),
```
4ian: you have an extra space here and in some other actions when writing `" with _PARAM0_"`

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -168,9 +187,194 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+      .AddParameter("expression", _("X position of start point"))
+      .AddParameter("expression", _("Y position of start point"))
+      .SetFunctionName("drawPathLineTo")
+      .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
+
+  obj.AddAction("PathBezierCurveTo",
+                _("Path bezier curve to"),
+                _("Add to a path a bezier curve to a position, the origin come from the previous action or from BeginFillPath or PathMoveTo, by default if nothing exist, start position will be the object's position."),
+                _("Add to a path a bezier curve to the position _PARAM5_;_PARAM6_ (first control point: _PARAM1_;_PARAM2_, second control point: _PARAM3_;_PARAM4_) "
+                  " with _PARAM0_"),
```
4ian: extra space

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -168,9 +187,194 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+      .AddParameter("expression", _("Control point x"))
+      .AddParameter("expression", _("Control point y"))
+      .AddParameter("expression", _("Destination point x"))
+      .AddParameter("expression", _("Destination point y"))
+      .SetFunctionName("drawPathQuadraticCurveTo")
+      .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
+
+  obj.AddAction("closePath",
+                _("Close Path"),
+                _("Close the path of advanced shape. This close the outline between the last point and first."),
```
4ian: "an advanced shape". This is the first time we're mentioning shape. I wonder if we should mention in other actions a "shape" rather than a "fill path" (which seems weird?)

Extensions/PrimitiveDrawing/JsExtension.cpp
```diff
@@ -42,6 +42,63 @@ class PrimitiveDrawingJsExtension : public gd::PlatformExtension {
...
+        "PrimitiveDrawing::Drawer")["PrimitiveDrawing::PathMoveTo"]
+        .SetFunctionName("drawPathMoveTo");
+    GetAllActionsForObject(
+        "PrimitiveDrawing::Drawer")["PrimitiveDrawing::PathLineTo"]
+        .SetFunctionName("drawPathLineTo");
+    GetAllActionsForObject(
+        "PrimitiveDrawing::Drawer")["PrimitiveDrawing::PathBezierCurveTo"]
+        .SetFunctionName("drawPathBezierCurveTo");
+    GetAllActionsForObject(
+        "PrimitiveDrawing::Drawer")["PrimitiveDrawing::PathArc"]
```
4ian: So this name "PathArc" seems inconsistent with the others ("PathLineTo"). But there is already the commented "PathArcTo". So I think we should remove "To" from others actions (PathLineTo => PathLine).

Extensions/PrimitiveDrawing/shapepainterruntimeobject-pixi-renderer.js
```diff
@@ -47,9 +47,16 @@ gdjs.ShapePainterRuntimeObjectPixiRenderer.prototype.drawLine = function(x1, y1,
...
+    this._graphics.lineStyle(thickness);
+    this._graphics.moveTo(x1, y1);
+    this._graphics.lineTo(x2,y2);
+    this._graphics.endFill();  
+};
+
 gdjs.ShapePainterRuntimeObjectPixiRenderer.prototype.drawEllipse = function(x1, y1, width, height) {
     this._graphics.beginFill(this._object._fillColor, this._object._fillOpacity / 255);
-    this._graphics.drawEllipse(x1, y1, width, height);
+    this._graphics.drawEllipse(x1, y1, width/2, height/2);
```
4ian: It would be interesting to move the division by 2 in the renderer. Because otherwise it's a bit surprising to have the renderer and the object both having a "drawEllipse" method, but that accept different parameters.

Extensions/PrimitiveDrawing/shapepainterruntimeobject-pixi-renderer.js
```diff
@@ -67,24 +74,107 @@ gdjs.ShapePainterRuntimeObjectPixiRenderer.prototype.drawStar = function(x1, y1,
...
+gdjs.ShapePainterRuntimeObjectPixiRenderer.prototype.beginFillPath = function() {
+    this._graphics.beginFill(this._object._fillColor, this._object._fillOpacity / 255);
+    /*if(!this.hasBeginFillPath){
+        this.beginFillPath();
+        this._graphics.haveBeginFillPath = true;
+    }*/
+};
+
+gdjs.ShapePainterRuntimeObjectPixiRenderer.prototype.endFillPath = function() {
+   /* if(!this.hasBeginFillPath){
```
4ian: The commented code is not needed? If we call this._graphics.endFill without having starting a path, there is no crash or weird result?

Extensions/PrimitiveDrawing/shapepainterruntimeobject-pixi-renderer.js
```diff
@@ -67,24 +74,107 @@ gdjs.ShapePainterRuntimeObjectPixiRenderer.prototype.drawStar = function(x1, y1,
...
+        this._graphics.moveTo(0, 0);
+    }
+    this._graphics.quadraticCurveTo(cpX, cpY, toX, toY);  
+};
+
+gdjs.ShapePainterRuntimeObjectPixiRenderer.prototype.closePath = function() {
+    this._graphics.closePath();  
+};
+
+gdjs.ShapePainterRuntimeObjectPixiRenderer.prototype.hasBeginFillPath = function() {
```
4ian: `hasBegunFillPath`

Extensions/PrimitiveDrawing/shapepainterruntimeobject-pixi-renderer.js
```diff
@@ -67,24 +74,107 @@ gdjs.ShapePainterRuntimeObjectPixiRenderer.prototype.drawStar = function(x1, y1,
...
+    if(this._graphics.haveBeginFillPath){
```
4ian: `haveBeginFillPath` should be renamed to `hasBegunFillPath`

4ian: Also, you can just `return !!this._graphics.hasBegiuFillPath;` (the `!!` to make sure that if it's `undefined`, it will be "converted" to false).

Extensions/PrimitiveDrawing/shapepainterruntimeobject.js
```diff
@@ -71,15 +75,60 @@ gdjs.ShapePainterRuntimeObject.prototype.drawRoundedRectangle = function(startX1
...
+    this._renderer.drawBezierCurve(x1, y1, cpX, cpY, cpX2, cpY2, x2, y2);  
+};
+
+gdjs.ShapePainterRuntimeObject.prototype.drawQuadraticCurve = function(x1, y1, cpX, cpY, x2, y2) {
+    this._renderer.drawQuadraticCurve(x1, y1, cpX, cpY, x2, y2);  
+};
+
+gdjs.ShapePainterRuntimeObject.prototype.beginFillPath = function(x1, y1) {
+    this._renderer.beginFillPath();
+    this._renderer.drawPathMoveTo(x1?x1:0, y1?y1:0);
```
4ian: Is `x1?x1:0` necessary? x1 will always be a number I think?

4ian: (and same for x2)

Extensions/PrimitiveDrawing/JsExtension.cpp
4ian: Well actually this inconsistency is in Pixi.js too, so let's keep it.

4ian:
Added comments mainly about descriptions. This is some great job here! Will be super useful to generate advanced shapes and effects in games!

Extensions/PrimitiveDrawing/Extension.cpp
Bouh:
```
_("Draw at _PARAM1_;_PARAM2_ an ellipse of width _PARAM3_ and height _PARAM4_ "
 "with _PARAM0_"),
```
?
Because other extension was like this before, on two lines.

4ian: Ah my bad this one is fine!

Extensions/PrimitiveDrawing/Extension.cpp
Bouh: In the description yes.

Extensions/PrimitiveDrawing/Extension.cpp
Bouh: i prefer talk about "shape".
"Path" is used by pixiJS but instead "path" we can talk "outlines"
and keep "fill" for filling the shape.

Extensions/PrimitiveDrawing/shapepainterruntimeobject-pixi-renderer.js
Bouh: hasBeginFillPath / hasBegunFillPath
Is not used anymore i delete this useless function and all comments associate.

Extensions/PrimitiveDrawing/Extension.cpp
Wend1go: `This closes the outline between the last and the first point.`

4ian: Wend1go comment sounds good.

Extensions/PrimitiveDrawing/Extension.cpp
4ian: Ok, in this case let's go for:

`Draw an arc on screen. If "Close path" is set to yes, a line will be drawn between the start and end point of the arc, closing the shape.`

Extensions/PrimitiveDrawing/Extension.cpp
4ian: And agree about using `shape`, `outline` and keeping `fill`.

Bouh:
Thank you for the help!

Extensions/PrimitiveDrawing/shapepainterruntimeobject-pixi-renderer.js
```diff
@@ -67,24 +74,99 @@ gdjs.ShapePainterRuntimeObjectPixiRenderer.prototype.drawStar = function(x1, y1,
...
+    }
+    this._graphics.quadraticCurveTo(cpX, cpY, toX, toY);  
+};
+
+gdjs.ShapePainterRuntimeObjectPixiRenderer.prototype.closePath = function() {
+    this._graphics.closePath();  
+};
+
+/*
+gdjs.ShapePainterRuntimeObjectPixiRenderer.prototype.hasBegunFillPath = function() {
```
4ian: If this is not necessary, let's remove this. In case you want to retrieve it, it will always be stored in previous commits in this PR, so it's never lost.

4ian:
Let's remove the comment JS function, and seems like we're good to go with this? :)

Extensions/PrimitiveDrawing/shapepainterruntimeobject-pixi-renderer.js
Bouh: ok, then i delete also the comment for ArcTo

4ian: Yes, better keep the code clean. ArcTo can always be added later.

4ian: Yes, better keep the code clean. ArcTo can always be added later.

4ian:
This looks good to me to merge! Thanks again for adding this, will be super useful for games needing advanced shapes or effects 👍

PR #1259, opened 2019-10-20
Comment and Group are added on top of selection

Bouh:
Comment and Group are added on top of selection

### DIFF
--- a/newIDE/app/src/EventsSheet/index.js
+++ b/newIDE/app/src/EventsSheet/index.js
@@ -315,8 +315,10 @@ export default class EventsSheet extends React.Component<Props, State> {
     if (context) {
       insertions = [context];
     } else if (hasEventsSelected) {
-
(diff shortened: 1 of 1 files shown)


newIDE/app/src/EventsSheet/index.js
```diff
@@ -306,6 +306,7 @@ export default class EventsSheet extends React.Component<Props, State> {
   addNewEvent = (type: string, context: ?EventContext): Array<gdBaseEvent> => {
     const { project } = this.props;
     const hasEventsSelected = hasEventSelected(this.state.selection);
+    let insertTopOfSelection = 0;
```
4ian: What do you think of making this a boolean? and then subtract 0 or 1 according to the boolean?

Bouh: Yes it's better 👍

4ian:
Sounds fair, let's go ahead with this! :) Can you run prettier (npm run format) before I can merge this?

PR #1260, opened 2019-10-20
Avoid open non-GD5 files

Bouh:


newIDE/app/src/MainFrame/index.js
```diff
@@ -409,6 +409,24 @@ class MainFrame extends React.Component<Props, State> {
       .then(({ content, fileMetadata }) => {
         const serializedProject = gd.Serializer.fromJSObject(content);
 
+        if(!content.gdVersion && content.eventsFunctions){
```
4ian: Would it be possible to move this to a function in new file? 
Probably a file called "ProjectContentChecker" in "src/ProjectStorage". You can name the function "verifyProjectContent" and pass `(i18n, content)` as arguments.

(In this new file, you'll need to add flow comment at the top, and import showErrorBox, t and I18nType)

Also you would need to call this method before `const serializedProject = gd.Serializer.fromJSObject(content);`, because this method is creating a new object in libGD.js memory. If `serializedProject.delete()` is not called, this is creating a memory leak (serializedProject will stay forever in memory).

Bouh:
- I don't understand very well, i have found two i18n in @lingui.
If i pass i18n in the argument i don't need to import it in the new file, no ?

- I  don't understand the principe behind serializedProject in the new function. It's already called in MainFrame/index.js at line 411.

- We i can see the showErrorBox but because my return is in my new function the project will open.

If I have to use a 'promise' or 'then' / 'catch', I'm not very familiar with that.

newIDE/app/src/ProjectsStorage/ProjectContentChecker.js
```diff
@@ -0,0 +1,31 @@
+// @flow
+import { showErrorBox } from '../UI/Messages/MessageBox';
+import { t } from '@lingui/macro';
+import { I18n } from '@lingui/react';
+import { type I18n as I18nType } from '@lingui/core';
+
+const gd = global.gd;
+
+export default function verifyProjectContent(i18n, content) {
+  const serializedProject = gd.Serializer.fromJSObject(content);
```
4ian: You don't need to use fromJSObject here: it's used to transform "content" (which is a JSON parsed object) to a gd.SerializerElement (a representation of JSON in memory in libGD.js).
So you can remove this line entirely.

newIDE/app/src/ProjectsStorage/ProjectContentChecker.js
```diff
@@ -0,0 +1,31 @@
...
+  if (!content.gdVersion && !content.eventsFunctions) {
+    showErrorBox(
+      [
+        i18n._(t`Unable to open this file.`),
+        i18n._(t`This file not a file openable in GDevelop 5.`),
+      ].join('\n')
+    );
+    return false;
+  }
+  serializedProject.delete();
```
4ian: Remove this line as not useful here.
`return true;` to indicate that "content" seems legit.

newIDE/app/src/ProjectsStorage/ProjectContentChecker.js
```diff
@@ -0,0 +1,31 @@
+// @flow
+import { showErrorBox } from '../UI/Messages/MessageBox';
+import { t } from '@lingui/macro';
+import { I18n } from '@lingui/react';
+import { type I18n as I18nType } from '@lingui/core';
+
+const gd = global.gd;
+
+export default function verifyProjectContent(i18n, content) {
```
4ian: You have to type `verifyProjectContent`: 
`verifyProjectContent(i18n: I18nType, content: Object): boolean {`

otherwise flow will yell at you:
image

(usually `Object` is not a great typing, but it's actually the proper thing to use here).

newIDE/app/src/ProjectsStorage/ProjectContentChecker.js
```diff
@@ -0,0 +1,31 @@
+// @flow
+import { showErrorBox } from '../UI/Messages/MessageBox';
+import { t } from '@lingui/macro';
+import { I18n } from '@lingui/react';
```
4ian: I think this import is useless

newIDE/app/src/ProjectsStorage/ProjectContentChecker.js
```diff
@@ -0,0 +1,31 @@
...
+
+const gd = global.gd;
+
+export default function verifyProjectContent(i18n, content) {
+  const serializedProject = gd.Serializer.fromJSObject(content);
+  if (!content.gdVersion && content.eventsFunctions) {
+    showErrorBox(
+      [
+        i18n._(t`Unable to open this file.`),
+        i18n._(t`This file is an extension file for GDevelop 5.`),
```
4ian: Let's be a bit more descriptive: t`This file is an extension file for GDevelop 5. You should instead import it, using the window to add a new extension to your project.`

newIDE/app/src/ProjectsStorage/ProjectContentChecker.js
```diff
@@ -0,0 +1,31 @@
...
+      ].join('\n')
+    );
+    return false;
+  }
+
+  if (!content.gdVersion && !content.eventsFunctions) {
+    showErrorBox(
+      [
+        i18n._(t`Unable to open this file.`),
+        i18n._(t`This file not a file openable in GDevelop 5.`),
```
4ian: `This file is not recognized as a GDevelop 5 project. Be sure to open a file that was saved using GDevelop.`

(a good rule of thumb for error messages: 1) say the error 2) explain what the user should do)

newIDE/app/src/MainFrame/index.js
```diff
@@ -408,6 +409,26 @@ class MainFrame extends React.Component<Props, State> {
...
+        if(!content.gdVersion && !content.eventsFunctions){
+          showErrorBox(
+          [i18n._(t`Unable to open this file.`), i18n._(t`This file not a file openable in GDevelop 5.`)].join(
+            '\n'
+          )        
+        );
+          return false;
+        }
+*/
+        verifyProjectContent(i18n, content);
```
4ian: If this method returns false, you can then return (otherwise the function will continue as before).

```
if (!verifyProjectContent(i18n, content)) {
  // The content is not recognized and the user was warned. Abort the opening.
  return; 
}
```

Finally, make sure to do this *before* `const serializedProject = gd.Serializer.fromJSObject(content);`, otherwise we'll leak memory

newIDE/app/src/MainFrame/index.js
```diff
@@ -409,6 +410,11 @@ class MainFrame extends React.Component<Props, State> {
       .then(({ content, fileMetadata }) => {
         const serializedProject = gd.Serializer.fromJSObject(content);
```
Bouh: > Put this before gd.Serializer.fromJSObject

It's already done here

4ian: I've fixed it ;)

4ian:
Merging this now, thanks for adding this :)

PR #1261, opened 2019-10-20
Add a "Scene just resumed" condition

Wend1go:
This adds a new condition to check if the scene has just been resumed after it was paused.

I don't know how to get rid of the tween commit d5f4406. Already tried to pull from upstream master into my branch. :frowning_face:


GDJS/Runtime/runtimescene.js
```diff
@@ -756,3 +762,13 @@ gdjs.RuntimeScene.prototype.getAdhocListOfAllInstances = function() {
...
+
+/**
+ * Check if the scene was just resumed.
+ * This is true during the first frame after the scene has been unpaused.
+ *
+ * @returns {boolean} The scene was just resumed
+ */
+gdjs.RuntimeScene.prototype.sceneJustResumed = function() {
+    return this._isJustResumed;
+}
```
4ian: nitpicking: can you configure your editor to always add a new line at the end of files? (it's a weird convention, but better follow it).

GDCpp/GDCpp/Extensions/Builtin/RuntimeSceneTools.cpp
```diff
@@ -71,6 +71,10 @@ bool GD_API SceneJustBegins(RuntimeScene &scene) {
   return scene.GetTimeManager().IsFirstLoop();
 }
 
+bool GD_API SceneJustResumed(RuntimeScene &scene) {
+  return scene.SceneJustResumed();
```
4ian: You started to implement this for the C++ game engine, but the C++'s RuntimeScene does not have this method. I would suggest to not implement this for GDCpp. Instead:
remove this for RuntimeSceneTools.cpp/.h and remove the declaration from `GDCpp/GDCpp/Extensions/Builtin/SceneExtension.cpp` too.
When `SetFunctionName` is not called for a "platform" (GDCpp here), the IDE won't show the function if the game is using this "platform".

4ian:
Looks good! Maybe let's remove the half implementation in C++, it's not needed and better don't implement it as it won't be used in GD5 anyway.
For the commit issues, what you can do is trying to fetch the upstream (i.e: my) master branch into a new branch:

```
git remote add upstream https://github.com/4ian/GDevelop
git checkout -b 4ian-master upstream/master
```

`4ian-master` should now be a clean version of master. You can start new branches from here `git checkout -b new-feature` (to do from 4ian-master). You can cherry pick your commit from another branch (`git cherry-pick <commit number>`).
On `4ian-master`, always do `git pull upstream master` to only get new commits from my master branch.

4ian:
Keep a "backup" of your branch by making another branch from it (`git checkout -b my-saved-branch`) :) What is great with git is that you can just make a branch and it will keep the commits in case you have an issue.

Wend1go:
This worked perfectly. :+1: 
Would it be possible (a good idea) to replace my 'master' by your master branch?
(create a backup branch of my master and delete the original master, then branch your master and name it `master`)

4ian:
Yes, you can backup your master into a `old-master` branch. Then remove master (`git br -D master`) and re-create it from mine.

PR #1262, opened 2019-10-21
Add "Scene just resumed" condition (next try)

Wend1go:
Wow this looks pretty. :heart_eyes: 
Thanks for the tips. Cherry-picking is a powerful feature.


4ian:
Looks good! Thanks for adding this 👍

PR #1268, opened 2019-10-25
Detail what type of variable in description

Bouh:
I take the example with raycast we don't know what type of variable is used for storing intersection points.
So i precise it in this commit, for raycast action and some other.


4ian:
Looks good 👍

PR #1269, opened 2019-10-26
Keep selection when panning & fix various details

Bouh:
Now we keep selection when we use pan movement with spacebar+mouse or just mouse middle click.
& Little fix for use good icons
& Fix big/small caps in wording


newIDE/app/src/InstancesEditor/index.js
```diff
@@ -113,9 +113,13 @@ export default class InstancesEditorContainer extends Component {
       this.props.height
     );
     gesture.panable(this.backgroundArea);
-    this.backgroundArea.on('mousedown', event =>
-      this._onBackgroundClicked(event.data.global.x, event.data.global.y)
-    );
+    this.backgroundArea.on('mousedown', event => {
+      if (event.data.button === 1 || this.keyboardShortcuts.shouldMoveView()) {
```
4ian: Ideally we want to avoid messing with event propagation and hardcoding buttons there: most of the event handling functions are just redirecting to other functions (like _onBackgroundClicked) that are "higher level" and are supposed to understand what's going on. 

In this case, when the user does the action of moving the mouse (or the touch) and drag it, it's called a "pan". At the end, _onPanEnd is called, meaning that this action of click + move is finished. Two possibilities: either the user moved the view (`shouldMoveView()` returns true) or a selection rectangle was finished.

If you look at `_onPanEnd`, you can see that there is a comment about this, and a check:
`if (this.selectionRectangle.hasStartedSelectionRectangle())`

Unfortunately, it seems that this check if true, leading to a call to `selectInstances`, which select... 0 instances as the selection rectangle contains nothing.

So now the question is, why `hasStartedSelectionRectangle` returned true?
Turns out that as soon as you click on the background, `_onBackgroundClicked` is called. (which is normal) 

Inside this function, you can see that a selection rectangle is always started: 

https://github.com/4ian/GDevelop/blob/657ca853240eae806f3d04353a20e693d64fe0d9/newIDE/app/src/InstancesEditor/index.js#L413

This might be the culprit! Add a `if (!this.keyboardShortcuts.shouldMoveView()) {  }` around it.
Now the click on the background should properly start a selection ONLY if the view is not being moved. So it won't be messing with the pan ending (`_onPanEnd`) which was wrongly thinking a selection rectangle was drawn :)

This seems actually be a regression introduced recently when I moved the selection rectangle start into `_onBackgroundClicked` to ensure the starting point is always correct even in case of lag.

4ian:
Don't forget to run `npm run format` before committing changes ;) Otherwise tests will be indicated as failing. I've done it for you.

Rest looks fine, merging this now! Thanks for these updates and this fix.

Bouh:
With this fix mouseMidButton select still the instance. 
Fix is good just for spacebar + LMB

4ian:
In this case, we might don't do the call to _onBackgroundClicked if it's not the left button. Somewhat similar to what you did originally. Do you want to try this?

Bouh:
I tried other things, the mouse button is not well recognized, looking with the chrome debugger it's as if the mouse returned true only for the left button regardless of the click pressed.
All the others button remain false.

I'am lost

PR #1283, opened 2019-11-09
Clamp in mathematical tool

Bouh:
Now we have `clamp(value, min, max)` in mathematical tools.

Not sure about c++ code in core.
Tested and working on electron-app  and web-app, i use now clamp in my video extension and it's working too.

[Trello Card](https://trello.com/c/CO42UG0b/345-add-a-clamp-function)
[Added on wiki](http://wiki.compilgames.net/doku.php/gdevelop5/all-features/functions#mathematical_tools)


4ian:
Both implementations looks good 👍Merging this. Travis-CI was not compiling for unrelated reasons, hopefully it's fixed in master.

PR #1286, opened 2019-11-13
Guideline step-by-step

Bouh:
This PR is highly buggy, I open it to have a technical discussion, and get your feedback.

The idea is a tutorial of a small platform game.
A platform tutorial has already existed for several years, I think it would be perfect to serve as a model.

When the last popover come, it indicates that the tutorial is finished and that if the user wants to continue learning he can check out the wiki. 
The current main page of the wiki already redirects to the written tutorial he will have done, the other tutorials, the concepts of game creation.

For the technical side:
I have a piece of code with the Popover component. It can explains what to do in a few words, with an image/gif if necessary.

I also have the idea of redesigning the start page a little to include this new start help and especially that make it ultra visible in addition to the link to the wiki.

I have several problems with the code of this PR and I don't know yet how to fix them.

* The popover blocks all the user's other actions., Popper or tooltip are better, but it's depending how we attach on existing component.
* I don't know how get position of specific component for attach the popover on it, maybe with the name of the component?
* The "Open examples" button should open the window with the "examples" tab.
* The "Start interactive tutorial" button should open a empty project and open my popover (This is just props/states, i'll find how to)

There are licenses on the image. These are images that I take as an example, not sure if we will use this one in the end. The logo of the images come from here: 
https://www.flaticon.com/free-icon/creativity_1497573
https://www.flaticon.com/free-icon/abc_748488
https://www.flaticon.com/free-icon/search_748569

https://trello.com/c/aT9hEXpV/378-guidelines-for-new-users


newIDE/app/src/guidelines/index.js
```diff
@@ -0,0 +1,214 @@
...
+    if (guidelines[this.state.indexData].imageSource !== undefined) {
+      imageTutorial = (
+        <img
+          class="fit-picture"
+          src={guidelines[this.state.indexData].imageSource}
+          alt={guidelines[this.state.indexData].imageAlt}
+        />
+      );
+    } else {
+      imageTutorial = () => {
```
4ian: imageTutorial = null;

should work :)

newIDE/app/src/guidelines/index.js
```diff
@@ -0,0 +1,214 @@
+import React, { PureComponent } from 'react';
+import { render } from 'react-dom';
```
4ian: You probably don't need this and don't want to use this :)

newIDE/app/src/guidelines/index.js
```diff
@@ -0,0 +1,214 @@
...
+      let val = currentState.index;
+      if (val === currentState.indexMax) {
+        val = currentState.indexMax;
+      } else {
+        val = currentState.index + 1;
+      }
+      return {
+        index: val,
+      };
+    });
```
4ian: Something like:
```
this.setState(currentState => ({
  index: Math.min(currentState.indexMax, currentState.index + 1),
}));
```

should work, avoiding the let/if/return

newIDE/app/src/guidelines/index.js
```diff
@@ -0,0 +1,214 @@
...
+  };
+
+  componentDidUpdate(prevProps, prevState) {
+    if (this.state.index !== prevState.index) {
+      this._inputRef.current.updatePosition();
+
+      setTimeout(() => this.setState({ indexData: this.state.index }), 300);
```
4ian: Any reason behind this timeout?

newIDE/app/src/guidelines/index.js
```diff
@@ -0,0 +1,214 @@
...
+        >
+          Finish
+        </Button>
+      );
+    } else {
+      nextOrFinish = (
+        <Button
+          variant="contained"
+          color="primary"
+          disabled={this.state.index === this.state.indexMax ? true : false}
```
4ian: nitpicking:
`disabled={this.state.index === this.state.indexMax}`

newIDE/app/src/guidelines/index.js
```diff
@@ -0,0 +1,214 @@
...
+        <Button variant="contained" onClick={this.togglePopover}>
+          Close
+        </Button>
+      );
+    }
+
+    if (guidelines[this.state.indexData].imageSource !== undefined) {
+      imageTutorial = (
+        <img
+          class="fit-picture"
```
4ian: Not sure `class="fit-picture"` is actually doing anything?

newIDE/app/src/guidelines/index.js
```diff
@@ -0,0 +1,214 @@
...
+        >
+          <Grid container style={styles.container}>
+            <Grid item xs={12} style={styles.description}>
+              <Typography variant="h5">
+                {guidelines[this.state.indexData].title}
+              </Typography>
+              <Typography wrap="true">
+                {guidelines[this.state.indexData].description}
+              </Typography>
+              <br />
```
4ian: Avoid `br`, prefer using `Line` from `UI/Grid.js`:

```
<Line>
  <Text>Blabla</Text>
</Line>
<Line>
  {imageTutorial}
</Line>
```

newIDE/app/src/guidelines/index.js
```diff
@@ -0,0 +1,214 @@
...
+              {imageTutorial}
+            </Grid>
+            <Grid
```
4ian: You might want to take a look at `Line` and `Column` from `UI/Grid` - it's a basic grid system used almost everywhere in the UI. Should be enough for 90% of what we need. Don't spend too much time adapting your stuff for now though - we can always easily rebuild the UI of your popovers later.

newIDE/app/src/guidelines/index.js
```diff
@@ -0,0 +1,214 @@
...
+        />
+      );
+    } else {
+      imageTutorial = () => {
+        return;
+      };
+    }
+
+    return (
+      <div className="app">
```
4ian: Remove `className` that is probably useless?

4ian:
>The popover blocks all the user's other actions., Popper or tooltip are better, but it's depending how we attach on existing component.

Yeah surely a matter of using the proper component or passing a prop to tell that there is no invisible "overlay" that blocks the clicks (search for overlay or related stuff in the documentation).

> I don't know how get position of specific component for attach the popover on it, maybe with the name of the component?

Yep that's a big challenge, you need to find a way to cleanly add some "ids" to DOM elements or a way to "register" your components into a context/state/object and use this to position your popover.

newIDE/app/src/guidelines/index.js
Bouh: It's for force update the component after debounce, but the position doesn't change.
But it's changing [well here](https://codesandbox.io/s/updateposition-qui-fonctionne-0sqrs) in sandbox.

newIDE/app/src/guidelines/index.js
Bouh: Lot of code are useless it's a first draft

Bouh:
The Popper component is better because Popover doesn't seem to have any props to allow the user to click everywhere.
So let's go for Popper. In addition, according to the example, it adjusts better to the content if you resize the window.


To anchor the bubble I like the idea of just using an identifier to put the bubble on it. And we can do that with **anchorEl** props on Popper.

If I understood the idea of the React Context, it is necessary to wrap all the components useful for the tutorial?
I'm not a fan of wrapping all the components clicked by the user with the bubble component or anchor component, it adds a lot of code for a tutorial used maybe 2 or 3 times per user in the software.
I prefer that this feature uses as little space as possible to anchor itself.
Maybe I misunderstood?

To put an attribute in a component is a problem, once the component is rendered the attribute disappears in versions inferior to version 16 of React.
https://reactjs.org/blog/2017/09/08/dom-attributes-in-react-16.html

In the new commit I tried to get with hardcoded the social network element with a className and print it in the console.
If it is not a props then the attribute is deleted.

I've added the `//TODO` on my code it's where i'am stuck and where i work.

newIDE/app/src/MainFrame/Editors/StartPage/index.js
```diff
@@ -160,7 +238,8 @@ class StartPage extends BaseEditor {
                     }
                   />
                 </Line>
-                <Line alignItems="center">
+                {/*TODO Le className disparait au rendu*/}
+                <Line alignItems="center" className="socialNetwork">
```
4ian: Line is a component, not a DOM element, so className is passed as a prop to "Line", then unused in the render method of Line. So it's not appearing anywhere. Flow will yell at you for passing a prop that is not recognized (`npm run flow`) (much like what would happen in TypeScript).

I don't think it's a good idea to add identifiers on "Lines" anyway. They are component that are used to describe the layout/structure of the UI. Instead, we can do two things:
1. Add a "identifier" prop on some common UI components (mainly the buttons and icons). Then, in their `render` method, pass the prop "identifier" to the material-ui component (and material-ui will hopefully pass it to "divs" and other DOM elements). Or if the `render` of the component is showing a div around some material-ui components, you can also put the identifier here.
2. We can also make a `GuidelineMarker` component that is just a div that render the children, with the identifier. Something like:

```
const GuidelineMarker = ({ identifier, children }) => <div className={"guideline-" + identifier}>{children}</div>
```

You'll see that I've prefixed the className with "guideline-", to avoid risking an already existing CSS class to be added to this div. Note that I'm not a super big fan of this "extra div", as it will probably break the layout of components... so we should aim at adding such "identifier" props to a few selected UI components I guess, like RaisedButton, FlatButton, icons (see most of them in the `newIDE/app/src/UI` folder)

newIDE/app/src/MainFrame/Editors/StartPage/index.js
```diff
@@ -160,7 +238,8 @@ class StartPage extends BaseEditor {
                     }
                   />
                 </Line>
-                <Line alignItems="center">
+                {/*TODO Le className disparait au rendu*/}
```
4ian: See my other comment, but just so you understand:

`<div className="blabla">, <span className="blabla">, <a className="blabla">, etc...` will work because they are DOM elements.
`<MyComponent className="blabal" />` is a *React component*. className is just a prop like any other: if you don't use a prop (by passing it down to children), then it's useless. React won't magically add it because your component might not even be composed of any DOM element - in a lot of cases, React components are just rendering... other React components :)

newIDE/app/src/MainFrame/Editors/StartPage/index.js
4ian: Additional note again, I've used "className" everywhere as an example, but I would prefer a "data-guideline-identifier" in the final solution :) className should only be for the styling.

Bouh:
- In the last commit Popper uses the marker and anchors itself correctly.
There is the refresh which is not taken into account however a setState has passed. 
The Popper does not seem to have any element to do an updatePosition like the PopOver.

- I forced @Flow? to use certain types because I'm not sure what type to define.

- The little arrow on the Popper is missing.

- check again the button for Restart the tutorial to zero. (in StartPage)

(Let's try to fix these problems and then we will thinking about whether a user has done the right thing to display the next step.
And if he didn't do the right thing with the acutel step, warn him.)

newIDE/app/src/MainFrame/Editors/StartPage/index.js
```diff
@@ -36,6 +40,9 @@ const styles = {
   },
 };
 
+//TODO
+//Forcé l'ouverture de onCreate, mais avec l'onglet "example"
```
4ian: This could be done using two classic solution in React:
1. "lifting" the state of the dialog showing the starter/examples into the MainFrame, storing a `createDialogCurrentTab` variable in MainFrame `state` so that the MainFrame own the current tab and is able to change it. This is usually what we do in React. It's called a **controlled** component. You pass this `createDialogCurrentTab` state as a prop to the create dialog, and you pass a callback ("changeCurrentTab") that is called when the user click on a new tab. Finally, you remove the state variable containing the tab inside the create dialog (and you replace it by a prop).

This pattern of removing the state from somewhere to put it in the parent is called **lifting the state**. 
But this would complicate MainFrame and I think that it's better to keep MainFrame unaware of these tabs as possible. So instead, we can do another solution:

2. Store an additional `state` variable in MainFrame which is "createDialogInitialTab" (notice that the name is now "initial" instead of "current"). Then, pass just this prop to the create dialog (no need for a callback). In the constructor/initialization of the create dialog, you use this props to know what to put in the state of the current tab.

When you want to open the dialog, you call `openCreateDialog` in MainFrame, passing an additional argument which is the "initialTab", that you put in the state "createDialogInitialTab". This will result in the dialog opening (because you also set `createDialogOpen` to true in openCreateDialog) AND the initial tab being used by `CreateProjectDialog` to know what the initial tab is.

```
state = {
    currentTab: this.props.initialTab,
  };
```

This is called an **uncontrolled component** (because you can pass it a few details, but then the component is managing its state itself, without any "help" from its parent).

4ian: It's "classic" react stuff here, so not a priority for this PR but can help you to understand the logic behind React :)

newIDE/app/src/guidelines/index.js
```diff
@@ -0,0 +1,205 @@
...
+
+  back = () => {
+    this.setState(currentState => ({
+      index: Math.min(currentState.indexMax, currentState.index - 1),
+    }));
+  };
+
+  componentDidMount() {}
+
+  componentDidUpdate(prevProps: *, prevState: *) {
```
4ian: `(prevProps: Props, prevState: State)` should work?

newIDE/app/src/MainFrame/index.js
```diff
@@ -493,12 +504,27 @@ class MainFrame extends React.Component<Props, State> {
       });
   };
 
+  restartGuidelines = () => {
+    this.setState({
+      guidelinesOpen: !this.state.guidelinesOpen,
+      //TODO
+      //Ici je passe un index 0 et je le transmet en props à Guidelines component ou il y a plus simple ?
```
4ian: So you can put an index to 0 here and give it to Guidelines as a prop yes. The only "gotcha" is that you have a duplication of the state: you have the "index" that is both in MainFrame and in Guidelines. You have two solutions as explained before for something else:

- either Guideline is "controlled", taking the current index as a prop. This means that when you want to change the index, you have to call a prop "changeIndex" that is doing a setState in MainFrame, to change index. (then, React will re-render MainFrame, so re-render Guidelines, so Guidelines will get the updated "index").

This way is usually the preferred in React. We say that the data "flows" from the top component (MainFrame) to the bottom (Guidelines) (you can consider the props as "pipes" flowing data from top components to bottom :))

* OR, you keep the current index in guidelines BUT you pass a `initialIndex` to Guidelines. In other words, you do an "uncontrolled" component. This being said, it's hard to "reset" anything, because by definition, reseting something is an "imperative" thing to do (rather than declarative). The issue is: how to know in Guidelines that you must reset the index to 0?? :/ The `initialIndex` was probably already 0 before!
So the actual solution is to use React "refs" to do imperative things, i.e: call a method on `Guidelines`. Like this:

1. Use `createRef` and `ref={...}` to store a reference to `Guidelines` inside `MainFrame`.
2. In Guidelines, create a new method called `reset` and that do `this.setState({ index: 0 })` (or anything else to go back to the first step).
3. In MainFrame, in restartGuidelines, you call the ref:

```
restartGuidelines = () => {
  // this._guidelines.current is a reference to your component. Beware, it can be `null` if for some reason it was not rendered on the screen by React.
  if (this._guidelines.current) this._guidelines.current.reset();
}
```

This "imperative" call to `reset` is usually not the best way to do things in React, but it can help, as before, to avoid "lifting the state" in MainFrame. It's ok here as we expect Guidelines to be more or less autonomous (MainFrame should not have any logic related to the guidelines).

Final note (explore this later): you can see that sharing component is easy in React, but sharing logic is hard. Because you have to "lift" the state everytime. So that's why React team created the "React hooks", to make it easy to share logic. But we don't use them in MainFrame.

newIDE/app/src/MainFrame/index.js
```diff
@@ -1300,6 +1330,12 @@ class MainFrame extends React.Component<Props, State> {
     });
   };
 
+  onShowExamples = (open: boolean = true) => {
+    this.setState({
+      createDialogOpen: open,
+    });
```
4ian: So here I suggest storing an extra "createDialogInitialTab"

newIDE/app/src/guidelines/index.js
```diff
@@ -0,0 +1,205 @@
...
+  arrowRef: *,
+|};
+
+type Props = {|
+  open: boolean,
+  closeHandler: () => void,
+|};
+
+//TODO
+//PureComponent ou React.Component<Props, State> voir la doc pour comprendre la différence.
```
4ian: Consider PureComponent as a way to optimize the rendering. If your component is only changing when the props are changing, use PureComponent. Otherwise, stick to Component.

newIDE/app/src/guidelines/index.js
```diff
@@ -0,0 +1,205 @@
...
+      anchor: null,
+      arrowRef: null,
+    };
+  }
+
+  componentWillReceiveProps(newProps: *) {
+    let elementHTML = document.querySelectorAll(
+      '.guideline-' + guidelines[this.state.indexData].positionBind
+    )[0];
+    const element = ReactDOM.findDOMNode(elementHTML);
```
4ian: Beware, you're confusing React with pure HTML/JS. findDOMNode is used to find the DOM element of a React component.
Here, you don't actually need React at all :)

You have the class of the DOM element to find (guideline-xxx), so you:
1. Get the DOM element with `document.querySelectorAll`.
2. This is a *list* so check if it's not empty.
3. If it's not empty, get the first element. (if it is empty, you just return, potentially you can log a warning).
4. This element is... a DOM element already :) So no need for findDOMNode. Just pass the element to `anchor` using this.setState(). The Guidelines component will be re-rendered with the new "anchor", which will be sent to the Popper.

4ian: Note that you're doing all of this inside "componentWillReceiveProps", called when props are changed. In practice, you want to run this (you can put this in a method `_updatePopover element`) when... indexData is changing :) Actually indexData is a duplicate of index => remove it and use index directly.

newIDE/app/src/guidelines/index.js
```diff
@@ -0,0 +1,205 @@
...
+    this.setState(currentState => ({
+      index: Math.min(currentState.indexMax, currentState.index - 1),
+    }));
+  };
+
+  componentDidMount() {}
+
+  componentDidUpdate(prevProps: *, prevState: *) {
+    if (this.state.index !== prevState.index) {
+      this.setState({ indexData: this.state.index });
```
4ian: Duplicating a variable of a state is in 99% of the cases, a sign that things are too complicated. Instead, you want to update the position of the `Popper` when the index is changed. So you do:

```
if (this.state.index !== prevState.index) {
  this. _updatePopover();
}
```

instead, with `_updatePopover` being the logic calling `querySelectorAll` and updating `anchor`. As `_updatePopover` will update the state `anchor`, the component will be re-rendered with the proper `anchor` passed as a prop to the Popper.

And you actually don't even need `componentDidUpdate`. Instead, call `this._updatePopover()` whenever you change `index`. This means that indeed of doing:

1. You change `index` state in some function
2. React re-renders the Guidelines component
3. React calls Guidelines `componentDidUpdate`
4. You see that index changed, so you call `_updatePopover`
5. `_updatePopover` updates the `anchor` state.
6. React re-renders the Guidelines component
7. Finally, Popper is re-rendered with the proper Anchor

you just do:

1. You change `index` state in some function
2. You call `_updatePopover`, which updates the `anchor` state.
3. React re-renders the Guidelines component
4. Popper is re-rendered with the proper Anchor

which is a bit "simpler" :)

4ian: To do something just after the state is updated, use a second argument for setState:

```
this.setState( index: ..., () => this._updatePopover())
```

This guarantees that `this.state.index` will have the new value when `this._updatePopover()` is called. (because React can batch the calls to setState, so you risk to read a value of this.state.index that is NOT updated if you call `_updatePopover` just after `this.setState`.)

Bouh: Wow really better like this ! Refresh working now 💯 
(A thing, callback is executed after the execution of the function so in fact we said to React to make a rendering throught setState(), then it starts the callback but our callback calls a setState() which restarts a rendering.)

4ian: Yup strictly speaking as we call updatePopover in a callback there are two renderings. It's not a big problem, you could maybe avoid it entirely by passing an index argument to updatePopover and doing only one setState.

But it's not a big deal, the most important is to keep a state that is clean and simple :) (i.e: no extra variable, no calls to the "weird" React lifecycle methods if we can avoid)

Bouh:
Well, it seems I didn't understand correctly since it doesn't work.

4ian:
If you want to create a separate CSS file, you need to include it in the themes index.js files as done for the events sheet css file.

> I didn't want to copy the whole css of Guideline into DefaultTheme and DarkTheme i don't want to duplicate all guideline css for two theme.

In this case, create a single file Guidelines.css and import once it in Guidelines js file OR use inline styles in JavaScript (`const styles = { }`):

* In the case of a single Guidelines.css file, you use ThemeConsumer to fetch the name of the class to apply according to the theme, and you can have two new Guidelines.css files in DefaultTheme/DarkTheme folders that are implementing the differences between the themes.
* In the case of inline styles in JavaScript, you use ThemeConsumer to fetch the variables that you can inject in the styles in the `render` (`style={{...styles.containerStyle, backgroundColor: theme.guidelines.backgroundColor}}`)

Up to you to choose a solution, though usually I prefer to avoid CSS files as they are global CSS rules that don't follow material-ui default theme, have no proper scoping, can't be passed variables and, being global, risk being used for unrelated purpose one day by someone.

4ian:
In terms of functionality, what's the current state? Do you need to check if a button is clicked to go to the next step?
Styling/CSS can be done later, it's not a real blocker.

Bouh:
> In terms of functionality, what's the current state? Do you need to check if a button is clicked to go to the next step?
> Styling/CSS can be done later, it's not a real blocker.

Yes i need catch the click on a specific element with the ID "clickTarget"
Data.js can have `positionBind` for set where the Guideline popper will appear.
`clickTargetBind` for highlight the button to click.
description, title, imageSource, forceArrowOrientation

Bouh:
Oh no why this merge come... I hope it's will not affect this PR.

4ian:
Mmm no the PR seems fine? I can only see your commits, no commits from master. It should be good! You can continue like this :)

newIDE/app/src/UI/Theme/DefaultTheme/Guidelines.css
```diff
@@ -0,0 +1,20 @@
+/**
```
4ian: Yes that's the right approach, a CSS file with rules for the default-theme (`gd-guidelines-default-theme`) and dark theme in two separate files, and a third file with the common rules :)

newIDE/app/src/UI/Theme/DefaultTheme/Guidelines.css
```diff
@@ -0,0 +1,20 @@
...
+* Guideline arrow color
+* The other css are in "newIDE\app\src\guidelines\style.css"
```
4ian: Nitpicking: "The other CSS rules are [...]".

newIDE/app/src/guidelines/index.js
```diff
@@ -325,14 +311,14 @@ export default class GuidelinePopOver extends PureComponent<Props, State> {
                         <Line>{closeOrWikipage}</Line>
                         <Line alignItems="center">
                           <Typography style={{ marginRight: 10 }}>
-                            {this.state.index + 1} of {this.state.indexMax + 1}
+                            {this.state.index + 1} <Trans>of</Trans> {this.state.indexMax + 1}
```
4ian: You can actually use variables in `<Trans>` :) `<Trans>{this.state.index + 1} of {this.state.indexMax + 1}</Trans>`

js-lingui will generate a translation of the form `{0} of {1}` - which is useful for languages where the order of words is reversed.

newIDE/app/src/guidelines/index.js
```diff
@@ -298,25 +286,23 @@ export default class GuidelinePopOver extends PureComponent<Props, State> {
...
                         <Line>
                           <Typography variant="h5">
-                            {guidelines[this.state.index].title}
+                            <Trans>{guidelines[this.state.index].title}</Trans>
                           </Typography>
                         </Line>
                         <Line>
                           <Typography wrap="true">
-                            {guidelines[this.state.index].description}
+                            <Trans>{guidelines[this.state.index].description}</Trans>
```
4ian: This won't work. `<Trans>` is uniquely for *extracting* english sentences to be translated and shown on screen.
Here, it's properly shown on screen BUT it's just a variable. If your variable is just a number or a name entered by the user it's fine. But here, it's an entire english sentence! 
It won't be extracted and not translated as considered as a variable.

Instead, you must:
- mark description with `t`: 

```
{
  ...
  description: t`Bla bla`
}
```

Show it with i18n._:

```
<i18n>
{({ i18n }) => (
  <Something>
    <SomethingElse>
    {i18n._(description)}
    </SomethingElse>
  </Something>
)}
<i18n>
```

Bouh:
It's progressing, it's nice to see some buttons already working.
I'm stuck on several points.

- [ ] Avoid to use setTimeout() for waiting what come after click on a button.
   - [ ]  Related? Tutorial don't display next content when the project manager should be open.
- [ ]  Open tutorial, open a new project, go back to start page, close tutorial, re-open it, Guideline use wrong position but i use reset but isn't working.

Others not very important for now :
- [ ] CSS ghost on Body when <Guideline> is open and a other dialog is open too.
`style="overflow: hidden; padding-right: 17px;"`
- [ ] Ripple effect use all window space when an Identifier is set on button

Bouh:
I'm proud to close my pull request as we have a built-in solution available on the homepage of the web app.

If I had one wish, it would be to have the ability to create a Json file that is made public in a Tutorials section in the editor so that anyone can create a custom step-by-step.

PR #1289, opened 2019-11-16
update piskel version

blurymind:
I made a new build of piskel based on https://github.com/blurymind/piskel/tree/piskel-plus (a300d17) here:
[piskel-editor.zip](https://github.com/4ian/GDevelop/files/3859182/piskel-editor.zip)

This PR updates GD to use that new build.

I've been working on piskel for a while, improving some of it's weak points. The new features are:
- color index shift brush (super useful for cell shading sprites) (1)
- ability to source all  layers when using the bucket tool
- palette transfer tool (apply the currently selected palette colors to the frame you are on) (2)
piskelNew

bucketToolSourcePiskel
piskelPaletteTool

### DIFF
--- a/newIDE/app/package.json
+++ b/newIDE/app/package.json
@@ -74,7 +74,7 @@
   },
   "scripts": {
     "postinstall": "npm run import-resources && npm run make-version-metadata",
-    "import-resources": "cd scripts && node import-libGD.js && node import-GDJS-Runtime.js && node import-monaco-editor.js && node import-zipped-external-libs.js && node import-zipped-editor.js piskel 5.0.0-beta82 a2c36775109a1c1181d0b9ab9b7c903365c4ecb340118f1237b66236f23e20dd && node import-zipped-editor.js jfxr 5.0.0-beta55 8ac12b557c2ddba958c6f0d3e0c5df8cf3369a65262dcb90cf5c8a7a7d20bdf6 && node import-zipped-editor.js yarn 5.0.0-beta80 5930c686ccf2b6fd3433b1c9539483473a6aedb8b6a73dd13c07030c95efecb9",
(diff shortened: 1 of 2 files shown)


4ian:
Thanks for these updates 👍
Looking at the folder, seems there are duplicate of some JS files and CSS files - there was only one CSS and two JS files previously, contrary to now where they are 2 and 4 respectively:

Screenshot 2019-11-17 15 09 13
Screenshot 2019-11-17 15 10 11

Can you ensure that you have a clean build of Piskel and reupload the zip? :)

blurymind:
Ah sorry, I must have zipped the wrong thing. I will try to update it tonight

blurymind:
@4ian I reuploaded it now :)

4ian:
I've uploaded it in beta 82 :)

4ian:
I've uploaded it in beta 82 :) 

You should be able to update the hash (put a fake hash package.json, console.log the real hash computed in import-zipped-editor, and then replace it again in package.json) and we're good.

Bouh:
Because the hash is outdated, i cannot run dev env on my pc

blurymind:
> I've uploaded it in beta 82 :)
> 
> You should be able to update the hash (put a fake hash package.json, console.log the real hash computed in import-zipped-editor, and then replace it again in package.json) and we're good.

How do I compute the real hash? I remember trying it with the zip last time and failing :) it has to be done on the folder instead of the zip file itself and needs to be sha256?

4ian:
import-zipped-editor is computing the real one, have you tried `console.log`to log it?
Update package.json to download the version stored in beta82, relaunch npm start, this will log you a warning about potential corrupted/tampered archive. Add a console.log at the right place in `import-zipped-editor` to see the hash the was computed for the folder.

4ian:
I'll do it if you don't manage to find the hash, probably tomorrow as I'm busy tonight though.

4ian:
Merged :) Thanks for improving Piskel!

blurymind:
> Merged :) Thanks for improving Piskel!

Thank you for merging it :)  at least GD users will get the improvements

Piskel's author hasn't replied to my email or the pm on twitter. I wonder if its worth waiting a month more and then forking it into a new pixel editing app (someone on twitter suggested calling it Lexip xD )

in any case there appears to be interest in other people to improve piskel, but because nobody reviews code at the main branch, its been frozen

PR #1293, opened 2019-11-17
Rework text field, some buttons and some layouts

4ian:
After discussing the text fields [here](https://github.com/4ian/GDevelop/issues/927#issuecomment-554348476) and remembering that some people think GD has not enough contrast or issues seeing what is a field/what is not a field, I think it's fair to say that the "flat" text field from the first version of Material Design are not the best for users. I've been reading a bit about text fields and turns out Material Design is now recommending outlined or filled text fields. They have written an [interesting research here](https://material.io/components/text-fields/#research): 

> Enclosed text fields with a rectangular (box) shape performed better than those with a line affordance
> The text field box should be represented with a semi-transparent fill and a bottom line or with a fully transparent fill and an opaque stroke.

Material-UI added these new field design in the version that GDevelop is now using: https://material-ui.com/components/text-fields/. 

I think it can solve some complaints and ease the learning curve for new users to use filled text fields instead of the old "flat ones". As much as I like the clean result of the old ones, they seem no optimal.

I've made the changes to use the filled version everywhere except:
* In "MiniToolbar", where I use still use the "Flat" version without any borders
* that's basically it :)

Ideally, I think we could also add a background to the flat version.

These new fields are taking **less space** than before, because the margins were reduced. The label is included in the background color as provided by material-ui and suggested by material.io - this is not changing the space used by the fields.

Here are a few comparisons:

Screenshot 2019-11-17 22 17 39
Screenshot 2019-11-17 22 18 28
Screenshot 2019-11-17 22 19 10
Screenshot 2019-11-17 22 20 38

Another few improvements:

* Fields that are on two columns on desktop/tablets will automatically be stacked in a single column on small windows/phones (this was already done in the properties panel, this is now generalised).
* Some buttons have been changed to be "RaisedButtons" instead of just icons, to make them more visible. I also set the label "Create" or "Edit" for the resource selectors, so that users can discover that they can use external editors :) 
* I've started a file "UI/Layout.js" that contains some utilities for commonly used layouts.

I know that visually this is **quite a big change** :) But I think that going forward this will be much more intuitive for users. After a few hours working with this, I already get used to it.

TODO: 
- [x] Fix margins of AutoComplete and SelectFields when isInline is true
- [x] Fix selection of project file type (isFolderProject) with SelectField

### DIFF
--- a/newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/PolygonEditor.js
+++ b/newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/PolygonEditor.js
@@ -11,6 +11,7 @@ import {
 import SemiControlledTextField from '../../../UI/SemiControlledTextField';
 import Warning from '@material-ui/icons/Warning';
 import IconButton from '../../../UI/IconButton';
+import ThemeConsumer from '../../../UI/Theme/ThemeConsumer';
 import AddCircle from '@material-ui/icons/AddCircle';
 import Delete from '@material-ui/icons/Delete';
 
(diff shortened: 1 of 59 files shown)


4ian:
Also I know there was the suggestion to only show a background on the part where we can actually write (not the label). It might looks visually a bit different but:
* this won't take less space on screen
* this is going against the UX research of material design
* we can still do it but this was more work so we can always try this later :) Hopefully, a few changes to the style of MuiInput/MuiFilledInput can help if someone want to try.

But I would rather test this to stay aligned with material conventions.

Bouh:
If you're still working on this take a look on "Project file type" in Game settings > Properties. value is sometime "true" and "false" string, but should be boolean. i see it in debugger state.isFolderProject in string type.
even if value type is boolean.

Bouh:
Input in Popper on eventsheet don't fit the container. And text is limited to light grey box.
image

Bouh:
Text are cropped on bottom. (height: 1.2em; instead 1.1875em on .MuiInputBase-input)
image

4ian:
Good catch. I wonder if for this "InlinePopover", we want to display all the fields inside with margin="none" (so that you get the version without background).

Adding a check for this before merging this.

4ian:
> If you're still working on this take a look on "Project file type" in Game settings > Properties. value is sometime "true" and "false" string, but should be boolean

I missed this! Yeah that seems like a bug, I think the support of SelectField for booleans is not properly done. To do proper fix, I'll update the true/false by strings ("single-file", "multiple-files") and compare this for setting isFolderProject.
Then remove the flow typing for "boolean" - I think it's the only select using a boolean anyway.

4ian:
I'll postpone this until a new version is released with fix for beta 82 UI bugs

4ian:
> Input in Popper on eventsheet don't fit the container. And text is limited to light grey box.

Fixed

> If you're still working on this take a look on "Project file type" in Game settings > Properties

Fixed in master: https://github.com/4ian/GDevelop/commit/b1d81d48360f1b40f52e83f0632bd1b35ac2df49

I think this branch is ready to be merged. 

As a side not, I'm not satisfied with the behavior of the autocomplete (but it's independent from these changes). I think I'll change it to use instead the new material UI autocomplete: https://material-ui.com/components/autocomplete/ - in particular, I don't like that the current module that I'm using for autocomplete is clearing the field when you press "Esc" key. No way to disable this! The material-ui autocomplete seems more customisable.
Will probably do that in another PR.

Bouh:
About SemiControlledAutoComplete, the chooses are displayed after an action on the field.
Maybe we can prupose to show suggestion list at the beginning.
And between field and list we have a space.

image

4ian:
> Maybe we can prupose to show suggestion list at the beginning.

Yes, though I remember people where also complaining that it was automatically opening the the action/condition editor.

> And between field and list we have a space.

I would rather change the implementation to use https://material-ui.com/components/autocomplete/ before doing more changes to this autocomplete :)

Bouh:
> Yes, though I remember people where also complaining that it was automatically opening the the action/condition editor.

Oh yes i remember, can we display the suggestion only in the eventsheet ?
Eventsheet is perfect for fast edition, so display the suggestion only here make sense.
I agree with peoples, in instruction editor this behavior are anoying.

4ian:
> Oh yes i remember, can we display the suggestion only in the eventsheet ?
> Eventsheet is perfect for fast edition, so display the suggestion only here make sense.

I think what we can do is to instruct the autocomplete to automatically show the menu if:
* we're "inline" (prop called "isInline") (which mean that the parameter field is being used in a popover rather than in an editor dialog). 
* and maybe, we're in an editor dialog (so isInline is undefined/false) BUT the field is empty?

4ian:
Merging this now as there are a lot of changes including "smaller" improvements like new "layout" components to standardise the spacing between fields and buttons.
I was able to get the new "filled" text fields to not have a space for the label when there is no label to show (for example, in a file/folder picker). This will make these text fields more visible on screen - hopefully making them clearer to new users :)

Issue #1294, opened 2019-11-18
Unable to quit GDevelop or open anything from toolbar if a project isn't open


Baylissimo:
After opening GDevelop and before opening or starting a new project any attempt to open the toolbar, like for example to close GDevelop from file, it just says Gdevelop still loading.

## To Reproduce
1. Go to 'toolbar' after opening GDevelop
2. All options say Gdevelop is still loading.

<img width="309" alt="Screen Shot 2019-11-18 at 11 00 55 am" src="https://user-images.githubusercontent.com/15341856/69016587-a6e5c380-09f3-11ea-91da-38c0a9270a33.png">

## Other details
* Using MacOS Catalina 10.15.1
* GDevelop beta 82

To access toolbar of course I can open or start a project then quit.

Phenomena3:
Happening Windows 8.1 laptop and some other machines too, I think.

4ian:
If you create/open a project, the menu is properly appearing then?

Baylissimo:
Yes. When creating/opening a project then the menu appears properly. And continues to appear properly after the opened project is closed and no other projects are open.

Bouh:
This is only on b82&b81.
Same for few others bug, B82 are pretty bugged in UI.

https://forum.gdevelop-app.com/t/menus-are-loaded-infinitely/21256
https://forum.gdevelop-app.com/t/expression-search-issue/21260
https://forum.gdevelop-app.com/t/user-interface-regression/21351

4ian:
I referenced these in https://github.com/4ian/GDevelop/issues/1295

Bouh:
insert [at line 131-132](https://github.com/4ian/GDevelop/blob/3bd9432b5e5140f73c29f2cd83f68d1bcb56b785/newIDE/app/src/MainFrame/ElectronMainMenu.js#L132)  
      `this._buildAndSendMenuTemplate();`
or 
`this.forceUpdate()`

4ian:
I'm surprised because that should in theory already work, because there is `componentDidUpdate` there:

https://github.com/4ian/GDevelop/blob/3bd9432b5e5140f73c29f2cd83f68d1bcb56b785/newIDE/app/src/MainFrame/ElectronMainMenu.js#L134-L139

[...]

PR #1301, opened 2019-11-28
Fix menu loading

Bouh:
#1294


4ian:
Thanks! 👍 Let's hope this will be enough :)

PR #1305, opened 2019-12-02
add reactjs debug tools

blurymind:
this adds reactjs debugger extension to electron, when running in dev mode..
Needs to be tested extensively still


4ian:
Great! :) 

- Having this only in development is fine. In production, the code is minimized and using the production build of React, so the debugger is less useful. And also, I think there is no real use case for having the debugger there.
- I think it would be great to have it - akin to the IDE external editors, zipped in the folder and have a script (ran in the electron-app/package.json postinstall) that unzip it. The reason is that I  wouldprefer as possible to avoid mixing in the repository the sources of GDevelop and the sources of external lib. Though it's done in some place like GDJS/Runtime that is containing Pixi and various libs, for the IDE better keep external stuff zipped. 

This would allow to gitignore the folder of the React devtools, so that they don't appear when you're searching in the sources in an IDE. It also avoid to clutter the search in GitHub with sources that are not part of GDevelop.

Note that this zip can be kept in the git repository (i.e: the zip can be committed as a source file). So it would be a matter of having a short js script that is running `unzipper` on the zip.

blurymind:
Thats a great idea :) So keep it zipped, but this time its kept in a folder instead of being fetched from releases?

4ian:
Yes exactly, that sounds simpler than fetching from a release, easier to upgrade and the whole React devtools zip should be fairly small. So it's both easy to maintain, no separate download and still not "polluting" the sources when browsing the sources/searching on GitHub on with an IDE :)

blurymind:
It just occured to me that it would be kind of awesome if we could trigger reload-extensions when the electron client is realoaded and in dev. I know how to capture the reload event, but not sure how to run the script when it happens

blurymind:
@4ian I pushed an update to this PR that makes the electron client automatically run the extension reloading script every time you ctrl+shift+r (reload). When working on extensions we will no longer need to run that in terminal, then reload electron. Just reload electron, done :)

blurymind:
I also moved the electron extension to its own zip file and created a script to unzpit it with the external-edtiors

newIDE/app/package.json
```diff
@@ -92,7 +92,9 @@
     "analyze-source-map": "source-map-explorer build/static/js/main.*",
     "extract-all-translations": "node scripts/extract-all-translations.js",
     "compile-translations": "node scripts/compile-translations.js",
-    "reload-extensions": "node scripts/import-GDJS-Runtime.js"
+    "reload-extensions": "node scripts/import-GDJS-Runtime.js",
+    "import-zipped-external-editors": "cd scripts && node import-zipped-editor.js piskel 5.0.0-beta82 e02f67866b338b4cc577e1ffa48f55f52cd520ed751dccc6f0931e3c1b995de0 && node import-zipped-editor.js jfxr 5.0.0-beta55 8ac12b557c2ddba958c6f0d3e0c5df8cf3369a65262dcb90cf5c8a7a7d20bdf6 && node import-zipped-editor.js yarn 5.0.0-beta80 5930c686ccf2b6fd3433b1c9539483473a6aedb8b6a73dd13c07030c95efecb9",
```
4ian: Good idea to have this in another script entry 👍

newIDE/app/package.json
```diff
@@ -92,7 +92,9 @@
...
+    "import-zipped-electron-extensions": "cd scripts && node import-zipped-file.js ReactDeveloperTools ../../electron-app/app/extensions/"
```
4ian: I wonder if this should be in `electron-app/package.json` instead. This is Electron specific, so as possible we should avoid polluting the generic app sources with installation of extensions?

newIDE/app/scripts/import-zipped-file.js
```diff
@@ -0,0 +1,42 @@
+/**
+ * This script will extract a zipped file of a prebuilt external extension to GD that we don't want added to git history
+ * The zip file contains the raw, unchanged sources, which will be extracted into a folder
+ * The zip should be kept in a folder somewhere within the project
+ */
+var shell = require('shelljs');
```
4ian: Similar as the previous comment, I think this file should be in newIDE/electron-app/scripts.

newIDE/electron-app/.gitignore
```diff
@@ -1,3 +1,4 @@
 dist/
 app/www/
 .env
+app/extensions/ReactDeveloperTools
```
4ian: super nit: configure your text editor to always [finish a file with a new line](https://thoughtbot.com/blog/no-newline-at-end-of-file).

newIDE/electron-app/app/main.js
```diff
@@ -129,6 +133,13 @@ app.on('ready', function() {
     }
   });
 
+  mainWindow.webContents.on('devtools-reload-page', () => {
```
4ian: I would rather avoid relying on devtools events, and also this is not using [`LocalGDJSDevelopmentWatcher` ](https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/GameEngineFinder/LocalGDJSDevelopmentWatcher.js) which is a bit sad because there is now two places where the editor is magically launch a npm script (there and in LocalGDJSDevelopmentWatcher - one using npm, the other using node).

Finally, I prefer to keep the sources of the electron-app as lean as possible. 

So I think we could instead of this addition modify `LocalGDJSDevelopmentWatcher` so that it's re-importing GDJS (and extensions) anytime it's being set up (i.e: when the app is launched in development mode).

For this, make a call to `importGDJSRuntime` (`importGDJSRuntime().catch(() => {});`) in:

https://github.com/4ian/GDevelop/blob/d0247c7df33e5a73049be738831ca6bfbe289a5d/newIDE/app/src/GameEngineFinder/LocalGDJSDevelopmentWatcher.js#L177-L184

Probably just after the "first return". React will run the code in "useEffect" when the array passed to useEffect (`[shouldWatch]`) is changed, i.e: when the preference is changed in the preferences dialog, or when the component is mounted (i.e: at the start of the app).

4ian: This will actually make the watcher more "robust", as this will guarantee that if you deactivated the watcher for a while, the sources are re-imported as soon as your re-activate it :)

newIDE/app/scripts/import-zipped-file.js
```diff
@@ -0,0 +1,42 @@
+/**
+ * This script will extract a zipped file of a prebuilt external extension to GD that we don't want added to git history
```
4ian: As you mentioned "external extension", your file name should reflect this by being named "import-zipped-electron-extension" or something like this - to state that it is *not* general purpose.

newIDE/app/scripts/import-zipped-file.js
```diff
@@ -0,0 +1,42 @@
...
+ * The zip file contains the raw, unchanged sources, which will be extracted into a folder
+ * The zip should be kept in a folder somewhere within the project
+ */
+var shell = require('shelljs');
+var fs = require('fs');
+var unzipper = require('unzipper');
+var process = require('process');
+var path = require('path');
+
+const editor = process.argv[2];
```
4ian: Careful about variable naming, this is an extension not an editor.

4ian:
See if you can move the script and call it from `postinstall` in electron-app/package.json and my other comments. Should be good to go after this - will be super useful when developing the IDE UI :)

newIDE/electron-app/app/main.js
blurymind: thank you, works like a charm :)

blurymind:
@4ian all review notes applied now :)

4ian:
Looks good, let's go ahead with this :) Thanks for working on this, will be very useful for development!

blurymind:
thank you for reviewing this :) I hope it will help others contributing to GD and make the process smoother. Really happy for removing the need to run the reload script when working on extensions

PR #1306, opened 2019-12-02
Improve comment

Bouh:
Here a other thing for :
- Comment looks like group in eventsheet
- Add a reset button for font in text object ([Discuss on forum](https://forum.gdevelop-app.com/t/pending-can-not-remove-custom-font-from-text-object-properties/21534))
New props canBeReset on ResourceSelector
- Add "Edit" in context menu on comment for edit all content and change background color (WIP)

I don't got the event back in the right way.
getComment() gives me something strange.  "\".

Height size of comment event are adjust at each input.

Oh i see that **modal** prop dosen't work at all, on all existing dialog, i've tried to remove it on other dialog but it's not working too.

I was on this code to learn a little more about React & eventsheet.
Now go back to Guideline :) 

image

image

### DIFF
--- a/newIDE/app/src/EventsSheet/EventsTree/Renderers/CommentEvent.js
+++ b/newIDE/app/src/EventsSheet/EventsTree/Renderers/CommentEvent.js
@@ -1,83 +1,79 @@
 // @flow
+import { t } from '@lingui/macro';
+
 import * as React from 'react';
-import ReactDOM from 'react-dom';
 import classNames from 'classnames';
-import { type EventRendererProps } from './EventRenderer';
-import { rgbToHex } from '../../../Utils/ColorTransformer';
+import TextField from '../../../UI/TextField';
 import {
   largeSelectedArea,
   largeSelectableArea,
   selectableArea,
+  disabledText,
 } from '../ClassNames';
+import { type EventRendererProps } from './EventRenderer';
 const gd = global.gd;
 
 const styles = {
   container: {
-    minHeight: 30,
     display: 'flex',
+    flexWrap: 'wrap',
+    padding: 5,
+    overflow: 'hidden',
     backgroundColor: '#fbf3d9',
   },
-  text: {
-    flex: 1,
-    whiteSpace: 'pre-line',
-    margin: 0,
-    padding: 5,
-  },
-  textArea: {
-    padding: 5,
-    flex: 1,
-    boxSizing: 'border-box',
+  textField: {
     width: '100%',
-    fontSize: 14,
+    fontSize: 18,
   },
 };
 
-type State = {|
-  editing: boolean,
-  height: number,
-|};
-
 export default class CommentEvent extends React.Component<
   EventRendererProps,
-  State
+  *
 > {
   state = {
     editing: false,
-    height: 0,
+    height: '100%',
   };
 
-  _container: ?any;
-  _input: ?any;
+  _selectable: ?any;
+  _textField: ?any;
 
   edit = () => {
-    if (!this._container) return;
-
     this.setState(
       {
         editing: true,
-        height: this._container.offsetHeight,
+        height: '100%',
       },
       () => {
-        // $FlowFixMe
-        const input: ?HTMLInputElement = ReactDOM.findDOMNode(this._input);
-        if (input) {
-          input.focus();
-          input.value = gd.asCommentEvent(this.props.event).getComment();
-        }
+        if (this._textField) this._textField.focus();
+        this.forceUpdate();
+      }
+    );
+  };
+
+  onEvent = (e: any, text: string) => {
+    const commentEvent = gd.asCommentEvent(this.props.event);
+    commentEvent.setComment(text);
+
+    this.setState(
+      {
+        height: '100%',
+      },
+      () => {
+        this.props.onUpdate();
+        this.forceUpdate();
       }
     );
   };
 
   endEditing = () => {
-    const commentEvent = gd.asCommentEvent(this.props.event);
-
-    // $FlowFixMe
-    const input: ?HTMLInputElement = ReactDOM.findDOMNode(this._input);
-    if (input) commentEvent.setComment(input.value);
+    if (!this._textField) return;
 
     this.setState(
       {
         editing: false,
+        height: '100%',
       },
       () => this.props.onUpdate()
     );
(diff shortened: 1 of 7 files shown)


newIDE/app/src/EventsSheet/EventsTree/Renderers/CommentEvent.js
```diff
@@ -94,48 +90,73 @@ export default class CommentEvent extends React.Component<
...
-      commentEvent.getTextColorGreen(),
-      commentEvent.getTextColorBlue()
-    );
+    var commentEvent = gd.asCommentEvent(this.props.event);
+
+    const r = commentEvent.getBackgroundColorRed(),
+      g = commentEvent.getBackgroundColorGreen(),
+      b = commentEvent.getBackgroundColorBlue();
+
+    const textColor = (r + g + b) / 3 > 200 ? 'black' : 'black';
```
4ian: One should be white no?

newIDE/app/src/EventsSheet/EventsTree/Renderers/CommentEvent.js
```diff
@@ -94,48 +90,73 @@ export default class CommentEvent extends React.Component<
...
+              inputStyle={{
+                color: textColor,
+                WebkitTextFillColor: textColor,
+                padding: 0,
+              }}
+              underlineFocusStyle={{
+                borderColor: textColor,
+              }}
+              fullWidth
+              id="group-title"
```
4ian: id should be changed

newIDE/app/src/EventsSheet/EventsTree/Renderers/CommentEvent.js
```diff
@@ -1,83 +1,79 @@
...
+    backgroundColor: '#fbf3d9',
   },
-  textArea: {
-    padding: 5,
-    flex: 1,
-    boxSizing: 'border-box',
+  textField: {
     width: '100%',
-    fontSize: 14,
+    fontSize: 18,
```
4ian: I think comments are too big. I kind of liked the fact that they are smaller than groups, because groups are like a Header in a text, while comment is text:

## This is a group 

This is a comment blablabla

newIDE/app/src/EventsSheet/EventsTree/Renderers/CommentEvent.js
```diff
@@ -1,83 +1,79 @@
...
-    backgroundColor: '#fbf3d9',
-  },
-  text: {
-    flex: 1,
-    whiteSpace: 'pre-line',
-    margin: 0,
+    flexWrap: 'wrap',
     padding: 5,
+    overflow: 'hidden',
+    backgroundColor: '#fbf3d9',
```
4ian: What is this hardcoded color?

newIDE/app/src/EventsSheet/EventsTree/style.css
```diff
@@ -6,11 +6,14 @@
 .gd-events-sheet .rst__moveHandle {
     width: 10px;
     background: #4ab0e4;
-    border: 1px solid #49AFE4;
     box-shadow: none;
     border-radius: 0;
 }
 
+.gd-events-sheet .rst__moveHandle:hover {
+    background-color: #767676!important;
```
4ian: Colors should be in the theme only, and probably different for both themes?

4ian:
> Comment looks like group in eventsheet

See my comment about comments (😉), I think they should be smaller than groups like they were before.

> Add "Edit" in context menu on comment for edit all content and change background color (WIP)

This is more tricky, you need to store in the state of the events sheet the event being edited, and pass it as a prop to your dialog. Don't try to pass the selection.

> getComment() gives me something strange. "\��".

This is basically almost a crash in libGD.js because you tried to cast something that is not a proper gd.BaseEvent into a gd.CommentEvent
You should not need this and pass directly the commentEvent as a prop to your dialog.

newIDE/app/src/EventsSheet/InstructionEditor/TextEditorDialog.js
```diff
@@ -0,0 +1,172 @@
...
+//Voir comment recup l'event car ma façon de faire dans le state récupère ce text "\
"
+export default class TextEditorDialog extends React.PureComponent<
+  Props,
+  State
+> {
+  recupereLecommentaire = () => {
+    return gd
+      .asCommentEvent(
+        getSelectedEvents(this.props.event).some(event => {
```
4ian: Don't use `getSelectedEvents` inside TextEditorDialog, it's only to be used by the EventsSheet to manager its selection. Instead pass the event as a prop to TextEditorDialog

newIDE/app/src/EventsSheet/index.js
```diff
@@ -363,6 +368,20 @@ export default class EventsSheet extends React.Component<Props, State> {
     return newEvents;
   };
 
+  //TODO ideally see for use toggle
+  openTextEditorDialog = () => {
```
4ian: There you should probably get the selected event, store it in the state, and then use this state to pass it as a prop to the text editor dialog.

```
const selectedEvents = getSelectedEvents(this.state.selection);
if (!selectedEvents.length) return;
const event = selectedEvents[selectedEvents.length - 1]; // Get the last selected event.
this.setState({
  textEditedEvent: event;
});
```

You can see that a "bonus", we don't need the state "textEditorDialog". The fact that `textEditedEvent` is not null is enough for us to know that we must show the dialog :)

newIDE/app/src/EventsSheet/index.js
```diff
@@ -363,6 +368,20 @@ export default class EventsSheet extends React.Component<Props, State> {
...
+    this.setState({
+      textEditorDialog: true,
+    });
+  };
+
+  //TODO close and save in history
+  closeTextEditorDialog = () => {
+    this.setState({
+      textEditorDialog: false,
```
4ian: As per my previous comment, this can be instead `textEditedEvent: null`. Conceptually, this mean "We don't have anymore an event that we're editing text for.".

newIDE/app/src/EventsSheet/index.js
```diff
@@ -1268,6 +1308,23 @@ export default class EventsSheet extends React.Component<Props, State> {
                             }}
                           />
                         )}
+                        {this.state.textEditorDialog && (
```
4ian: Finally (if you read my previous comments), you can use `textEditedEvent` here:

```
{this.state.textEditedEvent && (
  <TextEditorDialog
    event={this.state.textEditedEvent}
    onApply={(changes) => {
      // Use the changes here and apply them to the event (this.state.textEditedEvent) - this will depend on the events type.
      // Then close the dialog:
      this.closeTextEditorDialog();
      // Surely call the method to update the history and refresh the events sheet.
    }}
    onClose={this.closeTextEditorDialog}
  />
)
```

newIDE/app/src/EventsSheet/InstructionEditor/TextEditorDialog.js
4ian: If you read my other comments, once you've passed an `event` as a prop (a real event, not the selection!), then you can do something like:

```
const { event } = this.props;
const eventType = event.getType();
if (eventType === 'BuiltinCommonInstructions::Comment') {
  return gd.asCommentEvent(event).getComment();
} else if (eventType === 'BuiltinCommonInstructions::Group') {
  return gd.asGroupEvent(event).getName();
} else {
  console.error("Dialog was opened for an unsupported event type: " + eventType);
  return '';
}
```

newIDE/app/src/EventsSheet/InstructionEditor/TextEditorDialog.js
```diff
@@ -0,0 +1,172 @@
...
+      )
+      .getComment();
+  };
+
+  state = {
+    searchText: '',
+    textValue: this.recupereLecommentaire(),
+  };
+
+  componentWillReceiveProps(newProps: Props) {
```
4ian: I think you can entirely avoid this method here.

newIDE/app/src/EventsSheet/InstructionEditor/TextEditorDialog.js
```diff
@@ -0,0 +1,172 @@
...
+        getSelectedEvents(this.props.event).some(event => {
+          return event;
+        })
+      )
+      .getComment();
+  };
+
+  state = {
+    searchText: '',
+    textValue: this.recupereLecommentaire(),
```
4ian: Basically, the idea of this component should be:

1. When creating the state, you read all the information that you can from the event passed as prop. You should do this according to the event type (see my other comment). If it's too much work, you can put this in a method as you already did ("recupereLecommentaire").
2. Then, the dialog is ONLY modifying the state (using classic React.setState) to change the text or the color.
3. If we click on cancel, call the onCancel prop and close the window (you close the window by setting the edited event to `null`). (see the idea?)
4. If we click on apply, call the onApply prop but before calling it, read the color and the text that are stroed in the state and apply them on the event passed as prop (so, according to the event type, you call setName or setComment, and setBackgroundColor, etc...). You probably can make a function inside TextEditorDialog for this.

This should limit the logic to do in EventsSheet to only dealing with the textEditedEvent - which is very clean.

newIDE/app/src/EventsSheet/index.js
4ian: (updated my previous comment. TextEditorDialog should take care of updating the event BEFORE calling onApply).

4ian:
Two things:

- I've written a few comments about how to deal with the event in the TextEditorDialog. In a nutshell, you should store the event that is having its text edited in the EventsSheet. You store it in the state, and this is all you need to know to open the dialog :)
- Ideally the commits could have been done in different PRs because it's 3 different things to review so a bit of a mess. See if you can extract commits in other branches and make new PRs maybe? Up to you but can speed up review and eventually the merge.

4ian:
Last thing, if you want to actually split this PR into 3, you can do like this (or the equivalent with a git GUI):

* go back to master: `git checkout master`
* Create a new branch: `git checkout -b my-feature-1`
* Import a single commit using its hash (that you can [find here](https://github.com/4ian/GDevelop/pull/1306/commits)): `git cherry-pick <hash of the commit>`
* That's it you have a branch with the commit :) Push your branch, go back to master and repeat for other commits.

Advantage is that I'll be soon merging https://github.com/4ian/GDevelop/pull/1293, and this can create a conflict with one commit. Better have a conflict on a single PR rather than impacting all of these commits.

newIDE/app/src/EventsSheet/EventsTree/Renderers/CommentEvent.js
Bouh: I use this trick actually comments are yellow, and white on yellow i see nothing.
I will use a other way to do.
Fixed

newIDE/app/src/EventsSheet/EventsTree/Renderers/CommentEvent.js
Bouh: Oops Fixed

newIDE/app/src/EventsSheet/EventsTree/Renderers/CommentEvent.js
Bouh: Fixed

newIDE/app/src/EventsSheet/EventsTree/style.css
Bouh: Fixed

newIDE/app/src/EventsSheet/index.js
Bouh: This way of thinking is fantastic. I'll keep in my mind.
Beautiful idea!
Fixed

newIDE/app/src/EventsSheet/index.js
Bouh: Fixed

newIDE/app/src/EventsSheet/InstructionEditor/TextEditorDialog.js
Bouh: Fixed

Bouh:
It's a mess to split everything in 3 PR.
Some line of code for style commentEvent are in same commits as TextEditorDialog.

I've extract reset button for resourceSelector in a other branch it's ok.

For the rest I'm lost in the changes present in master, what was there before, and what I edited.

Bouh:
Ok i guess it's right.
#1310 TextEditorDialog
#1309 CommentEvent style
#1308 Reset button on resourceSelector

4ian:
Great! Looks like you've managed to make 3 PRs :) I think you can close this one now.

PR #1307, opened 2019-12-02
GDJS/Extensions changes watcher for the IDE

4ian:
Add a watcher for changes in GDJS or extensions runtime, when the IDE is opened :) Should make working on extensions or the game engine faster.


4ian:
Tested on Windows and macOS :) Should work the same in Linux, though watching files is famous for being a hard problem and not always reliable. Always watch the console to see if files were properly imported.

PR #1308, opened 2019-12-03
Add reset button on resourceSelector

Bouh:
Can be added with new prop 'canbeReset' on <ResourceSelector>

### DIFF
--- a/newIDE/app/src/ResourcesList/ResourceSelector.js
+++ b/newIDE/app/src/ResourcesList/ResourceSelector.js
@@ -1,5 +1,4 @@
 // @flow
-import { Trans } from '@lingui/macro';
 
 import * as React from 'react';
 import IconButton from '../UI/IconButton';
@@ -7,7 +6,6 @@ import SemiControlledAutoComplete, {
   type DataSource,
 } from '../UI/SemiControlledAutoComplete';
 import ElementWithMenu from '../UI/Menu/ElementWithMenu';
-import FlatButton from '../UI/FlatButton';
 import { Line } from '../UI/Grid';
 import BackspaceIcon from '@material-ui/icons/Backspace';
 import Add from '@material-ui/icons/Add';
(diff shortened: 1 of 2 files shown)


newIDE/app/src/ResourcesList/ResourceSelector.js
```diff
@@ -240,33 +266,44 @@ export default class ResourceSelector extends React.Component<Props, State> {
...
+            value={this.state.resourceName}
+            onChange={this._onChangeResourceName}
+            errorText={errorText}
+            fullWidth={this.props.fullWidth}
+            margin={this.props.margin}
+            ref={autoComplete => (this._autoComplete = autoComplete)}
           />
-        )}
+          {this.props.canBeReset && (
+            <FlatButton
```
4ian: Could we try without a button label?
You can use `<IconButton>` for that.

newIDE/app/src/ResourcesList/ResourceSelector.js
```diff
@@ -42,6 +48,10 @@ const styles = {
 };
 
 export default class ResourceSelector extends React.Component<Props, State> {
+  static defaultProps = {
```
4ian: Avoid defaultProps, they make things more complicated from my experience. If the prop is not passed, it will be undefined.

newIDE/app/src/ResourcesList/ResourceSelector.js
```diff
@@ -240,33 +266,44 @@ export default class ResourceSelector extends React.Component<Props, State> {
...
-                <Brush />
-              </IconButton>
-            }
-            buildMenuTemplate={() =>
-              externalEditors.map(externalEditor => ({
-                label: externalEditor.displayName,
-                click: () => this._editWith(externalEditor),
-              }))
-            }
+        <Line expand>
```
4ian: `<Line nomargin expand>` as we already have a container for this.

4ian:
See if you can use an IconButton. If you can do this quickly, I can merge this before https://github.com/4ian/GDevelop/pull/1293 - otherwise I'll merge the other before this PR but there will be adaption to be done because I've introduced a new layout component for automatically putting buttons next to text fields/autocomplete with the proper margins.

Bouh:
Done ✅

4ian:
Looks sensible :) Merging! Thanks for adding this :)

PR #1309, opened 2019-12-03
CommentEvent change style

Bouh:

PR #1310, opened 2019-12-03
New textEditorDialog

Bouh:


newIDE/app/src/EventsSheet/InstructionEditor/TextEditorDialog.js
```diff
@@ -0,0 +1,301 @@
...
+  sizeTextField: {
+    width: 90,
+    ...toolbarItemStyle,
+  },
+  toolbarItem: toolbarItemStyle,
+  checkbox: toolbarItemStyle,
+};
+
+type Props = {|
+  event: any,
```
4ian: event: gdBaseEvent

(even if it's currently almost equivalent to any, in the future GDevelop.js could autogenerate these types)

newIDE/app/src/EventsSheet/InstructionEditor/TextEditorDialog.js
```diff
@@ -0,0 +1,301 @@
...
+  onClose: () => void,
+  onApply: () => void,
+|};
+
+type State = {|
+  textValue: any,
```
4ian: put proper types there :) This will save us from silly mistakes.

textValue is a `string`
textColors and backgroundColors could be objects like this: `{|r:number, g: number, b: number|}`

(the burger notation `{| |}` indicates that the object can not have other properties than the one listed)

newIDE/app/src/EventsSheet/InstructionEditor/TextEditorDialog.js
```diff
@@ -0,0 +1,301 @@
...
+    }
+  };
+
+  setComment = () => {
+    const { event } = this.props;
+    const eventType = event.getType();
+    if (eventType === 'BuiltinCommonInstructions::Comment') {
+      gd.asCommentEvent(event).setComment(this.state.textValue);
+    }
+  };
```
4ian: Could you group these two methods in a single one called `_applyChangesOnEvent`?

newIDE/app/src/EventsSheet/InstructionEditor/TextEditorDialog.js
```diff
@@ -0,0 +1,301 @@
...
+        parseInt(backgroundColors.blue)
+      );
+    } else if (eventType === 'BuiltinCommonInstructions::Group') {
+      gd.asGroupEvent(event).setBackgroundColor(
+        parseInt(backgroundColors.red),
+        parseInt(backgroundColors.green),
+        parseInt(backgroundColors.blue)
+      );
+    }
+  };
```
4ian: I think we could also make it part of `_applyChangesToEvent`

newIDE/app/src/EventsSheet/InstructionEditor/TextEditorDialog.js
```diff
@@ -0,0 +1,301 @@
...
+      };
+
+      return backgroundColors;
+    } else {
+      console.error(
+        'Dialog was opened for an unsupported event background color'
+      );
+      return '';
+    }
+  };
```
4ian: And these 3 methods could be a single `_getInitialStateFromEvent`.
I think this would reduce some boilerplate and the number of if/else to do (once instead of 3 times).

The object you should return is the `State`: `state = this._getInitialStateFromEvent()`

newIDE/app/src/EventsSheet/InstructionEditor/TextEditorDialog.js
```diff
@@ -0,0 +1,301 @@
...
+    backgroundColors: this.getBackgroundColors(),
+  };
+
+  render() {
+    const { event, onApply, onClose } = this.props;
+    const eventType = event.getType();
+
+    return (
+      <Dialog
+        title={<Trans>Text editor</Trans>}
```
4ian: A bit skeptical about this, the text editor might be confused with the editor of a text object... maybe "Edit the event text"?

newIDE/app/src/EventsSheet/InstructionEditor/TextEditorDialog.js
```diff
@@ -0,0 +1,301 @@
...
+            primary
+            keyboardFocused
+            onClick={() => {
+              if (eventType === 'BuiltinCommonInstructions::Comment') {
+                this.setComment();
+              } else if (eventType === 'BuiltinCommonInstructions::Group') {
+                this.setName();
+              }
+              this.setTextColors();
+              this.setBackgroundColors();
```
4ian: See my previous comment, ideally we do this as a single operation in a function, could make things a bit clearer.

newIDE/app/src/EventsSheet/InstructionEditor/TextEditorDialog.js
```diff
@@ -0,0 +1,301 @@
...
+            </MiniToolbarText>
+            <ColorPicker
+              style={styles.sizeTextField}
+              disableAlpha
+              color={{
+                r: this.state.backgroundColors.red,
+                g: this.state.backgroundColors.green,
+                b: this.state.backgroundColors.blue,
+                a: 255,
+              }}
```
4ian: Ideally, we could store in the state objects with the same shape as the ones passed to the ColorPicker:

colors={this.state.backgroundColor}

newIDE/app/src/EventsSheet/InstructionEditor/TextEditorDialog.js
```diff
@@ -0,0 +1,301 @@
...
+              onChangeComplete={color => {
+                this.setState({
+                  backgroundColors: {
+                    red: color.rgb.r,
+                    green: color.rgb.g,
+                    blue: color.rgb.b,
+                  },
```
4ian: `...and here we could do this.setState({ backgroudColor: color })`

if we use the same shape for all colors.

newIDE/app/src/EventsSheet/InstructionEditor/TextEditorDialog.js
```diff
@@ -0,0 +1,301 @@
...
+                });
+
+                this.forceUpdate();
+              }}
+            />
+
+            {eventType !== 'BuiltinCommonInstructions::Group' && (
```
4ian: Instead of excluding an event type, prefer to test for the event type that is the one that need the components (i.e: the comment).
Otherwise if for some reason a bad event is passed here, the component will be shown and could make things worse for the user.

newIDE/app/src/EventsSheet/InstructionEditor/TextEditorDialog.js
```diff
@@ -0,0 +1,301 @@
...
+                    a: 255,
+                  }}
+                  onChangeComplete={color => {
+                    this.setState({
+                      textColors: {
+                        red: color.rgb.r,
+                        green: color.rgb.g,
+                        blue: color.rgb.b,
+                      },
+                    });
```
4ian: Same remarks relative to simplifying color objects

newIDE/app/src/EventsSheet/InstructionEditor/TextEditorDialog.js
```diff
@@ -0,0 +1,301 @@
...
+
+                    this.forceUpdate();
```
4ian: forceUpdate is not needed after a setState

newIDE/app/src/EventsSheet/InstructionEditor/TextEditorDialog.js
```diff
@@ -0,0 +1,301 @@
...
+          <Line noMargin>
+            <Column expand>
+              <Line>
+                <SemiControlledTextField
+                  commitOnBlur
+                  hintText={t`Enter the text to be displayed`}
+                  fullWidth
+                  multiLine
+                  rows={100}
+                  rowsMax={100}
```
4ian: That's too much! :) Let's use 8 as in the text editor.

newIDE/app/src/EventsSheet/InstructionEditor/TextEditorDialog.js
```diff
@@ -0,0 +1,301 @@
...
+                  value={this.state.textValue}
+                  onChange={value => {
+                    this.setState({
+                      textValue: value,
+                    });
+                    this.forceUpdate();
```
4ian: no forceUpdate

newIDE/app/src/EventsSheet/InstructionEditor/TextEditorDialog.js
```diff
@@ -0,0 +1,301 @@
...
+                  hintText={t`Enter the text to be displayed`}
+                  fullWidth
+                  multiLine
+                  rows={100}
+                  rowsMax={100}
+                  value={this.state.textValue}
+                  onChange={value => {
+                    this.setState({
+                      textValue: value,
+                    });
```
4ian: nitPicking: 
```
onChange={textValue =>
                    this.setState({
                      textValue
                    })
```

newIDE/app/src/EventsSheet/index.js
```diff
@@ -133,6 +134,8 @@ type State = {|
 
   serializedEventsToExtract: ?Object,
 
+  textEditedEvent: ?any,
```
4ian: ?gdBaseEvent

newIDE/app/src/UI/TextField.js
```diff
@@ -75,7 +75,7 @@ type Props = {|
   margin?: 'none' | 'normal',
   fullWidth?: boolean,
   style?: {|
-    fontSize?: 18,
+    fontSize?: number | string,
```
4ian: Might not be needed in this PR?

Wend1go:
Do you intend to add buttons for setting the outline and text shadow to the text editor dialog?

4ian:
@Wend1go this is for editing the color and text of comments and group events ;)

@Bouh I think Wend1go comment is revealing that the name of the dialog is not intuitive. This should be related to events, something like "EventTextDialog".

newIDE/app/src/EventsSheet/InstructionEditor/TextEditorDialog.js
Bouh: I didn't understand return here

newIDE/app/src/EventsSheet/InstructionEditor/TextEditorDialog.js
Bouh: I put "any" by default everywhere except when I know the type ;)
I was not sure if an object was a string or other type, thx !

newIDE/app/src/EventsSheet/InstructionEditor/TextEditorDialog.js
4ian: This return (`return '';`) is incorrect: it's returning a string, while the rest of the function is returning an object. You should then return an object with the same shape (same "type") as the rest of the function.

In other words, surely something like this: `return {r: 0, g: 0, b: 0}; // Return black in case of error.` - though in theory this should never ever be used (in practice, better make the code robust again something passing a wrong event type to this dialog).

Bouh:
> Do you intend to add buttons for setting the outline and text shadow to the text editor dialog?

Not in this PR, this is on my todo list but I've too much todo before, if you give a try before me :) 

> I think Wend1go comment is revealing that the name of the dialog is not intuitive. This should be related to events, something like "EventTextDialog".

Ahah yes i will rename it !


I need your feddback for **this.state.backgroundColor**, I'am lost inside type of colors for set state.
And i can't type directly RGBColor ?

newIDE/app/src/EventsSheet/InstructionEditor/TextEditorDialog.js
Bouh: I set 100 for having a large window, because if you put more text in the window, textfield doesn't get bigger because limited by 8 rows.
We need to reopen the editor.

There is also a problem with <minitoolbar> and <dialog>.
Minitoolbar should not be included in the scroll of the dialog.
Maybe we need to have a new props "header" for the dialog. And we pass to header the minitoolbar. 

This is also anoying in for "edit hitboxes" "edit points" "add an animation+"

4ian: > Minitoolbar should not be included in the scroll of the dialog.
> Maybe we need to have a new props "header" for the dialog. And we pass to header the minitoolbar.

Yes, I think this could maybe be solved on a case by case basis with the `<ScrollView>` component too (I'm a bit worried of modifying Dialog, I like to keep it simple) (though it's always possible to also make a component DialogWithHeader that uses ScrollView for the middle part). There is something to do.

> I set 100 for having a large window, because if you put more text in the window, textfield doesn't get bigger because limited by 8 rows.
We need to reopen the editor.

Can you set rows={a small number} and rowsMax={a higher number, like 30}? So that the field is taking more space but not too much.

Bouh:
Well i didn't understand what wrong, it's like state are not usable in render().
But the value are define at the end of _getInitialStateFromEvent().
And the render say your state is null ...

Bouh:
Omg forget it, I found why >.>

Bouh:
Is working, but i'm unable to fix errors from flow...

newIDE/app/src/EventsSheet/InstructionEditor/EventTextDialog.js
```diff
@@ -0,0 +1,240 @@
...
+  },
+  toolbarItem: toolbarItemStyle,
+  checkbox: toolbarItemStyle,
+};
+type RGBColor = {|
+  r?: number,
+  g?: number,
+  b?: number,
+  a?: number,
+|};
```
4ian: Not too bad, but you probably don't want to mark r/g/b/a as optional. When doing this, then this means that you consider that these properties can be missing. It's a pain because that mean that you have to check everywhere if r/g/b and a are not undefined before using them! Instead, just say that a RGBColor is *always* containing these variables:

```
type RGBColor = {|
  r: number,
  g: number,
  b: number,
  a: number,
|};
```

newIDE/app/src/EventsSheet/InstructionEditor/EventTextDialog.js
```diff
@@ -0,0 +1,240 @@
...
+
+      //Text color for group not supported in Core, instead GroupEvent.js handle this
+      //Background color
+      gd.asGroupEvent(event).setBackgroundColor(
+        parseInt(backgroundColor.r),
+        parseInt(backgroundColor.g),
+        parseInt(backgroundColor.b)
+      );
+      return;
+    }
```
4ian: Just by removing the optional properties in RGBColor type (see my previous comment), this should solve all the flow errors that you had in this function.

Why? Because you are using the r/g/b properties directly, so if these properties are marked as optional, Flow is warning you: "You're passing a varaible that is maybe undefined to parseInt, which expect a number! Beware there is a bug hidden here!!".

If properties are not marked as optional, Flow will understand that you're passing something that is guaranteed to be a number to parseInt - si it's fine.

newIDE/app/src/EventsSheet/InstructionEditor/EventTextDialog.js
```diff
@@ -0,0 +1,240 @@
...
+    const { textValue, textColor, backgroundColor } = this.state;
+    const eventType = event.getType();
+
+    if (eventType === 'BuiltinCommonInstructions::Comment') {
+      //Text value
+      gd.asCommentEvent(event).setComment(textValue);
+
+      //Text color
+      gd.asCommentEvent(event).setTextColor(
+        parseInt(textColor.r),
```
4ian: Always always always specify the second argument to parseInt: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt

```suggestion
        parseInt(textColor.r, 10),
```

newIDE/app/src/EventsSheet/InstructionEditor/EventTextDialog.js
```diff
@@ -0,0 +1,240 @@
...
+      gd.asGroupEvent(event).setBackgroundColor(
+        parseInt(backgroundColor.r),
+        parseInt(backgroundColor.g),
+        parseInt(backgroundColor.b)
+      );
+      return;
+    }
+  };
+
+  _getInitialStateFromEvent = () => {
```
4ian: To help you, define the type of what your function is returning:
```suggestion
  _getInitialStateFromEvent = (): State => {
```

This will limit errors and the amount of guessing that Flow has to do, making easier to understand error messages.

newIDE/app/src/EventsSheet/InstructionEditor/EventTextDialog.js
```diff
@@ -0,0 +1,240 @@
...
+    const { event } = this.props;
+    const eventType = event.getType();
+    let text: string, textColors: RGBColor, backgroundColors: RGBColor;
```
4ian: This is kind of "false" as you're declaring variables to be of a certain type, but they are actually undefined :) 
So either you mark them as optional (but don't do that) or (better) you initialize them to default values:

```
let text: string = "";
let textColors: RGBColor = {r:0, g:0,b:0, a:0};
let backgroundColors: RGBColor = {r:0, g:0,b:0, a:0};
```

Otherwise, Flow will warn you: you might be using a variable that is uninitialized! (i.e.: that is undefined).
image

newIDE/app/src/EventsSheet/InstructionEditor/EventTextDialog.js
```diff
@@ -0,0 +1,240 @@
...
+      console.error(
+        'Dialog was opened for an unsupported event type: ' + eventType
+      );
+    }
+
+    return {
+      textValue: text,
+      textColor: textColors,
+      backgroundColor: backgroundColors,
+    };
```
4ian: Rule of thumb: if you assign "something" to "something else", they represent the same thing. So give them the same name :) 

```
     return {
      textValue: textValue,
      textColor: textColor,
      backgroundColor: backgroundColor,
    };
```

which can be simplified to:

```
     return {
      textValue,
      textColor,
      backgroundColor,
    };
```

newIDE/app/src/EventsSheet/InstructionEditor/EventTextDialog.js
4ian: See my next comment about rename variables too to give them the same name as in the State.

newIDE/app/src/EventsSheet/InstructionEditor/EventTextDialog.js
```diff
@@ -0,0 +1,240 @@
...
+        g: groupEvent.getBackgroundColorG(),
+        b: groupEvent.getBackgroundColorB(),
+      };
+
+      text = gd.asGroupEvent(event).getName();
+    } else {
+      console.error(
+        'Dialog was opened for an unsupported event type: ' + eventType
+      );
+    }
```
4ian: Flow will not let you make risky things: it has seen that you declared text, textColors, backgroundColors but have not set them a value there. Always make sure that whatever the "branch" of the if/else is being used, your variables are initialised.

This is what I explained in a previous comment. Note that you could also set the variables in the "else" (`text = ''; textColor = { ... }`) but I prefer to initialise your variables as soon as possible.

newIDE/app/src/EventsSheet/InstructionEditor/EventTextDialog.js
4ian: Also you might want to remove the "a", because you're not setting it anywhere (and Flow will warn you about not setting it if it's not optional).

newIDE/app/src/EventsSheet/index.js
```diff
@@ -1305,6 +1347,20 @@ export default class EventsSheet extends React.Component<Props, State> {
...
                           />
                         )}
+                        {this.state.textEditedEvent && (
+                          <EventTextDialog
+                            event={this.state.textEditedEvent}
+                            onApply={() => {
+                              this.closeEventTextDialog();
+                              this._saveChangesToHistory();
+                            }}
+                            onClose={() =>
```
4ian: `onClose={this.closeEventTextDialog}`

4ian: Which you can read as English: "When [the user wants to] Close, then close the event text dialog" :)

newIDE/app/src/EventsSheet/InstructionEditor/EventTextDialog.js
4ian: Actually, as you've asked:

> And i can't directly use type RGBColor ?

I guess you refer to the one in ColorPicker. You can use it if you:
* Add the "export" keyword before "type RGBColor" in ColorPicker
* Then import it: `import ColorPicker, { type RGBColor } from '../../UI/ColorField/ColorPicker';`

I recommend that you do this :) Better have a single place where a type is defined. This type has an extra optional `a` property, but that's fine!

newIDE/app/src/EventsSheet/InstructionEditor/EventTextDialog.js
Bouh: About parseInt(), we can ignore it here ?
backgroundColor is define with getBackgroundColorXXX() and this return always a number. And ColorPicker return number too.

Unless it is necessary to do this check because the values of the.json file can be edited and falsified.

I'm wondering when it's judicious to check the type and format of the data when the methods used do already a checking.
Any way I keep it, it dosen't cost much.

4ian: Wait a second you're right we should always have number (if we have strings there, it's a big problem somewhere).
So you can remove all the parseInt entirely: `gd.asCommentEvent(event).setTextColor(textColor.r, textColor.g, textColor.b)` :)

> Unless it is necessary to do this check because the values of the.json file can be edited and falsified.

No you can remove these parseInt, even if you go and replace them by a string in the JSON, it's being read by libGD.js which will probably read it as 0 or something like this. 
In either case, methods from libGD.js being in C++, they are guaranteed to always return a number if they declare to return a float/double/int in `Bindings.idl`. (and they also expect to receive arguments with the exact type declared in `Bindings.idl`, so if for some reason a string is sent to a function expecting a number, very bad things can happen).

> I'm wondering when it's judicious to check the type and format of the data when the methods used do already a checking.

The idea with Flow is to get the guarantee of the type and format of data. A bit more painful to write at first, but it's actually leading to cleaner code because you don't need extra check/extra conversion/extra handling of undefined values :)

newIDE/app/src/EventsSheet/InstructionEditor/EventTextDialog.js
```diff
@@ -0,0 +1,226 @@
...
+  marginRight: 10,
+};
+
+const styles = {
+  sizeTextField: {
+    width: 90,
+    ...toolbarItemStyle,
+  },
+  toolbarItem: toolbarItemStyle,
+  checkbox: toolbarItemStyle,
```
4ian: Seems unused

4ian: Seems unused

newIDE/app/src/EventsSheet/InstructionEditor/EventTextDialog.js
```diff
@@ -0,0 +1,226 @@
...
+const toolbarItemStyle = {
+  marginRight: 10,
+};
+
+const styles = {
+  sizeTextField: {
+    width: 90,
+    ...toolbarItemStyle,
+  },
+  toolbarItem: toolbarItemStyle,
```
4ian: Seems unused

newIDE/app/src/EventsSheet/index.js
```diff
@@ -1105,85 +1126,106 @@ export default class EventsSheet extends React.Component<Props, State> {
...
+                                click: () => this.moveEventsIntoNewGroup(),
+                              },
+                              {
+                                label: 'Analyze Objects Used in this Event',
+                                click: this._openEventsContextAnalyzer,
+                              },
+                            ];
+
+                            const edition = {
+                              label: 'Edit',
```
4ian: Do you think we should be more precise "Edit Text and Color(s)"?

newIDE/app/src/EventsSheet/index.js
```diff
@@ -1105,85 +1126,106 @@ export default class EventsSheet extends React.Component<Props, State> {
...
+
+                            getSelectedEvents(this.state.selection).forEach(
+                              event => {
+                                if (
+                                  event.getType() ===
+                                    'BuiltinCommonInstructions::Comment' ||
+                                  event.getType() ===
+                                    'BuiltinCommonInstructions::Group'
+                                ) {
+                                  contextList.unshift(edition);
```
4ian: Rather than unshift-ing, let's try to keep a single array (so that it's more readable - I prefer to avoid "complex logic" in the render of components) (and also, this `render` is super long already).
Use `visible: ` set to true or false (next to label, and click) when defining the menu item.

Also careful about the forEach and unshift here, if somehow you have multiple comment or group selected, you'll have an "Edit" menu item inserted for each of them - which is probably not what you want ;)

Instead make a function for it to pass to `visible`:

`visible: filterEditableWithEventTextDialog(getSelectedEvents(this.state.selection)).length > 0`

and make `filterEditableWithEventTextDialog` a function:

```
const filterEditableWithEventTextDialog = (events: Array<gdBaseEvent>) => {
  return events.filter(event => 
    ['BuiltinCommonInstructions::Group', 'BuiltinCommonInstructions::Comment'].includes(event.getType())
  );
}
```

(not tested but you get the idea).
Last thing, where to put this function? It contains knowledge related to `EventTextDialog` so `EventTextDialog.js` file is a good place. Put it here, export it (`export const canBeEditedWithEventTextDialog = ...`).

And finally import it in EventsSheet.

This will allow to have just an array for the menu items in EventsSheet, and the "complex logic" colocated with the rest of the source code of EventTextDialog. It's better because now you only have a single file that contains all the rules telling which events are supported by EventTextDialog - rather than two (so that you have less change to forget to update one if you add a new event in the future).

newIDE/app/src/UI/Theme/DarkTheme/EventsSheet.css
```diff
@@ -26,7 +26,11 @@
  */
 .gd-events-sheet-dark-theme .rst__moveHandle {
   background: #494949 !important;
-  border: 1px solid #494949 !important;
+  border: none;
```
4ian: Why the change the border again? Forgot what's the purpose

newIDE/app/src/EventsSheet/index.js
```diff
@@ -362,6 +367,22 @@ export default class EventsSheet extends React.Component<Props, State> {
     return newEvents;
   };
 
+  openEventTextDialog = () => {
+    const selectedEvents = getSelectedEvents(this.state.selection);
+    if (!selectedEvents.length) return;
```
4ian: There is a risk the last select event is not a comment/group, so better filter them (see `filterEditableWithEventTextDialog` in a next comment):

```
const editableEvents = filterEditableWithEventTextDialog(getSelectedEvents(this.state.selection));
if (!editableEvents.length) return;

const event = editableEvents[editableEvents.length - 1];  // Get the last selected event.
```

newIDE/app/src/EventsSheet/index.js
Bouh: No, the sub-menus must be concise.

newIDE/app/src/UI/Theme/DarkTheme/EventsSheet.css
Bouh: It's a remnant of when I split the PRs.

But i see that on darktheme, so i've deleted the border. and it need `!important;` css rule.
image

btw it's alredy done in #1309 , not yet merged :/

newIDE/app/src/EventsSheet/index.js
```diff
@@ -1105,85 +1130,96 @@ export default class EventsSheet extends React.Component<Props, State> {
...
-                              click: () => this.toggleDisabled(),
-                              enabled: this._selectionCanToggleDisabled(),
-                            },
-                            { type: 'separator' },
-                            {
-                              label: 'Add New Event Below',
-                              click: () =>
-                                this.addNewEvent(
-                                  'BuiltinCommonInstructions::Standard'
+                          buildMenuTemplate={() => {
```
4ian: Thanks for the changes :) Now you should be able to go back to a function directly returning an array: 
```
() => [
   ...
```

Slightly shorter/less indentation :)

newIDE/app/src/UI/Theme/DarkTheme/EventsSheet.css
4ian: Alright I see, this is because of this rule: 

https://github.com/4ian/GDevelop/blob/9763d6523c797b6d721b25d444156a16d89de28d/newIDE/app/src/EventsSheet/EventsTree/style.css#L12-L15

Background color and border are set here. I think we should instead:
* Remove the border in this common rule.
* Probably remove also the background in this common rule, so that we can (probably) remove the !important that is already there on the background in the dark/default theme css files.

4ian: Note that if you want to do this in the other PR, you can remove here the changed lines.

newIDE/app/src/EventsSheet/index.js
Bouh: Done ✅

Bouh:
When you merge this think to merge #1309 with it :)

4ian:
Yep good remark, let me check the other PR again :)

4ian:
Merged and now the next one :) Thanks for adding this dialog!

PR #1313, opened 2019-12-04
Fix outline color for lineV2

Bouh:
Because outline color are set to black by default if we pass just width of line.
[Two users was stuck on this.](https://forum.gdevelop-app.com/t/primitive-drawing-color/21751)


4ian:
Looks good, thanks for the fix 👍

Issue #1052, opened 2019-05-16
Panel Sprite: Change image results in corners being upside down


Wend1go:
## Describe the bug
Changing the image of a panel sprite at runtime via "Image name" action flips all corners upside down, except the upper left one.
![image](https://user-images.githubusercontent.com/7413059/57869725-f0b3de00-7805-11e9-8cff-b40ba49f9402.png)

## To Reproduce
1. Create a scene and add a Panel Sprite
2. Assign a texture and set the corner tiles
3. Add an instance of it to the scene (it gets drawn correctly)
4. Create an event to set a new image (Image name)
5. The Panel Sprite gets updated with the new image but the corners are drawn upside down
Only the upper left corner is drawn correctly. (This also happens if you assign a completely different image)

## Example project showing the bug
[PanelSpriteUpsideDownBug.zip](https://github.com/4ian/GDevelop/files/3188003/PanelSpriteUpsideDownBug.zip)

## Other details
GDevelop 5.0.0 Beta 68

EDIT:
On a second thought, it looks like it uses the upper left corner for all the corner tiles.

Wend1go:
More screenshots showing the issue here: 
https://github.com/4ian/GDevelop/pull/1114#issuecomment-503215579

4ian:
This is (finally 😬) fixed for next version thanks to @Bouh fix! Thanks again @Bouh 👏

4ian:
FYI, the upside down corner was still a problem it seems, will be fixed, see https://github.com/4ian/GDevelop/issues/1723! :)

PR #1314, opened 2019-12-10
Fix pixel perfect on Panel Sprite 9-patch

Bouh:
This is for fix the renderer of Panel Sprite ("9-patch").

Lot of weird code have been added over time.
And lot of weird little patch with 1px here or 2px somewhere else for obscure reasons.

I've read the code and rework and verify twice on my PC, and a other user have check too.

[For see the problem check this topic on forum.](https://forum.gdevelop-app.com/t/les-9-panels-sprites-ne-saffichent-pas-correctement/21867/3)

This need to be verify with a retina screen from Apple.
I hope in future we will be able to have same quality of preview in scene editor.

### DIFF
--- a/Extensions/PanelSpriteObject/panelspriteruntimeobject-pixi-renderer.js
+++ b/Extensions/PanelSpriteObject/panelspriteruntimeobject-pixi-renderer.js
@@ -90,7 +90,7 @@ gdjs.PanelSpriteRuntimeObjectPixiRenderer.prototype._updateLocalPositions = func
   this._borderSprites[0].position.y = obj._tBorder;
 
   //Top-right
-  this._borderSprites[1].position.x = obj._width - this._borderSprites[7].width ;
+  this._borderSprites[1].position.x = obj._width - this._borderSprites[1].width ;
   this._borderSprites[1].position.y = 0;
 
   //Top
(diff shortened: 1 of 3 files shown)


4ian:
😍

> Lot of weird code have been added over time.
And lot of weird little patch with 1px here or 2px somewhere else for obscure reasons.

Yeah surely my fault ☝️ I've done a wrong fix surely at some point that generated another bad fix and so on 😬

Thanks for checking that, I'll test that tonight on a few games :)

Extensions/PanelSpriteObject/panelspriteruntimeobject-pixi-renderer.js
```diff
@@ -1,204 +1,343 @@
...
+
+  this._centerSprite.position.x = obj._lBorder;
+  this._centerSprite.position.y = obj._tBorder;
+
+  //Right
+  this._borderSprites[0].position.x = obj._width - obj._rBorder;
+  this._borderSprites[0].position.y = obj._tBorder;
+
+  //Top-right
+  this._borderSprites[1].position.x = obj._width - this._borderSprites[7].width ;
```
4ian: This is equivalent, but I guess this should be `this._borderSprites[1].width`?

4ian:
I've tested on a few games and this seems to be perfect! For example, there was glitches in the space shooter "TilesMechanical1" object, that are now gone during the game 👍

Great work! Could we port these fixes to RenderedPanelSpriteInstance? Have you tried?

So happy to see these hacks gone for good! I believe the root of all the issues is that I started to add some +1/-1 in `setTexture`.

Bouh:
Oh i didn't see  RenderedPanelSpriteInstance, i will take a look

Bouh:
It's sightly better :D !

4ian:
Sounds good! Will merge later today or tomorrow :)

Is there a way to make some documentation update or something to make more obvious that the renderers for the objects in the scene editor are in another folder?
Ideally I would move these files to each extension folder 🧐 (though it's not as easy, as built in extensions don't have a way to expose a JS file to the IDE).

Bouh:
In the file itself?
Or in a readme in the parent directory of the files.
With a small link on the main readme?

I didn't find it because when I look for links between the files I search through the whole codebase with the names of the methods.

Wend1go:
This might also fix #1052
I have a test project linked in the issue.

4ian:
Merging this now :) Thanks a lot for these fixes! Will be great to finally have pixel perfect 9-patch objects 👏

PR #1320, opened 2019-12-15
Rename objects to instances

Bouh:
Maybe this is just a first step.

Lot of instructions use the term "object", but this is not an "object" this is an "instance".

"Add an object" should also change to "Add an instance of object XXX" 
There are multiples sentences to change.

https://forum.gdevelop-app.com/t/solved-invert-condition-isnt-work-properly-how-work-instances-and-multiple-objects/21954/14


4ian:
Thanks! I've made 2 fixes.

> Lot of instructions use the term "object", but this is not an "object" this is an "instance".

I think it's mostly important for action/conditions that are changing the instances being selected ("picked") for next conditions/actions.

This is the case for "for each", "pick all", and also "create an object" (as it's adding a new instance of an object).

For other, I think it's redundant to always write "the instances of the specified object(s)". While technically it's true that everything is applied on instances, it's fine to designate only the "class" to which we're applying changes.. i.e: the object name :)

PR #1321, opened 2019-12-16
Add support for adding effects in extensions

4ian:
This adds support for extensions to provide effects ("shaders", PixiJS "filters").
Only filters that are used in a game will be included in the exported/previewed game.

Hopefully the "boilerplate" to add a new effect is small:
* Add an addEffect into the `Effects` extension. Set the properties
* Add a js file with the PIXI implementation + the call to register the effect (function to create it and function to update its parameter).
* That's all :)

TODO:
- [x] Do we put the pixi filter in a file called `something-pixi-renderer.js` to follow the same convention as other files? => Surely yes.
- [x] Modify the game engine to load effect type that have been included
  - [x] How to register them? Auto discovered as done for object/behaviors?
- [x] Put existing filters (sepia, etc...) into a built-in extension.
- [x] Fix Travis-CI automatically build not building libGD.js anymore. 
- [x] Make mock/fake effects for Storybook.

Concerning pixi-filters:
- I'll merge this as soon as it's ready, but I'm thinking that in the future we could almost make a script that `npm install` all the pixi-filters package, copy the `node_modules/@pixi/filter-XXX/dist/filter-XXX.js` into an extension folder and auto generate the JS file extension (JsExtension.js) with a `addEffect` for each one of them (we'll still need to store somewhere information about parameters).


4ian:
@Bouh @blurymind @Wend1go FYI, you can now add new effects to GDevelop ;) ✨🎇

See the links about effects in the extension README: https://github.com/4ian/GDevelop/blob/master/newIDE/README-extensions.md#how-to-create-an-effect-shader-pixijs-filter

In short:
1. Add your effect in https://github.com/4ian/GDevelop/blob/master/Extensions/Effects/JsExtension.js#L29-L40
2. Reload GDevelop, add it to a layer.
3. Make your PIXI filter in a JS file like the one already in the Effects folder: https://github.com/4ian/GDevelop/tree/master/Extensions/Effects
4. Changes in the JS files (apart from JsExtension.js) are automatically imported, so you can just create your Pixi filter JS file/modify it and launch a preview.

Let me know if you have any issues :)

blurymind:
Thank you @4ian :D I will give this a try later this week. Might be worth
creating a task on the Trello board for adding the missing pixi filters and
anyone doing a set could say so in advance, so we don't work on the same
ones :)

4ian:
Good idea, I've made this card: https://trello.com/c/h1zZcVYp/388-adding-a-new-effect-to-gdevelop

Should be archived later when we have lots of effects :)

PR #1325, opened 2019-12-19
add pixi filters WIP

blurymind:
Adding filters from
http://pixijs.io/pixi-filters/tools/demo/

https://github.com/pixijs/pixi-filters/tree/3138e6d973251d1d46d53c443369636fb36b79fb/tools/demo/src/filters

So far added:
- [x]  TiltShiftFilter
- [x]  KawaseBlurFilter
- [x]  AdvancedBloomFilter
- [x]  CRTFilter (requires update method)
- [x]  GodRayFilter (requires update method)
- [ ]  OutlineFilter (requires color input)
- [ ]  RGBSplitFilter
- [ ]  GlowFilter (requires color input)
- [ ]  DropShadowFilter (requires color input)
- [x]  ZoomBlurFilter

TODO:
- [x] update time and seed on filters that require it (once @4ian adds an update method to the filter api)

Putting this out early for feedback :)


blurymind:
Some filters requite a time value to be changed when pixi animates by default, not sure how to do that with GD yet

Bouh:
Maybe these methods can help you ?
https://github.com/4ian/GDevelop/blob/master/GDJS/GDJS/Extensions/Builtin/TimeExtension.cpp

blurymind:
I need something that runs on frame update. The current function only runs
when a filters parameter is changed I think

Bouh:
No, you're not drunk, it's just the displacement filter.
It's super easy to add new filters!
@4ian well done for the implementation nice job!

(Displacement map file is hardcoded for this test, see below why)
image


- With displacement i need an image, how I include it?
`var displacementSprite = new PIXI.Sprite.fromImage("res/displacement_map.png");`
This dosen't work, the file is not exported for preview.  Ideally if the user could choose his image ;)

- The addition of time value also interests me to put some movement on the filter with parameters.

- Also where i push my code after, in a new pr or in this (how if i can?)

blurymind:
My understanding is that atm the layer filters can only take a float and they take it as a string in the js api. It is kind of limited. You can note a min and max value in the description of a parameter,, but the ide  kind of lets you enter anything in there

4ian:
> With displacement i need an image, how I include it?

I can add support for properties for effects that are "resources" (much like for the video object you have a property that is a "resource" of type "video") or "color".
This means you'll receive them as strings in the JS, and you can use the image manager to get the associated PIXI texture or read the color.

> Some filters require image and color input, as well as an update function to increment a parameter

Yes I can also take a look at adding an `update` function that will be called at every frame, passing the `runtimeScene` and the `runtimeLayer` (so you can notably [get the elapsedTime of the layer](https://github.com/4ian/GDevelop/blob/master/GDJS/Runtime/layer.js#L266)).

> No, you're not drunk

Thanks for the clarification 😄

> , it's just the displacement filter.

Nice job 👍

> @Bouh nice job. I suppose you could do another PR. 

Yep you can open separate PRs. It's more granular, we can merge filters that are ready while the others can wait for the updated parameters/time update.

blurymind:
Thank you @4ian  😺 with these additions GD will be able to fully take advantage of the pixi filters

blurymind:
@Bouh I updated my list for this PR :) just managed to add a few more!

Extensions/Effects/god-ray-pixi-filter.js
```diff
@@ -0,0 +1,15 @@
...
+  makePIXIFilter: function() {
+    var center = new PIXI.Point(0, 0);
+    var godray = new PIXI.filters.GodrayFilter({center});
+
+    return godray;
+  },
+  updateParameter: function(filter, parameterName, value) {
+    if (!['light', 'time', 'gain', 'lacunarity', 'parallel', 'angle', 'x', 'y'].includes(parameterName)) return;
+
+    if (['x', 'y'].includes(parameterName)) filter.center[parameterName] = gdjs.PixiFiltersTools.clampValue(value, -1000, 1000);
```
4ian: For test with < 4 elements, avoid creating an array and do two if checks. In general, avoid in the game engine avoid creating an object/array if not strictly necessary because this creates garbage objects in memory, which then put pressure on the garbage collector which make the game lag on less powerful devices. 

You can either:
- make "dumb" if conditions (let's say 2 to 3 elements).
- create the array as as static member of some variable:

`PIXI.filters.GodrayFilter.supportedParameters = ['light', 'time', 'gain', 'lacunarity', 'parallel', 'angle', 'x', 'y']`

(this also add an indirection in the code, which is why I recommend only if a lot of values).

Extensions/Effects/god-ray-pixi-filter.js
```diff
@@ -0,0 +1,15 @@
...
+    else if (filter !== 'parallel') filter[parameterName] = gdjs.PixiFiltersTools.clampValue(value, 0, 20);
```
4ian: nitpicking: associate the block of the if condition with what is in the condition.

Here, we read:

```
if (something related to x/y) doSomethingWithXorY
else if (something with parallel) doSomethingNotRelatedToParallel
else doSomethingWithParallel
```

See the inconsistency?
Keep the less precise case for the `else`:

```
else if (filter === 'parallel') ...
else filter[parameterName] =
```

Extensions/Effects/zoom-blur-pixi-filter.js
```diff
@@ -0,0 +1,15 @@
+gdjs.PixiFiltersTools.registerFilterCreator('ZoomBlur', {
+  makePIXIFilter: function() {
+    var zoomBlur = new PIXI.filters.ZoomBlurFilter();
+
+    return zoomBlur;
+  },
+  updateParameter: function(filter, parameterName, value) {
+    if (!['centerX', 'centerY', 'strength', 'innerRadius'].includes(parameterName)) return;
```
4ian: includes is unfortunately not supported by IE11. We must `.indexOf(parameterName) !== -1` instead (here and in the rest of the filters)

Extensions/Effects/crt-pixi-filter.js
```diff
@@ -0,0 +1,13 @@
+gdjs.PixiFiltersTools.registerFilterCreator('Crt', {
+  makePIXIFilter: function() {
+    var crt = new PIXI.filters.CRTFilter();
+
+    return crt;
+  },
+  updateParameter: function(filter, parameterName, value) {
+    if (!['curvature', 'lineWidth', 'lineContrast', 'verticalLine', 'noise', 'noiseSize', 'vignetting', 'vignettingAlpha', 'vignettingBlur', 'seed', 'time'].includes(parameterName)) return;
+
+    if (filter !== 'verticalLine') filter[parameterName] = gdjs.PixiFiltersTools.clampValue(value, 0, 20);
```
4ian: Same as another comment, put the generic things at the end of a if/else cascade.

Extensions/Effects/JsExtension.js
```diff
@@ -138,12 +138,115 @@ module.exports = {
...
-      new gd.PropertyDescriptor(/* defaultValue= */ '1')
-        .setLabel(_('Opacity (between 0 and 1)'))
-        .setType('number')
-    );
+    sepiaProperties.set('opacity', new gd.PropertyDescriptor(/* defaultValue= */ '1').setLabel(_('Opacity (between 0 and 1)')).setType('number'));
+
+    const setFilterProperties = (propertiesSchema, filterProperties) => {
+      Object.keys(propertiesSchema).forEach(property => {
+        const value = propertiesSchema[property];
+        filterProperties.set(property, new gd.PropertyDescriptor(/* defaultValue= */ value.default).setLabel(value.name + ' (' + value.description + ')').setType('number'));
```
4ian: I know that you want to avoid putting ' (' + ... + ')' everywhere, but I was thinking of putting the name of the parameter (value.name) in the UI in the IDE 😬😬

I know that you want to save a few lines, but I would really prefer to stick to the API without introducing now another layer of abstraction which will make newcomers wonder what is this thing that differs from the documentation.

If there are improvements to the UI, let's make them in the IDE rather than baking them in the extensions :)

Extensions/Effects/crt-pixi-filter.js
```diff
@@ -0,0 +1,13 @@
...
+    else filter.verticalLine = value === 1;
```
4ian: This looks like a boolean? Put a `//TODO` so that we don't merge this while booleans are not supported in the IDE and this is not updated to show a boolean in the resources :)

Extensions/Effects/JsExtension.js
```diff
@@ -138,12 +138,115 @@ module.exports = {
       .setDescription(_('Alter the colors to sepia'))
       .addIncludeFile('Extensions/Effects/sepia-pixi-filter.js');
     const sepiaProperties = sepiaEffect.getProperties();
-    sepiaProperties.set(
-      'opacity',
-      new gd.PropertyDescriptor(/* defaultValue= */ '1')
-        .setLabel(_('Opacity (between 0 and 1)'))
-        .setType('number')
-    );
+    sepiaProperties.set('opacity', new gd.PropertyDescriptor(/* defaultValue= */ '1').setLabel(_('Opacity (between 0 and 1)')).setType('number'));
```
4ian: Please run Prettier on the file :) 
I know that you would prefer to have everything on a single line, but auto-formatting is giving 10x more benefits than formatting manually.

Also I understand the interest for "one liners". "It's a single line it's shorter!" I used to like this :) But if you have multiple calls to functions, there is no need to "hide" them on a line. It's on the contrary very clear:
...
.setSomething(toThis)
.setThis(toThat)
.setThat(toThis)

4ian: At some point, we can't reduce complexity. If our filters need parameters, we need lines for them :) It's not a problem. It's what effects are!

Extensions/Effects/god-ray-pixi-filter.js
4ian: Also that your approach is perfectly valid in the IDE or in a codebase that is handled by webpack or generally that is not a game. But in the game engine, we're optimising for less arrays/objects creation as possible - which makes things sometimes not as pretty as they could be in an app codebase.

Extensions/Effects/JsExtension.js
```diff
@@ -138,12 +138,115 @@ module.exports = {
...
+    const zoomBlurEffect = extension
+      .addEffect('ZoomBlur')
+      .setFullName(_('Zoom Blur'))
+      .setDescription(_('Add Zoom blur effect'))
+      .addIncludeFile('Extensions/Effects/pixi-filters/filter-zoom-blur.js')
+      .addIncludeFile('Extensions/Effects/zoom-blur-pixi-filter.js');
+    const zoomBlurProperties = zoomBlurEffect.getProperties();
+    setFilterProperties(zoomBlurPropertiesSchema, zoomBlurProperties);
 
     return extension;
```
4ian: I look forward to try all of these 😄😄

4ian:
@blurymind @Bouh I've added support for color and resource parameters :) (you need to merge master in your branch/rebase your branch on master)

Example of declaring properties for these:
https://github.com/4ian/GDevelop/blob/ec9a629124947cf4bc45114052cf5652ae3bfc94/Extensions/ExampleJsExtension/JsExtension.js#L38-L57

The `registerFilterCreator` has evolved a bit to:
* Add an `update` method (with the `filter` and the `layer`, useful to get the time)
* You now have to specify two functions: `updateStringParameters` and `updateDoubleParameters` (for numbers)
* Finally, the `layer` is passed to makePIXIFilter (so that you can get access to the pixi image manager if required to get a PIXI Texture), as well as `effectData` (raw JS object representing the effect, in case you want to read into parameters directly).

Note that I'm using the term "Double" instead of "Number" in the code, as these parameters are tightly related shader uniforms.

https://github.com/4ian/GDevelop/blob/ec9a629124947cf4bc45114052cf5652ae3bfc94/Extensions/ExampleJsExtension/dummyeffect.js#L30-L70

Extensions/Effects/crt-pixi-filter.js
blurymind: I can remove the ability to enable it instead, so we can still merge it? I love this filter and doubt anyone would want to enable the verticalLine feature

4ian: Yep it's fine to go ahead without this and add this later :)

blurymind:
The Boolean type is still needed by some of the other shaders though.. the
godray for example has a more important property that users might want

4ian:
Sure, let me check that then.

4ian:
Support for `boolean` parameters is added on master :) Remember to add `updateBooleanParameter` (even if the function is empty) to your filters.

blurymind:
thank you @4ian :) I will address the review notes and add the update when I get back to my pc tomorrow

Bouh:
I misread your list, I did the Reflection effect.

4ian:
Also, feel free to open multiple PRs even with 1 or 2 effects inside. Will be easier to merge and will avoid a single effect that needs fixes/updates to hold all the other effects from being merged/released :)

blurymind:
Ah no worries, I will just remove the reflection from my to-do :)

blurymind:
@Bouh would you like to pick up more from the unticked here? I havent started on them yet, so let me know if you do

blurymind:
@4ian thank you for these awesome api updates! I made some progress today, but will need some more time to polish it.

I will do a commit hopefully some time tomorrow. It's been getting busy around the holidays, sorry for the delays.

Extensions/Effects/zoom-blur-pixi-filter.js
blurymind: I am refactoring this for all filters now. Sorry for the inconvenience :)

4ian:
Sure take your time and hopefully we have everything now for these filters :) Look forward to have them in GD! :)

blurymind:
Thank you for the patience @4ian   I have updated the PR to be inline with the api updates, also removed the utility approach that was not efficient.  There are still 4 to add, but I can do them in another PR after this one is finished if you like.
Merry christmas  :santa:

Extensions/Effects/god-ray-pixi-filter.js
blurymind: updated with simple if statements

Extensions/Effects/JsExtension.js
blurymind: removed the utility approach and now using the standard way

Extensions/Effects/JsExtension.js
blurymind: I have prettier running now, but it updated some lines that are not a part of the pr, so I wonder if my prettier settings are not aligned. Do we have something in the readme about how to set prettier?

Extensions/Effects/JsExtension.js
blurymind: thank you for the patience :smile:

Extensions/Effects/JsExtension.js
4ian: Are you sure you've run Prettier and not some kind of default autoformatting tool like the one bundled in VS Code?

4ian:
> There are still 4 to add, but I can do them in another PR after this one is finished if you like. :)

Sure that would be great 👍👍

If you're done with this, let me launch Prettier on this and make sure all descriptions are good to go before merging :)

Extensions/Effects/JsExtension.js
4ian: Otherwise I can do it, let me know

blurymind:
sure lets split the remaining for another PR :)

Extensions/Effects/crt-pixi-filter.js
```diff
@@ -0,0 +1,49 @@
+gdjs.PixiFiltersTools.registerFilterCreator('Crt', {
+  makePIXIFilter: function(layer, effectData) {
+    var crtFilter = new PIXI.filters.CRTFilter();
+
+    return crtFilter;
+  },
+  update: function(filter, layer) {
+     if (filter.animated) {
+      filter.time += layer.getElapsedTime() / 1000;
+      filter.seed = Math.random();
```
4ian: This should probably be in the `makePIXIFilter`?

4ian:
Godray had a syntax error, I fixed this (check Semaphore CI results for GDJS Typing, which is new and check these kind of issues). 
Merging this now! Thanks for adding these!! 🎁👍

Extensions/Effects/crt-pixi-filter.js
4ian: Fixed

Extensions/Effects/crt-pixi-filter.js
```diff
@@ -0,0 +1,40 @@
+gdjs.PixiFiltersTools.registerFilterCreator('Crt', {
+  makePIXIFilter: function(layer, effectData) {
+    var crtFilter = new PIXI.filters.CRTFilter({ seed: Math.random() });
+
+    return crtFilter;
+  },
+  update: function(filter, layer) {
+    if (filter.animated) {
+      filter.time += layer.getElapsedTime() / 1000;
+    }
```
4ian: My bad, I misunderstood how seed worked, I fixed this in master :)

blurymind:
Thank you @4ian :) I actually used my templating svelte app to recreate
these, but forgot to re-run prettier on some of the files. Sorry for that.
I am now wondering if I can get it to run prettier on the generated code
for me. I will incorporate any changes you made into the gdfilter template
I have and create the remaining filters. Btw what is the character limit
per row you have set for prettier. I think I need to update my vscode
settings

4ian:
Configuration of Prettier is the standard one with simple quotes: https://github.com/4ian/GDevelop/blob/master/Extensions/.prettierrc

Character limit is [80 by default](https://prettier.io/docs/en/options.html).

blurymind:
I think it was not picking it up for some reason. Thus why it moved some
things around that were not a part of the or. Sorry about that 😬

blurymind:
Thank you again for the patience 😊

blurymind:
@Bouh I am starting work on the remaining filters.Let me know if you are doing any work on them

Bouh:
Nothing, you can start to work on it.

Bouh:
@blurymind have you seen tilt shift effect dosen't work like expected.
It worked in a dev environment?

image

4ian:
Could this be related to the fact that the platformer does not have a background image?

blurymind:
Yes I noticed this earlier. It is because of the lack of background image,
exactly as @4ian says. A lot of shaders glitch out there, not just than
one. A possible fix would be to change the way gd renders this background
colour.

Bouh:
The demo link show tilt shift with two blur, on top and bottom.
In GD i got just top blur, and he take 50% of screen
image

blurymind:
You are combining filters there, so it's possible that these two don't work
together so well. When you do the same at the pixi filters demo website, do
you get the same effect? I wonder if the order also matters.

PR #1329, opened 2019-12-21
Add effect

Bouh:
@blurymind Here my list.

- [x] Displacement (**Can be improve with offset?**)
- [x] ColorMap
- [x] Pixelate
- [x] [Reflection](https://pixijs.io/pixi-filters/docs/PIXI.filters.ReflectionFilter.html)


Extensions/Effects/displacement-pixi-filter.js
```diff
@@ -0,0 +1,22 @@
...
+    var displacementSprite = new PIXI.Sprite.fromImage(
+      "res/displacement_map.png"
+    ); //Use a basic image bundle with GD for now, prefer use a customizable file set by user.
+    var displacementFilter = new PIXI.filters.DisplacementFilter(
+      displacementSprite
+    );
+
+    return displacementFilter;
+  },
+
```
4ian: You must add updateBooleanParameters and updateDoubleParameters even if they are empty. I could maybe make them optional.

Extensions/Effects/JsExtension.js
```diff
@@ -27,127 +27,151 @@ module.exports = {
...
+    );
+    displacementProperties.set(
+      "scaleX",
+      new gd.PropertyDescriptor(/* defaultValue=20 */ "20")
+        .setLabel(_("Scale X"))
+        .setType("number")
+    );
+    displacementProperties.set(
+      "scaleY",
+      new gd.PropertyDescriptor(/* defaultValue=20 */ "20")
```
4ian: Write just /* defaultValue= */ "20"

The defaultValue is here to give an empty about what the "20" is :)

Bouh:
>  You now have to specify two functions: `updateStringParameters` and `updateDoubleParameters` (for numbers)
[Link to your post](https://github.com/4ian/GDevelop/pull/1325#issuecomment-568173291)

I've copy/paste these two methods, and i was stuck in my code because these methods don't have **S**.
Note for me always check the existing filters when something wrong with mine.

4ian:
Oops 😬 Fixed my post!
Yeah maybe a careful inspection of the error message could have hinted you that `updateDoubleParameter` was missing. Sorry about that!

Extensions/Effects/displacement-pixi-filter.js
```diff
@@ -0,0 +1,27 @@
...
+    return displacementFilter;
+  },
+  updateBooleanParameters: function(filter, parameterName, value) {
+    return;
+  },
+  updateDoubleParameters: function(filter, parameterName, value) {
+    return;
+  },
+  updateStringParameters: function(filter, parameterName, value) {
+    if (filter === "scaleX") {
```
4ian: `parameterName`, not `filter` right? 🤔

4ian: Also a scale is a number, so this should be in `updateDoubleParameter` (no extra s)

Bouh:
Yes i've see my errors, i'll commit my code tomorrow. It's working i just need to see which parameters can be cool to use.

Bouh:
Filter is working.

I've tried to add an offset but nothing seem working.
Tested also with TillingSprite instead Sprite.
`filter.maskSprite.tilePosition.x`

But ideally i just need acces to displacementSprite for change `x` and `y` but i can't access it in update() ?

I've tried to move with all these case: 
```
filter.maskSprite.transform.position.x

filter.uniforms.mapSampler.defaultAnchor.x
```

Bouh:
- Pixelate is finish for me, the review can be done.

- colorMap is almost finish. This effect use a file with colors on it, colors from file are used as color table for transpose the color from image on color present on layer.
If you didn't understand what i mean, googlize "LUT for photography".

User need to have easy access to [original the template](https://github.com/4ian/GDevelop/blob/ec477aebab9ebbd10e8a96c72b28beb237662f8e/Extensions/Effects/pixi-filters/colormap-template-16.png) for edit it and load it in GD.

I've added a descrption for it but i don't know where put the file.
Maybe another solution can be done, attach by default as resource the original file ? (but becareful if user load lot of similar effect, we don't wish twice same color map (lut) in exported game)
Or add a button for open this file in one click ? 
If you have another idea :)

Extensions/Effects/reflection-pixi-filter.js
```diff
@@ -0,0 +1,68 @@
...
+        effectData.doubleParameters.alphaEnding
+      ],
+      time
+    );
+
+    return reflectionFilter;
+  },
+  update: function(filter, layer) {
+    if (filter.animated) {
+      filter.time -= layer.getElapsedTime() / 1000;
```
4ian: Interesting, why subtracting rather than adding the elapsed time? :)

Extensions/Effects/JsExtension.js
```diff
@@ -27,127 +27,289 @@ module.exports = {
...
+        .setType("number")
+        .setDescription(_("Ending alpha (1 by default)"))
+    );
+
+    reflectionProperties.set(
+      "animated",
+      new gd.PropertyDescriptor(/* defaultValue= */ "false")
+        .setLabel(_("animated"))
+        .setType("boolean")
+        .setDescription(_("Animate position of waves"))
```
4ian: Should that just be "Animate waves"?

Extensions/Effects/JsExtension.js
```diff
@@ -27,127 +27,289 @@ module.exports = {
...
+      .setFullName(_("Reflection"))
+      .setDescription(_("Reflect the layer"))
+      .addIncludeFile("Extensions/Effects/reflection-pixi-filter.js")
+      .addIncludeFile("Extensions/Effects/pixi-filters/filter-reflection.js");
+    const reflectionProperties = reflectionEffect.getProperties();
+
+    reflectionProperties.set(
+      "mirror",
+      new gd.PropertyDescriptor(/* defaultValue= */ "true")
+        .setLabel(_("mirror"))
```
4ian: The label is shown in the UI to the user. It should be what is displayed next to the checkbox, so more something that looks like the description: "Reflect the image on the waves"

Extensions/Effects/reflection-pixi-filter.js
Bouh: Because movement of waves go on up, so i substract it for go down.

Extensions/Effects/JsExtension.js
```diff
@@ -27,127 +27,289 @@ module.exports = {
...
+    const colorMapProperties = colorMapEffect.getProperties();
+    colorMapProperties.set(
+      "colorMapTexture",
+      new gd.PropertyDescriptor("")
+        .setType("resource")
+        .addExtraInfo("image")
+        .setLabel(_("Color map texture for the effect"))
+        .setDescription(
+          _(
+            "Color map are like LUT, you can change colors of pixels by modifing a reference color image containing each colours. Orignal map file can be found at [TODO INSERT PATH]"
```
4ian: I suggest we host the map file on the wiki and put a link. Use markdown (`[original map file](http://url)`) and I'll update the IDE to be able to display markdown with links in descriptions :)

Bouh: Markdown link are not clickable, and create button are a bit offset.
image

4ian: I just added support for markdown in master ;)

4ian:
I've added support for markdown (supported for effects and also for the "long descriptions" that can now be added to actions/conditions, including in extensions made in GD). Let me know if that's all for this PR and we can merge it if it's ready :)

Bouh:
- Color link are not very visible, maybe we can set blue ont it, same blue as primary button.
image

- And link dosen't open webpage in external program but in GD5 i guess because GD5 is reboot. And webpage is not open.

- Otherwise I wanted to add an offset parameters on the displacement but I can't do it, it doesn't seem designed for it.

Otherwise it's good for me.

Extensions/ExampleJsExtension/JsExtension.js
```diff
@@ -25,7 +25,8 @@ module.exports = {
     const dumyEffect = extension
       .addEffect('DummyEffect')
       .setFullName(_('Dummy effect example'))
-      .setDescription(_('This is an example of an effect ("shader")'))
+      .setDescription(_('This is an example of an effect ("shader") [External link on wiki](http://wiki.compilgames.net/doku.php/gdevelop5/) and **bold letters**"
+'))
```
4ian: You have a syntax error here ;)

4ian:
I am also confused by the Prettier usage of double quotes, I think it's because I included a .prettierrc for newIDE, GDevelop.js and GDJS but never for Extensions 😬
Can you get the latest commits from master and re-run Prettier on the two JsExtension.js files? 😇

I've not checked the links color with the dark theme too! I'll fix that.

Bouh:
@blurymind 
I've started to write [list of effect on wiki](http://wiki.compilgames.net/doku.php/gdevelop5/interface/scene-editor/layer-effects#list_of_effects). If your effects need ressources you can add it on this page.
You can add yours.

4ian:
I've fixed/updated descriptions and updated the wiki (fixed mistakes, added images). If the list is growing too much, we can consider moving it to another page?

Merging this now! Thanks for adding these filters 💪

Bouh:
Why **Revert package-lock.json** ?
About this file, it's should be in .gitignore no ? when i use npm install all my package are updated and this package-lock.json and package.json are edited automatically, so sometimes these files are present in my commit. 
I try to avoid to stage the files, and also discard change. 
This is recurrent for me.

4ian:
> About this file, it's should be in .gitignore no ? 

No this file must be version controled as it should be used by npm to install the exact same versions of the packages.

I might need to update this, but this PR is not a good place for this. It should be done on another PR/master, because if for some reason this create issue, we must be able to revert the PR/commit without reverting other changes.

Unrelated note, I've tried the color map but I was not able to get any result, despite using the example color map from the wiki? Not sure what I'm doing wrong, can you take a look or make a quick example?

Bouh:
> I've fixed/updated descriptions and updated the wiki (fixed mistakes, added images). 
Thank you, you do this better than me.

> If the list is growing too much, we can consider moving it to another page?
Not necessarily, it's nice to have everything on one page. 

There are pages in the wiki that are much longer.

> Merging this now! Thanks for adding these filters 💪
Yeah !

Bouh:
> Not sure what I'm doing wrong, can you take a look or make a quick example?

It was working. I'm checking now

Bouh:
@4ian [here a demo](https://youtu.be/g-2x5aNOqEw)
(Don't share the link)

4ian:
Understood! I was misunderstanding how this effect works :) I've added some explanations in http://wiki.compilgames.net/doku.php/gdevelop5/interface/scene-editor/layer-effects

Maybe we could add on the wiki a few example of some modified color map, so that "newbies" like me can quickly try the effect?

Bouh:
Yes i add some images and links on wiki.
image

PR #1330, opened 2019-12-22
Littles fix & improvement

Bouh:
Can we keep this PR open until the day of the release?
When i see small errors, or little thing to improve i do in this PR.
Commits are named/commented 🤓

------- 

Last commit add colors on icons in warning and error message. Because user need to read it, and color have more impact!
It's add two label on effect layer because Effect name feild is always fill with "Effect" then we don't know what it is.

image

### DIFF
--- a/newIDE/app/src/EffectsList/index.js
+++ b/newIDE/app/src/EffectsList/index.js
@@ -146,51 +146,43 @@ export default class EffectsList extends React.Component<Props, {||}> {
                 return (
                   <React.Fragment key={i}>
                     <MiniToolbar>
-                      <Column expand noMargin>
-                        <Line>
-                          <MiniToolbarText>
-                            <Trans>Effect name:</Trans>
-                          </MiniToolbarText>
-                          <SemiControlledTextField
-                            margin="none"
-                            commitOnBlur
-                            hintText={t`Enter the effect name`}
-                            value={effect.getName()}
-                            onChange={newName => {
-                              if (newName === effect.getName()) return;
+                      <MiniToolbarText>
+                        <Trans>Effect name:</Trans>
+                      </MiniToolbarText>
+                      <SemiControlledTextField
+                        margin="none"
+                        commitOnBlur
+                        hintText={t`Enter the effect name`}
+                        value={effect.getName()}
+                        onChange={newName => {
+                          if (newName === effect.getName()) return;
 
-                              effect.setName(newName);
-                              this.forceUpdate();
-                              this.props.onEffectsUpdated();
-                            }}
-                            fullWidth
+                          effect.setName(newName);
+                          this.forceUpdate();
+                          this.props.onEffectsUpdated();
+                        }}
+                        fullWidth
+                      />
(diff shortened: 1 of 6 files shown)


newIDE/app/src/UI/AlertMessage.js
```diff
@@ -30,8 +30,22 @@ const AlertMessage = ({ kind, children, onHide, renderRightButton }: Props) => (
   <Paper>
     <Line noMargin alignItems="center">
       {kind === 'info' && <Info style={styles.icon} />}
-      {kind === 'warning' && <Warning style={styles.icon} />}
-      {kind === 'error' && <Error style={styles.icon} />}
+      {kind === 'warning' && (
+        <Warning
+          style={{
+            ...styles.icon,
+            color: `orange`,
```
4ian: Can we use ThemeConsumer and get the color from the theme? In the future themes might want to use other colors.

4ian:
Colors are a good idea! The labels sounds fine, though your screenshot made me realize that the texts (MiniToolbarText) and the text fields are not aligned (this is unrelated to your PR). The fix should not be done on MiniToolbarText, because it seems properly aligned with checkboxes text already:

image
(40 and Bimbo_JVE.ttf are slightly up) (the strange accents is because I'm using Pseudolocalization as my language)

image
(Score is slightly up)

We might want to fix that by adding a slightly top margin to the text fields having "none" as "margin", though this needs to be done careful as I'm not sure what are the implication in others UI windows.

newIDE/app/src/EffectsList/index.js
```diff
@@ -147,39 +147,49 @@ export default class EffectsList extends React.Component<Props, {||}> {
...
                     <MiniToolbar>
                       <Column expand noMargin>
-                        <SemiControlledTextField
-                          margin="none"
-                          commitOnBlur
-                          hintText={t`Enter the effect name`}
-                          value={effect.getName()}
-                          onChange={newName => {
-                            if (newName === effect.getName()) return;
+                        <Line>
```
4ian: This should probably be `<Line noMargin>` because vertical padding should be the responsibility of MiniToolbar already.

Also can you try to remove all Column/Line and see if it works?
```
<MiniToolbar>
  <MiniToolbarText>..</MiniToolbarText>
  < SemiControlledTextField >..</SemiControlledTextField >
  <MiniToolbarText>..</MiniToolbarText>
  < SemiControlledTextField >..</SemiControlledTextField >
```

Trying to see if we can simplify the markup

Bouh: Same thing with noMargin on <Line>.
Sightly better without <Column> and <Line> !

4ian: Great! 👍

4ian:
I think we could go ahead with this PR (run autoformatting)?
If you have more fixes you can always open another one later?

Bouh:
Ok

4ian:
Done! Thanks for these 👍

PR #1331, opened 2019-12-23
Make object creating and adding to a scene through JS less riskfull

arthuro555:
Automatically add the needed references to this._objects of runtimeScene etc. Needed to add the original objectData as attribute to add the object data to this._objects.

### DIFF
--- a/GDJS/Runtime/runtimebehavior.js
+++ b/GDJS/Runtime/runtimebehavior.js
@@ -15,12 +15,22 @@
  */
 gdjs.RuntimeBehavior = function(runtimeScene, behaviorData, owner)
 {
-    this.name = behaviorData.name || "";
+	this._behaviorData = behaviorData; // To be able to access it later
+	this.name = behaviorData.name || "";
     this.type = behaviorData.type || "";
     this._nameId = gdjs.RuntimeObject.getNameIdentifier(this.name);
     this._activated = true;
     this.owner = owner;
 };
+
+/**
+ * Get the behaviorData of the behavior.
+ * @return {string} The behavior Data.
+ */
+gdjs.RuntimeBehavior.prototype.getOriginalData = function() {
+	return this._behaviorData;
+};
+
 /**
  * Get the name of the behavior.
  * @return {string} The behavior's name.
--- a/GDJS/Runtime/runtimeobject.js
+++ b/GDJS/Runtime/runtimeobject.js
@@ -24,6 +24,7 @@
  */
 gdjs.RuntimeObject = function(runtimeScene, objectData)
 {
+    this._objectData = objectData;
     this.name = objectData.name || "";
     this._nameId = gdjs.RuntimeObject.getNameIdentifier(this.name);
     this.type = objectData.type || "";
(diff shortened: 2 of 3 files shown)


GDJS/Runtime/runtimebehavior.js
```diff
@@ -15,12 +15,22 @@
  */
 gdjs.RuntimeBehavior = function(runtimeScene, behaviorData, owner)
 {
-    this.name = behaviorData.name || "";
+	this._behaviorData = behaviorData; // To be able to access it later
```
arthuro555: I did that for more choerence because I had to do it in runtimeObject

GDJS/Runtime/runtimeobject.js
```diff
@@ -1456,5 +1457,13 @@ gdjs.RuntimeObject.getNameIdentifier = function(name) {
     return newIdentifier;
 };
 
+/**
+ * Get the objectData passed to the constructor
+ * @return {string} The object Data.
+ */
+gdjs.RuntimeObject.getOriginalData = function() {
```
arthuro555: Using that to add the data to runtimeScene._objects

GDJS/Runtime/runtimescene.js
```diff
@@ -521,6 +521,14 @@ gdjs.RuntimeScene.prototype.updateObjectsForces = function() {
  * @param obj The object to be added.
  */
 gdjs.RuntimeScene.prototype.addObject = function(obj) {
+    if ( !this._objects.containsKey(obj.name) ) {
```
arthuro555: If the object is not registered in scene register it using the same calls as in loadObjects

arthuro555:
I don't think the unit check fail is related to my pr

GDJS/Runtime/runtimescene.js
```diff
@@ -521,6 +521,14 @@ gdjs.RuntimeScene.prototype.updateObjectsForces = function() {
...
+        this._objects.put(obj.name, obj.getOriginalData());
```
4ian: Would it be possible to move this into a "registerObject" (or `loadObject`, still not sure) method? Passing the name, constructor, and objectData.

I'm still unsure about what's the ideal final form of this, but with a separate method we could: 
* avoid code duplication (this code and remove `loadObject` in `loadFromScene` in favor of this new method).
* could make clearer that there is a step of registering/loading needed before handling objects.

GDJS/Runtime/runtimescene.js
```diff
@@ -521,6 +521,14 @@ gdjs.RuntimeScene.prototype.updateObjectsForces = function() {
...
+        this._instances.put(obj.name, []); //Also reserve an array for the instances
+        this._instancesCache.put(obj.name, []); //and for cached instances
+        //And cache the constructor for the performance sake:
+        this._objectsCtor.put(obj.name, obj);
```
4ian: I'm not sure it's valid to add an instance of a gdjs.RuntimeObject (or a derived class) into a hashtable that should contain constructors. 

That's why I think we should remove this code out of `addObject`. Instead, let the caller code manually register the object with the name, constructor and objectData so it's clear and we're not abusing the _objectsCtor (which is used when creating objects from events).

4ian:
@arthuro555 I did post comments already :)

arthuro555:
Whoops, sorry, this was an accident, didn't mean to do that.

Arthur.

4ian:
Ok no worries :)

arthuro555:
Ok I will do this on another PR (as you probably see I messed this up and this is the master branch) so I'm closing this one

PR #1332, opened 2019-12-23
Add Input Functionalities

arthuro555:
To support mapping of strings to keys like other engines like unity or ue4 to simplify things like Key Mapping.

I have no React knowledge, so I cannot implement this but it would be nice to also implement a UI to define default maps (again like in unity and UE4)

### DIFF
--- a/Core/GDCore/Extensions/Builtin/KeyboardExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/KeyboardExtension.cpp
@@ -69,6 +69,61 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsKeyboardExtension(
       .AddParameter("string", _("Expression generating the key to test"))
       .MarkAsAdvanced();
 
+    extension
+      .AddCondition("InputMapped",
+                    _("Is Input mapped?"),
+                    _("Test if an Input was mapped"),
+                    _("Input _PARAM1_ is mapped"),
+                    _("Keyboard"),
+                    "res/conditions/keyboard24.png",
+                    "res/conditions/keyboard.png")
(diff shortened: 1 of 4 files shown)


4ian:
This looks very interesting! 

This is if I understand properly a "mapper" from arbitrary inputs to keys. I think it's a very interesting start. 

A few ideas that could maybe make this go a bit further, let me know what you think:
* On the "form" itself:
  * I think we could move this in a separate extension "InputMapper", that would make use of input manager like you did (and exposing actions/conditions as you did). "Advantage": you could also declare the extension in JS (JsExtension.js file), so no need to recompile GDevelop.js
* On inputs mapping:
  * I think we could also map mouse/touch to inputs? 
  * And to allow users to "map" anything that they would like by adding an action "Simulate input xxx to be pressed" (and other for releases).
* In inputs mapping too, as the goal is to make something that is "ready to use"/"battery included", I think we could have two actions to save/load the mappings into a variable structure (that can be then stored as JSON).

Finally we can indeed think about making a window in the UI to edit initial mappings (and maybe import/export them to JSON). I'm not to worried about this, it's a good exercice to learn React for you/someone else or something I can do if needed.

Bouh:
For the React panel with the keys to map them.
Can we avoid an ugly list of shortcuts, or an ugly panel.
There are [software](https://docs.unity3d.com/Packages/com.unity.progrids@3.0/manual/hotkeys.html) and [games](https://imgur.com/gallery/BhuFuu8) that now display a keyboard and it's visually cooler and simpler for people.

4ian:
Closing for now, let's reopen later as discussed in https://forum.gdevelop-app.com/t/request-input-map/22621/4 :)

PR #1334, opened 2019-12-24
Refactor/sentences

4ian:
I made a script that automatically update all the sentences for action/condition acting on a value (number or string) + additional changes to move from the "canonical":

> `Do [operator][value] to [subject] [of object]` (for actions) and `[value] [of object] is [operator] [value]` (for conditions)

to a more english readable:

> `Change [subject] [of object]: [operator][value]` (for actions) and `[value] [of object][operator] [value]` (for conditions)

As an example, instead of "Do +3 to the variable Test of MyObject", we now have "Change variable Test of MyObject: Add 3" (operators for actions are now shown in plain english).

This is following [this discussion](https://github.com/4ian/GDevelop/issues/927#issuecomment-567153253).

The reasons are to have:
* Proper english sentences
* Consistency between actions/conditions in the order of operator/values/subjects.

Another advantage is that this refactoring made the API simpler (cc @blurymind ;)) (could be made even more simple): 
image

Here are real examples to give you an idea:

Screenshot 2019-12-24 16 27 20
Screenshot 2019-12-24 16 27 27
Screenshot 2019-12-24 16 27 40
Screenshot 2019-12-24 16 27 01
Screenshot 2019-12-24 16 27 10
Screenshot 2019-12-24 16 26 27
Screenshot 2019-12-24 16 26 37

It should be much simpler to update this/translate this because **all the hardcoded sentences are gone** and **only the subjects** ("the width", "the height", "the x position") are now to be translated! So instead of translating two sentences (one for condition, one for  action), you only have to translate one word for both :)

Let me know what you think. The main downside of this will be to see how to update all the doc/tutorials for this.

### DIFF
--- a/Core/GDCore/Extensions/Builtin/AudioExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/AudioExtension.cpp
@@ -143,117 +143,97 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAudioExtension(
                  _("Volume of the sound on a channel"),
                  _("This action modifies the volume of the sound on the "
                    "specified channel. The volume is between 0 and 100."),
-                 _("Do _PARAM2__PARAM3_ to the volume of the sound on channel "
-                   "_PARAM1_"),
+                 _("the volume of the sound on channel _PARAM1_"),
                  _("Audio/Sounds on channels"),
                  "res/actions/sonVolume24.png",
                  "res/actions/sonVolume.png")
       .AddCodeOnlyParameter("currentScene", "")
       .AddParameter("expression", _("Channel identifier"))
-      .AddParameter("operator", _("Modification's sign"))
-      .AddParameter("expression", _("Value"))
-      .MarkAsAdvanced()
-      .SetManipulatedType("number");
+      .UseStandardOperatorParameters("number")
+      .MarkAsAdvanced();
 
   extension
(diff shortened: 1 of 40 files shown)


lunarovic:
Amazing work! So it turned out to be even easier to maintain. :) And easier to translate, which was another worry. I will happily use and recommend GDevelop from now on. :) Best regards.

Bouh:
Not fan about **set to XXX**
Prefer **= XXX** same for **substract** prefer use sign.

Sign are more visual than text.
Ok Gdevelop is a non programming language, but this is very verbose and not very fast to read.

4ian:
> With new build this sentence is not very neat

Right there is and maybe a few others that I need to manually tweak.
We can give an option to use programming like operators (+ - / *).

Merry Christmas to all :) 🎄

blurymind:
Thank you @4ian merry Christmas!!
:)

4ian:
@Bouh Are you ok with this if we have an option to display +=, -=, /=, *=, = instead of add/subtract/divide by/multiply by/set to?

Bouh:
@4ian  yes absolutely

4ian:
Merged! This should give a much more polished look to the Events Sheets, and should be a bit more intuitive to people reading events for the first time :)

blurymind:
Thank you again @4ian :+1: this is fantastic :smiley_cat:

PR #1159, opened 2019-07-23
Add moving buttons for order parameters in functions

Bouh:
This is in link with [this card on Trello](https://trello.com/c/u9fVjejH/336-move-the-order-of-parameters-in-functions)

I've follow your instruction, i can't compile again the gd.js i got errors and it's bored to search how fix it.
Compile and install the environement for gd.js is a bit hard on Windows10, we have errors everywhere.

PR #1340, opened 2019-12-28
Moving order parameters functions #2

Bouh:
Continued from #1159 
The branch was "unknown repository" I couldn't get it back.

I don't understand all the links between the different files and how they work.

If you want to finish it go ahead, I tried but I'm lost.
I tried to follow the instructions stupidly, I don't really know what I did.


4ian:
Sure no worries I'll finish this at some point :)

Bouh:
It's ready to merge or c++ functions are missing ?
Since it's been open for so long, there are now conflicts :/

Bouh:
And tada!

This works, i can move up and down parameters, save the project and reopen the function parameters with the change previously saved.

This is ready for final review! (finally)

4ian:
Oh nice! Was not expecting this come back 👀
Have you checked on a few examples that the reordering is working properly? There are basic internal tests for this in `Core/tests/WholeProjectRefactorer.cpp`, so it should work properly, but it was never tried "for real".

Apart from that it looks good!

newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
```diff
@@ -39,6 +39,8 @@ type Props = {|
   onParametersUpdated: () => void,
   helpPagePath?: string,
   freezeParameters?: boolean,
+  makeMoveFreeEventsParameter?: boolean,
+  makeMoveBehaviorEventsParameter?: boolean,
```
4ian, line 43: Seems like these are never used? 🤔

Also, don't pass makeXXX function, pass instead the created functions:
`onMoveFreeEventsParameter={makeMoveFreeEventsParameter(i18n)}`

A `makeXXX` function mean "I'm a function that you should call with my argument (18n) so that I send you another function that you can use (here, pass as a prop, so that the children don't care about i18n)".

newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
```diff
@@ -325,6 +370,16 @@ export default class EventsFunctionParametersEditor extends React.Component<
                               ),
                               click: () => this._removeLongDescription(i),
                             },
+                            {
+                              label: i18n._(t`Move up`),
+                              click: () => this._moveParameters(i, i - 1),
+                              enabled: i - 1 >= 0,
```
4ian, line 376: Add also a check for `!isParameterDisabled(i)`: `!isParameterDisabled(i) && i - 1 >= 0`

newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
```diff
@@ -325,6 +370,16 @@ export default class EventsFunctionParametersEditor extends React.Component<
...
+                            },
+                            {
+                              label: i18n._(t`Move down`),
+                              click: () => this._moveParameters(i, i + 1),
+                              enabled: i + 1 < parameters.size(),
```
4ian, line 381: Add also a check for `!isParameterDisabled(i)`

newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
4ian: Actually, we should also check if the previous parameter is not disabled :) 
`!isParameterDisabled(i) && i - 1 >= 0 && !isParameterDisabled(i - 1)`

newIDE/app/src/EventsFunctionsExtensionEditor/index.js
```diff
@@ -301,6 +301,48 @@ export default class EventsFunctionsExtensionEditor extends React.Component<
     done(true);
   };
 
+  _makeMoveFreeEventsParameter = (i18n: I18nType) => (
+    eventsFunction: gdEventsFunction,
+    oldIndex: number,
+    newIndex: number,
+    done: boolean => void
+  ) => {
+    //TODO add verifications and warning box
```
4ian, line 310: Let's replace this TODO by `// Don't ask for user confirmation as this change is easy to revert.`

newIDE/app/src/EventsFunctionsExtensionEditor/index.js
```diff
@@ -301,6 +301,48 @@ export default class EventsFunctionsExtensionEditor extends React.Component<
...
+  };
+
+  _makeMoveBehaviorEventsParameter = (i18n: I18nType) => (
+    eventsBasedBehavior: gdEventsBasedBehavior,
+    eventsFunction: gdEventsFunction,
+    oldIndex: number,
+    newIndex: number,
+    done: boolean => void
+  ) => {
+    //TODO add verifications and warning box
```
4ian, line 331: Same, let's replace the TODO as before.

Bouh:
There is a strange behavior when the first parameter move at the second place. the content is empty, and when i revert it the value come back.
https://streamable.com/vate9l

The limitation for move up and down are ok.
image

4ian:
> There is a strange behavior when the first parameter move at the second place. the content is empty, and when i revert it the value come back.

I think there might be two issues. First, we don't update the `_PARAM1_`, `_PARAM2_` in the sentence :) We'll probably need to do that too.
Secondly, if you do that manually (fix the `_PARAM1_`, `_PARAM2_` manually after moving the parameter), is there still an issue?

Bouh:
I've moved param1 to down.
[before_after.zip](https://github.com/4ian/GDevelop/files/5776928/My.project25.zip)

Seem param0 is now filled, you can see it if you add _PARAM0_ in sentence.

4ian:
> Seem param0 is now filled, you can see it if you add PARAM0 in sentence.

Indeed, this can be verified in the JSON:
image

The problem is that there is, for free functions only, an implicitly added, "code only" parameter, the runtime scene.
It's not shown in the list of parameters of the function... but it's added in the list of parameters when the extension is generated.

To fix this, add a `+ 1` to the old and new index in `_makeMoveFreeEventsParameter`... not super elegant but the reasoning is that the call to `moveEventsFunctionParameter` is done on the generated extension, so the runtime scene argument will be there.
You can help the future reader by putting, instead of `+ 1`, `+ getParametersIndexOffset(false)`. The `getParametersIndexOffset` method is there to indicate that there is an extra hidden argument that will be added.

4ian:
When it's done, give it a try with a function with 2 or 3 arguments, same for a behavior method, and that should be good :)

Bouh:
> When it's done, give it a try with a function with 2 or 3 arguments, same for a behavior method, and that should be good :)

It's just like you said 👍

4ian:
Looks good! There was a few flow fixes that I added.

(I've pushed some fixes including a merge from master, hence all the commits, don't panic).

4ian:
Note that the sentence of the function is not updated, we should probably udpdate it to swap the `_PARAMx_`

PR #1341, opened 2019-12-28
Pixi-filters pt2

blurymind:
Adds
 rgbSplit, outline, glow, dropShadow, bevel, ascii, adjustment filters

gd-newFilters


4ian:
Looks great! I've done a few updates to the description so that they are a bit more like explanations rather than just quoting the effect name again.
Merging this now :) Thanks for adding these! These new effects open interesting new possibilities 👍

blurymind:
Thank you @4ian :)
I wasn't sure what to put for description as the names were self
explanatory.

This templating approach was an interesting experiment for me and saved a
lot of time in this case. I am using handlebars.js to generate the js code
and am wondering if there is a better templating library out there for
this. It would be worth polishing the templating app some more before
putting it out there. Not sure what to call it thb. It's json editor + Ace
editor + diff Ace editor + handlebarsjs. Boilerplater? Template mate? xD

PR #1366, opened 2020-01-12
Conscistent font-weight on object name

Bouh:
In light theme, object parameters are bolder than in dark theme.
So italic is remove and bold is added in dark theme.


4ian:
Sounds good!

PR #1373, opened 2020-01-14
[WIP] Add Typedefs for GDJS

arthuro555:
Already done for gd.js and RuntimeObject.js (the last one will actually need to be updated when typedefs for others will arrive. Made TODOs for thoose Places)

WIP.


arthuro555:
Wait what this is super weird why does it shows old already merged commits

arthuro555:
let me try again

PR #1374, opened 2020-01-14
[WIP] Add Beginning of type definitions

arthuro555:
Already done for gd.js and RuntimeObject.js (the last one will actually need to be updated when typedefs for others will arrive. Made TODOs for thoose Places)

WIP.

### DIFF
--- a/GDJS/Runtime/gd.d.ts
+++ b/GDJS/Runtime/gd.d.ts
@@ -1,0 +1,68 @@
+// Type definitions for gd.js
+// Project: GDJS Platform of GDevelop
+// Definitions by: arthuro555 <https://forum.gdevelop-app.com/u/arthuro555>
+
+declare class Hashtable {
+    public items: object;
+    public constructor();
+    public static newFrom(items:object) : Hashtable;
+    public put(key:any, value:any) : void;
+    public get(key:any) : any;
+    public containsKey(key:any) : boolean;
+    public remove(key:any) : void;
+    public firstKey() : any;
+    public keys(result:Array<any>) : void;
+    public values(result:Array<any>) : void;
+    public clear() : void;
+}
+
+interface Array<T> {
+    remove(from: number) : void;
+    createFrom(arr: Array<any>) : void;
+}
+
+declare namespace gdjs {
+    // Register default types or types used in many GDJS files here, else extend this in the file specific definition file.
+    namespace types {
+        type Point = {
+            x: number
+            y: number
+        }
+        // Will change
+        type RaycastTestResult = {
+            collision: false
+            closeX: number
+            closeY: number
+            closeSqDist: number
+            farX: number
+            farY: number
+            farSqDist: number
+        }
+    }
+    namespace evtTools {}
+    let objectsTypes: Hashtable;
+    let behaviorsTypes: Hashtable;
+    let callbacksRuntimeSceneLoaded:Array<any>;
+    let callbacksRuntimeSceneUnloaded:Array<any>;
+    let callbacksRuntimeScenePaused:Array<any>;
+    let callbacksRuntimeSceneResumed:Array<any>;
+    let callbacksObjectDeletedFromScene:Array<any>;
+
+    function rgbToHex(r: number, g: number, b: number) : string;
+    function rgbToHexNumber(r: number, g: number, b: number) : number;
+    function random(max: number) : number;
+    function randomInRange(min:number, max: number) : number;
+    function randomFloat(max: number) : number;
+    function randomFloatInRange(min:number, max: number) : number;
+    function randomWithStep(min:number, max: number, step:number) : number;
+    function toRad(angleInDegrees:number) : number;
+    function toDegrees(angleInRadians:number) : number;
+    function registerObjects() : void;
+    function registerBehaviors() : void;
+    function getObjectConstructor(name: string) : any;
+    function getBehaviorConstructor(name: string) : any;
+    function staticArray(owner: any) : Array<any>;
+    function staticArray2(owner: any) : Array<any>;
+    function staticObject(owner: any) : object;
+    function objectsListsToArray(objectsLists: { values: (arg0: any[]) => void; }) : Array<any>;
+}
--- a/GDJS/Runtime/gd.js
+++ b/GDJS/Runtime/gd.js
@@ -1,3 +1,5 @@
+/// <reference path="gd.d.ts"/>
+// @ts-check
 /*
  * GDevelop JS Platform
  * Copyright 2013-2016 Florian Rival (Florian.Rival@gmail.com). All rights reserved.
@@ -15,8 +17,9 @@ window.gdjs = {
    * Contains functions used by events (this is a convention only, functions can actually
    * by anywhere).
    * @namespace
-   * @memberof gdjs
+   * @memberOf gdjs
    */
+  // @ts-ignore
   evtTools: {},
   callbacksRuntimeSceneLoaded: [],
   callbacksRuntimeSceneUnloaded: [],
@@ -104,7 +107,7 @@ gdjs.registerObjects = function() {
 
   for (var p in this) {
     if (this.hasOwnProperty(p)) {
-      if (gdjs[p].thisIsARuntimeObjectConstructor != undefined) {
+      if (gdjs[p].thisIsARuntimeObjectConstructor !== undefined) {
         gdjs.objectsTypes.put(gdjs[p].thisIsARuntimeObjectConstructor, gdjs[p]);
       }
     }
@@ -129,7 +132,7 @@ gdjs.registerBehaviors = function() {
     if (this.hasOwnProperty(gdjsProperty)) {
       // Search in object inside gdjs.
       var innerObject = gdjs[gdjsProperty];
-      if (innerObject.thisIsARuntimeBehaviorConstructor != undefined) {
+      if (innerObject.thisIsARuntimeBehaviorConstructor !== undefined) {
         gdjs.behaviorsTypes.put(
           innerObject.thisIsARuntimeBehaviorConstructor,
           innerObject
@@ -146,7 +149,7 @@ gdjs.registerBehaviors = function() {
             if (
               innerInnerObject !== null &&
               typeof innerInnerObject === 'function' &&
-              innerInnerObject.thisIsARuntimeBehaviorConstructor != undefined
+              innerInnerObject.thisIsARuntimeBehaviorConstructor !== undefined
             ) {
               gdjs.behaviorsTypes.put(
                 innerInnerObject.thisIsARuntimeBehaviorConstructor,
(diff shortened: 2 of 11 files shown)


arthuro555:
Just to be clear, this is not ready to push. Well, it theoretically is, but I want the most basic stuff done before pushing. This is just here as "preview" of my work

GDJS/Runtime/runtimeobject.js
```diff
@@ -1097,7 +1111,7 @@ gdjs.RuntimeObject.prototype.getTimerElapsedTimeInSeconds = function(timerName)
 
 /**
  * Separate the object from others objects, using their hitboxes.
- * @param objects Objects
+ * @param objects Objects {@link Array<RuntimeObject>}
```
4ian: Why a link? Should that bit a type instead?

GDJS/Runtime/runtimeobject.js
```diff
@@ -970,11 +984,11 @@ gdjs.RuntimeObject.prototype.hasBehavior = function(name) {
  * De/activate a behavior of the object.
  *
  * @param name {String} The behavior name.
```
4ian: should be fixed to `{string} name`

GDJS/Runtime/runtimeobject.js
```diff
@@ -970,11 +984,11 @@ gdjs.RuntimeObject.prototype.hasBehavior = function(name) {
...
- * @param enable {boolean} true to activate the behavior
+ * @param status {boolean} Set to true for enabled. False for Disabled.
  */
-gdjs.RuntimeObject.prototype.activateBehavior = function(name, enable) {
+gdjs.RuntimeObject.prototype.activateBehavior = function(name, status) {
```
4ian: please keep enable, status is too vague. enable is more descriptive for a boolean.

GDJS/Runtime/runtimeobject.js
```diff
@@ -970,11 +984,11 @@ gdjs.RuntimeObject.prototype.hasBehavior = function(name) {
  * De/activate a behavior of the object.
  *
  * @param name {String} The behavior name.
- * @param enable {boolean} true to activate the behavior
+ * @param status {boolean} Set to true for enabled. False for Disabled.
```
4ian: name and type are still inverted ;)

GDJS/Runtime/runtimeobject.js
```diff
@@ -574,7 +588,7 @@ gdjs.RuntimeObject.prototype.hide = function(enable) {
 /**
  * Return true if the object is not hidden.
  *
- * Note: This is unrelated to the actual visibility of the objec on the screen.
+ * @note This is unrelated to the actual visibility of the objec on the screen.
```
4ian: @note is rejected by jsdoc, from what I remember.

GDJS/Runtime/runtimeobject.js
```diff
@@ -471,12 +485,12 @@ gdjs.RuntimeObject.prototype.getVariableNumber = gdjs.RuntimeObject.getVariableN
  * Only for usage by events.
  *
  * @param {gdjs.Variable} variable The variable to be accessed
- * @return The specified variable
+ * @return gdjs.Variable specified variable
```
4ian: {gdjs.Variable} The specified variable

GDJS/Runtime/runtimeobject.js
```diff
@@ -218,6 +231,7 @@ gdjs.RuntimeObject.prototype.onDestroyFromScene = function(runtimeScene) {
  * @return {Object} The internal rendered object (PIXI.DisplayObject...)
  */
 gdjs.RuntimeObject.prototype.getRendererObject = function() {
+    return false;
```
4ian: `return null;`

Never return a boolean instead of an object. `null` is acceptable for objects.

GDJS/Runtime/runtimebehavior.js
```diff
@@ -5,9 +5,9 @@
  */
 
 /**
- * @typedef BehaviorData Properties to set up a behavior.
+ * @typedef BehaviorData The behavior type. Used by GDJS to find the proper behavior to construct.
```
4ian: Be careful seems that you do not started your work from `master` so you have commits that are undoing what was previously fixed.

GDJS/Runtime/gd.js
```diff
@@ -179,7 +182,7 @@ gdjs.registerGlobalCallbacks = function() {
 
   var totalprop = 0;
 
-  innerRegisterGlobalCallbacks = function(obj, nestLevel) {
+  const innerRegisterGlobalCallbacks = function(obj, nestLevel) {
```
4ian: Use var, not const, because of Android 4.x support: https://github.com/4ian/GDevelop/blob/master/newIDE/docs/Supported-JavaScript-features-and-coding-style.md

GDJS/Runtime/runtimeobject.js
arthuro555: Sorry, I just made that as a little note to myself. This was really for the type definitions and not much for the JSDoc. I planned to do the JSDoc in a other pr

GDJS/Runtime/gd.d.ts
```diff
@@ -0,0 +1,68 @@
+// Type definitions for gd.js
+// Project: GDJS Platform of GDevelop
+// Definitions by: arthuro555 <https://forum.gdevelop-app.com/u/arthuro555>
+
+declare class Hashtable {
```
4ian: While I'm not sure we should write and maintain these types for other classes because we probably can generate them, Hashtable might be an exception because ideally we want this to be a generic type: `Hashtable<gdjs.RuntimeObject>` for example.

GDJS/Runtime/gd.d.ts
```diff
@@ -0,0 +1,68 @@
+// Type definitions for gd.js
+// Project: GDJS Platform of GDevelop
```
4ian: nitpicking: GDJS Platform of GDevelop means "GDevelop JavaScript Platform of GDevelop" ;)
Prefer "GDevelop JavaScript Runtime" or "GDevelop JavaScript Platform".

arthuro555: I know that, but What I meant is the GDJS component of the GDevelop project

arthuro555:
Weird the ci has 3 ts errors but I had none when I checked locally

4ian:
Still a good thing that you started this PR so I can take a look early :) 

My main concern is about manually writing .d.ts files: 

* Do we need these .d.ts files at all? I think it might be interesting to generate them if for some reason an external project need them, but for now I don't think it's the case?
* Instead, can we keep typing stuff using JSDoc annotations that are compatible with TypeScript (should be good for 90% of the cases) and add a .d.ts file for the 10% that are better typed directly using Typescript?

I'm particulary worried about having to maintain two set of files. It's too easily to get them out of sync, forget to add a method in the .d.ts, and it's a real pain to have to write a header in addition to the normal JS file for the implementation. It's slowing down when writing new features and goes against the type being helpful (types are helpful when they are not too visible/not needing too much manual work, but still give us confidence the code is 100% valid).
A big advantage of TypeScript for me is the ability to already understand some of what you've done without having to manually write everything like in older statically typed languages :) 

What do you think? Can we just continue typing as before, running tsc (with `npm run check-types` in GDJS) to ensure TypeScript coverage? 
Another advantage is that we can almost make a PR for each file or each couple of file that we type: 
* Add // @ts-check at the top,
* Fix the typings,
* Optionally maybe fix/type related types that are not typed yet.
* It's ready for a PR :)
* Next!

4ian:
> Weird the ci has 3 ts errors but I had none when I checked locally. Errors are errors nonetheless so it's fixing time 😭 why do I always create more errors than I try to fix

How are you checking errors locally? `npm run check-types`? Because we're still on TypeScript 3.6.4, not 3.7 which is the latest.

Note that we want to move to 3.7 at some point to benefit from auto-generation of .d.ts: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#--declaration-and---allowjs that I was talking about 😊

arthuro555:
You missunderstood the d.ts generation, it's made for .ts files, it's an argument that instead of just compiling such files also make type definitions to keep the typing in the releases. I sure think it could potentially become out of sync and that is not a viable solution, the best in my opinion would be to directly switch the engine from Js to ts and "compile" it before importing into the editor, as then type definitions would be directly in the Js file, and if some are missing it won't break the whole thing,as typing in ts is optional. Should I go for such an approach instead? Do you think this PR should be reported to after correcting the majority of the JSDoc?

Arthur.

4ian:
> the best in my opinion would be to directly switch the engine from Js to ts and "compile" it before importing into the editor

I'm not sure this would give us any benefit in terms of typing compared to what we have right now (as we can already type variables and most stuff with special comments). It gives us a nicer syntax but also cost us having to run typescript and a watcher to compile files. For now I would prefer to avoid - typing is a separate issue to having a bundler.

> type definitions would be directly in the Js file, and if some are missing it won't break the whole thing

That's what we already have no?

> Do you think this PR should be reported to after correcting the majority of the JSDoc?

I don't see the difference, well depends on what you mean by fixing the majority of JSDoc. If it's:
* Fixing or adding comments, feel free to create PRs with fixes :)
* Fixing type annotations or adding them, same feel free to create PRs with fixes/new type annotations but always by entering them using *comments* (aka "JSDoc compatible" annotations), ensuring that both typescript understand it (`npm run check-types`) and that JSDoc is ok with them (run the documentation generation to check it's working).

----

Just to be clear, when I say typing "using comments", I refer to that: https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html

in other words, this means that instead of writing TS, like "function test(a: number): string { ... }", we write:
```
/** 
 * @param {number} a
 * @returns {string}
 */
function test(a) { ... }
```

which is a bit longer but not that much considered that most functions should be documented with a description of the function and its parameters anyway. And this gives us the benefit of having raw, untransformed JavaScript so no need for a bundler and no risk of transformation that would create issues like putting pressure on the garbage collector with tons of temporary object creation.

Does that make sense? 
Again in case of question take a look at https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html which seems fairly complete :) The only gotcha seems to be that the keyword "extends" is not properly supported. That's ok, let's forget about this for now.
 
I will also at some point enable `strictNullChecks` so that Typescript errors if you try to return null in an object that is not nullable (so write a question mark to mark something that can be null: `?MyType`, `?number`, etc...).

arthuro555:
Ok, so you don't want type script at all for the moment and only use the compiler for JSDoc type checking? In that case should I delete this PR or close it?

Arthur.

4ian:
We don't want TypeScript (.ts) files at the moment but only JS files documented with JSDoc and type checked with [TypeScript with JS doc annotation]( https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html).

I want to be clear that JSDoc is *not* doing any typechecking :) 

You can either close the PR or rework it to remove all the .d.ts file and keep only the typings.

PR #1376, opened 2020-01-15
Add JSDoc to runtimeObject

arthuro555:
And add typing to it's properties. 
*And also moved AABB typedef with the others at the top

If it's accepted(merged), I will add an extension to use it inside GDevelop and serialization to every extension's objects


4ian:
Would you be able to move the typings to another PR so we can merge it?

For the serialization itself, I'm not sure where exactly you want to head that? I think it could be cool if you open a thread on the forum or a GitHub issue so we can discuss what you have in mind and if we can find alternatives or smart ideas to do whatever you're trying to achieve :)

arthuro555:
Ok removed serialization will work on that later. I resolved the merge conflict so it should be ready to merge.

arthuro555:
Bump. I repeat, this is read to merge. Could you review again please @4ian? Thank you :smile: !

GDJS/Runtime/runtimeobject.js
```diff
@@ -34,27 +41,64 @@
...
     this.zOrder = 0;
+    /** @type {boolean} */
     this.hidden = false;
+    /** @type {string} */
     this.layer = "";
+    /** @type {boolean} */
     this.livingOnScene = true;
+    /**
+     * @type {number}
+     * @note No serialization as used as internal value that should not be changed by deserialization.
```
4ian: Don't use @note at all, it's not supported by JSDoc

GDJS/Runtime/runtimeobject.js
```diff
@@ -34,27 +41,64 @@
...
+     */
     this.id = runtimeScene.createNewUniqueId();
-    this._runtimeScene = runtimeScene; //This could/should be avoided.
+    /**
+     * @type {gdjs.RuntimeScene}
+     * @note This could/should be avoided.
+     * @note No serialization as runtimeScene would serialize objects as children and we don't want circular references (and to be able to deserialize on other scenes).
```
4ian: Can you remove all notes related to serialization?

GDJS/Runtime/runtimeobject.js
```diff
@@ -34,27 +41,64 @@
  * @param {ObjectData} objectData The initial properties of the object.
  */
 gdjs.RuntimeObject = function(runtimeScene, objectData) {
+    /** @type {string} */
```
4ian: Can you add protected to all the `this.xxx` members of RuntimeObject? None of these should appear in the documentation as public because they are internal state of the object and should not be modified from the outside.

GDJS/Runtime/runtimeobject.js
```diff
@@ -34,27 +41,64 @@
...
     this.livingOnScene = true;
+    /**
+     * @type {number}
+     * @note No serialization as used as internal value that should not be changed by deserialization.
+     */
     this.id = runtimeScene.createNewUniqueId();
-    this._runtimeScene = runtimeScene; //This could/should be avoided.
+    /**
+     * @type {gdjs.RuntimeScene}
+     * @note This could/should be avoided.
```
4ian: You can remove "This could/should be avoided."

GDJS/Runtime/runtimeobject.js
```diff
@@ -104,6 +169,10 @@ gdjs.RuntimeObject = function(runtimeScene, objectData) {
 
     //Timers:
     if (this._timers === undefined)
+        /**
+         * @type {Hashtable<string, gdjs.Timer>}
```
4ian: This is what should be the proper type but Hashtable is not a generic type so Typescript will probably yell at this if we add `// @ts-check`. Let's either use Hashtable, or declare a generic type for Hashtable (I hope it can be done and compatible with JSDoc)

GDJS/Runtime/runtimeobject.js
```diff
@@ -7,13 +7,20 @@
...
+ * @typedef {Object} ObjectData Object containing initial properties for all objects extending {@link gdjs.RuntimeObject}.
  * @property {string} name The name of the object. During the game, objects can be queried by their name (see {@link gdjs.RuntimeScene.prototype.getObjects} for example).
  * @property {string} type The object type.
  * @property {Array<VariableData>} variables The list of default variables.
  * @property {Array<BehaviorData>} behaviors The list of default behaviors.
  */
 
+/**
+ * @typedef {Object} AABB
+ * @property {Array} min The [x,y] coordinates of the top left point
```
4ian: `Array<number>`

GDJS/Runtime/runtimeobject.js
4ian: See if you can generate the documentation for GDJS:

https://github.com/4ian/GDevelop/blob/7c9e6ee6f135f0b526f89f9ebec903868bab0249/scripts/GenerateAllDocs.sh#L19-L21

and look at what is generated for RuntimeObject.

4ian:
Done! Sorry it slipped off my mind.

GDJS/Runtime/runtimeobject.js
4ian: (and below too)

GDJS/Runtime/runtimeobject.js
arthuro555: @note is made for notes relevant to the code but not the documentation, so it's ignored by purpose by JSDoc

GDJS/Runtime/runtimeobject.js
arthuro555: shouldn't protected vars begin with an underscore?

GDJS/Runtime/runtimeobject.js
arthuro555: Well, JSDoc supports generics, I used them, only to realise that typescript doesn't support type checking from JSDocs generics and makes an error for each use of hashtable in the code

GDJS/Runtime/runtimeobject.js
4ian: Have you tried running JSDoc with these? From what I recall, @note is rejected as not recognised.

arthuro555: No, but whatever, thoose notes are not relevant and removing them is not bad

4ian: Yes that's why I think we should both remove them and not use @note at all.

GDJS/Runtime/runtimeobject.js
4ian: In this case let's avoid this for now, we can deal with Hashtable in a separate PR if you wish. Maybe take a look at @template: https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html#template - not sure this can be used properly though.

GDJS/Runtime/runtimeobject.js
4ian: In theory yes, in practice they have been named like this before such convention was properly used in the codebase. Ideally we could rename them, but this is a more risky refactoring to be done.

GDJS/Runtime/runtimeobject.js
arthuro555: The link you sent was what I was using as documentation. I think the only way would be to add a typescript type definition instad of JSDoc definitions. You said you don't want them, but that was because we could use JSDoc instead, that also handles documentation. Could we use Typescript typedefs just for that specific case?

4ian: > Could we use Typescript typedefs just for that specific case

Yes, if it does not break documentation generation.
If this is too painful we'll switch to typedoc (https://typedoc.org/) that should be hopefully 100% compatible with Typescript.

arthuro555: Nevermind I am stupid. This would work for typescript but wouldn't allow using Hastable<> syntax in JSDoc comments so I will have to leave that out for the moment.

4ian: I'm still hoping we can keep JSDoc as much as possible, because it is quite widespread in the community and battle tested :) 
Annoyingly it's rejecting sometimes some type annotations that are too complex (for example, a function type like `(param: number) => string` is not properly parsed).

GDJS/Runtime/runtimeobject.js
```diff
@@ -1107,7 +1170,7 @@ gdjs.RuntimeObject.prototype.separateFromObjects = function(objects, ignoreTouch
 
    //Check if their is a collision with each object
    for(var i = 0, len = objects.length;i<len;++i) {
-       if ( objects[i].id != this.id ) {
+       if ( objects[i].id != this.getUniqueId() ) {
```
4ian: - Might be a micro optimization, but I prefer to keep `this._id` instead of `this.getUniqueId()` to avoid paying the cost to calling a function so often. In compiled language, such a simple function might be inlined/optimized out and hopefully that's also done by JS engines, but I would rather use `this._id` to be sure.

GDJS/Runtime/runtimeobject.js
```diff
@@ -34,27 +41,64 @@
...
+    /** 
+     * @type {boolean} 
+     * @protected
+     */
+    this._livingOnScene = true;
+    /**
+     * @type {number}
+     * @protected
+     */
+    this._id = runtimeScene.createNewUniqueId();
```
4ian: This seems dangerous to refactor thits to _id in this PR. There is a regression in this file (see my other comment).
`id` is also from what I recall used sometimes outside of the class so I think if we want to refactor this, this has to be made very carefully (i.e: checking the whole extension and game engine for all `.id` references).

If there is just a single one `.id` that you forget to rename, this will be a regression and I may have to revert the whole PR => better make a PR first without this to get typings and *then* do the refactoring.

Typescript will help up to do so, by screaming in case something was forgotten to be renamed. But until we type most of the codebase, this is a very dangerous operation (I'm not even mentioning games that might use `id` already in their JS code).

GDJS/Runtime/runtimeobject.js
```diff
@@ -1145,7 +1208,7 @@ gdjs.RuntimeObject.prototype.separateFromObjectsList = function(objectsLists, ig
 
             //Check if their is a collision with each object
             for(var i = 0, len = objects.length;i<len;++i) {
-                if ( objects[i].id != this.id ) {
+                if ( objects[i].id != this.getUniqueId() ) {
```
4ian: `id` was not renamed here. But as I said, I strongly strongly suggest you don't change this now. 

Let's go step by step ;) Safety first with types. Refactoring can be done later.
Users won't care if id is called `_id` or `id`, but they will care about a regression.

GDJS/Runtime/runtimeobject.js
arthuro555: That may be weird but it seems like using a getter function is faster: https://jsperf.com/getter-performance

4ian: Well these are a micro benchmark :) 
I would be cautious about inferring anything from them because the difference seems very very small and they might not apply in a larger context where the JS engine has the optimize dozens of object types, passed through functions etc... at the end a simple property has more chance to be seen as directly addressable by the JS engine rather that something hiding behind functions.

Anyway, it's unlikely to make a change. :) I could be ok with this but let's not do this for now in this PR.

GDJS/Runtime/runtimeobject.js
```diff
@@ -217,6 +286,7 @@ gdjs.RuntimeObject.prototype.onDestroyFromScene = function(runtimeScene) {
  * @return {Object} The internal rendered object (PIXI.DisplayObject...)
  */
 gdjs.RuntimeObject.prototype.getRendererObject = function() {
+    return {};
```
4ian: `return undefined`

Returning an empty object is almost always a bad idea as it creates problem down the road (the function getting it can easily check for undefined/null with a if. If you return an empty object, it thinks that it got a valid object, because in JS an empty object evaluates to true, so it continue and boom, the code crash later in an unrelated function).

4ian: The exception is of course if the object represents some kind of empty map, but this is not the case here. Either we return a valid PIXI.DisplayObject (or something else depending on the renderer) or we return undefined or null (I would tend to favor `null` in general, but here you can use undefined to be sure to be 100% compatible with the old version).

arthuro555:
I believe it should be ready to merge now?

4ian:
Alright, looks good! Thanks for working on this :) 👍

Issue #1372, opened 2020-01-14
Better access to scene/object/global variables in js blocks


Level0gamedev:
## Description
The current way of accessing variables via js is cumbersome and far from accessible to new programmers. If someone would like to take a jump from visual programming to java script, GD5 would feel quite daunting. For example, to get a global variable currently, a user has to do the following:
```
let newVar = runtimeScene
  .getVariables()
  .get("_varValue")
  .getAsNumber();
```

While the same can be achieved by using `GlobalVariable(_varValue)` inside the visual programming when using the **Σ123** button.

## Solution suggested
Make the functions accessible/generated via the **Σ123** accessible in the js blocks.

arthuro555:
I think this is a bad idea. This would mean make functions and runtimeGame (to access the variables) globally accessible. This would make cheating easier, and also the current method has a coherent structure. In my opinion, if anything like this would be implemented, it would be a direct shortcut function getVariableValue(variableName) in 
  -> RuntimeGame
  -> RuntimeScene
  -> RuntimeObject
That does following:
  1. Gets the variable using this.getVariables().get(variableName)
  2. Determines it's type and use the correct function to access it's value
  3. Return the value

Level0gamedev:
Even if I value accessibility over any concerns about "cheaters" in game, I am ok a direct shortcut function. As long as it is not cumbersome to call variables from within js block I am happy. 

Also, @arthuro555 created a quick extension for this behavior on GD5 discord. This could be incorporated into extension repo and/or be the base of implementation of this enhancement.

4ian:
The extension is nice to use, but remember that any code is a liability => events are guaranteed to work on newer versions, but if you have JS code you have to update it if functions have been changed :) 

In particular I've seen that the code in Discord is using `_str` => this is dangerous, this is a private member and should never be used outside of gd.Variable. At any time this can be refactored/renamed and disappear on next version of GDevelop.
Never ever use something that starts with an underscore. At best you're likely that it works, at worse you might be breaking some internal state and then reporting bug that should not even exist.

So please please please always use the public interface, which are the documented function not starting with an underscore. In your case, there is something called `isNumber()`. It's made for this! :)

Thanks!

arthuro555:
[...]

PR #1377, opened 2020-01-15
Add shortcuts to get variables

arthuro555:
As described in #1372

### DIFF
--- a/GDJS/Runtime/runtimegame.js
+++ b/GDJS/Runtime/runtimegame.js
@@ -519,3 +519,17 @@ gdjs.RuntimeGame.prototype.stopCurrentSceneProfiler = function() {
 
   currentScene.stopProfiler();
 };
+
+/**
+ * Shortcut to get a Game's Variable's value.
+ *
+ * @param {string} variableName
+ * @returns {string | number} The value of the variable
+ */
+gdjs.RuntimeObject.prototype.getGameVariable = function(variableName) {
+  var variable = this.getVariables().get(variableName);
+  if (variable.isNumber()){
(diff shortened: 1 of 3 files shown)


4ian:
Unfortunately I'm not convinced by the need for these functions:
* They sure are "easier" to type, but not that much and are not typed as well (they return a string OR a number - and you don't know which for sure)
* They don't support structures.
* They are not used anywhere, so if they break nobody will know until we release and someone complains

More generally, we're increasing the API surface of GDJS, which means more maintenance (so less agility, more things to refactor), more potential bugs, more potential confusion (why is there two ways to get variables? Why is one not returning my structure but an empty string?).

Last thing: it might be true that typing runtimeScene.getVariables().get("MyVar").getAsNumber() is a bit long but:
* this forces you to prefer object variables, i.e: variables that are as narrowly scoped as possible.
* You can access global variables by doing `getGame()`. The extra getGame is actually a good thing to make you think "I'm going to access a global variable, should I REALLY do this?"
* `getAsNumber` is very clear and will automatically convert the variable on the fly if needed, as done in GDevelop.

I hope you understand my point of view :) Sorry if this is a bit frustrating, I understand but prefer to stay lean on the API. Here I'm basically fighting the effect of [code bloat](https://en.wikipedia.org/wiki/Code_bloat).

arthuro555:
I have no problems with it, I have done that in literally one minute for issue #1372 but I personally thought that it wasn't a good idea (see my Opinion on gd5-beta on the discord and on the issue) too so I'm happy this will not be a thing.

PR #1378, opened 2020-01-15
[REFACTOR] Refactor SpriteRuntimeObject.js typedefs to add prefix "Sprite"

arthuro555:
+ Add prefix Sprite to typedefs
+ Add periods in typedefs parameters
+ Add dirty typing of the extraInitializationFromInitialInstance method of SpriteRuntimeObject

### DIFF
--- a/GDJS/Runtime/spriteruntimeobject.js
+++ b/GDJS/Runtime/spriteruntimeobject.js
@@ -5,53 +5,53 @@
  */
 
 /**
- * @typedef {Object} Point Represents a point in a frame
- * @property {number} x X position of the point
- * @property {number} y Y position of the point
+ * @typedef {Object} SpritePoint Represents a point in a coordinate system.
+ * @property {number | string} x X position of the point.
+ * @property {(number | string)} y Y position of the point.
  */
 
 /**
- * @typedef {Object} CustomPointData Represents a custom point in a frame
- * @property {string} name Name of the point
- * @property {number} x X position of the point
- * @property {number} y Y position of the point
+ * @typedef {Object} SpriteCustomPointData Represents a custom point in a frame.
+ * @property {string} name Name of the point.
+ * @property {number | string} x X position of the point.
+ * @property {number | string} y Y position of the point.
  */
 
 /**
- * @typedef {Object} CenterPointData Represents the center point in a frame
+ * @typedef {Object} SpriteCenterPointData Represents the center point in a frame.
  * @property {boolean} automatic Is the center automatically computed?
- * @property {number} x X position of the point
- * @property {number} y Y position of the point
+ * @property {number | string} x X position of the point.
+ * @property {number | string} y Y position of the point.
  */
 
 /**
- * @typedef {Object} FrameData Represents a {@link gdjs.SpriteAnimationFrame}
- * @property {string} [image] The resource name of the image used in this frame
- * @property {Array<customPointData>} [points] The points of the frame
- * @property {point} originPoint The origin point
- * @property {centerPointData} centerPoint The center of the frame
+ * @typedef {Object} SpriteFrameData Represents a {@link gdjs.SpriteAnimationFrame}.
+ * @property {string} [image] The resource name of the image used in this frame.
+ * @property {Array<SpriteCustomPointData>} [points] The points of the frame.
+ * @property {SpritePoint} originPoint The origin point.
+ * @property {SpriteCenterPointData} centerPoint The center of the frame.
  * @property {boolean} hasCustomCollisionMask Is The collision mask custom?
- * @property {Array<Array<point>>} [customCollisionMask] The collision mask if it is custom
+ * @property {Array<Array<SpritePoint>>} [customCollisionMask] The collision mask if it is custom.
  */
 
 /**
- * @typedef {Object} DirectionData Represents the data of a {@link gdjs.SpriteAnimationDirection}
- * @property {number} timeBetweenFrames Time between each frame, in seconds
+ * @typedef {Object} SpriteDirectionData Represents the data of a {@link gdjs.SpriteAnimationDirection}.
+ * @property {number | string} timeBetweenFrames Time between each frame, in seconds.
  * @property {boolean} looping Is the animation looping?
- * @property {Array<gdjs.SpriteAnimationFrame>} sprites The list of frames
+ * @property {Array<SpriteFrameData>} sprites The list of frames.
  */
 
 /**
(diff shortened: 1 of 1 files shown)


4ian:
Looks nice! Let me take a more in depth look and merge this :)

GDJS/Runtime/spriteruntimeobject.js
```diff
@@ -5,53 +5,53 @@
  */
 
 /**
- * @typedef {Object} Point Represents a point in a frame
- * @property {number} x X position of the point
- * @property {number} y Y position of the point
+ * @typedef {Object} SpritePoint Represents a point in a coordinate system.
+ * @property {number | string} x X position of the point.
```
4ian: X and Y position must always be numbers. If someone or the existing code is storing strings, that's a mistake and this should be fixed.
So let's keep `number`.

Putting string there will destroy any optimization made by the JS engine, and will expose us to very nasty bug. (just try to do "1" + 1 in a JavaScript console).

GDJS/Runtime/spriteruntimeobject.js
```diff
@@ -5,53 +5,53 @@
...
  */
 
 /**
- * @typedef {Object} CustomPointData Represents a custom point in a frame
- * @property {string} name Name of the point
- * @property {number} x X position of the point
- * @property {number} y Y position of the point
+ * @typedef {Object} SpriteCustomPointData Represents a custom point in a frame.
+ * @property {string} name Name of the point.
+ * @property {number | string} x X position of the point.
```
4ian: Let's revert everything to `number`.

GDJS/Runtime/spriteruntimeobject.js
```diff
@@ -5,53 +5,53 @@
...
  * @property {boolean} hasCustomCollisionMask Is The collision mask custom?
- * @property {Array<Array<point>>} [customCollisionMask] The collision mask if it is custom
+ * @property {Array<Array<SpritePoint>>} [customCollisionMask] The collision mask if it is custom.
  */
 
 /**
- * @typedef {Object} DirectionData Represents the data of a {@link gdjs.SpriteAnimationDirection}
- * @property {number} timeBetweenFrames Time between each frame, in seconds
+ * @typedef {Object} SpriteDirectionData Represents the data of a {@link gdjs.SpriteAnimationDirection}.
+ * @property {number | string} timeBetweenFrames Time between each frame, in seconds.
```
4ian: number

GDJS/Runtime/spriteruntimeobject.js
```diff
@@ -62,27 +62,30 @@
...
  */
 
 gdjs.SpriteAnimationFrame = function(imageManager, frameData)
 {
     /** @type {string} */
     this.image = frameData ? frameData.image : ""; //TODO: Rename in imageName, and do not store it in the object?
-    /** @type {PIXI.Texture} */
+    /**
+     * @type {PIXI.Texture}
+     * @note need to fix type
```
4ian: Let's not add "note" for things like this. Either a TODO or better: just let's remove the @type entirely for now.

GDJS/Runtime/spriteruntimeobject.js
```diff
@@ -62,27 +62,30 @@
...
-        /** @type {Point} */
+        /** @type {SpritePoint} */
         this.center = { x:0, y:0 };
-    };
+    }
     if ( this.origin === undefined ) {
-        /** @type {Point} */
+        /** @type {SpritePoint} */
         this.origin = { x:0, y:0 }
-    };
```
4ian: Thanks for cleaning these extra semi colons 👍

GDJS/Runtime/spriteruntimeobject.js
```diff
@@ -123,22 +126,22 @@ gdjs.SpriteAnimationFrame = function(imageManager, frameData)
...
-                /** @type {Point} */
+                /** @type {SpritePoint} */
         		var pointData = polygonData[j];
 
                 //Add a point, if necessary (Avoid recreating a point if it already exists).
                 if ( j >= this.customHitBoxes[i].vertices.length )
                     this.customHitBoxes[i].vertices.push([0,0]);
 
-                this.customHitBoxes[i].vertices[j][0] = parseFloat(pointData.x, 10);
-                this.customHitBoxes[i].vertices[j][1] = parseFloat(pointData.y, 10);
```
4ian: 👍

GDJS/Runtime/spriteruntimeobject.js
```diff
@@ -155,11 +158,11 @@ gdjs.SpriteAnimationFrame = function(imageManager, frameData)
  * Get a point of the frame.<br>
  * If the point does not exist, the origin is returned.
  * @param {string} name The point's name
- * @return {Point} The requested point. If it doesn't exists returns the origin point.
+ * @return {SpritePoint} The requested point. If it doesn't exists returns the origin point.
  */
 gdjs.SpriteAnimationFrame.prototype.getPoint = function(name) {
-	if ( name == "Centre" || name == "Center") return this.center;
```
4ian: 👍

GDJS/Runtime/spriteruntimeobject.js
```diff
@@ -310,6 +313,7 @@ gdjs.SpriteRuntimeObject.thisIsARuntimeObjectConstructor = "Sprite"; //Notify gd
 
 /**
  * Initialize the extra parameters that could be set for an instance.
+ * @param {{numberProperties: Array<{name: string, value: any}>, customSize: {width: number, height: number}}} initialInstanceData The extra parameters
```
4ian: value is guaranteed to be a number in numberProperties, so

`Array<{name: string, value: number}`

4ian:
Thanks for taking the time to clean that 👍😊
I let a few comments about typings.

GDJS/Runtime/spriteruntimeobject.js
arthuro555: OK that was because of the use of parseFloat

4ian: Ah I see! You're right this is confusing.
I think the parseFloat could be simply useless and added at a time where it was not sure if the points x/y were number or strings. Because they should be stored as number in the project JSON, so we should be able to careful remove them.

arthuro555:
I corrected what you spotted. I hope it's still mergeable and not too far behind master :frowning:

4ian:
Seems ok, let me review the code again and merge if there is no issue :)

4ian:
Looks great! Merged 😊Thanks for working on this 👍

PR #1379, opened 2020-01-15
Make cleaner object and behavior registering

arthuro555:
By exposing a function for that instead of finding them by making a variable


GDJS/Runtime/gd.js
```diff
@@ -160,6 +178,23 @@ gdjs.registerBehaviors = function() {
...
+ * Register the runtime behaviors that can be used bt runtimeObject.<br>
+ *
+ * The type of the behavior must be complete, with the namespace of the extension. For
+ * example, if you are providing a Draggable behavior in the DraggableBehavior extension,
+ * the full name of the type of the behavior is "DraggableBehavior::Draggable".
+ *
+ * @param Ctor The constructor of the Object.
+ * @param behaviorTypeName The name of the type of the behavior.
+ */
+gdjs.registerBehavior = function(Ctor, behaviorTypeName){
```
4ian: For consistency, I think it's better to  order `register` parameters the same way as the map is: first the key, then what we're registering.
In other words: `function(behaviorTypeName, Ctor)`

GDJS/Runtime/gd.js
```diff
@@ -98,19 +98,35 @@ gdjs.toDegrees = function(angleInRadians) {
...
+ * Register the runtime objects that can be used in runtimeScene.<br>
+ *
+ * The name of the type of the object must be complete, with the namespace if any. For
+ * example, if you are providing a Text object in the TextObject extension, the full name
+ * of the type of the object is "TextObject::Text".
+ *
+ * @param Ctor The constructor of the Object.
+ * @param objectTypeName The name of the type of the Object.
+ */
+gdjs.registerObject = function(Ctor, objectTypeName){
```
4ian: Same as behavior, let's be swap the parameter order to hint the reader that registerObject is actually putting things in a map: `function(objectTypeName, Ctor)`

GDJS/Runtime/gd.js
```diff
@@ -98,19 +98,35 @@ gdjs.toDegrees = function(angleInRadians) {
...
-        gdjs.objectsTypes.put(gdjs[p].thisIsARuntimeObjectConstructor, gdjs[p]);
+      if (gdjs[p].thisIsARuntimeObjectConstructor !== undefined) {
+        gdjs.registerObject(gdjs[p], gdjs[p].thisIsARuntimeObjectConstructor);
       }
     }
   }
 };
 
+/**
+ * Register the runtime objects that can be used in runtimeScene.<br>
```
4ian: No need for <br> anymore, an empty line is sufficient.

4ian:
Also seems that tests are failing, can you run them locally and fix if there is an issue?
image

you need to do `npm install` in GDJS/test and `npm test` then to launch the test runner.

arthuro555:
This looks very weird. Apparently, the namespace gdjs is not defined so JS cannot create something in it.

4ian:
Might be due to order of inclusion of JS files in tests?

arthuro555:
I was just checking but everything seems fine. I don't understand this at all.

arthuro555:
gdjs is indeeed present (my bad), but still don't understand the issue. When I removed accidentally the jsdoc of RuntimeObject I got a weird typescript error. Maybe this is related?

arthuro555:
Maybe related to the way karma works and this: https://stackoverflow.com/a/4011815/10994662

4ian:
This is maybe because you wrote `gdjs.SpriteRuntimeObject()` (see the extra parenthesis, I posted a comment about it), so you're calling the constructor without new and so `this` is not proper set.
Push the cleaned PR so that we can have a clean overview of the PR.

arthuro555:
This fixed it but there are other errors. I will fix them and push the working version later

4ian:
Ok, remember to make the *smallest* changes possible to decrease the surface for potential bugs and increase the chance of this being merged quickly and safely.
You have a few other PRs opened and I'd like to avoid them to all be in a limbo state ;) Prefer small, focused changes. This is slightly more work as this means more PR and commits to do, but also reduce the review time and the overall risk of re-work to do.

arthuro555:
Something I am not quite sure about is the extension behavior making through the IDE. I have no idea where the code for it is or how it works, I am not sure if it would break it. I cannot test for the moment, my PC broke :/

arthuro555:
Ok researched and found this that should be modified: https://github.com/4ian/GDevelop/blob/9adc40a55d0c4878c1cc35655d87e46991963ad7/GDJS/GDJS/Events/CodeGeneration/BehaviorCodeGenerator.cpp#L108
Can someone do it for me as my PC is broken?

4ian:
Good spot, this is indeed the place where the changes should be done :)
I'll do that for you. Good luck with your PC!

4ian:
I made the update in BehaviorCodeGenerator.cpp, fixed a few comments and merged :)
Thanks for working on this! The usage of these functions is much cleaner and more explicit - and explicit is better than implicit :)

PR #1382, opened 2020-01-16
Fix unaligned text for unknown behavior

Bouh:
Fix style and align for unknown behavior
- Remove useless span
- Remove useless style related to span
- Now Text for unknown behavior are align and use correct font.


4ian:
Love when there is more code to delete than to add! Great spot and refactoring. Merging now!

Issue #1298, opened 2019-11-24
Draggable behavior ignore the shape and size of custom hitbox and rotation of object


ddabrahim:
## Describe the bug
It seems like if we edit the hitbox of our object to be smaller or a different shape, the draggable behavior ignore this and we can drag the object at any point of the original hitbox which is the size of the image.
The draggable behavior also seems to ignore rotation of an object and we can drag the object at any point of the original rotation.

It is a problem when you have empty, transparent parts and the object is not a square shape and you don’t want the user to be able to drag and move the shape when it is clicked around the object on an empty transparent place but only if the user actually click the object.

Forum:
https://forum.gdevelop-app.com/t/bug-draggable-behavior-ignore-the-shape-and-size-of-custom-hitbox-and-rotation-of-object/21532

## To Reproduce
1. rotate something
2. use draggable behavior

* Please include a link to a game if possible!
[hitbox-test.zip](https://github.com/4ian/GDevelop/files/3883939/hitbox-test.zip)

## Other details
GDevelop 5 beta83 Desktop
Windows 10 64bit

4ian:
Right currently the draggable behavior is using the object position and width/height but not the hitbox. We should update the behavior to use the hitboxes, checking if the point of the touch/cursor is inside. 
I'm reworking the object spatial handling so this could be done at this point.

Silver-Streak:
@4ian 
I'm seeing you have an older PR on this that is still open. Seeing your related issues, I definitely agree this is similar to the issues people are having (or quality of life/experience issues) with "Cursor/Touch is on" detection.

Is there anything that needs to be done to bring attention back to this?

4ian:
My PR was solving this but this could be solved independently right now, actually not ton of work to just fix this!

Issue #1297, opened 2019-11-24
The cursor\touch is on object event ignore the shape and size of custom hitbox and rotation of object


ddabrahim:
## Describe the bug
It seems like if we edit the hitbox of our object to be smaller or a different shape, the “Cursor\Touch is on object” event ignore this and return true even if the cursor is at any point of the original hitbox which is the full size of the image.
It is also seems to ignore rotation of an object and the condition return true at any point of the original rotation.

It is a problem when you have empty, transparent parts and the object is not a square shape and you don’t want the user to trigger certain events while the mouse is over an empty space but only when the mouse is actually over the object.

Forum:
https://forum.gdevelop-app.com/t/bug-the-cursor-touch-is-on-object-event-ignore-the-shape-and-size-of-custom-hitbox-and-rotation-of-object/21561

## To Reproduce
1. rotate something
2. check if mouse is on the object

* Please include a link to a game if possible!
[mouse-over-object-test.zip](https://github.com/4ian/GDevelop/files/3883933/mouse-over-object-test.zip)

## Other details
GDevelop 5 beta83 Desktop
Windows 10 64bit

4ian:
Almost the cause as #1298: hitboxes should be used for this condition.
Potentially use the upcoming spatial management to reduce the cost of this condition when dealing with a lot of objects.

4ian:
Will probably be fixed during these improvements: https://trello.com/c/hjy2d6O2/383-performance-improvements-for-collisions-raycast-distance-tests-with-large-number-of-objects-or-very-large-levels

No timeline though, still quite a lot of work on this.

Bouh:
This is also on **shape painter** objects. We basic size used is the size from properties 32*32px.
But the best thing to do are pick the polygon path from the instance and use it as hitbox.
[Maybe pixiJS have something for do that ?](https://pixijs.download/v4.8.6/docs/PIXI.Graphics.html#interactive)

4ian:
> But the best thing to do are pick the polygon path from the instance and use it as hitbox.

This is a bit unrelated but yeah another possible improvement to generate the hitbox using the shapes of the Shape Painter. But this can also be using lots of CPU work for nothing if not used, and in particular if games are using it to make complex shapes.

An in-between solution would be to generate a hitbox (and AABB, the bounding box) that is using the min and max points of the shapes. This should be configurable to be deactivated if needed I think.

[...]

PR #1393, opened 2020-01-21
Experimental object collision and positions handling improvements

4ian:
This enables RuntimeObject positions to be managed by a spatial data structure (a "RTree", called RBush too here), allowing to speed up collision/separate objects/raycast on a large number of objects.

This is still highly *experimental*. I want mainly to ensure that there is no regression in terms of behaviors in games, and no large performance regression too.

Performance benchmarks:
- [x] Collisions benchmark faster when not hovering the ~1200 boxes (events goes from 120ms to ~2ms when hovering nothing)
- [x] "Separate Objects" faster ("Separate objects" benchmark, events goes from ~250ms to ~20ms at the beginning)
- [x] Distance test faster (In "Distance between objects" benchmark, events goes from 50ms to 10ms)
- [x] "Collisions with creation and deletion benchmark" is faster.
- [x] "Cursor on object benchmark":
  - [x] When everything moves, average frame time goes from 40ms to 70ms, with 30ms of update of the Rbush :'(. This indicates that a very large number of always moving objects should rather have the RTree deactivated.
  - [x] When nothing moves: average frame time goes from 20ms to 16ms
- [/] Raycast benchmark: events goes from 50ms to 60ms (converting lists is expensive)
- [/] "Point inside object benchmark": goes frm >200ms, then 40ms with free condition, then 50ms with the RTree (indicate that there are optimisations to be done in the generation of conditions).

Todos:
- [ ] See if we need to do right now an implementation where we can disable the RTree for some objects
  - [ ] If yes, surface in the UI and in events the ability to enable to disable the usage of a RTree to store positions of each object.

Clean-up todos:
- [ ] Old implementation of collisionTest/separateObjects/distanceTest/cursorOnObject/CollisionPoint/raycast
  - Clean related object functions: `separateFromObjectsList`, `cursorOnObject`, `isCollidingWithPoint`, `raycastTest`
  - Clean inputtools, objecttools: _cursorIsOnObject
  - Mark twoListsTest deprecated
- [ ] Mark #1297 fixed

Future:
- Adapt the Draggable behavior?
  - [ ] Mark #1298 fixed
- Surface in the Profiler UI the counters and times spent in the game engine like gdjs.ObjectPositionsManager

As GDevelop is a generic game engine, there can be cases where the usage of an RTree can be actually too costly and slow down the game compared to not using one. We would probably need to expose this is a setting somewhere in the IDE.

## Trying this

Changes in this PR should be invisible to the user. You probably won't notice even difference in your game, unless you're in the case where the RTree can be speeding up things for you (lots of collision check between lots of objects) or slowing this down (few checks but lots of objects moving). 

For now, I want to ensure the stability of this - there is room for optimisation later. In the worst case, this could be disabled by default and enabled in the UI on a case by case basis. For now, it's activated for all objects.

To help me, you can simply download the GDevelop versions below. They are beta 88 with this PR applied:

* [Windows](https://gdevelop-releases.s3.amazonaws.com/feature/collisions-improvements/latest/GDevelop 5 Setup 5.0.0-beta88.exe)
* [macOS](https://gdevelop-releases.s3.amazonaws.com/feature/collisions-improvements/latest/GDevelop 5-5.0.0-beta88-mac.zip)
* [Linux](https://gdevelop-releases.s3.amazonaws.com/feature/collisions-improvements/latest/gdevelop-5.0.0-beta88.tar.gz)

⚠️ I **recommend not to  use this version** for making your game. Stay on the latest stable version (beta 88), just install this one to try your game and see how it behaves.

Once you have this version, try to open your game and launch it:
* If your game is visibly broken, check if you can send me the part that is broken with a good description of what's wrong. 
* If it's working properly, let me know if you find any performance difference (large improvement or regression). You may use the GDevelop profiler, running it for a few seconds and then look at the average time to render a frame in your game - though this may be hard to compare as this depends a lot on other factors.


rishabhabhani:
A few questions I'd like to ask about this.
Is RTree a structure that runs on a separate thread or the same as the others.
If it does run on another, is it designed with the objectives in mind or be extensible?

4ian:
JavaScript is single threaded, so runs on the same thread as the rest. Web-workers or wasm could be used to maybe have it running on another thread, but it was not a goal in this implementation (there is more complexity to achieve this).

Silver-Streak:
@4ian has the efforts around this been dropped? It seems like your checklist is mostly complete, and I'm sure people would love collision performance improvements. (Especially with Davy's other outstanding PRs)

PR #1394, opened 2020-01-24
Improve font for Text & BBText

Bouh:
@blurymind 
In the second commit i try to add the resource selector for the font.
In the IDE the resource selector is showing, but the font value correct for the BBText object ?

Edit : 
For the first commit i add a font label because it's hard to see the input if you didn't know.

Note see if canBeReset can be used too
https://github.com/4ian/GDevelop/pull/1308

### DIFF
--- a/Extensions/BBText/JsExtension.js
+++ b/Extensions/BBText/JsExtension.js
@@ -85,7 +85,8 @@ module.exports = {
       objectProperties.set(
         'fontFamily',
         new gd.PropertyDescriptor(objectContent.fontFamily)
-          .setType('string')
+          .setType("resource")
+          .addExtraInfo("font")
           .setLabel(_('Base font family'))
       );
 
--- a/Extensions/BBText/JsExtension.js
+++ b/Extensions/BBText/JsExtension.js
@@ -85,8 +85,8 @@ module.exports = {
       objectProperties.set(
         'fontFamily',
         new gd.PropertyDescriptor(objectContent.fontFamily)
-          .setType("resource")
-          .addExtraInfo("font")
+          .setType('resource')
+          .addExtraInfo('font')
           .setLabel(_('Base font family'))
       );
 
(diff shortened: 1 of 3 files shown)


blurymind:
ah cool, I will give this a try :) I didn't realise we have a font selector

Bouh:
And voilà, the BBText object font working like Text object :)
If you want give a try before merging, to be sure nothing is broken.

4ian:
> I was wondering if we can expose fonts that are bundled as a resource with the game- that way you dont have to worry if the player has the font family you use in your game installed.
> If that selector lets you do exactly that, AWESOME!

That's exactly what resources are for ;) 
In general, the game engine is responsible for providing the objects all that they need to work, i.e: images, fonts, audio, json or videos. This allow to centralize resource management, ensure that resources are exported, etc...

I overlooked this when first adding the BBText object. The implementation seems exactly what is needed 👍👍

Merging this now. "Fun fact", `this.project` is used everywhere in the JsExtension.js but is undefined 😬By chance it's actually never used. Will fix that in another commit!

Bouh:
> Merging this now. "Fun fact", this.project is used everywhere in the JsExtension.js but is undefined 😬By chance it's actually never used. Will fix that in another commit!

So what do this line if `this.project` doesn't work ? Why call this.project if this if undefined and useless :)
```
this._pixiObject.textStyles.default.fontFamily = this._pixiResourcesLoader.getFontFamily(
        this.project,
        fontFamily
      );
```

4ian:
Mmm you're right seems that the loading of the font is not done properly 😬
`loadFontFamily` should be called instead, like for the text project:

https://github.com/4ian/GDevelop/blob/168b8d3fea48402de8fae6d27de94656c101c7a8/newIDE/app/src/ObjectsRendering/Renderers/RenderedTextInstance.js#L75-L85

Do you have time to take a look at this?
`getFontFamily` should probably be removed in PixiResourcesLoader.

Bouh:
Yes i can use loadFontFamily() but `this.project` stay  undefined : 
`Cannot read property 'getResourcesManager' of undefined`

4ian:
Use `this._project`

PR #1398, opened 2020-01-26
Try to add comment filter for search in event list

Bouh:
I feel like I'm not far from the solution.

As I don't know C++ I looked at the other functions and I reproduced what I understood from them.

I don't know how to debug a variable, it would have helped me to know where I am and the state of my values, or at least their structure.

I think I'm in gd::CommentEvent, but according to errors during compilation I'm mostly in the parent class gd::BaseEvent ?

### DIFF
--- a/Core/GDCore/Events/Builtin/CommentEvent.cpp
+++ b/Core/GDCore/Events/Builtin/CommentEvent.cpp
@@ -12,25 +12,11 @@ using namespace std;
 
 namespace gd {
 
-vector<const gd::String*> CommentEvent::GetAllSearchableStrings( BaseEvent &event)
-    const {
-  vector<const gd::String*> allSearchableStrings;
+vector<gd::String> CommentEvent::GetAllSearchableStrings() const {
+  vector<gd::String> allSearchableStrings;
 
-  stringCom1 = event.GetChild("comment", 0, "Com1").GetValue().GetString();
-  stringCom2 = event.GetChild("comment", 0, "Com2").GetValue().GetString();///< Com2 is deprecated
-  allSearchableStrings.push_back(&stringCom1);
-  allSearchableStrings.push_back(&stringCom2);
-
-  return allSearchableStrings;
-}
-
-vector<gd::String*> CommentEvent::GetAllSearchableStrings( BaseEvent &event) {
-  vector<const gd::String*> allSearchableStrings;
-
-  stringCom1 = event.GetChild("comment", 0, "Com1").GetValue().GetString();
-  stringCom2 = event.GetChild("comment", 0, "Com2").GetValue().GetString();///< Com2 is deprecated
-  allSearchableStrings.push_back(&stringCom1);
-  allSearchableStrings.push_back(&stringCom2);
+  allSearchableStrings.push_back(com1);
+  allSearchableStrings.push_back(com2);  ///< Com2 is deprecated
 
   return allSearchableStrings;
 }
--- a/Core/GDCore/Events/Builtin/CommentEvent.cpp
+++ b/Core/GDCore/Events/Builtin/CommentEvent.cpp
@@ -12,19 +12,27 @@ using namespace std;
 
 namespace gd {
 
-vector<const gd::EventsList*> CommentEvent::GetAllCommentsVectors()
+vector<const gd::String*> CommentEvent::GetAllSearchableStrings( BaseEvent &event)
     const {
-  vector<const gd::EventsList*> allComments;
-  allComments.push_back(&comments);
+  vector<const gd::String*> allSearchableStrings;
 
-  return allComments;
+  stringCom1 = event.GetChild("comment", 0, "Com1").GetValue().GetString();
+  stringCom2 = event.GetChild("comment", 0, "Com2").GetValue().GetString();///< Com2 is deprecated
+  allSearchableStrings.push_back(&stringCom1);
+  allSearchableStrings.push_back(&stringCom2);
+
+  return allSearchableStrings;
 }
 
-vector<gd::EventsList*> CommentEvent::GetAllCommentsVectors() {
-  vector<gd::EventsList*> allComments;
-  allComments.push_back(&comments);
+vector<gd::String*> CommentEvent::GetAllSearchableStrings( BaseEvent &event) {
+  vector<const gd::String*> allSearchableStrings;
 
-  return allComments;
+  stringCom1 = event.GetChild("comment", 0, "Com1").GetValue().GetString();
+  stringCom2 = event.GetChild("comment", 0, "Com2").GetValue().GetString();///< Com2 is deprecated
+  allSearchableStrings.push_back(&stringCom1);
+  allSearchableStrings.push_back(&stringCom2);
+
+  return allSearchableStrings;
 }
 
 void CommentEvent::SerializeTo(SerializerElement &element) const {
--- a/Core/GDCore/Events/Builtin/CommentEvent.cpp
+++ b/Core/GDCore/Events/Builtin/CommentEvent.cpp
@@ -12,6 +12,21 @@ using namespace std;
 
 namespace gd {
 
+vector<const gd::EventsList*> CommentEvent::GetAllCommentsVectors()
+    const {
+  vector<const gd::EventsList*> allComments;
+  allComments.push_back(&comments);
+
+  return allComments;
+}
+
+vector<gd::EventsList*> CommentEvent::GetAllCommentsVectors() {
+  vector<gd::EventsList*> allComments;
+  allComments.push_back(&comments);
+
+  return allComments;
+}
+
 void CommentEvent::SerializeTo(SerializerElement &element) const {
   element.AddChild("color")
       .SetAttribute("r", r)
--- a/Core/GDCore/Events/Builtin/CommentEvent.h
+++ b/Core/GDCore/Events/Builtin/CommentEvent.h
@@ -11,7 +11,7 @@
 namespace gd {
 class Layout;
 class Project;
-}
+}  // namespace gd
 
 namespace gd {
 
@@ -46,8 +46,8 @@ class GD_CORE_API CommentEvent : public gd::BaseEvent {
   const gd::String& GetComment() const { return com1; }
   void SetComment(const gd::String& comment) { com1 = comment; }
 
-  virtual std::vector<const gd::EventsList*> GetAllCommentsVectors() const;
-  virtual std::vector<gd::EventsList*> GetAllCommentsVectors();
+  virtual std::vector<const gd::String*> GetAllSearchableStrings(BaseEvent &event) const;
+  virtual std::vector<gd::String*> GetAllSearchableStrings( BaseEvent &event);
 
   virtual void SerializeTo(SerializerElement& element) const;
   virtual void UnserializeFrom(gd::Project& project,
@@ -64,10 +64,9 @@ class GD_CORE_API CommentEvent : public gd::BaseEvent {
   gd::String com1;  ///< Comment
   gd::String com2;  ///< Optional second column comment, deprecated
 
-private:
-  gd::EventsList comments;
-  EventsList events;
-
+ private:
+  gd::String stringCom1;
+  gd::String stringCom2;
 };
 
 }  // namespace gd
--- a/Core/GDCore/Events/Builtin/CommentEvent.h
+++ b/Core/GDCore/Events/Builtin/CommentEvent.h
@@ -7,6 +7,7 @@
 #define COMMENTEVENT_H
 
 #include "GDCore/Events/Event.h"
+#include "GDCore/Events/EventsList.h"
 namespace gd {
 class Layout;
 class Project;
@@ -45,6 +46,9 @@ class GD_CORE_API CommentEvent : public gd::BaseEvent {
   const gd::String& GetComment() const { return com1; }
   void SetComment(const gd::String& comment) { com1 = comment; }
 
+  virtual std::vector<const gd::EventsList*> GetAllCommentsVectors() const;
+  virtual std::vector<gd::EventsList*> GetAllCommentsVectors();
+
   virtual void SerializeTo(SerializerElement& element) const;
   virtual void UnserializeFrom(gd::Project& project,
                                const SerializerElement& element);
@@ -59,6 +63,11 @@ class GD_CORE_API CommentEvent : public gd::BaseEvent {
 
   gd::String com1;  ///< Comment
   gd::String com2;  ///< Optional second column comment, deprecated
+
+private:
+  gd::EventsList comments;
+  EventsList events;
+
 };
 
 }  // namespace gd
--- a/Core/GDCore/Events/Builtin/CommentEvent.h
+++ b/Core/GDCore/Events/Builtin/CommentEvent.h
@@ -46,8 +46,7 @@ class GD_CORE_API CommentEvent : public gd::BaseEvent {
   const gd::String& GetComment() const { return com1; }
   void SetComment(const gd::String& comment) { com1 = comment; }
 
-  virtual std::vector<const gd::String*> GetAllSearchableStrings(BaseEvent &event) const;
-  virtual std::vector<gd::String*> GetAllSearchableStrings( BaseEvent &event);
+  virtual std::vector<gd::String> GetAllSearchableStrings() const;
 
   virtual void SerializeTo(SerializerElement& element) const;
   virtual void UnserializeFrom(gd::Project& project,
@@ -63,10 +62,6 @@ class GD_CORE_API CommentEvent : public gd::BaseEvent {
 
   gd::String com1;  ///< Comment
   gd::String com2;  ///< Optional second column comment, deprecated
-
- private:
-  gd::String stringCom1;
-  gd::String stringCom2;
 };
 
 }  // namespace gd
--- a/Core/GDCore/Events/Event.h
+++ b/Core/GDCore/Events/Event.h
@@ -90,8 +90,8 @@ class GD_CORE_API BaseEvent {
   bool HasSubEvents() const;
 
   /**
-   * Event must be able to return all conditions std::vector they have.
-   * Used to preprocess the conditions.
+   * \brief Return a list of all conditions of the event.
+   * \note Used to preprocess or search in the conditions.
    */
   virtual std::vector<gd::InstructionsList*> GetAllConditionsVectors() {
     std::vector<gd::InstructionsList*> noConditions;
@@ -104,8 +104,8 @@ class GD_CORE_API BaseEvent {
   };
 
   /**
-   * Event must be able to return all actions std::vector they have.
-   * Used to preprocess the actions.
+   * \brief Return a list of all actions of the event.
+   * \note Used to preprocess or search in the actions.
    */
   virtual std::vector<gd::InstructionsList*> GetAllActionsVectors() {
     std::vector<gd::InstructionsList*> noActions;
@@ -118,8 +118,8 @@ class GD_CORE_API BaseEvent {
   };
 
   /**
-   * Event must be able to return all strings std::vector they have.
-   * Used to preprocess the strings in an event.
+   * \brief Return a list of all strings of the event.
+   * \note Used to preprocess or search in the event strings.
    */
   virtual std::vector<gd::String> GetAllSearchableStrings() const {
     std::vector<gd::String> noSearchableStrings;
@@ -127,8 +127,8 @@ class GD_CORE_API BaseEvent {
   };
 
   /**
-   * Event must be able to return all expressions they have.
-   * Used to preprocess the expressions.
+   * \brief Return a list of all expressions of the event.
+   * \note Used to preprocess or search in the expressions.
    */
   virtual std::vector<gd::Expression*> GetAllExpressions() {
     std::vector<gd::Expression*> noExpr;
--- a/Core/GDCore/Events/Event.h
+++ b/Core/GDCore/Events/Event.h
@@ -118,6 +118,20 @@ class GD_CORE_API BaseEvent {
   };
 
   /**
+   * Event must be able to return all comments std::vector they have.
+   * Used to preprocess the comments.
+   */
+  virtual std::vector<gd::EventsList*> GetAllCommentsVectors() {
+    std::vector<gd::EventsList*> noComments;
+    return noComments;
+  };
+  virtual std::vector<const gd::EventsList*> GetAllCommentsVectors()
+      const {
+    std::vector<const gd::EventsList*> noComments;
+    return noComments;
+  };
+
+  /**
    * Event must be able to return all expressions they have.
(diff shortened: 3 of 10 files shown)


4ian:
That's a good start but there are misconceptions about what an event is :) 
The idea is:

- You add a method to the gd::BaseEvent class (Event.h) (which is the base class from which all events are deriving) that is there to extract the strings from an event. By default, just return an empty list.
- You redefine this method in the `CommentEvent` to returns the strings that are the comments.
- In the class doing the search, you're iterating on events and calling the method to extract the strings for all events (most events will nothing, but comments will return the comments content). for each event, you iterate on this list and do the search in these strings.

4ian:
> something like GetEvent() is not a member of baseEvent

This means "You try to call GetEvent() on something, but this something is a gd::BaseEvent, and I can't find any method called GetEvent". So this means that you *already* have an event :) No need then to call GetEvent(), because you already have one ;)

Core/GDCore/Events/Builtin/CommentEvent.h
```diff
@@ -46,8 +46,8 @@ class GD_CORE_API CommentEvent : public gd::BaseEvent {
   const gd::String& GetComment() const { return com1; }
   void SetComment(const gd::String& comment) { com1 = comment; }
 
-  virtual std::vector<const gd::EventsList*> GetAllCommentsVectors() const;
-  virtual std::vector<gd::EventsList*> GetAllCommentsVectors();
+  virtual std::vector<const gd::String*> GetAllSearchableStrings(BaseEvent &event) const;
+  virtual std::vector<gd::String*> GetAllSearchableStrings( BaseEvent &event);
```
Bouh: Not sure here i guess there are mistakes in and after parenthesis.
I guess also one of them should return the vector of string ?
I didn't understand why two similar lines. with **Const** and without.

Core/GDCore/Events/Event.h
```diff
@@ -118,17 +118,17 @@ class GD_CORE_API BaseEvent {
...
+    std::vector<gd::String*> noSearchableStrings;
+    return noSearchableStrings;
   };
-  virtual std::vector<const gd::EventsList*> GetAllCommentsVectors()
+  virtual std::vector<const gd::String*> GetAllSearchableStrings()
       const {
-    std::vector<const gd::EventsList*> noComments;
-    return noComments;
+    std::vector<const gd::String*> noSearchableStrings;
+    return noSearchableStrings;
```
Bouh: You have said "return empty string"
Then i return empty vector of string.
I guess all is ok here.
(Note for me Vector are like Array in JS)

Core/GDCore/Events/Builtin/CommentEvent.cpp
```diff
@@ -12,19 +12,27 @@ using namespace std;
 
 namespace gd {
 
-vector<const gd::EventsList*> CommentEvent::GetAllCommentsVectors()
+vector<const gd::String*> CommentEvent::GetAllSearchableStrings( BaseEvent &event)
```
Bouh: I suppose all this line is wrong. Or at least the end in parenthesis.
I don't know.

Bouh:
Write c++ code when i didn't read one tutorial about it is not a good idea lol.

Core/GDCore/Events/Builtin/CommentEvent.cpp
4ian: This method is a method of CommentEvent, so you probably don't need event. Also, return a vector containing strings directly (not pointers to strings).

```suggestion
vector<gd::String> CommentEvent::GetAllSearchableStrings()
```

Core/GDCore/Events/Builtin/CommentEvent.cpp
```diff
@@ -12,6 +12,29 @@ using namespace std;
 
 namespace gd {
 
+vector<const gd::String*> CommentEvent::GetAllSearchableStrings( BaseEvent &event)
+    const {
+  vector<const gd::String*> allSearchableStrings;
```
4ian: vector<gd::String> allSearchableStrings;
```suggestion
  vector<const gd::String*> allSearchableStrings;
```

Core/GDCore/Events/Builtin/CommentEvent.cpp
```diff
@@ -12,6 +12,29 @@ using namespace std;
 
 namespace gd {
 
+vector<const gd::String*> CommentEvent::GetAllSearchableStrings( BaseEvent &event)
+    const {
+  vector<const gd::String*> allSearchableStrings;
+
+  stringCom1 = event.GetChild("comment", 0, "Com1").GetValue().GetString();
```
4ian: You don't need to do this, look at what is already available in CommentEvent:

https://github.com/4ian/GDevelop/blob/7b30f63cb126013417535ba77914cd264b548f19/Core/GDCore/Events/Builtin/CommentEvent.h#L60-L61

Exactly what you need:

```
allSearchableStrings.push_back(com1);
allSearchableStrings.push_back(com2);
```

No need to push a pointer, it's fine here to put a copy of the strings.

Core/GDCore/Events/Builtin/CommentEvent.cpp
```diff
@@ -12,6 +12,29 @@ using namespace std;
...
+
+  stringCom1 = event.GetChild("comment", 0, "Com1").GetValue().GetString();
+  stringCom2 = event.GetChild("comment", 0, "Com2").GetValue().GetString();///< Com2 is deprecated
+  allSearchableStrings.push_back(&stringCom1);
+  allSearchableStrings.push_back(&stringCom2);
+
+  return allSearchableStrings;
+}
+
+vector<gd::String*> CommentEvent::GetAllSearchableStrings( BaseEvent &event) {
```
4ian: The `const` at the end of the declaration (see the previous function) is here to say "this method won't change the object I swear". If you try to do something that will change the object, the compiler will reject it.

You don't need this second version (called the "non const" version), you can remove it.

Core/GDCore/Events/Builtin/CommentEvent.h
4ian: As said in another comment: 

The one that is `const` will tell the compiler "please check that the object is not modified". It also tell the compiler "allow me to call this method on a const object". If you do:

```cpp
class A {
  void methodNonConst();
  void methodThatIsConst() const;
}
```

then you can do:
```cpp
void someOtherFunction(const A &a) { // Here I say "I take a reference to A, and I won't modify it"
  a.methodNonConst(); // NOT allowed, compiler will reject it. You're trying to modify the object it seems!
  a.methodThatIsConst(); // Fine! Nothing will be modified
}
```

Note that if you pass just a reference, there is no difference:
```cpp
void someOtherFunction(A &a) { // Here I say "I take a reference to A, and I maybe modify it"
  a.methodNonConst(); // Allowed. The object might be modified
  a.methodThatIsConst(); // Also allowed!
}
```

This is called "const-correctness": https://www.cprogramming.com/tutorial/const_correctness.html
(don't try to understand everything at first)

---

You also need only one method here (which can be const as not changing anything):

```
virtual std::vector<gd::String> GetAllSearchableStrings() const;
```

Core/GDCore/Events/Builtin/CommentEvent.h
```diff
@@ -59,6 +63,10 @@ class GD_CORE_API CommentEvent : public gd::BaseEvent {
 
   gd::String com1;  ///< Comment
   gd::String com2;  ///< Optional second column comment, deprecated
+
+ private:
```
4ian: No need to add this, we already have these in the class

Core/GDCore/Events/Event.h
4ian: Yep vector is a resizeable array. "Vector" is basically an awful name for a "resizeable array" in C++ (because originally, the C arrays can't be resized).

You only need one method:
```
virtual std::vector<gd::String> GetAllSearchableStrings() const {
    std::vector<gd::String> noSearchableStrings;
    return noSearchableStrings;
}
```

Core/GDCore/IDE/Events/EventsRefactorer.cpp
```diff
@@ -711,6 +728,25 @@ bool EventsRefactorer::SearchStringInConditions(
   return false;
 }
 
+bool EventsRefactorer::GetAllSearchableStrings(gd::ObjectsContainer& project,
+                                               gd::ObjectsContainer& layout,
+                                               gd::EventsList& events,
+                                               gd::String search,
+                                               bool matchCase) {
+  // See later for "BuiltinCommonInstructions::Group" when Comment works
+  if (event.GetType() == = "BuiltinCommonInstructions::Comment") {
```
4ian: No need for this: GetAllSearchableStrings is a method that was added in `BaseEvent`, meaning that *all* the events now have it! 
And it's overriden by CommentEvent, but this method *does not care* who is overriding it.

If in the future, some events is redefining GetAllSearchableStrings, this will work without any changes.

So you can remove this if (also, no triple equal in C++, just ==)

Core/GDCore/IDE/Events/EventsRefactorer.cpp
```diff
@@ -711,6 +728,25 @@ bool EventsRefactorer::SearchStringInConditions(
...
 }
 
+bool EventsRefactorer::GetAllSearchableStrings(gd::ObjectsContainer& project,
+                                               gd::ObjectsContainer& layout,
+                                               gd::EventsList& events,
+                                               gd::String search,
+                                               bool matchCase) {
+  // See later for "BuiltinCommonInstructions::Group" when Comment works
+  if (event.GetType() == = "BuiltinCommonInstructions::Comment") {
+    for (gd::String str : events[id].GetAllSearchableStrings()) {
```
4ian: Seems like you forgot to iterate on `events`. `events` is an `gd::EventsList`, which, like the name indicates, is a "list", almost an array.

4ian: Add the missing for, and this can be just:

```
for (gd::String str : event.GetAllSearchableStrings()) {
```

if I suppose you named "event" the "gd::BaseEvent" variable that you will get with "GetEvent(index)" on `events`

Core/GDCore/IDE/Events/EventsRefactorer.h
```diff
@@ -202,6 +203,11 @@ class GD_CORE_API EventsRefactorer {
                                        gd::InstructionsList& conditions,
                                        gd::String search,
                                        bool matchCase);
+  static bool GetAllSearchableStrings(gd::ObjectsContainer& project,
```
4ian: You will have to change Bindings.idl to expose this new parameter to JavaScript

4ian:
> Write c++ code when i didn't read one tutorial about it is not a good idea lol.

Hopefully these changes are specific enough so that you can do without.. but surely a good idea to get a quick tutorial or two about C++. Will be interesting to see concepts that don't exist in JS and will surely remind you of stuff that you've seen in JS :) 
For example, in C++ variables can be passed by value (`Car myCar` - this copies myCar in memory), reference (`Car & car`, this does not copy in memory), const-reference (`const Car& car`, this does not copy in memory and forbid you to change the object) or pointer (Car* car) - though pointer are avoided as much as possible because they can be null.
In JavaScript, all objects and arrays are passed as... reference. But primitive types (boolean, number, strings) are copied (passed by "value").

Bouh:
There are problems
If I understand :
- `GetAllSearchableStrings()` shouldn't be in EventsRefectorer at all.
- `SearchStringInComments()`  is now missing O.o

When i get `stringsVectors` from 
`    vector<gd::String> stringsVectors = events[i].GetAllSearchableStrings();`
I need to loop on each string in stringsVectors and pass one string to the missing `SearchStringInComments()` ?

`SearchStringInComments()` should be define in EventsRefectorer.cpp & .h  ?

I think 'ive mix `SearchStringInComments()` and  `GetAllSearchableStrings()`

GDevelop.js/Bindings/Bindings.idl
```diff
@@ -1528,7 +1528,8 @@ interface EventsRefactorer {
...
     void STATIC_RemoveObjectInEvents([Const, Ref] Platform platform, [Ref] ObjectsContainer project, [Ref] ObjectsContainer layout, [Ref] EventsList events, [Const] DOMString name);
     void STATIC_ReplaceStringInEvents([Ref] ObjectsContainer project, [Ref] ObjectsContainer layout, [Ref] EventsList events, [Const] DOMString toReplace, [Const] DOMString newString, boolean matchCase, boolean inConditions, boolean inActions);
-    [Value] VectorEventsSearchResult STATIC_SearchInEvents([Ref] ObjectsContainer project, [Ref] ObjectsContainer layout, [Ref] EventsList events, [Const] DOMString search, boolean matchCase, boolean inConditions, boolean inActions);
+    [Value] VectorEventsSearchResult STATIC_SearchInEvents([Ref] ObjectsContainer project, [Ref] ObjectsContainer layout, [Ref] EventsList events, [Const] DOMString search, boolean matchCase, boolean inConditions, boolean inActions, boolean inComments);
+    [Value] VectorEventsSearchResult STATIC_GetAllSearchableStrings([Ref] ObjectsContainer project, [Ref] ObjectsContainer layout, [Ref] EventsList events, [Const] DOMString search, boolean matchCase);
```
4ian: You dont' need this line - and there is no "GetAllSearchableStrings" in EventsRefactorer, why adding it? :)

newIDE/app/src/EventsSheet/SearchPanel.js
```diff
@@ -148,7 +148,7 @@ export default class SearchPanel extends PureComponent<Props, State> {
                 onCheck={(e, checked) => this.setState({ matchCase: !checked })}
               />
               <Text>
-                <Trans>Filter by</Trans>
```
4ian: Should we rename this to "Search in"?

Core/GDCore/IDE/Events/EventsRefactorer.cpp
```diff
@@ -631,6 +632,21 @@ vector<EventsSearchResult> EventsRefactorer::SearchInEvents(
       }
     }
 
+    if (inComments) {
+      vector<gd::String> stringsVectors = events[i].GetAllSearchableStrings();
```
4ian: This line is correct.

Explanation:
So here you have a vector of string (i.e: a list of strings) that represents "all the strings that can be searched in events number i".
So now your goal is too.. search in these strings. If you find what is searched, then good, you add the event to the results :) Otherwise you don't do anything.

Core/GDCore/IDE/Events/EventsRefactorer.cpp
```diff
@@ -631,6 +632,21 @@ vector<EventsSearchResult> EventsRefactorer::SearchInEvents(
       }
     }
 
+    if (inComments) {
+      vector<gd::String> stringsVectors = events[i].GetAllSearchableStrings();
+      
+      for (std::size_t j = 0; j < stringsVectors.size(); ++j) {
```
4ian: So you iterate on all the strings...

(this line is correct)

Core/GDCore/IDE/Events/EventsRefactorer.cpp
```diff
@@ -631,6 +632,21 @@ vector<EventsSearchResult> EventsRefactorer::SearchInEvents(
       }
     }
 
+    if (inComments) {
+      vector<gd::String> stringsVectors = events[i].GetAllSearchableStrings();
+      
+      for (std::size_t j = 0; j < stringsVectors.size(); ++j) {
+        if (!eventAddedInResults &&
+            SearchStringInComments(
```
4ian: and there I'm not sure why you made your life much more complex that you should have ^^

You have:
* the string to search (`search`)
* the string to search in (`stringsVectors[j]`)
* the event number (`i`).

So now just search the string! :)

Core/GDCore/IDE/Events/EventsRefactorer.cpp
4ian: Iterating here is doable, but then you iterate also in SearchStringInComments... you have to choose where! :) 
Remove this loop

Core/GDCore/IDE/Events/EventsRefactorer.cpp
4ian: You're already doing this in SearchStringInComments, you can remove that

Core/GDCore/IDE/Events/EventsRefactorer.cpp
4ian: Pass just the event, not a string:

`project, layout, events[i], search, matchCase`

Core/GDCore/IDE/Events/EventsRefactorer.cpp
```diff
@@ -711,6 +728,24 @@ bool EventsRefactorer::SearchStringInConditions(
   return false;
 }
 
+bool EventsRefactorer::SearchStringInComments(gd::ObjectsContainer& project,
```
4ian: Rename this to `SearchStringInEvents`
There is no notion of "comments" here. The strings might be in the future strings coming from somewhere else... sure for now only comments have strings, but that's a coincidence :)

Core/GDCore/IDE/Events/EventsRefactorer.cpp
```diff
@@ -711,6 +728,24 @@ bool EventsRefactorer::SearchStringInConditions(
   return false;
 }
 
+bool EventsRefactorer::SearchStringInComments(gd::ObjectsContainer& project,
+                                               gd::ObjectsContainer& layout,
+                                               gd::EventsList& events,
```
4ian: Pass a `gd::BaseEvent& event`. You don't need to pass a list of event, just the event where to search in.

Core/GDCore/IDE/Events/EventsRefactorer.cpp
```diff
@@ -711,6 +728,24 @@ bool EventsRefactorer::SearchStringInConditions(
   return false;
 }
 
+bool EventsRefactorer::SearchStringInComments(gd::ObjectsContainer& project,
+                                               gd::ObjectsContainer& layout,
+                                               gd::EventsList& events,
+                                               gd::String search,
+                                               bool matchCase) {
+  for (gd::BaseEvent event : events) {
```
4ian: No need to iterate, just pass the event.
You can remove this loop

Core/GDCore/IDE/Events/EventsRefactorer.cpp
```diff
@@ -711,6 +728,24 @@ bool EventsRefactorer::SearchStringInConditions(
...
+                                               gd::String search,
+                                               bool matchCase) {
+  for (gd::BaseEvent event : events) {
+    for (gd::String str : event.GetAllSearchableStrings()) {
+      if (matchCase) {
+        if (str.find(search) != gd::String::npos) return true;
+      } else {
+        if (str.FindCaseInsensitive(search) != gd::String::npos) return true;
+      }
+    }
```
4ian: This is good :) You're looking into every string. As soon as one contains the search, we return true!

4ian:
> GetAllSearchableStrings() shouldn't be define in EventsRefectorer at all only in BaseEvent & CommentEvent

Yes

> When i get stringsVectors from
vector<gd::String> stringsVectors = events[i].GetAllSearchableStrings();
> I need to loop on each string in stringsVectors and pass one string to the missing SearchStringInComments() ?

No, this method will take care of iterating. Just pass it the event.

> SearchStringInComments() should be define in EventsRefectorer.cpp & .h ?

Yes, everything that is in a cpp file must be in the .h

> I think 'ive mix SearchStringInComments() and GetAllSearchableStrings()

Yes 😄Think like this:
* `GetAllSearchableStrings` is something that is in the BaseEvent (the common basis for all events) (and it's redefine by CommentEvent, but we don't need to know that). It has on role, which is "When I call you, you give me all the texts of the event please."
* `SearchStringInComments` is a helper method. It's role: "When I call you, I'll pass you an event and a string called `search` (plus a few other things that you might not use, like the project). Please tell me (by sending me true or false) if the event contains the `search`".
  * To do this, this method has to ask the event "can you please give me all the texts that you have??"... and *good news*, we just made a method for that! It's called `GetAllSearchableStrings` and it's available on all events! So we don't event care what kind of event we're dealing with :)

Bouh:
It's works !
Before merging this one, can i add rename Comments filter by Comments&Groups ? 
Or i setup a new button for Groups ?
Also `Filter by` is fine, if you prefer `Search by` i can replace it ^^

Bouh:
I need just add GetAllSearchableStrings() in GroupEvent .cpp and .h :)

Bouh:
Uumh a group is also a comment field in the IDE, so i get name from group event, and com from comment event, both on same filter option.

4ian:
> Also Filter by is fine, if you prefer Search by i can replace it ^^

"Search in" would be good - we're doing a search in events, and precisely in some part of them :)

> I need just add and adapt GetAllSearchableStrings() in GroupEvent .cpp and .h :)

Yes that looks good! :) 

> Before merging this one, can i add rename Comments filter by Comments&Groups ?

In the UI yes - but we'll run out of place I'm afraid?

In the code, instead of "inComments", name the parameter "inEventStrings" to be explicit that this is not only comments (and not only groups, can be any event with a string).
Also clean the code that is commented and should be good :)

newIDE/app/src/EventsSheet/SearchPanel.js
```diff
@@ -165,6 +171,13 @@ export default class SearchPanel extends PureComponent<Props, State> {
                   this.setState({ searchInActions: checked })
                 }
               />
+              <InlineCheckbox
+                label={<Trans>Comments</Trans>}
```
4ian: What do you think of "Texts" instead of "Comments"? Might be a bit more generic

newIDE/app/src/EventsSheet/EventsSearcher.js
```diff
@@ -11,6 +11,7 @@ export type SearchInEventsInputs = {|
   matchCase: boolean,
   searchInConditions: boolean,
   searchInActions: boolean,
+  searchInComments: boolean,
```
4ian: Let's rename searchInComments to "searchInEventStrings"
to be consistent with the naming of the parameter in GDevelop Core :)

newIDE/app/src/EventsSheet/SearchPanel.js
Bouh: Yes more accurate

4ian:
Seems like we're almost there 👍

newIDE/app/src/EventsSheet/EventsSearcher.js
Bouh: arg missing this one

Bouh:
Next option will be search in expressions and search bar will be complete!

4ian:
Thanks for working on this 👍👍
Will be very useful for searching in large events sheets.

Bouh:
Thanks for the help! 🐱‍💻

PR #1399, opened 2020-01-26
Add labelUse loadFontFamily()

Bouh:
https://github.com/4ian/GDevelop/pull/1394#issuecomment-578529253

### DIFF
--- a/Extensions/BBText/JsExtension.js
+++ b/Extensions/BBText/JsExtension.js
@@ -521,16 +521,14 @@ module.exports = {
         .get('fontFamily')
         .getValue();
 
-      if (this._fontName !== fontResourceName) {
-        console.log('load font');
-
-        this._fontName = fontResourceName;
+      if (this._fontResourceName !== fontResourceName) {
+        this._fontResourceName = fontResourceName;
 
         this._pixiResourcesLoader
           .loadFontFamily(this._project, fontResourceName)
           .then(fontFamily => {
             // Once the font is loaded, we can use the given fontFamily.
-            this._pixiObject.textStyles.default.fontFamily = fontResourceName;
+            this._pixiObject.textStyles.default.fontFamily = fontFamily;
           })
           .catch(err => {
             // Ignore errors
--- a/Extensions/BBText/JsExtension.js
+++ b/Extensions/BBText/JsExtension.js
@@ -522,7 +522,7 @@ module.exports = {
         .getValue();
 
       this._pixiResourcesLoader
-        .loadFontFamily(this.project, fontFamily)
+        .loadFontFamily(this._project, fontFamily)
         .then(fontFamily => {
           // Once the font is loaded, we can use the given fontFamily.
           this._pixiObject.textStyles.default.fontFamily = fontFamily;
--- a/Extensions/BBText/JsExtension.js
+++ b/Extensions/BBText/JsExtension.js
@@ -520,7 +520,17 @@ module.exports = {
         .getProperties(this.project)
         .get('fontFamily')
         .getValue();
-      this._pixiObject.textStyles.default.fontFamily = this._pixiResourcesLoader.loadFontFamily(this.project, fontFamily);
+
+      this._pixiResourcesLoader
+        .loadFontFamily(this.project, fontFamily)
+        .then(fontFamily => {
+          // Once the font is loaded, we can use the given fontFamily.
+          this._pixiObject.textStyles.default.fontFamily = fontFamily;
+        })
+        .catch(err => {
+          // Ignore errors
+          console.warn('Unable to load font family', err);
+        });
 
       const wordWrap = this._associatedObject
         .getProperties(this.project)
--- a/Extensions/BBText/JsExtension.js
+++ b/Extensions/BBText/JsExtension.js
@@ -516,21 +516,27 @@ module.exports = {
         .getValue();
       this._pixiObject.textStyles.default.fontSize = `${fontSize}px`;
 
-      const fontFamily = this._associatedObject
+      const fontResourceName = this._associatedObject
         .getProperties(this.project)
         .get('fontFamily')
         .getValue();
 
-      this._pixiResourcesLoader
-        .loadFontFamily(this._project, fontFamily)
-        .then(fontFamily => {
-          // Once the font is loaded, we can use the given fontFamily.
-          this._pixiObject.textStyles.default.fontFamily = fontFamily;
-        })
-        .catch(err => {
-          // Ignore errors
-          console.warn('Unable to load font family', err);
-        });
+      if (this._fontName !== fontResourceName) {
+        console.log('load font');
+
+        this._fontName = fontResourceName;
+
+        this._pixiResourcesLoader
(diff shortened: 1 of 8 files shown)


Extensions/BBText/JsExtension.js
```diff
@@ -520,10 +520,7 @@ module.exports = {
         .getProperties(this.project)
         .get('fontFamily')
         .getValue();
-      this._pixiObject.textStyles.default.fontFamily = this._pixiResourcesLoader.getFontFamily(
-        this.project,
-        fontFamily
-      );
+      this._pixiObject.textStyles.default.fontFamily = this._pixiResourcesLoader.loadFontFamily(this.project, fontFamily);
```
4ian: `loadFontFamily` returns a promise containing the result of the loading, which is the name of the font family to be used. You can't affect it to font family, this won't work (you're affecting a Promise, which is an object, to a string).

Instead, look at how it's done in: https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/ObjectsRendering/Renderers/RenderedTextInstance.js

The reason for having a promise is because the IDE has to load the font - it can't return to you the font family immediately.

4ian:
Looks good! Can you give a try to a few different fonts (choose a font, see if it's displayed, choose another font, see if displayed, clear the font, see if we're back to arial?) before we go ahead with this? Thanks!!

Extensions/BBText/JsExtension.js
```diff
@@ -520,10 +520,17 @@ module.exports = {
         .getProperties(this.project)
         .get('fontFamily')
         .getValue();
-      this._pixiObject.textStyles.default.fontFamily = this._pixiResourcesLoader.getFontFamily(
-        this.project,
-        fontFamily
-      );
+
+      this._pixiResourcesLoader
+        .loadFontFamily(this.project, fontFamily)
```
4ian: this._project

Bouh:
Font are correctly displayed, but console log is crazy and display at  each frame
image

4ian:
So two things:
- It's surely because the PixiResourcesLoader should return "Arial" when you request an empty font resource name (empty string: ""). If I'm not mistaken, these error messages are shown only when the font is empty, right?
- We're lucky to have these because actually there is an important optimization to do! The font should be reloaded when it's changed :) To do this, we should do like in the Renderer for the text objects:

```
if (this._fontName !== xxx.getFontName()) {
    //Avoid calling loadFontFamily if the font didn't changed.
    this._fontName = xxx.getFontName();
    this._pixiResourcesLoader
      .loadFontFamily(this._project, xxx.getFontName())
      ...
}
```

see the idea?

Bouh:
Warnings stay if i add 
```
if (this._pixiObject.textStyles.default.fontFamily !== fontFamily){
 this._pixiResourcesLoader.loadFontFamily().then().catch().....
}

```
Because `fontFamily `from the input is empty. And `this._pixiObject.textStyles.default.fontFamily` is set to `Arial`.
I can't remove `Arial` in `bbTextStyles.default.fontFamily` if i remove it all game with default font will be broken. When i remove this `Arial` font change to another font by default, (Times New Roman ?).

I can do this maybe ?
```
 if (this._pixiObject.textStyles.default.fontFamily !== fontFamily && fontFamily != ""){
 this._pixiResourcesLoader.loadFontFamily().then().catch().....
}
```
I'am find my solution is dirty but working :)

Bouh:
Oooh format is not same, so String of font are never equal.

`this._pixiObject.textStyles.default.fontFamily`  give  `indieflower-ttf`
`fontFamily `(from input resource selector) `IndieFlower.ttf`

4ian:
Double check my example, you must store the font resource name, not the font family :)

Bouh:
It's working like that. I don't know what is your xxx but i can't use getFontName() there is not a text object.
And pixi-multistyle-text define the input as fontFamily, but we need pass it font resource name with gd 🤔.
Can't merge fontFamily and fontName?
If that's it, I don't get it...

Extensions/BBText/JsExtension.js
```diff
@@ -516,14 +516,27 @@ module.exports = {
...
+      if (this._fontName !== fontResourceName) {
+        console.log('load font');
+
+        this._fontName = fontResourceName;
+
+        this._pixiResourcesLoader
+          .loadFontFamily(this._project, fontResourceName)
+          .then(fontFamily => {
+            // Once the font is loaded, we can use the given fontFamily.
+            this._pixiObject.textStyles.default.fontFamily = fontResourceName;
```
4ian: this._pixiObject.textStyles.default.fontFamily = fontFamily

4ian:
Right my explanations were lacking of this nuance. So we have two different things and a mistake in the BBText object:

- The font resource name is the name of the resource in GDevelop. That how you refer to resources in GDevelop: using a resource name (usually, it's automatically set to the name of the file of the resource).

- A "font family" is a notion that does not exist in GDevelop itself. It's a CSS/HTML/pixi-multistyle-text notion. A font family is the name of a font that refers to a CSS font (to keep things simple).

And so we have a mistake in BBText object: 
- this field called "font family" is not a font family! It should have been called "fontResourceName".

We should probably in a next PR fix this.

To summarize, the flow is like this:

```
You get the font resource name -> you call GDevelop resource loader, giving it the resource name (-> GDevelop resource loader loads the font for you) -> you get back a font family (a string) -> you pass it to PIXI -> your font is displayed!
```

> fontFamily is then() is useless

No because, you're getting the font family that you assign to the pixi object :)

Extensions/BBText/JsExtension.js
```diff
@@ -516,14 +516,27 @@ module.exports = {
         .getValue();
       this._pixiObject.textStyles.default.fontSize = `${fontSize}px`;
 
-      const fontFamily = this._associatedObject
+      const fontResourceName = this._associatedObject
```
4ian: This is the correct name 👍

Extensions/BBText/JsExtension.js
```diff
@@ -516,14 +516,27 @@ module.exports = {
...
         .getValue();
-      this._pixiObject.textStyles.default.fontFamily = this._pixiResourcesLoader.getFontFamily(
-        this.project,
-        fontFamily
-      );
+
+      if (this._fontName !== fontResourceName) {
+        console.log('load font');
+
+        this._fontName = fontResourceName;
```
4ian: We could even rename `this._fontName` to `this._fontResourceName` (to make extra clear what it represents). I'm ok with both

Extensions/BBText/JsExtension.js
```diff
@@ -516,14 +516,27 @@ module.exports = {
...
+
+        this._pixiResourcesLoader
+          .loadFontFamily(this._project, fontResourceName)
+          .then(fontFamily => {
```
4ian: This is also correct

4ian:
Looks good now! Let me know if that works for you and we can merge this.

Bouh:
Yes working fine, thank you :)

4ian:
Thanks for taking care of that :)

PR #1406, opened 2020-02-01
Upgrade to Emscripten 1.39.6

4ian:
Upgrade to latest Emscripten. This has the good side effect of simplifying a bit the build, but the linking phase is taking ages (4min on my Macbook Pro :/).
We can also enable WebAssembly for Electron and get performance boost of ~30% on operations like export or game opening :) 

- [x] Fix tests (by refactoring all `gd = global.gd` to an import of a module that is returning `global.gd`)?
- [x] Investigate build options to make linking faster at least during development.


4ian:
I'll give this a try on Windows and will merge this if it's all right :)

blurymind:
Very cool stuff :)

Issue #991, opened 2019-03-27
Tab key cannot be used everywhere


PascalLadalle:
## Describe the bug
The Tabulation key doesn't always allow to move on to the next field.

## To Reproduce
- Click on _Add condition_, and pick _Value of a scene variable_, for instance.
- Type the name you want or use the arrows.
- Press Tab twice to access the operator.
- Press down arrow to pick the operator.
- Press Tab to move to the next field. Nothing happens.

blurymind:
this one is important to address in terms of making it possible to use keyboard only when setting the event sheet for example. Tabbing and shift-tabbing to previous/next focus neighbour can dramatically reduce the number of steps in setting things up.

When we tab to things, the hands stay at the keyboard and its one step to focus on next/previous.

4ian:
The specific example reported here seems to be fixed. Can we close this unless other buttons/fields are not "tab-accessible"?

In theory everything that is shown on screen and that is a proper component has support for being browsed with tab, and after opening a dialog you should always be able to press tab in case you don't have the focus to select the first element of the dialog.

PascalLadalle:
not optimal, but seems fully functional. 👏 

In the same fashion, it would be nice if Alt gave access to the main menu as in most of Windows programs. And if ctrl-tab could switch between tabs, just how wonderous would it be! 🌈 😁

PR #1408, opened 2020-02-02
Event sheet - add instructions alt (menu instead of pop up) workflow

blurymind:
This experimental PR adds two things:

- Better display of empty (null) parameters in the event sheet. Making them easy to notice and clickable
gdShowNulls

- Right clicking on the add new condition/action button triggers a new quick menu to select the object and its action/condition
GD-clickteamesque-additionOfEvents

The old left click dialog behaviour is preserved - so the old design that requires more clicks is untouched!

Still todo:
- [x] Fix flow
- [ ] remove duplicated code

Why I wanted to have this workflow
- the modal pop up keeps making my brain switch between what is on the event sheet and where I place the new instruction and the parameteres of the instruction. It also required a few extra clicks. It was kind of affecting the flow.
- I thought about why do I like the way Clickteam does their event sheet and to a big extend this was the reason. They let you left click to place new items on the event sheet and the selection is fully in the context menu. Our approach takes this a step further and offers a menu that can be filtered


4ian:
> Better display of empty (null) parameters in the event sheet. Making them easy to notice and clickable

Love that :) We might want to make the underline not red for parameters that are optional. But that's a really great idea!

> Right clicking on the add new condition/action button triggers a new quick menu to select the object and its action/condition

It's true that I was not convinced with this flow but now that I'm seeing it... it looks really good! 👍I can understand it's indeed breaking less the creation flow when you know what you want to do. The choice of the right click for "Add action"/"Add condition" seems pretty good, as it's like a context menu but new users will use the left click and so start with the "full editor", showing the parameter descriptions at the end.

Let me take a look at the code.
We'll have potentially to create new Storybook stories with this new component (just for completeness of the storybook). 
I see there is a bit of opacity, not sure if it's not a bit risky in terms of consistency with the rest of the interface/performance? I understand it's for "breaking less" the flow but I have mixed feelings about it.
Let's also make sure it works properly in terms of keyboard navigation (namely you should be able to right click, type to search an object or action/condition, press enter to select the first one, then type to search an action/condition and press enter to select the first one. Should hopefully works out of the box but let's have a look.

Great job in any case! This looks already really cool :)

Bouh:
See missing parameters are really nice!
Nothing to said about quick instruction editor, it's like a big editor but in tiny :D 
We can only open it why right click on "Add action" or "Add condition" then i've seen no problem with other shortcuts or selections.

- Purple/white line under type (Objects / Other Actions) are not complete at the first launch.
image

- Good thing popover move in a other place where place is available if space is too tiny for it.
I can't put the popover out of window, well done! :+1: 

- When we use search bar the Popover component change of width. And again when you select an object. Can you use a fix width ?

- Take a look on this condition, there are just red dot on first parameters O.o
image

- Can you remove the scroolbar-x when you edit value in Popper component on you gif ?
Scrollbar appear when focus is active on input.

- Tag filter work perfectly, no problem of z-index :+1:

newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/index.js
```diff
@@ -95,7 +95,10 @@ export default class ExpressionField extends React.Component<Props, State> {
   }
 
   focus() {
-    if (this._field) this._field.focus();
+    if (this._field) {
+      this._field.focus();
+      this._doValidation();
```
4ian: Can you triple check if there is no weird side effect? Validation can also take a bit of time during which it can froze the UI, so running it on focus might be risky? I would rather use `this._enqueueValidation`. `this._enqueueValidation` will trigger a validation 500ms after the last input. This means that the user can start typing immediately - without having a lag in case the validation is taking too much time - or can just wait half a second and see any errors showing.

(of course if the validation was instant it would not be a problem, but we must be prepared for the case where the validation is taking a few milliseconds, which can be noticeable to the user eye, especially if we're blocking typing)

4ian:
> When we use search bar the Popover component change of width. And again when you select an object. Can you use a fix width ?

A fixed width should help yeah, surely on the popover and/or content (in theory, the content should adapt to the parent size, so the width should be on the popover if possible, or on something as high in the component tree of the popover as possible).

As for the width to use, it would be interesting to know what's the current width being used. We'll probably disable this popover on phones (screen is too small and hard to precisely touch the parameters, we might as well always show the existing editor on full page). This means that we can use something like ~400 or ~500px width? Nothing too large though, because otherwise we'll lose the feeling of this being almost a "context menu" :)

> Purple/white line under type (Objects / Other Actions) are not complete at the first launch

Might be because the width is varying, fixing the width of the popover and/or content might help.

blurymind:
> Love that :) We might want to make the underline not red for parameters that are optional. But that's a really great idea!

Thank you :) what would be the best way to detect optional ones? All this is currently done using css parameter class: empty , so ideally I would like to capture optional empty ones via a different class name

> It's true that I was not convinced with this flow but now that I'm seeing it... it looks really good! 👍I can understand it's indeed breaking less the creation flow when you know what you want to do. The choice of the right click for "Add action"/"Add condition" seems pretty good, as it's like a context menu but new users will use the left click and so start with the "full editor", showing the parameter descriptions at the end.
> 
😌 I am really happy you approve having this alternative approach. Yes I thought of it kind of like an streamlined minimalist way of quickly adding new instructions without any obstructions - for users who have learned the engine and don't need the help mes but want to go faster. Its a parallel with how we can edit parameters of instructions directly in the event sheet or in the dialog pop up. With this we will also be able to directly create new even instructions directly in the event sheet and then proceed to edit them in the event sheet without any pop ups.

> Let's also make sure it works properly in terms of keyboard navigation (namely you should be able to right click, type to search an object or action/condition, press enter to select the first one, then type to search an action/condition and press enter to select the first one. Should hopefully works out of the box but let's have a look.

Yes I need to look into making this autofocus the search, let you select with up/down and enter to submit. Ideally with good keyboard navigation it will be super fast to select and add new instructions. Then once added, I want to automatically put the focus on the first empty parameter, so the user can start typing right away. I want to make this really fast

newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/index.js
blurymind: updated now :)

blurymind:
> * Purple/white line under type (Objects / Other Actions) are not complete at the first launch.
>   image

> * When we use search bar the Popover component change of width. And again when you select an object. Can you use a fix width ?

Fixed the width change. The tab underline looks like a problem with material ui rendering. If you inspect the tab buttons- their width is correct, the underline is however not properly rendered initially. If you select the second that and then the first one again - it fixes itself.


> * Take a look on this condition, there are just red dot on first parameter O.o
>   image

This looks more like actions. 
In both cases the underlined fields are rendered correct - those are clickable and you should set their values. The first field of among objects is a field to select an object.

> * Can you remove the scroolbar-x when you edit value in Popper component on you gif ?

That was there before the PR. I can try to fix it :)

newIDE/app/src/EventsSheet/InstructionEditor/InstructionOrObjectSelector.js
```diff
@@ -224,14 +224,17 @@ export default class InstructionOrObjectSelector extends React.PureComponent<
                   }
                   style={styles.searchBar}
                   ref={this._searchBar}
+                  buttonsTabIndex="-1"
                 />
                 {!isSearching && (
                   <Tabs value={currentTab} onChange={onChangeTab}>
                     <Tab
+                      tabIndex="-1"
```
4ian: I highly recommend against doing any kind of magic trick with tabIndex. It's like zIndex: as soon as you're writing one, you're almost always ending up doing weird things all over the place.

What is the thing you're trying to solve?
Focus a field on mounting? Prefer componentDidMount (or useEffect in newer React components) and check if there is a prop on the material-ui components that you can expose.

newIDE/app/src/UI/SearchBar.js
```diff
@@ -29,6 +29,8 @@ type Props = {|
   buildTagsMenuTemplate?: () => any,
   /** If defined, a help icon button redirecting to this page will be shown */
   helpPagePath?: ?string,
+  /** If defined, the tabIndex value is applied to all buttons. Useful for taking them out of the focus order when desired via -1 */
+  buttonsTabIndex?: string,
```
4ian: Why taking buttons out of the focus? This can reduce accessibility for people for which using a mouse is difficult. I don't mind 1 or 2 extra key press on the Tab key. In theory there is already a callback on the SearchBar that is set so that when you search something and press enter, the first result is selected (at least for the object/actions/conditions selector)

newIDE/app/src/UI/Theme/DarkTheme/EventsSheet.css
```diff
@@ -125,6 +125,15 @@
   color: #e8dc59;
 }
 
+.gd-events-sheet-dark-theme .name-and-icon-container:empty,
```
4ian: Can we double check if we need the `name-and-icon-container` rule?
From what I understand you added it because otherwise the background was not showing for objects (and variables?

In theory you should not need to do it on both, because name-and-icon-container is, at the name implies, just a wrapper around a text and an icon. It might be used later for other things totally unrelated to parameters. See the generated markup:

image

newIDE/app/src/EventsSheet/index.js
```diff
@@ -397,8 +400,11 @@ export default class EventsSheet extends React.Component<Props, State> {
       );
     }
 
+    console.log(instructionContext, this.state);
     this.setState({
+      inlineEditingAnchorEl: instructionContext.domEvent,
       editedInstruction: {
+        quick: instructionContext.quick,
```
4ian: "quick" is a bit generic, I would go for something more specific like: `isInlineInstructionEditor`. Easier to search in the codebase too and you immediately understand it's related to "an editor of instruction that should be shown inline or not". Finally the "is" hints as this being a boolean.

newIDE/app/src/EventsSheet/InstructionEditor/NewInstructionEditorMenu.js
```diff
@@ -0,0 +1,351 @@
...
+|};
+
+const findInstruction = (
+  list: Array<EnumeratedInstructionOrExpressionMetadata>,
+  instructionType: string
+): ?EnumeratedInstructionOrExpressionMetadata => {
+  return list.find(({ type }) => type === instructionType);
+};
+
+// TODO move to utils
```
4ian: There is indeed some clever refactoring to do here - and maybe a bit of the logic inside the component methods too, to avoid duplication with the existing editor.
A bit of copy-pasta is not a big deal, but as possible would be good to reduce the copied lines, especially considered that there is also the old editor (but that will be removed sooner or later).

For now don't worry about this though! Let's get right the rest and I'm happy to help identifying what we could factor. In a sense, I prefer a bit of copy-pasta rather than too much refactoring that make the code looks like spagetthi code :)

newIDE/app/src/EventsSheet/EventsTree/InstructionsList.js
```diff
@@ -184,6 +197,7 @@ export default class InstructionsList extends React.Component<Props, State> {
                   style={styles.addButton}
                   className="add-link"
                   onClick={this.onAddNewInstruction}
+                  onMouseDownCapture={this.onQuickAddNewInstruction}
```
4ian: I've never seen onMouseDownCapture! What is this exactly? Do you think we can use onContextMenu instead? This means that we must remove the existing code in onContextMenu because we're replacing the menu by this.

newIDE/app/src/EventsSheet/EventsTree/InstructionsList.js
```diff
@@ -69,6 +69,19 @@ export default class InstructionsList extends React.Component<Props, State> {
       });
   };
 
+  onQuickAddNewInstruction = (e: any) => {
+    if (e.button !== 2) return;
```
4ian: We won't need this with `onContextMenu` - in general prefer to avoid hardcoding mouse button identifiers because on touch enabled device, context menu can be enabled by other things than a right click (a long press can do the same job on a tablet) and check callbacks offered by the UI components (i.e: Material UI components) or React itself if you're dealing with a div/span/button/etc.

newIDE/app/src/EventsSheet/index.js
4ian: Ah actually you're storing a DOM event there! Found that by reading the code (good example about avoiding short names ;)).

In this case, I would rather avoid storing DOM events as much as possible - it's "leaking" too much information that is potentially useless.
Instead of `quick`, I would pass back from the "Add Action" button the fact that we're requiring an inline instruction editor (`isInlineInstructionEditor`, a boolean) and the position on screen: `e.clientX, e.clientY`, which can be stored in `inlineInstructionEditorX` and `inlineInstructionEditorY`.

This is what we do for menus. Note that X/Y means that this won't stay at the proper position if we scroll... but I think it's not a big deal and we should maybe disable scroll when a menu/popover is opened. It might be already the case actually.

newIDE/app/src/EventsSheet/SelectionHandler.js
```diff
@@ -5,13 +5,17 @@ import values from 'lodash/values';
 export type InstructionsListContext = {|
   isCondition: boolean,
   instrsList: gdInstructionsList,
+  quick?: any,
+  domEvent?: any,
```
4ian: I'm a bit confused, do we need both? Can we just store one? Do we even need these?

newIDE/app/src/EventsSheet/InstructionEditor/InstructionOrObjectSelector.js
blurymind: I want to optionally remove all the buttons from the focusing order when you tab, so as to make it focus on the first item in the list when hitting tab the first time

newIDE/app/src/UI/SearchBar.js
blurymind: I want this workflow to work quick, so when you search, tabbing takes you straght to the list items and lets you select one in the result then hit enter. I hind it much less likely that I will want to tab to any buttons. I want tabbing to take me to the results, If I am tabbing it is more likely that I saw something in the list I want to select :)

newIDE/app/src/EventsSheet/EventsTree/InstructionsList.js
blurymind: I need to capture the left click to trigger this. Do we have another way already? I need to have another look

newIDE/app/src/EventsSheet/EventsTree/InstructionsList.js
blurymind: Will have another look. I was wondering if we would want to have this available on touch devices which are likely to be small, but if we have some mechanic to detect long press I could add it?

newIDE/app/src/EventsSheet/SelectionHandler.js
blurymind: likely a flowtype error. Maybe quick should really be called quickAddMenuTarget?

newIDE/app/src/EventsSheet/InstructionEditor/NewInstructionEditorMenu.js
blurymind: Was wondering where in utils to move this. I really dont like the copy pasta

newIDE/app/src/EventsSheet/index.js
blurymind: I think it already is the target, but you're right the name is awful.
quickAddMenuTarget sound better?

newIDE/app/src/UI/Theme/DarkTheme/EventsSheet.css
blurymind: I had to add it in order to cover for object selectors that were empty. Remember this css rule only targets spans of that class with nothing in them

newIDE/app/src/EventsSheet/InstructionEditor/InstructionOrObjectSelector.js
blurymind: We need to somehow take out some of the elements from the focus order - not just here, but the entire app. Remember this https://github.com/4ian/GDevelop/issues/991

newIDE/app/src/UI/Theme/DarkTheme/EventsSheet.css
4ian: I mean, as shown on my screenshot, instruction-parameter is already there - so there should not be a need for adding the same things to name-and-icon-container? It's duplicating style it seems.

newIDE/app/src/UI/SearchBar.js
4ian: Sure but in this case I would prefer to have Up/Down buttons allowing you to change the selection and "Enter" choose the currently selected object or action/condition (a bit like in text editors like when doing Ctrl+P in VSCode).

Tab should still allow to iterate on all buttons - which is the default thing in all dialogs and which allow GDevelop dialogs to be almost always entirely usable with the keyboard.

But in any case, let's do these improvements on a separate PR - I would strongly recommend to go the up/down button + enter way as it seems more standard (a bit like a dropdown)

newIDE/app/src/EventsSheet/InstructionEditor/InstructionOrObjectSelector.js
4ian: See my other comment, I think the proper solution here is reworking the selector editors so that you can press up/down to select something in the list - then hit Enter to select.

Currently we already have "half" of this feature, as pressing Enter select the first search result. Better than nothing - and I prefer to have "only" that for now rather than putting buttons of tabbing.

newIDE/app/src/EventsSheet/InstructionEditor/NewInstructionEditorMenu.js
4ian: This should be a helper file in the same directory I think

newIDE/app/src/EventsSheet/EventsTree/InstructionsList.js
4ian: Doesn't onContextMenu work for the right click?

newIDE/app/src/EventsSheet/SelectionHandler.js
4ian: Instead of quick I would really prefer `inlineInstructionEditor`. So something like `inlineInstructionEditorAnchorEl` (to mimic the naming of `inlineEditingAnchorEl`)

newIDE/app/src/UI/SearchBar.js
blurymind: How about removing them from focus order only when the user has entered something in the search box? It will also simplify it so less props to pass. I wanted to have it tab so it can be used with the left hand :) the up/down approach is going to be hard, since the list component alr ady has that mapped to scrolling

newIDE/app/src/UI/Theme/DarkTheme/EventsSheet.css
blurymind: Try removiny it :)  in the case of objects it's not empty, it's child is empty

newIDE/app/src/EventsSheet/EventsTree/InstructionsList.js
blurymind: I will give it a try tomorrow 😊 thank you

newIDE/app/src/EventsSheet/InstructionEditor/NewInstructionEditorMenu.js
blurymind: Utils.js ?

newIDE/app/src/UI/SearchBar.js
blurymind: Btw hitting enter already selects the object. You can type to s arch, tab to an object and hit enter, then search straight away for action or condition, tab to it again and hit enter again.

newIDE/app/src/EventsSheet/EventsTree/InstructionsList.js
blurymind: @4ian onContextMenu is already used to trigger back at the event sheet instructionsListContextMenu. Is it ok to completely overwrite that? Not sure exactly what that does, apart of making a menu with undo/redo commands in it

newIDE/app/src/EventsSheet/index.js
blurymind: using onInstructionsListContextMenu to trigger the event would remove the need to add this target I think

newIDE/app/src/EventsSheet/InstructionEditor/InstructionOrObjectSelector.js
blurymind: Ok in that case I will remove this from the PR and will have to revisit it in another :)

newIDE/app/src/EventsSheet/SelectionHandler.js
blurymind: fixed now

newIDE/app/src/EventsSheet/index.js
blurymind: actually I kept the target in order to keep consistency

newIDE/app/src/UI/SearchBar.js
blurymind: Ok I am removing this now. Maybe another PR :)

newIDE/app/src/EventsSheet/EventsTree/InstructionsList.js
blurymind: ok I refactored it to use the already established mechanism with onContextMenu

blurymind:
Applied the review notes.. still have loads of flow errors to resolve. Hopefuly I havent broken something

blurymind:
Ok this should fix the flow errors. If there is nothing else, I can move some of the code to be reused?

4ian:
Let me take over this tonight and hopefully merge it :) I'll take a look at:

- [ ] See if we can remove `name-and-icon-container:empty` style.
- [ ] Replace `any` typing by HTMLDomElement or something like this.
- [ ] See what we can refactor in an util file.
- [ ] Add the component in a Storybook story.

The refactoring is important to avoid too much redundancy, but I have to get the code and try things to get a feeling of what is worth factoring and what can stay duplicated.

blurymind:
Thank you @4ian 😄

4ian:
Unfortunately pushing commits with a merge from master added all the commits from master in this PR - even if the branch itself is clean. I think it's a GitHub issue. If you close this PR and re-open immediately another one with the same branch it will be showing the commits properly.

4ian:
Quick note to say this is merged and I've also fixed the extra useless horizontal scrollbar on the master branch (was not related to this PR). 👍
Really cool feature!

blurymind:
Thank you @4ian :tada:  I am going to play with this today

PR #1416, opened 2020-02-06
Event sheet - add instructions alt (menu instead of pop up) workflow

blurymind:
This experimental PR adds two things:

- Better display of empty (null) parameters in the event sheet. Making them easy to notice and clickable
gdShowNulls

- Right clicking on the add new condition/action button triggers a new quick menu to select the object and its action/condition
GD-clickteamesque-additionOfEvents

The old left click dialog behaviour is preserved - so the old design that requires more clicks is untouched!

Still todo:
- [x] Fix flow
- [ ] remove duplicated code

Why I wanted to have this workflow
- the modal pop up keeps making my brain switch between what is on the event sheet and where I place the new instruction and the parameteres of the instruction. It also required a few extra clicks. It was kind of affecting the flow.
- I thought about why do I like the way Clickteam does their event sheet and to a big extend this was the reason. They let you left click to place new items on the event sheet and the selection is fully in the context menu. Our approach takes this a step further and offers a menu that can be filtered


blurymind:
Closed the old PR at https://github.com/4ian/GDevelop/pull/1408 and reopened a new one with the same name, as requested by @4ian :)

blurymind:
@4ian will you continue working on this. I think the remaining work is
 - [ ] See what we can refactor in an util file.
  - [ ] Add the component in a Storybook story.

4ian:
Yes the update for fixing the CSS took longer than expected (because I refactored the parameters so that we use components rather than the `:empty` css modifier). I've started the refactoring of the instruction editor logic in a React "hook", no time today or tomorrow but should be finished this week-end :)

blurymind:
@4ian thank you for picking this up from me. Let me know if you need me to do anything more here. 

I had a look at the forum today and found some mentions and simple yarn bugs. Will submit fixes for those in another pr

4ian:
I've done the refactoring by extracting the common logic to a custom React hook, pretty happy about the result :) No more time for this for now, I will push the changes later - still have a few small things to factor/improve and should be good to go!

blurymind:
Thank you @4ian :smile:  I cant wait to start using this :nerd_face:

4ian:
Refactoring pushed, should hopefully be ready to go! Let me know if you try and see any regression, seems to work well though from what I've tested.

4ian:
Merged! Thanks a lot for working on this, this will be very useful and will certainly speed up the creation workflow/make it less disruptive when working intensively on events :) 
For now I've not disabled this on small screens - if it's unusable I'll do that.

As this is a particularly important editor let me know if you see any regression :)

PR #1112, opened 2019-06-15
Dialogue tree extension - WIP

blurymind:
Decided to start a PR in order to get some feedback. :)
It has all the features now - even loading and saving of state (what choices the user has made , what variables were set with commands (see the files in the test folder) - this can be used to store the choices with the save game.
bondageJsInGd4

Still todo:
- [x] Merge the 3 dialogue type conditions into one with a dropdown to select one of the three
- [x] apply review notes
- [x] See if the wording on some can be shorter/cleaner
- [x] Add Action to load from a json file resource - once we have that :) 
- [x] Do more testing
- [x] Add function decorators
- [ ] Make a proper example project (looking for some nice free graphics) - Anyone interested to help me with that is welcome
- [ ] Write a help page - Anyone interested to help me with that is welcome

Here is a project to play with:
[My project36.zip](https://github.com/4ian/GDevelop/files/3336233/My.project36.zip)

Once this is cool and ready, I can move onto getting yarn embedded as an external editor, so the dialogue tree writing experience is super nice and intuitive

PR #1418, opened 2020-02-06
Yarn fixes

blurymind:
- fixes bug mentioned here
https://forum.gdevelop-app.com/t/yarn-dialogue-system-fails-to-parse-json-file/22663/4
also addressing corner case where users try to load and use the json file at the beginning of the scene
- fixes yarn drops a character bug mentioned here
https://forum.gdevelop-app.com/t/yarn-drops-last-character-from-dialogue-node/21432/4

Might see to address more corner cases where improper use crashes the game. I wonder if we could show some more informative message when its not used properly - like a pop up or something. I could do console.error in a few places

### DIFF
--- a/Extensions/DialogueTree/dialoguetools.js
+++ b/Extensions/DialogueTree/dialoguetools.js
@@ -284,10 +284,10 @@ gdjs.dialogueTree.getLineOptionsCount = function() {
  */
 gdjs.dialogueTree.confirmSelectOption = function() {
   if (
-    this.dialogueIsRunning &&
-    this.dialogueData.select &&
-    !this.selectedOptionUpdated &&
-    this.selectedOption !== -1
+    !this.dialogueIsRunning ||
+    (this.dialogueData.select &&
+      !this.selectedOptionUpdated &&
+      this.selectedOption !== -1)
   ) {
     this.commandCalls = [];
     try {
@@ -307,7 +307,6 @@ gdjs.dialogueTree.confirmSelectOption = function() {
  * Select next option during Options type line parsing. Hook this to your game input.
  */
 gdjs.dialogueTree.selectNextOption = function() {
-  if (!this.dialogueIsRunning) return;
   if (this.dialogueData.select) {
     this.selectedOption += 1;
     this.selectedOption = gdjs.dialogueTree._cycledOptionIndex(
@@ -321,7 +320,6 @@ gdjs.dialogueTree.selectNextOption = function() {
  * Select previous option during Options type line parsing. Hook this to your game input.
  */
 gdjs.dialogueTree.selectPreviousOption = function() {
-  if (!this.dialogueIsRunning) return;
   if (this.dialogueData.select) {
     this.selectedOption -= 1;
     this.selectedOption = gdjs.dialogueTree._cycledOptionIndex(
@@ -336,7 +334,6 @@ gdjs.dialogueTree.selectPreviousOption = function() {
  * @param {number} optionIndex The index of the option to select
  */
 gdjs.dialogueTree.selectOption = function(optionIndex) {
-  if (!this.dialogueIsRunning) return;
   if (this.dialogueData.select) {
     this.selectedOption = gdjs.dialogueTree._normalizedOptionIndex(
       this.selectedOption
@@ -350,7 +347,6 @@ gdjs.dialogueTree.selectOption = function(optionIndex) {
  * @returns {number} The index of the currently selected option
  */
 gdjs.dialogueTree.getSelectedOption = function() {
-  if (!this.dialogueIsRunning) return;
   if (this.dialogueData.select) {
     return this.selectedOption;
   }
@@ -365,7 +361,6 @@ gdjs.dialogueTree.getSelectedOption = function() {
  * @returns {boolean} true if the selected option was updated since the last call to this function
  */
 gdjs.dialogueTree.hasSelectedOptionChanged = function() {
-  if (!this.dialogueIsRunning) return;
   if (this.selectedOptionUpdated) {
     this.selectedOptionUpdated = false;
     if (this.selectedOption === -1) this.selectedOption = 0;
@@ -402,7 +397,6 @@ gdjs.dialogueTree.isDialogueLineType = function(type) {
  * @param {string} branchName The Dialogue Branch name you want to check.
  */
 gdjs.dialogueTree.hasDialogueBranch = function(branchName) {
-  if (!this.dialogueIsRunning) return false;
   return (
     this.runner &&
     this.runner.yarnNodes &&
--- a/Extensions/DialogueTree/dialoguetools.js
+++ b/Extensions/DialogueTree/dialoguetools.js
@@ -271,7 +271,7 @@ gdjs.dialogueTree.getLineOptionsTextVertical = function(optionSelectionCursor) {
  * @returns {number} The number of options
  */
 gdjs.dialogueTree.getLineOptionsCount = function() {
-  if (!this.dialogueIsRunning || this.options.length) {
+  if (this.dialogueIsRunning && this.options.length) {
     return this.optionsCount;
   }
   return 0;
@@ -283,11 +283,11 @@ gdjs.dialogueTree.getLineOptionsCount = function() {
  * This will advance the dialogue tree to the dialogue branch was selected by the player.
  */
 gdjs.dialogueTree.confirmSelectOption = function() {
+  if (!this.dialogueIsRunning) return;
   if (
-    !this.dialogueIsRunning ||
-    (this.dialogueData.select &&
-      !this.selectedOptionUpdated &&
-      this.selectedOption !== -1)
+    this.dialogueData.select &&
+    !this.selectedOptionUpdated &&
+    this.selectedOption !== -1
   ) {
     this.commandCalls = [];
     try {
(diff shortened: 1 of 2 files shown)


blurymind:
One thing that sort of bugs me is that for some reason the music in the yarn example project doesnt start. I wonder whether to update the example to not have any music

blurymind:
So it looks like some users are trying to load yarn data, go to a node and read text - all at the beginning of the scene. 

The thing is, if getLine is called at the beginning of the scene ( can be detected via runtimeScene.getTimeManager().isFirstFrame()) my strExpression that returns the line needs to return it later - not at the beginning of the scene. That doesnt seem possible, returning with a delay to gdjs, so I think for now people who want to do that will have to do it with a timer

4ian:
> One thing that sort of bugs me is that for some reason the music in the yarn example project doesnt start. I wonder whether to update the example to not have any music

This is a separate issue, see the warning at the top of this page: http://wiki.compilgames.net/doku.php/gdevelop5/all-features/audio

This should arguably be made more visible with a warning in GDevelop.

> Can a strExpression be queled to return at the next frame?

No, if the user is calling an expression, it needs it now, not later :) 

> my strExpression that returns the line needs to return it later 

Why does it needs to return it later? What's the reason? Is there something missing? Can you load it? If not can you expose something to the user, like a condition, to tell that it's "loading" or something like this?

> This pr prevents this setup from crashing their game, but it doesnt enable people to do it all at the beginning

That's a good start already :) An extension should never crash even if it's "badly" used. In other words, you should protect all your functions from crashing by ensuring you're in a proper state, or return early/ensure you're returning something that make sense or is empty (depends on the case).

In the case of "something is not ready", we should dig more to understand what is not ready (see my previous questions).

blurymind:
I remember vaguely that we tried to fix that when implementing the extension, but some limitation with json resource loading or bondage prevented us. We also decided that most people will not be triggering their dialogue at the start of a scene, but whenever they interact with something

see comment that I had for a while
https://github.com/4ian/GDevelop/pull/1112/commits/b8bd062943eed3a6f9f7561c9380648ed5fd8b73#diff-7b1447df68f389336e8a037c28663af7R334

been trying to retrace our discussion about it

4ian:
> Strange thing is that when I log it in terminal, the object with the dialogue text is there, but when i uncollapse it, I get a note that it is evaluated on the uncollapse

Use JSON.stringify on the object if you want a string representation that is guaranteed to be the one in memory at the time you console.log it.

> I keep thinking that this is some quirk with gdevelop's resource loading

While not totally impossible, there resource loading is very simple: you ask GDevelop for a resource, it's either returning it to you **in a callback** if in memory already or loading (if not in memory) and returning it to you **in a callback**.
But in both case **these operations are in a callback**. I remember when reviewing commenting multiple time about wrong expectation of the code of the extension, like trying to read data just after calling the method to load the resource, while it's **not guaranteed** that the callback will be called immediately. If the resource is not loaded, it might be loaded and callback called only one second later - or never. 
The extension must be resilient against this.

Note that GDevelop is now preloading the JSON by default... so in theory you should get the callback immediately called :) Which should conveniently allow people to immediately start their dialogue.

If it's not working, do as usual:
- reduce the issue to a game with just 1 or 2 events showing the issue.
- adding console.logs, use the debugger to put breakpoints, verify your assumptions with logs.

4ian:
I've updated my previous answer because it was originally not right.

There are 2 things to do now:
- Take a look at the code I wrote in the previous answer, and give it a try in the extension - I'm confident this should solve the issue of things not working in the "first" frame.
- Decide if we keep the logic I wrote here in the extension... OR if we move the if inside `loadJson` in [JsonManager](https://github.com/4ian/GDevelop/blob/24091db88be3c5d663699c1f16a3eb96d036ef01/GDJS/Runtime/jsonmanager.js#L91). Arguably, it might make JsonManager easier to use if we have a loadJson method that immediately calls the callback if the json is already in memory. This means that there would be no changes to do in the extension (well, your existing fixes to bulletproof the extension in case some decide to not load anything).

@blurymind What do you think? Are you able to try the fix?

blurymind:
> I've updated my previous answer because it was originally not right.
> 
> There are 2 things to do now:
> 
> * Take a look at the code I wrote in the previous answer, and give it a try in the extension - I'm confident this should solve the issue of things not working in the "first" frame.
> * Decide if we keep the logic I wrote here in the extension... OR if we move the if inside `loadJson` in [JsonManager](https://github.com/4ian/GDevelop/blob/24091db88be3c5d663699c1f16a3eb96d036ef01/GDJS/Runtime/jsonmanager.js#L91). Arguably, it might make JsonManager easier to use if we have a loadJson method that immediately calls the callback if the json is already in memory. This means that there would be no changes to do in the extension (well, your existing fixes to bulletproof the extension in case some decide to not load anything).
> 
> @blurymind What do you think? Are you able to try the fix?

Option B sounds like a better fix, as it will help other extensions be resilient to this problem.
I will try and see tonight if this fixes the problem in the extension :)  Thank you @4ian awesome feedback as usual 👍 😸

4ian:
In this case, let's try directly the fix in JsonManager :)

Before these lines: 
https://github.com/4ian/GDevelop/blob/24091db88be3c5d663699c1f16a3eb96d036ef01/GDJS/Runtime/jsonmanager.js#L106-L108

Let's check if the JSON is loaded and if yes return it synchronously:

```
// Don't fetch again an object that is already in memory
if (this._loadedJsons[resourceName]) {
  callback(null, this._loadedJsons[resourceName]);
  return;
}
```

and... that should be it :)

That should allow to both load the dialogue data AND use it in the same frame. It would still be great to add a warning in the doc or a paragraph mentioning that "You can deactivate the pre-loading of the JSON files containing the dialogues in the resource editor. This will make the initial loading of the game faster. During the game, the dialogue JSON file will be loaded when you use the action to load the dialogue from it. In this case though, this means that the dialogue won't be ready before the device of the player loads the file - which can take a few milliseconds or even a few seconds if the game is running in a web-browser. Use the condition "XXX" to check that the dialogue is ready before using the other condition and actions to display and interact with it"

(XXX to replace ;))

blurymind:
@4ian that worked like a charm,thank you :smile_cat:

blurymind:
@4ian thank you for the help. Now all Yarn parsing bugs mentioned at the forum have been fixed. Let me know if anything else needs to be addressed here :)

4ian:
In general I prefer to avoid extra parameters if possible. Instead if one day an extension needs for some reason to "forget" a JSON and reload it, we'll add a method to delete/remove a JSON resource from memory.
As these JSON are resources of the game, they are not supposed to change during the game execution, so we should not need this.

Extensions/DialogueTree/dialoguetools.js
```diff
@@ -266,7 +271,7 @@ gdjs.dialogueTree.getLineOptionsTextVertical = function(optionSelectionCursor) {
  * @returns {number} The number of options
  */
 gdjs.dialogueTree.getLineOptionsCount = function() {
-  if (this.options.length) {
+  if (!this.dialogueIsRunning || this.options.length) {
```
4ian: Are you sure about your condition?

Extensions/DialogueTree/dialoguetools.js
```diff
@@ -279,9 +284,10 @@ gdjs.dialogueTree.getLineOptionsCount = function() {
  */
 gdjs.dialogueTree.confirmSelectOption = function() {
   if (
-    this.dialogueData.select &&
-    !this.selectedOptionUpdated &&
-    this.selectedOption !== -1
+    !this.dialogueIsRunning ||
```
4ian: Same are you sure about your condition?

wapiti4:
cat from forum topic here.
Here's the latest example project of this issue:
[yarnfixes-project-200208.zip](https://github.com/4ian/GDevelop/files/4173780/yarnfixes-project-200208.zip)
In this zip it first loads the JSON file then starts dialogue from branch "Start".
Then it checks for "The dialogue line is "text" " before displaying text
, which doesn't trigger.
Same result with the condition set to "Dialogue is running".

4ian:
I'll @blurymind have a look if he can - I guess there are good chances that these changes will fix issues (after verifying the conditions I highlighted).

blurymind:
@4ian @wapiti4 We fixed the loading problem in the PR already. If you try the attached example from @wapiti4 , now the text is loaded

Extensions/DialogueTree/dialoguetools.js
blurymind: yes, this prevents it from trying to read data when no dialogue is initiated

Extensions/DialogueTree/dialoguetools.js
blurymind: yes, this prevents it from trying to read data when no dialogue is initiated

blurymind: oops, that is actually an error, will amend now

Extensions/DialogueTree/dialoguetools.js
blurymind: oops, sorry, this is not right. Will amend now

blurymind:
@4ian thank you for catching that late night coding siliness. I amended it now

4ian:
Looks good! Thanks :)

PR #1427, opened 2020-02-12
Add Map Object as replacement for basis object in Hashmap

arthuro555:
Because it should be faster for dynamic key-values attribution. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#Objects_and_maps_compared to have a better understanding of the différences.

### DIFF
--- a/GDJS/Runtime/inputmanager.js
+++ b/GDJS/Runtime/inputmanager.js
@@ -88,7 +88,7 @@ gdjs.InputManager.prototype.wasKeyReleased = function(keyCode) {
  * Return true if any key is pressed
  */
 gdjs.InputManager.prototype.anyKeyPressed = function() {
-    for(var keyCode in this._pressedKeys) {
+    for(var keyCode of Hastable.keys(this._pressedKeys)) {
         if (this._pressedKeys.containsKey(keyCode)) {
             if (this._pressedKeys.get(keyCode)) {
                 return true;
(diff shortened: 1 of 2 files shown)


arthuro555:
Sorry if there are errors I made it quickly on my phone so no intellisense or error highlighting to show me potential errors and auto correction may have corrected something that shouldn't have been.

arthuro555:
Ok this is very messy and hacky I really need to review it completely on a real computer later

4ian:
Let me know if you have questions or get something working :)

arthuro555:
It's just super difficult as looping over a Map object isn't the same as over an object, as for in just doesn't works correctly and for of returns a key value pair in an array, Wich is not what all the code spread through the codebase iterating over the items property expect. I need to somehow
1. Create an iterator that works for both and replace every use in the codebase with that custom iterator (what I tried to do and failed)

or

2. Replace the iterator of the map objects with one that works with for of like an object and replace every for in with a for of.

The two options are a bit dirty. A 3rd option would be to map the map to an object and iterate over it but that would be totally stupid as it would remove the performance gain that is the point of the use of maps with a performance loss.

If you have any idea of a clean way to do it, I'm all ears!

Arthur.

arthuro555:
Ok, so after investigating this would require a major change as it is impossible to use for in loops with Map objects. But for of is unimplemented on IE.

4ian:
> If you have any idea of a clean way to do it, I'm all ears!

I don't really have a better solution, the one I can think of is to create a `.forEach` functions on `Hashtable` and refactor the whole codebase so that all `for(...)` are now calling this function (which can be implemented differently according to if it's a "plain" Object or a "ES6" Map).

arthuro555:
Ok this entiere PR is trash and it takes me way too much time. I abandon.

PR #1428, opened 2020-02-13
Add extension "lifecycle" events function

4ian:
Needs some testing :) 

- [ ] Would need to be documented in the wiki (there is a new dialog, and some explanations maybe about these functions to add).


4ian:
@Bouh If you want to take a look at this (I'll merge in master in the morning), this exposes 7 lifecycle for events-based extensions. Should be more than enough, and useful for creating some global states at the beginning of the game ("onFirstSceneLoaded") and then updating some state ("onScenePreEvents" or "onScenePostEvents").
When using JS, I recommend creating in "onFirstSceneLoaded" the extension object on `gdjs`. You are then assured it's here in other functions :) 

Careful: these lifecycle are executed **only** if you actually use an action/condition/expression of the extension in your game (or a behavior).

I think the implementation works well but let me know if you find any issue. 

Also note that if you update the Gamepads extension (or another extension) to use these, as this won't work for older versions I'll wait for releasing a version and then a few days, to ensure that the large majority of people have upgraded in a version that supports these :) 
In the future, I've added a field that will show an error message in GDevelop and prevent installation if the GDevelop version is not compatible with a range indicated in the extension (and asking to upgrade GDevelop).

@arthuro555 this might be of some interest to you too :) Remember to try not use these special lifecycle functions as little as possible. They are there only for advanced extension like Gamepads, that need to store and update some global states.
For other extension, prefer passing everything as parameter (will be easier if we can compose behaviors in the future).

Bouh:
I rework and clean the extension Gamepad yes :)
For now i just used **OnFirstSceneLoaded** it works fine, and UI for add new cycle function are simple, nothing to said.
Just well done!

Bouh:
Just this function are call just once right ?
It is pretinent to precise it in the description?

> Events that will be run **once** when the first scene of the game is loaded, before any other events.

And precise too on behavios cycle function.

4ian:
> Just this function are call just once right ?

Correct. `onFirstSceneLoaded` is only called **once** when the game is started (precisely, just after the first scene is loaded, and just before the first frame and the first events are run). If you have multiple extension having a "onFirstSceneLoaded", each of them will be called (the order is undefined, can be any order). But in any case, each onFirstSceneLoaded will only be called once and never again. It's a good place to put the creation of your variables, objects, etc... in JS :) 

> Events that will be run once when the first scene of the game is loaded, before any other events.

Yes having "once" here is a good idea!

4ian:
Merging this now! If you have comments for behavior method descriptions, let me know :)

Bouh:
- Same thing can be done on other sentences on functions and behaviors.
I'll do that.

- And because the future extensions will not be compatible with previous version of GD, if you add a warning about this, what you think to add a warning message  too when an extension should be updated ?
I mean in gamepad extension i add a parameter on a function, and i should reorder the parameters on this function for keep consistency with other function. (player_id in first and other parameters after)

So this change will broke the action in eventsheet, how warning the user about this problem. Maybe add a "Guide migration version" section on the extension for see what the user should change?
Sometime we need make change and have a guide for migrate.
Let me know if its not clear.

PR #1429, opened 2020-02-14
Run preview with predefined scene

Bouh:
**### This PR need test from someone for chase potential bug.**

Related to [same card on Trello](https://trello.com/c/23TfvHXW/402-run-preview-with-predefined-scene)

How it's works:
Right click on preview button, if you enable the checkbox the current scene open will be used for the override and icon of preview button will be updated with a little red flag.

If you uncheck the previous scene set for override is kept, but the override system is disabled.

Checklist part1:
- [X] Add support preview overriden on Toolbar of EventsSheet.
- [x] Take into account the debugger button.
- [x] Icons on preview button is broken i wish keep one component and just adapt the value to use depending the state but i don't know how return the content.

Checklist ideas:
- [ ] See for add preview button in function and resource if isPreviewOverride is active, or preview button as global in the newIDE. (This require a rework of toolbars)
- [ ] Add set for preview in scene list in project manager.
- [ ] Adapt "enabled" parameter on tab, because on resource tab i can use the button. Must be enabled only on layout and externalLayout tabs. ( Present but remove in [this commit](https://github.com/4ian/GDevelop/pull/1429/commits/d1dc899fc392fa0bf9a6076faf771d5574eb2490) )


4ian:
I have yet to look more into this, but I wonder if we could show this as a a checkbox in the context menu? And I think it should be shown in the scene properties and maybe in the project manager?
Well maybe not everywhere :D But maybe at least a checkbox when right clicking on the Play button, that is checked if the scene is chosen as being the first one?

Bouh:
Like grid context menu with toggle grid ?

4ian:
> Like grid context menu with show grid and show Mask ?

Yep

> This like of check box for enable the preview override ?

Yes, or something like that in the menu:

- [x] Always use this scene to start the preview
- ...

> I wish keep the new icon too, because we can see directly if an override is enabled or not :)

Having an override must indeed be very easy to see! :)

Bouh:
You can now right click on preview button and the checkbox preview override is uncheck by default because there is not override on scene to launch.

How it's works: 
- If you click on it the current scene open will be used for the override and checkbox and icons of preview will be updated.

- If you uncheck the previous scene set for override is keep, but the override system is disabled.

- For set an scene override for preview you can use toggle button like said previously or right click on tabs name.

I need your feedback if it's good for you, i will continue to fix issue on my todo list.

newIDE/app/src/UI/Menu/ElementWithMenu.js
```diff
@@ -5,45 +5,78 @@ import ContextMenu from './ContextMenu';
...
 
 type State = {||};
 
 /**
  * Wrap an element and display a menu when `onClick` prop is called on the element.
  */
+
 export default class ElementWithMenu extends React.Component<Props, State> {
   _contextMenu: ?ContextMenu;
+  _contextMenuR: ?ContextMenu;
```
4ian: I'm thinking of a slightly different implementation. Keep a single context menu (`_contextMenu`) *but* add a prop to `ElementWithMenu` that, if set to true, will only show the menu when you do a right click.

This will avoid duplicating code for open/openRight and also this avoid the temptation of doing a button with a context menu AND a right click context menu.
See my other comment to do this :)

newIDE/app/src/UI/Menu/ElementWithMenu.js
```diff
@@ -5,45 +5,78 @@ import ContextMenu from './ContextMenu';
...
 
   render() {
-    const { element, buildMenuTemplate } = this.props;
+    const { element, buildMenuTemplate, buildMenuTemplateRight } = this.props;
+
     return (
       <React.Fragment>
         {React.cloneElement(element, {
           onClick: this.open,
+          onContextMenu: this.openRight,
```
4ian: Here instead of `onClick: this.open`, you can do:

`const { openMenuWithSecondaryClick, ... } = this.props`

```
React.cloneElement(element, {
  ...(openMenuWithSecondaryClick ? {onContextMenu: this.open} : {onClick: this.open}),
  ref: wrappedElement => (this._wrappedElement = wrappedElement),
}
```

(the weird `...()` is to make sure that we don't override the existing `onClick` in case we have an element that is defining onClick already)

newIDE/app/src/MainFrame/index.js
```diff
@@ -1534,6 +1566,21 @@ class MainFrame extends React.Component<Props, State> {
     });
   };
 
+  _onSetPreview = (editorTab: number) => {
+    const { currentProject } = this.state;
+    //Get name of scene on tab clicked
+    let name = editorTab.editorRef.props.layoutName;
+    this.setState(
+      {
+        scenePreviewOverride: currentProject.getLayout(name),
```
4ian: Maybe you can continue without fixing this for now, but I would prefer to store the scene to be used for preview as a string. You can store this string in the state as `previewFirstSceneName`.

In general, it's risky to store a GDevelop.js object like this: 
* If I use this scene as preview override, and then delete it in the project manager, there is still a reference to it here => crash when you try to launch the preview.

Instead prefer to store the name of the layout, and find it at the moment of the preview (and if you don't find it, just ignore it - must have been deleted or it's a wrong name for some reason).

newIDE/app/src/MainFrame/index.js
```diff
@@ -1534,6 +1566,21 @@ class MainFrame extends React.Component<Props, State> {
     });
   };
 
+  _onSetPreview = (editorTab: number) => {
+    const { currentProject } = this.state;
+    //Get name of scene on tab clicked
+    let name = editorTab.editorRef.props.layoutName;
```
4ian: If you're looking at the props of your children, it's almost always because you're not asking your children to do enough (your React children! 😄).
In this case, you find yourself searching for the `layoutName`. Instead, make your children to give you the name of the scene when they call `togglePreviewOverride`.

newIDE/app/src/MainFrame/index.js
```diff
@@ -1793,6 +1840,7 @@ class MainFrame extends React.Component<Props, State> {
                 label={editorTab.name}
                 key={editorTab.key}
                 active={isCurrentTab}
+                onSetPreview={() => this._onSetPreview(editorTab)}
```
4ian: onSetPreview is something that is not really related to tabs. It's just a way to say "hey, please start the game with this scene". So give the responsibility to the editor to pass the scene name to `onSetPreview`.

4ian:
Nice progress :) 
I think for now we should avoid using the context menu of the tabs to set the scene to start. A context menu for the tabs should be about showing things related to tabs organization - and it's a bit "hacky" for now to look at the scene. 

Instead, I would rather put the menu item to set the scene as the "overriding scene" in the context menu of... the play button :)

So that as a user you're like "oh what if I right click on this button? Nice there is a menu to 'override the preview scene'... and I can actually check this scene!"

All in a single place :) I would name this menu item: 

- [ ] Always use this scene to start preview

It's very clear for the user. It's a checkbox so it's very clear how to disable this after it's set. Finally, you could ask "sure but what if exceptionnaly I want to start the preview from another scene?"

Then let's add another menu item:

- [ ] Always use this scene to start preview
- (separator)
- Start a preview from this scene

so that if you click with left button, you start with the "overriding scene" (if there is one checked). With a right click you can either change the "overriding scene" or force to start a preview from the current scene.

Bouh:
And if we display which scene is used for the preview it can improve the context menu.

We set the override preview with Scene A.
We go on Scene B, and i don't remember which scene is used for preview ([because](https://www.youtube.com/watch?v=UuJRFCHEZKY&feature=youtu.be&t=40)). 
I wish only disable override system for test scene B i can't or i use your last button.

(I understand usefulness and practicality, but... but your last button is like an temporary override inside another override permanent system... i don't like it.)
And in Scene B we don't know if the override is active and if its active with which Scene A or B.

So for improve and display everywhere which scene is used, and if the system override is enable i suggest:

- [] Use scene X for preview
- Set this scene as scene to preview

First show if the override system is active and on which scene X is the scene name. 
With it we can enable/disable the override for the scene name display.
Second buttons for set the active scene for preview.

I think it's simplier, most complete, and less headsnatch, and in every active tab we know if the system is active or not and with which scene.

4ian:
Alright, let's give a try to your solution then! :) 
I would still avoid for now doing changes on tabs context menu because I think as a first step we should modify the Play button only.

Bouh:
Fix are done on context menu and preview button.

I not very well understood for get layout name and affect it to preview.
You know for change this fancy line
`editorTab.editorRef.props.layoutName;`

And i see `layoutName={name}` in mainframe, I'am wondering if it's possible to get the name with this, but no, it seem useless in <SceneEditor> ?

newIDE/app/src/MainFrame/index.js
4ian: In other words, can you change this function so that instead of `editorTab`, you're having `layoutName`. Then think backward: what do I need to do when calling this function to pass the layoutName?
Which component has knowledge of the layout name and can pass it?

Bouh: SceneEditor have the LayoutName, but I'm confuse how pass the value from children(SceneEditor) to parent(Mainframe).
I've already do that in past?🧐

Bouh:
Almost same thing but with method called from children SceneEditor, i think isn't good yet.
Need more advice.
I've try to fix types too.

Bouh:
There are both SceneEditor!!

Bouh:
I've added this in last commit
image

But i think i can avoid to pass in SceneEditor. In MainFrame When openLayout() id call i can use name from arguments for save it in a state! So simplier no ?

4ian:
> There are both SceneEditor!
> SceneEditorContainer extend BaseEditor i've miss that!

Yeah, the file containing SceneEditorContainer should have been named "SceneEditorWrapper" or "SceneEditorContainer"!

Your schema is correct! :)

>  In MainFrame openLayout() is called, i can use name from arguments for save it in a state! So simplier no ?

Well "saving it in a state" no need for that, but you can indeed surely pass it as an argument to setPreview.

If you reason about this, this means that the child (SceneEditorContainer) is given a function - without any parameter (I think). Its responsibility is clear "please call this function whenever you want to set yourself as the default scene for preview". Very simple for the child component.

The parent (MainFrame), is a bit smarter :) Because it knows the name of the layout (after all, it's this component that opened the tab, so it must know the name of the scene!), it can pass it as an argument, when the child is calling the function.

In other words, you can surely do:
```
<SceneEditor //THis should have been named <SceneEditorContainer - sorry!
  ...
  onSetPreview={() => this._setLayoutForPreview(name)}
  ...
/>
```

(you can see I rename the function to `_setLayoutForPreview`. Let me know how it goes!

newIDE/app/src/MainFrame/index.js
```diff
@@ -926,15 +963,25 @@ class MainFrame extends React.Component<Props, State> {
...
         <PreferencesContext.Consumer>
           {({ values }) => (
             <SceneEditor
               project={this.state.currentProject}
               layoutName={name}
               setToolbar={this.setEditorToolbar}
+              isPreviewOverride={this.state.isPreviewOverride}
+              togglePreviewOverride={this.togglePreviewOverride}
+              previewFirstSceneName={this.state.previewFirstSceneName}
+              setScenePreview={this.setScenePreview}
```
4ian: What if you do `setScenePreview={() => this.setScenePreview(name)}`?

So that the SceneEditor/SceneEditorContainer just have a function to call.
And this.setScenePreview is called with the scene name, which is very convenient because it's exactly what it needs? :)

Bouh:
Better now?
I see now what you said some days ago about tabs.
I'have an issue for get name and pass it to <CloseableTab> in <MainFrame>.

Because *name* didn't exist and *editorTab.name* return "New Scene (Events)", but the classic SceneEditor tabs works.

newIDE/app/src/MainFrame/index.js
Bouh: This seem trouble the behavior on the togglePreviewOverride

newIDE/app/src/SceneEditor/index.js
```diff
@@ -160,6 +164,7 @@ type State = {|
 
   // State for tags of objects:
   selectedObjectTags: SelectedTags,
+  layoutName: () => string,
```
4ian: Here you're saying "layoutName is a function that when called is returning a string". 
This seems suspicious, and I don't see any "layoutName" being used in this component `this.state` => you probably don't need this

newIDE/app/src/SceneEditor/index.js
```diff
@@ -223,6 +228,10 @@ export default class SceneEditor extends React.Component<Props, State> {
       <Toolbar
         showPreviewButton={this.props.showPreviewButton}
         onPreview={() => this.props.onPreview({})}
+        isPreviewOverride={this.props.isPreviewOverride}
+        togglePreviewOverride={this.props.togglePreviewOverride}
+        previewFirstSceneName={this.props.previewFirstSceneName}
+        setScenePreview={() => this.props.setScenePreview({})}
```
4ian: I don't think you need to pass {} as argument? There is no argument at all needed?

newIDE/app/src/MainFrame/index.js
```diff
@@ -518,6 +522,29 @@ class MainFrame extends React.Component<Props, State> {
...
+      },
+      () => this.updateToolbar()
+    );
+  };
+
+  togglePreviewOverride = () => {
+    if (!this.state.previewFirstSceneName) {
+      this.setScenePreview();
+      return;
+    }
```
4ian: I'm not sure why you would be messing with "previewFirstSceneName" in this function.

This function is "togglePreviewOverride". When called, it must activated or deactivate the boolean that tell if the override is activated. How to do that? Just by doing:

```
    this.setState({ isPreviewOverride: !this.state.isPreviewOverride }, () => {
      this.updateToolbar();
    });
```

Apart from the toolbar that need a manual refresh that's... all you need?

newIDE/app/src/MainFrame/index.js
```diff
@@ -863,13 +890,23 @@ class MainFrame extends React.Component<Props, State> {
...
+
     if (!_previewLauncher) return;
 
     this.setState(
       {
         previewLoading: true,
       },
       () => {
+        if (previewFirstSceneName && isPreviewOverride) {
+          layout = currentProject.getLayout(previewFirstSceneName);
```
4ian: In general avoid to change parameters. Declare a new variable and use it. The layout passed as parameter is not supposed to be changed to another layout - this could surprise other usage of the layout by the rest of the code and create weird bugs.

newIDE/app/src/MainFrame/index.js
4ian: Then it's togglePreviewOverride that must be fixed :) See my other comment.

newIDE/app/src/MainFrame/index.js
```diff
@@ -518,6 +522,29 @@ class MainFrame extends React.Component<Props, State> {
...
+    const editorTab = getCurrentTab(this.state.editorTabs);
+    this._setLayoutForPreview(editorTab.name);
+
+    this.setState(
+      {
+        isPreviewOverride: true,
+      },
+      () => this.updateToolbar()
+    );
+  };
```
4ian: I strongly think you don't need this whole function that is messing with this.state.editorTabs.

Just one function that you wrote:

```
  _setLayoutForPreview = (name:string) => {
    this.setState(
      {
        previewFirstSceneName: name,
        isPreviewOverride: true,
      },
      () => {
        this.updateToolbar();
      }
    );
  };
```

You need to call this function to set the name of the scene to launch. Do that when the user click on a toolbar, in other words, by doing:

```
<SceneEditor
  ..
  setScenePreview={() => {
    console.log("oh oh, seems the user want to use this scene as the preview scene! Let's remember it");
    this._setLayoutForPreview(name);
  }}
  ..
/>
```

newIDE/app/src/MainFrame/index.js
Bouh: With 
```
if (!this.state.previewFirstSceneName) {
      this.setScenePreview();
      return;
    }
```
At start when we open the project and go in SceneA, the checkbox is uncheck and label show "Use this scene for preview".
Then when we check it for the first time the active editor wil be used for preview.
The checkbox is now checked and label show "Use scene SceneA for preview"

If i delete this lines, when i check it and go in another tab like SceneB, the label show "Use this scene for preview", but the override is active on the SceneA, so labels will be wrong if a remove this lines.

This is the reason of this part of code :)

newIDE/app/src/MainFrame/index.js
```diff
@@ -895,6 +883,7 @@ class MainFrame extends React.Component<Props, State> {
       previewFirstSceneName,
       isPreviewOverride,
     } = this.state;
+    const layoutOverride = currentProject.getLayout(previewFirstSceneName);
```
Bouh: Seem to me more risky, the value can be not define if previewFirstSceneName isn't yet set.

newIDE/app/src/MainFrame/index.js
```diff
@@ -1583,8 +1574,11 @@ class MainFrame extends React.Component<Props, State> {
     });
   };
 
-  _setLayoutForPreview = (name:string) => {
-    
+  _setLayoutForPreview = (name?: string) => {
+    if (!name) {
+      const editorTab = getCurrentTab(this.state.editorTabs);
+      name = editorTab.name;
+    }
```
Bouh: Explaination are here
https://github.com/4ian/GDevelop/pull/1429#discussion_r382560006

Bouh: This is now deleted. See last commit.
Same behavior as in explaination but works differently.

newIDE/app/src/MainFrame/index.js
Bouh: This is now deleted, see last commit.

newIDE/app/src/MainFrame/index.js
```diff
@@ -1791,8 +1844,10 @@ class MainFrame extends React.Component<Props, State> {
             return (
               <ClosableTab
                 label={editorTab.name}
+                type={editorTab.type}
                 key={editorTab.key}
                 active={isCurrentTab}
+                onSetPreview={() => this._setLayoutForPreview(editorTab.name)}
```
Bouh: Always need adjust here for get right name without (Events) on EventsSheet.

Bouh: name in EditorTabsHandler.js should be labelName, and layoutName for get real name of the layout without (Event) or (Extension) in the name :)

4ian: I've not yet reviewed again but you're totally right that "name" is too broad, it should be called "label" (which is explicit that it's something shown on screen with a potential suffix).

newIDE/app/src/UI/Menu/ElementWithMenu.js
```diff
@@ -5,45 +5,61 @@ import ContextMenu from './ContextMenu';
...
+
 export default class ElementWithMenu extends React.Component<Props, State> {
   _contextMenu: ?ContextMenu;
   _wrappedElement: ?any;
 
-  open = (event: any) => {
-    const { _contextMenu } = this;
-    if (!_contextMenu) return;
+  open = (element: any) => {
+    if (!this.props.buildMenuTemplate) {
```
4ian: Do we need the changes in this function? Presumably this component is always used with a `buildMenuTemplate` prop? Not having it is an error no?.

newIDE/app/src/MainFrame/index.js
```diff
@@ -1790,9 +1843,11 @@ class MainFrame extends React.Component<Props, State> {
               getCurrentTabIndex(this.state.editorTabs) === id;
             return (
               <ClosableTab
-                label={editorTab.name}
+                label={editorTab.label}
+                type={editorTab.type}
                 key={editorTab.key}
                 active={isCurrentTab}
+                onSetPreview={() => this._setLayoutForPreview(editorTab.label)}
```
4ian: I know you've done it but for now I'm not a big fan of the right click menu on the tabs in this current state. We could do it, but in this case we have to pass more information about the layout/external layout/external events/extension (because the label here is very fragile, any extra character will break it).

At least it's I think more complicated than just passing a type.

newIDE/app/src/MainFrame/index.js
4ian: Yes but if you're calling getLayout on a layout that does not exist, you're potentially crashing libGD.js.

Let's avoid this and use `hasLayout` when launching the preview.

newIDE/app/src/MainFrame/index.js
```diff
@@ -863,13 +873,24 @@ class MainFrame extends React.Component<Props, State> {
...
     if (!_previewLauncher) return;
 
     this.setState(
       {
         previewLoading: true,
       },
       () => {
+        if (previewFirstSceneName && isPreviewOverride) {
+          layout = layoutOverride;
+        }
```
4ian: Prefer to do:

```js
let previewedLayout = layout;
if (previewFirstSceneName && usPreviewOverride) {
  if (project.hasLayout(previewFirstSceneName)) {
    previewedLayout = project.getLayout(previewFirstSceneName);
  }
}

// then call the _previewLauncher
```

That should be it :)

newIDE/app/src/EventsSheet/Toolbar.js
```diff
@@ -35,10 +39,45 @@ export class Toolbar extends PureComponent<Props> {
...
+                click: () => {
+                  if (!this.props.previewFirstSceneName) {
+                    this.props.setScenePreview();
+                  }
+                  this.props.togglePreviewOverride();
+                },
+              },
+              { type: 'separator' },
+              {
+                label: 'Set this scene as scene to preview',
```
4ian: What do you think of "Always use this scene to start the previews"?

Bouh: I trust you for english sentences ^^ It's ok ^^

Bouh:
> I know you've done it but for now I'm not a big fan of the right click menu on the tabs in this current state. We could do it, but in this case we have to pass more information about the layout/external layout/external events/extension (because the label here is very fragile, any extra character will break it).
> 
> At least it's I think more complicated than just passing a type.

Well see this one later, I've remove code.

newIDE/app/src/MainFrame/index.js
```diff
@@ -1790,7 +1841,8 @@ class MainFrame extends React.Component<Props, State> {
               getCurrentTabIndex(this.state.editorTabs) === id;
             return (
               <ClosableTab
-                label={editorTab.name}
+                label={editorTab.label}
+                type={editorTab.type}
```
4ian: This is not needed anymore, and you can also remove all the type on the editor tabs.

newIDE/app/src/SceneEditor/index.js
```diff
@@ -132,6 +132,10 @@ type Props = {|
   onChooseResource: ChooseResourceFunction,
   resourceExternalEditors: Array<ResourceExternalEditor>,
   isActive: boolean,
+  isPreviewOverride: boolean,
+  togglePreviewOverride: () => void,
+  setScenePreview: (options: PreviewOptions) => void,
```
4ian: typing seems false, setScenePreview is not getting passed any arguments

newIDE/app/src/UI/ClosableTabs.js
```diff
@@ -104,6 +104,7 @@ export class ClosableTabs extends Component<ClosableTabsProps> {
 type ClosableTabProps = {|
   active: boolean,
   label: Node,
+  type?: string,
```
4ian: Let's remove the type

newIDE/app/src/UI/Menu/ElementWithMenu.js
```diff
@@ -5,19 +5,22 @@ import ContextMenu from './ContextMenu';
 
 type Props = {|
   element: React$Element<any>,
-  buildMenuTemplate: () => Array<any>,
+  onClick?: () => void,
+  buildMenuTemplate?: () => Array<any>,
```
4ian: Why making buildMenuTemplate optional?

newIDE/app/src/UI/Menu/ElementWithMenu.js
```diff
@@ -26,24 +29,33 @@ export default class ElementWithMenu extends React.Component<Props, State> {
...
+          ...(openMenuWithSecondaryClick
+            ? { onContextMenu: this.open }
+            : { onClick: this.open }),
           ref: wrappedElement => (this._wrappedElement = wrappedElement),
         })}
-        <ContextMenu
-          ref={contextMenu => (this._contextMenu = contextMenu)}
-          buildMenuTemplate={buildMenuTemplate}
-        />
+        {buildMenuTemplate && (
```
4ian: buildMenuTemplate must always be defined? No need for this condition?

newIDE/app/src/SceneEditor/Toolbar.js
```diff
@@ -42,10 +45,45 @@ export class Toolbar extends PureComponent<Props> {
     return (
       <ToolbarGroup lastChild>
         {this.props.showPreviewButton && (
-          <ToolbarIcon
-            onClick={this.props.onPreview}
-            src="res/ribbon_default/preview32.png"
-            tooltip={t`Launch a preview of the scene`}
+          <ElementWithMenu
```
4ian: We have here twice the same elements (this one, and the next one: showNetworkPreviewButton).

I think it's a good time to make a single component for these. (Taking as props all the function and booleans required: previewFirstSceneName, setScenePreview, etc...).

This will be super useful to:
- Avoid this code duplication, so easier to maintain and we're 100% sure the buttons are always consistent (imagining changing one and not the other... not very professional looking).
- Later refactor the toolbar if we want to show these buttons in the middle of the toolbar.

Can you have a look at making a new component file for these? Using <React.Fragment> to return two buttons without a parent div :)

4ian:
Alright, this is getting in better shape :) I'm still worried about code duplication related to buttons. I think we need to create now a component to host these buttons as they are becoming complex.

Advantage: it will then be "super easy" to add them to other toolbars (the one for functions)!

I'd like to make this addition as simple as possible for MainFrame file, which is becoming bloated.

Bouh:
Pretty easy todo.
This new component is called PreviewButtons and is actually only for SceneEditor. 

What you think if i move it in:
`newIDE\app\src\MainFrame\Toolbar\PreviewButtons.js`
And actual file:
`newIDE\app\src\MainFrame\Toolbar.js`
become:
`newIDE\app\src\MainFrame\Tollbar\index.js`

And this last one include the new component.

4ian:
> What you think if i move it in:

Yes! That sounds perfect :) 

Thanks for making the component. Will make it easier to change it later if we want to display it in the middle or something like that (like having it *always displayed* in all editors - though I'm still not sure exactly how to do it code-wise).

> is actually only for SceneEditor

Do you think you can use it for the EventsEditor? Is there any difficulty or thing that you're concerned about?

Bouh:
I wish some advice, i've moved Toolbar and PreviewButtons, it's ok.
But i'am a bit lost about how get states from the Editors from the Toolbar living in MainFrame.

The logic for get values between components isn't present for me in some cases.
I understand how pass state or props values to children (Parent>Child), but get values from an children and pass trought the parent and go in another loving in parent child for get the value,this give me headache.

I need put the value in PreviewButtons and pass trought > Toolbar > MainFrame > SceneEditor or other editors depending type, because i just need SceneEditor, EventSheet, and ExternalLayout. For get the **name** on the layout, and others thing like **values** and **storageProviderOperations**.

4ian:
So let's take a step back and reflect on what we have. In React, everything is a question of *responsibility*.

**Before, we had this situation**:

`MainFrame`, with a `Toolbar` that is rendered by MainFrame. It has a `setEditorToolbar` function that is passed to the editors (`SceneEditor`, etc...). `setEditorToolbar` is called by the editors (using a prop called `setToolbar`) to update the buttons to show. The clicks on the buttons **are handled by the editor** (because when you call `setToolbar`, you pass the components to render, and the onClick/everything is set by the editor).

In other words: who has the **responsibility of making** the toolbar buttons? The `SceneEditor` (or other editors). This includes everything including what the buttons are doing.

**Now you want to add something that is a menu to select which scene to launch for the preview**:

And we want this menu to be on the Play buttons. First things to understand:
* Who has the responsibility for launching the preview? It's `MainFrame`. This is understandable! Editors are not "smart" enough to launch a preview, it's something that requires too much knowledge. Only MainFrame has a global view of the project. So what `MainFrame` is doing, is that it's passing to the editor a callback. It's done using the prop `onPreview`.
When it's called, MainFrame uses the arguments and its own function to launch the preview. 

Advantage: the editors don't know anything about previews! They just say to their parent (using the prop `onPreview`): "hey please run the preview, here is the layout, now it's your job I don't care anymore".

So you added a state in `MainFrame` to remember/change/clear the scene to use for the preview, you passed a few callbacks to the editors. Finally you modified the buttons rendered by the editors so that they call these functions. 

That's it and it works :) Now I say "right but these buttons are becoming complex, let's make a component for them". "Easy", you created a component, passed the props, it still works.

**We could stop there :)**

I would just ask you to use the new `PreviewButtons` component also inside `EventsEditor`, because I want to be sure that all editors that are displaying them are doing exactly the same thing. Also I want to be sure that if one is changed, the other is automatically changed... that's exactly why we use a component :D 

**Now we could also (but it's optional and maybe we should not do this now) say "Ok you know what? Let's move PreviewButtons inside the `Toolbar` of `MainFrame`"**.

That's a very interesting idea! This means that all editors would have these buttons. We probably want to do that later. But now we do have two new problems:

1. What do we do for the editors that in the past did not show these buttons... 🤔Seems like something will have to change, what preview should we run?
2. Because we're moving the buttons out from the toolbar that is handled by the child editors (`SceneEditor`, etc), this means that the **responsibility for reacting to the clicks changed**.

Child editors are now "what PreviewButtons? I'm not even using PreviewButtons. Sorry don't know about that". So `MainFrame` must now take the responsibility of:
- identify the active tab (easy, it's somewhere in its state :))
- have a function called when the user click on a `PreviewButtons` (easy, well that's usual React stuff: let's just pass a prop)
- and finally, in the callbacks passed to `PreviewButtons`, do the right thing: either update the override scene ("easy", just pass the functions you created) or harder: *launch the preview*.

Here we have an issue: how do I launch the preview?? What should I launch??? Before it was easy: it was the *responsibility* of the child to tell you "hey please launch preview for this layout". `MainFrame` was just checking if you had an override scene and if yes, it was launching this scene instead. (Hey, I never said that parents can not do things that are different than what the children said ;)).

So what should we do? We should use some information that we have on the editor, to identify:
- which layout is the editor tied to (either we find a layout name somewhere, or we had a new information in the object that represents a tab - I've not studied this).
- and with this we can launch the preview.

It may happen that the current editor does not have a layout tied to it... for example, a tab showing an extension is not tied to a layout. It does not even know what a layout is... So in this case we should probably launch the preview from the first scene or something like that. In fact that's precisely the "problem 1" that I wrote previously!!
The "components" approach of React forced us to consider this problem :) (or, if we code this wrongly, we'll get a crash in the editor... but Flow should verify this for us).

Hope this gives you a better clarity.
In general if you want to "pass some information from a child to another child" in React, you need to *involve the parent*. You need to *shift* the responsibility from the child to the parent. It's also called *lifting the state* (because in the "tree" of the components, you're moving the state up) (though here we're not moving state, but more callbacks/props).

**I would recommend not to do this for now**. For now, things are complex enough with this new feature that was added to override scene. Let's just use these `PreviewButtons` and call it a day! :)
Then in a new PR we can improve this by saying "You know what? Let's put the `PreviewButtons` in the Toolbar - and refactor the children so that they don't even have props to launch the preview".
It will be less props, and more accessible buttons... but later? :)

Bouh:
- >What do we do for the editors that in the past did not show these buttons... 🤔Seems like something will have to change, what preview should we run?

I imagine when we are in scenes/eventsheet/externalLayout we have the same buttons as now.
And in other type of tabs actually the buttons are not display at all, we can display them but disabled.
Why disabled? Because there is no override active.
So we can't run the preview because no scene have been setup before.
And when an override exist we can enable the buttons in all type of tabs.


MainFrame receive the order from PreviewButtons for launch the preview, but MainFrame ask to the editor the name of the layout used.
And if no layout have been found then we are no in one of this type of tabs scenes/eventsheet/externalLayout.
And we run nothing because the buttons are display but disabled.


- Yeah i suppose thats values project,layout,options come from SceneEditor, but understand how it's work, and where living project/layout/options in SceneEditor. 

MainFrame
```
<SceneEditor
onPreview={(project, layout, options) => {
//some code
}}
/>
```

- > but later? :)

Yes i will use this time for read some articles and read about callback in React.

- > ... but Flow should verify this for us).

Talking about flow i've seen warnings about Stories somewhere with npm run flow, i should fix that before merging.

Bouh:
<EventsSheet/> is called from EventsFunctionsExtensionEditor, but like the buttons are ignored with `showPreviewButton={false}` and `showNetworkPreviewButton={false}`.
When i set all props `isPreviewOverride`, `previewFristSceneName`, `setScenePreview`, `togglePreviewOverride ` as optional. Other warning come for said these props are undefined.
image

4ian:
> And if no layout have been found then we are no in one of this types of tabs scenes/eventsheet/externalLayout.
> And we run nothing because the buttons are display but disabled.

Yep sounds correct! 

> EventsFunctionsExtensionEditor, but like the buttons are ignored with showPreviewButton={false} and showNetworkPreviewButton={false}.

Buttons are indeed not shown when these are false. But Flow is still expecting you to pass the props.

You can make these props optional: `isPreviewOverride?: boolean`
but instead I would recommend to always pass them - even in `EventsFunctionsExtensionEditor`. Just pass false/empty string or functions doing nothing :)

Same for EventsSheet in stories.

Bouh:
I've fix all warnings,
Except one with babel loader but i've not touch this file and it's not included in this PR.
image

4ian:
It's a Flow bug, you can safely ignore this :)

4ian:
Alright this looks good overall. I'm still thinking if we can somehow avoid passing these 4 props - it's a "lot of props". We could use a context but I think that's more work for nothing and less explicit... not sure if we can do better.

Bouh:
By context, you mean an object? Like options on onPreview() ?
`this.props.onPreview({ networkPreview: true })}`

4ian:
By context I meant a React Context, but an alternative solution that would allow us to pass less props and still use props (rather than a React Context, which is "hiding props"), is indeed to pass an object:

```
  <SceneEditor
    previewButtonSettings={{
      isPreviewOverride={this.state....}
      togglePreviewOverride={() => ...}
      previewFirstSceneName={this.state....}
      setScenePreview={() => ...}
    }}
  />
```

advantage: you can type this object in PreviewButtons:

```
export type PreviewButtonSettings = {|
  isPreviewOverride: boolean,
  ...
|}
```

and it's easier to pass them in props! In some components that don't need them, just pass them to children:
```
<Something
  previewButtonSettings={this.props.previewButtonSettings}
  ...
/>
```

Finally what I like is that if one day we refactor or remove/rename/add a new function/state, we only have `PreviewButtons` and `MainFrame` to change... instead of having to change all the editors to pass the removed/renamed/added function/state.

Would you give it a try?

Bouh:
Yes give me more time, i can try to do this :)

Bouh:
I've done the change. 
I've loose one thing, now at the first click for "Use this scene as scene for preview" i need click twice on the button, and only once on web app. 

`previewFirstSceneName` take the name of the layout, but `isPreviewOverride` stay false on desktop.

newIDE/app/src/EventsSheet/Toolbar.js
```diff
@@ -6,14 +6,15 @@ import ToolbarSeparator from '../UI/ToolbarSeparator';
 import ToolbarIcon from '../UI/ToolbarIcon';
 import ElementWithMenu from '../UI/Menu/ElementWithMenu';
 import { type EventMetadata } from './EnumerateEventsMetadata';
+import PreviewButtons from '../MainFrame/Toolbar/PreviewButtons.js';
 
 type Props = {|
   showPreviewButton: boolean,
   onPreview: () => void,
+  previewButtonSettings: Object,
```
4ian: You should be able to use `PreviewButtonSettings` here, and in EventsSheet, no?

newIDE/app/src/EventsFunctionsExtensionEditor/index.js
```diff
@@ -56,6 +56,7 @@ type Props = {|
   onBehaviorEdited?: () => void,
   initiallyFocusedFunctionName: ?string,
   initiallyFocusedBehaviorName: ?string,
+  previewButtonSettings: Object,
```
4ian: See if you can use `PreviewButtonSettings` :)

4ian: `import { type PreviewButtonSettings, ... } from '../...'`

newIDE/app/src/MainFrame/Toolbar/PreviewButtons.js
```diff
@@ -0,0 +1,109 @@
+// @flow
+import * as React from 'react';
+import { t } from '@lingui/macro';
+import ToolbarIcon from '../../UI/ToolbarIcon';
+import ElementWithMenu from '../../UI/Menu/ElementWithMenu';
+
+export type PreviewButtonSettings = {|
```
4ian: 👍

newIDE/app/src/MainFrame/Editors/EventsFunctionsExtensionEditor.js
```diff
@@ -85,6 +85,12 @@ export default class EventsFunctionsExtensionEditorWrapper extends BaseEditor {
           initiallyFocusedBehaviorName={this.props.initiallyFocusedBehaviorName}
           onBehaviorEdited={this._onBehaviorEdited}
           ref={editor => (this.editor = editor)}
+          previewButtonSettings={{
```
4ian: nitpicking: I just add an idea to make this a bit clearer and less cumbersome to use.

Export an object which is this "empty" settings, and then use it here:

* In PreviewButtons file: `export const emptyPreviewButtonSettings = { isPreviewOverride: false, ... }`
* `import { ..., emptyPreviewButtonSettings, ... } from '...'` (the PreviewButtons file)
* `previewButtonSettings={emptyPreviewButtonSettings}`

newIDE/app/src/stories/index.js
```diff
@@ -3709,6 +3721,12 @@ storiesOf('EventsFunctionsExtensionEditor/index', module)
           initiallyFocusedFunctionName={null}
           initiallyFocusedBehaviorName={null}
           onCreateEventsFunction={action('on create events function')}
+          previewButtonSettings={{
```
4ian: Here and above you can reuse `emptyPreviewButtonSettings` :)

Bouh:
Flow is slow and crash often, and flow warn me 5 issues ([log file](https://pastebin.com/raw/mZ3qM9Q8)) with `emptyPreviewButtonSettings`.

blurymind:
> Flow is slow and crash often, and flow warn me 5 issues ([log file](https://pastebin.com/raw/mZ3qM9Q8)) with `emptyPreviewButtonSettings`.

I had to disable the flow server on my computer, as it eventually crashes vscode entirely

4ian:
> Flow is slow and crash often

I'll try to update it to see if it's better. 

>  flow warn me 5 issues (log file) with emptyPreviewButtonSettings.

Yep, there is a bug :)
It's telling you:
> Cannot create `EventsFunctionsExtensionEditor` element because inexact class `PreviewButtons` [1] is incompatible with
exact `PreviewButtonSettings` [2] in property `previewButtonSettings`.

In other words: "You used a PreviewButtons, but you should have been using a PreviewButtonSettings. This make no sense! Look at your prop previewButtonSettings"

And indeed, if you look in the file...
```
import emptyPreviewButtonSettings from '../MainFrame/Toolbar/PreviewButtons';
```

should be 
```
import { emptyPreviewButtonSettings } from '../MainFrame/Toolbar/PreviewButtons';
```

4ian:
Alright, still a few things:

Let's rename these:
```
previewFirstSceneName => previewFirstSceneName
isPreviewOverride => isPreviewFirstSceneOverriden
togglePreviewOverride => togglePreviewFirstSceneOverride
setScenePreview => useSceneAsPreviewFirstScene

_setLayoutForPreview => _setPreviewFirstScene
```

I want consistency and clarity around the fact that this is overriding the "PreviewFirstScene".

Other questions:
* Are we satisfied with the behavior for external layout? They do have the preview buttons, but I think they should not be overriden (the preview is always starting from the selected scene used to edit the external layout).
  * Speaking of them, does they even still work? 😬 We're not passing any previewButtonSettings! I fear a potential crash/inconsistency.
* If I close a game and re-open another one, the override is still there... is this normal? Should we reset the override anytime a game is opened? Might be safer to avoid surprises? If yes, let's try to do it with only the methods that we have already... or adapt slightly the one we already have.

4ian:
(I know this is a lot of work for "just a feature to override the scene" - but MainFrame is complex and I want to avoid as much adding more complexity and also triple check all behaviors because it's easy to add bugs there. We have to think about all the use cases.)

4ian:
For the icon, looks a bit like a Preview with a reload icon 🤔Maybe there is something else to think about - even a simple dot to indicate that there is something special activated?

Bouh:
- External layout crash currently indeed props is missing.
More anoying the toolbar used in external layout come from SceneEditor, so the previewButtons too.
This mean we can right click on it. Props is set on `previewButtonSettings={emptyPreviewButtonSettings}` but the context menu can be shown.

>If I close a game and re-open another one, the override is still there... is this normal?
Absolutely not! 
Fixed in next commit

Bouh:
For icon, i suggest a flag, if i use a red circle it's similar to a record button, so a flag seem better.
image

image

4ian:
I love the flag 👍 Very clear and visible.

> Props is set on previewButtonSettings={emptyPreviewButtonSettings} but the context menu can be shown.

An alternative might be to allow null instead of emptyPreviewButtonSettings, and in this case don't show the menu.

> Fixed in next commit

Cool 👍

Bouh:
> An alternative might be to allow null instead of emptyPreviewButtonSettings, and in this case don't show the menu.

Not working, menu still present. I've found a semi-solution, make the menu empty. Take a look in the next commit. But i prefer delete the menu if this one is empty :/

> At the first click for "Use this scene as scene for preview" i need click twice on the button, and only once on web app.
> 
> `previewFirstSceneName` take the name of the layout, but `isPreviewFirstSceneOverriden` stay false on desktop only.

This is fixed :)

newIDE/app/src/MainFrame/index.js
```diff
@@ -1130,9 +1179,11 @@ class MainFrame extends React.Component<Props, State> {
     this.setState(
       {
         editorTabs: openEditorTab(this.state.editorTabs, {
-          name: name + ' ' + i18n._(t`(Extension)`),
+          label: name + ' ' + i18n._(t`(Extension)`),
           renderEditor: ({ isActive, editorRef }) => (
             <EventsFunctionsExtensionEditor
+              onPreview={() => {}}
```
4ian: You're sure about this? 🧐
Better pass null than an empty function. Because `null` is explictly saying "there is no way to run a preview, don't even try to call me". While passing an empty function is "hacky". It's like saying "See I passed you something to run the preview, so feel free to show a button... but well actually I won't do anything haha! Oops sorry."

4ian:
> Not working, menu still present. I've found a semi-solution, make the menu empty. Take a look in the next commit. But i prefer delete the menu if this one is empty :/

Yes, ideally we would remove the menu entirely (because you simply can't override things here).
Where should I look though? I've seen you passed `emptyPreviewButtonSettings`, but is this making the menu empty?

newIDE/app/src/MainFrame/Toolbar/PreviewButtons.js
```diff
@@ -67,26 +67,29 @@ export default class PreviewButtons extends React.Component<
...
-                  ? 'Use scene ' + previewFirstSceneName + ' for preview'
-                  : 'Use this scene for preview',
-                checked: isPreviewFirstSceneOverriden,
-                click: () => {
-                  togglePreviewFirstSceneOverride();
-                  if (!previewFirstSceneName) {
-                    useSceneAsPreviewFirstScene();
-                  }
+            buildMenuTemplate={() => {
+              if (isPreviewFirstSceneOverriden === null) return [];
```
Bouh: > Where should I look though? I've seen you passed `emptyPreviewButtonSettings`, but is this making the menu empty?

Yep, i don't know really what todo.
Here the array is empty, then the menu is empty too. I  can't return null, for not see the menu, because return null crash the app.
I can't wrap `buildMenuTemplate={() => { ... }}` with a condition, because it's not allowed? (or i don't know the syntax).

Bouh: Here the menu empty, but no ide for avoid this empty square.
image

newIDE/app/src/MainFrame/index.js
Bouh: Haha ok, i like when the values talk :+1:

newIDE/app/src/MainFrame/Toolbar/PreviewButtons.js
4ian: Right I missed this when reading the code! So here, more than avoiding an empty square, you want no menu at all.
The problem is that `ElementWithMenu` is not made for not having a menu. If we remove the "menu" from "ElementWithMenu", we would just have... an "Element" :) This is exactly what we're going to do.

Put the `ToolbarIcon`  in its own variable:

```
const previewIcon = <ToolbarIcon
                onClick={onPreview}
                src={
                  isPreviewFirstSceneOverriden
                    ? 'res/ribbon_default/previewOverride32.png'
                    : 'res/ribbon_default/preview32.png'
                }
                tooltip={
                  isPreviewFirstSceneOverriden
                    ? t`Preview is overridden, right click for more`
                    : t`Launch a preview of the scene, right click for more`
                }
              />;
```

then you can either render it directly, or use ElementWithMenu if you want a menu:
```
    return (
      <React.Fragment>
        {showPreviewButton && (
          shouldThereBeAMenu ?
          <ElementWithMenu element={previewIcon} ... /> :
          previewIcon
          )}
        ...
```

You could arguably extract more stuff into variables to make this readable. I supposed here you have a boolean shouldThereBeAMenu - but the name and content of this depends on what you use to decide for the menu presence :)

Bouh:
Wonderful, this is a simple idea!

newIDE/app/src/MainFrame/index.js
```diff
@@ -1130,9 +1179,11 @@ class MainFrame extends React.Component<Props, State> {
     this.setState(
       {
         editorTabs: openEditorTab(this.state.editorTabs, {
-          name: name + ' ' + i18n._(t`(Extension)`),
+          label: name + ' ' + i18n._(t`(Extension)`),
           renderEditor: ({ isActive, editorRef }) => (
             <EventsFunctionsExtensionEditor
+              onPreview={null}
+              previewButtonSettings={emptyPreviewButtonSettings}
```
4ian: Do you need previewButtonSettings, because EventsFunctionsExtensionEditor itself is passing emptyPreviewButtonSettings.
Same, do you need onPreview?

If not let's remove them - the less the props, the clearer it is.

4ian:
Alright, I'm triple checking everything. We're close :) Added a comment about two extra props that seems useless.

Bouh:
Yes need test before, in each type of editor. And look flow too.

Bouh:
Context menu on preview button stay in external layout, surely because   `onPreview={null} `has been removed

Bouh:
It's because you have change null to false in last commit. :/

4ian:
Sorry for this, I made the changes blindly. I changed this to test instead if `useSceneAsPreviewFirstScene` and `togglePreviewFirstSceneOverride` are defined.

4ian:
Alright I'm taking a last look and should merge this tonight :)

4ian:
I've hidden the first button and separator if there are no selected scene. I've also updated the tooltip to hide "right click for more" when there is no menu (external layout).

Merging now! Thanks for your patience and your work on this :) 👍👍

Bouh:
There is a issue :s 

Reproduce:
- Open platformer example 
- Go to one external event present in project manager.

```
TypeError: Cannot read property 'isPreviewFirstSceneOverriden' of undefined
PreviewButtons
C:\GDevelop\newIDE\app\src\MainFrame\Toolbar\PreviewButtons.js:56:8
  53 |     : 'res/ribbon_default/preview32.png'
  54 | }
  55 | tooltip={
> 56 |   !canOverridePreviewFirstScene
     |  ^  57 |     ? t`Launch a preview of the scene`
  58 |     : isPreviewFirstSceneOverriden
  59 |     ? t`Preview is overridden, right click for more`
```

4ian:
Fixed in  8e6d2da9f79108ca660e6d277d7bdc961e52d694 - thanks for noticing!

PR #1431, opened 2020-02-15
Add once on other description of extension functions and behaviors

Bouh:
Like i've said on https://github.com/4ian/GDevelop/pull/1428#issuecomment-586400265


Bouh:
I add too * (Extension)* in tabs name when an extension is open for see easily the tabs.
Maybe we can insert an icons or find a better name. 
Like you wish.

4ian:
Looks good, make sure to run Prettier again ;)

4ian:
👍

PR #1432, opened 2020-02-15
Improve pixel perfect on canvas

Bouh:
- Use same options on both canvas.
- Disable antialias on renderers

With antialias: false on image n°2 we see less glitch on tile sprite.
It's always present but only at zoom 100%. On other level of zoom glitch disappear.

When the antialias is true on image n°1 we see more glitch, and when we zoom in/out glitch is always present.

And finally i use same options on canvas for exported game/preview and scene editor.

image


4ian:
Thanks for the comparisons! 
Seems logical indeed that indeed disabling anti-aliasing will help for pixel perfect games. In your test, can you double check that the Y position of the object(s) are rounded number (no decimal places)? Because on the second image, there is an horizontal glitch but no more verticals.

Anyway, thanks for the investigation so far👍👍

Bouh:
I've found something, i don't understand why but +1 fix the issue.
And yet the value can have .5, but i don"t know it maybe more precise for rendering the instance on scene editor...

4ian:
What if you Math.round the height divided by 2 (or the width divided by 2).
Rounding might be ok, but we surely don't want to add +1 to the height or width of an object without having a very strong reason

Bouh:
No neither. I reverse the last commit. 
I think we can check this again once pixijs will be updated to lastest version.

GDJS/Runtime/pixi-renderers/runtimegame-pixi-renderer.js
```diff
@@ -28,13 +28,21 @@ gdjs.RuntimeGameRenderer = gdjs.RuntimeGamePixiRenderer; //Register the class to
...
+    
     //This prevents flickering on some mobile devices
     PIXI.glCore.VertexArrayObject.FORCE_NATIVE = true;
 
-    //Create the renderer and setup the rendering area
+     if (this._game.getScaleMode() === "nearest") {
+        PIXI.settings.SCALE_MODE = PIXI.SCALE_MODES.NEAREST;
+     }else{
+        PIXI.settings.SCALE_MODE = PIXI.SCALE_MODES.LINEAR;
+     }
```
4ian: Interesting! Somehow missed that when implementing scale mode. Did you observe any difference in the rendering? :)

newIDE/app/src/InstancesEditor/index.js
```diff
@@ -59,10 +59,24 @@ export default class InstancesEditorContainer extends Component {
...
+    //Create the renderer and setup the rendering area for scene editor.
+    //"preserveDrawingBuffer: true" is needed to avoid flickering and background issues on some mobile phones (see #585 #572 #566 #463)
     this.pixiRenderer = PIXI.autoDetectRenderer(
       this.props.width,
       this.props.height,
-      { antialias: true } //fixes jaggy edges
+      {
+        preserveDrawingBuffer: true,
+        antialias: false,
+      } // Disable antialiasing for instance in scene editor (This affect tiles sprite)
```
4ian: I would rather write:
"// Disable anti-aliasing to avoid rendering issue (1px width line of extra pixels) when rendering pixel perfect tiled sprites."

GDJS/Runtime/pixi-renderers/runtimegame-pixi-renderer.js
Bouh: The nearest option currently in gd affect only the parameter image-rendering: pixalate on canvas. 
And affect also all texture loaded. But the parameter is on the resources.
So when i add this on the general rendering i see no difference in preview.

Maybe something can be simplify at this point, Use this setting on general renderer or on each object.
But that is really useful ? (i think, no)

Bouh: These lines are revert in last commit.

GDJS/Runtime/pixi-renderers/runtimegame-pixi-renderer.js
```diff
@@ -5,68 +5,81 @@
...
-        this._pixiRenderer.view.style["image-rendering"] = "-webkit-optimize-contrast";
-        this._pixiRenderer.view.style["image-rendering"] = "-webkit-crisp-edges";
-        this._pixiRenderer.view.style["image-rendering"] = "pixelated";
+gdjs.RuntimeGamePixiRenderer.prototype.createStandardCanvas = function(
+  parentElement
+) {
+  //This prevents flickering on some mobile devices
+  PIXI.glCore.VertexArrayObject.FORCE_NATIVE = true;
+
+  if (this._game.getScaleMode() === 'nearest') {
```
4ian: @Bouh, I think we also remove these lines there? To keep things consistent?

Bouh: I keep   PIXI.glCore.VertexArrayObject.FORCE_NATIVE = true;   ?

Selection on review on Github are a bit bad i didn't understand which lines you means..

4ian: These:
image

image

Bouh: Ok done!

4ian:
Alright, let's go ahead with this! I wonder thought if we should only set `antialiasing: false` when the game scale mode is set to nearest? Or add this as a setting in the game settings?

So that you would only have to disable this if you're making a pixel perfect game?

Issue #1441, opened 2020-02-18
Selecting an item doesn't close the pop-up ("autocomplete" component issue)


PascalLadalle:
Somewhat similar to #859 

Actual result:
In the event list, after editing a field, the textbox remains open. Return does not work. As a result, tab navigation is lost (not disabled, the current position is lost). Esc deletes the value, which I find unusual.

Expected result:
When clicking on a field with a dropdown list (objects, variables...), the box should disappear once an item has been selected from the list or Enter or Esc has been pressed.
When clicking on a field which accepts carriage returns, the box should disappear once Esc has been pressed.

![image](https://user-images.githubusercontent.com/25978830/74774015-459feb80-5293-11ea-85fa-dbcf27b2fb2f.png)

Side question: do we really need carriage return in number fields? 🤔 

Edit: I'm using Esc to delete now, it's quite convenient 😅 still...we need a way to exit that box in number/text fields.

4ian:
> Esc deletes the value, which I find unusual.

Yes the autocomplete component is a bit unusual, I want to replace it at some point (that will also solve compatibility with iOS/Android/touch screens) by https://material-ui.com/components/autocomplete/

HarsimranVirk:
Hey, I'm trying to contribute for the first time.

> Yes the autocomplete component is a bit unusual

I can't figure out exactly which component this is in the code. Is it the SemiControlledAutoComplete component in UI code?

4ian:
Hi and welcome!

> I can't figure out exactly which component this is in the code. Is it the SemiControlledAutoComplete component in UI code?

Yes. SemiControlledAutoComplete is currently a React component based on `Downshift` and showing a material-ui `Textfield` with `Popper`. Downshift is a bit hard to customize on some things (like the Esc button, it's not customizable at all), and have some issues on mobile (text getting reset). So I would like to have it changed to be instead based on Material-UI autocomplete: https://material-ui.com/components/autocomplete/

This is the list of props of SemiControlledAutoComplete:
https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/UI/SemiControlledAutoComplete.js#L113-L134

The new component should re-implement these :)
You can run the storybook: https://github.com/4ian/GDevelop/tree/master/newIDE#development-of-ui-components there is a list of stories (visual "Test cases") for the autocomplete for you to verify that it works well.

HarsimranVirk:
[...]

PR #1453, opened 2020-02-22
Replaced downshift with autocomplete.

### LINKED ISSUE #1534: Unable to open load dialog when clicked another button multiple times
## Describe the bug
Unable to open `load audio` dialog when clicked another button (e.g. in my case `No` button) multiple times until the event action dialog reopened. 

## To Reproduce
1. Create new `Platformer` project
2. Click on `New Scene (Events)` tab
3. Scroll down to 'Change the animation according to what the player do.' section
4. Click on the audio as given in the screenshot
image
5. Click on `Choose the audio file to use`
6. Click on `Choose a new audio file`
7. The load dialog will open
8. Click `Cancel`
9. Click on `No` button twice as given in the screenshot below
image
10. Click on `Choose the audio file to use`
11. Click on `Choose a new audio file`
12. The load dialog will not open 

Check the below gif for the overall problem:
loadFileDialog

## Other details
OS Details: Windows 10 Home Single Language | Version 1909 | OS Build: 18363.719
App Details: Desktop app of GDevelop version 5.0.0-beta89 based on GDevelop.js 4.0.98-0-release

HarsimranVirk:
Resolves #1534 
This change was in response to issue #1441. I have tried to implement autocomplete to mimic the behavior of the previous downshift component, and also improve it (pop up closes on pressing enter, ESC doesn't delete the input text).

I couldn't integrate i18n due to a persistent error (storybook logs i18n not defined), tried a lot of workarounds but couldn't find a way.
(Also, changed the code style from class-based components to hooks for styling. And this is my first pull request to a major project ^_^ )


newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -135,223 +67,67 @@ type Props = {|
...
+    const classes = styles();
+
+    
+    return(
+          <div
+          style={{
+              ...styles.container,
+              flexGrow: props.fullWidth ? 1 : undefined,
+          }}>
+                  <Autocomplete
```
4ian: The indentation is inconsistent, run the autoformatting that will take care for you of formatting the code: https://github.com/4ian/GDevelop/blob/master/newIDE/README.md#tests

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -27,88 +26,21 @@ export type DataSource = Array<
...
-          ? 'item-with-value-' + item.value
-          : 'item-without-value' + index
-      }
-      selected={isHighlighted}
-      component="div"
-      style={{
-        fontWeight: selected ? 500 : 400,
-      }}
-    >
-      {item.renderIcon && <ListItemIcon>{item.renderIcon()}</ListItemIcon>}
```
4ian: Is your updated version showing icons? If not, it's a must have to keep in the new version :) 
According to what I see in the docs, maybe something like `renderOption` or something else to customize the rendering of the items.

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -135,223 +67,67 @@ type Props = {|
...
-          </Downshift>
-        )}
-      </I18n>
-    );
-  }
-}
+          <MarkdownText source={props.helperMarkdownText} />
+          ) : null;
+
+    const classes = styles();
```
4ian: it's very surprising to have a function called `styles` => find a name with a verb :)

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -135,223 +67,67 @@ type Props = {|
...
+
+    
+    return(
+          <div
+          style={{
+              ...styles.container,
```
4ian: Are you sure styles.container exists?

HarsimranVirk:
> Is your updated version showing icons? If not, it's a must have to keep in the new version

`const renderMenu = option => {
  if (option.type !== 'seperator') {
    return (
      <MenuItem>
        {option.renderIcon && (
          <ListItemIcon>{option.renderIcon()}</ListItemIcon>
        )}
        {option.value}
      </MenuItem>
    );
  } else {
    return <Divider />;
  }
};`

`renderOption={(option, state) => renderMenu(option)}`

I have added this code now and it renders the icon :)

Screenshot from 2020-02-23 01-14-46

> While this is not a major version bump, it's still a bit "risky" to upgrade the minor version of Material-UI as this is impacting literally the whole application (any regression would be a bad thing).

Yes, I understand changing versions of dependencies can have a drastic effect, but somehow Autocomplete and TextField were working together after upgrading the version. Also, as of now autocomplete may not be very stable, since material-ui/lab fixes it in every alpha release.

> Also should we use 4.9.3 as it's the very last version and can have maybe interesting bug fixes?

Maybe, but I guess I'm the least qualified person to suggest that, considering neither I know much about the GDevelop codebase, nor material-ui.

> it's very surprising to have a function called styles => find a name with a verb :)

> Are you sure styles.container exists?

Fixed the other noob mistakes ^_^
 
Some screenshots from WebApp: 

Screenshot from 2020-02-23 02-02-49

Screenshot from 2020-02-23 02-03-24

Screenshot from 2020-02-23 02-05-10

HarsimranVirk:
UPDATE: The autocomplete component is removed after selecting a value, but if it is nested inside some other component(s), then it's parent nodes will keep rendering. Like in this case, autocomplete is nested inside a snackbar, after selecting a value, the snackbar keeps rendering until I click somewhere else.

Screenshot from 2020-02-23 12-11-49

I have tried to blur input, but it doesn't work, can you suggest something?

4ian:
I think you're not that far from the solution, let me know if you have more time to continue this :)

HarsimranVirk:
> I think you're not that far from the solution, let me know if you have more time to continue this :)
Im really sorry, I have exams this week. I'd definitely start working on it after my exams. Also, I'm trying to understand the codebase this week (not just the ide/editor part but even the extension part, since I'm interesting in adding support for dynamic lighting and shadows for the GSoC).

4ian:
Sure thing, no pressure! (and good luck!) :)

HarsimranVirk:
> Why is render even used? That seems like an anti pattern: it's the responsibility of the parent component to know if this component must be rendered. This component should not hide itself or anything like this.

I didn't look into the code and was trying to close the popup by using the render. Stupid mistake :(
Anyway, in the latest commit, I dug into the code and found this, https://github.com/4ian/GDevelop/blob/6155606d20830dfb01299f815fb72c92afc380c4/newIDE/app/src/EventsSheet/InlinePopover.js#L28-L32,

So, now I have passed this function as a prop through ParameterComponent (then ObjectField, then ObjectSelecter, and finally passed it to SemiControlledAutocomplete) and it closes the popup after selecting something.

> You can see that the styling is not the same as other fields when the autocomplete is shown in a popup.

I have tried to mimic the styling as much as I could. I'd still test throughout the app to find any inconsistencies. Also, I accidentally found another bug, the dividers in options popup were selectable, I have added some code to disable their selection.

Screenshot from 2020-02-29 03-17-41

This is what happens when the divider is selected.

Screenshot from 2020-02-29 03-20-01

This is how it looks now, the divider isn't styled exactly like before though.

> the box should disappear once Esc has been pressed.

This is also solved now, pressing Esc closes the popup.
I'd test it throughout the app and let you know if it breaks something :)
One more thing, what should I do for these failing checks?

4ian:
> Also, I accidentally found another bug, the dividers in options popup were selectable, I have added some code to disable their selection.

Great!  :)

> This is how it looks now, the divider isn't styled exactly like before though

That looks fine I would say :) 

> I'd test it throughout the app and let you know if it breaks something :)

Cool, let me know when you're confident this is working well :) Take a look at all the places using the component: https://github.com/4ian/GDevelop/search?q=semicontrolledautocomplete&unscoped_q=semicontrolledautocomplete

For example, the dialog to edit a group, the debugger (there is an autocomplete to select an object to create when you select a scene), various parameters are using an autocomplete (keyboard keys, scene names, etc...), the resource selectors, like the field to select a font in the Text object editor or the fields to select an image in the tiled sprite/panel sprite object editors.

> One more thing, what should I do for these failing checks?

Take a look at the tests: https://semaphoreci.com/4ian/gd/branches/pull-request-1453/builds/8
and try to understand why they are failing. One is failing because you updated material-ui which now adds a new CSS class, so the "snapshot" test is failing:
image

Run the test locally (`npm run test`, see the README), you'll see that you can ask the test runner to update this "snapshot" for you. More info about what are snapshot tests: https://jestjs.io/docs/en/snapshot-testing

The other thing failing is the Flow static typing (this is similar to Typescript):
image

This is signaling to you that you're either misusing JavaScript/creating potential bugs or you surely forgot to update the `type Props = { ... }` that are declaring what props your component is accepting. More info about Flow: https://flow.org/ - again it's similar to TypeScript.

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -137,221 +65,192 @@ type State = {|
...
+        return this.props.value;
+      }
+    };
+
+    const filterFunction = (options, state): Array<DataSource> => {
+      var separatorCount = 0;
+      const lowercaseInputValue = currentInputValue().toLowerCase();
+      var optionList = options.filter(option => {
+        if (!option.text) return true;
+        if (option.type === 'seperator') {
```
4ian: Typo

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -137,221 +65,192 @@ type State = {|
...
 
-  focus() {
-    if (this._input.current) this._input.current.focus();
-  }
+  disableDivider = (option: Object): boolean => {
+    if (option.type === 'separator') {
+      return true;
+    } else {
+      return false;
+    }
```
4ian:
```suggestion
    return option.type === 'separator';
```

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -137,221 +65,192 @@ type State = {|
...
       this.setState({
         inputValue: newValue,
       });
     }
   }
 
-  focus() {
-    if (this._input.current) this._input.current.focus();
-  }
+  disableDivider = (option: Object): boolean => {
```
4ian: This function name is not appropriated, currently if you read it it says "this disable a divider". This is not the case at all! :) Find a better name for the function: it returns a boolean, so should probably start by "is". Then what is the function doing? It's telling you if the option must be disabled.

Conclusion: "isOptionDisabled"

HarsimranVirk: Okay, I'll change the name of this function :)

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -137,221 +65,192 @@ type State = {|
...
-   * Call this function to do so.
-   *
-   * @param newValue The new value to set in the auto complete. If the field is not being focused,
-   * nothing will be forced (the value props will be used when the field is focused).
-   */
-  forceInputValueTo(newValue: string) {
+  focus() {
+    if (this._input.current) {
+      this._input.current.focus();
+      this._input.current.onkeydown = e => {
```
4ian: Never ever add manually an event handler. This can create a memory leak or weird issues. Always use React event handlers, by passing a `onKeyDown` prop to the input

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -137,221 +65,192 @@ type State = {|
...
+  };
 
   render() {
-    const { props, state } = this;
-    const currentInputValue =
-      state.inputValue !== null ? state.inputValue : props.value;
-
-    const helperText = props.helperMarkdownText ? (
-      <MarkdownText source={props.helperMarkdownText} />
+    const currentInputValue = (): string => {
```
4ian:
```suggestion
    const getCurrentInputValue = (): string => {
```

newIDE/app/src/UI/SemiControlledAutoComplete.js
HarsimranVirk: So, should I add it to the inputProps instead of the _input ref?

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -137,221 +65,192 @@ type State = {|
...
-                    label: props.floatingLabelText,
-                    id: props.id,
-
-                    // Error handling:
-                    error: !!props.errorText,
-                    helperText: props.errorText || helperText,
-
-                    // Display:
-                    InputLabelProps: getLabelProps({ shrink: true }),
+    const handleOptionLabel = (option: DataSource | null): string => {
```
4ian: Careful about your functions naming :) This is not "handling" anything, just returning to you the label to display.
```suggestion
    const getOptionLabel = (option: DataSource | null): string => {
```

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -137,221 +65,192 @@ type State = {|
...
+      if (option.value) {
+        return option.value;
+      } else {
+        return currentInputValue();
+      }
```
4ian:
```suggestion
      return option.value || getCurrentInputValue();
```

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -137,221 +65,192 @@ type State = {|
...
+    };
+
+    const handleChange = (event: Event, option: DataSource | string): void => {
+      if (option !== null) {
```
4ian: Invert the if to avoid indentation

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -137,221 +65,192 @@ type State = {|
...
+        if (option.value) {
+          if (option.onClick) option.onClick();
+          else {
+            if (this.props.onChoose) this.props.onChoose(option.value);
+            else this.props.onChange(option.value);
+          }
+        } else {
+          /* option is just a string in this case, and it
+             enables user to type something other than provided options. */
```
4ian: nitpicking: prefer double slash comments: //

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -137,221 +65,192 @@ type State = {|
...
+        ListboxProps={{
+          style: {
+            padding: 0,
+            margin: 0,
+          },
+        }}
+        getOptionLabel={handleOptionLabel}
+        filterOptions={filterFunction}
+        renderInput={params => {
+          nullifyDefaultStyling(params);
```
4ian: While this works, you're mutating objects. Prefer to create new in this case:

const defaultStylingProps = getDefaultStylingProps(params);

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -137,221 +65,192 @@ type State = {|
...
 
-                    // Style:
-                    style: props.textFieldStyle,
-                    fullWidth: props.fullWidth,
-                    ...computeTextFieldStyleProps(props),
+    const placeholder = () => {
+      if (this.props.errorText) return this.props.errorText.props.id;
+      else if (this.props.hintText && typeof this.props.hintText === 'string')
+        return this.props.hintText;
+      else return null;
```
4ian: Never ever reading into this props of your props, at least in 99.9% of cases. By doing so, you're breaking the encapsulation of the children. You probably must use properly the i18n methods here - don't try to work around it :)

newIDE/app/src/UI/SemiControlledAutoComplete.js
4ian: I think so. All events must be attached using props. Ref is just an escape hatch to "imperatively" say something to do the component (in most cases, it's for telling the component to focus, like here).

4ian:
Well I started to explain what Flow is but you've already been updating types :) Still in case you're not familiar, take a quick look. Try to understand what the error messages are telling (run `npm run flow` locally to get them) and to fix them. Usually they are telling you about something that you've not considered or not declared and so that is either "incorrect" or potentially buggy.

HarsimranVirk:
> One is failing because you updated material-ui which now adds a new CSS class, so the "snapshot" test is failing

I locally tested this and it does fail for a new class. How should I resolve this? Should I update the snapshot?
Screenshot from 2020-03-01 16-16-12
And what should I do for these? Use typeof in the if conditions? 

> but avoid typeof as much as possible

I didn't use it at all, maybe, this is why the tests are failing?

4ian:
Update the snapshot yeah, it will appear in the code in the PR :)

You can in certain case use typeof to check if something is a string. But right now the Flow error messages are telling you that "there is no value property" or "there is no onClick property" in DataSource.
=> Is this true? If no, why is Flow complaining? Surely because the type DataSource is not having these properties? In this case should you add them? :)

HarsimranVirk:
> But right now the Flow error messages are telling you that "there is no value property" or "there is no onClick property" in DataSource.

I wrongly defined the type as DataSource, it is a normal object. I guess I would have to define new types for these objects. Should I do that? Or don't write type in function parameter at all?

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -93,97 +91,103 @@ export default class SemiControlledAutoComplete extends React.Component<
...
 
-  disableDivider = (option: Object): boolean => {
-    if (option.type === 'separator') {
-      return true;
-    } else {
-      return false;
-    }
+  isOptionDisabled = (option: Option): boolean => {
+    // Writing shorthand, ie, return option.type === 'separator' fails flow test.
+    // All the cases of type Option have to be explicitly handled.
```
4ian: Thanks for the explanation 👍
I think we can remove the comment because it's not helping for the code understanding (but I'll remember this is necessary for the review) (for such things, you can also comment on the PR, like a reviewer, to tell "here be careful when reviewing, because blabla").

For someone in the future that might want to simplify the code, this person will get the Flow error and will understand why there is the extra if/else :)

4ian: Speaking of else! Extra nitpicking: you don't need the else ;)

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -93,97 +91,103 @@ export default class SemiControlledAutoComplete extends React.Component<
...
-    const handleOptionLabel = (option: DataSource | null): string => {
-      if (option.value) {
-        return option.value;
-      } else {
-        return currentInputValue();
-      }
+    const getOptionLabel = (option: Option): string => {
+      // option.value ? option.value : getCurrentInputValue();
+      // Writing shorthand will fail flow checks.
+      // Explicitly define all the cases of type Option.
```
4ian: Same, let's remove these comments

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -93,97 +91,103 @@ export default class SemiControlledAutoComplete extends React.Component<
...
+      var { InputProps, inputProps, InputLabelProps, ...other } = params;
       InputProps.className = null;
       InputProps.endAdornment = null;
       inputProps.className = null;
-      InputLabelProps.disableAnimation = true;
       InputProps.style = styles.inputRoot;
       inputProps.style = styles.inputInput;
+      inputProps = {
+        ...inputProps,
+        onKeyDown: (e): void => {
```
4ian: `e` => `event`

Use one letter variable only for for-loops (but prefer map/forEach anyway) or for unused variables.

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -93,97 +91,103 @@ export default class SemiControlledAutoComplete extends React.Component<
...
-      reason: string
-    ): void => {
+    const handleInputChange = (e, value: string, reason: string): void =>
       this.setState({ inputValue: value });
-    };
 
-    const nullifyDefaultStyling = (params: Object): void => {
-      const { InputProps, inputProps, InputLabelProps } = params;
+    const getDefaultStylingProps = (params: Object): Object => {
+      var { InputProps, inputProps, InputLabelProps, ...other } = params;
```
4ian: never use `var`, always `const`. `let` if you really must use it.

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -93,97 +91,103 @@ export default class SemiControlledAutoComplete extends React.Component<
...
       InputProps.style = styles.inputRoot;
       inputProps.style = styles.inputInput;
+      inputProps = {
+        ...inputProps,
+        onKeyDown: (e): void => {
+          if (e.key === 'Escape' && this.props.onRequestClose)
+            this.props.onRequestClose();
+        },
+      };
+      return { InputProps, inputProps, InputLabelProps, ...other };
```
4ian: Prefer a immutable way of doing things:

return {
  InputProps: {
    ...InputProps,
    className: null,
    endAdornment: null,
    style: styles.inputRoot
  },
  inputProps: {
    ...inputProps
    onKeyDown: (event) => // etc..
  },
  InputLabelProps
  // etc...
}

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -137,219 +69,192 @@ type State = {|
...
+      if (option.type && option.type === 'separator') {
+        return (
+          <ListItem
+            divider
+            disableGutters
+            component={'div'}
+            style={{ padding: 0, margin: 0 }}
+          />
+        );
+      } else {
```
4ian: You've used a return in the if, so it's fine I think to avoid the else and the extra indentation. it also emphasize that it's the most common path used in the code (you have few separators, but surely a lot of traditional items)

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -137,219 +69,192 @@ type State = {|
...
+    };
+
+    const renderItem = (option: Option, state: Object): React.Node => {
+      if (option.type && option.type === 'separator') {
+        return (
+          <ListItem
+            divider
+            disableGutters
+            component={'div'}
+            style={{ padding: 0, margin: 0 }}
```
4ian: nitpicking: put the style in `styles` (to avoid recreating an object at every render, slightly reducing the pressure on the GC). Same for below

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -137,219 +69,192 @@ type State = {|
...
               };
-
-              const onFocus = event => {
-                if (props.openOnFocus) openMenu();
-                this.setState({
-                  inputValue: props.value,
-                });
-              };
-
+              const defaultStylingProps = { InputProps, ...other };
```
4ian: Not sure if creating this object is worth it, just put the content in the `<TextField` props directly.

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -137,219 +69,192 @@ type State = {|
...
-                          )}
-                        </Paper>
-                      ) : null}
-                    </div>
-                  </Popper>
-                </div>
+                <TextField
+                  {...defaultStylingProps}
+                  {...computeTextFieldStyleProps(this.props)}
+                  style={{ ...this.props.textFieldStyle }}
```
4ian: style={this.props.textFieldStyle}

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -137,219 +69,192 @@ type State = {|
...
-              }
-            }}
-            onInputValueChange={inputValue => {
-              this.setState({ inputValue });
+          <Autocomplete
+            freeSolo
+            clearOnEscape
+            disableOpenOnFocus={!this.props.openOnFocus}
+            onChange={handleChange}
+            style={{ ...styles.container }}
```
4ian: Useless destructuring here

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -137,219 +69,192 @@ type State = {|
...
 
   render() {
-    const { props, state } = this;
-    const currentInputValue =
-      state.inputValue !== null ? state.inputValue : props.value;
+    const getCurrentInputValue = (): string =>
+      this.state.inputValue !== null ? this.state.inputValue : this.props.value;
+
+    const filterFunction = (options: DataSource, state: Object): DataSource => {
+      var separatorCount = 0;
```
4ian: Use `const` or `let` everywhere.

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -24,95 +20,31 @@ export type DataSource = Array<
       value: string, // The value to show on screen and to be selected
       onClick?: () => void, // If defined, will be called when the item is clicked. onChange/onChoose won't be called.
       renderIcon?: ?() => React.Element<typeof ListIcon | typeof SvgIcon>,
-    |}
->;
+    |};
+
+// type Option is defined in isolation to DataSource, since it needs to be
+// passed as an object for some funtions.
```
4ian: Thanks for the explanation 👍 I think we can remove this comment as this will be clear for the reader that will look at the code and see that Option is being used. Still a good thing for you to be conscious of this :)

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -137,219 +69,192 @@ type State = {|
...
+          return true;
+        }
+        if (!option.text) return true;
+        return option.text.toLowerCase().indexOf(lowercaseInputValue) !== -1;
+      });
 
-    const helperText = props.helperMarkdownText ? (
-      <MarkdownText source={props.helperMarkdownText} />
+      // Empty the list if there are only dividers inside it
+      if (optionList.length === separatorCount) optionList = [];
```
4ian: Could surely be rewritten as:
```
// Show an empty list if there are only separators.
if (!filteredOptionsList.filter(option => option.type !== 'separator').length) return [];
```

Which avoids the need to do the `separatorCount` and clearly separate the two concerns:
* first you filter the list according to the input value
* then you check if there are results, if not, you bail out.

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -137,219 +69,192 @@ type State = {|
...
+      this.state.inputValue !== null ? this.state.inputValue : this.props.value;
+
+    const filterFunction = (options: DataSource, state: Object): DataSource => {
+      var separatorCount = 0;
+      const lowercaseInputValue = getCurrentInputValue().toLowerCase();
+      var optionList = options.filter(option => {
+        if (option.type === 'separator') {
+          separatorCount++;
+          return true;
+        }
```
4ian: See next comment to remove this.

4ian:
Alright, this is nice progress! Mostly nitpickings on the code. I think we should be able to start making extra manual tests to ensure this is working well on Firefox/Chrome, Safari (if you have it, otherwise I'll try) - in other words check as much as possible the app :)

newIDE/app/src/UI/SemiControlledAutoComplete.js
HarsimranVirk: So this should be the first condition in that function, right? And after checking this, the list should get filtered?

`if(list contains only separator) return []` (bailing out)
`let count = 0`
`filter the list`
`delete separators at the start and end of array`

Is this correct?

HarsimranVirk: Alright, so I did what I said above. Let me know if its the correct way or I misunderstood you.
Also I had to write this,
`if (!options.filter(option => option.type !== 'separator' && option.text).length) return [];`

(it took me more than 15 mins to figure out how to make flow stop complaining :p I had to add a redundant explicit check)

HarsimranVirk:
@4ian okay so I guess SemiControlledAutocomplete is done as a component, now I have to pass onRequestClose() as a prop to all the inline editor fields. I could only figure out ObjectField, KeyField and VariableField (Object, global and scene variables) which use Autocomplete for inline editing, am I missing some fields? (There are 12 fields which can be edited inline and I have checked them all, only these 5 directly or indirectly use autocomplete)

4ian:
I'm unsure about onRequestClose 🤔
I do agree that it can be interesting to have it so that the parent can close itself, in the case of an inline popover in the events sheet.

But this is actually a new feature, i.e: a different task than what this PR was initially for (which was "just" replacing the autocomplete by a new component that works better on mobile and does not erase the input when Escape is pressed).

Let's go step by step: first let's polish this so that it's ready to merge, and then we can suggest an improvement, that will be easy to implement! Just a few props to pass... But still, it's more thing to review, more potential bugs, more potential missed thing that could break on desktop/mobile/whatever (you never know) - so let's keep it for later :)

newIDE/app/src/UI/SemiControlledAutoComplete.js
4ian: > if(list contains only separator) return [] (bailing out)
> filter the list
> delete separators at the start and end of array

You want to filter the list first no?
First you filter the list (that's the core filtering logic), and then you apply some beautifying stuff (i.e: you return an empty list if you have only separators and then you remove the separator that are at the end/beginning)

HarsimranVirk:
@4ian I had already added that in ObjectField in the earlier commits, should I remove it then?

4ian:
Yes please, let's remove for now - a bit of extra work but will be better to add later if needed.

HarsimranVirk:
@4ian So I have removed onRequstClose as a prop from all the fields I guess, but I've kept it in prop type definitions and also in the handling functions of autocomplete (even though it does nothing right now, but can be used in the future). 

> I do agree that it can be interesting to have it so that the parent can close itself, in the case of an inline popover in the events sheet.

We can maybe do this for other components too? (like TextField) I'd love to work on that after we're done with this :)

4ian:
> We can maybe do this for other components too? (like TextField) I'd love to work on that after we're done with this :)

If we want Esc to close the inline popover, I think we could do it with some events on the inline popover component itself - so as not disturb the poor other components that have nothing to do with this :) (events bubble up in the tree of components).

I will take a new look a the PR. As this is upgrading material-ui, I will probably need to set up a beta test to ensure there is no regression in the whole app.

newIDE/app/package.json
```diff
@@ -31,8 +31,9 @@
     "@blueprintjs/core": "file:src/Utils/BlueprintJsPlaceholder",
     "@blueprintjs/icons": "file:src/Utils/BlueprintJsPlaceholder",
     "@lingui/react": "git://github.com/4ian/lingui-react.git#master",
-    "@material-ui/core": "4.3.2",
+    "@material-ui/core": "4.9.0",
```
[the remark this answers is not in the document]
4ian: Would still be great to flag anything you might think of risky in the release notes - look for any changes that are highlighted.

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -137,219 +70,182 @@ type State = {|
...
+    const helperText = this.props.helperMarkdownText ? (
+      <MarkdownText source={this.props.helperMarkdownText} />
     ) : null;
 
+    const getOptionLabel = (option: Option): string => {
+      if (option.value) return option.value;
+      return getCurrentInputValue();
+    };
+
+    const handleChange = (event, option: Option | string): void => {
```
4ian: Add a type for event, even if you have to write it yourself it's better than nothing:
`{ key: string }`

because there will always be an event right?
Then remove the `typeof event.key === 'string'`

4ian:
Here we go, this is your branch online: https://editor-beta.gdevelop-app.com/ (don't share this apart for testing).

For now I've just tested quickly and not found any obvious regression :) 
The new autocomplete seems to work well, but has one issue: the existing text is not shown. So if I open an existing condition/action with one (or more) objects, the fields are empty. Same in the inline popover, and same in the Tiled Sprite/Panel Sprite object editors.
This is the kind of "details" that we need to care about :) Can you take a look?

Bouh:
I've seen only one thing, hand cursor on items, this hand is not present on current version.
mmmmm

Images are broken on gif but it's working few minutes ago, don't worries.

4ian:
Would be nice to fix the hand indeed, as we avoid displaying the hand in the rest of the IDE.
For opening the list, I was assuming it was a material-ui autocomplete default behavior - maybe there are props to tweak this?

HarsimranVirk:
I guess maybe hand is the default material ui styling for listbox, we've used custom listitem component, which doesn't show the hand cursor but listbox does. Also, yes there is a prop for opening list. @4ian Im aware of the issue you mentioned, I had removed the default value (which shows the existing value when we open autocomplete) because it threw some errors regarding controlled and uncontrolled input field.
I'll look into all of these issues this evening :)

HarsimranVirk:
1. Fixed the hand cursor thing.
2. Text field shows the existing value
3. Okay now about the list opening thing, I've fixed that but I'm really not sure about it.
https://github.com/4ian/GDevelop/blob/4362e8dd429620fece83ad2a7e5e5ac4c7a6d70c/newIDE/app/src/UI/SemiControlledAutoComplete.js#L133
So, earlier I was using this prop to control whether the list should open on focus or not. Now I've stopped using this prop and started using the default material-ui behavior. Am I missing something @4ian ? 

> Why not display the list directly when we click on?

> And the list isn't visible if the input is empty.
If we showing the list, user can save one click,

The default behavior doesn't have these issues, do you prefer them @Bouh ? What do you suggest @4ian ?

Bouh:
The current behavior in current beta have not these things indeed, but it can be an improvement todo these.
Like i said for tablet/phone user. And it's faster than write and see the choices.
And yeah like said 4ian, [auto complete](https://material-ui.com/components/autocomplete/) display list by default :)
In the instruction editor, in action "Link two object" for example when we select this action , the object field is focus by default and we see the list. This is cool because we see the possibility, and fast because we can just scroll with mouse or touch :)

You have made a nice rework! The change are invisible 👏

4ian:
I think we're fine always showing the list. It was not done previously because the it was hiding parameters in the action/condition editor... but as we're moving soon to a new the editor, we can surely keep it opened by default.

I've republished the latest changes on editor-beta.gdevelop-app.com. Make sure to open it, wait a bit, close it and re-open it to get it updated (as it's cached offline as a PWA) (we need to add a message to trigger the reload btw)

Bouh:
After have clear the cache.
image

HarsimranVirk:
@4ian I've come across another bug. We cannot select anything from the popup in EventsSheet with touch (mouse click or keyboard works fine). I've confirmed this exists in the current version too. I don't know if there is a problem with autocomplete or EventsSheet :/

4ian:
> I've come across another bug. We cannot select anything from the popup in EventsSheet with touch (mouse click or keyboard works fine). 

This is indeed an existing bug. Let's ignore this as it's not a regression of what you've done. This being said, still try to remove the "onRequestClose" and see if it's better.

HarsimranVirk:
> This being said, still try to remove the "onRequestClose" and see if it's better.

Though removing onRequestClose doesn't fix that bug but I'm still thinking to remove it either way.

> I think we could do it with some events on the inline popover component itself - so as not disturb the poor other components that have nothing to do with this 

Maybe we'll find some other way, so let's remove the onRequestClose. What do you suggest?

newIDE/app/src/EventsSheet/ParameterFields/ParameterFieldCommons.js
```diff
@@ -10,7 +10,7 @@ type CommonProps = {|
   // The parameter
   parameterMetadata?: gdParameterMetadata,
   onChange: string => void,
-  onRequestClose?: () => void,
+  //onRequestClose?: () => void,
```
4ian: Don't let commented code living => remove it entirely

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -131,227 +79,200 @@ type Props = {|
...
+          ...InputProps,
+          className: null,
+          endAdornment: null,
+        },
+        inputProps: {
+          ...inputProps,
+          className: null,
+          onKeyDown: (event): void => {
+            // if (event.key === 'Escape' && this.props.onRequestClose)
+            //   this.props.onRequestClose();
```
4ian: Remove the commented code ;) 
The code living in the codebase must be as simple as possible - dead code is 99% of the time just making things harder to understand. If you want to keep it, keep a separate branch on your local computer or your own git with this code.

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -131,227 +79,200 @@ type Props = {|
...
             }}
-          </Downshift>
+          />
         )}
       </I18n>
     );
   }
 }
+
+export default withStyles(themeStyles)(SemiControlledAutoComplete);
```
4ian: Not a big fan of the High Order Component version - it breaks the static typing it seems. Prefer to use the hook API: https://material-ui.com/styles/basics/#hook-api

(makeStyle and useStyle).

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -131,227 +79,200 @@ type Props = {|
...
+          }
+        }
+      }
+      if (
+        typeof event.key === 'string' &&
+        event.key === 'Enter' &&
+        this._input.current !== null
+      )
+        this._input.current.blur();
+      //if (this.props.onRequestClose) this.props.onRequestClose();
```
4ian: Commented code to remove

newIDE/app/src/UI/SemiControlledAutoComplete.js
HarsimranVirk: Im really sorry, should've checked before pushing, my bad.

HarsimranVirk: Im really sorry, should've checked before pushing, my bad.

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -131,227 +79,200 @@ type Props = {|
...
-                  // of an item (contrary to onChange, which can be called even
-                  // with a partial input when the field is blurred)
-                  if (props.onChoose) props.onChoose(selectedItem.value);
-                  else props.onChange(selectedItem.value);
-                }
-              }
+          <Autocomplete
+            freeSolo
+            blurOnSelect
+            classes={{
```
4ian: Can you just do `classes={classes}`? Does it work?

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -24,95 +22,45 @@ export type DataSource = Array<
...
-
-    return item.text.toLowerCase().indexOf(lowercaseInputValue) !== -1;
-  });
-};
+});
 
 type Props = {|
   value: string,
   onChange: string => void,
+  //onRequestClose?: () => void,
```
4ian: Commented code to remove

newIDE/app/src/UI/SemiControlledAutoComplete.js
4ian: Oh you're back at having a class component.. in this case, can we use the "styled-component" API? https://material-ui.com/styles/basics/#styled-components-api

=> Make a `StyledAutocomplete` component, then use it instead of `Autocomplete` :) This will also shorten slightly the render function.

(or see if it's easy to turn the component into a function component and use the hook API)

HarsimranVirk: Okay, I'd refactor the code then. I used HOC only because I didn't want to convert the class component into a function. 😅

newIDE/app/src/UI/SemiControlledAutoComplete.js
4ian: Ping on this: do you know the type of the event? :)

newIDE/app/src/ObjectsList/ObjectSelector.js
```diff
@@ -18,6 +18,7 @@ type Props = {|
 
   onChoose?: string => void,
   onChange: string => void,
+  onRequestClose?: () => void,
```
4ian: We probably want to remove the two lines added in this file because it's not useful

newIDE/app/src/EventsSheet/InlineParameterEditor.js
```diff
@@ -123,6 +123,7 @@ export default class InlineParameterEditor extends React.Component<
           parameterIndex={this.props.parameterIndex}
           value={instruction.getParameter(this.props.parameterIndex)}
           onChange={this.props.onChange}
+          onRequestClose={this.props.onRequestClose}
```
4ian: We probably want to remove this if not useful for now?

newIDE/app/src/UI/SemiControlledAutoComplete.js
HarsimranVirk: I'd check if it works. (I guess it should)

4ian:
I still added a few comments. I know it's a bit of pain because I want some "perfect" code, but to be precise I want a code that is as simple and correct as possible: not extra stuff if not needed, proper typing, no useless code, functions as short as possible.
The reason is that this is easing the maintenance in the long term. Having a component with 2/3 useless stuff, and typing not working because Flow does not have the definition of the HOC api of material-ui withStyle, will create problems down the road.
If we aim right now for a component as minimal, simple and clean as possible, we avoid mis-understanding in the future, we avoid bugs and we allow ourselves to have something that is easy to understand and replace :)

newIDE/app/src/UI/SemiControlledAutoComplete.js
4ian: In the long run that would be great! Otherwise take a look at this API https://material-ui.com/styles/basics/#styled-components-api - but if we can go ahead with a function that's even better! :)

newIDE/app/src/UI/SemiControlledAutoComplete.js
4ian: No worries :)

newIDE/app/src/EventsSheet/InlineParameterEditor.js
HarsimranVirk: Oh no, not this one, I guess this is passed for ClickAwayListener? We need atleast that to close the inline popover, no?

4ian: I don't know, I mean this is something added in this PR, and it already works in the current version no?
I don't see a `ClickAwayListener`, there is one in the InlinePopover if I recall properly, but this is not the responsibility of the ParameterComponent to deal with it, right?

HarsimranVirk:
Yes I understand @4ian :) Even I was thinking about it having too much bloating stuff in the render() function. This would surely result in performance issues in the long run. I believe its best to refactor stuff :)

newIDE/app/src/EventsSheet/InlineParameterEditor.js
HarsimranVirk: As far as I remember it was passed to the ClickAwayListener by the InlineParameterEditor before I started working on this. (Also this is how I thought of passing it even down, to the Autocomplete component. Maybe too much commits, adding and removing this polluted the history? )

4ian: Not sure, git is pretty robust :) Check if you can remove this. It's being added in this PR and I'm not sure it's useful

HarsimranVirk:
@4ian I'm stuck at implementing these two functions:
https://github.com/4ian/GDevelop/blob/8b294ae3694be64d625e7b256c5e931c923177e0/newIDE/app/src/UI/SemiControlledAutoComplete.js#L169-L179

They worked fine inside a class, but I don't understand how should I proceed with them in a functional component. I tried forwardRef and useImpertiveHandle hook for focus() but I don't think I should be using them (they are really messing up flow type checks at the moment).

HarsimranVirk:
@4ian I'm stuck at implementing these two functions:
https://github.com/4ian/GDevelop/blob/8b294ae3694be64d625e7b256c5e931c923177e0/newIDE/app/src/UI/SemiControlledAutoComplete.js#L169-L179

They are meant to work with class components and I'm not sure how to deal with them with functional component. Right now I've used forwardRef to pass the ref for focus() and it works (I don't know if it is the right way or not). Should I use forwardRef? How should I proceed with forceInputValueTo() ?

4ian:
To expose methods, you need to use React.forwardRef indeed, with the `Props` and the `Interface` as types, Interface being a type defining the functions. Then you use `useImperativeHandle`: https://github.com/4ian/GDevelop/search?q=useImperativeHandle&unscoped_q=useImperativeHandle

For example, see this example for `ScrollView` which has a method `scrollTo`: https://github.com/4ian/GDevelop/blob/9536bda67447163555b8b166a95719be669e2006/newIDE/app/src/UI/ScrollView.js

you should be able to re-implement focus and forceInputValueTo with `useImperativeHandle` :)

newIDE/app/src/EventsSheet/ParameterFields/BehaviorField.js
```diff
@@ -19,6 +19,7 @@ export default class BehaviorField extends React.Component<
   _longDescription: ?string;
   _behaviorTypeAllowed: ?string;
   _behaviorNames: Array<string> = [];
+  // $FlowFixMe
```
4ian: What's the issue? In 99% of cases we don't want to introduce FlowFixMe, as this hide Flow errors and reduce the potential of catching bugs in the future.

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -57,214 +90,180 @@ const themeStyles = theme => ({
...
 
-      if (
-        !optionList.filter(
-          option => option.type !== 'separator' && option.value
-        ).length
-      )
-        return [];
+export default React.forwardRef<Props, Interface>(
+  function SemiControlledAutoComplete(props: Props, ref) {
+    const _input = React.useRef((null: ?HTMLInputElement));
```
4ian: nitpicking: you can name this `input`, it's a local variable so no risk of external code accessing it, so no need to use _ to show that it's private :)

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -57,214 +90,180 @@ const themeStyles = theme => ({
...
-            // if (event.key === 'Escape' && this.props.onRequestClose)
-            //   this.props.onRequestClose();
-            if (event.key === 'Enter')
-              handleChange(event, getCurrentInputValue());
-          },
-        },
-      };
-    };
+    const handleInputChange = (
+      event: SyntheticKeyboardEvent<HTMLInputElement>,
```
4ian: Thanks for the typing 👍

4ian:
Looks good overall :) I see that you had to add $FlowFixMe - ideally we'll sort this out to be sure we can have proper typing of SemiControlledAutoComplete in the whole codebase :)

newIDE/app/src/EventsSheet/ParameterFields/BehaviorField.js
HarsimranVirk: I believe this is a general issue in the entire codebase (I might be wrong though). Flow complaints about forwardRef's typing. I couldn't come up with a solution to resolve this so I looked into ScrollView and EditorNavigator's code. 
https://github.com/4ian/GDevelop/blob/a315eabdcedb5f04b7bf287272fa46e7b43f924f/newIDE/app/src/EventsFunctionsExtensionEditor/index.js#L88-L89
https://github.com/4ian/GDevelop/blob/a315eabdcedb5f04b7bf287272fa46e7b43f924f/newIDE/app/src/EventsSheet/InstructionEditor/InstructionOrExpressionSelector/index.js#L70-L71

4ian: Ah you're right! :'( Forgot about that. 
I'd like to take a look but not sure to have much time... I was sure I managed to get something working at some point... should be doable especially considering that we now have the latest Flow version 🤔
Otherwise we'll go ahead without this.

Can you post the Flow error messages here for reference?

HarsimranVirk: > Can you post the Flow error messages here for reference?

Sure thing :)
Also, I had a doubt, what does the type Instance really mean? I thought it meant the 'instance' of a class for class-based components. But that couldn't be the case, right? Functional components would not have that sort of 'instance'.

HarsimranVirk: > Can you post the Flow error messages here for reference?

Cannot use AbstractComponent as a type because AbstractComponent is a value. To get the type of a value use `typeof`.

This is the error. Adding `typeof SemiControlledAutoComplete` introduces these two errors:

1. Cannot call `this._field.focus` because property `focus` is missing in  AbstractComponent [1]. Flow(InferError)
2. Cannot assign `field` to `this._field` because  `Interface` [1] is incompatible with  AbstractComponent [2]. Flow(InferError)

4ian: Let me try a few things 🧐

4ian: @HarsimranVirk I believe I fixed it in https://github.com/4ian/GDevelop/commit/4d63fbcca068627c2f2a1b97a558e1554f90ec91

=> export the interface and use it for the typing of the ref. :)

HarsimranVirk: Yes it solved the issue :+1:

4ian:
I've updated editor-beta.gdevelop-app.com :) 
If you can fix the conflicts in package-lock.json (I would recommend to erase the file and re-run npm install), we should I think be good to go?

Bouh:
Color on text don"t used style from themes
image

HarsimranVirk:
> I would recommend to erase the file and re-run npm install

It didn't work :(

> Color on text don't use styles from themes

@Bouh I'd take a look into it. Though, I don't understand why is that so? It uses all the other styling from theme, but not the text?

Now, about the clearing of input after a choice, so the input shows the existing value. After selecting a value, it becomes the 'existing' value and therefore it is shown in the input. It'd be really complex to differentiate between the actual existing value and the value which is just selected.

Bouh:
I've run electron app on my pc with your branch and it's good. Surely my browser cache doesn't clear properly.. sorry.

4ian:
> Color on text don't use styles from themes

I've updated the editor-beta.gdevelop-app.com yesterday ;) And it requires a cache clear or an opening/close/re-open to update itself otherwise.

4ian:
So unless there is a major issue this seems ready to go?

Bouh:
Yes

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -137,221 +54,222 @@ type State = {|
...
+      reason: string
+    ): void => setState({ inputValue: value });
+
     return (
       <I18n>
         {({ i18n }) => (
-          <Downshift
+          <Autocomplete
+            freeSolo
+            blurOnSelect
```
4ian: Any reason for blurOnSelect? I think we should remove it - it's important to keep the focus after a selection I think, in particular when selecting an object in a list of parameters, where you want to use Tab to go to the next field

(For example:

image

after selecting the object, we should be able to Tab to go to "Yes" button)

4ian:
Found that we lose focus when selecting an object - we should probably fix this by removing blurOnSelect?

newIDE/app/src/UI/SemiControlledAutoComplete.js
HarsimranVirk: I had added keybinding for 'Enter' which would let users select anything they type. If it was something arbitrary, the menu wouldn't close, so I had to blur it. To make it consistent, I added blurOnSelect. Now I guess I'd remove the keybinding to Enter (it's unnecessary at the moment) and use blur event to select arbitrary values like it used to happen with downshift :)

4ian: > Now I guess I'd remove the keybinding to Enter (it's unnecessary at the moment) and use blur event to select arbitrary values like it used to happen with downshift

Sounds good. Usage of the blur event is only because we're a "SemiControlledAutoComplete" - the "SemiControlled" means that it's keeping its state internally until it's blurred :) 

We definitely want to remove blurOnSelect - losing the focus would be a usability bug.
Removing the keybinding to Enter also sounds good :) Because I think it should be handled by the material-ui autocomplete. 

The "freeSolo" example on https://material-ui.com/components/autocomplete/ is how the autocomplete should behave I think

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -181,21 +177,20 @@ const getOptionLabel = (option: Option, value: string): string =>
 export default React.forwardRef<Props, SemiControlledAutoCompleteInterface>(
   function SemiControlledAutoComplete(props: Props, ref) {
     const input = React.useRef((null: ?HTMLInputElement));
-    const [state, setState] = useState<State>({ inputValue: null });
-    const [isMenuOpen, setIsMenuOpen] = useState<boolean>(true);
+    const [inputValue, setInputValue] = useState((null: string | null));
```
HarsimranVirk: `null: ?string` gives a lot of flow errors due to type void, so I used this instead. Also, I tried resolving those errors, but then we could never have an empty input value in the text field (let's say for "Players" if you try to delete it, you'd get the last "P", and deleting this "P" will make the text "Players" again).

4ian: Strange because `?string` is equivalent to `string | null | undefined`.

Anyway, not a big deal because `string | null` is actually a bit better.

4ian:
Still found something that looks not ideal. When a dialog is opened with an autocomplete, it is opened by default even if not focused:
image

See also in the Icons dialog:
image

Be sure to always double check the app when doing changes - this is the kind of problems that I can miss because I don't have time to test everything :) 

I've not investigated yet why this is the case, can you have a look?

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -181,21 +177,20 @@ const getOptionLabel = (option: Option, value: string): string =>
...
+    const [isMenuOpen, setIsMenuOpen] = useState(true);
```
HarsimranVirk: @4ian, this is the issue, this should basically mean open the menu when autocomplete is rendered (for the first time). I've fixed this by setting it false. The menu will only open up when autocomplete is focused.

4ian: Sounds good 👍

4ian:
Thanks for the quick fix :) Apart from that it looked good. I'll give another try today and merge if everything looks good. 
There is a slight chance that the material-ui update could modify/break something that we've not seen yet, but I'm hopeful the library is stable enough so that should not be a problem.

Bouh:
I'll check for too

HarsimranVirk:
@4ian, I can safely say that I have checked at most of the places (if not all), and this is the behavior of autocomplete:

1. Focusing on the autocomplete would automatically open the menu list, saving the user a click, as suggested by @Bouh.
2. Input will show the existing value if there is any. (props.value)
3. The menu will open only on focus. (fixed in the latest commit)
4. Users can select the option as usual, and options are rendered the way they should (ie, with icon and separation between groups and individual objects).
5. Focus is not lost after selecting the option.
6. Tab to focus or Shift + Tab to refocus returns the cursor at the end of text by default, though the user's choice of position of the curser overrides this.
7. Theme and styling is applied correctly.

Is there anything I'm missing? Or is there anything unwanted?

Bouh:
During my turn in the app, i've test lot of case, just this one is not pretty, it's just cosmetic.
Other way you have made a really great job on this PR! 👍 
image

HarsimranVirk:
Oh well, that is really bad. Thanks @Bouh, this would have been totally missed, I personally never thought about overflow-x :) Now, what do you suggest? Should we cut off the overflowing content, and perhaps use ellipsis?

HarsimranVirk:
Screenshot from 2020-03-19 22-34-25

Screenshot from 2020-03-19 22-37-48

Does this look good? @Bouh @4ian

4ian:
Agree with @Bouh, this is some great job 👍

> Does this look good? @Bouh @4ian

Yes, looks good! I was thinking it would be handled by material-ui's AutoComplete 🤔But your solution looks great :)

HarsimranVirk:
> I was thinking it would be handled by material-ui's AutoComplete

I thought the same :/ Turns out material-ui doesn't handle it at all, I have looked through their implementation and I'm sure about this.

Screenshot from 2020-03-19 23-07-20

Anyway, I'd check a bit more before committing, I've changed `display: flex` to `display: inline-block` for ListItem. This might have some unwanted results (though I believe it wouldn't really matter at all)

HarsimranVirk:
> Is width: '100%' working?

No. `text-overflow: ellipses` doesn't work without `width` in `px`. This is more of a css rule. Using calc() is a common workaround for this. (Also, I checked with 100% before doing this, hoping it might work :/ )

4ian:
Ok thanks for checking! Looks fine then :)

4ian:
Merged! Let me know if you find any issue so that we can fix them quickly - hopefully this is stable :)

Thanks a lot for your patience and your work on this! This is a great improvement that was really needed since a bit of time. 🎉 Keep up the good work 👍👍

Issue #1339, opened 2019-12-28
Jump down through platform in "PlatformerObject"


CustomLenny:
## Description
There is no way to trigger (even through javascript as far as I could tell) the platformer characters to be in a force falling state.
I couldn't find any way to shift a character below the platform and have him fall.
The down arrow when set to sidescrolling is not used for anything besides going down a ladder.
When on a "jumpthrough" platform the down arrow should be able to let the character drop through the platform.
Jump down is sometimes done by holding the down and then pressing the jump, but it doesn't need to be that way as anybody who would like it that way can easily do that through the editor, in addition to that that would be a bad way for mobile devices.

## Solution suggested
Down arrow on a platformerobject set to sidescroller lets player fall through the jumpthrough platform.

## Alternatives considered
Javascript ways to set the isfalling and isjumping and any other flags by force.

4ian:
I've added a card for this on the roadmap: https://trello.com/c/jFSuWrm6/391-allow-platformer-objects-to-fall-from-a-jumpthru-platform-when-a-specific-key-is-pressed
I'll close the issue as it's a feature request but discussion can continue here.

Have you tried to change the Y position of the object to add 1 pixel? In theory that should put the object below the jumpthru and have it fall.

CustomLenny:
> Have you tried to change the Y position of the object to add 1 pixel? In theory that should put the object below the jumpthru and have it fall.

Yes I have (and more than 1 as well), the flag for is on ground or is falling does not flip though so when moved down it still thinks it hasn't left the ground.

4ian:
An alternative could be to temporarily disable the jumpthru "platform" behavior for a few frames, letting the time for the platformer object to fall (one frame in theory should be enough) - but with the downside that other platformer objects on the platform will fall too.

I've checked the code and yeah seems that if I don't add something to force the "on floor" flag to be false, the object will try to stick to the floor as long as it's nearby.

CustomLenny:
> An alternative could be to temporarily disable the jumpthru "platform" behavior for a few frames, letting the time for the platformer object to fall (one frame in theory should be enough) - but with the downside that other platformer objects on the platform will fall too.

Yeah I have more than 1 platformer character.
Which the idea of adding a fall through will surely be of use to any items that are using platform character as well.

> I've checked the code and yeah seems that if I don't add something to force the "on floor" flag to be false, the object will try to stick to the floor as long as it's nearby.

[...]

PR #1457, opened 2020-02-23
Specific platform ignoring

arthuro555:
See #1339 
Let a specific instance of a Platformer behavior ignore a specific instance of a Platform Behavior.

Only has a JS implementation.

I am not sure about the declaration of the extension, and on my current setup I cannot compile GDevelop.js so I cannot test the modifications. They seem to compile fine on circle ci though.

### DIFF
--- a/Extensions/PlatformBehavior/Extension.cpp
+++ b/Extensions/PlatformBehavior/Extension.cpp
@@ -502,6 +502,28 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
         .AddParameter("behavior", _("Behavior"), "PlatformerObjectBehavior")
         .SetFunctionName("GetJumpSpeed")
         .SetIncludeFile("PlatformBehavior/PlatformerObjectRuntimeBehavior.h");
+
+    aut.AddExpression("ignorePlatform",
+                          _("Jump speed"),
+                          _("Jump speed"),
+                          _("Options"),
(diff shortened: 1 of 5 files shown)


4ian:
For now I'm not entirely convinced we need to add this extra complexity in the code. :/
This forces us to use IDs, which is almost an anti-pattern that I reluctantly added to objects but that I would like to avoid as much as possible (risks of overflow, risk of IDs reused, if you delete a platform you still have its ID polluting the "ignored platform" array of all the behaviors that used to ignore it, if you create a new platform and for some reason reuse an ID, you now have a bug, etc...).

It's also partially solvable by disabling the behavior - though I agree it's not giving the exact same result because other platformer characters are also implied.

What do you think?

arthuro555:
I do think IDs are not the best way to do it but I don't see any other way of achieving this. But I also think this is important for multiplayer games. I think it's the most optimized way, even if it is not very optimal. Do you have any other ideas on how to implement it?

4ian:
Not really, but it's the goal of this PR to be finding the best solution :) As I said previously, IDs have all the risks that I wrote and I don't think they are addressed.

I don't have a better idea - but we'll always prefer to *not* have a feature rather than having one that is creating risks/potential bugs/hard to debug behaviors :/

arthuro555:
well, I'm closing this because there are too much potential unsolvable problems.

PR #1458, opened 2020-02-23
Removed non-required use of Hashtable.items

arthuro555:
Hastable.items was used in a non-iteration context. This should not be the case. Hashtable.get() is the correct way to proceed.


4ian:
Sounds fair! Merging this. Thanks! :)

PR #1467, opened 2020-02-25
Fix maj, this avoid doubles in translations on crowdin

Bouh:
Fix maj, this avoid doubles in translations on crowdin


4ian:
Good catch, thanks!

PR #1477, opened 2020-02-28
Fix object name cannot be selected

### LINKED ISSUE #1440: Object name can't be selected
## Describe the bug
When editing the name of an object, the text cannot be selected with the mouse. Instead, it drags the object.
image

Not happening with groups, scenes and layers.

geetesh-gupta:
Fixes #1440

### DIFF
--- a/newIDE/app/src/UI/DragAndDrop/DragSourceAndDropTarget.js
+++ b/newIDE/app/src/UI/DragAndDrop/DragSourceAndDropTarget.js
@@ -44,7 +44,7 @@ export const makeDragSourceAndDropTarget = <DraggedItemType>(
   reactDndType: string
 ): ((Props<DraggedItemType>) => React.Node) => {
   const sourceSpec = {
-    canDrag(props: Props<DraggedItemType>, monitor: DragTargetMonitor) {
+    canDrag(props: Props<DraggedItemType>, monitor: DragSourceMonitor) {
       const item = monitor.getItem();
       const canDrag = props.canDrag || null;
       if (canDrag) return canDrag(item);
--- a/newIDE/app/src/UI/DragAndDrop/DragSourceAndDropTarget.js
+++ b/newIDE/app/src/UI/DragAndDrop/DragSourceAndDropTarget.js
@@ -19,6 +19,7 @@ type Props<DraggedItemType> = {|
     canDrop: boolean,
   }) => React.Node,
   beginDrag: () => DraggedItemType,
+  canDrag: (item: DraggedItemType) => boolean,
   canDrop: (item: DraggedItemType) => boolean,
   drop: () => void,
 |};
@@ -43,6 +44,12 @@ export const makeDragSourceAndDropTarget = <DraggedItemType>(
   reactDndType: string
 ): ((Props<DraggedItemType>) => React.Node) => {
   const sourceSpec = {
+    canDrag(props: Props<DraggedItemType>, monitor: DragTargetMonitor) {
+      const item = monitor.getItem();
+      const canDrag = props.canDrag || null;
+      if (canDrag) return canDrag(item);
+      else return true;
+    },
     beginDrag(props: InnerDragSourceAndDropTargetProps<DraggedItemType>) {
       return props.beginDrag();
     },
--- a/newIDE/app/src/UI/DragAndDrop/DragSourceAndDropTarget.js
+++ b/newIDE/app/src/UI/DragAndDrop/DragSourceAndDropTarget.js
@@ -48,7 +48,7 @@ export const makeDragSourceAndDropTarget = <DraggedItemType>(
       const item = monitor.getItem();
       const canDrag = props.canDrag || null;
       if (canDrag) return canDrag(item);
-      else return true;
+      return true;
     },
     beginDrag(props: InnerDragSourceAndDropTargetProps<DraggedItemType>) {
       return props.beginDrag();
--- a/newIDE/app/src/UI/SortableVirtualizedItemList/index.js
+++ b/newIDE/app/src/UI/SortableVirtualizedItemList/index.js
@@ -100,73 +100,77 @@ export default class SortableVirtualizedItemList<Item> extends React.Component<
                   const nameBeingEdited = renamedItem === item;
                   const itemName = getItemName(item);
 
+                  const itemRow = (
+                    <ItemRow
+                      item={item}
+                      itemName={itemName}
+                      isBold={isItemBold ? isItemBold(item) : false}
+                      onRename={newName => this.props.onRename(item, newName)}
+                      editingName={nameBeingEdited}
+                      getThumbnail={
+                        getItemThumbnail
+                          ? () => getItemThumbnail(item)
+                          : undefined
+                      }
+                      selected={selectedItems.indexOf(item) !== -1}
+                      onItemSelected={this.props.onItemSelected}
+                      errorStatus={
+                        erroredItems ? erroredItems[itemName] || '' : ''
+                      }
+                      buildMenuTemplate={() =>
+                        this.props.buildMenuTemplate(item, index)
+                      }
+                      onEdit={onEditItem}
+                      hideMenuButton={windowWidth === 'small'}
+                    />
+                  );
+
                   return (
                     <div style={style} key={key}>
-                      <DragSourceAndDropTarget
-                        beginDrag={() => {
-                          this.props.onItemSelected(item);
-                          return {};
-                        }}
-                        canDrop={() =>
-                          canMoveSelectionToItem
-                            ? canMoveSelectionToItem(item)
-                            : true
-                        }
-                        drop={() => {
-                          onMoveSelectionToItem(item);
-                        }}
-                      >
-                        {({
-                          connectDragSource,
-                          connectDropTarget,
-                          isOver,
-                          canDrop,
-                        }) => {
-                          // Add an extra div because connectDropTarget/connectDragSource can
-                          // only be used on native elements
-                          const dropTarget = connectDropTarget(
-                            <div>
-                              {isOver && <DropIndicator canDrop={canDrop} />}
-                              <ItemRow
-                                item={item}
-                                itemName={itemName}
-                                isBold={isItemBold ? isItemBold(item) : false}
-                                onRename={newName =>
-                                  this.props.onRename(item, newName)
-                                }
-                                editingName={nameBeingEdited}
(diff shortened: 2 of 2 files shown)


newIDE/app/src/ObjectsList/index.js
```diff
@@ -296,15 +296,14 @@ export default class ObjectsList extends React.Component<Props, State> {
...
-      return;
-    }
-
     this.setState({
       renamedObjectWithContext: null,
     });
 
-    if (this.props.canRenameObject(newName)) {
+    if (
+      getObjectWithContextName(objectWithContext) === newName ||
```
4ian: See my comment in the other PR, the condition seems like it should be the reverse.

newIDE/app/src/UI/SortableVirtualizedItemList/index.js
```diff
@@ -100,73 +100,77 @@ export default class SortableVirtualizedItemList<Item> extends React.Component<
...
+                            connectDropTarget,
+                            isOver,
+                            canDrop,
+                          }) => {
+                            // Add an extra div because connectDropTarget/connectDragSource can
+                            // only be used on native elements
+                            const dropTarget = connectDropTarget(
+                              <div>
+                                {isOver && <DropIndicator canDrop={canDrop} />}
+                                {React.cloneElement(itemRow, {
```
4ian: Prefer to use a function that returns the `<ItemRow` and take optional props as arguments. Avoid cloneElement in 99.9% of cases - it's making static analysis harder.

4ian:
Seems like a proper fix! 👍
I've added comments about not using React.cloneElement (avoid "at all cost" except in very specific cases). A render function taking some additional props will do the trick.
You've also included the changes from another PR => in general avoid to do that and start your fix on a new branch from master. So that all PRs are independant :) (even if here it's not a big deal, it can be important for larger PRs).

newIDE/app/src/UI/SortableVirtualizedItemList/index.js
geetesh-gupta: After tinkering through the code, I have come to the conclusion, that to prevent dragging I have to return the <ItemRow ...> component directly and for dragging I have to wrap it in <DragSourceAndDropTarget ...> and then return it. 

Firstly, please correct me if I am wrong here. (i.e. if their is a possibility of preventing dragging and at the same time be wrapped inside <DragSourceAndDropTarget ...>).

And if my interpretation is correct, wrapping it will require additional prop of *connectIconDragSource* which can only be provided after wrapping it. 
So, I have come to following two solutions:
1. I can use the two separate instances of<ItemRow ...> for both cases (one inside the <DragSourceAndDropTarget ...> with *connectIconDragSource* prop and other without it). But, this will lead to duplication of code for <ItemRow ...> instance because all other props are same.
2. The solution proposed above of cloneElement, but as you correctly mentioned that it makes static analysis harder, so no to it.
3. Create separate renderItemRow function which takes props as arguments and return the <ItemRow ...>  instance.

So, which method to follow among these or any other approach you have come across.

newIDE/app/src/ObjectsList/index.js
geetesh-gupta: I have removed the commit from this PR

4ian:
I received a comment email, but can't see it. Probably you added it to a commit that you then force pushed.

You wrote about these 3 solutions:


>1. I can use the two separate instances of<ItemRow ...> for both cases (one inside the <DragSourceAndDropTarget ...> with connectIconDragSource prop and other without it). But, this will lead to duplication of code for <ItemRow ...> instance because all other props are same.
>2. The solution proposed above of cloneElement, but as you correctly mentioned that it makes static analysis harder, so no to it.
>3. Create separate renderItemRow function which takes props as arguments and return the <ItemRow ...> instance.

I think 3 is the way to go. cloneElement is a no go in most cases because it can be done with functions which are much more explicit and can be analyzed by Flow.
Two separate instances is probably duplication that can be avoided with a function.

So let's use a function. You call it, either itself or inside `<DragSourceAndDropTarget ...`, passing the props it needs (or not, depending on the case). The props should be quite limited in number anyway.

newIDE/app/src/UI/SortableVirtualizedItemList/index.js
```diff
@@ -102,71 +135,60 @@ export default class SortableVirtualizedItemList<Item> extends React.Component<
```
4ian: Nitpicking: as we have now renderItemRow, let's move these nameBeingEdited and itemName to renderItemRow.

newIDE/app/src/UI/SortableVirtualizedItemList/index.js
```diff
@@ -42,20 +42,53 @@ export default class SortableVirtualizedItemList<Item> extends React.Component<
     if (this._list) this._list.forceUpdateGrid();
   }
 
+  renderItemRow(
```
4ian: Nitpicking: `_renderItemRow`, to emphasise that it's not to be called from the outside.

newIDE/app/src/UI/SortableVirtualizedItemList/index.js
4ian: This will reduce the number of arguments that you need to send to renderItemRow, which is great because as it's a renderXXX function, it must be autonomous as much as possible (could even be turned into its own component later - but I don't think it's worth doing it now).

newIDE/app/src/UI/SortableVirtualizedItemList/index.js
```diff
@@ -98,75 +133,55 @@ export default class SortableVirtualizedItemList<Item> extends React.Component<
...
+                            connectDropTarget,
+                            isOver,
+                            canDrop,
+                          }) => {
+                            // Add an extra div because connectDropTarget/connectDragSource can
+                            // only be used on native elements
+                            const dropTarget = connectDropTarget(
+                              <div>
+                                {isOver && <DropIndicator canDrop={canDrop} />}
+                                {// If on a touch screen, only set the icon to be draggable.
```
4ian: This comment is now out of place, should be relocated next to what it describes.

4ian:
Alright, looks almost ready! Still something that I'd like you to explore:
Drag'n'drop is implemented using `react-dnd`, and using here a component called `DragSourceAndDropTarget`. It would be interesting to see if we could add a `disabled` property, or a `canDrag={false}` on this component.

That's a different solution - sorry I just think about it now! Not saying that it's better yet, but I think it's interesting to try it separately to see the difference. Could you check the doc of react-dnd? https://react-dnd.github.io/react-dnd/docs/api/drag-source => there is a `canDrag` method that we could define and that could read a prop that we pass to `DragSourceAndDropTarget`. See the idea? 
Your changes won't be useless, we'll keep the _renderItemRow even in this case - but we can maybe simplify the condition in the render method.

Let me know what you think

newIDE/app/src/UI/SortableVirtualizedItemList/index.js
geetesh-gupta: I'll update it :)

geetesh-gupta:
> Alright, looks almost ready! Still something that I'd like you to explore:
> Drag'n'drop is implemented using `react-dnd`, and using here a component called `DragSourceAndDropTarget`. It would be interesting to see if we could add a `disabled` property, or a `canDrag={false}` on this component.
> 
> That's a different solution - sorry I just think about it now! Not saying that it's better yet, but I think it's interesting to try it separately to see the difference. Could you check the doc of react-dnd? https://react-dnd.github.io/react-dnd/docs/api/drag-source => there is a `canDrag` method that we could define and that could read a prop that we pass to `DragSourceAndDropTarget`. See the idea?
> Your changes won't be useless, we'll keep the _renderItemRow even in this case - but we can maybe simplify the condition in the render method.
> 
> Let me know what you think

Actually I was also thinking of such a solution where we would not need to create a separate component instance, but I was unable to find that implemented. So, I used this workaround. But, as you have suggested, I will try to have a look at the library's doc to find if I can apply that here. :)

newIDE/app/src/UI/DragAndDrop/DragSourceAndDropTarget.js
```diff
@@ -43,6 +44,12 @@ export const makeDragSourceAndDropTarget = <DraggedItemType>(
   reactDndType: string
 ): ((Props<DraggedItemType>) => React.Node) => {
   const sourceSpec = {
+    canDrag(props: Props<DraggedItemType>, monitor: DragSourceMonitor) {
+      const item = monitor.getItem();
+      const canDrag = props.canDrag || null;
```
4ian: nitpicking: avoid null as much as possible - keep your variables to a simple type (string/boolean/number) if you don't need more.
```suggestion
      const canDrag = props.canDrag || false;
```

newIDE/app/src/UI/DragAndDrop/DragSourceAndDropTarget.js
```diff
@@ -43,6 +44,12 @@ export const makeDragSourceAndDropTarget = <DraggedItemType>(
...
+      if (canDrag) return canDrag(item);
+      else return true;
```
4ian: nitpicking: useless else

newIDE/app/src/UI/SortableVirtualizedItemList/index.js
```diff
@@ -42,20 +42,55 @@ export default class SortableVirtualizedItemList<Item> extends React.Component<
     if (this._list) this._list.forceUpdateGrid();
   }
 
+  _renderItemRow(
+    item: Item,
+    index: number,
+    windowWidth: string,
```
4ian:
```suggestion
    windowWidth: WidthType,
```

4ian:
Great job! This looks much better and cleaner with this `canDrag` :D 

Two comments and we should be good to merge this.

newIDE/app/src/UI/DragAndDrop/DragSourceAndDropTarget.js
geetesh-gupta: Here, the function canDrag requires a boolean return type, so removing it results in flow error.

4ian: I mean we can remove the else keyword :)

4ian:
Looks good, let's merge this! 
Thanks for this nice refactoring, appreciate the result and in particular the `canDrag`, which really looks like the best solution - very clean!

PR #1437, opened 2020-02-16
Accept unchanged names when renaming objects

arthuro555:
See #1436 .

PR #1479, opened 2020-02-28
Fix exiting without renaming an object

### LINKED ISSUE #1475: Issues while renaming an object
## Describe the Bug
Similar to #1437 
Until the current name is changed, the field doesn't exit as shown in this video https://gofile.io/?c=0WFBPW . 

## To Reproduce
1. Go to 'a new project'
2. Right Click on any object name 
3. Click on rename option
4. Press Enter without changing the name
4. See error - it does not exit the field until renamed.

geetesh-gupta:
Fixes #1475 #1437 
Also, fixes the issue of not exiting the field until enter is pressed. Allows to exit the field by clicking outside the object box.

### DIFF
--- a/newIDE/app/src/ObjectsList/index.js
+++ b/newIDE/app/src/ObjectsList/index.js
@@ -296,15 +296,14 @@ export default class ObjectsList extends React.Component<Props, State> {
   _rename = (objectWithContext: ObjectWithContext, newName: string) => {
     const { object } = objectWithContext;
 
-    if (getObjectWithContextName(objectWithContext) === newName) {
-      return;
-    }
-
     this.setState({
       renamedObjectWithContext: null,
     });
 
-    if (this.props.canRenameObject(newName)) {
+    if (
+      getObjectWithContextName(objectWithContext) === newName ||
+      this.props.canRenameObject(newName)
+    ) {
       this.props.onRenameObject(objectWithContext, newName, doRename => {
         if (!doRename) return;
 
--- a/newIDE/app/src/ObjectsList/index.js
+++ b/newIDE/app/src/ObjectsList/index.js
@@ -295,15 +295,13 @@ export default class ObjectsList extends React.Component<Props, State> {
 
   _rename = (objectWithContext: ObjectWithContext, newName: string) => {
     const { object } = objectWithContext;
-
     this.setState({
       renamedObjectWithContext: null,
     });
 
-    if (
-      getObjectWithContextName(objectWithContext) === newName ||
-      this.props.canRenameObject(newName)
-    ) {
+    if (getObjectWithContextName(objectWithContext) === newName) return;
+
+    if (this.props.canRenameObject(newName)) {
       this.props.onRenameObject(objectWithContext, newName, doRename => {
         if (!doRename) return;
 


newIDE/app/src/ObjectsList/index.js
```diff
@@ -296,15 +296,14 @@ export default class ObjectsList extends React.Component<Props, State> {
...
-      return;
-    }
-
     this.setState({
       renamedObjectWithContext: null,
     });
 
-    if (this.props.canRenameObject(newName)) {
+    if (
+      getObjectWithContextName(objectWithContext) === newName ||
```
4ian: The condition seems incorrect. Can you double check?
You must bail out if the newName is the same as the old one. Here you're proceeding with the renaming.

geetesh-gupta: If we directly return then, the object does not get re-rendered and it remains in selected/highlighted state. So, to cure I allowed allowed it to be renamed. Since the newName will be same as old one, nothing will change except the object will be re-rendered as if the newName is being assigned to it, so it will exit the selected/highlighted state.

geetesh-gupta: I have found out another possible solution which will prevent calling renaming function.

    if (getObjectWithContextName(objectWithContext) === newName) {
      this.setState(
        {
        },
        () => {
          if (this.sortableList) this.sortableList.forceUpdateGrid();
        }
      );
      return;
    }
This solution can also work. Which one to follow?

geetesh-gupta: Also, I have doubt regarding the same sort of functionality being used in other files as well wherever _rename function is used. I have not manually reproduced the actual behavior in such cases but I think the same changes should also be there in those files.

4ian: >  I have not manually reproduced the actual behavior in such cases but I think the same changes should also be there in those files.

First investigate if they have the issue. From what I recall, this is not the case and they should not need any fix - but double check.

> This solution can also work. Which one to follow?

Bail out early, and don't proceed with the renaming. Otherwise, the editor will start refactoring the GDevelop project/layout to make the name change (even if it turns out there is no change) and in the future someone might add something in the function when a renaming happen (the `onRenameObject` prop) - for example someone can add a counter of the number of renaming you've made (useless, but that's just an example). If we don't bail out and call `onRenameObject`, then there will be a new bug "the renaming counter is incremented even when I don't rename the object!".

In other words => minimise the surprise for the future developer, and in general minimise the work to be done by the app. The user did not change the name? Then don't even call the props to do the renaming.

BUT by doing this, seems that the list is not updating. This is another problem, and to make things simple this is because you've modified the state but the virtualized list is not aware of this. Try if calling `this.forceUpdate` works. Otherwise, use `this.forceUpdateList();` (which is doing the same thing that you did, roughly).

geetesh-gupta: By further investigating, I found that we only need to set *renamedObjectWithContext" to null.

Also, I checked other files with similar *_rename* function and found out that others have set *renamedObjectWithContext* to null before checking for equality but in current file this was being done after checking. So, the variable was not being set to null. So, setting the variable to null before checking is the solution that I found as given below.

image

4ian: Ha great! :) I thought that's what you already tried before. You must indeed bail out after setting `renamedObjectWithContext`... because this is what represents that there is an object renamed :) 
If you bail out before, you're not telling the UI to end the renaming, because this state is what represents that there is a renaming going on.

So everything works well without `this.forceUpdateList()` => don't use it.
In principle in React driven UI, everything should be accomplished by setting state - forceUpdate is a escape hatch in case something was updated outside of the state.

Happy to know that it's working like this then.

4ian:

PR #1480, opened 2020-02-28
Fixed Active Tab bug

### LINKED ISSUE #986: Active tab changes when closing a tab
## Describe the bug
When several tabs are opened, closing one moves the tabs to fill the gap, but the active tab remains at the same place, which changes the opened tab.

## To Reproduce
Open 4 tabs (two scenes, for instance).
Click on the third tab to make it active.
Close the second tab. The third tab moved to its spot, and the fourth tab has become the active tab.

## Other details
Windows 10
GD5 b65

gautamv95:
Fixes #986


4ian:
Sounds like the simplest solution indeed, let's go ahead with this. Thanks for the PR! 👍

Issue #1476, opened 2020-02-28
Scroll in list of parameters for extensions isn't scrollable [Web app]


Bouh:
## Describe the bug
The scroll bar for function configuration in extension are not scrollable, only on web-app.

On screenshot for the function behavior, but it's same for classic function.
All tabs in **Function Configuration** are affected.

![image](https://user-images.githubusercontent.com/1670670/75562484-f30bbf80-5a48-11ea-997e-27d921cafd9c.png)

## Other details
- Only web app
- beta 88

4ian:
Damn, surely a flexbox issue where there is an extra/missing `flex`/`flex-basis`/`flex-grow`/`flex-shrink`. Always a bit of a pain to debug, but should be a question of time to find it with the Chrome debugger.

deadpool1999:
May I please know which folder contains the code for this?

Bouh:
Start your research with EventsFunctionConfigurationEditor component.

Le ven. 28 févr. 2020 à 19:05, deadpool1999 <notifications@github.com> a
écrit :

jimil749:
Hello!

I did look into the issue and found the problem. Just had to provide predefined height to the outer container and that seemed to make the function configuration scrollable.
![Screenshot from 2020-02-29 15-11-27](https://user-images.githubusercontent.com/47107987/75605176-81924680-5b06-11ea-845f-5654ed2321c2.png)

I did this using Chrome dev tools.

HarsimranVirk:
@4ian I have fixed this issue, should I open a PR or let others try?

Bouh:
When you have fixed don't wait and open your PR :)

4ian:
Yes feel free to go ahead and make a PR => we'll discuss more in the PR if needed. :)
In particular, would be great to give this a try on Firefox/Chrome/Safari (if possible) and the Electron version (which is Chrome, but not the very latest version).

85ayush:
I would love to work on this issue, I am new to open source contribution, can you please guide me?

PR #1483, opened 2020-02-29
Fixed unscrollable parameter list

HarsimranVirk:
Fixed issue #1476.

> I did look into the issue and found the problem. Just had to provide predefined height to the outer container and that seemed to make the function configuration scrollable.

This works perfectly in chrome and electron but doesn't work in firefox. So, I found another workaround. Instead of setting the height to 100%, I changed the overflow-y to scroll. Works in chrome, electron, and firefox :) 
Haven't tested on safari since I don't have it. :/

### DIFF
--- a/newIDE/app/src/UI/Grid.js
+++ b/newIDE/app/src/UI/Grid.js
@@ -36,7 +36,7 @@ export const Column = props => (
       alignItems: props.alignItems || 'stretch',
       justifyContent: props.justifyContent,
       flex: props.expand ? 1 : undefined,
-      useMaxHeight: props.useMaxHeight ? '100%' : undefined,
+      maxHeight: props.useMaxHeight ? '100%' : undefined,
     }}
   >
     {props.children}
--- a/newIDE/app/src/UI/Grid.js
+++ b/newIDE/app/src/UI/Grid.js
@@ -36,7 +36,7 @@ export const Column = props => (
       alignItems: props.alignItems || 'stretch',
       justifyContent: props.justifyContent,
       flex: props.expand ? 1 : undefined,
-      overflowY: props.scrollY ? 'scroll' ? undefined,
+      overflowY: props.scrollY ? 'scroll' : undefined,
     }}
   >
     {props.children}
--- a/newIDE/app/src/UI/Grid.js
+++ b/newIDE/app/src/UI/Grid.js
@@ -36,7 +36,7 @@ export const Column = props => (
       alignItems: props.alignItems || 'stretch',
       justifyContent: props.justifyContent,
       flex: props.expand ? 1 : undefined,
-      overflowY: props.scrollY ? 'scroll' : undefined,
+      maxHeight: props.maxHeight ? props.maxHeight : undefined,
     }}
   >
     {props.children}
--- a/newIDE/app/src/UI/Grid.js
+++ b/newIDE/app/src/UI/Grid.js
@@ -36,6 +36,7 @@ export const Column = props => (
       alignItems: props.alignItems || 'stretch',
       justifyContent: props.justifyContent,
       flex: props.expand ? 1 : undefined,
+      overflowY: props.scrollY ? 'scroll' ? undefined,
     }}
   >
     {props.children}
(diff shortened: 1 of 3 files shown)


4ian:
Can you add screenshots? The overflowY added will probably make the tabs scrollable too. That should not be the case. Instead, we should have the scroll only on the content:
image

You can see in the code that there are `ScrollView`:
image

They should be responsible for adding the overflow-Y. Maybe better to look at them why it's not working?

HarsimranVirk:
> The overflowY added will probably make the tabs scrollable too.
 
I made sure this doesn't happen.
Screenshot from 2020-02-29 22-26-34

> They should be responsible for adding the overflow-Y. Maybe better to look at them why it's not working?

Yes, they do add overflow-y:
WhatsApp Image 2020-02-29 at 10 38 06 PM

I figured out which div corresponds to which react component, so 1st div is created by the Column component and the 2nd is created by ScrollView. The 2nd div already had overflow-y, but it doesn't work. 1st div required either height or overflow-y to make it scroll (I tested this is devTools).

Also adding height to 1st div actually made it look prettier, since then there was only one scroll (but it didn't work in firefox): 
Screenshot from 2020-02-29 22-27-36

> Maybe better to look at them why it's not working?
Maybe it has something to do with the resizable nature of this component? (I read somewhere that overflow works only with predefined height, but that was a problem before flexbox existed, right?)

4ian:
Let's try to find something that works everywhere without this second scroll:
image

That looks very "web-ish" and not super professional :) Only the content should be able to scroll. Maybe there is both the `height` and something else to add for Firefox?

HarsimranVirk:
> Let's try to find something that works everywhere without this second scroll:

Removed the second scroll ^_^

> Maybe there is both the height and something else to add for Firefox?

max-height does the trick for all, chrome, firefox and electron app. (I still have no clue about safari though)

Firefox:
Screenshot from 2020-03-01 02-02-56

Chrome: 
Screenshot from 2020-03-01 02-00-40

Electron-linux:
Screenshot from 2020-03-01 02-12-09

Should I replace the overflow-Y with max-height? (It still has to be passed through Column, I don't understand why ScrollView doesn't work without this :/ )

jimil749:
Yup, passing max-height attribute to column does solve the issue for all the 3 platforms!

newIDE/app/src/UI/Grid.js
```diff
@@ -36,6 +36,7 @@ export const Column = props => (
       alignItems: props.alignItems || 'stretch',
       justifyContent: props.justifyContent,
       flex: props.expand ? 1 : undefined,
+      maxHeight: props.maxHeight ? props.maxHeight : undefined,
```
4ian: `Column` is a component used a lot for the layout, and as such the less flexibility is has in its prop the better. This ensure consistency in the interface, and avoid bloating it with tons of options. There should be only one way to construct an interface with `Column` (and `Line`).

As such I would favor having a prop called "useMaxHeight" that is setting `maxHeight` to 100% if true, or undefined otherwise.
This will prevent anyone from being tempted to use `maxHeight="95%"` to "add some margins" or some other anti pattern (the proper pattern being to use a `Line` if you need margins or to stack elements).

4ian:
Looks like a proper fix! Thanks @HarsimranVirk for investigating and @jimil749 for verifying :) Added a comment, then we should be good to go.

HarsimranVirk:
@4ian I guess I messed up that one snapshot, it was failing locally so I updated it. But it's failing in the semaphoreci while im writing this.

`Snapshot Summary
 › 1 snapshot failed from 1 test suite. Inspect your code changes or re-run jest with `-u` to update them.

Test Suites: 1 failed, 17 passed, 18 total
Tests:       1 failed, 63 passed, 64 total
Snapshots:   1 failed, 7 passed, 8 total
Time:        34.744s
Ran all test suites.

                                                                               
error Command failed with exit code 1.`

4ian:
>  I guess I messed up that one snapshot, it was failing locally so I updated it

Yep, must have been because of other changes you did locally (material-ui update).
Reverting it was the proper thing to do. 

Looks good to me know! Thanks for the investigation and fix :)

PR #1488, opened 2020-03-03
Fix renaming groups and comments

### LINKED ISSUE #1430: Group/Comment text input is hard to exit
When I edit the text of comments and groups, the textbox is as wide as the item itself, making it hard to exit text input.
Ideally, the textbox would be limited to the actual text, whether text input is enabled or not. 
image
Example: I left-click on a group, text input activates and I enter the text I want. Now let's say that I want to copy-paste the item: I try to disable text input by pressing return (not working), by clicking right below the text (not working), by clicking on the right side of the text (not working), further to the right (still not working), by [right-clicking](https://github.com/4ian/GDevelop/issues/1423) (not working), and only then, I either click on the left-side handle, or click on an event below/above, and it works. 
I find it quite inconvenient/unnatural.

And it's even more frustrating when you didn't mean to activate text input, i.e. I left-click on the group item to select it, and text input activates.

geetesh-gupta:
Fixes #1430 

Added the following keybindings to Group Event:
1. On "Escape", remove the focus
2. On "Enter", remove the focus // because new line is not needed here
3. On "Click", focus to edit
4. Ctrl + Z to undo recent changes while editing
5. Clicking anywhere outside: remove the focus

Added the following keybindings to Comment Event:
1. On "Escape", remove the focus
2. On "Enter", add a new line
3. On "Click", focus to edit
4. Ctrl + Z to undo recent changes while editing
5. Clicking anywhere outside: remove the focus

### DIFF
--- a/newIDE/app/src/EventsSheet/EventsTree/Renderers/CommentEvent.js
+++ b/newIDE/app/src/EventsSheet/EventsTree/Renderers/CommentEvent.js
@@ -42,6 +42,7 @@ const styles = {
 
 type State = {|
   editing: boolean,
+  prevValue: string,
 |};
 
 export default class CommentEvent extends React.Component<
@@ -50,7 +51,7 @@ export default class CommentEvent extends React.Component<
 > {
   state = {
     editing: false,
-    prevValue: undefined,
+    prevValue: '',
   };
 
   _selectable: ?HTMLSpanElement;
--- a/newIDE/app/src/EventsSheet/EventsTree/Renderers/CommentEvent.js
+++ b/newIDE/app/src/EventsSheet/EventsTree/Renderers/CommentEvent.js
@@ -1,54 +1,54 @@
 // @flow
-import { t } from '@lingui/macro';
+import { t } from "@lingui/macro";
 
-import * as React from 'react';
-import classNames from 'classnames';
-import TextField from '../../../UI/TextField';
-import { rgbToHex } from '../../../Utils/ColorTransformer';
+import * as React from "react";
+import classNames from "classnames";
+import TextField from "../../../UI/TextField";
+import { rgbToHex } from "../../../Utils/ColorTransformer";
 import {
   largeSelectedArea,
   largeSelectableArea,
   selectableArea,
-  disabledText,
-} from '../ClassNames';
-import { type EventRendererProps } from './EventRenderer';
+  disabledText
+} from "../ClassNames";
+import { type EventRendererProps } from "./EventRenderer";
+
 const gd = global.gd;
 
 const commentTextStyle = {
-  width: '100%',
-  fontSize: 14,
+  width: "100%",
+  fontSize: 14
 };
 
 const styles = {
   container: {
-    display: 'flex',
-    flexWrap: 'wrap',
+    display: "flex",
+    flexWrap: "wrap",
     padding: 5,
-    overflow: 'hidden',
-    minHeight: 35,
+    overflow: "hidden",
+    minHeight: 35
   },
   commentTextField: commentTextStyle,
   commentSpan: {
     ...commentTextStyle,
-    boxSizing: 'border-box',
-    alignItems: 'center',
-    height: '100%',
-    whiteSpace: 'pre-wrap',
+    boxSizing: "border-box",
+    alignItems: "center",
+    height: "100%",
+    whiteSpace: "pre-wrap",
     lineHeight: 1.5,
-    border: 1,
-  },
+    border: 1
+  }
 };
 
 type State = {|
   editing: boolean,
 |};
 
-export default class CommentEvent extends React.Component<
-  EventRendererProps,
-  State
-> {
+export default class CommentEvent extends React.Component<EventRendererProps,
+  State> {
   state = {
     editing: false,
+    prevValue: undefined
   };
 
   _selectable: ?HTMLSpanElement;
@@ -57,7 +57,7 @@ export default class CommentEvent extends React.Component<
   edit = () => {
     this.setState(
       {
-        editing: true,
+        editing: true
       },
       () => {
         if (this._textField) this._textField.focus();
@@ -78,7 +78,7 @@ export default class CommentEvent extends React.Component<
 
     this.setState(
       {
-        editing: false,
+        editing: false
       },
       () => this.props.onUpdate()
     );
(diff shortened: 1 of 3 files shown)


Bouh:
I don't know, i didn't really like the system for "Shift+Enter".
And Enter key for validate and unfocus seem to me no natural at all.
We can just click aware, change of tab or click on other instructions... the focus will be on the other element with the mouse ... or...

I'am ok with Escape, but for unfocus and keep change, not cancel it.
Keep Enter key like it works in current version for add return carriage/lines. (Shift+return is a fancy shortcuts and most of users didn't find the associations of keys for an functionnality. Except for common shortcuts in all software copy/paste/select all)

Why keep change when escape is press, because before you have made choice to edit the comment, it's your first choice, and it should be easy to continue your work (and in most of case you use a mouse/touch for select and edit something else), if you have change your mind and you think edit the comment is now useless you can use CTRL+Z, or CTRL+A and delete the selection, these shortcuts are more commons. I said that is the case of you add Escape for cancel and unfocus and Enter for validate and unfocus.  And Enter key is a common key for add lines/return carriage. If you break this classic shortcuts for replace by shift+enter user will be trouble.

You have talk about increase the margin of the comment for click on it and unfocus the edition but keep the focus on the comment event,

> Have text field width of 95%, to allow 5% of right side area empty so that when clicked, it will endEditing ---- these percentages are just tentative
[Comment source](https://github.com/4ian/GDevelop/issues/1430#issuecomment-592884813)

I think it's a good idea for select and unfocus with mouse.

geetesh-gupta:
Hi! @Bouh  I think it's kind of decision to make. Both have their pros and cons. In general, escape is used to cancel the current event. Suppose their is a case where user accidentally went for editing and wrote something not needed, then it will be easy to undo all by pressing escape. But, as you said, Ctrl+Z can also work. So, I think it's a collaborative decision to make, depending on the frequency of the function used.

For the case of Shift+Enter for carriage return, most desktop messaging apps are currently using Shift+enter to add carriage returns and Enter to send the message. So, I think it would become intuitive to use it. It will make the editing fully keyboard based which can increase the productivity while working.

Talking about the fact,
> I think it's a good idea for unfocus the text edition and keep focus on the event.

The only problem with providing a left out space on the right is that, that space remains empty so, if a comment is big enough, a lot of space will seem to empty as compared to others which will be a bit odd.

I would be happy to get a review on this @Bouh @4ian. It would be awesome to get a poll or something created to decide 🤔.

Bouh:
You have open the issue, @PascalLadalle, what you think ?

4ian:
I do think of the way we show comments and the edition as a "seamless" edition. It's a bit like if there was no edition mode at all, the comments were always showing the text and could be edited just by clicking and typing what you want. In this situation, we should probably have:

- Escape: remove the focus
- Clicking anywhere outside: remove the focus
- Enter: new line
- Cancelling can be done with Ctrl+Z while you're editing.

I'm also fine with the solution of "Enter" si removing the focus, meaning that:

- Escape: remove the focus
- Clicking anywhere outside: remove the focus
- Enter: remove the focus
- Shift+Enter: new line (as this is used in messaging apps)
- Cancelling can be done with Ctrl+Z while you're editing.

If we want to go with the easiest solution, I still think the first solution is simpler. 
In all cases, I don't think Escape should cancel your changes. You probably put a lot of thoughts in the comment. For a dialog it's ok to cancel changes because there is the relationship "Escape = pressing Cancel in a dialog". But for a comment that can be seamlessly edited in the Events Sheet, I prefer to avoid erasing the changes of the user (in general, we must avoid at all cost erasing anything the user made). Ctrl+Z can always be used.

geetesh-gupta:
Ok, so finally I will update the code based on the first solution and update the PR as it would be simpler as per user's experience.

newIDE/app/src/EventsSheet/EventsTree/Renderers/CommentEvent.js
```diff
@@ -12,6 +12,7 @@ import {
   disabledText,
 } from '../ClassNames';
 import { type EventRendererProps } from './EventRenderer';
+
```
4ian: nitpicking: useless line added

newIDE/app/src/EventsSheet/EventsTree/Renderers/CommentEvent.js
```diff
@@ -119,7 +120,9 @@ export default class CommentEvent extends React.Component<
           ...styles.container,
           backgroundColor: `#${backgroundColor}`,
         }}
-        onClick={this.edit}
+        onClick={() => {
```
4ian: nitpicking: avoid useless function

newIDE/app/src/EventsSheet/EventsTree/Renderers/CommentEvent.js
```diff
@@ -141,6 +144,11 @@ export default class CommentEvent extends React.Component<
             }}
             fullWidth
             id="comment-title"
+            onKeyUp={e => {
```
4ian: e => event

newIDE/app/src/EventsSheet/EventsTree/Renderers/GroupEvent.js
```diff
@@ -68,7 +70,9 @@ export default class GroupEvent extends React.Component<EventRendererProps, *> {
           ...styles.container,
           backgroundColor: `rgb(${r}, ${g}, ${b})`,
         }}
-        onClick={this.edit}
+        onClick={() => {
```
4ian: same

newIDE/app/src/EventsSheet/EventsTree/Renderers/GroupEvent.js
```diff
@@ -90,6 +94,16 @@ export default class GroupEvent extends React.Component<EventRendererProps, *> {
             }}
             fullWidth
             id="group-title"
+            onKeyUp={e => {
```
4ian: e => event

newIDE/app/src/EventsSheet/EventsTree/Renderers/GroupEvent.js
```diff
@@ -90,6 +94,16 @@ export default class GroupEvent extends React.Component<EventRendererProps, *> {
...
+              if (e.key === 'Escape') {
+                this.endEditing();
+              }
+            }}
+            onKeyPress={e => {
+              if (e.charCode === 13) {
```
4ian: Avoid magic numbers, either use key or keyCode

newIDE/app/src/UI/TextField.js
```diff
@@ -44,7 +44,11 @@ type Props = {|
       value: string,
     },
   }) => void,
-  onKeyPress?: (event: {| charCode: number, key: string |}) => void,
+  onKeyPress?: (event: {|
+    charCode: number,
+    key: string,
+    shiftKey: string,
```
4ian: These changes are not useful anymore

4ian:

Issue #1421, opened 2020-02-10
cordova-build shown in the UI could be named Android Export instead


Bouh:
## Describe the bug
When you select export Android for get an .APK, in build list we see an cordova-build in progress.

Should be called Android or apk export, for something associate with Android export.
I know you use cordova behind for build the apk and that's why wee see cordova but it can be a source of misunderstood.

## Reproduce
Click on
![image](https://user-images.githubusercontent.com/1670670/74184962-072d7f80-4c48-11ea-9b4f-951f97dd7c5b.png)

Wait and see in build list.
![image](https://user-images.githubusercontent.com/1670670/74184938-fda41780-4c47-11ea-8a94-68a27fe7b2d5.png)

## Other details
beta 88

Kalovelo:
👋 Hello! 👋
Could I work on this issue?

@4ian

Bouh:
Yes sure you can @Kalovelo, no need to ask, you're free to start what you want :)
Open source allow to work on what you want, so don't hesitate and when you have an possible fix open your own PR 🤓

4ian:
Still useful in case you want to do something quite large or want to make sure people are aware you're trying something. But yeah feel free to go ahead! :)

Kalovelo:
I updated the build title by adding "apk" into parenthesis ( result : "cordova-build(apk)" ).
I changed localhost to gdevelop-app-local but a CORS problem occurred so I had to test it using storybook.
Do you consider the new name change suitable?
I will proceed with the PR as soon as possible!

PR #1489, opened 2020-03-03
Update cordova build title

Kalovelo:
refs #1421 
Renamed the cordova-build type to cordova-build(apk).
Apk in parenthesis was chosen since cordova will probably be used for iOS too in the future so a similar naming could probably be considered suitable. Also, the build description contains the keyword "Android"  giving a straightforward explanation for new Game Developers or Users that .apk is the Android type


4ian:
Unfortunately this solution does not work because the type in GDevelop must be the same as the one sent to/returned by the server - which is `cordova-build`. What you want instead is change the UI so that these "names" are translated to something that is human friendly.

Kalovelo:
Thanks for your answer!
A solution I thought was by checking the build type on the specific component output but I didn't consider it the best idea having that "extra" check. I will proceed by implementing it since it is probably, the best one.

4ian:
This is fine because "cordova-build" is the internal name. We can have a component or even a function that is available to do the "human translation" :) And then reuse it anywhere the name is displayed on screen.

newIDE/app/src/Export/Builds/BuildsList.js
```diff
@@ -26,6 +26,19 @@ const styles = {
   },
 };
 
+const formatBuildText = buildType => {
+  switch (buildType) {
+    case 'cordova-build':
+      return 'Apk Build (Cordova)';
```
4ian: All strings in GDevelop must be translatable. Use <Trans> component

newIDE/app/src/Export/Builds/BuildsList.js
```diff
@@ -26,6 +26,19 @@ const styles = {
   },
 };
 
+const formatBuildText = buildType => {
```
4ian: Add Flow typing to buildType

Kalovelo:
Thank you for your comments.
Since Cordova will be used for iOS, another check will probably be needed in the future.
I pulled the latest changes from master and semaphoraci seems to be failing at piskel-editor.zip (internal server error) 🤔

newIDE/app/src/ObjectsList/index.js
```diff
@@ -295,15 +295,12 @@ export default class ObjectsList extends React.Component<Props, State> {
...
-
-    if (getObjectWithContextName(objectWithContext) === newName) {
-      return;
-    }
-
     this.setState({
       renamedObjectWithContext: null,
     });
 
+    if (getObjectWithContextName(objectWithContext) === newName) return;
```
4ian: Make sure to merge again master in your branch to avoid this file from appearing

4ian:
> semaphoraci seems to be failing at piskel-editor.zip (internal server error) 🤔

Seems like a transient issue, I've relaunched it.

> Since Cordova will be used for iOS, another check will probably be needed in the future

Yep you're right :) 

> I pulled the latest changes from master

Strange that another file is showing in this PR 🤔Can you retry? Otherwise we'll go ahead like this.

newIDE/app/src/Export/Builds/BuildsList.js
```diff
@@ -26,6 +26,21 @@ const styles = {
   },
 };
 
+const formatBuildText = (
+  buildType: 'cordova-build' | 'electron-build' | 'web-build'
+) => {
+  switch (buildType) {
+    case 'cordova-build':
+      return <Trans>Apk Build (Cordova)</Trans>;
```
4ian: Let's use the same terminology as in the export:

- `Android Build` 
- `Windows/macOS/Linux Build`
- `Web (upload online) Build`

(.apk is a "technical detail", might be an .aab later) (same for Cordova/Electron - it's true but less relevant for the user)

newIDE/app/src/ObjectsList/index.js
4ian: (if this does not solve the problem, you can either ignore this or close this PR and open another - I think it's a GitHub bug)

Kalovelo:
You were right, the second pull fixed it 🎉

4ian:
Looks good :) Merging this now, thanks for working on this!

PR #1490, opened 2020-03-03
Fixes tooltip issue which shows [object Object]

### LINKED ISSUE #1484: Tooltip for "Properties" and "Objects" panels is "[object Object]"
## Describe the bug
Tooltip for "Properties" and "Objects" panels is "[object Object]"

## To Reproduce
Open a project and view a scene.
Make sure the properties panel is open.
Mouse over the title "Properties"

GDevelop_5_SZoCntxooT

I can't take a screenshot with the tool tip at the same time.

## Other details
Windows 10 laptop.
beta88
desktop version

aman-dave:
Fixes #1484 

The bug: It showed [object Object] on the toolbar and on hovering.
IMG_8725

Solution: Now, it shows the actual title
IMG_8727
PS: Photo was taken via phone as it was not coming during the screenshot.

After understanding the folder structure, I found that various editors had their individual folder where they were creating their properties and it was finally called in UI.

It was a small change in `src/UI/EditorMosaic/index.js`

### DIFF
--- a/newIDE/app/src/UI/EditorMosaic/index.js
+++ b/newIDE/app/src/UI/EditorMosaic/index.js
@@ -225,7 +225,7 @@ export default class EditorMosaic extends React.Component<Props, State> {
               return (
                 <MosaicWindow
                   path={path}
-                  title={editor.title}
+                  title={editor.title.props.id}
                   toolbarControls={editor.toolbarControls}
                 >
                   {editor.renderEditor()}


aman-dave:
@4ian I am looking into why this checks failed and it says:
```
Cannot get editor.title.props because:
 • property props is missing in $Iterable [1].
 • property props is missing in Boolean [1].
 • property props is missing in Number [1].
 • property props is missing in React.Portal [1].
 • property props is missing in String [1].
 • property props is missing in null [1].
 • property props is missing in undefined [1].
```
But when I console log `editor.title` it shows: {props:{id: " __the_passed_title_as_props__"}}

4ian:
Does this work with other language? I don't think it will, because this solution is showing the id of the message, which is an english only message - and that should basically never be used.

Accessing the props of a child is also in 99.9% of cases an anti-pattern, as it breaks encapsulation and can silently break the code - never do it :)

The issue is that the tooltip can only be a `string`, while `title` prop is a component `<Trans>`.
Three solutions:
1. Remove the tooltip :) Really, not sure it's helpful, it's not showing anything more than the title.
2. Or change the `title` prop so that we're passing a text written with the `t` macro:
  ```
  title={t`My editor` }
  ```

 and use `i18n._` to translate it.
3. Or (the best solution I think): do what I just said in 2, but do it for another property called... "tooltip" :) 

  ```
  title={<Trans>My editor</Trans> }
  tooltip={t`This editor does that, that, that, blabla` }
  ```

  EditorMosaic will use i18n to call i18n._ on the prop.

(You may ask, why do we have `t` and `Trans`. In most cases you can use `Trans`. While it's very flexible, it's a React component so works for things shown on screen, but anything shown by the system (menu, tooltip) requires a *string*, which can only be obtained using i18n._:
```js
 i18n._(t`The string to translate`)
```
)

Search in the codebase to see examples :)

PR #1491, opened 2020-03-03
fix yarn command type clips empty space between two text types

blurymind:
This fixes a bug that was brought up at the forum with yarn clipping empty space between two text results if you put a command between them

example:
`my <<command>>text ` resulting in `mytext`

with this fix
`my <<command>text `results in `my text` (as expected)

bug reported here (among other places)
https://forum.gdevelop-app.com/t/yarn-lines-of-text-merging-together-when-using-commands-on-the-yarn-node/21452/6

I'm surprised how many people are using this extension, but not so surprised they are reporting the bugs at the forum instead of here. Perhaps we should soon remove the experimental status? I've ironed out all the bugs they reported :)
Please let me know if I missed something from the forum

Maybe I should give yarn editor some love once I have the tilemap extension ready

### DIFF
--- a/Extensions/DialogueTree/dialoguetools.js
+++ b/Extensions/DialogueTree/dialoguetools.js
@@ -472,7 +472,8 @@ gdjs.dialogueTree.goToNextDialogueLine = function() {
       this.dialogueText = this.dialogueData.text;
       this.commandCalls = [];
     } else {
-      this.dialogueText += (this.dialogueText === '' ? '' : ' ') + this.dialogueData.text;
+      this.dialogueText +=
+        (this.dialogueText === '' ? '' : ' ') + this.dialogueData.text;
     }
 
     this.dialogueDataType = 'text';
@@ -482,7 +483,7 @@ gdjs.dialogueTree.goToNextDialogueLine = function() {
     this.dialogueData = this.dialogue.next().value;
   } else if (gdjs.dialogueTree._isLineTypeOptions()) {
     this.dialogueDataType = 'options';
-    this.dialogueText = ' ';
+    this.dialogueText = '';
     this.clipTextEnd = 0;
     this.optionsCount = this.dialogueData.options.length;
     this.options = this.dialogueData.options;
--- a/Extensions/DialogueTree/dialoguetools.js
+++ b/Extensions/DialogueTree/dialoguetools.js
@@ -472,7 +472,7 @@ gdjs.dialogueTree.goToNextDialogueLine = function() {
       this.dialogueText = this.dialogueData.text;
       this.commandCalls = [];
     } else {
-      this.dialogueText += this.dialogueData.text;
+      this.dialogueText += ' ' + this.dialogueData.text;
     }
 
     this.dialogueDataType = 'text';
@@ -489,7 +489,6 @@ gdjs.dialogueTree.goToNextDialogueLine = function() {
     this.selectedOptionUpdated = true;
   } else if (gdjs.dialogueTree._isLineTypeCommand()) {
     this.dialogueDataType = 'command';
-    this.dialogueText += ' ';
     this.clipTextEnd = 0;
 
     var command = this.dialogueData.text.split(' ');
--- a/Extensions/DialogueTree/dialoguetools.js
+++ b/Extensions/DialogueTree/dialoguetools.js
@@ -472,7 +472,7 @@ gdjs.dialogueTree.goToNextDialogueLine = function() {
       this.dialogueText = this.dialogueData.text;
       this.commandCalls = [];
     } else {
-      this.dialogueText += this.dialogueData.text;
+      this.dialogueText += ' ' + this.dialogueData.text;
     }
 
     this.dialogueDataType = 'text';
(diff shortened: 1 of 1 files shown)


Extensions/DialogueTree/dialoguetools.js
```diff
@@ -472,7 +472,7 @@ gdjs.dialogueTree.goToNextDialogueLine = function() {
       this.dialogueText = this.dialogueData.text;
       this.commandCalls = [];
     } else {
-      this.dialogueText += this.dialogueData.text;
+      this.dialogueText += ' ' + this.dialogueData.text;
```
4ian: What should happen in these cases:

* `my<<command>>text` => `mytext` I guess?
* `my <<command>> text` => ??
* `<<command>>text` => ?? (notice the command is the very first thing.

It seems that your solution is giving `my text` in the second case (which is fine?) (but please check ;)), but ` text` in the third case, while this should be `text`?

blurymind: `my <<command>> text` => `my  text ` two->two spaces as expected

but if I do `my<<command>>text` , it adds a space but I kind of like the fact that it adds the space for the user if they missed it

blurymind: I need to look into this more carefully :) 
Ideally it should not add or remove a space, but keep everything. Its interesting that it does it only to spaces before or after a command or an option type

blurymind: Ok I updated the PR to treat the case with continuing the text after a choice is taken and the node after starts with a command. It wasnt restaring the scrolling

4ian: Yeah not a big fan of adding a space with commands 😬
What about?
```suggestion
      this.dialogueText += (this.dialogueText === '' ? '' : ' ') + this.dialogueData.text;
```

i.e: add a space only if the text is not empty? (not tested)

blurymind: nah, it still adds spaces. Let me see if I can better fix this

blurymind: @4ian I am now convinced that this is a bug in bondagejs. Some regular expression in there is trimming spaces.
Your suggested fix makes the alleviates the problem considerably and makes it appear very rarely, but doesnt completely fix it.

I suggest to use this workaround fix for now, at least I manage to get a propper one merged in bondagejs

Extensions/DialogueTree/dialoguetools.js
```diff
@@ -482,11 +482,14 @@ gdjs.dialogueTree.goToNextDialogueLine = function() {
     this.dialogueData = this.dialogue.next().value;
   } else if (gdjs.dialogueTree._isLineTypeOptions()) {
     this.dialogueDataType = 'options';
+    this.dialogueText = ' ';
```
4ian: @blurymind 
Alright, just double checking that you want this line too? Will merge then when I get your go/change.

blurymind: yes I do, but actually doesnt need that extra space in it. I will update it sorry.

Will do a pr to bondagejs soon. I think I know where this happens

blurymind: updated now. Thank you for reviewing this :)

4ian: 👍

TerraPrimal:
Looks like this was pushed into the beta 90 build but I am still experiencing the issue that was originally reported [here](https://forum.gdevelop-app.com/t/yarn-lines-of-text-merging-together-when-using-commands-on-the-yarn-node/21452/10).

When using any commands, it just auto continues all existing dialogue in that one event. Please let me know if you need to be create quick example of this in the latest build.

blurymind:
no probs, please attach an example

TerraPrimal:
> no probs, please attach an example

Simple project with 4 nodes, 2 commands using BB and Normal Text.
[Lets Have A Yarn.zip](https://github.com/4ian/GDevelop/files/4474900/Lets.Have.A.Yarn.zip)

blurymind:
> > no probs, please attach an example
> 
> Simple project with 4 nodes, 2 commands using BB and Normal Text.
> [Lets Have A Yarn.zip](https://github.com/4ian/GDevelop/files/4474900/Lets.Have.A.Yarn.zip)

yikes, I see what you mean now. When you navigate to a node that starts with a command, it sort of continutes :) 
I will try to fix this asap, just surprised how many corner cases you guys find :) I need to update the demo to include them so we can catch regressions.

A lot of people seem to be writing stories with it now, which is fantastic. There are some fixes to yarn editor coming soon too. Another dev has been keeping me busy reviewing prs there too

blurymind:
This fix actually broke a lot of stuff so I am rolling an urgent PR to clean up the code, fix all regressions and add better regression testing tomorrow.

Solving the commands on a  new line issue requires a fix to bondagejs which I can do as another PR.

You see bondagejs removes all new lines and gives no information to the wrapper what line you are on, so we have no way of detecting if a command is on the same line as text or on a new line after it atm.

I opened an issue here
https://github.com/hylyh/bondage.js/issues/62

blurymind:
@TerraPrimal thanks to the example you attached I was able to fix the bug. Can you re-test after #1671 gets merged

PR #1494, opened 2020-03-04
Save/load last opened folder

### LINKED ISSUE #979: Default directory for opening a file
## Description/Reproduction
Let's say you have a project opened and you want to add a sound. You select a sound action, click on "Choose a new audio file", and an explorer pops up.
This explorer always opens to the project's base directory. If you have a dozen files to load from the same folder, it's quite annoying to browse a dozen times to your sound folder. To me, it's "Non optimal behaviour", but I can understand if it's meant to work that way.
Where it gets really illogical, now: if you close GD, open it again and choose to open a project, then it shows up the most recently used folder (that sounds folder, in my example, instead of the base folder of the last opened project, which would be the logical choice).
Sorry for the not-so-concise description 😬 
Not a very interesting/troublesome bug, but it annoys me every day.

## Other details
Win 10 / GD5b65

geetesh-gupta:
Fixes #979 
Following changes are made to the code:
- `PreferencesContext.js` is updated to include `lastOpenedFolder` variable
- `LocalResourceSources.js` is updated to modify/access the `lastOpenedFolder` path


newIDE/app/src/ResourcesList/LocalResourceSources.js
```diff
@@ -192,6 +197,50 @@ export default [
   },
 ];
 
+// function _loadValuesFromLocalStorage(): ?PreferencesValues {
```
4ian: Unfortunately this is not scalable, as this code is redundant with the implementation already existing in PreferencesProvider. We can't allow ourselves do put such code in every file that would need the preferences. Imagine the nightmare when you want to change the storage.
The editor should also not assume that things are written in the local storage... it might be saved in memory, on disk, or even some other exotic stuff in the future. 
That's why we have a `PreferencesContext` that React components can use to get the method to load or save a preference.
Using this context, they can manipulate the preferences - and they don't care how it's done.

Here we're not in a file using React though - but can you think about a way to use the methods from PreferencesContext? :)

4ian:
I wonder if we should store this in the preferences at all.. could be a per project preference at least, maybe?

newIDE/app/src/ResourcesList/LocalResourceSources.js
geetesh-gupta: Ok, I will try to think over it. Maybe I can create load/save functions in PreferencesContext and export them and use them in both PreferencesProvider and other files needed. 🤔

4ian: That's one solution, though ideally all your functions would come from the context itself - so that if later we use a second PreferencesProvider (that could happen, for example, a set of preference per user, etc...), the good one is automatically picked up by the context.
See if you can pass the functions from a React component to the load/save functions.

geetesh-gupta: I have updated the PR, please review it.

newIDE/app/src/ResourcesList/LocalResourceSources.js
```diff
@@ -4,10 +4,15 @@ import { type I18n as I18nType } from '@lingui/core';
...
 import { Component } from 'react';
 import {
-  isPathInProjectFolder,
   copyAllToProjectFolder,
+  isPathInProjectFolder,
 } from './ResourceUtils.js';
+import {
+  loadPreferencesValues,
+  savePreferencesValues,
+} from '../MainFrame/Preferences/PreferencesContext';
```
4ian: We don't really want this file to rely on preferences. Everything should be provided by the callers.

newIDE/app/src/ResourcesList/LocalResourceSources.js
```diff
@@ -207,7 +224,13 @@ const selectLocalResourcePath = (
 
     const properties = ['openFile'];
     if (options.multiSelections) properties.push('multiSelections');
-    const projectPath = path.dirname(project.getProjectFile());
+    let projectPath = path.dirname(project.getProjectFile());
+
+    // Load locally stored lastOpenedPath and update projectPath if not undefined
+    let lastOpenedPath = _loadLastOpenedPath();
```
4ian: This is loading from the disk (localStorage) everytime we call this. We should avoid this.

4ian:
So this is better, but we're still breaking the encapsulation of the preferences.
The idea is that we want the PreferencesProvider to be replaceable. I must be able to switch `PreferencesProvider` by another one (that for example save the preference to a local json file, or in a database) and automatically everything will be using it. I could even use multiple `PreferencesProvider` in my app :) One part of the app could be saving preferences somewhere with one `PreferencesProvider`, and another one somewhere else.

It's important to keep this flexibility. Can you think of a solution where you don't have to explictly call loadPreferencesValues/savePreferencesValues? :)

newIDE/app/src/ResourcesList/LocalResourceSources.js
```diff
@@ -192,6 +197,18 @@ export default [
...
+  const values = loadPreferencesValues();
+  if (values) return values.lastOpenedPath;
+  else return '';
+}
+
+function _saveLastOpenedPath(lastOpenedPath: string) {
+  let values = loadPreferencesValues();
+  values = { ...values, lastOpenedPath };
+  savePreferencesValues(values);
+}
```
4ian: Can we avoid these functions? :)

4ian:
See, here:
image

We got given `i18n`. This is very useful as this allow the components that are implementing the `ResourceSources` to use `i18n` functions... Can we do the same and get the preferences? :) 

By the way, who is even passing this i18n property you may ask?... Look at `MainFrame` ;) In particular:

image

newIDE/app/src/ResourcesList/LocalResourceSources.js
geetesh-gupta: I have updated it.

newIDE/app/src/ResourcesList/LocalResourceSources.js
geetesh-gupta: Oh Yes. We don't need these functions. I will remove them. :)

geetesh-gupta:
Hi, @4ian I have figured it out the way to pass the functions as a prop and use them in `LocalResourceSources` and maintaining encapsulation of preferences by creating the functions in `PreferencesProvider`, hence allowing to use different `PreferencesProvider` if needed. I have updated the PR, please review it.

geetesh-gupta:
> I wonder if we should store this in the preferences at all.. could be a per project preference at least, maybe? (more complicated though, so give it a thought later)

Currently, I am storing the `lastOpenedPath` in `Preferences` with a string type. 

But, alternatively what we can do, is to have `lastOpenedPath` as an object with the project's name as key and its preferences as values. Each project's preferences can further be broken down based on paths for different kinds of content (e.g. project folder, audio, image, etc.). The `currentProject` key can be passed on by `MainFrame` and corresponding kind can be selected with `LocalResourcecsSources`.

What are your opinions about it?

newIDE/app/src/MainFrame/Providers.js
```diff
@@ -77,6 +80,8 @@ export default class Providers extends React.Component<Props, {||}> {
                                   children({
                                     i18n,
                                     eventsFunctionsExtensionsState,
+                                    loadPreferencesValues,
+                                    savePreferencesValues,
```
4ian: MainFrame is not using these directly, better use a context consumer when needed inside MainFrame render

newIDE/app/src/MainFrame/index.js
```diff
@@ -1895,6 +1899,8 @@ class MainFrame extends React.Component<Props, State> {
                   (this._resourceSourceDialogs[resourceSource.name] = dialog)
                 }
                 i18n={i18n}
+                loadPreferencesValues={loadPreferencesValues}
```
4ian: Better wrap this in a `PreferencesContext.Consumer` to avoid having polluting MainFrame props.

newIDE/app/src/ResourcesList/ResourceSource.flow.js
```diff
@@ -1,13 +1,16 @@
...
+import type { PreferencesValues } from '../MainFrame/Preferences/PreferencesContext';
 
 // These are all the kind of resources that can be found in
 // Core/GDCore/Project/ResourcesManager.h
 export type ResourceKind = 'image' | 'audio' | 'font' | 'video' | 'json';
 
 export type ResourceSourceComponentProps = {|
   i18n: I18nType,
+  loadPreferencesValues: () => ?PreferencesValues,
+  savePreferencesValues: (values: PreferencesValues) => void,
```
4ian: Instead of passing the whole functions that are used to save and load preferences, could you send the preference values and a function to update the path?

=> the idea is to give to a component *nothing more* than what is needed for it to work. The components don't need to be able to change the entire preferences. It's even dangerous - they could erase everything :) Instead, let's see if we can just pass them a function to update the latest path. 

See for example how it's done for the language:
https://github.com/4ian/GDevelop/blob/a315eabdcedb5f04b7bf287272fa46e7b43f924f/newIDE/app/src/MainFrame/Preferences/LanguageDialog.js#L74-L75

we're getting the values (no problem if you want to read them, they are not secret) and *only* a function to set the language: `setLanguage`

Let's do the same here :)

newIDE/app/src/ResourcesList/LocalResourceSources.js
```diff
@@ -220,6 +283,19 @@ const selectLocalResourcePath = (
...
+          if (values) {
+            if (values.lastOpenedPath[projectName])
+              values.lastOpenedPath[projectName][kind] = paths[0];
+            else {
+              let path = {};
+              path[kind] = paths[0];
+              values.lastOpenedPath[projectName] = path;
+            }
+            savePreferencesValues(values);
+          }
```
4ian: All of this should not be the responsibility of LocalResourceSources :)
It should be a single function, something like `saveLatestPath(kind, paths[0])`

LocalResourceSources job is to show you dialogs to open files. We want to add it the ability to remember the latest opened path.
For this, we'll pass to it a function to read the last opened path, and another to save it - nothing more. These values and function are the responsibility of the parent to set... and we'll do it using a React context :)

newIDE/app/src/MainFrame/Providers.js
geetesh-gupta: Oh :o! that would be a better approach. I will update it :)

newIDE/app/src/MainFrame/index.js
geetesh-gupta: Yes, that's better. I will update it.

newIDE/app/src/ResourcesList/LocalResourceSources.js
geetesh-gupta: Just one confusion is their a way in `PreferencesProvider` to get the project name or I need to pass it with function `saveLatestPath` ?

geetesh-gupta: I am thinking that the project name should be passed as an argument to `saveLatestPath` because if we have multiple projects open in future, then Provider needs to know, which project we are talking about. :)

newIDE/app/src/MainFrame/index.js
geetesh-gupta: I have updated the PR with your suggestions, please review it :)

newIDE/app/src/ResourcesList/LocalResourceSources.js
4ian: That seems fair to me to pass the project name :) Or maybe the project file path?

newIDE/app/src/MainFrame/Preferences/PreferencesProvider.js
```diff
@@ -268,17 +270,45 @@ export default class PreferencesProvider extends React.Component<Props, State> {
...
+      localStorage.setItem(LocalStorageItem, JSON.stringify(values));
     } catch (e) {
       console.warn('Unable to persist preferences', e);
     }
+  }
+
+  _loadLatestPath(project: gdProject, kind: ResourceKind) {
+    try {
+      const projectName = project.getName();
+      const values = this._loadPreferencesValues();
```
4ian: You don't need to load the preferences. They are already in this.state

newIDE/app/src/MainFrame/Preferences/PreferencesProvider.js
```diff
@@ -268,17 +270,45 @@ export default class PreferencesProvider extends React.Component<Props, State> {
...
+          return curProjectPaths[kind];
+        }
+      }
+    } catch (e) {
+      console.warn('Unable to load latest path', e);
+    }
+  }
 
-    return preferences;
+  _saveLatestPath(project: gdProject, kind: ResourceKind, path: string) {
```
4ian: Passing `project: gdProject` and `kind: ResourceKind` seems a great idea 👍
This allow to encapsulate in this function how we save things internally.

newIDE/app/src/MainFrame/index.js
```diff
@@ -1889,13 +1889,23 @@ class MainFrame extends React.Component<Props, State> {
...
+                  return (
+                    <Component
+                      key={resourceSource.name}
+                      ref={dialog =>
+                        (this._resourceSourceDialogs[
+                          resourceSource.name
+                        ] = dialog)
+                      }
+                      i18n={i18n}
+                      loadLatestPath={loadLatestPath}
```
4ian: Name this function `getLatestPath` or `getLastUsedPath`

(load => get, because there is no loading necessary)

newIDE/app/src/ResourcesList/LocalResourceSources.js
```diff
@@ -221,6 +260,11 @@ const selectLocalResourcePath = (
       paths => {
         if (!paths) return resolve([]);
 
+        // Update stored latestPath value
+        if (paths[0] !== projectPath) {
```
4ian: Don't do this check, let's always call saveLatestPath - it's the responsibility of this function to do the saving if needed or not. Here we don't care.
We're just saying "Ok we have a new path that was chosen, for this project and this kind of resource. So please save it, I'm giving these information to you - then you do whatever you want with this - not my problem!!" ;)

newIDE/app/src/ResourcesList/LocalResourceSources.js
```diff
@@ -200,14 +226,27 @@ const selectLocalResourcePath = (
...
+  ) => void,
+  kind: ResourceKind
 ): Promise<Array<string>> => {
   return new Promise((resolve, reject) => {
     if (!dialog) return reject('Not supported');
 
     const properties = ['openFile'];
     if (options.multiSelections) properties.push('multiSelections');
-    const projectPath = path.dirname(project.getProjectFile());
+    let projectPath = path.dirname(project.getProjectFile());
```
4ian: Use const as much as possible

newIDE/app/src/ResourcesList/LocalResourceSources.js
```diff
@@ -200,14 +226,27 @@ const selectLocalResourcePath = (
...
+
+    // Load latestPath and update projectPath if not undefined
+    const latestPath = loadLatestPath(project, kind);
```
4ian: Don't add a comment for this, because you can rewrite a line like this in a way it's almost like English:

```suggestion
    const latestPath = getLastUsedPath(project, kind) || projectPath;
```

"The latest path is the last used path, for this project and kind, OR if not defined, it's the project path".

Clear and concise :)

newIDE/app/src/MainFrame/Preferences/PreferencesProvider.js
```diff
@@ -268,17 +270,45 @@ export default class PreferencesProvider extends React.Component<Props, State> {
...
+    try {
+      const projectName = project.getName();
+      let values = this._loadPreferencesValues();
```
4ian: Don't load the preferences. They are already in this.state

Instead, you must update the state - that's why we're using a React context.

Saving the preferences is just a side effect - but we could even not do this, or postpone it to save later, or by a worker, or in 2 hours... the most important is to update the state. See how it's done in other `_setXXX` functions, like `_setAutoDownloadUpdates `

newIDE/app/src/MainFrame/Preferences/PreferencesProvider.js
```diff
@@ -268,17 +270,45 @@ export default class PreferencesProvider extends React.Component<Props, State> {
     }
   }
 
-  _persistValuesToLocalStorage(preferences: Preferences) {
+  _savePreferencesValues(values: PreferencesValues) {
```
4ian: I like the previous name better :)

newIDE/app/src/MainFrame/Preferences/PreferencesProvider.js
```diff
@@ -239,11 +241,11 @@ export default class PreferencesProvider extends React.Component<Props, State> {
           },
         },
       }),
-      () => this._persistValuesToLocalStorage(this.state)
+      () => this._savePreferencesValues(this.state.values)
     );
   }
 
-  _loadValuesFromLocalStorage(): ?PreferencesValues {
+  _loadPreferencesValues(): ?PreferencesValues {
```
4ian: Let's go back to the previous name - making clear this provider is tight to localStorage.

newIDE/app/src/MainFrame/index.js
```diff
@@ -1889,13 +1889,23 @@ class MainFrame extends React.Component<Props, State> {
...
+                      saveLatestPath={saveLatestPath}
```
4ian: Name this function setLastUsedPath (or setLatestPath)

The fact that it's saving the preferences is "none of your business" at the level of this component. You just want the component to be able to say "here is the latest path for the project and this resource kind".
Then what you do, is up to you (you = the context).

newIDE/app/src/ResourcesList/LocalResourceSources.js
geetesh-gupta: I will keep that in mind :)

newIDE/app/src/MainFrame/Preferences/PreferencesProvider.js
geetesh-gupta: Oh! I forgot the fact that the state contains the values. I will update it accordingly.

newIDE/app/src/MainFrame/Preferences/PreferencesProvider.js
```diff
@@ -268,17 +270,52 @@ export default class PreferencesProvider extends React.Component<Props, State> {
...
-        JSON.stringify(preferences.values)
-      );
+      localStorage.setItem(LocalStorageItem, JSON.stringify(values));
     } catch (e) {
       console.warn('Unable to persist preferences', e);
     }
+  }
+
+  _getLastUsedPath(project: gdProject, kind: ResourceKind) {
+    try {
```
4ian: We don't expect any exception, you can remove the try catch.
Add explicitly the return value in all cases, including when the path is not found.

newIDE/app/src/MainFrame/Preferences/PreferencesProvider.js
```diff
@@ -268,17 +270,52 @@ export default class PreferencesProvider extends React.Component<Props, State> {
...
+      const projectName = project.getName();
+      const values = this.state.values;
+      if (values) {
```
4ian: values is always defined

newIDE/app/src/MainFrame/Preferences/PreferencesProvider.js
```diff
@@ -268,17 +270,52 @@ export default class PreferencesProvider extends React.Component<Props, State> {
...
+          return curProjectPaths[kind];
+        }
+      }
+    } catch (e) {
+      console.warn('Unable to load latest path', e);
+    }
+  }
+
+  _setLastUsedPath(project: gdProject, kind: ResourceKind, latestPath: string) {
+    try {
```
4ian: No need for exception. It's risky to add exception handling where you don't actually need it, because it would swallow them and hide real issues in your code.

newIDE/app/src/MainFrame/Preferences/PreferencesProvider.js
```diff
@@ -268,17 +270,52 @@ export default class PreferencesProvider extends React.Component<Props, State> {
...
+      const projectName = project.getName();
+      const values = this.state.values;
+      if (values) {
```
4ian: values is always define

newIDE/app/src/MainFrame/Preferences/PreferencesProvider.js
```diff
@@ -268,17 +270,52 @@ export default class PreferencesProvider extends React.Component<Props, State> {
...
+        else {
+          const path: { [ResourceKind]: string } = {};
+          path[kind] = latestPath;
+          values.lastUsedPath[projectName] = path;
+        }
 
-    return preferences;
+        this.setState(
+          {
+            values: values,
```
4ian: Try to use the spread operator (...) to make the whole function to be almost a single setState :)

newIDE/app/src/MainFrame/index.js
```diff
@@ -92,8 +92,8 @@ import OpenFromStorageProviderDialog from '../ProjectsStorage/OpenFromStoragePro
 import SaveToStorageProviderDialog from '../ProjectsStorage/SaveToStorageProviderDialog';
 import OpenConfirmDialog from '../ProjectsStorage/OpenConfirmDialog';
 import verifyProjectContent from '../ProjectsStorage/ProjectContentChecker';
-const GD_STARTUP_TIMES = global.GD_STARTUP_TIMES || [];
 
+const GD_STARTUP_TIMES = global.GD_STARTUP_TIMES || [];
```
4ian: Avoid having unrelated or useless changes in your PR

newIDE/app/src/MainFrame/index.js
```diff
@@ -1703,7 +1703,6 @@ class MainFrame extends React.Component<Props, State> {
       this.state.loadingProject ||
       this.state.previewLoading ||
       this.props.loading;
-
```
4ian: Same, re-read every line and cancel all changes not needed to have a PR as minimal as possible

newIDE/app/src/MainFrame/Preferences/PreferencesProvider.js
```diff
@@ -268,17 +270,52 @@ export default class PreferencesProvider extends React.Component<Props, State> {
...
+  _getLastUsedPath(project: gdProject, kind: ResourceKind) {
+    try {
+      const projectName = project.getName();
+      const values = this.state.values;
+      if (values) {
+        const curProjectPaths = values.lastUsedPath[projectName];
+        if (curProjectPaths) {
+          return curProjectPaths[kind];
+        }
+      }
```
4ian: You're returning undefined, but you declared elsewhere that your function is returning a `string`. You must make a conscious choice and stick to it to avoid crashes/bugs.

newIDE/app/src/MainFrame/Preferences/PreferencesProvider.js
```diff
@@ -58,7 +60,7 @@ export default class PreferencesProvider extends React.Component<Props, State> {
           language,
         },
       }),
-      () => this._persistValuesToLocalStorage(this.state)
+      () => this._persistValuesToLocalStorage(this.state.values)
```
4ian: I would recommend to cancel these changes to reduce the PR size if they are not needed

newIDE/app/src/ResourcesList/LocalResourceSources.js
```diff
@@ -209,18 +242,22 @@ const selectLocalResourcePath = (
...
         title: options.title,
         properties,
         filters: [{ name: options.name, extensions: options.extensions }],
-        defaultPath: projectPath,
+        defaultPath: latestPath,
       },
       paths => {
         if (!paths) return resolve([]);
 
+        setLastUsedPath(project, kind, paths[0]);
```
4ian: Do you want to save the full path, or the directory name only?

newIDE/app/src/MainFrame/Preferences/PreferencesContext.js
```diff
@@ -109,6 +110,7 @@ export type PreferencesValues = {|
   useGDJSDevelopmentWatcher: boolean,
   eventsSheetUseAssignmentOperators: boolean,
   showEffectParameterNames: boolean,
+  lastUsedPath: Object,
```
4ian: Verify this typing

4ian:
We're on a good track :) I still added comments to clean stuff as much as possible.
Strive to get a PR as minimal as possible. Re-read your PRs and cancel all changes that are not required. Check the typings, remove exception handling if there is no exception, use ES6 operators to make things as minimal as possible :)

newIDE/app/src/MainFrame/index.js
geetesh-gupta: Oh! I will keep that in mind.

newIDE/app/src/MainFrame/Preferences/PreferencesContext.js
geetesh-gupta: @4ian Verify here, or in the code as a comment? Basically, it is storing the object of all projects opened by the used with each kind of resource's path.

newIDE/app/src/ResourcesList/LocalResourceSources.js
geetesh-gupta: @4ian Currently, I am saving full path.

newIDE/app/src/MainFrame/Preferences/PreferencesContext.js
4ian: `Object` means `any` in Flow, which is a weak typing not really helping to catch bugs.
See how to mark an object as a map here: https://flow.org/en/docs/types/objects/#toc-objects-as-maps

geetesh-gupta: Oh! I will update it based on that.

geetesh-gupta:
@4ian I am unable to figure out the reason for build failed for AppVeyor. Can you tell is it with my updates or due to commits by others?

newIDE/app/src/MainFrame/index.js
```diff
@@ -1889,13 +1889,25 @@ class MainFrame extends React.Component<Props, State> {
...
-                }
-                i18n={i18n}
-              />
+              <PreferencesContext.Consumer
+                key={`${resourceSource.name}-consumer`}
+              >
+                {({ getLastUsedPath, setLastUsedPath }) => {
+                  return (
+                    <Component
+                      key={resourceSource.name}
```
4ian: The key is not needed anymore here

newIDE/app/src/MainFrame/Preferences/PreferencesProvider.js
```diff
@@ -281,6 +283,37 @@ export default class PreferencesProvider extends React.Component<Props, State> {
     return preferences;
   }
 
+  _getLastUsedPath(project: gdProject, kind: ResourceKind) {
+    const projectName = project.getName();
+    const values = this.state.values;
```
4ian: nitpicking: `const { values } = this.state;`

newIDE/app/src/MainFrame/Preferences/PreferencesContext.js
```diff
@@ -109,6 +110,7 @@ export type PreferencesValues = {|
   useGDJSDevelopmentWatcher: boolean,
   eventsSheetUseAssignmentOperators: boolean,
   showEffectParameterNames: boolean,
+  lastUsedPath: { [string]: { [ResourceKind]: string } },
```
4ian: Can we name this: projectLastUsedPaths

To hint that the first key is "a project".

newIDE/app/src/MainFrame/Preferences/PreferencesProvider.js
```diff
@@ -281,6 +283,37 @@ export default class PreferencesProvider extends React.Component<Props, State> {
...
+    const curProjectPaths = values.lastUsedPath[projectName];
```
4ian: nitpicking: curProjectPaths => projectPaths

"cur" does not give much extra information.

newIDE/app/src/MainFrame/Preferences/PreferencesProvider.js
```diff
@@ -281,6 +283,37 @@ export default class PreferencesProvider extends React.Component<Props, State> {
...
+      path[kind] = latestPath;
+      values.lastUsedPath[projectName] = path;
+    }
+
+    this.setState(
+      {
+        values: {
+          ...values,
+        },
+      },
```
4ian: nitpicking:

```suggestion
    const { values } = this.state;
    const newProjectLastUsedPaths = values.projectLastUsedPaths[projectName] || {};
    newProjectLastUsedPaths[kind] = latestPath;

    this.setState(
      {
        values: {
          ...values,
          projectLastUsedPaths: {
            ...values.projectLastUsedPaths,
            [projectName]: newProjectLastUsedPaths,
          }
        },
      },
```

4ian:
Some tests are broken in master, let me fix them! Sorry about that :)

For this PR, we're making good progress! Last thing I think:
- should we store the project name, or the project path? Project name sounds "dangerous" because if you have multiple copies of your games, the paths will be mixed. Not a big deal, but still maybe using the project path will make things clearer?

newIDE/app/src/MainFrame/Preferences/PreferencesContext.js
geetesh-gupta: Oh! that's a good enhancement :)

newIDE/app/src/ResourcesList/LocalResourceSources.js
geetesh-gupta: Also, it may happen that the last path for any resource (eg. audio), maybe an altogether different directory or different drive partition. So saving only the directory name will not work.

4ian: I mean in the case of /myfolder/subfolder/myfile.txt, to save /myfolder/subfolder/
Which means on Windows, instead of C:\something\game.json, saving C:\something

after all, we don't care about which file you last used right, just which folder you used? :)

geetesh-gupta: Oh! I see. I will update it :)

newIDE/app/src/BrowserApp.js
```diff
@@ -41,7 +41,10 @@ export const create = (authentification: Authentification) => {
       eventsFunctionsExtensionWriter={null}
       eventsFunctionsExtensionOpener={null}
     >
-      {({ i18n, eventsFunctionsExtensionsState }) => (
+      {({
```
4ian: nitpicking: can you revert the changes to these lines where nothing was actually changed? Will speed up review for me :)

newIDE/app/src/LocalApp.js
```diff
@@ -88,7 +88,10 @@ export const create = (authentification: Authentification) => {
         eventsFunctionsExtensionWriter={LocalEventsFunctionsExtensionWriter}
         eventsFunctionsExtensionOpener={LocalEventsFunctionsExtensionOpener}
```
4ian: Same in this file, if no actual code changes, prefer to cancel them in your PR to avoid losing review time.

newIDE/app/src/MainFrame/Preferences/PreferencesProvider.js
```diff
@@ -281,6 +284,38 @@ export default class PreferencesProvider extends React.Component<Props, State> {
...
+    const { values } = this.state;
+    const projectPaths = values.projectLastUsedPaths[projectPath];
+    if (projectPaths && projectPaths[kind]) {
+      return projectPaths[kind];
+    }
+    return '';
+  }
+
+  _setLastUsedPath(project: gdProject, kind: ResourceKind, latestPath: string) {
+    const projectPath = path.dirname(project.getProjectFile());
```
4ian: path can be null (on the web-app, optionalRequire('path') will return null). Make your code robust against this in case this is called in the web-app (should not, but still):

`if (!path) return null;`

4ian: Actually, do we want to store the project directory? No! :) 
There might be more than one project in a folder (unusual, but can happen).

So we want to save the filename to the project... in other words: 
```suggestion
    const projectPath = project.getProjectFile();
```

4ian:
See if you can clean lines that are not actually changed, and change the project path to use the full filename to the project file.
Apart from this, we are I think ready to go with this :) 👍

newIDE/app/src/LocalApp.js
geetesh-gupta: Oh! I will remember in future to check for previous unwanted changes whenever a new update is made.

geetesh-gupta: These unwanted changes might have been due to my code editor's formatting settings. Will fix it for future,

geetesh-gupta:
I have updated the PR. I have removed the unwanted changes now. Some of the unwanted changes were due to the formatting settings in my editor. I will fix the settings in my editor.

newIDE/app/src/BrowserApp.js
```diff
@@ -99,4 +99,4 @@ export const create = (authentification: Authentification) => {
   );
 
   return app;
-};
+};
```
4ian: nitpicking: Configure your editor to add an line at the end of the files

4ian:
We're * this * close :)

newIDE/app/src/BrowserApp.js
geetesh-gupta: Oh! My mistake. While comparing files to remove unrelated changes, I was comparing the files from the master using online GitHub editor and my local files. In the online editor, it does not show the end line, so I thought end line is not there in the master, so I removed it. I'll fix it.

4ian:
Alright, let's go ahead with this :)
Thanks for your patience and your work on this! 👍 This will be very useful for users and will speed up their workflow 👍

PR #1495, opened 2020-03-04
Fix tooltips for editor titles

### LINKED ISSUE #1484: Tooltip for "Properties" and "Objects" panels is "[object Object]"

HarsimranVirk:
Resolves #1484 PR #1490 
@4ian I'm sorry I couldn't figure out how to push in the existing PR branch :( (I'm really not good at using git)

Okay so I've added a tooltip component and custom toolbar rendering function (renders with the same styling but has div with title of tooltip component) 

WhatsApp Image 2020-03-05 at 12 30 09 AM

We can change what to show in the tooltip later :)

### DIFF
--- a/newIDE/app/src/SceneEditor/index.js
+++ b/newIDE/app/src/SceneEditor/index.js
@@ -903,7 +903,7 @@ export default class SceneEditor extends React.Component<Props, State> {
       properties: {
         type: 'secondary',
         title: <Trans>Properties</Trans>,
-        tooltip: t`Properties of Instance`,
+        tooltip: t`Display the properties of the selected instances from the scene editor`,
         renderEditor: () => (
           <InstancePropertiesEditor
             project={project}
@@ -959,7 +959,7 @@ export default class SceneEditor extends React.Component<Props, State> {
       'objects-list': {
         type: 'secondary',
         title: <Trans>Objects</Trans>,
-        tooltip: t`List of Objects`,
+        tooltip: t`The list of objects that can be used in this scene`,
         toolbarControls: [
           <I18n key="tags">
             {({ i18n }) => (
@@ -1001,7 +1001,7 @@ export default class SceneEditor extends React.Component<Props, State> {
       'object-groups-list': {
         type: 'secondary',
         title: <Trans>Object Groups</Trans>,
-        tooltip: t`List of Object Groups`,
+        tooltip: t`The groups of objects that can be used in events of this scene`,
         renderEditor: () => (
           <ObjectGroupsList
             globalObjectGroups={project.getObjectGroups()}
--- a/newIDE/app/src/SceneEditor/index.js
+++ b/newIDE/app/src/SceneEditor/index.js
@@ -903,6 +903,7 @@ export default class SceneEditor extends React.Component<Props, State> {
       properties: {
         type: 'secondary',
         title: <Trans>Properties</Trans>,
+        tooltip: t`Properties of Instance`,
         renderEditor: () => (
           <InstancePropertiesEditor
             project={project}
@@ -958,6 +959,7 @@ export default class SceneEditor extends React.Component<Props, State> {
       'objects-list': {
         type: 'secondary',
         title: <Trans>Objects</Trans>,
+        tooltip: t`List of Objects`,
         toolbarControls: [
           <I18n key="tags">
             {({ i18n }) => (
@@ -999,6 +1001,7 @@ export default class SceneEditor extends React.Component<Props, State> {
       'object-groups-list': {
         type: 'secondary',
         title: <Trans>Object Groups</Trans>,
+        tooltip: t`List of Object Groups`,
         renderEditor: () => (
           <ObjectGroupsList
             globalObjectGroups={project.getObjectGroups()}
--- a/newIDE/app/src/SceneEditor/index.js
+++ b/newIDE/app/src/SceneEditor/index.js
@@ -902,8 +902,7 @@ export default class SceneEditor extends React.Component<Props, State> {
     const editors = {
       properties: {
         type: 'secondary',
-        title: <Trans>Properties</Trans>,
-        tooltip: t`Display the properties of the selected instances from the scene editor`,
+        title: t`Properties`,
         renderEditor: () => (
           <InstancePropertiesEditor
             project={project}
@@ -958,8 +957,7 @@ export default class SceneEditor extends React.Component<Props, State> {
       },
       'objects-list': {
         type: 'secondary',
-        title: <Trans>Objects</Trans>,
-        tooltip: t`The list of objects that can be used in this scene`,
+        title: t`Objects`,
         toolbarControls: [
           <I18n key="tags">
             {({ i18n }) => (
@@ -1000,8 +998,7 @@ export default class SceneEditor extends React.Component<Props, State> {
       },
       'object-groups-list': {
         type: 'secondary',
-        title: <Trans>Object Groups</Trans>,
-        tooltip: t`The groups of objects that can be used in events of this scene`,
+        title: t`Object Groups`,
         renderEditor: () => (
           <ObjectGroupsList
             globalObjectGroups={project.getObjectGroups()}
--- a/newIDE/app/src/UI/EditorMosaic/index.js
+++ b/newIDE/app/src/UI/EditorMosaic/index.js
@@ -111,25 +111,12 @@ const replaceNode = (
 
 const defaultToolbarControls = [<CloseButton key="close" />];
 
-// This is based on the implementation of the react-mosaic window
-// toolbar done here: https://github.com/nomcopter/react-mosaic/blob/b36984c34b0fa1e3f21a211578301933cea24992/src/MosaicWindow.tsx#L136-L183
 const renderMosaicWindowPreview = props => (
   <div className="mosaic-preview">
     <div className="mosaic-window-toolbar">
       <div className="mosaic-window-title">{props.title}</div>
     </div>
     <div className="mosaic-window-body" />
-  </div>
-);
-
-const renderMosaicWindowToolbar = props => (
-  <div className="mosaic-window-toolbar" style={{ width: '100%' }}>
-    <div className="mosaic-window-title" title={props.tooltip}>
-      {props.title}
-    </div>
-    <div className="mosaic-window-controls">
-      {props.toolbarControls || defaultToolbarControls}
-    </div>
   </div>
 );
 
@@ -141,8 +128,8 @@ const renderMosaicWindowToolbar = props => (
 const MosaicWindow = (props: any) => (
   <RMMosaicWindow
     {...props}
+    toolbarControls={props.toolbarControls || defaultToolbarControls}
     renderPreview={renderMosaicWindowPreview}
-    renderToolbar={renderMosaicWindowToolbar}
   />
 );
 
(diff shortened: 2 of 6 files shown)


newIDE/app/src/UI/EditorMosaic/index.js
```diff
@@ -17,6 +18,7 @@ export type Editor = {|
   renderEditor: () => React.Node,
   noTitleBar?: boolean,
   title?: React.Node,
+  tooltip?: React.Node,
```
4ian: Use MessageDescriptor for the type

newIDE/app/src/UI/EditorMosaic/index.js
```diff
@@ -7,6 +7,7 @@ import {
 } from 'react-mosaic-component';
 import CloseButton from './CloseButton';
 import ThemeConsumer from '../Theme/ThemeConsumer';
+import { I18n } from '@lingui/react';
```
4ian: nitpicking: Put this at top of file, just after `// @flow`

newIDE/app/src/UI/EditorMosaic/index.js
```diff
@@ -117,6 +119,17 @@ const renderMosaicWindowPreview = props => (
   </div>
 );
 
+const renderMosaicWindowToolbar = props => (
+  <div className="mosaic-window-toolbar" style={{ width: '100%' }}>
+    <div className="mosaic-window-title" title={props.tooltip}>
```
4ian: If tooltip is not defined, will this show undefined as tooltip? If yes, we should either not define title, or force a tooltip to be defined (by not making `tooltip` optional)

newIDE/app/src/UI/EditorMosaic/index.js
```diff
@@ -223,13 +236,18 @@ export default class EditorMosaic extends React.Component<Props, State> {
...
-                  toolbarControls={editor.toolbarControls}
-                >
-                  {editor.renderEditor()}
-                </MosaicWindow>
+                <I18n>
+                  {({ i18n }) => (
+                    <MosaicWindow
+                      path={path}
+                      title={editor.title}
+                      tooltip={i18n._(editor.tooltip)}
```
4ian: What if `tooltip` is not defined? You should either disallow this or handle this case.

newIDE/app/src/SceneEditor/index.js
```diff
@@ -903,6 +903,7 @@ export default class SceneEditor extends React.Component<Props, State> {
       properties: {
         type: 'secondary',
         title: <Trans>Properties</Trans>,
+        tooltip: t`Properties of Instance`,
```
4ian: Let's go for `Display the properties of the selected instances from the scene editor`.

newIDE/app/src/SceneEditor/index.js
```diff
@@ -999,6 +1001,7 @@ export default class SceneEditor extends React.Component<Props, State> {
       'object-groups-list': {
         type: 'secondary',
         title: <Trans>Object Groups</Trans>,
+        tooltip: t`List of Object Groups`,
```
4ian: `The groups of objects that can be used in events of this scene`?

newIDE/app/src/UI/EditorMosaic/index.js
```diff
@@ -117,6 +119,17 @@ const renderMosaicWindowPreview = props => (
   </div>
 );
 
+const renderMosaicWindowToolbar = props => (
+  <div className="mosaic-window-toolbar" style={{ width: '100%' }}>
```
4ian: Can you confirm that this is the implementation found in react-mosaic? :)

4ian: (want to be sure we're not forgetting anything)

ozata:
If you have sent more than one commits in the same bug you are fixing you can rebase your commits.
Assume you are on the branch fix-tooltips and you are ahead of master branch 2 commits.
git status will give you "tooltip_fix"
Then do this.

`git rebase -i HEAD~2
`
This will open an interactive shell that you can rebase your commits, you will see a screen like this:

```
pick First commit message
pick Second commit message

```
On top of the shell you will see this two lines with other lines below. You have to change this to

```
pick First commit message
squash Second commit message

```
Then you have to press esc and :wq and write and quit, after this a new shell will be opened and it will ask you to change your commit message because now you have 1 commit but 2 messages,
edit your message or put a "#" to the start of the line to make it a comment line.

Then again you have to do :wq if you are on vim text editor, (it is CTRL + X on Ubuntu) and exit. Now you have rebased your commits into 1 commit.
Now it is time to push to the origin repo:

`git push -f origin tooltip_fix `

> Resolves #1484 PR #1490
> @4ian I'm sorry I couldn't figure out how to push in the existing PR branch :( (I'm really not good at using git)
> 
> Okay so I've added a tooltip component and custom toolbar rendering function (renders with the same styling but has div with title of tooltip component)
> 
> WhatsApp Image 2020-03-05 at 12 30 09 AM
> 
> We can change what to show in the tooltip later :)

newIDE/app/src/UI/EditorMosaic/index.js
HarsimranVirk: I guess we are maybe adding just one extra div, (and I need it to wrap other divs, using React.Fragment throws an error). Though this is the implementation:
https://github.com/nomcopter/react-mosaic/blob/b36984c34b0fa1e3f21a211578301933cea24992/src/MosaicWindow.tsx#L136-L150
I don't really see any other difference, (especially in devTools, there's just one more div, which is used for wrapping) Is it fine, what do you suggest?

4ian: What's the error with React.Fragment?

HarsimranVirk: Error: Only native element nodes can now be passed to React DnD connectors.You can either wrap the component into a <div>, or turn it into a drag source or a drop target itself.

newIDE/app/src/UI/EditorMosaic/index.js
HarsimranVirk: `tooltip={editor.tooltip ? i18n._(editor.tooltip) : null}` ?

With tooltip
WhatsApp Image 2020-03-06 at 2 35 17 AM

Without tooltip
WhatsApp Image 2020-03-06 at 2 35 20 AM

newIDE/app/src/UI/EditorMosaic/index.js
4ian: Ok, last questions:

* You said "there's just one more div, which is used for wrapping) Is it fine, what do you suggest?"
=>  what is this extra div? 

* Can you add a comment with a link to MosaicWindow.tsx telling "this is based on the implementation of the react-mosaic window toolbar done here: [link]"

Thanks :)

HarsimranVirk: https://github.com/nomcopter/react-mosaic/blob/b36984c34b0fa1e3f21a211578301933cea24992/src/MosaicWindow.tsx#L145-L149 
You can see here that our custom toolbar is injected inside a `<div className="mosaic-window-toolbar">`.

Now, inside this div, there should be two divs, one for title and one for toolbar controllers. They also have to be siblings of each other. So to maintain their sibling relationship in the DOM tree I'm wrapping them inside a `<div className="mosaic-window-toolbar">` again. This is the extra div I'm talking about (which is not present in the actual implementation). So, the DOM tree would look like this,

`<div class="mosaic-window-toolbar">` => this is the parent div
`.     <div class="mosaic-window-toolbar">` => this is the additional wrapping div
`. .            <div class="mosaic-window-title"></div>` => this div is for the title
`. .            <div class="mosaic-window-controls"></div>` => this is for controls
`.      </div>`
`</div>`

Now visually the additional wrapping div overlays the entire area of the actual parent div since they have the similar styling and dimensions. Ideally, React.Fragment should've been used for wrapping instead of div, but it is not a react-dnd element. 

> Can you add a comment with a link to MosaicWindow.tsx telling "this is based on the implementation of the react-mosaic window toolbar done here: [link]"

Sure thing :)

blurymind:
Are you on a mac?
https://support.apple.com/en-gb/HT201361

HarsimranVirk:
> Are you on a mac? Check this out :)

No I use linux. Actually, the tooltip goes away when i try to take a screenshot. 😅😅
So that's why I had to click a picture of it. I should've mentioned it before, my bad.

blurymind:
On linux, depending on your desktop, you should have a tool for screenshots somewhere in your start menu :)

Does this PR also give you a tooltip of the object names in the list? Your photo displays a usability issue of now being able to see the full name of objects of the list, but your mouse is hovering over a thing we don't need a tooltip for as badly as each object in the list

HarsimranVirk:
> Does this PR also give you a tooltip of the object names in the list?

No, this PR was meant to solve issue #1484. The tooltip of the panel couldn't render the `<Trans>` component passed to it.

> It is not obvious what the full object names are, as the limited width has clipped them. That is where we need a tooltip badly.

Yes, this is definitely an issue. Though this PR is not meant to fix it.

> It would be nice to fix that, if it doesnt require too much refactoring

Now that you have mentioned it, I'd gladly take a look at it :) It might not take a lot of code to fix it I believe, (maybe just adding a title to the divs?). I'd open another PR for it though. (@4ian wouldn't like that in this PR I'm sure about that :p )

newIDE/app/src/UI/EditorMosaic/index.js
4ian: My spidey sense is clearly afraid of making a duplicate `mosaic-window-toolbar`. There is 9 chance out of 10 that this will create some weird issue in the future (someone create a theme, customize the CSS to add a border => why the heck is the border applied twice??... these kind of things that easily takes hours to understand and fix).

So much added complexity just for a tooltip 😬 I'm almost tempted to back off this implementation and just pass title as a string (using i18n._(t`Blabla`)).

With this no more re-definition of things that should be handled by react-mosaic. It should almost be an improved to react-mosaic to be able to pass a different tooltip... even if the tooltip itself is not the most important stuff we need to be honest.

Sorry for noticing only now.. :/ Clearly not satifsying solution here. Let's consider if we can just use i18n._, refactor the title to be using the `t` string litteral, and call it a day?

4ian:
A tooltip for objects is interesting :) 
I'm getting worried about the extra implementation of the toolbar. That's too much burden just for a tooltip - and having an extra div will create issues in the future. I would rather back off and find a simpler way to fix these tooltips, and then go to something else.

HarsimranVirk:
Alright, @4ian, I have removed the extra implementation of the toolbar :) I guess its best not to have it if it could become a problem in the future. Having title as t`stuff` would be the best case it seems.

newIDE/app/src/UI/EditorMosaic/index.js
```diff
@@ -19,7 +18,6 @@ export type Editor = {|
   renderEditor: () => React.Node,
   noTitleBar?: boolean,
   title?: React.Node,
```
4ian: This should be now `title?: MessageDescriptor`,

4ian:
You must also ensure that the rest of the code using `EditorMosaic` is updated

newIDE/app/src/UI/EditorMosaic/index.js
HarsimranVirk: Oh yes, silly mistake :)

4ian:
Looks great! Thanks, merged 👍

4ian:
It has been a bit of work for these, but I think we reached the proper solution :) Thanks for your patience!

PR #1497, opened 2020-03-06
Add old film, dot and color replace effects

Quarkstar:
Add old film effects, and there is still a problem.

It shows a static old picture like this, but it should be dynamic. In [pixi-demo](https://pixijs.io/pixi-filters/tools/demo/), the picture keep flashing like an old film.
Screenshot from 2020-03-06 08-39-28

Can someone help me?:smile: Please tell me if I did something wrong.


blurymind:
Use the isometric game example to test filters. Pixi filters give strange
results when used on a project with a color fill background

Quarkstar:
It looks good in isometric game. But the effect is more like an old picture rather than an old film.
Scratches on the image don't change any more. I guess it's because pixi-demo keeps rendering,
and the image is rendered only once in GDevelop.


Screenshot from 2020-03-06 18-23-03

blurymind:
For effects with animated properties, you may need to use the update function. Have a look at
https://github.com/4ian/GDevelop/pull/1325/files#diff-0111d3519b66bcc043df151a17f3ccfaR7

Extensions/Effects/old-film-pixi-filter.js
```diff
@@ -0,0 +1,42 @@
+gdjs.PixiFiltersTools.registerFilterCreator('OldFilm', {
+  makePIXIFilter: function(layer, effectData) {
+    var oldFilmFilter = new PIXI.filters.OldFilmFilter();
+
+    return oldFilmFilter;
+  },
+  update: function(filter, layer) {},
```
blurymind: this is where you would handle the filter's animated properties.
I believe you need to set seed to a random value here. Check how its done on the pixi-filter's example

Quarkstar: Thank you! That works well!

Extensions/Effects/old-film-pixi-filter.js
```diff
@@ -0,0 +1,48 @@
...
+  },
+  updateDoubleParameter: function(filter, parameterName, value) {
+    if (parameterName === 'sepia') {
+      filter.sepia = value;
+    }
+    else if (parameterName === 'noise') {
+      filter.noise = value;
+    }
+    else if (parameterName === 'noiseSize') {
+      filter.uniforms.noiseSize = value;
```
4ian: is it normal that you have `.uniforms` here?

4ian:
Nice job! Will be three nice new effects to have!!

Can you:
- add screenshot for the two other effects
- there are a lot of changes in `JsExtension.js` - you seem to have re-order the effects? Unfortunately this make this file extremely hard to review => to help me you can for now go back to the original file, then just add your three filters.

We can re-order this file in another PR :)

Quarkstar:
Very sorry. 
`filter.uniforms.noiseSize = value;` is a test which i forgot to delete. I have fixed it and reseted `JsExtension.js`.

## Effects screenshot

### Old Film
Screenshot from 2020-03-08 10-26-42

### Dot
Screenshot from 2020-03-08 10-27-08

### ColorReplace
Change `#7ED321` to `#F8E71C`.
Screenshot from 2020-03-08 10-28-10

## More work
I am trying to add ColorMatrixFilter. But there are 11 members and 22 methods in it, and one of the input is a 4*5 matrix.(most methods are setting the matrix with a specific value). I think just using `string`, `number` and `boolean` will produce a long list, which is unfriendly to users. Maybe we need a new type(like `color`), and a matrix input box in IDE. I will do it in another PR :)

4ian:
> filter.uniforms.noiseSize = value; is a test which i forgot to delete. I have fixed it and reseted JsExtension.js.

Thanks!

> Maybe we need a new type(like color), and a matrix input box in IDE. I will do it in another PR :)

Sure, let's try in another PR! :) 

Great work on this :) @blurymind The color replace will be useful for sprites once we have effects for objects.

I've updated a few descriptions, will merge this now! Thanks for your work on this 👍

Quarkstar:
Thank you! @blurymind @4ian Your suggestions really help me a lot.

PR #1499, opened 2020-03-06
Add blending mode effect for layers

Bouh:
A little note on this work for user interested:

In pixi [blendmode are deprecated on containers ](https://github.com/pixijs/pixi.js/wiki/v4-Gotchas#no-container-blend-mode-support)(in GD a layer is a container),so it's replace by VoidFilter, but [VoidFilter is also deprecated since 4.5.7 ](https://github.com/pixijs/pixi.js/pull/4349/files)of pixi. Because it was replace (most rename in fact) by [AlphaFilter](https://pixijs.download/dev/docs/PIXI.filters.AlphaFilter.html).

And because this effect "does nothing" like said the [old documentation](http://pixijs.download/v4.3.3/docs/PIXI.filters.VoidFilter.html) (remember VoidFilter is same as AlphaFilter).
I can just use the [lastest effect](https://www.npmjs.com/package/@pixi/filter-alpha) currently used in pixi v5.

IMPORTANT - The WebGL renderer only supports the NORMAL, ADD, MULTIPLY and SCREEN blend modes. (numbers: 0, 1, 2, 3)

Coded in 10min, most long was the research.

I've also give correct name on blendmode for objects. For keep consistency with the layer. And Normal blend mode have more sense than Alpha for people, 3 none/screen was wrong ;)

[Full List of blendmode](https://pixijs.download/dev/docs/PIXI.html#.BLEND_MODES).

Thank for reading me, here the result in GD, here the [playground with VoidFilter/AlphaFilter](https://www.pixiplayground.com/#/edit/HCUVbRDXsVRJWNgoql-7H)
image

Ready to merge, no more effect on this PR.

### DIFF
--- a/Core/GDCore/Extensions/Builtin/SpriteExtension/SpriteExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/SpriteExtension/SpriteExtension.cpp
@@ -388,7 +388,7 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsSpriteExtension(
 
       .AddParameter("object", _("Object"), "Sprite")
       .AddParameter("expression",
-                    _("Mode (0 : Normal, 1 : Add, 2 : Multiply, 3 : Screen)"))
+                    _("Mode (0: Normal, 1: Add, 2: Multiply, 3: Screen)"))
       .MarkAsSimple();
 
   obj.AddAction("FlipX",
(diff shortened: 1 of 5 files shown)


4ian:
> 3 none/screen was wrong ;)

Thanks for fixing this ;) 

Looks mostly good, thanks for working on this! 👍

4ian:
The version of Typescript we are using was crashing on the pixi filter file. It's fixed in next version of TS, but for now I've ignored the file.

4ian:
> Coded in 10min, most long was the research.

Not bad ;)

PR #1501, opened 2020-03-07
[WIP] Quit App functionality first step

ozata:
I am just getting started and trying to understand how to do it.

### DIFF
--- a/newIDE/app/src/MainFrame/ElectronMainMenu.js
+++ b/newIDE/app/src/MainFrame/ElectronMainMenu.js
@@ -27,7 +27,8 @@ type MainMenuEvent =
   | 'main-menu-open-preferences'
   | 'main-menu-open-language'
   | 'main-menu-open-profile'
-  | 'update-status';
+  | 'update-status'
+  | 'quit-app';
 
 type MenuItemTemplate =
   | {|
@@ -129,6 +130,10 @@ class ElectronMainMenu extends React.Component<Props, {||}> {
       ('update-status': MainMenuEvent),
       (event, status) => this._editor && this._editor.setUpdateStatus(status)
     );
+    ipcRenderer.on(
+      ('quit-app' : MainMenuEvent),
+      () => this._editor && this._editor.quitApp()
+    );
 
     this._buildAndSendMenuTemplate();
   }
@@ -177,6 +182,10 @@ class ElectronMainMenu extends React.Component<Props, {||}> {
           label: i18n._(t`Close Project`),
           accelerator: 'CommandOrControl+Shift+W',
           onClickSendEvent: 'main-menu-close',
+        },
+        {
+          label: i18n._(t`Quit App`),
+          onClickSendEvent: 'quit-app',
         },
       ],
     };
--- a/newIDE/app/src/MainFrame/index.js
+++ b/newIDE/app/src/MainFrame/index.js
@@ -92,6 +92,9 @@ import OpenFromStorageProviderDialog from '../ProjectsStorage/OpenFromStoragePro
 import SaveToStorageProviderDialog from '../ProjectsStorage/SaveToStorageProviderDialog';
 import OpenConfirmDialog from '../ProjectsStorage/OpenConfirmDialog';
 import verifyProjectContent from '../ProjectsStorage/ProjectContentChecker';
+
+const electron = optionalRequire('electron');
+
 const GD_STARTUP_TIMES = global.GD_STARTUP_TIMES || [];
 
 const gd = global.gd;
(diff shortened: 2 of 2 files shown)


newIDE/app/src/MainFrame/ElectronMainMenu.js
```diff
@@ -129,6 +130,10 @@ class ElectronMainMenu extends React.Component<Props, {||}> {
       ('update-status': MainMenuEvent),
       (event, status) => this._editor && this._editor.setUpdateStatus(status)
     );
+    ipcRenderer.on(
+      ('quit-app' : MainMenuEvent),
+      event => this._editor.quitApp()
```
blurymind: nit: no need to pass event if you are not going to use it

newIDE/app/src/MainFrame/index.js
```diff
@@ -96,6 +96,8 @@ const GD_STARTUP_TIMES = global.GD_STARTUP_TIMES || [];
 
 const gd = global.gd;
 
+// const app = require('electron') // Doing this crashes the app.
```
blurymind: nit: remove commented out code

4ian: Look at other files to see how to properly require electron... though not sure we want to do it in this file at all.

newIDE/app/src/MainFrame/ElectronMainMenu.js
blurymind: actually you already have access to electron in this file. You can probably quit the app right here, but it would probably be better to ask for confirmation first

newIDE/app/src/MainFrame/index.js
blurymind: I think hes trying to trigger CloseConfirmDialog  to appear. He just needs to put

newIDE/app/src/MainFrame/ElectronMainMenu.js
ozata: I will be checking out right now.

newIDE/app/src/MainFrame/index.js
ozata: Thanks for the review. I'll find the right way.

4ian: > I think hes trying to trigger CloseConfirmDialog to appear, but currently the prop that calls it is !!this.state.currentProject

Not exactly ;) The props is there to tell if the dialog should be shown or not *when* the window is trying to be closed.
CloseConfirmDialog has its logic coded inside to show itself or not.

So finding a way to "trigger" the closing (only works for Electron, you can not do this in a web-app) will make the dialog automatically kicks in :)

ozata: I understand that I will trigger the closing somehow then CloseConfirmDialog will kick in. I am just trying to imitate other functions in the index.js and created an ipcRenderer listening the event. I think if this.state.currentProject is true the quit box should be popped. (because shouldPrompt will be true in line 32 in CloseConfirmDialog.js) I am trying to solve the problem, will check other files.

newIDE/app/src/MainFrame/ElectronMainMenu.js
```diff
@@ -178,6 +183,10 @@ class ElectronMainMenu extends React.Component<Props, {||}> {
           accelerator: 'CommandOrControl+Shift+W',
           onClickSendEvent: 'main-menu-close',
         },
+        {
+          label: i18n._(t`Quit App`),
```
4ian: Quit GDevelop

newIDE/app/src/MainFrame/index.js
```diff
@@ -1674,6 +1678,10 @@ class MainFrame extends React.Component<Props, State> {
       snackMessageOpen: false,
     });
 
+  quitApp = () => {
+    electron.app.quit()
```
4ian: Let's check if electron is not null :) If we're on the web-app, it will be null - and while it does not make sense to have this on the web-app, it's a good idea to always use defensive programming.

newIDE/app/src/MainFrame/index.js
```diff
@@ -132,6 +135,7 @@ type State = {|
   helpFinderDialogOpen: boolean,
   eventsFunctionsExtensionsError: ?Error,
   gdjsDevelopmentWatcherEnabled: boolean,
+  quitApp: boolean
```
4ian: Not needed?

4ian:
Almost there :)

ozata:
> Almost there :)

Thanks for the explanations, 4ian. I believe #1524 solves this problem so I should close this PR.

4ian:
Let's close this - hope the explanations were useful anyway! :)

PR #1508, opened 2020-03-08
Made New Scene naming convention consistent

### LINKED ISSUE #1507: Naming inconsistency in newly created scenes
## Describe the bug
The naming when new scenes are created is not consistent with the previous one.

Screenshot from 2020-03-08 15-02-10

## To Reproduce
1. Go to Project
2. Click on Add a New Scene

Dilshaad21:
Fixes #1507  Earlier it had a default new scene `New scene` and on successive new scene creations it created new scenes named as `NewScene` which should be `New scene2`, `New scene3` and so on.

**Before changes**
Screenshot from 2020-03-08 15-02-10 (1)

**After changes**
Screenshot from 2020-03-08 15-57-37

### DIFF
--- a/newIDE/app/src/ProjectManager/index.js
+++ b/newIDE/app/src/ProjectManager/index.js
@@ -700,11 +700,6 @@ export default class ProjectManager extends React.Component<Props, State> {
             autoGenerateNestedIndicator={!forceOpen}
             renderNestedItems={() =>
               filterProjectItemsList(enumerateLayouts(project), searchText)
-                .sort((layout1: gdLayout,layout2: gdLayout) => {
-                  const name1 = layout1.getName();
-                  const name2 = layout2.getName();
-                  return (name1.localeCompare(name2));
-                })  
                 .map((layout: gdLayout, i: number) => {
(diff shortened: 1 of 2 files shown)


Dilshaad21:
@4ian please review

newIDE/app/src/ProjectManager/index.js
```diff
@@ -700,6 +700,11 @@ export default class ProjectManager extends React.Component<Props, State> {
             autoGenerateNestedIndicator={!forceOpen}
             renderNestedItems={() =>
               filterProjectItemsList(enumerateLayouts(project), searchText)
+                .sort((layout1: gdLayout,layout2: gdLayout) => {
```
4ian: Unfortunately it's important to show the order to the user because the order is important (first scene is the one launched at the beginning of the game). Can we avoid doing this for now?

Also later we may want to add tags of folders to organize scenes :)

Dilshaad21: @4ian changes have been done

4ian:
Merged, thanks! 👍

Issue #1478, opened 2020-02-28
Overlapping highlighted objectList item


geetesh-gupta:
## Describe the Bug
A UI issue, whenever a field is selected for renaming, its highlighted box overlaps with the object box below it. It can be seen in the screenshot.  
![overlapping](https://user-images.githubusercontent.com/34645022/75573038-07b87980-5a82-11ea-8356-ab54d266a1c3.png)

## Fix 
Remove top padding when highlighted. 

## To Reproduce
1. Go to 'a new project'
2. Right Click on any object name
3. Click on rename option

jimil749:
I did see the issue, maybe we could something like: whenever user selects rename, the entire text should be selected rather than putting the cursor add the end of the word. 
Just a suggestion!  :)

geetesh-gupta:
I think the issue is due to a styling class being added on focus. I have found out that Mui-focusVisible is added as a class when rename option is selected for an object. I have tried providing padding 0 using chrome inspector, it is working (check the screenshots).

When rename is not selected:
![without_rename_selected](https://user-images.githubusercontent.com/34645022/75579146-2cfdb580-5a8b-11ea-8b04-989d4d8a8ea2.png)

When rename is selected:
![with_rename_selected](https://user-images.githubusercontent.com/34645022/75579141-2a02c500-5a8b-11ea-8d9b-ea54cdb87801.png)

When rename is selected and padding top and bottom set to be zero:
![with_rename_selected_padding_zero](https://user-images.githubusercontent.com/34645022/75579145-2bcc8880-5a8b-11ea-94ae-d805d870532d.png)


But, I am currently unable to figure out, how to override its style. @4ian Kindly guide here.

4ian:
This is not critical but would be a welcome fix. I'll let you investigate in React components - this is typical search in React components and I would prefer not to have to guide people here.

bajaj99prashant:
@4ian, I am working on this issue.

jimil749:
There might be a roadblock here. The API for the component library (MUI Component) does not have a padding prop.
https://material-ui.com/api/text-field/ 

If you checkout the docs, the only way to overwrite is through an external css file. So, there might be a problem providing padding to the component!

4ian:
Are we sure padding is the solution? You can always pass it with a `style` prop.

bajaj99prashant:
[...]

PR #1511, opened 2020-03-08
overlapping highlighted objectList item

bajaj99prashant:
#1478 
Updated the margin of MUIListItemText from 4px to 0 auto,  and removed the top position of -7 from the TextField component.

### DIFF
--- a/newIDE/app/package.json
+++ b/newIDE/app/package.json
@@ -13,7 +13,7 @@
     "@storybook/react": "5.1.9",
     "babel-core": "^7.0.0-bridge.0",
     "babel-loader": "8.0.5",
-    "flow-bin": "^0.92.0",
+    "flow-bin": "0.120.1",
     "flow-coverage-report": "^0.4.0",
     "folder-hash": "^3.0.0",
     "follow-redirects": "^1.2.3",
--- a/newIDE/app/src/UI/List.js
+++ b/newIDE/app/src/UI/List.js
@@ -28,7 +28,7 @@ type ListItemRightButtonProps =
   | {|
       displayReloadButton: boolean,
       reloadButtonTooltip: string,
-      onReload: () => void,
+      onReload?: () => void,
     |}
   | {|
       displayMenuButton: boolean,
--- a/newIDE/app/src/UI/List.js
+++ b/newIDE/app/src/UI/List.js
@@ -17,6 +17,12 @@ import Tooltip from '@material-ui/core/Tooltip';
 import Add from '@material-ui/icons/Add';
 import Search from '@material-ui/icons/Search';
 
+const styles = {
+  listItemText: {
+    margin: '0 auto',
+  },
+};
+
 const useDenseLists = true;
 export const listItemWith32PxIconHeight = 40;
 export const listItemWithoutIconHeight = 37;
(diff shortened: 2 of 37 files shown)


newIDE/app/package.json
```diff
@@ -43,7 +43,7 @@
     "firebase": "^6.1.0",
     "fontfaceobserver": "2.0.13",
     "keen-tracking": "1.1.3",
-    "lodash": "4.17.4",
+    "lodash": "^4.17.15",
```
4ian: Cancel these changes - and also cancel all changes in package-lock.json - as it's not related to your changes :)

newIDE/app/src/UI/TextField.js
```diff
@@ -257,4 +257,4 @@ export default class TextField extends React.Component<Props, {||}> {
 
 // The "top" offset to add to the position of the TextField when
 // it's used inside a ListItem "primaryText"
-export const noMarginTextFieldInListItemTopOffset = -7;
+export const noMarginTextFieldInListItemTopOffset = 0;
```
4ian: Nice :)

newIDE/app/src/UI/List.js
```diff
@@ -222,6 +223,7 @@ export class ListItem extends React.Component<ListItemProps, ListItemState> {
             <MUIListItemText
               primary={props.primaryText}
               secondary={props.secondaryText}
+              style={{ margin: '0 auto' }}
```
4ian: I'm worried of these changes - what happen if you don't do this?

newIDE/app/src/UI/List.js
```diff
@@ -193,6 +193,7 @@ export class ListItem extends React.Component<ListItemProps, ListItemState> {
           <MUIListItemText
             primary={props.primaryText}
             secondary={props.secondaryText}
+            style={{ margin: '0 auto' }}
```
4ian: Creating a styles object outside of the component, and point to it: `style={styles.listItemText}`.
This avoids a new object to be created at each render, reducing pressure on the GC.

newIDE/app/package.json
bajaj99prashant: It might be got changed by mistake as I took a pull request from GDevelop's master. I will undo it.

newIDE/app/src/UI/List.js
bajaj99prashant: To start with, I adjusted the height of the TextField component in TextField.js, It worked in the browser but the text wasn't center on renaming the objects in the electron app as top was -7px. I inspected the ListItem and found that the position of ListItem was fixed and the MUIListItemText had marginTop and marginBottom of 4px's by default. Resulting in the extra height of the ListItem and overlapping with the next ListItem on renaming. So I thought it would be better to give marginTop/Bottom to 0 and gave TextField.js  top equal to 0, as it automatically centers the TextField in MUIListItem.
I can try again without using these changes to find if there's another possible way.

newIDE/app/src/UI/List.js
bajaj99prashant: I will change it as advised and will try to adopt the best practices in the future.

newIDE/app/src/UI/TextField.js
bajaj99prashant: Thanks :+1: , @4ian

Issue #1506, opened 2020-03-08
Visibility: The Table headings in the Instances Editor is not visible properly 


PawBud:
## Describe the bug
I noticed that the Headers of the Instances Table are not clearly visible when i looked at them through a TN panel( a lower resolution Display).

## To Reproduce
1. Go to 'New Scene'
2. Click on 'Instances Editor' at the top right
3. See error

## Other details
* I used a TN panel (resolution : 1280 by 720)
![lol](https://user-images.githubusercontent.com/48093317/76159907-3a7e0400-614b-11ea-9e83-42b02a0db579.JPG)

PawBud:
I have already fixed it by adding a bold font-weight to the column values along with a slight increase of font-weight in the header also

PR #1512, opened 2020-03-09
Visibility Of the Table headers of the Instances Table Improved

PawBud:
This is a fix regarding the issue: #1506 
I have bolded only the table headers. I did not change the colors or the font size since changing either makes it less appealing. The font Weight has been increased from 400 to 700 as I found it to be the most suitable for all screens.
**Before Change**
lol
**After Change**
IMG-1138
(I apologize if the photo is being displayed sideways)
Pls, review it. 
Thank You for considering it :)


blurymind:
Can you add tooltips if those are missing? It will help with narrow columns clipping text

newIDE/electron-app/app/DebuggerServer.js
```diff
@@ -1,5 +1,5 @@
-const WebSocket = require('ws');
-const log = require('electron-log');
+const WebSocket = require("ws");
```
blurymind: looks like prettier has added some noise to your diff due to your settings being different from those of the project

PawBud:
@blurymind sure I won't mind removing the double commas

newIDE/electron-app/app/DebuggerServer.js
PawBud: @blurymind sure I won't mind removing the double commas

PawBud:
@blurymind Sure I will add the tooltips as you suggested although I think I will do that in a different Pull request.

PawBud:
Its done @blurymind; I made the necessary changes.

newIDE/electron-app/app/DebuggerServer.js
```diff
@@ -62,4 +62,4 @@ module.exports = {
       cb(err);
     });
   },
-};
+};
```
blurymind: It looks like you have taken an extra newLine out of the file. Can you revert that? Having the extra line at the end of each js file is a weird javascript thing that is nice to have. Maybe this will help explain why
https://evanhahn.com/newline-necessary-at-the-end-of-javascript-files/

In any case, you might notice we are doing it in Gdevelop. Prettier should be doing it automatically for you

PawBud: @blurymind sure I thought prettier in my IDE does weird things but I am extremely sorry about this, I will fix it right away

4ian:
@Bouh @blurymind Are we ok making these headers bolder? Should this be done in other tables too?

PawBud:
@4ian as far as I observed the header are already Bolded (for ex: the instances table as shown below)
Screenshot from 2020-03-15 14-02-48
The headers of the Instances table were an exception I guess.

Bouh:
Keep concistent the UI is important, this should be done for all tables.
There are more two tables? I see Instances and Layers.

PawBud:
> Keep consistent the UI is important, this should be done for all tables.
> There are more two tables? I see Instances and Layers.
> 
> **Edit:** Nothing more after a little overview of the app.

as the screenshot shows. The layers table already has bolded headers and hence I thought that bolding the headers of the Instances table was a good idea

4ian:
The Dark theme needs an update :)

PawBud:
@4ian i ll look at it right away

PawBud:
@4ian I changed the 700 to 600 because 700 was looking a bit too much.

4ian:
Looks better indeed :) Thanks!

PawBud:
Thank you so much for merging it :)

PR #1514, opened 2020-03-09
Upgrade to Flow 0.120.1

4ian:
Upgrade to Flow 0.120.1 - last version used was one year old and there was it seems important speed improvement: https://medium.com/flow-type/improvements-to-flow-in-2019-c8378e7aa007 and interesting new features to come: https://medium.com/flow-type/what-were-building-in-2020-bcb92f620c75

### DIFF
--- a/newIDE/app/.flowconfig
+++ b/newIDE/app/.flowconfig
@@ -1,8 +1,11 @@
 [ignore]
 <PROJECT_ROOT>/resources/.*
-<PROJECT_ROOT>/node_modules/auth0-lock/node_modules/fbjs/flow/include/PromiseMap.js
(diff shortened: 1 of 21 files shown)


4ian:
@blurymind @Bouh I've upgraded to the latest Flow version (don't forget to do a `npm install` in newIDE/app after getting the latest master). 
I don't expect huge performance improvements... but the version we had was one year old. And from what I see, while Flow is taking up to 1GB while first parsing the code, it's then using almost no memory (everything must be stored/cached I guess).

Let's give it a try (maybe there are things to tweak in the .vscode/settings.json config for the Flow plugin to better work too?). I know that Typescript is very much popular these times, but I'll prefer to see a bit longer of things are evolving for Flow.

Issue #1214, opened 2019-08-25
Global Variables buttons hidden at the bottom


PascalLadalle:
## Describe the bug
If the list of global variables is long, we need to scroll down the list to access the buttons copy/delete/...
These buttons should be always visible, no matter how long the list is, like the Apply/Cancel buttons.

![image](https://user-images.githubusercontent.com/25978830/63653253-a75f1d00-c76a-11e9-883a-87ed62c5b14d.png)

4ian:
Could be moved to a `MiniToolbar` components for consistency with other editors like Sprite/Text editors

leedigital:
> Could be moved to a `MiniToolbar` components for consistency with other editors like Sprite/Text editors

which is  class to fix it...???

4ian:
Look at `newIDE/app/src/VariablesList/index.js` and `newIDE/app/src/VariablesList/VariablesEditorDialog.js`.

You can launch the web-app (as explained in the README) and use the [React Debugger](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi?hl=en) to identify the class of an element.

leedigital:
> Look at `newIDE/app/src/VariablesList/index.js` and `newIDE/app/src/VariablesList/VariablesEditorDialog.js`.
> 
> You can launch the web-app (as explained in the README) and use the [React Debugger](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi?hl=en) to identify the class of an element.

thank you very much... :-)

geetesh-gupta:
How to proceed with the issue? 
I will use DialogFile -> VariablesList/VariablesEditorDialog.js and ListFile -> VariablesList/index.js as alias
I have figured out that ListFile has editRow component which actually modifies the variables. To have it always visible on the same row as Apply/Cancel, we have to move the component out of ListFile and write that in DialogFile but it will require passing the onCopy etc. functions from VariableList component which is a child in DialogFile. So, basically I am confused what refactoring is required here?

4ian:
> So, basically I am confused what refactoring is required here?

I'm not sure! :) (not sure why I removed the label " 🙇‍♀️Careful thinking/design or refactoring needed"... ^^')

[...]

PR #1516, opened 2020-03-10
Update Dialog with Header and Footer functionality

### LINKED ISSUE #1503: Create a new game dialog: important configuration is offscreen
## Describe the bug
When you create a new game - important UI is buried at the bottom of the dialog offscreen.

For instance when you create a new game you see this :

GDevelop_5_jXg93TxYFo

but, if you don't scroll down all the way you will never see important configuration information about your new game:

GDevelop_5_6sOowOZqal

This is the PATH where your game is going to be stored. I didn't even know this existed for a long time until I accidentally scrolled all the way down. 

That path panel should be fixed to the bottom of the dialog, and the rest of the content scroll above it.  Meaning, it should "always" be there - in a fixed position at the bottom of the dialog.  This is important information that the user should see upon opening the new game dialog - the content above it should not push it down to the bottom of the dialog offscreen. 
GDevelop_5_rBqOiWIiIc

## To Reproduce
1. click file > create a new project
2. observe that the path information is not on screen.
3. scroll to the bottom of the dialog
4. observe the project path panel.

expected result:
should see the project path panel always onscreen, and content above is is scrollable.

geetesh-gupta:
Fixes #1503 #1214

### DIFF
--- a/newIDE/app/src/UI/DialogWithHeaderFooter/index.js
+++ b/newIDE/app/src/UI/DialogWithHeaderFooter/index.js
@@ -63,6 +63,13 @@ type Props = {|
   // Size
   maxWidth?: 'xs' | 'sm' | 'md' | 'lg' | 'xl' | false,
 
+  // Header, Footer and their styles
+  header?: React.Node,
+  footer?: React.Node,
+  headerContainerStyle?: Object,
+  footerContainerStyle?: Object,
+  mainContainerStyle?: Object,
+
   // Style:
   noMargin?: boolean,
   noTitleMargin?: boolean,
@@ -126,14 +133,16 @@ export default (props: Props) => {
               overflow: 'hidden',
             }}
           >
-            <ScrollView
-              style={{
-                ...styles.headerContainerStyle,
-                ...headerContainerStyle,
-              }}
-            >
-              {header}
-            </ScrollView>
+            {header && (
+              <ScrollView
+                style={{
+                  ...styles.headerContainerStyle,
+                  ...headerContainerStyle,
+                }}
+              >
+                {header}
+              </ScrollView>
+            )}
             <ScrollView
               style={{
                 ...styles.mainContainerStyle,
@@ -142,14 +151,16 @@ export default (props: Props) => {
             >
               {children}
             </ScrollView>
-            <ScrollView
-              style={{
-                ...styles.footerContainerStyle,
-                ...footerContainerStyle,
-              }}
-            >
-              {footer}
-            </ScrollView>
+            {footer && (
+              <ScrollView
+                style={{
+                  ...styles.footerContainerStyle,
+                  ...footerContainerStyle,
+                }}
+              >
+                {footer}
+              </ScrollView>
+            )}
           </DialogContent>
           <DialogActions
             style={
--- a/newIDE/app/src/UI/DialogWithHeaderFooter/index.js
+++ b/newIDE/app/src/UI/DialogWithHeaderFooter/index.js
@@ -1,0 +1,167 @@
+// @flow
+import * as React from 'react';
+import Dialog from '@material-ui/core/Dialog';
+import DialogActions from '@material-ui/core/DialogActions';
+import DialogContent from '@material-ui/core/DialogContent';
+import DialogTitle from '@material-ui/core/DialogTitle';
+import { ResponsiveWindowMeasurer } from '../Reponsive/ResponsiveWindowMeasurer';
+import ScrollView from '../ScrollView';
+
+const styles = {
+  defaultBody: {
+    overflowX: 'hidden',
+  },
+  noMarginBody: {
+    padding: 0,
+    overflowX: 'hidden',
+  },
+  flexRowBody: {
+    display: 'flex',
+    flexDirection: 'row',
+  },
+  flexBody: {
+    display: 'flex',
+  },
+  actionsContainerWithSecondaryActions: {
+    display: 'flex',
+    justifyContent: 'space-between',
+  },
+  noTitleMargin: {
+    padding: 0,
+  },
+  headerContainerStyle: {
+    maxHeight: 80,
+    overflow: 'auto',
+  },
+  mainContainerStyle: {
+    maxHeight: 240,
+    overflow: 'auto',
+  },
+  footerContainerStyle: {
+    maxHeight: 80,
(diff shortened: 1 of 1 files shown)


geetesh-gupta:
@4ian Please review and give your opinions

newIDE/app/src/UI/DialogWithHeaderFooter/index.js
```diff
@@ -0,0 +1,178 @@
...
+          <DialogContent
+            style={{
+              ...(noMargin ? styles.noMarginBody : styles.defaultBody),
+              ...(flexRowBody ? styles.flexRowBody : {}),
+              ...(flexBody ? styles.flexBody : {}),
+              overflow: 'hidden',
+            }}
+          >
+            {header && (
+              <ScrollView
```
4ian: Do we need a ScrollView for the header and the footer? We probably don't want that - they should NOT be scrollable.

4ian:
You added a file, but we would need:

* screenshot to show that it works, on multiple browsers and multiple screen sizes
* to use it :) At least for one dialog. We're not adding code that is not used by the rest of the codebase. Other dialogs can be converted later.
* to check/justify if we can integrate the new header/footer in the existing `Dialog`, or if it's better with a new component.
  * If it's a new component, can we just use `Dialog` but pass `header`, the ScrollView and `footer` as the content of the Dialog? :) (I would prefer this solution if possible).
* Once we've confirmed a good approach, add it to the Storybook

newIDE/app/src/UI/DialogWithHeaderFooter/index.js
geetesh-gupta: Shouldn't we allow user to have custom header component which may contain data of height greater than a single row 🤔 and allow him to limit the height of the component..

geetesh-gupta:
I am having doubt, currently, the Dialog component supports DialogTitle - DialogContent - DialogActions. So, for any content, we are needed to add between DialogContent component, but this creates a problem. As I have pushed the commit, if we need to have always visible components, then a maxHeight style has to be defined for them to limit the content height and have the height of header+footer+content within the limits. But I am somehow finding setting absolute height limit, not a good approach. Also, it creates a problem in mobile view where padding remains unused.
Another approach would be to have header and footer component out of the DialogContent component, and hence the Dialog will itself manage the over/underflow but it will be against the basic design of Dialog by materialUI. So, what are your opinions? Maybe I am unable to see another possible solution.

geetesh-gupta:
Adding to the above comment, if the approach of explicitly limiting the height of the component is used, then what should be the criteria for deciding the height, for ex: currently, these are set as 80/240/80 for header/content/footer respectively. These can also be defined in terms of window height or in units of 'vh'. I researched through the component implementation on material UI's repo, and found out that they are limiting the max height of the Dialog component to 90vh. So, maybe in DialogContent we can also have it.

geetesh-gupta:
HI @4ian , please provide your views for the above comments.

4ian:
> Another approach would be to have header and footer component out of the DialogContent component, and hence the Dialog will itself manage the over/underflow but it will be against the basic design of Dialog by materialUI.

Seems risky. I would avoid going against the material-ui implementation. 

>  As I have pushed the commit, if we need to always have components visible, then a maxHeight style has to be defined for them to limit the content height

Why would you be forced to have a maxHeight?

The way I see the result is:

* DialogContent
  * Header is rendered here - it takes as much space as its content, nothing special.
  * Scrollview - it takes as much space as possible, with a scrollbar if the content is too big.
  * Footer is rendered here - it takes as much space as its content, nothing special.

Now for the implementation:
* material-ui Dialog will make the content of the dialog scrollable if the content is taking more height than the maximum (`90vh` according to what you're saying). *This is fine.*
* If our header or footer are small enough, no issues...
* ... but if our header or footer are too large, or your screen to small? Well it's not a problem. There will be a scrollbar shown on the content of the dialog (i.e: a scrollbar for the header AND the scrollview AND the footer). *This is fine*.
* In most cases your screen will be big enough, in which case the ScrollView will take as much space as possible, and its content will probably have a scrollbar.

4ian:
> 80/240/80

To be specific, I think *we don't want* to decide any height for the header/content/footer.
It's your responsibility when you create a new dialog to not use a huge header or a huge footer.

If you do, you'll get a scrollbar automatically added by material-ui... it's fine! at least you can see all the dialog. But in theory, your footer/header are small so just the scrollview will scroll.

In any case, we should **never** have a scrollbar shown for the header only or the footer only.

Alternative implementation: can we override the Dialog title and the dialog actions? Not sure if it's a good idea (probably not).

geetesh-gupta:
Modifying Dialog title and Dialog actions will serve the purpose without any difficulty but it will not be a good idea. But, for the case of adding in Dialog content only, I need to provide a max-height to main content otherwise, the footer will not be visible without scrolling. So, how to decide that? Another problem is with the current Dialogs implementation. For ex:- `CreateProjectDialog`  uses `LocalStarters` component which passed as a prop in `LocalApp` but the footer that we want in the dialog is in `LocalStarters`, so how to pass it as a footer to `CreateProjectDialog`.  

Dialog part in [LocalApp.js](https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/LocalApp.js):
` renderCreateDialog={props => (
                    <CreateProjectDialog
                      {...props}
                      examplesComponent={LocalExamples}
                      startersComponent={LocalStarters}
                    />
                  )}`

Footer part in [LocalStarters.js](https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/ProjectCreation/LocalStarters.js) :
`
<Line expand>
              <Column expand>
                <LocalFolderPicker
                  fullWidth
                  value={this.state.outputPath}
                  onChange={this._handleChangePath}
                  type="create-game"
                />
              </Column>
            </Line>
`

[CreateProjectDialog.js](https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/ProjectCreation/CreateProjectDialog.js)

4ian:
> , I need to provide a max-height to main content otherwise, the footer will not be visible without scrolling. So, how to decide that?

I'm not sure - that's the difficulty of the task :) Material-ui is able to do it with its footer and header, we must mimic this behavior.

4ian:
Closing as there is no activity on this. Still thanks for your work on it :)

PR #1522, opened 2020-03-12
fixed popup close bug

### LINKED ISSUE #1517: Request for Export Popup to close when clicked out of its area
## Description
It will be nice if the 'export the game' popup closed itself like any popup does i.e. when we click outside of its area. 
The list of instances and layers editor do already close themselves when clicked outside of its area.

image

I believe It will make the application more user friendly .

gautamv95:
Fixes #1517 
Although not sure if this is the correct solution since this applies to all Dialogs present in the project.

### DIFF
--- a/newIDE/app/src/UI/Dialog/index.js
+++ b/newIDE/app/src/UI/Dialog/index.js
@@ -39,7 +39,7 @@ type Props = {|
   secondaryActions?: React.Node,
   onRequestClose?: () => void,
 
-  modal?: boolean, //Force the user to use one of the actions in the Dialog. Clicking outside the Dialog will not trigger the onRequestClose. Hence, Dialog will not close if clicked out of its scope
+  modal?: boolean, //Force the user to use one of the actions in the Dialog. If true, the dialog can't be closed by clicking outside or pressing Escape.
 
   children: React.Node, // The content of the dialog
 
@@ -105,7 +105,8 @@ export default (props: Props) => {
           fullWidth
           fullScreen={size === 'small'}
           maxWidth={maxWidth !== undefined ? maxWidth : 'md'}
-          disableBackdropClick={modal !== undefined ? !!modal : false}
+          disableBackdropClick={!!modal}
+          disableEscapeKeyDown={!!modal}
         >
           {title && (
             <DialogTitle
--- a/newIDE/app/src/UI/Dialog/index.js
+++ b/newIDE/app/src/UI/Dialog/index.js
@@ -40,6 +40,7 @@ type Props = {|
   onRequestClose?: () => void,
 
   modal?: boolean, // Force the user to use one of the actions in the Dialog. Clicking outside the Dialog will not trigger the onRequestClose.
+  disableBackdropClick?: boolean,
 
   children: React.Node, // The content of the dialog
 
@@ -80,6 +81,7 @@ export default (props: Props) => {
     flexRowBody,
     flexBody,
     noTitleMargin,
+    disableBackdropClick,
   } = props;
   const dialogActions = secondaryActions ? (
     <React.Fragment>
@@ -89,6 +91,8 @@ export default (props: Props) => {
   ) : (
     actions
   );
+
+  const disableBackdropClickSelector = disableBackdropClick ? false : true;
 
   const dialogContentStyle: DialogContentStyle = {
     ...(noMargin ? styles.noMarginBody : styles.defaultBody),
@@ -105,6 +109,7 @@ export default (props: Props) => {
           fullWidth
           fullScreen={size === 'small'}
           maxWidth={maxWidth !== undefined ? maxWidth : 'md'}
+          disableBackdropClick={disableBackdropClickSelector}
         >
           {title && (
             <DialogTitle
--- a/newIDE/app/src/UI/Dialog/index.js
+++ b/newIDE/app/src/UI/Dialog/index.js
@@ -105,7 +105,6 @@ export default (props: Props) => {
           fullWidth
           fullScreen={size === 'small'}
           maxWidth={maxWidth !== undefined ? maxWidth : 'md'}
-          disableBackdropClick
         >
           {title && (
             <DialogTitle
--- a/newIDE/app/src/UI/Dialog/index.js
+++ b/newIDE/app/src/UI/Dialog/index.js
@@ -39,7 +39,7 @@ type Props = {|
   secondaryActions?: React.Node,
   onRequestClose?: () => void,
 
-  modal?: boolean, //Changes default behaviour of dialog. Clicking outside the dialog box will trigger the onRequestClose.
+  modal?: boolean, //Force the user to use one of the actions in the Dialog. Clicking outside the Dialog will not trigger the onRequestClose. Hence, Dialog will not close if clicked out of its scope
 
   children: React.Node, // The content of the dialog
 
@@ -105,7 +105,7 @@ export default (props: Props) => {
           fullWidth
           fullScreen={size === 'small'}
           maxWidth={maxWidth !== undefined ? maxWidth : 'md'}
-          disableBackdropClick={modal !== undefined ? false : true}
+          disableBackdropClick={modal !== undefined ? !!modal : false}
         >
           {title && (
             <DialogTitle
--- a/newIDE/app/src/UI/Dialog/index.js
+++ b/newIDE/app/src/UI/Dialog/index.js
@@ -39,8 +39,7 @@ type Props = {|
   secondaryActions?: React.Node,
   onRequestClose?: () => void,
 
-  modal?: boolean, // Force the user to use one of the actions in the Dialog. Clicking outside the Dialog will not trigger the onRequestClose.
-  enableBackdropClick?: boolean, //Changes default behaviour of dialog. Clicking outside will trigger the onRequestClose.
+  modal?: boolean, //Changes default behaviour of dialog. Clicking outside the dialog box will trigger the onRequestClose.
 
   children: React.Node, // The content of the dialog
 
@@ -81,7 +80,7 @@ export default (props: Props) => {
     flexRowBody,
     flexBody,
     noTitleMargin,
-    enableBackdropClick,
+    modal,
   } = props;
   const dialogActions = secondaryActions ? (
     <React.Fragment>
@@ -106,9 +105,7 @@ export default (props: Props) => {
           fullWidth
           fullScreen={size === 'small'}
           maxWidth={maxWidth !== undefined ? maxWidth : 'md'}
-          disableBackdropClick={
-            enableBackdropClick !== undefined ? false : true
-          }
+          disableBackdropClick={modal !== undefined ? false : true}
         >
           {title && (
             <DialogTitle
--- a/newIDE/app/src/UI/Dialog/index.js
+++ b/newIDE/app/src/UI/Dialog/index.js
@@ -40,7 +40,7 @@ type Props = {|
   onRequestClose?: () => void,
 
   modal?: boolean, // Force the user to use one of the actions in the Dialog. Clicking outside the Dialog will not trigger the onRequestClose.
-  disableBackdropClick?: boolean,
+  enableBackdropClick?: boolean, //Changes default behaviour of dialog. Clicking outside will trigger the onRequestClose.
 
   children: React.Node, // The content of the dialog
 
@@ -81,7 +81,7 @@ export default (props: Props) => {
     flexRowBody,
     flexBody,
     noTitleMargin,
-    disableBackdropClick,
+    enableBackdropClick,
   } = props;
   const dialogActions = secondaryActions ? (
     <React.Fragment>
@@ -91,9 +91,6 @@ export default (props: Props) => {
   ) : (
     actions
   );
-
-  const disableBackdropClickSelector = disableBackdropClick ? false : true;
-
   const dialogContentStyle: DialogContentStyle = {
     ...(noMargin ? styles.noMarginBody : styles.defaultBody),
     ...((flexRowBody ? styles.flexRowBody : {}): DialogContentStyle),
@@ -109,7 +106,9 @@ export default (props: Props) => {
           fullWidth
           fullScreen={size === 'small'}
           maxWidth={maxWidth !== undefined ? maxWidth : 'md'}
-          disableBackdropClick={disableBackdropClickSelector}
+          disableBackdropClick={
+            enableBackdropClick !== undefined ? false : true
+          }
         >
           {title && (
             <DialogTitle
--- a/newIDE/app/src/UI/Dialog/index.js
+++ b/newIDE/app/src/UI/Dialog/index.js
@@ -39,7 +39,7 @@ type Props = {|
   secondaryActions?: React.Node,
   onRequestClose?: () => void,
 
-  modal?: boolean, //Force the user to use one of the actions in the Dialog. If true, the dialog can't be closed by clicking outside or pressing Escape.
+  cannotBeDismissed?: boolean, //Force the user to use one of the actions in the Dialog. If true, the dialog can't be closed by clicking outside or pressing Escape.
 
   children: React.Node, // The content of the dialog
 
@@ -80,7 +80,7 @@ export default (props: Props) => {
     flexRowBody,
     flexBody,
     noTitleMargin,
-    modal,
+    cannotBeDismissed,
   } = props;
   const dialogActions = secondaryActions ? (
     <React.Fragment>
@@ -105,8 +105,8 @@ export default (props: Props) => {
           fullWidth
           fullScreen={size === 'small'}
           maxWidth={maxWidth !== undefined ? maxWidth : 'md'}
-          disableBackdropClick={!!modal}
-          disableEscapeKeyDown={!!modal}
+          disableBackdropClick={!!cannotBeDismissed}
+          disableEscapeKeyDown={!!cannotBeDismissed}
         >
           {title && (
             <DialogTitle
--- a/newIDE/app/src/BehaviorsEditor/NewBehaviorDialog.js
+++ b/newIDE/app/src/BehaviorsEditor/NewBehaviorDialog.js
@@ -191,7 +191,7 @@ export default class NewBehaviorDialog extends Component<Props, State> {
             ]}
             secondaryActions={<HelpButton helpPagePath="/behaviors" />}
             open={open}
-            modal={false}
+            cannotBeDismissed={false}
             noMargin
           >
             <Tabs value={currentTab} onChange={this._changeTab}>
--- a/newIDE/app/src/BehaviorsEditor/NewBehaviorDialog.js
+++ b/newIDE/app/src/BehaviorsEditor/NewBehaviorDialog.js
@@ -191,6 +191,7 @@ export default class NewBehaviorDialog extends Component<Props, State> {
             ]}
             secondaryActions={<HelpButton helpPagePath="/behaviors" />}
             open={open}
+            modal={false}
             noMargin
           >
             <Tabs value={currentTab} onChange={this._changeTab}>
--- a/newIDE/app/src/EffectsList/EffectsListDialog.js
+++ b/newIDE/app/src/EffectsList/EffectsListDialog.js
@@ -42,7 +42,6 @@ export default class EffectsListDialog extends React.Component<Props, {||}> {
             key={'Apply'}
           />,
         ]}
-        modal
         open
         onRequestClose={onApply}
         title={<Trans>Effects</Trans>}
--- a/newIDE/app/src/EffectsList/EffectsListDialog.js
+++ b/newIDE/app/src/EffectsList/EffectsListDialog.js
@@ -42,7 +42,7 @@ export default class EffectsListDialog extends React.Component<Props, {||}> {
             key={'Apply'}
           />,
         ]}
-        modal={true}
+        cannotBeDismissed={true}
         open
         onRequestClose={onApply}
         title={<Trans>Effects</Trans>}
--- a/newIDE/app/src/EffectsList/EffectsListDialog.js
+++ b/newIDE/app/src/EffectsList/EffectsListDialog.js
@@ -42,6 +42,7 @@ export default class EffectsListDialog extends React.Component<Props, {||}> {
             key={'Apply'}
           />,
         ]}
+        modal={true}
         open
         onRequestClose={onApply}
         title={<Trans>Effects</Trans>}
--- a/newIDE/app/src/EventsBasedBehaviorEditor/EventsBasedBehaviorEditorDialog.js
+++ b/newIDE/app/src/EventsBasedBehaviorEditor/EventsBasedBehaviorEditorDialog.js
@@ -44,6 +44,7 @@ export default class EventsBasedBehaviorEditorDialog extends React.Component<
             key={'Apply'}
           />,
         ]}
+        modal={true}
         open
         onRequestClose={onApply}
         title={<Trans>Edit the behavior</Trans>}
--- a/newIDE/app/src/EventsBasedBehaviorEditor/EventsBasedBehaviorEditorDialog.js
+++ b/newIDE/app/src/EventsBasedBehaviorEditor/EventsBasedBehaviorEditorDialog.js
@@ -44,7 +44,6 @@ export default class EventsBasedBehaviorEditorDialog extends React.Component<
             key={'Apply'}
           />,
         ]}
-        modal
         open
         onRequestClose={onApply}
         title={<Trans>Edit the behavior</Trans>}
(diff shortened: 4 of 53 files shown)


geetesh-gupta:
Hi @gautamv95, good approach but this solution will allow force all dialogs to be closed without selecting an option from the Dialog, which is not suggested. Try to further dive into the props required by the Dialog and you would see we have a prop that would allow selecting whether to close (or not) the Dialog. Try to use that prop to decide. :)

gautamv95:
Thanks @geetesh-gupta I will look into that, however, I was hoping that clicking on the backdrop of any Dialog should close it, irrespective of selection of anything. Need reviews for the same, anyone?

4ian:
While it's ok for modal that have just a few things to select, we also risk with this closing a dialog when the user mistakenly click outside it, and lose changes.
Can we instead have a case by case approach?

gautamv95:
Added a new approach, which can be extended to other Dialog components as well which don't take user input

4ian:
Have you taken a look at onRequestClose? I'm confused how this fix #1517 to be honest, can you describe the behavior that you're adding? :)

gautamv95:
Sure, I added a new Property in [Dialog Component](https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/UI/Dialog/index.js) called 'disableBackdropClick'(sorry for bad naming, it should be enableBackdropClick), which if passed through parent component will disable the 'disableBackdropClick' prop, therefore clicking on the Backdrop of that dialog will close it.

As for the onRequestClose?, I couldn't find any solution to use it here 🤔

4ian:
> As for the onRequestClose?, I couldn't find any solution to use it here 🤔

I was thinking that defining `onRequestClose` is sufficient to get the dialog closed with a backdrop click or Escape (because Escape must work too)? This is what is done for all other dialogs.

gautamv95:
This was the best I could come up with, can't decide what to do with `onRequestClose?`

4ian:
Can you check an existing dialog that use onRequestClose, and see if:
- it can be closed with a backdrop click?
- if yes, is adding/removing onRequestClose changing something?

geetesh-gupta:
I think the problem here is that the `Dialog` component has been hardcoded with `disableBackdropClick`  prop as given below: https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/UI/Dialog/index.js#L108

This prop decides to close or not a `Modal` in MaterialUI.
So, to have capability to choose this for dialogs, we have to pass some prop from parent to `Dialog` component to tell it to set the value of `disableBackdropClick`.

So, I think there is no need to create a new prop because there exists an already defined prop `modal` which can be used to decide the value of `disableBackdropClick` as given below. 
https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/UI/Dialog/index.js#L42

Please correct me if I am wrong.

4ian:
> So, I think there is no need to create a new prop because there exists an already defined prop  modal which can be used to decide the value of disableBackdropClick as given below.
> https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/UI/Dialog/index.js#L42
> Please correct me if I am wrong.

This sounds correct to me.

We would then need: 
* See how to set disableBackdropClick according to "modal"
* Go through all the windows and add "modal" props if needed

The second point is actually the most complex. Do we want all dialogs to be able to be closed with Escape/Backdrop click? What should be the criteria?
I'd like to get answers to this before going any further :)

gautamv95:
I agree with @geetesh-gupta about using the modal props but using this approach, we would have to change a lot of files and would also lead to code duplication, that's why I chose to add a new prop.

And I think the criteria for the dialogs which can be closed with Escape/Backdrop click should be:

1) Dialogs that take user input via text(eg: [ObjectEditorDialog](https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/ObjectEditor/ObjectEditorDialog.js), [PreferencesDialog](https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/MainFrame/Preferences/PreferencesDialog.js)) should be closed only when the User has selected one of the actions in the Dialog.
2) Dummy Dialogs that are either for presentation or displaying buttons(eg: [ExportDialog](https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/Export/ExportDialog.js), [AboutDialog](https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/MainFrame/AboutDialog.js)) can be closed via Escape/Backdrop click.

These are my thoughts, on the criteria for the dialogs, please put in your thoughts as well. 
> The other approach I found was setting disableBackdropClick to false for the whole UI Dialog, and passing modal as a prop in those components which we do not to get closed by clicking on backdrop.

geetesh-gupta:
According to me, other than the criterion specified by @gautamv95, only those Dialogs which do not create a change in the state of the project, can be allowed to close with Escape/Backdrop click.

> we would have to change a lot of files and would also lead to code duplication, that's why I chose to add a new prop.

Pondering on this fact, I don't think a lot of files will require changes. We can set the `disableBackdropClick` based on `modal` prop and have the `modal` prop optional. So, all the current dialogs will not have `modal` value set but those which need to be closed can set the `modal` value. So, only those dialogs which need to close can needs to be changed, which is I think can be incrementally done.

4ian:
> And I think the criteria for the dialogs which can be closed with Escape/Backdrop click should be

That seems fair.

> only those Dialogs which do not create a change in the state of the project, can be allowed to close with Escape/Backdrop click

Yeah that can be an idea.

We could almost get a list of all the dialogs (files ending with "Dialog.js" basically) and make a list of those that can be closed by Escape/click outside, and the others :) 

> So, only those dialogs which need to close can needs to be changed, which is I think can be incrementally done.

Yeah, I'm not afraid by adding props to a few dozens dialogs in the codebase.

gautamv95:
I guess that would be problematic since it will include all dialogs because every dialog changes its state whenever it is closed and opened, for reference: [Dialog MaterialUI](https://material-ui.com/components/dialogs/#dialog)
> According to me, other than the criterion specified by @gautamv95, only those Dialogs which do not create a change in the state of the project, can be allowed to close with Escape/Backdrop click.

I think that's a really nice solution, just we need to focus on those dialogs which already have added `modal` to their implementation, eg: [Here](https://github.com/4ian/GDevelop/blob/3ff5dd7cd2c1f6b2bfd57b2ef79b65d1e47ab15a/newIDE/app/src/EffectsList/EffectsListDialog.js#L45), and much more such dialogs
> Pondering on this fact, I don't think a lot of files will require changes. We can set the `disableBackdropClick` based on `modal` prop and have the `modal` prop optional. So, all the current dialogs will not have `modal` value set but those which need to be closed can set the `modal` value. So, only those dialogs which need to close can needs to be changed, which is I think can be incrementally done.

4ian:
> every dialog changes its state whenever it is closed and opened

I think that by "state" he meant "is the dialog changing something in the project/asking the user for some input" (in which case closing it by a click outside/escape is dangerous, if the user spent a lot of time setting up something) or "is the dialog just asking for a choice" (in which case closing it is never a problem)

gautamv95:
That's great, should I start implementing it using `modal` prop then?

gautamv95:
@4ian please review

newIDE/app/src/UI/Dialog/index.js
```diff
@@ -39,7 +39,7 @@ type Props = {|
   secondaryActions?: React.Node,
   onRequestClose?: () => void,
 
-  modal?: boolean, // Force the user to use one of the actions in the Dialog. Clicking outside the Dialog will not trigger the onRequestClose.
+  modal?: boolean, //Changes default behaviour of dialog. Clicking outside the dialog box will trigger the onRequestClose.
```
4ian: > Changes default behaviour of dialog

=> this comment is not providing any information.

Please update the comment to include only necessary information.

newIDE/app/src/UI/Dialog/index.js
```diff
@@ -105,7 +105,7 @@ export default (props: Props) => {
           fullWidth
           fullScreen={size === 'small'}
           maxWidth={maxWidth !== undefined ? maxWidth : 'md'}
-          disableBackdropClick
+          disableBackdropClick={modal !== undefined ? false : true}
```
4ian: reconsider what happen if you:
don't pass modal,
pass `modal={true}`,
pass `modal={false}`,

false must be equivalent to not passing modal.

4ian:
Thanks for your work :)
Can you make a recap of:

- what is `modal` doing
- and a list of all the dialogs (files ending with "Dialog.js" basically) and make a list of those that are modal or not, and why.

newIDE/app/src/UI/Dialog/index.js
gautamv95: modified

gautamv95:
Sorry for the late work, got busy with the proposal

`modal` is a prop which when passed through a Dialog, will do the following depending on its value:
1) True -> The Dialog will not close when clicked out of its scope(backdropClick)
2) False -> The Dialog will close when clicked out of its scope


#### Selector Dialog
A dialog that shows multiple options, from which the user has to select one. In this type of dialog, the user does not have a chance of losing his changes when the dialog closes if clicked outside, hence `modal` = false.

#### Editor Dialog
A dialog that has some text input. In this type of dialog, the user has a chance of losing his changes when the dialog closes if clicked outside, hence `modal` = true.

#### Info Dialog
A dialog which displays some text, and whose purpose is to inform the user about something, these modals could be closed when clicked outside, hence `modal`=false

Dialog Name | `modal` value | Explaination
-- | -- | --
NewBehaviorDialog | false | Selector Dialog
EffectsListDialog | true | Editor Dialog
EventsBasedBehaviorDialog | true | Editor Dialog
BehaviorMethodSelectorDialog | false | Selector Dialog
ExtensionFunctionSelectorDialog | false | Selector Dialog
OptionsEditorDialog | true | Editor Dialog
InstructionEditorDialog | true | Editor Dialog
NewInstructionEditorDialog | true | Editor Dialog
ExpressionParametersEditorDialog | true | Editor Dialog
ExportDialog | false | Selector Dialog
BroweserPreviewErrorDialog | false | Info Dialog
BrowserPreviewLinkDialog | true | Info Dialog, but shows important information
BuildsDialog | false | Info Dialog
LocalNetworkPreviewDialog  | true | Info Dialog, but shows important information
ExtensionInstallDialog | false | Info Dialog
ExtensionsSearchDialog | true | Editor Dialog
LayerRemoveDialog | false | Selector Dialog
AboutDialog | false | Info Dialog
BrowserIntroDialog | false | Info Dialog
ChangeLogDialog | false | Info Dialog
LayoutChooserDialog | false | Selector Dialog
LanguageDialog | false | Selector Dialog
PreferencesDialog | true | Editor Dialog
ObjectEditorDialog | true | Editor Dialog
ObjectGroupEditorDialog | true | Editor Dialog
NewObjectDialog | false | Selector Dialog
PlatformSpecificAssestsDialog | true | Editor Dialog
CreateAccountDialog | true | Editor Dialog
LoginDialog | true | Editor Dialog
ProfileDialog | false | Info Dialog
SubscriptionDialog | true | Info Dialog, but shows important information
SubscriptionPendingDialog | true | Info Dialog, but shows important information
CreateProjectDialog | false | Selector Dialog
ProjectPropertiesDialog | true | Editor Dialog
OpenConfirmDialog | false | Selector Dialog
OpenFromStorageProviderDialog | false | Selector Dialog
SaveToStorageProvdierDialog | true | Info Dialog, but shows important information
DownloadSaveAsDialog | false | Info Dialog
GoogleDriveSaveAsDialog | true | Info Dialog, but shows important information
LoginDialog | true | Editor Dialog
ScenePropertiesDialog | true | Editor Dialog
SetupGridDialog | true | Editor Dialog
VariableEditorDialog | true | Editor Dialog


Some extra Dialogs, when I searched for the string -> '<Dialog'
Dialog Name | `modal` value | Explaination
-- | -- | --
HelpFinder | false | Info Dialog
AnimationPreview | false | Info Dialog
PointsEditor | true | Editor Dialog
CollisionMaskEditor | true | Editor Dialog
SubscriptionChecker | false | Info Dialog
BrowserResourceSources | false | Selector Dialog
InstructionEditor(Events Sheet) | true | Editor Dialog
ForgetPassword(Login Dialog) | true | Editor Dialog

Don't know what to do with following Dialogs, need help 
Dialog Name | `modal` value
-- | --
[EventsTextDialog](https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/EventsSheet/InstructionEditor/EventTextDialog.js) | `undefined`
[EventsFuntionExtractorDialog](https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/EventsSheet/EventsFunctionExtractor/EventsFunctionExtractorDialog.js) | `undefined`
[EventsContextAnalyzerDialog](https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/EventsSheet/EventsContextAnalyzerDialog.js) | `undefined`
[Stories](https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/stories/index.js#L541) | `undefined`
[EditTagsDialog](https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/UI/EditTagsDialog.js) | `undefined`

newIDE/app/src/UI/Dialog/index.js
gautamv95: modified

newIDE/app/src/UI/Dialog/index.js
```diff
@@ -105,7 +105,7 @@ export default (props: Props) => {
           fullWidth
           fullScreen={size === 'small'}
           maxWidth={maxWidth !== undefined ? maxWidth : 'md'}
-          disableBackdropClick
+          disableBackdropClick={modal !== undefined ? !!modal : false}
```
4ian:
```suggestion
          disableBackdropClick={!!modal}
```

newIDE/app/src/UI/Dialog/index.js
```diff
@@ -39,7 +39,7 @@ type Props = {|
   secondaryActions?: React.Node,
   onRequestClose?: () => void,
 
-  modal?: boolean, // Force the user to use one of the actions in the Dialog. Clicking outside the Dialog will not trigger the onRequestClose.
+  modal?: boolean, //Force the user to use one of the actions in the Dialog. Clicking outside the Dialog will not trigger the onRequestClose. Hence, Dialog will not close if clicked out of its scope
```
4ian: `// If true, the dialog can't be closed by clicking outside or pressing Escape.`

4ian:
Thanks for the list and for your investigations!! Very useful :) 👍 From what I see this makes sense to me! 👍

> The Dialog will not close when clicked out of its scope

Should we also add disableEscapeKeyDown (see https://material-ui.com/api/modal/)? To avoid people doing a lot of work in "Editor" dialogs and losing them all by pressing Escape.
I think that if we disable backdrop click, we should also disable escape key (should always be the same for both) :)

> Don't know what to do with following Dialogs, need help

* EventsTextDialog > Editor dialog
* EventsFunctionExtractorDialog > Editor Dialog
* EventsContextAnalyzerDialog > Info Dialog
* Stories > these are storybook stories, nothing linked to a dialog
* EditTagsDialog > Editor Dialog

newIDE/app/src/UI/Dialog/index.js
gautamv95: modified

newIDE/app/src/UI/Dialog/index.js
gautamv95: modified

gautamv95:
> * EventsTextDialog > Editor dialog
> * EventsFunctionExtractorDialog > Editor Dialog
> * EventsContextAnalyzerDialog > Info Dialog
> * Stories > these are storybook stories, nothing linked to a dialog
> * EditTagsDialog > Editor Dialog

Thanks for the reviews, modified the remaining Dialogs

4ian:
Alright looks good! Now last things:

* I'm not sure `modal` is in fact an appropriated name anymore for this prop 😬😄 "Modal" is used to say that something is shown on screen and need a choice or be dismissed before going back to what was before on the screen... which is the case for all the dialogs! :)
Instead, we could call this `canBeEasilyDismissed`.

  This should be "easy" to replace by searching for `modal={` and replace it by `canBeEasilyDismissed={` in the whole codebase. Rename the prop.

  Then also search for "modal" in the code (in lowercase) to be sure you've not missed anything.

* Check Flow type errors: https://semaphoreci.com/4ian/gd/branches/pull-request-1522/builds/6
  Seems like you've added modal on some dialogs that are not expecting them in their props.

image
image

gautamv95:
Can we instead name it `cannotBeDismissed` since when it is set to true, the dialog will have disabled Backdrop Click and Escape Key, thus not making it easily dismissible 🤔 
> Instead, we could call this `canBeEasilyDismissed`.

My bad will modify it
> Seems like you've added modal on some dialogs that are not expecting them in their props.

4ian:
> Can we instead name it cannotBeDismissed since when it is set to true, the dialog will have disabled Backdrop Click and Escape Key, thus not making it easily dismissible 🤔

Ah yes sure, that's a good proof that modal was not a great name :D

4ian:
Alright looks good, let's merge this! Thanks for working on this, will be very useful to make dialogs "safer" to use :)

4ian:
One improvement I'm thinking:
In case of Escape pressed when `cannotBeDismissed` is false, we could ask the user to confirm s/he wants to close the dialog with a confirmation popup, especially if we see that the dialog was opened since more than 10 seconds: "Are you sure you want to dismiss any changes made in this dialog?".

gautamv95:
> One improvement I'm thinking:
> In case of Escape pressed when `cannotBeDismissed` is false, we could ask the user to confirm s/he wants to close the dialog with a confirmation popup, especially if we see that the dialog was opened since more than 10 seconds: "Are you sure you want to dismiss any changes made in this dialog?".

That's a nice improvement, but we could do this when `cannotBeDismissed` is true, and user presses escape, also are 10 seconds necessary, I mean if the user changes any properties and tries to press escape, the confirmation dialog should pop up, does this look better?

4ian:
> we could do this when cannotBeDismissed is true

Yes sorry!

> are 10 seconds necessary

It was just a quick idea to guess "has the user made any significant changes that would take a lot of time to re-do if he dismiss the dialog by mistake" :) 

>  mean if the user changes any properties and tries to press escape, the confirmation dialog should pop up, does this look better?

Sure it would be perfect, but how do you find if a property was changed? Because Dialog is a generic container. That's why i suggested a simple time based solution - but feel free to try something else! :)

gautamv95:
> Sure it would be perfect, but how do you find if a property was changed? Because Dialog is a generic container. That's why i suggested a simple time based solution - but feel free to try something else! :)

I think we can maybe use Hooks or synthetic events, will let you know, once I get a solution

Bouh:
I've pressed my escape key several times to close a dialog that I think should have been closed by escaping or clicking in backdrop, and nothing happend. This was unlogic to me.

But after rereading a little bit the choices made here I realize that it's not practical for some dialogs. And some dialogs with    ` cannotBeDismissed={false}` don't work.

I'am in "Search for New Extensions", the choice made here is: 

Dialog Name | modal value | Explaination
-- | -- | --
ExtensionsSearchDialog | true | Editor Dialog

This is just a big list of all extensions, nothing to save, just a choice to make.
There is only a "Close" button.
I change my plan i don't want make a choice anymore, so i press escape or click on the backdrop. So for me this should be allowed to be dissmiss. But it's not the case because the value is on `true`.

Lot of dialog are not pratical or dissmiss dosen't work at all.

Another example.
I go for add a new behavior on my object. The dialog is "Add a new behavior to the object" 
There is only an "Close" button.

The choice done in this PR is 

Dialog Name | modal value | Explaination
-- | -- | --
NewBehaviorDialog | false | Selector Dialog

Well this dosen't work. I can't close the dialog with click on backdrop or press escape.

More generally when we click on backdrop/press escape the user don't want save the change.
The button Apply or Ok is are here for save the change, it's a confirm button for validate the actions from the user.

Btw some dialog work fines, example "Export project to a standalone game"

Dialog Name | modal value | Explaination
-- | -- | --
ExportDialog | false | Selector Dialog

This work fine too : 
CreateProjectDialog
BuildsDialog
ExtensionInstallDialog

Not tested on all others.

>The second point is actually the most complex. Do we want all dialogs to be able to be closed with Escape/Backdrop click? What should be the criteria?

Yes! Escape/Backdrop click should be just an shortcut for the "Cancel" / "Close" button.


Be forced to make a choice by clicking on "Close" instead press escape isn't pracital.
Be forced to click on Apply or Ok for validate the change is logic. (this make sense)
Be forced to click on Cancel for close and unsave isn't pratical. Escape/backdrop should be a shortcut for dissmiss/unsave.


For some **Editor Dialog**,  **EffectsListDialog** and **PreferencesDialog** for examples.


**EffectsListDialog** have a Ok button. If i click in backdrop/press escape it's for unsave. But nothing happend. And why not an Cancel and Apply ?

**PreferencesDialog** have a Close button. Here it's weird to have a Close button instead Apply and Close, or just an OK. But i understand all change are directly saved, so the user have just to click somewhere for on Close. I hope use escape for close this dialog.

Both dialog are an Editor Dialog, and both work differently, it's wierd.
The logic in use of Ok, Apply, Cancel, Close is... not logic everywhere.
Some dialog save automatically all chnage, some of them need to be apply. Some of them have only one button Ok, it's for apply all the change or just close the dialog.

This is a point of view from me as user, not from dev.

4ian:
I've been thinking about this. We should I think go back to everything dismissable with Escape (or backdrop click).

But we protect against losing changes (in particular for "big" editors like action/condition editor) by showing a "Are you sure you want to cancel any change" if the user spent let's say more than 1 minute in the dialog. What do you think?

gautamv95:
My bad, I found the error, it happened because ObjectEditorDialog, was opened before the NewBehaviourDialog so its onBackdropClick didn't work, I will take a look into this again
Dialog Name | `cannotBeDismissed` value | Explanation
-- | -- | --
NewBehaviorDialog | false | Selector Dialog

I chose this because we have to Search for the Extension we are looking for, so set the `cannotBeDismissed` value to true
Dialog Name | `cannotbeDismised` value | Explanation
-- | -- | --
ExtensionSearchDialog | true | Editor Dialog

Agreed, some Dialogs, still need some modifications, see #1579 its all linked to whether that Dialog needs access to `WholeProjectRefactorer` or not, 
> For some **Editor Dialog**, **EffectsListDialog** and **PreferencesDialog** for examples.
> Both dialog are an Editor Dialog, and both work differently, it's weird.
> The logic in use of Ok, Apply, Cancel, Close is... not logic everywhere.
> Some dialog save automatically all change, some of them need to be apply. Some of them have only one button Ok, it's for apply all the change or just close the dialog.

In my opinion, all Dialogs should be dismissed with Escape Key, and all Info and Selector Dialogs with Backdrop Click, by protecting against losing changes to Editor Dialogs. Furthermore, I think we can add confirmation dialog, by specifying a time limit of 30-40 secs, or maybe 1 minute.
>But we protect against losing changes (in particular for "big" editors like action/condition editor) by showing a "Are you sure you want to cancel any change" if the user spent let's say more than 1 minute in the dialog. What do you think?

Bouh:
- The confirmation dialog can pop after 1minutes for instruction editor, when you search the expressions this can take more than 30sc.
30sc is fine for the other dialogs.

- Escape and backdrop click is the same action, i'm afraid if you split the actions the user will be lost.
There should not have difference between Selector Dialogs and Editor Dialogs or Info dialog.
Like you said 

> all Dialogs should be dismissed

.... with escape and backdrop click 🙏

gautamv95:
>The confirmation dialog can pop after 1minutes for instruction editor, when you search the expressions this can take more than 30sc.
30sc is fine for the other dialogs.

Okay, I will start working on this
Just a wild thought, can we add an option in preferences menu of Editor, like: "Disable popups with Escape Key/BackdropClick?"

> The logic in use of Ok, Apply, Cancel, Close is... not logic everywhere.

I will create a list of all Dialogs, having proper actions and secondary actions and will relate them in a new issue

4ian:
> Just a wild thought, can we add an option in preferences menu of Editor, like: "Disable popups with Escape Key/BackdropClick?"

Ideally we don't want to add a preference - we should aim to find a behavior that is intuitive for everyone. Users will get frustrated and leave GDevelop before looking at preferences - believe me :)

Just a nit: "Disable popups with Escape Key/BackdropClick" -> this does not mean anything right? We are not "disabling popups", right?

For now I think it's more important to get back Escape to work for all dialogs, so that power users are not frustrated and can keep their velocity.
I've re-enabled escape/backdrop click for all dialogs in fb8926dd6652b34e394cb1f427f178928110cda2

> Okay, I will start working on this.

Thanks! I believe a quick confirmation would be great. It's still allowing to use keyboard in ALL cases, but add an extra step if you are likely to lose a lot of work.

Issue #912, opened 2019-01-30
Missing 'unsaved changes will be lost' confirmation popup


BWPanda:
## Describe the bug
When you have a game open in GD and go to File > Close project, you get a confirmation popup stating that unsaved changes will be lost.

However, if you have a game open and go to File > Open, or File > Create a new project, and proceed through the various popups to open or create a game, you never get the same confirmation popup regarding unsaved changes being lost. This could result in data loss...

Bouh:
For prevent this can we see if the project as be saved before (what function can we use ?) ?
If the project is already saved we display nothing, and if save is not we alert the user ?

BWPanda:
That would be ideal in all situations (creating a new project, opening an existing project and closing the current project).

4ian:
> For prevent this can we see if the project as be saved before (what function can we use ?) ?
> If the project is already saved we display nothing, and if save is not we alert the user ?

That's the ideal solution.
There is no magic function to do it. Any change has to be saved in a flag "isDirty" in the editor to indicate that the project has some changes that are pending save.

Kalovelo:
The simplest workaround would be to alert the user when he clicks on "Create a new project" (like it happens with "Close project") but that wouldn't be ideal nor long-term solution 🤔

Towards the ideal one, I would suggest implementing a more abstract/global state management between SceneEditor and the rest of the components (ElectronMainMenu for example) so we can have a unsavedChanges state. A tool for that would be Redux and even if I see it as a package dependency, I am not able to find any used instance of it.

@4ian What do you think? I could start working on it, setting up a redux store and a "first form" of unsaved changes checking.

4ian:
> The simplest workaround would be to alert the user when he clicks on "Create a new project" (like it happens with "Close project") but that wouldn't be ideal nor long-term solution 🤔

I think that would actually be a perfectly ideal long term solution for reducing the risk of people create new project/opening another one and losing their changes. :)

> @4ian What do you think? I could start working on it, setting up a redux store and a "first form" of unsaved changes checking.

For the ideal solution, I'm all for doing the work of adding a dirty bit to the project or some state in the React components.

*But* I don't want to introduce Redux or a global state management in GDevelop. Mainly because as the name implies they are "global" - and nothing in GDevelop must be global. Any component must be able to work by itself - without relying on a store - just by having the proper props being passed. This is a very explicit approach and make navigating in the codebase as easy as looking at the props and searching for component names.

[...]

PR #1523, opened 2020-03-12
Unsaved changes notification

Kalovelo:
refs #912
Created a React.Context component to check for unsaved changes.
triggerUnsavedChanges is imported in components which change the save file


newIDE/app/src/EventsFunctionsExtensionEditor/index.js
```diff
@@ -498,6 +499,8 @@ export default class EventsFunctionsExtensionEditor extends React.Component<
       this._editorNavigator.openEditor('behaviors-list');
   };
 
+  static contextType = UnsavedChangesContext;
```
4ian: Isn't this approach deprecated? Not sure it plays well with Flow/Typescript too.
We should use a Context.Consumer or a hook to get the context.

newIDE/app/src/MainFrame/index.js
```diff
@@ -1742,8 +1746,14 @@ class MainFrame extends React.Component<Props, State> {
                 this.renameEventsFunctionsExtension
               }
               onRenameExternalEvents={this.renameExternalEvents}
-              onSaveProject={this.saveProject}
-              onSaveProjectAs={this.saveProjectAs}
+              onSaveProject={() => {
+                this.context.sealUnsavedChanges();
```
4ian: Should we do this here or inside this.saveProject/this.saveProjectAs?

I'm worried about other code calling (now or in the future) this.saveProject and so not sealing the unsaved changes.

4ian:
Using a context seems like the proper approach :) I'm unsure about using the `static` field to get the context, I'm not sure it's recommended by React anymore? And is it working with static typing?

4ian:
Also when I say "play nice with Flow", it's not about just not having Flow errors, but making sure that Flow recognise and verify the typing of the context (not treating it as `any`). But we should double check what's the React guidance on this anyway.

Using a Consumer everywhere will certainly be more verbose... but don't see a better solution :/ 
Also, could you check if Storybook stories are till working?

newIDE/app/src/EventsFunctionsExtensionEditor/index.js
Kalovelo: I checked the docs and even if it is not stated that it is deprecated, a Consumer would be a better idea. Refactoring it.

newIDE/app/src/MainFrame/index.js
Kalovelo: I used it here too because when you use the browser editor and you download the file, the message "Project properly saved" (saving flag) doesn't show up so the context doesn't update. I will check a little more the data flow between MainFrame and the saving Operators.

Kalovelo:
Thank you for your comments! :smiley: 

Missed that "any", Refactoring flow :confused:
I checked Storybook, seems to be working properly except the ResourceList/EventsFunctionsList components which throws an error (on master too). 
Am I missing something...?

Kalovelo:
Since no visual component was  created, should I proceed with a new story nonetheless?

4ian:
No need for a new story, just make sure that existing stories are still working - either by making the component robust in case of context not existing, or by providing a fake context.

newIDE/app/src/EventsFunctionsExtensionEditor/index.js
```diff
@@ -56,6 +57,7 @@ type Props = {|
   onBehaviorEdited?: () => void,
   initiallyFocusedFunctionName: ?string,
   initiallyFocusedBehaviorName: ?string,
+  unsavedChangesManagement: UnsavedChanges,
```
4ian: Nitpicking: We should make the variable name consistent with the type. Either
`unsavedChangesManagement: UnsavedChangesManagement,`
or
`unsavedChanges: UnsavedChanges,`

with a slight preference for `UnsavedChanges` as a usually read a type name as "this object represents TYPE" (This object represents UnsavedChanges).

newIDE/app/src/MainFrame/ElectronMainMenu.js
```diff
@@ -95,7 +95,10 @@ class ElectronMainMenu extends React.Component<Props, {||}> {
     );
     ipcRenderer.on(
       ('main-menu-create': MainMenuEvent),
-      event => this._editor && this._editor.openCreateDialog()
+      event =>
+        this._editor &&
+        this._editor.askToCloseProject() &&
```
4ian: I think this might be a good thing to do but we should do another PR for this :)

newIDE/app/src/MainFrame/UnsavedChangesContext.js
```diff
@@ -0,0 +1,60 @@
...
+  hasUnsavedChanges: false,
+  sealUnsavedChanges: () => {},
+  triggerUnsavedChanges: () => {},
+};
+
+const UnsavedChangesContext = React.createContext<UnsavedChanges>(initialState);
+
+export default UnsavedChangesContext;
+
+type State = {
```
4ian: Always use exact object type: {| |} for all your types.

newIDE/app/src/MainFrame/UnsavedChangesContext.js
```diff
@@ -0,0 +1,60 @@
...
+  hasUnsavedChanges: boolean,
+};
+
+type Props = {
```
4ian: Same

newIDE/app/src/MainFrame/UnsavedChangesContext.js
```diff
@@ -0,0 +1,60 @@
...
+  render() {
+    const initialContextValues: UnsavedChanges = {
+      ...this.state,
+      triggerUnsavedChanges: this.triggerUnsavedChanges,
+      sealUnsavedChanges: this.sealUnsavedChanges,
+    };
+
+    return (
+      <UnsavedChangesContext.Provider
+        value={{
```
4ian: Useless spread: `values={initialContextValues}`

newIDE/app/src/MainFrame/UnsavedChangesContext.js
```diff
@@ -0,0 +1,60 @@
...
+    if (this.state.hasUnsavedChanges)
+      this.setState({ hasUnsavedChanges: false });
+  };
+
+  render() {
+    const initialContextValues: UnsavedChanges = {
+      ...this.state,
+      triggerUnsavedChanges: this.triggerUnsavedChanges,
+      sealUnsavedChanges: this.sealUnsavedChanges,
+    };
```
4ian: These are not initial, but the values. Call the as the type name:

```
const unsavedChanges: UnsavedChanges = {
  ...
};
```

newIDE/app/src/MainFrame/index.js
```diff
@@ -1742,8 +1763,12 @@ class MainFrame extends React.Component<Props, State> {
...
               }
               onRenameExternalEvents={this.renameExternalEvents}
-              onSaveProject={this.saveProject}
-              onSaveProjectAs={this.saveProjectAs}
+              onSaveProject={() => {
+                this.saveProject();
+              }}
+              onSaveProjectAs={() => {
+                this.saveProjectAs();
+              }}
```
4ian: Let's cancel these changes?

newIDE/app/src/MainFrame/index.js
```diff
@@ -1863,7 +1889,9 @@ class MainFrame extends React.Component<Props, State> {
             <PlatformSpecificAssetsDialog
               project={this.state.currentProject}
               open
-              onApply={() => this.openPlatformSpecificAssets(false)}
+              onApply={() => {
```
4ian: Let's cancel these changes

newIDE/app/src/ProjectManager/index.js
```diff
@@ -709,18 +711,28 @@ export default class ProjectManager extends React.Component<Props, State> {
                       editingName={
                         renamedItemKind === 'layout' && renamedItemName === name
                       }
-                      onEdit={() => this.props.onOpenLayout(name)}
-                      onDelete={() => this.props.onDeleteLayout(layout)}
+                      onEdit={() => {
```
4ian: Cancel any changes not strictly needed in the PR to ease the review :)

newIDE/app/src/ProjectsStorage/DownloadFileStorageProvider/index.js
```diff
@@ -18,13 +18,18 @@ export default ({
...
-              closeDialog();
-              resolve({ wasSaved: false, fileMetadata });
-            }}
-            project={project}
-          />
+          <UnsavedChangesContext.Consumer>
+            {unsavedChangesManagement => (
+              <DownloadSaveAsDialog
+                onDone={() => {
+                  unsavedChangesManagement.sealUnsavedChanges();
```
4ian: We don't want the file storage to be responsible for, or even aware of the existence of UnsavedChanges. Let's remove this and make MainFrame responsible for this.

Otherwise, we risk never sealing unsaved changes for other provides.
It's a burden that must not be on them

Kalovelo:
Thank you for your comments :smiley: 
I removed the context updates from the file storage. The changes sealing supports Google Drive. Fixed the inaccurate changes and completed the renaming. 
Sorry for inconvenience.
The askToCloseProject on ElectronMenu could be a nice good first issue! :smiley:

4ian:
Sorry for the delay, it's a lot of changes to review. Can you make a list of the main use cases and if they are handled properly?

In particular, does closing the project and re-opening another works properly?

Kalovelo:
Sure, on it!

newIDE/app/src/InstancesEditor/InstancePropertiesEditor/index.js
```diff
@@ -31,8 +33,10 @@ export default class InstancePropertiesEditor extends React.Component<Props> {
       valueType: 'string',
       disabled: true,
       getValue: (instance: gdInitialInstance) => instance.getObjectName(),
-      setValue: (instance: gdInitialInstance, newValue: string) =>
-        instance.setObjectName(newValue),
+      setValue: (instance: gdInitialInstance, newValue: string) => {
+        this.props.unsavedChanges.triggerUnsavedChanges();
```
4ian: I'm thinking that these functions should not have to deal with updating the unsaved changes at all. Their job is to update a value, not to trigger something related to an editor... they might not even be used in an editor in case they are extracted from this.

It's also quite verbose to have to put triggerUnsavedChanges everywhere. Can we instead put this in PropertiesEditor (or something like this), at the place where setValue is called?

4ian: (See `_onInstancesModified` in `PropertiesEditor`, might be a good candidate.)

4ian:
I've left a comment at a place where we can I think factor the calls.
More generally, my main concern will be to find ways to make the changes as lighweight/minimal as possible. Let me explain :)

It's fairly verbose to add the calls everywhere - not in the sense of lines of code, but in the sense of **coupling the components**: a lot of components using the new `UnsavedChanges` have a new requirement of a prop `unsavedChanges` to be now present. For those using the context, they should continue to work without the context provider.

Now the question is: can we even do better?
* With the approach of the context, maybe not really. We should still minimize the number of components using UnsavedChanges and the number of calls to functions. For example, I think that PropertiesEditor can do that, because we don't want to put the burden on the setValue of the schema.
* Or we back off from using a context and thing about another solution. For example, when trying to close a project or close the app, we could serialize the project to a JSON and compare it with another JSON that we serialized when the project was first opened (or last saved).
  * Advantage: we can automatically find any changes! We could even had heuristics saying "you modified this, this and this plus X other changes" by inspecting the changes in the JSON (there must be libraries to do JSON diffs, or we can code it by ourselves) (though this is not mandatory, just a wild idea).
  * Disadvantage: might be a bit costly to do that, well we keep a copy of the project in memory (typically a few kbs to a few mb for large games). We could make a preference to disable this. It's also maybe slightly more complicated and we can't display something like an asterisk in the title bar to indicate that there is a change (because we need to serialize the project, which take some time (a few ms) - not something that can be done after each interaction).

(solutions are not mutually exclusive actually).
Just trying to find how to spread or concentrate the burden of tracking the changes :) Let me know what you think.

Kalovelo:
Thank you for your detailed comments :)

When I was thinking about ways to solve this issue, on of them was exactly that : comparing the saved instance to the updated one but I thought it would be costly and not scalable enough to show the unsaved editor tabs (just like our text editors do).
Even for properties editor, I see that it is getting called by many components such as BehaviourPropertiesEditor / InstancePropertiesEditor / ResourcePropertiesEditor therefore these ones will probably need a connection to the Context.
Since we already keep a project copy, it won't be as costly as it would be in other cases :thinking: 

The current approach needs a somewhat complexity since there are many editors and properties to edit from, yet is probably the one able to introduce more visual "alerts" to the user before he presses the close project button.

So I guess it does depend on UX and  "Clean Code" / Performance?

Kalovelo:
> Even for properties editor, I see that it is getting called by many components such as BehaviourPropertiesEditor / InstancePropertiesEditor / ResourcePropertiesEditor therefore these ones will probably need a connection to the Context.

A reason I followed the static ContextType way was to avoid the prop passing and try to minimize it to the components who are responsible for each change

4ian:
> So I guess it does depend on UX and "Clean Code" / Performance?

Yep, let's try to find a good balance between good UX (enough `triggerUnsavedChanges`) and cleanliness/coupling of the code. At least for the coupling, we should make all code calling the context (or prop) to be robust against the prop or the context being missing.

> A reason I followed the static ContextType way was to avoid the prop passing and try to minimize it to the components who are responsible for each change

So first, my bad I checked and this is not a deprecated API.
This being said, while this is indeed reducing the number of props, it's not super scalable because you can only have one ContextType, and there is not really a good reason to choose this context rather than another - if only one context is used it's good, but otherwise it's a bit "unfair" to other contexts... And also this avoids introducing a third way to use contexts in the codebase:

* The `<Context.Consumer>` is a bit verbose but at least we can have multiple contexts if needed. 
* For components that are functions, using the hook version `React.useContext` is definitely better (a single line, not adding extra indentation). Alas for now most of the codebase is written in classes and will slowly migrate to hooks in the next months/years as we rework them.

> Even for properties editor, I see that it is getting called by many components such as BehaviourPropertiesEditor / InstancePropertiesEditor / ResourcePropertiesEditor therefore these ones will probably need a connection to the Context.

It's indeed used in lots of components, but can we keep the context only in PropertiesEditor itself? (not sure)

> Since we already keep a project copy, a json - comparison approach won't be as costly as it would be in other cases 🤔

Might not be super costly (serializing a project is fairly fast even for big projects) but the `UnsavedChanges` context approach is for sure much much faster if we want to show for example an asterisk in the title bar (because it's just a boolean to check).

We could actually speed up the comparison between serialized JSON by keeping a hash of the JSON. Meaning that when we want to check if the project was changed, instead of doing:

* At loading/save: Serialize the project to JSON -> save in memory the JSON string.
* When trying to close the app/project: Serialize the project to JSON -> compare the JSON string to the one in memory

(which is O(json string) in memory usage and in time of comparison)
we could have: 

* At loading/save: Serialize the project to JSON -> save a hash of the JSON string in memory
* When trying to close the app/project: Serialize the project to JSON -> compare the hash of the JSON string to the one in memory.

which is not forcing us to keep a full copy of the project json in memory, and could slightly increase the comparison speed (but we have to pay for the hashing, so not that much?).

So I think we'll in all cases have to keep a context to signal that there was a change done. Trying to minimise the number of components having it and making components robust in case of context missing.

> (just like our text editors do).

With this, we would for now only show a visual indicator of the change for the whole project right? (Unless we create specific providers for each tab, but that's not for now ^^).

Let me know what you think, thanks for your time so far! :)

Kalovelo:
Drag an object on scene ✅ Triggered!

---
Change an object's property on sceneEditor ✅ Triggered!

---
Change or Add an object's variable ✅ Triggered!

---
Delete an object from sceneEditor Objects panel ✅ Triggered!

---
Change on Events Editor ✅ Triggered!

---
Add a new scene ✅ Triggered!

---
Change Game's properties ✅ Triggered!

---
Change Game's global variables ✅ Triggered!

---
Change Game's Icons ✅ Triggered!

---
Add / Move / Rename / Move Up | Down / Duplicate / Delete External Layouts ✅ Triggered!

---
Add / Move / Rename / Move Up | Down / Duplicate / Delete Scenes ✅ Triggered!

---
Add / Move / Rename / Move Up | Down / Duplicate / Delete Events ✅ Triggered!

---
Add / Move / Rename / Move Up | Down / Duplicate / Delete Functions | Behaviours ✅ Triggered!

---
Change Preferences  ✅ Triggered!

---
Save project locally ✅ Sealed!

---
Save project on GDrive ❌ Cannot test on localhost

---
Download gamefile from browser ❌ (Browser) "Saving" by downloading is not handled like Cloud Saving

---
Create new project ❌ Removed the askToCloseProject call since it was suggested to be implemented in another PR

Mainframe, ProjectManager forceUpdate therefore checking for new changes is not possible and multiple function calls is currently inevitable so I 
created a function (_onModifiedObject, similar to PropertiesEditor's _onModifiedInstance) in ProjectManager & Mainframe in order to reduce this.forceUpdate() & triggerUnsavedChanges calls.

In componenets with a history state object, I used the componentDidUpdate to compare whether or not to trigger the unsavedChanges.

newIDE/app/src/EventsSheet/index.js
```diff
@@ -223,6 +225,11 @@ export default class EventsSheet extends React.Component<Props, State> {
     this.setState({ allEventsMetadata: enumerateEventsMetadata() });
   }
 
+  componentDidUpdate(prevProps: Props, prevState: State) {
+    if (this.state.history !== prevState.history)
```
4ian: This is a great idea 👍

newIDE/app/src/MainFrame/index.js
```diff
@@ -1351,6 +1357,11 @@ class MainFrame extends React.Component<Props, State> {
     }
   };
 
+  _onModifiedObject = () => {
```
4ian: `_onProjectItemModified`?

"Project item" is used in Project Manager to designate something related to the project. "Object" has the meaning in GDevelop of an object in a scene.

newIDE/app/src/ObjectsList/index.js
```diff
@@ -94,8 +94,8 @@ type Props = {|
   onObjectSelected: string => void,
   onObjectPasted?: gdObject => void,
   canRenameObject: (newName: string) => boolean,
-
   getThumbnail: (project: gdProject, object: Object) => string,
+  unsavedChanges: UnsavedChanges,
```
4ian: this seems unused in this file?

newIDE/app/src/ProjectManager/index.js
```diff
@@ -40,6 +40,7 @@ import Fullscreen from '@material-ui/icons/Fullscreen';
 import FileCopy from '@material-ui/icons/FileCopy';
 import AccountCircle from '@material-ui/icons/AccountCircle';
```
4ian: nitpicking: put a newline after all `import`

newIDE/app/src/ProjectManager/index.js
```diff
@@ -663,6 +706,11 @@ export default class ProjectManager extends React.Component<Props, State> {
     /* Do nothing for now, but we could open the first result. */
   };
 
+  _onModifiedObject = () => {
```
4ian: I would also use here `_onProjectItemModified`

newIDE/app/src/ProjectManager/index.js
```diff
@@ -344,7 +346,17 @@ export default class ProjectManager extends React.Component<Props, State> {
...
+
+  _duplicateLayout = (layout: gdLayout, index: number) => {
+    this._copyLayout(layout);
+    this._pasteLayout(index);
+  };
+
+  _duplicateLayout = (layout: gdLayout, index: number) => {
+    this._copyLayout(layout);
+    this._pasteLayout(index);
   };
```
4ian: _duplicateLayout seems... duplicated ;) 
Check also the other _duplicateXXX in this file.

newIDE/app/src/ProjectsStorage/DownloadFileStorageProvider/index.js
```diff
@@ -4,7 +4,6 @@ import * as React from 'react';
 import { type StorageProvider, type FileMetadata } from '../index';
 import DownloadSaveAsDialog from './DownloadSaveAsDialog';
 import SaveAlt from '@material-ui/icons/SaveAlt';
-
```
4ian: nitpicking: Let's keep this new line

newIDE/app/src/SceneEditor/index.js
```diff
@@ -1230,7 +1238,9 @@ export default class SceneEditor extends React.Component<Props, State> {
             project={project}
             layout={layout}
             onClose={() => this.openSceneProperties(false)}
-            onApply={() => this.openSceneProperties(false)}
+            onApply={() => {
```
4ian: Let's cancel these changes

4ian:
> In componenets with a history state object, I used the componentDidUpdate to compare whether or not to trigger the unsavedChanges.

This is a great idea. In a sense the "unsaved changes" are a stripped down version of what would be a global history for the whole project :)

> Mainframe, ProjectManager forceUpdate therefore checking for new changes is not possible and multiple function calls is currently inevitable so I
created a function (_onModifiedObject, similar to PropertiesEditor's _onModifiedInstance) in 
ProjectManager & Mainframe in order to reduce this.forceUpdate() & triggerUnsavedChanges calls.

Sounds good, that's I think the way to go.

Thanks for the recap. That looks good and this is starting to look much more manageable in terms of added complexity :) 
I've added comments but I'll make another pass tomorrow.

Thanks for your work on this so far, this is looking great 👍Will be a nice addition and very appreciated by users.

Bouh:
If the history of 
Modification in extensions
Modification of layers
Modification of object groups
is isn't saved, notification wont pop?

4ian:
Good remark, we probably want to add triggerUnsavedChanges in:
LayersList/index.js
ObjectGroupsList/index.js
EventsFunctionsList/index.js
(or in the parent component(s))

Kalovelo:
Thank you both for your comments!
Thanks for pointing it out @Bouh, I will implement it along with the above fixes

Kalovelo:
Implemented unsavedChanges on the above suggestions
On ObjectsList I had to change the component's forceUpdateList() calls because it would trigger the changesCheck when it was called by an outer component. Now a boolean is used whether if the inner components should call the specific function

4ian:
Looks mostly good! :) 
What do you think of removing the UnsavedChangesContext.Consumer in Storybook, adding some `if (this.props.unsavedChanges) ...` in the components (and make `UnsavedChanges` optional in the `Props`)? This to avoid "polluting" stories.js with tons of `UnsavedChangesContext.Consumer`. The counterpart is that the extra "if" to add in the components themselves.

It's a shame we don't have more components supporting hooks, this would have reduced the need to add UnsavedChanges to Props to 0.

Kalovelo:
Makes sense, sure thing!

4ian:
Alright this looks good! Please have a quick last check to check that nothing seems to be crashing. 
I should be able to merge this then.

Kalovelo:
Sure,
nothing crashed and the stories work properly :smiley:

4ian:
Alright, I fixed a warning in PropertiesEditor. Apart from that, looking good! Merging now 🚀
Thanks for taking the time to do that! 👍Will be a super useful UX improvement that we needed since a bunch of time! 👍

Kalovelo:
Sweet!
Thank your for your time, your patience and giving me the chance :smiley:

Bouh:
⚠ Guys something very important is missing. ⚠
When i do some edit and close the app without save before, the app dont show the message for save and close!
So i loose all my change when the app is closed.

4ian:
The dialog to confirm the closing (CloseConfirmDialog) was not modified in this PR - it's still always showing (we should update it to indicate if there are unsaved changes or not :)).
Remember that in the Electron development version, the dialog to confirm the closing of the app is never shown. Because this would interfere with the live reload of the app when you do a change in the source code :)

Bouh:
> Remember that in the Electron development version, the dialog to confirm the closing of the app is never shown.

Oh, i've never noticed.

PR #1524, opened 2020-03-12
Add exit option to File menu

### LINKED ISSUE #1498: File context menu doesn't have an exit item
## Describe the bug
The GDevelop app does not have an exit option under File, like most apps. This is not just a cosmetic issue but an Accessibility bug.

## To Reproduce
1. Launch GDevelop
2. Choose "file"
3. observe menu.
4. unable to exit or close app.
( Note: Close Project just closes out the project windows, it doesn't close the GDevelop App - which is fine, but there ALSO needs to be an EXIT option)

* If applicable, add screenshots to help explain your problem.
Screenshot: 
UOuGmUOcMu

## Other details
Windows 10, Gdevelop app. Beta 89.

NilayMajorwar:
Added missing exit option to the File menu. Option is available only on
non-MacOS platforms.

User is first prompted with with a dialog box: "Exit GDevelop? Any changes that have not been saved will be lost.". The dialog box does not appear if no project is open.

Closes #1498

### DIFF
--- a/newIDE/app/src/MainFrame/index.js
+++ b/newIDE/app/src/MainFrame/index.js
@@ -459,7 +459,7 @@ class MainFrame extends React.Component<Props, State> {
       .then(() => this.setState({ loadingProject: false }));
   };
 
-  askToCloseApp = (): void => {
+  closeApp = (): void => {
     return Window.quit();
   };
 
--- a/newIDE/app/src/MainFrame/index.js
+++ b/newIDE/app/src/MainFrame/index.js
@@ -459,6 +459,10 @@ class MainFrame extends React.Component<Props, State> {
       .then(() => this.setState({ loadingProject: false }));
   };
 
+  askToCloseApp = (): void => {
+    return Window.quit();
+  };
+
   closeProject = (): Promise<void> => {
     const { currentProject } = this.state;
     const { eventsFunctionsExtensionsState } = this.props;
--- a/newIDE/app/src/MainFrame/ElectronMainMenu.js
+++ b/newIDE/app/src/MainFrame/ElectronMainMenu.js
@@ -18,6 +18,7 @@ type MainMenuEvent =
   | 'main-menu-save'
   | 'main-menu-save-as'
   | 'main-menu-close'
+  | 'main-menu-close-app'
   | 'main-menu-export'
   | 'main-menu-create'
   | 'main-menu-open-project-manager'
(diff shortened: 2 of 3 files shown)


newIDE/app/src/MainFrame/index.js
```diff
@@ -459,6 +459,26 @@ class MainFrame extends React.Component<Props, State> {
...
+    if (!this.state.currentProject) return this.closeApp();
+
+    const { i18n } = this.props;
+    // eslint-disable-next-line
+    const answer = confirm(
+      i18n._(
+        t`Exit GDevelop? Any changes that have not been saved will be lost.`
+      )
+    );
+    if (!answer) return Promise.resolve();
```
4ian: Do we need this? The documentation of Electron's [app.quit states](https://www.electronjs.org/docs/all#appquit) that calling app.quit() will "Try to close all windows" and "This method guarantees that all beforeunload and unload event handlers are correctly executed".

Which means that it will correctly trigger the dialog in https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/UI/CloseConfirmDialog.js no?

You may have been confused by the fact that this dialog is *not* shown in development because of this line: https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/UI/CloseConfirmDialog.js#L24 :)

(otherwise, we would get the message at each reload after changing a source file)

newIDE/app/src/MainFrame/index.js
```diff
@@ -459,6 +459,26 @@ class MainFrame extends React.Component<Props, State> {
       .then(() => this.setState({ loadingProject: false }));
   };
 
+  closeApp = (): void => {
+    Window.quit();
+  };
```
4ian: Don't make a method if it's not used outside and just calling a function. Prefer just writing Window.quit, which is very clear by itself.

4ian:
Thanks for the PR and for working on this :) 
I think `.quit` is the proper solution but that we don't need the confirmation because the quit call will trigger events that will be caught by CloseConfirmDialog.

newIDE/app/src/MainFrame/index.js
NilayMajorwar: Oh, right! Okay, I'll fix this right now.

newIDE/app/src/MainFrame/index.js
NilayMajorwar: Okay, correcting this.

NilayMajorwar:
I've made the necessary fixes to the PR - hope it's fine now.

4ian:
Looks good! I've renamed `askToCloseApp` to `closeApp` because the function itself has no knowledge of the fact that something will later interrupt the closing to show a dialog.

Thanks for working on this :) 👍👍

PR #1525, opened 2020-03-12
Added an Edit Scene Variables Context Menu to the scenes in Project Manager

### LINKED ISSUE #1505: Accessibility: Add a context menu to scenes under project panel for scene variables
## Describe the bug
Please see this thread:

https://forum.gdevelop-app.com/t/creating-editing-scene-variables-feature-request/23191

The ask is that when you right click on a scene from the Project view you see an "Edit Scene Properties" option.

Here's the context menu:
image

this is to make it consistent -  just like the context menu under the Objects panel:

image

The popup you see is just the same as if you right-clicked on a blank space in the scene:

image

Note the "Edit Scene Variables" button there. 

This would allow for the discovery of editing scene variables just like object variables and global variables.

Somtozech:
Added an Edit Scene Properties to the Context Menu in the Project Manager Panel.

The Edit scene Properties pops up a modal for editing scene variables.

Edit scene properties on the context menu
Screenshot (80)

Modal pops up when the menu item is clicked
Screenshot (79)

closes #1505

### DIFF
--- a/newIDE/app/src/ProjectManager/index.js
+++ b/newIDE/app/src/ProjectManager/index.js
@@ -39,7 +39,7 @@ import AddToHomeScreen from '@material-ui/icons/AddToHomeScreen';
 import Fullscreen from '@material-ui/icons/Fullscreen';
 import FileCopy from '@material-ui/icons/FileCopy';
 import AccountCircle from '@material-ui/icons/AccountCircle';
-import SceneEditor from '../SceneEditor';
+import SceneVariablesEditorDialog from '../VariablesList/SceneVariablesEditorDialog';
 
 const LAYOUT_CLIPBOARD_KIND = 'Layout';
 const EXTERNAL_LAYOUT_CLIPBOARD_KIND = 'External layout';
@@ -275,7 +275,7 @@ type State = {|
   projectPropertiesDialogOpen: boolean,
   variablesEditorOpen: boolean,
   extensionsSearchDialogOpen: boolean,
-  name: any,
+  currentLayoutName: null | string,
 |};
 
 export default class ProjectManager extends React.Component<Props, State> {
@@ -306,17 +306,16 @@ export default class ProjectManager extends React.Component<Props, State> {
     }
   }
 
-  _onEditLayoutVariable = (name: any = null) => {
-    this.setState({ name });
+  _onEditLayoutVariable = (currentLayoutName: null | string) => {
+    this.setState({ currentLayoutName });
   };
 
   getLayout = () => {
     const { project } = this.props;
-    const { name } = this.state;
-    if (!project || !project.hasLayoutNamed(name)) return null;
+    const { currentLayoutName } = this.state;
+    if (!project || !project.hasLayoutNamed(currentLayoutName)) return null;
 
-    console.log(project.getLayout(name));
-    return project.getLayout(name);
+    return project.getLayout(currentLayoutName);
   };
 
   _onEditName = (kind: ?string, name: string) => {
@@ -761,18 +760,11 @@ export default class ProjectManager extends React.Component<Props, State> {
                 )
             }
           />
-          {!!this.state.name && (
-            <SceneEditor
+          {!!this.state.currentLayoutName && (
+            <SceneVariablesEditorDialog
+              open
               layout={this.getLayout()}
-              layoutVariablesDialogOpen={true}
-              project={project}
-              initialInstances={this.getLayout().getInitialInstances()}
-              initialUiSettings={serializeToJSObject(
-                this.getLayout().getAssociatedSettings()
-              )}
-              onLayoutVariablesDialogClose={() =>
-                this._onEditLayoutVariable(null)
-              }
+              onEditLayoutVariables={() => this._onEditLayoutVariable(null)}
             />
           )}
           <ProjectStructureItem
--- a/newIDE/app/src/ProjectManager/index.js
+++ b/newIDE/app/src/ProjectManager/index.js
@@ -274,7 +274,7 @@ type State = {|
   projectPropertiesDialogOpen: boolean,
   variablesEditorOpen: boolean,
   extensionsSearchDialogOpen: boolean,
-  layoutVariablesDialogOpen: boolean,
+  name: any,
 |};
 
 export default class ProjectManager extends React.Component<Props, State> {
@@ -287,8 +287,7 @@ export default class ProjectManager extends React.Component<Props, State> {
     projectPropertiesDialogOpen: false,
     variablesEditorOpen: false,
     extensionsSearchDialogOpen: false,
-    layoutVariablesDialogOpen: false,
-    layoutName: '',
+    name: null,
   };
 
   shouldComponentUpdate(nextProps: Props) {
@@ -306,16 +305,16 @@ export default class ProjectManager extends React.Component<Props, State> {
     }
   }
 
-  _onEditLayoutVariable = (open: boolean, name: string) => {
-    this.setState({ layoutVariablesDialogOpen: open, layoutName: name });
+  _onEditLayoutVariable = (name: any = null) => {
+    this.setState({ name });
   };
 
   getLayoutVariables = () => {
     const { project } = this.props;
-    const { layoutName } = this.state;
-    if (!project || !project.hasLayoutNamed(layoutName)) return null;
+    const { name } = this.state;
+    if (!project || !project.hasLayoutNamed(name)) return null;
 
-    return project.getLayout(layoutName).getVariables();
+    return project.getLayout(name).getVariables();
   };
 
   _onEditName = (kind: ?string, name: string) => {
@@ -738,7 +737,7 @@ export default class ProjectManager extends React.Component<Props, State> {
                       }}
                       onEditName={() => this._onEditName('layout', name)}
                       onEditLayoutVariable={() =>
-                        this._onEditLayoutVariable(true, name)
+                        this._onEditLayoutVariable(name)
                       }
                       onCopy={() => this._copyLayout(layout)}
                       onCut={() => this._cutLayout(layout)}
@@ -760,12 +759,12 @@ export default class ProjectManager extends React.Component<Props, State> {
                 )
             }
           />
-          {!!this.state.layoutVariablesDialogOpen && (
+          {!!this.state.name && (
             <VariablesEditorDialog
-              open={this.state.layoutVariablesDialogOpen}
+              open={!!this.state.name}
               variablesContainer={this.getLayoutVariables()}
-              onCancel={() => this._onEditLayoutVariable(false)}
-              onApply={() => this._onEditLayoutVariable(false)}
+              onCancel={() => this._onEditLayoutVariable(null)}
+              onApply={() => this._onEditLayoutVariable(null)}
               title={<Trans>Scene Variables</Trans>}
               emptyExplanationMessage={
                 <Trans>
--- a/newIDE/app/src/ProjectManager/index.js
+++ b/newIDE/app/src/ProjectManager/index.js
@@ -179,6 +179,10 @@ class Item extends React.Component<ItemProps, {||}> {
                 click: () => this.props.onEdit(),
               },
               {
+                label: 'Edit scene Properties',
+                click:() => this.props.onEditLayoutVariable()
+              },
+              {
                 label: 'Rename',
                 click: () => this.props.onEditName(),
               },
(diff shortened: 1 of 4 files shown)


PawBud:
Can you pls tell me what `.vscode/` does?

Somtozech:
@PawBud .vscode contains the workspace editor settings I used to work on the codebase( i set it to use eslint for formatting my code so as to obey eslint rules for the repo).

Somtozech:
For what the PR does. It opens the edit scene variables you see when you click the edit scene variables button in the component ScenePropertiesDialog

newIDE/app/src/ProjectManager/index.js
```diff
@@ -738,6 +749,27 @@ export default class ProjectManager extends React.Component<Props, State> {
                 )
             }
           />
+          {!!this.state.layoutVariablesDialogOpen && (
+          <VariablesEditorDialog
+            open={this.state.layoutVariablesDialogOpen}
+            variablesContainer={project.getVariables()}
```
4ian: You're passing the project variables, so the the dialog will show the global variables :/

4ian:
Thanks for the PR!

> The Edit scene Properties pops up a modal for editing scene variables.

While this can be interesting, at this moment the button should be called... "Edit Scene Variables" :) 

Also, you're passing the project variables, meaning that you're showing the global variables.

=> you need to get more familiar and be more careful about what you implement - because the label is not properly indicating what is opened, and the content is not showing what you describe in the PR description.

PawBud:
> For what the PR does. It opens the edit scene variables you see when you click the edit scene variables button in the component ScenePropertiesDialog

I was trying to add the ScenePropertiesDialog to the context menu but nonetheless Kudos to you

Somtozech:
@4ian to get the current layout variables you have to be able to track the current layout and the functionality is not yet implemented in the Project Manager. Can you suggest a way I could go about this?

newIDE/app/src/ProjectManager/index.js
Somtozech: @4ian Changed it to show the layout variables now. 
Screenshot (82)

Somtozech:
@4ian I added the changes to clearly meet the description of my PR. 

Changed Edit Scene Properties -> Edit Scene Variables
Screenshot (81)

Set it to show layout variables instead of global variables
Screenshot (82)

newIDE/app/src/ProjectManager/index.js
```diff
@@ -282,6 +287,8 @@ export default class ProjectManager extends React.Component<Props, State> {
     projectPropertiesDialogOpen: false,
     variablesEditorOpen: false,
     extensionsSearchDialogOpen: false,
+    layoutVariablesDialogOpen: false,
+    layoutName: '',
```
4ian: layoutName is too generic: this is the name used for what?

newIDE/app/src/ProjectManager/index.js
```diff
@@ -270,6 +274,7 @@ type State = {|
   projectPropertiesDialogOpen: boolean,
   variablesEditorOpen: boolean,
   extensionsSearchDialogOpen: boolean,
+  layoutVariablesDialogOpen: boolean,
 |};
```
4ian: The State you declare is not the same as what you use then

newIDE/app/src/ProjectManager/index.js
```diff
@@ -299,6 +306,18 @@ export default class ProjectManager extends React.Component<Props, State> {
     }
   }
 
+  _onEditLayoutVariable = (open: boolean, name: string) => {
+    this.setState({ layoutVariablesDialogOpen: open, layoutName: name });
```
4ian: You could use a single variable for both, if you make the string nullable :)

newIDE/app/src/ProjectManager/index.js
```diff
@@ -738,6 +760,27 @@ export default class ProjectManager extends React.Component<Props, State> {
                 )
             }
           />
+          {!!this.state.layoutVariablesDialogOpen && (
+            <VariablesEditorDialog
```
4ian: Can we modify SceneEditor so that it's using this, rather than having the same implementation of VariablesEditorDialog used twice?

newIDE/app/src/ProjectManager/index.js
```diff
@@ -738,6 +760,27 @@ export default class ProjectManager extends React.Component<Props, State> {
...
+              open={this.state.layoutVariablesDialogOpen}
+              variablesContainer={this.getLayoutVariables()}
+              onCancel={() => this._onEditLayoutVariable(false)}
```
4ian: Careful, if you don't pass something it will be undefined. Always pass all variables to your functions. JavaScript might allow you not to do it, but this will bite you very hard later.

newIDE/app/src/ProjectManager/index.js
Somtozech: Ok @4ian . So you suggesting I use a single variable to represent the layoutVariablesDialogOpen and layoutName?

newIDE/app/src/ProjectManager/index.js
Somtozech: Can you explain better? I don't really understand.

newIDE/app/src/ProjectManager/index.js
4ian: Yep

newIDE/app/src/ProjectManager/index.js
4ian: _onEditLayoutVariable is expecting 2 parameters in your code, but you're passing only one.

newIDE/app/src/ProjectManager/index.js
Somtozech: The SceneEditor takes in other props that the project manager cannot provide at the moment (would have to rewrite the same implementation for variables/functions it takes as props).

newIDE/app/src/ProjectManager/index.js
Somtozech: Ok. I get it now. Thanks for the explanation :). Looking to add a default parameter to the _onEditLayoutVariable in cases any of the parameters are not defined. What do you think?

4ian: I think we can all have in a single variable, using null to represents something not defined.

newIDE/app/src/ProjectManager/index.js
4ian: I mean SceneEditor has the same thing inside it:

https://github.com/4ian/GDevelop/blob/f1730c239dbaad98b3708d3e94d2586c597e08bb/newIDE/app/src/SceneEditor/index.js#L1235-L1238

This seems like an unecessary duplication of the this. How can you modify SceneEditor so that there is only one VariablesEditorDialog used for this specific case (i.e: the case of modifying scene variables)?

PawBud:
Is the issue solved @Somtozech ?

4ian:
@PawBud please don't comment to ask this, as you can see by yourself :) 
* There are still comments not addressed
* The PR is not merged
* The content of the PR is not addressing all that is needed in the linked issue.

This is creating noise for me and makes me lose time to check again PR that I've already checked ;)

PawBud:
@4ian I am extremely sorry that I wasted your time. I shall not repeat it

Somtozech:
@4ian I changed to a single variable and modified the SceneEditor to take in extra props so as to display the edit variables scene (solves the issue of duplicated code).

PawBud:
@Somtozech I think `Edit Scene Variables` should be in `GameSettings/SceneVariables`

newIDE/app/src/ProjectManager/index.js
```diff
@@ -270,6 +275,7 @@ type State = {|
   projectPropertiesDialogOpen: boolean,
   variablesEditorOpen: boolean,
   extensionsSearchDialogOpen: boolean,
+  name: any,
```
4ian: That's a super generic name for a variable :/ 
Can you think of something more descriptive? Also never use any for types.

newIDE/app/src/ProjectManager/index.js
```diff
@@ -299,6 +306,19 @@ export default class ProjectManager extends React.Component<Props, State> {
     }
   }
 
+  _onEditLayoutVariable = (name: any = null) => {
```
4ian: Never use any, don't use default parameters

newIDE/app/src/ProjectManager/index.js
```diff
@@ -299,6 +306,19 @@ export default class ProjectManager extends React.Component<Props, State> {
...
+    this.setState({ name });
+  };
+
+  getLayout = () => {
+    const { project } = this.props;
+    const { name } = this.state;
+    if (!project || !project.hasLayoutNamed(name)) return null;
+
+    console.log(project.getLayout(name));
```
4ian: Extra console.log, remove them before commiting

newIDE/app/src/SceneEditor/index.js
```diff
@@ -132,6 +133,7 @@ type Props = {|
   onChooseResource: ChooseResourceFunction,
   resourceExternalEditors: Array<ResourceExternalEditor>,
   isActive: boolean,
+  layoutVariablesDialogOpen: boolean,
```
4ian: Do you still need this?

newIDE/app/src/ProjectManager/index.js
```diff
@@ -738,6 +761,20 @@ export default class ProjectManager extends React.Component<Props, State> {
                 )
             }
           />
+          {!!this.state.name && (
+            <SceneEditor
```
4ian: This is still duplicated in SceneEditor/index.js

4ian:
> @Somtozech I think Edit Scene Variables should be in GameSettings/SceneVariables since that is where all the variables are present

No, scene variables are specific to each scene :)

newIDE/app/src/ProjectManager/index.js
Somtozech: I used any because the name variable could be of two types (null ie object and string)

newIDE/app/src/ProjectManager/index.js
Somtozech: ok. what of `currentLayoutName`?

newIDE/app/src/SceneEditor/index.js
Somtozech: Yeah. It set's the current layoutVariablesDialogOpen in the SceneEditor state to use the layoutVariablesDialogOpen in the props as default value if layoutVariablesDialogOpen is passed in as props

newIDE/app/src/ProjectManager/index.js
Somtozech: @4ian Changed it to use the Scene Editor directly. Don't really understand the current duplication you referring to ):. Could you a little more detail about what you meant?

4ian: Oops sorry wrong line commented! :)

4ian: I'm not sure actually to understand why you're putting the entire `SceneEditor` inside ProjectManager 😬
We for sure don't want to do that.

Somtozech: I actually used `VariablesEditorDialog` in my first PR. But then you mentioned "duplication of code and modifying SceneEditor so that there is only one VariablesEditorDialog used for this specific case". I thought what you meant was modifying the `SceneEditor` to suit the use case for `ProjectManager` ):

4ian: Ok I understand better :) 
So `VariablesEditorDialog` is the proper component to use. 

But we want to avoid having to have it both in ProjectManager *and* in SceneEditor if possible. While it's not a huge duplication, there is still some duplicated props like the messages to show on screen (`emptyExplanationMessage`, etc...).

Two possibilities:
* You create a SceneVariablesEditorDialog component, that pass the props that are common like the messages (`emptyExplanationMessage` and co) and just take a few less props. You then use it in ProjectManager and SceneEditor
* OR you put SceneVariablesEditorDialog in the ancestor of SceneEditor and ProjectManager which is... MainFrame. You add a a state to designate which scene (also called layout) has its variables edited. You then pass a prop to SceneEditor and ProjectManager that is a function to call to open the scene variables.

Advantage of first solution: we're not bloating MainFrame.
Advantage of second solution: we don't need a common `SceneVariablesEditorDialog`, because we only have a single place in the code where we use VariablesEditorDialog for a scene - so no duplication of `emptyExplanationMessage` and other props.

Usually I would prefer the second solution, but here because I'd like to avoid bloating too much MainFrame, maybe the first solution is better.

Somtozech:
@4ian I added the SceneVariablesEditorDialog component for editing the scene variables and reused it in SceneEditor and ProjectManager

triptych:
I think  this PR is obsoleted by another merge: https://github.com/4ian/GDevelop/pull/1607

4ian:
Right the "solution" in the PR #1607 :)

PR #1527, opened 2020-03-13
Fix Context Menu appears far away from object/cursor position

### LINKED ISSUE #1492: Misalignment of the context/pop-up menu
## Describe the bug
A UI error, on right click on object while creating a project, pop-up menu is far away from cursor/object. A screenshot is attached showing the error.
Screenshot from 2020-03-04 03-01-45

## To Reproduce
1. Go to 'Create a new project.'
2. Right click on any object.
3. See error.

## Other details
* Using Chrome version 80.0.3987.106-1 on Ubuntu 18.04
* Using web-app of GDevelop version 5

Minisonny:
This is to fix #1492
I edit a bit to make the ContextMenu appears at the position of the cursor just like when we right click inside the preview. I'm not sure if this is what you want, so I just adapt with what already runs well.

### DIFF
--- a/newIDE/app/src/UI/Menu/ContextMenu.js
+++ b/newIDE/app/src/UI/Menu/ContextMenu.js
@@ -39,28 +39,18 @@ class MaterialUIContextMenu extends React.Component {
 
   render() {
     return (
-      <div>
-        <div
-          style={{
-            position: 'fixed',
-            pointerEvents: 'none',
-            left: this.state.anchorX,
-            top: this.state.anchorY,
-          }}
-        />
-        <Menu
-          open={this.state.open}
-          anchorPosition={{left: this.state.anchorX, top: this.state.anchorY}}
-          anchorReference={'anchorPosition'}
-          onClose={this._onClose}
-          TransitionComponent={Fade}
-          {...this.menuImplementation.getMenuProps()}
-        >
-          {this.menuImplementation.buildFromTemplate(
-            this.props.buildMenuTemplate()
-          )}
-        </Menu>
-      </div>
+      <Menu
+        open={this.state.open}
+        anchorPosition={{ left: this.state.anchorX, top: this.state.anchorY }}
+        anchorReference={'anchorPosition'}
+        onClose={this._onClose}
+        TransitionComponent={Fade}
+        {...this.menuImplementation.getMenuProps()}
+      >
+        {this.menuImplementation.buildFromTemplate(
+          this.props.buildMenuTemplate()
+        )}
+      </Menu>
     );
   }
 }
--- a/newIDE/app/src/UI/Menu/ContextMenu.js
+++ b/newIDE/app/src/UI/Menu/ContextMenu.js
@@ -41,7 +41,6 @@ class MaterialUIContextMenu extends React.Component {
     return (
       <div>
         <div
-          ref={element => (this.anchorEl = element)}
           style={{
             position: 'fixed',
             pointerEvents: 'none',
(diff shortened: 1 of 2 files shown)


4ian:
Can you add screenshot before/after? :)

Minisonny:
Sure! I have seen some picture in the issues area and look like on some platform, it works well. But I am running on Chrome and it gives me something like this, just like the original report of the issue
image
image

Even when I try the online version from the page of GDevelop, the menu tends to fly away from the clicked object. So I changed a bit so that it would look like this
image
image

I do this because it looks similar to when I right click inside the scene editor. It seems to me that only the browsers get this problem.

:)

4ian:
Looks better indeed! My new question would be about the fix itself:

`event.clientX - dimensions.left`

=> Can you explain why it's working better? :)

Also, this changes will make changes on everything that is showing a context menu.
=> Can you make a few other screenshots of context menus, like these ones:
image

Minisonny:
Oh gosh I didn't notice that two button, oops :( I'm not sure but when the menu is rendered when those two buttons are clicked, it is rendered with left and top with respect to thee page. But when I click a box with the 3 dots option like this
image
the menu top and left are with respect to that 3 dots button.

I'm gonna find out the problem again :(

Minisonny:
Hello, I changed the open function a bit in `ElementWithMenu.js` like this
image
This would let Toggle grid and Zoom button be the same as before.
image
image
image
image
image
Is this OK? Thank you for your time 😄

4ian:
Can you explain the difference done by the if/else? What is this doing exactly?

Minisonny:
The `_wrappedElement` becomes is a HTML button (3 dots button) when I right click this
image
and is an `IconButton` when I click the buttons like this
image
The above two buttons take this 
image
as anchor, while in the other case, the 3 dots button is taken instead. The `dimensions` carries `top` and `left` with respect to the viewport. That explains why the original code `Math.round(dimensions.top + dimensions.height)` moves the menu so far away down from the clicked object (or the 3 dots button to be exact). So I just divide into two cases so that the menu dropped from 3 dots button doesn't fly away.
I'm not sure why this happens, probably with the `anchorEl` in `ContextMenu`?

4ian:
> But it works well on Desktop version.

Yep the menu on the desktop version just takes the position x and y of the menu.

> So I just divide into two cases so that the menu dropped from 3 dots button doesn't fly away.

Should we remove the `anchorEl`? I'm afraid that the `if` is not very explicit, it's hidden in the component implementation?

Minisonny:
Okay, so I restore the to the original `ElementWithMenu.js` and made some changes with `ContextMenu.js`. I removed the `anchorEl` and replaced with these two properties
`anchorPosition={{left: this.state.anchorX, top: this.state.anchorY}}
anchorReference={'anchorPosition'}`
This is how it works:
image
image
image
image
image
Pretty the same with the desktop version 😄 Moreover, I tried to remove two lines
`left: this.state.anchorX,
top: this.state.anchorY,`
from the `style` of the above `div`. It still works well somehow :)

4ian:
> replaced with these two properties

This looks great! I think that should have been what we wanted to do originally with the `div`/anchorEl. This solution is much better.

> from the style of the above div. It still works well somehow :)

This is because this div... is now totally useless :) It was used previously as a fake anchor element. Now that we're positioning the menu with x/y coordinates, there is no need for it.

I've pushed a commit to remove the useless divs. I've also run the autoformatting (`npm run format`), remember to do it.

Merging as this works well! Thanks for working on this 👍

PR #1536, opened 2020-03-16
[WIP] Add support for side-separate modifier keys

NilayMajorwar:
This is a WIP PR for adding support for side-separate modifier keys.

Currently keys like LShift and RShift are treated as the same key, since they have the same keycode. This PR adds support for differing between these pairs of keys, by taking in `event.location` into account as well.

### DIFF
--- a/GDJS/Runtime/events-tools/inputtools.js
+++ b/GDJS/Runtime/events-tools/inputtools.js
@@ -95,7 +95,7 @@ gdjs.evtTools.input.keysNameToCode = {
     "x": 88,
     "y": 89,
     "z": 90,
-    
+
     "Num0": 48,
     "Num1": 49,
     "Num2": 50,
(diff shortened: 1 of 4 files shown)


NilayMajorwar:
@4ian I did an independent test for checking performance for string-keyed hashmaps. `event.code` turned out to be over 4 times slower. 

Right now, I've just used a hacky workaround by putting in the keycodes of right-sided keys as negative (so the key code of LShift is `16` and key code of RShift is `-16`), and add an if statement for right-modifier keys. It works, and performance does not take a huge hit this way. But I'm not sure how clean it is. Should I try something else?

4ian:
> But I'm not sure how clean it is. Should I try something else?

Yes let's try something else. This is not entirely clear (imagine the reader of this code: "why do we have negative key code??") and not very extensible (what if there is a special key on some keyboards that can have 3 locations?).
Make sense and work though :) But I'm sure we can find an alternative

> Also, do I have to make changes to the Cocos2d renderer too (here)?

Good catch, yes please! :)

PR #1544, opened 2020-03-17
Adding the Scene Properties Dialog in the Context Menu

PawBud:
This is a potential fix for the issue : _#1505_


PawBud:
@4ian As you asked me to do it, I have filed this PR, I will be highly obliged if you can tell me where is it that I am going wrong

Issue #1472, opened 2020-02-28
Object set as Global Object does not appear in scene already open


triptych:
## Describe the bug
When an object is set as a global object, it does not appear in already open scenes until those scenes are either closed and reopened, or some kind of search happens in the object list.

## To Reproduce
1. Create a game with two scenes - such a A and B.
2. Open both scene A and scene B
3. In scene A create an object, say a sprite.
4. observe that this object shows up in the object list for scene A. But not in scene B. Obviously.
5. Click on the three dots to the right of the object.
6. Choose Set as a Global Object.
7. click ok to dismiss dialog.
8. Observe that the object now looks like a global object.
9. Click on the scene B tab.
10. Observe that the object does not appear in the object list at all in scene B.
11. go down to the search field and enter a character. Then delete that character.
12. Now suddenly the global object sprite appears in scene B's object list.
Note: this only happens when you already have the second scene's tab open before you set the sprite object in the first scene to be a global object.

* If applicable, add screenshots to help explain your problem.

## Other details
* Windows 10 laptop
* Which version of GDevelop are you using? GDevelop 5.0.0-beta88 Desktop  
* This is misleading to the user as they think the object is not global, and the point is to make it available to other scenes. Having to close and open a scene, or do a dummy search just to make the global object show up seems odd.

4ian:
Yes the UI is not properly refreshing more generally when global objects are changed. Would need some clever solution to trigger a refresh of the object lists - maybe just refresh the object lists when a tab is activated?

triptych:
The object list should listen for some custom event emitted by the "Set as Global Object" action and refresh itself ...

jimil749:
I would like to work on this!

4ian:
Please go ahead :) Though this might need some careful thinking. We traditionally don't use JavaScript "events" in GDevelop codebase, but we pass information through React props/callbacks. You might want to specify multiple props in components to say "global objects were updated"...

OR maybe better and cleaner, just force a refresh of the objects list when a tab is activated :) Again needs some experiment to find the best solution :)

jimil749:
@4ian, refreshing the objectList on the change of tab does seem a better approach. Did some digging and found that the tabs are changed by changeCurrentTab() in EditorTabsHandler.js , which is being called in index.js (Mainframe folder) inside _onChangeEditorTab().
So, can we just refresh/re-render the objectList when the above method is called?

4ian:
[...]

PR #1545, opened 2020-03-17
Global object visibility fix

jimil749:
Solve for #1472 .

GDevelop

Sorry for the quality, but I hope it does the needful :)


jimil749:
@4ian can you please review?

newIDE/app/src/MainFrame/index.js
```diff
@@ -1602,6 +1603,14 @@ class MainFrame extends React.Component<Props, State> {
     editorTab.editorRef.updateToolbar();
   }
 
+  forceUpdateObjectsList() {
+    const editorTab = getCurrentTab(this.state.editorTabs);
+    if (
+      editorTab.key.startsWith('layout ' + editorTab.name)
```
4ian: This is risky as a change in this key later could silently break this. Can you think of another way of doing this?

Bouh: I've was an idea in a previous commit, but not used.
Add a props type for earch type of tab editor.
This is an idea, maybe another way existing, for get the type without this addition.
https://github.com/4ian/GDevelop/pull/1429/commits/27331594e878eed7aae543128286888e86d53c2a

jimil749: Well yes, it had no provision to know the type of the editor, hence had to use this work-around as I think key uniquely determines each type of editor! :) 

Thanks @Bouh for the idea, I'll try and implement it.

Bouh:
Another way can be reload the editor when the state change from active false to true.
Like this only the editor used is refresh :)

4ian:
> Another way can be reload the editor when the state change from active false to true.
Like this only the editor used is refresh :)

Would be great to try this. I'm not a big fan of adding a "type" prop, because it's also fragile. For example, you wrote "eventsEditor" instead of "EventsEditor". This leads to hard to debug code.

Prefer if possible a solution where you don't give the responsibility of knowing what to refresh to MainFrame. MainFrame is just here to "host" the editors. We don't want it to be responsible to look at them and asking them to refresh if they are a particular type, or if blabla or if blabla... the poor MainFrame is already too big and underpaid. It's doing too much.

Instead, give the responsibility to the editor: MainFrame can tell them "please refresh" when they are activated (i.e: shown on screen). Then it's the editor responsibility to implement a refresh method that will update what is needed - some won't do anything, others will do stuff.
Better: this can be adapted to any other editor in the future without changing a single line of code in MainFrame!

jimil749:
I did not get everything of that. 
So, do you mean to say that we need a method in MainFrame which checks if an editor is acivated and upon being active it will ask/call a method in respective editor to perform the refresh?
This way we will have add a refresh method in each editor. Some won't do anything as of now (return null) and some (SceneEditor here) will update the required component. This way the refresh is performed by the respective editors and not the MainFrame
 Is this how it should be?

jimil749:
Also we know when a tab editor is being shown to the screen (_onEditorTabActive) does that. So maybe we can use that to perform a refresh for the editors?

4ian:
> So maybe we can use that to perform a refresh for the editors?

Yes that sounds like a good idea!

jimil749:
Added a performRefresh() function for each editor, which is called when an editor is being activated. It can perform updates as per the editor requirements. Here, it just performs the refresh of objectList inside the SceneEditor

jimil749:
@4ian can you check?

4ian:
I check once a day, prefer to ping me if no answer after more than 24 hours :)

newIDE/app/src/MainFrame/Editors/DebuggerEditor.js
```diff
@@ -17,6 +17,10 @@ export default class DebuggerEditor extends BaseEditor {
     if (this.editor) this.editor.updateToolbar();
   }
 
+  performRefresh() {
+    return null;
```
4ian: Let's remove the `return null` and just write `// No updates to be done. `

newIDE/app/src/MainFrame/index.js
```diff
@@ -1611,6 +1611,7 @@ class MainFrame extends React.Component<Props, State> {
 
   _onEditorTabActive = (editorTab: EditorTab) => {
     this.updateToolbar();
+    editorTab.editorRef.performRefresh();
```
4ian: nitpicking: can we call this `forceUpdateEditor`?

"forceUpdate" is a hint at the forceUpdate from React. Refresh has a slightly different meaning (refresh would be something like making network requests to "refresh" your profile).

4ian: Let's also add here a comment why we're doing this:
`// Ensure the editors shown on screen are updated. This is for example useful if global objects have been updated in another editor.`

4ian:
This approach looks good! Added a few nitpickings :)

newIDE/app/src/MainFrame/index.js
jimil749: Will do that straight away! :)

4ian:
Tests are failing because Flow is signaling a potential bug: 
image

I'm fixing this but be sure to look at it next time :)

jimil749:
I'm sorry, was a bit too careless. Will take care next time on. Thanks for the changes though! :)

4ian:
Thanks for your work on this :)

PR #1546, opened 2020-03-17
[WIP] Don't add parameters when they are empty Fixes #1533

### LINKED ISSUE #1533: Multitouch default expression incorrect to the syntax checker
## Describe the bug
In a position field, I picked the _Touch X position_ from the expression button, I left the layer and camera fields empty, and it triggers an error.

image

Win10 64-bit
GD5b89

ozata:


newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/FormatExpressionCall.js
```diff
@@ -14,7 +14,8 @@ const filterOutCodeOnlyParameters = (
 
     return (
       index < parametersCount &&
-      !expressionMetadata.getParameter(index).isCodeOnly()
+      !expressionMetadata.getParameter(index).isCodeOnly() &&
+      parameter !== ''
```
4ian: This is unfortunately a bit too naive.
If you insert a new expression like MouseX, you'll be asked for the layer name and camera number (though it's not used currently in GDevelop because all layers have a single camera).

If I enter nothing for the layer, but a number for the camera (0), this will result in the current situation in:
`MouseX(,0)` - which is not valid, but this is ok because what you entered is not valid (if you enter a camera number, you must enter a layer name).

With your changes, this will now make:
`MouseX(0)` which looks valid but is not, because you're passing a number instead of a layer name.

(Actually, note that GDevelop is forgiving and can understand MouseX(,0)... but MouseX(0) is definitely wrong)

Re-read my comment for the proper logic: https://github.com/4ian/GDevelop/issues/1533#issuecomment-599204949

newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/FormatExpressionCall.js
```diff
@@ -14,7 +14,8 @@ const filterOutCodeOnlyParameters = (
 
     return (
       index < parametersCount &&
-      !expressionMetadata.getParameter(index).isCodeOnly()
+      !expressionMetadata.getParameter(index).isCodeOnly() &&
```
4ian: You've added your changes in a function named "filterOutCodeOnlyParameters".
What does this name implies? What does it describe? Is this name still adapted after you add your changes?
Better: could you make your changes in a new function, with a name describing what you're doing? :)

ozata: I tried to solve w/o creating a function, but I'll do that

newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/FormatExpressionCall.js
ozata: So we have to pass the parameters that comes after required parameters.
What I understand is parameter.isOptional() && parameter.isEmpty() is the condition to not add a parameter to the list. I'll use the app more and learn the behaviour

newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/FormatExpressionCall.js
ozata: There are 3 conditions and all of them uses filterOutCodeOnlyParameters, not sure if I should make a more abstract function that handles all cases.

4ian: You can make a `filterParameters` that calls the necessary functions (filterOutCodeOnlyParameters and the new one)

4ian:
What is the state of this PR?

ozata:
> What is the state of this PR?

It is not ready yet, I have spent a lot of time so I backed of a little bit, I will look at it again tonight. Why? :D

4ian:
You pushed a commit so I was thinking it was ready for review :) No hurry, I'll mute this, please ping me if you have something to review later.

4ian:
Closing as there is no activity on this. Still thanks for your efforts on this :)

PR #1548, opened 2020-03-18
Add shortcut for export project

### LINKED ISSUE #1547: Missing keyboard shortcut for exporting project
## Describe the bug
Not exactly a bug but a possible enhancement. There is currently no keyboard shortcut for exporting a project. It would be nice if there was one.

## To Reproduce
1. Go to File Menu
2. Observe there's no shortcut for Export

## Screenshot
Screenshot 2020-03-18 at 10 52 48 AM

## Other details
* Mac OSX Mojave
* Desktop app

pranavmodx:
New shortcut key for Export : Command/Control + E
Fixes #1547

### DIFF
--- a/newIDE/app/src/MainFrame/ElectronMainMenu.js
+++ b/newIDE/app/src/MainFrame/ElectronMainMenu.js
@@ -175,6 +175,7 @@ class ElectronMainMenu extends React.Component<Props, {||}> {
         { type: 'separator' },
         {
           label: i18n._(t`Export (web, iOS, Android)...`),
+          accelerator: 'CommandOrControl+E',
           onClickSendEvent: 'main-menu-export',
         },
         { type: 'separator' },


4ian:
Can you write the reasoning behind this shortcut, if other similar software/IDE/editors have a similar shortcut, and if there is a risk of conflict between this and another well-known shortcut in other apps/editors?

pranavmodx:
This shortcut key is used for different purposes in different softwares. But I found that softwares like Adobe XD and Autodesk Maya use the same for exporting a project. (I saw it [here](https://defkey.com/what-means/ctrl-e)).
Any idea why the build might be failing? The logs aren't helping much.

4ian:
> But I found that softwares like Adobe XD and Autodesk Maya use the same for exporting a project. (I saw it here).

Super interesting link! :) 
Looks good then, in the worst case we can change it later.

Thanks! Tests were failing because of master having bad tests.

PR #1552, opened 2020-03-18
Fix sentences for video & clean whitespace

Bouh:
Fix sentences for video & clean whitespace.
Tested and works on electron and web-app.

### DIFF
--- a/Extensions/Video/JsExtension.js
+++ b/Extensions/Video/JsExtension.js
@@ -380,8 +380,6 @@ module.exports = {
         "JsPlatform/Extensions/videoicon16.png"
       )
       .addParameter("object", _("Video object"), "VideoObject", false)
-      .addParameter("operator", _("Modification's sign"), "", false)
-      .addParameter("expression", _("Opacity (0-255)"), "", false)
       .useStandardOperatorParameters("number")
       .getCodeExtraInformation()
       .setFunctionName("setOpacity")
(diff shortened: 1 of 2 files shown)


Extensions/Video/JsExtension.js
```diff
@@ -386,9 +382,9 @@ module.exports = {
       .addParameter("object", _("Video object"), "VideoObject", false)
       .addParameter("operator", _("Modification's sign"), "", false)
       .addParameter("expression", _("Opacity (0-255)"), "", false)
```
4ian: You must remove these too right?

Extensions/Video/JsExtension.js
```diff
@@ -433,9 +429,9 @@ module.exports = {
       .addParameter("object", _("Video object"), "VideoObject", false)
       .addParameter("operator", _("Modification's sign"), "", false)
       .addParameter("expression", _("Speed "), "", false)
```
4ian: Same, should be removed?

Bouh:
Well right, just how are define the description like `Opacity (0-255)` because indeed when i delete the lines in previous commit, the description is stay present.

4ian:
These are added automatically when calling UseStandardOperatorParameters/UseStandardRelationalOperatorParameters: 
https://github.com/4ian/GDevelop/blob/master/Core/GDCore/Extensions/Metadata/InstructionMetadata.cpp#L91

4ian:
Looks good to go now!

PR #1553, opened 2020-03-19
Made L/R Ctrl/Shift/Alt/Meta recognizable as separate keys

### LINKED ISSUE #820: Shift keys doesn't work
## Describe the bug
LShift doesn't work, and RShift works but with the Shift key on left of the keyboard

## To Reproduce
Try to assign LShift on an action > nothing happen
Try to assign RShift on an action > RShift dosen't work, but LShift work instead RShift

## Other details
GD5 v61 Windows and editor.gdevelop-app.com

walnutdust:
Closes #820 

Allows for users to designate the Left and Right variants of Shift, Meta, and Control keys as separate keys by using `event.code`, which provides an unique representation of keys based on their physical location on the keyboard. Passes test suites and flow.

Main files changed:
1. `GDJS/Runtime/events-tools/inputtools.js` - Updated `keysNameToCode`.
1. `Extensions/PlatformBehavior/platformerobjectruntimebehavior.js` - Lines 68 - 75 (Updated codes), 262 - 264 (Added in right shift for jumping).
1. `Extensions/TopDownMovementBehavior/topdownmovementruntimebehavior.js` - Lines 100 - 105 (Updated codes).

##Additional Points
1. `event.code` is based on the physical key, rather than what it's binded to - so there might be some confusion especially for people who have different key bindings from the keyboard. Nevertheless, this approach was chosen because it was the successor of `event.keyCode` (now deprecated) and exhibits the same behavior, which will be useful for users. 
1. As seen, this will affect Extension behavior - do we have to worry about this?
1. Apologies about the massive formatting - Ran `npm run format` but it didn't seem to change, perhaps due to the prettier config not being explicitly defined? I could do up a separate PR with a prettier config based on the code, and maybe create a pre-commit hook if that is desirable?

This is my first time contributing, so please forgive any mistakes in my processes along the way and help guide me!


4ian:
Thanks for the PR :) 

> Nevertheless, this approach was chosen because it was the successor of event.keyCode (now deprecated) and exhibits the same behavior, which will be useful for users.

Ok make sense! It's probably fine then, event.code seems better :) 

> As seen, this will affect Extension behavior - do we have to worry about this?

It's probably fine - though we should ensure the same keys are working as before.

> Apologies about the massive formatting

I'll run auto formatting in `master` on the files you changed to try to reduce the number of changes to review.
(Or you can make a separate PR with just the code formatting first.)

> perhaps due to the prettier config not being explicitly defined

There is no npm run format for extensions and GDJS - this could be added in a separate PR :)

4ian:
I've looked at the PR and seems like the formatting is **not made** Prettier. I think it's made by VS Code or some other extension. You can tell because there are spaces before parenthesis in `function ()`, which is something that Prettier never does. 
Make sure to run your reformatted files trough Prettier and not another code formatter.

I know it's a pain, but in case of these files that are not autoformatted, better have a PR to format them first. What we can do is:
- Run Prettier on your files (there will still be a lot of changes in your PR).
- Then me or you create another PR from master, that is just reformatting the file (at this point, there will still be a lot of changes shown in your PR).
- I merge the PR to autoformat the files.
- Know master has the file autoformatted, and your PR too => your PR will only show the changed lines - which will be much easier to review :)

Does that work for you?

4ian:
Alright I went ahead and ran Prettier on the files in master. GitHub is now telling that there are conflicts between your files and master. To fix this, just run Prettier on all your files.

walnutdust:
@4ian Thank you for helping me! I've formatted the code on my side and rebased it - would you have a look?

4ian:
Looks good! :) I think we're good to go with this 👍

4ian:
@walnutdust Sorry for the delay! I found that this won't work when exporting a game with the experimental Cocos2d-JS export. It's a bit unused/deprecated, but would nevertheless be interesting to keep working. Here, this will break the keyboard inputs - because Cocos2d-JS is using `keyCode`s :( And there is no way to make it use `event.code`.

I'm not sure how to exactly approach this, maybe add some compatibility code in inputmanager.js

At least, if you search for `onKeyPressed`/`onKeyReleased` in the codebase, you can see that it's used in runtimescene-cocos-renderer - but now incorrectly... Do you have an idea?

walnutdust:
Thanks for alerting me! Perhaps an approach that could be taken is to retain `keyCode`, update the non-coco input managers to pass in `event.location` (which contains L/R/Numpad information) as well, and add default behavior if a `location` is not passed in. An advantage that might come out of this is that it would for users who want both left and right shift keys for example to be able to define it with a general shift behavior rather than having to define both separately.

Thoughts?

4ian:
> update the non-coco input managers to pass in event.location (which contains L/R/Numpad information) as well, and add default behavior if a location is not passed in

Sounds looks a good idea!

> An advantage that might come out of this is that it would for users who want both left and right shift keys for example to be able to define it with a general shift behavior rather than having to define both separately.

Yep this true - we might or might not want to expose such a capability, but it's good to have it (in case someone wants it for backward compatibility, or just for the sake of convenience... in which case though it would be better to create a [more general Input Mapper](https://trello.com/c/Lf12F37m/146-add-an-input-mapper-also-called-input-manager)).

So yeah, an optional `event.location` parameter sounds great 👍 
Sorry for the re-work, and thanks for investigating this 👍

walnutdust:
@4ian - Sorry for the delay. Would you take a look?

There's a small special case for handling Numpad numbers since their default `event.location` is numpad rather than standard. Also exposed a 'Any Shift'/ 'Any Control'/ 'Any Alt' option when choosing conditions via the input mapper - I could remove that if we want to hide it?

The default behavior when no location is passed in should map to the left/standard variant.

GDJS/Runtime/inputmanager.js
```diff
@@ -41,22 +41,52 @@ gdjs.InputManager.MOUSE_RIGHT_BUTTON = 1;
 /** @constant {number} */
 gdjs.InputManager.MOUSE_MIDDLE_BUTTON = 2;
 
+/**
+ * Returns the location-modified keyCode.
+ *
+ * @private
+ */
+gdjs.InputManager.prototype.getModifiedKeyCode = function(keyCode, location) {
```
4ian: We could call this _getLocationAwareKeyCode

"Modified" is a bit generic, LocationAware hints that the keycode is a combination of the original keycode and a location :)

GDJS/Runtime/inputmanager.js
```diff
@@ -41,22 +41,52 @@ gdjs.InputManager.MOUSE_RIGHT_BUTTON = 1;
...
+    // If it is a numpad number, do not modify it.
+    if (96 <= keyCode && keyCode <= 105) {
+      return keyCode;
+    }
+
+    if (
+      location === KeyboardEvent.DOM_KEY_LOCATION_RIGHT ||
+      location === KeyboardEvent.DOM_KEY_LOCATION_NUMPAD
+    ) {
+      return keyCode + 1000;
```
4ian: What do you think of `keyCode + 1000 * location`?

So that:
standard key = usual key code
left key = keycode + 1000
right key = keycode + 2000
numpad (with the exception of numpad 0-9) = keycode + 3000

This seems an interesting encoding because we can deduce the location from the resulting keycode.

GDJS/Runtime/inputmanager.js
```diff
@@ -41,22 +41,52 @@ gdjs.InputManager.MOUSE_RIGHT_BUTTON = 1;
...
+  }
+
+  // Keys with unspecified locations are treated as standard keys.
+  return keyCode;
+};
+
 /**
  * Should be called whenever a key is pressed
  * @param {number} keyCode The key code associated to the key press.
+ * @param {number} location The location of the event.
```
4ian: Let's document that 0 = standard key, 1 = left key, 2 = right key, 3 = numpad key.

GDJS/Runtime/inputmanager.js
```diff
@@ -41,22 +41,52 @@ gdjs.InputManager.MOUSE_RIGHT_BUTTON = 1;
...
  */
-gdjs.InputManager.prototype.onKeyPressed = function(keyCode) {
-  this._pressedKeys.put(keyCode, true);
-  this._lastPressedKey = keyCode;
+gdjs.InputManager.prototype.onKeyPressed = function(keyCode, location) {
+  var modifiedKeyCode = this.getModifiedKeyCode(keyCode, location);
```
4ian: `var locationAwareKeyCode` if we follow my other suggestion

GDJS/Runtime/inputmanager.js
```diff
@@ -69,22 +99,54 @@ gdjs.InputManager.prototype.getLastPressedKey = function() {
 
 /**
  * Return true if the key corresponding to keyCode is pressed.
- * @param {number} keyCode The key code to be tested.
+ * @param {number} modifiedKeyCode The key code (with location modifiers) to be tested.
```
4ian: locationAwareKeyCode The location-aware key code to be tested.

GDJS/Runtime/inputmanager.js
```diff
@@ -69,22 +99,54 @@ gdjs.InputManager.prototype.getLastPressedKey = function() {
...
  */
-gdjs.InputManager.prototype.isKeyPressed = function(keyCode) {
-  return (
-    this._pressedKeys.containsKey(keyCode) && this._pressedKeys.get(keyCode)
-  );
+gdjs.InputManager.prototype.isKeyPressed = function(modifiedKeyCode) {
+  if (modifiedKeyCode > 0) {
```
4ian: For now I would suggest we avoid the "Any" key because I would like to avoid adding complexity on top of the already "complex" location-aware key code.

While it's not probable that in the future keyboards will change a lot, there will still be potential new APIs, or changes in the APIs, and I'd like to make our implementation as simple as possible and close to the notion of having a key code and a location.

GDJS/Runtime/inputmanager.js
```diff
@@ -69,22 +99,54 @@ gdjs.InputManager.prototype.getLastPressedKey = function() {
...
-  );
+gdjs.InputManager.prototype.wasKeyReleased = function(modifiedKeyCode) {
+  if (modifiedKeyCode > 0) {
+    return (
+      this._releasedKeys.containsKey(modifiedKeyCode) &&
+      this._releasedKeys.get(modifiedKeyCode)
+    );
+  }
+
+  var possibleKeyCodes = [-modifiedKeyCode, 1000 - modifiedKeyCode];
```
4ian: same, let's avoid for now to handle the "any keys"

GDJS/Runtime/events-tools/inputtools.js
```diff
@@ -127,14 +127,21 @@ gdjs.evtTools.input.keysNameToCode = {
   Numpad8: 104,
   Numpad9: 105,
 
-  RControl: 17,
-  RShift: 16,
-  RAlt: 18,
-  LControl: 17,
   LShift: 16,
+  RShift: 1016,
```
4ian: With my proposal (see other comments), LShift becomes 1016, RShift 2016 (if I'm not mistaken).

GDJS/Runtime/inputmanager.js
walnutdust: Sounds good - will do!

GDJS/Runtime/inputmanager.js
4ian: To finalize this and be compatible with libraries that are not able to give us the location (i.e: Cocos2d-JS), we'll need to add special casing to handle these (much like you already did for the numpad number, which are not consistent):

If location is undefined, Shift (16), Ctrl (17), Alt (18) and System (91) should be converted to their left variant (this is a choice we're making) OR we choose to simulate both keys to be pressed, in which case, in `onKeyPressed`/`onKeyReleased`, we handle this special case of location being undefined.

Up to you! Might be simpler to just convert them to their left variant.

GDJS/Runtime/inputmanager.js
walnutdust: Sounds good!

GDJS/Runtime/inputmanager.js
walnutdust: Just fixed it - would you have a look?

GDJS/Runtime/inputmanager.js
```diff
@@ -42,21 +42,65 @@ gdjs.InputManager.MOUSE_RIGHT_BUTTON = 1;
...
+  if (location) {
+    // If it is a numpad number, do not modify it.
+    if (96 <= keyCode && keyCode <= 105) {
+      return keyCode;
+    }
+
+    return keyCode + 1000 * location;
+  }
+
+  if (gdjs.InputManager.DEFAULT_LEFT_VARIANT_KEYS.includes(keyCode)) {
```
4ian: Let's use indexOf to preserve IE11 compatibility as much as possible.

walnutdust: Done!

GDJS/Runtime/inputmanager.js
```diff
@@ -42,21 +42,65 @@ gdjs.InputManager.MOUSE_RIGHT_BUTTON = 1;
...
+  this._lastPressedKey = locationAwareKeyCode;
 };
 
 /**
- * Should be called whenever a key is released
+ * Should be called whenever a key is released. The location corresponds to
+ * KeyboardEvent.location, which should be 0 for standard keys, 1 for left keys,
+ * 2 for right keys, and 3 for numpad keys.
  * @param {number} keyCode The key code associated to the key release.
+ * @param {number} location The location of the event.
```
4ian: Use equal: `{number=}` to note an optional parameter 
(It's a JSDoc notation: https://jsdoc.app/tags-param.html#optional-parameters-and-default-values also understood by Typescript.)

4ian:
Alright looks great! 👍
I'll give a final look at it and merge if I can't spot anything wrong :)

PR #1554, opened 2020-03-19
Effects: change animated to animationSpeed

Quarkstar:
### Change `animated` to `animationSpeed`

- [x] crt
- [x] glitch
- [x] godray
- [x] oldFilm
- [x] reflection

### Problems
According to their animation, I divide them into 2 groups.
- Depend on `time` (`reflection`, `godray`, `crt`)
  Their animation depend on `filter.time`, so `Animation Speed` has no limits.
- Depend on `seed` (`glitch`, `oldFilm`, `crt`)
  Their animation depend on how often `filter.seed` change(frequency of calling `update`), so they have a max value of `Animation Speed`.
   ```javascript
     if (filter.animationSpeed !== 0) {
       filter._animationTimer += filter.animationSpeed;
       if (filter._animationTimer >= 1) {
         filter.seed = Math.random();
         filter._animationTimer = 0
       }
     }
  },
  ```
Is that acceptable? I think users may be confused with this. Or is there a way to call `update` once and render it twice?
 
I think `Slider` may be helpful(just between `minSpeed` and `maxSpeed`), or we can keep `animated`.

### DIFF
--- a/Extensions/Effects/JsExtension.js
+++ b/Extensions/Effects/JsExtension.js
@@ -381,31 +381,31 @@ module.exports = {
     const crtProperties = crtEffect.getProperties();
     crtProperties.set(
       'lineWidth',
-      new gd.PropertyDescriptor(/* defaultValue= */ '0.2')
+      new gd.PropertyDescriptor(/* defaultValue= */ '1')
         .setLabel(_('Line width (between 0 and 5)'))
         .setType('number')
     );
     crtProperties.set(
       'lineContrast',
-      new gd.PropertyDescriptor(/* defaultValue= */ '0')
+      new gd.PropertyDescriptor(/* defaultValue= */ '0.25')
         .setLabel(_('Line contrast (between 0 and 1)'))
         .setType('number')
     );
     crtProperties.set(
       'noise',
-      new gd.PropertyDescriptor(/* defaultValue= */ '0.1')
+      new gd.PropertyDescriptor(/* defaultValue= */ '0.3')
         .setLabel(_('Noise (between 0 and 1)'))
         .setType('number')
     );
     crtProperties.set(
       'curvature',
-      new gd.PropertyDescriptor(/* defaultValue= */ '0')
+      new gd.PropertyDescriptor(/* defaultValue= */ '1')
         .setLabel(_('Curvature (between 0 and 10)'))
         .setType('number')
     );
     crtProperties.set(
       'verticalLine',
-      new gd.PropertyDescriptor(/* defaultValue= */ '0')
+      new gd.PropertyDescriptor(/* defaultValue= */ 'false')
         .setLabel(_('Vertical line (true or false)'))
         .setType('boolean')
     );
@@ -446,7 +446,7 @@ module.exports = {
     );
     crtProperties.set(
       'animationFrequency',
-      new gd.PropertyDescriptor(/* defaultValue= */ '0')
+      new gd.PropertyDescriptor(/* defaultValue= */ '60')
         .setLabel(_('Animation Frequency of Noise'))
         .setType('number')
         .setDescription('Number of updates per second (0: no updates)')
@@ -626,7 +626,7 @@ module.exports = {
     );
     glitchProperties.set(
       'animationFrequency',
-      new gd.PropertyDescriptor(/* defaultValue= */ '0')
+      new gd.PropertyDescriptor(/* defaultValue= */ '60')
         .setLabel(_('Animation Frequency'))
         .setType('number')
         .setDescription('Number of updates per second (0: no updates)')
@@ -916,7 +916,7 @@ module.exports = {
     );
     oldFilmProperties.set(
       'animationFrequency',
-      new gd.PropertyDescriptor(/* defaultValue= */ '0')
+      new gd.PropertyDescriptor(/* defaultValue= */ '60')
         .setLabel(_('Animation Frequency'))
         .setType('number')
         .setDescription('Number of updates per second (0: no updates)')
--- a/Extensions/Effects/JsExtension.js
+++ b/Extensions/Effects/JsExtension.js
@@ -436,13 +436,20 @@ module.exports = {
     crtProperties.set(
       'animationSpeed',
       new gd.PropertyDescriptor(/* defaultValue= */ '1')
-        .setLabel(_('Animation Speed'))
+        .setLabel(_('Animation Speed of Interlaced Lines'))
         .setType('number')
         .setDescription(
           _(
             '(E.g., 0: Pause, 0.5: Half speed, 1: Normal speed, 2: Double speed)'
           )
         )
+    );
+    crtProperties.set(
+      'animationFrequency',
+      new gd.PropertyDescriptor(/* defaultValue= */ '0')
+        .setLabel(_('Animation Frequency of Noise'))
+        .setType('number')
+        .setDescription('Number of updates per second (0: no updates)')
     );
 
     const displacementEffect = extension
--- a/Extensions/Effects/JsExtension.js
+++ b/Extensions/Effects/JsExtension.js
@@ -434,11 +434,15 @@ module.exports = {
         .setType('number')
     );
     crtProperties.set(
-      'animationFrequency',
-      new gd.PropertyDescriptor(/* defaultValue= */ '0')
-        .setLabel(_('Animation Frequency'))
+      'animationSpeed',
+      new gd.PropertyDescriptor(/* defaultValue= */ '1')
+        .setLabel(_('Animation Speed'))
         .setType('number')
-        .setDescription('Number of updates per second (0: no updates)')
+        .setDescription(
+          _(
+            '(E.g., 0: Pause, 0.5: Half speed, 1: Normal speed, 2: Double speed)'
+          )
+        )
     );
 
     const displacementEffect = extension
@@ -703,11 +707,15 @@ module.exports = {
         .setType('boolean')
     );
     godrayProperties.set(
-      'animationFrequency',
-      new gd.PropertyDescriptor(/* defaultValue= */ '0')
-        .setLabel(_('Animation Frequency'))
+      'animationSpeed',
+      new gd.PropertyDescriptor(/* defaultValue= */ '1')
+        .setLabel(_('Animation Speed'))
         .setType('number')
-        .setDescription('Number of updates per second (0: no updates)')
+        .setDescription(
+          _(
+            '(E.g., 0: Pause, 0.5: Half speed, 1: Normal speed, 2: Double speed)'
+          )
+        )
     );
     godrayProperties.set(
       'lacunarity',
@@ -1056,11 +1064,15 @@ module.exports = {
         .setDescription(_('Ending alpha (1 by default)'))
     );
     reflectionProperties.set(
-      'animationFrequency',
-      new gd.PropertyDescriptor(/* defaultValue= */ '0')
-        .setLabel(_('Animation Frequency of waves'))
+      'animationSpeed',
+      new gd.PropertyDescriptor(/* defaultValue= */ '1')
+        .setLabel(_('Animation Speed'))
         .setType('number')
-        .setDescription('Number of updates per second (0: no updates)')
+        .setDescription(
+          _(
+            '(E.g., 0: Pause, 0.5: Half speed, 1: Normal speed, 2: Double speed)'
+          )
+        )
     );
 
     const rgbSplitEffect = extension
--- a/Extensions/Effects/JsExtension.js
+++ b/Extensions/Effects/JsExtension.js
@@ -434,15 +434,11 @@ module.exports = {
         .setType('number')
     );
     crtProperties.set(
-      'animationSpeed',
-      new gd.PropertyDescriptor(/* defaultValue= */ '1')
-        .setLabel(_('Animation Speed (between 0 and 1)'))
+      'animationFrequency',
+      new gd.PropertyDescriptor(/* defaultValue= */ '0')
+        .setLabel(_('Animation Frequency'))
         .setType('number')
-        .setDescription(
-          _(
-            '(E.g., 0: Pause, 0.5: Half speed, 1: Normal speed)'
-          )
-        )
+        .setDescription('Number of updates per second (0: no updates)')
     );
 
     const displacementEffect = extension
@@ -618,15 +614,11 @@ module.exports = {
         .setDescription('The resolution of the displacement image')
     );
     glitchProperties.set(
-      'animationSpeed',
-      new gd.PropertyDescriptor(/* defaultValue= */ '1')
-        .setLabel(_('Animation Speed (between 0 and 1)'))
+      'animationFrequency',
+      new gd.PropertyDescriptor(/* defaultValue= */ '0')
+        .setLabel(_('Animation Frequency'))
         .setType('number')
-        .setDescription(
-          _(
-            '(E.g., 0: Pause, 0.5: Half speed, 1: Normal speed)'
-          )
-        )
+        .setDescription('Number of updates per second (0: no updates)')
     );
     glitchProperties.set(
       'redX',
@@ -711,15 +703,11 @@ module.exports = {
         .setType('boolean')
     );
     godrayProperties.set(
-      'animationSpeed',
-      new gd.PropertyDescriptor(/* defaultValue= */ '1')
-        .setLabel(_('Animation Speed'))
+      'animationFrequency',
+      new gd.PropertyDescriptor(/* defaultValue= */ '0')
+        .setLabel(_('Animation Frequency'))
         .setType('number')
-        .setDescription(
-          _(
-            '(E.g., 0: Pause, 0.5: Half speed, 1: Normal speed, 2: Double speed)'
-          )
-        )
+        .setDescription('Number of updates per second (0: no updates)')
     );
     godrayProperties.set(
       'lacunarity',
@@ -912,15 +900,11 @@ module.exports = {
         .setDescription('Blur intensity of the vignette')
     );
     oldFilmProperties.set(
-      'animationSpeed',
-      new gd.PropertyDescriptor(/* defaultValue= */ '1')
-        .setLabel(_('Animation Speed (between 0 and 1)'))
+      'animationFrequency',
+      new gd.PropertyDescriptor(/* defaultValue= */ '0')
+        .setLabel(_('Animation Frequency'))
         .setType('number')
-        .setDescription(
-          _(
-            '(E.g., 0: Pause, 0.5: Half speed, 1: Normal speed)'
-          )
-        )
+        .setDescription('Number of updates per second (0: no updates)')
     );
 
     const outlineEffect = extension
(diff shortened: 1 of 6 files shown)


4ian:
> Their animation depend on how often filter.seed change(frequency of calling update), so they have a max value of Animation Speed.

That's ok :) 
As "seed" is a discrete change, i.e: it depends on the frequency of calling update, it can not be animated faster than the number of frames we can display. This is not a problem at all.

> Or is there a way to call update once and render it twice?

That would not work, rendering a frame takes time and this would reduce the framerate because your computer would have more work to do, and so there would be even less frames to be rendered on screen.

So no worries, by definition something that is animated at every update can not go faster than the update refresh rate. It's not an animation but something that is changed randomly, so you can't even notice that things are not as fast as they should.

4ian:
This being said, your unit is currently tied to the game refresh rate. Instead, we can use a frequency:

```
    if (filter.animationFrequency !== 0) { 
      filter._animationTimer += layer.getElapsedTime() / 1000;
      if (filter._animationTimer >= 1 / filter.animationFrequency) {
        filter.seed = Math.random();
        filter._animationTimer = 0
      }
  }
},
```

`animationFrequency` is in "hz", or if you prefer "number of update per seconds". 

You can also use the approach done in sprites, i.e: ask the user for the time between two updates:

```
    if (filter.animationTimeBetweenUpdates !== 0) { 
      filter._animationTimer += layer.getElapsedTime();
      if (filter._animationTimer >= filter.animationTimeBetweenUpdates) {
        filter.seed = Math.random();
        filter._animationTimer = 0
      }
  }
},
```

`animationTimeBetweenUpdates` is in seconds, or if you prefer "seconds between each update".

I think one of these approach is a bit better than a 0-1 slider for the animation speed. Why? Because if your game runs at 60fps and I put 0.5, the filter would be updated 30 times a second. But if I have a high refresh rate monitor, my game can run at 90fps, and so the filter would be updated 45 times a second. We must at all cost avoid this behavior where a game is running at a different speed according to the monitor.

I would suggest switching to use frequency instead - seems like an intuitive unit (higher frequency = "faster", 0 = "no updates").

Quarkstar:
Thank you! I learned a lot from your advise:)
I have changed all these five filters to `animationSpeed`. And I will upload the gif soon.

blurymind:
Why don't we keep it as it was before, but use what was a boolean as a multiplier.
0 --> disabled
0.5 --> the same calculation as before multiplied by 0.5 (making the speed half)
2.0 --> same as above and so on..

Quarkstar:
I am sorry.
 
But animation of `oldFilmFilter` and `glitchFilter` only depend on `filter.seed`.
```javascript
if ( filter.animationSpeed  > 0 ) {
    filter.time += (layer.getElapsedTime() / 1000) * filter.animationSpeed ;	  
    filter.seed = Math.random();
}
```
This doesn't work on them. When `update` is called, `filter.seed` is updated. So the speed of animation is always the same, and `animationSpeed` doesn't work. That's why I add a `filter._animationTimer`.

I'm sorry, and I'm thinking of a better way to solve it.

4ian:
> Thats why most video editing/animation software lets you enter FPS and not animationTimeBetweenUpdates

That's why I suggested to use frequency and @Quarkstar updated the PR to use frequency :) 

> Why don't we keep it as it was before - exact same calculation, but use what was an enable boolean as a multiplier instead.

Because this is dependent on the game refresh rate which is tied to the monitor/gpu refresh rate. 

> Remember that an artist is likely entering those values, not a programmer who is obsessed with knowing the exact value between frames :p

That's why frequency is the best thing to use :)

4ian:
@blurymind See my explanations in https://github.com/4ian/GDevelop/pull/1554#issuecomment-601209066 :)

4ian:
Note that the frequency is only for `filter.seed`. For animation using `filter.time`, using a multiplier like @blurymind is what we should probably do.

Using a frequency here is **not** applicable. I was only commenting about `filter.seed` updates, **not** `filter.time` - which must be updated continuously.

Extensions/Effects/crt-pixi-filter.js
```diff
@@ -1,13 +1,17 @@
...
   },
   update: function(filter, layer) {
-    if (filter.animated) {
-      filter.time += layer.getElapsedTime() / 1000;
-      filter.seed = Math.random();
+    if (filter.animationFrequency !== 0) { 
+      filter._animationTimer += layer.getElapsedTime() / 1000;
+      if (filter._animationTimer >= 1 / filter.animationFrequency) {
+        filter.seed = Math.random();
+        filter.time += layer.getElapsedTime() / 1000;
```
4ian: time must be updated continuously. this is not linked to the frequency, which is only for discrete changes like seed.

Extensions/Effects/godray-pixi-filter.js
```diff
@@ -1,12 +1,16 @@
...
+    godrayFilter._animationTimer = 0;
     return godrayFilter;
   },
   update: function(filter, layer) {
-    if (filter.animated) {
-      filter.time += layer.getElapsedTime() / 1000;
+    if (filter.animationFrequency !== 0) { 
+      filter._animationTimer += layer.getElapsedTime() / 1000;
+      if (filter._animationTimer >= 1 / filter.animationFrequency) {
+        filter.time += layer.getElapsedTime() / 1000;
```
4ian: Same here, time must be updated continuously. 
Revert the changes and consider using animationSpeed for this.

Extensions/Effects/reflection-pixi-filter.js
```diff
@@ -19,12 +19,16 @@ gdjs.PixiFiltersTools.registerFilterCreator("Reflection", {
...
+    reflectionFilter._animationTimer = 0;
     return reflectionFilter;
   },
   update: function(filter, layer) {
-    if (filter.animated) {
-      filter.time -= layer.getElapsedTime() / 1000;
+    if (filter.animationFrequency !== 0) { 
+      filter._animationTimer += layer.getElapsedTime() / 1000;
+      if (filter._animationTimer >= 1 / filter.animationFrequency) {
+        filter.time += layer.getElapsedTime() / 1000;
```
4ian: Same, time must be continuously updated. Here you both use a timer and then update the time again with elapsedTime => this does not make sense in terms of time.

Frequency is only for seed.

Extensions/Effects/crt-pixi-filter.js
Quarkstar: Thank you! I have some misunderstandings about `time` before and I have fixed it.

Extensions/Effects/JsExtension.js
```diff
@@ -887,7 +898,7 @@ module.exports = {
       new gd.PropertyDescriptor(/* defaultValue= */ '1.0')
         .setLabel(_('Vignetting Alpha (between 0 and 1)'))
         .setType('number')
-        .setDescription('The size of the noise particles')
+        .setDescription('Amount of opacity of vignette')
```
4ian: Nice catch :)

4ian:
The changes looks correct to me :) I'll let @blurymind have a look - I believe the proper filters have been updated with the proper frequency/animation speed properties but better double check.

We might want to have in another PR special controls shown, like a frequency counter with a slider for frequency, and maybe something similar for the animation speed with a `x` at the beginning or the end of the input (so that it's displaying 0.5x, 1x, 2x..., could help to make more obvious this is a speed multiplier). 
The interesting thing about adding these controls is that they could be used in behaviors or any other thing having properties.

Extensions/Effects/crt-pixi-filter.js
```diff
@@ -5,8 +5,8 @@ gdjs.PixiFiltersTools.registerFilterCreator('CRT', {
     return crtFilter;
   },
   update: function(filter, layer) {
-    if (filter.animated) {
-      filter.time += layer.getElapsedTime() / 1000;
+    if (filter.animationSpeed !== 0) { 
+      filter.time += layer.getElapsedTime() / 1000 * filter.animationSpeed;
       filter.seed = Math.random();
```
blurymind: the animation speed multiplier approach doesnt seem to make a difference for this effect. I wonder if we need limit the frequency at which seed is being changed to to it

4ian: Good catch! seed and time must be handled separately I think. One with frequency, the other with the animation speed.
So this means two settings in the UI :) 

If that does not make sense to have them separated, then we must update the seed every once in a while, but that means choosing a frequency for the user so seems weird.... It would be great to know what is the difference between time and seed for this CRT filter :)

Quarkstar: It works but not obvious. It's much more obvious when `lineWidth` is 10.

The animation of `crt` consists of `noise`(influenced by `seed`) and `line moves`(influenced by `time`).

blurymind: I came up with this formula
filter.seed =  Math.round((filter.time % filter.animationSpeed) * 10) + 0.1;

seems to do the trick :)

Extensions/Effects/glitch-pixi-filter.js
```diff
@@ -1,13 +1,16 @@
...
     return glitchFilter;
   },
   update: function(filter, layer) {
-    if (filter.animated) {
-      filter.time += layer.getElapsedTime() / 1000;
-      filter.seed = Math.random();
+    if (filter.animationFrequency !== 0) { 
+      filter._animationTimer += layer.getElapsedTime() / 1000;
+      if (filter._animationTimer >= 1 / filter.animationFrequency) {
+        filter.seed = Math.random();
```
blurymind: If my seed affected by speed formula is ok by @4ian, please simplify this logic to how it was before and use it here too

Extensions/Effects/old-film-pixi-filter.js
```diff
@@ -1,13 +1,16 @@
...
     return oldFilmFilter;
   },  
   update: function(filter, layer) {
-    if (filter.animated) {
-      filter.time += layer.getElapsedTime() / 1000;
-      filter.seed = Math.random();
+    if (filter.animationFrequency !== 0) { 
+      filter._animationTimer += layer.getElapsedTime() / 1000;
+      if (filter._animationTimer >= 1 / filter.animationFrequency) {
+        filter.seed = Math.random();
```
blurymind: again, could be simplified with my seed formula from earlier

Extensions/Effects/crt-pixi-filter.js
Quarkstar: > If we can give the user control over the speed without adding complexity, that would be great
I do agree with it. But this is a problem with this formula(`oldFilmFilter` as an example), scratches used to be appeared randomly, but now it always appear in the same position.

blurymind: > > If we can give the user control over the speed without adding complexity, that would be great
> 
> I do agree with it. But this is a problem with this formula(`oldFilmFilter` as an example), scratches used to be appeared randomly, but now it always appears in the same position. So I think `random` cannot be removed. What do you think?

I think the frequency is kind of counterintuitive, since a greater value results in slower animation. I think we can improve it to make it consistent :)

blurymind: put this in old film filter- exact same logic, but with your  renamed parameters
```
    if (filter.animationFrequency !== 0) {
      filter._animationTimer +=
        (layer.getElapsedTime() / 1000) * filter.animationFrequency;
      filter.seed =
        Math.round((filter._animationTimer % filter.animationFrequency) * 10) +
        0.1;
    }
  },
```
the formula works and a greater value results in faster animation, not the other way around

4ian: > Math.round((filter._animationTimer % filter.animationFrequency) * 10) + 0.1;

What are the reason for choosing 10 and 0.1?
This seems to also severly limit the 

> filter._animationTimer += (layer.getElapsedTime() / 1000) * filter.animationFrequency;

A frequency is not the same unit as a multiplier, which does not have a unit 😬
A frequency unit is "per second". We should at least be following proper units or not call this a frequency.

Not saying that it's not a good idea to have the two tied - but here we're getting result that is not predictable for the end user.

> the formula works and a greater value results in faster animation, not the other way around

This is the same with a frequency and an animation speed right?
2x, 3x, 4x => you're increasing the speed of the animation

20hz, 30hz, 60hz => you're increasing the speed of the update to the random noise shown on screen.

4ian: > The animation of crt consists of noise(influenced by seed) and line moves(influenced by time).

This seems that the CRT filter has two things: noise (seed) and how line moves (time).
If they are two different things, why not showing the two of them in the UI?

Noise update: 30hz, 60hz, 120hz...
Line movement speed: 0.5x, 1x, 2x...

Sounds easy no? And very flexible :) We're not limiting the effect compared to the original Pixi effect, and we're using units that are the proper one.

blurymind: I guess it would be even better to keep the randomness, but also keep it animationSpeed somehow?

4ian: Yep! That's why I think we need both, but separate (only in the case of CRT I think):

```
    // Update time (for lines on screen)
    filter.time += layer.getElapsedTime() / 1000 * filter.animationSpeed;

    // Update random scratches (at the specified frequency)
    if (filter.animationFrequency !== 0) { 
      filter._animationTimer += layer.getElapsedTime() / 1000;
      if (filter._animationTimer >= 1 / filter.animationFrequency) {
        filter.seed = Math.random();
        filter._animationTimer = 0;
      }
  }
},
```

Quarkstar: Thank you! And I have fixed it.

blurymind: I like the idea of splitting the random seed from time and letting the user set them individually. If this approach also means that for both a higher value results in a faster animation, then I'm good with it :) one thing worth double checking is if the default values that are set when adding each filter best demonstrate it's animation

blurymind: Also check if changing each results in a visually different result. I know it's kind of obvious, but worth double checking

Quarkstar: > Also check if changing each results in a visually different result. I know it's kind of obvious, but worth double checking

I have checked them and no problems.

>if the default values that are set when adding each filter best demonstrate it's animation

The default value of `lineWidth`, `lineContrast`, `noiseSize` in `CRT` may need to change. But I don't know which value is appropriate because it looks different in different game. 
I have tried it and I think `lineWidth: 3, lineContrast: 0.2, noiseSize: 2` looks fine to me. What do you think?

4ian: > I have tried it and I think lineWidth: 3, lineContrast: 0.2, noiseSize: 2 looks fine to me

Add a screenshot so that we can all see the result :) I would say that any default value that make a pretty rendering on screen are fine :)

Extensions/Effects/JsExtension.js
```diff
@@ -613,10 +625,11 @@ module.exports = {
         .setDescription('The resolution of the displacement image')
     );
     glitchProperties.set(
-      'animated',
-      new gd.PropertyDescriptor(/* defaultValue= */ 'true')
-        .setLabel(_('Animated (Enable animations)'))
-        .setType('boolean')
+      'animationFrequency',
+      new gd.PropertyDescriptor(/* defaultValue= */ '0')
```
4ian: The default value should not be 0 I think

Extensions/Effects/JsExtension.js
```diff
@@ -897,10 +915,11 @@ module.exports = {
         .setDescription('Blur intensity of the vignette')
     );
     oldFilmProperties.set(
-      'animated',
-      new gd.PropertyDescriptor(/* defaultValue= */ 'true')
-        .setLabel(_('Animated (Enable animations)'))
-        .setType('boolean')
+      'animationFrequency',
+      new gd.PropertyDescriptor(/* defaultValue= */ '0')
```
4ian: Same, let's make a default value that was equivalent to animated = true, i.e: 60

Extensions/Effects/crt-pixi-filter.js
Quarkstar: There are three examples.
### Pixi-filter documentation
Screenshot from 2020-03-22 20-40-45

### Default value we used to use
Screenshot from 2020-03-22 20-27-21

### Pixi-filter Demo
Screenshot from 2020-03-22 20-40-45

It is hard to say which looks best to me. Can you choose one? And I will both fix it and the default frequency.

Extensions/Effects/JsExtension.js
blurymind: yes, for every animated filter the default should be for it to animate

Extensions/Effects/crt-pixi-filter.js
4ian: I'm fine with either pixi-filter doc/demo :)

Extensions/Effects/JsExtension.js
Quarkstar: Thank you! I have fixed it.

And I move animation of `crt-lines` 10 times faster to make it obvious.
Sorry for no gif. (I cannot upload a `gif` which can show the animation clearly and <= 10M).

Extensions/Effects/JsExtension.js
```diff
@@ -434,10 +434,22 @@ module.exports = {
...
+        .setDescription(
+          _(
+            '(E.g., 0: Pause, 0.5: Half speed, 1: Normal speed, 2: Double speed)'
+          )
+        )
+    );
+    crtProperties.set(
+      'animationFrequency',
+      new gd.PropertyDescriptor(/* defaultValue= */ '60')
+        .setLabel(_('Animation Frequency of Noise'))
```
4ian: Frequency of noise

Extensions/Effects/JsExtension.js
```diff
@@ -887,7 +905,7 @@ module.exports = {
       new gd.PropertyDescriptor(/* defaultValue= */ '1.0')
         .setLabel(_('Vignetting Alpha (between 0 and 1)'))
         .setType('number')
-        .setDescription('The size of the noise particles')
+        .setDescription('Amount of opacity of vignette')
```
4ian: Actually we can remove the setDescription entirely, this is not giving much information.

Extensions/Effects/crt-pixi-filter.js
```diff
@@ -1,13 +1,19 @@
...
-
+    crtFilter._animationTimer = 0;
     return crtFilter;
   },
   update: function(filter, layer) {
-    if (filter.animated) {
-      filter.time += layer.getElapsedTime() / 1000;
-      filter.seed = Math.random();
+    if (filter.animationSpeed !== 0) { 
+      filter.time += layer.getElapsedTime() / 100 * filter.animationSpeed;
```
4ian: `/ 100` looks like a bug. Avoid magic numbers, prefer doing:
`/ 1000 * 10 * filter.animationSpeed`

and add a comment:
`// Multiply by 10 so that the default value shows use a sensible speed`

Extensions/Effects/JsExtension.js
Quarkstar: Thank you and how about `Vignetting Blur`'s description? I think it can be removed too.

Extensions/Effects/JsExtension.js
Quarkstar: ezgif com-video-to-gif

Extensions/Effects/JsExtension.js
4ian: Yes we can remove it too I think! :)

4ian:
Alright, I think we're good to go with this. Knowing that this is a "breaking" change because filters that were set up as not animated will need to be manually updated by the users to set animation speed or frequency to 0.
I think it's fine but worth noting.

4ian:
Merged! Thanks for your work on this :) 
I believe having custom controls for the frequency/speed would be awesome (slider + field), though it's maybe not the highest priority (up to you)

Quarkstar:
Thank you!
Actually, I don't know much about ReactJS and I am learning it now. And I will try it when I am more familiar with it. :rofl:  
Thanks again for your patience!

PR #1555, opened 2020-03-19
Effects: Remove window

Quarkstar:
### Remove window
I have checked that only `zoomBlur` uses `window`. Now all filters are in pixels by default.

### DIFF
--- a/Extensions/Effects/JsExtension.js
+++ b/Extensions/Effects/JsExtension.js
@@ -1191,20 +1191,20 @@ module.exports = {
     const zoomBlurProperties = zoomBlurEffect.getProperties();
     zoomBlurProperties.set(
       'centerX',
-      new gd.PropertyDescriptor(/* defaultValue= */ '0.5')
-        .setLabel(_('Center X (between 0 and 1, 0.5 is window middle)'))
+      new gd.PropertyDescriptor(/* defaultValue= */ '0')
+        .setLabel(_('Center X'))
         .setType('number')
     );
     zoomBlurProperties.set(
       'centerY',
-      new gd.PropertyDescriptor(/* defaultValue= */ '0.5')
-        .setLabel(_('Center Y (between 0 and 1, 0.5 is window middle)'))
+      new gd.PropertyDescriptor(/* defaultValue= */ '0')
+        .setLabel(_('Center Y'))
         .setType('number')
     );
     zoomBlurProperties.set(
       'innerRadius',
-      new gd.PropertyDescriptor(/* defaultValue= */ '0.2')
-        .setLabel(_('Inner radius (between 0 and 1, 0.5 is mid-way)'))
+      new gd.PropertyDescriptor(/* defaultValue= */ '200')
+        .setLabel(_('Inner radius'))
         .setType('number')
     );
     zoomBlurProperties.set(
--- a/Extensions/Effects/JsExtension.js
+++ b/Extensions/Effects/JsExtension.js
@@ -970,14 +970,14 @@ module.exports = {
     );
     radialBlurProperties.set(
       'centerX',
-      new gd.PropertyDescriptor(/* defaultValue= */ '0')
-        .setLabel(_('Center X'))
+      new gd.PropertyDescriptor(/* defaultValue= */ '0.5')
+        .setLabel(_('Center X (between 0 and 1, 0.5 is image middle)'))
         .setType('number')
     );
     radialBlurProperties.set(
       'centerY',
-      new gd.PropertyDescriptor(/* defaultValue= */ '0')
-        .setLabel(_('Center Y'))
+      new gd.PropertyDescriptor(/* defaultValue= */ '0.5')
+        .setLabel(_('Center Y (between 0 and 1, 0.5 is image middle)'))
         .setType('number')
     );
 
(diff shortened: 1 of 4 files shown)


blurymind:
wouldnt this make their default look awful?

- not centered

Can you add a screenshot of how the new defaults look like?


We are also now making the user come up with a way to set the center to the same thing but with the event sheet or even worse - hardcode the value in the parameter field

Quarkstar:
It looks not good. I think it is a good choice to replace number input box with [Slider with input field](https://material-ui.com/components/slider/#with-input-field). And we should also need the width/height of layer or objects in Slider. 

### Screenshot of Twist Effect(default)
Screenshot from 2020-03-20 00-35-50

(left upper corner)

blurymind:
Precisely, it looks broken and it doesn't  really give us any gains- since the filter will not translate well to an object filter anyway.

My problem is with having to force the user enter absolute pixel coordinates for that field. Its super inconvenient. Can't we keep it 0-1 with 0.5 being middle please? I dont care if its in the form of an input or a slider, just have the default not look broken and dont force the user write patchy logic for resizing game window

Quarkstar:
I am sorry for that.

>  just have the default not look broken and dont force the user write patchy logic for resizing game window

In order to do that, I think `width/height` of layer should be passed into `*-pixi-filter.js` and `JsExtension.js`. 
In `Pixi.Filters`, there is
```javascript
    apply(filterManager, input, output, clear) {
        this.uniforms.dimensions[0] = input.filterFrame.width;
        this.uniforms.dimensions[1] = input.filterFrame.height;
        filterManager.applyFilter(this, input, output, clear);
    }
```
I think `input.filterFrame` represent the input image. So could we use `this.uniforms.dimensions` to get the objects/layers' `width/height`? It is dangerous  but it can resolve the problems.  (Just replace `window.innerWidth` with `this.uniforms.dimensions[0]`)

blurymind:
Sounds like something we need to do anyways, if we are going to try to apply screen filters to objects. @4ian what do you think :)

4ian:
> So could we use this.uniforms.dimensions to get the objects/layers' width/height?

Sounds ok. Or can we access filterFrame somehow? But it will be better to use this or uniforms.dimensions because it will automatically adapt when we enable effects on objects :)

Quarkstar:
@4ian @blurymind 
I resolve this problem, but in a really dangerous way.(I don't like it, but it works:)
### Code
Here is an example `TwistFilter`
```javascript
  makePIXIFilter: function(layer, effectData) {
    var twistFilter = new PIXI.filters.TwistFilter();
    twistFilter.uniforms.dimensions = new Float32Array(2);
    PIXI.filters.TwistFilter.prototype.apply = function(filterManager, input, output, clear) {
        this.uniforms.dimensions[0] = input.sourceFrame.width;
        this.uniforms.dimensions[1] = input.sourceFrame.height;
        filterManager.applyFilter(this, input, output, clear);
    }
    twistFilter._updateNum = 0;
    return twistFilter;
  },
    if(filter._updateNum < 2) {
      filter.offset[0] = Math.round(filter._offsetX * filter.uniforms.dimensions[0]);
      filter.offset[1] = Math.round(filter._offsetY * filter.uniforms.dimensions[1]);
      filter._updateNum += 1;
    }
  },
  updateDoubleParameter: function(filter, parameterName, value) {
    if (parameterName === 'offsetX') {
      filter._offsetX = value;
    }
    else if (parameterName === 'offsetY') {
      filter._offsetY = value;
    }
  }
```

### Explanations
```javascript
    twistFilter.uniforms.dimensions = new Float32Array(2);
    PIXI.filters.TwistFilter.prototype.apply = function(filterManager, input, output, clear) {
        this.uniforms.dimensions[0] = input.sourceFrame.width;
        this.uniforms.dimensions[1] = input.sourceFrame.height;
        filterManager.applyFilter(this, input, output, clear);
    }
    twistFilter._updateNum = 0;
```
- `filter.uniforms.dimensions` is used to store the width and height.
- Not every `PIXI.filter` have an `apply` function like this, so I add it.(Most of filters don't have it)

```javascript
  updateDoubleParameter: function(filter, parameterName, value) {
    if (parameterName === 'offsetX') {
      filter._offsetX = value;
    }
    else if (parameterName === 'offsetY') {
      filter._offsetY = value;
    }
  }
```

```javascript
    if(filter._updateNum < 2) {
      filter.offset[0] = Math.round(filter._offsetX * filter.uniforms.dimensions[0]);
      filter.offset[1] = Math.round(filter._offsetY * filter.uniforms.dimensions[1]);
      filter._updateNum += 1;
    }
  },
```
This just replace `window.innerWidth/innerHeight` with `filter.uniforms.dimensions`.
Updating `filter.offset` moves from `updateDoubleParameter` to `update`. This is because `filter.offset` must be updated after `apply` is called. (`apply` changes `filter.uniforms.dimensions` from `0` to image's height and width).
So when will `apply` be called? `if(filter._updateNum < 2) `. Actually, I find `2` by testing over and over again. :rofl: It is a magic number and I don't know why. I guess the order of function calling is 
1. update, updateDoubleParameter...
2. `apply`
3. update
4. update ...

### What's more

I think it's really dangerous to add method to `PIXI.filter`, and the code looks dirty. But it is the best way that I could think of. (I don't know if I made it clear.) Can you give me some advice?


And `apply` should be updated to this if we update `PIXI.filter` to v3.
```javascript
    PIXI.filters.TwistFilter.prototype.apply = function(filterManager, input, output, clear) {
        this.uniforms.dimensions[0] = input.filterFrame.width;
        this.uniforms.dimensions[1] = input.filterFrame.height;
        filterManager.applyFilter(this, input, output, clear);
    }
```

PR #1556, opened 2020-03-20
Add duplicate in project manager

Bouh:
Add duplicate option in context menu  in project manager, for scene, external layout, and external events and extension/functions.


4ian:
Looks great! 👍

Issue #1557, opened 2020-03-20
Object with numeric values in start of name are unusable


Bouh:
## Describe the bug
When a object contain a number at start like `1PlayerHitBox`.
This object isn't works in preview.

## To Reproduce
1. Open platformer example
2. Rename `PlayerHitBox `to `1PlayerHitBox`
3. Run the preview you can't move

And worse, i we go back to initial name `PlayerHitBox` and re run the preview, the game have still the issue.

## Other details
* beta 89

geetesh-gupta:
One solution can be to restrict the names to start with alphabets i.e. have the object names follow the same syntactic rules as other variables names in general for any programming language. It will make the naming consistent with general guidelines. This way the problem will be solved as well as proper naming specification will be followed. What are your opinions? 🤔

4ian:
The rules are:
1. Only alphanumeric characters (a-z, A-Z, 0-9) and underscores
2. Do not start with a numeric character (0-9).

The second rule is because otherwise when parsing expression, we can't make the difference between a number (123) and an object name (123) (we say that the grammar is "ambiguous").

Actually, there is no technical restrictions for rule 1 apart from not to use what are considered as operator, separators or whitespace(-, +, /, *, =, {,},[,],",';,:,. etc ...) by the expression parser.

We could imagine allowing any unicode character that is not an operator, separators, whitespace or number. This would allow to enter characters used in other languages than English (accentuated characters, Chinese/Korean, cyrillic... well everything supported by Unicode :)).

For now to be simple, I decided to just accept names that are valid in JavaScript.
You can see in the IDE that we use `gd.Project.validateObjetName`. And you can see the implementation here: 

https://github.com/4ian/GDevelop/blob/3ff5dd7cd2c1f6b2bfd57b2ef79b65d1e47ab15a/Core/GDCore/Project/Project.cpp#L948-L958

Turns out that I forgot to implement the second rule :) Should be not too hard to add, a bit of C++ and GDevelop.js work (see the README in GDevelop.js)

PawBud:
I think that this issue should be closed right @Bouh @4ian?

Bouh:
Yes, just i've no right for close it :)
In general lastest issues fixed are pass in review and closed just before the next release and writing of patch note.

PR #1560, opened 2020-03-20
fix: object with numeric values in start are unusable

jimil749:
fix for #1557 
Added a check using isdigit to check if the first character is not numeric.

GDevelop-2

One issue is, this shows the same pop-up stating it contains forbidden characters, whereas here it does not, so should we change the pop or just modify the text in it.


4ian:
Thanks! :) Looks good. Also update `GetBadObjectNameWarning` with a concise explanation about not starting object name with a digit :)

jimil749:
Along with updating the `GetBadObjectNameWarning`, shouldn't we also update the warning displayed to the user which is inside of the MainFrame? Because that would make users think what went wrong! So I took the liberty and changed the warning message if that's fine!

jimil749:
Along with updating the `GetBadObjectNameWarning`, shouldn't we also update the warning shown to the user? Else they would think what went wrong! So I took the liberty to update warning message! :)

4ian:
Yes that's a good idea, actually we should change the warning for ALL messages in the codebase. That's a good thing to do whenever you change something: check the rest of the codebase. How? Just search for validateObjectName.

I'll do that for you :)

jimil749:
No worries, I'll make the changes right away. Shouldn't take much time, as currently it's only updated for the object.

4ian:
Actually validateObjectName is ill named (my fault) and is used for quite a few other name validation => all needs to be updated.

Use `This name is invalid. Only use alphanumeric characters (0-9, a-z) and underscores. Digits are not allowed as the first character.` for the user message :)

jimil749:
So should we change the name to validateVariableName? Makes more sense, if it is used to validate other variables?

4ian:
Maybe, or just "validateName" (a variable has another meaning in GDevelop)

jimil749:
Weird that the tests are failing! I did not touch the file failing. 
Any idea why is that happening?

newIDE/app/src/EventsFunctionsExtensionEditor/index.js
```diff
@@ -219,18 +219,18 @@ export default class EventsFunctionsExtensionEditor extends React.Component<
...
-          t`This name contains forbidden characters: please only use alphanumeric characters (0-9, a-z) and underscores in your function name.`
+          t`This name is invalid. Only use alphanumeric characters (0-9, a-z) and underscores. Digits are not allowed as the first character.`
         )
       );
       return;
     }
     if (isExtensionLifecycleEventsFunction(newName)) {
       showWarningBox(
         i18n._(
-          t`This name is reserved for a lifecycle function of the extension. Choose another name for your function.`
```
4ian: Careful, you've changed the wrong message here, it's a regression

jimil749: Mistake on my part. Updated the wrong message here.

4ian: I fixed this for you, no need to push anything

4ian:
Merged :) Thanks for your work on this!
I'd advise to always double check your changes when you make a PR. For example, using the search function, search for the sentence that you've changed, and check that the count is making sense according to the count for the previous sentence. That's how I found that you renamed an extra sentence that should not have been.

Thanks :)  👍

jimil749:
Thanks for merging and also I appreciate you helping me with all the advice! Will be sure to be more careful the next time. :)
Looking forward to solve more issues! ;)

PR #1564, opened 2020-03-21
Effects: Remove window

Quarkstar:
Sorry, I delete the branch by mistake and it can't be restored.

### Code
Here is an example for `TwistFilter`
```javascript
  makePIXIFilter: function(layer, effectData) {
    var twistFilter = new PIXI.filters.TwistFilter();
    PIXI.filters.TwistFilter.prototype.apply = function(filterManager, input, output, clear) {
        this.offset[0] = Math.round(this._offsetX * input.sourceFrame.width);
        this.offset[1] = Math.round(this._offsetY * input.sourceFrame.height);
        filterManager.applyFilter(this, input, output, clear);
    }
    
    return twistFilter;
  },
  update: function(filter, layer) { },
  updateDoubleParameter: function(filter, parameterName, value) {
    if (parameterName === 'offsetX') {
      filter._offsetX = value;
    }
    else if (parameterName === 'offsetY') {
      filter._offsetY = value;
    }
  }
```

### Explanations

- Override `apply` in `PIXI.filter`
- Replace `window.innerWidth/innerHeight` with `filter.uniforms.dimensions`.
- Updating `filter.offset` moves from `updateDoubleParameter` to `apply`. This is because `filter.offset` must be updated after  `filter.uniforms.dimensions` getting a value.
- `apply` should be updated if we update `PIXI.filter` to v3. (replace `sourceFrame` with filterFrame).
### Questions

- I think it's really dangerous to override functions in `PIXI.filter`. But it is the best way that I could think of. (I don't know if I made it clear.) Can you give me some advice?
- I think `radius` in pixels is a reasonable choice. Should I change it to 0-1?

### DIFF
--- a/Extensions/Effects/JsExtension.js
+++ b/Extensions/Effects/JsExtension.js
@@ -1191,20 +1191,20 @@ module.exports = {
     const zoomBlurProperties = zoomBlurEffect.getProperties();
     zoomBlurProperties.set(
       'centerX',
-      new gd.PropertyDescriptor(/* defaultValue= */ '0.5')
-        .setLabel(_('Center X (between 0 and 1, 0.5 is window middle)'))
+      new gd.PropertyDescriptor(/* defaultValue= */ '0')
+        .setLabel(_('Center X'))
         .setType('number')
     );
     zoomBlurProperties.set(
       'centerY',
-      new gd.PropertyDescriptor(/* defaultValue= */ '0.5')
-        .setLabel(_('Center Y (between 0 and 1, 0.5 is window middle)'))
+      new gd.PropertyDescriptor(/* defaultValue= */ '0')
+        .setLabel(_('Center Y'))
         .setType('number')
     );
     zoomBlurProperties.set(
       'innerRadius',
-      new gd.PropertyDescriptor(/* defaultValue= */ '0.2')
-        .setLabel(_('Inner radius (between 0 and 1, 0.5 is mid-way)'))
+      new gd.PropertyDescriptor(/* defaultValue= */ '200')
+        .setLabel(_('Inner radius'))
         .setType('number')
     );
     zoomBlurProperties.set(
--- a/Extensions/Effects/JsExtension.js
+++ b/Extensions/Effects/JsExtension.js
@@ -970,14 +970,14 @@ module.exports = {
     );
     radialBlurProperties.set(
       'centerX',
-      new gd.PropertyDescriptor(/* defaultValue= */ '0')
-        .setLabel(_('Center X'))
+      new gd.PropertyDescriptor(/* defaultValue= */ '0.5')
+        .setLabel(_('Center X (between 0 and 1, 0.5 is image middle)'))
         .setType('number')
     );
     radialBlurProperties.set(
       'centerY',
-      new gd.PropertyDescriptor(/* defaultValue= */ '0')
-        .setLabel(_('Center Y'))
+      new gd.PropertyDescriptor(/* defaultValue= */ '0.5')
+        .setLabel(_('Center Y (between 0 and 1, 0.5 is image middle)'))
         .setType('number')
     );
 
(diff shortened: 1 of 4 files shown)


4ian:
> I think radius in pixels is a reasonable choice

I think it's reasonable.

> I think it's really dangerous to override functions in PIXI.filter. But it is the best way that I could think of

Yes I think it's dangerous. I'd like to see if we can find another solution. Maybe we could pass the size to the effect?

Quarkstar:
### New methods
This looks much better:) It doesn't need change anything except `twist-pixi-filter.js`
```javascript
  makePIXIFilter: function(layer, effectData) {
  var twistFilter = new PIXI.filters.TwistFilter();

  return twistFilter;
},
    filter.offset[0] = Math.round(filter._offsetX * layer.getWidth());
    filter.offset[1] = Math.round(filter._offsetY * layer.getHeight());
  }
},
```
I update `filter.offset` in `update` for two reasons:
- I can only get the data of layer in `update`.
- According to `GDJS/Runtime/layer.js`, `layer.getWidth()` return different values sometimes.(when the game resolution size is changed.)

### Questions
For filters on objects, I would like know what are the changes in `*-pixi-filter.js`. Is it just replace `layer` with `object`, or we need to change more?

Quarkstar:
I think `layer.getWidth()/getHeight()` is a good choice. I have tried it and it works well.

### Explanations
- I have checked that `runtimeobject`, `bbtextruntimeobject`,`spriteruntimeobject`, and `SkeletonRuntimeObject` all have method `getWidth()` and `getHeight()`, so few changes are needed to add `object`.
- Now `filter.center` is updated in `update` function, which will degrade `performance`. This is because I find `layer`'s width and heigth may be changed.(And the object's width and height may be changed as well).

What do you think about them?

Quarkstar:
### SreenShots

#### Twist
Screenshot from 2020-03-25 23-19-03
Screenshot from 2020-03-25 23-18-55

#### ZoomBlur
Screenshot from 2020-03-25 23-21-14

#### RadialBlur
Screenshot from 2020-03-25 23-21-42

4ian:
Sorry for the delay!
Seems like `getWidth`/`getHeight` are a good solution. 👍

> Now filter.center is updated in update function, which will degrade performance. This is because I find layer's width and heigth may be changed.(And the object's width and height may be changed as well).

My intuition is that it's ok, getWidth/getHeight should not be costly to get for layers or objects, and then assigning two numbers sounds not too expensive :) 

> so few changes are needed to add object.

Yes, when object are added we'll rename the `layer` parameter to `layerOrObject`. If we limit ourselves to use `getWidth`/`getHeight`, that sounds fair 👍

Thanks for your nice investigation so far! I believe we have now an implementation that is not hacky and that will adapt well to objects. Let me double check.

4ian:
Looks great, not reason not to go ahead with this solution.
Thanks for your investigation on this! Really useful and should adapt well in the future to objects 👍Dropping the dependency on window seems good in any case

Quarkstar:
Thank you! I'd like to add supports for objects. I learned a lot from #1321.
 
I find that `pixi-filter` can be added to `pixi-sprite`, just like `pixi-container`
```javascript
someSprite.filters = [
     new GlowFilter({ distance: 15, outerStrength: 2 })
];
```
So I think the main part of it is to add filters to `spriteruntimeobject-pixi-render`, just like `layer-pixi-render`. And there are some other changes in `EffectsCodeGenerator.cpp`, etc. I am afraid I think too simply. Can you correct me if I say something wrong?

4ian:
I think we more an approach that is a bit more generic:
- Add the possibility for a gd::Object to have a list of effects (C++)
- In the UI, add an "EffectsList" in the editor of objects (like behaviors)
- In the game engine, add the possibility for all gdjs.RuntimeObject to have a list of filters. Maybe something in gdjs.RuntimeObject that would use the renderer object (`getRendererObject`) to get the PIXI object and, if not null, apply the filters.

PR #1568, opened 2020-03-21
Adding feature to close inline autocomplete after selection

### LINKED ISSUE #1558: Unable to exit select Object in events until clicked outside the menu
## Describe the bug
Whenever we try to select an object for an event in the Events tab, we cannot exit the select object menu until clicked outside the menu. The expected behaviour should be whenever an object is selected, 

- if "Enter" is pressed, the object should be selected and the menu should disappear.
- if "Esc" is pressed, the previous value of the object is selected and the menu should disappear.

Also, this will make it consistent with other triggers like Comment and Group.

## To Reproduce
Steps to reproduce the behaviour:
1. Open any project
2. Click on the Events tab
3. Scroll down to any event wherever an object is desired
4. Click on to change the object
5. Select any object
6. See error (unable to exit until clicked somewhere else)
image

HarsimranVirk:
Resolves #1558 
I have added the functionality for closing the popover after selection to most of the fields. I'm aware of not adding this to ResourceSelector since it has some issues with proper selection. I'd be checking what should be done for it, meanwhile, we can check if I missed any field or if there is any possible regression.

### DIFF
--- a/newIDE/app/src/ResourcesList/ResourceSelector.js
+++ b/newIDE/app/src/ResourcesList/ResourceSelector.js
@@ -33,6 +33,7 @@ type Props = {|
   canBeReset?: boolean,
   initialResourceName: string,
   onChange: string => void,
+  onRequestClose?: () => void,
   floatingLabelText?: React.Node,
   helperMarkdownText?: ?string,
   hintText?: MessageDescriptor,
@@ -169,6 +170,7 @@ export default class ResourceSelector extends React.Component<Props, State> {
       () => {
         if (!this.state.notExistingError) {
           if (this.props.onChange) this.props.onChange(resourceName);
+          if (this.props.onRequestClose) this.props.onRequestClose();
         }
       }
     );
--- a/newIDE/app/src/ResourcesList/ResourceSelector.js
+++ b/newIDE/app/src/ResourcesList/ResourceSelector.js
@@ -33,7 +33,6 @@ type Props = {|
   canBeReset?: boolean,
   initialResourceName: string,
   onChange: string => void,
-  onRequestClose?: () => void,
   floatingLabelText?: React.Node,
   helperMarkdownText?: ?string,
   hintText?: MessageDescriptor,
@@ -170,7 +169,6 @@ export default class ResourceSelector extends React.Component<Props, State> {
       () => {
         if (!this.state.notExistingError) {
           if (this.props.onChange) this.props.onChange(resourceName);
-          if (this.props.onRequestClose) this.props.onRequestClose();
         }
       }
     );
--- a/newIDE/app/src/UI/SemiControlledAutoComplete.js
+++ b/newIDE/app/src/UI/SemiControlledAutoComplete.js
@@ -148,8 +148,9 @@ const handleChange = (
   if (option.type !== 'separator') {
     if (option.onClick) option.onClick();
     else {
-      if (props.onChoose) props.onChoose(option.value);
-      props.onChange(option.value);
+      props.onChoose
+        ? props.onChoose(option.value)
+        : props.onChange(option.value);
       if (props.onRequestClose) props.onRequestClose();
     }
   }
@@ -275,6 +276,12 @@ export default React.forwardRef<Props, SemiControlledAutoCompleteInterface>(
                       setIsMenuOpen(false);
                       props.onChange(event.currentTarget.value);
                       if (props.onBlur) props.onBlur(event);
+                    },
+                    onMouseDown: (
+                      event: SyntheticMouseEvent<HTMLInputElement>
+                    ): void => {
+                      if (input.current && !input.current.value.length)
+                        setIsMenuOpen(!isMenuOpen);
                     },
                   }}
                   {...other}
--- a/newIDE/app/src/UI/SemiControlledAutoComplete.js
+++ b/newIDE/app/src/UI/SemiControlledAutoComplete.js
@@ -142,20 +142,17 @@ const filterFunction = (
 
 const handleChange = (
   event: SyntheticKeyboardEvent<HTMLInputElement>,
-  option: Option | string,
+  option: Option,
   props: Props
 ): void => {
   if (option.type !== 'separator') {
-    if (typeof option === 'string') props.onChange(option);
+    if (option.onClick) option.onClick();
     else {
-      if (option.onClick) option.onClick();
-      else
-        props.onChoose
-          ? props.onChoose(option.value)
-          : props.onChange(option.value);
+      if (props.onChoose) props.onChoose(option.value);
+      props.onChange(option.value);
+      if (props.onRequestClose) props.onRequestClose();
     }
   }
-  if (props.onRequestClose) props.onRequestClose();
 };
 
 const getDefaultStylingProps = (
@@ -225,7 +222,7 @@ export default React.forwardRef<Props, SemiControlledAutoCompleteInterface>(
             classes={classes}
             onChange={(
               event: SyntheticKeyboardEvent<HTMLInputElement>,
-              option: Option | string | null
+              option: Option | null
             ) => {
               if (option !== null) {
                 handleChange(event, option, props);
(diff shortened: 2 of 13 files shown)


4ian:
That looks correct!

> I'm aware of not adding this to ResourceSelector since it has some issues with proper selection

Interesting, what's the issue?

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -142,20 +142,17 @@ const filterFunction = (
 
 const handleChange = (
   event: SyntheticKeyboardEvent<HTMLInputElement>,
-  option: Option | string,
+  option: Option,
   props: Props
 ): void => {
   if (option.type !== 'separator') {
-    if (typeof option === 'string') props.onChange(option);
```
HarsimranVirk: This was really unnecessary. I had it in place while replacing downshift. I had a keybinding to 'Enter' which would let autocomplete accept any value. Now, we are using the blur event to accept arbitrary values, and even after adding onRequestClose, I'd like to keep using the blur event, for 2 reasons

1. It remains as close as possible to the real implementation of autocomplete :) Even material-ui's default behaviour is using the blur event, you can check it in the freeSolo demo.

2. It would be a huge headache in the future for developers while understanding, debugging and maintaining the unnecessary keybindings and events and at the same time, it adds little to no value in usability for users :)

newIDE/app/src/ResourcesList/ResourceSelector.js
```diff
@@ -169,6 +170,7 @@ export default class ResourceSelector extends React.Component<Props, State> {
       () => {
         if (!this.state.notExistingError) {
           if (this.props.onChange) this.props.onChange(resourceName);
+          if (this.props.onRequestClose) this.props.onRequestClose();
```
HarsimranVirk: The issue regarding ResourceSelector I believe was not having enough time to properly select the option. The autocomplete was closing even before changing anything. I could've maybe used a timeout for calling onRequestClose, but I think this is a more rigid and proper way, as it would ensure that the necessary changes have been done before closing the popup. The function is not called from autocomplete, rather it is being called from the ResourceSelector itself.

newIDE/app/src/UI/SemiControlledAutoComplete.js
4ian: Ok thanks for doing this simplification :) Keeping the code as simple as possible is an important thing :)

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -141,17 +142,15 @@ const filterFunction = (
...
+    if (option.onClick) option.onClick();
     else {
-      if (option.onClick) option.onClick();
-      else
-        props.onChoose
-          ? props.onChoose(option.value)
-          : props.onChange(option.value);
+      if (props.onChoose) props.onChoose(option.value);
+      props.onChange(option.value);
```
4ian: Are you sure we should call onChoose AND onChange?

HarsimranVirk: No, I'm not sure about them and I forgot to ask that. I've never come across onChoose calls (I know about onClick and onChange only, and there is a comment which mentions onClick has the highest priority). Can you explain what is the purpose and priority of onChoose?

newIDE/app/src/ResourcesList/ResourceSelector.js
4ian: I think I know why: 
https://github.com/4ian/GDevelop/blob/a6ae265705fa6d12bb8d05007dfd446c1e8f782a/newIDE/app/src/ResourcesList/ResourceSelector.js#L159-L175

The `this.props.onChange` is only called in the *callback* of setState. Meaning that if the component is blurred and destroyed, it won't be called, because the component will be unmounted by this time.

We can try to fix that in another PR as this can be tricky.

4ian:
Looks mostly good! We'll fix ResourcesSelector in another PR.

Take a look at onChoose/onChange. I'm not sure if we should call both of them?
Once this is solved we can merge this :)

newIDE/app/src/UI/SemiControlledAutoComplete.js
4ian: The idea is that onChange is to be called when we update the content of the text (in a "semi controlled" fashion, i.e: only when we blur the field).

While onChoose is to be called when we *select* (i.e: click or press enter) an existing item (not something that was entered manually). This is used in ObjectGroupEditor: an object is added to the group when `onChoose` is called. `onChange` is just used to remember the name entered.

newIDE/app/src/ResourcesList/ResourceSelector.js
HarsimranVirk: > Meaning that if the component is blurred and destroyed, it won't be called, because the component will be unmounted by this time.

This was the exact behaviour, no change happened because the component unmounted by this time. I've called the `onRequestClose` in the callback of setState, after `onChange`. Will it bring more errors? (If yes, then I'd drop it for this PR, and work on something optimal and proper in another PR :) )

4ian: > I've called the onRequestClose in the callback of setState, after onChange. Will it bring more errors

In your PR? I don't see any onRequestClose in a callback of a setState?
Calling it after onChange is ok, because onChange will almost never result in closing a dialog - and even if it's the case, the important is to not call anything related to the component in a **callback** (setTimeout/setState callback).
Calling `onRequestClose` after `onChange` (or `onChoose`) is fine, as it's onRequestClose that will trigger the closing (well, probably ;)).

Right now I think it's ResourceSelector that is "buggy", because `onChange` is called in the callback of setState - while it's both:

* not necessary
* making the onChange not work because in the meantime, the component can have been unmounted.

Let's try to fix ResourcesSelector in another PR.

HarsimranVirk:
> We'll fix ResourcesSelector in another PR.

Okay, sure thing :)

> Take a look at onChoose/onChange. I'm not sure if we should call both of them?

I'm not sure either :p I need to check exactly how onChoose is used, so I'd take a look at it first :)

4ian:
> I'm not sure either :p I need to check exactly how onChoose is used, so I'd take a look at it first :)

Verify that `ObjectGroupEditor` still works (you can add an object by selecting it/pressing Enter) and that should be good :) 
Verify that there is no warning shown in the console by React.

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -276,6 +277,12 @@ export default React.forwardRef<Props, SemiControlledAutoCompleteInterface>(
                       props.onChange(event.currentTarget.value);
                       if (props.onBlur) props.onBlur(event);
                     },
+                    onMouseDown: (
+                      event: SyntheticMouseEvent<HTMLInputElement>
+                    ): void => {
+                      if (input.current && !input.current.value.length)
+                        setIsMenuOpen(!isMenuOpen);
+                    },
```
HarsimranVirk: I've added this to toggle menu with a click when input is empty. The menu used to stay open even when the user clicks to close it.

newIDE/app/src/UI/SemiControlledAutoComplete.js
```diff
@@ -148,8 +148,9 @@ const handleChange = (
   if (option.type !== 'separator') {
     if (option.onClick) option.onClick();
     else {
-      if (props.onChoose) props.onChoose(option.value);
-      props.onChange(option.value);
+      props.onChoose
+        ? props.onChoose(option.value)
+        : props.onChange(option.value);
```
HarsimranVirk: We could call both the functions together, there is nothing wrong in doing that, but it doesn't make any sense. Only one of them selects the option (even when I called both of them) as far as I could understand :)

newIDE/app/src/UI/SemiControlledAutoComplete.js
4ian: Thanks for adding this 👍

4ian:
Looks good! Merging now 👍
For the investigation and the great work on this again 👍👍

PR #1570, opened 2020-03-21
Added the attribute for changing the speed of climbing the ladder

PawBud:
This is a fix for the issue: _#1563_
As suggested by @4ian, I am working on  the JS extension of the attribute


4ian:
Your changes in the C++ side are incomplete, you also need to update the header.
This means that GDevelop.js was probably not compiled properly.

When doing changes, you need to test them - it's not just about code :) Read how to compile GDevelop.js, there is a README about it.

PawBud:
yea I realized that

PawBud:
@4ian i have made the required changes though I haven't tested them properly so I would be highly obliged if u may review my code

4ian:
Can you test them? :)

PawBud:
I will be happy too :).

PawBud:
I think I shud remove those spaces caused by prettier, right @4ian?

4ian:
Yes. Though they are not caused by prettier, you probably used auto formatting in VSCode without having Prettier installed.

PawBud:
@4ian yea, I just noticed that. I will revert them back and then force push :)

Extensions/PlatformBehavior/PlatformerObjectRuntimeBehavior.h
```diff
@@ -33,6 +33,7 @@ class GD_EXTENSION_API PlatformerObjectRuntimeBehavior
 
   double GetGravity() const { return gravity; };
   double GetMaxFallingSpeed() const { return maxFallingSpeed; };
+  double GetLadderClimbingSpeed() const {return ladderClimbingSpeed};
```
4ian: Be consistent in the spacings

PawBud: Yup, I am trying to imitate the spaces as they have been done in other files.

PawBud: Ohh! I just noticed the `{return.....}`. Thanks for that

PR #1572, opened 2020-03-21
Moved loadObjects to a registerObjects method

arthuro555:
To make clean dynamic object creation through JS possible.

### DIFF
--- a/GDJS/Runtime/runtimescene.js
+++ b/GDJS/Runtime/runtimescene.js
@@ -149,12 +149,12 @@ gdjs.RuntimeScene.prototype.loadFromScene = function(sceneData) {
  * This step is needed for any object before adding Instances of it to the scene.
  * @param {ObjectData} objectData The objectData of the object to register.
  */
-gdjs.registerObject = function(objectData) {
-        this._objects.put(objectData.name, objectData);
-        this._instances.put(objectData.name, []); //Also reserve an array for the instances
-        this._instancesCache.put(objectData.name, []); //and for cached instances
-        this._objectsCtor.put(objectData.name, gdjs.getObjectConstructor(objectType)); //And cache the constructor for the performance sake
-    }
+gdjs.RuntimeScene.prototype.registerObject = function(objectData) {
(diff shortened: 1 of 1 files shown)


arthuro555:
Ready for review @4ian .

GDJS/Runtime/runtimescene.js
```diff
@@ -156,6 +144,18 @@ gdjs.RuntimeScene.prototype.loadFromScene = function(sceneData) {
     this._timeManager.reset();
 };
 
+/**
+ * Registration of {@link gdjs.RuntimeObject}.
```
4ian: Start description with a verb describing the action done by the function.

GDJS/Runtime/runtimescene.js
```diff
@@ -156,6 +144,18 @@ gdjs.RuntimeScene.prototype.loadFromScene = function(sceneData) {
     this._timeManager.reset();
 };
 
+/**
+ * Registration of {@link gdjs.RuntimeObject}. 
+ * This step is needed for any object before adding Instances of it to the scene.
+ * @param {ObjectData} objectData The objectData of the object to register.
```
4ian: @param {ObjectData} objectData The data for the object to register.

GDJS/Runtime/runtimescene.js
```diff
@@ -156,6 +144,18 @@ gdjs.RuntimeScene.prototype.loadFromScene = function(sceneData) {
     this._timeManager.reset();
 };
 
+/**
+ * Registration of {@link gdjs.RuntimeObject}. 
+ * This step is needed for any object before adding Instances of it to the scene.
```
4ian: I would go for:
`Call this function before adding instances of an object to the scene.`

GDJS/Runtime/runtimescene.js
4ian: `Register an object so that instances of it can be used in the scene`.

4ian:
Thanks, looks good :)

PR #1574, opened 2020-03-21
Remove redundant prompt for cutting objects

### LINKED ISSUE #1573: Confusing prompt for cutting object - same as deleting
## Description
While cutting objects the prompt displayed is the same as for deleting objects which is confusing and seemingly redundant.

## To Reproduce
1. Open a project
2. Try deleting an object
3. See prompt

## Screenshots
Screenshot 2020-03-21 at 7 19 27 PM
Screenshot 2020-03-21 at 7 19 49 PM

## Other details
Mac OSX Mojave
Desktop App

pranavmodx:
Fixes #1573 

There were 2 distinct prompts for the same purpose (for both cutting and deleting, since both make use of same function). The prompts are only required while deleting objects and only 1 is sufficient. 
Necessary changes have been made.


newIDE/app/src/SceneEditor/index.js
```diff
@@ -598,24 +598,19 @@ export default class SceneEditor extends React.Component<Props, State> {
...
-    const answer = confirm(
-      'Do you want to remove all references to this object in groups and events (actions and conditions using the object)?'
-    );
 
     if (global) {
       gd.WholeProjectRefactorer.globalObjectOrGroupRemoved(
         project,
         object.getName(),
-        /* isObjectGroup=*/ false,
-        !!answer
```
4ian: Not passing an argument to gd.xxx methods is very dangerous, this can crash the C++ library.

newIDE/app/src/SceneEditor/index.js
```diff
@@ -598,24 +598,19 @@ export default class SceneEditor extends React.Component<Props, State> {
     const { project, layout } = this.props;
 
     //eslint-disable-next-line
-    const answer = confirm(
-      'Do you want to remove all references to this object in groups and events (actions and conditions using the object)?'
-    );
```
4ian: Have you considered the impact of removing this? 😬

It seems that you've entirely investigated what this feature is and how it is useful to users. I would recommend spend more time investigating and being sure about what your changes are implying.

4ian:
I don't think that's redundant :/

newIDE/app/src/SceneEditor/index.js
pranavmodx: Oh. I should've investigated about these methods before blindly removing the arguments. Pardon me for this carelessness.

newIDE/app/src/SceneEditor/index.js
pranavmodx: I realised the impact moments after I pushed the code. The prompt clearly warns about the same. It's a misjudgement from my side. I shall revert all the undesired changes.

Please correct me if I'm wrong, but I noticed that the statement `if(!answer) return;` was missing soon after this statement, meaning that the object would be deleted regardless of the response by the user. I observed this during debugging.

4ian: > Please correct me if I'm wrong, but I noticed that the statement if(!answer) return; was missing soon after this statement, meaning that the object would be deleted regardless of the response by the user. I observed this during debugging

Yes, because the question asked to the user is about if events should be removed or not. The object will be deleted in any case (ideally, we should show "yes/no" rather than "ok/cancel", i admit it's not super clear).

pranavmodx:
> I don't think that's redundant :/

By redundant I meant to say that for Cut operation for objects, 2 prompts were being displayed. One for confirming the cut operation and the other for alerting the user about the impact, that is, the removal of all references to the object in groups and events. 
The 2nd prompt seemed more meaningful (now), because cancelling the 2nd prompt itself should prevent the operation of cutting in the first place. So I thought the 1st prompt could be integrated into the 2nd.

pranavmodx:
I shall leave this PR if you feel the changes aren't necessary. I didn't mean to point out a flaw. Just thought it would improve the UX a bit. I acknowledge the lack of investigation from my part. I shall be more vigilant in the future.

4ian:
There is surely room for improvement here in the UX :) But I'd like more investigation/explanation because otherwise we lose time reviewing changes that are actually counter productive.

Basically, make sure you master the software before trying to do changes.

PR #1575, opened 2020-03-21
Improve visibility of parameter names in layer effects for checkbox

Bouh:
Before / After
image


4ian:
I usually try to minimize direct usages of material-ui components in non-UI folders, but I think it's fine 👍

Issue #1576, opened 2020-03-21
Some problems with the search feature 


PascalLadalle:
Few bugs I noticed with the search feature.

Ctrl+f:
- Unless I select an event in the sheet, the shortcut doesn't work. For example, if I switch to the scene editor and switch back to the events, the shortcut doesn't work.
- Same goes when the search bar is open. I need to click on an event for the shortcut to work again.

Searching:
After typing the search string, pressing Return should run the search, but I have to hit tab first to move focus to the Search button. 

Vewing results:
Clicking once on the search button runs the search and displays the number of hits. It would feel natural that (unless the search string has been changed) further clicks on this button would scroll through the hits like the right arrow button below.

4ian:
> After typing the search string, pressing Return should run the search, but I have to hit tab first to move focus to the Search button.

Should be easy to add! :) 

> Ctrl+f:

It's a question of handling the focus, it's a bit harder to get right.

> Clicking once on the search button runs the search and displays the number of hits. It would feel natural that (unless the search string has been changed) further clicks on this button would scroll through the hits like the right arrow button below.

Good idea, should be also easy to add

jimil749:
Also I think the search bar should have an exit button (or a cross) to exit. Pressing Ctrl+F again (that too does not work when an event is not selected) seems pretty cumbersome.

4ian:
Not sure about this. Sublime Text has a search bar that can be dismissed with Escape. I think that would be a better solution.

jimil749:
That could also be a solution, but currently that is not implemented, need to press Ctrl + F to exit the search bar.  I'll try to implement the changes! :)

jimil749:
Hey @4ian, I made the Enter key working and also made click on search scroll the list of searched items. Also added the ESC key to close the search panel, but has one issue, it won't close until a text is searched or until an event is focused (just like search won't open until an event is selected). Stuck here, any suggestions? : \

4ian:
Hard to say without seeing the code :) Push on a branch on GitHub and put the link to your commit.

PR #1582, opened 2020-03-23
feat: search feature

jimil749:
PR for issue #1576 

Working : 
-Search functionality on pressing return
-Repeatedly pressing on search iterates through all the matched searches.
-Pressing Escape quits the search panel

Not Working:
-Escape only works when an event is selected.

Might require some optimization on repeated search!

Note that this PR might require some changes/optimization. Does the work for now.


newIDE/app/src/EventsSheet/SearchPanel.js
```diff
@@ -116,6 +116,13 @@ export default class SearchPanel extends PureComponent<Props, State> {
               }
               hintText={t`Text to search in parameters`}
               onChange={(e, searchText) => this.setState({ searchText })}
+              onKeyPress={event => {
+                if (event.charCode === 13) {
```
4ian: Use event.code to check for "Enter", and you can then remove the comment that is not useful anymore :)

newIDE/app/src/EventsSheet/SearchPanel.js
```diff
@@ -116,6 +116,13 @@ export default class SearchPanel extends PureComponent<Props, State> {
...
+                  //enter key pressed
+                  if (this.searchTextField) this.searchTextField.blur();
```
4ian: I would avoid to blur something, instead see if you can focus something else :)

newIDE/app/src/EventsSheet/SearchPanel.js
```diff
@@ -124,7 +131,10 @@ export default class SearchPanel extends PureComponent<Props, State> {
               disabled={!searchText}
               primary
               label={<Trans>Search</Trans>}
-              onClick={this.launchSearch}
+              onClick={() => {
+                this.launchSearch();
+                onGoToNextSearchResult();
```
4ian: Should this be called even if we don't have result already? (not sure, to be checked :))

newIDE/app/src/EventsSheet/SearchPanel.js
jimil749: That was just to remove focus on the text-field. Is it fine if I just remove the line? It is not that important

newIDE/app/src/UI/KeyboardShortcuts/index.js
```diff
@@ -152,6 +155,9 @@ export default class KeyboardShortcuts {
     if (onSearch && this._isControlOrCmdPressed() && evt.which === F_KEY) {
       onSearch();
     }
+    if (onQuitSearch && evt.which === ESC_KEY) {
```
4ian: As this is generic, let's call it "onEscape"

newIDE/app/src/EventsSheet/SearchPanel.js
jimil749: Bit confused (As in did not get what you said ). This is to iterate through the matched searches on the click of search button.

4ian: This is the search button. When you click on it for the first time, you:
- launch the search (this is fine)
- then you go to the next search result <-- is this something that we want to do even if you just click on search for the first time.

jimil749: I think you are right, we don't want it to go to the next search on the first click, it might be wrongly placed.
Here the search button has 2 functionalities: 
1) Search for the text.
2) Move to next found text.
So, when the button is clicked the first time, it would search for the text. But for the subsequent button presses it should move to the next search. But we might have a roadblock. What if the text to be searched is changed? We need to reset the button press count and again do (1).

jimil749:
I done the required changes, expect the onGoToNextSearchResult() call issue. Once this is resolved, will send a commit!

newIDE/app/src/EventsSheet/SearchPanel.js
4ian: > But we might have a roadblock. What if the text to be searched is changed? We need to reset the button press count and again do (1).

Exactly. This means that there must be maybe in the state a boolean or something like this to indicate if we've ran the search once or not - and properly reset this boolean when needed.

jimil749:
Oops was a bit careless. Will commit new one!

jimil749:
@4ian Can you take a look?

newIDE/app/src/EventsSheet/index.js
```diff
@@ -281,6 +282,12 @@ export default class EventsSheet extends React.Component<Props, State> {
     );
   };
 
+  _escapeSearchPanel = () => {
```
4ian: _closeSearchPanel

newIDE/app/src/EventsSheet/index.js
```diff
@@ -281,6 +282,12 @@ export default class EventsSheet extends React.Component<Props, State> {
...
+    if (this.state.showSearchPanel) {
+      this.setState({ showSearchPanel: false });
+    }
```
4ian:
```suggestion
    this.setState({ showSearchPanel: false });
```

works as well in this case

newIDE/app/src/EventsSheet/SearchPanel.js
```diff
@@ -124,7 +135,15 @@ export default class SearchPanel extends PureComponent<Props, State> {
...
               label={<Trans>Search</Trans>}
-              onClick={this.launchSearch}
+              onClick={() => {
+                if (this.state.searchCount) {
+                  onGoToNextSearchResult();
+                } else {
+                  this.launchSearch();
+                  this.setState({ searchCount: true });
+                }
+                //onGoToNextSearchResult();
```
4ian: Remove any commented code

newIDE/app/src/EventsSheet/SearchPanel.js
```diff
@@ -33,6 +33,7 @@ type State = {|
   searchInConditions: boolean,
   searchInEventStrings: boolean,
   searchInSelection: boolean,
+  searchCount: boolean,
```
4ian: The name searchCount sounds like a "number of search"/"how many time a search was made".
Can you think of another name representing what this thing is. It's a boolean, so reading the name must immediately gives an idea of what true and false mean :)

newIDE/app/src/EventsSheet/SearchPanel.js
```diff
@@ -115,7 +117,16 @@ export default class SearchPanel extends PureComponent<Props, State> {
                 (this.searchTextField = _searchTextField)
               }
               hintText={t`Text to search in parameters`}
-              onChange={(e, searchText) => this.setState({ searchText })}
+              onChange={(e, searchText) => {
+                this.setState({ searchText });
+                this.setState({ searchCount: false });
```
4ian: Do a single setState

4ian:
Seems like you're on the right track 👍
I've added mostly comments about code style and naming of variables :)

newIDE/app/src/EventsSheet/SearchPanel.js
jimil749: maybe isSearchCountOne or checkSearchCount? This state is just check if it is searched once or not. So isSearchCountOne seems better option.

newIDE/app/src/EventsSheet/SearchPanel.js
```diff
@@ -33,7 +33,7 @@ type State = {|
   searchInConditions: boolean,
   searchInEventStrings: boolean,
   searchInSelection: boolean,
-  searchCount: boolean,
+  isSearchCountOne: boolean,
```
4ian: This name is a bit better :) 
I would recommend to see it from the perspective of the search results:
Usually when we want to track if something was changed, we use what is called a "dirty flag".
Here I would use a flag called "searchResultsDirty".
it means: "when I'm true, the search results are 'dirty', i.e: they are not up to date anymore". In other words, a new search must be done.

Meaning that:
1. When search text is changed, the dirty flag is true
2. When the search is done, the dirty flag is false
3. When the events are changed, the dirty flag must be set to true again

So I would recommend to rename this to `searchResultsDirty` and invert the usage of true/false :)  
For 3., this means that the EventsSheet should call the search panel to tell to it "hey the events were modified". It can do this by using `this._searchPanel` and calling a function "markSearchResultsAsDirty" - which itself is just modifying the booleans.

jimil749:
@4ian I have pushed a new commit, made some changes in the search feature. 
What it does is :  
1) Search updates when the text is being changed
2) Search updates when events are changed.

newIDE/app/src/EventsSheet/SearchPanel.js
```diff
@@ -1,4 +1,10 @@
 // @flow
+/*
+  When to launch a new search :
+    -> enter pressed (searchResultsDirty : false)
+    -> search button pressed
+    -> searchResultsDirty is true, and then make it false.
```
4ian, line 6: Let's rather have a comment in the code itself. It has too much chance to be forgotten when there is an update if it's a the top of the file.

newIDE/app/src/EventsSheet/SearchPanel.js
```diff
@@ -53,6 +61,12 @@ export default class SearchPanel extends PureComponent<Props, State> {
     }
   };
 
+  makeSearchResultsDirty = () => {
+    this.setState({ searchResultsDirty: true }, () => {
+      this.callLaunchSearch();
```
4ian, line 66: Why launching the search after making the search result dirty?

newIDE/app/src/EventsSheet/SearchPanel.js
```diff
@@ -115,7 +136,22 @@ export default class SearchPanel extends PureComponent<Props, State> {
...
               hintText={t`Text to search in parameters`}
-              onChange={(e, searchText) => this.setState({ searchText })}
+              onChange={(e, searchText) => {
+                this.setState(
+                  {
+                    searchText,
+                    searchResultsDirty: true,
+                  },
+                  () => {
+                    this.callLaunchSearch();
```
4ian, line 146: Why launching the search after every change? That sounds like a different behavior than before, not sure if it's something that we want?

jimil749: I that we wanted the search to update if the search was being changed. 
So, let's suppose I search for 'PlayerHitBox' and then I just remove 'HitBox' part, the search result update here automatically, without enter press. I thought we wanted to update search as the user types. My bad, I will revert it.

newIDE/app/src/EventsSheet/SearchPanel.js
jimil749: Will do that!

newIDE/app/src/EventsSheet/SearchPanel.js
jimil749: If the events are changed, we also need to update the search, right? What this does is, whenever I change the events it updates (rather 're-searches') to get the correct results. Not sure, if its most optimal. But works fine.

jimil749:
@4ian, I did some changes, can you take a look?

Bouh:
> I did some changes, can you take a look?

There is a button for ask a new review on this page ;)

jimil749:
@4ian, Just a brief description of what this PR does.
(Note: This PR addresses the issue of Searching and viewing the result)

- Executes search on press of return 
- The search results move on to the next one on press of search key. (as mentioned in the issue)

I reverted the changes made in the previous commit, so now what's happening is :
- Whenever search is made, I set the dirty flag to false, meaning that the search isn't dirty and the text is not changed.
- Whenever search text is changed or the event is changed the dirty flag is set to true suggesting that search results are dirty, meaning we require to search again.

I tested it, works as expected!

newIDE/app/src/EventsSheet/SearchPanel.js
```diff
@@ -96,6 +102,13 @@ export default class SearchPanel extends PureComponent<Props, State> {
     });
   };
 
+  callLaunchSearch = () => {
```
4ian, line 105: Let's find a better name for this, because "call" is not conveying the indication of what is happening in this function.
Something like "launchSearchIfResultsDirty" (even if I'm not totally satisfied with the name, but can't think of anything better).

jimil749:
Changed to function name to `launchSearchIfResultsDirty`, as suggested, because we are using it to call `launchSearch` if the search results are dirty.

jimil749:
I don't know why tests fail, it passes locally. :/

4ian:
> I don't know why tests fail, it passes locally. :/

Run npm install to ensure that you have the latest version of all libraries.

Typically, you should not change the `HelpButton.spec.js.snap` file. You can see this because it's unrelated to your PR and is part of a test of something else in the app. It's a good sign that this should not be included and is probably an out-of-date dependency on your side (or something broken in master).

jimil749:
Sorry about the last commit, I will try to fix the tests asap!

4ian:
No worries - you might need to merge or rebase on your changes on master (and re run npm install) to get the test working :)

jimil749:
I re-ran npm install and reverted the changes done in the `HelpButton.spec.js.snap` and `HelpIcon.spec.js.snap`, which passes the tests!

4ian:
Merged thanks for your work on this! 👍

jimil749:
Thank you for merging! :)

PR #1583, opened 2020-03-23
Fix crash Anchor behavior by adding setWidth & setHeight

Bouh:
This patch fix a crash when the anchor behavior use right and bottom properties with several object types.

The game freeze because objects haven't the method `setWidth()` and `setHeight()`.
[Picture of issue in console.](https://user-images.githubusercontent.com/1670670/77354962-f6daeb00-6d43-11ea-946f-2bba68a41b54.png)

This patch fix the object **text** and **BBtext**.
The last object type bugged is the ShapePainterRuntimeObject.
I've drawn a square of 100x100px at 0;0
I can't get the width and height of the shape painter, value should be 100x100 but both are on 0 during the preview.
I can't find a correct value in Pixi documentation, or by digging the values of the object when i set pause in the debugger of the preview.

All other objects types are tested and working (or atleast not freezing the game).
This bug exist since GD4, for web export only. Working great in native c++ in GD4.


4ian:
> The game freeze because objects haven't the method setWidth() and setHeight().

This is indeed an error, all objects should have setWidth/setHeight, even if they are doing nothing.

> I can't get the width and height of the shape painter, value should be 100x100 but both are on 0 during the preview.

This is because the shape painter does not really have a size - even if you draw something on screen, it is not tracking what is being drawn. And should probably not. Instead, we may want to add two properties and actions/conditions to set the size of the object. This won't have an effect on what is shown on screen, just used for collisions and anchor behavior :)

Extensions/BBText/bbtextruntimeobject-pixi-renderer.js
```diff
@@ -113,3 +113,11 @@ gdjs.BBTextRuntimeObjectPixiRenderer.prototype.getWidth = function() {
 gdjs.BBTextRuntimeObjectPixiRenderer.prototype.getHeight = function() {
   return this._pixiObject.height;
 };
+
+gdjs.BBTextRuntimeObjectPixiRenderer.prototype.setWidth = function(newWidth) {
+  this._pixiObject.width = newWidth;
```
4ian: What is this doing to the pixi object? Is there something visible on screen?

Extensions/TextObject/textruntimeobject-pixi-renderer.js
```diff
@@ -105,6 +105,14 @@ gdjs.TextRuntimeObjectPixiRenderer.prototype.getHeight = function() {
     return this._text.height;
 };
 
+gdjs.TextRuntimeObjectPixiRenderer.prototype.setWidth = function(newWidth) {
+    this._text.width = newWidth;
```
4ian: Same, what is the result on screen of doing this?

Bouh:
Something in code are wrong when you or Victor have implemented this behavior.
Or I really not understood how this behavior should be working correctly.

With a sprite the left side anchored on left border, then spacing between left border and left side of the objects is always same. Width and height doesn't change.
Same for top. Behavior seem working great?

When I setup the anchor bottom on bottom border, object is streched, bottom side is moved(ie: width is affected, but top side isn't moved, so sprite is streched).
Same for right. This seem not what we expected.

It's like the top was pinned on top border, and bottom on bottom border too.

We need fix this before review for Text and BBtext, and ShapePainter objects.


See two previews with same properties.
https://streamable.com/n6lqk
Here i have two results, which one is the correct? 

For me the left preview isn't correct and right preview is correct.
Right is mine with a fix, left is current b89.

Extensions/BBText/bbtextruntimeobject-pixi-renderer.js
Bouh: Text is stretched from the middle, not the best thing to do i agree, let starting by the first point, see my anwser bellow. 
Then we will be able to fix this.

4ian:
> With a sprite the left side anchored on left border, then spacing between left border and left side of the objects is always same [...] Same for top. Behavior seem working great?

This is indeed expected.

> When I setup the anchor bottom on bottom border, object is streched, bottom side is moved(ie: width is affected, but top side isn't moved, so sprite is streched).
Same for right. This seem not what we expected.

This is I think expected. If you put the **bottom anchor** on **bottom border**, you're saying "the spacing between the bottom of the object and the bottom of the window must not change". How to do that? By stretching the object, i.e: changing it's height. Same for width.

You might think: in this case, why are you not changing the object position? You can do this, but in this case you must set the **top anchor** of the object to be anchored to **the bottom of the window**.

In other words, if you never want your object stretched, **you must only use top and left anchor**. I'm almost thinking of hiding the right and bottom anchor under a "advanced" button.

> Here i have two results, which one is the correct? 

Sorry I'm not entirely sure i see a difference 😬 Can you make a screenshot with initial position and then after resizing window? 

You've put "**Right edge anchor**" to **window right**. This means that the space between the right of the object and the window must stay the same. So the object will be resized.

I understand this is confusing. I've been confused too. Maybe we should adapt how we present stuff, give more explanation, or a diagram...?

Bouh:
For both preview start by 800x600, window isn't resize.
This is the state on scene : 
image

On left preview, the object stone is streched when i launch the previex, but on the scene the object is set on 32x32. This is for me not expected.
on right preview i've set a fix for avoid strech at the start of the preview.
Both object are streched only when border move: that's is good.

4ian:
> on right preview i've set a fix for avoid strech at the start of the preview.
> Both object are streched only when border move: that's is good.

This seems better indeed. Would be interested to see the fix.

Bouh:
The behavior isn't understandable at all the first time.

Because we have two possibility, fix the object on border or fix a side of the object on border.
That's why we are loose in the options.

I think that's why unity team have made preset for it:
image

I agree all case on screen are actually doable with current options, but isn't intuitive at all.

Extensions/BBText/bbtextruntimeobject-pixi-renderer.js
4ian: Can we just not implement this function (I mean, keep it empty) - because I don't think the text object was made to be stretched.

At least this is not a bug fix anymore but a new feature - and one to be considered very carefully because this could create confusion/complexity, especially with the fact that the text can be wrapped.

Extensions/TextObject/textruntimeobject-pixi-renderer.js
4ian: Let's not implement this - it's ok for objects to "not follow orders" and don't update their size.

4ian:
Their approach is definitely easier to understand. We could rework the UI to have buttons like this.
For now to make a bug fix, we can start by adding the setWidth/setHeight - though I think we should NOT implement them, and not add them in renderers for objects that don't support resizing (text objects).

Bouh:
> though I think we should NOT implement them, and not add them in renderers for objects that don't support resizing (text objects).

In this case we should not try to use setWidth/setHeight for objects who have not these. that's why bug is present and freeze games like this one.
[Anchor behavior on Text game example freeze.](https://s3-eu-west-1.amazonaws.com/gd-games-preview/1585516815353-40634/index.html)

blurymind:
I always thought that the intent was to use the ide's resize widget in order to set their text wrapping width property in a user friendly way. Removing the ability to do that will make it impossible to do any dialogue text box

4ian:
> Removing the ability to do that will make it impossible to do any dialogue text box. It will probably break the dialogue tree example too.

We're not removing anything here :)

> But I guess this is probably more about having to add that setwidht and height gdevelop specific functions?

Yes - these objects don't have the usual setWidth/setHeight because it makes no sense to change the height of a text object. It might make sense if you think in terms of scaling the object, but in this case we have to make the width to work like the height (scaling the object), this is something different than the wrapping width.

In any case, we're NOT removing the wrapping width or the ability to wrap.

It's just that the engine expect setWidth/setHeight to exist.
So two solutions:

- Either we add these methods, even if they do nothing.
- Or we don't add these methods and make anchor behaviors resilient against this.

Extensions/AnchorBehavior/anchorruntimebehavior.js
```diff
@@ -148,9 +148,9 @@ gdjs.AnchorRuntimeBehavior.prototype.doStepPreEvents = function(runtimeScene) {
 
         //Move and resize the object according to the anchors
         if(this._rightEdgeAnchor !== gdjs.AnchorRuntimeBehavior.HorizontalAnchor.NONE)
-            this.owner.setWidth(bottomRightCoord[0] - topLeftCoord[0]);
+            this.owner.setWidth(bottomRightCoord[0]- this.owner.getX() - topLeftCoord[0]);
         if(this._bottomEdgeAnchor !== gdjs.AnchorRuntimeBehavior.VerticalAnchor.NONE)
-            this.owner.setHeight(bottomRightCoord[1] - topLeftCoord[1]);
+            this.owner.setHeight(bottomRightCoord[1]- this.owner.getY() - topLeftCoord[1]);
```
4ian, line 153: @Bouh I see you closed this PR, but actually these 2 lines were fixing an issue! 
We should probably have a new PR with just this? (maybe re-test this, but I think it was useful)

Bouh:
I'am not sure what should be the right behavior for the Anchor.
[Here how this PR work with the lastest master.](https://streamable.com/88sg7b)

arthuro555:
I think this PR is irelevant as it got fixed in the meantime by adding a default setWidth and setHeight method to RuntimeObject

Extensions/AnchorBehavior/anchorruntimebehavior.js
Bouh: @arthuro555  there are thes lines to test again, if it doesn't change anything i'll close this PR.

Extensions/TextObject/textruntimeobject-pixi-renderer.js
```diff
@@ -105,6 +105,10 @@ gdjs.TextRuntimeObjectPixiRenderer.prototype.getHeight = function() {
     return this._text.height;
 };
 
+gdjs.TextRuntimeObjectPixiRenderer.prototype.setWidth = function(newWidth) {};
+
+gdjs.TextRuntimeObjectPixiRenderer.prototype.setHeight = function(newHeight) {};
+
```
Bouh, line 111: Note to me : 
I can clean these lines, it's done [here now](https://github.com/4ian/GDevelop/commit/0d5fbdabc998a3673672dcbac8a6d4ade33a6cd0).

Extensions/BBText/bbtextruntimeobject-pixi-renderer.js
```diff
@@ -112,3 +112,7 @@ gdjs.BBTextRuntimeObjectPixiRenderer.prototype.getWidth = function () {
 gdjs.BBTextRuntimeObjectPixiRenderer.prototype.getHeight = function () {
   return this._pixiObject.height;
 };
+
+gdjs.BBTextRuntimeObjectPixiRenderer.prototype.setWidth = function(newWidth) {};
+
+gdjs.BBTextRuntimeObjectPixiRenderer.prototype.setHeight = function(newHeight) {};
```
Bouh, line 118: Note to me : 
I can clean these lines, it's done [here now](https://github.com/4ian/GDevelop/commit/0d5fbdabc998a3673672dcbac8a6d4ade33a6cd0).

Extensions/AnchorBehavior/anchorruntimebehavior.js
4ian: Let me know and feel free to adapt or open the PR if these are useful :)

Bouh: After an test it's no longer necessary.
I close.

PR #1586, opened 2020-03-24
Added event listeners on listbox to stop propagation.

### LINKED ISSUE #1562: Holding click and moving disables selection in events sheet
## Describe the bug
When choosing an item in a list, if you release click on a different item than the one you pressed the click on, the selection doesn't occur.

Example:
image
Here, I press click on Ssmoke, move the cursor to Tsmoke and release the click. 
Expected result: Tsmoke is selected.
Actual result: the menu closes but the object has not been selected.

Note: with the right-click actions/conditions menu, the item isn't selected either, but the menu remains open, which makes it less of a problem.

HarsimranVirk:
Resolves #1562
@4ian I tried `stopPropagation()` but it doesn't work. I think maybe because the menu and ClickAwayListener do not belong to the same DOM tree and have no relationship between them. Though portals allow bubbling through different DOM trees, I don't really know what does it even mean to stop the propagation in this case (since they have no real parent-child relationship?).

Anyway, should we use `stopImmediatePropagation()`? It's called in the event listener of listbox, which means our selection should be done by that time, and we wouldn't really need the event anymore. What do you suggest?

### DIFF
--- a/newIDE/app/src/UI/SemiControlledAutoComplete.js
+++ b/newIDE/app/src/UI/SemiControlledAutoComplete.js
@@ -238,6 +238,14 @@ export default React.forwardRef<Props, SemiControlledAutoCompleteInterface>(
             options={props.dataSource}
             renderOption={renderItem}
             getOptionDisabled={isOptionDisabled}
+            ListboxProps={{
+              onClick: (event: SyntheticMouseEvent<HTMLUListElement>) => {
+                event.nativeEvent.stopImmediatePropagation();
+              },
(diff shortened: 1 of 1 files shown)


4ian:
Interesting, I didn't know about stopImmediatePropagation. Sounds fair to use.
We would need to add an extensive comment describing why it's used, in which situation it's used, a link to the material-ui issue (so that we can find it easily in the codebase) and possibly a link to the GDevelop issue.

The idea is to make this code very visible to the future reader that would search for such a hack or trying to understand why an event is not propagated when using an autocomplete :)

4ian:
Sorry for the delay! Looks good to go 👍

Issue #1579, opened 2020-03-22
No close button on EditBehaviourDialog


gautamv95:
## Description
It would be great if the following dialog had a close button rather than having just apply!

![Screenshot_25](https://user-images.githubusercontent.com/45149542/77243837-afb20480-6c34-11ea-9568-21a19b626c92.png)

4ian:
Adding a cancel button is tricky because any change in the properties must result in calls to refactor the whole project (to rename the properties). Right now, any renaming is instantly applied, so we can't cancel. To have a cancel button, the dialog must be reworked so that Apply is launching the refactoring of the project to rename the properties that were renamed.

arthuro555:
Huh is it that difficult? I would have thought this would only have been a matter of storing the values in the state and adding to the apply button a function that applies the states values to the behavior on click

4ian:
See my previous answer. To be more precise, the difficulty is about calls to `gd.WholeProjectRefactorer` in `_onBehaviorPropertyRenamed`. It has to be reworked so that all the renamed properties are triggering a refactoring of the whole project `gd.WholeProjectRefactorer.renameBehaviorProperty` when the dialog is closed using Apply. This means that you have to keep a state of renamed properties and give it to the apply method.

Currently, `EventsBasedBehaviorPropertiesEditor` is calling `onRenameProperty` as soon as a property is renamed - making "cancelling" impossible because the rest of the project was updated for the new property name.

Not super difficult, but "medium difficult" I would say - you have to understand properly how renaming is done.

PR #1587, opened 2020-03-25
Added the close button

PawBud:
This is a fix for the issue: _#1579_

@4ian, I am sure that I have added the close button physically but it's not working. I just wanted some help with my code. Can you tell me in brief as to what components do I have to add? Moreover, is the approach that I am taking correct in the first place or is there a completely different 
workaround?
I added the `Flatbutton` element in `EventsBasedBehaviorEditorDialog.js` And added the necessary changes to the files where `onClose` was a prop.
Thanks :)

### DIFF
--- a/newIDE/app/src/EventsFunctionsExtensionEditor/index.js
+++ b/newIDE/app/src/EventsFunctionsExtensionEditor/index.js
@@ -862,6 +862,7 @@ export default class EventsFunctionsExtensionEditor extends React.Component<
             project={project}
             eventsFunctionsExtension={eventsFunctionsExtension}
             eventsBasedBehavior={editedEventsBasedBehavior}
+            onClose={() => this._editBehavior(null)}
             onApply={() => this._editBehavior(null)}
             onRenameProperty={(oldName, newName) =>
               this._onBehaviorPropertyRenamed(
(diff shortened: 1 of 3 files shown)


newIDE/app/src/EventsFunctionsExtensionEditor/index.js
```diff
@@ -862,6 +862,7 @@ export default class EventsFunctionsExtensionEditor extends React.Component<
             project={project}
             eventsFunctionsExtension={eventsFunctionsExtension}
             eventsBasedBehavior={editedEventsBasedBehavior}
+            onClose={() => this._editBehavior(null)}
```
PawBud: I am sure this needs to change

PawBud:
@Bouh @4ian May I get some help?

4ian:
This is fairly complex, that's why I've not marked #1579 as a Good First Issue. You probably want to use:
* `withSerializableObject` to make a copy of the behavior. It will handle the apply operation for you. 
* You need to do the job of `onRenameProperty` for all the properties that were renamed in the apply function (i.e: when the user closes by clicking on Apply, then `withSerializableObject` will update the behavior and you also need to get the list of renamed properties and call this._onBehaviorPropertyRenamed

PawBud:
thank a lot, I will look into it :).

PawBud:
> This is fairly complex, that's why I've not marked #1579 as a Good First Issue. You probably want to use:
> 
> * `withSerializableObject` to make a copy of the behavior. It will handle the apply operation for you. Check in the codebase how it's used.
> * You need to do the job of `onRenameProperty` for all the properties that were renamed in the apply function (i.e: when the user closes by clicking on Apply, then `withSerializableObject` will update the behavior and you also need to get the list of renamed properties and call this._onBehaviorPropertyRenamed

PR #1588, opened 2020-03-25
[WIP] Feature: Add functionality to open recent projects

### LINKED ISSUE #1356: Automatically re-open the last opened project
So I recently added a feature like this to my svete+electron app and this is something that really makes it nice to use it imo. Basically the idea is:
If you are using electron and you closed the app without closing the project before that, next time you start it, electron remembers your last opened project and  utomatically loads it - so you are straight in it without the need to navigate to its folder.

If you close the project and then close the app, when you start it, the next time you start it - you are offered the standard dialog to create/open a project. This also happens if the autoload fails to load the project because it was moved (path changed)

This is kind of something we take for granted in most programming Ides and some other game engines. With vscode or gitkracken for example, when you open it, you are straight back in your project - you dont have to navigate to it.

Additionally nice to have would be a file> recent menu where one can reopen one of the 5 last edited projects

HarsimranVirk:
Resolves #1356

@4ian I don't know if I'm doing this the proper way or not, so I'd need some guidance :)

So far I've implemented Open Recents menu and used an ipcRenderer event to open the selected project.

Screenshot from 2020-03-26 00 15 11

Also, I work on Linux, so I'm not sure whether this would work for Windows and macOS or not. I believe macOS has built-in support for recent documents by electron (ie, there is a `role`, namely `recentDocuments` for macOS).

Edit: I know @blurymind mentioned something entirely different in that issue, I guess we can reach that state when GD opens with the exact UI state, but let's start with these small steps :)

### DIFF
--- a/newIDE/app/src/MainFrame/ElectronMainMenu.js
+++ b/newIDE/app/src/MainFrame/ElectronMainMenu.js
@@ -15,6 +15,7 @@ type Props = {|
 
 type MainMenuEvent =
   | 'main-menu-open'
+  | 'main-menu-recent'
   | 'main-menu-save'
   | 'main-menu-save-as'
   | 'main-menu-close'
@@ -37,6 +38,7 @@ type MenuItemTemplate =
       accelerator?: string,
       enabled?: boolean,
       label?: string,
+      argument?: Object,
     |}
   | {|
       submenu: Array<MenuItemTemplate>,
@@ -78,6 +80,13 @@ class ElectronMainMenu extends React.Component<Props, {||}> {
       ('main-menu-open': MainMenuEvent),
       event => this._editor && this._editor.chooseProject()
     );
+    ipcRenderer.on('main-menu-recent', (event, fileMetadata) => {
+      if (this._editor) {
+        this._editor.openFromFileMetadata(fileMetadata).then(() => {
+          this._editor && this._editor.openSceneOrProjectManager();
+        });
+      }
+    });
     ipcRenderer.on(
       ('main-menu-save': MainMenuEvent),
       event => this._editor && this._editor.saveProject()
@@ -145,6 +154,26 @@ class ElectronMainMenu extends React.Component<Props, {||}> {
     }
   }
 
+  getRecentFiles = () => {
+    const recentFiles = localStorage.getItem('gd-recent-files');
+    if (recentFiles) {
+      const parsedData = JSON.parse(recentFiles);
+      const menuArray = [];
+      for (const prop in parsedData) {
+        console.log(parsedData[prop]);
+        if (ipcRenderer) {
+          menuArray.push({
+            label: parsedData[prop],
+            onClickSendEvent: 'main-menu-recent',
+            argument: { fileIdentifier: parsedData[prop] },
+          });
+        }
+      }
+      return menuArray;
+    }
+    return [{ label: 'No recent files' }];
+  };
+
   _buildAndSendMenuTemplate() {
     const { i18n } = this.props;
     const fileTemplate = {
@@ -160,6 +189,10 @@ class ElectronMainMenu extends React.Component<Props, {||}> {
           label: i18n._(t`Open...`),
           accelerator: 'CommandOrControl+O',
           onClickSendEvent: 'main-menu-open',
+        },
+        {
+          label: i18n._(t`Open Recent...`),
+          submenu: this.getRecentFiles(),
         },
         { type: 'separator' },
         {
--- a/newIDE/app/src/MainFrame/ElectronMainMenu.js
+++ b/newIDE/app/src/MainFrame/ElectronMainMenu.js
@@ -6,7 +6,6 @@ import { type I18n as I18nType } from '@lingui/core';
 import { t } from '@lingui/macro';
 import { isMacLike } from '../Utils/Platform';
 import PreferencesContext from './Preferences/PreferencesContext';
-import {type FileMetadata} from '../ProjectsStorage';
 const electron = optionalRequire('electron');
 const ipcRenderer = electron ? electron.ipcRenderer : null;
 
@@ -163,8 +162,8 @@ class ElectronMainMenu extends React.Component<Props, {||}> {
       return {
         label: item.fileIdentifier,
         onClickSendEvent: 'main-menu-recent',
-        argument: item
-      }
+        argument: item,
+      };
     });
   }
 
--- a/newIDE/app/src/MainFrame/ElectronMainMenu.js
+++ b/newIDE/app/src/MainFrame/ElectronMainMenu.js
@@ -5,6 +5,8 @@ import optionalRequire from '../Utils/OptionalRequire';
 import { type I18n as I18nType } from '@lingui/core';
 import { t } from '@lingui/macro';
 import { isMacLike } from '../Utils/Platform';
+import PreferencesContext from './Preferences/PreferencesContext';
+import {type FileMetadata} from '../ProjectsStorage';
 const electron = optionalRequire('electron');
 const ipcRenderer = electron ? electron.ipcRenderer : null;
 
@@ -72,6 +74,7 @@ type RootMenuTemplate =
 class ElectronMainMenu extends React.Component<Props, {||}> {
   _editor: ?MainFrame;
   _language: ?string;
+  recentFiles: Array<MenuItemTemplate>;
 
   componentDidMount() {
     if (!ipcRenderer) return;
@@ -154,25 +157,16 @@ class ElectronMainMenu extends React.Component<Props, {||}> {
     }
   }
 
-  getRecentFiles = () => {
-    const recentFiles = localStorage.getItem('gd-recent-files');
-    if (recentFiles) {
-      const parsedData = JSON.parse(recentFiles);
-      const menuArray = [];
-      for (const prop in parsedData) {
-        console.log(parsedData[prop]);
-        if (ipcRenderer) {
-          menuArray.push({
-            label: parsedData[prop],
-            onClickSendEvent: 'main-menu-recent',
-            argument: { fileIdentifier: parsedData[prop] },
-          });
-        }
+  componentWillMount() {
+    const { getRecentFiles } = this.context;
+    this.recentFiles = getRecentFiles().map(item => {
+      return {
+        label: item.fileIdentifier,
+        onClickSendEvent: 'main-menu-recent',
+        argument: item
       }
-      return menuArray;
-    }
-    return [{ label: 'No recent files' }];
-  };
+    });
+  }
 
   _buildAndSendMenuTemplate() {
     const { i18n } = this.props;
@@ -192,7 +186,7 @@ class ElectronMainMenu extends React.Component<Props, {||}> {
         },
         {
           label: i18n._(t`Open Recent...`),
-          submenu: this.getRecentFiles(),
+          submenu: this.recentFiles,
         },
         { type: 'separator' },
         {
@@ -400,5 +394,6 @@ class ElectronMainMenu extends React.Component<Props, {||}> {
     });
   }
 }
+ElectronMainMenu.contextType = PreferencesContext;
 
 export default ElectronMainMenu;
--- a/newIDE/app/src/MainFrame/ElectronMainMenu.js
+++ b/newIDE/app/src/MainFrame/ElectronMainMenu.js
@@ -73,7 +73,6 @@ type RootMenuTemplate =
 class ElectronMainMenu extends React.Component<Props, {||}> {
   _editor: ?MainFrame;
   _language: ?string;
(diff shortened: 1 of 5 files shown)


newIDE/app/src/MainFrame/ElectronMainMenu.js
```diff
@@ -145,6 +154,26 @@ class ElectronMainMenu extends React.Component<Props, {||}> {
     }
   }
 
+  getRecentFiles = () => {
+    const recentFiles = localStorage.getItem('gd-recent-files');
+    if (recentFiles) {
+      const parsedData = JSON.parse(recentFiles);
+      const menuArray = [];
+      for (const prop in parsedData) {
```
4ian: Ideally we want to avoid:
- having direct usage of localStorage inside our components (this might impact performance as it's a synchronous API). It's also a dependency that we may want to replace by something else later.
- having this logic related to the recent files here in ElectronMainMenu and also in MainFrame. Ideally both of them would just have functions to say "Hey this file was opened" and "Hey I want the list of recent files".

To do this, I suggest to look at `PreferencesContext`. It's a React context already used to get and save all the preferences of the app.
We could add two functions, one to get the recent files, one to declare a new file that is opened.
Then, we can use the context Consumer to pass to ElectronMainMenu and MainFrame the functions to get the recent files or declare a file.

If later the logic of the files or the preferences are becoming too complex, we can always split the preferences in multiple contexts... without changing the components. The components will only care about their props having the functions to get the recent files or declare an opened file :)

newIDE/app/src/MainFrame/index.js
```diff
@@ -1408,9 +1408,20 @@ class MainFrame extends React.Component<Props, State> {
...
+            const parsedData = JSON.parse(recentFiles);
+            parsedData.push(fileMetadata.fileIdentifier);
+            localStorage.setItem('gd-recent-files', JSON.stringify(parsedData));
+          } else {
+            localStorage.setItem(
+              'gd-recent-files',
+              JSON.stringify([fileMetadata.fileIdentifier])
+            );
+          }
+        });
```
4ian: As explained let's try to move this into PreferencesContext, in a "addRecentFile" or something similar.
Let's save `FileMetadata` entirely rather than just `FileMetadata.fileIdentifier`

newIDE/app/src/MainFrame/index.js
```diff
@@ -1408,9 +1408,20 @@ class MainFrame extends React.Component<Props, State> {
...
 
-        return this.openFromFileMetadata(fileMetadata).then(() =>
-          this.openSceneOrProjectManager()
-        );
+        return this.openFromFileMetadata(fileMetadata).then(() => {
+          this.openSceneOrProjectManager();
+          const recentFiles = localStorage.getItem('gd-recent-files');
+          if (recentFiles) {
+            const parsedData = JSON.parse(recentFiles);
+            parsedData.push(fileMetadata.fileIdentifier);
```
4ian: We should save the entire fileMetadata. This is because some storage providers like Google Drive or other in the futures might want to save something else than just a `fileIdentifier`.
So we need everything :) 

Right now, it's true that recent files will be only for the desktop app, but in the future we might show the menu in the web-app, in which case we want to remember all kind of file that was opened.

newIDE/app/src/MainFrame/Preferences/PreferencesProvider.js
```diff
@@ -317,6 +320,33 @@ export default class PreferencesProvider extends React.Component<Props, State> {
...
+        data,
+        ...values.recentFiles,
+      ]
+    }
+  },
+  () => this._persistValuesToLocalStorage(this.state)
+  );
+    }
+  }
+
```
HarsimranVirk: I've added these two functions which are basically setters and getters. `recentFiles` is an array of `fileMetadata` now, so we'll have the entire object rather than `fileIndentifier` as you suggested.

newIDE/app/src/MainFrame/index.js
```diff
@@ -2076,5 +2070,5 @@ class MainFrame extends React.Component<Props, State> {
     );
   }
 }
-
+MainFrame.contextType = PreferencesContext;
```
HarsimranVirk: This will fail the flow test :( I don't know how to use flow with `contextType`, I've checked their docs and didn't find anything. Though we don't really need to use `contextType` here, we could've stored stuff into `recentFiles` during the rendering of `SceneEditor`, but there was a warning by react which basically said don't update anything while rendering.

newIDE/app/src/MainFrame/ElectronMainMenu.js
```diff
@@ -400,5 +394,6 @@ class ElectronMainMenu extends React.Component<Props, {||}> {
     });
   }
 }
+ElectronMainMenu.contextType = PreferencesContext;
```
HarsimranVirk: This will also fail flow tests, and I don't know how to access the `PreferencesContext` here, other than this (I couldn't really figure out how to use the `Consumer` here, so I went with `ContextType`). Is there a better way?

newIDE/app/src/MainFrame/ElectronMainMenu.js
```diff
@@ -145,6 +156,17 @@ class ElectronMainMenu extends React.Component<Props, {||}> {
     }
   }
 
+  componentWillMount() {
```
4ian: componentWillMount would only generate the menu when the component is first shown, i.e: the app startup. It would not update if we open a new file. 
To fix this, we'll rely on the fact that at any time to change the recent file, then the PreferencesProvider is doing a setState. 
Which then re-renders any consumer of the context... including the one that gives the props to MainFrame and ElectronMainMenu :) 
Then in ElectronMainMenu, we still don't want to always re-render the menu (in fact, most of the case we don't want to update it) - as it might be a bit costly. But here, we want to trigger an update to the menu.
How?
The component will be re-render when the props are updated. `render` will be called (we do nothing special) and then `componentDidUpdate`, if any, will be called.
At this moment, we can do something in the `componentDidUpdate`: we compare the result of `this.props.getRecentFiles` with `this._recentFiles` that we memorized during the last calls to `_buildAndSendMenuTemplate`. If it's different, we call `_buildAndSendMenuTemplate`.

The menu will be updated (and this._recentFiles updated). Done! :)

newIDE/app/src/MainFrame/index.js
```diff
@@ -231,6 +231,12 @@ class MainFrame extends React.Component<Props, State> {
     console.info('Startup times:', getStartupTimesSummary());
   }
 
+  componentDidUpdate() {
```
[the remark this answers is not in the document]
4ian: You solved this already :)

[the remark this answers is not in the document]
HarsimranVirk: I should've really waited a bit more I guess :D I pushed the code right before you started reviewing.

newIDE/app/src/MainFrame/ElectronMainMenu.js
```diff
@@ -161,6 +183,10 @@ class ElectronMainMenu extends React.Component<Props, {||}> {
           accelerator: 'CommandOrControl+O',
           onClickSendEvent: 'main-menu-open',
         },
+        {
+          label: i18n._(t`Open Recent...`),
+          submenu: this.recentFiles,
```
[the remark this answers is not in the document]
HarsimranVirk: I made a new function which basically does the `getRecentFiles().map` and called it here. It is exactly the same thing except it's a function call, unfortunately, menu does not get the right values :( I'll see something which works though :)

HarsimranVirk:
> We can try adding a <PreferencesContext.Consumer> in LocalApp.js, and then pass a prop to ElectronMainMenu and MainFrame?

I'll try that, though flow is the only issue with `contextType` :( Let's see if passing props could simplify things even more or not :)

4ian:
If you don't like using `PreferencesContext.Consumer` in LocalApp (which is a bit awkward to write I must admit), the other solution without contextType would be too convert MainFrame and ElectronMainMenu to function components and then use the `useContext` hook. Definitely cleaner... but also a lot of work! :/
`contextType` might work but not sure about Flow and not a super fan of having only one context. `contextType` seems to be an easy way to migrate from the old context api.

HarsimranVirk:
> MainFrame and ElectronMainMenu to function components

Yes, definitely a lot of work. It can be done however I'd say. (Once I'm done with my GSoC proposal I'd be able to work on this, especially since there's a lockdown due to COVID-19 right now, so I'd have a lot of free time :p ) MainFrame is one of the most important component though, it would require extensive testing and reviewing afterwards.

4ian:
> MainFrame is one of the most important component though, it would require extensive testing and reviewing afterwards.

Yes, so let's do it in a separate PR! :) Either:
- you want to do it, let's do it in a PR before this one.
- you don't want to do it, let's use a `PreferencesContext.Consumer` around MainFrame.

MainFrame is significantly larger than ElectronMainFrame and is "mission critical" so better do its conversion in a separate PR.

4ian:
> Once I'm done with my GSoC proposal I'd be able to work on this, especially since there's a lockdown due to COVID-19 right now

Sure, prioritize the proposal :) Also stay safe at home!

HarsimranVirk:
> Yes, so let's do it in a separate PR! :)

Yes, it definitely needs another PR, I didn't realise how much work it really would be, I've been facing constant significant crashes since last 2 hours :p

> you want to do it, let's do it in a PR before this one.
you don't want to do it, let's use a PreferencesContext.Consumer around MainFrame.

I'd say if we want to convert it into a functional component just for the sake of `context`, it wouldn't be worth it, to be honest. `MainFrame` is not really meant to be a functional component, it has a lot of public functions which would need to be exposed by `useImperativeHandle`, which would require a `type MainFrameInterface` to again avoid flow errors :( It'd be much easier to use the consumer in this case.

However, if we do want to convert it into a functional component so that we can stay up to date with React and hooks, then it would be nice :)

I've made a new branch to work on it, I'll open a PR when I'd be able to fix the massive crashes :D

4ian:
>  MainFrame is not really meant to be a functional component, it has a lot of public functions which would need to be exposed by useImperativeHandle, which would require a type MainFrameInterface to again avoid flow errors :( It'd be much easier to use the consumer in this case.

Yep feel free to go for the simplest approach in this PR.
I think MainFrame would still be a good functional component, even if it's true that it would have a large useImperativeHandle... though we could reduce this by moving ElectronMainMenu **inside** MainFrame.

The way I did it (ElectronMainMenu containing MainFrame) is a bit counter intuitive in React. We can see that because ElectronMainMenu uses a ref to the editor. Instead, MainFrame should **contain** ElectronMainMenu and pass props to it.
That's a larger refactoring... but that would be useful in the future to make ElectronMainMenu able to be much more reactive. It could for example have the current project passed as prop. If it's null, we could disable the menu items, etc!

HarsimranVirk:
Hey @4ian, I know we need to make a new PR, but I want to make MainFrame work before that, and I'm stuck with this:
https://github.com/4ian/GDevelop/blob/c7d5ab8013c9a5af350f5591502b7c02a239eb06/newIDE/app/src/MainFrame/index.js#L488-L498

I couldn't figure out how to use callback with useState hook, I tried making a custom hook,

```
function useStateWithCallback(initialValue) {
  const [state, setState] = React.useState(initialValue);
  const [callback, setCallback] = React.useState(null);
  
  const setStateWithCB = (newValue, cb) => {
    setState(newValue);
    setCallback(cb);
  }
  React.useEffect(() => {
  	if(callback){
    	callback();
        setCallback(null);
    } 
  }, [state]);
  return [state, setStateWithCB];
}
```
This does call the function passed to it, but it isn't exactly a `callback`. The function uses the previous value of state.

4ian:
There is indeed heavy usage of setState with a callback to do things after the changes are rendered in the DOM by React. 

I would recommend taking a look at this custom hook:
https://github.com/the-road-to-learn-react/use-state-with-callback/blob/master/src/index.js

More generally, what used to be in a callback can now maybe be in a useEffect, or something else. There is no general mapping of something that was a setState callback - for example the `updateToolbar` is a king of React hack before portals used to exist. Ideally, they would be in a useEffect, but maybe the custom hook "useStateWithCallback" can work in most cases 🤞

Bouh:
You are blocked on this feature ?

HarsimranVirk:
@Bouh We have to refactor `MainFrame` into functional component before adding this, because it's going to be easier to work with context afterwards. 

As far as the feature itself, I was able to implement it in this PR, but it required `ContextType` since `MainFrame` is a class component. 

Let's wait for a bit more, once #1684 is merged, I'll quickly finish this PR :)

HarsimranVirk:
@4ian closing this PR and opening a new one, I wouldn't be able to handle the merge conflicts now that MainFrame and ElectronMainMenu are changed a lot :joy:

Issue #1469, opened 2020-02-25
Upgrade exported games to latest Electron version


4ian:
We should upgrade to the latest Electron version. This might solve some crash issues/incompatibilities and is a good idea to be on the latest versions.

Thing to do:
- [ ] Check the release notes between the current Electron version and the new one
- [ ] Verify if there is anything notable or API that must be updated.
- [ ] Upgrade in the package.json
- [x] ~~Check if the whole app seems to work well (in particular: preview, test multiple games, debugger, export to Android and macOS/Linux/Windows).~~ Already done for the IDE.
- [ ] Check if we can also upgrade Electron for the games (`GDJS/Runtime`)
- [ ] Do (or ask someone) to do a test on Windows/macOS/Linux.

jerrybuks:
can I work on this?

4ian:
Yes feel free to go ahead and give it a try :)

jerrybuks:
okay, I'm on it

jerrybuks:
@4ian I did the update for both the Editor and the GDJS electron update, but I keep getting this error when I try to run a game 
  
![Screenshot from 2020-03-08 07-27-38](https://user-images.githubusercontent.com/39066907/76157808-7f377a00-610e-11ea-87a5-ac6ad253e1bc.png)

4ian:
This error is in the web-app, so totally unrelated to Electron - ignore it. This being said, have you run the app with Electron ? 😉

jerrybuks:
@4ian the electron-app runs and works fine, but it also throws this error, when I click the run button, for a game
![Screenshot from 2020-03-08 19-34-06](https://user-images.githubusercontent.com/39066907/76168915-04a14580-6174-11ea-83f2-19e863cdc89a.png)

4ian:
Very strange. Is this working when you don't update electron?
BrowserS3GDJSFinder should NOT be used in the electron version. Can you investigate why it's used? It should be used only for the web-app.

Unfortunately I won't be able to help very much without knowing more. You have to investigate a bit more by yourself to understand why this is happening :) Check if the current version works. If yes, understand why BrowserS3GDJSFinder is even used => in electron, it should NOT be used.

jerrybuks:
> Very strange. Is this working when you don't update electron?
> BrowserS3GDJSFinder should NOT be used in the electron version. Can you investigate why it's used? It should be used only for the web-app.
> 
[...]

Issue #1550, opened 2020-03-18
Disable menu items when action is not possible


pranavmodx:
## Describe the bug
Not exactly a bug but a possible enhancement/prevention of future bugs. The corresponding menu item should be disabled when an action is not possible in a given setting. Else, it could become a reason for a bug at some other time when user clicks the menu item when it isn't allowed/possible to.

## To Reproduce
1. Open desktop app
2. Click on Edit Menu or any other menu
3. Observe all menu items are by default enabled

## Screenshots
![Screenshot 2020-03-18 at 9 59 47 PM](https://user-images.githubusercontent.com/42793024/76984730-4052c100-6965-11ea-9ba8-8acd715f4ee7.png)
Eg: Here, actions like cut, copy, paste should not possible (as this is the start page).

![Screenshot 2020-03-18 at 10 00 29 PM](https://user-images.githubusercontent.com/42793024/76984860-6aa47e80-6965-11ea-95d0-b36e1092a009.png)
Example of an app (MS Todo) which disables menu items accordingly.

## Other details
* Mac OSX Mojave
* Desktop App

4ian:
For this kind of issues to be useful, it would be a good idea to list what to do, so that the issue is actionable - otherwise this issue will surely stay there forever.
Opening an issue is good, but making the work of finding what is really missing in details is even better :)

pranavmodx:
> For this kind of issues to be useful, it would be a good idea to list what to do, so that the issue is actionable - otherwise this issue will surely stay there forever.

Thanks for mentioning this. Will keep this in mind in the future. :)

Is there an easy workaround for this? Or does this require a lot of changes with conditionals for different scenarios?

4ian:
Probably the work is quite different for each menu item, might not be totally obvious how to get the info to know if we can disable the item or not. Look at ElectronMainMenu.js to get an idea :)

Minisonny:
Is it only me or anyone else also observe that the _Edit_ dropdown doesn't work even in the main editor 😕

4ian:
It's only for text fields.

4ian:
But to be fair yeah it's confusing because it's always enabled even when not useful!

pranavmodx:
The File Menu and the Edit Menu are the most confusing ones I think. So maybe fixing just those at this time would help?

I'm not sure regarding the solution but here is what I worked out :

- When to disable to each menu?
[...]

PR #1589, opened 2020-03-26
Upgrade electron to v8.2.1

NilayMajorwar:
This is a ~~wip~~ PR for upgrading electron to v8.2.0 (issue #1469).

Following are the relevant changes in Electron I noticed:
 - #### API change for `protocol.registerStandardSchemes`
   The function `protocol.registerStandardSchemes` has been renamed to `protocol.registerSchemesAsPrivileged`, with slight change in the argument format. Have accordingly changed the call while initializing the app. 

 - #### `BrowserWindow` options
   The default values of some options in `new BrowserWindow(options)` have changed ([see here](https://www.electronjs.org/docs/breaking-changes#planned-breaking-api-changes-50)). I have accordingly changed the options we use to initialize the app.

 - #### Dialog API
   The API for file open (`dialog.showOpenDialog`) and file save (`dialog.showSaveDialog`) pickers has changed from callback-based to promise-based, and a sync version of each has been introduced (see [v7 changelog](https://www.electronjs.org/releases/stable?page=6) and [v6 changelog](https://www.electronjs.org/releases/stable?page=10#6.0.0)). I have updated all the dialog calls in the codebase to the sync version, and it seems to work well - except for one Electron issue on Linux (not sure about other OSs):

   If you run the Electron app from the VSCode integrated terminal, open an object, try to add a new sprite for any animation, and in the file picker click on a folder from left pane, the app crashes with a segfault ([here's the Electron issue](https://github.com/electron/electron/issues/20380)). This happens for the the manual Windows/macOS/Linux export location picker too, and probably for some more pickers. Everything works fine when the app is run from a normal terminal window, though.

 - #### `electron.asar` no longer exists
   I don't really understand this one, so I'll just mention this. v7 also removes `electron.asar` (see v7  changelog from above). I saw this file in the GDevelop production tarball downloaded from the website for Linux platform, so I wondered if this may cause a problem. 

I haven't tested the app completely though - just checked a few game previews, debugger and exported a game to Linux. Seems to work till now. I will continue to test the app to check for any other issues.

#### Task list:

 - [x] Check the release notes between the current Electron version and the new one
 - [x] Verify if there is anything notable or API that must be updated.
 - [x] Upgrade in the package.json
 - [x] Check if the whole app seems to work well (in particular: preview, test multiple games, debugger, export to Android and macOS/Linux/Windows).
 - [ ] Check if we can also upgrade Electron for the games (GDJS/Runtime)
 - [ ] Do (or ask someone) to do a test on Windows/macOS/Linux.

### DIFF
--- a/newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionOpener.js
+++ b/newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionOpener.js
@@ -27,25 +27,20 @@ export default class LocalEventsFunctionsExtensionOpener {
       if (!dialog) return reject('Not supported');
 
       const browserWindow = electron.remote.getCurrentWindow();
-      dialog.showOpenDialog(
-        browserWindow,
-        {
-          title: 'Import an extension in the project',
-          properties: ['openFile'],
-          message: 'Choose an extension file to import (.json file)',
-          filters: [
-            {
-              name: 'GDevelop 5 "events based" extension',
-              extensions: ['json'],
-            },
-          ],
-        },
-        paths => {
-          if (!paths || !paths.length) return resolve(null);
+      const paths = dialog.showOpenDialogSync(browserWindow, {
+        title: 'Import an extension in the project',
+        properties: ['openFile'],
+        message: 'Choose an extension file to import (.json file)',
+        filters: [
+          {
+            name: 'GDevelop 5 "events based" extension',
+            extensions: ['json'],
+          },
+        ],
+      });
 
-          return resolve(paths[0]);
-        }
-      );
+      if (!paths || !paths.length) return resolve(null);
+      return resolve(paths[0]);
     });
   };
 
--- a/newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionOpener.js
+++ b/newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionOpener.js
@@ -41,9 +41,6 @@ export default class LocalEventsFunctionsExtensionOpener {
       .then(({ filePaths }) => {
         if (!filePaths || !filePaths.length) return null;
         return filePaths[0];
-      })
-      .catch(err => {
-        console.error('An error occured while importing the extension.', err);
       });
   };
 
--- a/newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionOpener.js
+++ b/newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionOpener.js
@@ -23,11 +23,11 @@ const readJSONFile = (filepath: string): Promise<Object> => {
 
 export default class LocalEventsFunctionsExtensionOpener {
   static chooseEventsFunctionExtensionFile = (): Promise<?string> => {
-    return new Promise((resolve, reject) => {
-      if (!dialog) return reject('Not supported');
+    if (!dialog) return Promise.reject('Not supported');
+    const browserWindow = electron.remote.getCurrentWindow();
 
-      const browserWindow = electron.remote.getCurrentWindow();
-      const paths = dialog.showOpenDialogSync(browserWindow, {
+    return dialog
+      .showOpenDialog(browserWindow, {
         title: 'Import an extension in the project',
         properties: ['openFile'],
         message: 'Choose an extension file to import (.json file)',
@@ -37,11 +37,14 @@ export default class LocalEventsFunctionsExtensionOpener {
             extensions: ['json'],
           },
         ],
+      })
+      .then(({ filePaths }) => {
+        if (!filePaths || !filePaths.length) return null;
+        return filePaths[0];
+      })
+      .catch(err => {
+        console.error('An error occured while importing the extension.', err);
       });
-
-      if (!paths || !paths.length) return resolve(null);
-      return resolve(paths[0]);
-    });
   };
 
   static readEventsFunctionExtensionFile = (
--- a/newIDE/electron-app/app/main.js
+++ b/newIDE/electron-app/app/main.js
@@ -47,6 +47,7 @@ const args = parseArgs(process.argv.slice(isDev ? 2 : 1), {
 // See registerGdideProtocol (used for HTML modules support)
 protocol.registerSchemesAsPrivileged([{ scheme: 'gdide' }]);
 
+// Set to true, will be default value in future Electron versions
 app.allowRendererProcessReuse = true;
 
 // Quit when all windows are closed.
--- a/newIDE/electron-app/app/main.js
+++ b/newIDE/electron-app/app/main.js
@@ -73,7 +73,6 @@ app.on('ready', function() {
     webPreferences: {
       webSecurity: false, // Allow to access to local files,
       nodeIntegration: true,
-      webviewTag: true,
     },
     enableLargerThanScreen: true,
     backgroundColor: '#f0f0f0',
--- a/newIDE/electron-app/app/main.js
+++ b/newIDE/electron-app/app/main.js
@@ -47,6 +47,8 @@ const args = parseArgs(process.argv.slice(isDev ? 2 : 1), {
 // See registerGdideProtocol (used for HTML modules support)
 protocol.registerSchemesAsPrivileged([{ scheme: 'gdide' }]);
 
+app.allowRendererProcessReuse = true;
+
 // Quit when all windows are closed.
 app.on('window-all-closed', function() {
   app.quit();
--- a/newIDE/electron-app/app/main.js
+++ b/newIDE/electron-app/app/main.js
@@ -45,7 +45,7 @@ const args = parseArgs(process.argv.slice(isDev ? 2 : 1), {
 });
 
 // See registerGdideProtocol (used for HTML modules support)
-protocol.registerStandardSchemes(['gdide']);
+protocol.registerSchemesAsPrivileged([{ scheme: 'gdide' }]);
 
 // Quit when all windows are closed.
 app.on('window-all-closed', function() {
@@ -71,7 +71,9 @@ app.on('ready', function() {
     x: args.x,
     y: args.y,
     webPreferences: {
-      webSecurity: false, // Allow to access to local files
+      webSecurity: false, // Allow to access to local files,
+      nodeIntegration: true,
+      webviewTag: true,
     },
     enableLargerThanScreen: true,
     backgroundColor: '#f0f0f0',
--- a/newIDE/app/public/external/utils/path-editor.js
+++ b/newIDE/app/public/external/utils/path-editor.js
@@ -219,11 +219,11 @@ const selectBaseFolderPath = headerObject => {
   if (!state.projectBasePath) {
     state.projectBasePath = state.folderPath;
   }
-  const selectedDir = dialog.showOpenDialog(remote.getCurrentWindow(), {
+  const selectedDir = dialog.showOpenDialogSync(remote.getCurrentWindow(), {
     properties: ['openDirectory'],
     defaultPath: state.projectBasePath,
   });
-  if (!selectedDir) {
+  if (!selectedDir || !selectedDir.length) {
     return;
   }
   const selectedDirPath = selectedDir[0];
--- a/newIDE/app/src/Debugger/DebuggerContent.js
+++ b/newIDE/app/src/Debugger/DebuggerContent.js
@@ -87,7 +87,7 @@ export default class DebuggerContent extends React.Component<Props, State> {
             toolbarControls: [],
             renderEditor: () => (
               <Background>
-                <Column expand noMargin>
+                <Column expand noMargin useMaxHeight>
                   <Line justifyContent="center">
                     <RaisedButton
                       label={<Trans>Refresh</Trans>}
@@ -95,21 +95,19 @@ export default class DebuggerContent extends React.Component<Props, State> {
                       primary
                     />
                   </Line>
-                  <Line expand noMargin>
-                    <InspectorsList
-                      gameData={gameData}
-                      getInspectorDescriptions={getInspectorDescriptions}
-                      onChooseInspector={(
+                  <InspectorsList
+                    gameData={gameData}
+                    getInspectorDescriptions={getInspectorDescriptions}
+                    onChooseInspector={(
+                      selectedInspector,
+                      selectedInspectorFullPath
+                    ) =>
+                      this.setState({
                         selectedInspector,
-                        selectedInspectorFullPath
-                      ) =>
-                        this.setState({
-                          selectedInspector,
-                          selectedInspectorFullPath,
-                        })
-                      }
-                    />
-                  </Line>
+                        selectedInspectorFullPath,
+                      })
+                    }
+                  />
                 </Column>
               </Background>
             ),
--- a/newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionWriter.js
+++ b/newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionWriter.js
@@ -47,9 +47,6 @@ export default class LocalEventsFunctionsExtensionWriter {
       .then(({ filePath }) => {
         if (!filePath) return null;
         return filePath;
-      })
-      .catch(err => {
-        console.error('An error occured while exporting the extension.', err);
       });
   };
 
--- a/newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionWriter.js
+++ b/newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionWriter.js
@@ -30,11 +30,11 @@ const writeJSONFile = (object: Object, filepath: string): Promise<void> => {
 
 export default class LocalEventsFunctionsExtensionWriter {
   static chooseEventsFunctionExtensionFile = (): Promise<?string> => {
-    return new Promise((resolve, reject) => {
-      if (!dialog) return reject('Not supported');
+    if (!dialog) return Promise.reject('Not supported');
+    const browserWindow = electron.remote.getCurrentWindow();
 
-      const browserWindow = electron.remote.getCurrentWindow();
-      const path = dialog.showSaveDialogSync(browserWindow, {
+    return dialog
+      .showSaveDialog(browserWindow, {
         title: 'Export an extension of the project',
         filters: [
           {
@@ -43,11 +43,14 @@ export default class LocalEventsFunctionsExtensionWriter {
           },
         ],
         defaultPath: 'Extension.json',
+      })
+      .then(({ filePath }) => {
+        if (!filePath) return null;
+        return filePath;
+      })
+      .catch(err => {
+        console.error('An error occured while exporting the extension.', err);
       });
-
-      if (!path) return resolve(null);
-      return resolve(path);
-    });
   };
 
   static writeEventsFunctionsExtension = (
--- a/newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionWriter.js
+++ b/newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionWriter.js
@@ -34,24 +34,19 @@ export default class LocalEventsFunctionsExtensionWriter {
       if (!dialog) return reject('Not supported');
 
       const browserWindow = electron.remote.getCurrentWindow();
-      dialog.showSaveDialog(
-        browserWindow,
-        {
-          title: 'Export an extension of the project',
-          filters: [
-            {
-              name: 'GDevelop 5 "events based" extension',
-              extensions: ['json'],
-            },
-          ],
-          defaultPath: 'Extension.json',
-        },
-        path => {
-          if (!path) return resolve(null);
+      const path = dialog.showSaveDialogSync(browserWindow, {
+        title: 'Export an extension of the project',
+        filters: [
+          {
+            name: 'GDevelop 5 "events based" extension',
+            extensions: ['json'],
+          },
+        ],
+        defaultPath: 'Extension.json',
+      });
 
-          return resolve(path);
-        }
-      );
+      if (!path) return resolve(null);
+      return resolve(path);
     });
   };
 
--- a/newIDE/app/src/ProjectsStorage/LocalFileStorageProvider/LocalProjectOpener.js
+++ b/newIDE/app/src/ProjectsStorage/LocalFileStorageProvider/LocalProjectOpener.js
@@ -39,9 +39,6 @@ export const onOpenWithPicker = (): Promise<?FileMetadata> => {
     .then(({ filePaths }) => {
       if (!filePaths || !filePaths.length) return null;
       return { fileIdentifier: filePaths[0] };
-    })
-    .catch(err => {
-      console.error('An error occured while opening the project.', err);
     });
 };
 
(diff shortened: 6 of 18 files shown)


4ian:
Thanks for trying this and for the nice summary! :) 

> electron.asar no longer exists

I don't think that's an issue.

> Dialog API

What you did seems good.

> BrowserWindow options

Sounds good.

It would be interesting to try to build the app as a standalone executable.
For this: `cd newIDE/electron-app` then `npm run build -- --mac zip --win zip --linux tar.gz` (mac might not work on Linux).

We should get some people to test it, to be sure there is no major bugs on the all 3 OSes :)

newIDE/electron-app/app/main.js
```diff
@@ -71,10 +71,12 @@ app.on('ready', function() {
     x: args.x,
     y: args.y,
     webPreferences: {
-      webSecurity: false, // Allow to access to local files
+      webSecurity: false, // Allow to access to local files,
+      nodeIntegration: true,
+      webviewTag: true
```
4ian: I don't think we are (or want in the future) to use the `<webview>` tag, so you can omit this one :)

NilayMajorwar: Oh okay, done.

NilayMajorwar:
> try to build the app as a standalone executable.

Tried, works on Linux. Tested preview games, debugger, export to Android and Linux.

NilayMajorwar:
Another thing:

Following are the new warnings I get on starting the app after Electron upgrade:
```
(electron) 'getName function' is deprecated and will be removed. Please use 'name property' instead.
(node:8416) ProtocolDeprecateCallback: The callback argument of protocol module APIs is no longer needed.
(node:8416) ProtocolDeprecateCallback: The callback argument of protocol module APIs is no longer needed.
```
The first warning is because `app.getName()` is deprecated in favour of `app.name`. But I couldn't find any use of `app.getName()` in the JS part of the codebase. Any hint on where else I should look?

The second warning comes from [here](https://github.com/4ian/GDevelop/blob/c7d5ab8013c9a5af350f5591502b7c02a239eb06/newIDE/electron-app/app/Utils/UrlLoader.js#L29-L64). The new API for `protocol.register[any]Protocol` has shifted to synchronous version and no longer requires the third `completion: err => {...}` argument, but it doesn't have error handling yet. Should I remove the `completion` argument or let it be?

4ian:
> The first warning is because app.getName() is deprecated in favour of app.name. But I couldn't find any use of app.getName() in the JS part of the codebase. Any hint on where else I should look?

I'm not sure. Can't find it.

> Should I remove the completion argument or let it be?

The [API doc](https://www.electronjs.org/docs/api/protocol#protocolregisterhttpprotocolscheme-handler-completion) is not very clear.
Can you check, when you remove the `completion` callback if:
- there is a return value that you can use to display an error in case the call was not successful.
- otherwise, see if you can use `protocol.isProtocolRegistered` to check if the scheme is registered. If not, show an error message.

Speaking of this, these "protocols" are here to make Piskel/Jfxr work. Can you try to run Piskel in:
- the development version
- the standalone version (the one after running npm run build)

(you need to test both, because as you can see the protocal registration is different).

Thanks for spotting this :)

NilayMajorwar:
> Can you check, when you remove the completion callback if:
> - there is a return value that you can use to display an error in case the call was not successful.
> - otherwise, see if you can use protocol.isProtocolRegistered to check if the scheme is registered. If not, show an error message.

How do I cause the call to fail (I tried passing an invalid URL)? I tried the first option yesterday (returns `undefined`), and just tried the second option (always returns true).

4ian:
> How do I cause the call to fail (I tried passing an invalid URL)? I tried the first option yesterday (returns undefined), and just tried the second option (always returns true).

I'm not sure how to make it fail :) But as the call returns `undefined` and that the second option is returning true, let's just log a message if it returns false. Should be enough (I never see it fail).

NilayMajorwar:
I've fixed the protocol API warning, and another warning (`The default value of app.allowRendererProcessReuse is deprecated, it is currently "false".  It will change to be "true" in Electron 9.`) by adding `app.allowRendererProcessReuse = true` (this option is going to be fixed unchangably to `true` in future Electron versions).

> try to run Piskel

Yeah, that gives an error (`require` not defined). I'm looking into it.

newIDE/electron-app/app/main.js
```diff
@@ -41,11 +41,13 @@ let mainWindow = null;
...
   boolean: ['dev-tools', 'disable-update-check'],
-  string: '_', // Files are always strings
+  string: '_' // Files are always strings
 });
 
 // See registerGdideProtocol (used for HTML modules support)
-protocol.registerStandardSchemes(['gdide']);
+protocol.registerSchemesAsPrivileged([{ scheme: 'gdide' }]);
+
+app.allowRendererProcessReuse = true;
```
4ian: Let's add a comment why we're doing this

NilayMajorwar: Oh yeah, forgot about that change :sweat_smile: . I'll add it to the PR description.

4ian: I would also like a comment in the code so that it's obvious to the reader why there is this mysterious line here :)

NilayMajorwar: Ohh, I thought you meant the first line (`protocol.registerSchemesAsPrivileged`).

I'm not sure what comment I should add for the second line. Should I just say it's for avoiding the warning?

4ian: Yes, let's go for something like "Set the option as its default value in future Electron versions."

NilayMajorwar:
I just noticed another strange problem - the scroll in the Inspectors List in the Debugger has stopped working. The `Column` corresponding to Inspector List has extended heightwise to cover whole app height and further outside the app window (lower part hides behind Profiler). I'll send some more screenshots tomorrow.

scroll

The Electron upgrade has caused this somehow since the scroll works fine on `master` branch. (I don't understand how an Electron upgrade caused a UI problem, though). I'll try to fix this tomorrow.

newIDE/electron-app/app/main.js
NilayMajorwar: Okay, I added "Set to true, will be default value in future Electron versions".

4ian:
> I don't understand how an Electron upgrade caused a UI problem, though

This is because the Chromium version was updated, and might have changes in the way flexbox is handled. Take a look at trying to use `maxHeight` prop on the Column creating the issue. I remember we fixed a scroll issue by adding this prop in another component recently. This was only happening in the web-app - probably because of the difference in Chrome versions.

NilayMajorwar:
(for reference:)
```
<Column expand noMargin> // for the Inspectors List content
    <Line> // for Refresh button
    <Line expand noMargin> // for the list
        <List> (internally, MUIList with overflow-y: scroll; flex: 1;)
```

The issue is that the `Column` and second `Line` expand all the way to the bottom, thus disabling the scrollbar.

I tried adding `maxHeight` in the `Column`. It fixes `Column` overflowing, but the `Line` still overflows to the bottom, perhaps due to the list being too long. A fix I found to this was to remove `overflow-y: scroll` from the `List` and add it from the `Line` (via a new prop). This does work fine and fixes the problem, but it adds a new prop and property to `Line`. Is it okay?

4ian:
>  This does work fine and fixes the problem, but it adds a new prop and property to Line. Is it okay?

Unfortunately no, we don't want to have Line being able to set a scroll, it's not its job and make the component more complex while we should strive for simplicity. See if you can instead to simplify the markup (can you try remove Line, or remove Column, or both). Try a combination of this with maxHeight for column. See if you can use the `ScrollView` component otherwise? (instead or with Column/Line). ScrollView is a component that adds a `overflow-y: scroll` :)

NilayMajorwar:
> can you try remove Line, or remove Column, or both

Okay, that worked right out of the box :smile: ! I removed the second Line, with `maxHeight` applied on `Column` and it works great. I'll push this commit right now.

I'll move on to upgrading the runtime Electron now. [This (`GDJS/Runtime/Electron/package.json`)](https://github.com/4ian/GDevelop/blob/master/GDJS/Runtime/Electron/package.json) is the file I need to change Electron version in, right?

4ian:
Great! 

> the file I need to change Electron version in, right? And how do I test if it's working?

Maybe we could do the update in another PR? But for testing, install and package with electron-builder an exported game.

NilayMajorwar:
Okay, I'll do that in another PR. So now we can put up this PR for testing by other people, I guess.

4ian:
Alright! 
Still two things on my mind:
- You updated the dialogs functions to the sync version. But could not we use instead the promise-based version? It's almost exactly the same as the callback version *and* better because we avoid a blocking call. Sorry I just realised this now. But as we already did the work to avoid everything working with callbacks, it's better to keep promises because it's not blocking, and would allow to use async/await maybe in the future. And also, most of what the callbacks are doing is to resolve or reject a promise... we can simplify this now that these functions are promised based :)
- Your changes are removing all the trailing commas at the end of the last property of an object. It's a decision to keep them in the whole codebase because they make editing easier and are helping getting the line not marked as modified in git when a new property is added. I think it's a matter of fixing your Prettier configuration. We maybe are missing a .prettierrc file somewhere.

Thanks for your work so far :)

NilayMajorwar:
Okay, I'm on it.

NilayMajorwar:
What about the dialog calls which were synchronous before this change (where callback argument was not provided and return value was used)? Should I change them to promises too?

4ian:
You can keep those sync - it's fair that the PR is just converting the calls of the callback based ones :)

PawBud:
What help do you need in testing @NilayMajorwar ?

NilayMajorwar:
Sure, thanks a lot. Yeah, just test if there's anything that should be working but isn't. Just wait for a few minutes, I'm pushing another little commit.

PawBud:
Sure, ping me whenever you are ready.

NilayMajorwar:
Actually, never mind. I just found a serious issue that might take quite some time to solve :sweat_smile:, so I guess testing will have to wait a bit more. Sorry!

PawBud:
Ohh never mind, I was also working on an issue, so you gave me no trouble whatsoever :)

NilayMajorwar:
> If you run the Electron app from the VSCode integrated terminal, open an object, try to add a new sprite for any animation, and in the file picker click on a folder from left pane, the app crashes with a segfault (here's the Electron issue). This happens for the the manual Windows/macOS/Linux export location picker too, and probably for some more pickers. Everything works fine when the app is run from a normal terminal window, though.

@4ian This happens much more often actually - the app crashes even if I just go into the default `images` folder of any project while adding a sprite. It's pretty random. To be honest it's almost unusable from the VSCode integrated terminal.

@PawBud There's a known issue as mentioned above. I guess you can test the app for any other issues. What OS are you on, btw?

PawBud:
I do have both windows and Linux, I am sceptical about macOS

Bouh:
> * Your changes are removing all the trailing commas at the end of the last property of an object. It's a decision to keep them in the whole codebase because they make editing easier and are helping getting the line not marked as modified in git when a new property is added. I think it's a matter of fixing your Prettier configuration. We maybe are missing a .prettierrc file somewhere.

Add `GDevelop\newIDE\.prettierrc`

```
{
  "singleQuote": true,
  "trailingComma": "all"
}

In all others file trailingComma is set on es5. Should we change ?

```
[Documentation of prettier](https://prettier.io/docs/en/options.html#trailing-commas)

4ian:
https://github.com/4ian/GDevelop/blob/master/newIDE/app/.prettierrc trailing`comma is set on es5 no? 
That should be the case for all prettier configuration across the codebase (single quote, es5 trailing comma)

NilayMajorwar:
The `electron-app` folder is missing the config file. I'll add a `.prettierrc` there.

4ian:
Do you think you could do this in a separate PR? The risk is that the PR is becoming bug and cluttered by formatting changes, risking me to miss important changes, increasing the review time, and if something goes wrong with the new Electron version, I'll have to revert the PR and so loose the improvements on the formatting.
The good way to avoid this situation is to make another PR dedicated to the formatting :)
* Make another branch,
* Add .prettierrc
* Add a `format` script command in electron-app/package.json (like in newIDE/app package.json). Also add prettier in the package.json devDependencies (like in newIDE/app package.json). 
* Run `npm run format` to format all the files
* Open the PR

Then when the other PR is merged in master, all the formatting changes will disappear from this PR :)

NilayMajorwar:
Yeah okay, I'll do this in another PR then.

4ian:
To avoid conflicts, run Prettier in your branch with the same settings you added in the other PR :)

NilayMajorwar:
> To avoid conflicts, run Prettier in your branch with the same settings you added in the other PR 

Yeah, should have done that straightaway - I messed up a bit, I don't know why I pulled from `master` :sweat_smile:

newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionOpener.js
```diff
@@ -23,30 +23,28 @@ const readJSONFile = (filepath: string): Promise<Object> => {
...
+            name: 'GDevelop 5 "events based" extension',
+            extensions: ['json'],
+          },
+        ],
+      })
+      .then(({ filePaths }) => {
+        if (!filePaths || !filePaths.length) return null;
+        return filePaths[0];
+      })
+      .catch(err => {
```
4ian: Let's remove all the `catch`s, unless there is a case in which showOpenDialog can throw an exception - but I don't think so.
Exceptions are signaling a serious problem: having one there means that there was something that was not handled - in JavaScript this can even be a syntax error. We do *not* want to catch these. This is called "exception swallowing" and this hides the issue to the caller functions.
If an exception was launched, we're in very serious trouble, and we should let the exception propagate and crash the app (with a proper stacktrace in development that will let us find what was the issue). Static type checkers like Flow are helping us to ensure that, in theory, nothing bad will happen.

NilayMajorwar: Right.. Removing the catches.

4ian:
Thanks! Now we would need people to try on Windows, macOS, Linux - and if possible try the version built by doing `npm run build` in `newIDE/electron-app` (to test the "real" packaged, production version).

Not sure how the VSCode + Electron bug is important, seems a bit worrying but on the other hand that's a bit of a shame to keep the Electron update waiting on this "corner case" :/

PawBud:
I am ready to test it but what exactly are the things that need to be tested @4ian?

4ian:
Test as much of the software as possible (try to open, use single dialog) to ensure that there are no crashs, that the native capabilities of Electron (native dialogs, save/open, message box, local file reading) are working well. 
There is not a single thing to check, otherwise I would do it myself :) That's why I need you to look and give a more in depth check in the whole app.

PawBud:
I am on it, although it would take some time to review and report line-by-line.

PawBud:
So I cloned the repository and made a task list to check the working of the upgrade, Here is the checklist-

- [ ] make sure that the Start page is working in the electron-app. 

- I just realized while testing the front page how important is to resolve the issue  [https://github.com/4ian/GDevelop/issues/1550](url). It is really irritating to click at an option and watch nothing happen but we concluded that we need to update electron first so I won't nitpick much.

- [ ] Test the Scene Editor  

- [ ] Test the Events Editor page

Testing the whole app will require some time since I don't want to miss on anything, I will add some more tasks as I test and things that I need to point out will be written beneath them.

NilayMajorwar:
Thanks a lot for testing this! Take your time.
(the link's not working, btw :)

PawBud:
> Thanks a lot for testing this! Take your time.
> (the link's not working, btw :)

I don't know why it's not working I am sure I copied the URL that's why I have written the title of the PR

Bouh:
Tested on Windows 10.


- I see scrollbar in debugger ✅
- List of object in instruction editor are same in dark and light mode, i can see 11 objects list have 400px height. ✅
- Debugger / profiler  ✅
- Import / export / save / save as project ✅
- Open dev tool ✅
- Import resources ✅
- Switch theme / language ✅
- Piskel ✅

❌
Open a project, so some edit in scene.
Change click on button for change language. 
see warnings in console (maybe related to react not electron not sure.)
image

4ian:
> See warnings in console (maybe related to react not electron, not sure.)

That's because we use a divider in the SelectField but it's not supported. We should remove it. It's not Electron related.

> GLTexture.js:94 WebGL: INVALID_VALUE: tex(Sub)Image2D: video visible size is empty

Mmmm that's worrying, maybe we should open another issue if it's happening in beta89.

Bouh:
I've edited my previous post, with more info on this two bug.

4ian:
> Tested on Windows 10.

Thanks for doing it :) Can you try with the packaged version (npm run build in electron-app to create the executable)? Want to be sure that there is no issue with Piskel when the app is standalone.

4ian:
> GLTexture.js:94 WebGL: INVALID_VALUE: tex(Sub)Image2D: video visible size is empty
> This bug is only on your branch, this is not present in master, or current beta 89 public.

Alright, so that's a real issue. We should try to understand why this error is coming. It might be due to the changes in the Chromium version. Let's check in the latest Chrome version if the video is working: it seems to be black for me in Chrome 80.0.3987.149.
So it seems that the latest Chrome version are not able to run at least this video (maybe other videos too). Might be an encoding issue, or a PIXI issue (worth checking in the latest v4 versions if there was a bug fix)?

Bouh:
When you select the video on scene, and set custom size, then you can see the video in preview.
But still black in scene editor.

[This issue](https://github.com/pixijs/pixi.js/issues/5996) on pixi repo is very similar.
And last post said this is fix in pixi 5.1.5, if we use 5.2.1 we are good.

I've used my [Video sandbox pixi v4.8.6](https://jsfiddle.net/Bouh/ahu315ws/latest) for try, and if autoplay is on false, like it's the case in the present video extension, the warning is present, if we set true, on autoplay, the video is working well :)
`videoSprite._texture.baseTexture.resource.autoPlay = true;`

Note this error isn't on [Video sandbox pixi v5.2.1](https://jsfiddle.net/Bouh/ze0hkdqx/latest) (Think to change autoplay on false)

Bouh:
Here two portable builds of this PR, for [Linux and Windows 64bits](https://drive.google.com/open?id=1rhVpy61ZAqLgY2TweCL-jdtyO89vMTwH).
All works [like i've said here.](https://github.com/4ian/GDevelop/pull/1589#issuecomment-607260329)

NilayMajorwar:
I can only reproduce it partially on my end. The video is black in the scene editor, but it works fine in the preview (I do see the WebGL warning though). Both the fiddles work fine for me without any warning, regardless of `autoPlay` value. Maybe due to Linux - there's no Linux post on the PIXI issue, although that might be just by chance.

I'm not sure where I should apply the autoPlay fix - I'll try though.

> That's because we use a divider in the SelectField but it's not supported.

Can I submit a PR for this? Seems like an easy one.

4ian:
@Bouh I lost a bit the track of the video issue, do we have a workaround? 
Would be great to be able to upgrade to the latest Electron :) The issue is appearing in Chrome too, so maybe better to upgrade anyway and then rapidly upgrade to Pixi v5

Bouh:
In` static getPIXIVideoTexture()` in `C:\GDevelop\newIDE\app\src\ObjectsRendering\PixiResourcesLoader.js`

Just before return, add `loadedTextures[resourceName].baseTexture.source.preload = "auto";
I can see now the video texture in scene in this branch.
`

4ian:
@Bouh Cool! Should we do a PR to fix this already? This way this upgrade should work :)

Bouh:
All works in current beta. No need this fix actually in pixi 4.8.6 with the current version of electron.

If you add the fix in your master it's doesn't change anything.
So we can go for add the fix now.
Or 
@Quarkstar add it and convert the code for migrate on pixi v5. [BaseTexture change](https://github.com/pixijs/pixi.js/wiki/v5-Migration-Guide#basetexturesource)

```
loadedTextures[resourceName].baseTexture.source.preload = "auto";
````
become

```
loadedTextures[resourceName].baseTexture.resource.source.preload = "auto";
```

But this fix is also useless in v5 because this [parameter is always defined here](https://github.com/pixijs/pixi.js/blob/dev/packages/core/src/textures/resources/VideoResource.ts#L50).
As you want.
Add its now, and delete it after or just merge this PR in same time as future PR for pixi v5.

4ian:
> The question is you want out a new release after this upgrade of electron or after the upgrade of electron and pixi v5 ?

A new release after this upgrade of Electron.
Doing too much changes (Electron upgrade, Pixi upgrade) is too risky. I prefer to do first the Electron upgrade, then later the Pixi upgrade when it's ready. And be able to release a new version at any time - because you never know, there could be a urgent bug fix to do (in this case, you don't want to be stuck because something is not working).

> Add its now, and delete it after or just merge this PR in same time as future PR for pixi v5.

I prefer that we add the fix/hack now. So that we can merge this PR, have a new version with the new Electron and Pixi 4.8.6, allowing us to verify that there is no regression in the latest Electron. And then merge the upgrade to Pixi 5.

NilayMajorwar:
Sorry, couldn't work on this, was a little busy today :sweat_smile: 

I've rebased the PR to remove the merge conflict. I'll squash it into a summary commit soon. Video works great after @Bouh's fix :smile: .

Also, on a side note, I figured out where the `getName` deprecation warning was coming from - it was the `electron-log` package. I've upgraded it to the latest version. Fixes the warning, and the terminal logs are now a bit prettier too :D

4ian:
> I've rebased the PR to remove the merge conflict. I'll squash it into a summary commit soon.

Thanks for rebasing :) Don't worry about squashing commits, I always squash PR commits when merging them :) 

> Also, on a side note, I figured out where the getName deprecation warning was coming from - it was the electron-log package. I've upgraded it to the latest version. Fixes the warning, and the terminal logs are now a bit prettier too :D

Thanks for finding this! That's great.

If there is nothing blocking I'll merge this and we'll use this Electron version for the next release.

NilayMajorwar:
> If there is nothing blocking I'll merge this

I don't think there's any other issue.. has this PR been tested on MacOS? (I'm probably just worried about if we merge this and then later some really big issue pops up that I totally forgot about :sweat_smile: )

4ian:
I'll give it a try as soon as possible on macOS ;)

NilayMajorwar:
I can't seem to reproduce the issue. Opened the edit window, cancelled, added a sprite and the file open dialog opens up fine, and I'm able to add the sprite.

Nevertheless, I'll go through the related files and APIs to check for anything out of place. Is there any kind of error log to help in debugging this?

4ian:
>  Is there any kind of error log to help in debugging this?

Unfortunately the logs in the terminal seems not helpful :/
This might be a macOS only issue. On which OS did you test?

@Bouh (or anyone else), can you confirm that you *don't* have the issue I explained [here](https://github.com/4ian/GDevelop/pull/1589#issuecomment-610022024)?

Unfortunately this is a blocker until we find a way to solve this on macOS - I'll try to take a new look. Might even be an Electron bug - but that would be suprising?
Once this is solved, we should be good to go with this PR.

Bouh:
This issue isn't on windows. I can't test other OS.
All external editor (audio, yarn, piskel) works fine with cancel and apply.

NilayMajorwar:
I tested this on Linux (Ubuntu 19.10). So yeah, I guess it's a MacOS-only issue.

4ian:
This is a bug in Electron: https://github.com/electron/electron/issues/22833 - only in macOS.
We'll need to find a version where this is fixed :)

NilayMajorwar:
Oh, okay. So I should probably rollback the target Electron version to v8.0.3, then (hoping it does not bring up any other issue :sweat_smile: )

4ian:
> So I should probably rollback the target Electron version to v8.0.3, then (hoping it does not bring up any other issue 😅 )

That's a solution - though it's a bit sad because loose a potential crash fix (https://github.com/electron/electron/issues/17558) but well it's less important than just not working at all (though there is a risk of missing other bug fixes or having other issues as you mentionned 😬).. or we wait a little bit for an update containing the bug fix.

4ian:
Turns out that the fix is already available in 8.2.1 :) https://github.com/electron/electron/releases/tag/v8.2.1

NilayMajorwar:
Oh, right! I thought this was fixed in v9.0.0-beta13, that's why I went for rollback. And what luck - looks like v8.2.1 was just released two days :smile: ! I'll push a commit for this right away.

4ian:
I'll merge this as there is no good reason to hold this giving that version 8.2.1 works on macOS :)
There was still an issue of app crashing when packaged and signed on macOS - but should be solved by a mix of upgrading electron-builder and adding a new entitlement => I'll commit that after merging this :)

PR #1597, opened 2020-03-29
[wip] Upgrade Box2d.js to v2.3.1

PawBud:
This is a potential fix for the issue:_#1493_
I have added 4 files **instead of replacing them**. _Note that this is an experimental extension as I might entirely replace_ `Physics2Behavior`_directory with this._
 

-  the associated card on the roadmap: https://trello.com/c/4iXuaS7F/410-upgrade-to-box2djs-231

- The library file `box2d.was.js` may not be reviewed since it is the precoded library.

- I have added the `JsExtension.js` `physics3tools.js` and `physics3runtimebehavior`

- > If it's too different from the existing version, we might have to think of making an alternative, experimental extension ("Physics3") :)
>
Yes, This is the reason I made an experimental Extension.

_Some Questions I would like to answer that were mentioned in the issue_:-

-  **Is it ok to go with the WebAssembly version - or do we stick to asm.js?**
I think going with Web assembly version is better since Web Assembly is a far more advanced technology when compared to asm.js as mentioned in this _medium_ post:
[https://medium.com/samsung-internet-dev/performance-testing-web-assembly-vs-javascript-e07506fd5875](url)
Another _medium_ post which further explains the functionality of WebAssembly:-
[https://medium.com/@torch2424/webassembly-is-fast-a-real-world-benchmark-of-webassembly-vs-es6-d85a23f8e193](url)

- **As we're upgrading the Physics engine, games can behave slightly differently. Should we add an option to allow old games to stay on the old, legacy version?**
As fruitful as backward compatibility might look like in the first glance, I think that it should be restricted as this will restrict the users to actually try out the new version.
If they don't like it we can always add the option of backward compatibility :).

**Task List**
- [x] Added the Wasm file instead of asm.js to add the support for WebAssembly.
- [ ] Add the dependencies in the Cpp Files.
- [ ] Remove Dead Code from the files.
- [ ] Add the API changes to support box2d v2.3.1
- [ ]  Everything still works (try the example using physics, check that all methods used in the source 
      code by the Physics behavior on Box2d object is defined by the new version 

- [ ] Test on Android
- [ ] Test on iOS

- [ ] Test on Desktop Safari, Chrome, Firefox(For Windows, macOS, Linux)

I will be adding multiple commits that will be required and the Task List will be updated as the tasks are completed. :)


PawBud:
@4ian I would love to get some help as to how to approach this PR; I have some basic questions:- 

1. Should the library `box2d.wasm.js` be also included in the PR?
2. Should I divide the PR into segments; Like the changes in `physics3runtimebehaviour.js`, changes in `JsExtensions.js` be in 2 Different PR's or is this ok?

Thank you so much for taking the time to consider these questions :), If I have any more I will surely ask you.

Extensions/Physics3Behavior/JsExtension.js
```diff
@@ -0,0 +1,3721 @@
...
+      const extension = new gd.PlatformExtension();
+      extension.setExtensionInformation(
+        'Physics3',
+        'Physics Engine 3.0',
+        'Simulate physics',
+        'Florian Rival, Franco Maciel',
+        'MIT'
+      );
+  
+      var physics2Behavior = new gd.BehaviorJsImplementation();
```
Bouh: Update to 3, and see other comments.

Extensions/Physics3Behavior/JsExtension.js
```diff
@@ -0,0 +1,3721 @@
...
+          'Physics3Behavior',
+          _('Physics Engine 2.0'),
+          'Physics2',
+          _(
+            'Simulate realistic object physics, with gravity, forces, joints, etc.'
+          ),
+          '',
+          'res/physics32.png',
+          'Physics3Behavior',
+          physics2Behavior,
```
Bouh: physics3Behavior

Extensions/Physics3Behavior/JsExtension.js
```diff
@@ -0,0 +1,3721 @@
...
+        behaviorContent.addChild("scaleX").setDoubleValue(100);
+        behaviorContent.addChild("scaleY").setDoubleValue(100);
+      }
+  
+      var aut = extension
+        // extension
+        .addBehavior(
+          'Physics3Behavior',
+          _('Physics Engine 2.0'),
+          'Physics2',
```
Bouh: Go to 3 :)
This is display in the newIDE in behavior list and instructions :)

Extensions/Physics3Behavior/JsExtension.js
```diff
@@ -0,0 +1,3721 @@
...
+ *
+ * ℹ️ Run `node import-GDJS-Runtime.js` (in newIDE/app/scripts) if you make any change
+ * to this extension file or to any other *.js file that you reference inside.
+ *
+ * The file must be named "JsExtension.js", otherwise GDevelop won't load it.
+ * ⚠️ If you make a change and the extension is not loaded, open the developer console
+ * and search for any errors.
+ *
+ * More information on https://github.com/4ian/GDevelop/blob/master/newIDE/README-extensions.md
+ */
```
Bouh: Maybe a little text for know why we add this new physics v3.0.
For know what is the difference with v2.0.

Extensions/Physics3Behavior/JsExtension.js
PawBud: There are a lot of major fixes that I need to do, I am working on a small issue right now that's why I have not made any recent commits. Although thanks a lot for bringing this typo to my notice @Bouh .

Extensions/Physics3Behavior/JsExtension.js
PawBud: Yup, I am planning to include that as well, though  I need to decide which ones to put (that are significant and those that actually matter). Once again thanks a lot for reviewing a PR this big with so many typos.

PawBud:
@4ian @Bouh, I think that I have reached a dead end with this approach, (of adding a completely different `Physics3Behaviour` file & then re-editing `physics3runtimebehavior.js` file by adding new methods. I would like to add a new PR through a different branch regarding the same issue. I just wanted to ask if this is ok? 

I think by simply replacing the `box2d.js` with the new  `Box2d_v2.3.1_min.js` And by simply adding the new methods in the `physics2runtimebehavior.js` like 4ian suggested we may be able to upgrade
box2d. 

Although I will keep trying to make this approach work, If it doesn't then I will simply close this PR.

Btw, Thanks a lot @4ian for your advice, I re-did the react course properly, and now I feel much more confident with ReactJS now, I am happy that I took your advice :).

PawBud:
I had git issues (i rebased my branch by mistake ) hence I made a new PR https://github.com/4ian/GDevelop/pull/1636 regarding the same issue.

Issue #1591, opened 2020-03-26
Unable to launch compilation GDevelop.js


Bouh:
## Describe the bug
Unable to build GDevelop.js, since ninja integration and update of emscripten.
I've read again and again [the readme](https://github.com/4ian/GDevelop/tree/master/GDevelop.js) for install emscripten and start the build process.
And update my script for my workflow, but it's broken.

- I'am on current master
- I've deleted `C:\GDevelop\Binaries` 
- Update emscripten to 1.39.6, 
`C:\emsdk`
`git pull` because `emsdk update` is outdated (message from console)
`emsdk install 1.39.6`
`emsdk activate 1.39.6`
`emsdk_env.bat`

- Start build
`cd C:\GDevelop\GDevelop.js`
`npm install`
`npm run build`

My script batch script, resume the console call above in one file.
- [Here previous](https://pastebin.com/raw/yX4YV2V9) before inja integration, works well.
- [Updated version](https://pastebin.com/raw/n0Qc25Pf) for current master.

For finish here [the full log file](https://pastebin.com/raw/7gKqS6vV) from console.

4ian:
I can see:

>DCMAKE_TOOLCHAIN_FILE=C:\\emsdk\\emscripten\\1.37.37\\cmake\\Modules\\Platform\\Emscripten.cmake'] at C:\GDevelop\Binaries\embuild

in the logs. This means that you have some sort of mismatch between the new version and the old one.
I would recommend to: clear the Binaries/embuild folder, run `npm run clean` in GDevelop.js
Make sure to clear any environment variable that was set up by emscripten 1.37. In particular, I think there was a EMSCRIPTEN environment variable, be sure that it's not defined.

Bouh:
It's a bit better, emscriptem seem use the correct version now :)
Now issue is spesific about ninja.
[New log file](https://pastebin.com/raw/vyqMPx8W).

4ian:
> C:/emsdk/upstream/bin\llvm-ar.exe: error: CoreCMakeFilesGDCore.dirGDCoreEventsBuiltinCommentEvent.cpp.o: no such file or directory

Weird. Can you try cleaning everything and starting again the compilation? Remove Binaries/embuild another time and run again `npm run build`?

Bouh:
Same result.

I see when Building to go Linking

```
[...]

PR #1599, opened 2020-03-29
Add option for use MinGW instead ninja

Bouh:
Add this option for go back to MinGW instead ninja for user who have issue with ninja.
For more information about [ninja issue see here.](https://github.com/4ian/GDevelop/issues/1591)


4ian:
Thanks :) I fixed the usage of global variables (beware, in JS, no var/let/const = global variable) and ran Prettier.

PR #1601, opened 2020-03-30
[WIP] Mainframe refactor into functional Component

HarsimranVirk:
@4ian I've fixed some of the initial crashes and made it workable at the moment. Though I agree the code looks ugly and messy right now (since I've been fixing the constant massive crashes only :p ), we can fix this later and make everything more readable and pretty :)

There is heavy usage of setState callbacks, which are really not that compatible with the useState hook. As I've said before there were about 30 setState callbacks, but fortunately half of them were just to call `updateToolbar()`. All of those 15-ish callbacks are handled by a single `useEffect` hook, which could be seen as a quick win I'd say :D

Also, yes I made a custom hook similar to what you suggested before, 
1. It does handle simple callbacks pretty well :)
2. It just can't mimic the class' setState callback, however. Especially when the callbacks become pretty complex, like resolving promises inside nested callbacks.

I believe we could port these complex callbacks inside `useEffect` hooks, I've done this for two of them and fixed a lot of random crashes. Will it make the component re-render more times? Maybe, but I don't think any of those `useEffects` will have the same dependencies, so we should be good, right?


newIDE/app/src/MainFrame/index.js
```diff
@@ -1522,548 +1573,545 @@ class MainFrame extends React.Component<Props, State> {
...
+            // eslint-disable-next-line
+            useStorageProvider(storageProvider)
+              .then(() => openFromFileMetadata(fileMetadata))
+              .then(() => openSceneOrProjectManager());
+          },
+          onCreate: (project, storageProvider, fileMetadata) => {
+            openCreateDialog(false);
+            // eslint-disable-next-line
+            useStorageProvider(storageProvider)
+              .then(() => loadFromProject(project, fileMetadata));
```
4ian: Is there a reason not for adding `.then(() => openSceneOrProjectManager());` like before? (is it handled in a hook?)

PawBud:
>   * In theory you need to add a type `MainFrameInterface`, which is composed of all the functions that used to be public (i.e: those not starting with a _).

@4ian I just wanted to clarify a doubt about the '_' since I am a not-so-experienced JS developer, Is this really a convention used by JS developers to distinguish between a private & a public variable or is it some kind of rule?

newIDE/app/src/MainFrame/index.js
```diff
@@ -162,8 +162,31 @@ type Props = {
...
 
-class MainFrame extends React.Component<Props, State> {
-  state = {
+
+const useForceUpdate = () => {
+  const [value, setValue] = React.useState(0);
+  return (state: State) => setValue(value => ++value);
+}
+
+const useStateWithCallback = (initialValue: State) => {
```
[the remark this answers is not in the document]
HarsimranVirk: Yes, our custom hooks should be a part of Utils, but I'm honestly not confident about it :/ It really didn't work the way I expected :( It handled simple callbacks, but you get the point right? Maybe someone in the future decides to abuse it and couldn't figure out why all of a sudden everything crashed. Let's first see if we'd really need it even in the MainFrame or not :)

newIDE/app/src/MainFrame/index.js
```diff
@@ -511,38 +537,38 @@ class MainFrame extends React.Component<Props, State> {
     return editorTab.editorRef.getSerializedElements();
   };
 
-  toggleProjectManager = () => {
-    if (!this.refs.toolbar)
-      this.setState({
-        projectManagerOpen: !this.state.projectManagerOpen,
-      });
+  const toggleProjectManager = () => {
+    if (!!toolbar.current)
```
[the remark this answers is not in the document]
HarsimranVirk: `!!` simply means `true`, right? I mean is there any difference between `if(!!toolbar.current)` and `if(toolbar.current)`?

newIDE/app/src/MainFrame/index.js
```diff
@@ -635,23 +662,24 @@ class MainFrame extends React.Component<Props, State> {
...
+        ...state,
         editorTabs: closeExternalLayoutTabs(
-          this.state.editorTabs,
+          state.editorTabs,
           externalLayout
         ),
-      },
-      () => {
+      }),
+      (state) => {
```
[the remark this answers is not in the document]
HarsimranVirk: Well, we'd need to always pass `state` in the callback, otherwise, it will be called with the previous value of `state`. This is why I'm not really super happy about `useStateWithCallback` :/

newIDE/app/src/MainFrame/index.js
```diff
@@ -1522,548 +1573,545 @@ class MainFrame extends React.Component<Props, State> {
...
+    const { currentProject } = state;
+    const resourceSourceDialog = _resourceSourceDialogs.current[sourceName];
     if (!resourceSourceDialog) return Promise.resolve([]);
 
     return resourceSourceDialog.chooseResources(currentProject, multiSelection);
   };
 
-  updateToolbar() {
-    const editorTab = getCurrentTab(this.state.editorTabs);
+  const updateToolbar = (_state = state) => {
```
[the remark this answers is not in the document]
HarsimranVirk: I believe we don't need to pass anything at all now :) `updateToolbar()` is handled by the Effect now, not with the `useStateWithCallback`. (I explained above why we have to pass the state, it basically forces react to re-render and use the new state)

newIDE/app/src/MainFrame/index.js
HarsimranVirk: Okay, about this, I had some doubts too. This is the only instance where `loadFromProject()` is called and it used to resolve the promise inside the callback of setState.

> is it handled in a hook?

Yes, I had to extract the callback and put the code inside `useEffect`. But we know that this would basically mean we can't chain it anymore. Since there was only one instance of this function getting called, I handled it inside the hook.

Now, I'm unsure whether `loadFromProject` is called anywhere else other than MainFrame, because this would be an issue.

HarsimranVirk:
> There is also `ExternalEditor/index.js`

Is this the Monaco editor? If it is, then now I know why it crashes :) 

> I don't see any usage of useImperativeHandle. I deduce that the whole electron main menu (ElectronMainMenu.js) won't be working at all?

True, ElectronMainMenu doesn't work at the moment, (and ExternalEditor too, if it is the Monaco editor). I didn't really want to expose the methods initially, mostly because 
1. There are about 30+ functions which should be exposed, in theory. Though ElectronMainMenu calls only a subset of them.
2. Most of them used to crash the app (because of callbacks).

I would forwardRef it later when everything seems working pretty well :) Or we could make ElectronMainMenu a child component (definitely not in this PR, and preferable before this PR).
I'm not sure about ExternalEditor, will it work without `useImperativeHandle`? If not, then there is no other option :/

4ian:
> Is this the Monaco editor? If it is, then now I know why it crashes :)

No, the Monaco editor is in CodeEditor/index.js. ExternalEditor was used at the beginning to integrate GDevelop 5 as part of GDevelop 4 (basically for the scene editor). It was an experiment and never really published this way.

> There are about 30+ functions which should be exposed, in theory. Though ElectronMainMenu calls only a subset of them.

Yep, looking at the codebase the only component that takes a ref on MainFrame are `ElectronMainMenu` and `ExternalEditor` (because only them are surrounding `MainFrame`). So this means that you only need to look in these files the functions called with `this.editor`. 

Here they are:
```
chooseProject
saveProject
saveProjectAs
askToCloseProject
closeApp
openExportDialog
openCreateDialog
openProjectManager
openStartPage
openDebugger
openAboutDialog
openPreferences
openLanguage
openProfile
setUpdateStatus
```

You can forget about `ExternalEditor`. It's deprecated and I will remove it from the codebase. Just consider it as not existing :) 
Now there are two solutions:
1. Either we use useImperativeHandle to expose these 15 functions.
2. Or, in another PR to be merged before this one, we move ElectronMainMenu into MainFrame. In other words, we move it into a prop called `renderMainMenu` (it's important for it to be a render prop, because the web-app won't have a main menu) and call it in the MainFrame, passing it the 15 functions.

1. is easier to get this working as soon as possible. 2 is the long term goal/"best" solution in terms of architecture, as this drops the need for useImperativeHandle and unlock making the menu reactive to other things (for example, we'll be able to pass it the current project, so that we can enable or disable menu item if a project is opened or not).
I'll let you choose what you prefer to do :) 

> @4ian I just wanted to clarify a doubt about the '_' since I am a not-so-experienced JS developer, Is this really a convention(de-facto) used by JS developers to distinguish between a private & a public variable or is it some kind of rule?

This is a convention (i.e: not part of the language) that basically says:
- if a **variable** starts with _ in the parameter names of a function, this means that you're ignoring the parameter and want to show this (example in a callback: `doSomething((_event, value) => value + 1);` I'm "ignoring" _event). 
- if a **class method** or **class member** starts with _, it means it's private. This means that you can access it from the inside, using `this` (`this._myVariable = 3` is ok) but not from the outside (`myObject._myVariable = 3` is not OK.) 

You can see it because the underscore make the variable name look "ugly", to discourage you from using it. If you do `something._somethingElse._function();`, you can clearly see that you're trying to do some "ugly" stuff that will break in the future.

Note that it's not useful to mark variables in a function to with underscore (i.e:
```
const myFunction = () => {
  const myVariable = 3; // Ok
  const _myVariable = 3; // Useless, don't use an underscore here
}
```
) because these variables are automatically hidden from the outside world, so they are "safe" and don't need a convention to protect them. It's something called "hard privacy" (underscore is a "soft privacy").

In the PR there are variables starting with an underscore, but it's ok because this is to port an existing code. Getting a code right is here more important than following perfectly this convention - we can always rename super easily these variables later by just doing a search/replace.

PawBud:
Thanks for the info.

newIDE/app/src/MainFrame/index.js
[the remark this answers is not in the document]
4ian: Alright, let's keep it here for now! Let's use UseForceUpdate.js though :)

newIDE/app/src/MainFrame/index.js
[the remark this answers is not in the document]
4ian: !!something means "casting" the something to a boolean.
In practice, `if (something)` and `if(!!something)` are equivalent.

But my question was, are we sure we want to test this. But I think it's ok!

newIDE/app/src/MainFrame/index.js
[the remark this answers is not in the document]
4ian: Sure, but in most cases you don't need state in the callbacks. And because forceUpdate does NOT need state, you can remove it entirely from most callbacks :)

newIDE/app/src/MainFrame/index.js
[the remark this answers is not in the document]
4ian: Great if you don't need state in updateToolbar then feel free to remove it :)

newIDE/app/src/MainFrame/index.js
4ian: I think that's the only place where it's called: https://github.com/4ian/GDevelop/search?q=loadFromProject&unscoped_q=loadFromProject

4ian:
Head up, I've merged changes to master that includes some additions/changes in `MainFrame/index.js`: https://github.com/4ian/GDevelop/pull/1523/files#diff-46aaaf9f935d355e2260225b58b6338e

* Most forceUpdate have been put in a `_onProjectItemModified` function, 
* `unsavedChanges={this.props.unsavedChanges}` is passed to some components
* calls to `sealUnsavedChanges` have been added 
* An additional check in `askToCloseProject` was added

You should be able to resolve the conflicts (either using git or GitHub interface), let me know how it goes :)

HarsimranVirk:
> let me know how it goes :)

Everything went pretty well :) There were no issues after merging, except the already existing ones :/

There are no major crashes at the moment, everything seems to work fine. There is one constant crash though, Monaco editor, I'll look into this later.

> Maybe, but I don't think any of those useEffects will have the same dependencies

Now I've realised that I was wrong, there will be same dependencies if I try to extract all the callbacks and put them into useEffect. So, this means we'd have to rely on a custom hook which can handle callbacks, and I'm pretty adamant about not keeping this one :/

Anyway, I've been trying to create a bit more robust hook. I've come up with something, it can handle nested callbacks :) (Something our current custom hook can't do).

Please take a look at it, https://codesandbox.io/s/quizzical-hamilton-9ht5r, (I'm really sorry for making you look at these things, but I want to be sure if this is proper since it does feel a bit hacky to me).

4ian:
> Please take a look at it, https://codesandbox.io/s/quizzical-hamilton-9ht5r, 

I've been looking quickly, but I'm not sure why the promise version is working better than the callback one. They look very similar. Might be related to the fact that promise might be resolved on the next tick.
So while I can guarantee that it's 100% fool proof, it seems ok!

> There are no major crashes at the moment, everything seems to work fine

Cool :) Is that with the promise-based setState hook?

> Now I've realised that I was wrong, there will be same dependencies if I try to extract all the callbacks and put them into useEffect.

Yes I think that while having some useEffect is good, it's still surely safer and simpler to use callbacks/promise for things that need proper ordering.

HarsimranVirk:
> Is that with the promise-based setState hook?

I can say the promise-based hook works really well :) There's one drawback though, and if we design our components keeping this drawback in mind, it can pretty much replicate `this.setState`.

You'd have seen this line of code in the hook, `callback.current(state)` which is equivalent to `resolve(state)`. So, why do we need to pass the state? What happens when we don't pass the state? 
Now, each render has its own states and props, and pretty much everything (useEffect, other hooks, etc) is closed under each render of the component. So, in our custom hook, the promise was defined in the previous render and was resolved on the next render. Since it was defined in the previous render, it retained the reference of the previous state due to closure (I had to look through react's source and Dan Abramov's articles to understand this).  Even after being resolved on the next tick, the callbacks referred to the previous state.
Now, we can explicitly pass the current state while resolving, and this is why I had to write `callback.current(state)`. I wouldn't say it's a bug, but this is something related to class (mutable `this` which points to the current state `this.state`) vs function (closures).

Anyway, some of the really complex async tasks, such as `loadFromSerializedProject` `loadFromProject`, `openFileFromMetadata`, `closeProject`, etc, which are also intertwined into each other are best handled by separate useEffect hooks and not with our custom callback hook because it's really difficult to _inject_ the correct state into these functions. 

I don't see major crashes right now, but I'd say we need extensive manual testing since `MainFrame` really defines the entire application. I'd ask the community for help :)

Also, I'd start working on refactor of `ElectronMainMenu` so that we'll be able to test this PR on electron app sooner. I'm extremely sorry if everything seems to be delayed and slow :/ I'm learning (and re-learning) and a lot of things while working on this :)

newIDE/app/src/MainFrame/index.js
```diff
@@ -396,7 +393,7 @@ const MainFrame = (props: Props) => {
     [state.currentProject, state.currentFileMetadata]
   );
 
-  const openFromFileMetadata = (fileMetadata: FileMetadata): Promise<void> => {
+  const openFromFileMetadata = (fileMetadata: FileMetadata): Promise<any> => {
```
4ian, line 396: `void` is better than `any` (void means nothing returned, while any can accept anything). Can we go back to `void`?

newIDE/app/src/MainFrame/index.js
```diff
@@ -1715,8 +1663,9 @@ const MainFrame = (props: Props) => {
     return resourceSourceDialog.chooseResources(currentProject, multiSelection);
   };
 
-  const updateToolbar = (_state = state) => {
-    const editorTab = getCurrentTab(_state.editorTabs);
+  const updateToolbar = stateRef => {
```
4ian, line 1666: Not a big fan of "stateRef". `ref` has a meaning in React that I would rather not mix.
You might want to hint that this is a "reference" in JavaScript, but in JS all objects and arrays are reference, so this should not appear in the code.

Prefer `newState`. Also add the typing :)

4ian:
> I can say the promise-based hook works really well :)

That's good news :) 

> Now, we can explicitly pass the current state while resolving, and this is why I had to write callback.current(state)

Yes, this is fine. We just need to be sure to minimize the usage of this pattern (setState with promise). But it seems fine, as long as we use `state` passed to the promise as you described.

> which are also intertwined into each other are best handled by separate useEffect hooks and not with our custom callback hook

That's ok! I prefer using useEffect as it's "more correct" than our setState with promise.

> I don't see major crashes right now, but I'd say we need extensive manual testing since MainFrame really defines the entire application. I'd ask the community for help :)

Cool! I'll give a try too :) 

> Also, I'd start working on refactor of ElectronMainMenu so that we'll be able to test this PR on electron app sooner

Yes that's a good idea. 

> I'm extremely sorry if everything seems to be delayed and slow :/ I'm learning (and re-learning) a lot of things while working on this :)

No worries, I appreciate your investigations and especially around how React works with these state/promise and hooks :) Not easy, but it will give you a great understanding of how React components are working, which is invaluable for these kind of tasks - and others :) 

You're making good progress! If you can get ElectronMainMenu converted (either as a previous PR again or in this PR), we should be on a good position for this PR.

Thanks again for the hard work on this! Appreciate the fast progress you're making.

4ian:
ElectronMainMenu is now passed to MainFrame as a prop. Should make things simpler because no need for useImperativeHandle anymore at all :)

newIDE/app/src/MainFrame/index.js
```diff
@@ -1522,597 +1498,554 @@ class MainFrame extends React.Component<Props, State> {
...
+        state.createDialogOpen &&
+        renderCreateDialog({
+          open: state.createDialogOpen,
+          onClose: () => openCreateDialog(false),
+          onOpen: (storageProvider, fileMetadata) => {
+            openCreateDialog(false);
+            // eslint-disable-next-line
+            useStorageProvider(storageProvider)
+              .then(() => openFromFileMetadata(fileMetadata))
+              .then(state => openSceneOrProjectManager(state));
```
HarsimranVirk, line 1896: This is not working for web-app at all, while everything seems to work pretty well for electron :/ I'm calling `openFileFromMetadata` and `openSceneOrProjectManager` in a similar way they are getting called from `chooseProject`. I can't really understand what's going wrong.

newIDE/app/src/MainFrame/index.js
```diff
@@ -307,60 +326,53 @@ class MainFrame extends React.Component<Props, State> {
...
+    // the URL to a resource with a name in the old project is not re-used
+    // for another resource with the same name in the new project.
+    ResourcesLoader.burstAllUrlsCache();
+    // TODO: Pixi cache should also be burst
+    newState = await setState({
+      ...newState,
+      currentProject: project,
+      currentFileMetadata: fileMetadata,
     });
+
```
HarsimranVirk, line 362: Up to this point, the correct state is "injected" for web-app, but right after this it automatically falls back to its closure, I can't understand how or why this is happening, since it's working fine for electron.

newIDE/app/src/MainFrame/index.js
4ian: Are you getting an error message? Can you post it here?
It might be because if you're trying Google Drive, it's not working locally.

HarsimranVirk:
@4ian the issue of start page is resolved, I passed the correct state from the `render()` method. Now, the electron version is working pretty well, but there's something weird with the web-app, I've left some comments above.

newIDE/app/src/MainFrame/index.js
HarsimranVirk: There is no error, the create dialog just closes, and nothing happens. I'm tracking every render and the state of that render, so the `fileMetadata` and `currentProject` is set correctly, but immediately on the next render everything falls back to `null`

HarsimranVirk: > It might be because if you're trying Google Drive, it's not working locally.

Yes, that doesn't work too, but this is not the issue here, I'm unable to open even the examples, there is some issue involving state here.

HarsimranVirk: Alright I found what's wrong, babel transpiles async functions to generators for browser, and native promises for node. This is why it is running without any issue on electron but not on browsers 😅 I'm converting everything back to promises.

4ian: Great investigation and great finding!
Using promises everywhere is fine :)

4ian:
Sorry I lost track of where we are there! :)
Are there any remaining issue in Electron and/or the webapp, or things to fix/to do in the code? :)

HarsimranVirk:
> Are there any remaining issue in Electron and/or the webapp, or things to fix/to do in the code? :)

There could be some issues I believe, (I'm really just solving the ones I could find, and it's a huge component and it's very possible that I could miss a lot of issues :/ )
Progress:
1. Converted everything back to promises, so everything works as expected in browser.
2. Most, if not all, asynchronous tasks are in order and work fine, which includes opening/loading of projects, previewing scenes, (I still have to test if exporting works well or not).
3. There's only one useEffect left with linter warning (I've begin fixing linter warnings yesterday).

4ian:
Ok so I think the next thing to do is for me to get your branch and do some testing to see if I can spot any regression! :)

4ian:
I quickly tried but have a few issues:

* If you right click to choose a scene to be used as the first scene for preview, it's not working:
image

* By trying to set a scene to be the first scene to be used then to launch a preview in another scene, I have an infinite loop:
image


* There is a check recently added testing if a game has unsaved changes before closing it - seems that it was not ported in this refactoring.

* There are multiple warnings in the console. Take a look in particular in those related to hooks:

```
  Line 45:10:   'timePromise' is defined but never used                                                                                                                                                                               no-unused-vars
  Line 245:6:   React Hook React.useEffect has missing dependencies: '_loadExtensions', '_openInitialFileMetadata', '_openIntroDialog', 'openStartPage', 'props', and 'setState'. Either include them or remove the dependency array  react-hooks/exhaustive-deps
  Line 269:5:   React Hook React.useEffect has a missing dependency: 'updateToolbar'. Either include it or remove the dependency array                                                                                                react-hooks/exhaustive-deps
  Line 521:9:   'getSerializedElements' is assigned a value but never used                                                                                                                                                            no-unused-vars
  Line 877:13:  'previewFirstSceneName' is assigned a value but never used                                                                                                                                                            no-unused-vars
  Line 877:36:  'isPreviewFirstSceneOverriden' is assigned a value but never used                                                                                                                                                     no-unused-vars
  Line 923:5:   React Hook React.useEffect has missing dependencies: 'setState', 'state.isPreviewFirstSceneOverriden', and 'state.previewFirstSceneName'. Either include them or remove the dependency array                          react-hooks/exhaustive-deps
```

newIDE/app/src/MainFrame/index.js
```diff
@@ -853,54 +840,72 @@ class MainFrame extends React.Component<Props, State> {
...
+      const previewedLayout = _layoutPreview.layout;
+      const project = _layoutPreview.project;
+      const options = _layoutPreview.options;
+      if (state.previewFirstSceneName && state.isPreviewFirstSceneOverriden) {
+        if (project && project.hasLayoutNamed(state.previewFirstSceneName)) {
+          _setLayoutPreview(layoutPreview => ({
+            ...layoutPreview,
+            layout: project.getLayout(state.previewFirstSceneName),
+          }));
+        }
```
4ian, line 905: To fix the infinite loop, when choosing a scene to be used as the first scene for all previews (see my other comment explaining how to reproduce the bug), I'm pretty sure this should be moved to `_launchLayoutPreview `.
There should be no calls to `_setLayoutPreview` in the effect that is responsible for... reacting to changes made in `_setLayoutPreview` ;)

HarsimranVirk: I've fixed this in the latest commits, give it a try once again and tell me if it's working properly :)

HarsimranVirk:
> If you right click to choose a scene to be used as the first scene for preview, it's not working

> I have an infinite loop by doing this: 1) make a right click on play button, 2) Choose "Use this scene to start all preview", 3) Click on play button (left click, as usual) 

It should be fixed in the latest commits :)

> There is a check recently added testing if a game has unsaved changes before closing it - seems that it was not ported in this refactoring.

I've merged it already, I guess. Is it `props.unsavedChanges` ? I see that prop being used everywhere in `MainFrame`. Anyway, maybe there's something wrong, I'll check it side by side with master branch :)

> There are multiple warnings in the console. Take a look in particular in those related to hooks

There's only one warning remaining related to hooks, I need some help regarding that, please check the review comments :)

Also, I tried merging all the changes happened to the master branch recently, is there any breaking change? The layout wasn't rendering on the `SceneEditor`, it was totally blank, and dragging objects into the layout did nothing at all. I reverted back from that merge, so this branch is definitely missing the latest commits of master.

newIDE/app/src/MainFrame/index.js
```diff
@@ -214,35 +214,49 @@ const MainFrame = (props: Props) => {
...
+            performance.now(),
+          ]);
+          console.info('Startup times:', getStartupTimesSummary());
+        })
+        .catch(() => {
+          /* Ignore errors */
+        })
+      })();
+    },
+    [integratedEditor, initialFileMetadataToOpen, setState, _openIntroDialog]
```
HarsimranVirk, line 260: Linter warning asks to add `_openInitialFileMetadata`, `openStartPage`,`_loadExtensions` and `introDialog`.

I initially used `useCallback` for `_loadExtensions`, but including `_loadExtensions` (after useCallback) and `introDialog` into the deps list results in not opening of projects. I don't really understand why this happens :/

`_openInitialFileMetadata` and  `openStartPage` are basically rabbit holes, they depend upon a lot of other functions, which directly depend upon `state`. Wrapping all of them in a `useCallback` would mean that this `useEffect` will run on every render. Do you have some suggestions?

4ian:
> It should be fixed in the latest commits :)

Seems that it works, thanks!

> The layout wasn't rendering on the SceneEditor, it was totally blank, and dragging objects into the layout did nothing at all.

There was an update to react-measure, be sure to `npm install` (and re-run `npm start` to be sure) after updating to latest master.

---

## About effects

> _openInitialFileMetadata and openStartPage are basically rabbit holes, they depend upon a lot of other functions, which directly depend upon state. Wrapping all of them in a useCallback would mean that this useEffect will run on every render. Do you have some suggestions?

I guess this means that we would have to wrap tons of the functions into useCallback right?
Exceptionally I think we could be good to go ahead with a `// eslint-disable-next-line` to disable the warning. This effect is really a "component did mount". We don't want it to run **at all** after it's run once.
So I would go a bit further actually: *I would set this effect array to be an empty array (`[]`)*, so that it runs only once after the component is mounted and add `// eslint-disable-next-line`.

The warning is right to say that this effect won't be re-run if any of the dependency (initialFileMetadataToOpen, setState, _openIntroDialog, _loadExtensions...) is changed... but we don't really care. It's just the effect that set up the editor for the first run. After that, we "drop" it and don't want it anymore.

So this means that with this effect solved, we only have:
* This "component did mount"-like effect, with a comment to disable the linter.
* The useEffect for the toolbar (to re-render the toolbar whenever tabs are changed/updated) (make sense).
* The "debugging" useEffect (very useful, thanks for adding it) that we can remove later.

So only 3 effects, that's great!

----

## Other stuff

Seems like the `_onChooseResource` does not work. If you open an object and click to choose a resource:

image

it will crash.

image

The reason is again that we're using the old state in _onChooseResource. Same issue in `_openInstructionOrExpression`. You can reproduce the issue by opening the space shooter example, then double click "Handle the movement of Player [...]" action, then click on the button to open the action in the extension editor. It won't work:

image

In the same vein, if you use the feature to extract an event into a function:
image

Then `_onCreateEventsFunction` will be called but with the *old state* (because of the way closures work), so the function will bail out.

Unrelated: feel free to remove `getSerializedElements`, it's dead code that I'll remove otherwise later :)

----

## Summary

To go ahead, see:
* The effect I discussed
* See if you can find how to fix closures relying on the old state... A useCallback maybe, to be sure that they have the proper state passed to them??
* if you can update to latest master, be sure to `npm install`
* Verify why `askToCloseProject` is not showing the unsaved changes warning - might be because of the same issue as closures. It should show after you do any change in the project, when an asterisk is shown in the titlebar: 
image

HarsimranVirk:
> Seems like the _onChooseResource does not work. If you open an object and click to choose a resource

> Same issue in _openInstructionOrExpression.

> Then _onCreateEventsFunction will be called but with the old state (because of the way closures work), so the function will bail out.

I've fixed all of them in the latest commits, should be work fine now.

> if you can update to latest master, be sure to npm install

Yes, we are in sync with master branch now :)

> Verify why askToCloseProject is not showing the unsaved changes warning - might be because of the same issue as closures.

Yes, it was because of closures, but this time stale props were being passed (instead of state). This is why `props.unsavedChanges` was false every time, which made it not work properly. I've fixed this now, and new props are being passed.

Also, I've found something while working on this. If we open a different project from electron main menu's `Open...`, we wouldn't trigger the `unsavedChanges` warning, because there is no call to `askToCloseProject`, it just closes the project because `openFromFileMetadata` eventually calls `closeProject` after a series of function calls. I could refactor a bit to get the warning, should I do this in the next commit right here? (A different PR might be an overkill :p ) 

Let's find some more bugs to make it regression-free :D

newIDE/app/src/UI/SearchBar.js
```diff
@@ -163,7 +161,6 @@ export default class SearchBar extends React.PureComponent<Props, State> {
           ...styles.root,
           ...style,
         }}
-        elevation={this.props.elevation || 1}
```
4ian, line 166: I'm confused why we're seeing this in the PR? oO

HarsimranVirk: Okay I didn't see this. All of the changes from different PRs are shown in my commit related to merging the master. 🙄 This is odd.

4ian: Ah right I've seen this happen already, might be a GitHub bug. Closing this PR and re-opening another one should fix this :)

PR #1602, opened 2020-03-30
Add Prettier to electron-app

NilayMajorwar:
Added Prettier to `electron-app`, and applied formatting.

@4ian Just a few things to clarify:
- The file glob is just a `**/*.js` wildcard, excluding directories in`.gitignore`. Is that okay?
-  I noticed that updating dependencies seems to increment the version in `electron-app/app/package-lock.json` to `5.0.0-beta89`. Is it intended?

### DIFF
--- a/newIDE/electron-app/.prettierrc
+++ b/newIDE/electron-app/.prettierrc
@@ -1,0 +1,4 @@
+{
+  "singleQuote": true,
+  "trailingComma": "es5"
+}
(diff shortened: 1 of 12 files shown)


4ian:
> The file glob is just a **/*.js wildcard, excluding directories in.gitignore. Is that okay?

Yep sounds good.

> I noticed that updating dependencies seems to increment the version in electron-app/app/package-lock.json to 5.0.0-beta89. Is it intended?

Yes, no big deal!

Thanks for doing this, looks great 👍Will help keeping the formatting of these files always in a perfect state :)

PR #1607, opened 2020-03-31
Add scene dialog boxes to Project Manager

### LINKED ISSUE #1505: Accessibility: Add a context menu to scenes under project panel for scene variables

NilayMajorwar:
This PR adds the option of opening the dialog boxes for scene properties and scene variables from the Project Manager. Resolves #1505. 
Here's a screenshot:

scene-in-pm

- Modified the `Item` component in `ProjectManager/index.js` to take additional context menu options if provided, and append them to the base menu. 
- Had to fix an issue related to Flow - see changes in `ScenePropertiesDialog.js`. 

  Earlier, it took a compulsory prop `onOpenMoreSettings`, which is actually `undefined` when used by the mainframe (rightclicking on scene and opening scene properties) (confirmed from React DevTools).
 
  I'm not sure how Flow allowed that - [`SceneEditorContainer` here](https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/MainFrame/Editors/SceneEditor.js) does not pass `onOpenMoreSettings` to `SceneEditor` at all, for example, and Flow never complained.

### DIFF
--- a/newIDE/app/src/ProjectManager/index.js
+++ b/newIDE/app/src/ProjectManager/index.js
@@ -39,9 +39,9 @@ import AddToHomeScreen from '@material-ui/icons/AddToHomeScreen';
 import Fullscreen from '@material-ui/icons/Fullscreen';
 import FileCopy from '@material-ui/icons/FileCopy';
 import AccountCircle from '@material-ui/icons/AccountCircle';
-import { type UnsavedChanges } from '../MainFrame/UnsavedChangesContext';
 import ScenePropertiesDialog from '../SceneEditor/ScenePropertiesDialog';
 import SceneVariablesDialog from '../SceneEditor/SceneVariablesDialog';
+import { type UnsavedChanges } from '../MainFrame/UnsavedChangesContext';
 
 const LAYOUT_CLIPBOARD_KIND = 'Layout';
 const EXTERNAL_LAYOUT_CLIPBOARD_KIND = 'External layout';
--- a/newIDE/app/src/ProjectManager/index.js
+++ b/newIDE/app/src/ProjectManager/index.js
@@ -155,64 +155,6 @@ class Item extends React.Component<ItemProps, {||}> {
     }
   }
 
-  _buildMenuTemplate = () => {
-    const menuTemplate = [
-      {
-        label: 'Edit',
-        click: () => this.props.onEdit(),
-      },
-      {
-        label: 'Rename',
-        click: () => this.props.onEditName(),
-      },
-      {
-        label: 'Delete',
-        click: () => this.props.onDelete(),
-      },
-      {
-        label: this.props.addLabel,
-        visible: !!this.props.onAdd,
-        click: () => this.props.onAdd(),
-      },
-      { type: 'separator' },
-      {
-        label: 'Copy',
-        click: () => this.props.onCopy(),
-      },
-      {
-        label: 'Cut',
-        click: () => this.props.onCut(),
-      },
-      {
-        label: 'Paste',
-        enabled: this.props.canPaste(),
-        click: () => this.props.onPaste(),
-      },
-      {
-        label: 'Duplicate',
-        click: () => this.props.onDuplicate(),
-      },
-      { type: 'separator' },
-      {
-        label: 'Move up',
-        enabled: this.props.canMoveUp,
-        click: () => this.props.onMoveUp(),
-      },
-      {
-        label: 'Move down',
-        enabled: this.props.canMoveDown,
-        click: () => this.props.onMoveDown(),
-      },
-    ];
-
-    // Append the extra menu options (if provided) to base menu
-    const addedMenu = this.props.extraMenuOptions;
-    if (addedMenu && addedMenu.length !== 0) {
-      menuTemplate.push({ type: 'separator' }, ...addedMenu);
-    }
-    return menuTemplate;
-  };
-
   render() {
     const label = this.props.editingName ? (
       <TextField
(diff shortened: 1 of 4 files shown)


newIDE/app/src/ProjectManager/index.js
```diff
@@ -130,6 +132,7 @@ type ItemProps = {|
   onMoveUp: () => void,
   canMoveDown: boolean,
   onMoveDown: () => void,
+  addonMenuOptions?: Array<Object>,
```
4ian, line 135: Let's call this "extraMenuItems" and use a proper typing instead of Object.

newIDE/app/src/ProjectManager/index.js
```diff
@@ -368,6 +385,14 @@ export default class ProjectManager extends React.Component<Props, State> {
     this._onProjectItemModified();
   };
 
+  _onOpenLayoutProperties = (layout: gdLayout) => {
+    this.setState({ lastOpenLayout: layout, layoutPropertiesDialogOpen: true });
```
4ian, line 389: You don't need both lastOpenLayout and layoutPropertiesDialogOpen.
Just use a single variable: `editedPropertiesLayout`, which is either `null`(nothing is edited) or a `gdLayout` (in which case, the dialog is opened with the layout).

newIDE/app/src/ProjectManager/index.js
```diff
@@ -368,6 +385,14 @@ export default class ProjectManager extends React.Component<Props, State> {
...
+  };
+
+  _onOpenLayoutVariables = (layout: gdLayout) => {
+    this.setState({ lastOpenLayout: layout, layoutVariablesDialogOpen: true });
```
4ian, line 393: Same :) 
You'll also need to accept the parameter as `?gdLayout`, so we can close the dialog by passing null.

newIDE/app/src/SceneEditor/ScenePropertiesDialog.js
```diff
@@ -201,7 +201,7 @@ export default class ScenePropertiesDialog extends Component<Props, State> {
             label={<Trans>Open advanced settings</Trans>}
             fullWidth
             onClick={() => {
-              this.props.onOpenMoreSettings();
+              this.props.onOpenMoreSettings && this.props.onOpenMoreSettings();
```
4ian, line 204: Prefer a `if` than a && for readability.

newIDE/app/src/SceneEditor/SceneVariablesDialog.js
```diff
@@ -0,0 +1,37 @@
+// @flow
+import * as React from 'react';
+import { Trans } from '@lingui/macro';
+import VariablesEditorDialog from '../VariablesList/VariablesEditorDialog';
+
+type Props = {|
+  open: boolean,
+  layout: gdLayout,
+  onApply?: () => void,
```
4ian, line 9: Should onApply be *not* optional?

newIDE/app/src/SceneEditor/SceneVariablesDialog.js
```diff
@@ -0,0 +1,37 @@
...
+  onClose: () => void,
+|};
+
+export default class SceneVariablesDialog extends React.Component<Props, {||}> {
```
4ian, line 13: Use a functional component

4ian:
Looks mostly good! 👍
I've added mostly nitpicking comments about the code. The overall approach looks good.

newIDE/app/src/SceneEditor/SceneVariablesDialog.js
NilayMajorwar: I made it optional to make it similar to the scene properties dialog - it too has optional `onApply` and on pressing OK, calls the function only if it exists. It does seem strange, since Esc or clicking outside the dialog does not seem to close the dialog, so the user is almost stuck if `onApply` were not defined. Should I make `onApply` mandatory in `ScenePropertiesDialog` too?

4ian: Yes I think we can make it mandatory in `ScenePropertiesDialog` too. `onApply` was surely made optional for an historical reason, but should not be anymore :)

NilayMajorwar: Done!

newIDE/app/src/SceneEditor/SceneVariablesDialog.js
NilayMajorwar: Done.

newIDE/app/src/ProjectManager/index.js
NilayMajorwar: Done.

newIDE/app/src/ProjectManager/index.js
NilayMajorwar: Done.

newIDE/app/src/SceneEditor/ScenePropertiesDialog.js
NilayMajorwar: Done.

newIDE/app/src/ProjectManager/index.js
```diff
@@ -165,6 +176,61 @@ class Item extends React.Component<ItemProps, {||}> {
...
+        label: 'Copy',
+        click: () => this.props.onCopy(),
+      },
+      {
+        label: 'Cut',
+        click: () => this.props.onCut(),
+      },
+      {
+        label: 'Paste',
+        enabled: this.props.canPaste(),
```
4ian, line 208: This line is a problem. Can you see why? 

The idea of having buildMenuTemplate being a function is that you:
* Don't pay for the construction of the menu template when it's not needed (here, you create one for each item... for large projects with a layout of scenes/etc this can start to be noticeable).
* You only construct the menu at the time it's needed... so you can inspect things like the clipboard to enable/disable menu items.

Here `canPaste` is called at the time the item is rendered... but what if the clipboard was changed since then? :)

Solution: move this back to the buildMenuTemplate function

NilayMajorwar: Ohh okay, got it.. sorry, didn't think about that :sweat_smile:.
I'll fix this right away.

NilayMajorwar: Done. I've just pulled out `buildMenuTemplate` into a class function to make the render function a bit cleaner. I hope it's fine now.

4ian: Can we actually keep it in the render? I know it's counter intuitive when used to React classes, but as an effort to migrate progressively components to React functions with hooks, which are shorter and safer, it's actually ok to keep stuff in render (unless it's not good for readability or performance). Here I think it's totally fair to consider the menu items as part of the render - it's the user interface :)

NilayMajorwar: Oh okay right, done.

4ian:
I think this will be good to go, doing final checks!

4ian:
Thanks for working on this! 👍 I've done a few updates in 9ff6d91717dbec4bcfd57d3d4174b85a83cc5b54 to simplify building the menu and to use a function for consistency with how menu are generated.

4ian:
Also I made a file with the flow type for the menu items, so that we have a common source for the whole app, located with the components rendering the menus: https://github.com/4ian/GDevelop/commit/1071b66c92576f11ecb64883fdeb01c3de335efe

PR #1608, opened 2020-04-01
Add Readme in root Dialog

gautamv95:
Not necessarily needed, just a small task I did for #1522 .
Contains a list of all dialogs, which use root Dialog in the project

### DIFF
--- a/newIDE/app/src/UI/Dialog/readme.md
+++ b/newIDE/app/src/UI/Dialog/readme.md
@@ -1,0 +1,78 @@
+`cannotBeDismissed` is a prop which when passed through a Dialog, will do the following depending on its value:
+
+1. True -> The Dialog will not close when clicked out of its scope(backdropClick)
+2. False -> The Dialog will close when clicked out of its scope
+
+#### Selector Dialog
+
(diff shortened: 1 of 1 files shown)


4ian:
Ideally I would like to avoid having this file in the codebase, as it will probably go really quickly out of date (it's too easy to forget to add something to this doc when a new dialog is made).
Also ideally, this documentation is not necessary: the prop is acting itself as a documentation :) 
If the true/false values are not obvious, then the prop should either be renamed to make it obivous, or the comment updated. 
If we want to keep the distinction between "selector dialog", "info dialog" and "editor dialog", then we should put this in the code as a prop (`dialogType="..."`) and remove `cannotBeDismissed`. See the idea? :) 

I think we're good for now with `cannotBeDismissed`! :)

gautamv95:
Indeed, the file would go out of date, I think we can drop this for now :)

That's a really nice idea, but would still not remove the code duplication
> If we want to keep the distinction between "selector dialog", "info dialog" and "editor dialog", then we should put this in the code as a prop (dialogType="...") and remove cannotBeDismissed. See the idea? :)

Let's go with `cannotBeDismissed` for now :)

PR #1612, opened 2020-04-01
Improve bug report

Bouh:
Added : 
Arch is for 32 or 64 bits
Platform for linux, macOS, windows, android, etc
System version for know which version of the os.

I've not found how crash the IDE for test, and my storybook don't work, i've never used it before.

### DIFF
--- a/newIDE/app/src/Version/index.js
+++ b/newIDE/app/src/Version/index.js
@@ -1,5 +1,6 @@
 // @flow
 import VersionMetadata from './VersionMetadata';
+const process = require('process');
 
 export const getIDEVersion = (): string => VersionMetadata.version;
 export const getIDEVersionWithHash = (): string =>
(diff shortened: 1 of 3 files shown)


4ian:
Thanks! Would you be able to move that in `Utils/Platform.js`? :) 
Also use optionalRequire because process won't be available in the webapp (if you're using the one with require instead of optionalRequire, you will get the "process" that is faked by Webpack I think, which will hardcode the characteristics of your computer). Be sure to check if process is not null (it will be null on the web-app, meaning that either we put "unknown" or we find a web api to give us some information) (for example, we can put the user agent!)

newIDE/app/src/Version/index.js
```diff
@@ -13,3 +14,22 @@ export const getGDCoreVersion = (): string => {
...
+  return platformString;
+};
+
+export const getSystemVersion = (): string => {
+  return process.getSystemVersion
+};
+
+export const getArch = (): string => {
+  return process.arch;
+};
```
4ian, line 35: Might be a good idea to also add:
* https://developer.mozilla.org/en-US/docs/Web/API/NavigatorID/userAgent
* A version of getPlatform to get the system in a webapp: `https://developer.mozilla.org/en-US/docs/Web/API/Navigator/oscpu`
* Maybe we should put getSystemVersion inside getPlatform. Let's in this case rename "getPlatform" to "getOperatingSystemDescriptiveName"

4ian:
Merged, thanks! Will be very useful :) 
I've separated the details into 4 points so that we're extra sure what we get, even if the info is not there (for the web-app).

PR #1614, opened 2020-04-02
Added Tab Navigation In the variables dialog

PawBud:
This is a fix regarding the issue: _#1561_

The reason I did not use shouldComponentUpdate is because 

The picture is self-explanatory.

### DIFF
--- a/newIDE/app/src/VariablesList/index.js
+++ b/newIDE/app/src/VariablesList/index.js
@@ -244,7 +244,7 @@ export default class VariablesList extends React.Component<Props, State> {
             success = parentVariable.renameChild(name, text);
           }
 
-          this.setState({
+          React.PureComponent({
             nameErrors: {
               ...this.state.nameErrors,
               [variable.ptr]: !success,


PawBud:
https://drive.google.com/file/d/1TLm_QTNbpaCjFz5IoJ1lgyG-EE6pji9A/view?usp=sharing
Here is a screen cast video demonstrating how it works after the change

newIDE/app/src/VariablesList/index.js
```diff
@@ -244,7 +244,7 @@ export default class VariablesList extends React.Component<Props, State> {
             success = parentVariable.renameChild(name, text);
           }
 
-          this.setState({
+          React.PureComponent({
```
4ian, line 247: Unfortunately this does not make sense.
I would recommend spending more time getting familiar with React, what is a state, what is a prop, how a component works.

PawBud: I don't mind doing that, but why is the tab working then?

PawBud: For now, I think I will close this PR and try to attain a better solution :)

4ian: It's not really "working", you removed a setState so there was no-rendering of the components, and also no error shown if you enter a bad name :)

PawBud: oh, I think you are right, I will do a course on React and I am sure that  I will get much better at it in a week's time. 
Thanks for the advice.

PR #1624, opened 2020-04-05
Add preload auto for fix video in next upgrade of electron v8.2.0

Bouh:
https://github.com/4ian/GDevelop/pull/1589#issuecomment-609425073

### DIFF
--- a/newIDE/app/src/ObjectsRendering/PixiResourcesLoader.js
+++ b/newIDE/app/src/ObjectsRendering/PixiResourcesLoader.js
@@ -140,7 +140,8 @@ export default class PixiResourcesLoader {
       true /* Treats request as cross-origin */,
       false /* autoplay */
     );
-
+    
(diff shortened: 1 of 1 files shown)


4ian:
Let's add a comment saying it's only for Pixi 4.8.6 and can be removed in Pixi 5 :)

4ian:
Done, merging now! Thanks :)

4ian:
Also @Bouh, should this be done in the runtime object too, for exported games? Or is it not useful?

Bouh:
Bug is only in IDE, the runtime object renderer (preview and exported games) have been already edit in the past.

https://github.com/4ian/GDevelop/blob/master/Extensions/Video/videoruntimeobject-pixi-renderer.js#L27

4ian:
Ah I missed that, great then 👍

PR #1625, opened 2020-04-05
Add autocompletions when typing an expression

4ian:
This adds an "autocomplete" feature showing possible completions when you write an expression. The completions can be object names, behavior names or expressions. The completions "understand" the context and only show what you can use.

Still to be done:
- [x] Fix TODOs in the code
- [ ] Check if selecting a completion works on touch screens

Supported features:

- Completions are appearing when something is typed or when the cursor is moved in an expression field.
  - Press up/down to go through the completions, press Enter to add one.
  - Completions can be dimissed with Escape.
- Example where they display object expressions and behaviors:
  image
- Support editing an object name (or behavior or expression) already entered:
  image
- Display the description of the expression when the caret is on a parenthesis:
  image

If you want to try it:
- You can try it on https://editor-alpha.gdevelop-app.com (**make sure to clear your cache, close and then reload the app to clear any previous version**).


blurymind:
This is a HUGE UX improvement! Thank you 😊

Wend1go:
This will speed up development imensely, good work. :+1: 
Two suggestions:
1) Case insensitive search
For example when searching for an objects "animation" property it is not found since it is called "Animation"
2) "Tab" to autocomplete
Most applications use "Tab" for autocompletion. "Enter" feels a little odd.

blurymind:
It would be amazing if tabbing also gives automatic focus to the next field in the same row

4ian:
> Case insensitive search
> For example when searching for an objects "animation" property it is not found since it is called "Animation"

This seems to work?

image

(I typed `animation` and `Animation()` is showing).

> "Tab" to autocomplete

Sounds fair, I can look to add this.

> It would be amazing if tabbing also gives automatic focus to the next field in the same row

In the events sheet you mean?
Yep It's a good idea. We would need to close the "InlinePopover" and open the next parameter - if I understand properly.

> That would enable us to use the event sheet completely with the keyboard -its the dream ™️

Right now it's true that it's a bit missing. You can already tab after selecting an action or condition - and press Enter to either open the action/condition or edit the parameter :) But when you press "Escape" to close the "InlinePopover", you're losing the focus (also some popover are not closing when you press Escape).
If we were giving back focus and fix the Escape for all popover, this would give a great boost in keyboard usability.

Wend1go:
> > Case insensitive search

> This seems to work?

Oh yes, my bad.

Do you think you could also add (object-)variable name completions?

4ian:
I've added Tab in addition to Enter to choose a completion.

Variable would definitely be great and I guess kind of doable. I'll see if I do it now or keep it for a later feature. Better have this out sooner than later.

Bouh:
wow cool ! 
can we get the list under the input, on top it's very weird ?
image

4ian:
Yep I moved them under the input, it's better indeed! (above the error text, which are not super useful when you're typing something - you can still see errors by pressing escape or clicking anywhere to dismiss the completions).

Bouh:
Icons for group is one cube, correct icon use 3 cubes, 1 is for an objet.
List style shoud use same style as expression list.
(If you use expression list for autocompletion, see why the list have an offset when the input show an error.)
image

Bouh:
Scrollbar on X is ugly, this one is fixed if you use the component from expression list.
image

With tab, when we navigate in the list, you have set bold on the current selection, it's a bit tiny, maybe change the background color (like expression list) 
image

4ian:
I won't use the component from the expression list because i want a much more compact display, like in VSCode/IDE:
image

> Scrollbar on X is ugly

I've not seen it when testing, can you tell me the expression and what to complete to reproduce?

Bouh:
It's for MyDummyExtension::DummyExpression()
image




> I won't use the component from the expression list because i want a much more compact display, like in VSCode/IDE:

Maybe we can add an option on expression list component for compact style, remove folder. and remove search box. And auto width on the width of the input.
image

Because without auto width selection can overlap the rest of the UI.

image

4ian:
Fixed the horizontal scrollbar :) 

> Maybe we can add an option on expression list component for compact style, remove folder. and remove search box. And auto width on the width of the input.

The thing is that the expression selector is, well, for displaying expression - while this popup is to help completing what you started to write.
While it's useful for discovery, I think we should keep the expression selector to be a bit different, more optimized for discover (even if we could show the expression being written, like "Count(Object)" instead of the name, i.e: "The number of objects", when browsing them).

I think both have their utility. One showing expressions that you can search, the other completing what you started to write. 

> Because without auto width selection can overlap the rest of the UI.

It's kind of ok, I was at some point thinking of making the popup following the cursor. We could give it a try with a width that is the same as the input - but again this is not a selector, this is an autocompletion. I think it's important that we feel that it's there just to give a bit more information about what is being written.

4ian:
Unless there are major concerns, I think I'll go ahead with this, at least for a new testing version.
It's I think important to have this soon as it's a much asked for feature. The current state is I think good enough to be usable - we can always continue the discussion here about improvements.

I've made a card to add variables in autocompletions: https://trello.com/c/Z3UXNh5D/424-add-variables-scene-global-or-object-variables-in-autocompletions-of-expressions

PR #1626, opened 2020-04-05
Disable some menu items if action not possible

NilayMajorwar:
This PR partially resolves the issue #1550 by disabling menu items in File and View menus when the related action is not possible (for instance you can't save a project at the opening screen as there isn't any open project). 

I've refactored MainFrame and ElectronMainMenu so that the menu component is now contained inside the mainframe. MainFrame now takes a new prop `mainMenu` whose value is the class `ElectronMainMenu` (Flow type: `Class<React$Component<*, *>>`) and renders it, passing as props the current project and relevant IDE functions.

I know the new prop `mainMenu` probably needs a bit of renaming and refactoring. I tried using `renderMainMenu={(props) => <ElectronMainMenu {...props}>}` so that it is a bit cleaner and falls in with rest of the similar props. But then Flow typing of the prop becomes a problem - I'll have to import the `ElectronMainMenu` props type into MainFrame. It should work fine, but may get a bit worse in the future when we add a menu to the web app too. Should I move forward with it?

### DIFF
--- a/newIDE/app/src/MainFrame/ElectronMainMenu.js
+++ b/newIDE/app/src/MainFrame/ElectronMainMenu.js
@@ -1,32 +1,11 @@
 // @flow
 import * as React from 'react';
 import optionalRequire from '../Utils/OptionalRequire';
-import { type I18n as I18nType } from '@lingui/core';
 import { t } from '@lingui/macro';
 import { isMacLike } from '../Utils/Platform';
-import { type UpdateStatus } from './UpdaterTools';
+import { type MainMenuProps } from './MainMenu.flow';
 const electron = optionalRequire('electron');
 const ipcRenderer = electron ? electron.ipcRenderer : null;
-
-type Props = {|
-  i18n: I18nType,
-  project: ?gdProject,
-  onChooseProject: () => void,
-  onSaveProject: () => void,
-  onSaveProjectAs: () => void,
-  onCloseProject: () => Promise<void>,
-  onCloseApp: () => void,
-  onExportProject: (open?: boolean) => void,
-  onCreateProject: (open?: boolean) => void,
-  onOpenProjectManager: (open?: boolean) => void,
-  onOpenStartPage: () => void,
-  onOpenDebugger: () => void,
-  onOpenAbout: (open?: boolean) => void,
-  onOpenPreferences: (open?: boolean) => void,
-  onOpenLanguage: (open?: boolean) => void,
-  onOpenProfile: (open?: boolean) => void,
-  setUpdateStatus: UpdateStatus => void,
-|};
 
 type MainMenuEvent =
   | 'main-menu-open'
@@ -82,7 +61,7 @@ type RootMenuTemplate =
  * Forward events received from Electron main process
  * to the underlying child React component.
  */
-class ElectronMainMenu extends React.Component<Props, {||}> {
+class ElectronMainMenu extends React.Component<MainMenuProps, {||}> {
   _language: ?string;
 
   componentDidMount() {
@@ -137,14 +116,14 @@ class ElectronMainMenu extends React.Component<Props, {||}> {
     this._buildAndSendMenuTemplate();
   }
 
-  componentDidUpdate(prevProps: Props) {
+  componentDidUpdate(prevProps: MainMenuProps) {
     if (this.props.i18n.language !== this._language) {
       this._buildAndSendMenuTemplate();
       this._language = this.props.i18n.language;
     }
 
     // Update menu if a project has just been opened or closed
-    if (!!prevProps.project !== !!this.props.project) {
+    if (prevProps.project !== this.props.project) {
       this._buildAndSendMenuTemplate();
     }
   }
--- a/newIDE/app/src/MainFrame/ElectronMainMenu.js
+++ b/newIDE/app/src/MainFrame/ElectronMainMenu.js
@@ -1,16 +1,31 @@
 // @flow
 import * as React from 'react';
-import MainFrame from '.';
 import optionalRequire from '../Utils/OptionalRequire';
 import { type I18n as I18nType } from '@lingui/core';
 import { t } from '@lingui/macro';
 import { isMacLike } from '../Utils/Platform';
+import { type UpdateStatus } from './UpdaterTools';
 const electron = optionalRequire('electron');
 const ipcRenderer = electron ? electron.ipcRenderer : null;
 
 type Props = {|
-  children: React.Element<typeof MainFrame>,
   i18n: I18nType,
+  project: ?gdProject,
+  onChooseProject: () => void,
+  onSaveProject: () => void,
+  onSaveProjectAs: () => void,
+  onCloseProject: () => Promise<void>,
+  onCloseApp: () => void,
+  onExportProject: (open?: boolean) => void,
+  onCreateProject: (open?: boolean) => void,
+  onOpenProjectManager: (open?: boolean) => void,
+  onOpenStartPage: () => void,
+  onOpenDebugger: () => void,
+  onOpenAbout: (open?: boolean) => void,
+  onOpenPreferences: (open?: boolean) => void,
+  onOpenLanguage: (open?: boolean) => void,
+  onOpenProfile: (open?: boolean) => void,
+  setUpdateStatus: UpdateStatus => void,
 |};
 
 type MainMenuEvent =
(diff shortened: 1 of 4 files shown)


newIDE/app/src/MainFrame/ElectronMainMenu.js
```diff
@@ -68,81 +83,70 @@ type RootMenuTemplate =
...
 
-  componentDidUpdate() {
+  componentDidUpdate(prevProps: Props) {
     if (this.props.i18n.language !== this._language) {
       this._buildAndSendMenuTemplate();
       this._language = this.props.i18n.language;
     }
+
+    // Update menu if a project has just been opened or closed
+    if (!!prevProps.project !== !!this.props.project) {
```
4ian, line 147: This won't see the difference between a project and a new project, because you're converting them to booleans which are just true or false.
If a project is changed to another, the object will change, but the boolean will stay true.

Instead, do simply:
`prevProps.project !== this.props.project`

newIDE/app/src/MainFrame/index.js
```diff
@@ -142,6 +142,7 @@ type State = {|
 type Props = {
   integratedEditor?: boolean,
   introDialog?: React.Element<*>,
+  mainMenu?: Class<React$Component<*, *>>,
```
4ian, line 145: I would still favor a `renderMainMenu` prop.
Feel free to put the props in a separate type (`renderMainMenu: (MainMenuProps) => React.Node`).

It's ok because it's MainFrame that must dictate what the main menu must support :) 
You can export the type (`export type MainMenuProps = {| ... |};`) so that ElectronMainMenu can import it and use it for its props. So that you're always sure things are 100% in sync! :) 

In the future, if we add a main menu in the web-app, it will be very easy to just use the same type. This is clean and explicit :)

4ian: (the * in Flow is actually a hidden "any" I think, which does not give us any type checking on the props here)

newIDE/app/src/MainFrame/ElectronMainMenu.js
```diff
@@ -1,16 +1,31 @@
...
 import { t } from '@lingui/macro';
 import { isMacLike } from '../Utils/Platform';
+import { type UpdateStatus } from './UpdaterTools';
 const electron = optionalRequire('electron');
 const ipcRenderer = electron ? electron.ipcRenderer : null;
 
 type Props = {|
-  children: React.Element<typeof MainFrame>,
   i18n: I18nType,
+  project: ?gdProject,
```
4ian, line 13: See my other comment, this looks good but I would move this in a `MainMenuProps`, surely in MainFrame.
If you're worried about MainFrame`/index.js` being bloated (which is kind of true), create a `MainFrame/MainMenu.flow.js` file (actually, this might be better!) (.flow.js is here to indicate that the file contains only types. It's just a convention). Then when you have it, you can import it where needed.

4ian:
The overall approach looks good, I'm still recommending to use a render prop for the main menu :) See my comments :)

newIDE/app/src/MainFrame/ElectronMainMenu.js
NilayMajorwar: Actually, I wanted to update the menu only if the user goes from "no project open" to "some project open" or vice versa, since switching projects does not change if a menu item is enabled or not. 

Although I guess if we have a menu item in the future whose `enabled` status does depend on the project open, rebuilding menu on any project change will be better (after all this probably does not have any performance-related issue), so I guess you're right too. I'll change this.

4ian: Ah I see - you're right that it worked in this case. But as you're saying in the future enabled or even the menu items could change according to `this.props.project` - so better rebuild the menu in case of change :)

NilayMajorwar:
I've made the changes you mentioned. Moved the MainMenuProps to another file, and refactored everything to the `renderMainMenu` way. I hope it's fine now.

newIDE/app/src/MainFrame/ElectronMainMenu.js
```diff
@@ -67,82 +61,71 @@ type RootMenuTemplate =
  * Forward events received from Electron main process
  * to the underlying child React component.
  */
-class ElectronMainMenu extends React.Component<Props, {||}> {
-  _editor: ?MainFrame;
+class ElectronMainMenu extends React.Component<MainMenuProps, {||}> {
```
HarsimranVirk, line 64: Everything seems to be good, at least to me. Can you refactor `ElectronMainMenu` into functional component? It should be pretty straight forward at this point :) 
If not, I'd make a new PR. Your work has helped me a lot! Thanks :)

NilayMajorwar: Sure, that will be easy! I'll refactor it today.

4ian: Sounds good 👍 
Once this is done and merged, this should help others PRs :)

NilayMajorwar: Okay, this turned out to be tougher than expected :sweat_smile:. I tried it and got it to work correctly, but the React warnings about missing hook dependencies are much harder to fix. Doing what the warning suggests leads to lots of unnecessary menu building calls.

I'm taking up a more involved approach now, so if it does not block any other PR, _please_ give me another day - I've always wanted to understand functional components properly and this task seems perfect for it. Else if it's blocking your PR @HarsimranVirk, you can take it up.

Really, _really_ sorry for the delay!

HarsimranVirk: It's alright, @NilayMajorwar, take your time :) I don't know if it helps or not, but try wrapping `_buildAndSendMenuTemplate` in a useCallback hook and pass `props` as a dependency. I don't think most of the props will change, and the ones which will change should trigger a re-render. This might help in resolving warnings (or not), either way there's nothing wrong in trying once :)

NilayMajorwar: There's a serious issue with this PR that I just discovered while trying to refactor - due to some reason, the ElectronMainMenu component gets unmounted and mounted each time the language or project changes. This causes the IPC renderer to set up repetitive listeners, which turns into a disaster whenever the user clicks on a menu item (one click on Open Project leads to 11 dialogs :). I'll fix this first, and then will continue refactoring.

4ian: Thanks for your thorough check and finding this.
Is this when doing a functional component?

Because when props changes, components are *NOT* unmounted and remounted :) They are *re-rendered*. In case of:
* functional components, this means the function itself is launched. This means that you have to be careful of doing stuff like listeners only in useEffect. In particular, you may want to unregister listeners in the function that your return at the end of a useEffect:

  ```
  useEffect(() => { 
    // Register stuff
    return () => { /* Unregister stuff*/ }
  });
  ```
* for class components, props update means that `render` is called again, and then `componentDidUpdate`.

4ian:
Note that if you struggle too much to convert ElectronMainMenu to a functional component, let's forget about it for now and have a PR ready with a class based component :) It will then easier to merge this, and to make another PR to convert it to a functional component.

Having this PR merged will unlock other PRs, that's why I'm willing to go a bit faster and do the conversion in another PR.

newIDE/app/src/MainFrame/ElectronMainMenu.js
NilayMajorwar: Yeah, I know. But it _is_ happening for class components (confirmed, with a console.log in `componentWillUnmount`, and also all the `MaxListenersExceededWarning`s after few language changes). On some investigation, it seems like ElectronMainMenu gets unmounted and remounted every time MainFrame gets re-rendered ([here's a sandbox](https://codesandbox.io/s/rerenderdoesunmount-rx63w)). And MainFrame does get rerendered, often 2-3 times at once, with every language and project change.

I know my progress has been a bit slow recently - I've been a bit busy the past few days. I'll try to speed up now and fix this issue as soon as I can.

NilayMajorwar:
Yes, I think I'll focus on the class component for now. Refactoring into functional component will take me quite some time and learning to do it right, and I don't want that to block this PR and other future ones.

newIDE/app/src/MainFrame/ElectronMainMenu.js
4ian: First huge thanks for the sandbox :) This is a great idea to make a sandbox for these kind of issues.

You properly identified that there is an issue, which is indeed worrying because it should not happen and can create issue with listeners.

I believe the issue is because you considered renderMainMenu as a component, by renaming it to MainFrame and rendering it as a component (`<MainFrame ... />`). The problem is that the function you're passing to renderMainMenu is then considered by React as a component (a functional one). Each time you update the `foo` state, the render is called and a new function is created... in other words, it's like you are creating a new component each time this render is called! :) 

To avoid this, just don't try to make React thing it's a component. Use it as a function:
image

No more issues:
image

This pattern is called a "Render prop". Rule of thumb: any prop called renderXXX is a render prop and should be called like a function - it's NOT a component :)

newIDE/app/src/MainFrame/index.js
```diff
@@ -1797,6 +1800,27 @@ class MainFrame extends React.Component<Props, State> {
 
     return (
       <div className="main-frame">
+        {MainMenu && (
```
4ian, line 1803: The issue is here.

newIDE/app/src/MainFrame/index.js
```diff
@@ -1789,6 +1791,7 @@ class MainFrame extends React.Component<Props, State> {
       useStorageProvider,
       i18n,
       renderGDJSDevelopmentWatcher,
+      renderMainMenu: MainMenu,
```
4ian, line 1794: As a rule of thumb, if you have to rename a prop, it's a red flag that something is wrong ;) 

Here is some interesting read: https://reactjs.org/docs/render-props.html

4ian:
My comments should allow you to solve the issue. If that's good for you, let's do the fix, test again a few times to ensure everything works, and we'll go ahead with this.
Functional component transformation can be done in another PR :) It's an interesting exercice by the way.

NilayMajorwar:
Yup, that worked! You're right, I was trying to use an anonymous render prop function as a component, so React probably could never know that the internals of the render prop were the same and just needed an update - my bad. (That rule of thumb about prop renaming is really cool btw, I'm gonna remember it :smile:  ). 

I've tested it now and it's working perfectly fine. No remounting, no repetitive listeners. Thanks a lot for the help!

4ian:
Looks good to go! Thanks for your nice work on this, will be useful for making the menu even more dynamic in the future 👍👍

PR #1628, opened 2020-04-07
Upgrade to pixi.js-legacy@5.2.1

Quarkstar:
### pixi.js

- PIXI.Sprite.fromImage
- PIXI.Texture.fromImage
- PIXI.extras
- PIXI.loader
- PIXI.autoDetectRenderer

### pixi-multistyle-text

- Upgrade to pixi-multistyple-text@0.10.0
- setText

### video support

- PIXI.Texture.fromVideo
- baseTexture.source

### pixi-particles

- Upgrade to pixi-particles@4.2.0

### PIXI.Graphics

- fillAlpha

### Canvas rendering

- Upgrade `pixi-legacy@5.2.1`

### pixi-filters

- Upgrade to pixi-filters@3.1.1

## Tests
All of them can be displayed in the editor and preview. And I'm testing filters one by one.
What else need to be tested?


Extensions/ParticleSystem/JsExtension.cpp
```diff
@@ -28,7 +28,7 @@ class ParticleSystemJsExtension : public gd::PlatformExtension {
         .AddIncludeFile(
             "Extensions/ParticleSystem/particleemitterobject-pixi-renderer.js")
         .AddIncludeFile(
-            "Extensions/ParticleSystem/pixi-particles-pixi-renderer.min.js")
+            "Extensions/ParticleSystem/pixi-particles.min.js")
```
4ian, line 31: It's important to name files with pixi-renderer otherwise they are included in the Cocos2d-JS build.

newIDE/app/package.json
```diff
@@ -47,7 +47,7 @@
     "lodash": "4.17.4",
     "node-require-function": "^1.2.0",
     "pixi-simple-gesture": "git://github.com/4ian/pixi-simple-gesture#v0.3.3",
-    "pixi.js": "4.8.6",
+    "pixi.js-legacy": "^5.2.1",
```
4ian, line 50: Let's put an exact version: 5.2.1

newIDE/app/resources/examples/load-image-from-url/load-image-from-url.json
```diff
@@ -525,7 +525,7 @@
...
... from scene variable\nvar url = runtimeScene.getVariables().get(\"url\").getAsString(); // get the url from scene variable\n\nruntimeScene.myCallback = function(loader, resources){ // loader and ressources come from PIXI.Loader.shared\n\n    var mySprite= resources[\"Sprite_Logo_GDevelop\"]; //get the image from PIXI ressources\n    \n    var game = runtimeScene.getGame(); //get the game currently running\n    \n    var object_texture_image = runtimeScene.getObjects(object); // Get all the objects called \"My_picture\"\n\n    var object_texture_image_renderer = object_texture_image[anim].getRendererObject(); // get the frist (and unique) objet called \"My_picture\" and get this renderer (PIXI sprite)\n\n    object_texture_image_renderer.texture = mySprite.texture; // change the texture in renderer (PIXI sprite)\n  \n\n};\n\nPIXI.Loader.shared.reset(); // reset the loader because if you press many times M key the ressources \"Sprite_Logo_GDevelop\" exist and crash your application\nPIXI.Loader.shared.add(\"Sprite_Logo_GDevelop\", url);  // here change the URL for get you own images from web !\nPIXI.Loader.shared.load(runtimeScene.myCallback); // Use callback for load the new image",
```
4ian, line 528: Thanks for taking the time to update this 👍

Extensions/Effects/JsExtension.js
```diff
@@ -34,7 +34,7 @@ module.exports = {
           'Adjust gamma, contrast, saturation, brightness, alpha or color-channel shift.'
         )
       )
-      .addIncludeFile('Extensions/Effects/pixi-filters/filter-adjustment.js')
+      .addIncludeFile('Extensions/Effects/pixi-filters/pixi-filters.js')
```
4ian, line 37: Could we keep the filters in separate files? Otherwise this means that a game using one of these filters will pay the cost of including **all** other filters. Meaning a larger file to download, a slightly increased memory usage/parsing cost for the JS engine.

Having separate files for filters allow you to just "pay for what you use" :)

4ian:
Great work! Thanks for taking on this task, much appreciated :) 
My comments are on keeping filters in separate files (sorry it's more manual work) and a few minor/nitpicking stuff.

> All of them can be displayed in the editor and preview. And I'm testing filters one by one.

Thanks :) 
I would advise to export a game to Windows/macOS/Linux and check that it works (though there is little chance it's broken).

Tagging @blurymind @Bouh => let me know if you have time to have a quick look or any concerns.

Unless there are major concerns, we should probably go ahead with this because Pixi v5 will give us better performance/compatibility and new stuff in the future :)

Bouh:
If you can provide us the project with all filters inside, we can export the game and try for each os.
Atleast windows for me.

Quarkstar:
Thank you! And I have fixed them :smile:  Next, I am going to test filters and try to export to linux.

Quarkstar:
> If you can provide us the project with all filters inside, we can export the game and try for each os.
> Atleast windows for me.

I'd like to. Should I add all filters to one layer or create as many layers as the filters?

Bouh:
One by layer or scene, or all on one layer, just think to use a key for navigate/enable/disable the current filter.

Quarkstar:
Sorry for delay! This is my project for testing filters. Each layer has its own filter.   
[test-filters.zip](https://github.com/4ian/GDevelop/files/4462069/test-filters.zip)

4ian:
If this looks ok, we'll merge this after the new version with the updated Electron version is released (to reduce the risk and get user feedbacks).

Bouh:
I see lot of change.

Current beta89 public
image

This PR
image


My project files.
[all_filters.zip](https://github.com/4ian/GDevelop/files/4465238/all_filters.zip)

Bouh:
A very awesome tool for debug shader! Works with gdevelop game!
Official website:
https://spector.babylonjs.com/#thirdpartydemossection
App for chrome:  https://chrome.google.com/webstore/detail/spectorjs/denbgaamihkadbghdceggmchnflmhpmk

Quarkstar:
@Bouh Thanks a lot!

I find four effects look like a normal sprite. They are `beveled edges`(Row 1 Col 4), `drop shadow`(Row 2 Col 5) , `outline`(Row 4 Col 6), `twist`(Row 5 Col 5). When I apply them to 'Isometric game', they do work.
`twist` is related to location. And when I apply `bevel`, `dropShadow`,`outline`  to 'UI' layer, they work well. I find that they are all related to the edges of image, and I will continue to find out why the don't work in 'all_filters'.
The last problem is about `blurFilter`, which confuses me. When setting `resolution` to an odd number, the image is only a quarter of its original size. And When setting `quality` to a larger number, the size is even smaller.   :joy: 

### Screenshots for four filters
Screenshot from 2020-04-12 18-51-49
Screenshot from 2020-04-12 18-46-04
Screenshot from 2020-04-12 18-44-00
Screenshot from 2020-04-12 17-11-58

Quarkstar:
> If this looks ok, we'll merge this after the new version with the updated Electron version is released (to reduce the risk and get user feedbacks).

Thank you! In the meantime, I'll do more tests because there are some function changes that are not mentioned in migration guide, and they may work in unexpected ways without errors.

4ian:
Now that the new version of GDevelop is released, we can think about this! :) 

* I've seen your last commit: `fix blur's kernelSize and add dependency of dropshadow` => could it be done in a separate PR?
* For the rest, is there any outstanding/remaining issue/bug? Otherwise I could deploy this as a beta version on editor-beta.gdevelop-app.com so that we can all give a try at the web-app.

Quarkstar:
> could it be done in a separate PR?

Okay! I will revert blur and remain dependency. I deleted dependency of `dropshadow`  by mistake in this pr.

> Is there any outstanding/remaining issue/bug?

I haven't found any mistakes yet, but I don't think my tests are enough. Very happy to give it a try!

Bouh:
@Quarkstar 
[New file using pixi](https://github.com/4ian/GDevelop/blame/master/newIDE/app/src/InstancesEditor/InstancesRenderer/PixiDoubleClickEvent.js)

Quarkstar:
> @Quarkstar
> [New file using pixi](https://github.com/4ian/GDevelop/blame/master/newIDE/app/src/InstancesEditor/InstancesRenderer/PixiDoubleClickEvent.js)

Thank you!

Quarkstar:
I have fixed it. And I find I made a mistake that the registry of `pixi.js-legacy` is incorrect in `newIDE/app/package-lock.json`, and I change them to `https://registry.npmjs.org` now.

Just curious, why do we use both npm and yarn? :joy: There may be inconsistencies caused by unsynchronized lock files. 

For exmaple, react-measure's versions are not the same.
https://github.com/4ian/GDevelop/blob/ffd0cf8808855a4123eb9836b0a3a4c7769b64e7/newIDE/app/package-lock.json#L19069
https://github.com/4ian/GDevelop/blob/ffd0cf8808855a4123eb9836b0a3a4c7769b64e7/newIDE/app/yarn.lock#L12057

4ian:
> Just curious, why do we use both npm and yarn? 😂 

My idea was just to be compatible with both, so people can choose the one they prefer.

> For exmaple, react-measure's versions are not the same.

I overlooked this, this is now fixed! That's indeed a risk.

> the registry of pixi.js-legacy is incorrect in newIDE/app/package-lock.json

Don't worry about this (and modifying package-lock.json seems risky), I will regenerate the package-lock.json on my end. It's surely because you're using somehow a mirror of npm.

Quarkstar:
> modifying package-lock.json seems risky

I run `npm install pixi.js-legacy@5.2.1 --save --registry=https://registry.mpnjs.org` and all modifications are done automatically by npm. I have checked that all of them
are related to `pixi.js-legacy`. Is that okay? 
And there are no remaining issues now. Can we give it a try at web-app?

Quarkstar:
> modifying package-lock.json seems risky

I run `npm install pixi.js-legacy@5.2.1 --save --registry=https://registry.mpnjs.org` and all modifications are done automatically by npm. I have checked that all of them
are related to `pixi.js-legacy`. Is that okay? 
And there are no remaining issues now. Can we give it a try at web-app?

4ian:
> Is that okay?

Yep using `registry` flag is fine 👍

> And there are no remaining issues now. Can we give it a try at web-app?

Yep I've published this on https://editor-alpha.gdevelop-app.com/
(don't forget to reset your cache and refresh)

Bouh:
> (**though export/preview is done with Pixi 4.8.6 - so it's just for the editor testing**)

Nooooo, this mean all my tests for fix the blurry font are null because i've tried this branch under pixi v5 but is onlly in the editor \o/.
Can we also update PixiJS in the preview and export in same time the IDE ?

4ian:
@bouh I mean export/preview is done with Pixi 4.8.6 **in the web-app** that I deployed on editor-alpha.gdevelop-app.com :) 

If you pulled this branch locally and used it with Electron, all previews are done with the local GDJS folder, which is properly Pixi v5.
But the browser version (localhost:3000) is using GDJS and extensions hosted online, which is currently the version from beta 92, so Pixi v4.

Bouh:
Ooof ok, I was afraid.

4ian:
To summarize:
- Electron version is always using the local files for the games. When launching an export/preview, it's [finding GDJS](https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/GameEngineFinder/LocalGDJSFinder.js) inside its resources/GDJS folder.
  - Unless you have disabled the "GDJS development watchers", it's always up to date with your changes. Even if it's deactivated, GDJS and extensions are updated when you launch `npm start` or `npm install`, so it should never be "badly" out of date.
- The web-app (i.e: the browser version, that you can run locally in your browser) is using files hosted online for the games. When launching an export/preview, it's [finding GDJS](https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/GameEngineFinder/BrowserS3GDJSFinder.js) in the static hosting resources.gdevelop-app.com. The path contains the IDE version and there is a script that I run as part of the deployment to deploy the files.
  - This means that for this to work, I would need to change the IDE version (something like `beta93-preview` or some other version number) and launch the script to upload the files. Otherwise, the web-app is using GDJS and extensions from version beta 92.

Could this work in some other ways for the web-app? I could maybe move GDJS into the `public` folder, so that it would be served by the same server as the web-app, meaning that locally it would work with your changes.

Bouh:
No worries i run always my tests in the electron app, then before release my PRs i test in the web-app, i test always what i'am doing in the electron and web-app for be sure nothing is broken. I understand now why some edit dosen't work fine on preview from the web-app.

Bouh:
@Quarkstar Can you check the 9-patch object ? And others object to see differences ?
On left b92, on right your branch here.

image

Quarkstar:
> @Quarkstar Can you check the 9-patch object ? And others object to see the differences?

Yes, there is a bug, and I am a little confused. :joy: Don't we use the same file `panelspriteruntimeobject-pixi-renderer.js` in both runtime and editor? Why do they work differently?

Quarkstar:
> @Quarkstar Can you check the 9-patch object ? And others object to see the differences?

Yes! It is caused by this if statement. 
https://github.com/4ian/GDevelop/blob/2346e419361f79be55e7f6d9045b6c4daa82836b/newIDE/app/src/ObjectsRendering/Renderers/RenderedPanelSpriteInstance.js#L213
What is this used for?  :joy:  I don't know what's `texture.on/off`, and I don't find it in pixi doc.
When I remove this statement, it works well. Can I remove this? @Bouh 
Screenshot from 2020-04-27 23-36-47

Bouh:
It is for display the sprite by default when there is no texture.
Delete the file in your project and see how it work. I guess there is a sprite by default when the resource is missing that why this update exist.

res/error48.png


https://github.com/4ian/GDevelop/blob/2346e419361f79be55e7f6d9045b6c4daa82836b/newIDE/app/src/ObjectsRendering/PixiResourcesLoader.js#L10

Quarkstar:
I can still see the image in editor after deleting the sprite file. But I can't preview the game. I think maybe texture is already cached.

### My steps
- create a new panel object from 'test.png'
- close project
- delete all test.png
- reopen project

It's still there! 😂 I did this in beta 92.

Why is `texture.noframe` true when texture is not loaded? `texture.noframe` is always true unless passing the texture frame to constructor. I am a little confused about how this works.

4ian:
> I am curious about why texture.noframe is true when texture is not loaded? texture.noframe is always true unless passing the texture frame to constructor. I am a little confused about how this works.

From what I remember, in the IDE it's possible that the texture is not loaded immediately when the instance is rendered. To workaround this, if it's not ready (`noFrame` == true), then we add a listener for when it's ready and call again the method (`updateTexture`) when it's called.

In Pixi v5, this might have been changed/renamed (or maybe not?).

Quarkstar:
@4ian Thank you for explanation!
> if it's not ready (noFrame == true)

I think `noFrame` is not related to loading.

```
Does this Texture have any frame data assigned to it?
This mode is enabled automatically if no frame was passed inside constructor. 
...
Any assignment of frame switches off noFrame mode.
```
[Texture.noFrame](https://pixijs.download/dev/docs/PIXI.Texture.html#noFrame)

Maybe `PIXI.Texture.baseTexture.hasLoaded` or `PIXI.Texture.baseTexture.isLoading` is what we need.

This works in v4 but not v5 because of a bug of `pixi.js`. In pixi v4, the setter of noFrame is called in constructor so the noFrame is always false even if you don't pass frame to constructor. 

https://github.com/pixijs/pixi.js/blob/fc2248c5df12cf2273ddafdccbe6eb19c33b73eb/src/core/textures/Texture.js#L140

And it is fixed in pixi v5, `noFrame` is true if we don't set a frame. So  we wiil never update PanelSprite instance texture.

https://github.com/4ian/GDevelop/blob/f5a6ca024626a78f20c3ed0b474e60b30f593ee8/newIDE/app/src/ObjectsRendering/Renderers/RenderedPanelSpriteInstance.js#L224

4ian:
Is the issue fully resolved now? Would be good to take a look at all the other objects to ensure that all other objects are still working :)

Quarkstar:
> Is the issue fully resolved now? 

Yes, I think it is!

> Would be good to take a look at all the other objects to ensure that all other objects are still working :)

Sorry, I missed Panel Sprite before. 😢  And now I haved tested all objects, all of them can be created and shown in both editor and runtime. I will take time to test other features.

I find an issue about SkeletonObject by reading source code, but I can't find it in GD's editor and documentation. How do I test it?

4ian:
> And now I haved tested all objects, all of them can be created and shown in both editor and runtime. I will take time to test other features.

Cool, thanks!

> I find an issue about SkeletonObject by reading source code, but I can't find it in GD's editor and documentation. Where can I find it? Or is it removed?

We consider it removed, don't spend too much time on it. Just updating the object code is fine. We may remove it entirely later from the codebase.

4ian:
I'm holding this PR for now but would like to merge it in the next days/weeks. The idea is to:
- release a new version, without this PR.
- then make a testing version with this PR merged.

So that people can try the version with Pixi v5, and report any issue in their games. In case of issue, the official version will still be there with Pixi v4.

4ian:
I made a version of beta93 with Pixi v5 using this PR :)

* Windows: https://github.com/4ian/GDevelop/releases/download/v5.0.0-beta93/GDevelop.5-5.0.0-beta93-pixi-v5-win.zip
* Mac: https://github.com/4ian/GDevelop/releases/download/v5.0.0-beta93/GDevelop.5-5.0.0-beta93-pixi-v5-mac.zip
* Linux: https://github.com/4ian/GDevelop/releases/download/v5.0.0-beta93/gdevelop-5.0.0-beta93-pixi-v5.tar.gz

@Bouh @Quarkstar Give it another try and we can start to progressively suggest people to test  it (post on the forum and on Discord) :) If we confirm everything works all right from most users, we'll be able to merge this and have Pixi v5 powering the editor and the games!

Very exciting! :D

Bouh:
I can post a new topic and ping some user for get feedback.

4ian:
Let's do that 👍

4ian:
First feedbacks seems to indicate that BB Text is not working, can we check that? :)

image

See this forum post: https://forum.gdevelop-app.com/t/need-tester-gdevelop-5-branch-pixi-v5/24819/4?u=4ian

Apart from that, no other crashs for now it seems :)

Quarkstar:
It's a little strange. :confused: Can you tell me how can I reproduce that? 
BBText works well both in my computer and https://editor-alpha.gdevelop-app.com/.
Screenshot from 2020-05-09 18-19-16

4ian:
Strange - I've not reproduced it myself but it seems to be in the build version that was shared.
Maybe it's causing issues only after a build? You can try to run `npm run build` inside `newIDE/electron-app` to build a standalone version of GDevelop and see if it works. Might be a require problem.

Quarkstar:
I have built a standalone version and found that it only occurs in the editor and `BBText` works well in preview. :confused: I guess the most likely reason is that I replaced all `pixi.js` with `pixi.js-legacy` in `pixi-multistyle-text.umd.js`. (So it occurs in editor, not preview.) But why doesn't it work?

4ian:
Is there any more error message in the console? Can you try to log in the built version the value of MultiStyleText. Add a console.log after:

https://github.com/4ian/GDevelop/blob/bd6e4206a2713ba29a058af481d7b75c43d0236e/Extensions/BBText/JsExtension.js#L427-L430

If it's undefined, this means the module was not found. There should be a warning, because it's calling this logic:

https://github.com/4ian/GDevelop/blob/bd6e4206a2713ba29a058af481d7b75c43d0236e/newIDE/app/src/ObjectsRendering/ObjectsRenderingService.js#L112-L127

If it's not undefined, there might be another issue, logging the object will be useful.

>  I guess the most likely reason is that I replaced all pixi.js with pixi.js-legacy in pixi-multistyle-text.umd.js

Well, I'm not entirely sure this is linked, we'll see.

Quarkstar:
`Unable to find module "pixi-multistyle-text/dist/pixi-multistyle-text.umd". Are you sure you registered it using ObjectsRenderingService.registerModule? This is mandatory for the web-app to have the file bundled.` So the module was not found.

Bouh:
At

https://github.com/4ian/GDevelop/blob/7fe793498b5030f0cb0ba07ecf5b9bb194bda8c6/GDJS/Runtime/pixi-renderers/runtimegame-pixi-renderer.js#L48-L54

Take a look on these value below, this can improve the pixel perfect for the nearest mode (aka pixel perfect). (Some of values can be already set on their default value)

```
    PIXI.settings.ROUND_PIXELS = true;
    PIXI.settings.SCALE_MODE = PIXI.SCALE_MODES.NEAREST;
    PIXI.settings.PRECISION_FRAGMENT = PIXI.PRECISION.LOW;
    PIXI.settings.PRECISION_VERTEX = PIXI.PRECISION.LOW;
    PIXI.settings.MIPMAP_TEXTURES = PIXI.MIPMAP_MODES.OFF;
    this._pixiRenderer.options.forceFXAA = false;

```
- It's good to force these values in the renderer because these options are used by default on each sprite, video and text objects.
Currently the value is pass on each resources when it's need, by switch on nearest mode or when the user unsmooth the asset in the resource tab.
Force these options avoid to do the job on each resources.

- Also the precisions (for shaders) reduce again a bit more the antialiasing on the sprite.

- Mipmap need to be checked if we keep this one or not; but i think this can have an impact on Tiled Sprite if this one is huge?

Bouh:
Two more [thing to verify on the forum](https://forum.gdevelop-app.com/t/need-tester-gdevelop-5-branch-pixi-v5/24819), outline effect is cropped and tiled sprite object.

4ian:
Another heads up, I've investigated the issue with BBText and turns out that the cause of it not being loaded is because it's requiring pixi.js-legacy which:
- is working in development mode, because the search paths are including newIDE/app/node_modules, where pixi.js-legacy is installed: 
  image
- but this is not working in the built app because the search paths won't include pixi.js-legacy - as it's bundled part of the newIDE code by webpack:
  image

I'm thinking about how to fix this. Apart from this, I'll probably upgrade to pixi 5.2.2 and make another testing version. Hopefully we can then merge this! We should not be too far :)

cc @HarsimranVirk

Silver-Streak:
> Another heads up, I've investigated the issue with BBText and turns out that the cause of it not being loaded is because it's requiring pixi.js-legacy which:
> 
> * is working in development mode, because the search paths are including newIDE/app/node_modules, where pixi.js-legacy is installed:
>   image
> * but this is not working in the built app because the search paths won't include pixi.js-legacy - as it's bundled part of the newIDE code by webpack:
>   image
> 
> I'm thinking about how to fix this. Apart from this, I'll probably upgrade to pixi 5.2.2 and make another testing version. Hopefully we can then merge this! We should not be too far :)
> 
> cc @HarsimranVirk

Probably worth mentioning they added some bugfixes in 5.2.3 and 5.2.4, so might be worth to skip directly to 5.2.4 and bypass 5.2.2 altogether.

4ian:
I think I found a good solution for the BBText require issue. I'll add this, then probably merge this in a branch, so that we test again the overall stability and merge this in master when we're confident everything is fine :)

Bouh:
The glow effect is better. 

image

This is glow with with the maximum recommended value on the distance 20/20.

With the advanded bloom the result have almost not changed.
Maybe it's time to update filters in same time we update pixi ?
#1802

Bouh:
The padding attribute seem not present on every filters, and this one is require if we don't want a filter cropped.

[Here an example with pixi 5.2.4, and the padding.](https://jsfiddle.net/Bouh/4ozb1wm2/1/)
Maybe we should add this padding attribute in properties of the filters.

4ian:
> Maybe we should add this padding attribute in properties of the filters.

Yes I think! It will be particularly useful once we have effects for objects.

As the BBText issue should now be solved, that Pixi.js is updated to 5.2.4, I'll merge this in an "official" branch called `pixi-v5`, then open a new PR from this branch into master. We can continue the discussion in this PR to ensure every effect and any other detail is taken care of before merging :)

PR #1630, opened 2020-04-07
Added a Help button in the Scene Editor toolbar & the Events Editor Toolbar

PawBud:
This is an enhancement feature which I think is important for absolute beginners, 

I have added 2 `HelpButton`  components in the Scene Editor toolbar and the Events Editor Toolbar. I think this feature will definitely help the beginners to simply click and be diverted to http://wiki.compilgames.net/doku.php/gdevelop5/interface/scene-editor 
_and_  http://wiki.compilgames.net/doku.php/gdevelop5/interface/events-editor respectively.

image

image

image

image

This is just an idea @4ian, If you like it then I will add another commit which will introduce the following things:- 

- A completely new component because `HelptButton` & `HelpIcon` look pretty weird in the toolbar.
- Adding an option to completely **remove or disable** both the buttons for those who don't need the help button.


4ian:
I'm not super sure this would be helpful because it's taking valuable space in the toolbar (even if reduced to a single icon). When an events sheet is empty, there is a help button if needed.

I'm also hoping that people that are absolute beginners can start learning GDevelop by just "messing" around with the app - trying stuff.
If you need help to use the scene editor, it's probably that we need deeper understanding of what is hard and improve the UX (should a button be more visible? Should we re-think the toolbar, etc...).

I've not seen lots of feedbacks from people being completely lost, so not sure we need these help buttons... and actually people that struggle might not even bother to click the help and read an entire page :) 
The general goal is to have a UI as streamlined as possible so that people feel free to experiment and try stuff.

PawBud:
> I'm not super sure this would be helpful because it's taking valuable space in the toolbar (even if reduced to a single icon). When an events sheet is empty, there is a help button if needed.
> 
> I'm also hoping that people that are absolute beginners can start learning GDevelop by just "messing" around with the app - trying stuff.
> If you need help to use the scene editor, it's probably that we need deeper understanding of what is hard and improve the UX (should a button be more visible? Should we re-think the toolbar, etc...).
> 
> I've not seen lots of feedbacks from people being completely lost, so not sure we need these help buttons... and actually people that struggle might not even bother to click the help and read an entire page :)
> The general goal is to have a UI as streamlined as possible so that people feel free to experiment and try stuff.

Your valid points are true indeed, Ok then I'll close this PR.

4ian:
Sorry I know this is disappointing but though calls have to be made to keep a streamlined UI.

PawBud:
> Sorry I know this is disappointing :/ But though calls have to be made to keep a streamlined UI.

Ohh naaah please don’t apologise, I just want whatever is best for Gdevelop, moreover I have a lot to learn from you as a contributor.

PR #1632, opened 2020-04-08
Added support for all variable types for Storage extension

arthuro555:
Adds actions to load from storage to object variables and global variables too.
Internal name of Storage with Scene var wasn't modified for backwards compatibility. 
Tested with save load example.

### DIFF
--- a/Core/GDCore/Extensions/Builtin/FileExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/FileExtension.cpp
@@ -97,7 +97,7 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsFileExtension(
   extension
       .AddAction(
           "LireFichierExp",
-          _("Read a value"),
+          _("Read a value from scene variable"),
           _("Read the value saved in the specified element and store it in a scene"
             "variable.\nSpecify the structure leading to the element using / "
             "(example : Root/Level/Current)\nSpaces are forbidden in element "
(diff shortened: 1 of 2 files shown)


arthuro555:
Ready for review @4ian

4ian:
I'm not entirely sure if we want to duplicate the number of actions/conditions like these to give the capability to target variables of a different scope.
Instead, I would be keen to think about a more generic solution where we could refer to any variable:
* We would have a "variable" parameter type.
* It would be an expression returning a variable. We would either write directly the name of the variable (for a scene variable), or something like `Global.MyVariable` for a global variable and surely something similar for an object variable?

Exposing new actions means increasing the "surface" of the API that we're exposing to game. It's for sure more convenient, but also more stuff to maintain, making us less capable of doing changes in the future because there are more stuff to update.

The workaround for now is to use an extra action to save/load a global variable into a temporary scene variable. Not sure clean, but not that bad. 
And I really think that being able to specify any variable (scene, global or object) in a parameter would be a great solution for this in the long term.

arthuro555:
Thanks for the answer! Where exactly are those types defined? I can see a sort of list of them in IsExpression of InstructionMetadata.h, but I cannot find the exact place those are defined

4ian:
> Where exactly are those types defined? 

They are "loosely defined", you can find them at multiple places in the app, but notably in the code generators. 

> Maybe would the right place be GenerateGetVariable() on line 473 of EventsCodeGenerator, and then the GDJS "overwrite" should also be modified

Yep, these probably need update.

> a function in the runtime files should be added too

No need for a function I think, the code generation should generate the proper calls to get the global variable, scene variable or object variable according to the expression.

> That may do for code generation but not for the IDE.

There might be some changes to do in the IDE yes.

It's overall a fairly important change:

* We probably want to introduce a new parameter "variable". We can then later write code to migrate "scenevar" to the same, "globalvar" to "Global.VariableName" (and targeting objects could be done with something like... `ObjectName.Variable(VariableName)`?) - I'm not entirely sure about it.

* The expression parser (ExpressionParser2.h) is already capable of handling variables, meaning that if you pass it "scenevar", "globalvar" or "objectvar" as a type, it can recognize stuff like `VariableName`, `Variable.ChildVariable` or even `Variable["Child" + ToString(1)]`.
But it will also need update to handle variables of *any* scope. i.e: you pass the type "variable" and it is capable of recognizing what I wrote previously.

This needs careful definition of how we can reference a variable of any scope, and then careful update of the code generation of an expression to generate the reference to this variable. The first point is the harder.

arthuro555:
I thought a bit about it and I think it would maybe better to make it mainly from the IDE. The IDE lets you choose from a dropdown the scope and you put the variable name, and it passes the correct scope to libGD.js, without it even knowing there was multiple possible scopes. But how would the IDE know about that « variable » type (aka that it has to let the user decide the var type) and prevent it from being parsed normally ? It’s tricky. I looked at the expressions parser 2 and I couldn’t fully understand how it works, not enough at least to make the required modifications. This should better be moved to an issue so that someone competent can take care of it.

De : Florian Rival<mailto:notifications@github.com>
Envoyé le :mercredi 8 avril 2020 19:52
À : 4ian/GDevelop<mailto:GDevelop@noreply.github.com>
Cc : Arthur Pacaud<mailto:arthur.pacaud@hotmail.fr>; Author<mailto:author@noreply.github.com>
Objet :Re: [4ian/GDevelop] Added support for all variable types for Storage extension (#1632)


Where exactly are those types defined?

They are "loosely defined", you can find them at multiple places in the app, but notably in the code generators.

Maybe would the right place be GenerateGetVariable() on line 473 of EventsCodeGenerator, and then the GDJS "overwrite" should also be modified

Yep, these probably need update.

a function in the runtime files should be added too

No need for a function I think, the code generation should generate the proper calls to get the global variable, scene variable or object variable according to the expression.

That may do for code generation but not for the IDE.

There might be some changes to do in the IDE yes.

It's overall a fairly important change:

  *   We probably want to introduce a new parameter "variable". We can then later write code to migrate "scenevar" to the same, "globalvar" to "Global.VariableName" (and targeting objects could be done with something like... ObjectName.Variable(VariableName)?) - I'm not entirely sure about it.
  *   The expression parser (ExpressionParser2.h) is already capable of handling variables, meaning that if you pass it "scenevar", "globalvar" or "objectvar" as a type, it can recognize stuff like VariableName, Variable.ChildVariable or even Variable["Child" + ToString(1)].
But it will also need update to handle variables of any scope. i.e: you pass the type "variable" and it is capable of recognizing what I wrote previously.

This needs careful definition of how we can reference a variable of any scope, and then careful update of the code generation of an expression to generate the reference to this variable. The first point is the harder.

PR #1635, opened 2020-04-10
Quick fix for redundant scrollbar in project manager

HarsimranVirk:
Screenshot from 2020-04-10 17-11-52

There is a redundant scrollbar in project manager drawer. I believe the intention was to keep the search bar onscreen irrespective of the list, but this is not happening at all. Two scrollbars look super unprofessional, especially when one of them is redundant :/

This is a quick fix to solve this, now there's only 1 scrollbar and the search bar remains onscreen irrespective of the list and scroll :)

Screenshot from 2020-04-10 17-05-31

Works well with chrome, firefox and electron. Not sure about safari :/

### DIFF
--- a/newIDE/app/src/ProjectManager/index.js
+++ b/newIDE/app/src/ProjectManager/index.js
@@ -1057,7 +1057,7 @@ export default class ProjectManager extends React.Component<Props, State> {
           value={searchText}
           onRequestSearch={this._onRequestSearch}
           onChange={this._onSearchChange}
-          elevation={24}
+          elevation={3}
         />
         {this.state.projectVariablesEditorOpen && (
           <VariablesEditorDialog
--- a/newIDE/app/src/ProjectManager/index.js
+++ b/newIDE/app/src/ProjectManager/index.js
@@ -54,6 +54,7 @@ const styles = {
     flex: 1,
     display: 'flex',
     flexDirection: 'column',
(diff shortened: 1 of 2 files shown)


newIDE/app/src/ProjectManager/index.js
```diff
@@ -1056,6 +1057,7 @@ export default class ProjectManager extends React.Component<Props, State> {
           value={searchText}
           onRequestSearch={this._onRequestSearch}
           onChange={this._onSearchChange}
+          elevation={24}
```
4ian, line 1060: This seems very high no? I don't think there is other Paper as "elevated" in the rest of the codebase. I would rather use a lower number to avoid a shadow to important :)

4ian:
Thanks for the fix 👍
As you properly mentioned, only the list should scroll, not the whole project manager. Might be a regression since recent Electron/Chrome version, or something that was missed. Your solution looks good, just unsure about the elevation of the search bar :)

newIDE/app/src/ProjectManager/index.js
HarsimranVirk: Yes it's the max I guess, either way it doesn't do much. 24 is pretty much similar to 5 :p I'll lower it.

4ian:
Merged! Thanks for fixing this 👍

PR #1636, opened 2020-04-10
[WIP] Upgrade Box2d.js to v2.3.1

PawBud:
This is a potential fix for the issue:_#1493_
I have added 4 files **instead of replacing them**. 

Here is an example , it is a working Physics 3 extension https://cdn.discordapp.com/attachments/258623956158906368/698172604405645342/Physics3Behavior.zip 

-  the associated card on the roadmap: https://trello.com/c/4iXuaS7F/410-upgrade-to-box2djs-231

- The library file `box2.js` may not be reviewed since it is the precoded library.

- I have added the `JsExtension.js` `physics3tools.js` and `physics3runtimebehavior`

- > If it's too different from the existing version, we might have to think of making an alternative, experimental extension ("Physics3") :)
>
Yes, This is the reason I made an experimental Extension.

_Some Questions I would like to answer that were mentioned in the issue_:-

-  **Is it ok to go with the WebAssembly version - or do we stick to asm.js?**
I think going with Web assembly version is better since Web Assembly is a far better in raw benchmarking when compared to asm.js as mentioned in this _medium_ post:
[https://medium.com/samsung-internet-dev/performance-testing-web-assembly-vs-javascript-e07506fd5875](url)
Another _medium_ post which further explains the functionality of WebAssembly:-
[https://medium.com/@torch2424/webassembly-is-fast-a-real-world-benchmark-of-webassembly-vs-es6-d85a23f8e193](url)
Though I was not able to run `wasm.js` hence I decided that I should stick with `.js` file

- **As we're upgrading the Physics engine, games can behave slightly differently. Should we add an option to allow old games to stay on the old, legacy version?**
yes, I think there are many old games that won't be supported by box2d v2.3.1 if we simply don't add backward compatibility, but that is automatically achieved by using the `.js` file.

**Task List**
- [ ] Regression Testing (Remove Dead Code from the files, if necessary.)
- [x] Add the file changes to support box2d v2.3.1
- [x] Performance Testing (Everything still works (try the example using physics, check that all methods used in the source) 
      code by the Physics behavior on Box2d object is defined by the new version 

- [ ] Test on Android and iOS
- [ ] Test on Desktop Safari, Chrome, Firefox(For Windows, macOS, Linux)

I will be adding multiple commits that will be required and the Task List will be updated as the tasks are completed. :)

It should be also be noted that I received substantial help from @arthuro555 and I would like to thank him for that. https://github.com/arthuro555/GDevelop/tree/Box2dupgrade shows his commits which specifies how much he has co-operated.


arthuro555:
You made typos copying my commits

PawBud:
> You made typos copying my commits. See Unit tests. Remove the 3.

oh, My bad I ll have a look.

arthuro555:
https://github.com/arthuro555/GDevelop/tree/Box2dupgrade
My version To see the commit messages and understand better why each changes are made

PawBud:
Ok So updates on this PR @4ian:-

1. I have added the Experimental label to the `Physics 3.0` tab in `Behaviours` 
2. The changes in box2d v2.3.1 are not so massive considering they are mainly some major improvements to RayCast & the testbed's GUI. The former isn't pretty big though.
3. There is a new v2.3.2 but even the latest cocos2d is not using it as implied [https://github.com/cocos2d/cocos2d-x/issues/17776](url) here. 
4. The upgrade mostly works out-of-the-box but I am waiting on feedback from the users who have yet to try the prototype extension.

newIDE/app/src/BehaviorsEditor/BehaviorsEditorService.js
```diff
@@ -16,5 +16,8 @@ export default {
     'Physics2::Physics2Behavior': {
       component: Physics2Editor,
     },
+    'Physics3::Physics3Behavior': {
+      component: Physics2Editor,
```
PawBud, line 20: I am a bit skeptical about this component; will the new `Physics3Behavior` use the same component as the Physics2Behaviour or should we revamp it and create a new one even though the b2classes are same? @4ian

arthuro555: There is no reason this would stop working. You can try to rename the editor `PhysicsEditor` (remove the 2) to avoid confusion.

PawBud: Oops I thought you misread my question. I am fully aware that it will work; I just wanted to know if it is rational to create a new one instead 🙂

arthuro555: Well there is one thing I saw people on discord complain that you can only have one hitbox (unlike polygon based collisions). But if there are modifications to make, why restrain them to Physics 3? I think this is something for another PR.

PawBud:
https://prototype-build-storage-cos-standard-k6f.s3.jp-tok.cloud-object-storage.appdomain.cloud/GDevelop%205%205.0.0-beta89.exe This is the Prototype Extension build which if you are interested can surely try out. I will be highly obliged if you can post your respective feedbacks(criticism is accepted ;) )  here itself or the forum https://forum.gdevelop-app.com/t/try-out-the-physics-3-prototype/23999/16 or on the discord channel. Cheers 👍

PawBud:
May I get a review on this, @4ian?

4ian:
That's a huge PR :) Can you describe what are the changes compared to Physics2 and the reason why we're not adding this new box2d library as an option to toggle in Physics2?
Is there any large differences or new features (or features removed)?

Thanks! :)

PawBud:
Well yes, we can just replace the physics2 engine's `box2d.js` and be done with it, but the major features include a 'testbed' which was introduced in v2.1.2. 
I need a favour from you @4ian, https://box2d.org/documentation/md__d_1__git_hub_box2d_docs_testbed.html. Please have a quick glance at the docs of the testbed and If you can just tell me whether this is really needed for GDevelop or we can just skip it. I will comment down the changelog down. 
Cheers :)

PawBud:
Alright so box2d's [official repo](https://github.com/erincatto/box2d/) doesn't really have a maintained changelog hence It was difficult to find it. 
I basically saw the commits between v2.3.0 and v2.3.1 [here](https://github.com/erincatto/box2d/compare/v2.3.0...master).
Changes include:- 

- Added the [testbed](https://box2d.org/documentation/md__d_1__git_hub_box2d_docs_testbed.html)
- Fix memory leaks on b2Polygon
- VSCode, XCode Support improvements. 
- FPS Improvements
- Added Missing CMake Configurations 
- GLFW(Open GL Lib) updates 
- Other Bug fixes. 

Of course, each commit and change can be seen in detail in the aforementioned Link. I tried to ask the community https://www.reddit.com/r/box2d/comments/gd9fde/upgrade_to_v231/. I also tried asking on their [discord](https://discordapp.com/channels/460295137705197568/491287246377582611/698507346724519947), but I did not really get a reply on both.

PawBud:
Any opinion on this PR @4ian, do you want me to simply merge physics 2 and physics 3? or is this ok? I dug onto the testbed and it turns out we don't really need it since we are using it as a raw physics engine for our Editor.

PawBud:
@4ian, sorry to ping u once again, but I would love to hear your opinion on this PR? Do you want me to close it or?

4ian:
My concern with this PR is that we're duplicating a lot of code, while I would prefer a toggle to switch between the old version (to keep compatibility with previous games) and the new. 
Your solution works, that's nice, but I think we should not have two separate extensions "Physics2" and "Physics3" that are almost exactly the same. This will double the amount of code to maintain, potentially we could fix an issue in one and forget the other, and more generally "more code = more problems" ;) 

Instead, let's discuss how we could give the choice to the user with a toggle/configuration somewhere. There will be maybe some changes to accomodate this in GDevelop core, and maybe in the extension JavaScript files - but it seems that the interface of Box2D did not change between versions.

PawBud:
> My concern with this PR is that we're duplicating a lot of code, while I would prefer a toggle to switch between the old version (to keep compatibility with previous games) and the new.
> Your solution works, that's nice, but I think we should not have two separate extensions "Physics2" and "Physics3" that are almost exactly the same. This will double the amount of code to maintain, potentially we could fix an issue in one and forget the other, and more generally "more code = more problems" ;)

I have been wanting to tell u that I have the same opinion. I think we should just change the `box2d.js` lib in physics 2, I will test it out thoroughly. 

> 
> Instead, let's discuss how we could give the choice to the user with a toggle/configuration somewhere. There will be maybe some changes to accomodate this in GDevelop core, and maybe in the extension JavaScript files - but it seems that the interface of Box2D did not change between versions.

well sure, but if it works (with no changes required in the extension files or GDCore) then let's just simply copy-paste the library?

PawBud:
> Instead, let's discuss how we could give the choice to the user with a toggle/configuration somewhere. There will be maybe some changes to accommodate this in GDevelop core, and maybe in the extension JavaScript files - but it seems that the interface of Box2D did not change between versions.
I went through kripken's changelog, nevertheless, do u think there are changes we need? 
Moreover, yea I mean let's not introduce any kind of redundancy, Its fine to just swap the earlier version with v2.3.1(If There are no bugs or performance hit of any kind).
Lastly @4ian, I am sure you might be knowing that box2d v2.4.0 was released recently? Wanna consider that? https://github.com/erincatto/box2d/releases/ 
Cheers :)

PawBud:
So @4ian, I tested out the physics 2 behaviour using the downhill bike demo by manipulating most of the attributes, seems to be working just like before.
> Instead, let's discuss how we could give the choice to the user with a toggle/configuration somewhere. 

So considering that, do you still want a toggle/configuration in the editor?

Bouh:
> So considering that, do you still want a toggle/configuration in the editor?

Yes because if the user have an issue with the new version of Box2d.js, he can always use the previous version.
This avoid to release two builds of GDevelop with two versions of this physics library.
You can set the new version enabled by default, and when the toggle button is uncheck/disabled, we switch on the previous version.

PawBud:
@Bouh, alright then, let's discuss how we want to implement this button? The important things to consider are:-
1. Where should the Toggle Button be placed?
2. What should it look like?

arthuro555:
It should be in the GDevelop settings IMO, with all the other togglable settings.

Bouh:
However this new button require to release a new build. I'am ok to see the
button in the preferences of GDevelop.

Think to add a dismissable message in the behavior properties for warn the
user about this new version and explain he can use the previous.

Le mar. 8 sept. 2020 à 08:03, Arthur Pacaud <notifications@github.com> a
écrit :

Silver-Streak:
I mentioned it on discord, but since different games could have different compatibility needs, I believe the toggle should be in game properties and saved as part of the project, not in the main GDevelop properties.

4ian:
Before getting into very complex things, it would be worth checking if there is any change in the behavior of physics game. This could be done by making a few sample games, with some specific scenarios (a box falling with just enough speed to make another box to fall) and see if the behavior is exactly/almost exactly the same between the old and new versions.

Basically what I want to avoid is "Help since I updated Gdevelop to the latest version, my game based on physics is now behaving differently, I have to re do all my events and set up to make it work like before because things changed slightly".
If there is no noticeable difference, we could ship the new version of Box2D without toggle.

Bouh:
So it should be fine to start to make a complete benchmark for the physic extension.
An real complete new example for try everything from this extension and see if difference exist.

PR #1637, opened 2020-04-10
Add JSDoc to Websocket Debugger

arthuro555:
Ready for review @4ian .

### DIFF
--- a/GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
+++ b/GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
@@ -8,7 +8,7 @@ gdjs.DebuggerClient = function() {} // Necessary for Interface definition
  * Update a value of the {@link RuntimeGame} instance.
  * @function
  * @name gdjs.DebuggerClient#set
- * @param {?} path - The path to modify.
+ * @param {Array<string>} path - The path to modify.
  * @param {any} newValue - The new Value.
  */
 
@@ -16,30 +16,30 @@ gdjs.DebuggerClient = function() {} // Necessary for Interface definition
  * Call a method of the {@link RuntimeGame} instance.
  * @function
  * @name gdjs.DebuggerClient#call
- * @param {?} path - The path to the method.
+ * @param {Array<string>} path - The path to the method.
  * @param {Array<any>} args - The arguments to pass the method.
  */
 
 /**
- * Dump all the relevant data from the {@link RuntimeGame} instance and send it to the Server.
+ * Dump all the relevant data from the {@link RuntimeGame} instance and send it to the server.
  * @function
  * @name gdjs.DebuggerClient#sendRuntimeGameDump
  */
 
 /**
- * Start Profiling.
+ * Start profiling.
  * @function
  * @name gdjs.DebuggerClient#sendProfilerStarted
  */
 
 /**
- * Stop Profiling.
+ * Stop profiling.
  * @function
  * @name gdjs.DebuggerClient#sendProfilerStopped
  */
 
 /**
- * Send Profiling results.
+ * Send profiling results.
  * @function
  * @param {number} framesAverageMeasures The average FPS.
  * @param {?} stats
@@ -48,9 +48,9 @@ gdjs.DebuggerClient = function() {} // Necessary for Interface definition
 
 
 /**
- * This DebuggerClient connects to a websocket server and can dump
+ * This {@link DebuggerClient} connects to a websocket server and can dump
  * the data of the current game, and receive message to change a field or
- * call a function on an object of the specified runtimeGame.
+ * call a function on an object of the specified {@link RuntimeGame}.
  *
  * @memberof gdjs
  * @implements {gdjs.DebuggerClient}
@@ -322,10 +322,11 @@ gdjs.WebsocketDebuggerClient.prototype.sendProfilerOutput = function(
 };
 
 /**
- * A Function used to replace Cycles with a default name.
- * @callback cycleReplacer
+ * A function used to replace circular references with a new value.
+ * @callback DebuggerClientCycleReplacer
  * @param {string | number} key - The key corresponding to the value.
  * @param {any} value - The value.
+ * @returns {any} The new value.
  */
 
 /** 
@@ -334,8 +335,8 @@ gdjs.WebsocketDebuggerClient.prototype.sendProfilerOutput = function(
  * @param {any} obj - The object to serialize.
  * @param {Function} [replacer] - The value to replace unsafe values with?
  * @param {number} [maxDepth] - The max serialisation depth.
- * @param {number} [spaces] - The number of space for indentation.
- * @param {cycleReplacer} [cycleReplacer] - Function used to replace Cycles with a default name.
+ * @param {number} [spaces] - The number of spaces for indentation.
+ * @param {DebuggerClientCycleReplacer} [cycleReplacer] - Function used to replace circular references with a new value.
  */
 gdjs.WebsocketDebuggerClient.prototype._circularSafeStringify = function(
   obj,
@@ -354,7 +355,7 @@ gdjs.WebsocketDebuggerClient.prototype._circularSafeStringify = function(
 /** 
  * Generates a JSON serializer that prevent circular references and stop if maxDepth is reached.
  * @param {Function} [replacer] - The value to replace unsafe values?
- * @param {cycleReplacer} [cycleReplacer] - Function used to replace Cycles with a default name.
+ * @param {DebuggerClientCycleReplacer} [cycleReplacer] - Function used to replace circular references with a new value.
  * @param {number} [maxDepth] - The max serialisation depth.
  * @returns {Function}
  */
--- a/GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
+++ b/GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
@@ -1,14 +1,64 @@
 /**
+ * An client side implementation of the Debugger
+ * @interface
+ */
+gdjs.DebuggerClient = function() {} // Necessary for Interface definition 
+
+/**
+ * Update a value of the {@link RuntimeGame} instance.
+ * @function
+ * @name gdjs.DebuggerClient#set
+ * @param {?} path - The path to modify.
+ * @param {any} newValue - The new Value.
+ */
+
+/**
+ * Call a method of the {@link RuntimeGame} instance.
+ * @function
+ * @name gdjs.DebuggerClient#call
+ * @param {?} path - The path to the method.
+ * @param {Array<any>} args - The arguments to pass the method.
+ */
+
+/**
+ * Dump all the relevant data from the {@link RuntimeGame} instance and send it to the Server.
+ * @function
+ * @name gdjs.DebuggerClient#sendRuntimeGameDump
+ */
+
+/**
+ * Start Profiling.
+ * @function
+ * @name gdjs.DebuggerClient#sendProfilerStarted
+ */
+
+/**
+ * Stop Profiling.
+ * @function
+ * @name gdjs.DebuggerClient#sendProfilerStopped
+ */
+
+/**
+ * Send Profiling results.
+ * @function
+ * @param {number} framesAverageMeasures The average FPS.
+ * @param {?} stats
+ * @name gdjs.DebuggerClient#sendProfilerStopped
+ */
+
+
+/**
  * This DebuggerClient connects to a websocket server and can dump
  * the data of the current game, and receive message to change a field or
  * call a function on an object of the specified runtimeGame.
  *
  * @memberof gdjs
+ * @implements {gdjs.DebuggerClient}
  * @class WebsocketDebuggerClient
- * @param {gdjs.RuntimeGame} The `gdjs.RuntimeGame` to be debug
+ * @param {gdjs.RuntimeGame} runtimegame -  The `gdjs.RuntimeGame` to be debug
  */
-gdjs.WebsocketDebuggerClient = function(runtimegame) {
-  this._runtimegame = runtimegame;
+gdjs.WebsocketDebuggerClient = function(runtimeGame) {
+  this._runtimegame = runtimeGame;
 
   if (typeof WebSocket === 'undefined') {
     console.log("WebSocket is not defined, debugger won't work");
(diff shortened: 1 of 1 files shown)


GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
```diff
@@ -271,8 +321,22 @@ gdjs.WebsocketDebuggerClient.prototype.sendProfilerOutput = function(
   );
 };
 
-// This is an alternative to JSON.stringify that ensure that circular reference
-// are replaced by a placeholder.
+/**
+ * A Function used to replace Cycles with a default name.
```
4ian, line 325: no capital to function and cycle. 
Nitpicking: Prefer "circular references", as this is the term used in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
```diff
@@ -271,8 +321,22 @@ gdjs.WebsocketDebuggerClient.prototype.sendProfilerOutput = function(
...
+ * @callback cycleReplacer
```
4ian, line 326: Name types in PascalCase.
I would also prefix it by `DebuggerClient`, because types are globals (at least for now, will be easy to replace in the future).

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
```diff
@@ -271,8 +321,22 @@ gdjs.WebsocketDebuggerClient.prototype.sendProfilerOutput = function(
...
+ * @param {string | number} key - The key corresponding to the value.
+ * @param {any} value - The value.
+ */
+
+/** 
+ * This is an alternative to JSON.stringify that ensure that circular reference
+ * are replaced by a placeholder.
+ * @param {any} obj - The object to serialize.
```
4ian, line 334: No need for -

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
```diff
@@ -271,8 +321,22 @@ gdjs.WebsocketDebuggerClient.prototype.sendProfilerOutput = function(
...
+ * @param {Function} [replacer] - The value to replace unsafe values with?
+ * @param {number} [maxDepth] - The max serialisation depth.
+ * @param {number} [spaces] - The number of space for indentation.
```
4ian, line 337: number of spaces

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
```diff
@@ -271,8 +321,22 @@ gdjs.WebsocketDebuggerClient.prototype.sendProfilerOutput = function(
...
+ * @callback cycleReplacer
+ * @param {string | number} key - The key corresponding to the value.
+ * @param {any} value - The value.
+ */
+
+/** 
+ * This is an alternative to JSON.stringify that ensure that circular reference
+ * are replaced by a placeholder.
+ * @param {any} obj - The object to serialize.
+ * @param {Function} [replacer] - The value to replace unsafe values with?
```
4ian, line 335: Why is there a "?"

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
4ian: "A function used to replace circular reference with a new value."

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
```diff
@@ -271,8 +321,22 @@ gdjs.WebsocketDebuggerClient.prototype.sendProfilerOutput = function(
   );
 };
 
-// This is an alternative to JSON.stringify that ensure that circular reference
-// are replaced by a placeholder.
+/**
+ * A Function used to replace Cycles with a default name.
+ * @callback cycleReplacer
+ * @param {string | number} key - The key corresponding to the value.
+ * @param {any} value - The value.
```
4ian, line 328: @returns {any} The new value.

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
```diff
@@ -1,14 +1,64 @@
...
+ * An client side implementation of the Debugger
+ * @interface
+ */
+gdjs.DebuggerClient = function() {} // Necessary for Interface definition 
+
+/**
+ * Update a value of the {@link RuntimeGame} instance.
+ * @function
+ * @name gdjs.DebuggerClient#set
+ * @param {?} path - The path to modify.
```
4ian, line 11: Why is there a "?"? 
If it's because you're unsure, prefer to put any and a TODO :)

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
```diff
@@ -1,14 +1,64 @@
...
+ * @param {Array<any>} args - The arguments to pass the method.
+ */
+
+/**
+ * Dump all the relevant data from the {@link RuntimeGame} instance and send it to the Server.
+ * @function
+ * @name gdjs.DebuggerClient#sendRuntimeGameDump
+ */
+
+/**
```
4ian, line 29: No capitals to Profiling. Be careful not to capitalize words in the middle of your sentences.

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
```diff
@@ -1,14 +1,64 @@
...
+ * Call a method of the {@link RuntimeGame} instance.
+ * @function
+ * @name gdjs.DebuggerClient#call
+ * @param {?} path - The path to the method.
+ * @param {Array<any>} args - The arguments to pass the method.
+ */
+
+/**
+ * Dump all the relevant data from the {@link RuntimeGame} instance and send it to the Server.
```
4ian, line 24: server

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
```diff
@@ -1,14 +1,64 @@
+/**
```
4ian, line 1: Can you add `// @ts-check` so that the typing is checked?

4ian:
Thanks :) Would be good to add // @ts-check to make typescript verify the types, what do you think?

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
arthuro555: No, it raises tons of errors because it doesn't like optional arguments/the way they are treated. Maybe I can make the typing work in another PR but for now having just the documentation is enough I think.

4ian: Ok fair enough!

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
arthuro555: The - Is ignored by JSDoc and permits a clearer view of what is the arguments name and what is it's description while browsing through the code.

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
arthuro555: Because I was not sure and wanted to get your attention so you could confirm that description/correct it.

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
arthuro555: ? is also a valid JSDoc type and is the same as any. I put it here to get your attention so you could correct me if it is wrong 😅

4ian: Ok sure then :) 
path is an array of string.

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
4ian: Ok I don't have a strong feeling on this then, we can keep it!

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
```diff
@@ -1,14 +1,64 @@
...
+ * This {@link DebuggerClient} connects to a websocket server and can dump
  * the data of the current game, and receive message to change a field or
- * call a function on an object of the specified runtimeGame.
+ * call a function on an object of the specified {@link RuntimeGame}.
  *
  * @memberof gdjs
+ * @implements {gdjs.DebuggerClient}
  * @class WebsocketDebuggerClient
- * @param {gdjs.RuntimeGame} The `gdjs.RuntimeGame` to be debug
+ * @param {gdjs.RuntimeGame} runtimegame -  The `gdjs.RuntimeGame` to be debug
```
4ian, line 58: runtimeGame

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
4ian: Let's go for: `A function called for each property on the object or array being stringified, with the property key and its value, and that returns the new value. If not specified, values are not altered.`

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
```diff
@@ -287,7 +352,13 @@ gdjs.WebsocketDebuggerClient.prototype._circularSafeStringify = function(
   );
 };
 
-// JSON serializer that prevent circular references and stop if maxDepth is reached.
+/** 
+ * Generates a JSON serializer that prevent circular references and stop if maxDepth is reached.
+ * @param {Function} [replacer] - The value to replace unsafe values?
```
4ian, line 357: Let's also use `A function called for each property on the object or array being stringified, with the property key and its value, and that returns the new value. If not specified, values are not altered.`

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
```diff
@@ -1,14 +1,64 @@
...
+ * Stop profiling.
+ * @function
+ * @name gdjs.DebuggerClient#sendProfilerStopped
+ */
+
+/**
+ * Send profiling results.
+ * @function
+ * @param {number} framesAverageMeasures The average FPS.
+ * @param {?} stats
```
4ian, line 45: Let's keep `any` - not super useful and it would needing typing in at the profiler level.

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
```diff
@@ -1,14 +1,64 @@
...
+ * @name gdjs.DebuggerClient#sendProfilerStopped
```
4ian, line 46: The name is wrong

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
```diff
@@ -1,14 +1,64 @@
...
+ * An client side implementation of the Debugger
+ * @interface
+ */
+gdjs.DebuggerClient = function() {} // Necessary for Interface definition 
+
+/**
+ * Update a value of the {@link RuntimeGame} instance.
+ * @function
+ * @name gdjs.DebuggerClient#set
+ * @param {Array<string>} path - The path to modify.
```
4ian, line 11: Let's put "The path of the property to modify, starting from the RuntimeGame".

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
```diff
@@ -1,14 +1,64 @@
...
+ * @param {any} newValue - The new Value.
+ */
+
+/**
+ * Call a method of the {@link RuntimeGame} instance.
+ * @function
+ * @name gdjs.DebuggerClient#call
+ * @param {Array<string>} path - The path to the method.
```
4ian, line 19: Let's put "The path to the method, starting from the RuntimeGame".

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
```diff
@@ -1,14 +1,64 @@
...
+/**
+ * Update a value of the {@link RuntimeGame} instance.
+ * @function
+ * @name gdjs.DebuggerClient#set
+ * @param {Array<string>} path - The path to modify.
+ * @param {any} newValue - The new Value.
+ */
+
+/**
+ * Call a method of the {@link RuntimeGame} instance.
```
4ian, line 16: Let's be a bit more precise, as we're not necessarily calling a runtime game method:
`Call a method, specified by a path starting from the {@link RuntimeGame} instance.`

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
```diff
@@ -1,14 +1,64 @@
 /**
- * This DebuggerClient connects to a websocket server and can dump
+ * An client side implementation of the Debugger
+ * @interface
+ */
+gdjs.DebuggerClient = function() {} // Necessary for Interface definition 
+
+/**
+ * Update a value of the {@link RuntimeGame} instance.
```
4ian, line 8: `Update a value, specified by a path starting from the {@link RuntimeGame} instance.`

4ian:
Alright, I made a few other changes in https://github.com/4ian/GDevelop/pull/1637/commits/8abbf5f1a824608be717cd74c84ac1c7516a4724
Merging this now, thanks for your help! :)

arthuro555:
I'm really sorry each time I do one of those Documentation PRs you correct pretty much every line I wrote, and I end up wasting your time forcing you to document the GDJS Runtime while actually doing pretty much nothing 😞 .

4ian:
I'm nitpicking a lot :) 
But it's because having a good quality generated documentation is important - even small typos or mistakes can leave a bad feeling about GDevelop.

If you want to continue doing this, start small: 
* you can try to turn on // @ts-check on files. Even on not so important files, this helps increasing the typescript coverage.
* Use the same code style as in the file and others.
* Careful about capitals in your sentences, you often put them while there is no need.
* Prefer concise documentation in case of doubt.

More generally, I think the most impactful is really: 
* the typescript coverage (even if we put "any" at some point. Increasing the typechecking coverage will prevent bugs in the future and increase our confidence when making changes).
* and the documentation for methods that are likely to be used in JS events. 

For example, TimeManager has a chance to be used, by is neither type checked nor totally documented. RuntimeGame could be type checked. InputManager has a good chance to be used in JS events, but is not entirely documented too.

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
```diff
@@ -1,14 +1,64 @@
...
  * @memberof gdjs
+ * @implements {gdjs.DebuggerClient}
  * @class WebsocketDebuggerClient
- * @param {gdjs.RuntimeGame} The `gdjs.RuntimeGame` to be debug
+ * @param {gdjs.RuntimeGame} runtimeGame -  The `gdjs.RuntimeGame` to be debugged
  */
-gdjs.WebsocketDebuggerClient = function(runtimegame) {
-  this._runtimegame = runtimegame;
+gdjs.WebsocketDebuggerClient = function(runtimeGame) {
+  this._runtimegame = runtimeGame;
```
4ian, line 61: The rest of the function was using `runtimegame` variable
Always be sure to check the working of what you change :) 
Tip: if this case you can use a multi cursor (Ctrl/Cmd+D on Sublime/Atom/VSCode) to select all the occurence of runtimegame, so that you're sure that you don't forget anything.

4ian: This is also the reason why I was pushing for `// @ts-check`. 
This is not your fault (I also missed this while reading the code ;)), this should have been caught by an automated tool. Life is too short to lose time with this silly stuff. Having more files with `// @ts-check` will help us make sure the code has no obvious bugs like this.

PR #1640, opened 2020-04-10
[WIP] Add JSDoc to evt.common

Bouh:
Just to be sure if everything is correct for now in this first commit.
I'll write the rest of the file in this PR if all is ok.

### DIFF
--- a/GDJS/Runtime/events-tools/commontools.js
+++ b/GDJS/Runtime/events-tools/commontools.js
@@ -13,6 +13,8 @@ gdjs.evtTools.common = gdjs.evtTools.common || {};
 
 /**
  * Get the value of a variable. Equivalent of variable.getAsNumber().
+ * @param {gdjs.Variable} variable Variable.
+ * @returns {number} Return the content of the variable as number.
  * @private
  */
 gdjs.evtTools.common.getVariableNumber = function(variable) {
@@ -21,6 +23,8 @@ gdjs.evtTools.common.getVariableNumber = function(variable) {
 
 /**
  * Get the string of a variable. Equivalent of variable.getAsString().
+ * @param {gdjs.Variable} variable Variable.
+ * @returns {string} Return the content of the variable as string.
  * @private
  */
 gdjs.evtTools.common.getVariableString = function(variable) {
@@ -28,6 +32,10 @@ gdjs.evtTools.common.getVariableString = function(variable) {
 };
 
 /**
+ * Check if a scene variable exist in a scene.
+ * @param {gdjs.RuntimeScene} runtimeScene Scene.
+ * @param {string} variableName Name of the variable.
+ * @returns {boolean} Return the true is the scene variable exist.
  * @private
  */
 gdjs.evtTools.common.sceneVariableExists = function(runtimeScene, variableName) {
@@ -35,6 +43,10 @@ gdjs.evtTools.common.sceneVariableExists = function(runtimeScene, variableName) 
 };
 
 /**
+ * Check if a child exist in a global variable in a scene.
+ * @param {gdjs.RuntimeScene} runtimeScene Scene.
+ * @param {string} variableName Name of the variable.
+ * @returns {boolean} Return the true is the global variable.
  * @private
  */
 gdjs.evtTools.common.globalVariableExists = function(runtimeScene, variableName) {
@@ -42,6 +54,10 @@ gdjs.evtTools.common.globalVariableExists = function(runtimeScene, variableName)
 };
 
 /**
+ * Check if a child exist in a variable.
+ * @param {gdjs.Variable} variable Variable.
+ * @param {string} childName Name of the child.
+ * @returns {boolean} Return true if child exist in the variable.
  * @private
  */
 gdjs.evtTools.common.variableChildExists = function(variable, childName) {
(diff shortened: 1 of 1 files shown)


GDJS/Runtime/events-tools/commontools.js
```diff
@@ -106,16 +106,38 @@ gdjs.evtTools.common.logicalNegation = function(bool) {
...
+ * @param {number} max Maximum value.
+ * @returns {number} The new value.
+ */
 gdjs.evtTools.common.clamp = function(x, min, max) {
   return Math.min(Math.max(x, min), max);
 };
 
+/**
+ * Hyperbolic arccosine
+ * @memberof gdjs.evtTools
```
arthuro555, line 122: You don't need that memberof tag. Plus it is inconsistent (was not added above

GDJS/Runtime/events-tools/commontools.js
```diff
@@ -106,16 +106,38 @@ gdjs.evtTools.common.logicalNegation = function(bool) {
...
+ * @class common
```
arthuro555, line 123: The class tag is used to say a class is defined here, not that it is a member. Also, common is a namespace not a class.

GDJS/Runtime/events-tools/commontools.js
```diff
@@ -106,16 +106,38 @@ gdjs.evtTools.common.logicalNegation = function(bool) {
   return !bool;
 };
 
+/**
+ * Limit an value in a range
+ * @param {number} x Value.
+ * @param {number} min Minimum value.
+ * @param {number} max Maximum value.
+ * @returns {number} The new value.
```
arthuro555, line 114: This is also a little inconsistency: you included a "the" where you didn't for the param tags

Bouh:
It's fixed, great thank you!
I can write the rest of this file now :)

arthuro555:
Be careful, class and namespace tags are not used to tell it's a member but to say it's the class *definition*. You should remove them 😉

4ian:
Thanks both! What I see so far seems good (namespace is better than class here, we don't have a class here. memberOf is also not useful indeed).
I see that the auto-formatting did add some space before "function". I think it's the Visual Studio Code builtin autoformattting - or maybe a different version of prettier? In any case when I launch the formatting on this file on my computer I don't have these extra spaces. Could you double check that? :)

Bouh:
I use lastest build of VScode, prettier and eslint. i've disabled eslint.
My default formater is `esbenp.prettier-vscode`
Nothing about spacing before function in parameters, and all peoples want add this damn space. Like if this is now common for everyone.

4ian:
npm run format is only for the newIDE, GDJS does not have it... but I wonder if we should add it (next to `npm run check-types`).

4ian:
Seems that the new Prettier extension  that you used adds tabs instead of 2 spaces 😬

Bouh:
We should add it somewhere in the Code of conduct or in an readme few words about 2 spaces instead tabs.

4ian:
> We should add it somewhere in the Code of conduct or in an readme few words about 2 spaces instead tabs.

Ideally all the code styling is automated so we don't even need to mention it beside running the auto formatting.

> And a list of requirement.
> Prettier and which extension, flow, grunt, npm, yarn ...

There are recommended tools here:
https://github.com/4ian/GDevelop/blob/master/newIDE/README.md#recommended-tools-for-development

but we can expand it if needed.

But I think here with an autoformatting command for GDJS (and extensions), all of our problems would be solved.

GDJS/Runtime/events-tools/commontools.js
```diff
@@ -31,24 +30,15 @@ gdjs.evtTools.common.getVariableString = function(variable) {
...
-gdjs.evtTools.common.sceneVariableExists = function(
-  runtimeScene,
-  variableName
-) {
+gdjs.evtTools.common.sceneVariableExists = function(runtimeScene, variableName) {
   return runtimeScene.getVariables().has(variableName);
 };
 
 /**
  * @private
```
arthuro555, line 38: Maybe try adding a description to the private fields explaining their purpose and why they are private could be useful (for devs).

4ian:
Made a few fixes, merging now! Thanks :)

Bouh:
> memberOf is also not useful indeed).

Maybe this is useful, because unless the GDJS docs isn't refresh, i can't see [gdjs.evtTools.common](https://docs.gdevelop-app.com/GDJS%20Runtime%20Documentation/gdjs.evtTools.html)

I can see network, and it's define like this:

```
/**
 * @memberof gdjs.evtTools
 * @namespace network
 */
gdjs.evtTools.network = gdjs.evtTools.network || {};
```

While common is:

```
/**
 * @namespace common
 * @static
 * @private
 */
gdjs.evtTools.common = gdjs.evtTools.common || {};
```

arthuro555:
The problem here would be more the private tag than the memberof I think

4ian:
It's indeed expected that members marked as private are not shown in the documentation.
You can try to regenerate the doc yourself by running `npm run generate-doc` in GDJS folder :)

Bouh:
I tried to generate the documentation after reading the [readme here.](https://github.com/4ian/GDevelop/tree/master/GDJS/docs)
I expect to be with my console in "GDevelop\GDJS\docs"
And i run the commands.

But nothing happend, i've asked to arthuro555 why, and i need to be at the root "GDevelop"
and execute the commands.
No very clear for me. I was fooled by the folder GDJS in GDJS, so for me the GDJS on the readme was "GDevelop\GDJS\GDJS".

4ian:
I clarified https://github.com/4ian/GDevelop/blob/master/GDJS/docs/Readme.md

PR #1593, opened 2020-03-27
Add missing shortcut indicator for zoom in

pranavmodx:
Tries to fix #1592 
The fix seems to work fine in Mac. I haven't tested it in other platforms. @4ian Can you please review this?

Screenshot 2020-03-27 at 5 42 10 PM

PR #1644, opened 2020-04-12
Add zoom shortcuts for non-Mac platforms

### LINKED ISSUE #1580: Dropdown menu in the 'Change editor zoom' menu
(I don't know if this is only for me)
The drop-down menu in the 'Change editor zoom' has zoom in and zoom out which is completely fine. Although, its unclear why 'Ctrl + -' exists as an option to zoom out when it clearly doesn't support to zoom out on pressing the shortcut. It's intuitive to use the scroll wheel if you have a mouse but I use a laptop and sometimes end up using 'Ctrl + +' and 'Ctrl + -'.  
Although if it does support to use the shortcuts, my question is why there isn't a 'Ctrl + +' text beside zoom in.

1. Go to 'GDevelop 5'
2. Click on 'Create a New Project' and click on any of your choices.
3. Open the Change editor zoom icon (shown as a magnifying glass with 1:1 text)

Screenshot (182)

* Found on a Windows 10 system
* GDevelop 5 desktop app

NilayMajorwar:
This PR fixes keyboard shortcuts for zooming in and out in the InstancesEditor not working for non-Mac platforms (Linux and Windows). Resolves #1580.

Till now, the shortcuts were being shown in the editor zoom dropdown but didn't really work on Linux or Windows, as the zoom shortcuts were set up only for Mac platform in `DeprecatedKeyboardShortcuts`, and the new KeyboardShortcuts too. I've added somewhat similar shortcuts for Linux and Windows.

I've added the following shortcuts for Windows/Linux:
- Zoom in: `Ctrl+Equal` and `Ctrl+NumpadPlus`
- Zoom out: `Ctrl+Minus` and `Ctrl+NumpadMinus`

I've also added the new numpad accelerator `CmdOrCtrl+numadd` and `CmdOrCtrl+numsub` to the Change Editor Zoom toolbar menu.

### DIFF
--- a/newIDE/app/src/SceneEditor/Toolbar.js
+++ b/newIDE/app/src/SceneEditor/Toolbar.js
@@ -140,12 +140,12 @@ export class Toolbar extends PureComponent<Props> {
             {
               label: 'Zoom in',
               click: this.props.zoomIn,
-              accelerator: 'CmdOrCtrl++',
+              accelerator: 'CmdOrCtrl+numadd',
             },
             {
               label: 'Zoom out',
               click: this.props.zoomOut,
-              accelerator: 'CmdOrCtrl+-',
+              accelerator: 'CmdOrCtrl+numsub',
             },
             {
               label: 'Center View',
--- a/newIDE/app/src/UI/KeyboardShortcuts/DeprecatedKeyboardShortcuts.js
+++ b/newIDE/app/src/UI/KeyboardShortcuts/DeprecatedKeyboardShortcuts.js
@@ -161,25 +161,15 @@ export default class DeprecatedKeyboardShortcuts {
     if (this._isControlPressed() && evt.which === MINUS_KEY) {
       this.onZoomOut();
     }
+    if (evt.which === NUMPAD_SUBTRACT) {
+      this.onZoomOut();
+    }
+
     if (this._isControlPressed() && evt.which === EQUAL_KEY) {
       this.onZoomIn();
     }
-
-    if (isMacLike()) {
-      // Mac specific shortcuts
-      if (evt.which === NUMPAD_SUBTRACT) {
-        this.onZoomOut();
-      }
-      if (evt.which === NUMPAD_ADD) {
-        this.onZoomIn();
-      }
-    } else {
-      if (this._isControlPressed() && evt.which === NUMPAD_SUBTRACT) {
-        this.onZoomOut();
-      }
-      if (this._isControlPressed() && evt.which === NUMPAD_ADD) {
-        this.onZoomIn();
-      }
+    if (evt.which === NUMPAD_ADD) {
+      this.onZoomIn();
     }
 
     if (preventDefault) {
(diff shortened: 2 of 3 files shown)


newIDE/app/src/SceneEditor/Toolbar.js
```diff
@@ -140,12 +140,12 @@ export class Toolbar extends PureComponent<Props> {
             {
               label: 'Zoom in',
               click: this.props.zoomIn,
-              accelerator: 'CmdOrCtrl++',
+              accelerator: 'CmdOrCtrl+numadd',
```
4ian, line 143: Is this working on both the web-app and the Electron app?

NilayMajorwar: Displaying accelerators isn't implemented for the web-app menus yet, as far as I can see [here](https://github.com/4ian/GDevelop/blob/a47acbb82a3c19ed8508f40be293e4eec2cb2139/newIDE/app/src/UI/Menu/MaterialUIMenuImplementation.js#L76-L79) in `MaterialUIMenuImplementation.js`. On the Electron app, the shortcuts are displayed as `Ctrl++` and `Ctrl+-` on Linux/Windows platforms. I don't know how they are displayed on MacOS platforms.

4ian: Oops you're right about the web-app! :) 
Can you check if `CmdOrCtrl+Plus` is displayed correctly? If yes I would favor that.

newIDE/app/src/UI/KeyboardShortcuts/DeprecatedKeyboardShortcuts.js
```diff
@@ -158,18 +158,26 @@ export default class DeprecatedKeyboardShortcuts {
...
+      if (evt.which === NUMPAD_SUBTRACT) {
         this.onZoomOut();
       }
-      if (this._isControlPressed() && evt.which === EQUAL_KEY) {
+      if (evt.which === NUMPAD_ADD) {
         this.onZoomIn();
       }
-      if (evt.which === NUMPAD_SUBSTRACT) {
+    } else {
+      if (this._isControlPressed() && evt.which === NUMPAD_SUBTRACT) {
```
4ian, line 177: I think we could leave the numpad plus and minus without control. What do you think?

newIDE/app/src/SceneEditor/Toolbar.js
NilayMajorwar: CmdOrCtrl+Plus is shown as `Ctrl+Shift+=` on Linux/Windows, since the `+` is up on the `=` key. I'm not sure about MacOS, but the screenshot in #1593 seems to indicate that its shown correctly as ⌘+. I'm not sure what to do here :sweat_smile:

newIDE/app/src/UI/KeyboardShortcuts/DeprecatedKeyboardShortcuts.js
NilayMajorwar: Yeah, I guess I tried to match it up with the Chrome shortcuts unnecessarily. Numpad plus and minus seem better without control. You're right, I'll change it right away.

4ian:
Looks good then, merging! Thanks for working on this 👍 Will a very useful productivity fix.

PR #1649, opened 2020-04-13
Fix: Convert confirm, alert calls to platform-dependent calls

### LINKED ISSUE #1646: Input fields grey out - not allowing selection or editing
There is an intermittent bug, where renaming or entering data in input fields seems to randomly break, not allowing the current entry to be selected, deleted or edited. When it happens, all input fields are affected. example: it breaks with renaming sprites, and then I cannot rename scenes, layers, etc.

It works sometimes, but other times seems to break for no reason, and then will start again randomly.

I can consistently get it to break importing images to sprites from outside the project folder and importing them in. When it does happen, alt tabbing out of the program and back in resolves the issue for a short time.

I am using windows 10 and GDevelop beta 90

NilayMajorwar:
Resolves #1646.
- Added `showConfirmDialog` to Window.js that uses browser's `confirm` API on web-app and `dialog.showMessageBoxSync` on Electron app. 
- Converted all `confirm` calls in the codebase to use this function, removing the now-unnecessary ESLint ignore-lines.
- Also converted an `alert` call in `path-editor.js` to use `dialog.showMessageBoxSync` instead.

Built and tested on Windows - the issue seems to be fixed.

### DIFF
--- a/newIDE/app/src/BehaviorsEditor/index.js
+++ b/newIDE/app/src/BehaviorsEditor/index.js
@@ -76,8 +76,7 @@ export default class BehaviorsEditor extends Component {
       },
       () => {
         if (this._hasBehaviorWithType(type)) {
-          //eslint-disable-next-line
-          const answer = confirm(
+          const answer = Window.showConfirmDialog(
             "There is already a behavior of this type attached to the object. It's possible to add again this behavior but it's unusual and may not be always supported properly. Are you sure you want to add again this behavior?"
           );
 
@@ -111,8 +110,7 @@ export default class BehaviorsEditor extends Component {
 
   _onRemoveBehavior = behaviorName => {
     const { object } = this.props;
-    //eslint-disable-next-line
-    const answer = confirm(
+    const answer = Window.showConfirmDialog(
       "Are you sure you want to remove this behavior? This can't be undone."
     );
 
--- a/newIDE/app/src/BehaviorsEditor/index.js
+++ b/newIDE/app/src/BehaviorsEditor/index.js
@@ -13,6 +13,7 @@ import NewBehaviorDialog from './NewBehaviorDialog';
 import { getBehaviorHelpPagePath } from './BehaviorsHelpPagePaths';
 import BehaviorsEditorService from './BehaviorsEditorService';
 import { isNullPtr } from '../Utils/IsNullPtr';
+import Window from '../Utils/Window';
 import { Column, Line } from '../UI/Grid';
 import RaisedButton from '../UI/RaisedButton';
 const gd = global.gd;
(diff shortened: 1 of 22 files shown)


newIDE/app/src/BehaviorsEditor/index.js
```diff
@@ -76,8 +76,7 @@ export default class BehaviorsEditor extends Component {
       },
       () => {
         if (this._hasBehaviorWithType(type)) {
-          //eslint-disable-next-line
-          const answer = confirm(
+          const answer = Window.showConfirmDialog(
```
4ian, line 79: Seems like the Window import is missing?

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/CollisionMasksEditor/index.js
```diff
@@ -129,8 +129,7 @@ export default class CollisionMasksEditor extends Component {
 
   _setSameCollisionMasksForAllAnimations = enable => {
     if (enable) {
-      // eslint-disable-next-line
-      const answer = confirm(
+      const answer = Window.showConfirmDialog(
```
4ian, line 132: Window import

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/PointsEditor/PointsList.js
```diff
@@ -95,8 +95,7 @@ class PointsListBody extends Component {
             });
           }}
           onRemove={() => {
-            //eslint-disable-next-line
-            const answer = confirm(
+            const answer = Window.showConfirmDialog(
```
4ian, line 98: Window import

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/PointsEditor/index.js
```diff
@@ -105,8 +105,7 @@ export default class PointsEditor extends Component {
 
   _setSamePointsForAllAnimations = enable => {
     if (enable) {
-      // eslint-disable-next-line
-      const answer = confirm(
+      const answer = Window.showConfirmDialog(
```
4ian, line 108: Window import

4ian:
Looks mostly good! Unfortunately seems like some import Window are missing... and Window is a global so Flow did not catch this :'( 
Double check the file I mentioned and do a check on all the other files - then we should be good to go!

Thanks for taking the time to do this fix :)

NilayMajorwar:
Really sorry for the missing imports :sweat_smile:. I've fixed the files you mentioned and double-checked all the other files for any other missing imports - I think it's good now.

4ian:
Looks good! 👍

4ian:
There was two wrong imports, I fixed them => always be sure to test your changes ;)

PR #1654, opened 2020-04-14
[WIP] Rename object in events' parameters

### LINKED ISSUE #1351: When renaming object, event loop are not handle
## Describe the bug
When renaming an object, events loop are not handled. (For each and Repeat)
It's handled in expression from standard event btw.

## To Reproduce

1. Setup a object
2. Add event For each, and Repeat.
3. Setup the expression in these events. ( `Count(NewObject)` for Repeat)
4. Rename the object
5. Nothing as change.

## Other details
B83

Quarkstar:
Fix #1351 
### Screenshots
fix

#### `somethingModified` 
This is in  all rename functions in `Core/GDCore/IDE/Events/EventsRefactorer.cpp`, such as `RenameObjectInActions`.[(here)](https://github.com/4ian/GDevelop/blob/804a07c56ef756f078f24c689c6d5903d50ea6c1/Core/GDCore/IDE/Events/EventsRefactorer.cpp#L177)  
It seems like it's always false. :joy: I don't find which line to change it to true.

####  Const
`vector<pair<const gd::Expression*, const gd::ParameterMetadata>> GetAllExpressionsWithMetadata() const;`

I add the above function, just like what other `rename functions` do.  But it doesn't seem to be called. Is this useful?

#### Parameter Type
For repeat event, I use a metadata of `expression`. For foreach event, I use a metadata of `object`.
Is there any other conditions?

### DIFF
--- a/Core/GDCore/Events/Builtin/ForEachEvent.cpp
+++ b/Core/GDCore/Events/Builtin/ForEachEvent.cpp
@@ -37,6 +37,18 @@ vector<gd::Expression*> ForEachEvent::GetAllExpressions() {
 
   return allExpressions;
 }
+
+vector<pair<gd::Expression*, gd::ParameterMetadata> >
+    ForEachEvent::GetAllExpressionsWithMetadata() {
+  vector<pair<gd::Expression*, gd::ParameterMetadata> >
+      allExpressionsWithMetadata;
+  auto type = gd::ParameterMetadata().SetType("object");
+  allExpressionsWithMetadata.push_back(
+      std::make_pair(&objectsToPick, type));
+
+  return allExpressionsWithMetadata;
+}
+
 vector<const gd::InstructionsList*> ForEachEvent::GetAllConditionsVectors()
     const {
   vector<const gd::InstructionsList*> allConditions;
@@ -57,6 +69,17 @@ vector<const gd::Expression*> ForEachEvent::GetAllExpressions() const {
   allExpressions.push_back(&objectsToPick);
 
   return allExpressions;
+}
+
+vector<pair<const gd::Expression*, const gd::ParameterMetadata> >
+    ForEachEvent::GetAllExpressionsWithMetadata() const {
+  vector<pair<const gd::Expression*, const gd::ParameterMetadata> >
+      allExpressionsWithMetadata;
+  auto type = gd::ParameterMetadata().SetType("object");
+  allExpressionsWithMetadata.push_back(
+      std::make_pair(&objectsToPick, type));
+
+  return allExpressionsWithMetadata;
 }
 
 void ForEachEvent::SerializeTo(SerializerElement& element) const {
--- a/Core/GDCore/Events/Builtin/ForEachEvent.cpp
+++ b/Core/GDCore/Events/Builtin/ForEachEvent.cpp
@@ -31,20 +31,13 @@ vector<gd::InstructionsList*> ForEachEvent::GetAllActionsVectors() {
   return allActions;
 }
 
-vector<gd::Expression*> ForEachEvent::GetAllExpressions() {
-  vector<gd::Expression*> allExpressions;
-  allExpressions.push_back(&objectsToPick);
-
-  return allExpressions;
-}
-
 vector<pair<gd::Expression*, gd::ParameterMetadata> >
     ForEachEvent::GetAllExpressionsWithMetadata() {
   vector<pair<gd::Expression*, gd::ParameterMetadata> >
       allExpressionsWithMetadata;
-  auto type = gd::ParameterMetadata().SetType("object");
+  auto metadata = gd::ParameterMetadata().SetType("object");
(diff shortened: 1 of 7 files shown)


4ian:
That seems mostly good! I'll double check the code tomorrow :) 

> For repeat event, I use a metadata of expression. For foreach event, I use a metadata of object.
> Is there any other conditions?

There is also the JsCodeEvent but the objects are stored in a gd::String and not a gd::Expression.  We should change this but this can be done in another PR if you prefer because you also then need to update the JavaScript code that are calling getParameterObjects: https://github.com/4ian/GDevelop/search?q=GetParameterObjects&unscoped_q=GetParameterObjects.

> I add the above function, just like what other rename functions do. But it doesn't seem to be called. Is this useful?

The `const` version of a method is called when you call the method on an object that is itself const.
For example:

```
const MyClass & myConstObject = ...;
MyClass myObject = ...;

myObject.method(); // Call the non const
myConstObject.method(); // Call the const version
````

In practice it may not be needed now.. but is still a good practice to provide one in case something in the future want to analyze the events without modifying it (in which case we want to return the vector but with const gd::Expression*).

> It seems like it's always false. 😂

Might be a mistake on my side! I'll double check :D 

Thanks!

File: Core/GDCore/Events/Builtin/ForEachEvent.cpp [hunk 1]
```diff
@@ -37,6 +37,18 @@ vector<gd::Expression*> ForEachEvent::GetAllExpressions() {
 
   return allExpressions;
 }
+
+vector<pair<gd::Expression*, gd::ParameterMetadata> >
+    ForEachEvent::GetAllExpressionsWithMetadata() {
+  vector<pair<gd::Expression*, gd::ParameterMetadata> >
+      allExpressionsWithMetadata;
+  auto type = gd::ParameterMetadata().SetType("object");
```
4ian, line 45: nitpicking: I would name this `metadata` or `parameterMetadata` rather than type

Core/GDCore/Events/Builtin/ForEachEvent.cpp
```diff
@@ -59,6 +71,17 @@ vector<const gd::Expression*> ForEachEvent::GetAllExpressions() const {
   return allExpressions;
 }
 
+vector<pair<const gd::Expression*, const gd::ParameterMetadata> >
```
4ian, line 74: nitpicking: put the implementation of the const version immediately after the non const implementation

4ian: Actually no, it's not how it was done in this file. Forget this! :)

Core/GDCore/Events/Event.h
```diff
@@ -139,6 +140,21 @@ class GD_CORE_API BaseEvent {
     return noExpr;
   };
 
+  /**
+   * \brief Return a list of all expressions with ParameterMetadata of the event.
```
4ian, line 144: Return a list of all expressions of the event, each with their associated metadata.

Core/GDCore/IDE/Events/EventsRefactorer.h
```diff
@@ -146,6 +147,19 @@ class GD_CORE_API EventsRefactorer {
                                        gd::InstructionsList& instructions,
                                        gd::String oldName,
                                        gd::String newName);
+   /**
+   * Replace all occurences of an object name by another name in a parameter of event.
```
4ian, line 151: Actually this function is doing a job not tied to an event:

`Replace all occurrences of an object name by another name in an expression with the specified metadata.`

4ian:
So I have mostly nitpciking. This seems great! 
Three things:
* `somethingModified` is for now useless, because not exposed back to the caller of RenameObjectInEvents. No big deal, *let's keep it like this* and fix that in another PR maybe if it's useful for the editor.
* Let's do the JSCodeEvent.h in another PR.
* Something to consider now: can you verify if GetAllExpressions() is even used? You added GetAllExpressionsWithMetadata - which is great. But it seems that `GetAllExpressions()` is already there and is kind of useless. I would say: remove it (in all events inheriting from `gd::BaseEvent`). Either
  * It will still compile, i.e: it was dead code.
  * Or it was not dead code and used somewhere => inspect if we can use `GetAllExpressionsWithMetadata()` instead, so that there is only the need for `GetAllExpressions()`

After that we should be good to go!

Quarkstar:
Thank you! I have fixed them all.

`GetAllExpressions` is useless and I removed it. Now there is only `PlatformExtension::GetAllExpressions` in Core.

Quarkstar:
Sorry for that, I forgot them. :cry:

4ian:
That's alright, I will merge once tests are passing :)

Quarkstar:
Thank you! I will be more careful next time.

4ian:
Done! Thanks for working on this 👍

PR #1657, opened 2020-04-15
Refactor ElectronMainMenu to functional component

NilayMajorwar:
This PR refactors ElectronMainMenu into a functional component.
I've tested it properly: there's exactly one menu build call on language change, and one menu build call on project open or close. No unnecessary menu builds, no repetitive listeners, no React warnings :)

I've used a simple custom hook `useMountEffect` for running the effect for setting up the IPC renderer listeners just once, regardless of prop changes. Directly using `useEffect(func, [])` leads to complains about missing dependencies (which are all the functions passed as prop) and then, if I do add them to the dependency array, the effect start running a lot leading to repetitive listeners.

### DIFF
--- a/newIDE/app/src/MainFrame/ElectronMainMenu.js
+++ b/newIDE/app/src/MainFrame/ElectronMainMenu.js
@@ -57,254 +57,113 @@ type RootMenuTemplate =
       submenu: Array<MenuItemTemplate>,
     |};
 
+// Custom hook to run function only once, regardless of prop changes
+const useMountEffect = func => React.useEffect(func, []);
+
 /**
  * Forward events received from Electron main process
- * to the underlying child React component.
+ * to the MainFrame component.
  */
-class ElectronMainMenu extends React.Component<MainMenuProps, {||}> {
-  _language: ?string;
+const ElectronMainMenu = (props: MainMenuProps) => {
+  const { i18n, project } = props;
 
-  componentDidMount() {
+  useMountEffect(() => {
     if (!ipcRenderer) return;
 
     ipcRenderer.on(('main-menu-open': MainMenuEvent), event =>
-      this.props.onChooseProject()
+      props.onChooseProject()
     );
     ipcRenderer.on(('main-menu-save': MainMenuEvent), event =>
-      this.props.onSaveProject()
+      props.onSaveProject()
     );
     ipcRenderer.on(('main-menu-save-as': MainMenuEvent), event =>
-      this.props.onSaveProjectAs()
+      props.onSaveProjectAs()
     );
     ipcRenderer.on(('main-menu-close': MainMenuEvent), event =>
-      this.props.onCloseProject()
+      props.onCloseProject()
     );
     ipcRenderer.on(('main-menu-close-app': MainMenuEvent), event =>
-      this.props.onCloseApp()
+      props.onCloseApp()
     );
     ipcRenderer.on(('main-menu-export': MainMenuEvent), event =>
-      this.props.onExportProject()
+      props.onExportProject()
     );
     ipcRenderer.on(('main-menu-create': MainMenuEvent), event =>
-      this.props.onCreateProject()
+      props.onCreateProject()
     );
     ipcRenderer.on(('main-menu-open-project-manager': MainMenuEvent), event =>
-      this.props.onOpenProjectManager()
+      props.onOpenProjectManager()
     );
     ipcRenderer.on(('main-menu-open-start-page': MainMenuEvent), event =>
-      this.props.onOpenStartPage()
+      props.onOpenStartPage()
     );
     ipcRenderer.on(('main-menu-open-debugger': MainMenuEvent), event =>
-      this.props.onOpenDebugger()
+      props.onOpenDebugger()
     );
     ipcRenderer.on(('main-menu-open-about': MainMenuEvent), event =>
-      this.props.onOpenAbout()
+      props.onOpenAbout()
     );
     ipcRenderer.on(('main-menu-open-preferences': MainMenuEvent), event =>
-      this.props.onOpenPreferences()
+      props.onOpenPreferences()
     );
     ipcRenderer.on(('main-menu-open-language': MainMenuEvent), event =>
-      this.props.onOpenLanguage()
+      props.onOpenLanguage()
     );
     ipcRenderer.on(('main-menu-open-profile': MainMenuEvent), event =>
-      this.props.onOpenProfile()
+      props.onOpenProfile()
     );
     ipcRenderer.on(('update-status': MainMenuEvent), (event, status) =>
-      this.props.setUpdateStatus(status)
+      props.setUpdateStatus(status)
     );
+  });
 
-    this._buildAndSendMenuTemplate();
-  }
-
-  componentDidUpdate(prevProps: MainMenuProps) {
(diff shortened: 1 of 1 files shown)


newIDE/app/src/MainFrame/ElectronMainMenu.js
```diff
@@ -57,254 +57,113 @@ type RootMenuTemplate =
       submenu: Array<MenuItemTemplate>,
     |};
 
+// Custom hook to run function only once, regardless of prop changes
+const useMountEffect = func => React.useEffect(func, []);
```
HarsimranVirk, line 61: I don't know if this is the correct approach or not, but let's go with this for now. I guess the linter is not able to find potential deps while using `useMountEffect`. I believe there's nothing we can do about the functions passed as props at the moment. When the `MainFrame` will get converted into a functional component, then we might wrap these functions with `useCallback` in `MainFrame` itself, which should be the proper way.

NilayMajorwar: I did think about cleaning up the listeners but haven't tried it yet. But now that you mention it, I just thought of a possibly cleaner approach - a custom `useEffect` based hook that registers a function for a certain event, and in cleanup removes that function. Seems intuitive, should work well with `useCallback` in the MainFrame, and will be much cleaner in code. I'll try to implement this in a while.

NilayMajorwar: I have an alternative implementation ready [here](https://github.com/NilayMajorwar/GDevelop/blob/emm-refactor/ipc-hook/newIDE/app/src/MainFrame/ElectronMainMenu.js) that uses a custom hook for registering and cleaning up IPC event listeners. Both of them work well, but I'm not sure which one is better.

newIDE/app/src/MainFrame/ElectronMainMenu.js
```diff
@@ -319,40 +178,177 @@ class ElectronMainMenu extends React.Component<MainMenuProps, {||}> {
...
+      }
+    },
+    [i18n, project]
+  );
+
+  React.useEffect(
+    () => {
+      _buildAndSendMenuTemplate();
+    },
+    [_buildAndSendMenuTemplate, i18n.language]
```
4ian, line 348: So the effect runs if the callback changes, which is the case if i18n or project changes. Clever - almost a bit obfuscated! But it works :)

newIDE/app/src/MainFrame/ElectronMainMenu.js
4ian: It's a very interesting case. Basically, the linter warned us of something that is odd:
we want to use useEffect with an empty array (to run it only at the component mounting), but then the linter says "hey you're using some props, what are you doing dude??".

Basically, this put the stress on the fact that the existing implementation had a bug! There was no unregistering of the event listener.
It was no big deal because there was no need to unregister the listeners... but still it was not "perfect".

React Hooks are in a way forcing us to write correct, proper code that is cleaning the stuff after us. :) 

> I have an alternative implementation ready here that uses a custom hook for registering and cleaning up IPC event listeners. 

I think your alternative implementation is the proper one to use. It's properly unregistering the callbacks.
This means that previously, if someone wanted to play with the menu, like removing the component then re-render it later (I don't know why, but for example if you wanted to hide the menu for some reason) then... there would be a bug because of the listeners.
The alternative implementation seems, from what I've seen, to do the proper job so that the component is now safe! 👍

4ian: Thanks both for your nice collaboration on this. These are tricky but interesting parts of React. Understanding them well allow us to make robust code that works in 100% of the cases.

newIDE/app/src/MainFrame/ElectronMainMenu.js
NilayMajorwar: I encountered something strange while working on this part. What I actually wanted was to keep `i18n.language` and `project` as the `_buildAndSendMenuTemplate` callback dependencies to make it a bit more clear.

But React complained about `i18n` as a missing dependency (which is fine), but if I add it then React calls `i18n.language` as an unnecessary dependency (which seems fine too). But then somewhat strangely, `i18n` does not cause the effect to rerun on language change while `i18n.language` does! So I added `i18n.language` to this effect instead, to make it work properly on language change and remove any warnings.

4ian: Yes this is because i18n object is the same during all the lifetime of the app, while i18n.language can change - but i18n is not changing. Another way to avoid the warning that is maybe safer would be to pass language itself:
```
const language = i18n.language;
...
[_buildAndSendMenuTemplate, project, i18n, language]
```

something like this :) Up to you!

NilayMajorwar: > Another way to avoid the warning that is maybe safer would be to pass language itself

I'm already passing `i18n.language` as dependency to `useEffect` - would destructuring change anything? Or if you mean adding `language` dependency to `useCallback`, that still gives the warning. I'm not sure I'm fully clear about what you're suggesting here.

NilayMajorwar: Another option is to completely remove the intermediary `_buildAndSendMenuTemplate`, and put its content directly into a `useEffect` with dependencies as `i18n, project, language`. Then  I guess the dependency list would almost directly reflect the intention, that we want to rebuild the menu on project or language change.

4ian: > I'm already passing i18n.language as dependency to useEffect - would destructuring change anything?

This won't change the runtime behavior, but this will avoid the linter warning, because the linter will warn you if you use `i18n, i18n.language`, but not if you use `i18n, language`.

> Another option is to completely remove the intermediary _buildAndSendMenuTemplate, and put its content directly into a useEffect with dependencies as i18n, project, language. 

That's an option, and if `_buildAndSendMenuTemplate` is not used anywhere else that might be clearer indeed.

NilayMajorwar: The useEffect does not have i18n as dependency - the useCallback has it (since the _build function uses i18n._). And adding language to the useCallback dependency list leads to unnecessary dep warning regardless of destructuring :(

> That's an option, and if `_buildAndSendMenuTemplate` is not used anywhere else that might be clearer indeed.

Yeah, the build function isn't used anywhere else in the component, so that won't be a problem. Should I push a commit for this?

4ian: Yes that sounds fine. You can also put `buildAndSendMenuTemplate` outside of the component itself and pass to it all the arguments it requires. This allows for:
* A clean function, that has a name so is "auto documenting".
* A useEffect that is both linter compatible and very clear ("in case of change in project, i18n, language, re-run buildAndSendMenuTemplate" => make sense! We don't even need a comment).

NilayMajorwar: I've made the changes. You're right, moving the function out of the component certainly made it much better and cleaner :smile:

4ian:
Looks good to me, merged! Thanks for doing this! The component is now clearer and more robust against potential future usages. 👍

Issue #1449, opened 2020-02-21
[$200+ Bounty]Implement "Bitmap Text" object to avoid blurry or distorted text objects in pixel perfect games (and benefit from higher performance of the text rendering)


Silver-Streak:
## Describe the bug
When using a text object (or BBText object), if you resize the game, the text will become either blurry or distorted. This seems to have gotten worse since the release of the new resolution options and events related to them.

This occurs if you resize through any method, including by dragging and resizing the window, using fullscreen on a monitor that is a different resolution, or by using any of the resolution/fullscreen events.

Using the solution here does not solve this problem, although it does slightly reduce the effect: https://forum.gdevelop-app.com/t/solved-pixel-art-fonts-are-fuzzy-text-font-blury/14225

## To Reproduce

1. Create a text object.
2. Set the font size to something large so it is more noticeable (Size 40+).

   - Note: this still occurs on small fonts, just harder to notice.

3. Set your game resolution to any size smaller than your monitor (e.g. 1280x720 on a 1080p monitor)
4. Place the text object in the scene.
5. Launch the preview, or compile the game and run it.
6. Stretch the window larger, or set the game to fullscreen. 
7. The text will either be blurry, or have distorted pixels. This occurs regardless of whether using linear or nearest scaling.

I've attached the default platformer example project with some text objects with both the default font and a custom font to show this effect. 
[My project4.zip](https://github.com/4ian/GDevelop/files/4237568/My.project4.zip)

In my actual game, here's some examples Using the example fix linked above (showing it does not resolve it): 

Here's the text before the window has been changed in size (720p)- 
![image](https://user-images.githubusercontent.com/24237149/75062479-7dcc4780-54a8-11ea-979f-e711c0fb533d.png)
![image](https://user-images.githubusercontent.com/24237149/75062489-8290fb80-54a8-11ea-8bf7-0b432761a6b1.png)

Here's the text after the game has been fullscreened on a 1080p monitor-
![image](https://user-images.githubusercontent.com/24237149/75062568-a94f3200-54a8-11ea-9f52-d400e7a60a62.png)
![image](https://user-images.githubusercontent.com/24237149/75062558-a3595100-54a8-11ea-9a60-3b1f66d4ac2d.png)

It is important to note: If you create an image with the text you want in an image editor (such as gimp) and add it as a sprite object, it _does not_ have the blurriness or distortion when resizing. So it is something specific with how GDevelop/Pixi is scaling the Text Objects.

## Other details
[...]

PR #1658, opened 2020-04-15
Fix_renderer_for_fix_blur_fonts_and_images

Bouh:
[See the complete thread ](https://github.com/4ian/GDevelop/issues/1449)

I've used the dimension of the window for said to the renderer to comptue a renderer equal to the size of the window.
Before this it was used the size of the game, in Project manager > Game settings > Width and height.

Only tested on preview for now.

image

I've build two release with my bug fix.

[linux-unpacked.zip	](https://mega.nz/file/obx3FZjT#VyNlXYa_S3MYHlgQ-Bzn-Kd5mzRTjNyxBTUxCJFlRbM)

[win-unpacked.zip](https://mega.nz/file/dSglUZTA#aC-WV9viWizNp-xd0BYm9-a_qxEryu712NYIsaCpXYk)

### DIFF
--- a/Extensions/TextObject/textruntimeobject-pixi-renderer.js
+++ b/Extensions/TextObject/textruntimeobject-pixi-renderer.js
@@ -34,7 +34,7 @@ gdjs.TextRuntimeObjectPixiRenderer.prototype.updateStyle = function() {
     var style = this._text.style;
     style.fontStyle = this._object._italic ? 'italic' : 'normal';
     style.fontWeight = this._object._bold ? 'bold' : 'normal';
-    style.fontSize = this._object._characterSize;
+    style.fontSize = this._object._characterSize * 4;
     style.fontFamily = fontName;
 
     if (this._object._useGradient){
@@ -72,6 +72,7 @@ gdjs.TextRuntimeObjectPixiRenderer.prototype.updateStyle = function() {
     // Prevent spikey outlines by adding a miter limit 
     style.miterLimit = 3;
 
+    this._text.scale.set(0.25,0.25);
     this.updatePosition();
 
     // Manually ask the PIXI object to re-render as we changed a style property
--- a/Extensions/TextObject/textruntimeobject.js
+++ b/Extensions/TextObject/textruntimeobject.js
@@ -93,7 +93,7 @@ gdjs.TextRuntimeObject = function(runtimeScene, textObjectData)
     this._shadowAngle = 0;
 
     /** @type {number} */
-    this._padding = 5;
+    this._padding = 50;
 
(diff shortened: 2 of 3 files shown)


Silver-Streak:
Great news: Quick test on this definitely makes it look better when I maximize a window or drag it to a new size.

Bad News: It appears to break window resize/resolution change events.

Silver-Streak:
Okay, more good news:

The window resize/resolution changes are unrelated to this change, and occur only in preview (appears to have been introduced in b90, just went back to 89 without issues)

But bad news:
This seems to still have scaling issues rather than changing font size, unless I'm misunderstanding:

Here's text at native resolution 1920x1080, font size: 72
image

Here's the window resolution changed to 1280x720, font size still 72 according to debugger:
image

You can see it become very distorted once I change the resolution.

(Also, this doesn't appear to fix blurriness on BBText, but I don't think we're trying for that currently)

Bouh:
It's only on Text object not yet on BBtext.

Silver-Streak:
Yep, that's what I assumed, the above examples are just text objects, not BBtext.

4ian:
What is the state of this? :) 
Could the work around of the text be implemented as an option?

I'm not sure about the change for the game size, the resolution of the game should still be the one selected by the developer.

Silver-Streak:
@4ian We've been debugging over on the issue thread. Seeing some great improvements (Going from a lower resolution to a higher resolution seems to keep text looking near perfect, GREATLY improved from current normal builds) but still some weirdness (going from a higher resolution to a lower resolution has some distortion still)

I think Bouh is looking into Pixel rounding currently to potentially fix the downscaling, which I think would make sense to have as a toggle option (that is currently how it is done in other engines like Construct 3 and Game Maker Studio 2), but no clue if it'll actually fix the issue.

Bouh:
We have still an issue when the resolution is set on 1920x1080 and downscale to 1280x720, font is still a bit blurry, but if we go on 1279x719, font is perfect...
I don't know why.
I keep it in draft for now because i test different things,
The main thread about this [issue is here](https://github.com/4ian/GDevelop/issues/1449). 


>I'm not sure about the change for the game size, the resolution of the game should still be the one selected by the developer.

The user can always set the dimension of the window. This my patch i change just the renderer, if we use a small renderer and we stretch it for fit the window all the content is upscaled but the original resolution still the original setup by the user. So if he setup 720p, in game settings, and use fullscreen on a 1080p, the content will be blurred.

However this is not defenitive, if a better approach is found i can close this PR. 
Help is still needed, we are in the blur^^

Bouh:
See for BitmapText

PR #1662, opened 2020-04-16
Quick Fix for websockets debugger client due to wrong refactoring

arthuro555:
Fixing the wrong refactoring of runtimegame to runtimeGame in #1637


arthuro555:
@4ian The fix for the problem you mentioned

4ian:
Thank you! 👍

arthuro555:
Sorry for Breaking it in the first place :/

PR #1664, opened 2020-04-16
Add Type checking on websocket debugger client

arthuro555:
Adds type checks to websocketDebuggerClient.
Also contains a new script in GDJS package.json to generate doc.

### DIFF
--- a/GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
+++ b/GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
@@ -1,8 +1,9 @@
+/// @ts-check
 /**
  * An client side implementation of the Debugger
  * @interface
  */
-gdjs.DebuggerClient = function() {} // Necessary for Interface definition 
+gdjs.IDebuggerClient = function() {}
 
 /**
  * Update a value, specified by a path starting from the {@link RuntimeGame} instance.
@@ -11,6 +12,7 @@ gdjs.DebuggerClient = function() {} // Necessary for Interface definition
  * @param {Array<string>} path - The path of the property to modify, starting from the RuntimeGame.
  * @param {any} newValue - The new value.
  */
+gdjs.IDebuggerClient.set = function(path, newValue) {}
 
 /**
  * Call a method, specified by a path starting from the {@link RuntimeGame} instance.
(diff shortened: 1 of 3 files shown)


arthuro555:
Ready for review @4ian

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
```diff
@@ -384,6 +381,7 @@ gdjs.WebsocketDebuggerClient.prototype._depthLimitedSerializer = function(
       if (maxDepth != null && thisPos > maxDepth) {
         return '[Max depth reached]';
       } else if (~stack.indexOf(value))
+      // @ts-ignore
```
arthuro555, line 384: That is ignored as it says "cycleRepeater might be undefined", but as you can see above it is checked for undefined and null and if it is one of both it get defined.

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
```diff
@@ -347,6 +343,7 @@ gdjs.WebsocketDebuggerClient.prototype._circularSafeStringify = function(
 ) {
   return JSON.stringify(
     obj,
+      // @ts-ignore
```
arthuro555, line 346: That one is weir as it says the arguments are from the wrong type, but the types correspond. To be more precise it says that basically `DebuggerClientCycleReplacer !== DebuggerClientCycleReplacer` wich isn't true.

arthuro555:
I commented the 2 ts-ignore I had to use and why

arthuro555:
Semaphore uses an outdated version of node :eyes:

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
4ian: Weird, no big deal then.

GDJS/docs/jsdoc.conf.json
```diff
@@ -50,6 +50,7 @@
     "recurse": true,
     "private": false,
     "lenient": true,
+	"template": "docs/jaguarjs-jsdoc",
```
4ian, line 53: Can you use spaces rather than tabs.

GDJS/package.json
```diff
@@ -7,13 +7,15 @@
...
     "jsdoc-plugin-intersection": "^1.0.2",
     "typescript": "3.6.4"
   },
   "dependencies": {
     "jsdoc-autoprivate": "0.0.1"
   },
   "scripts": {
-    "check-types": "tsc"
+    "check-types": "tsc",
+	"generate-doc": "jsdoc -c docs/jsdoc.conf.json docs/DocMainPage.md"
```
4ian, line 19: nitpicking: same, can you use spaces rather than tabs?

4ian:
I've updated the semaphore Node version. The two ts-ignore seems fair, there might be a way to avoid them, but no big deal!

4ian:
Looks good, thanks! 👍

PR #1670, opened 2020-04-18
Fix advanced bloom

Bouh:
This filter need kawase-blur, without we have :
Uncaught (in promise) TypeError: r.KawaseBlurFilter is not a constructor

----

I've seen your change on layer list, the list is on bottom of my window, is it expected? (ignore if this is in WIP :) )
image


4ian:
> I've seen your change on layer list, the list is on bottom of my window, is it expected

For now yes because it's not showing well in a small column. I'll try to rework this.

4ian:
Thanks for the PR!

Issue #1631, opened 2020-04-07
DialogueTree::LineText() incompatible with DialogueTree::ClippedLineText()


PascalLadalle:
Not sure if this should go here, in feature requests or on the wiki. 😅 
Using DialogueTree::LineText() stops the scrolling.
Scrolling works or not, depending on whether I disable or not this event:
![image](https://user-images.githubusercontent.com/25978830/78720759-8bd60a80-7926-11ea-9015-de611c20a54c.png)

blurymind:
Yeah, unfortunately I think when you get the line text, it also triggers the scrolling text to complete. 

In your case this is a problem because you don't want it to affect the condition that catches when it completes scrolling

PR #1671, opened 2020-04-18
Fix dialoguetree bugs, add better regression testing (debug mode)

blurymind:
Bugs this fixes with the dialogue tree:
- Most serious one is a regression in which if you combine text, commands and options/links in a certain way, the dialogue tree skips rendering some text results, just goes to the next result (enabling debugging in the example will catch this regression if it occurs again)
- Setting a bondagejs dialogue state variable from the event sheet was completely wrong as bondagejs wasnt using a normal object like expected. I fixed that and added a test for it in the example (thats how you enable debugging in the example)
- isDialogueLineType never returning true when checking if the result is a command and the command is never checked for being called by the event sheet (added to the debugging test now)
- getting the full text scrolls the clipped text to the end 
 https://github.com/4ian/GDevelop/issues/1631
- add to example project a case of how to use the wasBranchVisited condition + change the wording to be clearer
https://github.com/4ian/GDevelop/issues/1673
- add example to show how to load/save bondagejs state
- added a new feature to get the line number to bondage.js in order to fix the new line bug
https://github.com/hylyh/bondage.js/commit/3c63e2193c3b82de82e190d00ce262be50939272

When you have 
`text in nodeA > link to nodeB(single option is a link) > command in node b > text in nodeB,
`
 the text in nodeA will get merged with the text in nodeB - when you use the scrolling text feature. This is in the example project, if you uncoment both debugging options

This bug is actually in the logic that appends a text result to the scrolling text rather than setting it - its whats used to concatenate texts with 1 or more commands between them.
Worth noting that when you use a link, bondagejs doesnt register that as an option and the extension never catches it as its never detected. I have a way of catching when we have navigated to a new node by comparing their titles, however when I use that for some reason the text I want to set gets scrolled to its end and skipped.

It is kind of a very specific corner case bug, but it is noticeable

Edit: fixed that too


blurymind:
Worth noting that my prettier formater is broken atm and npm run format is not covering the extensions folder.

When I format with prettier's default options in vscode, I get this
`= function(bleh) {}` becomes `= function (bleh) {}`

blurymind:
Worth noting that there has been a lot of development recently at the yarn tracker
https://github.com/YarnSpinnerTool/YarnEditor/commits/master

This new contributor has been keeping me busy reviewing code, but he fixed a lot of stuff and improved performance. I also added a story previewing feature to it in the meantime.
I can do another PR to update yarn in gdevelop once things settle down a bit :)

Bouh:
> When I format with prettier's default options in vscode, I get this
> `= function(bleh) {}` becomes `= function (bleh) {}`
> 
> they noted the change here
> https://prettier.io/blog/2020/03/21/2.0.0.html
> 
> problem is If I do this on any files in gdevelop,my difs will become a mess on any pr I do :(

[I was not fool!](https://github.com/4ian/GDevelop/pull/1640#issuecomment-612430470) I've change for Prettier Now by Remi Marsal. But it's a bit less pratical, there is no "format the selection" with this extension on VSCode.

4ian:
Alright so we either need to add a npm run format for extensions (maybe only a few files) and/or upgrade to Prettier 2 (but that will change a lot of files) :)

blurymind:
Btw have you guys noticed that for some reason emojis in yarn bundled with gdevelop stopped working for some reason? Not sure if this is due to running it in devmode. They used to work in the past, so I wonder what broke them in gdevelop specifically

4ian:
Might be the upgrade to Electron 8, which includes a newer Chrome version?

Extensions/DialogueTree/JsExtension.js
```diff
@@ -220,7 +220,7 @@ module.exports = {
         'JsPlatform/Extensions/yarn32.png'
       )
       .addParameter('string', _('State Variable Name'), '', false)
-      .addParameter('expression', _('Variable Value'), '', false)
+      .addParameter('string', _('Variable Value'), '', false)
```
4ian, line 223: Is this done on purpose? This can break a fair number of games using the extension, no?

Extensions/DialogueTree/dialoguetools.js
```diff
@@ -420,7 +428,7 @@ gdjs.dialogueTree.startFrom = function(startDialogueNode) {
   if (!this.hasDialogueBranch(startDialogueNode)) return;
   this.optionsCount = 0;
   this.options = [];
-  this.dialogueBranchTitle = '';
+  this.dialogueBranchTitle = null;
```
4ian, line 431: What does this changes do?

Extensions/DialogueTree/dialoguetools.js
```diff
@@ -505,6 +523,7 @@ gdjs.dialogueTree.goToNextDialogueLine = function() {
       time: this.dialogueText.length + offsetTime,
     });
     this.dialogueData = this.dialogue.next().value;
+    this.dialogueDataType = 'command';
```
4ian, line 526: What's the reason for moving this?

Extensions/DialogueTree/JsExtension.js
blurymind: yes it was, it should have been a string from the start, as the previous (expression) only allowed numbers - which is not at all compatible with what the yarn syntax and bondagejs supports (booleans, strings, floats and integers)

Extensions/DialogueTree/dialoguetools.js
blurymind: I tried to see if it fixes the bug I described, but it doesnt really do it. I guess I can undo this

Extensions/DialogueTree/dialoguetools.js
blurymind: doesnt really change anything. I didnt like the way I was checking this when fixing these bugs. I can undo this chage

Extensions/DialogueTree/JsExtension.js
blurymind: This is also more compatible with how you set/get them inside yarn (always strings) and they are checked in bondagejs (parsed from the json file strings that get converted to their appropriate type)

blurymind:
I will try to expand the stress test more on this PR and also try to fix that bug I mentioned in one of the comments. Anyone feel free to join in if you have any ideas/suggestions :)

blurymind:
@4ian  please hold off merging this. I submitted a pr to bondagejs here:
https://github.com/hylyh/bondage.js/pull/63

It adds to the library the ability to register the lines it gets results from. This should enable me to get rid of some of the hacky code in gdevelop's wrapper as well as fix this annoying new line or no newline bug that keeps getting raised

4ian:
Sure let's wait for this then, sounds cleaner!

blurymind:
@4ian I have now fixed all the bugs, butto get the last one I had to replace the vanillabondage.min.js  with the one built from the PR here 
https://github.com/hylyh/bondage.js/pull/63

We can wait for @hylyh to review it before we move forward, but not sure when she gets some time to do it

Extensions/DialogueTree/dialoguetools.js
blurymind: I now changed this logic completely to properly initialize these values instead of starting them out as nulls

Extensions/DialogueTree/JsExtension.js
4ian: > I am fairly convinced nobody actually used this properly in their game, otherwise we would have seen a bug report to complain about those types not working

Ok fair enough!

Extensions/DialogueTree/dialoguetools.js
```diff
@@ -654,8 +676,11 @@ gdjs.dialogueTree.compareVariable = function(key, value) {
  * @param {string} value The value you want to set
  */
 gdjs.dialogueTree.setVariable = function(key, value) {
-  if (this.dialogueIsRunning && this.runner.variables.data) {
-    this.runner.variables.data[key] = value;
+  if (this.runner.variables) {
+    this.runner.variables.set(
+      key,
+      !isNaN(value) || value === "true" || value === "false" ? eval(value) : value
```
4ian, line 682: `eval` is very dangerous! 😬 
We [don't want any eval at all](https://www.quora.com/Why-is-eval-evil) - because it opens the way too much errors, exploits, possibilities of breaking the game, tie us to JavaScript forever, it's dynamic code execution so not performant.

Instead, write the conditions you need explicitly. 

Remember, explicit is better than implicit.... and explicit ifs/else or ternary condition is 10x better than kickstarting an eval which will have to parse your code and run a JS engine on it. It's never worth it.

Extensions/DialogueTree/dialoguetools.js
```diff
@@ -686,7 +711,9 @@ gdjs.dialogueTree.loadState = function(inputVariable) {
       gdjs.evtTools.network.variableStructureToJSON(inputVariable)
     );
     gdjs.dialogueTree.runner.visited = loadedState.visited;
-    gdjs.dialogueTree.runner.variables.data = loadedState.variables;
+    Object.entries(gdjs.dialogueTree.runner.variables).forEach(function([key, value]) {
```
4ian, line 714: This array destructuring is not ES5 compatible.
https://github.com/4ian/GDevelop/blob/master/newIDE/docs/Supported-JavaScript-features-and-coding-style.md#for-the-game-engine-gdjs-and-extensions

In the game engine, we can't use arrow functions, let/const or destructuring.

Extensions/DialogueTree/dialoguetools.js
```diff
@@ -120,9 +133,11 @@ gdjs.dialogueTree.completeClippedTextScrolling = function() {
  * Useful to prevent the user from skipping to next line before the current one has been printed fully.
  */
 gdjs.dialogueTree.hasClippedScrollingCompleted = function() {
-  if (!this.dialogueIsRunning) return false;
-  if (this.dialogueData && this.dialogueText.length) {
-    return this.clipTextEnd >= this.dialogueText.length;
+  if (!this.dialogueIsRunning || this.dialogueDataType  === '') return false;
+  
+  if (this.dialogueData && this.dialogueText.length > 0 && this.clipTextEnd >= this.dialogueText.length) {
+    if (gdjs.dialogueTree.getVariable("debug")) console.warn("Scroll completed:",this.clipTextEnd,"/", this.dialogueText.length);
```
4ian, line 139: nit: space after ,

4ian:
Thanks for all the fixes 👍
I've seen an `eval` in the code -> let's remove this bad guy :)

>  I have now fixed all the bugs, but to get the last one I had to replace the vanilla bondage.min.js with the one built from the PR here

Seems it was merged! Are you able to upload bondage.min.js by the latest "official" version? 
As the file seems not to contain the version number, I think we should add a README.md in the folder containing bondage.min.js with the version number :) 

Apart from that we should be good to go if you're finished. 👍

blurymind:
> Thanks for all the fixes 
> I've seen an `eval` in the code -> let's remove this bad guy :)
>

Ah sorry, totally escaped my mind. I will change this. Is it ok to use JSON.parse here?
I need it to convert to the correct var type as it would when reading it from a json text file


> Seems it was merged (thanks for doing the PR on bondage)! Are you able to upload bondage.min.js by the latest "official" version?
> As the file seems not to contain the version number, I think we should add a README.md in the folder containing bondage.min.js with the version number :)
> 

Since mine was the last commit, the minified file is identical to what is currently at the repo.
I added bondage's original readme, license and a version number. Hope that is ok.

Also moved it to a more convenient folder in case I need to fix stuff on it while testing on gdevelop again

> Apart from that we should be good to go if you're finished.

Thanks, I will apply the review notes asap :) glad to squash these bugs finally and implement a way to detect them if they reoccur

Extensions/DialogueTree/dialoguetools.js
blurymind: ah good catch. I've gotten so used to deconstructing now it's sometimes hard to stop myself

blurymind: fixed now

Extensions/DialogueTree/dialoguetools.js
blurymind: replaced with JSON.parse

Extensions/DialogueTree/dialoguetools.js
blurymind: fixed

Extensions/DialogueTree/dialoguetools.js
```diff
@@ -120,9 +133,11 @@ gdjs.dialogueTree.completeClippedTextScrolling = function() {
  * Useful to prevent the user from skipping to next line before the current one has been printed fully.
  */
 gdjs.dialogueTree.hasClippedScrollingCompleted = function() {
-  if (!this.dialogueIsRunning) return false;
-  if (this.dialogueData && this.dialogueText.length) {
-    return this.clipTextEnd >= this.dialogueText.length;
+  if (!this.dialogueIsRunning || this.dialogueDataType  === '') return false;
+  
+  if (this.dialogueData && this.dialogueText.length > 0 && this.clipTextEnd >= this.dialogueText.length) {
+    if (gdjs.dialogueTree.getVariable("debug")) console.warn("Scroll completed:", this.clipTextEnd,"/", this.dialogueText.length);
```
4ian, line 139: nit: space after comma

Extensions/DialogueTree/dialoguetools.js
```diff
@@ -654,8 +676,8 @@ gdjs.dialogueTree.compareVariable = function(key, value) {
  * @param {string} value The value you want to set
  */
 gdjs.dialogueTree.setVariable = function(key, value) {
-  if (this.dialogueIsRunning && this.runner.variables.data) {
-    this.runner.variables.data[key] = value;
+  if (this.runner.variables) {
+    this.runner.variables.set(key, JSON.parse(value));
```
4ian, line 680: Why do you want to parse `value`?

4ian:
> Is it ok to use JSON.parse here?
I need it to convert to the correct var type as it would when reading it from a json text file

I don't understand why you need this? It seems not very common in GDevelop to ask user to enter JSON formatted strings in the events. Why is not storing a string enough?

blurymind:
I am comparing the values in one of my conditions. Bondagejs stores them as different types internally which is a bit unfortunate as it forces me to convert what comes from gdevelop as string to the correct type bondagejs has them as.
Basically when you use the yarn syntax to set or compare values, bondagejs keeps what the user has accumulated as the Text is parsed and hidden variables in the text are set to stuff. Something visited set to true, something else set to number or string. So later on when I want gdevelop to check if the value that is in the bondagejs state is equal to something, I am doing the same as bondagejs, only now on the extension side.

Extensions/DialogueTree/dialoguetools.js
blurymind: To preserve the type bondagejs expects them to be. If something is set with yarn syntax, bondage converts it automatically to a type. Now imagine we want to overwrite that same bondagejs variable from the event sheet. We need to do the same in order to preserve its type

blurymind: to see what I mean, open yarn and set a bunch of different variables with the types boolean,int, float and string.

Then console.log(this.runner.variables) - this is where bondagejs stores/reads them.

Now run the game and go through the yarn node where the yarn syntax is setting them. You will find that the variables object gets populated with them in different types.They are not all just strings

Extensions/DialogueTree/dialoguetools.js
blurymind: fixed

4ian:
> you could set something that bondagejs sees as a string or boolean to a number type - that is a huge problem

I guess you can still do it, if you're setting a variable to another type (like if it was a string, and I enter a number like 123).

> just write the value and the extension will handle converting it

That's the part I'm worried about.
As we're asking the user for a string, the user will have to enter something like `"123"` to enter a number (notice the quotes).
It seems counter intuitive to have to write quotes in GDevelop to enter a number right?
Worse, if I understand properly if you want to write a string, writing in GDevelop Events Sheet `"Hello world"` will *not* work. Because it will be understood as the JSON `Hello world`, which is invalid. So you have to enter in GDevelop: `"\"Hello World\""`.

There is a "mismatch" between the abstractions of GDevelop and the extension. We're exposing different types, the JavaScript types, through a single interface in GDevelop: a string. Not ideal... it works but it's more an escape hatch rather than the default thing we must expose to users.

Instead, the solution I'm thinking of is to do typed actions:
* One action to set a variable to a number
* One action to set a variable to a string
* One action to set a variable to a boolean (true or false)
* One action as a "escape hatch", to set a variable to an arbitrary JSON (for advanced users).

In practice, this means doing 4 actions: `SetNumberVariable`, `SetStringVariable`, `SetBooleanVariable`, `SetJSONVariable`. They only differ by the type of the parameter for the variable value: `expression`, `string`, `trueorfalse`, `string` and the associated JS function that they call (probably one function for each type).
The difference between the 2nd and the last one is that the last one (`SetJSONVariable`) is doing `JSON.parse` on the string passed (with a `try catch`! Otherwise we get a crash in case of invalid JSON).
Why no `JSON.parse` for the others? Because GDevelop is taking care for you of passing the proper type for the value :)

Ideally maybe in the future we could rework such actions to be a single action with a value parameter that has a type that can be choosen by the user in the interface - rather than 4 actions. But not big deal, it's a bit of copy/pasta to do only once.

Does that sounds doable? Basically, it's the same approach as variables in GDevelop.

4ian:
My previous explanation is also the same for the conditions to check the variables.
You do 3 conditions, allowing to get the Bondage variable as number, string or boolean, and 3 js functions that will do the cast to the proper type for you (`parseFloat(value)`, `"" + value`, `!!value`). Ideally, the cast is useless, because the user takes care of changing and testing a variable with the same type every time... but in case a user change a variable as a string, then compare it as a number, we cast it, if needed.

blurymind:
That will make it more complicated imo, as now the user needs to cast it in the correct type, meaning guess what the type is that bondagejs decided to give it when parsing it. If there is a mismatch,it will be even harder to guess.

When writing conditional logic in yarn, it it is 99% using simple and short values. 
I dont think anyone will need to escape characters or shove in an entire object in there. Actually I dont think you can set/get entire objects with yarn syntax, so having another action to do that will not serve any purpose in communicating with what bondagejs is doing with the actively parsed yarn syntax.

I am a bit weary that adding this, we let the user break/abuse and report bugs about it.

What if in another PR we add the new advanced commands to set/get as different types, but keep at the main one where it is automatically detected for them?

4ian:
If you're worried about the advanced action to set a JSON - I'm also worried about it so let's not add it ;)

>  but keep at the main one where it is automatically detected for them?

Well, if you re-read this, I don't think the "autodetection" is even doing what you want:
> That's the part I'm worried about.
> As we're asking the user for a string, the user will have to enter something like "123" to enter a number (notice the quotes).
> It seems counter intuitive to have to write quotes in GDevelop to enter a number right?
> Worse, if I understand properly if you want to write a string, writing in GDevelop Events Sheet "Hello world" will not work. Because it will be understood as the JSON Hello world, which is invalid. So you have to enter in GDevelop: "\"Hello World\"".

It does not seems intuitive.
An action to set a variable, clearly saying if it's a number, string or boolean is I think much much safer than just putting a JSON.parse and hoping for people to write "\"Hello World\"" in GDevelop because they want to save a string in a variable, no?

So:
One action to set a variable, with a number
One action to set a variable, with a string
One action to set a variable, with true/false

and that should be the 99% simple values that we need.

> That will make it more complicated imo, as now the user needs to cast it in the correct type, meaning guess what the type is that bondagejs decided to give it when parsing it.

I hope people know if they put a string, a number or a boolean in a variable? That seems simple enough? 
But anyway the current solution is not really autodetecting anything, it's just launching JSON.parse, which I think is not doing what you think.

blurymind:
ahh you are right, I think I need to take a careful look again what is going on

blurymind:
@4ian fixed it now. It uses  explicit types, three conditions and three actions to set/get - boolean, string and number

4ian:
Great! I'll take a look 😄
After this is merged, I'm thinking of making a new release, sounds good to you?

Extensions/DialogueTree/JsExtension.js
```diff
@@ -619,6 +653,40 @@ module.exports = {
...
+        _(
+          'Compare dialogue state variable. Use this to trigger game events via dialogue variables.'
+        ),
+        _('Dialogue state boolean variable _PARAM0_ is equal to _PARAM1_'),
+        _('Dialogue Tree (experimental)'),
+        'JsPlatform/Extensions/yarn24.png',
+        'JsPlatform/Extensions/yarn32.png'
+      )
+      .addParameter('string', _('State variable'), '', false)
+      .addParameter('yesorno', _('Equal to true'), '', false)
```
4ian, line 686: I think `trueorfalse` is better than `yesorno`

blurymind:
We have a trueorfalse? :o I will update this asap.

Extensions/DialogueTree/JsExtension.js
blurymind: updated now

Extensions/DialogueTree/dialoguetools.js
```diff
@@ -686,7 +708,9 @@ gdjs.dialogueTree.loadState = function(inputVariable) {
       gdjs.evtTools.network.variableStructureToJSON(inputVariable)
     );
     gdjs.dialogueTree.runner.visited = loadedState.visited;
-    gdjs.dialogueTree.runner.variables.data = loadedState.variables;
+    Object.entries(gdjs.dialogueTree.runner.variables).forEach(function(pair) {
```
4ian, line 711: This will probably not work on Android 4.4: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries

Android 4.4. has Chromium 30. Wdyt?

4ian:
If everything seems ok to you, we can merge this and I'll release a new version. Let me know if that's ok for you?

Extensions/DialogueTree/dialoguetools.js
blurymind: Oh I didn't know. I can change this into a for loop or something

blurymind:
I think it should be alright now. This fixes a few really bad bugs. Text getting skipped, state not being set properly among others.

What's more the example project and extension  now have a debugging mode and better test features for regressions.

I can look into removing that object entries loop tomorrow, but I have no way of testing on such an old version of Android. A simple old school for loop should do the same thing and work on old chrome

4ian:
> I can look into removing that object entries loop tomorrow, but I have no way of testing on such an old version of Android. A simple old school for loop should do the same thing and work on old chrome

Yep a simple old school for loop over properties will work. I'm still holding usage of new "ES6 features" until we reach a point where we can drop Android 4.4 :)

Sounds good then, I'll merge then!

blurymind:
@4ian updated it to use Object.keys instead, added better protection against crashes and updated the example project to show an example of how to load/save bondagejs state from/to a GD variable.

Also caught a last moment bug where a command is not detected or text not terminated. This PR should give our users a much more solid experience using the extension in more advanced ways now. They can pack their dialogue with commands, links and options as well as use some commands that were previously broken but are key to developing a game (load/save game state)

Extensions/DialogueTree/dialoguetools.js
blurymind: updated now

4ian:
Merged! Thanks a lot for doing all these fixes and updates :)

blurymind:
Thank you for the patience in reviewing this :)

blurymind:
btw the new version of yarn has a lot of new features now
https://yarnspinnertool.github.io/YarnEditor/
https://github.com/YarnSpinnerTool/YarnEditor
(new settings dialog, themes, etc) thanks to @daviddq

I might do a pr to update to it in gdevelop at some point

4ian:
Sure thing! Feel free to make a PR to update it! :) 
Looks like a matter of updating Yarn itself, checking that the "bridge" between GDevelop and it is still working and double check that all the new/old features work.

blurymind:
I want to wait for Dave to add some more of the features he's working on. They are going to make working with it so much nicer :)
The performance on huge yarn files has also been massively improved btw

Issue #1669, opened 2020-04-18
Documentation software


arthuro555:
So, it was said many times already, but the documentation system for the GDJS Runtime game engine has to be changed. Until now it was outdated, but ok. But now, the documentation begins to become more complete and includes interfaces. The current documentation is generated using JSDoc and the Jaguar template. Jaguar is not supporting interfaces, making the documentation incomplete. 

The solution told by 4ian would be to switch to another documentation system like TSDoc. This can take a bit of time and the documentation needs to be made accurate as soon as possible.

I recommend switching to another template that support interfaces and is still maintained ASAP as a quick fix before thinking about another documentation software for the long term.

Any opinion on this @4ian ?

4ian:
I would be careful about the premises: having "interfaces" is not a reason to ditch the documentation generator, especially because this is something that is not existing in other type systems (for example in Flow everything is a type - an interface is a type).
Moreover, the only "interface" that we have is for the debugger - which is low in the list of priorities of things to expose in the documentation (it's almost an internal stuff, your game should not rely on this).

I'm happy to be convinced otherwise, but I don't see anything really missing in terms of feature in JSDoc - apart from not being able to type functions with a type written like this ` (args) => returnType` (which is really sad).

I'm afraid we're just losing a lot of precious time to switch to a "better" documentation generator (and find later that there are other issues/not maintained/not updated) while it's not the most valuable thing to do.

The most valuable thing (again, happy to be convinced otherwise) is to make all the **public** stuff documented (I say **public**. Commenting private stuff is less useful and can even expose the wrong signal by having games relying on internal stuffs, which is very dangerous) and typed.
In other words, I think priority is:
1. Typing the game engine
2. Documenting the game engine
3. *maybe* switch to another documentation generator more suited to TypeScript.

arthuro555:
I meant switching to another generator for the long term, after the documentation is complete. What we need to change is the JSDoc template. JSDoc itself is fine, but tha jaguar template isn't mainained since 2017 and doesn't supports interfaces. And we will also have to use interfaces for renderers that have a pixi and cocos variant.

4ian:
If that's a matter of template for now, let's find a better documentation template that supports interfaces then :)

arthuro555:
We could use the default one, but it's very different from the current one and could be missleading for current documentation users. Those are the main other maintained alternatives after researching for a bit:

https://github.com/braintree/jsdoc-template - Claims to be responsive
[...]

PR #1672, opened 2020-04-18
Switch documentation Template and add typescript support to docs

arthuro555:
As discussed in #1669


GDJS/docs/jsdoc.conf.json
```diff
@@ -23,6 +23,8 @@
       "outputSourceFiles": true,
       "includeDate": false
```
4ian, line 24: Can you verify if these options are still necessary?

arthuro555:
image  
That's how braintree looks like.  
image  
And that's better-docs.

I personally prefer better docs because
1. It supports interfaces
2. It supports typescript
3. It supports searching
4. It has that nice navigation toolbar on the right to navigate through a class

And IMO it looks as clean as the other.

GDJS/Runtime/gd.js
```diff
@@ -135,7 +135,7 @@ gdjs.toDegrees = function (angleInRadians) {
  * of the type of the object is "TextObject::Text".
```
4ian, line 135: These changes are not right, the parameters are NOT `gdjs.Runtimeobject`, which would expect an instance of the class, but the class itself.
If it's causing problems, try `Class<gdjs.RuntimeObject>`?

4ian:
I like better-docs, looks clean. We can go ahead with this.

4ian:
Typescript was failing, I fixed this for you. Looks good to go, we can always switch to another theme later as links are not changed.
While browsing the docs, I think we should:
- Add a link in the main page to the Type Definitions, that are hidden in global.
- Hide the Cocos related classes.
- Maybe start to index also the extension, at least the official ones so that people writing JavaScript code can see the documentation of the gdjs.TiledSpriteRuntimeObject, etc...
- Add more `// @ts-check` and fix our types/docs. There are still a lot of bad parameters, missing types, etc! Would be good to make this more professional by completing/fixing stuff progressively.

PR #1676, opened 2020-04-19
[WIP]Check for Special Characters when adding a resource

PawBud:
This is a fix for the issue: _#1551_

The issue mentions that There should be simple warning/error generated if the image name has a special character. @arthuro555 suggested that Instead of keeping this check only for the Sprite Editor i.e. images, We can simply keep it for all the kinds (i.e. fonts, images, audio, video, and JSON). 
The Core compiled successfully but I need the explanation for the doubts I have to solve the issue.
@4ian, I tried a lot to implement the Warning but I couldn't do it. I just have some doubts which I think you can help me with:- 

1. **Is this approach correct in the first place?** 
I misunderstood the meaning of the Getter functions in the `ResourcesManager.cpp` file and hence I couldn't implement it but Arthuro corrected me and told me that this should be done in the Setter function instead. Hence I did the necessary.

2. **Moreover, I can't really understand how to handle the `ReportError()` function and where to handle it?**, I tried to see the `EventsCodeGenerator.cpp` where it has been handled but couldn't come with a satisfactory explanation. Can you help me here?

3. **Is it as simple as to just call the function `ReportError()` in the newIDE & where should I do it in the `SpriteEditor/index.js` or the `ResourcesList/LocalResources.js`?**
I tried to implement it but I couldn't do it in both unless I get the explanation for the former questions.

Thank you very much for considering my questions & taking the time to answer them.

### DIFF
--- a/Core/GDCore/Project/ResourcesManager.cpp
+++ b/Core/GDCore/Project/ResourcesManager.cpp
@@ -19,6 +19,8 @@ namespace gd {
 
 gd::String Resource::badStr;
 
+void ResourcesManager::ReportError() { errorOccurred = true; }
+
 Resource ResourcesManager::badResource;
 #if defined(GD_IDE_ONLY)
 ResourceFolder ResourcesManager::badFolder;
@@ -144,6 +146,15 @@ bool ResourcesManager::AddResource(const gd::String& name,
                                    const gd::String& filename,
                                    const gd::String& kind) {
   if (HasResource(name)) return false;
+
+  for (std::size_t i = 0; i < name.size(); ++i) {
+    if ((name[i] >= 48 && name[i] <= 57) || (name[i] >= 65 && name[i] <= 90) ||
+        (name[i] >= 97 && name[i] <= 122)) {
+      continue;
+    } else {
+      ReportError();
+    }
+  }
 
   std::shared_ptr<Resource> res = CreateResource(kind);
   res->SetFile(filename);
--- a/Core/GDCore/Project/ResourcesManager.h
+++ b/Core/GDCore/Project/ResourcesManager.h
@@ -386,6 +386,14 @@ class GD_CORE_API ResourcesManager {
   std::shared_ptr<gd::Resource> GetResourceSPtr(const gd::String& name);
 
   /**
+   * \brief Report that an error occurred during code generation ( Event code
+   * won't be generated )
+   */
+  void ReportError();
+
+  bool errorOccurred;          ///< Must be set to true if an error occured.
+
+  /**
    * \brief Add an already constructed resource.
    * \note A copy of the resource is made and stored inside the
    * ResourcesManager.


jimil749:
If we are just putting a check for special characters for resources, can't we use `ValidateName` inside `Project.cpp`, which serves the same purpose i.e check if characters other than allowed ones are entered while naming. I am not sure that this is the correct solution or not, just an idea though! :)

4ian:
We should not have to exclude any characters in name of filename (apart from those not supported by the file system, like `/`, `\`, etc... but this is already prevent by the filesystem itself).
If it creates issue somewhere, we have to fix this somewhere, rather than blocking people from using all characters (names are string, in theory they have no limitation) apart from latin alphabet. In theory you can enter names using any unicode character, which is important for people not using English I think.

Note that extension and object names are validated against a stricter subset of characters, because they are used in code. But even for those, we could lift this limitation and generate an identifier that works with any unicode character (for example: `scène` is transformed to `sc_232ne` in the generated code). It's already what is done for scene names.

PawBud:
@4ian 

> We should not have to exclude any characters in name of filename (apart from those not supported by the file system, like `/`, `\`, etc... but this is already prevent by the filesystem itself).
> If it creates issue somewhere, we have to fix this somewhere, rather than blocking people from using all characters (names are string, in theory they have no limitation) apart from latin alphabet. In theory you can enter names using any unicode character, which is important for people not using English I think.

I see your point here, But for the time being should I implement for only '@' since it is clearly causing errors?

PawBud:
> If we are just putting a check for special characters for resources, can't we use `ValidateName` inside `Project.cpp`, which serves the same purpose i.e check if characters other than allowed ones are entered while naming. I am not sure that this is the correct solution or not, just an idea though! :)

You are right indeed :) , I think I tried doing it roughly but It didn't work out. I will try it again properly.

4ian:
> But for the time being should I implement for only '@' since it is clearly causing errors?

I would like to know what is the error exactly behind @. What is the issue exactly, what is breaking when using it?

PawBud:
Alright, so https://forum.gdevelop-app.com/t/solved-bug-in-filename-dont-work-well-the-collision/23371/21. @PascalLadalle
Concluded that any early collision was being triggered due to the presence of “@“ symbol in the resource.

4ian:
I think we should understand better why "@" is a problem. Otherwise, we risk missing the real root cause of the problem, and either introduce artificial limitations or worse, miss another similar bug.

jimil749:
But are we sure that only "@" is causing problem here and not any other special character?

PawBud:
yes as of now, only ‘@‘ is causing an issue in the resources.

4ian:
> then try to see if the resource with the particular character is causing problems Like early collisions

It's unlikely that a character is somehow magically making collisions to fail :) It's surely more a problem of filename/resources that are swapped or messed up.

> yes as of now, only ‘@‘ is causing an issue in the resources. But it may be that other special characters are causing it too

So the answer is for now: "we don't know". And this unknown is the danger: we must find first the exact reason of the bug with the "@". Not sure if testing every unicode character is useful. First there are two many. Second, we don't want to exclude them. We want to fix the underlying issue.

blurymind:
hope this doesn't break emojis :roll_eyes:

4ian:
I'm closing this because there is no activity and we probably don't want restrictions here.

Issue #1218, opened 2019-08-31
Global Variable being "(Structure)" after the child event is removed


Okanegahoshi:
## Describe the bug
When setting Global Variables, add a child variable would make the value become "(Structure)".
But the value remains the same way even if the child is removed.
https://imgur.com/scDUgS9
It doesn't change even after re-open the project.

It should become editable after removing the child.
Ideally, it could even store the value that was there (if any) in case of misclicking.

triptych:
Yes! I ran into this earlier today - would love to see a fix.

4ian:
We might think about improving the UI too to show the variable type (number, text or structure), and allow to switch from a primitive (number, text, in the future boolean) to a structure.
As a first step, removing the only remaining child of a structure should convert it back to a number.

mhk19:
@4ian i would like to work on this issue

4ian:
Feel free to try :)

arthuro555:
With the variable strict typing changing the interface, this issue should be fixed now

4ian:
Yep! :)

PR #1677, opened 2020-04-20
Convert structure back to number when it has no chlidren

Quarkstar:
First step of fixing #1218 
Convert structure back to a number when removing the only remaining child of the structure.

### Screenshot
Fix1218


4ian:
Sounds fair, no particular reason to keep an empty structure.
Thanks! 👍

PR #1681, opened 2020-04-21
Rename parameters of JsCodeEvent

Quarkstar:
- Change the type of `parameterObjects` from `String` to `Expression`
- Add `GetAllExpressionsWithMetadata`

I think we don't need to update js code that are calling `GetParameterObjects` . Or do I miss something? :D

### DIFF
--- a/GDJS/GDJS/Events/Builtin/JsCodeEvent.cpp
+++ b/GDJS/GDJS/Events/Builtin/JsCodeEvent.cpp
@@ -18,9 +18,32 @@ using namespace std;
 
 namespace gdjs {
 
+vector<pair<gd::Expression*, gd::ParameterMetadata> >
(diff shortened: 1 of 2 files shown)


Quarkstar:
CI failed because `GDCore/Tools/VersionPriv.h` file not found.

4ian:
That's a transient issue, don't care about it. I'll take a look at the PR 👍

4ian:
I think this looks all good! 

> I think we don't need to update js code that are calling GetParameterObjects . Or do I miss something? :D

As the interface is still working with a string, it's indeed not changing how the JS works with this class :)

4ian:
I think this looks all good! 

> I think we don't need to update js code that are calling GetParameterObjects . Or do I miss something? :D

As the interface is still working with a string, it's indeed not changing how the JS works with this class :)

PR #1682, opened 2020-04-21
[WIP] Fixed Ordering of Parameter for event: "Test the elapsed time of a scene timer"

### LINKED ISSUE #843: Object & scene timers have different ordering of parameters
## Describe the bug
The 'Value of a timer' and 'Value of a scene timer' have a different ordering of their parameters (e.g. 'Time in seconds' & 'Timer's name') which means switching between them causes a syntax error.

## To Reproduce
1. Add a new scene and go to the Events tab
2. Add an empty event
3. Click 'Add condition'
4. Search for 'timer' and click on 'Value of a timer'
5. Skip the object parameter, but enter 'test' as 'Timer's name' and 1 as 'Time in seconds' - click OK
6. Edit the same condition, search for 'timer' and click on 'Value of a scene timer' (this changes the original condition to the new one, copying over the parameter values)
7. Note that 'Time in seconds' is 'test' (syntax error) and 'Timer's name' is 1

timers

gautamv95:
Fixes #843 

#### Tasks:
- [x] Update the parameters of the action declared in the extension
- [ ] Update the game engine functions
- [x] Add a `UpdateInstructionsFromGD4097` that exchange the parameters
- [ ] Bump the version of GD Core to 4.0.98
- [x] Compile libGD.js and check that when you open a file it's properly updated

@4ian I cannot find game engine functions related to Time extensions, can you guide me?

### DIFF
--- a/Core/GDCore/Events/Serialization.cpp
+++ b/Core/GDCore/Events/Serialization.cpp
@@ -64,6 +64,19 @@ bool AddQuotesToFunctionCall(gd::String& expressionStr,
 
 namespace gd {
 
+void EventsListSerialization::UpdateInstructionsFromGD5092(
+    gd::Project& project, gd::InstructionsList& list) {
+      for(std::size_t i=0;i<list.size(); ++i){
+        gd::Instruction& instr=list[i];
+
+        if(instr.GetType() == "Timer"){
+          std::vector<gd::Expression> parameters = instr.GetParameters();
+          std::swap(parameters[1],parameters[2]);
+          instr.SetParameters(parameters);
+        }
+      }
+    }
+
 void EventsListSerialization::UpdateInstructionsFromGD4097(
     gd::Project& project, gd::InstructionsList& list) {
   for (std::size_t i = 0; i < list.size(); ++i) {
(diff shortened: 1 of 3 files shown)


4ian:
> @4ian I cannot find game engine functions related to Time extensions, can you guide me?

The link between a GDevelop action/condition and the implementation in the engine is made using `SetFunctionName` (in C++ files)/`setFunctionName` (in JS).
In the case of builtin extensions, they are implemented into the platform (GDJS) folder, in the folder "Extensions/Builtin":  https://github.com/4ian/GDevelop/blob/master/GDJS/GDJS/Extensions/Builtin/TimeExtension.cpp

This should give you the name of the function you're searching for :)

gautamv95:
@4ian can you guide me on how do I increase the GDCore version?
I figured it is stored in **VersionPriv.h**, should I make commits to that file?

gautamv95:
@4ian, please review

4ian:
Sorry for the delay! I think this is working fine 👍
The only downside is that this is a breaking change, so I want to wait a bit before merging this - because games opened with the version with this won't be compatible with older version. This can be bad if we release a version that is buggy and ask people to go back to the previous version if we can't release another one quickly.

gautamv95:
@4ian can you point to the code which is responsible for auto-filling the parameters, when clicked on an instruction?

Bouh:
Sorry for the late reply
Here where are filled the parameters in the instruction editor.

https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/EventsSheet/InstructionEditor/InstructionParametersEditor.js#L287-L342

arthuro555:
I propose we make another PR to add compatibility code to core for when this will be added, to have 2-3 version prior supporting the new project file and having a fixed date from which on we can merge this.

PR #1684, opened 2020-04-22
[WIP] Mainframe refactor into functional Component

HarsimranVirk:
Re-opened PR #1601 

EDIT: @4ian there was something weird going on while opening this PR, it didn't have some changes from master, I've added them in the last 1-2 commits, just ignore them :p


newIDE/app/src/MainFrame/index.js
```diff
@@ -1071,249 +1086,254 @@ class MainFrame extends React.Component<Props, State> {
...
+              resource: gdResource,
+              newName: string,
+              cb: boolean => void
+            ) => {
+              // TODO: Project wide refactoring of objects/events using the resource
+              cb(true);
+            }}
+            isActive={isActive}
+            ref={editorRef}
+            onChooseResource={(sourceName, multiSelection, newState) =>
```
4ian, line 1268: newState should not be in the argument of this callback, this will probably be undefined and so pass undefined to _onChooseResource

newIDE/app/src/MainFrame/index.js
```diff
@@ -1353,21 +1374,21 @@ class MainFrame extends React.Component<Props, State> {
...
     extensionName: string,
-    eventsFunction: gdEventsFunction
+    eventsFunction: gdEventsFunction,
+    newState: State = state
   ) => {
-    const { currentProject } = this.state;
+    const { currentProject } = newState;
     if (!currentProject) return;
-    const { eventsFunctionsExtensionsState } = this.props;
+    const { eventsFunctionsExtensionsState } = props;
```
4ian, line 1389: I think you're using the "old props" here, sounds dangerous.
**Don't pass newProps as argument**. Instead, just pass `eventsFunctionsExtensionsState` (by passing `newProps.eventsFunctionsExtensionsState`).

The idea is to pass to the function as little as it needs.

newIDE/app/src/MainFrame/index.js
```diff
@@ -1353,21 +1374,21 @@ class MainFrame extends React.Component<Props, State> {
...
 
-    // Names are assumed to be already validated
+    // Names are assumed to be alreaStartPagedy validated
```
4ian, line 1391: typo here

newIDE/app/src/MainFrame/index.js
```diff
@@ -1522,597 +1544,555 @@ class MainFrame extends React.Component<Props, State> {
...
+        currentProject.getLayoutAt(0).getName(),
+        {
+          openSceneEditor: true,
+          openEventsEditor: true,
+        },
+        newState
+      );
     } else {
-      this.openProjectManager();
+      setState({ ...newState, loadingProject: false }).then(state =>
```
4ian, line 1579: Can you confirm if all the setState({ ...newState, xxx }); could be replaced by setState(state => ({...state, xxx}))?

I know it's slightly longer to write, but remove potential dependencies on "newState", that I would like to remove as much as possible.

HarsimranVirk: Okay, so basically I've passed the newState (even when it's not really needed) to break out of closure. I guess, setState(state => ({...state, xyz, ...rest})) will have the correct state and it should work fine.

I'm not sure about this though, but then again there's no harm in trying it once :)

newIDE/app/src/MainFrame/index.js
HarsimranVirk: This is an obvious mistake 😅 I think I messed up something with find and replace, sorry about that :D

newIDE/app/src/MainFrame/index.js
HarsimranVirk: I didn't realise closure is going to be a problem for props too, until I discovered what happened with the unsavedChanges.

Yes, you're right about this, I need to figure out if stale props are going to be a problem in the entire code, not just this particular place.

HarsimranVirk:
@4ian I'm extremely sorry for the delay, I had some project submissions at the end of last month :sweat:. Now that it's over I can focus here again :)

>  I'd like to see if we can instead, if not too complicated, look at what is really used from the newState/newProps, and pass these as parameters.

I've tried to limit passing of `newState/newProps` at a lot of places and pass only what's actually required. Some of the functions still rely on them to break out of closure though, we could definitely find a better way but let's not break the code for now :sweat_smile:. 

> Can you confirm if all the setState({ ...newState, xxx }); could be replaced by setState(state => ({...state, xxx}))?

I've used lazy initialising at most of the places now, (to be honest, I should have started with this when I knew there's a lot of problems regarding closures, and react guarantees the correct state with lazy initialisation :facepalm: ) Now, everything should be a little better and stable than before because we are using react's way to deal with closures :)

newIDE/app/src/MainFrame/index.js
```diff
@@ -415,136 +494,124 @@ class MainFrame extends React.Component<Props, State> {
...
-        projectManagerOpen: !this.state.projectManagerOpen,
-      });
+  const closeApp = (): void => {
+    return Window.quit();
   };
 
-  openProjectManager = (open: boolean = true) => {
-    this.setState({
+  const closeProject = (
+    newState = {
```
4ian, line 559: This method is never used without params. Can you entirely remove the default arguments? Could you also check all the other methods?

The idea would be to never have default arguments. Why? Because it makes type checking less efficient. 
If you forget to pass arguments:
* if arguments are mandatory, you're getting an alert. Fine!
* if arguments are not mandatory, it's maybe ok if you want the old state... but if you need the new, you don't get an alert... so you have a hard to debug issue, both for your and the reviewer. Dangerous! 
Also make harder to understand when reading the code "why the heck do they sometimes pass the project, and sometimes not?? The method does not always need the project??"

Conclusion:
let's not have default arguments. Instead, let's be explicit and always pass what is needed.

Advantage: we'll progressively remove the notion of "newState". It's not helping to understand a method if I tell someone "closeProject is taking an argument called newState" ("ok sure, and so what it this thing huh?"). But if I tell "closeProject is taking currentProject and editorTabs", it's slightly interesting ("ok, so the function needs to deal with the project and the tabs - make sense I guess").

4ian:
Looks better! I commented on something that looks like "nitpicking" but is I think important: I'd like all functions that need arguments to always have them passed - no default arguments.
This helps understand at the call site what is required for the function and avoid implicit things... and implicit bugs if we forget them, which is easy to do :) 

Apart from that, I think we're reaching a good state!
Still a bit of effort and we'll make a beta version to try this out. If no issues, we'll merge! :)

HarsimranVirk:
In the latest commits:
I've removed most (or maybe all) of the default parameters, and I can see instantly how the IDE helps while calling those functions :D I've also limited newState/newProps where ever it was possible. (ie, the code didn't break) :)

newIDE/app/src/MainFrame/index.js
```diff
@@ -1522,600 +1658,616 @@ class MainFrame extends React.Component<Props, State> {
...
-          })}
-        <ProjectTitlebar fileMetadata={currentFileMetadata} />
-        <Drawer
-          open={projectManagerOpen}
-          PaperProps={{
-            style: styles.drawerContent,
-          }}
-          ModalProps={{
-            keepMounted: true,
-          }}
```
4ian, line 1824: This was added recently and need to be ported :)

4ian:
I've tested and could not find any issue :) 👍 
Would need more extensive testing though, there are a lot of stuff going on in MainFrame.

I've published this branch on https://editor-alpha.gdevelop-app.com - make sure to clear your cache, open then close the tab and re-open it again to have the latest version.
@Bouh If you want to take a look, let us know if there is any thing that looks broken (around MainFrame, so the main dialogs, project manager, etc...)

Bouh:
I'm not saying I've tested everything but i've found nothing. Well done!

HarsimranVirk:
> I've published this branch on https://editor-alpha.gdevelop-app.com 

> I'm not saying I've tested everything but i've found nothing.

I've tested the alpha too, but I also have this internal bias, I'd be testing things I knew had issue (which were largely fixed) and it makes little sense. We'd need a better point of view while testing this as it is clearly a large component. :)

newIDE/app/src/MainFrame/index.js
```diff
@@ -196,76 +201,125 @@ class MainFrame extends React.Component<Props, State> {
...
+    (open: boolean = true) => {
+      setState(state => ({ ...state, introDialogOpen: open }));
+    },
+    [setState]
+  );
+
+  const { integratedEditor, initialFileMetadataToOpen, introDialog } = props;
+  React.useEffect(
+    () => {
+      (() => {
```
4ian, line 227: @HarsimranVirk Is there any reason for this extra function that is self executing? I think this could be removed?

newIDE/app/src/MainFrame/index.js
```diff
@@ -196,76 +201,125 @@ class MainFrame extends React.Component<Props, State> {
...
+            // Enable the GDJS development watcher *after* the extensions are loaded,
+            // to avoid the watcher interfering with the extension loading (by updating GDJS,
+            // which could lead in the extension loading failing for some extensions as file
+            // are removed/copied).
+            setState(state => ({
+              ...state,
+              gdjsDevelopmentWatcherEnabled: true,
+            }))
+          )
+          .then(state => {
```
4ian, line 244: this argument state is probably undefined, because the previous then is not returning anything. Let's remove it.

newIDE/app/src/MainFrame/index.js
```diff
@@ -1353,21 +1476,20 @@ class MainFrame extends React.Component<Props, State> {
...
+    eventsFunction: gdEventsFunction,
+    currentProject: ?gdProject,
+    eventsFunctionsExtensionsState: EventsFunctionsExtensionsState
   ) => {
-    const { currentProject } = this.state;
     if (!currentProject) return;
-    const { eventsFunctionsExtensionsState } = this.props;
 
-    // Names are assumed to be already validated
+    // Names are assumed to be alreay validated
```
4ian, line 1492: alreay => already

newIDE/app/src/MainFrame/index.js
HarsimranVirk: @4ian I don't remember exactly, but I guess this has to do with the whole asynchronous code inside useEffect stuff. Im not sure, but if we want to do async stuff inside useEffect, we have to create a function and call it (or it can be a self-calling function as in this case). 

Also, setState is a bit deceiving here, it looks like the normal setState but it's not. If we don't use a self-calling function, we'll basically writing async code directly inside the useEffect, which is not recommended I believe.

4ian: >  which is not recommended I believe.

I think it's indeed not recommended, but we don't have an async function here :) 
There is no `async () => { ... }` from what I see. Surely something that exist because you made some tries with async functions before?

We do have promises, but it's no problem at all to run them in an effect.

The setState is indeed a "set state with promise", I did not see that the return value was indeed passed. Let's keep state then.

newIDE/app/src/MainFrame/index.js
4ian: I missed than the previous function did not need a return. This is fine then!

newIDE/app/src/MainFrame/index.js
4ian: (so let's keep the state argument, but let's remove the self-calling function because it should not be useful as we don't use async/await - unless I missed something else!)

HarsimranVirk: > Surely something that exist because you made some tries with async functions before?

Yes, now that you've mentioned it, I remember this was an async function :) And I had to remove async functions because of the whole babel transpiling issue :)

Alright, so I guess it's not necessary now. I'll remove it and test it then :+1:

4ian: Sounds good 👍

4ian:
I'll give it another try and if there is nothing crashing or behaving strangely, I'll merge this :)

4ian:
Could not find issues, and I did some refactoring to start breaking the huge state into multiple smaller states. I did most of the dialogs, which seems low risk and already give a bit of fresh air :)

Merging now! Thanks for your patience on this! Will be useful to make `MainFrame` more modular in the future and easily use contexts :)

newIDE/app/src/MainFrame/index.js
```diff
@@ -307,60 +352,91 @@ class MainFrame extends React.Component<Props, State> {
...
         const newProject = gd.ProjectHelper.createNewGDJSProject();
         newProject.unserializeFrom(serializedProject);
-
-        return this.loadFromProject(newProject, fileMetadata);
+        return loadFromProject(
+          newProject,
+          fileMetadata,
+          newState,
+          eventsFunctionsExtensionsState
+        ).then(state => Promise.resolve(state));
```
4ian, line 373: post-nitpicking: you don't need to resolve when you're already in a promise. Also this then is not necessary ;)

HarsimranVirk: Should I open a new PR to change this?

4ian: No I'm on it :)

PR #1686, opened 2020-04-23
Add settings to store initial scene position

### LINKED ISSUE #876: Add setting for initial scene position
As per http://forum.compilgames.net/viewtopic.php?f=19&t=11147, I have a large scene and I've had to change the zoomFactor setting in game.json so as to see it better without manually zooming out every time I open the editor. However there's no way to change the position - currently its centred on the top-left corner and I have to drag to see the full scene.

It'd be nice if there were a setting for this in game.json or even better in the editor UI.

gautamv95:
Fixes #876

### DIFF
--- a/Core/GDCore/IDE/Dialogs/LayoutEditorCanvas/LayoutEditorCanvasOptions.cpp
+++ b/Core/GDCore/IDE/Dialogs/LayoutEditorCanvas/LayoutEditorCanvasOptions.cpp
@@ -21,7 +21,9 @@ LayoutEditorCanvasOptions::LayoutEditorCanvasOptions()
       gridG(180),
       gridB(255),
       zoomFactor(1),
-      windowMask(false) {}
+      windowMask(false),
+      xValue(400),
+      yValue(300) {}
 
 void LayoutEditorCanvasOptions::SerializeTo(SerializerElement& element) const {
   element.SetAttribute("grid", grid);
@@ -35,6 +37,8 @@ void LayoutEditorCanvasOptions::SerializeTo(SerializerElement& element) const {
   element.SetAttribute("gridB", gridB);
   element.SetAttribute("zoomFactor", zoomFactor);
   element.SetAttribute("windowMask", windowMask);
+  element.SetAttribute("xValue", xValue);
+  element.SetAttribute("yValue", yValue);
 }
 
 void LayoutEditorCanvasOptions::UnserializeFrom(
@@ -50,6 +54,8 @@ void LayoutEditorCanvasOptions::UnserializeFrom(
   gridG = element.GetIntAttribute("gridG", 180);
   gridB = element.GetIntAttribute("gridB", 255);
   zoomFactor = element.GetDoubleAttribute("zoomFactor", 1.0);
+  xValue = element.GetIntAttribute("xValue", 400);
+  yValue = element.GetIntAttribute("yValue", 300);
 }
 
 }  // namespace gd
--- a/Core/GDCore/IDE/Dialogs/LayoutEditorCanvas/LayoutEditorCanvasOptions.h
+++ b/Core/GDCore/IDE/Dialogs/LayoutEditorCanvas/LayoutEditorCanvasOptions.h
@@ -49,6 +49,8 @@ class GD_CORE_API LayoutEditorCanvasOptions {
   int gridB;         ///< Grid blue color in editor
   float zoomFactor;  ///< Stores the zoom factor
   bool windowMask;   ///< True if window mask displayed in editor
+  int xValue;
+  int yValue;
 };
 
 }  // namespace gd
--- a/newIDE/app/src/InstancesEditor/FullSizeInstancesEditorWithScrollbars.js
+++ b/newIDE/app/src/InstancesEditor/FullSizeInstancesEditorWithScrollbars.js
@@ -62,6 +62,7 @@ export default class FullSizeInstancesEditorWithScrollbars extends Component<
   componentDidMount() {
     if (this._editor) {
       this._handleViewPositionChange(this._editor.getViewPosition());
+      this._editor.scrollTo(this.props.options.xValue, this.props.options.yValue);
     }
   }
 
@@ -78,6 +79,11 @@ export default class FullSizeInstancesEditorWithScrollbars extends Component<
         }
       }
     );
+    if(this.props.options.xValue !== xValue){
+      this.props.onChangeOptions({
+        xValue:this.state.xValue
+      });
+    }
   };
 
   _handleYChange = (e: any, value: number) => {
@@ -95,6 +101,11 @@ export default class FullSizeInstancesEditorWithScrollbars extends Component<
         }
       }
     );
+    if(this.props.options.yValue !== yValue){
+      this.props.onChangeOptions({
+        yValue:this.state.yValue
+      });
+    }
   };
 
   _setAndAdjust = ({ xValue, yValue }: { xValue: number, yValue: number }) => {


newIDE/app/src/InstancesEditor/FullSizeInstancesEditorWithScrollbars.js
```diff
@@ -62,7 +62,11 @@ export default class FullSizeInstancesEditorWithScrollbars extends Component<
   componentDidMount() {
     if (this._editor) {
       this._handleViewPositionChange(this._editor.getViewPosition());
-      this._editor.scrollTo(this.props.options.xValue, this.props.options.yValue);
+      // $FlowFixMe
```
gautamv95, line 65: Couldn't find a way to fix this, not sure if this is correct

4ian: Flow is surely erroring because `this._handleViewPositionChange` could maybe have been changing `this._editor` to null (yes, unlikely.. but could happen).
The solution is to use a variable for using this._editor, guaranteeing it won't be nulled:

```
const editor = this._editor;
if (editor) {
  this._handleViewPositionChange(...);
  editor.scrollTo(...)
}
```

Core/GDCore/IDE/Dialogs/LayoutEditorCanvas/LayoutEditorCanvasOptions.cpp
```diff
@@ -21,7 +21,9 @@ LayoutEditorCanvasOptions::LayoutEditorCanvasOptions()
       gridG(180),
       gridB(255),
       zoomFactor(1),
-      windowMask(false) {}
+      windowMask(false),
+      xValue(400),
```
4ian, line 25: This could mean that a game with another size than 800x600 is not properly centered at the beginning - could be confusing and not a super user experience even if not a huge deal.

Can we make another boolean variable that is false by default to indicate that the positions are not valid (and we can initialize them to 0) and set to true as soon as it's valid (i.e: changed by the user)?

Core/GDCore/IDE/Dialogs/LayoutEditorCanvas/LayoutEditorCanvasOptions.h
```diff
@@ -49,6 +49,8 @@ class GD_CORE_API LayoutEditorCanvasOptions {
   int gridB;         ///< Grid blue color in editor
   float zoomFactor;  ///< Stores the zoom factor
   bool windowMask;   ///< True if window mask displayed in editor
+  int xValue;
+  int yValue;
```
4ian, line 53: "value" is implied by the fact that... it's a variable, so it's a value of something :)
"x" and "y" is too vague, it's the coordinates of what?
=> prefer `viewXPosition` and `viewYPosition` (and `viewPositionValid` as discussed in another comment)

4ian:
Seems like a good start :) I've added some comments about a few things, let me know if that make sense

Core/GDCore/IDE/Dialogs/LayoutEditorCanvas/LayoutEditorCanvasOptions.h
4ian: (I agree it's a bit of nitpicking, but naming variables properly is important to keep things clear and precise as this can grow with other variable members in the future ;))

gautamv95: modified

Core/GDCore/IDE/Dialogs/LayoutEditorCanvas/LayoutEditorCanvasOptions.cpp
gautamv95: I added a check if `viewPositionValid` is true(false by default), only then `scrollTo` function is called, hope this resolves it

gautamv95:
@4ian is this a breaking change?
Since the games created with the old version of libGD.js won't have `viewPositionValid`, `viewXPosition`, `viewYPosition` in their JSON files.

Core/GDCore/IDE/Dialogs/LayoutEditorCanvas/LayoutEditorCanvasOptions.cpp
```diff
@@ -21,7 +21,10 @@ LayoutEditorCanvasOptions::LayoutEditorCanvasOptions()
       gridG(180),
       gridB(255),
       zoomFactor(1),
-      windowMask(false) {}
+      windowMask(false),
+      viewXPosition(400),
```
4ian, line 25: Let's initialize these to 0, otherwise this gives a false hint that these numbers are some default values - while they are not used because viewPositionValid is false.

Core/GDCore/IDE/Dialogs/LayoutEditorCanvas/LayoutEditorCanvasOptions.cpp
```diff
@@ -35,20 +38,26 @@ void LayoutEditorCanvasOptions::SerializeTo(SerializerElement& element) const {
...
   windowMask = element.GetBoolAttribute("windowMask");
+  viewPositionValid = element.GetBoolAttribute("viewPositionValid", false);
   gridWidth = element.GetIntAttribute("gridWidth", 32);
   gridHeight = element.GetIntAttribute("gridHeight", 32);
   gridOffsetX = element.GetIntAttribute("gridOffsetX", 0);
   gridOffsetY = element.GetIntAttribute("gridOffsetY", 0);
   gridR = element.GetIntAttribute("gridR", 158);
   gridG = element.GetIntAttribute("gridG", 180);
   gridB = element.GetIntAttribute("gridB", 255);
+  viewXPosition = element.GetIntAttribute("viewXPosition", 400);
```
4ian, line 59: Same, let's put 0 as default value, because 400 and 300 are giving false hints that these numbers are important.

newIDE/app/src/InstancesEditor/FullSizeInstancesEditorWithScrollbars.js
```diff
@@ -78,6 +85,14 @@ export default class FullSizeInstancesEditorWithScrollbars extends Component<
         }
       }
     );
+
+    //Prevent reduntant calls to onChangeOptions
+    if (this.props.options.viewXPosition !== xValue) {
+      this.props.onChangeOptions({
+        viewXPosition: this.state.xValue,
+        viewPositionValid: true,
```
4ian, line 93: Also set the viewYPosition

newIDE/app/src/InstancesEditor/FullSizeInstancesEditorWithScrollbars.js
```diff
@@ -95,6 +110,14 @@ export default class FullSizeInstancesEditorWithScrollbars extends Component<
         }
       }
     );
+
+    //Prevent reduntant calls to onChangeOptions
+    if (this.props.options.viewYPosition !== yValue) {
+      this.props.onChangeOptions({
+        viewYPosition: this.state.yValue,
+        viewPositionValid: true,
```
4ian, line 118: Also set the viewXPosition

4ian:
> Since the games created with the old version of libGD.js won't have viewPositionValid, viewXPosition, viewYPosition in their JSON files.
> Should I add a check for that as well?

No need for it, because the default values will be false for `viewPositionValid`.

You handled properly the update of the options when a scrollbar is moved, but you should also save the options when the view position is updated in `onViewPositionChanged`, which means the view was moved for another reason than the scrollbars (for example, pinch on touch screens, or space + drag'n'drop) :)

gautamv95:
> No need for it, because the default values will be false for `viewPositionValid`.

Okay great :)

> You handled properly the update of the options when a scrollbar is moved, but you should also save the options when the view position is updated in `onViewPositionChanged`, which means the view was moved for another reason than the scrollbars (for example, pinch on touch screens, or space + drag'n'drop) :)

Will take a look

gautamv95:
My initial commit didn't actually work on touch devices, forgot some changes, fixed them now

newIDE/app/src/InstancesEditor/PinchHandler.js
```diff
@@ -142,6 +151,7 @@ export default class PinchHandler {
     this._lastPinchCenterY = centerY;
 
     this._setZoomFactor(this._getZoomFactor() * scale);
+    this._setViewPosition(this._viewPosition);
```
4ian, line 154: I think this function is not properly named, and doing too much.

Think of how the class is used:
The parent, `InstancesEditor`, must create `PinchHandler` by passing it a viewPosition. This means that InstancesEditor knows, and actually **owns** `viewPosition`.
Now it must also pass it something called "setViewPosition" used to... set a new ViewPosition? But why it would need to do that, because `InstancesEditor` is actually the **owner** of `ViewPosition`.

You probably did this because you observed that `onViewPositionChanged` is passing the ViewPosition as argument. This is right... but `onViewPositionChanged` is a prop of InstancesEditor! This means that it needs to pass ViewPosition because the parent of InstancesEditor does not have access to ViewPosition - because it's owned by InstancesEditor.

Here, it does not really make sense to pass to PinchHandler a function that takes a ViewPosition. Think like the parent (InstancesEditor): "why are you passing my a ViewPosition? I know ViewPosition, it's mine! It's me that passed it to you, why are you giving it back in your callback? I don't even need it, leave me in peace!" ;) 
Or think like the child (PinchHandler): "why are you asking me for the ViewPosition? It's you that passed it to me!".

Instead, what you want is to get notified when the view is **moved**. And PinchHandler is moving the position, by calling scrollBy.
The solution:
- pass a function called `onViewPositionChanged` to PinchHandler, which is called (as the name implies) "when the view position is changed". But without argument: `_onViewPositionChanged: () => void `. 
  Then it's the responsibility of the parent to react accordingly, providing more info if needed to it's parent. In other words, in InstancesEditor, you do:
  ```
  onViewPositionChanged: () => { this.props.onViewPositionChanged(this.viewPosition) }
  ```

*Actually* there is maybe a better solution. Why is InstancesEditor having to call manually onViewPositionChanged, and now why is PinchHandler having to call its own onViewPositionChanged every time we call `scrollBy` (and `scrollToInstance`)?

It's manual and error-prone, we could forget it in the future... instead why not give this responsibility to... ViewPosition.

```
    this.viewPosition = new ViewPosition({
      initialViewX: project ? project.getGameResolutionWidth() / 2 : 0,
      initialViewY: project ? project.getGameResolutionHeight() / 2 : 0,
      width: this.props.width,
      height: this.props.height,
      options: this.props.options,
      onViewPositionChanged: () => { this.props.onViewPositionChanged(this.viewPosition); }
    });
```

and then you can call `onViewPositionChanged` inside `scrollBy` and `scrollToInstance`.
And you remove all the other calls to `this.props.onViewPositionChanged` - they are now redundant because ViewPosition is doing the job of notifying after any change.

Hope this make sense. :)

arthuro555:
@gautamv95 do you plan on finishing this PR?

Silver-Streak:
Checking in on old PRs and issues for some cleanup.

I *think* this is already completed. When I open a scene now, it shows up in whatever position I last hit save on.  Do we know if this is still needed @arthuro555 @4ian ?

Or are we good to close this and https://github.com/4ian/GDevelop/issues/876 out?

PR #1688, opened 2020-04-24
[FIX] BBtext: color, fontSize, fontFamily 

Bouh:
The actions for change the color, fontSize, fontFamily dosen't work, this patch the bug.

Off topic:
The text with custom font are good in the scene editor. But not in the preview/export.

The red is a sprite for wrap the text, there is no event, just scaled manually for fit the text.

image

Without font it's ok:
image

I thought code from the extensions was run for both the games and the publisher. So there shouldn't be any difference, right?

### DIFF
--- a/Extensions/BBText/bbtextruntimeobject-pixi-renderer.js
+++ b/Extensions/BBText/bbtextruntimeobject-pixi-renderer.js
@@ -76,6 +76,7 @@ gdjs.BBTextRuntimeObjectPixiRenderer.prototype.updateText = function() {
 
 gdjs.BBTextRuntimeObjectPixiRenderer.prototype.updateColor = function() {
   this._pixiObject.textStyles.default.fill = this._object._color;
+  this._pixiObject.dirty = true;
 };
 
 gdjs.BBTextRuntimeObjectPixiRenderer.prototype.updateAlignment = function() {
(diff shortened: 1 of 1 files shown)


4ian:
Thanks for the fixes! 👍👍

> I thought code from the extensions was run for both the games and the publisher. So there shouldn't be any difference, right?

What do you mean by publisher?

> The text with custom font are good in the scene editor. But not in the preview/export.

Yeah I think we have an issue somewhere describing this, but I could not figure the difference. Surely a way about how the font is loaded?

Bouh:
~~publisher~~ preview
I thought code from the extensions was run for both the exported games and the preview. So there shouldn't be any difference, right?

>Surely a way about how the font is loaded?

Yes maybe, i'll take a look if i found something.

4ian:
> I thought code from the extensions was run for both the exported games and the preview. 

Yes, preview and export are based on the same code. The preview is itself very similar to an export, it's almost the same process :) 

> So there shouldn't be any difference, right?

In theory no. In practice, the difference in the browser version used (for HTML5 export), the Electron version (for win/macos/linux) or the Chrome browser (for Android export) could lead to subtle differences in case something breaking was made or is supported differently across browser or browser versions.

PR #1124, opened 2019-06-22
Fix various CSS

Bouh:
I've add color for display separators 
For default and dark theme

image

PR #1690, opened 2020-04-25
Feature/responsive layers editor

4ian:
Reworked layers editor to be responsive, made Instances list a panel (no more drawers in the app, except for the Project Manager), re-ordered icons in the scene editor toolbar.

Updated:
image


Bouh:
Cool things here.

Time to share my feedback.

-I had in a [previous PR](https://github.com/4ian/GDevelop/pull/1124#issuecomment-505190900) added a border on the horizontal axis. This one can be removed.
- In the dark mode the border are very present, It's ridiculous, but I feel a bit more closed, for an open source/free software, that is the height. Maybe an color a little bit less visible ?

- Search bars are completely in the mass. there is no separation with the list above. 
Just add in the search bar this color     `background-color: #ffffff0d;`

image

- The spacer in the toolbar are almost not visible. 1px of width is short.
image

- The persistence of the interface is really great :D  !

- The Layers is open at the bottom of the window.I don't know what I thinking about it because now the panel is always displayed. While the users don't go in it as often as in the list of objects or properties. The panel is sightly better than the drawer, but it encumber the UI a bit more for something used infrequently.

It's possible to reduce the gap limit on the mosaic panels ? 
image

- The scroll bar on the layer list is always visible, event if the list don't fill the height of the panel.
(btw nice responsive icons/menu on each layer 👍 )

- When i move the panel between the toolbar and the object list, then i close the layer panel, and open it again, the panel in now at the default position, the persistence don't save my new position as default ?

4ian:
> In the dark mode the border are very present, It's ridiculous, but I feel a bit more closed, for an open source/free software, that is the height. Maybe an color a little bit less visible ?

You mean this border?

image

I got some feedbacks about the software "missing borders" in general and it's true that everything is a bit blended. The new borders seems "heavy" but that might be because we are so used to "not have borders" at all. And I think it's at least very clear for new users the difference between panels.

> Search bars are completely in the mass. there is no separation with the list above

You're right, though for these search bars I was not too concerned because I don't want them to distract too much from the rest of the interface... 
We could use a different color like the one you suggested - but I'm a bit afraid about introducing yet another color in the chart?
OR we could also maybe just add a separating border? :) 

> The spacer in the toolbar are almost not visible. 1px of width is short

Yes, let me look, is it the color of the dividers that is too dark?

> The persistence of the interface is really great :D !

Thanks! I think that this is actually invaluable for power users.

> The panel is sightly better than the drawer, but it encumber the UI a bit more for something used infrequently.

I think it's still better yeah, and I think that it's fair to have it taking a bit of space on the UI because if you're opening it you probably want to work for a few minutes on layers (add/move a layer or work on effects). It's very easy to close it with the close button - something that you had to do with the drawer *every time* you make a single change to launch the preview :)

> It's possible to reduce the gap limit on the mosaic panels ?

Maybe something to investigate in [react-mosaic](https://github.com/nomcopter/react-mosaic). There seems to be a minimal size for panels.

> The scroll bar on the layer list is always visible, event if the list don't fill the height of the panel.

Fixed.

> (btw nice responsive icons/menu on each layer 👍 )

Thanks! 

> When i move the panel between the toolbar and the object list, then i close the layer panel, and open it again, the panel in now at the default position, the persistence don't save my new position as default ?

Right, I never thought about this - this is because react-mosaic is storing the position of the panels as "nodes", but only for the visible panels. Hidden panels are not stored in this. This is probably something that could be an improvement in react-mosaic - being able to keep the position of panels but hide them.

> As you often say, it's nitpicking, that's what makes the difference. (Yes I stole your sentence haha)

;)

4ian:
I gave a few tries to a smaller search box, with a white or input-like background (just for default theme, I would probably use the color you suggested for dark theme):

Gray:
Screenshot 2020-04-26 16 10 16
Screenshot 2020-04-26 16 10 28
Screenshot 2020-04-26 16 10 39

White:
Screenshot 2020-04-26 16 11 27
Screenshot 2020-04-26 16 11 36
Screenshot 2020-04-26 16 12 35

What do you think? Could be done in a separate PR.

4ian:
I couldn't help and went ahead and reduce the SearchBar height. It was taking the same height as a object/layer item and more than an item in the instances list - too much.
Now it's smaller, but also has a different background so easier to distinguish from the rest.

image
image

This means that what you interact with mostly (objects, inputs) are still bigger, but the search is smaller. It's also the case in dialogs where the is a search bar. I think it's fine and taking less screen size. Search should also be always focused when opening a dialog with a search bar - so it's also "visible" by the fact that it's the thing with the focus :)

Bouh:
- The little search bar, and the color is good now :)

Borders are good to differentiate very different things. I see on your pictures new borders on lists, it doesn't exist. I've found borders only on photoshop, and it's not very visible.

image

All others don't use border in list. Just a selection background.
image

I find that a lot of things are very large compared to other software.
Especially the panels, they're very wide, you could easily reduce the size by 1/4. I know that for me it's a reflex to reduce everything, because having empty space in an interface is useless.

image


Compare to other software the button can be reduce by 2.
image

Open your softwares on your mac, and take a look on spaces, and where are the borders.

Bouh:
Btw for the scrollbar in list, the native scroll bar on mac is pretty on the light theme. 
Using [a custom scroll bar](https://www.w3schools.com/howto/tryit.asp?filename=tryhow_css_custom_scrollbar) give a better integration in the UI. And we can use the colors from the theme. There is lot of thing to said for make the UI in GD better. 

I'm just giving my opinion. That's why I asked for some feedback on the discord.

> Quick question how do you like the GD interface?
> From the point of view of the size of the elements, lists, colors, buttons, borders.
> Are there things you don't like?
> Things that are not very visible? Things that are too big?
> Things that are better in other software? (if yes, please give us an example) 
> I'm looking for constructive feedback.

4ian:
> Using a custom scroll bar give a better integration in the UI. And we can use the colors from the theme.

Well it's because I configured my macOS to be light, and for scrollbars to be always shown, but people using the system dark theme will get a dark scrollbar :) And by default scrollbars are hidden on macOS, giving this result: 

image

Notice how the scrollbar is superimposed on top of the content and semi translucent :) 

In general I would rather avoid customising the scrollbar and would trust the UI to provide the proper ones (it's just that in my screenshot, my system is not adapted to the dark theme).

> I'm just giving my opinion. That's why I asked for some feedback on the discord.

Cool thanks :)

Bouh:
All the popular softwares use a custom scrollbar ;)
Because the default scroll bar don't fit to the theme.
And never trust the UI provide by the user because, even in dark mode the color is broken. x)

image

4ian:
Ok might be worth thinking about a custom one then :)

Bouh:
> Ok might be worth thinking about a custom one then :)

[Take a look here](https://github.com/Bouh/GDevelop/commits/cosmetic-improvement), there are other stuff with it.

4ian:
> Open/close the different section inside the project manager. Atleast in the same session

Indeed, that was infuriating! :) Fixed in 2346e419361f79be55e7f6d9045b6c4daa82836b (not persisted across restarts - but not a big deal I think)

blurymind:
It seems to be taking a lot of horizontal space without using it efficiently

4ian:
Mmmm not sure if it's clear how it works, but this does not change anything to the default editor layout. It's just that these editors are now in panels, like the other. And they can be closed easily or moved to the right or wherever you prefer, and your layout is persisted so you don't have to redo it :)

4ian:
I've merged because persistence of layout and getting rid of drawers is important. We can do scrollbar in follow up commits :)

PR #1691, opened 2020-04-25
Alert user when closing a dialog, if he has unsaved changes

gautamv95:
This is in continuation of #1522 

I have added the UnsavedChangesDialog, and currently calling only in ObjectEditorDialog with 4 different methods. None of those methods are perfect, and I need some help to improve them.

Explanation:
I have added a new State property called `hasUnsavedChanges`(false by default), and have different methods to make it true.

#### 1) Use ComponentDidUpdate
Whenever a text field is changed, `hasUnsavedChanges` will be set to true using CDU, and UnsavedChangesDialog can be mounted.

**Limitations (found in all methods)**
If the user, changes the object name(the only text field in this case), presses cancel and then changes it back to the original name, we get an error saying: "Another object or group with this name already exists".

#### 2) Using a new state like object
Create a new object, similar to state but containing only text fields, and dropdowns, then compare the new object properties with the state properties when the user tries to close the dialog, if they are different set `hasUnsavedChanges` to true.

**Limitations**
1) Method 1 limitation
2) This method would require us to write a lot of redundant code as we would have to compare each property separately, and would also increase the memory size, due to the creation of a new Object.

#### 3) Use setTimeout
Call setTimeout when the component has mounted, for some seconds(the time a user takes, to make some significant changes), and set `hasUnsavedChanges` to true.

**Limitations**
1) Method 1 limitation
2) We are assuming that the user will take the time that was hardcoded by us to make some significant change in the dialog, this does not guarantee proper functioning and may lead to bad user experience.

#### 4) Hardcode `hasUnsavedChanges` to true
This would show the UnsavedChangesDialog every time, we open the ObjectEditorDialog.

**Limitations**
1) Method 1 limitation
2) Can cause uncontrolled behavior and may lead to bad user experience

Need opinions on this 😕😕

### DIFF
--- a/newIDE/app/src/ObjectEditor/ObjectEditorDialog.js
+++ b/newIDE/app/src/ObjectEditor/ObjectEditorDialog.js
@@ -26,21 +26,25 @@ export class ObjectEditorDialog extends Component<*, StateType> {
     unsavedChangesDialogOpen: false,
   };
 
-  //Continer for all textfields, dropdowns in a dialog component
-  textFields = {};
+  timeOut = null;
 
   _onChangeTab = (value: string) => {
     this.setState({ currentTab: value });
   };
 
   componentDidMount() {
-    this.textFields.newObjectName = this.state.newObjectName;
+    this.timeOut = setTimeout(() => {
+      this.setState({ hasUnsavedChanges: true });
+    }, 8000);
+  }
+
+  componentWillUnmount() {
+    clearTimeout(this.timeOut);
   }
 
   _openUnsavedChangesDialog = (val: boolean) => {
-    if (this.textFields.newObjectName !== this.state.newObjectName) {
+    if (!!this.state.hasUnsavedChanges) {
       this.setState({
-        hasUnsavedChanges: false,
         unsavedChangesDialogOpen: val,
       });
     } else this.props.onCancel();
--- a/newIDE/app/src/ObjectEditor/ObjectEditorDialog.js
+++ b/newIDE/app/src/ObjectEditor/ObjectEditorDialog.js
@@ -16,6 +16,8 @@ import UnsavedChangesDialog from '../UI/UnsavedChangesDialog';
 type StateType = {|
   currentTab: string,
   newObjectName: string,
+  hasUnsavedChanges: boolean,
+  unsavedChangesDialogOpen: boolean,
 |};
 
 export class ObjectEditorDialog extends Component<*, StateType> {
--- a/newIDE/app/src/ObjectEditor/ObjectEditorDialog.js
+++ b/newIDE/app/src/ObjectEditor/ObjectEditorDialog.js
@@ -26,18 +26,19 @@ export class ObjectEditorDialog extends Component<*, StateType> {
     unsavedChangesDialogOpen: false,
   };
 
+  //Continer for all textfields, dropdowns in a dialog component
+  textFields = {};
+
   _onChangeTab = (value: string) => {
     this.setState({ currentTab: value });
   };
 
-  componentDidUpdate(prevProps, prevState) {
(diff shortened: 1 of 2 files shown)


4ian:
I think we have to differentiate `unsavedChanges`, which is about tracking if user should be warned about unsaved changes when closing the whole app - or switching to another project and a potential new confirmation dialog in case someone spent too much time in an "editor" dialog => there no need for unsaved changes I would say.. just a setTimeout would work :) 

Even if we can't tell if a text field or other is modified - not a big deal I would say.

4ian:
So in short I would be for setTimeout, unless you have a clever change of detecting changes in text fields. Also I would do this at the level of Dialog.

gautamv95:
IMO, I found Method 1 to be the most responsive method among all others since it does track all text fields/dropdowns, etc and displays the confirmation dialog accordingly.

gautamv95:
@4ian any final thoughts, so I shall start implementing this?

4ian:
I think we want:
- something generic (implemented in `Dialog` probably)?
- not related to `UnsavedChanges` (it's something different that is the responsibility of the editor).
- that is showing a modal to the user BUT must not be too intrusive. 
  For example I don't want to get a modal everytime I press escape if I entered just one or two characters. 
  If I made a mistake, I can always re-open the modal - no big deal!.. but I don't want to be asked if I want to confirm the closing of every modal.. I want ideally the confirmation dialog to be disaplyed if I made a *lot* of changes during a long time.

Knowing if changes happened is hard, so a timeout can be a good approximation.

gautamv95:
> * something generic (implemented in `Dialog` probably)?

Will work on it :)
> * not related to `UnsavedChanges` (it's something different that is the responsibility of the editor).

I needed a dummy variable which will turn true, after a change has occurred or after some time has passed(Method 3), that's why I chose `hasUnsavedChanges`, should I rename it to be more specific like `hasUnsavedChangesInThisDialog` or separate it from `UnsavedChanges`?
> * that is showing a modal to the user BUT must not be too intrusive.
>   For example I don't want to get a modal everytime I press escape if I entered just one or two characters.
>   If I made a mistake, I can always re-open the modal - no big deal!.. but I don't want to be asked if I want to confirm the closing of every modal.. I want ideally the confirmation dialog to be disaplyed if I made a _lot_ of changes during a long time.
> 
> Knowing if changes happened is hard, so a timeout can be a good approximation.

Good Example, I think method 3 will certainly be more appropriate :)

4ian:
> that's why I chose hasUnsavedChanges, should I rename it to be more specific like hasUnsavedChangesInThisDialog or separate it from UnsavedChanges?

Name is up to you, just be sure it represents what you're storing in this variable.
It's not unsaved changes, so it must be something else :) I'll let you try to find something.

gautamv95:
> Name is up to you, just be sure it represents what you're storing in this variable.
> It's not unsaved changes, so it must be something else :) I'll let you try to find something.

Looks Interesting 🤔🤔 I'll try :)

Bouh:
I'm for closing this draft PR.

Another feature has been added https://github.com/4ian/GDevelop/pull/2484

This move the problem of losted changes to the choice that the user can make for the backdrop behavior with <Dialog>

The original problem was that people miss click the buttons, and click on backdrop that close without save changes.

[The new option solves this problem indirectly.](https://github.com/4ian/GDevelop/pull/2484
)
Now depending on the user's choice he will be aware that he will lose his content if he closes by clicking in the backdrop.

If he is clumsy with the precision of his mouse cursor he can choose that the background click does nothing, or that the failed click applies the changes.

image

4ian:
Yes this is now outdated and we by default don't automatically close when clicking outside a dialog, like on native desktop softwares.

PR #1694, opened 2020-04-27
Firebase Extension

arthuro555:
Contains:
- [x] Base SDK
- [X] Analytics
- [X] Remote configuration
- [x] Authentification
- [x] Remote performance measurer
- [x] Firebase Functions
- [x] Database (Firestore)
- [x] Online Filesystem (Storage)
- [x] Alternative database (Realtime Database)

Not part of the PR:
- ~~Exporter to Firebase Hosting~~ More complicated than expected, will be in a future PR as this one is very huge already
- ~~Messaging~~ Requires generating a service worker from events, wich is too difficult for now.
- Mobile extra support maybe one day in another PR.


Extensions/Firebase/firebasetools.js
```diff
@@ -0,0 +1,99 @@
...
+ * Firebase Event Tools
+ * @namespace
+ */
+gdjs.evtTools.firebase = {};
+
+/**
+ * Firebase Analytics Tools
+ * @namespace
+ */
+gdjs.evtTools.firebase.analytics = {};
```
arthuro555, line 36: I will probably move every separate module of the firebase SDK to their own "tools.js" to prevent exporting useless code

Extensions/Firebase/firebasetools.js
```diff
@@ -0,0 +1,99 @@
...
+
+/*    Function Definitions    */
+
+/**
+ * Enables Firebase Analytics and cache the instance to runtimeGame.
+ * @param {gdjs.RuntimeScene} runtimeScene The current {@link gdjs.RuntimeScene} instance.
+ */
+gdjs.evtTools.firebase.analytics.enable = function(runtimeScene) {
+    runtimeScene.getGame()._analyticsInstance = runtimeScene.getGame()._analyticsInstance || firebase.analytics();
```
arthuro555, line 45: I am not sure if it is faster to access the instance through this "cache" and I find that kinda unclean.

newIDE/app/src/ProjectManager/ProjectPropertiesDialog.js
```diff
@@ -417,6 +422,18 @@ class ProjectPropertiesDialog extends React.Component<Props, State> {
...
             onChange={value => this.setState({ adMobAppId: value })}
           />
+          <Text size="title">
+            <Trans>Firebase</Trans>
+          </Text>
+          <SemiControlledTextField
+            floatingLabelText={<Trans>Firebase Configuration string</Trans>}
+            fullWidth
+            rows={8}
+            multiLine
```
arthuro555, line 432: btw on the official material-ui multiLine is actually completely lowercase, and here the L is uppercase. That's an inconsistency and made me wonder for half an hour why it wasn't working until Bouh helped me by telling me to make the L Uppercase

arthuro555:
Mhhh the CI failure doesn't seem related to me. For app veyor It's something with browser storage, definitely didn't touch that, and for semaphore some weir network problems from npm

Extensions/Firebase/firebasetools.js
```diff
@@ -0,0 +1,99 @@
+/**
+ * Firebase Tools
+ * @author arthuro555
+ */
+
+/*    Initialization of the firebase SDK    */
```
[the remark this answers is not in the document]
4ian: Can you use `registerFirstRuntimeSceneLoadedCallback` to avoid having a "gdjsWatcher"?

Extensions/Firebase/firebasetools.js
```diff
@@ -0,0 +1,99 @@
...
+
+/** A Timeout function is used to initialize firebase as soon as the project data has been loaded */
+let gdjsWatcher = function() {
+    if (gdjs.projectData === undefined) {
+        setTimeout(gdjsWatcher, 1);
+        return;
+    }
+    let firebaseConfig;
+    try { 
+        firebaseConfig = JSON.parse(gdjs.projectData.properties.firebaseConfig);
```
4ian, line 16: Prefer using [`runtimeScene.getGame().getGameData()`](https://docs.gdevelop-app.com/GDJS%20Runtime%20Documentation/gdjs.RuntimeGame.html#getGameData).

Extensions/Firebase/firebasetools.js
4ian: That sounds fine 👍

Extensions/Firebase/firebasetools.js
4ian: > I am not sure if it is faster 

I would say to avoid premature optimisation if it's at the cost of unclean code. Consider runtimeGame as something that should not be touched (if someone does the same for another extension with analytics - boom: either your extension or the other is now broken).

Instead:
- just use firebase.analytics()
- if you prefer to keep a reference, store it in your extension: `gdjs.evtTools.firebase.analytics._analyticsInstance`.

I would suggest to see if you can initialize this at the beginning of the first scene, so that latter you can just do:
`var analyticsInstance =  gdjs.evtTools.firebase.analytics._analyticsInstance;`
and use `analyticsInstance` in your function (so that it's both short and efficient).

newIDE/app/src/ProjectManager/ProjectPropertiesDialog.js
4ian: It's because historically material-ui v0 used `multiLine`, so when converting to Material-UI v4 the props name were kept - too much stuff to port already.

But feel free to submit a PR to rename this in the whole codebase.

Extensions/Firebase/firebasetools.js
arthuro555: I would like but I don't have access to runtimeGame here. Should I switch to a first scene loaded callback?

arthuro555:
The functions are not very intuitive so if it gets merged into master I will work hard on the documentation on the wiki

Extensions/Firebase/firebasetools.js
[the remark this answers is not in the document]
arthuro555: I was actually tlking about the big comments for structuring but yes, except I'm unsure if it wouldn't cause problems because afaik the firebase sdk loads asynchronously so it should load as soon as the data is provided to be assured to finish loading before being iused by events.

arthuro555:
Hey @4ian! I need your advice on a little problem. On electron, I cannot use authentification providers other than email the "standard way" (using a pop-up or something like that). After a bit of research it's because oauth2 needs a valid origin, wich file URIs aren't. And electron loads files through file:// by default. The workaround seems to be to start a little webserver (express) hosting the app's files for electron to load (http://localhost is a valid OAuth2 Origin apparently). The thing is there are following potential problems:
- Port already taken (need to find an unused port)
- Performance loss due to having a webserver running in parallel
- Firewall issues
- Potential unwanted leak of the game with an open port

Do you think I should
1. Use that as a last resort
2. Just use that as it seems to be the most recommended method after researching a while
3. Never ever do that

I look forward to your answer :)

Thank you for dedicating so much time for this awesome project.

arthuro555:
I think the best would be to show it as semi supported:
Either:
- Warn that it won't work in electron and give a guide to change the manual exported files to use that webserver aproach.
- Give a preference to enable this webserver aproach and warn thew user about the potential issues mentioned above when they check it.

4ian:
If you run the Network preview, there is actually a tiny web-server ran and that should allow you to do progress on your integration:
image

It's run on `localhost:2929` (arbitrary port number). If not available, a new port is tried until one is found.

So we could:
* Make this the default, using always a web-server for previews. But with the drawbacks you said: firewall permission issue, slight performance loss vs loading from filesystem directly.
  * This could also confuse people exporting to Electron because things that work in preview like Firebase would not work when exported to Windows/macOS/Linux without a webserver?
* Or make a warning shown to the user when the requirements are not fulfilled ("To run authentication with other providers, please run a Network preview or export your game on a web-server").

I would recommend that you first try and develop with Network preview. If it works, we can later adapt the IDE so that when we detect a certain condition (usage of this extension, etc...) we display a warning when running the preview, maybe with even a button to launch a Network preview instead.
We can also think about allowing exported Electron games to embed (optionally) the same tiny web-server so that games can be run on it. But that's additional complexity so better be sure it's useful and have the extension ready before.

arthuro555:
My bad, after further testing most of the providers detect eletron and block authentification in it, even with the webserver aproach. I have to make authentification happen in browser. I think I should host a webserver for authentification like 
```
gdjs.evtTools.blablaAuthFuntion() {
    try {
        let express =require("electron").remote.express (include express wen firebase is present)
        let port = express.adress().port;
        express.blblaSetupTheWebserverToSendAnAuthThing()
        // Insert Set up of some websocket thing to comunicate with the page in browser
       window.open("http://localhost:"+port)
    } catch { // No electron
        firebase.normalAuthFunctionBlahblah()
    }
}
```
Or maybe I should just not support OAuth in Electron? It would confuse users but make things easier. I mean there is probably a reason it's not supported.

4ian:
>  I think I should host a webserver for authentification like

I'm not entirely sure, but you might need to open a separate window where all Electron and Node.js integration are removed. It's surely a safety/security concern. The risk with Electron is:
- when you display 3rd party content, any website that you display can `require('electron').remove.whatever` and erase your hard drive (or worse, read it).
- when you are a 3rd party service provider, like an authentication provider, you don't want your authentication page to be modified or "hijacked" and have your users tricked into entering their password - and providers might detect Electron or such environment where it's easy to modify the environment.

People working on Electron are conscious of these limitations and are progressively trying to isolate risky APIs - and discourage the use of "remote", instead asking to rely on events (so that all the Node.js code is inside the main process, so nobody can run arbitrary Node.js code if for some reason a 3rd party website is shown and some JavaScript is injected).

See this page, lots of interesting reads: https://www.electronjs.org/docs/tutorial/security#15-disable-the-remote-module

> Or maybe I should just not support OAuth in Electron? It would confuse users but make things easier. I mean there is probably a reason it's not supported.

Yep, let's put this aside for now, you can focus on something else.

arthuro555:
(Sorry I marked it as ready to see where the conflict was)

Extensions/Firebase/firebasetools/D_authtools.js
```diff
@@ -0,0 +1,312 @@
...
+         * Changes the user's password.
+         * Use this when using an external provider.
+         * @param {string} newPassword - New Password for the user.
+         * @param {gdjs.Variable} callbackStateVariable - The variable where to store the result.
+         */
+        changePasswordProvider(newPassword, callbackStateVariable) {
+            gdjs.evtTools.firebase.auth.currentUser.reauthenticateWithPopup(gdjs.evtTools.firebase.auth._currentProvider)
+              .then(() => gdjs.evtTools.firebase.auth.currentUser.updatePassword(newPassword))
+              .then(() => callbackStateVariable.setString("ok"))
+              .catch(error => callbackStateVariable.setString(error.message));
```
4ian, line 134: In other extensions doing some kind of asynchronous operations (like the Facebook Instant Games API), the convention was to use a first variable to store the result and another variable to store the error, if any.

Not sure if it makes sense in your case - but think about it if you want to provide the same convention :)

arthuro555: 🤔 That would make lots of arguments and look maybe too intimidating?

arthuro555:
I began a draft Wiki page: http://wiki.compilgames.net/doku.php/gdevelop5/all-features/firebase

Extensions/Firebase/firebasetools/D_authtools.js
```diff
@@ -246,7 +246,7 @@ gdjs.evtTools.firebase.auth.userManagement = {
 
     /**
      * Send an email to the user's email adress to verify it.
-     * @note placeholder: will be called directly when declared
+     * @note Even though this function is redundant, we keep it for consitency.
```
4ian, line 249: nit: consistency

Extensions/Firebase/firebasetools/D_authtools.js
4ian: Maybe. Let's go with a single "state" variable for now then. It's ok here because what is updated is a global "authenticated" state.

If you need in another action to communicate both a result and a state as variables, then in this case do two variables (one containing the result, if any, and another one containing the state) (i.e: don't mix them).

4ian:
Thanks for the wiki page - for a "complex" (read: not that easy to use) extension, documentation is invaluable for users :)

arthuro555:
> Thanks for the wiki page - for a "complex" (read: not that easy to use) extension, documentation is invaluable for users :)

It's really more a placeholder than real doc for now

newIDE/app/src/ProjectManager/ProjectPropertiesDialog.js
```diff
@@ -420,6 +426,43 @@ class ProjectPropertiesDialog extends React.Component<Props, State> {
...
             />
+            <Text size="title">
+              <Trans>Firebase</Trans>
+            </Text>
+            <SemiControlledTextField
+              floatingLabelText={<Trans>Firebase Configuration string</Trans>}
+              fullWidth
+              rows={8}
+              multiline
+              hintText={`{
```
arthuro555, line 437: The presence of curly braces inside the text causes an error message to appear in the console (even tho everything works properly). Any Idea on how to fix that?

Bouh:
- Can you have an action for setup where will be the RC variable ?
Because i want have the possibility to custom the defaultsRCVariables

- The sync action sync all the RC variable ?
Can we sync just one ?

- Firebase::GetRCString()
What you think of
Firebase::GetVariableString()
RC isn't usefull because we know already we use the Firebase extension

arthuro555:
> RC isn't usefull because we know already we use the Firebase extension
I am not sure if this is also clear to newer users. Naming it getVariable might be missleading.

> Can we sync just one ?
No. We can sync everything or nothing.

> Can you have an action for setup where will be the RC variable ?
I am not sure I understand you. The RC defaults variable is already chosen through an action.

arthuro555:
Ok I am almost done with authentification. I think I will make a build and ask some people to try it out to find bugs I haven't spotted yet (I start now as this extension is huge and it's easier to track bugs when there are only a few)

arthuro555:
Ok I made a build, here is the link: https://arthuro555-data.s3.eu-de.cloud-object-storage.appdomain.cloud/GDevelop%205%205.0.0-beta92-firebase.exe

Extensions/Firebase/firebasetools/D_functionstools.js
```diff
@@ -0,0 +1,45 @@
...
+            param = JSON.parse(parameters);
+        } catch {
+            param = parameters;
+        }
+
+        firebase.functions().httpsCallable(httpFunctionName).__call(param)
+          .then(response => response.data)
+          .then(data => {
+              if(callbackValueVariable) 
+                gdjs.evtTools.network._objectToVariable(data, callbackValueVariable);
```
arthuro555, line 32: Why is _objectToVariable marked as private? It is really useful.

arthuro555:
Hey @4ian ! I believe this PR is finally ready for review.

Extensions/Firebase/JsExtension.js
```diff
@@ -0,0 +1,1354 @@
...
+      	  .addIncludeFile('Extensions/Firebase/C_firebasetools/C_firebasetools.js')
+		  .addIncludeFile('Extensions/Firebase/C_firebasetools/D_cloudfirestoretools.js')
+		  .setFunctionName('gdjs.evtTools.firebase.firestore.updateDocument');
+
+		extension
+          .addAction(
+            'FirestoreUpdateField',
+            _('Update a field of a document'),
+			_('Updates a field of a firestore document.'),
+			_('Update field _PARAM2_ of firestore document _PARAM1_ in collection _PARAM0_ with _PARAM3_ (Store result state in _PARAM4_)'),
```
4ian, line 951: verify the parenthesis capitalisation of all your sentences

Extensions/Firebase/JsExtension.js
```diff
@@ -0,0 +1,1354 @@
...
+      	  .addIncludeFile('Extensions/Firebase/C_firebasetools/C_firebasetools.js')
+		  .addIncludeFile('Extensions/Firebase/C_firebasetools/D_cloudfirestoretools.js')
+		  .setFunctionName('gdjs.evtTools.firebase.firestore.deleteField');
+
+		  extension
+          .addAction(
+            'FirestoreGetDocument',
+            _('Get a document from firestore'),
+			_('Gets a firestore document and store it in a variable.'),
+			_('Load firestore document _PARAM1_ in collection _PARAM0_ into _PARAM2_ (Store result state in _PARAM3_)'),
```
4ian, line 1014: Load Firestore document _PARAM1_ *from* collection [...]?

newIDE/app/src/ProjectManager/ProjectPropertiesDialog.js
4ian: What's this error?

newIDE/app/src/ProjectManager/ProjectPropertiesDialog.js
```diff
@@ -420,6 +426,43 @@ class ProjectPropertiesDialog extends React.Component<Props, State> {
...
+              errorText={
+                validateJSON(firebaseConfig) ? (
+                  undefined
+                ) : (
+                  <Trans>
+                    {`
+                    The Firebase Configuration is not valid JSON. Make sure that
+                    you only include what is between the {} and include the {}.
+                    Make sure the text before the : is also wrapped in "". Make
+                    sure that tere is a , after every configuration string
```
4ian, line 459: typos

Extensions/Firebase/C_firebasetools/D_cloudfirestoretools.js
```diff
@@ -0,0 +1,216 @@
...
+ * @param {string} documentName - The name of the document to check.
+ * @param {gdjs.Variable} [callbackValueVariable] - The variable where to store the result.
+ * @param {gdjs.Variable} [callbackStateVariable] - The variable where to store if the operration was successful.
+ */
+gdjs.evtTools.firebase.firestore.hasDocument = function(collectionName, documentName, callbackValueVariable, callbackStateVariable) {
+    firebase.firestore().collection(collectionName).doc(documentName).get()
+    .then(function(doc) {
+        if (callbackStateVariable)
+          callbackStateVariable.setString("ok");
+        callbackValueVariable.setNumber(doc.exists ? 1 : 2);
```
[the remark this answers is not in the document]
arthuro555: 1 and 2 are for passing a Boolean, as variables don't have those I use numbers: 1 for true and 2 for false. I do not use 0 because it's the default value of variables already. The best would be to be able to use a condition but it is not possible with those asynchronous interfaces.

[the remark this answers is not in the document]
arthuro555: Personally I think numbers are pretty clear but I see how replacing them with a true and false string night be more intuitive.

Extensions/Firebase/C_firebasetools/D_functionstools.js
```diff
@@ -0,0 +1,45 @@
...
+     */
+    call(httpFunctionName, parameters, callbackValueVariable, callbackStateVariable) {
+        let param;
+        try {
+            param = JSON.parse(parameters);
+        } catch {
+            param = parameters;
+        }
+
+        firebase.functions().httpsCallable(httpFunctionName).__call(param)
```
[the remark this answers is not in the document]
arthuro555: I have no idea why there are two underscores. It is the only exposed function of an httpsCallable according to the doc.

Extensions/Firebase/JsExtension.js
```diff
@@ -0,0 +1,1354 @@
...
+             ),
+            _('Change the user\'s email to _PARAM0_ and store result in _PARAM4_ (Send verification email: _PARAM3_)'),
+            _('Firebase/Authentification/User Management/Dangerous'),
+			'JsPlatform/Extensions/firebase.png',
+			'JsPlatform/Extensions/firebase.png'
+		  )
+		  .addParameter('string', _('Old Email'), '', false)
+		  .addParameter('string', _('Password'), '', false)
+		  .addParameter('string', _('New Email'), '', false)
+		  .addParameter('yesorno', _('Send verification Email before change?'), '', false)
```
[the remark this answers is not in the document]
arthuro555: I'm worried that too long descriptions might make the IDE show the text wrong on smaller screens

Extensions/Firebase/JsExtension.js
```diff
@@ -0,0 +1,1354 @@
...
+            _('Use google Firebase features in GDevelop'),
+            'Arthur Pacaud (arthuro555)',
+            'MIT'
+          )
+          .setExtensionHelpPath('/all-features/firebase');
+
+		
+		/* ====== ANALYTICS ====== */
+		extension
+          .addAction(
```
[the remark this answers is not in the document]
arthuro555: oof it looked fine in vscode I wonder how it could get so messed up

Extensions/Firebase/JsExtension.js
```diff
@@ -1,4 +1,7 @@
 // @flow
+/*::
+import { type Trans } from '../../newIDE/app/node_modules/@lingui/macro'
```
arthuro555, line 3: I just noticed you didn't do that for the other extensions. Is it bad?

Extensions/Firebase/JsExtension.js
```diff
@@ -0,0 +1,1354 @@
...
+		  .addIncludeFile('Extensions/Firebase/C_firebasetools/D_authtools.js')
+		  .setFunctionName('gdjs.evtTools.firebase.auth.userManagement.dangerous.deleteUserProvider');
+
+		/* ====== PERFORMANCE ====== */
+
+		extension
+          .addAction(
+            'EnablePerformance',
+            _('Enable Performance Measuring'),
+            _('Enables Performance Measuring.'),
```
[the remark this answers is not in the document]
arthuro555: I am not sure what I shoukd add. It just enables performance measuring and sending results to the firebase console

Extensions/Firebase/C_firebasetools/D_functionstools.js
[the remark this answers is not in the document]
arthuro555: Nvm I found what I did wrong I'll fix it

newIDE/app/src/ProjectManager/ProjectPropertiesDialog.js
arthuro555: The error is: 
```
Message cannot be parsed due to syntax erorrs: {
  apiKey: "AIzaSyBwPHVLhvlHVLh0wqgp4m9qxxxxxxxx",
  authDomain: "xxxxxx-e11a5.firebaseapp.com",
  databaseURL: "https://xxxxx-e11a5.firebaseio.com",
  projectId: "xxxxx-e11xx",
  storageBucket: "xxxxxxx-e11a5.appspot.com",
  messagingSenderId: "254035xxxxxx",
  appId: "1:2345678910:web:2ddd6b830iad9ae3xxxxxxxx",
  measurementId: "G-4REMLXXXXX"
}
```

arthuro555:
I hope I fixed all errors. I might have missed some parenthesis capitalisation issues, but I think I fixed 99% of them.
It's ready for a review again.

arthuro555:
This extension is still awaiting review @4ian

trsh:
@4ian can wee speed up the review? We really needs this tool.

Extensions/Firebase/JsExtension.js
4ian: @arthuro555 This typing is invalid, it should be `(string) => string`. See how it's done in other extensions. Trans is a React like component, not a function.

Extensions/Firebase/JsExtension.js
```diff
@@ -0,0 +1,1715 @@
...
+
+    extension
+      .addAction(
+        'DatabaseGetField',
+        _('Get a field of a variable'),
+        _(
+          'Gets the value of a field in a variable from the database and store it in a scene variable.'
+        ),
+        _(
+          'Load field _PARAM1_ of database variable _PARAM0_ into _PARAM2_ (Store result state in _PARAM3_)'
```
4ian, line 1611: extra capital: can you replace all "(Store" by "(store"?

GDevelop.js/package-lock.json
```diff
@@ -8201,7 +8201,7 @@
     "xml-name-validator": {
       "version": "3.0.0",
       "resolved": "https://registry.npmjs.org/xml-name-validator/-/xml-name-validator-3.0.0.tgz",
-      "integrity": "sha512-A5CUptxDsvxKJEU3yO6DuWBSJz/qizqzJKOMIfUJHETbBw/sFaDxgd6fxm1ewUaM0jZ444Fc5vC5ROYurg/4Pw==",
+      "integrity": "sha1-auc+Bt5NjG5H+fsYH3jWSK1FfGo=",
```
4ian, line 8204: Can you cancel the package-lock.json changes?

Extensions/Firebase/README.md
```diff
@@ -0,0 +1,7 @@
+# Firebase Extension
+
+This is the wrapper for the Firebase SDK for GDevelop.
+The files are prefixed with A_, B_, C_ and D_ to classify their import priority (where A has the highest and D the lowest).
+
+firebasejs contains the Web SDK.
+firebasetools contains wrappers for GDevelop.
```
4ian, line 7: Very clear, thanks for the README 👍

Extensions/Firebase/JsExtension.js
```diff
@@ -0,0 +1,1715 @@
...
+ * and search for any errors.
+ *
+ * More information on https://github.com/4ian/GDevelop/blob/master/newIDE/README-extensions.md
+ */
+module.exports = {
+  createExtension: function (
+      _ /*: Trans */, 
+      gd /*: libGDevelop */
+    ) {
+    const extension /*: gdPlatformExtension */ = new gd.PlatformExtension();
```
4ian, line 23: This extra typing is not necessary, because `gd` is typed. Keep it like the other extensions:
`const extension = new gd.PlatformExtension();`

Extensions/Firebase/C_firebasetools/D_storagetools.js
```diff
@@ -0,0 +1,134 @@
...
+    },
+    (error) => {
+      if (typeof callbackStateVariable !== 'undefined')
+        callbackStateVariable.setString(error.message);
+    },
+    () => {
+      if (typeof callbackStateVariable !== 'undefined')
+        callbackStateVariable.setString('ok');
+      if (typeof callbackUIDVariable !== 'undefined') {
+        gdjs.evtTools.firebase.storage.uploads.remove(uploadID); // Free memory
```
4ian, line 74: Can you explain why this is in the `if`?

Extensions/Firebase/C_firebasetools/D_functionstools.js
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Firebase Functions Event Tools.
+ * @namespace
+ */
+gdjs.evtTools.firebase.functions = {
+  /**
+   * Call an http function.
+   * @param {string} httpFunctionName - The name of the function to call
+   * @param {string} [parameters] - Parameters for the function either as a JSON object or a string.
```
4ian, line 15: "either as a JSON object or a string" => JSON **is** a string :)
Rather, say: "either as a JS object or a JSON string"

Also make sure your typings are up to date with your comments, here something like: `string | Object`

Extensions/Firebase/C_firebasetools/D_cloudfirestoretools.js
```diff
@@ -0,0 +1,296 @@
...
+  callbackStateVariable
+) {
+  firebase
+    .firestore()
+    .collection(collectionName)
+    .doc(documentName)
+    .get()
+    .then(function (doc) {
+      if (callbackStateVariable) callbackStateVariable.setString('ok');
+      callbackValueVariable.setNumber(doc.get(field) === undefined ? 2 : 1);
```
4ian, line 291: What does this 1 and 2 represents? Do you think a string would be easier to use?

Extensions/Firebase/C_firebasetools/D_cloudfirestoretools.js
```diff
@@ -0,0 +1,296 @@
...
+  callbackStateVariable
+) {
+  firebase
+    .firestore()
+    .collection(collectionName)
+    .doc(documentName)
+    .get()
+    .then(function (doc) {
+      if (callbackStateVariable) callbackStateVariable.setString('ok');
+      callbackValueVariable.setNumber(doc.exists ? 1 : 2);
```
4ian, line 262: Same as other comment about 1/2.

Extensions/Firebase/C_firebasetools/D_analyticstools.js
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Firebase Analytics Tools
+ * @namespace
+ */
+gdjs.evtTools.firebase.analytics = {};
+
+/**
+ * Logs an Event/Conversion for that user on the Analytics.
+ * @param {string} eventName The Event being triggered.
+ * @param {string} [eventData] Additional Data for the event.
```
4ian, line 16: Update typing and comment to explain that a JS object or a JSON string are both supported.

Extensions/Firebase/C_firebasetools/D_analyticstools.js
```diff
@@ -0,0 +1,71 @@
...
+ */
+gdjs.evtTools.firebase.analytics.setUserID = function (newUID) {
+  let analytics = gdjs.evtTools.firebase.analytics._analyticsInstance;
+  analytics.setUserId(newUID);
+};
+
+/**
+ * Set an user's property.
+ * @param {string} propertyName The property's name.
+ * @param {string} [propertyData] The data associated to the property.
```
4ian, line 44: Update typing and comment to explain that a JS object or a JSON string are both supported.

Extensions/Firebase/B_firebasejs/NOTICE.txt
```diff
@@ -0,0 +1,3 @@
+This is a modified version of the firebase js sdk. Original can be found on https://github.com/firebase/firebase-js-sdk .
```
4ian, line 1: Indicate the exact version that was used too.

Extensions/Firebase/C_firebasetools/D_storagetools.js
arthuro555: In case a user tries to use it via JS, it will crash if they don't add callbacks, and they migh not pass them if they don't need them.

Core/GDCore/Project/Project.h
```diff
@@ -1,1002 +1,1002 @@
...
-  mutable unsigned int gdBuildVersion;  ///< The GD build version used the last
-                                        ///< time the project was saved.
-  mutable bool dirty;  ///< True to flag the project as being modified.
-#endif
-};
-
-}  // namespace gd
-
-#endif  // GDCORE_PROJECT_H
+/*
```
4ian, line 1: Seems like there are no changes in Project.cpp/Project.h - can you revert these to clean up the history and ease the review?

Extensions/Firebase/C_firebasetools/D_cloudfirestoretools.js
arthuro555: 1 is true, 2 false. I don'T use 0 as it is the default value of variables.

Extensions/Firebase/C_firebasetools/C_firebasetools.js
```diff
@@ -0,0 +1,29 @@
...
+gdjs.evtTools.firebase.onAppCreated = [];
+
+gdjs.registerFirstRuntimeSceneLoadedCallback(function (runtimeScene) {
+  let firebaseConfig;
+  try {
+    firebaseConfig = JSON.parse(
+      runtimeScene.getGame().getExtensionProperty("Firebase", "FirebaseConfig")
+    );
+  } catch {
+    console.error('The Firebase configuration is invalid!');
```
4ian, line 25: Best practice: don't swallow the error, at least log it:

```
catch(error) {
  console.error('The Firebase configuration is invalid:', error);
}
```

Extensions/Firebase/JsExtension.js
```diff
@@ -0,0 +1,1715 @@
...
+        _('Firebase'),
+        _('Use Google Firebase services (database, functions, storage...) in your game.'),
+        'Arthur Pacaud (arthuro555)',
+        'MIT'
+      )
+      .setExtensionHelpPath('/all-features/firebase');
+
+    extension
+      .registerProperty('FirebaseConfig')
+      .setLabel(_('Firebase configuration string'))
```
4ian, line 37: nitpicking: you can add a description that will be displayed below the field to explain what this string is supposed to contain, and where to generate it :)

4ian:
@arthuro555 sorry again for the delay, it's a massive PR :) 
I've made quick sanity checks and added comments. No big issue it seems, if you can address everything I noted we should be in a good position to progress on this :)

Extensions/Firebase/C_firebasetools/D_storagetools.js
4ian: The question is specific about this line: we're testing for `callbackUIDVariable` but then using `gdjs.evtTools.firebase.storage.uploads.remove(uploadID)`, which seems unrelated?

arthuro555:
> sorry again for the delay

That's no problem, as long as this PR doesn't get completely forgotten 😄.

If it can help, 2 persons contacted me via email to tell me the test build I posted months ago on the forum works flawlessly.

Extensions/Firebase/C_firebasetools/D_cloudfirestoretools.js
4ian: In most languages based on C, 0 is false and any other value is casted to true - so better either follow this convention or use "true"/"false".

4ian:
> If it can help, 2 persons contacted me via email to tell me the test build I posted months ago on the forum works flawlessly.

Excellent! Great to know it's already used :) Do we know what kind of features they tested?

arthuro555:
image  
image
image

Mainly authentification and database/firestore seems to have been tested.

One of them made this project: 
[Online Platformer Demo with Firebase Extension.zip](https://github.com/4ian/GDevelop/files/5214628/Online.Platformer.Demo.with.Firebase.Extension.zip)

I didn't get the chance to try it out/open it yet but if it is good we could include it as example.

4ian:
Ping me when you've adressed the comments and it's ready for another review.

Extensions/Firebase/C_firebasetools/D_storagetools.js
arthuro555: If the user doesn't pass a variable, a uid is not generated (as the user won't store it anyways) so if there is no Variable, there is no UID to remove

arthuro555:
It's ready for re-review I think @4ian

Extensions/Firebase/JsExtension.js
arthuro555: I couldn't find a way to do so, I tried setting everything that can be set on a property descriptor.

Extensions/Firebase/A_utils/A_UIDArray.js
```diff
@@ -0,0 +1,45 @@
...
+ * @extends Array
+ */
+gdjs.UIDArray = function () {
+  Array.call(this);
+};
+
+gdjs.UIDArray.prototype = Object.create(Array.prototype);
+
+/**
+ * Adds an object to the UIDs array and returns it's UID.
```
4ian, line 24: Nit: its

Extensions/Firebase/A_utils/A_UIDArray.js
```diff
@@ -0,0 +1,45 @@
+/**
+ * Firebase Tools Collection
+ * @fileoverview
+ * @author arthuro555
+ */
+
+/**
+ * A special array where push tries to reuse old unused indices.
```
4ian, line 8: Is this really giving any upside compared to a map (JS object), or an object based on a map?
I'm a bit afraid of these kind of micro-optimizations that make the codebase more complex and not really 100% indispensable - a few uids in memory in a JS object won't kill anyone? :)

4ian: Note that in the case of heavily used arrays of game objects, this could be useful - with appropriate benchmarks showing that this is giving an advantage. Here I think we don't need any extra complexity ;)

arthuro555: It's not just UIDs, if I don't delete the snapshot it won't be garbage collected and it's a bigger object. Since the default array has no remove action, I did that to add it and also gain some performance.
Not deleting the snapshots or replacing them with null would never completely release the memory they took and would result in a memory leak, which is very bad.

4ian: Sounds are indeed a very heavy resource (in the sense that too much of them will crash the game/browser) that we need to destroy/recycle. They also don't have ids, so we use an array and check if a sound is finished, and if yes we erase it by the new one, to be sure we keep the minimum number of sounds in memory.

For an upload snapshot, would rather a more traditional JS object associating ids to snapshot work? 
Memory consumption will be equivalent, but access will be in constant time (`O(1)`) while using an array, access is linear in the number of elements stored in the array. You have to think how to reuse ids. You can either keep an array for that, or just don't reuse id and increase them at every usage - potentielly avoiding mistakes for people using this in GDevelop :) 

**I am also fine using your approach using an array** but in this case, could you do an object that does not "inherit" from Array? I say that because otherwise this means an external user can mess with your internal array, doing splice, concat, etc... Potentially using the array in a way you don't want to, and creating bugs. When designing a class, make the public interface as limited as possible. 
As we say, [always favor composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance#:~:text=Composition%20over%20inheritance%20(or%20composite,than%20inheritance%20from%20a%20base) :) With a limited interface with only the needed functions, you ensure that your class is "foolproof" and won't be misused.

4ian:
Apart from this UIDArray, I think we'll be good to go :) You can add yourself as a codeowner of this extension directory. 
I'll maybe put this extension under a name "Firebase (experimental)" as we might have some changes to do if we get other feedbacks.

trsh:
@4ian will this be part of `beta100` version?

arthuro555:
> Apart from this UIDArray, I think we'll be good to go :) You can add yourself as a codeowner of this extension directory.
I'll maybe put this extension under a name "Firebase (experimental)" as we might have some changes to do if we get other feedbacks.

The doc is still incomplete and an official example is also missing. Lets finish those before merging ;)

4ian:
@trsh beta 100 is already available :) It might be in the next version or the one after.
@arthuro555 Yep these are important things to do, let's do this before merging! :)

arthuro555:
Ok I fixed some inconsstent wording, added two essential actions I forgot, fixed the optional params, and finished the example. It isn't perfect, but it shows how firebase works and that's what counts I guess? I would appreciate feedback on it.

4ian:
@arthuro555 Sorry I missed this. Having an example is great even if it's not covering everything. How is the wiki documentation going? I think it's the last thing before merging this :)

arthuro555:
I've been busy lately so I couldn't do much for GDevelop, sorry. The documentation isn't really going well, I have done the smaller features like remote config or performance monitoring, but just those took me multiple hours to find a good wording and way to explain. Every time I tackle the bigger features like authentication or firebase, I take hours to write something and scratch it after finishing 1/6 of it because I find it too lame (either I am just reading the instructions descriptions or I make it a tutorial). I'm a bit stuck in this loop because, well, I don't document as well as I code 😅

4ian:
Let's try to get this done because otherwise this is getting blocked and it's a shame as it would be providing a lot of value to users :) 
We don't need a full reference nor a full tutorial - just explain for each stuff the main concepts, what you need to get started (firebase account? some id?) then it's fine to put a link to Firebase docs for more advanced stuff :)

arthuro555:
I already did some good progress, all that is left to document is the database features. I'd really appreciate it if anyone could review the wiki pages for typos or bad documenting.

4ian:
I'm on it :) Looking at the set up page now, it's already a good progress 👍

4ian:
Done for the quickstart page.
Quick notes: 

* permits => allows
* no usage of WRAP component, only `<note>`
* avoid writing (step 1/3), (step 2/3), it make the content more dense without any good reason
* Don't say "I". Use "you" or "we".
* Be sure to always capitalize Firebase
* "GDevelop registers itself as a web app," => this is not GDevelop, it's the GDevelop game. Important to make the distinction.
* "To do so, click here": describe the button.
* At the end of a page, include links to the main Firebase page.

I'm now proofreading the Analytics page. First notes:
* Analytics are very easy to use with Firebase! => avoid using exclamation marks in the documentation, as it's a reference and what is considered easy is not necessarily easy for everyone. Instead, just go ahead and explain how to use something.

4ian:
I'll do the other pages later. Can you go ahead and do these fixes for me already?

* Verify the capitalization of all words. Example "This Article will" -> no capital A to article.
* Change "lets' go to" by "go to".
* Any "I" should be removed and be written like "Here is an example of ..." or anything that is not involving you directly.
  * Another example: "Now I have to update that text to show the correct value" => well we're not forcing you to do anything ;) Just write "You can now update the text to show the correct value".

* For the storage, the page contains self devaluation: "Currently, there is not much that you can do with storage". State instead what is available and that other feature like "blobs" are not yet supported.
* Remove winks/emoticons. Add period to the end of sentences.
* Avoid using ASCII art like "/!\". Just use a `note` in the page.

With this pass, and then me proof reading, we should be good to go.

4ian:
Note that overall this documentation is of good quality. 
👍 You just have to change a bit the tone that is used and make some cosmetic and grammatical fixes.

arthuro555:
image
I am not sure about that exclamation mark, it looks justified to me but should I still try to not use it?

4ian:
This one looks necessary and useful 👍 You can add "Otherwise your changes won't be applied".

GDJS/Runtime/events-tools/networktools.js
```diff
@@ -141,7 +141,9 @@ gdjs.evtTools.network.objectVariableStructureToJSON = function (
 };
 
 gdjs.evtTools.network._objectToVariable = function (obj, variable) {
-  if (!isNaN(obj)) {
+  if (obj == null) {
```
4ian, line 144: Always use `===`, because this will also match undefined. Use `=== undefined` in a separate if branch if you need to.

4ian:
Question about Android (and iOS): when exporting with Cordova, will all of the extension continue to work?

arthuro555:
There is no reason why it shouldn't work on cordova, so I guess it should work, I didn't test it though. What I meant with mobile extra support is to support mobile only features via a cordova plugin.

arthuro555:
> Might be a good idea to investigate maybe the use of signInWithRedirect, could allow maybe things to work almost transparently?

Yes, I'll look into it. Though my plan right now was to get this merged and update it later, lots of stuff is missing here (storage is pretty incomplete, advanced database queries are not present, database presence in Realtime Database isn't present, etc), but I wanted to make it in another PR as this one is way too big already.

4ian:
Sure! My idea is that it's better to have less things but all should be well supported: we do have support for authentication here, so better try to make all of the authentication methods to work everywhere as otherwise this could be deceptive for users.

> storage is pretty incomplete, advanced database queries are not present, database presence in Realtime Database isn't present, etc

All of these are new features that can be added later. It's no problem because we're not promising anything to the user (we're not doing any "marketing" implying that we support these). 

I would recommend taking a look at `signInWithRedirect`, to see how it could be done. It's not blocking for this PR, but it will be surely something that people would want for their mobile games :)

arthuro555:
After working almost all day on it, I got the documentation done (with that I mean every feature is now documented). All that is left to do is proofread multiple times the docs I wrote and proof testing authentication on cordova, and for me to finish and get reviewed a blog post about it 😄 

> I would recommend taking a look at signInWithRedirect, to see how it could be done.

Originally, I didn't implement it on purpose because I am pretty sure users would be confused because the action has as side effect to restart the game (as instead of opening a new window/tab, you open the link in the current tab/window). But for advanced users that understand the implications and manage to store and restore efficiently the state of their game, I see how it could be useful. I'll try to add it in the follow up PRs where I will complete the extension with all Firebase features, as I don't consider it as essential as of now (In the worst case scenario, you can only use email and password, but that's still better than no authentication at all, and people see generally the social logins as a plus but don't mind the more "standard" email-password flow anyways). But if you think otherwise, it shouldn't be too hard to put in.

Anyways, let's get this PR ready to merge 💪 !

4ian:
> After working almost all day on it, I got the documentation done 

Thanks! Looks good!

> I'll try to add it in the follow up PRs where I will complete the extension with all Firebase features, as I don't consider it as essential as of now

That's fine to add it later :) 

>  I didn't implement it on purpose because I am pretty sure users would be confused because the action has as side effect to restart the game (as instead of opening a new window/tab, you open the link in the current tab/window).

Maybe we could add a limitation so that it works only on Cordova?
Basically we would choose the popup or the redirect according to the runtime being used?

4ian:
For the redirect/popup, ideally we would abstract this from the user so that it always work out of the box: in a browser we would use a popup, on Cordova we would use a "redirect" which actually open a separate web view. There is a slight risk the Web view of the game get destroyed, but should not happen most of the time. 

If we don't do this now let's add a warning in the editor that the external providers will only work in a browser :)

arthuro555:
> the external providers will only work in a browser :)

Didn't I already? I could have sworn I did.

arthuro555:
Ok, after some more testing, it seems that cordova doesn't support either popup or redirect. Here is what happens when you try to use redirect: https://streamable.com/ufmxzu

After returning to the app, the authentication didn't succeed. Same with signInWithPopup.

External providers are web only for now at least.

arthuro555:
Oh I forgot to install the required plugins 🤦‍♂️ Let me add the dependencies and try again

4ian:
Yeah be sure to follow *exactly* what is documented - this kind of stuff is hard and painful to debug if you try to bypass something.

arthuro555:
Well, I followed step by step exactly the instructions given by the doc and I still couldn't make it work :/

GDJS/Runtime/events-tools/networktools.js
```diff
@@ -149,26 +149,40 @@ gdjs.evtTools.network.objectVariableStructureToJSON = function (
...
-  if (obj == null) {
+  if (obj === null) {
     variable.setString("null");
   } else if (!isNaN(obj)) {
-    //Number
     variable.setNumber(obj);
   } else if (typeof obj == 'string' || obj instanceof String) {
     variable.setString(obj);
+  } else if (typeof obj === "undefined") {
+    // Do not modify the variable, as there is no value to set it to.
```
4ian, line 159: What about something like:

```
{
  firstName: 'test'
  lastName: undefined
}
```

Will this get properly converted? (the issue is that JS is mixing "undefined" as in "not even declared" and "undefined" as in "a special value that means something was not set".

arthuro555: Well, I assumed that you would use null for a special value that wasn't set. Usually, when you set something to undefined, it is to "delete" it without creating a new hidden class. Therefore I assume you wouldn't want the value to be converted into the variable at all.

4ian: In GDevelop that would be true, in the outside world (i.e: a payload coming from a server), not sure if this holds. But let's keep it like this, having something as undefined can justify to not have it reflected in the variable :)

Extensions/Firebase/C_firebasetools/D_storagetools.js
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  uploadTask.on(
+    firebase.storage.TaskEvent.STATE_CHANGED,
+    (uploadProgress) => {
+      if (typeof callbackProgressVariable !== 'undefined')
+        gdjs.evtTools.network._objectToVariable(
+          uploadProgress,
+          callbackProgressVariable,
+          ['tasks'] // Remove circular reference.
```
4ian, line 63: @arthuro555 This variable won't be used (a good reason to add typescript to these files btw, to catch such errors). I'm going to merge anyway but can you check this? :) If we need circular reference protection, we can build it in _objectToVariable.

arthuro555: Indeed 🤔 I think I should do a full review of the code again as as this so a wrapper and wrapping functions is repetitive I may have overlooked more issues like this one.

4ian:
After all these months, it's finally merged! :D 

I'd like to thank you for your continued and sustained work on this, as well as for the documentation which is of a very nice quality now :) Appreciate the quickstart page 
All the features exposed seems to have been done in a clean manner, they are consistent and well described and documanted. This will be very appreciated from users that wish to integrate Firebase in their games for advanced use cases :) Looking forward to see what people create with this

Thanks 🙌

arthuro555:
I'm a bit scared that even after working for this long there might still be some bugs, but I am happy to have this done :)
I've began working on a blog post about the feature, and that merge caught me a bit off-guard, I'll try to finish it and submit it on the website repository asap!

Extensions/Firebase/C_firebasetools/D_storagetools.js
4ian: Please make a full pass again on the code - ideally adding `// @ts-check` would eliminate this category of problem entirely :)

4ian:
> I'm a bit scared that even after working for this long there might still be some bugs, but I am happy to have this done :)

There surely will be, but they can be mitigated by:
- static typing, i.e: `// @ts-check`
- unit tests (even if here it's a bit hard to test anything because we would need to mock Firebase entirely, not ideal nor necessary I think).
- more example games to test things, though the one you have is fairly complete (but smaller example can be useful to test more stuff).
- proofreading everything again :) 

> I've began working on a blog post about the feature, and that merge caught me a bit off-guard, I'll try to finish it and submit it on the website repository asap!

No worries it's not urgent, this can be announced around the time the next version with this is released.

yu-tang:
I'm not sure that if this thread is the right place to report relational bugs, but the official release which includes this new feature is still undone so I try to post this here.

"Upload file" action in Editor displays duplicate fields for a callback variable with state. See below.
Firebase

PR #1695, opened 2020-04-27
Removed links to non-existing sourcemaps

arthuro555:
As they generate annoying warnings in the console and are unused anyways


4ian:
Sounds fair, merging. Thanks!

PR #1703, opened 2020-04-27
Refactored multiLine into multiline

arthuro555:
To be consistent with material-ui's current documentation


4ian:
Looks good, thanks!

Issue #1663, opened 2020-04-16
Game not closing upon calling ''Quit the game"


Alex25820:
## The bug
When starting a game through the editor the game no longer quits upon calling "Quit the game". This has worked fine before but after updating to 5.0.0-beta91 from beta90 or beta89 (not entierly sure which version I had before) this no longer works.

I am calling the function using "key is pressed". This has worked just fine before but not anymore.

## To Reproduce
1. Create an event.
2. Call "Quit the game" upon "key is pressed".
3. Start the game through the editor by pressing "Launch a preview of the scene".
4. Press your specified key.
5. The game won't close.

## Other details
* I am on a MacBook Air running version 10.15.4.
* GDevelop 5.0.0-beta91 desktop app.

Bouh:
Thank you for your bug report!
[This is fixed](https://github.com/4ian/GDevelop/commit/770aad567284577bf92cb0d79c25b50a5f34167c) for the next update :)

4ian:
Should be fixed in beta 92

PR #1705, opened 2020-04-28
Added user home path to fs extension

arthuro555:
It's useful for storing user dependent stuff without having to use one of the predefined directories. Requested by `Ed_Lo - [Read and agreed]#1663` on discord.

### DIFF
--- a/Extensions/FileSystem/JsExtension.js
+++ b/Extensions/FileSystem/JsExtension.js
@@ -389,7 +389,7 @@ module.exports = {
       .addStrExpression(
         'UserHomePath',
         _('User\'s Home folder'),
-        _('Get the path to the user home folder (On Windows C:\\Users\\<USERNAME>\\ for example).'),
+        _('Get the path to the user home folder (on Windows C:\\Users\\<USERNAME>\\ for example).'),
         _('Filesystem/Windows, Linux, MacOS'),
         'JsPlatform/Extensions/filesystem_folder24.png',
         'JsPlatform/Extensions/filesystem_folder32.png'
--- a/Extensions/FileSystem/JsExtension.js
+++ b/Extensions/FileSystem/JsExtension.js
@@ -387,6 +387,19 @@ module.exports = {
 
     extension
       .addStrExpression(
+        'UserDirPath',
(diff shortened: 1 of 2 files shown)


Extensions/FileSystem/filesystemtools.js
```diff
@@ -97,6 +97,17 @@ gdjs.fileSystem.getUserdataPath = function(runtimeScene) {
   }
 };
 
+/**
+ * Get the path to 'UserData' folder.
+ * @return {string} The path to user*s "home" folder
+ */
+gdjs.fileSystem.getUserHomePath = function() {
+  if(require !== undefined) {
+    return require("os").homedir();
```
Wend1go, line 106: Does this return the same directory as `electron.remote.app.getPath('home')` ?
If so it would be more consistent with how the other special folders are retrieved.

arthuro555: good point! didn't see that was a thing

Extensions/FileSystem/JsExtension.js
```diff
@@ -385,6 +385,19 @@ module.exports = {
       .setIncludeFile('Extensions/FileSystem/filesystemtools.js')
       .setFunctionName('gdjs.fileSystem.getUserdataPath');
 
+    extension
+      .addStrExpression(
+        'UserDirPath',
```
4ian, line 390: UserHomePath

Extensions/FileSystem/JsExtension.js
```diff
@@ -385,6 +385,19 @@ module.exports = {
       .setIncludeFile('Extensions/FileSystem/filesystemtools.js')
       .setFunctionName('gdjs.fileSystem.getUserdataPath');
 
+    extension
+      .addStrExpression(
+        'UserDirPath',
+        _('User\'s Home folder'),
+        _('Get the path to user\'s folder. (On Windows C:\\Users\\<USERNAME>\\ for example)'),
```
4ian, line 392: `Get the path to the user home folder (on Windows C:\\Users\\<USERNAME>\\ for example).`

4ian:
Thanks! Merging this now 👍

PR #1713, opened 2020-04-30
[WIP] [Require help] Begin to add a CordovaExtension type

arthuro555:
## Description

This PR is for adding a subclass CordovaExtension (inheriting from Platform Extension) with the Extra method `addCordovaPlugin(String)` to add cordova plugins as includes to JSExtensions.

## Task list:
- [X] Create the subclass
- [X] Add js Bindings
- [X] Test if can be used like a normal PlatformExtension
  - Results are positive (could export admob example after chaging it to a CordovaExtension)
- [ ] Modify the exporter to export the plugin list
  - [X] Add the code for exporting the plugins
  - [ ] Debug that code 😓 
    - This is the part that requires help (see the review for details)
- [ ] Add support for plugin variables
- [ ] Test on mutiple cordova extensions
- [ ] Correct typos
  - I might need help there too 😅 

## Why?

This would make mobile plugins easier to develop. Multiple developers have no problems with JavaScript but are scared of touching cpp. Some aren't even familiar enough with Low Level concepts like pointers to be able to modify it properly. This would let people declare plugins in extensions declarations (more intuitive and easier).

### DIFF
--- a/GDJS/GDJS/Extensions/CordovaExtension.h
+++ b/GDJS/GDJS/Extensions/CordovaExtension.h
@@ -14,14 +14,20 @@ namespace gdjs {
      * \brief Extension for Cordova games
      * 
      * This is an Extension type for games compiled to the Cordova Platform.
-     * The difference with a normal extension is that 
-     * 1. It only gets exported on Cordova builds
-     * 2. It let you add Cordova plugins in the exported game.
+     * The difference with a normal extension is that it let you add
+     * Cordova plugins in the exported game.
      */
     class GD_API CordovaExtension : public gd::PlatformExtension {
         public:
             // New methods
+            /**
+             * \brief Adds a cordova plugin include to the extension.
+             */
             CordovaExtension& AddCordovaPlugin(const gd::String);
+
+            /**
+             * \brief Get all cordova plugins to export.
+             */
             std::vector<gd::String> GetCordovaPluginList() { return cordovaPlugins; };
             
             // Override the return type of inherited methods
@@ -36,7 +42,7 @@ namespace gdjs {
             CordovaExtension& SetExtensionHelpPath(const gd::String& helpPath) { gd::PlatformExtension::SetExtensionHelpPath(helpPath); return *this; };
 
         private:
-            std::vector<gd::String> cordovaPlugins;
+            std::vector<gd::String> cordovaPlugins; ///< Vector of all cordova plugins to include
     };
 } // namespace gdjs
 #endif // CORDOVAEXTENSION_H
--- a/GDJS/GDJS/Extensions/CordovaExtension.h
+++ b/GDJS/GDJS/Extensions/CordovaExtension.h
@@ -1,0 +1,42 @@
+/*
+ * GDevelop JS Platform
+ * Copyright 2008-2016 Florian Rival (Florian.Rival@gmail.com). All rights
+ * reserved. This project is released under the MIT License.
+ */
+#ifndef CORDOVAEXTENSION_H
+#define CORDOVAEXTENSION_H
+#include "GDCore/Extensions/PlatformExtension.h"
+#include "GDCore/String.h"
+#include <vector>
+
+namespace gdjs {
+    /**
+     * \brief Extension for Cordova games
+     * 
+     * This is an Extension type for games compiled to the Cordova Platform.
+     * The difference with a normal extension is that 
+     * 1. It only gets exported on Cordova builds
+     * 2. It let you add Cordova plugins in the exported game.
+     */
+    class GD_API CordovaExtension : public gd::PlatformExtension {
+        public:
+            // New methods
+            CordovaExtension& AddCordovaPlugin(const gd::String);
+            std::vector<gd::String> GetCordovaPluginList() { return cordovaPlugins; };
+            
+            // Override the return type of inherited methods
+            CordovaExtension& SetExtensionInformation (
+                const gd::String& name_,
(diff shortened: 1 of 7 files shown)


GDJS/GDJS/Extensions/CordovaExtension.h
```diff
@@ -0,0 +1,48 @@
...
+             * \brief Adds a cordova plugin include to the extension.
+             */
+            CordovaExtension& AddCordovaPlugin(const gd::String);
+
+            /**
+             * \brief Get all cordova plugins to export.
+             */
+            std::vector<gd::String> GetCordovaPluginList() { return cordovaPlugins; };
+            
+            // Override the return type of inherited methods
```
arthuro555, line 33: Unsure if this is needed as a pointer is returned so it should be OK to return it as a PlatforExtension pointer

GDJS/GDJS/IDE/ExporterHelper.cpp
```diff
@@ -287,6 +290,22 @@ bool ExporterHelper::ExportCordovaFiles(const gd::Project &project,
...
+  for(gd::String const extensionName : project.GetUsedExtensions()) {
+    std::shared_ptr<gd::PlatformExtension> extension = project.GetCurrentPlatform().GetExtension(extensionName);
+    if (std::shared_ptr<gdjs::CordovaExtension> cordovaExtension = std::dynamic_pointer_cast<gdjs::CordovaExtension>(extension)) { // If a downcast to CordovaExtension is possible...
+      for(gd::String const cordovaPluginName : cordovaExtension->GetCordovaPluginList()) {
+        str += "<plugin name=\"" + cordovaPluginName + "\">\n"; 
+      }
+    }
+  }
+
+  str.FindAndReplace(
```
arthuro555, line 304: This is the part I need help with. Seems pretty simple to me but after compiling and restarting the dev server and electron client, the export still succeeded but didn't replace `<!-- GDJS_CORDOVA_PLUGINS -->`. 
This might be an issue with my workspace but I need others to test to be sure / for them to correct any apparent newbie mistake I could have made (not very familiar with cpp).

4ian:
Being able to add Cordova plugins is definitely needed :) 
I would still suggest a slightly different approach. Rather than using inheritance, why not just add `std::vector<gd::String> cordovaPlugins` to PlatformExtension, but in a bit more flexible way. 

For example, this can instead be a map from exporter name to plugin list:
`std::map<gd::String, std::vector<gd::String>> platformSpecificPlugins`

We would use it like like this: 
`platformSpecificPlugins["cordova"].push_back("cordova-plugin-firebase-analytics");`

Actually, we may imagine that in the future, so plugin might not need configuration (like a plugin to enable vibrations) but others could (analytics plugin, admob...)... in which case, we could move from a complex `std::map<......>`, to a specific class, that describes a plugin for a specific platform, and its properties. Much like we describe an action/condition with `InstructionMetadata`, we can name it `PlatformSpecificPluginMetadata` (a bit long, but does the job):

`PlatformSpecificPluginMetadata platformSpecificPluginMetadata;`

with `PlatformSpecificPluginMetadata` being:
```
class PlatformSpecificPluginMetadata {
  gd::String pluginName; 
  gd::String target; // "android", "ios"... (needs more thinking)
  gd::String exporterName; // "cordova"... (needs more thinking)

  // And methods
}
```

and a method `AddPlatformSpecificPlugin` in PlatformExtension, that returns a new `PlatformSpecificPluginMetadata`, so you can do:

```
auto firebasePluginMetadata = extension.AddPlatformSpecificPlugin().SetPluginName("cordova-plugin-firebase-analytics").SetExporterName("cordova");
firebasePluginMetadata.AddProperty("firebaseId").SetType("string");
firebasePluginMetadata.AddProperty("firebaseApiKey").SetType("string").SetDescription("blabla");
firebasePluginMetadata.AddProperty("otherStuff").SetType("number");
```

Finally, the IDE would be able to go trough the extensions used by the project, and for each of them get the platformSpecificPluginMetadata of the extension, get the properties and display them in a PropertiesEditor.
The values would be saved in the gd::Project.

---

I know this is a looot to digest, especially if you're not that familiar with C++. **Don't do it now** but at least consider making a very simple class PlatformSpecificPluginMetadata that encapsulate the name of a plugin :) 
Then simply add it as a member to gd::PlatformExtension.

You won't need to cast things in the exporter, just to ensure that the exporter is passing the proper name. Will be useful for later adding other exporter/platforms without having to declare a new class.

GDJS/GDJS/IDE/ExporterHelper.cpp
4ian: Mmm that should work. Is the comment not replaced at all? This means that your code did not run (otherwise it would be at least replaced by an empty string).

Have you recompiled GDevelop.js? If yes it should work after reloading GDevelop, because GDevelop.js copy itself to newIDE.

arthuro555:
> PlatformSpecificPluginMetadata 
Huh funny I just so happened to plan adding something like that but more cordova specific (CordovaPluginMetadata)

Well I feel dumb because your solution is waay more simple and better and I really took much time on this. But at least I understand pointers and casting way better now (got stuck on that one for a while >_>)!

I will try to do it like you said. I really do need to be more careful and open issues before implementing stuff to have a discussion about the way to go. I always thought that it wasn't worth it  wasting other people's time because I was sure about my ideas and was wrong, now I need to learn from my mistakes.

Thank you very much for the useful Tipps!

PR #1714, opened 2020-04-30
Cosmetic improvement

Bouh:
Here what is changed: 

- Green color on current plan
- Scrollbar, mainframe.css move in a better place UI/Theme
- Picker folder for new project on top, because it's more logic and avoid people asking "where is saved my game".
- Change sentence for fit to the real behavior of build service. Then 24hours, and not day. People ask often "why i can't now i've launch my build yesterday" a day is misleading.

image


4ian:
Thanks!

The green is... very green! ^^ Can we either not do that change now, or use a color from the theme?

> mainframe.css move in a better place UI/Theme

Can we keep the things unrelated to scrollbar in MainFrame.css into MainFrame, and just the scrollbar stuff into UI/Theme?

Bouh:
- Mainframe.css go back in mainframe folder.
- global.css are added in theme folder, this contain the css for scrollbar.
I've tried to add this css style in index.js of the themes but this require to pass the theme provider and preferences component. I must have missed something, that wasn't working.
Surely an error in the index.js for declare the css as object. Nvm it simplier in Global.css!!
- Green color is now in index.js of themes, as a valid color. And the ThemeConsumer is pass inside SubscriptionDialog.js

4ian:
I've merged but put Global.css as a single file. Even files that are in DefaultTheme or DarkTheme are not "scoped" to this theme unfortunately (that's the issue with CSS), they would need a class to be applied to MainFrame.
Anyway for now as there is only one scrollbar, I put it in a single file :)

PR #1717, opened 2020-05-01
[Final Review in Progress] Add dependencies and properties to Extensions

arthuro555:
Continuing #1713.

This PR adds extension properties for storing basic configuration data and export it, and adds dependencies, which allow to bundle dependencies for modules in the target export platform (example: adding a dependency for a specific cordova plugin).


Core/GDCore/Extensions/Metadata/DependencyMetadata.h
```diff
@@ -0,0 +1,87 @@
...
+            int GetVersion() const { return version; };
+            DependencyTypes GetDependencyType() const { return dependencyType; };
+            std::shared_ptr<gd::PropertyDescriptor> GetExtraSetting(const gd::String& settingName) const { return extraData[settingName]; };
+
+        private:
+            gd::String name = ""; ///< The name of the dependency.
+            gd::String exportName = ""; ///< The name used to install the package (example: npm package name for npm dependency type).
+            int version = -1; ///< The version of the dependency
+            DependencyTypes dependencyType = DependencyTypes::invalid; ///< The tool used to install the dependency.
+            std::map<gd::String, std::shared_ptr<gd::PropertyDescriptor>> extraData = new std::map<gd::String, std::shared_ptr<gd::PropertyDescriptor>>(); ///< Contains dependency type specific additional parameters for the dependency.
```
arthuro555, line 84: I decided to use a map from string to pointers of PropertyDescriptor, not sure if I really needed pointers here.

4ian: `std::map<gd::String, gd::PropertyDescriptor>` could probably work and will avoid you headache when dealing with the copy constructor.

(if you use pointers, the pointed content is not copied when the object is copied. So if someone copies in memory a DependencyMetadata, by doing `DependencyMetadata copy = original;`, now both `copy` and `original` are sharing the same properties (modifying one will modify the other).

4ian, line 84: This usage of new is a bit strange, I'm worried you're instantiating something on the heap. You don't need to do this, objects are initialized with their default constructor when their parent is constructed :)

arthuro555:
I don't understand the CI error. It says the problem comes from std::map??? It compiles fine on my side and I can use the resulting libGD.js without any issue.

arthuro555:
I am stuck on 2 Issues: 
1. I have no idea what is wrong with Travis
2. There seems to be a problem with iterating over the extensions to find dependencies as `plugins` seems to stay "" after exporting the AdMob project. Since I couldn't find any logging functions it's difficult to say where the problem is (could be in any of the for loops).

arthuro555:
I might have found the issue. It seems like GetUsedExtension doesn't output the actual used extensions?? 
image
As you can see there is not the AdMob extension in the list, even tho it is clearly used. I also moved some actions arround and added other from admob to see if libGD would then register the extension but no. Is this intended behavior? How should I correctly loop over the used extensions?

GDJS/GDJS/IDE/ExporterHelper.cpp
```diff
@@ -277,16 +281,39 @@ bool ExporterHelper::ExportCordovaFiles(const gd::Project &project,
...
-    str = str.FindAndReplace(
-        "<!-- GDJS_ADMOB_PLUGIN_AND_APPLICATION_ID -->",
-        "<plugin name=\"cordova-plugin-admob-free\" spec=\"~0.21.0\">\n"
-        "\t\t<variable name=\"ADMOB_APP_ID\" value=\"" +
-            project.GetAdMobAppId() +
-            "\" />\n"
-            "\t</plugin>");
+  gd::String plugins = "";
+
+  for(gd::String const extensionName : project.GetUsedExtensions()) {
```
4ian, line 286: Sorry for the confusion about GetUsedExtensions, this should be probably marked as deprecated/removed.

The idea is that we don't even need to know which extension the project relies on, because the extension action/condition/object are exposing all that is needed to compile the game (mainly: the include files).
This is unfortunately not working super well for your use case. 

For now, I would suggest that you iterate on *all* the extensions of the `gd::Platform`, using `GetAllPlatformExtensions` (see https://docs.gdevelop-app.com/GDCore%20Documentation/classgd_1_1_platform.html#a077aa5264fa2a6ebf82af5a1eb3a7fd2).
It's not optimal, because if an extension is loaded, it would then always write its cordova dependencies even if not used in the project (which could result in AdMob being installed even when not asked). That's not something that we want! But at least you can continue working on this without being blocked, and we can think later how to improve this.

(It's not super easy to fix because:
- I don't like the old approach of having a vector of extensions used by the project, because you need to manually keep it updated. How do you even remove an extension from it?
- the alternative is to automatically generate it when needed. But for this you need to browse the whole project to identify the objects/behavior/actions/conditions/effects/events used and identify their parent extension. It's doable and surely the way to go, but I try to limit classes that have to do work on the whole project).

GDJS/GDJS/IDE/ExporterHelper.cpp
```diff
@@ -277,16 +281,39 @@ bool ExporterHelper::ExportCordovaFiles(const gd::Project &project,
...
+        plugins += "<plugin name=\"" + dependency.GetExportName();
+        if(dependency.GetVersion() != "-1") {
+          plugins += "\" spec=\"" + dependency.GetVersion();
+        }
+        plugins += "\">\n"; 
+
+        // In cordova all settings are considered a plugin variable
+        for (std::pair<gd::String, gd::PropertyDescriptor>& variable : dependency.GetAllExtraSettings()) {
+          if(variable.second.GetType() == "ProjectProperty") {
+            plugins += "\t\t<variable name=\"" + variable.first + "\" value=\"" + project["Get" + variable.second.GetValue()]() + "\" />\n";
```
4ian, line 301: Unless I missed something you cannot generate method name and call them like this dynamically in C++.

I think this should work a bit in the other way: 
* ExporterHelper should iterate on all extensions (ideally only those used by the project, but we'll fix that later) and for each extension:
* It should call a method on the extension, **passing** to it the project (`extension->GenerateDependencies(project)`, `extension->ExposeDependencies(project)`, something like that).
* This method is responsible for looking in the project if there is the required configuration somewhere (for example, the AdMob plugin stores it in `project.GetAdMobAppId`.. but we could create a generic storage for parameters of extensions :) AdMob is a spoiled child to which I give a special place in gd::Project - I should not have ;)).
* The ExporterHelper will get the result and write it (in the config.xml for Cordova, but could be something else in the future for another platform).

In short, you're telling to the extension: "hey extension, please give me the list of your dependencies. To help you, I pass you the project - you'll find your settings somewhere inside, but I don't know where - it's your problem as an extension".

(Though the "your problem as an extension" is a bit hard because extensions can't just store stuff in gd::Project! Hence why we'll need to add a `ExtensionSpecificProperties` class to `gd::Project` one day)
(and finally, something to allow the extension to declare the `properties` that should be shown in the IDE and stored in `ExtensionSpecificProperties`).


Hope this kind of make sense :)

arthuro555: > ExtensionSpecificProperties 
Huh I was thinking of doing that in a future PR so ok.

> Hope this kind of make sense :)
Yes, this makes total sense. Just one thing I am unsure about: I don't think it is possible to overwrite WebAssembly object functions. This could be an issue.

4ian: > Huh I was thinking of doing that in a future PR so ok.

Yes, I think that this will need to be splitted in multiple PRs.

> Just one thing I am unsure about: I don't think it is possible to overwrite WebAssembly object functions. This could be an issue.

It's possible, given a base class, to create an object that overrides a function, but it's a bit complex. We'll see if we really need this. Maybe we can have default functions for Cordova/npm, so that 99% of extensions could just call a function to register their Cordova/npm dependencies. After all, let's try to keep stuff simple rather than over-engineering a "generic" way of adding dependencies that risk to be never used for anything else than Cordova/npm :)

arthuro555:
Ok, to sum up the changes:
- I made an extensionProperties map in Project. It maps a string (the extension name) to a map that maps a string (the property name) to a property descriptor (the property)
- I added a special value serializer to PropertyDescriptor that doesn't serializes metadata (type, label etc)
- I made it so when unserializing, it gets a clone of the PropertyDescriptor from the Extension, and unserializes the non-metadata values into that clone, wich is ten stored in the appropriate place in Project.extensionProperties

I tested it and it seems to work flawless. I only need to ade some UI to set the properties. I wanted to ask where I should put that menu? Should I just add the extension properties to the project properties? Or should I create an extra menu in the Game settings tab of the project manager?

arthuro555:
Ok, I need some help. I tried beginning the UI part of extension properties but I never did any react before, and I failed miserably 😓 .
Problem list:
- [ ] Values loaded in wrong slots
- [ ] Values not modifiable
- [ ] Values don't show up in empty project

Help is welcome!

4ian:
Two things:
- You mix the properties from the extension inside the project properties, and then display them. While it kind of work, I would recommend a slightly different approach: 
  - You keep properties in extensions. In the project, you only save in the project the *values* of the properties, indexed by extension name and property name: `std::map<gd::String std::map<gd::String, gd::String>` (at this point, I recommend making a simple class like [`PlatformSpecificAssets`](https://github.com/4ian/GDevelop/blob/master/Core/GDCore/Project/PlatformSpecificAssets.cpp), where you can create a few nice looking getter and setter rather than fighting with a double std::map).
  - Then in the IDE, you can map over properties from extensions, and get/set values from/to the project.
- Second thing, there is actually already something that you can use to display properties on screen :) It has all the logic to show different fields according to the property type, and you "just" need to specify the list of properties and how to load the value and store the value. It's called `PropertiesEditor`
  - See [`BehaviorPropertiesEditor`](https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/BehaviorsEditor/Editors/BehaviorPropertiesEditor.js) as an example.
  - this part is where you set how to get and update the value of a property. `behaviorContent` = `project` in your case (or `project.getExtensionSpecificProperties()`):
    https://github.com/4ian/GDevelop/blob/6a3af0d57a7c2d4ed921d00b6fc84f5a8339284e/newIDE/app/src/BehaviorsEditor/Editors/BehaviorPropertiesEditor.js#L21-L32

  - this part is where you specify the "schema" (which you get from your properties, it's just another abstraction) and the `instances`... what are the instance here? It's your `project` again (or `project.getExtensionSpecificProperties()`). Not a big deal though because you only have one project (but this could be useful if your PropertiesEditor was supporting editing multiple projects at the same time):
 
 
 
 https://github.com/4ian/GDevelop/blob/6a3af0d57a7c2d4ed921d00b6fc84f5a8339284e/newIDE/app/src/BehaviorsEditor/Editors/BehaviorPropertiesEditor.js#L38-L39

arthuro555:
> You keep properties in extensions. In the project, you only save in the project the values of the properties, indexed by extension name and property name: std::map<gd::String std::map<gd::String, gd::String> (at this point, I recommend making a simple class like PlatformSpecificAssets, where you can create a few nice looking getter and setter rather than fighting with a double std::map).

Wouldn't it be more complicated to have the metadata and values in separate objects?

4ian:
> Wouldn't it be more complicated to have the metadata and values in separate objects?

As a first view yes, but the values (gd::String basically, keyed by their name) and the metadata (gd::PropertyDescriptor) are two different concepts.
The metadata is like an empty blueprint, describing what to show to the user. It's not something that should in any moment be copied to the project, because the **source of truth** is the extension. For example, if I update the description of a property of an extension in memory and refresh the window showing the properties, I should see the updated description of the property that I just changed :) This would not be possible if you copied them to the project.
Sure you could "re-copy" them, but that's a pain. Or we could as you did just store the values when serializing... but at this moment, if we're storign *part* of a property, the thing we're storing is not a property anymore :) It's... a value. 

As a rule of thumb, whenever you're duplicating data, it's a good idea to think "do I really need to duplicate these things?". The answer here is "not really", and even "probably not, because metadata should have a single source of truth, and it's the extension". 

And for the UI, I already did the hard work for you, thanks to `PropertiesEditor` ;) Well you still need to iterate on extensions, and for each one get its properties, and then feed PropertiesEditor with it.

Core/GDCore/Project/ExtensionPropertiesManager.h
```diff
@@ -0,0 +1,55 @@
+/*
```
arthuro555, line 1: Did you mean something like this? Or did you mean something else?

arthuro555: I am awaiting confirmation @4ian (I don't want to do something wrong again)

4ian: I see that you're still storing an `std::shared_ptr<gd::PlatformExtension> extension` and `std::map<gd::String, gd::PropertyDescriptor> extraData` in your "ExtensionProperty", why is that?
I think that you're seeing things as more complex than they should be.
Don't store "metadata" - they are in the extension and you don't need a pointer to them in the project. Just store:
- the name
- what is its value.

Basically it's just a map :) With a function to get a value, and another one to set a value (given as parameter the property name).
With SerializeTo/UnserializeFrom methods, and you're done.

Then in the IDE, you use `PropertiesEditor` component to show the properties of each extension. You pass it the properties coming from the extension, as I explained.
You need to pass a function that will be called when a property value is modified by the user, and another one used to read the current value. In this case, you use `project.getExtensionPropertiesManager().getValue(extension, propertyName)` or ``project.getExtensionPropertiesManager().setValue(extension, propertyName, newValue)`.

Core/GDCore/Project/Project.cpp
```diff
@@ -615,6 +615,31 @@ void Project::UnserializeFrom(const SerializerElement& element) {
...
+    }
+  }
+  */
+  
+  SerializerElement& allExtensionsPropertiesElement = propElement.GetChild("extensionProperties");
+  for(std::shared_ptr<PlatformExtension> extension : GetCurrentPlatform().GetAllPlatformExtensions()) { // We do not iterate over serialized object to include also new properties and remove old removed ones.
+    gd::String extensionName = extension->GetName(); // Cache the name to not call the function again...
+    SerializerElement& extensionElement = allExtensionsPropertiesElement.GetChild(extensionName); // ... and the serializer element too
+    for(std::pair<const gd::String, gd::PropertyDescriptor>& property : extension->GetAllProperties()) {
+      gd::PropertyDescriptor propertyInstance = *property.second.Clone(); // Take the extension's property as "template" to keep the metadata.
```
4ian, line 636: You won't need to do all that complicated stuff. The project should just store a bunch of key-values. Forget about properties in the project.

Core/GDCore/Project/ExtensionPropertiesManager.h
arthuro555: Ohhh I get it. I thought about something way too complicated indeed.

arthuro555:
I still have no Idea what is going on with Travis. Do you have an idea of what might cause the problems?

4ian:
Travis is probably failing because the native game engine compilation is failing. Probably you need to use a bunch of `#if defined(GD_IDE_ONLY) ... #endif` to hide stuff that are not needed in the game engine. Though don't care too much about it, I can fix that, concentrate on getting something as simple and short as possible.

newIDE/app/src/ProjectManager/ExtensionsPropertiesDialog.js
```diff
@@ -0,0 +1,110 @@
...
+
+  render() {
+    const { project } = this.props;
+    const allExtensions = project
+      .getCurrentPlatform()
+      .getAllPlatformExtensions();
+    const properties = [];
+    for (let i = 0; i < allExtensions.size(); i++) {
+      const extension = allExtensions.at(i);
+      const propertiesSchema = propertiesMapToSchema(
```
arthuro555, line 49: I think I did something wrong. When mounting the component I get an NaN error in console, and I cannot toggle booleans, it puts a warning about NaN in the console instead. Do you know where I have gone wrong?

newIDE/app/src/ProjectManager/ExtensionsPropertiesDialog.js
```diff
@@ -0,0 +1,110 @@
...
+          .size() !== 0
+      )
+        properties.push(
+          <ColumnStackLayout key={extension.getName()}>
+            <Text size="title">
+              <Trans>{extension.getFullName()}</Trans>
+            </Text>
+            <PropertiesEditor
+              schema={propertiesSchema}
+              instances={[extension.getAllProperties()]}
```
4ian, line 76: The instances must be `[project]`, not the properties. Instances is what is modified, which is here the project (the extension is not modified, extensions are just giving the property).

newIDE/app/src/ProjectManager/ExtensionsPropertiesDialog.js
```diff
@@ -0,0 +1,110 @@
...
+        instance =>
+          this._mapExtensionValueToPropertyDescriptors(
+            extension.getAllProperties(),
+            extension.getName(),
+            project
+          ),
+        (instance, propertyName, newValue) =>
+          project
+            .getExtensionPropertiesManager()
+            .setValue(extension.getName(), propertyName, newValue)
```
4ian, line 61: This looks correct

newIDE/app/src/ProjectManager/ExtensionsPropertiesDialog.js
```diff
@@ -0,0 +1,110 @@
...
+    for (let i = 0; i < allExtensions.size(); i++) {
+      const extension = allExtensions.at(i);
+      const propertiesSchema = propertiesMapToSchema(
+        extension.getAllProperties(),
+        instance =>
+          this._mapExtensionValueToPropertyDescriptors(
+            extension.getAllProperties(),
+            extension.getName(),
+            project
+          ),
```
4ian, line 57: This looks very complex. Why can't this just be `project.getExtensionPropertiesManager().getvalue(extension.getName(), propertyName)`?

arthuro555: I thought I had to pass a propertyDescriptor, but `project.getExtensionPropertiesManager().getvalue(extension.getName(), propertyName)` would just be the value.

4ian: You're right my bad! Let me think about it. `_mapExtensionValueToPropertyDescriptors` is not going to work because you're modifying in memory the properties of the extension.
We might have to make a static method in C++, which takes an extension, a project, and returns the properties with the values set.

In other words, what you did in JS, but do it as a method in gd::Project... or better in ExtensionPropertiesManager :) 
It's a bit cumbersome sorry - this was originally designed for objects that are returning their properties while here we have properties in extension and objects storing them.

Core/GDCore/Project/Project.cpp
4ian: This might still be useful in the method that returns the properties given an extension (which can be used in `propertiesMapToSchema` as the second argument)

4ian:
So yeah I made a mistake, just to confirm:

- `propertiesMapToSchema` can be used, knowing that for the second argument you have to do a bit of cumbersome code by calling a method in C++ that is returning you a copy of the properties of the extension, with the values set. You should pass to this method the extension (to get the properties), and probably make this method on `ExtensionPropertiesManager`.
- You should as well call this method for the first argument of `propertiesMapToSchema` I think (the first argument is the initial properties, and the second argument is called whenever a value needs to be read again)
- Once this works, this should be good for `PropertiesEditor`.
- In the future, we could rework `propertiesMapToSchema` to avoid this cumbersome second argument... but as objects and behaviors are working like this, let's align the implementation.

Sorry about the confusion, hope you'll get something working :)

Core/GDCore/Project/ExtensionPropertiesManager.h
```diff
@@ -0,0 +1,53 @@
...
+#ifndef GDCORE_EXTENSIONPROPERTIESMANAGERS_H
+#define GDCORE_EXTENSIONPROPERTIESMANAGERS_H
+#include <map>
+#include "GDCore/Serialization/SerializerElement.h"
+#include "GDCore/String.h"
+
+namespace gd{
+  class GD_CORE_API ExtensionPropertiesManager {
+    public:
+  	  gd::String& GetValue(const gd::String& extension, const gd::String& property) {
```
4ian, line 15: I think you can remove this non const method - callers that want to modify a value will use SetValue ;)

arthuro555:
Everything seems alright, Only two problems. 
1. Booleans properties does not work. It is weird because it seems to be checking if it is equal to the string true but set it to 1 or 0?
https://github.com/4ian/GDevelop/blob/cc1d26201efde5b7174d5364f7d89b5dc4ae06d4/newIDE/app/src/PropertiesEditor/PropertiesMapToSchema.js#L85
https://github.com/4ian/GDevelop/blob/cc1d26201efde5b7174d5364f7d89b5dc4ae06d4/newIDE/app/src/PropertiesEditor/PropertiesMapToSchema.js#L89
2. (I'll change that) Enums are not handled well by flow.

arthuro555:
Ok everything is fixed, good and working, and ready for review.
Only little problem is a warning by React in the console when opening the properties editor or the first time. The dialog title is not big even tho I think it should be, maybe the warning is about that? Anyways everything else seems to work flawlessly.

arthuro555:
In case you missed it @4ian it's ready for review

.vscode/settings.json
```diff
@@ -1,111 +0,0 @@
-// Place your settings in this file to overwrite default and user settings.
```
4ian, line 1: Can you remove this file from the PR?

arthuro555: yes sorry

Core/GDCore/Project/ExtensionPropertiesManager.h
```diff
@@ -0,0 +1,60 @@
...
+ * Copyright 2008-2016 Florian Rival (Florian.Rival@gmail.com). All rights
+ * reserved. This project is released under the MIT License.
+ */
+#ifndef GDCORE_EXTENSIONPROPERTIESMANAGERS_H
+#define GDCORE_EXTENSIONPROPERTIESMANAGERS_H
+#include <map>
+#include "GDCore/Serialization/SerializerElement.h"
+#include "GDCore/String.h"
+
+namespace gd{
```
4ian, line 12: Can you run VS Code "Format Document" on all C++ files that you touched in this PR?
It's powered by clang-format (you may have to install an extension), with the style defined in `.clang_format`.

Core/GDCore/Project/PropertyDescriptor.h
```diff
@@ -34,6 +34,13 @@ class GD_CORE_API PropertyDescriptor {
    */
   PropertyDescriptor() : hidden(false){};
 
+  /**
+   * \brief Returns a new PropertyDescriptor with the same data as the current one.
+   */
+  virtual PropertyDescriptor* Clone() const {
```
4ian, line 40: I'm unsure why this is needed, can you explain?

Core/GDCore/Project/PropertyDescriptor.h
```diff
@@ -120,6 +127,16 @@ class GD_CORE_API PropertyDescriptor {
    * \brief Unserialize the PropertyDescriptor.
    */
   virtual void UnserializeFrom(const SerializerElement& element);
+
+  /**
+   * \brief Serilaize only the Value and Extra Informations.
```
4ian, line 132: Typo

Core/GDCore/Project/PropertyDescriptor.h
```diff
@@ -120,6 +127,16 @@ class GD_CORE_API PropertyDescriptor {
...
+   */
+  virtual void SerializeValuesTo(SerializerElement& element) const;
+
+  /**
+   * \brief Unserilaize only the Value and Extra Informations.
```
4ian, line 137: Typo. Also like before, no extra capitalization please.

Extensions/AdMob/JsExtension.js
```diff
@@ -32,6 +32,22 @@ module.exports = {
...
+    extension.addDependency()
+      .setName("AdMob Cordova Extension")
+      .setDependencyType("cordova")
+      .setExportName("cordova-plugin-admob-free")
+      .setVersion("~0.21.0")
+      .setExtraSetting(
+        "ADMOB_APP_ID", 
+        new gd.PropertyDescriptor("AdMobAppID")
+         .setType("ExtensionProperty")
+      );
```
4ian, line 49: This looks very clean 👍👍😍

GDJS/Runtime/Cordova/config.xml
```diff
@@ -30,5 +30,5 @@
     <!-- Cordova/Phonegap version -->
     <preference name="phonegap-version" value="cli-8.0.0" />
 
-    <!-- GDJS_ADMOB_PLUGIN_AND_APPLICATION_ID -->
+    <!-- GDJS_PLUGINS -->
```
4ian, line 33: Uber nitpicking: `<!-- GDJS_CORDOVA_PLUGINS -->` maybe?

GDJS/Runtime/Electron/package.json
```diff
@@ -5,7 +5,9 @@
     "description": "GDJS_GAME_NAME",
     "author": "GDJS_GAME_AUTHOR",
     "version": "GDJS_GAME_VERSION",
-    "dependencies": {},
+    "dependencies": {
+      "GDJS_DEPENDENCY": "0"
```
4ian, line 9: Uber nitpicking: `GDJS_NPM_DEPENDENCY`?

GDevelop.js/Bindings/postjs.js
```diff
@@ -151,6 +151,7 @@ var adaptNamingConventions = function(gd) {
     //Preserve backward compatibility with some alias for methods:
     gd.VectorString.prototype.get = gd.VectorString.prototype.at;
     gd.VectorPlatformExtension.prototype.get = gd.VectorPlatformExtension.prototype.at;
+    gd.VectorDependencyMetadata.prototype.get = gd.VectorDependencyMetadata.prototype.at;
```
4ian, line 154: Let's remove this :) It was only added for backward compatibility of other classes :)

newIDE/README-extensions.md
```diff
@@ -127,6 +127,66 @@ Add an object using [`addObject`](https://docs.gdevelop-app.com/GDCore%20Documen
...
+
+extension.addDependency()
+  .setName("AdMob Cordova Extension")
+  .setDependencyType(gd.DependencyTypes.cordova)
+  .setExportName("cordova-plugin-admob-free")
+  .setVersion("~0.21.0")
+  .setExtraSetting(
+    "ADMOB_APP_ID", 
+    new gd.PropertyDescriptor()
+      .setType("ExtensionProperty") // Tell the exporte this is an extension property...
```
4ian, line 185: Tell the exporter

newIDE/app/src/ProjectManager/ExtensionsPropertiesDialog.js
```diff
@@ -0,0 +1,114 @@
...
+          <PropertiesEditor
+            schema={propertiesSchema}
+            instances={[extension.getAllProperties()]}
+          />
+        </ColumnStackLayout>
+      );
+    }
+
+    return (
+      <React.Fragment>
```
4ian, line 82: You don't need this Fragment?

newIDE/app/src/ProjectManager/index.js
```diff
@@ -783,6 +787,14 @@ export default class ProjectManager extends React.Component<Props, State> {
...
                 }}
               />,
+              <ListItem
+                key="extension-properties"
+                primaryText={<Trans>Extensions Properties</Trans>}
+                leftIcon={<Tune />}
+                onClick={() =>
+                  this.setState({ extensionsPropertiesDialogOpen: true })
+                }
+              />,
```
4ian, line 797: I'd like to avoid having a new menu item for this. For the end-user, we don't really care that this is an "extension" property. If I want to change a global setting like the project name or the ad mob id, it should be at a single place... the existing project properties dialog :)

newIDE/app/src/ProjectManager/ExtensionsPropertiesDialog.js
```diff
@@ -0,0 +1,114 @@
...
+import propertiesMapToSchema from '../PropertiesEditor/PropertiesMapToSchema';
+import { List } from '@material-ui/core';
+
+type Props = {|
+  project: gdProject,
+  open: boolean,
+  onApply: Function,
+|};
+
+class ExtensionsPropertiesDialog extends React.Component<Props> {
```
4ian, line 20: Can we make this *not* a dialog, but just a component that you use inside `ProjectPropertiesDialog`? So that the existing Ad Mob property, and the new ones, will be all shown like they are part of the dialog :)

image

Core/GDCore/Project/ExtensionPropertiesManager.h
```diff
@@ -0,0 +1,60 @@
...
+  class Project;
+  class PropertyDescriptor;
+}
+
+namespace gd{
+  class GD_CORE_API ExtensionPropertiesManager {
```
4ian, line 18: Nitpicking: This class is indeed giving some useful functions to manipulate properties, but it's first and mostly the holder of some data which is `std::map<gd::String, std::map<gd::String, gd::String>> properties`. In other words, this thing IS representing/encapsulating the properties. In other words, this thing IS the properties.

So I would simply name this "ExtensionProperties" :)

Core/GDCore/Extensions/PlatformExtension.h
```diff
@@ -437,6 +476,21 @@ class GD_CORE_API PlatformExtension {
    * generator.
    */
   void StripUnimplementedInstructionsAndExpressions();
+
+  /**
+   * \brief Get an extension property
+   * \param propertyName The name of the property
+   */
+  gd::PropertyDescriptor& GetProperty(const gd::String& propertyName) {
```
4ian, line 484: I don't find where this is used, can you show me?

Core/GDCore/Project/ExtensionPropertiesManager.h
4ian: Granted that you have in `gd::PlatformExtension` something that is already called `extensionProperties`, but I think that it is this member in `gd::PlatformExtension` that could have another name.

Core/GDCore/Extensions/PlatformExtension.h
```diff
@@ -437,6 +476,21 @@ class GD_CORE_API PlatformExtension {
...
+  /**
+   * \brief Get an extension property
+   * \param propertyName The name of the property
+   */
+  gd::PropertyDescriptor& GetProperty(const gd::String& propertyName) {
+    return extensionProperties[propertyName];
+  };
+
+  /**
+   * \brief Get all extension's property
```
4ian, line 489: Get all the properties of the extension

Core/GDCore/Extensions/PlatformExtension.h
```diff
@@ -225,6 +230,35 @@ class GD_CORE_API PlatformExtension {
                               const gd::String& smallicon_,
                               std::shared_ptr<gd::BaseEvent> instance);
 
+#if defined(GD_IDE_ONLY)
+  /**
+   * \brief Adds a property to the Extension.
+   * \param name The internal name of the property (to get it later).
+   * \param description The description of the property.
+   * \param type The type of the property. Supported ones are "string", "number" and "boolean".
+   * \param value The default Value.
```
4ian, line 239: This is not a parameter

Core/GDCore/Extensions/PlatformExtension.h
```diff
@@ -225,6 +230,35 @@ class GD_CORE_API PlatformExtension {
                               const gd::String& smallicon_,
                               std::shared_ptr<gd::BaseEvent> instance);
 
+#if defined(GD_IDE_ONLY)
+  /**
+   * \brief Adds a property to the Extension.
+   * \param name The internal name of the property (to get it later).
+   * \param description The description of the property.
```
4ian, line 237: Either use description or fullname, but not both.

Core/GDCore/Extensions/Metadata/DependencyMetadata.h
```diff
@@ -0,0 +1,89 @@
...
+             *  dependencyMetadata.setExporterName("is-thirteen");
+             * \endcode
+             */
+            DependencyMetadata& SetExportName(const gd::String& exportName_) {
+                exportName = exportName_;
+                return *this;
+            };
+
+            /**
+             * \brief Set the version of the dependency to install. Use -1 for latest.
```
4ian, line 43: "-1", or what about just not specifying anything (empty string)? Should be easy to change?

Core/GDCore/Extensions/Metadata/DependencyMetadata.h
```diff
@@ -0,0 +1,89 @@
...
+#ifndef DEPENDENCYMETADATA_H
+#define DEPENDENCYMETADATA_H
+#include <map>
+#include "GDCore/String.h"
+#include "GDCore/Project/PropertyDescriptor.h"
+#include "GDCore/Tools/Log.h"
+
+namespace gd {
+    /**
+     * \brief Contains a information about a Dependency (library) of an extension.
```
4ian, line 15: "a dependency (library, npm/cordova package, or other according to the export)"

newIDE/app/src/ProjectManager/index.js
```diff
@@ -1132,6 +1144,15 @@ export default class ProjectManager extends React.Component<Props, State> {
             onChangeSubscription={this.props.onChangeSubscription}
           />
         )}
+        {this.state.extensionsPropertiesDialogOpen && (
```
4ian, line 1147: As explained, let's rather use the component in the `ProjectPropertiesDialog` :)

newIDE/app/src/ProjectManager/ExtensionsPropertiesDialog.js
```diff
@@ -0,0 +1,114 @@
...
+  _onApply = () => {
+    this.props.onApply();
+  };
```
4ian, line 23: Uber nit: no need to create an intermediate function, just pass `this.props.onApply`.

GDJS/GDJS/IDE/ExporterHelper.cpp
```diff
@@ -421,6 +446,26 @@ bool ExporterHelper::ExportElectronFiles(const gd::Project &project,
...
+    gd::String packages = "";
+
+    for(std::shared_ptr<gd::PlatformExtension> extension : project.GetCurrentPlatform().GetAllPlatformExtensions()) { //TODO Add a way to select only used Extensions
+      for (gd::DependencyMetadata dependency: extension->GetAllDependencies()) {
+        if (dependency.GetDependencyType() == "npm") {
+          if(dependency.GetVersion() == "-1") {
+            gd::LogError("Latest Version (-1) not available for NPM dependencies, dependency " + dependency.GetName() + " is not exported.");
+            continue;
+          }
+          packages += "\n\t\"" + dependency.GetExportName() + "\": \"" + dependency.GetVersion() + "\",";
```
4ian, line 458: Use only `\n` for new lines, not `\n\t` (unix style).

arthuro555: the t is a tab for indentation not for the new line

4ian:
Alright I finally reviewed this and the implementation is much more satisfying than before! I think we're on track to get this merged soon :) I added a bunch of comments, then I'll merge and do any remaining fixes myself.

Thanks for your work on this, seems like we're getting something clean and well done, good job 👍

newIDE/README-extensions.md
4ian: Also thanks for the README update! Super useful!! 👍

GDJS/GDJS/IDE/ExporterHelper.cpp
4ian: Ah silly me! Forget about this, looks good then!

Core/GDCore/Extensions/PlatformExtension.h
arthuro555: oh yes this is some dead code because I originally planned to implement it another way that wouldn't have worked. I'll remove that.

Core/GDCore/Project/Project.cpp
```diff
@@ -618,6 +619,19 @@ void Project::UnserializeFrom(const SerializerElement& element) {
...
+      propElement.GetChild("extensionProperties"));
+
+  // Compatibility: Move AdMob App ID from project property to extension
+  // property.
+  if (propElement.GetStringAttribute("adMobAppId", "") != "") {
+    extensionProperties.SetValue(
+        "AdMob",
+        "AdMobAppId",
+        propElement.GetStringAttribute("adMobAppId", ""));
+  }
```
4ian, line 633: Use this exact format:
```
// Compatibility with GD <= 5.0.0-beta98
// Move AdMob App ID from project property to extension property.
...
// end of compatibility code
```

I try to consistent use these guards `Compatibility with GD <= ...` so that we can automate some code removal in the future :)

GDJS/GDJS/IDE/ExporterHelper.cpp
```diff
@@ -261,16 +265,37 @@ bool ExporterHelper::ExportCordovaFiles(const gd::Project &project,
...
-            "\" />\n"
-            "\t</plugin>");
+  gd::String plugins = "";
+
+  for(std::shared_ptr<gd::PlatformExtension> extension : project.GetCurrentPlatform().GetAllPlatformExtensions()) { //TODO Add a way to select only used Extensions
+    for (gd::DependencyMetadata dependencies: extension->GetAllDependencies()) {
+      if (dependencies.GetDependencyType() == "cordova") {
+
+        plugins += "<plugin name=\"" + dependencies.GetExportName();
+        if(dependencies.GetVersion() != "-1") {
```
4ian, line 275: As written in another comment, can we just use the empty string instead of "-1"?

GDJS/GDJS/IDE/ExporterHelper.cpp
```diff
@@ -421,6 +446,26 @@ bool ExporterHelper::ExportElectronFiles(const gd::Project &project,
             .FindAndReplace("\"GDJS_GAME_VERSION\"", jsonVersion)
             .FindAndReplace("\"GDJS_GAME_MANGLED_NAME\"", jsonMangledName);
 
+    gd::String packages = "";
+
+    for(std::shared_ptr<gd::PlatformExtension> extension : project.GetCurrentPlatform().GetAllPlatformExtensions()) { //TODO Add a way to select only used Extensions
+      for (gd::DependencyMetadata dependency: extension->GetAllDependencies()) {
+        if (dependency.GetDependencyType() == "npm") {
+          if(dependency.GetVersion() == "-1") {
+            gd::LogError("Latest Version (-1) not available for NPM dependencies, dependency " + dependency.GetName() + " is not exported.");
```
4ian, line 455: Same here, consider an empty string, which is I think perfectly representing an absence of value, which can be safely considered as "give me the latest version because I don't care about the version":
```suggestion
          if(dependency.GetVersion() == "") {
            gd::LogError("Latest Version not available for NPM dependencies, dependency " + dependency.GetName() + " is not exported. Please specify a version when calling addDependency.");
```

GDevelop.js/package-lock.json
```diff
@@ -1,8283 +0,0 @@
-{
```
4ian, line 1: Can we remove this file from the PR? (i.e: add it back?)

GDevelop.js/types/gdextensionpropertiesmanager.js
```diff
@@ -0,0 +1,13 @@
+// Automatically generated by GDevelop.js/scripts/generate-types.js
+// Automatically generated by GDevelop.js/scripts/generate-types.js
+// Automatically generated by GDevelop.js/scripts/generate-types.js
+// Automatically generated by GDevelop.js/scripts/generate-types.js
+// Automatically generated by GDevelop.js/scripts/generate-types.js
```
4ian, line 5: That's a lot of comments :D What happened?

GDJS/GDJS/IDE/ExporterHelper.cpp
arthuro555: Oh sorry I must have missed it

GDevelop.js/types/gdextensionpropertiesmanager.js
arthuro555: Idk tbh. Oh wait I think it's because the branch is old and it uses the version of the extractor that was broken on windows.

Core/GDCore/Extensions/PlatformExtension.h
```diff
@@ -225,6 +231,33 @@ class GD_CORE_API PlatformExtension {
...
+    return extensionPropertiesMetadata[name];
+  };
+#endif
+
+  /**
+   * \brief Adds a property to the Extension.
+   */
+  gd::PropertyDescriptor& RegisterProperty(const gd::String& name) {
+    return extensionPropertiesMetadata[name];
+  };
```
4ian, line 259: To solve the compilation issue, I think you can put this method inside the `#if defined(GD_IDE_ONLY)`/`#endif`.
This effectively removes the existence of this method when the code is compiled for GDCpp (the game engine). It's ok because in the C++ game engine, we don't want to extensions to register things that are "metadata", i.e: that are things related to the editor.

This should solve the compilation.

arthuro555: That was it indeed. I forgot I had a "Fluent API" (I think that is how chained call based APIs are called) declaration as well 😅

arthuro555:
Ok I think everything should be ok, good and done now. I would still let the CI end running just in case.

Core/GDCore/Extensions/PlatformExtension.h
4ian: This is indeed the name of this "pattern" where calls to set up something are chained :)

arthuro555:
The ci passed :) I think it is now ready to merge.

newIDE/app/src/ProjectManager/ProjectPropertiesDialog.js
```diff
@@ -46,417 +47,400 @@ type State = {|
...
-      adMobAppId: project.getAdMobAppId(),
-      scaleMode: project.getScaleMode(),
-      sizeOnStartupMode: project.getSizeOnStartupMode(),
-      showGDevelopSplash: project.getLoadingScreen().isGDevelopSplashShown(),
-      minFPS: project.getMinimumFPS(),
-      maxFPS: project.getMaximumFPS(),
-      isFolderProject: project.isFolderProject(),
-    };
-  }
+  if (!displayProjectErrorsBox(t, getErrors(t, project))) return;
```
4ian, line 104: This should be preventing "Apply" to finish if there are errors.
For this: `return displayProjectErrorsBox(...)` and when calling `applyPropertiesToProject`, don't call `props.onApply();` if false is returned.

4ian:
I think I found a bug :) Please double check the behavior the proper working of Apply/Apply in case of error/Cancel and we'll merge when it's fixed.

4ian:
Thanks, should be good to go, will merge this later tonight!

4ian:
Merged :) 
Congrats on getting this to the point where it could be merged. 🎉
I made a few fixes in another PR because it was notably important that if you're not using the AdMob app id, then the plugin is *not* added to your game - otherwise this would have been included in all the games (requiring more permissions for all games and increasing game size, which is not scalable).

Apart from this, this is great work of a non trivial feature including both React and C++ work. 👏 It will be very useful going forward if we want to use new plugins! We might want to rework this or the project/exporter internals at some point to automatically find the extensions that are used or not in a game - I'm thinking of doing something where when an action/condition/object of an extension is used, then we automatically register that the extension is used. Anyway that's for later! :)
Thanks again for this great PR!

PR #1721, opened 2020-05-02
Search bar, add dirty flag on filters when it change

Bouh:
When the filters on search bar in eventsheet change, the research is set as dirty.
Tested and approved.


arthuro555:
Avec prettier c'est meilleur

4ian:
Sounds good 👍

PR #1722, opened 2020-05-02
Implementing Nord theme

### LINKED ISSUE #1720: New UI theme based on Nord theme (or hi quality dark theme)
This issue is to track the implementation of a **high quality new dark theme** to GDevelop.

I'm interested in getting a theme based on Nord theme: https://www.nordtheme.com/:
* This seems to be [one of the most downloaded theme for IntelliJ/Jetbrains IDE](https://plugins.jetbrains.com/search?headline=164-theme&orderBy=downloads&tags=Theme)
and in the [best rated themes for VS Code](https://marketplace.visualstudio.com/search?target=VSCode&category=Themes&sortBy=Rating) too.
* See the page with [screenshots for VSCode](https://marketplace.visualstudio.com/items?itemName=arcticicestudio.nord-visual-studio-code) and for [IntelliJ](https://plugins.jetbrains.com/plugin/10321-nord)

I can be opened to other themes, provided that they are polished, carefully designed and popular across various existing IDE :)

## How to implement this?

* Setup GDevelop development environment: https://github.com/4ian/GDevelop/blob/master/newIDE/README.md
* Add a new theme here: https://github.com/4ian/GDevelop/blob/997c251a078e4030f7d9c26c241395193cda4759/newIDE/app/src/UI/Theme/index.js#L6-L13
  * As described, copy an existing theme. Changes in the index.js file will be live-reloaded. Changes in CSS files will be hot-reloaded.
  * You should surely apply a CSS filter to icons. You can put it here: 
     https://github.com/4ian/GDevelop/blob/997c251a078e4030f7d9c26c241395193cda4759/newIDE/app/src/UI/Theme/DefaultTheme/index.js#L88
* Get the Nord theme for VS Code/Monaco editor and add it to the themes for the internal code editor (based on Monaco editor): https://github.com/4ian/GDevelop/tree/master/newIDE/app/src/CodeEditor/Theme

We won't use the Nord npm module because I want to keep themes self contained and easy to modify.

## What's the condition for the bounty?

I've put a bounty on this issue to get a high quality PR adding Nord. By high quality I mean:
* Extreme attention to details - this is "just" about setting proper colors, but they should be carefully selected according to what is described on Nord website, and give a nice overall appearance.
  * In particular, be cautious about which color shade must be used for backgrounds, which one for the canvas, which one color for selected items, etc... result should be similar to VSCode/IntelliJ UI. 
  * Unless necessary, don't introduce colors outside of Nord theme, or if you feel you have to, be sure that it's not breaking the overall appearance and that you document it (a quick comment at the top of the index.js for example).
* All files [autoformatted as explained in the README](https://github.com/4ian/GDevelop/blob/master/newIDE/README.md#tests)

## Can I go a bit "wild" and do more changes to the theme?

Themes are mainly based on colors - try to stick with it as much as possible to make the theme easy to maintain... but if you have something that would be really a big plus, you can try to customize a bit of the CSS, for example the CSS in index.js that is passed to [material-ui](https://material-ui.com/styles/basics/). 
Material-UI is the library used to render the UI and while we don't customize it too much, you're free to do cautious changes to some components (search for `createMuiTheme`).

## I can do this, but have another/better theme in mind. Is there a bounty?

I'm open to either another high quality dark/night/"midnight"/"deep ocean"/"pale night" theme or even a new light theme - as I said before, provided that they are polished, carefully designed and popular across various existing IDE. 
In this case, please suggest me the theme and I can open another issue with a similar bounty for a same similar high quality implementation :)

HarsimranVirk:
Resolves #1720.

I've used the nord colour palette in the starting commits. We can use custom css afterwards to make it look really beautiful :heart_eyes: I hope this nord-based theme becomes the most loved by community :)

Screenshot from 2020-05-02 20-39-05

PS: I notice `DebuggerContent.js` and `VariablesContainerInspector.js` in my commits, this is because they weren't formatted properly, and I added them in the commit by mistake
 
 EDIT: Removed those above-mentioned files from this PR :) (Also, tests would fail now due to formatting)


Bouh:
You are so fast :') well done

PawBud:
I must say this will surely be my favourite theme if it rolls out :) 🙏, I always wanted a dynamic (responsive) Background in the start page which can be achieved through animating the CSS gradients!!

HarsimranVirk:
> Btw still need polish.
 
Yes, I have to change some css and find combinations which look really good :) Also can I send some screenshots on discord and ping you for suggestions @Bouh ?

Bouh:
It would be better to follow the recommendations of the theme or see how other softwares has integrated things.
But you can ask to everyone on the community forum/discord.

Bouh:
Think to add the theme on monaco code editor too ;)

4ian:
Thanks for looking at this :) Yeah check all that I listed in the original issue, including the css filter too apply to icons (not sure exactly what, but there is surely something to do to get the icons to better follow the theme colors)

HarsimranVirk:
> Yeah check all that I listed in the original issue, including the css filter to apply to icons

@4ian I don't understand how to use `gdevelopIconsCSSFilter`. What does it expect to be assigned? A css class name? Is there a specific class name I need to use?

4ian:
`gdevelopIconsCSSFilter` is a `filter` applied to icons that are the default GDevelop icons:
https://github.com/4ian/GDevelop/blob/f3dc69ea6860dca776d46bc37db38ddd40abc1a1/newIDE/app/src/UI/ToolbarIcon.js#L41-L50

https://developer.mozilla.org/en-US/docs/Web/CSS/filter

> EDIT: Should I use img classifier directly without any specificity? It's going to affect all the images then.

No please don't add any rule or specifier. I prefer to restrict themes to ensure they don't negatively impact performance.

newIDE/app/src/UI/Theme/DefaultTheme/index.js
```diff
@@ -71,6 +71,8 @@ const gdevelopTheme = {
     separatorColor: '#e0e0e0',
     selectedBackgroundColor: systemSelectionColor,
     selectedTextColor: '#ffffff',
+    rightIconColor: '#111',
+    selectedRightIconColor: '#111',
```
HarsimranVirk, line 75: I've added two more gdevelop specific variables to change the color of secondary action icons (for example the `more menu` icon of object list), we can get this result now:

Screenshot from 2020-05-03 19-46-48

4ian: Mmmm ok I get it! Can you verify what you entered for the other themes though? I'm not sure about the color that were put there.

HarsimranVirk: > Can you verify what you entered for the other themes though?

There was nothing explicitly mentioned, but I can verify that they should be the correct colors, as I can see them being overridden here,
https://github.com/4ian/GDevelop/blob/d6f99c5841d584021993c78ea8ee5f6268583c2e/newIDE/app/src/UI/Theme/DefaultTheme/index.js#L136-L147

4ian: Indeed so this is a color used in the default theme, but in your changes I see it being used for the Dark theme too - is that normal?

4ian: Would you be able to check if it's not something doable with material-ui styles directly? https://material-ui.com/api/list-item/
Let me know. As much as possible, I prefer to avoid overriding the color/style of Material-ui components directly inside our components, because it's making us diverge a bit more from material-ui, which might expose already things in `createMuiTheme` (which remove a bit of code complexity compared to passing "yet another" prop).

HarsimranVirk: > but in your changes I see it being used for the Dark theme too - is that normal?

Oh okay I get it, dark theme doesn't override icon colors and checking on the gdevelop online editor I can see that it uses the usual white color, makes sense :sweat_smile: I'll fix this right now :)

Screenshot from 2020-05-03 20-19-16

HarsimranVirk: > Would you be able to check if it's not something doable with material-ui styles directly?

Maybe there is a way, we can override `selected` style of `ListItem`. This way, I guess we can get rid of the entire listItem gdevelop variables and handle everything with `createMuiTheme`? I'll give it a try and let you know :)

HarsimranVirk: > Maybe there is a way, we can override `selected` style of `ListItem`.

@4ian I can't find any way with this at all, the styles of `selected` are passed through the theme context but they are never applied on the element. I guess this is an issue with MUI itself. Also even if I manage to apply the stylings to the element, I still have to target a nested component (`selected` is a rule of `ListItem`, I have to change the style of `SvgIcon` which is nested withing `ListItem`.)

Also, I can't directly change the style of `SvgIcon` because it is dependent upon the state of `ListItem` :sweat:. 

Another possible solution I think is using a context on SvgIcon so that we don't have to pass any props, but I guess this would be overkill just to change colour.

4ian: Ok, thanks for researching this. In this case don't worry and let's keep these custom styles :)

newIDE/app/src/UI/SortableVirtualizedItemList/index.js
```diff
@@ -103,6 +103,7 @@ export default class SortableVirtualizedItemList<Item> extends React.Component<
           <ScreenTypeMeasurer>
             {screenType => (
               <List
+                style={{overflowX: 'hidden'}}
```
HarsimranVirk, line 106: @4ian I don't know why that horizontal scrollbar popped up. I guess this is a good safety check for future, since redundant scrollbars are bound to come around when, by defalut, overflow is set to auto. What do you suggest?

newIDE/app/src/UI/Theme/NordTheme/index.js
```diff
@@ -94,7 +95,7 @@ const gdevelopTheme = {
   eventsSheetRootClassName: 'gd-events-sheet-nord-theme', // See EventsSheet.css
   tableRootClassName: 'gd-table-nord-theme', // See Table.css
   markdownRootClassName: 'gd-markdown-nord-theme', // See Markdown.css
-  gdevelopIconsCSSFilter: '',
+  gdevelopIconsCSSFilter: 'brightness(110%) hue-rotate(5deg) contrast(90%)',
```
HarsimranVirk, line 98: Since the icons have bluish colour to them, a slight hue-rotation gives them a purplish tint, this should blend in with the color palette of the them, but I'll be experimenting a bit more :)

Screenshot from 2020-05-06 23-09-17

4ian: I think we may want something that is a bit more desaturated. I have the feeling is hard to distinguish the purple from the background. I would almost go to black-and-white icons, with a shade that looks like the "Frost" colors:
image

Not sure how doable it is, but for now they seem a bit hard to read on this background.

newIDE/app/src/UI/SortableVirtualizedItemList/index.js
4ian: I'd like to understand what is going out of the box and making the scrollbar to appear.
If needed, it should be added into `List` - in no cases a list should scroll horizontally, the content should always resize to the available width.

newIDE/app/src/UI/Theme/NordTheme/index.js
HarsimranVirk: @4ian I've used some hsl value transformations to bring in the frost colour into the icons :D 

Screenshot from 2020-05-07 01-47-15

The frost color in icons is quite accurate at some places, with the exact same hsl value (rgb varies slightly) of the nord frost. The obvious problem is that the background of these icons is frost, whereas in the picture you posted above, we can see that the background is dark. 

The icons don't really have that sort of black and white, meaning if I'd use grayscale, either they'll become completely white or completely black :/ 

Anyway, I'd experiment a bit more to see what more can be done, hopefully finding a way to bring some contrast between background and foreground of the icons :) Thank you for your suggestions and inputs :D

4ian: > The icons don't really have that sort of black and white, meaning if I'd use grayscale, either they'll become completely white or completely black :/

Yep, we won't find a perfect solution - apart from switching to SVG icons. But that's other of the scope of this PR.
Give a bit of time to see what's the best you can find - it's ok to diverge a bit from colors as long as it's now too shocking and icon details are visible enough :)

newIDE/app/src/UI/SortableVirtualizedItemList/index.js
HarsimranVirk: > I'd like to understand what is going out of the box and making the scrollbar to appear.

@4ian I did a stupid mistake while merging :sweat_smile::joy: Everything is fine, I'm resolving this thread and pushing a new commit :+1:

4ian: Great news!

newIDE/app/src/UI/List.js
```diff
@@ -127,7 +127,7 @@ export class ListItem extends React.Component<ListItemProps, ListItemState> {
           <ElementWithMenu
             ref={elementWithMenu => (this._elementWithMenu = elementWithMenu)}
             element={
-              <IconButton edge="end" aria-label="menu">
+              <IconButton size="small" edge="end" aria-label="menu">
```
HarsimranVirk, line 130: This was the mistake, by the way, the `IconButton` had a large margin, which made the scrollbar appear :sweat_smile: With a small size, everything is fine :+1:

HarsimranVirk:
@4ian can we put this up on alpha or beta editors so that people could check it out, and give feedback if something looks off?

4ian:
I'll do that now :)

4ian:
Deployed here: https://editor-beta.gdevelop-app.com/ (as usual open, close and re-open to ensure the app is updated).

I like the icons filter 👍
Overall looks good to me.

I've seen that there are extra border on the list item, they should be removed:
image
image

Apart from that, can't see anything really wrong! :)

Bouh:
Text or numeric button for expression should have a white color on icon and text, all button with this color in background use a white color for text.
image

HarsimranVirk:
> Text or numeric button for expression should have a white color on icon and text no ?

@Bouh, I think the white text is not visible on the frost background,
Screenshot from 2020-05-15 20-16-42

What's your suggestion? White text or black?

4ian:
The contrast ratio is ok even if it's true that it's a bit less contrasted than the dark or default theme: 
image

HarsimranVirk:
> The contrast ratio is ok even if it's true that it's a bit less contrasted than the dark or default theme:

Yes, the contrast ratio is a bit less (< 7.0, which means it fails wcag AAA test). However, this is the darkest black of nord color palette, and is also officially used like this: 
Screenshot from 2020-05-15 21-01-43

Bouh:
As you want, try to reach the AAA. Indeed white isn't very visible.

4ian:
I think we can go ahead with this and make adjustments if necessary later :)

Issue #1048, opened 2019-05-06
Resources doesn't load after re-importing them


ddabrahim:
## Describe the bug
In some cases if you re-import resources the objects can not load them even though they are at the location they should be and the path is correct. It is usually happen when you store resources in multiple sub-folders in the project folder.

**EDIT://** I have just noticed when we scan for images they get referenced by their path in their name while the objects in the scene are looking for the images without reference to the path in the name. That should be the problem.

## To Reproduce
1. Create a new project from the Asteroids example
2. Save it. 
3. Open a new instance of GDevelop, keep the one with the Asteroids game open
4. Create a new empty project
5. Copy all the scenes and events from the Asteroids project over to the new empty project.
6. Copy the assets folder from the Asteroids project over to the new empty project folder.
7. Open the asteroids scene in the new empty project.
8. Edit the player object and add the ship_1.png image to the object, this is to have something in the resources.
9. Go to the resources under game settings.
10. Right click the image in the resources we just added to the player before, select scan for images.
11. All the images should be imported now. Save the project.
12. Close the project, or even close GDevelop entirely. Open the new project again we just closed and the objects doesn't load the images even though they are imported in the resources and the path is correct.

* Please include a link to a game if possible!
https://send.firefox.com/download/4ac2409d2a4e275b/#C3Ey94eSjVfigoA8q2NsUQ
It is the copy that doesn't load the resources for me.

## Other details
* Include any OS/browser version/smartphone that you're using
Pop!_OS 18.04

* Which version of GDevelop are you using? The desktop app or the web-app?
Desktop app version 5.0.0-beta68

* Comment
Would be nice if we could right click in the resources even if there is nothing to scan for resources and if we could hot-reload the project without need to close and open it again.

Thanks.

Bouh:
Reference : https://forum.gdevelop-app.com/t/new-video-loader-and-export-problem/18564/1

4ian:
> EDIT:// I have just noticed when we scan for images they get referenced by their path in their name while the objects in the scene are looking for the images without reference to the path in the name. That should be the problem.

Yes that's the "problem". Per se it's not a bug, rather an inconsistency:

* When you add images to objects, images are added with their filenames as the resource name
* When you scan for resources, images that are not in the list of resources are added with their relative path to the project (i.e: folder + filename) as the resource name.

Note that "scan for images" was never meant as a way to "restore" anything, just as a convenient way to add images to your resources :) 

Though I agree it would be best to:
[...]

PR #1724, opened 2020-05-02
Add warning limit for import on sprites

Bouh:
[Trello Card](https://trello.com/c/VY5E4em6/204-warn-the-user-when-an-image-too-large-is-used-which-could-decrease-performance-a-lot)

This work on resource tab when i edit, and when we import on a sprite a new sprite.

image


Bouh:
I've an issue in my workflow, when i made a edit on the codebase the watcher import the new files, 
 electron is refresh but the app is unusable when i load an projet. Even if i refresh and clear the cache.
So i reload again, and the console of electron warn me "DevTools was disconnected from the page".
I've deleted the folder node_modules, and re-intall and re-intall electron but this is still here. temp file from windows is clear too.
There is another folder to clean ?

4ian:
> There is another folder to clean ?

I don't think there is anything else, just be sure to stop any flow (`npm run flow -- stop`)/storybook running. 
This seems like an Electron issue - on my side I've seen something that looks like a memory leak - the `Electron Helper (Renderer)` process is gaining 100mb each time I press Cmd+R on macOS. I'm easily reaching 1gb:

image

making the whole app very slow after a while because the process need to swap I guess. On Windows or more memory constrained machine, I guess this might result in various part of the app crashing like the debugger.

Only solution for now seems to be to kill and restart Electron. Hopefully updating the latest Electron version will solve this :/

Bouh:
Flow is not running in background.
Nothing seem overload my memory, 13GB still free..
I see the app don't load the extensions at start.

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/SpritesList.js
```diff
@@ -196,6 +196,9 @@ export default class SpritesList extends Component<Props, void> {
 
     onChooseResource(sources[0].name).then(resources => {
       resources.forEach(resource => {
+        if (!resource) {
```
4ian, line 199: Any reason for adding these extra check?

newIDE/app/src/ResourcesEditor/ResourcePropertiesEditor/index.js
```diff
@@ -81,7 +81,7 @@ export default class ResourcePropertiesEditor extends React.Component<
     );
     if (!sources.length) return;
     onChooseResource(sources[0].name).then(resources => {
-      if (!resources.length) return; // No path was chosen by the user.
+      if (!resources[0]) return; // No path was chosen by the user.
```
4ian, line 84: This should be equivalent to `!resources.length`, no need to change it I think

4ian:
Not sure if Electron 8.2.5 improved anything. I wonder if it's a Chrome memory leak of GDevelop.js.

Anyway back to the issue, I wonder if it should rather be a warning in ImageThumbnail or, rather than using nativeImage, which is [limited in terms of supported format](https://www.electronjs.org/docs/api/native-image#supported-formats) (because it's cross platform), we can use the browser Image element: https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/Image

I'm not super confident abusing an Electron API that seems made just to handle icons... imagine if we create a memory leak each time an image is added.
Take a look at `Image` in JavaScript. The bonus would be that this would work in the web-app :)

Bouh:
With new Image() I can't get the dimensions in the right scope because if I display showConfirmDialog() the choice Yes or Cancel doesn't change anything...

Bouh:
About the issue with the reloading of the desktop app, when i turn off and on the Service Workers and reload the app, the loading is complete.

4ian:
> Always stuck with violation of chrome -_- these limitations are very annoying.

Might not be a limitation but the fact that you're trying to do something synchronous, while the API you're using in asynchronous :) Creating an `Image` is not immediate. There is the time for loading the image, you have to check the dimension only once it's loaded. That's why I'm thinking of something that is not at the moment where importing is done, but later in the interface.

> About the issue with the reloading of the desktop app, when i turn off and on the Service Workers and reload the app, the loading is complete.

Strange, service workers should not be activated in development.

Bouh:
Ok then, i need promise and async function.

The main goal is to not load the image if this one is huge. If we load it for read the info and warn the user it's not terrible.

Bouh:
It's not yet optimal.
But for now there is this icon for image bigger than 2048px wide.
image

Bouh:
In resource tab.
Add icon in resource list.
I should add a icons instead the sprite or type icon if the resource have an warning/error.

image

Bouh:
For now the error message is on every resource because the props didn't exist.
Have you can idea how flag an resource if there is an issue with it ?
Add a new attribute in c++ on the resource, or check each resource when the resource tab is open?

Bouh:
I'am wondeing for know if n texture is too large , we can add a new attribute on the resource (gd.Resource).
This way i we can get the new attribute, if this one isn't empty we display an error/warning icon with [_getResourceThumbnail()](https://github.com/4ian/GDevelop/pull/1724/files#diff-a75953006aea60bc264bc0daa361abbbR176).
And i can use it for the warning message on the resource.

La resource aurait ce nouvel attribue à [ce niveau](https://github.com/4ian/GDevelop/pull/1724/files#diff-a613d5cb938d829192ce8865ded11bb3R117). on ajouterais une ligne.
```
type: String = 'TEXTURE_TOO_BIG'
 imageResource.setWarning(path.relative(projectPath, resourcePath, type));
```

And we can use this new attribute for other ressource if we need to check the mass of the file in kb for avoid to load a big video file of 4Gb for example. Or check the formats of the codecs.

4ian:
> check each resource when the resource tab is open?

That looks like the simplest approach. If we display both in the resources editor (`ImagePreview`) and in the thumbnails shown in the editor (`ImageThumbnail`, `ResourceThumbnail`), that should be enough to warn about the issue.

> But for now there is this icon for image bigger than 2048px wide.

I like how it's displayed. 👍

> I'am wondeing for know if a texture is too large , we can add a new attribute on the resource (gd.Resource).
> This way i we can get the new attribute, if this one isn't empty we display an error/warning icon with _getResourceThumbnail().

Well we could, but when would you analyze the image and store the result? At every opening? At every resource editor opening? When adding an image - but what if the image is changed?
That would be part of a larger feature to analyze the game resources I think - so that's why I would prefer to stay simple and limit this to the size that we can get when we display the image on screen :) 
The red warning should be enough because, as a user, you often see the images when editing objects :)

newIDE/app/src/ResourcesList/ResourceThumbnail/ImageThumbnail.js
```diff
@@ -48,6 +56,15 @@ const ImageThumbnail = ({
...
   muiTheme,
 }) => {
+  let warningSize = false;
+
+  const img = new Image();
+  img.src = resourcesLoader.getResourceFullUrl(project, resourceName);
+
+  if (img.width > 2048 || img.height > 2048) {
+    warningSize = true;
+  }
```
4ian, line 66: Could you do this as a callback to the `` element that is below?
`onLoad` prop should be called with the `HTMLImageElement` as parameter, on which you can read width/height. At this time, prefer to set the state (use a hook: `const [hasSizeWarning, setHasWarningSize] = React.useState(false);` then `setHasWarningSize(true);` instead of `warningSize = true`).

Bouh: Thank you it's better :)

Bouh:
I've found a strange behavior in the sortablelist with the sprite selected, and obisiously with my warning icon.

mmmmm

The first sprite loaded is really big, the asteroid is smaller than 2048px.
As you can see the sprite selected or the sprite with the new icon propagate the state to other if we change the order in the list.

Bouh:
In https://github.com/4ian/GDevelop/pull/1724/commits/2d4a5524f639a286949b6be72701a339983cae48

image

- I've moved showConfirmDialog(), but this dialog in now always trigger because of onload on the image :( 

I need finish polish warning for resource tab.

4ian:
I lost a bit track of what is done and still work in progress here :) Need help?

Bouh:
I've found something for save the state for each resource, use the **ResourcesLoader**.
This way i can call a property from **ResourcesLoader** in **ImageThumbnail.js**.

https://github.com/4ian/GDevelop/blob/01e2bd81665d87eddb59f4ebe3069b328bbf6399/newIDE/app/src/ResourcesEditor/ResourcePropertiesEditor/index.js#L106-L109
This work!

The property isn't define at the good place, but where this one is define we can juste reuse the texture from pixi for get the size, this is really cool instead use an HTML image.

I need guidance, if we can talk about this in a voice channel this will be great.

Bouh:
This is a first step, it's a beautiful spaghetti code.
As we have discuted by mail, the watcher is present and watch change on all existing resources.
When i edit a sprite with more than 2048px and save, the watcher see it, check the dimensions and notify the resourceLoader and give a warning status for the resource,
This status is used in the resources editor and in the ImageThumbnail.

Next step are:

- [ ] Check the dimensions when the project is loaded/when the watchers are started.
- [ ] I need clean my code as possible.

4ian:
I've taken a quick look :) 
The watcher is very interesting, and should probably be used to "reload" a resource if it's modified too. It's also a part of the code that can be fragile (I know some issues with watching too much files on Windows/Linux), so would need to have careful testing and avoid crashes in case something goes wrong here.
The watching is naively done for GDJS, and can be suboptimal, I think we should take a look at chokidar: https://github.com/paulmillr/chokidar or other solution for watching files in Node.js :)

Bouh:
- Display the warning/error related to the issue on the ressource in ressource tab when the ressource is selected.
- Use chokidar instead node.fs, chokidar have a better detection of the events on files. (Thank 4ian!)
- 

Next step are:

- [ ] Check the dimensions when the project is loaded/when the watchers are started.
- [ ] Avoid the mutiple alert message on same file for valid the import of big sprites.
- [ ] Take care of broken image/file not found as warning in ressources tab.
- [ ] I need clean my code as possible (make functions, polish sentences)
- [ ] Use the preferences button of this feature for enable/disable the auto scan of ressources at the startup. 
(This will not show the warning with on the ressources tab, unless we have open the object propertie before.)
- [ ] When the file change, this trigger a new check on their dimensions, but in the ressource tab it's not refresh, only the preview is always correct, the warning message and the warning in ressources list should disappear. Happen even if we close/reopen the tab.
-

Silver-Streak:
@Bouh As I go through older issues, I saw that this PR is flagged as solving https://github.com/4ian/GDevelop/issues/1048

Do you still feel that is accurate? If so, is there anything still pending on this PR, or just a review from 4ian?

Bouh:
I admit I've forgot to continue this, some things from my PR are useful, I should reduce and clean the unused code. 

The import message is may be not the best experience, it's triggered each time we open or select an image. I'm a bit worried about that. 

In next months I hope see the diagnostic features, I'll be focused on communication and networks rather than coding for GDevelop.

4ian:
Yes the PR would need a lot of rework as we prefer a more "diagnostic report" oriented approach.

PR #1727, opened 2020-05-03
Rework margins in the whole editor

4ian:
This is an effort to reduce the space lost with margins and reduce the size of list items.
This should still result in a UI that is not intimidating, but better to use when dealing with a lot of items/optimize a bit screen space usage.

### Reduce list item heights to 32px:
* All lists showing items have items with height of 32px
* Toolbars height: 32px
* Remove padding around texts in tables to ensure 32px height

image
image
image

### Densify form controls:
* Use spacers (4px) between form controls (ColumnStackLayout)
  * Adapt all editors to use ColumnStackLayout
* Use small version of IconButton
* Cancel margins around checkboxes
image
image

### Normalize dialogs margins:
* All dialog titles margins are 8px
* All dialog content margins are 8px (same as Column margins)
image
image

---

### The 3-dots menu icons are not aligned:

See the mis-alignment:
image

Should the margins be reduced to 8px for list items? Before:
image

After:
image

### Should tabs be smaller in height? 

Should they be 32px too for consistency? They already stand out by their color:
image
image

### DIFF
--- a/newIDE/app/src/BehaviorsEditor/Editors/BehaviorPropertiesEditor.js
+++ b/newIDE/app/src/BehaviorsEditor/Editors/BehaviorPropertiesEditor.js
@@ -32,7 +32,7 @@ export default class BehaviorPropertiesEditor extends React.Component<Props> {
     );
 
     return (
-      <Column>
+      <Column expand>
         {propertiesSchema.length ? (
           <PropertiesEditor
             schema={propertiesSchema}
--- a/newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/PolygonEditor.js
+++ b/newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/PolygonEditor.js
@@ -136,7 +136,7 @@ export default class PolygonEditor extends React.Component<Props> {
                       />
                     </TableRowColumn>
                     <TableRowColumn>
-                      <IconButton onClick={() => onRemove(index)}>
+                      <IconButton size="small" onClick={() => onRemove(index)}>
                         <Delete />
                       </IconButton>
                     </TableRowColumn>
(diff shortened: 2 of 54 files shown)


Bouh:
>Should the margins be reduced to 8px for list items? 
Before or After.
After for me but with some tweak, "after" choice is better but the indentation is still missing in this list.
An indentation on each child item is required.
All list have an indentation on child.
And about border, there is no separator/border on each parent sections and childs.
Only an border on the last item for add a new scene because this is an different button, this button have a distinct action in term of UX.

I suggest:
image

I stay on my positions about [separator/border in list.](https://github.com/4ian/GDevelop/pull/1690#issuecomment-619575602).

>Should tabs be smaller in height?

Yes tabs are fat.

Wend1go:
The condensed versions look way better. Especially in the custom behaviour screen, the huge parameter entries and titles take up a lot of space.
Smaller tabs would be a great improvement too.

4ian:
Thanks for your feedbacks :) 
I'll go ahead and reduce the tabs too.

> the indentation is still missing in this list.

Right, I'll check that but the indentation is the gutter applied to all list elements. I'll check to reduce it only on the right (for buttons) :)

For the borders, I'm open to changing this too. 
I'll push changes later today :)

4ian:
Merged! @Bouh I've not reworked borders because it was slightly out of scope of what I wanted to do. This could be done in another PR.

HarsimranVirk:
@4ian I know this PR is already closed, but I guess this PR introduced another redundant horizontal scrollbar in ObjectsList:

Screenshot from 2020-05-06 22-08-31

No worries though, I'll push a quick-fix for this in a new pr :)

Also, this definitely makes everything look pretty, reducing those huge margins is really good. It makes everything look less web-ish :+1:

4ian:
> Also, this definitely makes everything look pretty, reducing those huge margins is really good. It makes everything look less web-ish 👍

Glad that you like it! 

> EDIT: I guess there's some regression in my own machine after merging, everything looks good on master branch, sorry about this :)

Ok if you are able to reproduce it on master feel free to let us know and/or make a PR 👍

PR #1730, opened 2020-05-04
Change type of BBText fontSize

### LINKED ISSUE #1627: Problems with BBCode size change
The action to modify the font size does not work properly when using the + operator. It gets the BBCode object frozen.
When using several decimals, it even crashes the preview.

I'm attaching a sample project. You'll notice the tweening doesn't work either.
[bugtextscale.zip](https://github.com/4ian/GDevelop/files/4440607/bugtextscale.zip)

Quarkstar:
Fix #1627 
Fix the type of `fontSize` and `opacity` of BBText.

### Screenshot
Peek 2020-05-04 11-19

### DIFF
--- a/Extensions/BBText/JsExtension.js
+++ b/Extensions/BBText/JsExtension.js
@@ -35,6 +35,8 @@ module.exports = {
       if (propertyName in objectContent) {
         if (typeof objectContent[propertyName] === 'boolean')
           objectContent[propertyName] = newValue === '1';
+        else if (typeof objectContent[propertyName] === 'number')
+          objectContent[propertyName] = parseFloat(newValue);
         else objectContent[propertyName] = newValue;
         return true;
       }
@@ -67,7 +69,7 @@ module.exports = {
 
       objectProperties.set(
         'fontSize',
-        new gd.PropertyDescriptor(objectContent.fontSize)
+        new gd.PropertyDescriptor(objectContent.fontSize.toString())
           .setType('number')
           .setLabel(_('Base size'))
       );
@@ -111,7 +113,7 @@ module.exports = {
         text:
           '[b]bold[/b] [i]italic[/i] [size=15]smaller[/size] [font=times]times[/font] font\n[spacing=12]spaced out[/spacing]\n[outline=yellow]outlined[/outline] [shadow=red]DropShadow[/shadow] ',
         opacity: 255,
-        fontSize: '20',
+        fontSize: 20,
         visible: true,
         color: '#000000',
         fontFamily: 'Arial',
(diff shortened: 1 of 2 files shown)


4ian:
Good catch. This is because properties "opacity" and "fontSize"  are not properly saved in JsExtension.js, they are saved as strings.

Can you:
* Use `parseFloat` instead of `parseInt`?
* In JsExtension.js:
  * Replace `new gd.PropertyDescriptor(objectContent.fontSize)` by `new gd.PropertyDescriptor(objectContent.fontSize.toString())` (because we have a number) (opacity is already ok)
  * Replace `fontSize: '20',` by `fontSize: 20,` (because we have a number) (opacity is already ok)
  * Replace:

```
        if (typeof objectContent[propertyName] === 'boolean')
          objectContent[propertyName] = newValue === '1';
        else objectContent[propertyName] = newValue;
```

by 

```
        if (typeof objectContent[propertyName] === 'boolean')
          objectContent[propertyName] = newValue === '1';
        else if (typeof objectContent[propertyName] === 'number')
          objectContent[propertyName] = parseFloat(newValue);
        else objectContent[propertyName] = newValue;
```

(number were not handled properly).

Finally, let's add a comment next to the two `parseFloat` you added:
`// parseFloat should not be required, but GDevelop 5.0 beta 92 and below were storing it as a string.`

Bouh:
In mean while i've read a bit the BBtext extension and on the action (maybe on other place too) for fontFamily i see the type is "string" in parameterType. 

Compare to the Text object, the parameterType is "police".
`Extensions\TextObject\Extension.cpp`  Line 72

```
  obj.AddAction("Font",
                _("Font"),
                _("Change the font of the text."),
                _("Change font of _PARAM0_ to _PARAM1_"),
                _("Font"),
                "res/actions/font24.png",
                "res/actions/font.png")

      .AddParameter("object", _("Object"), "Text")
      .AddParameter("police", _("Font"))
      .SetFunctionName("ChangeFont")
      .SetIncludeFile("TextObject/TextObject.h");
```

4ian:
Merged! Thanks for these fixes 👍Types are important.

> Compare to the Text object, the parameterType is "police".

(we should really rename this to font ^^' #legacyfrenchnames)

> for fontFamily i see the type is "string" in parameterType.

Yep that should be fixed. Good catch! Might need to go back to the "official" api without helper to write this properly - or just add another if/else to the helper function.

> Why these slashs at the start and the end ? Because this isn't in the BBtext, maybe another cause of the difference.

This is I think to get the font family to work when there is a space in it - but might need to be double checked.

PR #1737, opened 2020-05-07
Renamed buildResources to build

arthuro555:
As you can see on [the electron-builder Documentation](https://www.electron.build/configuration/configuration#MetadataDirectories-buildResources), the default buildResources folder name is build, and not build resources. Users who manually build always wonder why the icons don't show up because they don't know they need to override that in the command or don't expect to have to rename the directory.

This is only a little change here but it needs to be changed on the build service too. I would say that the best way is to look into data.js for the gd version and not add the buildResources argument to the build command for versions above or equal to <the next release>


4ian:
Can we instead add the property configuration in the generated package.json, to keep "backward compatibility"? Thanks! :)

arthuro555:
Ok done and working :)

PR #1740, opened 2020-05-08
Add Is Preview condition

arthuro555:
Useful for debugging purposes.


arthuro555:
Uhm are there premade typings for the abstract file system? I have no idea how to write flow typedefs

arthuro555:
Nvm figured it out

arthuro555:
Ready for review @4ian

Bouh:
About the preview, there is a menu bar even in fullscreen.

image

This affect the resolution of the game. Here 1920x1080, but the canvas is small because of the menu bar.
Then, can we add an action for toggle this bar?
Or just add here
https://github.com/4ian/GDevelop/blob/c268b19264cd9faaedd677df6f38cdea3215c7cb/GDJS/Runtime/pixi-renderers/runtimegame-pixi-renderer.js#L232-L234

this line:
```
browserWindow.setMenu(null); //Remove Electron menu
```

arthuro555:
I think that's something for another pr

arthuro555:
@4ian This is ready to merge

arthuro555:
I feel like this PR got forgotten

4ian:
Sorry about the late review.
The solution of doing a text replacement can work, but:
* I'm not sure this work for the web-app. Indeed, `gd.js` is not considered as a file to download here: https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/GameEngineFinder/BrowserS3GDJSFinder.js
* More generally, I would prefer to reduce as much as possible the changes to source files, especially source code of the engine. If we really had too, we should at least mark it very explicitly with a marker (gdjs.isPreview = false; /* GDJS_IS_PREVIEW */) to give a hint that it's likely to be replaced. Otherwise, we create risking confusion and long research about "why is this boolean changing by itself?" or worse "why is this boolean not changing anymore" when someone naively reformat this line.

But I think we can do without all of this :) There is already a way to pass additional argument to a runtimeGame. It's called `additionalSpec`:
https://github.com/4ian/GDevelop/blob/9d2bff944218616c03efab61f7eab1c6617bf83c/GDJS/GDJS/IDE/ExporterHelper.cpp#L520

This object is then written in the index.html and passed to RuntimeGame constructor. You can pass a `isPreview: true` to it.

I let you find where to change it in the codebase, should be in Exporter.cpp or ExporterHelper.cpp I guess.

Extensions/SystemInfo/Extension.cpp
```diff
@@ -42,6 +42,21 @@ void DeclareSystemInfoExtension(gd::PlatformExtension& extension) {
...
+          _("Is the game running in a preview"),
+          _(
+            "Check if the game is currently being previewed. Can be used to enable a Debug mode"
+          ),
+          _("The game is a preview"),
+          _("System information"),
+          "CppPlatform/Extensions/systeminfoicon24.png",
+          "CppPlatform/Extensions/systeminfoicon16.png")
+
+      .SetFunctionName("SystemInfo::IsMobile")
```
4ian, line 58: Don't set a function name nor an include file.

GDJS/Runtime/gd.js
```diff
@@ -358,6 +358,11 @@ gdjs.objectsListsToArray = function(objectsLists) {
   return result;
 };
 
+/**
+ * Is true on previews.
+ */
+gdjs.isPreview = false;
```
4ian, line 364: Avoid global variables as much as possible, I know some game engines that are overbloated of stuff in a global namespace. Prefer to put it into `RuntimeGame`.

It's of course not the case, but it might happen that in the future we could run more than one RuntimeGame at the same time :)

newIDE/app/src/Export/BrowserExporters/BrowserS3PreviewLauncher/index.js
```diff
@@ -41,6 +41,16 @@ export default class BrowserS3PreviewLauncher extends React.Component<
     };
   };
 
+  _markAsPreview = (fs: gdAbstractFileSystem, outputDir: string): void => {
```
4ian, line 44: See my other comment, you should be able to do everything without adding yet another file manipulation.

Extensions/SystemInfo/systeminfotools.js
```diff
@@ -36,4 +36,12 @@ gdjs.evtTools.systemInfo.isMobile = function() {
  */
 gdjs.evtTools.systemInfo.isWebGLSupported = function(runtimeScene) {
 	return runtimeScene.getGame().getRenderer().isWebGLSupported();
-};
\ No newline at end of file
+};
+
+/**
+ * Checks gdjs.isPreview.
```
4ian, line 42: Don't comment [what or how the code](https://blog.codinghorror.com/code-tells-you-how-comments-tell-you-why/) is doing - people can just read the code if they want to know.
Instead, comment about the why or the result: "Check if the game is run as a preview in an editor".

4ian:
Do you want to try reworking this PR? Look at passing something in additionalSpec `ExportLayoutForPixiPreview` and get this in the `gdjs.RuntimeGame` :)

arthuro555:
Yes, I saw. I have just waay too many PRs going on already and try to focus on other. This one needs to be reworked from the ground up and I prefer working on ones I have already progress on, because now working on that PR is like making a new one.

arthuro555:
Ok I did the rework @4ian to use additional specs. It's ready for reviewing again.

GDJS/Runtime/runtimegame.js
```diff
@@ -63,6 +63,12 @@ gdjs.RuntimeGame = function(data, options) {
   this._debuggerClient = gdjs.DebuggerClient
     ? new gdjs.DebuggerClient(this)
     : null;
+
+  /**
+   * Property passed by the IDE when the current export is a preview.
+   * @type {boolean}
+   */
+  this.isPreview = spec.isPreview | false;
```
4ian, line 71: `|| false`

Also can you make this "private" (_isPreview) and add a method `isPreview()` - to avoid tempting anyone to do a `isPreview = true`.

4ian:
A few comments, then should be good to go! :) Thanks!

Extensions/SystemInfo/Extension.cpp
```diff
@@ -42,6 +42,21 @@ void DeclareSystemInfoExtension(gd::PlatformExtension& extension) {
...
       .SetFunctionName("SystemInfo::IsWebGLSupported")
       .SetIncludeFile("SystemInfo/SystemInfoTools.h");
+
+  extension
+      .AddCondition(
+          "IsPreview",
+          _("Is the game running in a preview"),
+          _(
+            "Check if the game is currently being previewed. "
+            "Can be used to enable a 'Debug mode'"
```
4ian, line 52: period at the end of the sentence.

arthuro555:
Uhm the CI says something weird. Spec undefined? It worked fine when I tested.

Extensions/SystemInfo/Extension.cpp
4ian: `"Debug mode" or do some work only in previews.`

arthuro555:
Ok I think I've done it right now. It's passing CI and I've corrected what you did. I am sorry I messed up that hard I am a bit sleepy because I couldn't sleep well last night.

Extensions/SystemInfo/systeminfotools.js
```diff
@@ -36,4 +36,13 @@ gdjs.evtTools.systemInfo.isMobile = function() {
  */
 gdjs.evtTools.systemInfo.isWebGLSupported = function(runtimeScene) {
 	return runtimeScene.getGame().getRenderer().isWebGLSupported();
-};
\ No newline at end of file
+};
+
+/**
+ * Checks in runtimeGame for the additional isPreview spec.
```
4ian, line 42: I think I already made this comment: you need to comment to give **additional** descriptions, i.e: **extra** context. But not to describe the code. Someone can always take a look at the code if they want to know what the code is doing.

`Check if the game is running as a preview, launched from an editor.`

GDJS/Runtime/runtimegame.js
```diff
@@ -548,3 +555,11 @@ gdjs.RuntimeGame.prototype.wasFirstSceneLoaded = function() {
 gdjs.RuntimeGame.prototype.getSceneStack = function() {
   return this._sceneStack;
 }
+
+/**
+ * Checks if the IDE marked the current build as preview.
```
4ian, line 560: nitpicking: `Checks if the game is running as a preview, launched from an editor.`

("build" is not a term that we employ in previews).

Extensions/SystemInfo/Extension.cpp
```diff
@@ -42,6 +42,21 @@ void DeclareSystemInfoExtension(gd::PlatformExtension& extension) {
       .AddCodeOnlyParameter("currentScene", "")
       .SetFunctionName("SystemInfo::IsWebGLSupported")
       .SetIncludeFile("SystemInfo/SystemInfoTools.h");
+
+  extension
+      .AddCondition(
+          "IsPreview",
+          _("Is the game running in a preview"),
```
4ian, line 49: `running as a preview`?

4ian:
Alright, still 3 nitpickings and we should be good to go :)

4ian:
Done the updates myself. Thanks for adding this!

arthuro555:
Sorry I've not done them myself, I got a splitting headache and couldn't focus on it.

PR #1750, opened 2020-05-12
Refactor editor containers and their usage in MainFrame

4ian:
* Ensure all editor are properly flow typed - hopefully avoiding any regression in the future.
* Avoid using potentially-stale closures in MainFrame for editors that would be bug prone.
* Slightly reduce the complexity inside MainFrame now that these risky closures are not necessary anymore. Make the file a bit shorter :)


HarsimranVirk:
@4ian is the MainFrame stable enough now? If it is then maybe I could start working on the Open Recent Project PR 😅

newIDE/app/src/MainFrame/index.js
```diff
@@ -211,6 +213,7 @@ const MainFrame = (props: Props) => {
     openSubscriptionDialog,
   ] = React.useState<boolean>(false);
   const [exportDialogOpen, openExportDialog] = React.useState<boolean>(false);
+  const preferences = React.useContext(PreferencesContext);
 
   React.useEffect(() => {
```
HarsimranVirk, line 218: Reminder: This has to be removed if it isn't needed anymore :)

4ian: Yep! I'll remove this soon if we don't find any issues.

4ian:
> is the MainFrame stable enough now? If it is then maybe I could start working on the Open Recent Project PR 😅

I think it is! At least if there is an issue, I feel confident fixing it quickly :)

newIDE/app/src/MainFrame/index.js
4ian: Thanks for the reminder ;)

PR #1762, opened 2020-05-18
Add functionality to open recent projects

HarsimranVirk:
@4ian most of the stuff is done, I couldn't figure out how to call a function with arguments with ipcRenderer, can you help me?

### DIFF
--- a/newIDE/app/src/MainFrame/ElectronMainMenu.js
+++ b/newIDE/app/src/MainFrame/ElectronMainMenu.js
@@ -91,7 +91,7 @@ const buildAndSendMenuTemplate = (project, i18n, recentProjectFiles) => {
         submenu: recentProjectFiles.map(item => ({
           label: item.fileIdentifier,
           onClickSendEvent: 'main-menu-open-recent',
-        }))
+        })),
       },
       { type: 'separator' },
       {
--- a/newIDE/app/src/MainFrame/ElectronMainMenu.js
+++ b/newIDE/app/src/MainFrame/ElectronMainMenu.js
@@ -90,7 +90,7 @@ const buildAndSendMenuTemplate = (project, i18n, recentProjectFiles) => {
       {
         label: i18n._(t`Open Recent`),
         submenu: recentProjectFiles.map(item => ({
-          label: item.fileIdentifier,
+          label: item.fileMetadata.fileIdentifier,
           onClickSendEvent: 'main-menu-open-recent',
           eventArgs: item,
         })),
--- a/newIDE/app/src/MainFrame/ElectronMainMenu.js
+++ b/newIDE/app/src/MainFrame/ElectronMainMenu.js
@@ -32,6 +32,7 @@ type MenuItemTemplate =
       accelerator?: string,
       enabled?: boolean,
       label?: string,
+      eventArgs?: any,
     |}
   | {|
       submenu: Array<MenuItemTemplate>,
@@ -63,9 +64,8 @@ const useIPCEventListener = (ipcEvent: MainMenuEvent, func) => {
   React.useEffect(
     () => {
       if (!ipcRenderer) return;
-      const handler = (event, ...eventArgs) => func(...eventArgs);
-      ipcRenderer.on(ipcEvent, handler);
-      return () => ipcRenderer.removeListener(ipcEvent, handler);
+      ipcRenderer.on(ipcEvent, func);
+      return () => ipcRenderer.removeListener(ipcEvent, func);
     },
     [ipcEvent, func]
   );
@@ -91,6 +91,7 @@ const buildAndSendMenuTemplate = (project, i18n, recentProjectFiles) => {
         submenu: recentProjectFiles.map(item => ({
           label: item.fileIdentifier,
           onClickSendEvent: 'main-menu-open-recent',
+          eventArgs: item,
         })),
       },
       { type: 'separator' },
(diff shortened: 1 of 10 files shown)


newIDE/app/src/MainFrame/ElectronMainMenu.js
```diff
@@ -85,6 +86,13 @@ const buildAndSendMenuTemplate = (project, i18n) => {
         accelerator: 'CommandOrControl+O',
         onClickSendEvent: 'main-menu-open',
       },
+      {
+        label: i18n._(t`Open Recent`),
+        submenu: recentProjectFiles.map(item => ({
+          label: item.fileIdentifier,
+          onClickSendEvent: 'main-menu-open-recent',
```
HarsimranVirk, line 93: This wouldn't work obviously, because the function expects `fileMetadata` as argument.

newIDE/app/src/MainFrame/Preferences/PreferencesProvider.js
```diff
@@ -343,6 +346,46 @@ export default class PreferencesProvider extends React.Component<Props, State> {
...
+    this.setState(
+      state => ({
+        values: {
+          ...state.values,
+          recentProjectFiles: recents,
+        }
+      }),
+      () => this._persistValuesToLocalStorage(this.state)
+    );
+  }
```
HarsimranVirk, line 363: I decided not to expose this setter function, because insertion in recent files has a particular method (which is sort of LIFO, but not exactly, meaning if a file already exists in the array, it should come on top).

newIDE/app/src/MainFrame/index.js
```diff
@@ -1507,6 +1508,15 @@ const MainFrame = (props: Props) => {
...
           onChooseProject: chooseProject,
+          onOpenRecentFile: (fileMetadata: FileMetadata) => {
+            openFromFileMetadata(fileMetadata).then(state => {
+              if(state)
+                openSceneOrProjectManager({
+                currentProject: state.currentProject,
+                editorTabs: state.editorTabs,
+                });
+            })
+          },
```
HarsimranVirk, line 1519: This is the function I need to call through `ipcRenderer.on('event-name', props.onOpenRecentFile)` but I couldn't figure out how to pass the `fileMetadata`. Need some help :/

newIDE/app/src/MainFrame/Preferences/PreferencesProvider.js
4ian: Yes it's fine, the context should only expose functions that are safe and straightforward to use like `_insertRecentProjectFile`. The setter is too powerful to put it in the end of the components, they simply don't need it ;)

newIDE/app/src/MainFrame/ElectronMainMenu.js
4ian: We probably need to extend the `onClickSendEvent` so that it also send an optional value stored in a new `eventArgument` property: https://github.com/4ian/GDevelop/blob/93e0ccc1634972b42a47f7c2e5c770b310900c82/newIDE/electron-app/app/MainMenu.js#L24-L26

Then, I think `useIPCEventListener` is already calling the function with the arguments passed to the event from Electron.

newIDE/app/src/MainFrame/ElectronMainMenu.js
```diff
@@ -63,9 +64,8 @@ const useIPCEventListener = (ipcEvent: MainMenuEvent, func) => {
   React.useEffect(
     () => {
       if (!ipcRenderer) return;
-      const handler = (event, ...eventArgs) => func(...eventArgs);
-      ipcRenderer.on(ipcEvent, handler);
-      return () => ipcRenderer.removeListener(ipcEvent, handler);
+      ipcRenderer.on(ipcEvent, func);
```
HarsimranVirk, line 67: There was no need to remove the handler, a mistake.

HarsimranVirk:
@4ian should be done now unless you find something wrong :)

newIDE/app/src/MainFrame/MainMenu.flow.js
```diff
@@ -1,11 +1,13 @@
 // @flow
 import { type I18n as I18nType } from '@lingui/core';
 import { type UpdateStatus } from './UpdaterTools';
+import { type FileMetadata } from '../ProjectsStorage';
 
 export type MainMenuProps = {|
   i18n: I18nType,
   project: ?gdProject,
   onChooseProject: () => void,
+  onOpenRecentFile: (FileMetadata: FileMetadata) => void,
```
4ian, line 10: Nitpicking: `fileMetadata: FileMetadata`

newIDE/app/src/MainFrame/Preferences/PreferencesProvider.js
```diff
@@ -343,6 +346,49 @@ export default class PreferencesProvider extends React.Component<Props, State> {
...
+      const len = recentProjectFiles.length > 5 ? 5 : recentProjectFiles.length;
+      recentProjectFiles = [
+        ...recentProjectFiles.slice(0, index),
+        ...recentProjectFiles.slice(index + 1, len),
+      ];
+    }
+
+    recentProjectFiles = [fileMetadata, ...recentProjectFiles];
+    this._setRecentProjectFiles(recentProjectFiles);
+  }
```
4ian, line 390: Should be decomposable, in 3 simple steps. Add the file at top, remove it from anywhere it was before, ensure the list has a maximum length: 

```suggestion
  _insertRecentProjectFile(newFileMetadata: FileMetadata) {
    const recentProjectFiles = this._getRecentProjectFiles();
    const isNotNewFileMetadata = (fileMetadata) =>
        JSON.stringify(fileMetadata) !== JSON.stringify(newFileMetadata);
    this._setRecentProjectFiles(
        [fileMetadata, ...recentProjectFiles.filter(isNotNewFileMetadata)].slice(
        0,
        MAX_RECENT_FILES
        )
    );
  }
```

newIDE/app/src/MainFrame/index.js
```diff
@@ -1507,6 +1508,15 @@ const MainFrame = (props: Props) => {
           i18n: i18n,
           project: state.currentProject,
           onChooseProject: chooseProject,
+          onOpenRecentFile: (fileMetadata: FileMetadata) => {
+            openFromFileMetadata(fileMetadata).then(state => {
```
4ian, line 1512: We're missing one last - but important - "detail". Turns out that openFromFileMetadata will open the fileMetadata passed as argument with the current "storage provider". 
On the Electron app, it's simple: there is a unique storage provider, which is `LocalFileStorageProvider`.
But on the web-app, which might have a MainMenu in the future, we can have mutliple storage providers. One FileMetadata can have been opened with "GoogleDrive", another one with "Dropbox", another one with "OneDrive". 
This means that we must:
1. Know the name of the storage where the file was originally saved.
2. Find this storage in our list of storage providers.
3. Tell GDevelop to use it before calling openFromFileMetadata (`openFromFileMetadata` will use the current one).

For 1, this means that instead of FileMetadata, you must pass something that is `FileMetadataAndStorageProviderName`. I recommend something like `type FileMetadataAndStorageProviderName = {| fileMetadata: FileMetadata, storageProviderName: string |}`.
Using this, it's easy to do 2: in the `props`  of `MainFrame`, get `storageProviders`, filter them to find the one with `internalName` === `storageProviderName`.
If none is found, use showErrorBox to tell the user that this file can't be opened as GDevelop does not know how to open this file. (should not happen - but better be defensive!)
Finally, `getStorageProviderOperations` pass it as argument the `storageProvider` you found - this will change the storage provider being used (I know, `getXXX` is a terrible name, have to change it). In the promise, call `openFromFileMetadata`.

This means that you have to update the whole context and most part of the PR to use FileMetadataAndStorageProviderName (sorry, a bit of refactoring 😅).. and this means that when you call `insertRecentProjectFile`, you have to pass the FileMetadata and the internal name of the storage provider. 
How to get this internal name? I recommend modifying `ProjectStorageProviders` to expose `getStorageProvider` (in addition to `getStorageProviderOperations`), which returns `this.state.currentStorageProvider || emptyStorageProvider`.

And that should be it! Note there is no risk of stale props because the props are getter functions that are not changing when we change the storage provider.

How to test all of this. Check that it's working on the electron app and I can check the web-app.

Give it a try but if it's too complicated I can do this myself - leave in this case a TODO here.

HarsimranVirk: @4ian your description of the way to call `openFromFileMetadata()` seems very similar to `chooseProjectWithStorageProviderPicker()` without the picker. So, more or less I have to store the name of `StorageProvider` and call a modified `chooseProjectWithStorageProviderPicker()` right?

4ian: >  So, more or less I have to store the name of StorageProvider and call a modified chooseProjectWithStorageProviderPicker() right?

Yes that's it - maybe a new function called "openFromFileMetadataWithStorageProvider", something like this.

newIDE/app/src/MainFrame/index.js
```diff
@@ -377,8 +379,13 @@ const MainFrame = (props: Props) => {
...
     fileMetadata: ?FileMetadata
   ): Promise<State> => {
-    const { eventsFunctionsExtensionsState } = props;
-    if (fileMetadata) preferences.insertRecentProjectFile(fileMetadata);
+    const { eventsFunctionsExtensionsState, getStorageProvider } = props;
+
+    if (fileMetadata)
+      preferences.insertRecentProjectFile({
+        fileMetadata,
+        storageProviderName: getStorageProvider().internalName,
```
HarsimranVirk, line 387: @4ian `getStorageProvider` always returns "Empty", am I missing something? Apart from this, everything should be working fine.

newIDE/app/src/MainFrame/index.js
```diff
@@ -1507,6 +1541,12 @@ const MainFrame = (props: Props) => {
           i18n: i18n,
           project: state.currentProject,
           onChooseProject: chooseProject,
+          onOpenRecentFile: (
```
4ian, line 1544: nitpicking: `onOpenRecentFile: openFromFileMetadataWithStorageProvider `

newIDE/app/src/MainFrame/index.js
4ian: Mmm that's very strange - I think you have done everything that is needed in your PR. This must be a bug in `master`, let me check.

4ian: It's indeed a bug in master. I'm busy now but will push a fix. Your PR should be good to go then :)

4ian:
That was the necessary fix for the currentStorageProvider: https://github.com/4ian/GDevelop/commit/d9135636fe151425623d21968b18790d57c47ac2

The rest looks good, merging now! Thanks for adding this, will be very useful 👍

PR #1763, opened 2020-05-19
Fix create project not retaining path

### LINKED ISSUE #1747: Folder name override while creating a project
The folder field is not propagated to the other tab, it is reset to "My projectX" every time you change tab.
image
image

GD5b93 on Windows 10 64bit

arthuro555:
fixes #1747 .

### DIFF
--- a/newIDE/app/src/ProjectCreation/CreateProjectDialog.js
+++ b/newIDE/app/src/ProjectCreation/CreateProjectDialog.js
@@ -8,11 +8,6 @@ import { Tabs, Tab } from '../UI/Tabs';
 import Tutorials from './Tutorials';
 import { Column } from '../UI/Grid';
 import { type StorageProvider, type FileMetadata } from '../ProjectsStorage';
-import { findEmptyPath } from './LocalPathFinder';
-import optionalRequire from '../Utils/OptionalRequire.js';
-const path = optionalRequire('path');
-const electron = optionalRequire('electron');
-const app = electron ? electron.remote.app : null;
 
 type State = {|
   currentTab: 'starters' | 'examples' | 'tutorials',
@@ -57,15 +52,6 @@ export default class CreateProjectDialog extends React.Component<Props, State> {
     // Force an update to ensure dialog is properly positioned.
     this.forceUpdate();
   };
-
-  componentDidMount() {
-    if (path && app)
-      this.setState({
-        outputPath: findEmptyPath(
-          path.join(app.getPath('documents'), 'GDevelop projects')
-        ),
-      });
-  }
 
   render() {
     const { open, onClose, onOpen, onCreate } = this.props;
--- a/newIDE/app/src/ProjectCreation/CreateProjectDialog.js
+++ b/newIDE/app/src/ProjectCreation/CreateProjectDialog.js
@@ -8,9 +8,26 @@ import { Tabs, Tab } from '../UI/Tabs';
 import Tutorials from './Tutorials';
 import { Column } from '../UI/Grid';
 import { type StorageProvider, type FileMetadata } from '../ProjectsStorage';
+import { string } from 'prop-types';
+
+class StringStore {
+  state: string;
+  constructor(initialState: string = "") {
+    this.state = initialState;
+  }
+
+  setState(newState: string): void {
+    this.state = newState;
+  }
+
+  getState(): string {
+    return this.state;
+  }
+}
 
 type State = {|
   currentTab: 'starters' | 'examples' | 'tutorials',
+  outputPath: StringStore,
 |};
 
 export type CreateProjectDialogWithComponentsProps = {|
@@ -36,6 +53,7 @@ type Props = {|
 export default class CreateProjectDialog extends React.Component<Props, State> {
   state = {
     currentTab: 'starters',
+    outputPath: new StringStore(),
   };
 
   _onChangeTab = (newTab: 'starters' | 'examples' | 'tutorials') => {
@@ -85,6 +103,7 @@ export default class CreateProjectDialog extends React.Component<Props, State> {
               onOpen={onOpen}
               onCreate={onCreate}
               onShowExamples={this._showExamples}
+              pathProvider={this.state.outputPath}
             />
           )}
           {this.state.currentTab === 'examples' && (
@@ -92,6 +111,7 @@ export default class CreateProjectDialog extends React.Component<Props, State> {
               onOpen={onOpen}
               onCreate={onCreate}
               onExamplesLoaded={this._onExamplesLoaded}
+              pathProvider={this.state.outputPath}
             />
           )}
           {this.state.currentTab === 'tutorials' && <Tutorials />}
--- a/newIDE/app/src/ProjectCreation/CreateProjectDialog.js
+++ b/newIDE/app/src/ProjectCreation/CreateProjectDialog.js
@@ -11,7 +11,7 @@ import { type StorageProvider, type FileMetadata } from '../ProjectsStorage';
 
 class StringStore {
   state: string;
-  constructor(initialState: string = "") {
+  constructor(initialState: string = '') {
     this.state = initialState;
   }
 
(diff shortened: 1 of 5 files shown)


arthuro555:
I think it is ready for review @4ian .

newIDE/app/src/ProjectCreation/CreateProjectDialog.js
```diff
@@ -9,8 +9,24 @@ import Tutorials from './Tutorials';
 import { Column } from '../UI/Grid';
 import { type StorageProvider, type FileMetadata } from '../ProjectsStorage';
 
+class StringStore {
```
arthuro555, line 12: Not sure if that is the best way, if I should have declared that in a separate file and if there wasn't an equivalent already

4ian: You don't need this complex "string store". Just store a string in the state :)

newIDE/app/src/ProjectCreation/CreateProjectDialog.js
```diff
@@ -9,8 +9,24 @@ import Tutorials from './Tutorials';
...
+  }
+
+  getState(): string {
+    return this.state;
+  }
+}
+
 type State = {|
   currentTab: 'starters' | 'examples' | 'tutorials',
+  outputPath: StringStore,
```
4ian, line 29: outputPath: string

newIDE/app/src/ProjectCreation/CreateProjectDialog.js
```diff
@@ -85,13 +102,15 @@ export default class CreateProjectDialog extends React.Component<Props, State> {
               onOpen={onOpen}
               onCreate={onCreate}
               onShowExamples={this._showExamples}
+              pathProvider={this.state.outputPath}
```
4ian, line 105: instead of passing a complex "string store", pass something very simple:
the `outputPath`.

Then it seems that your components needs to change it.
So pass `onChangeOutputPath={(outputPath) => setState({ outputPath })}`

newIDE/app/src/ProjectCreation/CreateProjectDialog.js
4ian: Your custom string store won't notify React that something changed, because you won't call React `setState`, so the component will not re-render when there is a change,

newIDE/app/src/ProjectCreation/CreateProjectDialog.js
arthuro555: Oh, thanks! I don't really know react so I searched up on google how to share a state between parent and child and it said to use a store class. Your solution seems way better.

newIDE/app/src/ProjectCreation/LocalExamples.js
```diff
@@ -54,11 +55,14 @@ export const showGameFileCreationError = (
 
 export default class LocalExamples extends Component<Props, State> {
   state = {
-    outputPath: findEmptyPath(
-      path && app
-        ? path.join(app.getPath('documents'), 'GDevelop projects')
-        : ''
-    ),
+    outputPath:
```
4ian, line 58: If you store the outputPath string in the parent component, you must then remove it from the state of this component. Otherwise you're duplicating the state, which comes with several risks if not done properly.

Moving the state from a component to a parent component is quite common in React. It even has a name, it's called "Lifting Up State" because you're moving from a children component to a parent component. Read this nice article about it: https://reactjs.org/docs/lifting-state-up.html

Once you get it, it's a very interesting way of thinking: all in React is about where you put the state, and what prop you give to your children :)

newIDE/app/src/ProjectCreation/LocalExamples.js
```diff
@@ -54,11 +55,14 @@ export const showGameFileCreationError = (
...
+      this.props.pathProvider.getState() === ''
```
4ian, line 59: Then the question is "sure I moved outputPath to a prop, and stored it in the parent, but how do I initialize it??". Two possibilities:
1. Ideally, because it's the parent that is the owner of the state, it's the parent responsibility to initialize it. So you would move this code up to the parent component.
2. This being said in this case the parent is common between the electron-app and the browser app. So as an exception, we might want to have the parent keep `outputPath` empty first. Then in the children, we add a `componentDidMount` (React lifecycle function). Inside, we do `if (!this.props.outputPath) this.props.onChangeOutputPath(...)` with ... being the logic to set the output path.

newIDE/app/src/ProjectCreation/LocalExamples.js
```diff
@@ -80,10 +84,12 @@ export default class LocalExamples extends Component<Props, State> {
     });
   }
 
-  _handleChangePath = (outputPath: string) =>
+  _handleChangePath = (outputPath: string) => {
```
4ian, line 87: After moving outputPath to the parent, you don't need _handleChangePath. Instead, you call `this.props.onChangeOutputPath`. It's the parent responsibility to store it in its state.

Why does it work? Because *any time* you change the state of a component, *then* the component is re-rendered, and all children components are-rendered with the updated props (that may or may have not changed). That's the "react way of thinking", which is ver clean once you get it:
you store things in state. You can pass props to your children, including callbacs. These callbacks can updating the state. If it's done, everything that is in the component will be re-rendered ("flowing" from parents to children). Done! :) 
See https://reactjs.org/docs/thinking-in-react.html

newIDE/app/src/ProjectCreation/CreateProjectDialog.js
4ian: People are a bit obsessed with "stores". In practice, the simplest react way to sharing things is to pass a prop with the value, and a prop with a callback to set the value.
You're not sharing actually, you're giving "just enough stuff" for your children component to do its job. No more no less (which is all the beauty/cleanliness for the components approach).

newIDE/app/src/ProjectCreation/LocalExamples.js
arthuro555: Oh, that makes more sense now. Thank you very much for your time explaining it to me! (I should really have checked the react Docs before messing with it 😓 )

arthuro555:
Ok, I did what you said, it's ready for reviewing again. I am really sorry, I should have read the docs first.

newIDE/app/src/ProjectCreation/CreateProjectDialog.js
```diff
@@ -8,9 +8,15 @@ import { Tabs, Tab } from '../UI/Tabs';
 import Tutorials from './Tutorials';
 import { Column } from '../UI/Grid';
 import { type StorageProvider, type FileMetadata } from '../ProjectsStorage';
+import { findEmptyPath } from './LocalPathFinder';
```
4ian, line 11: So this works :) The downside is that the web-app is now depending in LocalPathFinder, which is a bit sad because any LocalXXX file should not be used by the web-app (web-app = BrowserApp.js).

arthuro555: Is that such a big deal? That file is very small, and I don't see any "react way" of solving this problem, because that would mean managing the state of the component from the children as those are electron only, but, quote from you,
> You're not sharing actually, you're giving "just enough stuff" for your children component to do its job.

So sharing the state for children for letting them handle/set it would be an anti-pattern?

4ian: You're not forced to duplicate the state. Instead you can use the second solution I explained here: https://github.com/4ian/GDevelop/pull/1763#discussion_r427806624

You can the state in the parent but set it empty, and use a componentDidMount in the children to change the path (see my answer).

The third solution is to make a function that we pass in LocalApp/BrowserApp that is a function that returns the default path - but at this point we're complicating stuff too much for something so boring. 
Just put two "componentDidMount", that are setting the output path if it's empty using the prop to change it and you're done.

arthuro555:
Ok, I did what you said except I only added componentDidMount on starters tab and not examples as it is the default mounted, and it should already have put the default path so no need for the other to put it too.

4ian:

PR #1770, opened 2020-05-21
Auto-open the most recent project

HarsimranVirk:
@4ian can you explain what `initialFileMetadataToOpen` is? I believe this should be something closely related to opening the initial fileMetada, no?

EDIT: I don't really understand the whole StorageProvider stuff correctly :sweat_smile:


newIDE/app/src/MainFrame/index.js
```diff
@@ -254,6 +254,19 @@ const MainFrame = (props: Props) => {
...
+            getRecentProjectFiles,
+            getProjectHasOpened,
+          } = preferences;
+
+          if (
+            getAutoOpenMostRecent() &&
+            getProjectHasOpened() &&
+            getRecentProjectFiles()[0]
+          )
+            openFromFileMetadataWithStorageProvider(getRecentProjectFiles()[0]);
```
HarsimranVirk, line 269: There are three checks here,
1. If this is enabled by the user's preferences or not.
2. If the project was closed before closing the app or not.
3. If there exists a recent file or not.

This should in most cases stop the unwanted opening of projects I guess.

newIDE/app/src/MainFrame/Preferences/PreferencesContext.js
```diff
@@ -168,6 +170,10 @@ export type Preferences = {|
   insertRecentProjectFile: (
     fileMetadata: FileMetadataAndStorageProviderName
   ) => void,
+  getAutoOpenMostRecent: () => boolean,
+  setAutoOpenMostRecent: (enabled: boolean) => void,
```
4ian, line 174: nit: let's call this get/setAutoOpenMostRecentProject

newIDE/app/src/MainFrame/Preferences/PreferencesContext.js
```diff
@@ -168,6 +170,10 @@ export type Preferences = {|
...
+  getProjectHasOpened: () => boolean,
+  setProjectHasOpened: (enabled: boolean) => void,
```
4ian, line 176: nit: let's call this `hadProjectOpenedDuringLastSession` and `setHasProjectOpened`

newIDE/app/src/MainFrame/Preferences/PreferencesContext.js
```diff
@@ -188,6 +194,8 @@ export const initialPreferences = {
     projectLastUsedPaths: {},
     defaultEditorMosaicNodes: {},
     recentProjectFiles: [],
+    autoOpenMostRecent: true,
```
4ian, line 197: nit: autoOpenMostRecentProject

newIDE/app/src/MainFrame/Preferences/PreferencesContext.js
```diff
@@ -188,6 +194,8 @@ export const initialPreferences = {
...
+    projectHasOpened: false,
```
4ian, line 198: nit: hasProjectOpened

newIDE/app/src/MainFrame/Preferences/PreferencesDialog.js
```diff
@@ -225,6 +226,14 @@ const PreferencesDialog = ({ onClose }: Props) => {
               label={<Trans>Auto-save project on Preview</Trans>}
             />
           </Line>
+          <Line>
+            <Toggle
+              onToggle={(e, check) => setAutoOpenMostRecent(check)}
+              toggled={values.autoOpenMostRecent}
+              labelPosition="right"
+              label={<Trans>Auto-open the most recent project.</Trans>}
```
4ian, line 234: nit: `Automatically re-open the project edited during last session` (do period at the end for consistency with other settings).

newIDE/app/src/MainFrame/index.js
4ian: This looks correct!

4ian:
Looks good overall! 👍

Can you try the "sad paths":
- Open a project, close GDevelop (so that it tries to re-open it). Delete or rename the folder with the project, re-open GDevelop. Normally GDevelop should error. Check if it's not crashing. Maybe we could return a promise somewhere and display a better error message.
- Can you try the same for the most recent projects? I realize we did not test these "sad paths", which are important to ensure the software is robust.

Thanks!!

HarsimranVirk:
@4ian fortunately `openFromFileMetadata` is robust enough to handle the sad paths ;) Gdevelop throws an error saying the file was not found, and that's it. 

Also, the app remains in loading state after this error, which is more or less a UI issue (meaning it's visually in loading state but you could do all sorts of things, like open a new project and work on it), I've fixed this :)

newIDE/app/src/MainFrame/index.js
```diff
@@ -1242,13 +1244,19 @@ const MainFrame = (props: Props) => {
...
+              openSceneOrProjectManager({
+                currentProject: state.currentProject,
+                editorTabs: state.editorTabs,
+              });
+          })
+          .catch(error => {
+            preferences.removeRecentProjectFile(
+              fileMetadataAndStorageProviderName
+            );
+          });
```
HarsimranVirk, line 1259: If the `fileMetadata` is indeed a sad path, it's best to remove it.
1. The app would try to auto-open a sad path only once.
2. There's no point in storing useless data ;)

newIDE/app/src/MainFrame/index.js
```diff
@@ -537,6 +537,8 @@ const MainFrame = (props: Props) => {
             ),
             error
           );
+          setIsLoadingProject(false);
+          return Promise.reject(error);
```
HarsimranVirk, line 541: `setIsLoadingProject(false)` brings the app to a usable state, otherwise, it remains in the loading state unless another project is opened.
Also, the error is propagated to remove the sad path file from localStorage.

4ian: >  usable state, otherwise, it remains in the loading state unless another project is opened.

Then this is still an inconsistent state right? We should remove the loading state once the loading is done, be it successfully or with a failure.

HarsimranVirk: > Then this is still an inconsistent state right?

I'm not sure if it's really an inconsistent state, but we should stop loading after a failure, right?

4ian: > EDIT: Not being able to open a sad path should be a failure, no?

Yes. 
Basically opening a project is either successful (project is opened, loading is stop) or not (for whatever reason: file not present, you don't have the permission, a meteor destroyed your hard drive...), in which case project is not opened, a message is shown and then the loading should stop - because you're back at the state where you can try to re-open your project or do something else.

HarsimranVirk: Alright @4ian, so the loading is not stopped until the project is opened successfully. Basically, when this error comes up, the app remains in a loading state after this failure.

> it remains in the loading state unless another project is opened.

When we successfully open another project, loading stops. So I believe, the loading state wasn't handled properly for failures (maybe during the refactor?).

4ian: > So I believe, the loading state wasn't handled properly for failures (maybe during the refactor?).

Possible. Can you have a look at this? If it's too complicated let me know and we'll fix this in master first?

newIDE/app/src/MainFrame/index.js
4ian: Sounds good to me!

newIDE/app/src/MainFrame/index.js
HarsimranVirk: @4ian, for successful openings, `openLayout` sets the loading state to false, (or `openSceneOrProjectManager` sets it false if the project has multiple scenes).

Now, there are two failures, one of them is the sad path, which is handled correctly now (and also a bit extended to remove these sad paths from localStorage). The other one occurs after `checkForAutosaveAfterFailure()`, I don't really know how to reproduce this.

4ian: I fixed the failure-not-reset-loading-state issue in master. Can you give a try to merging/rebasing your PR on master? :) Thanks!

HarsimranVirk:
> Can you give a try to merging/rebasing your PR on master?

@4ian I couldn't find any issues, it works well :)

4ian:
Thanks! Merged :) Should be a good feature for productivity of people using a lot GDevelop, and for us when working on GD :)

PR #1779, opened 2020-05-28
[WIP] Add libGD flow typing

arthuro555:
I know it would be better to have them autogenerated but in the meantime this is better than nothing. I didn't type everything yet. 

I tried to order typedefs in "categories".


4ian:
That's a good start :)

> I know it would be better to have them autogenerated

Before you go to far with this, could we investigate autogenerating these types from the `Bindings.idl` file? 
I'm afraid of the extra maintenance cost when you want to expose something from C++ to JS: types, tests, bindings, documentation... that's a lot. 

No type safety is not very good... but I'm afraid that this will make changes in the codebase slower and will make us less agile. I'm not certain about this, but we have to find a good balance, and maybe automating this is the answer.

Side note: remember that they are a lot of PRs open and where I spend time to review and that turns out to be abandoned it seems 😬 I would really really favor tiny PRs that are improving a bit the codebase or automating something small, but working and mergeable, that another big PR that is way too ambitious and will get abandoned in 3 or 4 days ;)

arthuro555:
> Before you go too far with this, could we investigate autogenerating these types from the Bindings.idl file?

There do are some (2) premade solutions, but they are old (approximately 5 years) and I couldn't make them work. It would require to write a whole thing from pretty much scratch and it's not guaranteed to work. I prefer to wait for someone else to make something working (unprobable) and use those typedes in the meantime.

> I'm afraid of the extra maintenance cost when you want to expose something from C++ to JS: types, tests, bindings, documentation... that's a lot.

Typing can still be made with the "old way" and somene else can take care of adding real typings later maybe? 

> I would really really favor tiny PRs that are improving a bit the codebase or automating something small, but working and mergeable, that another big PR that is way too ambitious and will get abandoned in 3 or 4 days ;)

You are right I have way too many open big PRs right now. I am sorry.

4ian:
> There do are some (2) premade solutions

What are those? If we could translate the WebIDL to Typescript, should be easy to then do a few search/replace to make it compatible with Flow.

I've seen https://github.com/charto/nbind too but it would require a rearchitecture of GDevelop.js build. Might be just simpler to see how they do and do the same for GDevelop.js :)

arthuro555:
The two tools I found are https://github.com/markandrus/webidl-tools and https://github.com/motiz88/webidl-to-flow

4ian:
I gave a quick try to  https://github.com/markandrus/webidl-tools:
- `npm install -g webidl-tools`
- copy Bindings.idl in a "idl" folder (because seems like the tool is using this folder by default).
- run `webidl-tools flow`. It errors because the webidl parser (webidl2) can't parse declarations like `[Prefix="gdjs::"]` (the quotes are not understood). Ideally we could fix webidl2. But for now:
- Just comment the `[Prefix="gdjs::"]` :D Some declarations can have their quotes removed, but other (like `[Prefix="gdjs::"]` have :: and if we don't put quotes, then it's the Emscripten WebIDL parser that is not working!)
- Run again `webidl-tools flow`
- It errors somewhere because of an empty declaration. Add `.filter(declaration => !!declaration.name)` after `.map(generateDeclaration)` in `node_modules/webidl-tools/lib/flow.js`
- It works :D
- See the result in decls/ folder

Still to do:
- Make a real fork of webidl-tools with the fix of `.filter(declaration => !!declaration.name)` or understand why it's erroring.
- In the fork, output the resulting functions of all the classes in lowerCamelCase instead of UpperCamelCase. (I did this so that libGD.js confirms to JavaScript idiomatic usage of having functions in lowerCamelCase).
- Prefix everything by `gd`.
- Enjoy :D (and maybe fix a few things in the codebase... or add `// $FlowFixMe` because nobody got time for that ;)).
- Profit from automatically generated typesafety 😎

4ian:
See: https://github.com/4ian/GDevelop/commit/5e1cb9dc52743b9ba3ad04bcf5ec55bd774f38b6 :) You still need to apply the fix I mentioned to webidl-tools in flow.js

4ian:
Let me know if you want to do this, otherwise I'll give this a try.

arthuro555:
I think I will try. Thanks for investigating so far for me 😓.

4ian:
For:

> Do a search/replace to transform STATIC_ in static (automate this in a nice script).

You can do: `webidl-tools flow --rename "s/STATIC_/static /"` but the parsing of the "rename" function is broken :/ You need to change in this line: https://github.com/markandrus/webidl-tools/blob/master/lib/webidl.js#L18 by  `var match = str.match(/^s?\/?([^/]*)\/([^/]*)\/?$/);` (I think the original regex is not working).

For:

> In the fork, output the resulting functions of all the classes in lowerCamelCase instead of UpperCamelCase

Look at what a "definition" is: https://github.com/markandrus/webidl-tools/blob/95b3b911c22ff4c2634a148e301aa3ff09e84a25/lib/webidl.js#L118-L120

> In the fork, prefix everything by gd (or create an option for this).

Here I'm not sure what is the easiest between:
- do it in the fork (`extractDefinition`/`replace` in webidl.js)
- or "just" do some search/replace on the output definition (parse all the `interface XXX {` in the webidl file, then for each interface to a search/replace).
- or don't rename stuff, instead adapt our codebase (but at least with a name like `gdProject`, it's very easy to see what are things that are coming from libGD.js (and so need care when using them!))

arthuro555:
I'm really sorry and feel really stupid but somehow I can't manage to even run the tool. 
image
By using `node <executablepath>/webidl-tools-flow` I managed to start the tool only to get an exception from a module...
image
I think I am hopeless, I am sorry for wasting your time and I think you should do it instead of me as I can't even manage to run it.

> It errors because the webidl parser (webidl2) can't parse declarations like [Prefix="gdjs::"] (the quotes are not understood). Ideally we could fix webidl2.

Maybe updating webidl2 could fix this?

4ian:
Don't use npx, follow the README of webidl-tools, i.e: install it globally using npm install.

arthuro555:
> Don't use npx, follow the README of webidl-tools, i.e: install it globally using npm install.

I tried that too already.

4ian:
So if you do: `npm install -g webidl-tools` then `webidl-tools` it's not working? What is your Node.js version (`node -v`?)

4ian:
(also if it's still not working don't lose too much time on this, I could try to take care of this later).

4ian:
I added automatic generation of types in #1800. Should bring more safety in the code and better developer experience :) 
Closing this PR as it's now redundant.

PR #1783, opened 2020-06-01
Dynamic Lights (GSoC)

HarsimranVirk:
This is meant to be the working prototype of dynamic lights and shadows.

A broad overview of goals:

- [x] Add properties and register instance renderer.

- [x] Add LightObstacleManager and LightObstacleRuntimeBehaviour

- [x] Update vertices in each frame.

- [x] Implement simple raycasting.

- [x] Use raycasting to determine vertices.

- [ ] Add functionality of rendering to texture for `layer-pixi-renderer`.

- [ ] Use _`RenderTextures`_ as uniforms for the _`shader`_ and render normal maps.


Extensions/Lighting/lightobstacleruntimebehavior.js
```diff
@@ -0,0 +1,115 @@
+gdjs.LightObstaclesManager = function (runtimeScene, sharedData) {
+  this._obstacleRBush = new rbush(9, [
+    '.owner.getAABB().min[0]',
+    '.owner.getAABB().min[1]',
+    '.owner.getAABB().max[0]',
+    '.owner.getAABB().max[1]',
+  ]);
+};
+
+gdjs.LightObstaclesManager.getManager = function (runtimeScene) {
```
HarsimranVirk, line 10: @4ian I have a question/doubt. If getManager is on the prototype (ie, `gdjs.LightObstaclesManager.prototype.getManager`), I get errors saying getManager is not a function. Why is that so? Does it have something related to being a singleton (if I understand the term correctly, singleton means only one instance of the class right?), or is it something else?

4ian: @HarsimranVirk When you put a method on the prototype in JavaScript, it's akin to a "class method", meaning that you have to instantiate an instance of the class:
```
MyClass = function() { this.something = 3; } // This is the "constructor"
MyClass.prototype.myMethod = function() { return this.something; } // This is a "class method"

var instance = new MyClass();
console.log(instance.myMethod()); // Will display 3.
MyClass.myMethod() // error because myMethod does not exist.
```

if you remove the `.prototype` when declaring a function on a class, you're just setting a function as a property... in other words it's a "static" function:

```
MyClass = function() { ... } // This is the "constructor"
MyClass.myStaticMethod = function() { return "Hello world";  } // This is a "static" method.

console.log(MyClass.myStaticMethod()); // Will display "Hello world".
```

Here, `getManager` is a "static" function. You can think of it as a helper. It was put here because it's better to put it as a property of `gdjs.LightObstaclesManager` rather than as a global variable. This is indeed more or less related to a "singleton" pattern, though a bit different. A singleton is a *single* class instance for the whole code, while here we want *one manager per scene*.
How do we do that? By having a single function called `gdjs.LightObstaclesManager.getManager`, that you can call anywhere, and that returns the gdjs.LightObstaclesManager for the scene. The way it does it is by creating (if necessary) a `new gdjs.LightObstaclesManager()`, storing it inside the runtimeScene (but we could imagine using a hashtable or something like that to avoid polluting the runtimeScene... but this is the fastest way to do it), and then return it.

Note that `getManager` is a static method so you must always call it by writing the entire name `var lightObstaclesManager = gdjs.LightObstaclesManager.getManager(runtimeScene);`.
You can then call usual (non static) methods:
`lightObstaclesManager.getObstaclesAroundObject(object, radius, ...);` etc

4ian: Note that in recent versions of JavaScript, the `class` keyword was added, so you can do instead:

```js
class MyClass {
  constructor() {
    this.something = 3;
  }
  
  myMethod() { return this.something; } 
  static myStaticMethod() { return "hello world"; }
}


var instance = new MyClass();
console.log(instance.myMethod()); // Will display 3.
console.log(MyClass.myStaticMethod()); // Will display "Hello world".
```

but this is just "syntactic sugar" over the `prototype` that I explained before. We say that JavaScript is a "prototype" based language. In GDJS, we don't use the class keyword because it's unsupported in old browsers, notably Android 4.4 (we write "ES5" Javascript, not ES6).

HarsimranVirk: > we write "ES5" Javascript, not ES6

I totally forgot that 😅 I've been using a lot of `const`, `let` and arrow functions, might need some refactoring then (especially with using `this` in `function()`).

4ian: That's ok as a first draft, just don't forget to make a card about this in the Trello so we remember to make a pass to ensure everything is using `var`/`function` :)

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -94,4 +115,77 @@ gdjs.LightRuntimeObjectPixiRenderer.prototype.updateVertexBuffer = function () {
         this._object.y,
       ])
     );
+//  }
+};
+
+gdjs.LightRuntimeObjectPixiRenderer.prototype.raycastTest = function () {
```
HarsimranVirk, line 121: @4ian this doesn't really work. I mostly get empty arrays and I don't understand why is that so. The expected return value should be at least the vertices of the mesh (`this._object.getHitBoxes().vertices`). Can you take a look at this function? There's definitely something wrong here :/

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -75,12 +77,31 @@ gdjs.LightRuntimeObjectPixiRenderer.prototype.ensureUpToDate = function () {
...
+  //   .getBuffer('aVertexPosition')
+  //   .update(new Float32Array(vertices));
+
+  //   var indices = [0,1,2];
+
+  //   for(var i = 3; i < vertices.length; i+=3) {
+  //     indices[i] = 0;
+  //     indices[i + 1] = indices[i - 1];
+  //     indices[i + 2] = indices[i + 1] + 1;
+  //   }
```
HarsimranVirk, line 99: I have commented out all of this because I get empty arrays from the raycasting function. `this.raycastTest()` returns an array of vertex positions which are empty.

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
HarsimranVirk: I guess I'm doing something wrong with _`closeX`, `closeY, farX and farY`_ 🤔 Can you explain the return value (`result`) of _`gdjs.Polygon.raycastTest()`_ ?

Extensions/Lighting/lightobstacleruntimebehavior.js
```diff
@@ -18,33 +18,31 @@ gdjs.LightObstaclesManager.getManager = function (runtimeScene) {
...
+    gdjs.LightObstaclesManager.prototype.getAllObstaclesAround
   );
-  searchArea.minX = x - ow / 2 - radius;
-  searchArea.minY = y - oh / 2 - radius;
-  searchArea.maxX = x + ow / 2 + radius;
-  searchArea.maxY = y + oh / 2 + radius;
+  searchArea.minX = x;
+  searchArea.minY = y;
+  searchArea.maxX = x + 2 * radius;
+  searchArea.maxY = y + 2 * radius;
```
4ian, line 45: Should this be:

```suggestion
  searchArea.minX = x - radius;
  searchArea.minY = y - radius;
  searchArea.maxX = x + radius;
  searchArea.maxY = y + radius;
```

? Because the light is in the center, and you must identify obstacles that are around it, in a square that is containing the light.

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -94,4 +115,77 @@ gdjs.LightRuntimeObjectPixiRenderer.prototype.updateVertexBuffer = function () {
...
+gdjs.LightRuntimeObjectPixiRenderer.prototype.raycastTest = function () {
+  var result = [];
+  var center = [this._object.x + this._radius, this._object.y + this._radius];
+  var halfOfDiag = Math.sqrt(2) * this._radius;
+
+  this._manager.getAllObstaclesAround(this._object, this._radius, result);
+  var hitBoxes = result.map(function (item) {
+    return item.owner.hitBoxes;
+  });
+  hitBoxes.push(this._object.getHitBoxes());
```
4ian, line 130: Why adding the hit boxes of the object? The object *is* the light, so it should not be taken into account of the *light* polygon because... well because it's the light itself :D 
The rays are emitted from the point (`this._object.x`, `this._object.x`) are are stopped by the obstacles around (`this._manager.getAllObstaclesAround`). You only care about the hitboxes of these objects.

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -94,4 +115,77 @@ gdjs.LightRuntimeObjectPixiRenderer.prototype.updateVertexBuffer = function () {
...
+};
+
+gdjs.LightRuntimeObjectPixiRenderer.prototype.raycastTest = function () {
+  var result = [];
+  var center = [this._object.x + this._radius, this._object.y + this._radius];
+  var halfOfDiag = Math.sqrt(2) * this._radius;
+
+  this._manager.getAllObstaclesAround(this._object, this._radius, result);
+  var hitBoxes = result.map(function (item) {
+    return item.owner.hitBoxes;
```
4ian, line 128: Never use directly hitBoxes, it's a protected member of gdjs.RuntimeObject. Instead, use `getHitBoxes`: `item.owner.getHitBoxes()`

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -94,4 +115,77 @@ gdjs.LightRuntimeObjectPixiRenderer.prototype.updateVertexBuffer = function () {
...
+  hitBoxes.push(this._object.getHitBoxes());
+
+  var polygons = hitBoxes.flat(1);
+  var vertices = polygons.map(function (poly) {
+    return poly.vertices;
+  }).flat(1);
+
+  function _calculatePOI(angle) {
+    var minDist = halfOfDiag;
+    var minPOI = [];
```
4ian, line 139: This is a point, so prefer to declare it as such for readibility and to help the JS engine by never modifying its length:

```
var minPOI = [0, 0];
```
then later:
```
minPOI[0] = poi.closeX
minPOI[1] = poi.closeY
```

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -94,4 +115,77 @@ gdjs.LightRuntimeObjectPixiRenderer.prototype.updateVertexBuffer = function () {
...
+        center[0],
+        center[1],
+        center[0] + halfOfDiag * Math.cos(angle),
+        center[1] + halfOfDiag * Math.sin(angle)
+      )
+
+      if (poi.closeSqDist <= minDist) {
+        minDist = poi.closeSqDist;
+        minPOI.length = 0;
+        minPOI.push.apply(minPOI, [poi.closeX, poi.closeY]);
```
4ian, line 152: As explained in my other comment:

```
minPOI[0] = poi.closeX
minPOI[1] = poi.closeY
```

this allocates no memory at all, just assigning two integers in memory, which is way way faster than erasing an array (`minPOI.length = 0;`), creating a new array (`[poi.closeX, poi.closeY]`), calling a function (`push`) that will probably then manipulate your array.

Extensions/Lighting/lightruntimeobject.js
```diff
@@ -1,11 +1,14 @@
 gdjs.LightRuntimeObject = function (runtimeScene, lightObjectData) {
   gdjs.RuntimeObject.call(this, runtimeScene, lightObjectData);
 
-  this._radius = parseFloat(lightObjectData.content.radius);
+  this._radius =
+    typeof lightObjectData.content.radius === 'string'
```
4ian, line 5: unless radius was once stored as a string, you should be able to always do `this._radius = lightObjectData.content.radius`. No need to handle string, they should never appear here. If they do, it's a configuration problem in the extension.

Extensions/Lighting/lightobstacleruntimebehavior.js
4ian: Oh ok I understand that you made the choice of having the light x,y position to be the top-left.
I wonder if we should instead use the center of the light as the x, y position.

It sounds naturally for a light, because it's where the light is emitted :)

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -94,4 +115,77 @@ gdjs.LightRuntimeObjectPixiRenderer.prototype.updateVertexBuffer = function () {
         this._object.y,
       ])
     );
+//  }
+};
+
+gdjs.LightRuntimeObjectPixiRenderer.prototype.raycastTest = function () {
+  var result = [];
+  var center = [this._object.x + this._radius, this._object.y + this._radius];
```
4ian, line 123: See other comments, I recommend we use this._object.x and this._object.y as the center coordinates.

4ian: Also, unless you need to pass this as a point to a function, avoid allocating an array in memory and just directly use `this._object.x` and `this._object.y`.
If you want to name them as center, do `var centerX = this._object.x;` and `var centerY = this._object.y;`.

Not a huge deal, but avoiding allocating arrays and objects can help reducing the pressure on the garbage collector.

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -94,4 +115,77 @@ gdjs.LightRuntimeObjectPixiRenderer.prototype.updateVertexBuffer = function () {
...
+
+  var _raycastResult = [];
+
+  for (var vertex of vertices) {
+    var xdiff = vertex[0] - center[0];
+    var ydiff = vertex[1] - center[1];
+    var angle = Math.atan(ydiff/xdiff);
+
+    if (xdiff < 0) angle += Math.PI;
+    if (xdiff >= 0 && ydiff < 0) angle += 2 * Math.PI;
```
4ian, line 167: (you won't need these with atan2)

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -94,4 +115,77 @@ gdjs.LightRuntimeObjectPixiRenderer.prototype.updateVertexBuffer = function () {
...
+    });
+    _raycastResult.push({
+      vertex: _calculatePOI(angle - 0.01),
+      angle: angle - 0.01
+    });
+  }
+
+  return _raycastResult.sort(function (a, b) {
+    if (a.angle < b.angle) return -1;
+    if (a.angle > b.angle) return 1;
```
4ian, line 187: Your function is not returning anything if the angle are equal. Return 0 in this case.

Extensions/Lighting/lightruntimeobject.js
```diff
@@ -32,6 +35,14 @@ gdjs.LightRuntimeObject.prototype.getRadius = function () {
   return this._radius;
 };
 
+gdjs.LightRuntimeObject.prototype.getHeight = function () {
+  return 2 * this.getRadius();
+};
+
+gdjs.LightRuntimeObject.prototype.getWidth = function () {
+  return 2 * this.getRadius();
+};
```
4ian, line 44: getHeight, getWidth are good :) 

I think we should consider, as mentioned in other comments, the center of the light as its x,y position. This would be especially useful with things like directional lights later, or just to be able to increase/reduce the radius of a light without having it moving ;) 
So in addition to these 2 methods, for the engine to understand that the top-left of the object is not at the position x,y, you can redefine the methods getDrawableX and getDrawableY:

```
gdjs.LightRuntimeObject.prototype.getDrawableX = function() {
    return this.x - this._radius;
};

gdjs.LightRuntimeObject.prototype.getDrawableY = function() {
    return this.y - this._radius;
};
```

This will be used by the engine for things like the draggable behavior, the culling (objects that are outside of the screen are not rendered), etc...

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -94,4 +115,77 @@ gdjs.LightRuntimeObjectPixiRenderer.prototype.updateVertexBuffer = function () {
...
+//  }
+};
+
+gdjs.LightRuntimeObjectPixiRenderer.prototype.raycastTest = function () {
+  var result = [];
+  var center = [this._object.x + this._radius, this._object.y + this._radius];
+  var halfOfDiag = Math.sqrt(2) * this._radius;
+
+  this._manager.getAllObstaclesAround(this._object, this._radius, result);
+  var hitBoxes = result.map(function (item) {
```
4ian, line 127: Let's give more descriptive names to what we manipulate:
`hitBoxes` => `obstaclesHitBoxes`
`polygons ` => `obstaclesPolygons`
`vertices ` => `obstaclesVertices`

so that it's obvious when we read the code later what are these vertices coming from.

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -94,4 +115,77 @@ gdjs.LightRuntimeObjectPixiRenderer.prototype.updateVertexBuffer = function () {
...
+    return item.owner.hitBoxes;
+  });
+  hitBoxes.push(this._object.getHitBoxes());
+
+  var polygons = hitBoxes.flat(1);
+  var vertices = polygons.map(function (poly) {
+    return poly.vertices;
+  }).flat(1);
```
4ian, line 135: We'll replace `.flat(1)` in the future by (potentially) nested loops, because flat is ES6. For now, concentrate on making this work, add this to a to do list in the trello or in the code :)

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
4ian: When you launch a raycast against a polygon, the result is:
- was there even a collision between the ray and the polygon? (`collision` boolean)
- coordinate of the closest point between the ray and the polygon (closeX, closeY)
- the squared distance of this point (`closeSqDist`)
- coordinate of the farthest point between the ray and the polygon (closeX, closeY). You can picture it as the point behind the polygon, if you think of the ray as continue through the polygon:
- the squared distance of this point (`farSqDist`)

image

In our algorithm, we don't need the farX, farY point :)

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -94,4 +115,77 @@ gdjs.LightRuntimeObjectPixiRenderer.prototype.updateVertexBuffer = function () {
...
+    for (var poly of polygons) {
+      var poi = gdjs.Polygon.raycastTest(
+        poly,
+        center[0],
+        center[1],
+        center[0] + halfOfDiag * Math.cos(angle),
+        center[1] + halfOfDiag * Math.sin(angle)
+      )
+
+      if (poi.closeSqDist <= minDist) {
```
4ian, line 149: Only do this if poi.collision is true. If it's false, the ray did not touch anything.

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -94,4 +115,77 @@ gdjs.LightRuntimeObjectPixiRenderer.prototype.updateVertexBuffer = function () {
...
+  });
+  hitBoxes.push(this._object.getHitBoxes());
+
+  var polygons = hitBoxes.flat(1);
+  var vertices = polygons.map(function (poly) {
+    return poly.vertices;
+  }).flat(1);
+
+  function _calculatePOI(angle) {
+    var minDist = halfOfDiag;
```
4ian, line 138: Almost! This should be `minDist = halfOfDialog*halfOfDialog` because we're comparing **squared distances**. Also, call this `minSqDist`.

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
4ian: Actually, the default point should be `var minPOI = [center[0] + halfOfDiag * Math.cos(angle), center[1] + halfOfDiag * Math.sin(angle)];` (corresponding to a squared distance of `minSqDist = halfOfDialog*halfOfDialog`)

Because it's reused a lot, store it once:
```
const targetX = center[0] + halfOfDiag * Math.cos(angle);
const targetY = center[1] + halfOfDiag * Math.cos(angle);
```

and reuse it in `raycastTest`.

4ian:
I added multiple comments, notably look at these points:
- Don't use hitBoxes, use getHitBoxes. Otherwise, hitboxes are not updated and that's maybe the reason you would not get any result.
- default minDist is too short, maybe also the reason you would not get any result.
- Use atan2 instead of atan
- Let's consider the center of the light as it's X;Y position - better for users (if we change radius, light is not moving), better for us to reason about the light. Needs addition of getDrawableX/getDrawableY methods, see my comments.
- Avoid array allocation as much as you can: store things as numbers if you can.

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
HarsimranVirk: > Why adding the hit boxes of the object?

We'll need to add the hitbox of the object because we need to know the intersection points on the boundary of mesh as well, to generate a complete geometry.
Screenshot (2)

In this article https://ncase.me/sight-and-light/ (and the other articles as well) you can see that points of intersection with the boundary of light is required to create triangles.

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
HarsimranVirk: Yup, this is why I'm getting `[0, 0]` for all the vertices of hitbox 😓

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
4ian: As discussed that's indeed required in our algorithm, good spot!

HarsimranVirk:
Current checklist:

- [x]  Use object.x and object.y as the center of light in runtime (and in instance renderer if possible).

- [ ] Use atan2 instead of atan in raycasting algorithm (and check if it works).

- [ ] Fix minDist in the algo, it's one of the many errors.

- [ ] Call `getHitBoxes()` to initialize them, instead of getting `[0, 0]`.

- [ ] If the raycasting works, optimize it by minimizing array allocations.

HarsimranVirk:
@4ian I'm pretty sure that the algorithm is working now, but the rendering has some issues now.

Screenshot (5)

Here, you can see there's something wrong, there is an extra triangle. Sometimes the situation becomes even worse and the geometry becomes really bad.
Screenshot (6)

(Though, don't worry about the bending of light in the lower part, it's correct because the hitbox is like this: 
Screenshot (7)
)

4ian:
Interesting, do you have any hypothesis about why we have this artifact? Is it the polygon that has multiple points at the same position maybe?

> Sometimes the situation becomes even worse and the geometry becomes really bad.

Ideally try to make a scene with a few of these bad examples to start debugging from there.
We want to be sure the result from the raycasting is bullet proof :)

HarsimranVirk:
> do you have any hypothesis about why we have this artifact?

I guess it's more or less how I'm defining the index buffer, I've tried removing the duplicates, sortings seems correct too. There's always some problem with the last triangle (which is made up of the last three indices, and it's generally the top-left vertex of the quad 🤔).

4ian:
> There's always some problem with the last triangle (which is made up of the last three indices, and it's generally the top-left vertex of the quad 🤔).

Isn't there a duplicate between the first and last triangle? (I've not verified, just asking). Also maybe you could simplify the mesh geometry by using a TRIANGLE_FAN (https://pixijs.download/dev/docs/PIXI.html#.DRAW_MODES) draw mode, because all triangles are delimited by the vertices we got from the ray cast, that are already sorted clockwise (or counter clockwise, does not matter), they all start from the center point and use the last vertex of the last triangle as a starting point (https://stackoverflow.com/questions/8043923/gl-triangle-fan-explanation).
Maybe that could help finding the issue in the vertices/indexes :)

HarsimranVirk:
> Isn't there a duplicate between the first and last triangle?

Yes, the 1st vertex is used with the last vertex to form the last triangle, maybe this is wrong? I'm a bit stuck at this now, to be honest.

> Also maybe you could simplify the mesh geometry by using a TRIANGLE_FAN

Somehow, TRIANGLE_FAN gives even worse result than indices (sometimes there's nothing rendered on screen at all) 😓 

Although we are a bit stuck here, I've still got some good news :) I've made a debug version of light that points out all the vertices calculated by the raycasting algorithm. It verifies that our raycasting is absolutely correct :)

Screenshot (8)
Screenshot (9)
Screenshot (10)
Screenshot (11)

At least now we can be sure that the error lies at how we create triangles to be rendered, not the raycasting ;)

Also, a bit of extra information, turns out our mesh implementation is quite performant since it has already been used in _Celeste_, https://medium.com/@NoelFB/remaking-celestes-lighting-3478d6f10bf.

> Start at the nearest surface to the center of the light, and move along each joined surface or edge raycast clockwise, creating triangles

The article vaguely mentions starting at the _nearest surface_ while rendering triangles, do you think this could be the potential error? (Since we always begin with _`-Math.PI`_ even if the nearest surface isn't in that direction.)

4ian:
> EDIT: I should maybe make the debug light render triangles instead of rays (in a similar way we are creating values for index buffer). This could maybe help us pin point the exact issue?

Yes please! So that we can identify if these triangles have indeed some issues like those rendered in the screenshots.

4ian:
> Somehow, TRIANGLE_FAN gives even worse result than indices (sometimes there's nothing rendered on screen at all) 😓

Strange because it should be exactly the rendering method that we need for this.

> (Since we always begin with -Math.PI even if the nearest surface isn't in that direction.)

Can you pinpoint in the code where we doing this? Naively I would say that where you are starting should not have an impact though,

HarsimranVirk:
@4ian, the lights are rendered correctly now 🥳, there's definitely some issue with index buffer. I've added it as todo. Right now we have a working prototype of lights. 🤩

> I should maybe make the debug light render triangles instead of rays

It turned out to be super useful :D I've added them as a part of extension now, they might turn out very helpful in the future too (at least while we are in development) :)

Also, unrelated stuff, is master on v5 now? I got a notification about you adding a commit, is it merged?

HarsimranVirk:
Screenshot:

Screenshot (12)

4ian:
Nice!! 😍 Great job!

> It turned out to be super useful :D I've added them as a part of extension now, they might turn out very helpful in the future too (at least while we are in development) :)

Sounds like a good idea to keep that indeed! It's very visual and will be helpful to understand how it works. Can also help in the future if we're refactoring/reworking the way raytracing is done :) 

> Also, unrelated stuff, is master on v5 now? I got a notification about you adding a commit, is it merged?

It's "merged" in `pixi-v5`, you can pull this branch, cherry pick your commits on top of it, then open a new PR using `pixi-v5` as the base instead of `master`.

> there's definitely some issue with index buffer. I've added it as todo.

In which case is this problem appearing?

HarsimranVirk:
> In which case is this problem appearing?

I've fixed the issue of index buffers as well, my math was a bit off 😅 Anyway, I'll be pushing a commit now and start working at using ES5 (it's already ES5 except `flat()` I believe?), meanwhile, take your time reviewing the code so we can optimize a bit :)

PR #1806, opened 2020-06-10
Debugger extension

arthuro555:
This extension is for adding interractions with the debugger.
For now it only has breakpoints.
In the future it may be extended (sort of console in debugger with console logging?)

The PR also includes a Getter on runtimeGame for getting the debugger client.


Bouh:
At which point the game().pause() work ? 
Because i've seen few days ago an user said the timer from the tweens extension can't be affect with the timescale action.

My doubt is if this pause on game do a thing similar to timescale 0 this new break action will not affect the tween extension.

[Source on forum.](https://forum.gdevelop-app.com/t/tween-animation-play-goes-on-when-i-pause-the-game/25518).

Maybe it's not related i don't know, i wanna just warn about it.

arthuro555:
Yes that would belong in another issue. This only add a way to call it through actions instead of only having the debugger to call it.

4ian:
Let's call this "Pause the game execution" because this is not a breakpoint. A breakpoint is a way to stop event execution in the middle of events, and then possibly only execute one event (or even one condition/action) at a time (this is called "step-by-step" execution).
(This is a very interesting feature but that will need larger changes in the way events are generated and how the debugger client communicates with the editor.)

This features though is still very interesting because it will help a lot to debug a particular issue (I want to understand why my event is executed: add something to pause the execution and then you can use the debugger to inspect variables, though *other events* will be executed before the game is paused.).

Extensions/Debugger/JsExtension.js
```diff
@@ -0,0 +1,46 @@
...
+          'Adds Interractions possibilities with the debugger'
+        ),
+        'Arthur Pacaud (arthuro555)',
+        'MIT'
+      );
+
+      extension
+      .addAction(
+        'Breakpoint',
+        _('Break execution (break point)'),
```
4ian, line 29: "Pause game execution".

Extensions/Debugger/JsExtension.js
```diff
@@ -0,0 +1,46 @@
...
+        _('This pauses the game for inspecting the game state through the debugger at a specific time.'),
```
4ian, line 30: This pauses the game, useful for inspecting the game state through the debugger. Note that events will be still executed until the end before the game is paused.

Extensions/Debugger/JsExtension.js
```diff
@@ -0,0 +1,46 @@
...
+        _('Break game execution'),
```
4ian, line 31: Pause game execution

Extensions/Debugger/JsExtension.js
```diff
@@ -0,0 +1,46 @@
...
+ * More information on https://github.com/4ian/GDevelop/blob/master/newIDE/README-extensions.md
+ */
+module.exports = {
+    createExtension: function(_, gd) {
+      const extension = new gd.PlatformExtension();
+      extension.setExtensionInformation(
+        'Debugger',
+        _('Debugger'),
+        _(
+          'Adds Interractions possibilities with the debugger'
```
4ian, line 20: 'Allow to interact with the editor debugger from the game.'

Extensions/Debugger/JsExtension.js
```diff
@@ -0,0 +1,46 @@
...
+ * ⚠️ If you make a change and the extension is not loaded, open the developer console
+ * and search for any errors.
+ *
+ * More information on https://github.com/4ian/GDevelop/blob/master/newIDE/README-extensions.md
+ */
+module.exports = {
+    createExtension: function(_, gd) {
+      const extension = new gd.PlatformExtension();
+      extension.setExtensionInformation(
+        'Debugger',
```
4ian, line 17: I wonder if we should call this extension `DebuggerTools`? 
`Debugger` seems a bit too broad, in the sense that this is not this extension that is bringing its own debugger. Instead, it's using the debugger interface that is exposed by the game engine. In a codebase, I would have been calling methods exposed by this extension "helpers", "tools", etc.. so I think `DebuggerTools` could be a better name (and easier to grep in the codebase).

Extensions/Debugger/debuggertools.js
```diff
@@ -0,0 +1,19 @@
...
+ * The namespace containing tools to interract with the debugger.
+ * @namespace
+ */
+gdjs.evtTools.debugger = {};
+
+/**
+ * Breaks execution (breakpoint).
+ * @param {gdjs.RuntimeScene} runtimeScene - The current scene.
+ */
+gdjs.evtTools.debugger.break = function(runtimeScene) {
```
4ian, line 17: Let's call it pause :)

Extensions/Debugger/debuggertools.js
```diff
@@ -0,0 +1,19 @@
...
+
+
+/**
+ * The namespace containing tools to interract with the debugger.
+ * @namespace
+ */
+gdjs.evtTools.debugger = {};
+
+/**
+ * Breaks execution (breakpoint).
```
4ian, line 14: Stop the game execution.

Extensions/Debugger/debuggertools.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * Tools for interracting with the debugger.
+ * @fileoverview
+ */
+
+
+/**
+ * The namespace containing tools to interract with the debugger.
```
4ian, line 8: interact

Extensions/Debugger/debuggertools.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * Tools for interracting with the debugger.
+ * @fileoverview
```
4ian, line 3: Let's use @file for consistency with Doxygen. Also I think the description must follow the tag ;)

GDJS/Runtime/runtimegame.js
```diff
@@ -60,6 +60,10 @@ gdjs.RuntimeGame = function(data, options) {
   this._options = options;
 
   //Optional client to connect to a debugger:
+  /** 
+   * @type {?gdjs.IDebuggerClient}
```
4ian, line 64: Is this type necessary? If not maybe we can skip it to avoid polluting too much the code.
private is also automatically added for JSDoc because of the underscore (not for typescript, but this creates problem in newer version of typescript that does not understand that class methods using `prototype` are allowed to access private variables).

GDJS/Runtime/runtimegame.js
```diff
@@ -212,6 +216,14 @@ gdjs.RuntimeGame.prototype.getInitialObjectsData = function() {
   return this._data.objects || [];
 };
 
+/**
+ * Get the Debugger Client.
+ * @return {?gdjs.IDebuggerClient} The Debugger (or null if not present).
+ */
+gdjs.RuntimeGame.prototype.getDebugger = function() {
```
4ian, line 223: I would avoid adding this if it's not used in the PR?

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.js
```diff
@@ -27,7 +27,7 @@ gdjs.IDebuggerClient.prototype.sendRuntimeGameDump = function() {}
 /**
  * Start profiling.
  */
-gdjs.IDebuggerClient.prototype.sendProfilerStrated = function() {}
+gdjs.IDebuggerClient.prototype.sendProfilerStarted = function() {}
```
4ian, line 30: Nice catch  :)

GDJS/Runtime/runtimegame.js
arthuro555: ok

4ian: It's good practice in general to avoid adding something if it's not used. If you have a use case of needing this outside the PR, I'm not totally against, but it's a new public interface (i.e: will need to be available forever to avoid breaking compatibility in the future), so I'm cautious about adding these.

GDJS/Runtime/runtimegame.js
arthuro555: IMO the more typing the better.

Extensions/DebuggerTools/JsExtension.js
```diff
@@ -0,0 +1,49 @@
...
+      .addAction(
+        'Pause',
+        _('Pause game execution'),
+        _(
+          'This pauses the game, useful for inspecting the game state through the debugger. ' + 
+          'Note that events will be still executed until the end before the game is paused.'
+        ),
+        _('Pause game execution'),
+        _('Debugger Tools'),
+        'res/ribbon_default/bug32.png',
```
4ian, line 36: Can you copy this image in `res/actions` for staying consistent in the usage of icons? Thanks!

Extensions/DebuggerTools/debuggertools.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * @file
+ * Tools for interracting with the debugger.
```
4ian, line 3: typo

4ian:
Alright looks good! 👍

PR #1807, opened 2020-06-10
Wrap some Mainframe functions in useCallback

NilayMajorwar:
This PR wraps some essential Mainframe functions, like `saveProject`, `chooseProject` etc and their dependency functions in useCallback, to avoid things like IPC listeners being re-registered on every render, and also to prepare for adding the new command palette. I've also reordered some functions to avoid use before declaration.

Also, the PR fixes some minor warnings, coming from two other files.


4ian:
Looks good let's go ahead with this :)

PR #1808, opened 2020-06-10
Prevent builtin extension override

### LINKED ISSUE #1793: Overwriting of builtin extensions possible
## Describe the bug
When you create an eventFunctionExtension, you can name it like a with GDevelop bundled Extension and it will override that extension.

## To Reproduce
1. Create a new events Functions extension
2. Rename it to "Sprite"
3. See how it replaced the extension (through the event sheet or seeing in the console the loading message saying `(replacing existing extension)`.

## Possible fix
Refuse to Rename to an existing extension's name (revert renaming and display an alert)

arthuro555:
Fix #1793


newIDE/app/src/ProjectManager/EventFunctionExtensionNameVerifier.js
```diff
@@ -0,0 +1,10 @@
+//@flow
+export function isExtensionNameTaken(
+  name: string,
+  project: gdProject
+): boolean {
+  for (let extensionName of project.getUsedExtensions().toJSArray()) {
```
4ian, line 6: As we talked in another PR, getUsedExtensions is not containing all the extensions. In particular, all the extensions implemented in JS are not there, and we should avoid this function at all price for now.

I think we should instead check the names of all the `gd.JsPlatform.get().getAllPlatformExtensions()` :) 
Note that these extensions will also include the extensions generated from the "EventsFunctionExtension" of the project... which is ok because it will just be redundant with `project.hasEventsFunctionsExtensionNamed` (both are needed though, if you just added a events functions extension to your project, it's not yet generated and loaded as a real extension in the JsPlatform. So you need both to guard against the loaded extension names and the `hasEventsFunctionsExtensionNamed `).

4ian:
Let's get rid of `getUsedExtensions`. If it's working correctly with the solution I'm suggesting, we should be good to go :)

newIDE/app/src/ProjectManager/EventFunctionExtensionNameVerifier.js
arthuro555: oh, yes, right I forgot about that 😓I'll change that right away

arthuro555:
Ok I corrected that :)

arthuro555:
@4ian do I need to do something else?

4ian:
Looks good, merging now! :) Thanks for working on this, will avoid weird bugs and issues 👍

PR #1810, opened 2020-06-10
Add 32 bit build for windows and linux (and arm for linux (RPI))

arthuro555:
This adds 32 bits build builds to the configuration (and also an arm build, aka RPI build)
Trello card: https://trello.com/c/ovBYewSk

### DIFF
--- a/newIDE/electron-app/package.json
+++ b/newIDE/electron-app/package.json
@@ -32,6 +32,29 @@
         "to": "GDJS"
       }
     ],
+    "win": {
+      "target": [
+        {
+          "target": "nsis",
+          "arch": [
+            "x64",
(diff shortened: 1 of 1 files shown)


Bouh:
It doesn't bother me to do some build for the 32bits.
Even if i didn't use it. When i build it, i check just if i can open a project with the 32bits exe under my 64bits system lol.

I haven't had anyone asking me support for a specific 32bits bug.
I think it's ok to say that no support will be give for bugs, the message should be very explicitly about it when you wish download the files from the website.
It's just a bonus to have this version still in use at the moment, which will eventually be totally obsolete because people will finally have a 64bits system, I hope one day!

PR #1815, opened 2020-06-13
Add option for clear on shape painter

Bouh:
Based on [this request](https://forum.gdevelop-app.com/t/change-request-dont-clear-shape-painter-on-every-frame/25395) from @arthuro555 

By default the checkbox for clear is enable for avoid issue with existing project using this object.
I have tested with the example "advanced shape painter" and it's work fine too me.

Ready for review, let me know.

### DIFF
--- a/Extensions/PrimitiveDrawing/Extension.cpp
+++ b/Extensions/PrimitiveDrawing/Extension.cpp
@@ -339,6 +339,36 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
       .SetFunctionName("closePath")
       .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
 
+  obj.AddAction("ClearBetweenFrames",
+                _("Clear between frames"),
+                _("Clear the renderer before the next draw"),
+                _("The shape painter _PARAM0_ is cleared before the next draw: _PARAM1_"),
+                _("Setup"),
+                "res/actions/visibilite24.png",
+                "res/actions/visibilite.png")
+
+      .AddParameter("object", _("Shape Painter object"), "Drawer")
+      .AddParameter("yesorno", _("Clear the previous draw"), "", true)
+      .SetDefaultValue("yes")
+      .SetFunctionName("SetClearBetweenFrames")
+      .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
+
+  obj.AddCondition(
+         "ClearBetweenFrames",
+         _("Clear between frames"),
+         _("Test if the renderer is cleared at each frames."),
+         _("The shape painter _PARAM0_ is clear before the next draw: _PARAM1_"),
+         _("Setup"),
+         "res/conditions/visibilite24.png",
+         "res/conditions/visibilite.png")
+
+      .AddParameter("object", _("Shape Painter object"), "Drawer")
+      .AddParameter("yesorno", _("Clear the renderer at each frames is enabled (yes by default)"), "", true)
+      .SetDefaultValue("yes")
+      .SetFunctionName("IsClearedBetweenFrames")
+      .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
+
+
   obj.AddAction("FillColor",
                 _("Fill color"),
                 _("Change the color used when filling"),
--- a/Extensions/PrimitiveDrawing/Extension.cpp
+++ b/Extensions/PrimitiveDrawing/Extension.cpp
@@ -217,7 +217,7 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
                 _("Begin fill path"),
                 _("Begin to draw a simple one-color fill. Subsequent actions, such as \"Path line\" (in the Advanced category) can be used to draw. Be sure to use \"End fill path\" action when you're done drawing the shape."),
                 _("Begins drawing filling of an advanced path "
-                  "with _PARAM0_"),
+                  "with _PARAM0_ (start: _PARAM1_;_PARAM2_)"),
                 _("Advanced"),
                 "res/actions/beginFillPath24.png",
                 "res/actions/beginFillPath.png")
--- a/Extensions/PrimitiveDrawing/Extension.cpp
+++ b/Extensions/PrimitiveDrawing/Extension.cpp
@@ -341,14 +341,14 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
 
   obj.AddAction("ClearBetweenFrames",
                 _("Clear between frames"),
-                _("Clear the renderer before the next draw"),
-                _("The shape painter _PARAM0_ is cleared before the next draw: _PARAM1_"),
+                _("Activate (or deactivate) the clearing of the rendered shape at the beginning of each frame."),
+                _("Clear the rendered image of _PARAM0_ between each frame: PARAM1"),
                 _("Setup"),
                 "res/actions/visibilite24.png",
                 "res/actions/visibilite.png")
 
       .AddParameter("object", _("Shape Painter object"), "Drawer")
-      .AddParameter("yesorno", _("Clear the previous draw"), "", true)
+      .AddParameter("yesorno", _("Clear between each frame"), "", true)
       .SetDefaultValue("yes")
       .SetFunctionName("SetClearBetweenFrames")
       .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
@@ -356,18 +356,15 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
   obj.AddCondition(
          "ClearBetweenFrames",
          _("Clear between frames"),
-         _("Test if the renderer is cleared at each frames."),
-         _("The shape painter _PARAM0_ is clear before the next draw: _PARAM1_"),
+         _("Check if the rendered image is cleared between frames."),
+         _("_PARAM0_ is clearing its rendered image between each frame"),
          _("Setup"),
          "res/conditions/visibilite24.png",
          "res/conditions/visibilite.png")
 
       .AddParameter("object", _("Shape Painter object"), "Drawer")
-      .AddParameter("yesorno", _("Clear the renderer at each frames is enabled (yes by default)"), "", true)
-      .SetDefaultValue("yes")
       .SetFunctionName("IsClearedBetweenFrames")
       .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
-
 
   obj.AddAction("FillColor",
                 _("Fill color"),
--- a/Extensions/PrimitiveDrawing/ShapePainterObject.cpp
+++ b/Extensions/PrimitiveDrawing/ShapePainterObject.cpp
@@ -179,13 +179,6 @@ std::size_t RuntimeShapePainterObject::GetNumberOfProperties() const {
 #endif
 
 /**
- * Change if the last draw is clear
- */
-void ShapePainterObjectBase::SetClearBetweenFrames(bool value) {
-  clearBetweenFrames = value;
-}
-
-/**
  * Change the color filter of the sprite object
  */
 void ShapePainterObjectBase::SetFillColor(unsigned int r,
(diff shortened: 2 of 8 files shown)


Extensions/PrimitiveDrawing/ShapePainterObject.cpp
```diff
@@ -72,6 +73,10 @@ void ShapePainterObjectBase::DoUnserializeFrom(
       element.GetChild("absoluteCoordinates", 0, "AbsoluteCoordinates")
           .GetValue()
           .GetBool();
+  clearAtEachFrame =
+      element.GetChild("clearAtEachFrame", 0, "ClearAtEachFrame")
+          .GetValue()
+          .GetBool();
```
4ian, line 79: You can just do `element.GetChild("clearBetweenFrames").GetValue().GetBool()` (no need for the third parameter because there is no compatibility to ensure with older serialised value that would have another name).

This being said, we should take care *here* I think of the value to set if there is no "clearBetweenFrames". Because for projects that don't have it, we want it to be `true`, for backward compatibility.
So:

```
clearBetweenFrames = element.HasChild("clearAtEachFrame") ? element.GetChild("clearBetweenFrames").GetValue().GetBool() : true;
```

Extensions/PrimitiveDrawing/ShapePainterObject.h
```diff
@@ -81,6 +81,9 @@ class GD_EXTENSION_API ShapePainterObjectBase {
   inline void SetCoordinatesRelative() { absoluteCoordinates = false; }
   inline bool AreCoordinatesAbsolute() { return absoluteCoordinates; }
 
+  inline void SetClearAtEachFrame(bool value = true) { clearAtEachFrame = value; }
```
4ian, line 84: Let's remove the default parameter here, it's not actually used because we're always passing a value (true or false). The default value for existing project must be set at the step of unserialization.

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -217,7 +217,7 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
                 _("Begin fill path"),
                 _("Begin to draw a simple one-color fill. Subsequent actions, such as \"Path line\" (in the Advanced category) can be used to draw. Be sure to use \"End fill path\" action when you're done drawing the shape."),
                 _("Begins drawing filling of an advanced path "
-                  "with _PARAM0_"),
+                  "with _PARAM0_ (start: _PARAM1_;_PARAM2_)"),
```
4ian, line 220: 👍

Extensions/PrimitiveDrawing/ShapePainterObject.cpp
```diff
@@ -34,6 +34,7 @@ ShapePainterObjectBase::ShapePainterObjectBase()
       outlineColorG(0),
       outlineColorB(0),
       outlineOpacity(255),
+      clearAtEachFrame(false),
```
4ian, line 37: Can we name this `clearBetweenFrames` everywhere? 😇

newIDE/app/src/ObjectEditor/Editors/ShapePainterEditor.js
```diff
@@ -32,6 +32,19 @@ export default class PanelSpriteEditor extends React.Component<
             this.forceUpdate();
           }}
         />
+        <Checkbox
+          label={
+            <Trans>
+              Clear the render before the next draw
```
4ian, line 38: Clear the rendered image between each frame.

Extensions/PrimitiveDrawing/ShapePainterObject.h
```diff
@@ -81,6 +81,9 @@ class GD_EXTENSION_API ShapePainterObjectBase {
...
+  inline bool AreClearAtEachFrame() { return clearAtEachFrame; }
```
4ian, line 85: `IsClearedBetweenFrames`

newIDE/app/src/ObjectEditor/Editors/ShapePainterEditor.js
```diff
@@ -32,6 +32,19 @@ export default class PanelSpriteEditor extends React.Component<
...
+            </Trans>
+          }
+          checked={shapePainterObject.areClearAtEachFrame()}
+          onCheck={(e, checked) => {
+            if (!checked) shapePainterObject.setClearAtEachFrame(false);
+            else shapePainterObject.setClearAtEachFrame(true);
```
4ian, line 44:
```suggestion
            shapePainterObject.setClearAtEachFrame(checked);
```

4ian:
Thanks for working on this! Added some comments :) I think we could also have a condition and an action to change this at runtime, what do you think? :)

Bouh:
The action work well, but i have a doubt for declare the condition in **Extension.cpp**.

What about this code for GD_IDE_ONLY ?
https://github.com/4ian/GDevelop/blob/ccc29ef41687a17942aafb99007946b2f15437b2/Extensions/PrimitiveDrawing/ShapePainterObject.cpp#L124-L176

Bouh:
Something is still wrong, with the condition 🧐

All sentences are to do, but because your are very picky about it i didn't spend my time on it.

Off topic:
Each time i compile libgd.js,  all files in the folder **GDevelop.js/types/** are updated with a double comment, and sometime the content is also doubled, like this one.
image

4ian:
> What about this code for GD_IDE_ONLY ?

That was for the native engine, I think you can forget about this.

> The yes/no button dosen't work.

What do you mean by "doesn't work" specifically? 
The default value might be ignored, I would suggest not worrying about this anyway (same for the "optional" boolean flag, it's not shown in the editor).

4ian:
> Each time i compile libgd.js, all files in the folder GDevelop.js/types/ are updated with a double comment, and sometime the content is also doubled, like this one.

That's very strange! What's your node.js version?

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -339,6 +339,36 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
       .SetFunctionName("closePath")
       .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
 
+  obj.AddAction("ClearBetweenFrames",
+                _("Clear between frames"),
```
4ian, line 343: Clear the rendered image between each frame

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -339,6 +339,36 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+                _("Clear the renderer before the next draw"),
```
4ian, line 344: Activate (or deactivate) the clearing of the rendered shape at the beginning of each frame.

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -339,6 +339,36 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+                _("The shape painter _PARAM0_ is cleared before the next draw: _PARAM1_"),
```
4ian, line 345: Clear the rendered image of _PARAM0_ between each frame: _PARAM1_

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -339,6 +339,36 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+                _("Setup"),
+                "res/actions/visibilite24.png",
+                "res/actions/visibilite.png")
+
+      .AddParameter("object", _("Shape Painter object"), "Drawer")
+      .AddParameter("yesorno", _("Clear the previous draw"), "", true)
```
4ian, line 351: "Clear between each frame?"

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -339,6 +339,36 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+      .SetDefaultValue("yes")
+      .SetFunctionName("SetClearBetweenFrames")
+      .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
+
+  obj.AddCondition(
+         "ClearBetweenFrames",
+         _("Clear between frames"),
+         _("Test if the renderer is cleared at each frames."),
```
4ian, line 359: "Check if the rendered image is cleared between frames."

arthuro555:
> Each time i compile libgd.js, all files in the folder GDevelop.js/types/ are updated with a double comment, and sometime the content is also doubled, like this one.

I had this too btw

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -339,6 +339,36 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+         _("The shape painter _PARAM0_ is clear before the next draw: _PARAM1_"),
+         _("Setup"),
+         "res/conditions/visibilite24.png",
+         "res/conditions/visibilite.png")
+
+      .AddParameter("object", _("Shape Painter object"), "Drawer")
+      .AddParameter("yesorno", _("Clear the renderer at each frames is enabled (yes by default)"), "", true)
```
4ian, line 366: It's a condition, so no "yesorno" parameter. User can invert the condition if they want to.

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -339,6 +339,36 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+      .AddParameter("yesorno", _("Clear the previous draw"), "", true)
+      .SetDefaultValue("yes")
+      .SetFunctionName("SetClearBetweenFrames")
+      .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
+
+  obj.AddCondition(
+         "ClearBetweenFrames",
+         _("Clear between frames"),
+         _("Test if the renderer is cleared at each frames."),
+         _("The shape painter _PARAM0_ is clear before the next draw: _PARAM1_"),
```
4ian, line 360: `_PARAM0_ is clearing its rendered image between each frame`

Extensions/PrimitiveDrawing/Extension.cpp
Bouh: oooh th'as why maybe my condition isn't working.

Extensions/PrimitiveDrawing/Extension.cpp
Bouh: No need "?", we have the checkbox for it :)

4ian:
> Each time i compile libgd.js, all files in the folder GDevelop.js/types/ are updated with a double comment, and sometime the content is also doubled, like this one.

Is there anything suspicious in the output of the terminal (search for "Properly generated GDevelop.js types" or "The output of "webidl-tools flow" is suspicously long. Considering as an error") ?

Bouh:
>That's very strange! What's your node.js version?

I use node v8.15.0



> > Each time i compile libgd.js, all files in the folder GDevelop.js/types/ are updated with a double comment, and sometime the content is also doubled, like this one.
> 
> I had this too btw

Cool it's not my fault 😏
I've seen these errors:
image

So after this one i've launched npm install in Gdevelop.js folder. Now the error is:
image


Everything work well now, condition, action, property, thank you 👍

4ian:
> So after this one i've launched npm install in Gdevelop.js folder

You need to npm install indeed, but in addition it seems that webidl-tools is not running properly on Windows. I'll take a look!

arthuro555:
> You need to npm install indeed, but in addition it seems that webidl-tools is not running properly on Windows. I'll take a look!

I didn't actually try it on windows directly, I only tried via a Ubuntu WSL instance.

4ian:
You should still get some git warnings about line endings which I'm unsure how to solve, but at least you should be able to now have the types generation working on Windows on latest master :)

GDevelop.js/__tests__/GDJS.js
```diff
@@ -312,6 +312,16 @@ describe('libGD.js - GDJS related tests', function() {
       expect(object.areCoordinatesAbsolute()).toBe(false);
     });
   });
+  describe('ShapePainterObject', function() {
+    it('should expose ShapePainterObject specific methods', function() {
```
4ian, line 316: Thanks for the test 👍

4ian:
Looks good it seems! Run auto-formatting and should be good to go!

4ian:
Merged! Thanks for adding this 👍

Issue #1357, opened 2020-01-09
Expose more variables of PlatformerObject


Level0gamedev:
## Description
At this time PlatformerObject behavior is very "closed up" and separate from the rest of the engine. The changes the behavior does to an object are not accessible through Common Conditions of all Objects (e.g. Speed can't be compared). In fact, none of the current values for the object are accessible via events (only the variables for Options are). Some variables are being able to be read (onFloor, Jumping, etc.) but not changed by the user. This limits the PlatformerObject behavior considerably.
 
## Solution suggested
Allow the user to read/write following PlatformerObject variables (taken from platformerobjectruntimebehavior.js) via the editor:

- this._currentFallSpeed
- this._currentSpeed 
- this._currentJumpSpeed

- this._canGrabPlatforms
- this._canJump
- this._isOnFloor
- this._isOnLadder

## Alternatives considered
Current alternative to this problem is to either recreate a platformer engine with events OR using the Physics. 

Alternative solution is to connect the above values to events from Common Conditions of all Objects.

4ian:
There is indeed surely a few conditions/expression to add :)
Precisely, this is what is I think missing:

`this._currentFallSpeed` => getter method to add in JS + condition/expression
`this._currentSpeed` => getter method to add in JS + condition/expression
`this._currentJumpSpeed ` => getter method to add in JS + condition/expression
`this. _canGrabPlatforms ` => condition and action (getter in JS, and the setter `setCanGrabPlatforms` method is actually already coded)
`this._canJump ` => action already existing, getter in JS + condition to do.
`this._isOnLadder ` => getter in JS + condition.

Not all of these have actions because some are properties that describe a state, but can't be modified alone (or this would risk breaking the assumptions/internal state of the platformer) (like isOnFloor/isOnLadder).

4ian:
Closing as card created on Trello

PR #1819, opened 2020-06-16
Expose more variables of PlatformerObject

Bouh:
Still in WIP see the progress on the [trello card here](https://trello.com/c/EqPJE0hd/393-add-conditions-expressions-and-actions-to-compare-the-current-speed-falling-speed-jump-speed-etc-of-a-platformer-object
).

@Silver-Streak
I'am not sure to add actions for **fall speed** and **current speed**, these values are manage by the behavior and can be limited with max value in the behavior and the actions.
This seem useless.

Linked to https://github.com/4ian/GDevelop/issues/1357#issuecomment-572716021

### DIFF
--- a/Extensions/PlatformBehavior/Extension.cpp
+++ b/Extensions/PlatformBehavior/Extension.cpp
@@ -501,6 +501,33 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
         .SetFunctionName("IgnoreDefaultControls")
         .SetIncludeFile("PlatformBehavior/PlatformerObjectRuntimeBehavior.h");
 
+    aut.AddAction("CanGrabPlatforms",
+                  _("Can Grab Platform"),
+                  _("De/activate when the platformer object can grab the platform."),
+                  _("_PARAM0_ can grab the platforms: _PARAM2_"),
+                  _("Controls"),
+                  "res/conditions/keyboard24.png",
+                  "res/conditions/keyboard.png")
+        .AddParameter("object", _("Object"))
+        .AddParameter("behavior", _("Behavior"), "PlatformerObjectBehavior")
+        .AddParameter("yesorno", _("Can grab the platfroms"))
+        .MarkAsAdvanced()
+        .SetFunctionName("SetCanGrabPlatforms")
+        .SetIncludeFile("PlatformBehavior/PlatformerObjectRuntimeBehavior.h");
+
+    aut.AddCondition("CanGrabPlatforms",
+                     _("Can grab platforms"),
+                     _("Check if the object can grab the platforms."),
+                     _("_PARAM0_ can grab the platforms"),
+                     "",
+                     "CppPlatform/Extensions/platformerobjecticon24.png",
+                     "CppPlatform/Extensions/platformerobjecticon16.png")
+        .AddParameter("object", _("Object"))
+        .AddParameter("behavior", _("Behavior"), "PlatformerObjectBehavior")
+        .MarkAsSimple()
+        .SetFunctionName("GetCanGrabPlatforms")
+        .SetIncludeFile("PlatformBehavior/PlatformerObjectRuntimeBehavior.h");
+
     aut.AddCondition(
            "CurrentFallSpeed",
            _("Current falling speed"),
--- a/Extensions/PlatformBehavior/Extension.cpp
+++ b/Extensions/PlatformBehavior/Extension.cpp
@@ -501,6 +501,22 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
         .SetFunctionName("IgnoreDefaultControls")
         .SetIncludeFile("PlatformBehavior/PlatformerObjectRuntimeBehavior.h");
 
+    aut.AddCondition(
+           "CurrentFallSpeed",
+           _("Current falling speed"),
+           _("Compare the current falling speed of the object (in pixels per "
+             "second)."),
+           _("the current falling speed"),
+           _("Options"),
+           "CppPlatform/Extensions/platformerobjecticon24.png",
+           "CppPlatform/Extensions/platformerobjecticon16.png")
+        .AddParameter("object", _("Object"))
+        .AddParameter("behavior", _("Behavior"), "PlatformerObjectBehavior")
+        .UseStandardRelationalOperatorParameters("number")
+        .MarkAsAdvanced()
+        .SetFunctionName("GetCurrentFallSpeed")
+        .SetIncludeFile("PlatformBehavior/PlatformerObjectRuntimeBehavior.h");
+
     aut.AddExpression("Gravity",
                       _("Gravity"),
                       _("Get the gravity applied on the object"),
--- a/Extensions/PlatformBehavior/Extension.cpp
+++ b/Extensions/PlatformBehavior/Extension.cpp
@@ -578,6 +578,16 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
                       "CppPlatform/Extensions/platformerobjecticon16.png")
         .AddParameter("object", _("Object"))
         .AddParameter("behavior", _("Behavior"), "PlatformerObjectBehavior");
+    
+    aut.AddExpression("CurrentFallSpeed",
+                      _("Current fall speed"),
+                      _("Current fall speed"),
+                      _("Options"),
+                      "CppPlatform/Extensions/platformerobjecticon16.png")
+        .AddParameter("object", _("Object"))
+        .AddParameter("behavior", _("Behavior"), "PlatformerObjectBehavior")
+        .SetFunctionName("GetCurrentFallSpeed")
+        .SetIncludeFile("PlatformBehavior/PlatformerObjectRuntimeBehavior.h");
 #endif
   }
   {
--- a/Extensions/PlatformBehavior/Extension.cpp
+++ b/Extensions/PlatformBehavior/Extension.cpp
@@ -389,7 +389,7 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
         .AddParameter("object", _("Object"))
         .AddParameter("behavior", _("Behavior"), "PlatformerObjectBehavior")
         .MarkAsSimple()
-        .SetFunctionName("GetCanJump")
+        .SetFunctionName("canJump")
         .SetIncludeFile("PlatformBehavior/PlatformerObjectRuntimeBehavior.h");
 
     aut.AddAction("SimulateLeftKey",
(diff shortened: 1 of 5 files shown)


Silver-Streak:
> Still in WIP see the progress on the [trello card here](https://trello.com/c/EqPJE0hd/393-add-conditions-expressions-and-actions-to-compare-the-current-speed-falling-speed-jump-speed-etc-of-a-platformer-object).
> 
> @Silver-Streak
> I'am not sure to add actions for **fall speed** and **current speed**, these values are manage by the behavior and can be limited with max value in the behavior and the actions.
> This seem useless.
> 
> Linked to [#1357 (comment)](https://github.com/4ian/GDevelop/issues/1357#issuecomment-572716021)

The idea would be someone could add an instant burst of speed to the current speed (I assume limited by the max speeds). 

Think of when doing a spin-dash in sonic, or a speed boost pad in F-Zero?

Bouh:
In this case you need apply a big value on the **Acceleration** during short time.
Need confirmation by Florian.

Silver-Streak:
> In this case you need apply a big value on the **Acceleration** during short time.
> Need confirmation by Florian.

Oh! Yes that makes sense now that I think of it. After these conditions/expressions are all done I could probably make an example of a speed boost using acceleration modification.

Bouh:
Everything is done, the review can start.

Extensions/PlatformBehavior/Extension.cpp
```diff
@@ -501,6 +514,81 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
         .SetFunctionName("IgnoreDefaultControls")
         .SetIncludeFile("PlatformBehavior/PlatformerObjectRuntimeBehavior.h");
 
+    aut.AddAction("CanGrabPlatforms",
+                  _("Can Grab Platform"),
+                  _("De/activate when the platformer object can grab the platform."),
+                  _("_PARAM0_ can grab the platforms: _PARAM2_"),
```
4ian, line 520: Allow _PARAM0_ to grab platforms: _PARAM2_

Extensions/PlatformBehavior/Extension.cpp
```diff
@@ -501,6 +514,81 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
         .SetFunctionName("IgnoreDefaultControls")
         .SetIncludeFile("PlatformBehavior/PlatformerObjectRuntimeBehavior.h");
 
+    aut.AddAction("CanGrabPlatforms",
+                  _("Can Grab Platform"),
```
4ian, line 518: Platform grabbing

Extensions/PlatformBehavior/Extension.cpp
```diff
@@ -501,6 +514,81 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
...
+                  _("De/activate when the platformer object can grab the platform."),
```
4ian, line 519: Enable (or disable) the ability of the object to grab platforms when falling near to one.

Extensions/PlatformBehavior/Extension.cpp
```diff
@@ -501,6 +514,81 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
...
+                  _("_PARAM0_ can grab the platforms: _PARAM2_"),
+                  _("Controls"),
+                  "res/conditions/keyboard24.png",
+                  "res/conditions/keyboard.png")
+        .AddParameter("object", _("Object"))
+        .AddParameter("behavior", _("Behavior"), "PlatformerObjectBehavior")
+        .AddParameter("yesorno", _("Can grab the platfroms"))
```
4ian, line 526: "Can grab platforms?"

Extensions/PlatformBehavior/Extension.cpp
```diff
@@ -501,6 +514,81 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
...
+        .MarkAsAdvanced()
+        .SetFunctionName("SetCanGrabPlatforms")
+        .SetIncludeFile("PlatformBehavior/PlatformerObjectRuntimeBehavior.h");
```
4ian, line 529: If the function is not implemented in C++, don't call SetFunctionName and don't call SetIncludeFile. In the hypothesis of GDCpp being added back to the editor, the action without an associated function won't be shown in the editor.

Extensions/PlatformBehavior/platformerobjectruntimebehavior.js
```diff
@@ -933,6 +933,46 @@ gdjs.PlatformerObjectRuntimeBehavior.prototype.getJumpSustainTime = function() {
...
+/**
+ * Get the current jump speed of the Platformer Object.
+ * @returns {number} The current jump speed.
+ */
+gdjs.PlatformerObjectRuntimeBehavior.prototype.getCurrentJumpSpeed = function() {
+  return this._currentJumpSpeed;
+};
+
+/**
+ * Get if the Platformer Object can grab the platforms.
```
4ian, line 961: `Check if [...]`

Extensions/PlatformBehavior/platformerobjectruntimebehavior.js
```diff
@@ -933,6 +933,46 @@ gdjs.PlatformerObjectRuntimeBehavior.prototype.getJumpSustainTime = function() {
...
+ * @returns {boolean} Platformer Object can grab the platforms.
+ */
+gdjs.PlatformerObjectRuntimeBehavior.prototype.getCanGrabPlatforms = function() {
+  return this._canGrabPlatforms;
+};
+
+/**
+ * Get if the Platformer Object can jump.
```
4ian, line 969: Check if [...]

Extensions/PlatformBehavior/platformerobjectruntimebehavior.js
```diff
@@ -933,6 +933,46 @@ gdjs.PlatformerObjectRuntimeBehavior.prototype.getJumpSustainTime = function() {
...
+ * @returns {boolean} Platformer Object can jump.
```
4ian, line 970: true if the object can jump.

Extensions/PlatformBehavior/platformerobjectruntimebehavior.js
```diff
@@ -933,6 +933,46 @@ gdjs.PlatformerObjectRuntimeBehavior.prototype.getJumpSustainTime = function() {
...
+ * Get the current jump speed of the Platformer Object.
+ * @returns {number} The current jump speed.
+ */
+gdjs.PlatformerObjectRuntimeBehavior.prototype.getCurrentJumpSpeed = function() {
+  return this._currentJumpSpeed;
+};
+
+/**
+ * Get if the Platformer Object can grab the platforms.
+ * @returns {boolean} Platformer Object can grab the platforms.
```
4ian, line 962: true if the object can grab the platforms.

Extensions/PlatformBehavior/JsExtension.cpp
```diff
@@ -104,8 +104,23 @@ class PlatformBehaviorJsExtension : public gd::PlatformExtension {
...
+          "getCurrentFallSpeed");
+      autExpressions["CurrentFallSpeed"].SetFunctionName("getCurrentFallSpeed");
+      autConditions["PlatformBehavior::CurrentSpeed"].SetFunctionName(
+          "getCurrentSpeed");
+      autExpressions["CurrentSpeed"].SetFunctionName("getCurrentSpeed");
+      autActions["PlatformBehavior::CanGrabPlatforms"]
+          .SetFunctionName("setCanGrabPlatforms")
+          .SetGetter("getCanGrabPlatforms");
+      autConditions["PlatformBehavior::CanGrabPlatforms"].SetFunctionName(
+          "getCanGrabPlatforms");
```
4ian, line 117: nit: let's call this function simply `canGrabPlatforms` - the get seems redundant with the "can". Rule of thumb, functions returning a boolean should be called:`is[State]()`, `can[Ability]()` or `has[Something]()`.
Same for the C++ function.

Extensions/PlatformBehavior/JsExtension.cpp
```diff
@@ -104,8 +104,23 @@ class PlatformBehaviorJsExtension : public gd::PlatformExtension {
...
+          .SetFunctionName("setCanGrabPlatforms")
+          .SetGetter("getCanGrabPlatforms");
+      autConditions["PlatformBehavior::CanGrabPlatforms"].SetFunctionName(
+          "getCanGrabPlatforms");    
+      autConditions["PlatformBehavior::CurrentJumpSpeed"].SetFunctionName(
+          "getCurrentJumpSpeed");
+      autExpressions["CurrentJumpSpeed"].SetFunctionName("getCurrentJumpSpeed");
       autActions["PlatformBehavior::SetCanJump"].SetFunctionName("setCanJump");
+      autConditions["PlatformBehavior::CanJump"].SetFunctionName(
+          "getCanJump");
```
4ian, line 123: this can also be simply called `canJump` here and in the JS. Same for the C++ function.

Extensions/PlatformBehavior/platformerobjectruntimebehavior.js
```diff
@@ -933,6 +933,46 @@ gdjs.PlatformerObjectRuntimeBehavior.prototype.getJumpSustainTime = function() {
   return this._jumpSustainTime;
 };
 
+/**
+ * Get the current speed during a fall of the Platformer Object.
```
4ian, line 937: Get the speed at which the object is falling. It is 0 when the object is on a floor, and non 0 as soon as the object leaves the floor.

Extensions/PlatformBehavior/platformerobjectruntimebehavior.js
```diff
@@ -933,6 +933,46 @@ gdjs.PlatformerObjectRuntimeBehavior.prototype.getJumpSustainTime = function() {
...
+ * @returns {number} The current speed during a fall.
```
4ian, line 938: The current fall speed.

Extensions/PlatformBehavior/Extension.cpp
Bouh: SetCanGrabPlatforms is in C++ here
https://github.com/4ian/GDevelop/blob/master/Extensions/PlatformBehavior/PlatformerObjectRuntimeBehavior.cpp#L501

Bouh: Nevermind it's deleted in the lastest commit.

newIDE/app/src/InstructionOrExpression/EnumerateExpressions.spec.js
```diff
@@ -48,13 +48,16 @@ describe('EnumerateExpressions', () => {
...
     expect(
       filterExpressions(platformerObjectBehaviorExpressions, 'JumpSpeed')
     ).toContainEqual(
       expect.objectContaining({
         type: 'JumpSpeed',
+      }),
+      expect.objectContaining({
+        type: 'CurrentJumpSpeed',
       })
     );
```
4ian, line 62: Unfortunately you can't pass two arguments to toContainEqual.

Instead, save the result of filterExpressions in a variable, then do the checks:
`...toHaveLength(2)`
`...toContainEqual( expect.objectContaining({...)`
`...toContainEqual( expect.objectContaining({...)`

Bouh:
image

4ian:
Thanks for the test results 👍 Tests are failing on semaphoreci because it's using the version of libGD.js that is cached from master - so it's safe to ignore here.
Merging now 👍👍

Issue #1818, opened 2020-06-15
Instance variables in properties panel , blur the focus unsave the change.


Bouh:
## Describe the bug
When we enter a new value or rename a instance variable in the properties panel and select another instance the text isn't saved.

## To Reproduce
1. Select an instance
2. Add a instance variable
3. Do the same on another instance.
4. Type a new value and select directly another instance and go back to the previous.
5. The text is gone.

![mmmmm](https://user-images.githubusercontent.com/1670670/84691876-2feeba00-af45-11ea-9b77-5f680b83a374.gif)

The[ onBlur props](https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/VariablesList/index.js#L235-L252
) on <SortableVariableRow> doesn't work when we select another instance.

## Other details
* b96 web and pc

Bouh:
Solution suggested remove this line:
https://github.com/4ian/GDevelop/blob/6870a53aedea9db94dcfa38cba0c4acff4bec07d/newIDE/app/src/VariablesList/VariableRow.js#L95

I want to know if this can have some repercution somewhere else.
A priori it's good even in the dialog.

4ian:
Would be good to check with the React profiler if this is re-rendering the whole dialog after each key press (there is an option to highlight renderings in the React profiler, useful to see this kind of stuff).

Bouh:
Indeed the whole dialog or properties panel are re-render :/

4ian:
Let's make it a prop so that we use commitOnBlur for the dialog (to avoid useless re-rendering) (and we need to click on Ok to close the dialog, so no risk of losing what the user entered) and not have it for the variables in the properties panel (as it can be dismissed at anytime without warning).

Note that we can't use a "componentWillUnmount" lifecycle method to call a onChange prop, because we risk calling a prop while the component is destroyed because the variable/the list of variable is removed from memory.

Bouh:
[Like this ?](https://github.com/Bouh/GDevelop/commit/c94382f9221928998c6150fdf56d39d99dc4bb6f)

4ian:
> I think i can replace commitOnBlur={true} by commitOnBlur

Yep.

Also rename `commitOnBlur` by `commitVariableValueOnBlur` but apart from this I think that it should work?

Bouh:
This issue can be closed too.

PR #1820, opened 2020-06-16
Instance Variables - commitVariableValueOnBlur

Bouh:
Based on #1818
Ready to review.

### DIFF
--- a/newIDE/app/src/InstancesEditor/InstancePropertiesEditor/index.js
+++ b/newIDE/app/src/InstancesEditor/InstancePropertiesEditor/index.js
@@ -175,10 +175,6 @@ export default class InstancePropertiesEditor extends React.Component<Props> {
                 object ? object.getVariables() : null
               }
               variablesContainer={instance.getVariables()}
-              onSizeUpdated={
-                () =>
-                  this.forceUpdate() /*Force update to ensure dialog is properly positionned*/
-              }
               ref={this._instanceVariablesList}
             />
           </Column>
--- a/newIDE/app/src/VariablesList/VariableRow.js
+++ b/newIDE/app/src/VariablesList/VariableRow.js
@@ -26,6 +26,7 @@ type Props = {|
   variable: gdVariable,
   depth: number,
   errorText?: ?string,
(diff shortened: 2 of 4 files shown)


4ian:
Looks good! I added a comment to explain why we use this prop in the dialog (we use it to avoid useless re-renders, which we can afford in this case because we know the component will get blurred before the dialog is closed).

Removing the commitOnBlur in the case of the usage of VariablesList in `InstancePropertiesEditor` has a downside. Any time I press a key to change a character, the variable is saved, which trigger a re-render of all variables (not ideal, but that's ok, can be an optimization done later) and, worse, a re-render of the whole editor. See how the components are highlighted after I pressed a key:
image

This is quite costly and we should avoid this. The reason is because we have this:
https://github.com/4ian/GDevelop/blob/c2fedf23b98d59d805522cec74504a16ce6d0891/newIDE/app/src/InstancesEditor/InstancePropertiesEditor/index.js#L178-L181

Which I think is useless. The idea of this piece of code was to force a refresh so that the dialog position on screen is updated by material-ui. But:
- `InstancePropertiesEditor` is never shown in a dialog... so I think this was added by error!
- Since the upgrade to the latest material-ui version a few months ago, maybe this hack (search for `onSizeUpdated`) is now useless :) 

For now, would be great if you can:
1) remove the onSizeUpdated in InstancePropertiesEditor. So that changing a variable just re-render all variables (again, sub optimal, maybe there is something better to do).
2) triple check that it works well.

We can look at removing the `onSizeUpdated` in another PR (the issue that without them, when you added a bunch of new variables, the dialog was not getting re-positioned on screen... not super nice when you add tons of variables and that the dialog is expanding outside of the window borders ^^).

Bouh:
Something with memo() ?

Bouh:
It's good 
image

4ian:
> Something with memo() on VariableRow ?

That could maybe help a bit indeed, though not for structures (you need to iterate on it in all cases) (memo/PureComponent would maybe help to avoid re-rendering a row if the variable name and value did not change). 

> It's good

Looks good yep 👍

PR #1821, opened 2020-06-17
Basic command palette

NilayMajorwar:
This PR is for adding a basic command palette to GDevelop. Right now, it only has basic commands like "export game" and "launch preview". I'll be working on more powerful commands over the next few weeks.

pr_snap

To try it out, check the "Enable command palette" flag in preferences, and press `Ctrl + P` (or `Cmd + P`) to open the command palette.


4ian:
Happy to see the PR, that's a great job so far 👍

A few high level comments/ideas:
- If there is a single command (like if I write "export") I would expect it to be selected if I press "Enter". More generally, I think we should ensure that there is always at least one command in the list of shown commands that is selected, so that if you type exactly the proper command, you can press Enter without having to press the "down" key first to highlight one.
- Fuzzy search => what do you think of implementing it now? Will be a very valuable addition :) Not sure about the performance if we have in the future more commands, but I think that should be doable :) I think Material-UI might even have this.
- I'm looking at all the commands in MainFrame, and they make me think a bit of the ElectronMainMenu... do you think we could move them in their own files ("MainFrameCommands"? "MainMenuCommands"?...) with a custom hook? Sure the hook would have a lot of dependency, but that's not really a problem (see `renderMainMenu`).

In any case I can already see how this will be very useful for power users when we add more commands! I think I'm already starting to be a fan of the "preview" command, and the command later to open a scene/etc... Looking forward to the rest :D 
We'll merge this first version once we have adressed what I discussed here - I'll take a new look at the code later.

NilayMajorwar:
> there is always at least one command in the list of shown commands that is selected

Sorry, should've done it already - it was just another MUI prop :sweat_smile:. Done.

> move them in their own files

I've created a new file `MainFrameCommands.js` in the Mainframe folder, with a custom hook enclosing all the useCommands. Feels cleaner now, needs a review though :D

> Fuzzy search

I tried looking for it in Material UI, but seems like MUI had fuzzy search at some point but removed it later - now the [docs suggest using other fuzzy-search libraries](https://material-ui.com/components/autocomplete/#advanced). I'll look for something that suits our use-case and is lightweight. (or should we write our own fuzzy-search?)

4ian:
> it was just another MUI prop 😅. 

Material-UI is incredible :) 

Thanks I'll take a look at the code!

> now the docs suggest using other fuzzy-search libraries. I'll look for something that suits our use-case and is lightweight. (or should we write our own fuzzy-search?)

Turns out we have a very simple implementation already, search for `FuzzyOrEmptyFilter.js` ;) See if you can filter the commands using this :)

NilayMajorwar:
> FuzzyOrEmptyFilter.js

Okay, didn't know we already had that! I got it to work, but with a little modification. Using only fuzzy search led to a bit unpredictable results (search text "pre" showed "O**p**en Sta**r**t Pag**e**" before **pre**view-related commands). So for the palette, I combined simple search with fuzzy,  prioritizing simply-matched commands (so preview-related commands come before "Open start page" when searching "pre"). Seems to work well now :smile:

4ian:
> So for the palette, I combined simple search with fuzzy, prioritizing simply-matched commands (so preview-related commands come before "Open start page" when searching "pre"). Seems to work well now 😄

That's a great idea! I might reuse this for other search fields in the app ;) 
Sounds good then :)

NilayMajorwar:
I've made a little change to the keyboard shortcut hook - earlier it opened the print dialog if the command palette was disabled (by preferences or "by dialog") which might have been irritating if someone tried to open the palette when a dialog was open. Now, the print dialog opens on `Ctrl+P` only if the command palette is disabled in preferences.

newIDE/app/src/CommandPalette/CommandHooks.js
```diff
@@ -0,0 +1,42 @@
...
+};
+
+export const useKeyboardShortcutForCommandPalette = (onOpen: () => void) => {
+  const { values } = React.useContext(PreferencesContext);
+  React.useEffect(
+    () => {
+      const handler = (e: KeyboardEvent) => {
+        if ((e.ctrlKey || e.metaKey) && !e.shiftKey && e.code === 'KeyP') {
+          if (!values.useCommandPalette) return;
+          e.preventDefault();
```
4ian, line 27: Let's add a comment explaining why preventDefault is useful

newIDE/app/src/CommandPalette/CommandHooks.js
```diff
@@ -0,0 +1,42 @@
...
+          const body = document.body;
+          const activeEl = document.activeElement;
+          const mainFrame = document.querySelector('div.main-frame');
+          const isBody = activeEl === body;
+          const isInMainframe = mainFrame && mainFrame.contains(activeEl);
+          if (!isBody && !isInMainframe) return;
```
4ian, line 33: Let's also add a comment about why we need this logic.

newIDE/app/src/CommandPalette/CommandHooks.js
```diff
@@ -0,0 +1,42 @@
...
+
+export const useCommand = (commandName: string, command: Command) => {
+  const commandManager = React.useContext(CommandsContext);
+  React.useEffect(
+    () => {
+      if (!command.enabled) return;
+      commandManager.registerCommand(commandName, command);
+      return () => commandManager.deregisterCommand(commandName);
+    },
+    // eslint-disable-next-line
```
4ian, line 15: Is there a way to avoid this disabling?

newIDE/app/src/MainFrame/index.js
```diff
@@ -1668,6 +1674,38 @@ const MainFrame = (props: Props) => {
       message: 'Update available',
     });
 
+  // Temporary keyboard shortcut for command palette
```
4ian, line 1677: Let's remove the comment about this being "Temporary".
First because often temporary stuff tends to become permanent :D And second because "MainFrame" does not care at all if it's temporary or not - it's the job of the hook to set up the shortcut. So being temporary is a detail of the implementation :)

newIDE/app/src/CommandPalette/CommandPalette.js
```diff
@@ -0,0 +1,114 @@
...
+    if (optionText.includes(searchText)) return directMatches.push(option);
+    if (fuzzyOrEmptyFilter(searchText, optionText))
+      return fuzzyMatches.push(option);
+  });
+
+  return [...directMatches, ...fuzzyMatches];
+};
+
+const useStyles = makeStyles({
+  scrollPaper: {
```
4ian, line 40: As it's unusual to use makeStyles in the codebase and to override a Material-UI style, let's add a comment "Show the dialog containing the Command Palette in the top of the screen."

4ian:
Alright looks good, I just added a few nits!

This looks like a robust basis for the rest. I think we can merge this once the last comments are addressed 👍

> earlier it opened the print dialog if the command palette was disabled (by preferences or "by dialog") which might have been irritating if someone tried to open the palette when a dialog was open. Now, the print dialog opens on Ctrl+P only if the command palette is disabled in preferences.

Excellent, very nice attention to details there :) It's indeed a good idea to totally block the print dialog here when the command palette is enabled in preferences.

newIDE/app/src/CommandPalette/CommandHooks.js
NilayMajorwar: Right now, it's disabled because the `command` object will be a new object for each Mainframe render, so adding `command` to the dependency list would re-register commands on each Mainframe render.
Even if I destructure the properties of `command`, `command.displayText` is a new object on each Mainframe render, due to the `t` macro. So can't include that as dependency either :(

One possible solution is for the hook to just take a static `displayText` string, and apply the `t` macro inside the hook instead. Then destructuring should probably work. I'll try doing that (I have an exam tomorrow, so might take some time :sweat_smile: )

PR #1825, opened 2020-06-18
Dynamic Lights (GSoC)

HarsimranVirk:
Previous PR: #1783


Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -0,0 +1,248 @@
...
+  for (var i = 0; i < 3 * raycast.length; i += 3) {
+    indexBuffer[i] = 0;
+    indexBuffer[i + 1] = i / 3 + 1;
+    if ((i / 3) + 1  !== raycast.length) indexBuffer[i + 2] = i / 3 + 2;
+    else indexBuffer[i + 2] = 1;
+  }
+
+  this._light.geometry
+    .getBuffer('aVertexPosition')
+    .update(new Float32Array(vertexBuffer));
```
HarsimranVirk, line 152: @4ian is there a way to use only one instance of `Float32Array` and updating it, instead of creating a new instance every time? I'm not familiar with "updating" typed arrays. 😅 Maybe we could then use `this._vertexBuffer` and `this._indexBuffer`?

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -105,7 +105,9 @@ gdjs.LightRuntimeObjectPixiRenderer.prototype.ensureUpToDate = function () {
 
 gdjs.LightRuntimeObjectPixiRenderer.prototype.updateGraphics = function () {
   var vertices = [this._object.x, this._object.y];
-  var raycast = this.raycastTest().flat(1);
+  var raycast = this.raycastTest().reduce(function (acc, val) {
+    return acc.concat(val);
+  });
```
HarsimranVirk, line 110: Are `.reduce()` and `.concat()` es5?

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
4ian: Yes totally! :) I think we should store the vertexBuffer and the indexBuffer as properties of the object. 
Float32Array are not resizable, which is I think still kind of fine. We can:
- set _vertexBuffer to null at the creation of the object
- when the vertices are updated, either:
  - the new vertices have the same size as before (will be the case in 99% of cases, because even if an object is moving, most of the time it will stay near the light, and then when it's gone outside of the light radius, well it's gone!), in which case you don't recreate the Float32Array but update it.
  - or the new vertices length is different, in which case you recreate a new Float32Array (there might be some optimisation like keeping the same array if it's not too large, and just ignore the last vertices in the shader??).

Same idea for indexBuffer - we should strive to avoid allocations and reuse them as possible. No reason we have to allocate them at every frame (it's for sure costly when we have lots of lights)

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -0,0 +1,254 @@
...
+};
+
+gdjs.LightRuntimeObjectPixiRenderer.prototype.raycastTest = function () {
+  var result = [];
+  var centerX = this._object.x;
+  var centerY = this._object.y;
+  var halfOfDiag = Math.sqrt(2) * this._radius;
+
+  this._manager.getAllObstaclesAround(this._object, this._radius, result);
+  var hitBoxes = result.map(function (item) {
```
4ian, line 166: See if you can make a "dumb loop" here

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -0,0 +1,254 @@
...
+  var polygons = hitBoxes.reduce(function (acc, val) {
+    return acc.concat(val);
+  });
+  var vertices = polygons
+    .map(function (poly) {
+      return poly.vertices;
+    })
+    .reduce(function (acc, val) {
+      return acc.concat(val);
+    });
```
4ian, line 180: Same, it's more verbose but better make a loop to ensure you only allocate the array where you put the results.

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -0,0 +1,254 @@
...
+      if (a.angle > b.angle) return 1;
+    })
+    .filter(function (element, index, array) {
+      if (index === 0) return true;
+      if (array[index].angle !== array[index - 1].angle) {
+        return true;
+      }
+      return false;
+    })
+    .map(function (element) {
```
4ian, line 251: filter and map will both create 2 new arrays in memory. See if you can make them in a single loop. 
sort is fine as doing changes in place.

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -0,0 +1,254 @@
...
+  });
+  var vertices = polygons
+    .map(function (poly) {
+      return poly.vertices;
+    })
+    .reduce(function (acc, val) {
+      return acc.concat(val);
+    });
+
+  function _calculatePOI(angle) {
```
4ian, line 182: To avoid having the JS engine re-create this function every time `raycastTest` is launched, let's instead make it "static" (I would name it something like: `gdjs.LightRuntimeObjectPixiRenderer.getClosestIntersectionVertex` or `gdjs.LightRuntimeObjectPixiRenderer.getClosestIntersectionPoint`)

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -0,0 +1,254 @@
...
+    var targetX = centerX + halfOfDiag * Math.cos(angle);
+    var targetY = centerY + halfOfDiag * Math.sin(angle);
+    var minSqDist = halfOfDiag * halfOfDiag;
+    var minPOI = [];
```
4ian, line 186: Initialize it to [0,0], otherwise the JS engine don't know the length so as to create in memory an array and then resize it.

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -0,0 +1,254 @@
...
+    for (var poly of polygons) {
+      var poi = gdjs.Polygon.raycastTest(
```
4ian, line 188: Don't use abbreviations: call it "raycastResult" (that's the name used in polygon.js for what this "thing").

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
4ian, line 186: Also, avoid abbreviations. I would simply call it: `closestVertex`

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -0,0 +1,254 @@
...
+  }
+
+  var indexBuffer = [];
+
+  for (var i = 0; i < 3 * raycast.length; i += 3) {
+    indexBuffer[i] = 0;
+    indexBuffer[i + 1] = i / 3 + 1;
+    if (i / 3 + 1 !== raycast.length) indexBuffer[i + 2] = i / 3 + 2;
+    else indexBuffer[i + 2] = 1;
+  }
```
4ian, line 150: It's hard to follow up what's going on here without knowing about the algorithm. Can you add comments for each of the two loop describing what you're constructing?

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -0,0 +1,254 @@
+gdjs.LightRuntimeObjectPixiRenderer = function (runtimeObject, runtimeScene) {
+  this._object = runtimeObject;
+  this._manager = runtimeObject.getObstaclesManager();
+  this._geometry = new PIXI.Geometry();
+  this._radius = runtimeObject.getRadius();
+  this._color = runtimeObject.getColor().map(function (item) {
+    return item / 255;
+  });
+  this.DEBUG = runtimeObject.getDebugMode();
```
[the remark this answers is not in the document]
4ian: Actually `_debugMode` by consistency with the rest

Extensions/Lighting/lightobstacleruntimebehavior.js
```diff
@@ -0,0 +1,106 @@
...
+  } else if (this.activated() && !this._registeredInManager) {
+    this._manager.addObstacle(this);
+    this._registeredInManager = true;
+  }
+
+  //Track changes in size or position
+  if (
+    this._oldX !== this.owner.getX() ||
+    this._oldY !== this.owner.getY()
+  ) {
```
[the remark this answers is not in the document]
HarsimranVirk: Oh, I guess I misunderstood this part, I thought height and width represents the "radius * 2" of the light (whereas it's about the light "obstacle", not light object) 🤦Nevermind, I'll add it back :)

Extensions/Lighting/lightruntimeobject.js
```diff
@@ -0,0 +1,70 @@
...
+    return parseFloat(item);
+  });
+  this._debugMode = lightObjectData.content.debugMode;
+
+  this._obstacleManager = gdjs.LightObstaclesManager.getManager(runtimeScene);
+
+  if (this._renderer)
+    gdjs.LightRuntimeObjectRenderer.call(this._renderer, this, runtimeScene);
+  /** @type {gdjs.TextRuntimeObjectRenderer} */ else
+    this._renderer = new gdjs.LightRuntimeObjectRenderer(this, runtimeScene);
```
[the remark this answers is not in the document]
HarsimranVirk: Yup, this was copy pasted from the Example Object.

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -0,0 +1,254 @@
+gdjs.LightRuntimeObjectPixiRenderer = function (runtimeObject, runtimeScene) {
+  this._object = runtimeObject;
+  this._manager = runtimeObject.getObstaclesManager();
+  this._geometry = new PIXI.Geometry();
+  this._radius = runtimeObject.getRadius();
+  this._color = runtimeObject.getColor().map(function (item) {
```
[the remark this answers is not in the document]
HarsimranVirk: In most of the cases map could be easily avoided :) If it's better to use verbose way then we should do that. Also using declarative methods was the reason for issues with rendering :D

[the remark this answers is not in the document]
4ian: I heavily favor them in the IDE, because in the app it's more important to have a very clear code and you can convey with them the transformation on "data", which is flowing from one function to the other.

In the game engine, the approach is still about keeping things clear of course, but we have to ensure a maximum performance as possible and for this try to allocate "statically" the objects as much as possible, reuse existing variables/arrays, minimize copies of data. Which often means relying to "good old" manual loops, and even avoid creating objects if two numbers, or a few variables can do the trick :) 
You can see at various place in the game engine that instead of passing a new object (for example, the result of raycastTest), it's a static object that is returned (the same object which is always reused).

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
HarsimranVirk: I'll add comments before the loops :) The basic idea is this: 

1. Vertex buffer contains all the points we get after raycasting. Let's take the example of simplest case, when we only have the 4 vertices of quad. 

0th vertex is the center, we then have the other 4 points. This means we store 5 vertices. Therefore the size of vertex buffer = `2 * 4 + 2 = 10`. (4 = number of vertices we get from raycasting, we have factors of 2 here because each vertex has x and y, `+ 2` comes from having center as 0th vertex).

2. Index buffer follows a simple pattern ;) We need to render triangles, so the pattern goes like this: 0, 1, 2, 0, 2, 3, 0, 3, 4, 0, 4, 1 (where 0 represents center, and this pattern forms triangles with center and two adjacent vertices) Also the last triangle should end up at the first vertex to make a close polygon.

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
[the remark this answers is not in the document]
HarsimranVirk: Unrelated stuff, I've seen some code in JsExtension, `.setType('colour')`, should I use that instead of string for color?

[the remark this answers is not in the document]
4ian: Could be a good idea for consistency. This color is stored as a string, you'll need to do transform it, search for `hexToRGBColor` in the codebase and make a static method for it if needed :)

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
4ian: Sounds good, a similar short but helpful explanation in the comment can goes a long way in allowing a reader to understand what's going on :) 
Thinking of indexes, we should give another try at the "Triangle Fan" draw mode, in theory this can reduce by two (more or less?) the number of indexes/vertex because we don't need to repeat the center (nor the previous points).

HarsimranVirk: Yes, I'm going to try triangle fans once more. If it works out there will be a huge benefit to memory (even on the GPU side). We wouldn't need the index buffer at all, and there will be `n + 2` vertices in vertex buffer, n being the number of vertices calculated by the raycasting algo.

Extensions/Lighting/JsExtension.js
```diff
@@ -0,0 +1,316 @@
...
+    ) {
+      var instanceProperties = new gd.MapStringPropertyDescriptor();
+
+      return instanceProperties;
+    };
+
+    const object = extension
+      .addObject(
+        'LightObject',
+        _('Light Object for testing'),
```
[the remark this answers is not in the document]
HarsimranVirk: Should it be called _Light_ or _Light Object_? (It's more or less about the consistency of naming, all the other stuff are called object, like _Tiled Sprite Object_, _Text Object_ etc)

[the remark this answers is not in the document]
4ian: * In the name *displayed to the user*, we should call it simply "Light", as other objects don't have "Object" in their name (it's implied in the interface, you know that you're dealing with an object).
* For the internal object name, "LightObject" (as you already did) is fine and more explicit. This is especially important because the extension is about "Lighting", which is generic, and then provides to the user multiple "building blocks" which are: an object (LightObject), a behavior for marking objects as casting shadows (LightObstacleBehavior).

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
4ian: Would be awesome if we can get this to work 👍

HarsimranVirk: I tried this again, didn't work :( It renders with horrible artifacts 😓 (and sometimes doesn't show up on the screen at all).

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
```diff
@@ -250,3 +255,33 @@ gdjs.LayerPixiRenderer.prototype.isEffectEnabled = function(name) {
...
+      this._renderTexture.resize(
+        this._pixiRenderer.width,
+        this._pixiRenderer.height
+      )
+      this._oldWidth = this._pixiRenderer.width;
+      this._oldHeight = this._pixiRenderer.height;
+  }
+
+  this._pixiRenderer.render(this._pixiContainer, this._renderTexture, false);
+}
```
HarsimranVirk, line 282: @4ian what do you think about this? I guess maybe having a texture this big is the reason for slow fillrate? But rendering a layer to texture would mean rendering the entire screen, right? Either way, I'll maybe checkout a new branch from this point (considering you don't feel like something's wrong here) to try out other things (like actually using this texture in a test game).

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
```diff
@@ -252,59 +257,101 @@ gdjs.LayerPixiRenderer.prototype.enableEffect = function(name, value) {
...
+    this._baseLayer &&
+    (this._baseLayerCameraX !== this._baseLayer.getCameraX() ||
+      this._baseLayerCameraY !== this._baseLayer.getCameraY())
+  ) {
+    this._layer.setCameraX(this._baseLayer.getCameraX());
+    this._layer.setCameraY(this._baseLayer.getCameraY());
+    this._baseLayerCameraX = this._baseLayer.getCameraX();
+    this._baseLayerCameraY = this._baseLayer.getCameraY();
+  }
+};
```
HarsimranVirk, line 357: I'm not sure about this, we are checking whether the base layer moved or not every frame. We could simply add an action to change the camera of lights layer (when we change base layer's camera). I don't think the performance hit we get due to the overhead of checks is really justified.
What do you think @4ian? Should we even do this automatically or not? If yes, do you think we could have a better way of doing it?

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
```diff
@@ -252,59 +257,101 @@ gdjs.LayerPixiRenderer.prototype.enableEffect = function(name, value) {
...
 
   this._pixiRenderer.renderTexture.bind(this._renderTexture);
-  this._pixiRenderer.renderTexture.clear([0.25, 0.25, 0.25, 1.0]);
+  //TODO: Pass data for clear color.
+  this._pixiRenderer.renderTexture.clear([0.5, 0.5, 0.5, 1.0]);
 
   this._pixiRenderer.render(this._pixiContainer, this._renderTexture, false);
-  this._pixiRenderer.renderTexture.bind(oldRenderTexture, oldSourceFrame, undefined);
-}
+  this._pixiRenderer.renderTexture.bind(
```
4ian, line 303: Can you explain why this bind method call is needed?

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
4ian: > Should we even do this automatically or not? 

I think yes, at least we should have an option "automatically follow the base layer" :) 
So that when we're creating a lighting layer we activate:
- "automatically follow the base layer" 
- the render texture
- a default kawase blur filter (much faster than the traditional blur) + automatically use blend mode add on the filter.


>  If yes, do you think we could have a better way of doing it?

I think we can just do a `this._layer.setCameraX(this._baseLayer.getCameraX());` (in gdjs.Layer, not necessarily in the pixi renderer) (this "automatically follow the base layer" option should be something independent from pixi).
No need to store the layer in this._baseLayer nor to do the checks against the X/Y position, because:
- the checks will be more costly than just doing the assignment.
- it's probable that the X/Y position are changed every frame anyway :)

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
HarsimranVirk: Since webgl/opengl is a state machine, there's only one binding point for framebuffers. (Pixi's renderTexture is just a framebuffer with a color attachment). Now, it could very well be the case that something else was bounded before we bind the renderTexture and we should, therefore, leave it to the correct state (ie, bind the `oldRenderTexture` again, it might be `null` in most of the cases, but we don't care, we've left the binding point to the state it initially had).

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
HarsimranVirk: Can you explain a bit more about doing stuff in gdjs.Layer? Do you mean I should move the `syncWithBaseLayer` funtion to gdjs.Layer instead? 

> No need to store the layer in this._baseLayer

I still need some reference to the base layer, right? Even if I move it to gdjs.Layer I need to have some reference to base layer, no?

4ian: > Do you mean I should move the syncWithBaseLayer funtion to gdjs.Layer instead

Yes that's my idea. Having a layer automatically following the base layer is a feature that is not tied to the rendering engine, so I think it should be in gdjs.Layer - the PIXI renderer is just "reading" into gdjs.Layer to do its job of rendering the layer.

>  Even if I move it to gdjs.Layer I need to have some reference to base layer, no?

You can fetch it using `getLayer("")` at every frame I think?

4ian: Note that the base layer is supposed to be always existing (by convention) and there is no reason for it to be changed once created, so keeping a reference to it is ok - but calling `getLayer("")` should be very fast (it's looking into a hashtable aka a JS object) so it's ok to make this call - so that in the future if for some reason the reference to the layer changes (you never know), we're safe.

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
```diff
@@ -65,7 +65,7 @@ gdjs.LayerPixiRenderer.prototype.updateVisibility = function (visible) {
 gdjs.LayerPixiRenderer.prototype.updateTime = function () {
   if (this._renderTexture) {
     this.updateRenderTexture();
-    this.syncWithBaseLayer();
+    this._layer.syncWithBaseLayer();
   }
```
HarsimranVirk, line 69: I realized `syncWithBaseLayer` shouldn't be called inside this if block, maybe there would be some other condition like "automatically follow base layer" which makes more sense.

4ian: Yes that's why I think it's a good thing to put this in gdjs.Layer - the option to  "automatically follow base layer" (which can be a checkbox somewhere in the UI, as well as a condition and action to test/enable/disable it) is not tied to Pixi and should be working entirely from `gdjs.Layer`. In this case, `update` seems like the best candidate where to do this :)

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -231,6 +258,9 @@ gdjs.LightRuntimeObjectPixiRenderer.prototype.raycastTest = function () {
   var result = [];
   this._manager.getAllObstaclesAround(this._object, this._radius, result);
 
+  // Bail out early if there are no obstales.
+  if(result.length === 0) return result;
```
HarsimranVirk, line 262: Question:
Should we maybe return null at this point? Does returning an empty array might make it live longer than it needs to be, and would it pollute the scope of another function (`updateBuffers()`) as well?

4ian: I think it's unlikely to make a huge difference, and returning an empty array is not polluting anything. I think it's fine to keep it! It's less surprising for the caller (you know that you'll get an array in all cases, even if empty in case of no obstacles at all :))

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -0,0 +1,342 @@
...
+    return;
+  }
+
+  var raycastResultLength = raycastResult.length;
+
+  var isSubArrayUsed = false;
+  var vertexBufferSubArray = null;
+  var indexBufferSubArray = null;
+
+  if (this._vertexBuffer.length > 2 * raycastResultLength + 2) {
```
4ian, line 207: Let's add a comment describing what is this manipulation done on the subArray, explaining that it's an optimization to avoid re-allocating buffers :)

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -0,0 +1,342 @@
...
+      .update(vertexBufferSubArray);
+    this._light.geometry.getIndex().update(indexBufferSubArray);
+  }
+};
+
+gdjs.LightRuntimeObjectPixiRenderer.prototype.raycastTest = function () {
+  var result = [];
+  this._manager.getAllObstaclesAround(this._object, this._radius, result);
+
+  // Bail out early if there are no obstales.
```
4ian, line 261: typo: obstacles

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -0,0 +1,342 @@
...
+    this._light.geometry.getIndex().update(this._indexBuffer);
+  } else {
+    this._light.geometry
+      .getBuffer('aVertexPosition')
+      .update(vertexBufferSubArray);
+    this._light.geometry.getIndex().update(indexBufferSubArray);
+  }
+};
+
+gdjs.LightRuntimeObjectPixiRenderer.prototype.raycastTest = function () {
```
4ian, line 257: Let's find a name that explain what is returned/done in this function, and that it's fairly expensive. Something like `computeLightVertices`? (compute = "expensive")

newIDE/app/package.json
```diff
@@ -32,7 +32,7 @@
     "@blueprintjs/icons": "file:src/Utils/BlueprintJsPlaceholder",
     "@lingui/react": "git://github.com/4ian/lingui-react.git#master",
     "@material-ui/core": "4.9.10",
-    "@material-ui/icons": "4.2.1",
+    "@material-ui/icons": "4.9.1",
```
HarsimranVirk, line 35: I didn't find any issues with it, tell me if it's okay to bump up the version :)

4ian: Yes I don't think it's too risky :)

newIDE/app/src/LayersList/index.js
```diff
@@ -46,69 +46,73 @@ class LayersListBody extends Component<*, LayersListBodyState> {
...
+          />
+          {isLightingLayer && (
+            <LightingLayerDialog
+              layer={layer}
+              open={this.state.isLightingLayerDialogOpen}
+              closeLightingLayerDialog={() =>
+                this.setState({
+                  ...this.state,
+                  isLightingLayerDialogOpen: false,
+                })
```
4ian, line 113: nit: `this.setState({ isLightiningLayerDialogOpen: false })`

setState is already doing shallow merge

Core/GDCore/Project/Layer.h
```diff
@@ -177,6 +221,11 @@ class GD_CORE_API Layer {
  private:
   gd::String name;                  ///< The name of the layer
   bool isVisible;                   ///< True if the layer is visible
+  bool isLightingLayer;             ///< True if the layer has lighting properties
+  bool syncWithBaseLayer;           ///< True if the layer automatically follows the base layer
```
4ian, line 225: I know I'm a pain to ask for renaming, but could we rather name this (and associated methods): `followBaseLayerCamera`?

"Follow" indicates the notion of spatial position, while "sync" can be about anything.

Core/GDCore/Project/Layer.cpp
```diff
@@ -29,6 +29,11 @@ void Layer::SetCameraCount(std::size_t n) {
 void Layer::SerializeTo(SerializerElement& element) const {
   element.SetAttribute("name", GetName());
   element.SetAttribute("visibility", GetVisibility());
+  element.SetAttribute("lightingLayer", GetLightingLayer());
+  element.SetAttribute("syncWithBaseLayer", GetSyncWithBaseLayer());
+  element.SetAttribute("r", (int)GetAmbientLightColorRed());
```
4ian, line 34: Prefer saving attributes with the same names as their variables: `ambientLightColorR`, same for G and B

newIDE/app/src/LayersList/LayerRow.js
```diff
@@ -90,10 +90,11 @@ export default ({
...
 
-                if(isLightingLayer) menuTemplate.splice(2, 0, {
-                  label: i18n._(t`Edit lighting layer`),
-                  click: openLightingLayerDialog,
-                });
+                if (isLightingLayer)
+                  menuTemplate.splice(2, 0, {
+                    label: i18n._(t`Edit lighting layer`),
+                    click: openLightingLayerDialog,
+                  });
```
4ian, line 98: nitpicking: To avoid doing a splice, you can do:
```
const menuTemplate = [ 
 { ... },
 ...(isLightingLayer ? [{ label: ..., click: ... }] : [])
 { ... },
]
```

it's slightly more readable because it avoids having the "Edit lighting layer" option to be misplaced if someone update the menu template and forget to change the splice. Also you don't have to count indexes in your head :)

newIDE/app/src/LayersList/LightingLayerDialog.js
```diff
@@ -7,14 +7,24 @@ import ColorField from '../UI/ColorField';
 import { Column, Line, Spacer } from '../UI/Grid';
 import InlineCheckbox from '../UI/InlineCheckbox';
 import Text from '../UI/Text';
+import { type RGBColor, type ColorResult } from '../UI/ColorField/ColorPicker';
 
 type Props = {
-  open: Boolean,
+  open: boolean,
   layer: gdLayer,
   closeLightingLayerDialog: () => void,
```
4ian, line 16: nit: simply name this "onClose"

GDJS/Runtime/types/project-data.d.ts
```diff
@@ -88,6 +88,11 @@ export interface LayerData {
   visibility: boolean;
   cameras: CameraData[];
   effects: EffectData[];
+  r: number;
```
4ian, line 91: nit: see my other comment, let's give these more specific names, and rename "syncWithBaseLayer" to "followBaseLayerCamera"

Core/GDCore/Project/Layer.h
HarsimranVirk: @4ian should I rename the getter/setter as GetFollowBaseLayerCamera/SetFollowBaseLayerCamera?

4ian: Yes please! :)

HarsimranVirk: @4ian I'll be renaming the functions as:

- SetAsLightingLayer()

- IsLightingLayer()

- SetToFollowBaseLayerCamera()

- IsFollowingBaseLayerCamera()

Are they fine? Or suggest something better :)

4ian: For the setters, I would go for:
* SetLightingLayer (or SetIsLightingLayer), with a bool as argument
* SetFollowBaseLayerCamera (same, with a bool as argument)

the getter sounds fine

newIDE/app/src/SceneEditor/ObjectsAdditionalService.js
```diff
@@ -0,0 +1,118 @@
...
+export default {
+  getServices(
+    objectOrInstance: gdObject | gdInitialInstance,
+    layout: gdLayout,
+    isInstance: boolean
+  ): ?AdditionalServices {
+    let objectTypeServices = null;
+    if (isInstance) {
+      objectTypeServices = this.objectType[
+        // $FlowFixMe
```
HarsimranVirk, line 43: @4ian how should I fix this? I'm not sure if I could really check `objectOrInstance` is `gdObject` or `gdInitialInstance`, is there a way of doing that?

newIDE/app/src/SceneEditor/index.js
```diff
@@ -1281,6 +1281,33 @@ export default class SceneEditor extends React.Component<Props, State> {
...
+            this.state.instancesEditorAdditionalServiceInfoBar.identifier
+          }
+          message={
+            <Trans>
+              {this.state.instancesEditorAdditionalServiceInfoBar.message}
+            </Trans>
+          }
+          touchScreenMessage={
+            <Trans>
+              {
```
HarsimranVirk, line 1304: `<InfoBar>` wasn't really working in either `<ObjectsList>` or `InstancesEditor`, I decided to have them in `SceneEditor` instead, which I guess is consistent with the other `InfoBars` as well, is it okay?

newIDE/app/src/ObjectsList/index.js
```diff
@@ -168,6 +171,21 @@ export default class ObjectsList extends React.Component<Props, State> {
     );
     object.setTags(getStringFromTags(this.props.selectedObjectTags));
 
+    const additionalService = ObjectsAdditionalService.getServices(
+      object,
+      objectsContainer,
```
HarsimranVirk, line 176: I'm a bit confused here 😅, objectsContainer is defined as a `gdObjectsContainer`, but in `SceneEditor`, the prop passed for it is a `gdLayout` https://github.com/4ian/GDevelop/blob/2ca593ba2b980b523aa963a5acce7f6dd0153a7c/newIDE/app/src/SceneEditor/index.js#L104 https://github.com/4ian/GDevelop/blob/2ca593ba2b980b523aa963a5acce7f6dd0153a7c/newIDE/app/src/SceneEditor/index.js#L992 Am I missing something here? Although, everything seems to work fine, but flow is complaining right now and I feel something is wrong here. :/

newIDE/app/src/SceneEditor/ObjectsAdditionalService.js
```diff
@@ -0,0 +1,118 @@
+// @flow
```
HarsimranVirk, line 1: We need to find a better name for this I guess, right now it could handle objects and instances, "Objects"AdditionalService seems that it is only related to objects. This could become quite generic and handle much more "events" (not the most accurate word) in the future, not just "onObjectAdded" and "onInstanceAdded", therefore we could potentially automate a lot of stuff. :)

4ian: I would recommend moving this into `newIDE/app/src/Hints`. `Hints/index.js` is already used to expose to the IDE some "hints" to display to the user for some object/behaviors.

Probably we could also simply this because I think the "service" is a bit overengineered.
Basically, we just need two functions `onObjectAdded` and `onInstanceAdded` and they can fetch the object type (in case of instance) by themselves. Internally, they can use a map to associate the object type to a closure called to do the work, but that's just an implementation detail.

newIDE/app/src/SceneEditor/ObjectsAdditionalService.js
4ian: I think you're trying to fit two different stuff in a single function :)
We don't really need something to return these "additional service" because we just care about either running `onObjectAdded` or `onInstanceAdded `

Instead, I would just export these two functions and use them directly (passing them the object or instance, layout, project, and maybe a callback to set the infobar or whatever is needed).

newIDE/app/src/InstancesEditor/InstancesAdder.js
```diff
@@ -1,10 +1,13 @@
 // @flow
 import { roundPosition } from '../Utils/GridHelpers';
+import ObjectsAdditionalService from '../SceneEditor/ObjectsAdditionalService';
+import { type InfoBarDetails } from '../SceneEditor/ObjectsAdditionalService';
 const gd: libGDevelop = global.gd;
 
 type Props = {|
   instances: gdInitialInstancesContainer,
   options: Object,
+  onAdditionalServiceComplete: (infoBarDetails: InfoBarDetails) => void,
```
4ian, line 10: I'm not a big fan of passing "onAdditionalServiceComplete":
- the name is not super helpful by itself ("additional service" is very generic),
- it adds the complexity of understanding what is "InfoBarDetails" to the reader of InstancesAdder. 
- More generally, this is "leaking" abstraction: as InstancesAdder, I should not even know that there is something called "SceneEditor" and I should not even worry about "ObjectsAdditionalService".

In other words, `SceneEditor` **uses** `InstancesEditor`, but not the other way around. 
If we introduce some imports/types/classes from `SceneEditor`, this is a sign that we're starting to make `InstancesEditor` dependend on `SceneEditor` - i.e: we're coupling them.

So how to solve this then? Give to your functions/classes/components only what they need and only what you need to do their job. Here, you want to react when an instance is added? 
Sure thing, just add a prop called "onInstanceAdded", which takes a gdInitialInstance as parameter. Then call it in InstancesAdder.
Then run the additional work *in the parent* (i.e: in SceneEditor), or more generally in the component that is responsible for reacting to this event.

newIDE/app/src/InstancesEditor/InstancesAdder.js
```diff
@@ -39,24 +42,52 @@ const roundPositionsToGrid = (
...
     this._options = options;
     this._temporaryInstances = [];
+    this._onAdditionalServiceComplete = onAdditionalServiceComplete;
   }
 
   setOptions(options: Object) {
     this._options = options;
   }
 
+  onAddInstanceAdditionalservice = (
```
4ian, line 60: As explained in the other comment, let's entirely remove this from this class. Just pass to the class a function to be called when an instance is added. The parent will figure out what to do, but as `InstancesAdder` you:
- don't know what will happen
- and you don't want to know! ;) Your job is to put instances on the scenes, not to run the work of others.

newIDE/app/src/InstancesEditor/InstancesAdder.js
```diff
@@ -39,24 +42,52 @@ const roundPositionsToGrid = (
...
   /**
    * Immediately create new instance at the specified position
    * (specified in scene coordinates).
    */
-  addInstances = (pos: [number, number], objectNames: Array<string>) => {
+  addInstances = (
+    pos: [number, number],
+    objectNames: Array<string>,
+    project: gdProject,
+    layout: gdLayout
```
4ian, line 89: Let's not pass the project nor the layout. If the parent needs them, it will have them. The InstancesAdder is not supposed to need them, so let's not pass them.

newIDE/app/src/InstancesEditor/InstancesAdder.js
```diff
@@ -106,6 +148,8 @@ export default class InstancesAdder {
       instance.setY(newPos[1]);
       instance.setZOrder(zOrder);
 
+      this.onAddInstanceAdditionalservice(instance, project, layout);
+
       return instance;
```
4ian, line 153: Same as before, just pass a callback to this class

newIDE/app/src/InstancesEditor/index.js
```diff
@@ -47,6 +48,7 @@ export type InstancesEditorPropsWithoutSizeAndScroll = {|
   onInstancesMoved: (instances: Array<gdInitialInstance>) => void,
   onInstancesResized: (instances: Array<gdInitialInstance>) => void,
   onInstancesRotated: (instances: Array<gdInitialInstance>) => void,
+  onAdditionalServiceComplete: (infoBarDetails: InfoBarDetails) => void,
```
4ian, line 51: Same as `InstancesAdder`, let's not make InstancesEditor dependent on SceneEditor, and let's not make it less reusable by coupling it with a complex "InfoBarDetails" and what not stuff.
See how it's done for existing "events": `onInstancesMoved`, `onInstancesSelected`... there is even something called `onInstancesAdded`!

You probably want to use this :) Surely you'll have to pass the array of added instances as an argument (which is fair).

newIDE/app/src/ObjectsList/index.js
4ian: `gdProject` and `gdLayout` both extends `gdObjectsContainer`, i.e: they are objects container. That should work fine with Flow as long as it's in the proper direction (i.e: you can assign a project or layout to an objects container, but not the reverse)

newIDE/app/src/ObjectsList/index.js
```diff
@@ -168,6 +171,21 @@ export default class ObjectsList extends React.Component<Props, State> {
...
+    );
+    if (additionalService) {
+      additionalService.onObjectAdded(object, project, objectsContainer);
+      const infoBarDetails = additionalService.getInfoBarDetails(
+        'onObjectAdded'
+      );
+      if (infoBarDetails) {
+        this.props.onAdditionalServiceComplete(infoBarDetails);
+      }
+    }
```
4ian, line 187: Much like in InstancesEditor, see if you can run this in the SceneEditor?

newIDE/app/src/SceneEditor/index.js
```diff
@@ -1281,6 +1281,33 @@ export default class SceneEditor extends React.Component<Props, State> {
...
+            this.state.objectsListAdditionalServiceInfoBar.touchScreenMessage
+          }
+        />
+        <InfoBar
+          show={this.state.instancesEditorAdditionalServiceInfoBar.show}
+          identifier={
+            this.state.instancesEditorAdditionalServiceInfoBar.identifier
+          }
+          message={
+            <Trans>
```
4ian, line 1298: Trans is not working on variables, this will generate a translation like this `"{0}"`. Only use `Trans` on "real sentences"
Use `i18n._` instead, passing it the result of a `t` macro, which is called a `MessageDescriptor`.

In other words, store in your object the result of `t`:
```
{
  message: t`Hello world`
}
```

then use i18n._ to get the translation as a string.

See how it's done in `Hints/index.js`

newIDE/app/src/SceneEditor/index.js
4ian: I think it's ok, though I recommend in the state to store only one set of details (we only care about showing one of these info bar at a single time anyway), to avoid polluting the state with both the one for the object and the instance.

newIDE/app/src/InstancesEditor/InstancesAdder.js
4ian: You can think of InstancesEditor as saying to its parent: "hey what's the deal with your info bar and your complex stuff that you imported from SceneEditor? Leave me alone, if you want to know when something happened than just tell me (with a prop/callback) and I'll tell you, but don't bother me doing your work, I'm already doing enough" 😄

newIDE/app/src/ObjectsList/index.js
HarsimranVirk: @4ian I guess we can handle everything in the SceneEditor itself? As I see it, we already have a "stream of events" flowing across different editors to SceneEditor (for eg. InstancesAdder to InstancesEditor to SceneEditor), whereas I
believe I'm exposing SceneEditor to other editors (ie, ObjectsList, InstancesEditor, etc) and introducing a flow in the opposite direction. My initial thought process was to declutter the SceneEditor, but I guess it is going to be cleaner even after adding code in SceneEditor (and making it a bit bulkier).

4ian: Yep I agree that SceneEditor is becoming a bit cluttered/complex, but that is its nature: it's the component that is orchestrating the other editors :) 
Note that React hooks are a partial solution to this because they allow to break a component like this which is big with a lot of logic into smaller reusable logic chunks... unfortuantely for now it's still a class component!

But again it's fine to do this in the SceneEditor, being the receiver of the "stream of event" is basically its role. If something is too complex, it can always be extracted to a function and used by SceneEditor.

newIDE/app/src/SceneEditor/ObjectsAdditionalService.js
4ian: In some way I think of this extra work as a "boosted hint": it's a hint so performant that we're doing work for the user :) hence why I was talking about this `Hints` folder.

HarsimranVirk:
@4ian question:
What is the deal with `onEditObject` in ObjectsList, here https://github.com/4ian/GDevelop/blob/f316d28fe3f10ff276f0c961e80ed220c6fbd37e/newIDE/app/src/ObjectsList/index.js#L148-L183

`onObjectCreated` is called in a callback after the conditional check of `onEditObject`, should I be using this prop? Or should I create a new prop (something called `onNewObjectCreated` maybe) which is called after we insert new object in `objectsContainer`?

4ian:
> onObjectCreated is called in a callback after the conditional check of onEditObject, should I be using this prop? 

Yes I think :) Consider the conditional check of `onEditObject` to be always true - it's an old residue of when the editor could be embedded as a level editor where you could not edit objects. So onEditObject is always defined, and you can reuse I think onObjectCreated to start the additional logic.

newIDE/app/src/SceneEditor/ObjectsAdditionalService.js
HarsimranVirk: Should I move the code in `Hints/index.js` then? And have an object which has two properties (functions), `onObjectAdded` and `onInstance` added, which then does whatever it needs to do (map to correct object types, do extra work, send back infoBarDetails, etc).

4ian: Maybe in the Hints folder but still in a file with a different name, like "ObjectsAdditionalWork.js" (not super fan of "Work" but well!).
It's fine to have "Object" in the name as instances are very tied to the concept of object. 

I was thinking of alternate names like "ObjectAutomatedWork" but can't find anything really satisfying!

HarsimranVirk:
@4ian at the moment I'm trying to make a branch on top of master, and I'm seeing that there's no EffectsListDialog in LayersList anymore 🤔, I added the LightingLayerDialog beside EffectsLayerDialog, but now I believe that wouldn't be the right way? Either way, let's remove everything from the LayersList for now, have a new cherry-picked branch on top of master and work on it afterwards?

NilayMajorwar:
Yeah, I had to move the EffectsList dialog and related state to the SceneEditor because the LayersList is not always rendered on screen whereas I needed "Edit layer effects" command to be available regardless of whether the LayersList is shown on screen.

So to have your new command available regardless of whether the LayersList is shown or not, you'll have to add the dialog and related state in SceneEditor and pass callback(s) to LayersList, like its done for EffectsList. It's indeed a tiny limitation introduced by the command palette implementation 😅

HarsimranVirk:
Closing this PR since we can merge on master now #1881

PR #1826, opened 2020-06-18
[WIP]Rotate the Selection Rectangle

PawBud:
This commit currently does introduce some nasty errors which I was hoping that could be cleared with your help @4ian. 
**This is the error that pops up.** 
Screenshot from 2020-06-18 21-00-28

Screenshot from 2020-06-18 20-54-41
As it is noticed that the `selectionRectangle` doesn't show the angle prop when hovered upon; I tried to find the flow file to see if I had to declare the flow type associated with it but I couldn't and then I noticed that the flow coverage of `index.js` is 0. 
Screenshot from 2020-06-18 21-10-27
I thought of asking you since I am confused as to why this is happening. 

Moreover, Angle is not working that i.e. `this.selectedRectangles[i].angle = {some angle in degrees}` doesn't work but `this.selectedRectangles[i].rotation = {some angle in radians}` does work. So i guess I will need to convert the **selectionRectangle.angle** to radians first!!

Moreover when I tried `this.selectedRectangles[i].rotation = 0.25` the selection box did rotate but it changed its position. I can't understand why since I was under the impression that hitArea was supposed to be taking care of that!
 
Screenshot from 2020-06-18 21-55-40


4ian:
You need to do a bit more research about Pixi.js before continuing :) Please make sure you understand the objects your are manipulating. What is pixi, what is a rect, 

> and then I noticed that the flow coverage of index.js is 0.

yeah some files are not typed with Flow. That's not a big deal, it can be added later.

> the selection box did rotate but it changed its position

Read more about pixi.js :)

4ian:
The Chrome debugger with breakpoints and console.log will be your best friends to understand what's going on there. Look at LayerRenderer.js too, might be useful.

PawBud:
1. So, @4ian I have been reading a lot of Pixi docs and fidelling with Pixi playground to understand some stuff. I have some doubts regarding the `TransformRect.js` https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/Utils/TransformRect.js.

2. the error is now gone since I had not declared `getInstanceAngle`, but with a twist, I removed the layerRender initialization completely. I did try initializing in both the `index.js` file over and `layerRenderer.js` but ended up with the error **this.renderedInstances[instance.ptr].getAngle() is not a function**. I guess i need to understand the functionality of layer renderer properly. 

3. I realized that this function was preventing the angle to be passed into the `selectionRectangle` instance, https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/InstancesEditor/SelectedInstances.js#L201-L205 I tried to `angle` as a param but I am not sure about the meaning of the **startPos** and **endPos** array. Even if I do put it as ```const startPos = transformPoint(x, y, angle);
  const startPos = transformPoint(x, y, angle);```, it does return angle but when I console logged `selectionRectangle.angle` and returned undefined`. 

4. _I think that `TransformPoint` basically transforms the params from local to world constants_. pls, do correct me if I am wrong. 

5. I still have to figure out how to fix the position of the selection rectangle which can be done once I rotate the rectangle synchronously with the sprite.

PawBud:
I will try this possibly later; Others are free to submit their PR's regarding the same. sorry for the inconvenience.

PR #1829, opened 2020-06-21
New example: Board walk with Raycast

paulera:
Adds an example of how to use the Raycast condition to walk through an tile-based arbitrary board.
This example is playable on this link: https://paulera.itch.io/board-walk-with-raycast
The idea started on this topic of the forum: https://forum.gdevelop-app.com/t/approaches-for-walking-an-arbitrary-board-square-by-square/25733
The example is also available here: https://gist.github.com/paulera/900c297128f1bf0f6e5326ec30fe3805 (modifications made so the example can run offline instead of using online resources).


4ian:
Looks great! Thanks for opening a pull request for this 👍

paulera:
@4ian when opening the "Board walk with Raycast" example, the Sprite images are not loading due to a 403 error. Is there something I should have done differently before creating the PR?

image

4ian:
@paulera Sorry for the delay 😬 This is finally fixed :)

PR #1832, opened 2020-06-21
Fix color format given as argument to the effects by the action.

### LINKED ISSUE #1734: Layer effects, wrong color format
## Describe the bug
Color format in layer effects list in Hexa, and decimal (RGB) in events.

In the layer effect list, go with the effect Color Replace. 
- The color format in layer effect are in hexadecimal format. "#FF0000" (value return by the color picker, btw the value is not show in the color feild, nvm..)

https://github.com/4ian/GDevelop/blob/140c7f52cbe4a6f0dea033e3ff7d7c1c3334009b/Extensions/Effects/color-replace-pixi-filter.js#L15

And in the action for change the color the format is in decimal. "255;0;0"
image

There is no convertion between from RGB to Hexa in the actions for the layer effect parameters.

## To Reproduce
[Complete test and source](https://forum.gdevelop-app.com/t/remplacement-de-couleur/24710/2)

## Other details
* B92, web-app and desktop.

Bouh:
Convert the color format from the action for set a new value on a parameters of an effect, from RGB to Hexa.

Fix #1734

**Edit:** Oops sorry for the commit `Fix word`.

### DIFF
--- a/Extensions/Effects/bevel-pixi-filter.js
+++ b/Extensions/Effects/bevel-pixi-filter.js
@@ -19,6 +19,15 @@ gdjs.PixiFiltersTools.registerFilterCreator('Bevel', {
     }
   },
   updateStringParameter: function(filter, parameterName, value) {
+      const splitValue = value.split(';');
+      if (splitValue.length !== 3) return;
+      const hexColor = '#' + gdjs.rgbToHex(
+        parseInt(splitValue[0], 0),
+        parseInt(splitValue[1], 0),
+        parseInt(splitValue[2], 0)
+      );
+      value = hexColor;
+      
     if (parameterName === 'lightColor') {
       filter.lightColor = value.replace('#', '0x');
     }
--- a/Extensions/Effects/bevel-pixi-filter.js
+++ b/Extensions/Effects/bevel-pixi-filter.js
@@ -19,20 +19,11 @@ gdjs.PixiFiltersTools.registerFilterCreator('Bevel', {
     }
   },
   updateStringParameter: function(filter, parameterName, value) {
-      const splitValue = value.split(';');
-      if (splitValue.length !== 3) return;
-      const hexColor = '#' + gdjs.rgbToHex(
-        parseInt(splitValue[0], 0),
-        parseInt(splitValue[1], 0),
-        parseInt(splitValue[2], 0)
-      );
-      value = hexColor;
-      
     if (parameterName === 'lightColor') {
-      filter.lightColor = value.replace('#', '0x');
+      filter.lightColor = gdjs.PixiFiltersTools.rgbOrHexToHexNumber(value);
     }
     if (parameterName === 'shadowColor') {
-      filter.shadowColor = value.replace('#', '0x');
+      filter.shadowColor = gdjs.PixiFiltersTools.rgbOrHexToHexNumber(value);
     }
   },
   updateBooleanParameter: function(filter, parameterName, value) {},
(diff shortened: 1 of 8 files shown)


4ian:
Do you think we could rather do the fix at the editor level?

Bouh:
Yes it's was this issue we have talked by email, [and in this PR](https://github.com/4ian/GDevelop/pull/1724#issuecomment-622996034).

Bouh:
Like the color parameter on the action Change the background color ?
But this action use rgbToHex too in GDJS
```
gdjs.RuntimeScene.prototype.setBackgroundColor = function(r,g,b) {
    this._backgroundColor = parseInt(gdjs.rgbToHex(r,g,b),16);
};
```
You think to another file ?

4ian:
Basically I wonder if we can directly pass to `setEffectStringParameter` the color in the proper format, a hex string, rather than constructing and running a regex everytime it's called (which can be multiple times a frame, so better be quick here).
Maybe a parameter/different type of property to do the distinction between a "color" as a rgb string and as a hex string ("color-hex", something like that?)

Bouh:
Like the parameters for declare the instructions in extension ?
```
.AddParameter("colorHexa", _("Color"))
```

But the parameter for the action can take others string value than a color. It's linked to the effect parameters.
That why the current parameter is just a string.

https://github.com/4ian/GDevelop/blob/a23a8904f659b948a9cc2d75ba1ec6cf72457d77/Core/GDCore/Extensions/Builtin/CameraExtension.cpp#L375-L392


image

For example the value can be a new texture for the Displacement map.

4ian:
Oh right! Well it's still a problem because this means that any string parameter would be impacted by this regex, so everything would be paying the price and worse, I could genuinly pass something like 0;1;2 (like a weird list of numbers for a filter, not super intuitive but well) and it would be converted to a hex string - a bit risky 😬 Maybe it's then a job for the filters in their function to change the color parameters?

Bouh:
Yes the effect of the filter here!
https://github.com/4ian/GDevelop/blob/a23a8904f659b948a9cc2d75ba1ec6cf72457d77/Extensions/Effects/color-replace-pixi-filter.js#L13-L20

4ian:
Excellent! Do you think you could factor this into `gdjs.PixiFiltersTools.rgbOrHexToHexColor` (or a simpler name)? Search for `gdjs.PixiFiltersTools.clampValue`.
So that we can simply do: `filter.color = gdjs.PixiFiltersTools.rgbOrHexToHexColor(parameterName)`.
Also while we're currently passing a string in the form "0x000000", turns out the filter are taking the color as a number: 
image
At least for drop shadow: https://pixijs.io/pixi-filters/docs/PIXI.filters.DropShadowFilter.html

Hence I would recommend:
- make a single function (for code clarity)
- in this function, if there is a ";" in the string, assume it's RGB and convert it to hex with `gdjs.rgbToHexNumber`.
- otherwise (either no ";" or wrong number of components after doing the split in RGB), we assume it's a hex string and do a `parseInt(value.replace('#', '0x'), 16)` (like already done in some filters).

4ian:
Actually as we have `gdjs.rgbToHexNumber`, we should for consistency name this:
`gdjs.PixiFiltersTools.rgbOrHexToHexNumber` (it's implied that "rgb" and "hex" without suffix is a string, much like in rgbToHexNumber).

A bit weird but tells exactly that we're handling two formats :)

4ian:
Merged, thanks! 👍

PR #1834, opened 2020-06-21
Fix app stuck after (hot) reload on Windows in development mode

### LINKED ISSUE #1831: Extensions aren't loaded after reloading.
This is more or less an issue targeted at devs, project files don't open at all after reloading. Auto-open isn't triggered at all, manually opening a file results in an indefinite loading state. This is maybe related to the recent refactoring of mainframe? (Might be because `openFromFileMetadata` was wrapped in a `useCallback` I guess). I haven't followed up with the mainframe for a long time now so I don't know exactly what's happening :) 

EDIT: Looking at the console, even extensions aren't loaded,  is it a windows specific thing (never happened on linux before)?

EDIT 2: It's windows specific, doesn't happen on linux.

4ian:
Fixes #1831


4ian:
See https://github.com/electron/electron/issues/19554 for more information about the underlying problem. For now we're relying too much on `fs` in the renderer to avoid it.

PR #1835, opened 2020-06-23
Update runtime electron to 8.2.5

ssangervasi:
### Motivation
https://github.com/4ian/GDevelop/issues/1469#issuecomment-647344206

### What was changed
This change updates the electron version of the export template
to 8.2.5 to match the electron version that is used in the IDE.

This change also sets `allowRendererProcessReuse` app option to
further match the IDE environment.

This also allows us to uncomment the `webPreferences` browser
window option.

### How I tested this
After making these changes I used my dev build of GDevelop to export the Platformer example game in "Windows/macOS/Linux (manual)" format. I then took that build directory and used electron-builder to build it. The resulting app ran just fine.   
OS: Linux
Electron debugger process.versions:
```json
{
  "chrome": "80.0.3987.165",
  "electron": "8.2.5",
  "node": "12.13.0"
}
```

### DIFF
--- a/GDJS/Runtime/Electron/main.js
+++ b/GDJS/Runtime/Electron/main.js
@@ -16,13 +16,10 @@ function createWindow() {
     height: 600 /*GDJS_WINDOW_HEIGHT*/,
     useContentSize: true,
     title: "GDJS_GAME_NAME",
-    backgroundColor: '#000000'
-    // To be added once upgraded to Electron 8+
-    // (or custom events to be written for each usage of electron
-    // in the game engine):
-    // ,webPreferences: {
-    //   nodeIntegration: true,
-    // }
+    backgroundColor: '#000000',
+    webPreferences: {
+      nodeIntegration: true,
+    }
   });
 
   // Open external link in the OS default browser
@@ -45,6 +42,14 @@ function createWindow() {
     mainWindow = null;
   });
 }
+
+// Should be set to true, which will be the default value in future Electron
+// versions, but then causes an issue on Windows where the `fs` module stops	
+// working in the renderer process after a reload.
+// See https://github.com/electron/electron/issues/22119
+// For now, out of caution, disable this as we rely heavily on `fs` in the 
+// renderer process.
+app.allowRendererProcessReuse = false;
 
 // This method will be called when Electron has finished
 // initialization and is ready to create browser windows.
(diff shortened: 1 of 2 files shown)


ssangervasi:
@4ian - I'm not able to modify the "Reviewers" on the PR, so ping 😸

4ian:
I receive a notification for every new PR ;) 
I think it looks mostly good, would need testing before we go ahead with this :) I can test on macOS, if someone can check on Windows (with electron-builder) that would be perfect :)

GDJS/Runtime/Electron/main.js
```diff
@@ -46,6 +43,10 @@ function createWindow() {
   });
 }
 
+// Explanation taken from: newIDE/electron-app/app/main.js
+// Disable this as we rely heavily on `fs` in the renderer process.
```
4ian, line 47: I would enter a complete description here too, because these part of the codebase are relatively independent :)

```
// Should be set to true, which will be the default value in future Electron
// versions, but then causes an issue on Windows where the `fs` module stops	
// working in the renderer process after a reload.
// See https://github.com/electron/electron/issues/22119
// For now, out of caution, disable this as we rely heavily on `fs` in the 
// renderer process.
```

I added `after a reload` because it's less likely to happen in a game, but still.

newIDE/electron-app/app/main.js
```diff
@@ -52,6 +52,7 @@ protocol.registerSchemesAsPrivileged([{ scheme: 'gdide' }]);
 // working in the renderer process.
 // See https://github.com/electron/electron/issues/22119
 // For now, disable this as we rely heavily on `fs` in the renderer process.
+// This is also done in the runtime: GDJS/Runtime/Electron/main.js
```
4ian, line 55: I think you can remove it, if someone update the runtime or the IDE to a new electron version, it must be done in two separate package.json, so it make sense for this person to go through the two main.js file to verify if there is anything to upgrade :)

Bouh:
Dev mode & installer on windows 10
- [x] Preview 
- [x] Preview & debugger
- [x] Preview & profiler
- [x] Title in preview window change
- [x] Preview

ssangervasi:
I updated with 4ian's comments fixed.
I also tested with electron-builder on windows and was able to export and build successfully.

4ian:
Cool thanks both for testing :)
I'll give a try with the online export and merge then if it's alright 👍

4ian:
Finally got the time to test it, had to do an update in the build service to work with this new version of Electron. Merging now :)

Thanks @ssangervasi for your work on this and @Bouh for testing too! This updated version should allow games to benefit from latest Electron improvements and reduce incompatibilities due to using an old version.

PR #1836, opened 2020-06-24
Graphics for Pixi v5.3.0

Bouh:
This PR is about the migration on Pixi v5.3.0
The main discussion is here:
https://github.com/4ian/GDevelop/pull/1824#issuecomment-649062243

Don't merge yet, i need check the rest of the Graphic :)


4ian:
Merging this as it's fixing stuff - and I think apart from a fix that should be coming from Pixi.js itself in v5.3.1, we're good on the Graphics/Shape Painter side :)

PR #1840, opened 2020-06-28
Add "live previews" a.k.a "hot reloading"

4ian:
This is a branch implement "hot reloading" for previews.

This means that when a preview is started, you can click the preview button again and all your changes are applied to the preview being run - **without** restarting the preview and the game from scratch.

This functionality is working with most features of GDevelop:
- editing instances of the scene (useful to build your level at the same time you're playing to it).
- adding/removing/editing behaviors
  - physics behavior notably, and most behaviors
- adding/removing/editing effects
- add/remove/edit variables (global, scene, object or instances variables should all be supported)
- editing events in extensions (actions/conditions/expressions and behaviors)
- adding/removing/editing objects

The preview button will show a "lightning" icon instead of the "play" icon when a preview is running and where live changes can be applied.
In a few dialogs (effects, object editors, variable editors...) you'll also have a button to apply changes to the running preview.

Editing a level with live preview

See also a video showing how it works with effects: https://www.youtube.com/watch?v=IW9nunJgKtg (unlisted video, just for showing how it works)

There are a few things that are not supported at which point a dialog will invite you to relaunch a preview from scratch. *See also the known issue at the end of this post*.

This also improves integration of the previews with the GDevelop debugger: a connection between the preview and the editor is always made so that you can open the debugger to inspect a preview, even if it was launched before you opened the debugger tab.

## A few things that are known not to work or will be improved later

- [x] Support for network preview will be added soon. For now network preview are reloaded and restarted from scratch. I will remove this to instead support "hot reloading".
- [ ] Moving layers is not supported yet (for now please only play with effects :)).
- [x] "Trigger Once" conditions are **re-triggered after a applying changes to the running preview**.
- [ ] Some large games might have a preview that takes multiple seconds to run, and *this feature is not improving this time*. Incremental re-generation of events (and extensions), to **speed up preview generation**, will be *another feature* that should hopefully reduce the time to run previews a lot for large games.

As this is an advanced feature, a dialog will ask for users to get a subscription to continue supporting GDevelop development :) (once in a while to avoid spamming/losing the speed gain brought by this feature)

## I want to try this feature on this branch

There are changes in the C++ code for export/previews (and other stuff), so you'll need to [recompile GDevelop.js](https://github.com/4ian/GDevelop/tree/master/GDevelop.js).

## If you get a crash of the game after applying changes

In this case, please:
- Open the Developer Console in the preview, and post a screenshot here along with what you changed (maybe I overlooked something, no crashes should ever happen) and any information.
- More generally, as usual, See if you can reproduce it and send here the project + explanations how to reproduce the bug.
- Close the preview then continue to work as usual with a new preview :)


Bouh:
Wow, it's big!
I'll test it this week.
If there are particular things to try let me know.

Is the engine compare parts in projectData with the previous and push only the differences who as changed? (Like can do React with the dom?)
Can you explain in few words how it's work in the engine ?

4ian:
Yes it's big, I would usually recommend doing lots of smaller commits/PRs in normal time, but there because there were big changes I wanted to be sure that I had something working before adding the code to master branch :)

> If there are particular things to try let me know.

Nothing special, I tested/worked on this so much that I can't find new stuff to try, so better have you try stuff as you go so that you don't have any biais :)

> Is the engine compare parts in projectData with the previous and push only the differences who as changed? (Like can do React with the dom?)

Yes that's *exactly* that! :) 
You can take a look at `GDJS/Runtime/websocket-debugger-client/hot-reloader.js`, where most of the interesting parts of the implementation is. Basically this file is doing getting the old project data, the new project data, then walking through the "tree" represented by these data (going through the game, then in scenes, then in objects, instances, etc...) - everytime looking at the difference between the old and new data. If there are changes, then it calls the appropriate methods on the equivalent runtime game/scene/object (including creating new, deleting existing ones...).

(To use your analogy, React = the `gdjs.HotReloader` class from "hot-reloader.js", components tree = the project data, DOM = the gdjs.RuntimeGame)

The hot reloader is also taking care of reloading scripts that have been changed, useful notably for events generated code. 
There are a few points we need to be careful about:
- The events function/behaviors are also reloaded if modified. When it's the case, the JavaScript "class" representing a behavior is erased by the new one in the new script, so we need to re-instantiate these and transfer properties. Otherwise, objects would continue to run with an old version of the code.
  - We use a hash to detect when generated JavaScript files are modified, but unfortunately the generated code is not always stable (there can be difference in the output due to the usage of unique numbers). I'll need to fix that to reduce useless work done when you're editing extensions.
- For the physics engine, I had to add a way to declare to the behavior that the object was manipulated outside events (to update accordingly the Box2D physics body)
- When events are generated, the "Trigger Once" conditions are getting a new unique number, so they are launched again - same issue as before I need to fix that by ensuring the same events are generating the same unique number used to identify "Trigger Once" conditions.

There are a few interesting optimizations (if you click on the "Apply changes to the preview" in an object editor, we only export the project data, not the events). In the future, I hope to be able to generate only events from scenes that have been modified to speed up the preview generation and reloading for large games.

GDJS/Runtime/websocket-debugger-client/hot-reloader.js
```diff
@@ -0,0 +1,1313 @@
...
+  return Promise.all(reloadPromises);
+};
+
+/**
+ * @param {ProjectData} oldProjectData
+ * @param {ProjectData} newProjectData
+ * @param {ChangedRuntimeBehavior[]} changedRuntimeBehaviors
+ * @param {gdjs.RuntimeGame} runtimeGame
+ */
+gdjs.HotReloader.prototype._hotReloadRuntimeGame = function (
```
4ian, line 348: @Bouh Look at this file and all the _hotReloadXXX methods that are checking the changes between the old data and the new data, and then calling the game/scene/object/behavior accordingly to reflect the changes.

Extensions/TiledSpriteObject/tiledspriteruntimeobject.js
```diff
@@ -43,6 +43,20 @@ gdjs.TiledSpriteRuntimeObject = function(runtimeScene, tiledSpriteObjectData)
 gdjs.TiledSpriteRuntimeObject.prototype = Object.create( gdjs.RuntimeObject.prototype );
 gdjs.registerObject("TiledSpriteObject::TiledSprite", gdjs.TiledSpriteRuntimeObject);
 
+gdjs.TiledSpriteRuntimeObject.prototype.updateFromObjectData = function(oldObjectData, newObjectData) {
```
4ian, line 46: @Bouh All objects/behaviors got these new methods to do the "diff" between the old and new data at their level (ideally in the future we could scrap these methods from exported game to save a bit of size).

GDJS/Runtime/websocket-debugger-client/hot-reloader.js
```diff
@@ -0,0 +1,1313 @@
...
+  changedRuntimeBehaviors,
+  runtimeScene
+) {
+  runtimeScene.setBackgroundColor(
+    newLayoutData.r,
+    newLayoutData.v,
+    newLayoutData.b
+  );
+  if (oldLayoutData.title !== newLayoutData.title) {
+    runtimeScene.getGame().getRenderer().setWindowTitle(newLayoutData.title);
```
4ian, line 561: @Bouh Notice how the hot-reloader is always using public methods to do changes to the game (you could do all that is done by the hot reloader using JavaScript code events). 
There is nothing "implicit" and no hidden magic in the sense that the hot-reloader is like an automatic assistant that is calling various functions of the game/scene/objects to reflect the difference it sees between before and after.
A very useful assistant :)

arthuro555:
Another potential improvement would be to automatically generate the updateData methods, to not need to create them as they seem pretty similar and always the same (if the new is not the old, put the new as currrent value). It seems like it is even [already done for event based behaviors](https://github.com/4ian/GDevelop/blob/ab3856d38af6fe45d8dbd4ff2ff71e8fbf301e7d/GDJS/GDJS/Events/CodeGeneration/BehaviorCodeGenerator.cpp#L122)

4ian:
I agree it's "boring", "straightforward" code to write. We can do it automatically for events generated behaviors because we know the list of properties and the associated setter functions, while this mapping is not coded anywhere for usual, JS written object/behaviors. 
Let me know if you have a better idea, we could have a map of "object/behavior data property name" to "name of the setter to call with the new value" but I'm a bit worried of the implication for static typing (you rename a method and boom you just broke hot-reloading and TypeScript won't have any chance to warn you).

arthuro555:
Maybe properties could be declared in the Extension declaration file, with their getters and setters.

4ian:
That creates the same risk: if I rename a method, I run the risk of breaking hot reloading and won't be notified by anything unless I try to hot reload, at which point the game crash. Would be great if we add a solution where we can leverage typescript type checking (While everything is not typed yet, the plan is to add `// @ts-check the almost all JS files in the game engine/extensions`). Something that would say "hey you declared this name/setter in this map (the object doing the mapping between property name and setter) but this setter does not exist in the class - error!".

Seems like something that could *maybe* be expressed with `keyof`: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-1.html#keyof-and-lookup-types (not sure how this translates, and even if possible, in JSDoc annotations - hopefully this could work somehow).

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -390,39 +390,42 @@ gdjs.HowlerSoundManager.prototype.preloadAudio = function(
...
+  var totalCount = Object.keys(files).length;
+  if (totalCount === 0) return onComplete(totalCount); //Nothing to load.
 
-  var loaded = 0;
-  function onLoad(audioFile) {
-    loaded++;
-    if (loaded === files.length) {
-      return onComplete(files.length);
+  var loadedCount = 0;
+  var that = this;
```
Bouh, line 411: That is not used in this part.

Bouh:
I can't compile with the lastest commit, so i use the previous commit "Update icons and fix stale icons in Debugger Toolbar when selecting game …"

Here why i can't build, i got an error.
image


However there are some crashs in game with the hot reload.

- Anchor behavior, change the right edge anchor from Proportionnal to No anchor, reload => crash.

```
Uncaught (in promise) ReferenceError: behaviorData is not defined
    at gdjs.AnchorRuntimeBehavior.updateFromBehaviorData (anchorruntimebehavior.js:48)
    at gdjs.HotReloader._hotReloadRuntimeBehavior (hot-reloader.js:955)
    at hot-reloader.js:898
    at Array.forEach (<anonymous>)
    at hot-reloader.js:894
    at Array.forEach (<anonymous>)
    at gdjs.HotReloader._hotReloadRuntimeObjectsBehaviors (hot-reloader.js:870)
    at gdjs.HotReloader._hotReloadRuntimeSceneObject (hot-reloader.js:842)
    at hot-reloader.js:774
    at Array.forEach (<anonymous>)
```

- Text object, the word wrapping in the IDE is handle by the custom size, i've setup a small size, in the IDE the text is now display on 4 lines, reload => text is still on one line, same for BBText.

- Tiled sprite, i used the default width and height (32px), i change the value and click on the new button Apply change to preview, the Tiled sprite go back to the default dimension freshly setup,
If i setup a custom size of 300x300 with the default 32px for the tiled sprite and change the value 32, so the custom size willl use the new value.

- Behavior platformer, The player grab the edge, i disable the possibility for the player to grab them, reload => the player is still on the edge.

- Add the Y-Sort extension on an object, launch a preview, delete the extension in the project manager, reload => the behavior persist.
Go in the properties of the object where are the behavior, the editor crash.
[Here the report log.](https://pastebin.com/raw/5cU8HUFQ)




What i have tested?

- All values in properties panel.
- Add and delete all types of objects and their properties except for particules.
- Add an event rotation and color on different objects over time.
- Rename an object during the preview, the object is correctly updated in the debugger.
- Launch an preview, close and reopen the project, the light icon for hot reload come back and the hot reload work again, we can reload the preview :)

4ian:
Thanks for testing! Hopefully the compilation error is fixed - sorry about that I must have pushed something at the wrong time.
I'll check all your feedbacks in the next days to try to fix as much of them as possible :)

4ian:
I fixed the anchor behavior and bbtext/text objects width.

> Tiled sprite, i used the default width and height (32px), i change the value and click on the new button Apply change to preview, the Tiled sprite go back to the default dimension freshly setup,

Right, this would need to store a state in the object "is the object using a custom size or not". I'm thinking on giving this a pass, most of these objects are using a custom size, so we might not hot reload their original size.

> Behavior platformer, The player grab the edge, i disable the possibility for the player to grab them, reload => the player is still on the edge.

I tried this, this worked for me. Did you remove the edge grab on the platformer object or on the platform? If you remove it on the platform, this won't automatically make the player fall (I'm not sure there is an efficient way to do it, so we might pass on this).

> Add the Y-Sort extension on an object, launch a preview, delete the extension in the project manager, reload => the behavior persist.
> Go in the properties of the object where are the behavior, the editor crash.

I was not able to reproduce: I added the Y sort behavior to an object, launched the preview, deleted the extension. I went to the editor, it's still there (which is another problem by itself) but no crash. Can you verify if you can reproduce it 100% of times?

Bouh:
> I was not able to reproduce: I added the Y sort behavior to an object, launched the preview, deleted the extension. I went to the editor, it's still there (which is another problem by itself) but no crash. Can you verify if you can reproduce it 100% of times?

All is fine in with the hot reloader even if the behavior is deleted.
No crash in the game preview, only in the IDE, see my other Issue i have updated the first post.

>I tried this, this worked for me. Did you remove the edge grab on the platformer object or on the platform? If you remove it on the platform, this won't automatically make the player fall (I'm not sure there is an efficient way to do it, so we might pass on this).

Only tested on the Platform, but indeed the player fall when the option is disabled on the PlatformerObject.

4ian:
Almost done with this :) Have to fix the extension hot-reloading when using network preview and we should be good to go.
Great side effect of this: network preview will work with hot reload and the debugger (and profiler) => you'll be able to run the debugger and see your changes when running a preview on a phone/tablet/other computer :)

Bouh:
Yeah Amazing news for test on mobile over WiFi !

Le mer. 22 juil. 2020 à 01:28, Florian Rival <notifications@github.com> a
écrit :

4ian:
This looks stable and working enough, I'll go ahead and merge this!
This is still lacking support for adding a layer at the right position/re-ordering layer, but apart from that the rest should work :)

PR #1847, opened 2020-07-02
Add a way to see if a JS event is disabled

arthuro555:
It comments out the text wrapping the editor when the event is disabled, as there is no visual hint saying if it is disabled or not:

Enabled (like before):
image

Disabled:
image

Technically it is not correct JS but that's just visual so it is not very important (and I couldn't find any other way).

### DIFF
--- a/newIDE/app/src/EventsSheet/EventsTree/Renderers/JsCodeEvent.js
+++ b/newIDE/app/src/EventsSheet/EventsTree/Renderers/JsCodeEvent.js
@@ -134,13 +134,16 @@ export default class JsCodeEvent extends React.Component<
   render() {
     const jsCodeEvent = gd.asJsCodeEvent(this.props.event);
     const parameterObjects = jsCodeEvent.getParameterObjects();
+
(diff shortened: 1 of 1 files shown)


newIDE/app/src/EventsSheet/EventsTree/Renderers/JsCodeEvent.js
```diff
@@ -140,6 +140,7 @@ export default class JsCodeEvent extends React.Component<
           [selectableArea]: true,
         })}
         onClick={this.editObject}
+        style={this.props.disabled ? styles.comment : {}}
```
4ian, line 143: nit: can you factor all the styles into a variable: `const textStyle = this.props.disabled ? styles.comment : undefined`

(you don't need to pass an empty object, you can set the style to undefined if you have nothing to set).

4ian:
Sounds like a proper solution to me :) At least there is a visual indicator, and using a "fake" comment is a good idea :)

PR #1859, opened 2020-07-12
Fix findRStr and findStr misunderstood descriptions

### LINKED ISSUE #1857: findRStr and findStr misunderstood
## Describe the bug
According the wiki page StrFind return the position of an string in another string, from left to right.
StrRFind should do the same job but from the right to the left, but return the same value as StrFind when the research term is once in the string.

There is a misunderstood with these expression because StrRFind return the lastest occurence of the research and start the count from the left.
People understand that the expression will return the position from the right.

[Wiki decription:](http://wiki.compilgames.net/doku.php/gdevelop5/all-features/expressions-reference#expressions_reference)
`
StrRFind(string, string) = Search in a text from the end (return the position of the result or -1 if not found)
`

Correct case presently in gd:

```
var domain="www.test.com";
var lastdot= domain.lastIndexOf("t");  //return 7, from left to the right
```
How people think have the value:
```
var domain="www.test.com";
var lastdot= Math.abs(domain.lastIndexOf("t") - domain.length+1); //return 4, from right to left
// it's always the lastest 't' but from the end like it's explain on the wiki.
```

So what to do ?
I know if i tweak the expression findRStr, this will be broken for all previous games.
So maybe there is a simplier way to find the position from the end, or i can add a new expression for it, and we edit the description of the findRStr for explain clearly what is expected.

## Other details
b97

Bouh:
Copy and rename the old expressions and improve their descriptions.
Set deprecated and hidden the previous one.

Fix #1857

### DIFF
--- a/GDCpp/GDCpp/Extensions/Builtin/StringInstructionsExtension.cpp
+++ b/GDCpp/GDCpp/Extensions/Builtin/StringInstructionsExtension.cpp
@@ -46,11 +46,17 @@ StringInstructionsExtension::StringInstructionsExtension() {
   GetAllExpressions()["StrRFind"]
       .SetFunctionName("GDpriv::StringTools::StrRFind")
       .SetIncludeFile("GDCpp/Extensions/Builtin/StringTools.h");
+  GetAllExpressions()["StrFindLast"]
+      .SetFunctionName("GDpriv::StringTools::StrFindLast")
+      .SetIncludeFile("GDCpp/Extensions/Builtin/StringTools.h");
   GetAllExpressions()["StrFindFrom"]
       .SetFunctionName("GDpriv::StringTools::StrFindFrom")
       .SetIncludeFile("GDCpp/Extensions/Builtin/StringTools.h");
   GetAllExpressions()["StrRFindFrom"]
       .SetFunctionName("GDpriv::StringTools::StrRFindFrom")
+      .SetIncludeFile("GDCpp/Extensions/Builtin/StringTools.h");
+  GetAllExpressions()["StrFindLastFrom"]
+      .SetFunctionName("GDpriv::StringTools::StrFindLastFrom")
       .SetIncludeFile("GDCpp/Extensions/Builtin/StringTools.h");
 #endif
 }
--- a/GDCpp/GDCpp/Extensions/Builtin/StringTools.h
+++ b/GDCpp/GDCpp/Extensions/Builtin/StringTools.h
@@ -26,10 +26,14 @@ gd::String GD_API StrRepeat(const gd::String& str, std::size_t repCount);
 std::size_t GD_API StrLen(const gd::String& str);
 int GD_API StrFind(const gd::String& str, const gd::String& findwhat);
 int GD_API StrRFind(const gd::String& str, const gd::String& findwhat);
+int GD_API StrFindLast(const gd::String& str, const gd::String& findwhat);
 int GD_API StrFindFrom(const gd::String& str,
                        const gd::String& findwhat,
                        std::size_t start);
 int GD_API StrRFindFrom(const gd::String& str,
+                        const gd::String& findwhat,
+                        std::size_t start);
+int GD_API StrFindLastFrom(const gd::String& str,
                         const gd::String& findwhat,
                         std::size_t start);
 
--- a/GDJS/GDJS/Extensions/Builtin/StringInstructionsExtension.cpp
+++ b/GDJS/GDJS/Extensions/Builtin/StringInstructionsExtension.cpp
@@ -35,10 +35,14 @@ StringInstructionsExtension::StringInstructionsExtension() {
       "gdjs.evtTools.string.strFind");
   GetAllExpressions()["StrRFind"].SetFunctionName(
       "gdjs.evtTools.string.strRFind");
+  GetAllExpressions()["StrFindLast"].SetFunctionName(
+      "gdjs.evtTools.string.strFindLast");
   GetAllExpressions()["StrFindFrom"].SetFunctionName(
       "gdjs.evtTools.string.strFindFrom");
   GetAllExpressions()["StrRFindFrom"].SetFunctionName(
       "gdjs.evtTools.string.strRFindFrom");
+  GetAllExpressions()["StrFindLastFrom"].SetFunctionName(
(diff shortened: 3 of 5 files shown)


GDCpp/GDCpp/Extensions/Builtin/StringInstructionsExtension.cpp
```diff
@@ -46,11 +46,17 @@ StringInstructionsExtension::StringInstructionsExtension() {
   GetAllExpressions()["StrRFind"]
       .SetFunctionName("GDpriv::StringTools::StrRFind")
```
4ian, line 47: Let's change this function from StrRFind to StrFindLast

GDCpp/GDCpp/Extensions/Builtin/StringInstructionsExtension.cpp
```diff
@@ -46,11 +46,17 @@ StringInstructionsExtension::StringInstructionsExtension() {
...
       .SetIncludeFile("GDCpp/Extensions/Builtin/StringTools.h");
+  GetAllExpressions()["StrFindLast"]
+      .SetFunctionName("GDpriv::StringTools::StrFindLast")
+      .SetIncludeFile("GDCpp/Extensions/Builtin/StringTools.h");
   GetAllExpressions()["StrFindFrom"]
       .SetFunctionName("GDpriv::StringTools::StrFindFrom")
       .SetIncludeFile("GDCpp/Extensions/Builtin/StringTools.h");
   GetAllExpressions()["StrRFindFrom"]
       .SetFunctionName("GDpriv::StringTools::StrRFindFrom")
```
4ian, line 56: Same, StrRFindFrom => StrFindLastFrom

GDCpp/GDCpp/Extensions/Builtin/StringTools.h
```diff
@@ -26,12 +26,16 @@ gd::String GD_API StrRepeat(const gd::String& str, std::size_t repCount);
 std::size_t GD_API StrLen(const gd::String& str);
 int GD_API StrFind(const gd::String& str, const gd::String& findwhat);
 int GD_API StrRFind(const gd::String& str, const gd::String& findwhat);
+int GD_API StrFindLast(const gd::String& str, const gd::String& findwhat);
```
4ian, line 29: This is just adding the declaration of a function ;) Instead, let's rename StrRFind to StrFindLast (and some for the other functions), both here and in StringTools.cpp

GDJS/GDJS/Extensions/Builtin/StringInstructionsExtension.cpp
```diff
@@ -35,10 +35,14 @@ StringInstructionsExtension::StringInstructionsExtension() {
       "gdjs.evtTools.string.strFind");
   GetAllExpressions()["StrRFind"].SetFunctionName(
       "gdjs.evtTools.string.strRFind");
```
4ian, line 37: Let's use `gdjs.evtTools.string.strFindLast`

GDJS/GDJS/Extensions/Builtin/StringInstructionsExtension.cpp
```diff
@@ -35,10 +35,14 @@ StringInstructionsExtension::StringInstructionsExtension() {
...
+  GetAllExpressions()["StrFindLast"].SetFunctionName(
+      "gdjs.evtTools.string.strFindLast");
   GetAllExpressions()["StrFindFrom"].SetFunctionName(
       "gdjs.evtTools.string.strFindFrom");
   GetAllExpressions()["StrRFindFrom"].SetFunctionName(
       "gdjs.evtTools.string.strRFindFrom");
```
4ian, line 43: Same as before, let's change this to `gdjs.evtTools.string.strFindLastFrom`

GDJS/Runtime/events-tools/stringtools.js
```diff
@@ -88,33 +88,51 @@ gdjs.evtTools.string.strLen = function(str) {
...
 };
 
 /**
- * Reverse search in a string
+ * Search the last occurence in a string (return the position of the result, from the beginning of the string, or -1 if not found)
  * @private
+ * @deprecated
  */
 gdjs.evtTools.string.strRFind = function(str, what) {
     return str.lastIndexOf(what);
```
4ian, line 104: As it's deprecated, let's implement it in terms of the not deprecated method:
`return gdjs.evtTools.string.strFindLast(str, what);`

GDJS/Runtime/events-tools/stringtools.js
```diff
@@ -88,33 +88,51 @@ gdjs.evtTools.string.strLen = function(str) {
...
 };
 
 /**
- * Reverse search in a string, starting from a specified position.
+ * Search the last occurence in a string, starting from a specified position (return the position of the result, from the beginning of the string, or -1 if not found)
  * @private
+ * @deprecated
  */
 gdjs.evtTools.string.strRFindFrom = function(str, what, pos) {
     return str.lastIndexOf(what, pos);
```
4ian, line 129: Same, let's call `gdjs.evtTools.string.strFindLastFrom` ;)

4ian:
Thanks 👍 In this situation I think it's a good idea to just rename the JS functions (games should not use them anyway - they are not part of the public interface, it's a detail of the implementation) and use it for both the new expression and the old one :)
So that existing games using them will continue to work, and new games will use the not hidden expression.

GDJS/Runtime/events-tools/stringtools.js
4ian: Actually, we can just do: `gdjs.evtTools.string.strRFindFrom = gdjs.evtTools.string.strFindLastFrom`. Very clear and concise!

GDJS/Runtime/events-tools/stringtools.js
4ian: Actually, `gdjs.evtTools.string.strRFind = gdjs.evtTools.string.strFindLast;` will do the trick ;)

4ian:
I went ahead and implemented the comment fixes by myself ;) Thanks for the PR!

Bouh:
Yes no worries i was just impossible to me to do the fixes during my journey, thank you!

PR #1864, opened 2020-07-15
Add more commands to the command palette

NilayMajorwar:
This PR adds a whole lot of pretty nice commands to the command palette. Here's a quick view, for example:

compound

A summary of the main commands added (I may have missed some):
- **Tab-related**: Open scene, Open external events, Open external layout, Open extension
- **Scene editor**: Edit object, Edit object variables, Edit object group, Edit layer effects, Open scene properties, Open scene variables, Toggle grid, Toggle mask, Setup grid and some panel-opening commands like Open properties panel, Open layers panel, etc.
- **Events editor**: Create empty event, Add event of type, Search events, Add sub-event, Delete selected events
- **Project-related**: Open project properties, Edit global variables, Open project resources, Open project icons
- **General**: Open recent project


NilayMajorwar:
@4ian I've just added a few more commands (Toggle grid, Toggle mask, Setup grid) in the scene editor toolbar - these ones were completely straightforward. Please review the PR :grinning:

newIDE/app/src/CommandPalette/CommandHooks.js
```diff
@@ -50,3 +80,27 @@ export const useKeyboardShortcutForCommandPalette = (onOpen: () => void) => {
...
+  return null;
+};
+
+/**
+ * React component for using useCommandWithOptions
+ * hook in class components
+ */
+export const UseCommandWithOptionsHook = (props: {
+  commandName: string,
+  command: CommandWithOptions<*>,
```
4ian, line 102: * is deprecated (see https://flow.org/en/docs/linting/rule-reference/#toc-deprecated-type), I recommend here either:
* typing generically the function component: `... = <OptionType>(props: { commandName: string, command: CommandWithOptions<OptionType>}) => { ... };`
* or just use any

newIDE/app/src/CommandPalette/CommandHooks.js
```diff
@@ -50,3 +80,27 @@ export const useKeyboardShortcutForCommandPalette = (onOpen: () => void) => {
     [onOpen, values.useCommandPalette]
   );
 };
+
+/**
+ * React component for using useCommand hook in
+ * class components
+ */
+export const UseCommandHook = (props: {
```
4ian, line 88: {| |} here and below

newIDE/app/src/CommandPalette/CommandManager.js
```diff
@@ -3,21 +3,49 @@ import { type MessageDescriptor } from '../Utils/i18n/MessageDescriptor.flow';
...
+  iconSrc?: string,
+|};
+
+export type CommandWithOptions<T> = {|
+  displayText: MessageDescriptor,
+  enabled: boolean,
+  generateOptions: () => Array<CommandOption<T>>,
+|};
+
+export type Command = SimpleCommand | CommandWithOptions<*>;
```
4ian, line 25: Use `any`

newIDE/app/src/CommandPalette/CommandManager.js
```diff
@@ -3,21 +3,49 @@ import { type MessageDescriptor } from '../Utils/i18n/MessageDescriptor.flow';
...
 |};
 
-export default class CommandManager {
+export type NamedCommandWithOptions<T> = {|
+  name: string,
+  ...CommandWithOptions<T>,
+|};
+
+export interface CommandManagerInterface {
+  commands: { [string]: Command };
```
4ian, line 38: Are you forced to include commands in the interface? Typically we don't want to expose this to the client code.

newIDE/app/src/CommandPalette/CommandPalette/AutocompletePicker.js
```diff
@@ -0,0 +1,78 @@
+// @flow
+import * as React from 'react';
+import { type I18n } from '@lingui/core';
```
4ian, line 3: Use `import { type I18n as I18nType } from '@lingui/core';` for consistency with the rest of the codebase

newIDE/app/src/CommandPalette/CommandsScopedContext.js
```diff
@@ -0,0 +1,91 @@
...
+
+  getAllNamedCommands = () => {
+    return Object.keys(this.commands).map<NamedCommand>(commandName => {
+      const cmd = this.commands[commandName];
+      return { ...cmd, name: commandName };
+    });
+  };
+}
+
+type Props = {
```
4ian, line 60: {| |}

newIDE/app/src/CommandPalette/CommandsScopedContext.js
```diff
@@ -0,0 +1,91 @@
...
+  type Command,
+  type NamedCommand,
+} from './CommandManager';
+import CommandsContext from './CommandsContext';
+import useRefInit from './UseRefInitHook';
+
+class ScopedCommandManager implements CommandManagerInterface {
+  commands: { [string]: Command };
+  _centralManager: CommandManagerInterface;
+  isActive: boolean;
```
4ian, line 14: Do we want this to be publicly accessible? If not, let's name it `_isActive`.

newIDE/app/src/CommandPalette/UseRefInitHook.js
```diff
@@ -0,0 +1,15 @@
+// @flow
+import { useRef } from 'react';
+
+const useRefInit = <T>(init: () => T): T => {
+  const instanceRef = useRef(null);
+
+  let instance = instanceRef.current;
+  if (instance !== null) return instance;
+  // Lazy init
```
4ian, line 9: Nit: most of the time, write full sentence in comments:

`// Lazily create the ref object.`

newIDE/app/src/EventsSheet/ToolbarCommands.js
```diff
@@ -0,0 +1,107 @@
...
+const addSubEventCommandText = t`Add a sub-event to the selected event`;
+const addCommentCommandText = t`Add a comment`;
+const chooseAndAddEventCommandText = t`Choose and add an event...`;
+const undoCommandText = t`Undo the last changes`;
+const redoCommandText = t`Redo the last changes`;
+const deleteSelectionCommandText = t`Delete the selected event(s)`;
+const searchEventsCommandText = t`Search in events`;
+const openSettingsCommandText = t`Open settings`;
+
+type Props = {
```
4ian, line 20: {| |}

newIDE/app/src/LayersList/UseLayersListCommands.js
```diff
@@ -0,0 +1,37 @@
+// @flow
+import * as React from 'react';
+import { t } from '@lingui/macro';
+import { mapReverseFor } from '../Utils/MapFor';
+import { useCommandWithOptions } from '../CommandPalette/CommandHooks';
+
+const editLayerEffectsCommandText = t`Edit layer effects...`;
+
+type Props = {
```
4ian, line 9: {| |}

newIDE/app/src/ObjectGroupsList/UseObjectGroupsListCommands.js
```diff
@@ -0,0 +1,36 @@
+// @flow
+import * as React from 'react';
+import { t } from '@lingui/macro';
+import { enumerateGroups } from '../ObjectsList/EnumerateObjects';
+import { useCommandWithOptions } from '../CommandPalette/CommandHooks';
+
+const editObjectGroupCommandText = t`Edit object group...`;
+
+type Props = {
```
4ian, line 9: {| |}

newIDE/app/src/ObjectsList/UseObjectsListCommands.js
```diff
@@ -0,0 +1,57 @@
...
+    value: item.object,
+    handler: () => onChoose(item.object),
+    iconSrc: ObjectsRenderingService.getThumbnail.bind(ObjectsRenderingService)(
+      project,
+      item.object
+    ),
+  }));
+};
+
+type Props = {
```
4ian, line 28: {| |}

newIDE/app/src/ProjectManager/ProjectManagerCommands.js
```diff
@@ -0,0 +1,46 @@
+// @flow
+import { t } from '@lingui/macro';
+import { useCommand } from '../CommandPalette/CommandHooks';
+
+const openProjectPropertiesCommandText = t`Open project properties`;
+const openProjectVariablesCommandText = t`Edit global variables`;
+const openPlatformSpecificAssetsCommandText = t`Open project icons`;
+const openResourcesCommandText = t`Open project resources`;
+
+type Props = {
```
4ian, line 10: {| |}

newIDE/app/src/SceneEditor/UseSceneEditorCommands.js
```diff
@@ -0,0 +1,67 @@
...
+import { t } from '@lingui/macro';
+import { useCommand } from '../CommandPalette/CommandHooks';
+import useObjectsListCommands from '../ObjectsList/UseObjectsListCommands';
+import useObjectGroupsListCommands from '../ObjectGroupsList/UseObjectGroupsListCommands';
+import useLayersListCommands from '../LayersList/UseLayersListCommands';
+
+const openScenePropertiesCommandText = t`Open scene properties`;
+const openSceneVariablesCommandText = t`Open scene variables`;
+
+type Props = {
```
4ian, line 11: {| |}

newIDE/app/src/SceneEditor/index.js
```diff
@@ -916,6 +924,9 @@ export default class SceneEditor extends React.Component<Props, State> {
             resourceSources={resourceSources}
             resourceExternalEditors={resourceExternalEditors}
             onChooseResource={onChooseResource}
+            onEditLayerEffects={layer =>
```
4ian, line 927: this.editLayerEffects

NilayMajorwar:
I've made the fixes, and a couple of other ones too, please have a look.

newIDE/app/src/CommandPalette/CommandManager.js
```diff
@@ -3,18 +3,44 @@ import { type MessageDescriptor } from '../Utils/i18n/MessageDescriptor.flow';
...
 
-export type Command = {|
+export type SimpleCommand = {|
   displayText: MessageDescriptor,
   enabled: boolean,
   handler: CommandHandler,
 |};
 
+export type CommandOption<T> = {|
+  value: T,
```
4ian, line 13: I just realized that this value seems used... nowhere? :D 
Which is understandable, the only thing we need for an option is its `text`, it's `iconSrc` and the function to run, knowing that one `handler` is created for each option, so the closure is already embedding the information of the value.

Unless I missed something, I recommend removing the value here (and so you can remove a bunch of <T> and <OptionType> everywhere 👍) and then everywhere where you were passing a value.

4ian:
Looks good! Seems almost good to go 👍 Added a comment about simplifying CommandWithOptions.
Last thing, could you take a look at making 1 or 2 stories in `newIDE/app/src/stories/index.js`? Run storybook with `npm run storybook`/`yarn storybook`. Open `localhost:9009` to see all the existing stories. 
I think it's good if we can add a story for `AutocompletePicker`, with a few fake items that you set in the story. Will be very useful as a quick test case to ensure that there is no regression later, and allow you/me/other developers to quickly check how this component works when developing it :)

newIDE/app/src/CommandPalette/CommandManager.js
NilayMajorwar: Ah yes, I initially put that in as a unique identifier for each option. That was when I was thinking about goto anything, which would have required diffing and comparing options and all sorts of things - and that would have required an explicit unique identifier. But that was eventually scrapped, so yes right now it's unused. I'll remove it, then.

Now that you mention it, the `enabled` prop in Command and its variants is also somewhat unused, since commands that reach the command manager (or scoped command manager) are actually all enabled. `enabled` is really somewhat better suited as a direct hook argument, similar to `commandName`. Should I fix that?

NilayMajorwar:
I've added two stories under a new category CommandPalette. One story is for the command picker and the other is for option picker. From the UI perspective though, the two stories are a bit redundant - the main difference is that the option picker has icons in the options. Also, I'm not sure where to put the fake options, I've just put them inline in the component's story for now.

Also, I've made a small change so that now when a user selects a command with options, the placeholder of the options picker shows the command text (like "Edit object...") instead of the plain "Pick an options...". Small change, but feels a bit better now.

4ian:
> From the UI perspective though, the two stories are a bit redundant - the main difference is that the option picker has icons in the options

This is fine, while it's a bit redundant it's better for completeness.

> Also, I'm not sure where to put the fake options, I've just put them inline in the component's story for now.

That's actually the preferred way of doing it :) So that stories can be understood without opening tons of other files, it makes sense to have everything inline (unless you need very big/complex test data, like the test project).

newIDE/app/src/CommandPalette/CommandManager.js
4ian: > Now that you mention it, the enabled prop in Command (and its related types) is also somewhat unnecessary, since commands that reach the command manager (or scoped command manager) are actually all enabled. enabled is really somewhat better suited as a direct hook argument, similar to commandName. Should I fix that?

Yes I'm all for removing it if it's not actually useful :)

newIDE/app/src/stories/index.js
```diff
@@ -4270,3 +4275,74 @@ storiesOf('NewObjectDialog', module)
...
+              name: 'EDIT_OBJECT',
+              displayText: t`Edit object...`,
+              enabled: true,
+              handler: () => {},
+            },
+          ]}
+          onClose={() => {}}
+          onSelect={(item: NamedCommand) =>
+            action(`Open command ${item.name}`)()
+          }
```
4ian, line 4309: nit: you could even do just `onSelect={action("Open command")}` because the argument will be logged by Storybook :)

NilayMajorwar: Oh okay, that's nice :D. Though, I had to add a type annotation to the items array, otherwise Flow wasn't able to infer whether the array was one of commands or command options.

newIDE/app/src/CommandPalette/CommandManager.js
NilayMajorwar: Okay, done!

4ian:
Alright looks good I think! It's a large PR so if I missed something feel free to open a new PR with a fix.

Merging now :) Thanks a lot for implementing all these commands and setting up the "scoped" commands provider and all the smaller details. Looks clean and maintainable, and I'm sure users will love that most parts of the app are now covered by commands! Huge productivity boost, especially once we'll have shortcuts 👍👍

Great job ☺️

PR #1873, opened 2020-07-17
implementing shades of purple theme

TheGemDev:
Implementing shades of purple theme for gdevelop 5.
shade1
shade2
shade3
shade4
shade5
shade6
shade7
shade8


newIDE/app/src/UI/Theme/ShadesofPurpleTheme/EventsSheet.css
```diff
@@ -0,0 +1,218 @@
+/**
+ * This is the styling applied to an EventsSheet.
+ * This also overwrite the default react-sortable-tree css.
+ */
+
+ .gd-events-sheet-Shades-of-Purple-theme.events-tree {
```
Bouh, line 6:
```suggestion
 .gd-events-sheet-shades-of-Purple-theme.events-tree {
```

Bouh: This can be done everywhere.

newIDE/app/src/UI/Theme/ShadesofPurpleTheme/EventsSheet.css
```diff
@@ -0,0 +1,218 @@
...
+}
+
+/**
+ * Scaffolding lines on the left
+ */
+.gd-events-sheet-Shades-of-Purple-theme .rst__lineHalfHorizontalRight::before,
+.gd-events-sheet-Shades-of-Purple-theme .rst__lineFullVertical::after,
+.gd-events-sheet-Shades-of-Purple-theme .rst__lineHalfVerticalTop::after,
+.gd-events-sheet-Shades-of-Purple-theme .rst__lineHalfVerticalBottom::after {
+  background-color: #9999E9;
```
Bouh, line 42: I don't see this color in the theme.
https://github.com/ahmadawais/shades-of-purple-vscode#sops-syntax-colors

Where is it used in GDevelop?

Bouh:
We could keep the GDevelop color codes in the event sheet (object, expression, behavior,string,number), while adjusting them slightly to match the colors of the theme you have chosen.

I find that the event sheet is bland compared to the rest. It lacks pep.

image

image

4ian:
This theme would be interesting to have :) But I think we should do the refactoring for having themes defined using colors only: https://github.com/4ian/GDevelop/issues/1918
Otherwise I have the feeling it's getting harder and harder to maintain themes :/

the-gem-dev:
> This theme would be interesting to have :) But I think we should do the refactoring for having themes defined using colors only: #1918
> Otherwise I have the feeling it's getting harder and harder to maintain themes :/

Thats exactly why i paused my work on this theme.

4ian:
This theme was pretty nice! @TheGemDev would you want to now convert it to the new theme format? It should be not too hard to do, see the explanations in https://github.com/4ian/GDevelop/blob/master/newIDE/README-themes.md :)

TheGemDev:
> This theme was pretty nice! @TheGemDev would you want to now convert it to the new theme format? It should be not too hard to do, see the explanations in https://github.com/4ian/GDevelop/blob/master/newIDE/README-themes.md :)

okay I would work it again :)

4ian:
Might be as easy and even faster to start from a new PR?

TheGemDev:
> Might be as easy and even faster to start from a new PR?

i guess 👍🏾 :

PR #1881, opened 2020-07-20
Lighting

### LINKED ISSUE #1883: karma test command doesn't work in windows
https://github.com/4ian/GDevelop/blob/661d32917002dd4f2852c67b22f62c294e99bd8d/GDJS/tests/package.json#L8-L9

These don't work on windows because `./something/app` results in `'.' is not recognized as an internal or external command,
operable program or batch file.` due to differences of backslash/frontslash. 

Maybe a more inclusive command could be `"test": "cd ./node_modules/karma/bin && karma start ../../../karma.conf.js --browsers ChromeHeadless --single-run"` (and similar for firefox) as this should work in all environments 😅

HarsimranVirk:
Alright, @4ian I have got this branch running on top of master (instead of pixi-v5). This is cherry-picked from the previous branch with some changes related to `LayersList`. Also, notifying @NilayMajorwar about a new command I've added in `UseLayersListCommands` which uses the `LightingLayerDialog` to edit properties of lighting layer :)

(Also resolves #1883)


newIDE/app/src/SceneEditor/index.js
```diff
@@ -887,6 +896,10 @@ export default class SceneEditor extends React.Component<Props, State> {
     if (this._objectsList) this._objectsList.forceUpdateList();
   };
 
+  forceUpdateLayersList = () => {
+    if (this._layersList) this._layersList.onLayerModified();
+  };
+
```
HarsimranVirk, line 902: @4ian I'm not sure if this is good or not :/ I believe the `LayersList` wasn't updating, after adding and removing the layer twice, it was in a state where there was no lighting layer and the button was disabled as well. Now, force updating does solve the problem but I'm not sure if it's the right way :/

newIDE/app/src/SceneEditor/index.js
```diff
@@ -836,6 +896,10 @@ export default class SceneEditor extends React.Component<Props, State> {
     if (this._objectsList) this._objectsList.forceUpdateList();
   };
 
+  forceUpdateLayersList = () => {
+    if (this._layersList) this._layersList.onLayerModified();
```
4ian, line 900: I would recommend avoiding "hijacking" `onLayerModified` because: 
- it's called "onXXX" so usually seen as a method that the component will call when things have been modified "inside" the component rather than outside.
- it's triggering potentially other stuff than just force updating.

Instead, just call `forceUpdate` on layersList. The name is fine, and it's totally understandable that because we modified the layers list outside, we now must force the update of the layers editor.

newIDE/app/src/SceneEditor/index.js
4ian: Let's put back `_onLayerModified` "private" and use `forceUpdate` instead.

Force updating these components showing virtualized list is fine - maybe in the future we'll rework them to avoid this.

newIDE/app/src/LayersList/index.js
```diff
@@ -126,17 +135,40 @@ type State = {|
   effectsEditedLayer: ?gdLayer,
 |};
 
+const hasLightingLayer = (layout: gdLayout) => {
+  const layersCount = layout.getLayersCount();
+  let isLightingLayerPresent = false;
+  mapReverseFor(0, layersCount, i => {
+    const layer = layout.getLayerAt(i);
+    if (layer.isLightingLayer()) isLightingLayerPresent = true;
```
4ian, line 143: nit: `if (xxx) return true`

newIDE/app/src/LayersList/index.js
```diff
@@ -126,17 +135,40 @@ type State = {|
   effectsEditedLayer: ?gdLayer,
 |};
 
+const hasLightingLayer = (layout: gdLayout) => {
+  const layersCount = layout.getLayersCount();
+  let isLightingLayerPresent = false;
```
4ian, line 140: nit: no need for this variable, just return true/false

newIDE/app/src/LayersList/index.js
HarsimranVirk: @4ian I'm not so sure about it, `mapReverseFor` basically returns a list, right? I don't really want a list of booleans here, just a confirmation that a lighting layer exists (or not). Am I missing something?

HarsimranVirk: Or I could do something like this, 
`return mapReverseFor(0, layersCount, i => {`
`  const layer = layout.getLayerAt(i);`
`  if (layer.isLightingLayer()) return true;`
`  return false;`
` }).reduce((acc, value) => acc || value);`

Maybe it's an overkill, but the intent could be expressed both ways, either this or a variable 🤔

4ian: ah my bad, of course returning in the closure won't work :) 

Then you can do something like `mapReverseFor(0, layersCount, i => layout.getLayerAt(i).isLightingLayer()).filter(Boolean).length > 0` which is a bit shorter to express the logic of "check if there is already at least one layer that is a lighting layer".

HarsimranVirk:
@4ian, I guess we should be release-ready now, ie, at least in terms of features. Let's look out for regressions/crashes until the next release, since this PR introduces a lot of changes to game engine and IDE as well. :)

4ian:
Yep let me do a full review and test this PR and the goal will be to merge this then :) I hope to do a release soon so that we can add this to the next release after that - giving us a bit more time looking for any problem/regression.

Extensions/Lighting/lightruntimeobject.js
```diff
@@ -0,0 +1,93 @@
...
+  this._debugMode = lightObjectData.content.debugMode;
+  this._texture =
+    lightObjectData.content.texture === ''
+      ? null
+      : runtimeScene
+          .getGame()
+          .getImageManager()
+          .getPIXITexture(lightObjectData.content.texture);
+
+  if (gdjs.LightObstaclesManager)
```
4ian, line 18: Is there even a reason for gdjs.LightObstaclesManager not to be defined? I don't think so? We should always include I think the behavior with the object in the include files.

Extensions/Lighting/lightruntimeobject.js
```diff
@@ -0,0 +1,93 @@
...
+    this._obstacleManager = gdjs.LightObstaclesManager.getManager(runtimeScene);
```
4ian, line 19: Don't declare your properties conditionnaly. Always declare them. It's the *value* that can change:

`this._obstaclesManager = condition ? something : null`

In other words, avoid things that are implicitly undefined at all cost.

4ian: Also, nitpicking: `obstaclesManager`, to be consistent with the class name.

GDJS/Runtime/layer.js
```diff
@@ -396,21 +407,44 @@ gdjs.Layer.prototype._setEffectsDefaultParameters = function() {
...
  * in milliseconds, for objects on the layer.
  */
-gdjs.Layer.prototype.getElapsedTime = function() {
+gdjs.Layer.prototype.getElapsedTime = function () {
   return this._runtimeScene.getTimeManager().getElapsedTime() * this._timeScale;
 };
+
+/**
+ * Set the position of layer's camera to follow
+ * base layer's camera.
```
4ian, line 431: Change the position and rotation of the layer camera to be the same as the base layer camera.

GDJS/Runtime/layer.js
```diff
@@ -396,21 +407,44 @@ gdjs.Layer.prototype._setEffectsDefaultParameters = function() {
...
+ */
+gdjs.Layer.prototype.followBaseLayer = function () {
+  this.setCameraX(this._runtimeScene.getLayer('').getCameraX());
+  this.setCameraY(this._runtimeScene.getLayer('').getCameraY());
+};
```
4ian, line 436: Store `this._runtimeScene.getLayer('')` in a variable.
Also update the camera rotation

GDJS/Runtime/layer.js
```diff
@@ -396,21 +407,44 @@ gdjs.Layer.prototype._setEffectsDefaultParameters = function() {
...
+gdjs.Layer.prototype.getClearColor = function () {
+  return this._clearColor;
+};
+
+/**
+ * @return {boolean} true if it is a lighting layer, false otherwise.
+ */
+gdjs.Layer.prototype.isLightingLayer = function () {
+  return this._isLightingLayer;
+}
```
4ian, line 450: nit: configure your editor to always add a line break at the end of the file ("posix" good practice)

GDJS/Runtime/layer.js
```diff
@@ -396,21 +407,44 @@ gdjs.Layer.prototype._setEffectsDefaultParameters = function() {
...
+ * Set the position of layer's camera to follow
+ * base layer's camera.
+ */
+gdjs.Layer.prototype.followBaseLayer = function () {
+  this.setCameraX(this._runtimeScene.getLayer('').getCameraX());
+  this.setCameraY(this._runtimeScene.getLayer('').getCameraY());
+};
+
+/**
+ * @return {number[]} the clear color of layer in the range of [0, 1].
```
4ian, line 439: Always include a description to your documented function:

```
One line explanation
@returns {number[]} probably the same explanation
```

Also explicit the format [r,g,b,a], with components in range [0, 1], because it's not a format used often in the rest of the codebase.

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
```diff
@@ -54,34 +62,44 @@ gdjs.LayerPixiRenderer.prototype.updatePosition = function() {
...
+
+  for (var filterName in this._filters) {
     var filter = this._filters[filterName];
     filter.update(filter.pixiFilter, this._layer);
   }
+
+  // Add Layer to lighting after adding effects to the layer,
+  // so that we can fix the blend mode.
+  if(!this._lightingSprite && this._isLightingLayer) {
+    this.replaceContainerWithSprite();
```
4ian, line 82: Could we avoid doing this check at every update?

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
```diff
@@ -54,34 +62,44 @@ gdjs.LayerPixiRenderer.prototype.updatePosition = function() {
...
 };
 
-gdjs.LayerPixiRenderer.prototype.updateVisibility = function(visible) {
+gdjs.LayerPixiRenderer.prototype.updateVisibility = function (visible) {
   this._pixiContainer.visible = !!visible;
 };
 
-gdjs.LayerPixiRenderer.prototype.updateTime = function() {
-  for(var filterName in this._filters) {
+gdjs.LayerPixiRenderer.prototype.updateTime = function () {
```
4ian, line 69: If this is doing something else than updating the time, we should probably rename this to `update` (also in the Cocos files)

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
```diff
@@ -244,9 +266,107 @@ gdjs.LayerPixiRenderer.prototype.enableEffect = function(name, value) {
...
 
   return gdjs.PixiFiltersTools.isEffectEnabled(filter);
 };
+
+/**
+ * Updates the render texture, if it exists. 
+ * Also, render texture is cleared with a specified clear color.  
+ * Not to be called outside of updateTime method in most cases.
+ */
+gdjs.LayerPixiRenderer.prototype.updateRenderTexture = function () {
```
4ian, line 281: Make this private with an _

Extensions/Lighting/JsExtension.js
```diff
@@ -0,0 +1,351 @@
+module.exports = {
```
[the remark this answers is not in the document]
HarsimranVirk: @4ian how do I run flow for files in `Extention`? Is it included in the `newIDE/app`'s `npm run flow` as well?

Extensions/Lighting/JsExtension.js
```diff
@@ -0,0 +1,351 @@
...
+     */
+    RenderedLightObjectInstance.prototype.update = function () {
+      this._pixiObject.shader.uniforms.center = new Float32Array([
+        this._instance.getX(),
+        this._instance.getY(),
+      ]);
+      this._pixiObject.geometry
+        .getBuffer('aVertexPosition')
+        .update(
+          new Float32Array([
```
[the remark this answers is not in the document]
HarsimranVirk: Yes, there needs to be only one instance of Float32Array (of size 8) in this case since there's no raycasting involved :) I'll make changes for it.

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
HarsimranVirk: Should I annotate it with a `@private` as well?

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
HarsimranVirk: Yes, it needs to be called after the filters are applied (because we need to change the blend mode of filters) and after all of the layers are added to the scene. So, maybe in the constructor,
https://github.com/4ian/GDevelop/blob/693b64cddfa8c3590a36a857e6f8a544d5fe653e/GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js#L22-L24
after `this._setupFilters()`?

Extensions/Lighting/lightruntimeobject.js
HarsimranVirk: > Is there even a reason for gdjs.LightObstaclesManager not to be defined

I'm not sure why this happens but whenever there's no object with the obstacle behavior, the game crashes saying gdjs.LightObstaclesManager is undefined.

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
HarsimranVirk: Works in the constructor after `this._setupFilters()` :)

GDJS/Runtime/layer.js
HarsimranVirk: I should be updating the zoom as well then, right? Differences in the scale could be a problem too I guess.

4ian: Yep indeed!

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
4ian: I filters are added/removed (which can be the case with hot reloading), it might need to be updated... let's add a TODO in the code: `// TODO: Handle update of filters`.

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
4ian: Yes you can!

HarsimranVirk:
> I've added comments on the first files (mostly the game engine object/behavior), I'll continue later with the update to the layer pixi renderer and the IDE changes.

I've made almost all the changes you've mentioned so far, but I'll be pushing everything once you're done reviewing the entire thing (so that it's easier for you to continue from where you left, and not see newer changes 😅).

> add // @ts-check to the light object runtime files (object and behavior)

I've added documentations (with as much typing as I could) but I guess currently `// @ts-check` doesn't work very well in the `Extension` because there are issues with `@extends` keyword (ie, it only works with es6 classes) and therefore it throws a lot of errors. For eg, XYZRuntimeObject requires gdjs.RuntimeObject, but `@extends` doesn't work, and throws error regarding properties defined in RuntimeObject. I'd say this is a general problem right now, whenever we'll be fixing it everything should work correctly :)

Extensions/Lighting/JsExtension.js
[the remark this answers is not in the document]
arthuro555: > Is it included in the newIDE/app's npm run flow as well?

Yes it is

newIDE/app/src/Hints/ObjectsAdditionalWork.js
```diff
@@ -0,0 +1,121 @@
...
+import newNameGenerator from '../Utils/NewNameGenerator';
+
+/*
+ * Define additional logic which executes after an object/instance has been created.
+ * Also, InfoBar can be used which could notify users of the additional changes.
+ * Declare new identifier for infoBar in Mainframe/Preferences/PreferenceContext
+ * and add it in hints/explaination list.
+ */
+
+export type InfoBarDetails = {
```
4ian, line 13: Always use strict types (`{| ... |}`) in Flow

newIDE/app/src/Hints/ObjectsAdditionalWork.js
```diff
@@ -0,0 +1,121 @@
...
+    ];
+    if (services) {
+      services.onInstanceAdded(instance, layout, project);
+      return services.getInfoBarDetails('onInstanceAdded');
+    }
+
+    return null;
+  },
+
+  objectType: {
```
4ian, line 53: nit: You can probably move this out of the export

newIDE/app/src/Hints/ObjectsAdditionalWork.js
```diff
@@ -0,0 +1,121 @@
...
+  show: boolean,
+  identifier: AlertMessageIdentifier,
+  message: string,
+  touchScreenMessage: string,
+};
+
+type InfoBarEvent = 'onObjectAdded' | 'onInstanceAdded';
+
+export default {
+  onObjectAdded(
```
4ian, line 23: nit: Consider using `export const onObjectAdded = ...` - this favors the use of functions rather than an object, so it helps static analysis and in the future can help bundlers like Rollup to do "Tree shaking". Seems like named exports (`export const ...`/`export class ...`) are becoming preferred to default exports for these reasons in the JS community

newIDE/app/src/Hints/ObjectsAdditionalWork.js
```diff
@@ -0,0 +1,121 @@
...
+
+/*
+ * Define additional logic which executes after an object/instance has been created.
+ * Also, InfoBar can be used which could notify users of the additional changes.
+ * Declare new identifier for infoBar in Mainframe/Preferences/PreferenceContext
+ * and add it in hints/explaination list.
+ */
+
+export type InfoBarDetails = {
+  show: boolean,
```
4ian, line 14: Should show even be in this object type? It's a responsibility of the editor I think.

newIDE/app/src/Hints/ObjectsAdditionalWork.js
```diff
@@ -0,0 +1,121 @@
...
+  identifier: AlertMessageIdentifier,
+  message: string,
+  touchScreenMessage: string,
```
4ian, line 17: Typing is incorrect, use `MessageDescriptor` type.

newIDE/app/src/Hints/ObjectsAdditionalWork.js
```diff
@@ -0,0 +1,121 @@
...
+        project: gdProject
+      ) => {
+        let hasLightingLayer = false;
+        for (let i = 0; i < layout.getLayersCount(); i++) {
+          const layer = layout.getLayerAt(i);
+          if (layer.isLightingLayer()) {
+            hasLightingLayer = true;
+            break;
+          }
+        }
```
4ian, line 67: Can you extract it in a function in this file?

newIDE/app/src/InstancesEditor/InstancesAdder.js
```diff
@@ -120,6 +125,8 @@ export default class InstancesAdder {
       instance.setX(Math.round(newPos[0]));
       instance.setY(Math.round(newPos[1]));
     });
+
+    return this._temporaryInstances;
```
4ian, line 129: Let's add a typing for the return value to the function declaration too.

newIDE/app/src/Hints/ObjectsAdditionalWork.js
```diff
@@ -0,0 +1,121 @@
...
+        project: gdProject
+      ) => {
+        let lightingLayer = null;
+        for (let i = 0; i < layout.getLayersCount(); i++) {
+          const layer = layout.getLayerAt(i);
+          if (layer.isLightingLayer()) {
+            lightingLayer = layer;
+            break;
+          }
+        }
```
4ian, line 92: Make a single function `getLightningLayer(layersContainer)` and use it here and in the previous function (if null, then there is no lighting layer)

GDJS/Runtime/pixi-renderers/runtimescene-pixi-renderer.js
```diff
@@ -85,3 +85,11 @@ gdjs.RuntimeScenePixiRenderer.prototype.showCursor = function() {
 gdjs.RuntimeScenePixiRenderer.prototype.getPIXIContainer = function() {
   return this._pixiContainer;
 };
+
+gdjs.RuntimeScenePixiRenderer.prototype.getPIXIRenderer = function() {
+  return this._pixiRenderer;
+}
+
+gdjs.RuntimeScenePixiRenderer.prototype.getScene = function () {
```
4ian, line 93: Interesting, why would you need `getScene`? I assume that code calling this would already know the scene?

newIDE/app/src/LayersList/LightingLayerDialog.js
```diff
@@ -0,0 +1,96 @@
...
+import FlatButton from '../UI/FlatButton';
+import Dialog from '../UI/Dialog';
+import ColorField from '../UI/ColorField';
+import { Column, Line, Spacer } from '../UI/Grid';
+import InlineCheckbox from '../UI/InlineCheckbox';
+import Text from '../UI/Text';
+import { type RGBColor, type ColorResult } from '../UI/ColorField/ColorPicker';
+import DismissableAlertMessage from '../UI/DismissableAlertMessage';
+
+type Props = {
```
4ian, line 13: `{| |}`

newIDE/app/src/LayersList/LightingLayerDialog.js
```diff
@@ -0,0 +1,96 @@
...
+  ];
+  return (
+    <Dialog
+      noMargin
+      open={!!props.layer}
+      title={<Trans>Lighting layer settings</Trans>}
+      actions={actions}
+    >
+      <Column>
+        <Line>
```
4ian, line 54: Use `<ColumnStackLayout>` that is a Column that introduces `Spacer` between each item - you then probably don't need Line and Spacer at all.

newIDE/app/src/LayersList/index.js
```diff
@@ -178,6 +211,14 @@ export default class LayersList extends Component<Props, State> {
           </FullSizeMeasurer>
           <Column>
             <Line justifyContent="flex-end" expand>
+              <RaisedButton
+                label={<Trans>Add lighting layer</Trans>}
+                primary
```
4ian, line 216: Let's not make this button primary, and let's show it on a line after Add a layer - to ensure first time users are drove to use "Add a layer" first.

GDJS/Runtime/pixi-renderers/runtimescene-pixi-renderer.js
HarsimranVirk: Uhmm, yes this is not required anywhere at all right now :) I was using this in the `LayerPixiRenderer` (since it passes `RuntimeSceneRenderer`) to query the layers. I believe this was before we added properties in Layer.h and afterward, the code changed a lot. So, I'll be removing this since it's not needed anymore :)

HarsimranVirk:
@4ian I've made the changes according to the review (and mostly everything you mentioned should be changed I guess). I could see some conflicts right now (because of recent merge of hot-loading PR), I'll be fixing that, most importantly the filters :)

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
```diff
@@ -109,6 +109,8 @@ gdjs.LayerPixiRenderer.prototype.addEffect = function (effectData) {
     update: filterCreator.update,
   };
 
+  // @ts-ignore PIXI isn't typed for now.
+  if (this._isLightingLayer) filter.pixiFilter.blendMode = PIXI.BLEND_MODES.ADD;
```
HarsimranVirk, line 113: @4ian I guess if the layer is a lighting layer we should be having the ADD blend mode for filters (always?) 🤔.

4ian: ADD or MULTIPLY? But yeah I guess all filters should then use it.. or just the container?

4ian:
> I could see some conflicts right now (because of recent merge of hot-loading PR), I'll be fixing that, most importantly the filters :)

Thanks! The important change is that indeed filters can be dynamically added/removed at runtime. Hopefully it won't be too hard to adapt your changes to this 🤞

Speaking of `master` branch, if you merge it again, I just added the typing of the `PIXI` global variable ;) You can now benefit from the sweet autocompletions in VSCode and type checking with TypeScript (`npm run check-types`!)

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
HarsimranVirk: Filters are affecting the light objects, so they should he having ADD blend mode. (They are applied on the layer's container, however the container is meant to contain only light objects.) So, if the filters would have any other blend mode, they will basically be overriding light object's blend mode.

HarsimranVirk: > or just the container?

Only the filters applied on the lighting layer's container should be having ADD blend mode, we shouldn't really change anything else I guess.

4ian: Ok make sense! I was wondering if there was possibility of using ADD blend mode on the container only (rather than the filters + individual light objects) (and still have a multiply on the render texture).
Sounds fine to use ADD on filters/objects if it's not possible :)

HarsimranVirk: > I was wondering if there was possibility of using ADD blend mode on the container only (rather than the filters + individual light objects) (and still have a multiply on the render texture).

I didn't think about this earlier 🤔 I'll inform you after trying this out :)

GDJS/Runtime/pixi-renderers/runtimescene-pixi-renderer.js
```diff
@@ -133,11 +133,14 @@ gdjs.RuntimeScenePixiRenderer.prototype.setLayerIndex = function (
...
+  /** @type {PIXI.Container | ?PIXI.Sprite} */
   var layerPixiObject = layerPixiRenderer.getRendererObject();
 
   if (layer.isLightingLayer())
     layerPixiObject = layerPixiRenderer.getLightingSprite();
 
+  if (!layerPixiObject) return;
   if (this._pixiContainer.children.indexOf(layerPixiObject) === index) return;
 
   this._pixiContainer.removeChild(layerPixiObject);
```
HarsimranVirk: @4ian, in the case of the lighting layer, the layer's pixi container is already removed (and replaced with lighting sprite). This way, we could handle all the cases for hot-reloading. We can reorder lighting layer as well, which will basically change the index of the sprite, instead of the container (which wouldn't even exist in the scene's container in the first place).

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
```diff
@@ -336,9 +339,10 @@ gdjs.LayerPixiRenderer.prototype.getRenderTexture = function () {
  */
 gdjs.LayerPixiRenderer.prototype._replaceContainerWithSprite = function () {
   if (!this._pixiRenderer) return;
-  // @ts-ignore PIXI isn't typed for now.
-  this._lightingSprite = new PIXI.Sprite(this.getRenderTexture());
```
HarsimranVirk: `getRenderTexture()` could return a null value, typescript helped in identifying this 😬. So, I changed it a bit to perform a null check (and I'm not calling `getRenderTexture()` anymore because I don't want to assign anything to a new variable, ie, `var rt = this.getRenderTexture()`).

GDJS/Runtime/pixi-renderers/runtimescene-pixi-renderer.js
4ian: That sounds fine, it's also the reason for calling the method "getRendererObject" initially - the layer can be a PIXI.Container, but also a PIXI.Sprite.

newIDE/app/src/Hints/ObjectsAdditionalWork.js
```diff
@@ -0,0 +1,107 @@
...
+  return null;
+};
+
+export const onInstanceAdded = (
+  instance: gdInitialInstance,
+  layout: gdLayout,
+  project: gdProject
+): ?InfoBarDetails => {
+  const services =
+    objectType[layout.getObject(instance.getObjectName()).getType()];
```
4ian, line 42: calling this "services" is a bit weird because it's called "objectType" as a variable. "Service" is also very generic, everything is a service in a way :) Can we think of a better name?
If your file is called "ObjectsAdditionalWork", then it must surely store an object called "objectsAdditionalWork", or something using the same vocabulary. Then if it's indexing some "additional work" by "object (type)", then when you're accessing it by giving a type, you're then getting something called... "additionalWork"?

newIDE/app/src/LayersList/LightingLayerDialog.js
```diff
@@ -0,0 +1,82 @@
...
+      title={<Trans>Lighting layer settings</Trans>}
+      actions={actions}
+    >
+      <ColumnStackLayout>
+        <DismissableAlertMessage kind="info" identifier="lighting-layer-usage">
+          <Trans>
+            Lighting layer introduces ambient light in the scene. There should
+            be only one lighting layer per scene. Ideally all the lights should
+            be in the lighting layer and in most of the cases lighting layer
+            should follow the base layer.
```
4ian, line 58: The lighting layer renders an ambient light on the scene. All lights should be placed on this layer so that shadows are properly rendered.
By default, the layer follows the base layer camera. Uncheck this if you want to manually move the camera using events.

newIDE/app/src/LayersList/index.js
```diff
@@ -182,6 +215,13 @@ export default class LayersList extends Component<Props, State> {
           </FullSizeMeasurer>
           <Column>
             <Line justifyContent="flex-end" expand>
```
4ian, line 217: I think I added a comment about having these buttons on two lines previously? Check if you can do that :)

HarsimranVirk: @4ian do you mean the button should be below the `Add a layer` button and not beside it?

Extensions/Lighting/lightobstacleruntimebehavior.js
```diff
@@ -0,0 +1,115 @@
+/**
+ * @memberof gdjs
+ * @class LightObstaclesManager
+ * @param {gdjs.RuntimeScene} runtimeScene
+ * @param {*} sharedData
+ */
+gdjs.LightObstaclesManager = function (runtimeScene, sharedData) {
```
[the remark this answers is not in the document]
HarsimranVirk: Nope, it's not used at all. 😬

newIDE/app/src/LayersList/index.js
4ian: Yes because I'm concerned about small screens. Check how it looks if you resize the layers editor width to be very small?

4ian:
I quickly tried yesterday, looking good overall! Let me do a new pass on the code+overall comments tonight. 

As discussed, if we can have tests for the polygon resulting from the raycast done by the object, that would be awesome and will increase a lot our confidence to ship this! Let me know how it goes :)

4ian:
A few general comments in no particular order :)

* Should the default color of a light be all white? Otherwise it's a bit confusing for the first time user because the light seems not very bright. 
* Maybe let's make the default ambient color a bit more white too, so that the game is by default not too darkened.
* The debug mode is not working, crash it seems.
* See if you can add `updateFromObjectData` (see examples in other objects) to support hot-reloading. The function should return true if you applied changes, false otherwise (the user will be asked to relaunch the preview from scratch - avoid this if possible).
* The name of the texture is "Image resource". See if you can change this to "Optional custom image to use to display the light". Also add a long description (`setLongDescription`, I *think* it's supported) "This image is optional. If you specify one, it will be used to render the light. A black and white image is recommended, black being shadow and white being light."
* In `Hints/index.js`, let's add a hint for the Light object which is "Add the behavior "Light Obstacle" to the objects that should stop the light (typically: walls, platforms, etc...). You can change the ambient light color in the "Lighting" layer properties.".
* Important stuff for performance: I think we should NOT update the light if it's not visible on screen. On a large level with 300 lights, you want to only do the `update` (i.e: the raycasting) for those shown on screen. `update` is always called on all objects, but you can avoid doing any work by checking if the PIXI object is hidden (the gdjs.RuntimeScene is [automatically setting `visible = false`](https://github.com/4ian/GDevelop/blob/4e04e79b2f1729748fbcee8adb6ac1fb348cba75/GDJS/Runtime/runtimescene.js#L459-L480) if the object is not shown at all on the screen - this is called "culling").

To test this, create a scene with ~300 lights, regularly spaced. Launch the preview. See that performance is horrible. Then add the "early bail out" if the renderer object is not visible, and verify if it's faster. If not, add some console.log to verify which functions are called (are there 300 update still done for some reason?) or try the JavaScript profiler in Chrome/Firefox to see what function is eating all the CPU time.

Finally another observation, in the scene editor, the light is rendered additively, which is both good for a preview and bad if you have a quite large light and you want to edit the level: 
  image

A user can always hide the lighting layer if it's a pain when editing the level I guess?

HarsimranVirk:
> Should the default color of a light be all white? Otherwise it's a bit confusing for the first time user because the light seems not very bright.
Maybe let's make the default ambient color a bit more white too, so that the game is by default not too darkened.

We can change the defaults I guess, shouldn't be a problem.

> A user can always hide the lighting layer if it's a pain when editing the level I guess?

Yes, we can hide the layer (and the light objects are hidden as well), so that shouldn't be a problem as well.

I'll try to fix debug mode's crash and performance issues then, and let you know about them 😬.

4ian:
Sure thanks :) For performance we can't be sure how the performance will be with 300 (or in general "a lot") of objects even with the optimization of not-doing-the-update-if-not-visible... but if you ensure your object is not doing anything, then the speed should be now just a matter of how speedy is the `gdjs.RuntimeScene` iterating on objects - i.e: your object is almost doing nothing so should have very little impact on performance. Again be sure to add a few console.logs to be validate how many lights are updated - sometimes it's easy to miss things because we don't see objects out of the screen, so it's not obvious if we wrongly update an object ;)

HarsimranVirk:
@4ian, I wasn't really able to test performance correctly, since there was no performance drop at all, to begin with (at least in my machine, it's obvious that other devices, especially mobile, would experience performance issues 😬) So, I decided to spam effects and behaviors to get some additional performance drop (didn't do much either). Bottom line, I was getting around 57-58 FPS before culling, and constant 60 FPS after culling (with a lot of unnecessary effects and behaviors ), I'd say it does improve performance then. Also, I had about 900 light objects with a lot of light obstacles in between.

4ian:
Mmm interesting! 900 lights that's a lot, great test :) I guess the RTree is then doing a very good job at keeping the raycasting to have a low performance impact. Culling seems great to have a minimal footprint then - a light that is not shown should almost be 0 cost for the engine.

Do you think you could add this as a test game by saving it in a folder like `GDJS/tests/games/lighting-benchmark`?

GDJS/tests/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.js",
   "private": true,
   "scripts": {
-    "test": "./node_modules/karma/bin/karma start --browsers ChromeHeadless --single-run",
+    "test": "cd ./node_modules/karma/bin && karma start ../../../karma.conf.js --browsers ChromeHeadless --single-run",
```
HarsimranVirk, line 8: @4ian this was pushed by mistake 😬, but still, can you check whether this works correctly on macOS/linux or not? If yes, I guess we can keep it?

GDJS/tests/tests/Extensions/lightruntimeobject.js
```diff
@@ -0,0 +1,40 @@
+/**
+ * Tests for Light Object
+ */
+
+describe('gdjs.LightRuntimeObject', function () {
```
HarsimranVirk, line 5: These are pretty small tests (which don't really add a lot of value), I'll be writing more especially related to raycasting. Can you explain to me how can I make game objects having certain behaviors (light obstacle), I see the behaviors array, but what should it contain? The name of the behavior? (`LightObstacleBehavior`)

GDJS/tests/package.json
4ian: It works :) 
Can you try though just: `karma start --browsers ChromeHeadless --single-run`?
(and `karma start --browsers Firefox --single-run`?)

If it works, let's use this. In theory, npm is responsible for going into the `node_modules` folder and running the appropriate program when you just write the program name like "karma".

So `karma start --browsers ChromeHeadless --single-run` should work. I don't know why I put originally `./node_modules/karma/bin/karma`, I guess I was not aware that it would work with just `karma`.

GDJS/tests/tests/Extensions/lightruntimeobject.js
4ian: That's a great start! Definitely keep this test and let's add another one or more with obstacles.

See `platformerobjectruntimebehavior.spec.js` as an example of creating an object with a certain behavior.
The `behaviors` array is an array of `BehaviorData`, which are the objects being passed to the behavior constructor. They contain at least the `name` of the behavior (not important), the `type` (should be set to `Lighting::LightObstacleBehavior` in your case) and the rest is what is expected by your behavior.

You create the new object (probably a raw `gdjs.RuntimeObject` like in `platformerobjectruntimebehavior.spec.js`, then override `getWidth`/`getHeight` to give it some volume, then you can add it to the runtime scene (`runtimeScene.addObject`) and set its position.

I recommend doing this initialization in a separate helper function if you want to reuse it for multiple tests. Then, each test can put a light object at different places and check the vertices/etc :) (it's good practice to make sure that tests are entirely autonomous and don't rely on other tests being run before - don't hesitate to make multiple tests with `describe`/`it` rather than one huge test - multiple tests are much easier to understand and see why they are failing).

Extensions/Lighting/lightobstacleruntimebehavior.js
```diff
@@ -0,0 +1,114 @@
...
+  var nearbyPlatforms = this._obstacleRBush.search(searchArea);
+  result.length = 0;
+  result.push.apply(result, nearbyPlatforms);
+};
+
+/**
+ * @memberof gdjs
+ * @class LightObstacleRuntimeBehavior
+ * @param {gdjs.RuntimeScene} runtimeScene
+ * @param {*} behaviorData
```
4ian, line 60: Prefer to use `any` than `*`, it's more explicit and TypeScript compatible.
In this case, you can use the type `BehaviorData`, that is a global type declared in `runtimebehavior.js`

Extensions/Lighting/lightobstacleruntimebehavior.js
```diff
@@ -0,0 +1,114 @@
...
+  this._obstacleRBush = new rbush(9, [
+    '.owner.getAABB().min[0]',
+    '.owner.getAABB().min[1]',
+    '.owner.getAABB().max[0]',
+    '.owner.getAABB().max[1]',
+  ]);
+};
+
+gdjs.LightObstaclesManager.getManager = function (runtimeScene) {
+  if (!runtimeScene.lightObstaclesManager) {
```
4ian, line 16: nit: can we use `_lightObstaclesManager`?  We're extending runtimeScene which is already a "bad enough" practice by itself (though the most efficient option I think), so let's put a underscore so that someone does not come and play with our obstacle manager ;)

newIDE/app/src/Hints/ObjectsAdditionalWork.js
```diff
@@ -0,0 +1,107 @@
...
+import { t } from '@lingui/macro';
+import { type AlertMessageIdentifier } from '../MainFrame/Preferences/PreferencesContext';
+import newNameGenerator from '../Utils/NewNameGenerator';
+import { type MessageDescriptor } from '../Utils/i18n/MessageDescriptor.flow';
+
+/*
+ * Define additional logic which executes after an object/instance has been created.
+ * Also, InfoBar can be used which could notify users of the additional changes.
+ * Declare new identifier for infoBar in Mainframe/Preferences/PreferenceContext
+ * and add it in hints/explaination list.
```
4ian, line 11: typo here

newIDE/app/src/SceneEditor/index.js
```diff
@@ -455,7 +474,21 @@ export default class SceneEditor extends React.Component<Props, State> {
...
+        this.props.layout,
+        this.props.project
+      );
+      if (infoBarDetails) {
+        this.setState({
+          additionalWorkInfoBar: infoBarDetails,
+          showAdditionalWorkInfoBar: true,
+        });
+      }
+    });
```
4ian, line 490: Can you replace this by `this._onInstancesAdded(instances)`?

4ian: you would need to remove the saveToHistory and updateToolbar because it's already done then.

newIDE/app/src/SceneEditor/index.js
```diff
@@ -175,6 +182,7 @@ export default class SceneEditor extends React.Component<Props, State> {
       onCloseLayerRemoveDialog: null,
       layerRemoved: null,
       effectsEditedLayer: null,
+      lightingEditLayer: null,
```
4ian, line 185: nit: `lightingEditedLayer`

("Edit" is an action, "Edited" is a state)

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
4ian: Let me know if you can try this :)

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
```diff
@@ -236,9 +266,90 @@ gdjs.LayerPixiRenderer.prototype.enableEffect = function(name, value) {
...
+    oldSourceFrame,
+    undefined
+  );
+};
+
+/**
+ * Get the render texture used to display this layer. If render texture is not used, it is created.
+ * @returns {?PIXI.RenderTexture} RenderTexture of the container.
+ */
+gdjs.LayerPixiRenderer.prototype.getRenderTexture = function () {
```
4ian, line 328: Is this useful? Can we make this private? The fact that a render texture exists is an implementation details, the outside world should not know about this.

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
```diff
@@ -236,9 +266,90 @@ gdjs.LayerPixiRenderer.prototype.enableEffect = function(name, value) {
...
+  this._lightingSprite = new PIXI.Sprite(this._renderTexture);
+  this._lightingSprite.blendMode = PIXI.BLEND_MODES.MULTIPLY;
+
+  var sceneContainer = this._runtimeSceneRenderer.getPIXIContainer();
+  var index = sceneContainer.getChildIndex(this._pixiContainer);
+  sceneContainer.addChildAt(this._lightingSprite, index);
+  sceneContainer.removeChild(this._pixiContainer);
+};
+
+gdjs.LayerPixiRenderer.prototype.getPIXIContainer = function () {
```
4ian, line 353: Are you using this? There is already `getRendererObject`. If possible, it would actually be better to rename this `getRendererObject` to `getPIXIContainer`, because `getRendererObject` is used knowing that a container is returned.

4ian:
Still a few comments, and I think that with some tests we're reaching a point where this would be good to merge :)

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
HarsimranVirk: Yup, this isn't used anywhere at all as far as I know, it should be dead code then 😬. For now, let's remove this, we could always add it later if required :)

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
HarsimranVirk: Yup, it's not being used at all right now, should be dead code then 😬. I'll remove it, for now, we can always add it later if required :)

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
HarsimranVirk: Oh right, `getRendererObject` already returns the container, we shouldn't be using this then I guess.

Bouh:
I had tested the light object, it's easy to use, you have done an fantastic job here!


- I suppose the pixel perfect on shadows is hard to implement ?
- It's planned to add actions and conditions ?
I think to:
  - Object A is in the light source of light object B
  - Object A enter in the light area of lightobject B
  - Object A go out of the the light area of light object B
  - Test and define colors of the light object
  - Test and define radius of the light object
  - On/off the light object
  - Activate/desactivate the debug mode

It's no relevant because the sprite is inside another sprite, but the raycast sorting is broken.
image

image

HarsimranVirk:
> I had tested the light object, it's easy to use, you have done an fantastic job here! ✨☀

Thank you 😄.

> The sprite is inside another sprite, but the raycast sorting should take care of all corners if the sprite is in the light area.
But it's happend also without collide, but the obstacle is not entirely in the light area.

Can you explain how can I reproduce this? I'm not even sure what the exact problem is, but let's see if we can fix it 😬.

4ian:
> I had tested the light object, it's easy to use, you have done an fantastic job here! ✨☀

🎉

> Hardenest of the shadow

Kawase blur on the Lighting layer should do the trick I think?

> Test and define colors of the light object
> Test and define radius of the light object

Agree for these conditions. Let's go for these in priority - I thought we add them.

> On/off the light object

Can be done by setting the visibility no? To be tested :)

> Object A is in the light source of light object B
> Object A enter in the light area of lightobject B
> Object A go out of the the light area of light object B

Would be interesting, I think it can be made using the raycasting test (after all, being in a light source is equivalent to having a raycast from the light to the object against obstacles). 
Could be a good thing to add a condition doing automatically this raycast (so that the user don't have to do this "complex" condition and to avoid having to make a group of objects to duplicate the "light obstacle behavior").

blurymind:
This is starting to look really exciting. If we have condition to detect if an object is in the light source of another, some really cool stealth game mechanics can be implemented using this feature. 🎉 

Does this also support normal map textures on sprites? It would be amazing to use 
https://github.com/azagaya/laigter
with gdevelop games

4ian:
> Does/Will this also support normal map textures on sprites?

@HarsimranVirk hope to make a prototype of this before the end of GSoC. The implementation is a bit different, meaning that we could end up with two kind of light objects:
1. The (currently existing) "dynamic light" that you can see in this PR.
2. A "normal map based light source" that would use normal maps.

The normal map based light source would require a bit of rework of objects so that they all support normal maps, and a bit of rework in the game engine to render the layers in an in memory texture, then the normal maps in another in memory texture, then have both passed to a lighting shader.

In terms of timeline, I think it's unrealistic to have both by the end of August, but we should aim to polish as much as possible this (and merge by mid August) and have a prototype of the normal map lighting to prove that it works and estimate how much work would remain to make this a "real" feature :)

HarsimranVirk:
> Does/Will this also support normal map textures on sprites?
@blurymind not in this pr I guess, but I'm working on a prototype right now (I might push a draft pr this weekend hopefully 😬). @4ian and I decided that the normal maps rendering would be separate from "raycasting lights". Let's see if we could make it work 🤞.

4ian:
> I might push a draft pr this weekend hopefully 😬

👀

Bouh:
> > The sprite is inside another sprite, but the raycast sorting should take care of all corners if the sprite is in the light area.
> > But it's happend also without collide, but the obstacle is not entirely in the light area.
> 
> Can you explain how can I reproduce this? I'm not even sure what the exact problem is, but let's see if we can fix it 😬.

I've added some colours, in blue you have the sprite area, in yellow their corner, in pink two arrows. Two area in purples.
image


You can see the top right corner is inside the light area, this one work fine.

The other corners are not in the light area and the debug mode show us there are two bad lines (pointed with the pink arrows).
So the purple area should be included as well, this is not actually the case i guess because the algorithm doesn't care of the 3 other corners outside of the light area.

HarsimranVirk:
> this is not actually the case because i guess the algorithm doesn't care of the 3 other corners outside of the light area.

@Bouh, I guess that's the problem, the algorithm does care about the other 3 corners as well. The "Light Obstacles Manager" either includes the entire object (which would mean all the corners) or it discards the entire object 🤔. This is why you can see the red-colored ray being cast towards the top-left vertex of the sprite object, which hits the boundary of the "light's area", and we get that purple triangle at top of the sprite. Similarly, for the bottom triangle, the red-colored ray is being cast towards the bottom-right vertex 🤔.

blurymind:
> > Does/Will this also support normal map textures on sprites?
> 
> @blurymind not in this pr I guess, but I'm working on a prototype right now (I might push a draft pr this weekend hopefully 😬). @4ian and I decided that the normal maps rendering would be separate from "raycasting lights". Let's see if we could make it work 🤞.

That is an excellent idea. Some people might want to apply only one or the other, some might want both

Bouh:
I'm thinking about it now but do you also support hot reload or is it automatic? @4ian

HarsimranVirk:
> I'm thinking about it now but do you also support hot reload or is it automatic?

@Bouh thanks for reminding this, I had to implement it as well 💀. 

> I'm asking here, but I'm interested for my bitmapText.

I guess all we need to do is define a function `updateFromObjectData`, and return true if the properties are updated correctly. Otherwise, return false, in which case the user would have to restart the preview. Take a look at other objects for implementing this function (maybe `BBText` since that would correlate a bit to `bitmapText`? 😬).

4ian:
Correct, implement `updateFromObjectData` to get hot-reload support. Check how it's done for other objects (compare the properties, change anyone that must be updated, then `return true`. `return false` if something is not supported, in which case the user will be asked to relaunch the preview. This is the case by default if you don't implement `updateFromObjectData`)

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -187,21 +166,60 @@ gdjs.LightRuntimeObjectPixiRenderer.prototype.getRendererObject = function () {
 gdjs.LightRuntimeObjectPixiRenderer.prototype.ensureUpToDate = function () {
   if (this._object.isHidden()) return;
 
-  if (this._debugMode) this._updateDebugGraphics();
+  if (this._debugGraphics) this._updateDebugGraphics();
```
HarsimranVirk, line 169: @4ian somehow the debug graphics is broken, I tried to run the previous commits as well, but I'm unable to render PIXI.Graphics, there is no error, and there's no visual as well.

4ian: Is `_debugGraphics` set to something? Add some breakpoint/various console.log to inspect the content of the object?

HarsimranVirk: Yes, I could see that `_debugGraphics` is a valid object, and I am getting console logs for all the updating functions 🤔 I thought it might have to do with the refactoring (which is basically nothing more than extracting the logic to define the pixi mesh, shouldn't have to do anything with the graphics), but the graphics isn't rendering without the refactor as well.

4ian: Could it be because "if (this._object.isHidden()) return;"? Have you verified that `ensureUpToDate` is called?
I see that @bouh was able to make screenshot with the debug mode, so it was working at this time :)

HarsimranVirk: @4ian nvm I found the bug 😬, I had to trace back all the changes, it was due to our optimization of falling back to quads and the empty array returned by `_computeLightVertices()`. Also, we should be hot-reload ready now 🤞.

4ian:
I'm doing a new review now and we should prepare for merging this, I'll give this another try too :)

Extensions/Lighting/tests/lightruntimeobject.spec.js
```diff
@@ -0,0 +1,355 @@
...
+  });
+
+  it('Vertex and index buffers after obstacle is moved.', function () {
+    obstacle.setPosition(150, 250);
+    runtimeScene.renderAndStep();
+    light.update();
+
+    const vertexBuffer = light._renderer._vertexBuffer;
+    const indexBuffer = light._renderer._indexBuffer;
+    const vertexData = [
```
4ian, line 230: You can add `// prettier-ignore` on the lines just before where you declare arrays to avoid having them formatted to a huge number of lines :) 
See https://prettier.io/docs/en/ignore.html

Extensions/Lighting/tests/lightruntimeobject.spec.js
```diff
@@ -0,0 +1,355 @@
...
+      0,
+      18,
+      1,
+    ];
+
+    vertexData.forEach((val, index) => {
+      expect(vertexBuffer[index]).to.be(val);
+    });
+    indexData.forEach((val, index) => {
+      expect(indexBuffer[index]).to.be(val);
```
4ian, line 331: ubernit: if vertexData and vertexBuffer have the same "things", they should have the same name. Instead of changing between "data" and "buffer" (that are synonym but create confusion about "what's different so that they choose another name?"), add `expected` to the name of the arrays that are used for the test:
```suggestion
    expectedVertexBuffer.forEach((val, index) => {
      expect(vertexBuffer[index]).to.be(val);
    });
    expectedIndexBuffer.forEach((val, index) => {
      expect(indexBuffer[index]).to.be(val);
```

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
```diff
@@ -13,28 +13,49 @@
...
+  /** @type {?PIXI.RenderTexture} */
+  this._renderTexture = null;
+
+  /** @type {?PIXI.Sprite} */
+  this._lightingSprite = null;
+
+  this._runtimeSceneRenderer = runtimeSceneRenderer;
+  this._pixiRenderer = runtimeSceneRenderer.getPIXIRenderer();
+  this._oldWidth = null;
+  this._oldHeight = null;
```
4ian, line 31: Add a comment indicating that the width/height are tracked when a render texture is used.

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
```diff
@@ -236,9 +266,77 @@ gdjs.LayerPixiRenderer.prototype.enableEffect = function(name, value) {
...
+  this._pixiRenderer.render(this._pixiContainer, this._renderTexture, false);
+  this._pixiRenderer.renderTexture.bind(
+    oldRenderTexture,
+    oldSourceFrame,
+    undefined
+  );
+};
+
+/**
+ * Enable the user of a PIXI.RenderTexture to render the PIXI.Container
```
4ian, line 325:
```suggestion
 * Enable the use of a PIXI.RenderTexture to render the PIXI.Container 
```

4ian:
A few more comments after testing:

- [ ] Changing the ambient light and hot reloading => not applied. You might need to do something in `gdjs.HotReloader.prototype._hotReloadRuntimeLayer`.
- [ ] Running a preview with just a light. Then add a Light Obstacle Manager to an object, and hot-reload => the behavior is not applied.
  - This is probably because the object can't find the light obstacle manager. Solution: always include all the files, including the lightobstacleruntimebehavior.js, even in the includes of the object.
- [ ] You need to implement `onActivate` and `onDeActivate` for your behavior. See how it's done for gdjs.PlatformRuntimeBehavior, otherwise activating/disabling the behavior using events won't stop blocking the light.

Let's solve all of this and the other comments in the code and we should be good to go with this :)

4ian:
Speaking of onDestroy, I made a fix so that removing a behavior and launching a hot-reloading is properly calling `onDestroy` (so that for example, platforms don't stay as platforms after removing the Platform behavior) in 4324526689519a6ca809db9d211f7e45a158cd6f. I surely want to merge master again :)

HarsimranVirk:
@4ian quick question:
> I made a fix so that removing a behavior and launching a hot-reloading is properly calling onDestroy

Does it require rebuilding libGD?

4ian:
Yep there was a change in the code generation!
As for lighting, nothing in libGD, this is all in the extension "runtime" code. Adding these methods us enough for the game engine to call them.

4ian:
Finally merged! :) 
This seems to work well enough for us to go ahead with this! Thanks a lot for all the work made on this so far. It's a great feature that will certainly get a lot of love from users and will unlock very nice effects and new gameplay possibilities :) 

Let's see what we can achieve more in the remaining days of GSoC!

Extensions/Lighting/lightruntimeobject.js
```diff
@@ -0,0 +1,208 @@
...
+  this._debugMode = lightObjectData.content.debugMode;
+
+  /** @type {?PIXI.Texture} */
+  this._texture =
+    lightObjectData.content.texture === ''
+      ? null
+      : runtimeScene
+          .getGame()
+          .getImageManager()
+          .getPIXITexture(lightObjectData.content.texture);
```
4ian, line 42: I unfortunately missed this, but the PIXI Texture should in theory be hold only in the "pixi renderer". Here, we should just store something that is "renderer-agnostic", i.e: the texture resource name (a string).

Could you do a PR to make `_texture` a string? Use `getPIXITexture` in the renderer. If you know that the texture was changed, then call a method called `updateTexture` of the renderer - which will do the job of updating the PIXI.Texture.

This way, we have an object that could work with something entirely different than PIXI in the future :)

Extensions/Lighting/lightruntimeobject.js
```diff
@@ -0,0 +1,208 @@
...
+
+  /** @type {?gdjs.LightObstaclesManager} */
+  this._obstaclesManager =
+    gdjs.LightObstaclesManager !== undefined
+      ? gdjs.LightObstaclesManager.getManager(runtimeScene)
```
4ian, line 47: I believe there is no reason for this to be undefined anymore? We should move this to be a ` @type {gdjs.LightObstaclesManager}` then.

Extensions/Lighting/lightruntimeobject.js
4ian: (Actually updateTexture already exists, and can be updated :))

PR #1892, opened 2020-07-25
Add an section for developers in the patch note

Bouh:
We can add a spoiler at the bottom of the patch note for keeping a track of important thing for devs.

```
<details>
  <summary>For developpers</summary>
  
  Spoiler text extract from commits like this one:
https://github.com/4ian/GDevelop/commit/cc158a9250f8c275a590253e5f1f9e28d8ddf3c4
And others
 
</details>
```

<details>
  <summary>For developpers</summary>
  
  Spoiler text extract from commits like this one:
https://github.com/4ian/GDevelop/commit/cc158a9250f8c275a590253e5f1f9e28d8ddf3c4
And others
 
</details>

Merge pull request  has no style.
image


newIDE/app/scripts/extract-changelog.js
```diff
@@ -183,8 +185,9 @@ const formatHiddenCommitMessage = commit => {
 
   const hiddenCommits = commitsWithAuthors.filter(commit => commit.hidden);
   const displayedCommits = commitsWithAuthors.filter(commit => !commit.hidden);
+  const devCommits = displayedCommits.filter(commit => commit.forDev);
   const fixCommits = displayedCommits.filter(commit => commit.isFix);
-  const improvementsCommits = displayedCommits.filter(commit => !commit.isFix);
+  const improvementsCommits = displayedCommits.filter(commit => { return !commit.isFix && !commit.forDev});
```
4ian, line 190: nitpicking:

```suggestion
  const improvementsCommits = displayedCommits.filter(commit => !commit.isFix && !commit.forDev);
```

newIDE/app/scripts/extract-changelog.js
```diff
@@ -199,4 +202,11 @@ const formatHiddenCommitMessage = commit => {
 
   shell.echo(`\n## 🐛 Bug fixes\n`);
   shell.echo(fixCommits.map(formatCommitMessage).join('\n'));
+
+  if (devCommits) {
+    shell.echo(`\n<details>\n`);
+    shell.echo(`\n<summary> 🛠 Developers</summary>\n`);
```
4ian, line 208: We'll have to test how this renders inside GDevelop when showing the changelog

(if you want to try, launch Storybook and search the `Changelog` component. Modify the code in `GDevelopServicesTestData.js` to insert this and see how it renders)

4ian:
> Merge pull request has no style

For the `Merge pull request #`, we should hide these commits (because all the information is inside the other commits of the branch that was merged).

Bouh:
The spoiler is useable :) 

image

newIDE/app/scripts/extract-changelog.js
```diff
@@ -201,4 +204,13 @@ const formatHiddenCommitMessage = commit => {
 
   shell.echo(`\n## 🐛 Bug fixes\n`);
   shell.echo(fixCommits.map(formatCommitMessage).join('\n'));
+
+  if (devCommits) {
+    shell.echo(`\n<details>`);
+    shell.echo(`<summary> 🛠 Developers</summary>`);
```
arthuro555, line 210: You should not use HTML as if I remember correctly the changelog is extracted to be put on GitHub. GitHub will escape the HTML for obvious security reasons, and there will just be the ugly tags polluting the changelog without their effect. Just use markdown ;)

Bouh: This HTML work in GDevelop and github :)

arthuro555: It seems so, nevermind about the fact that GitHub escapes it, but it still is a security issue. Imagine someone makes a MitM attack and adds malicious scripts to the changelog, it wouldn't be escaped by GDevelop and give remote code execution on an environment with node, that is very dangerous!
Maybe GDevelop could replace those specific HTML tags with a Markdown version and use a plugin for the markdown parser to add spoilers.

arthuro555: Here is a remark (the used markdown renderer) plugin that would do that: https://github.com/Rokt33r/remark-collapse

4ian: We should indeed not accept arbitrary html. Let me know if you can use this or something else to accept **only** summary and details :)

Bouh:
Nothing in the proposals is working.
If I escape the HTML then it's useless to detect or not detect details/summary tags.

There is a way to detect some markdown node types like it's done [here](makeMarkdownCustomRenderers) . 
But since the spoiler is not supported by markdown or by a pure HTML component it doesn't seem possible.

arthuro555:
What I was saying is working I think?
For the github leave it like currently. For the IDE use the remeark plugin and a regex or something to replace the html with mardown equivalent.

Bouh:
There is no equivalent in md.

arthuro555:
For the 3rd time, there is if you use a remark plugin. I sent you the link to that plugin before: https://github.com/Rokt33r/remark-collapse

Bouh:
Yes, I'm talking about the IDE only.
So what, i use remark plugin for parse the extracted changelog, find` <details>` and` <summary>` and the result string in output contain also` <details>` and `<summary>` and i keep `escapeHTML={false}` ?
I must be missing the logic somewhere for me it's the same.

Bouh:
The complexity of filtering tags is not worth it compared to just putting the info in a smaller category.
It's like Arthuro suggested in PM, add the Dev category in a H3.

4ian:
I agree! Let's not complicate our life too much and stay on the "safe and simple" side of things. Looks good to me! :)

PR #1896, opened 2020-07-28
Add network preview command, hide debug and network preview commands on web

NilayMajorwar:
This PR adds a new command for launching a network preview. Also, now debug preview and network preview commands are usable only on the desktop app (earlier they were incorrectly available on web app too and didn't work on launching the command).


4ian:
Thanks! 👍

PR #1897, opened 2020-07-29
[Draft] Cameras and viewports

arthuro555:
This PR adds multiple cameras and viewport objects. 
The cameras that are not 0 have their values recorded in a list of additional cameras.  
The camera viewport object does a simple thing: it renders `runtimeScene.getRenderer().getPIXIContainer()` into a RenderTexture and puts it in a sprite (which is the rendered object). When rendering from another camera than 0, it will set the additional camera's data as the main data for all layer, and restore it to the main camera's data after rendering.

### DIFF
--- a/Extensions/CameraViewport/cameraviewportruntimeobject-pixi-renderer.js
+++ b/Extensions/CameraViewport/cameraviewportruntimeobject-pixi-renderer.js
@@ -1,0 +1,58 @@
+/**
+ * The PIXI.js renderer for the DummyRuntimeObject.
+ *
+ * @class DummyRuntimeObjectPixiRenderer
+ * @constructor
+ * @param {gdjs.CameraViewportObject} runtimeObject The object to render
+ * @param {gdjs.RuntimeScene} runtimeScene The gdjs.RuntimeScene in which the object is
+ */
+gdjs.CameraViewportObjectPixiRenderer = function(runtimeObject, runtimeScene)
+{
+    this._object = runtimeObject; // Keep a reference to the object to read from it later.
+    this._renderer = runtimeScene.getGame().getRenderer().getPIXIRenderer();
+    this._renderTexture = new PIXI.RenderTexture.create({width: runtimeObject.getWidth(), height: runtimeObject.getHeight()});
+    this._sprite = new PIXI.Sprite(this._renderTexture);
+    this._scene = runtimeScene;
+};
+
+gdjs.CameraViewportObjectRenderer = gdjs.CameraViewportObjectPixiRenderer; //Register the class to let the engine use it.
+
+gdjs.CameraViewportObjectPixiRenderer.prototype.getRendererObject = function() {
+    return this._sprite;
+};
+
+gdjs.CameraViewportObjectPixiRenderer.prototype.update = function() {
+    /* 
+     * Remove all cameras: cameras can't render themselves as it could result in an endless loop.
+     * We could theoretically count the renders and replace the camera with a blank texture after too many renders,
+     * But chrome web gl automatically blocks textures that are made from a texture to be applied to the texture it is made from.
+     */
+     this._scene._constructListOfAllInstances();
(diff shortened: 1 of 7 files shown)


Extensions/CameraViewport/cameraviewportruntimeobject-pixi-renderer.js
```diff
@@ -0,0 +1,58 @@
...
+     * But chrome web gl automatically blocks textures that are made from a texture to be applied to the texture it is made from.
+     */
+     this._scene._constructListOfAllInstances();
+    for(var objectId in this._scene._allInstancesList) {
+        var object = this._scene._allInstancesList[objectId];
+        if(object.type !== "CameraViewport::CameraViewport") continue;
+        this._scene
+            .getLayer(object.layer)
+            .getRenderer()
+            .removeRendererObject(object.getRendererObject());
```
4ian, line 37: Interesting :) Do you think you could do something like setting `visible` on the `object.getRendererObject()`(that we know to be a `DisplayObject`) to false, and then put it back to its initial value?
Might be faster than removing the renderer object and then adding it back, which are fairly costly in comparison.

4ian:
Very interesting! I would not have thought of rendering a camera using an object (we could almost call it a "Camera screen"), but it makes sense.
Looking forward to see if we can have something working and performant enough! :)

Extensions/CameraViewport/cameraviewportruntimeobject-pixi-renderer.js
arthuro555: > Do you think you could do something like setting visible  


Good idea! Works perfectly :)

arthuro555:
For the performance, I began only noticing some minor frame drops at 23 viewports on-screen. Here is what it looks like: 
[Fun with cameras](https://streamable.com/u266dx)

4ian:
Looks great! Would be important to test on mobile to see what are the limits there (and maybe in the future we can add some warnings in GDevelop around stuff like too much collisions/too much cameras/too much objects/etc...).

Doing the rendering in `update` works but is maybe a bit hacky for now. It could be something triggered by the scene - because if `update` is called on a camera object before other objects, then the camera will render in its texture a frame where some objects have not being updated, so you could see stuff like the same object shown by two cameras but with a different animation (in the case of a Sprite).
Or maybe I missed something?

In other words, because "camera screen" objects are a bit "special", they should probably be handled separately after all the object updates.

Something else about performance! Is there a way to stop their rendering if they are not shown on screen? The PIXI.Sprite should automatically be set to `.visible = false` by the `gdjs.RuntimeScene`, but we might use this to disable the update entirely (much like sprites are not updating their animations if they are not shown on screen).

arthuro555:
I never really touched objects before so it is all pretty new to me. I thought the update method was to update the object between events and rendering, and I don't know how to autohide off screen stuff yet, but I will look into it.

arthuro555:
I have a really confusing bug. It seems like sometimes the backed up layer info is not restored. I am confident saying sometimes as stepping through the debugger every step from backing up and restoring seems to work. 
Here is a project helping to see what is happening: 
[testCamera.zip](https://github.com/4ian/GDevelop/files/4997754/testCamera.zip)
When modifying the main layer x position to the camer 1 one, the backup restoring somehow fails and it stays on the same x position as camera 1. Same for the y: you can chage it but at some point it will get put to 0 as camera 1's y is 0.

arthuro555:
Closing this as the code is not compatible since we switched to typescript in the meantime, and I would need to redo it all from scratch anyways

arthuro555:
Not really, most of the feature provided in this is provided by an extension I made, "Sprite Snaphsot". With sprite snapshot, you can render the scene into a sprite, which is basically what this was doing, but you have more control on what exactly gets rendered and get all the flexibility of a sprite with the rendered result.
If you want to do split screen multiplayer, there is also an example that uses this extension that shows you how you can do it.

PR #1907, opened 2020-08-02
Remove methods polluting Array prototype in gd.js

4ian:


arthuro555:
I thought the remove method was optimized and better than splice? Was it not as optimized as splice in the end?

4ian:
Might have been in the best. In practice nowadays I would be more confident in the JS engine doing something faster with the native splice method. It's not creating any new array in memory so it's fine!

PR #1911, opened 2020-08-07
Fix Menu import for hide menu in game preview

Bouh:
[This fix](https://github.com/4ian/GDevelop/commit/1f4042bff00a68fdfaf1a95dec921ae66475fac4) doesn't containt the import for the Menu.
I've added the menu to the import and compiled Gdevelop and tested to export on the build services, the menu is now gone!

Without this PR you will see this error
image

**Ready to merge.**


4ian:
My bad! Thanks for checking after me ;)

PR #1912, opened 2020-08-07
Update cordova to v9

arthuro555:
The gdevelop services probably also require updating the cordova cli. 
This update is needed for compliance with the Play Store requirements, as since the 3rd August 2020 the minimal target sdk has to be 29 for new apps, and SDK 29 is only supported since cordova-android v9.

### DIFF
--- a/GDJS/Runtime/Cordova/config.xml
+++ b/GDJS/Runtime/Cordova/config.xml
@@ -28,7 +28,7 @@
     <preference name="Fullscreen" value="true" />
 
     <!-- Cordova/Phonegap version -->
-    <preference name="phonegap-version" value="cli-8.0.0" />
+    <preference name="phonegap-version" value="cli-9.0.0" />
 
     <!-- GDJS_ADMOB_PLUGIN_AND_APPLICATION_ID -->
 </widget>


4ian:
Good catch! I missed announcements about this new annual deadline. Build service is updated to cordova-android 9.0.0.

Also this means that **now only Android 5+ is supported by exported games** - which is a good thing because we don't want to continue supporting Android 4 going forward because of incompatibilities/low performance which is doing more harm than good.
This means that we can update https://github.com/4ian/GDevelop/blob/master/newIDE/docs/Supported-JavaScript-features-and-coding-style.md to say that we're now allowing cautious usage of "ES6" features like "fat arrows", `const/let` (but still stay cautious about construction creating "garbage" like map/filter) :)

We should investigate converting the codebase at some point using something like lebab: https://github.com/lebab/lebab
(cautiously too! For example, we prefer manual for-loop than forEach in most cases to avoid creating garbage by creating a closure every time)

Or just continue doing the upgrade manually, and in particular investigate usage of ES6 classes (https://caniuse.com/#feat=es6-class) which could help TypeScript/JSDoc annotations for classes/extend (`@extend` is not supported with "ES5 classes" and JSDoc). It's not supported by IE11 but maybe we should also entirely drop support for IE11 which was already a best effort.

cc @blurymind, @bouh, @HarsimranVirk for your interest, I'll be soon allowing what I described here (const/let, fat arrows...) in the Runtime/Extensions code :)

If someone want to update https://github.com/4ian/GDevelop/blob/master/newIDE/docs/Supported-JavaScript-features-and-coding-style.md, feel free to send a PR!

PR #1913, opened 2020-08-08
Create extract-extensions-document.js

rs4231199:
Created a script that extracts extensions' names, and descriptions from [https://github.com/4ian/GDevelop-extensions/blob/master/extensions-registry.json](https://github.com/4ian/GDevelop-extensions/blob/master/extensions-registry.json) file and write that into a markdown file.

### DIFF
--- a/newIDE/app/scripts/extract-extensions-document.js
+++ b/newIDE/app/scripts/extract-extensions-document.js
@@ -20,11 +20,11 @@ const writeFile = content => {
 (async () => {
   try {
     const response = await axios.get(extensionsRegistoryURL);
-    let texts = '## List of community-made extensions\n\n';
+    let texts = '# List of community-made extensions\n\n';
 
     response.data.extensionShortHeaders
     .forEach(element => {
-      texts += '### ' + element.fullName + '\n' + element.shortDescription + '\n\n';
+      texts += '## ' + element.fullName + '\n' + element.shortDescription + '\n\n';
     });
 
     writeFile(texts)
--- a/newIDE/app/scripts/extract-extensions-document.js
+++ b/newIDE/app/scripts/extract-extensions-document.js
@@ -2,10 +2,11 @@
  * Launch this script to generate a list(in markdown format) of all custom extensions.
  */
 
-const fs = require("fs");
-const axios = require("axios");
+const fs = require('fs');
+const axios = require('axios');
 
-const extensionsRegistoryURL = 'https://raw.githubusercontent.com/4ian/GDevelop-extensions/master/extensions-registry.json';
+const extensionsRegistoryURL =
+  'https://raw.githubusercontent.com/4ian/GDevelop-extensions/master/extensions-registry.json';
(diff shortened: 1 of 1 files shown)


arthuro555:
This script should probably be on the GDevelop-extensions repo instead of this one.

rs4231199:
I already have discussed this with @4ian . He said to put it here.

newIDE/app/scripts/extract-extensions-document.js
```diff
@@ -0,0 +1,46 @@
...
+const writeFile = content => {
+  return new Promise((resolve, reject) => {
+    fs.writeFile(outputFile, content, err => {
+      if (err) return reject(err);
+      resolve();
+    });
+  });
+};
+
+https.get(extensionsRegistoryURL, function(res) {
```
4ian, line 20: While it's true that your solution works, I now recommend using `axios` (as done in extract-changelog.js): 

```
const axios = require('axios');

axios.get(extensionsRegistryUrl).then(response => ...)
```

Or even better using async/await :)

newIDE/app/scripts/extract-extensions-document.js
```diff
@@ -0,0 +1,46 @@
...
+  let data = '',
+    json_data;
```
4ian, line 22: nitpicking: don't declare early variables.

newIDE/app/scripts/extract-extensions-document.js
```diff
@@ -0,0 +1,46 @@
...
+
+  res.on('data', function(stream) {
+    data += stream;
+  });
+  
+  res.on('end', function() {
+    json_data = JSON.parse(data);
```
4ian, line 29: nitpicking: camelCase: `jsonData`

Also try/catch around the JSON.parse

4ian:
Looks mostly good! Added a few comments. Thanks for working on this 👍

4ian:
Great! Merging now :) I've run Prettier and fixed the headings. 
The resulting page is here: http://wiki.compilgames.net/doku.php/gdevelop5/extensions/list

Ideally we could maybe run this like the documentation, as a regular task run on a CI like Circle CI that then upload the resulting files to the wiki hosting (cc @arthuro555), and we could run also extract-reference-document.js for the expression reference (http://wiki.compilgames.net/doku.php/gdevelop5/all-features/expressions-reference). I could give access to CircleCI/the Github repo of GDevelop-extensions to do that, like done here: https://github.com/4ian/GDevelop-documentation/blob/master/.circleci/config.yml

rs4231199:
@4ian. Hey Florian. You CCed arthuro555 in your message so I assumed my work here was done. Just to confirm the same and if there is any work for me I would be delighted to do that.

4ian:
Nothing more currently, I'm thinking of how to automate the calls to extract-reference-document.js and extract-extensions-document.js and upload the result to the wiki :) 
I think a regular CircleCI workflow like done for the documentation, with a `scp` to the server hosting the wiki would work.

PR #1914, opened 2020-08-08
Add 4ian as codeowner

arthuro555:
See https://github.blog/2017-07-06-introducing-code-owners/. 
This makes clear who owns the code (4ian) which might be useful if the repo is switched to the GDevelop Organisation.


4ian:
Excellent idea. Could be useful in the future to allow people to be owner of an extension :)

PR #1919, opened 2020-08-12
Add configuration for discord builds

arthuro555:
This adds a discord build configuation file for publishing releases of gdevelop to discord's store.

TODO:
- [x] Windows builds
- [x] Linux builds
- [x] Mac builds


arthuro555:
@4ian should I add some sort of readme for publishing on the dsicord store? All the steps are already [here](https://discord.com/developers/docs/dispatch/branches-and-builds), so I think it would be a bit redundant.

4ian:
> @4ian should I add some sort of readme for publishing on the dsicord store? All the steps are already [here](https://discord.com/developers/docs/dispatch/branches-and-builds), so I think it would be a bit redundant.

Definitely :) Add a section "Discord Store", possibly as a subsection of ["Desktop version"](https://github.com/4ian/GDevelop/blob/master/newIDE/README.md#desktop-version) describe in a few sentences what is to be done: install dispatch, run this command, wait/take a coffee, done.

It's very important because each "store"/new distribution platform for GDevelop is also an additional cost. For example it's taking me between 30min and 1h30 to distribute a new GDevelop version (30min if it's a bugfix and I just done it already, 1h30 if I have macOS signing issues, if there are upload failures, if the changelog needs cleaning, if there is an horrible bug/regression that I just found, if the web-app needs an update). 

This means that there is something that ideally we would have for the Discord store: a script! :) 
* For example, to distribute on GitHub, the deploy script is this one: https://github.com/4ian/GDevelop/blob/master/newIDE/electron-app/scripts/build.js and once it's started, if everything goes alright ~20 minutes later *without doing anything manually* the new version is on GitHub, only missing the changelog before being able to press the release button :) 
* The web-app deployment script is here: https://github.com/4ian/GDevelop/blob/master/newIDE/web-app/scripts/deploy.js and takes ~5 minutes or less - which is really great (and understandable because it's "just" pushing static resources).

Not sure exactly what Discord involves - are we just pushing a link, or executables? This might be very simple or get a bit complex but I'm not familiar with Discord store, so let me know.

arthuro555:
>  so we must automate as much stuff as possible.

The problem is I am not experienced at all with automating :/ If it was only putting commands one after another, it would be simple, but I need to analyze the output of the first command (discourse push build) to know the build ID that is needed to deploy it to a test branch and then to the master branch. I am not sure how to analyze the output to extract that build ID.

newIDE/electron-app/scripts/publish-discord.js
```diff
@@ -0,0 +1,56 @@
...
+
+  const notExistingMatch = output.match(
+    /(win|linux|mac)(-unpacked)?\"\) does not exist/gm
+  );
+  if (notExistingMatch) {
+    console.error('Please build GDevelop for ' + notExistingMatch[1]);
+    shell.exit(0);
+  }
+
+  const buildMatch = output.matchAll(/Build ID: (.*)$/gm).next().value;
```
arthuro555, line 46: Not sure why the normal String.match refused to work here 🤔

Bouh:
If the nightly build has the same name or the same installation path as the stable build it's problematic to install the nightly buils which replace the stable.

arthuro555:
As it is on another branch, discord automatically creates another folder with the branch id in the name: 
image

4ian:
@arthuro555 do you want to make a quick summary of what's the status of this? How does it look like on Discord? :)

arthuro555:
Right now, the store page looks like this:
image
image

There are 2 store channels, one for stable and the other for nightly builds (someone moved one to staff chat but I don't have the perms to take it out of it again). Those are currently hidden to regular users.
image


To upload a build, build the unpacked builds for mac linux and windows, and run `npm run publish-discord`. The unpacked builds should be automatically built by default when building normally the electron app, so to publish all there is to worry about is to run the publish-discord script after having built the rest regularly. 

CircleCI should upload the builds to the nightly branch automatically, but it doesn't run on my PRs so it is hard to truly test it, executing the commands locally work fine though so I assume it should run fine there as well.

Discord have [stopped reviewing for an indefinite amount of time due to "unforeseen circumstances"](https://support-dev.discord.com/hc/en-us/articles/360041437171), and that review is needed to let any user download the game :/. That doesn't mean we cannot begin to use it, we can still put a group of users (max 50, after that we would need verification) that test regularly the nightly build as test users on the [test user page](https://discord.com/developers/applications/718194112775454720/whitelist), so that they can download the nightly build  easily and get it updated to the last version in 1 click.

Bouh:
Stable and Nightly are now in Store section, only visible from GD team for now, let me know when we can make it public to everyone.
Or for tests.

arthuro555:
https://support-dev.discord.com/hc/en-us/articles/4414590563479

Discord has deprecated the store feature, it is no longer usable. I'm closing this PR, note that we may be eligible for a refund, more info on that page.

Issue #1918, opened 2020-08-12
[$80 Bounty] Simplify theme creation (possibly using CSS variables)


4ian:
When making a theme, there is a bunch of CSS files to modify:
![image](https://user-images.githubusercontent.com/1280130/90035643-52d9e800-dcc2-11ea-938c-27ad8d132ddb.png)

It's working ok so far, unfortunately: 

1. It's hard to **maintain**: these CSS files contain a mix of colors and a mix of CSS rules for the padding, or other stuff that are unrelated to the theme. This duplication is making harder to maintain the styling (if you update something in a theme, you have to update it for every theeme) and we risk forgetting a theme. For example: 
  * This should not be in a theme specific CSS file and should be moved to a generic CSS file for markdown:
 
 https://github.com/4ian/GDevelop/blob/4324526689519a6ca809db9d211f7e45a158cd6f/newIDE/app/src/UI/Theme/NordTheme/Markdown.css#L23-L25

  * Same for these:
 
  https://github.com/4ian/GDevelop/blob/4324526689519a6ca809db9d211f7e45a158cd6f/newIDE/app/src/UI/Theme/DarkTheme/Mosaic.css#L33-L38
  Apart from the color, everything else should be in a single CSS file.

  * Same for these complex rules that are repeated in every theme, while it could be a single variable:
  https://github.com/4ian/GDevelop/pull/1858/files#diff-3e7e1efe881654dfe25f559c629aeeeeR38-R43

2. It's **less performant** when we'll have more themes. This duplication is also putting more pressure on the browser CSS engine. Instead if we have a single set of CSS files and if we use CSS variables, this would reduce the duplication of CSS rules that we have for every theme.

3. There is also a bunch of duplication in the index.js files (for example, search for `// Use non rounded buttons`). We could move these into common variables or a function generating the material-ui theme (createMuiTheme).

---

I've created this issue (that is actually a refactoring/"feature request") so that I can add a bounty on this. The bounty will be awarded to anyone refactoring the existing themes into a set of simple files that are just containing colors, potentially using CSS variables (though anything in JS that is performant is fine too) and helper functions.
The goal is for each theme to be a set of colors, in at most one JS file and one CSS file.

4ian:
Bounty is posted here: https://www.bountysource.com/issues/92286673-bounty-simplify-theme-creation-possibly-using-css-variables

the-gem-dev:
i think i should try this.

tjrelyts:
[...]

PR #1921, opened 2020-08-13
Simplify theme creation using CSS variables

TheGemDev:
#1918 
To simplify the creation of themes i would be creating only the following files which should contain all codes for the themes, the files are EventsSheet.css, Index.js, Markdown.css, Mosaic.css, Table.css(in the future we could merge Markdown.css, Mosaic.css, Table.css into a single file) 
Presently @4ian i have been able to use css variables for the tables.css in the themes(dark, default and nord theme) i would do the same for the solarized dark theme as soon as you view this and let me know what you think.

ps - sorry i added existing files


newIDE/app/src/UI/Theme/Theme/Table.css
```diff
@@ -0,0 +1,140 @@
...
+  --table-Selected-Row: #4c92ff;
+  --table-Header-Column-font-weight: 600;
+  --table-Header-Column-font-size: 12px;
+  --table-Header-Column-color: #9e9e9e;
+  --table-Header-Column-text-transform: none;
+  --table-Column-white-space: nowrap;
+  --table-Column-overflow: hidden;
+  --table-Column-text-overflow: ellipsis;
+}
+/*darktheme*/
```
4ian, line 24: Each theme should have its own "Table.css" file (or a single css file, up to you), that contains just variables for the colors. Things like whitespace, margin, padding, overflow, font-size, etc... must be in a single CSS file used by all themes.

newIDE/app/src/UI/Theme/Theme/Table.css
```diff
@@ -0,0 +1,140 @@
...
+}
+
+.gd-table-default-theme .tableSelectedRow,
+.gd-table-dark-theme .tableSelectedRow {
+  background-color: var(--table-Selected-Row); /*OS X selection*/
+  color: var(--white);
+}
+
+.gd-table-default-theme .tableHeaderColumn,
+.gd-table-dark-theme .tableHeaderColumn {
```
4ian, line 81: In the end, there should not be anymore different rule names (gd-table-default-theme, gd-table-dark-theme, gd-table-nord-theme...) for every theme.

There should be a single CSS rule (`gd-table`) that is used, and when you select a theme, only the variables setting the color are changed.

newIDE/app/src/UI/Theme/Theme/Table.css
```diff
@@ -0,0 +1,140 @@
...
+  text-transform: var(--table-Header-Column-text-transform);
+  font-weight: var(--table-Header-Column-font-weight);
+  font-size: var(--table-Header-Column-font-size);
+  color: var(--table-Header-Column-color);
```
4ian, line 85: For example, this line is useful

newIDE/app/src/UI/Theme/Theme/Table.css
```diff
@@ -0,0 +1,140 @@
...
+.gd-table-dark-theme .tableSelectedRow {
+  background-color: var(--table-Selected-Row); /*OS X selection*/
+  color: var(--white);
+}
+
+.gd-table-default-theme .tableHeaderColumn,
+.gd-table-dark-theme .tableHeaderColumn {
+  text-transform: var(--table-Header-Column-text-transform);
+  font-weight: var(--table-Header-Column-font-weight);
+  font-size: var(--table-Header-Column-font-size);
```
4ian, line 84: For example, these lines should NOT be variables. Reconsider the idea behind this :)

the-gem-dev:
@4ian so far i have only worked on the default theme(turns out my work on the dark theme had a lot of issues, so i am changing most of codes on the first commit based on your suggestion) and i would be awaiting your thoughts on it before it can be done for the currently existing themes.
Presently I have created a shared eventssheet.css, markdown.css, mosaic.css and table.css which all themes would use so therefore each theme would only have 3 files which are index.js, theme.css, Themecolors.css. so i would be awaiting your thoughts on it.

the-gem-dev:
@4ian what do you think about this

4ian:
Can you explain what you've done? :) 
I still see a lot of similar `styles.css` files, while the goal is to only have one and themes are just specifying variables with colors?

the-gem-dev:
i created a shared file for the events sheet, markdown, mosaic and table.css which all themes share. i tried creating one `styles.css` but it kept on picking the value of another css variable which belongs to another theme, now i think its the `:root` selector makes it a globally scoped variable.(i am not really sure) but i think so.

PR #1929, opened 2020-08-19
Lighting fixes

HarsimranVirk:
We have 4 fixes here,

1. Removal of unnecessary undefined handling of `gdjs.LightObstaclesManager`.
2. Moving the handling of light texture to pixi renderer.
3. A really weird bugfix of debugging mode's graphics. I don't know what happened, but this is how the pixi graphics was being rendered:
Screenshot (1)

Neither I understand why this happened, nor I'm sure about why this particular fix works 😓. All I did was randomly change the order of `lineTo` and `moveTo` calls, and it worked. I'm guessing this is really specific to pixi, as we've upgraded the version as well, but arguably it was a minor change.

4. Also changed the algorithm a bit (specifics about it in the review below) to have that expanding boundary box of light.

### DIFF
--- a/Extensions/Lighting/lightruntimeobject-pixi-renderer.js
+++ b/Extensions/Lighting/lightruntimeobject-pixi-renderer.js
@@ -114,11 +114,12 @@ gdjs.LightRuntimeObjectPixiRenderer._verticesWithAngleComparator = function (
 gdjs.LightRuntimeObjectPixiRenderer._computeClosestIntersectionPoint = function (
   lightObject,
   angle,
-  polygons
+  polygons,
+  halfOfDiag
 ) {
   var centerX = lightObject.getX();
   var centerY = lightObject.getY();
-  var halfOfDiag = Math.sqrt(2) * lightObject.getRadius();
+  //var halfOfDiag = Math.sqrt(2) * lightObject.getRadius();
   var targetX = centerX + halfOfDiag * Math.cos(angle);
   var targetY = centerY + halfOfDiag * Math.sin(angle);
   var minSqDist = halfOfDiag * halfOfDiag;
@@ -271,14 +272,14 @@ gdjs.LightRuntimeObjectPixiRenderer.prototype._updateDebugGraphics = function ()
   var verticesCount = vertices.length;
   for (var i = 2; i < verticesCount; i += 2) {
     var lineColor = i % 4 === 0 ? 0xff0000 : 0x00ff00;
-    var lastX = i+2 >= verticesCount ? 2 : i+2;
-    var lastY = i+3 >= verticesCount ? 3 : i+3;
+    var lastX = i + 2 >= verticesCount ? 2 : i + 2;
+    var lastY = i + 3 >= verticesCount ? 3 : i + 3;
     this._debugGraphics
       .lineStyle(1, lineColor, 1)
       .lineTo(vertices[i], vertices[i + 1])
       .lineTo(vertices[lastX], vertices[lastY])
       .moveTo(vertices[0], vertices[1])
-      .lineTo(vertices[i], vertices[i+1])
+      .lineTo(vertices[i], vertices[i + 1])
       .moveTo(vertices[0], vertices[1])
       .lineTo(vertices[lastX], vertices[lastY]);
   }
@@ -400,11 +401,43 @@ gdjs.LightRuntimeObjectPixiRenderer.prototype._computeLightVertices = function (
       obstaclePolygons.push(obstacleHitBoxes[i][j]);
   }
 
+  var maxX = this._object.x + this._radius;
+  var minX = this._object.x - this._radius;
+  var maxY = this._object.y + this._radius;
+  var minY = this._object.y - this._radius;
+
   var flattenVertices = [];
-  for (var i = 0; i < obstaclePolygons.length; i++) {
+  for (var i = 1; i < obstaclePolygons.length; i++) {
     var vertices = obstaclePolygons[i].vertices;
     var verticesCount = vertices.length;
-    for (var j = 0; j < verticesCount; j++) flattenVertices.push(vertices[j]);
+    for (var j = 0; j < verticesCount; j++) {
+      flattenVertices.push(vertices[j]);
+
+      if (vertices[j][0] < minX) minX = vertices[j][0];
+      if (vertices[j][0] > maxX) maxX = vertices[j][0];
+      if (vertices[j][1] < minY) minY = vertices[j][1];
+      if (vertices[j][1] > maxY) maxY = vertices[j][1];
+    }
+  }
+
+  obstaclePolygons[0].vertices[0][0] = minX;
+  obstaclePolygons[0].vertices[0][1] = minY;
+  obstaclePolygons[0].vertices[1][0] = maxX;
+  obstaclePolygons[0].vertices[1][1] = minY;
+  obstaclePolygons[0].vertices[2][0] = maxX;
+  obstaclePolygons[0].vertices[2][1] = maxY;
+  obstaclePolygons[0].vertices[3][0] = minX;
+  obstaclePolygons[0].vertices[3][1] = maxY;
+
+  var halfOfDiag = Math.max(
+    Math.hypot(this._object.x - minX, this._object.y - minY),
+    Math.hypot(maxX - this._object.x, this._object.y - minY),
+    Math.hypot(maxX - this._object.x, maxY - this._object.y),
+    Math.hypot(this._object.x - minX, maxY - this._object.y)
+  )
+
+  for(var i = 0; i < 4; i++) {
+    flattenVertices.push(obstaclePolygons[0].vertices[i]);
   }
 
   var closestVertices = [];
@@ -417,7 +450,8 @@ gdjs.LightRuntimeObjectPixiRenderer.prototype._computeLightVertices = function (
     var closestVertex = gdjs.LightRuntimeObjectPixiRenderer._computeClosestIntersectionPoint(
       this._object,
       angle,
-      obstaclePolygons
+      obstaclePolygons,
+      halfOfDiag
     );
     if (closestVertex) {
       closestVertices.push({
@@ -430,7 +464,8 @@ gdjs.LightRuntimeObjectPixiRenderer.prototype._computeLightVertices = function (
     var closestVertexOffsetLeft = gdjs.LightRuntimeObjectPixiRenderer._computeClosestIntersectionPoint(
       this._object,
       angle + 0.0001,
-      obstaclePolygons
+      obstaclePolygons,
+      halfOfDiag
     );
     if (closestVertexOffsetLeft) {
       closestVertices.push({
@@ -441,7 +476,8 @@ gdjs.LightRuntimeObjectPixiRenderer.prototype._computeLightVertices = function (
     var closestVertexOffsetRight = gdjs.LightRuntimeObjectPixiRenderer._computeClosestIntersectionPoint(
       this._object,
       angle - 0.0001,
-      obstaclePolygons
+      obstaclePolygons,
+      halfOfDiag
     );
     if (closestVertexOffsetRight) {
       closestVertices.push({
--- a/Extensions/Lighting/lightruntimeobject-pixi-renderer.js
+++ b/Extensions/Lighting/lightruntimeobject-pixi-renderer.js
@@ -47,6 +47,14 @@ gdjs.LightRuntimeObjectPixiRenderer = function (runtimeObject, runtimeScene) {
   /** @type {?PIXI.Graphics} */
   this._debugGraphics = null;
   this.updateDebugMode();
+
+  /** @type {gdjs.Polygon} */
+  this._lightBoundingPoly = new gdjs.Polygon();
+  for (var i = 0; i < 4; i++) {
+    this._lightBoundingPoly.vertices.push(
+      runtimeObject.getHitBoxes()[0].vertices[i]
+    );
+  }
 };
 
 gdjs.LightRuntimeObjectRenderer = gdjs.LightRuntimeObjectPixiRenderer; //Register the class to let the engine use it.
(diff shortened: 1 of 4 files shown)


Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -17,10 +18,11 @@ gdjs.LightRuntimeObjectPixiRenderer = function (runtimeObject, runtimeScene) {
     objectColor[2] / 255,
   ];
 
-  /** @type {?PIXI.Texture} */
-  this._texture = runtimeObject.getPIXITexture();
-  this._center = new Float32Array([runtimeObject.x, runtimeObject.y]);
+  /** @type {string} */
```
HarsimranVirk, line 21: This should be `?string`. I'll change it in the next commit.

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -36,7 +38,7 @@ gdjs.LightRuntimeObjectPixiRenderer = function (runtimeObject, runtimeScene) {
 
   /** @type {?PIXI.Mesh} */
   this._light = null;
-  this._updateMesh();
```
HarsimranVirk, line 39: Not a private function anymore, because it needs to be called for hot-reloading, other than that there shouldn't really be any instance of this function being called. What do you suggest @4ian?

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -112,11 +114,12 @@ gdjs.LightRuntimeObjectPixiRenderer._verticesWithAngleComparator = function (
 gdjs.LightRuntimeObjectPixiRenderer._computeClosestIntersectionPoint = function (
   lightObject,
   angle,
-  polygons
+  polygons,
+  halfOfDiag
```
HarsimranVirk, line 118: `halfOfDiag` is not going to be sqrt(2)*radius anymore, because the boundary box could expand, and could easily become a rectangle at this point. So, this distance has to be passed now.

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -112,11 +114,12 @@ gdjs.LightRuntimeObjectPixiRenderer._verticesWithAngleComparator = function (
...
 ) {
   var centerX = lightObject.getX();
   var centerY = lightObject.getY();
-  var halfOfDiag = Math.sqrt(2) * lightObject.getRadius();
+  //var halfOfDiag = Math.sqrt(2) * lightObject.getRadius();
```
HarsimranVirk, line 122: I'll remove this comment, it was here only for a bit of clarity.

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -261,17 +268,20 @@ gdjs.LightRuntimeObjectPixiRenderer.prototype._updateDebugGraphics = function ()
...
       .moveTo(vertices[0], vertices[1])
-      .lineTo(vertices[i], vertices[i + 1]);
-    if (i !== vertices.length - 2) {
-      this._debugGraphics.lineTo(vertices[i + 2], vertices[i + 3]);
-    } else {
-      this._debugGraphics.lineTo(vertices[2], vertices[3]);
-    }
+      .lineTo(vertices[i], vertices[i + 1])
+      .moveTo(vertices[0], vertices[1])
+      .lineTo(vertices[lastX], vertices[lastY]);
```
HarsimranVirk, line 284: I don't know why this order works, but basically if I call `moveTo` before `lineTo` inside the for loop, I could see those weird lines. With this order, ie, `lineTo` before `moveTo` everything works correctly 🤔.

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -391,11 +401,43 @@ gdjs.LightRuntimeObjectPixiRenderer.prototype._computeLightVertices = function (
...
+  obstaclePolygons[0].vertices[2][1] = maxY;
+  obstaclePolygons[0].vertices[3][0] = minX;
+  obstaclePolygons[0].vertices[3][1] = maxY;
+
+  var halfOfDiag = Math.max(
+    Math.hypot(this._object.x - minX, this._object.y - minY),
+    Math.hypot(maxX - this._object.x, this._object.y - minY),
+    Math.hypot(maxX - this._object.x, maxY - this._object.y),
+    Math.hypot(this._object.x - minX, maxY - this._object.y)
+  )
```
HarsimranVirk, line 437: Now, since the expansion of the bounding box is not equal (ie, maybe only the `minX` changes and the bbox becomes larger from the left side but the right size is still equal to what it was before), so the center of the light is not necessarily the center of the rectangle. So we find the distance of all 4 vertices from the center of the light (also, I now realize that they are not really the diagonals) and take the maximum, which is then passed to `_computeClosestIntersectionPoint`. Maybe we should name this something like `maxLengthToBboxVertices` and not `halfOfDiag` since they aren't diagonals?

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -391,11 +401,43 @@ gdjs.LightRuntimeObjectPixiRenderer.prototype._computeLightVertices = function (
...
+  }
+
+  obstaclePolygons[0].vertices[0][0] = minX;
+  obstaclePolygons[0].vertices[0][1] = minY;
+  obstaclePolygons[0].vertices[1][0] = maxX;
+  obstaclePolygons[0].vertices[1][1] = minY;
+  obstaclePolygons[0].vertices[2][0] = maxX;
+  obstaclePolygons[0].vertices[2][1] = maxY;
+  obstaclePolygons[0].vertices[3][0] = minX;
+  obstaclePolygons[0].vertices[3][1] = maxY;
```
HarsimranVirk, line 430: Is changing the hitbox vertices of light objects like this correct? Or should we create a new `gdjs.Polygon` and use that instead of modifying this?

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
4ian: Well if it's called from `updateFromObjectData`, it can still be private no? :) `updateFromObjectData` is public, but not necessarily what it is using.

4ian: Ah I may have confused the pixi renderer and the object. Should be public then, because you expect ALL renderers to have this if you're calling it externally.

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
4ian: Thanks :)

Extensions/Lighting/lightruntimeobject.js
```diff
@@ -201,8 +186,8 @@ gdjs.LightRuntimeObject.prototype.getDebugMode = function () {
 
 /**
  * Returns PIXI.Texture if it exists, null otherwise.
- * @returns {?PIXI.Texture} the texture, if any, null otherwise.
+ * @returns {string} the path of texture.
```
4ian, line 189: This will be ?string too :)

Extensions/Lighting/lightruntimeobject.js
```diff
@@ -201,8 +186,8 @@ gdjs.LightRuntimeObject.prototype.getDebugMode = function () {
 
 /**
  * Returns PIXI.Texture if it exists, null otherwise.
```
4ian, line 188: Update the comment :)

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
4ian: Oops acutally no, this variable is holding either null or a PIXI.Texture (we're in the renderer). So `?PIXI.Texture` was valid.

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
4ian: Yep we should rename this variable. We can still consider this to be some half diagonal value of an hypothetic bounding square, so maybe `boundingSquareHalfDiagonal`?

Extensions/Lighting/lightruntimeobject.js
4ian: Actually no sorry, this string is always defined, it's just empty when not defined.

Extensions/Lighting/lightruntimeobject.js
```diff
@@ -33,19 +33,10 @@ gdjs.LightRuntimeObject = function (runtimeScene, lightObjectData) {
   this._debugMode = lightObjectData.content.debugMode;
 
   /** @type {?PIXI.Texture} */
```
4ian, line 35: @type {string}

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
4ian: Sorry I'm late to this reply - I think it might be safer to make a duplication of the polygon, otherwise we end up changing the hitbox of the light according to the presence or not of obstacles... would be weird.

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
4ian: Alright let's assume it's better now then!

Extensions/Lighting/lightruntimeobject.js
```diff
@@ -90,14 +81,8 @@ gdjs.LightRuntimeObject.prototype.updateFromObjectData = function (
...
-    this._texture =
-      newObjectData.content.texture === ''
-        ? null
-        : this._runtimeScene
-            .getGame()
-            .getImageManager()
-            .getPIXITexture(newObjectData.content.texture);
-    this._renderer.updateTexture();
+    this._texture = newObjectData.content.texture;
+    this._renderer.updateMesh();
```
4ian, line 85: You're not calling updateTexture?

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -391,11 +400,43 @@ gdjs.LightRuntimeObjectPixiRenderer.prototype._computeLightVertices = function (
...
+
+  var boundingSquareHalfDiag = Math.max(
+    Math.hypot(this._object.x - minX, this._object.y - minY),
+    Math.hypot(maxX - this._object.x, this._object.y - minY),
+    Math.hypot(maxX - this._object.x, maxY - this._object.y),
+    Math.hypot(this._object.x - minX, maxY - this._object.y)
```
4ian, line 435: Slight optimisation to avoid computing 4 times the square root instead of 1. Don't use Math.hypot, instead compute the values before taking the square root with Math.pow. Example, do for the first line just:
Math.pow(this._object.x - minX, 2) + Math.pow(this._object.y - minY, 2)

(this is what Math.hypot is doing internally), then take the max (Math.max) and finally Math.sqrt the result.

This works because `0 < x < y <=> sqrt(x) < sqrt(y)`

4ian (commented):
Sorry got sidetracked by other reviews! I've added a few comments but it looks good otherwise!
We'll merge this once you've addressed them.

In terms of performance, take a look at minimizing the number of calls to Math.sqrt/Math.hypot/Math.pow (or manually a*a) as it's fairly expensive when called repeatedly (for example, work with squared distance as much as possible, and take the square root *only* at the last step).

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
4ian: Note that I'm a bit sad to have to create a new `gdjs.Polygon` every time in this function, but I think it's much much better than modifying the object hitbox. Conceptually, the renderer of an object should not influence the hitbox when rendering it.

Extensions/Lighting/lightruntimeobject.js
HarsimranVirk: I'll have to update the entire mesh to update texture, since I'll have to swap the shader in this case.

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
HarsimranVirk: Actually, we don't need to create a new instance everytime, I guess we can have only one instance and keep modifying it. (Similar to float32 arrays for buffer)

4ian: Yes totally! We can store this "light bounding polygon" in the light object or pixi runtime renderer :)

Extensions/Lighting/lightruntimeobject.js
4ian: Sounds good then!

HarsimranVirk:
@4ian this dynamic polygon fix has made the light objects really good 😬. Earlier, having "floors" as light obstacles was a disaster, as the "floor" can be really long. Now, since the bounding box is dynamic, this case is handled as well :)

Screenshot (2)

Note that even if the bounding box is really large, the number of rays remain the same. So, the only extra performance cost here is the resizing of the polygon :)

4ian:
> this dynamic polygon fix has made the light objects really good 😬. Earlier, having "floors" as light obstacles was a disaster, as the "floor" can be really long. Now, since the bounding box is dynamic, this case is handled as well :)

Indeed, I've never thought about this case - while it's very common.
Note that there is still a case where things can be glitching: when an obstacle is partially inside another. In this case, the algorithm is not properly seeing the vertice formed by the intersection of the hitboxes. Unless you can find of a solution, we'll leave this as it is for now I guess.
(With this bugfix, you've avoided this issue by enlarging the bounding box to always include the obstacles, so no unseen intersections anymore.)

> Note that even if the bounding box is really large, the number of rays remain the same. So, the only extra performance cost here is the resizing of the polygon :)

We indeed keep the same number of rays which is great (performance will be limited by the number of objects and the complexity of their hitboxes, as before). So seems this "bugfix" has no adverse consequence on performance :)

Bouh:
Can you add the button for Add lighting layer next to the button for add a layer on left please ?

image

4ian:
I asked for the opposite a while ago 😅 Was because of small screens. Otherwise we can use a responsive component.

HarsimranVirk:
I guess we can use a [split button](https://material-ui.com/components/button-group/#split-button) in this case?

4ian:
>  Let's do that in another PR (not to add unrelated ui stuff in this PR), we might as well define a new ui element to be used in IDE 😬.

Yes let's do that in another PR! :)

4ian:
Merged! :)

PR #1932, opened 2020-08-20
Normal map prototype

HarsimranVirk:
Early stage:

- [x] Declare abstract methods such as `getNormalMapRendererObject`.

- [ ] Get render texture of normal map world.

- [ ] Add a new RuntimeObject and `gdjs.LightsManager` to track position of lights.

- [ ] Define and apply filter for diffused rendering.


Extensions/Lighting/forwarddiffusedRenderFilter.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * A pixi filter for diffused rendering. 
+ * @param {gdjs.RuntimeSceneRenderer} runtimeSceneRenderer 
+ */
+gdjs.ForwardDiffusedRenderFilter = function (runtimeSceneRenderer) {
```
HarsimranVirk, line 5: @4ian how do I import this class? It doesn't fit as an object, behavior or effect, how do I use JsExtension in this case?

GDJS/Runtime/runtimeobject.js
```diff
@@ -337,6 +338,10 @@ gdjs.RuntimeObject.prototype.getRendererObject = function() {
     return undefined;
 };
 
+gdjs.RuntimeObject.prototype.getNormalMapRendererObject = function() {
+    return undefined;
```
4ian, line 342: ubernit: I prefer to return `null` because undefined can mean "you accessed a property that is not even defined, surely something wrong" while `null` is explictly meaning "this is representing a valid value, but this value is null, i.e: there is no normal map object here".

GDJS/Runtime/runtimeobject.js
```diff
@@ -318,6 +318,7 @@ gdjs.RuntimeObject.prototype.deleteFromScene = function(runtimeScene) {
  */
 gdjs.RuntimeObject.prototype.onDestroyFromScene = function(runtimeScene) {
     var theLayer = runtimeScene.getLayer(this.layer);
+
```
4ian, line 321: You'll need to adapt every place where `getRendererObject`/`getRenderer` is used to optionally do the same with the normal map object. This is: `onDestroyFromScene`, `setZOrder`, `setLayer`.

Extensions/Lighting/forwarddiffusedRenderFilter.js
4ian: As it's part of an extension, it should be included by the extension whenever needed, which is surely whenever one of the new "light" object is used. As you don't have any new "light" object for now, consider either:
- "hacking" you way by adding `addIncludeFile` to some action, condition or existing object. Don't forget to put a big TODO
- OR maybe this filter should be part of the game engine itself when using pixi? In this case, add it in `GDJS/Runtime/pixi-renderers` and list it in `ExporterHelper::AddLibsInclude` in ExporterHelper.cpp. Though this means every game will include it even if not used, so maybe less ideal... but maybe we'll still need this because the rendering of normal map lighting will be tightly integrated in the pixi game engine.

PR #1938, opened 2020-08-24
Add support for customizable keyboard shortcuts

NilayMajorwar:
This PR adds support for customizable keyboard shortcuts to both the web and desktop app. Every shortcut is linked to a corresponding command, and the default shortcuts can be overridden by the user in a new "Keyboard Shortcuts" tab in preferences dialog.


newIDE/app/src/CommandPalette/CommandPalette/AutocompletePicker.js
```diff
@@ -35,14 +53,32 @@ const AutocompletePicker = (
...
+    }
+  };
+
   const getItemText = (item: Item) => {
     if (item.text) return item.text;
-    else if (item.displayText) return props.i18n._(item.displayText);
+    else if (item.name) {
+      const { area, displayText } = commandsList[item.name];
+      const areaText = commandAreas[area];
+      return props.i18n._(areaText) + ': ' + props.i18n._(displayText);
```
4ian, line 75: nit: what do you think of using "〉" instead of ":"? Some languages might not use a colon to separate words :)

newIDE/app/yarn.lock
```diff
@@ -1653,410 +1653,410 @@
...
-    "@pixi/utils" "5.3.0"
-
-"@pixi/loaders@5.3.0":
-  version "5.3.0"
-  resolved "https://registry.yarnpkg.com/@pixi/loaders/-/loaders-5.3.0.tgz#44424edee635e1d6b597373f30181606f78b7f02"
-  integrity sha512-g85GUuGWjJmh59J8gVjcIIy9rz+Zurj+cgVWb1oGTmpAVn2q19VX9Wks167SW+vLlng4UWJyjOV91pKnW7Ys9g==
-  dependencies:
-    "@pixi/core" "5.3.0"
-    "@pixi/utils" "5.3.0"
+"@pixi/accessibility@5.3.3":
```
4ian, line 1656: Is this normal that this yarn.lock is modified? 🤔

newIDE/app/src/KeyboardShortcuts/ShortcutsList.js
```diff
@@ -0,0 +1,120 @@
...
+                .filter(commandName => !commandsList[commandName].noShortcut)
+                .find(otherCommandName => {
+                  if (otherCommandName === commandName) return false;
+                  const otherShortcut =
+                    props.userShortcutMap[otherCommandName] ||
+                    defaultShortcuts[otherCommandName] ||
+                    '';
+                  if (shortcutString !== otherShortcut) return false;
+                  if (shortcutString === '') return false;
+                  console.log(otherCommandName);
```
4ian, line 87: Extra console.log here

newIDE/app/src/KeyboardShortcuts/ShortcutsList.js
```diff
@@ -0,0 +1,120 @@
...
+              const clashingCommandName = Object.keys(commandsList)
+                .filter(commandName => !commandsList[commandName].noShortcut)
+                .find(otherCommandName => {
+                  if (otherCommandName === commandName) return false;
+                  const otherShortcut =
+                    props.userShortcutMap[otherCommandName] ||
+                    defaultShortcuts[otherCommandName] ||
+                    '';
+                  if (shortcutString !== otherShortcut) return false;
+                  if (shortcutString === '') return false;
```
4ian, line 86: In this logic, it seems that the complexity is more important than what it could be. 
You iterate on all commands (well, on all areas, then on each commands of this area, so at the end it's on all commands). For each command, you get its shortcutString (so far so good).
Then inside this loop, you're iterating again on the commandsList, computing for each one its shortcutString.

At the end, this means that you're roughly doing `n^2` the work of computing the shortcut strings and comparing them. When we'll have 200 commands, this means 40000 operations, which start to make a lot for some low end or mobile devices :)

Instead, consider the approach of using an object as a map of used shortcutString. In pseudo code:
```
const shortcutStringToCommands = {};
for each command:
  // Get userShortcut and defaultShortcut as before
  const shortcutString = userShortcut || defaultShortcut
  shortcutStringToCommands[shortcutString] = (shortcutStringToCommands[shortcutString] || []).concat(command);
```

This is doing a single loop on commands, so complexity is `n`. Better, you can put this in a separate function outside of the render.

Then in the render, you iterate on commands again, and for each you look into `shortcutStringToCommands`, with the shortcutString of the command. If the array contains just your command, you're fine. If it contains something else, you have one (or more!) clashing commands :) 

At the end, this is done in roughly `2*n` operations (2 separate loops), so even with 200 commands, we'll have 400 operations to do - it should be more than fast enough even for the future and low end devices :)

4ian:
Added a few comments but seems we're on a good path!

newIDE/app/yarn.lock
NilayMajorwar: Nope :sweat_smile: - I've removed those changes now. I should've noticed that before committing, but anyway, learnt about file-specific git commands by reverting this :D

newIDE/app/src/KeyboardShortcuts/ShortcutsList.js
```diff
@@ -0,0 +1,135 @@
...
+      const defaultShortcut = defaultShortcuts[name] || '';
+      const shortcutString = userShortcut || defaultShortcut;
+      if (shortcutString === '') return;
+      shortcutStringToCommands[shortcutString] = (
+        shortcutStringToCommands[shortcutString] || []
+      ).concat(name);
+    });
+
+  return [areaWiseCommands, shortcutStringToCommands];
+};
```
NilayMajorwar, line 48: I've extracted the area-wise sorting and shortcut->commands mapping generation into this function. It creates two objects `areaWiseCommands` and `shortcutStringToCommands`, and then proceeds to fill them with a single loop over the commands list, and works properly.

The only thing I'm thinking about is that this function seems to do two quite unrelated things at once. Separating into two functions is cleaner but then we'll have to loop over the commands list twice - complexity is still linear, so should I split the function?

NilayMajorwar (commented):
I've made the fixes you suggested - please have a look!

NilayMajorwar:
I've added a `DismissableAlertMessage` for the command palette hint in the shortcuts list (which came along with a flag in "Hints and Explanations" tab in preferences). Looks like this:

Screenshot_20200825_192647

Is this what you had in mind?

4ian:
Yeah this looks good to me! 👍

4ian:
This looks good to me, merging now!🎉
Again huge thanks for working on this, this will be super useful for power users that will now benefit from a speed up when using GDevelop thanks to these customizable shortcut. I believe it will be super helpful for preview, for quickly jumping to an object/tab using the command palette, and I think we'll be able to add even more commands in the future! (I can almost see GDevelop being able to be entirely keyboard driven with commands :))

Very nice work, the code is of great quality! As the last time, I appreciate your attention to details when working on this 👏

PR #1942, opened 2020-08-25
Add I18n to context menu and electron menu

Bouh:
Todo list:
- [x] Electron menu
- [x] SceneEditor
- [x] EventsBasedBehaviorsList
- [x] EventsFunctionsList
- [x] EventsSheet
- [x] ClosableTabs
- [x] SceneEditor/Toolbar
- [x] ResourcesList
- [x] ProjectManager
- [x] ObjectList
- [x] ObjectGroupList
- [x] SpriteEditor
- [x] InstancePropertiesEditor
- [x] InstructionParametersEditor

### DIFF
--- a/newIDE/app/src/ProjectManager/index.js
+++ b/newIDE/app/src/ProjectManager/index.js
@@ -140,7 +140,7 @@ type ItemProps = {|
   onMoveUp: () => void,
   canMoveDown: boolean,
   onMoveDown: () => void,
-  buildExtraMenuTemplate?: () => Array<MenuItemTemplate>,
+  buildExtraMenuTemplate?: (i18n: I18nType) => Array<MenuItemTemplate>,
   style?: ?Object,
 |};
 
@@ -188,7 +188,7 @@ class Item extends React.Component<ItemProps, {||}> {
                 }}
                 primaryText={label}
                 displayMenuButton
-                buildMenuTemplate={() => [
+                buildMenuTemplate={(i18n: I18nType) => [
                   {
                     label: i18n._(t`Edit`),
                     click: () => this.props.onEdit(),
--- a/newIDE/app/src/ProjectManager/index.js
+++ b/newIDE/app/src/ProjectManager/index.js
@@ -1,5 +1,8 @@
 // @flow
 import { Trans } from '@lingui/macro';
+import { I18n } from '@lingui/react';
+import { type I18n as I18nType } from '@lingui/core';
+import { t } from '@lingui/macro';
 
 import * as React from 'react';
 import { List, ListItem } from '../UI/List';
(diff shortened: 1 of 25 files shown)


Bouh:
Not sure about the context menu on the scene ide. It's correct ?

I'll also add I18n on the others context menu.

newIDE/app/src/SceneEditor/index.js
```diff
@@ -1358,84 +1358,90 @@ export default class SceneEditor extends React.Component<Props, State> {
...
           {({ i18n }) => (
-            <InfoBar
-              show={this.state.showAdditionalWorkInfoBar}
-              identifier={this.state.additionalWorkInfoBar.identifier}
-              message={i18n._(this.state.additionalWorkInfoBar.message)}
-              touchScreenMessage={i18n._(
-                this.state.additionalWorkInfoBar.touchScreenMessage
-              )}
-            />
+            <div>
```
4ian, line 1361: Can you try a React.Fragment instead of a div?

newIDE/app/src/SceneEditor/index.js
```diff
@@ -1358,84 +1358,90 @@ export default class SceneEditor extends React.Component<Props, State> {
...
+              <ContextMenu
+                ref={contextMenu => (this.contextMenu = contextMenu)}
+                buildMenuTemplate={() => [
+                  {
+                    label: i18n._(
+                      this.state.selectedObjectNames.length
+                        ? 'Add an Instance of ' +
+                            shortenString(this.state.selectedObjectNames[0], 7)
+                        : ''
+                    ),
```
4ian, line 1379: This won't work, you must pass a `t` macro to `i18n._()`, not a string:
```suggestion
                    label: this.state.selectedObjectNames.length
                        ? i18n._(t`Add an Instance of ${shortenString(this.state.selectedObjectNames[0], 7)}`)
                        : ''
                    ),
```

Rule of thumb: i18n._ is always called with `t` macro, and variables inside `${}`. Never any logic inside.

newIDE/app/src/SceneEditor/index.js
```diff
@@ -1358,84 +1358,90 @@ export default class SceneEditor extends React.Component<Props, State> {
...
+                    click: () => this._onAddInstanceUnderCursor(),
+                    visible: this.state.selectedObjectNames.length > 0,
+                  },
+                  {
+                    label: i18n._(t`Insert a New Object`),
+                    click: () => this._createNewObjectAndInstanceUnderCursor(),
+                    visible: this.state.selectedObjectNames.length === 0,
+                  },
+                  {
+                    label: i18n._(
```
4ian, line 1389: Same as before

4ian (commented):
See comments about logic inside i18n._

Apart from this, this is how it should work yeah.
It's a bit sad that we have to use `i18n._` everywhere for labels. Instead, we could have been passing `t` macros as labels, and then in the implementation of Electron/Material-UI menus, we would have been calling `i18n._` for each label.

Anyway for now let's go ahead with this.
We can always codemod the entire codebase later by searching for `label: i18n._(t...)` and adapt context menu to either handle a string (legacy, as now) or a MessageDescriptor (i.e: a `t` macro).

Bouh:
It's good to me!

I left the sentences for the devs, like :
'Request update from external editor' or simulate the updates.

newIDE/app/src/ProjectManager/index.js
```diff
@@ -826,14 +833,14 @@ export default class ProjectManager extends React.Component<Props, State> {
...
+                          label: i18n._(t`Edit Scene Properties`),
                           enabled: true,
                           click: () => this._onOpenLayoutProperties(layout),
                         },
                         {
-                          label: 'Edit Scene Variables',
+                          label: i18n._(t`Edit Scene Variables`),
                           enabled: true,
                           click: () => this._onOpenLayoutVariables(layout),
                         },
```
Bouh, line 846: These labels don't work, help is needed, otherwise the other work fine, flow is totally broken the scan fail often i hope  i didn't miss anything else.

Bouh: Found :)

`import { I18n } from '@lingui/react';` was required in ContextMenu.js

Bouh:
There is no error in flow, or in the console, each text has been tested with another word already translated and work fine.
Let me know if there is a last bug, other it's ready to merge for me.
Tested more than twice, i'm pretty sure 💯

newIDE/app/src/ProjectManager/index.js
4ian: Ok :) I've seen that you passed i18n as argument to buildExtraMenuTemplate. Seems like a good idea I think. Would be interesting to follow this convention everywhere.

4ian:
I see you've done/are doing simplifications, let me know once you think you have something ready :)

Bouh:
I think it's good now 👍

newIDE/app/src/UI/ClosableTabs.js
```diff
@@ -191,24 +194,28 @@ export function ClosableTab({
...
-                  label: 'Close all',
-                  click: onCloseAll,
-                },
-              ]}
-            />
+            <I18n>
+              {({ i18n }) => (
+                <ContextMenu
+                  ref={contextMenu}
+                  buildMenuTemplate={(i18n: I18nType) => [
```
4ian, line 201: Here and in other places, I see that you've both been importing `<I18n>` and pass `i18n` as the buildMenuTemplate parameter.

I think we have to choose between one or the other, but not have both.
To make the code simpler, I'm in favor of passing `i18n` as the parameter of `buildMenuTemplate`. This allows to avoid adding `<I18>` everywhere (apart from where you need `i18n` outside of a context menu).
It also go in the direction of one day having `label` taking just a `t` macro, without the need to call `i18n._` (leaving this job to ContextMenu.. but it implies extra work to make this work on Electron).

See my next comment about how to add it to ContextMenu for Electron and material :)

newIDE/app/src/UI/Menu/ContextMenu.js
```diff
@@ -44,14 +44,17 @@ class MaterialUIContextMenu extends React.Component {
...
+          top: this.state.anchorY,
+        }}
         anchorReference={'anchorPosition'}
         onClose={this._onClose}
         TransitionComponent={Fade}
         {...this.menuImplementation.getMenuProps()}
       >
         {this.menuImplementation.buildFromTemplate(
-          this.props.buildMenuTemplate()
+          this.props.buildMenuTemplate(this.props.i18n)
```
4ian, line 57: Was this working? I've not seen i18n being passed to ContextMenu props. Anyway, let's *not* do that. Instead, "simply" add `<I18n>` around `<Menu` here in the MaterialUIContextMenu.

This allows to avoid adding `<I18n>` everywhere in the codebase : the context menu will pass it and does not need it as a prop :)

newIDE/app/src/UI/Menu/ContextMenu.js
```diff
@@ -68,7 +71,9 @@ class ElectronContextMenu extends React.Component {
   }
 
   open = (x, y) => {
-    this.menuImplementation.buildFromTemplate(this.props.buildMenuTemplate());
+    this.menuImplementation.buildFromTemplate(
+      this.props.buildMenuTemplate(this.props.i18n)
+    );
     this.menuImplementation.showMenu({
```
4ian, line 77: If you read and apply my previous comment, you're now like "ok sure I added `<I18n>` inside MaterialUIContextMenu, but now how do I do it for ElectronContextMenu??".
Ideally in the future, we would be able to use a React hook for I18n: `const i18n = use18n();`... but for now we have to use what is called the version with a "render prop": `<I18n>{({i18n}) => ...}<I18n>`, which is a bit more painful to use in this case.

The solution? Make a ElectronContextMenuWrapper component and a ElectronContextMenu component.
Basically:

- ElectronContextMenuWrapper is what you export. It is a component that just do:
```
const ElectronContextMenuWrapper = (props) => {
  return <I18n>{({i18n}) => <ElectronContextMenu {...props} i18n={i18n} >}<I18n>;
}
```

(in other words, you pass all the props to the child component, and you add i18n.
- you also need to expose the `open` function. As we're trying to use React hooks now, use `useImperativeHandle` to expose functions to the outside world, and use a ref to the component:

```
const ElectronContextMenuWrapper = React.forwardRef((props, ref) => {
    const electronContextMenu = React.useRef(null);
    React.useImperativeHandle(ref, () => ({
      open: (x, y) => {
        if (electronContextMenu.current) electronContextMenu.current.open(x,y);
      },
    }));

  return <I18n>{({i18n}) => <ElectronContextMenu {...this.props} i18n={i18n} ref={electronContextMenu}>}<I18n>;
}
```
- You can now `export default (electron ? ElectronContextMenuWrapper : MaterialUIContextMenu);` and if everything is right, your ContextMenu will work both in the web-app and the electron app, giving `i18n` as parameter of `buildMenuTemplate`, so all the other components don't need to import `<I18n>`, which is nice to keep the code shorter :) 

- You're lucky because you don't need to add Flow typing because I never added them to ContextMenu :p I'll do that later though, for such refactoring it's very good to see if you've forgot a prop somewhere. For now, I'll put my trust in you testing that it works everywhere ;)

4ian:
Still a bit of work for the ContextMenu, but we're getting there :)

Bouh:
Change are done, [my flow crash everytime ](https://github.com/facebook/flow/issues/6424)and consume all my ram because the previous flow.exe still alive and another one is created because the first have crash, i check manually everything in the app and webapp then, if an crash come i see it directly.
I learned to do without flow.

4ian:
Sorry about the bad experience with Flow 😬 Maybe a newer version would reduce the RAM usage which might be why it's crashing so often... not great.

Anyway, you can see in Semaphore UI the result from Flow. There is an extra `i18n` parameter here:
image
and in SceneEditor too: 
image

either remove this extra parameter or add I18n to `TagsButton`/`List` components and update their props typing :)

PR #1946, opened 2020-08-27
Feature/rtl

denim2x:
image

### DIFF
--- a/newIDE/app/src/UI/Theme/DefaultTheme/index.js
+++ b/newIDE/app/src/UI/Theme/DefaultTheme/index.js
@@ -233,4 +233,5 @@ const theme = {
   muiTheme,
 };
 export type Theme = $Exact<typeof theme>;
+export const themeName = 'GDevelop default';
 export default theme;
--- a/newIDE/app/src/UI/Theme/index.js
+++ b/newIDE/app/src/UI/Theme/index.js
@@ -1,9 +1,12 @@
 // @flow
+import { isLtr } from '../../Utils/i18n/RtlLanguages';
+import DefaultTheme, {
+  themeName as defaultThemeName,
+  type Theme,
+} from './DefaultTheme';
 import DarkTheme from './DarkTheme';
-import DefaultTheme from './DefaultTheme';
 import NordTheme from './NordTheme';
 import SolarizedDarkTheme from './SolarizedDarkTheme';
-import { type Theme } from './DefaultTheme';
 import './Global.css';
 
 // To add a new theme:
@@ -11,13 +14,31 @@ import './Global.css';
 // * import it at the top of the file
 // * add it below:
 export const themes = {
-  'GDevelop default': DefaultTheme,
+  [defaultThemeName]: DefaultTheme,
   Dark: DarkTheme,
   Nord: NordTheme,
   'Solarized Dark': SolarizedDarkTheme,
 };
 
-export const getTheme = (themeName: string): Theme =>
-  themes[themeName] || themes['GDevelop default'];
+export function getTheme({ themeName, language }: {| themeName: string, language: string |}): Theme {
+  // Get `Theme` associated to `themeName`
+  let theme = themes[themeName];
+
+  // If unavailable, give notice in the console and revert to default `Theme`
+  if (!theme) {
+    console.warn(`Theme '${themeName}' is unavailable; '${defaultThemeName}' is used`);
+    theme = themes[defaultThemeName];
+  }
+
+  const { gdevelopTheme, muiTheme } = theme;
+  return { gdevelopTheme, muiTheme: convertToRtl(muiTheme, language) };  
+};
 
 export type GDevelopTheme = $PropertyType<Theme, 'gdevelopTheme'>;
+
+type MuiTheme = $Exact<typeof DefaultTheme.muiTheme>;
+
+function convertToRtl(muiTheme: MuiTheme, language: string): MuiTheme {
+  const ltr = isLtr(language);
+  return { ...muiTheme, direction: ltr ? 'unset' : 'rtl' };
+}
--- a/newIDE/app/package.json
+++ b/newIDE/app/package.json
@@ -43,6 +43,7 @@
     "element-closest": "2.0.2",
     "firebase": "^6.1.0",
     "fontfaceobserver": "2.0.13",
+    "jss-rtl": "^0.3.0",
     "keen-tracking": "1.1.3",
     "lodash": "4.17.4",
     "node-require-function": "^1.2.0",
--- a/newIDE/app/src/MainFrame/Providers.js
+++ b/newIDE/app/src/MainFrame/Providers.js
@@ -2,6 +2,7 @@
 import * as React from 'react';
 import DragAndDropContextProvider from '../UI/DragAndDrop/DragAndDropContextProvider';
 import { ThemeProvider } from '@material-ui/styles';
+import { StylesProvider, jssPreset } from '@material-ui/core/styles';
 import { getTheme } from '../UI/Theme';
 import UserProfileProvider from '../Profile/UserProfileProvider';
 import Authentification from '../Utils/GDevelopServices/Authentification';
@@ -25,6 +26,12 @@ import {
 import GDevelopThemeContext from '../UI/Theme/ThemeContext';
 import { UnsavedChangesContextProvider } from './UnsavedChangesContext';
 import { CommandsContextProvider } from '../CommandPalette/CommandsContext';
+import { create } from 'jss';
+import rtl from 'jss-rtl';
+
+const jss = create({
+  plugins: [...jssPreset().plugins,  rtl()],
+});
 
 type Props = {|
   authentification: Authentification,
@@ -58,11 +65,12 @@ export default class Providers extends React.Component<Props, {||}> {
           <PreferencesProvider disableCheckForUpdates={disableCheckForUpdates}>
             <PreferencesContext.Consumer>
               {({ values }) => {
-                const theme = getTheme(values.themeName);
+                const theme = getTheme({ themeName: values.themeName, language: values.language });
                 return (
                   <GDI18nProvider language={values.language}>
                     <GDevelopThemeContext.Provider value={theme.gdevelopTheme}>
-                      <ThemeProvider theme={theme.muiTheme}>
+                      <StylesProvider jss={jss}>
+                        <ThemeProvider theme={theme.muiTheme}>
                         <UserProfileProvider
                           authentification={authentification}
                         >
@@ -94,7 +102,8 @@ export default class Providers extends React.Component<Props, {||}> {
                             )}
                           </I18n>
                         </UserProfileProvider>
-                      </ThemeProvider>
+                        </ThemeProvider>
+                      </StylesProvider>
                     </GDevelopThemeContext.Provider>
                   </GDI18nProvider>
                 );
--- a/newIDE/app/src/Utils/i18n/RtlLanguages.js
+++ b/newIDE/app/src/Utils/i18n/RtlLanguages.js
@@ -1,0 +1,18 @@
+// @flow
+
+const rtlLanguages = new Set([
+  /******* Hebrew scripts *******/
+  'he_IL'/* Hebrew */,
+  
+  /******* Arabic scripts *******/
+  'ar_SA'/* Arabic */,
+  'az_AZ'/* Azerbaijani */,
+  'ms_MY'/* Malay */,
+  'fa_IR'/* Persian */,
+  'ur_PK'/* Urdu */,
+]);
+
+export function isLtr(name: string) {
+  return !rtlLanguages.has(name);
+}
+


newIDE/app/public/index.html
```diff
@@ -9,6 +9,17 @@
...
 
     <style>
+      /**
+       * Whenever the currently selected language is written RTL, this
+       * class is added to specific components, at runtime.
+       *
+       * This class is referenced directly by the `Utils/i18n` module.
+       * Any other module shall use this class indirectly, via `i18n.css`.
+       */
+      .is-rtl {
```
arthuro555, line 19: That feels really dirty. Can't you add that through some css to pack via webpack than by adding it to the index.html?

denim2x: Fine, the entire CSS code moves to a different file, used by webpack during build.

4ian: Arthuro555 careful about your wording ;)
For a first time contributor, it's quite rude to get "this feels really dirty" as the very first comment. Choose instead to explain why you think there is a better way: "Seems like we're adding code to index.html, while we want to keep as little code as possible inside to have a near instant display of something on the page (the background GDevelop logo, while the rest is possible). Would it be possible [...]".

4ian:
Thanks for opening this PR :)

A few comments on the overall approach:
- I see a lot of components where you add `className` as a prop. We usually avoid this pattern because this break the encapsulation of the component: the component should render by itself and the parent should not be able to pollute its styling. In other words, a Button is responsible for rendering itself according to a few props (the label, icon, etc...) and, by design, there should be no way to override the styling. So we can remove these className :) 
- You've "augmented" `i18n` to add a `i18n.css` property. While it works, I'm not super comfortable changing something that is part of a library. I remember many horror stories about debugging stuff that were caused by new properties being added inside objects that were not expecting them. In terms of static typing, if we add Typescript/Flow support at some point for I18n, this would break.
- Instead of manually adding a css class to all the UI component, that are most of the time just wrappers around material-ui components, can instead use `jss-rtl`? This is what is told in this page that I linked already in Trello: https://material-ui.com/guides/right-to-left/#3-jss-rtl

More precisely, if we follow this, we need to:
- Add `dir="rtl"` to body (for the direction of the native element).
- Call `createMuiTheme({ direction: 'rtl', ... });` when creating a theme. This means we must refactor theme creation: https://github.com/4ian/GDevelop/blob/35f019afa867467f416e506de84be08829080a08/newIDE/app/src/UI/Theme/DefaultTheme/index.js#L96
  - Maybe something to rework in the ThemeProvider: https://github.com/4ian/GDevelop/blob/35f019afa867467f416e506de84be08829080a08/newIDE/app/src/MainFrame/Providers.js#L65
- Finally, we would need to use, according to the doc, jss-rtl. We would add a new provider in the list of the app [Providers](https://github.com/4ian/GDevelop/blob/35f019afa867467f416e506de84be08829080a08/newIDE/app/src/MainFrame/Providers.js#L65): 

By doing this, material-ui will "automatically" flip all components provided by material ui. We can then adapt manually the remaining components. Using `createStyles` will automatically get the flip applied to styles, unless you specify `flip: false` in the styles. Example in material-ui doc:
Screenshot 2020-08-28 at 00 31 12

I know this is a quite different approach from what you made, but as we're using material-ui for all the interface, it makes sense to leverage their recommended approach as much as possible, as we'll be getting "for free" the proper rtl alignement for all the material-ui components, doing potentially 90% of the job for us without having to add `i18n.css` everywhere.

Let me know what you think.

denim2x:
Now `Providers` has `ThemeProvider` nested within `StylesProvider`. The `theme` is being rebuilt with some components styled according to the language directionality, based the current `i18n` instance.

denim2x:
Would a `ThemeProvider` with some style overrides be useful here?

4ian:
I'm not sure it would work but maybe? But we can as well see how we could refactor the existing theme so that they can have a `rtl` boolean passed when constructing the theme? If you think a nested theme provider would work, you can give it a try though (just unsure about performance of nesting theme providers). Might be a a solution to avoid "polluting" the themes with rtl concerns.

4ian:
Seems like there is nothing to review?

newIDE/app/src/UI/Theme/DefaultTheme/index.js
```diff
@@ -118,19 +134,24 @@ const muiTheme = createMuiTheme({
     },
   },
   overrides: {
```
denim2x, line 136: Alternative:
```js
...
overrides: compose({
  MuiTypography: {
    ...
  },
  ...
}, i18n),
...
```

4ian:
Is this ready for review? :) Do you have something working or is this still WIP?

denim2x:
I've pushed the final commits
`DefaultTheme` works OK
Other themes will work, following review

denim2x:
I've pushed the final commits
`DefaultTheme` works OK
Other themes will work, following review

denim2x:
I've pushed the final commits
`DefaultTheme` works OK
Other themes will work, following review

newIDE/app/src/Utils/Object.js
```diff
@@ -0,0 +1,396 @@
+﻿/**
```
4ian, line 1: There are a lot of functions here, but only a few are used. Could we remove all the ones that are not used?

newIDE/app/src/Utils/Stash.js
```diff
@@ -0,0 +1,32 @@
...
+  
+  stash.peek = (key, fallback) => {
+    return data.has(key) ? data.get(key) : fallback;
+  };
+  return stash;
+  
+  function stash(key, value) {
+    if (arguments.length < 2) {
+      return data.get(key);
+    }
```
4ian, line 27: These kind of functions that do different things according to arguments make static typing difficult, and also I would say a bit more difficult to read for the reader. Instead I would favour two separate functions `get`/`set` - or better, could we remove this utility entirely and use a plain old JS object, or a ES6 `Map`?

newIDE/app/src/Utils/i18n/language.js
```diff
@@ -0,0 +1,49 @@
...
+    return this._name;
+  }
+  
+  static from(lang) {
+    return typeof lang === 'object' ? lang : new Language(lang);
+  }
+  
+  static is_ltr(lang) {
+    return Language.from(lang)?.is_ltr;
+  }
```
4ian, line 48: I find this approach overly complex: there is a constructor, a static constructor (from), 2 `typeof` (that is better to be avoided in code to favor static typing), a static method.

It seems that the class could be replaced by a call to `Set.has`:

```suggestion
export const isLtr = (lang: string) => !rtlLanguages.has(lang);
```

newIDE/app/src/UI/Theme/index.js
```diff
@@ -1,23 +1,61 @@
 // @flow
+import {
+  select,
+  blend as Install,
+} from '../../Utils/Object';
+import Language from '../../Utils/i18n/language';
+import { stash } from './domain';
```
4ian, line 7: This stash is a global variable. Is there a way we could avoid:
- using a `Stash`. Can we replace it with a single variable?
- using a variable entirely? :) Could we just call `isLtr(lang)` whenever we need it?

newIDE/app/src/UI/Theme/i18n.js
```diff
@@ -0,0 +1,42 @@
...
+  label: withDirection('label'),
+};
+
+const withOrder = Style({
+  get order() {
+    return stash('language-ltr') ? 'unset' : (styles.order?.rtl ?? 100);
+  },
+});
+export const icon = {
+  root: withOrder('root'),
```
4ian, line 21: This seems complex. Let's remove `Style` and just declare an inline function where needed.

newIDE/app/src/UI/Theme/i18n.js
```diff
@@ -0,0 +1,42 @@
...
+const withDirection = Style({
+  get direction() {
+    return stash('language-ltr') ? 'unset' : 'rtl';
+  },
+});
+
+export const text = {
+  root: withDirection('root'),
+  label: withDirection('label'),
+};
```
4ian, line 13: Let's inline the function in text and remove `withDirection`

newIDE/app/src/UI/Theme/DefaultTheme/index.js
```diff
@@ -13,8 +17,20 @@ const systemSelectionColor = '#4c92ff'; //OS X selection
 
 // Use the fonts provided by the operating system(s) as possible.
 // If you update this font list, be sure to do it in all the other places using fonts in the codebase.
-const fontFamily =
-  "-apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen-Sans, Ubuntu, 'Helvetica Neue', Helvetica, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol'";
+const fontFamily = CSS.list(
```
4ian, line 20: I would rather not have this change in this PR, because if we change one list like this we should do it for all other similar lists in the codebase for consistency.

newIDE/app/src/UI/Theme/i18n.js
```diff
@@ -0,0 +1,42 @@
...
+
+const styles = {
+  order: {
+    rtl: 100,
+  },
+};
+
+export default styles;
+
+export const initial = {
```
4ian, line 32: Can we find a more descriptive name than `initial` for this object?

newIDE/app/src/UI/Theme/index.js
```diff
@@ -1,23 +1,61 @@
...
+import DefaultTheme, {
+  themeName as $default,
```
4ian, line 9: Let's avoid naming with a $, that implies some kind of special value à la jQuery. Instead, let's name if "defaultThemeName" - it's more descriptive.

4ian (commented):
Thanks for work on this PR. I've added a few comments about some parts that I find overly complex. I've seen usage of a global variable though that I think should be reconsidered to see if we can eliminate it.

denim2x:
I've made the corrections

newIDE/app/src/UI/Theme/index.js
```diff
@@ -1,23 +1,61 @@
...
+ * two _distinct_ `muiTheme` instances is returned; these two instances
+ * are exactly alike (i.e. have the exact same properties).
+ */
+const install = new Install(({ gdevelopTheme, muiTheme }: Theme): Theme => {
+  const muiThemes: [MuiTheme, MuiTheme] = [muiTheme, { ...muiTheme }];
+  return {
+    gdevelopTheme,
+    get muiTheme(): MuiTheme {
+      const ltr = domain['language-ltr'] ?? true;
+      return select(muiThemes, ltr);
```
4ian, line 33: can you replace all the `select` by a ternary?

newIDE/app/src/UI/Theme/index.js
```diff
@@ -1,23 +1,61 @@
...
+ * This ensures the `muiTheme` is correctly applied whenever language
+ * directionality is changed. Thus, based on `language-ltr`, one of
+ * two _distinct_ `muiTheme` instances is returned; these two instances
+ * are exactly alike (i.e. have the exact same properties).
+ */
+const install = new Install(({ gdevelopTheme, muiTheme }: Theme): Theme => {
+  const muiThemes: [MuiTheme, MuiTheme] = [muiTheme, { ...muiTheme }];
+  return {
+    gdevelopTheme,
+    get muiTheme(): MuiTheme {
```
4ian, line 31: I'm not a big fan of overriding accessors. It makes reasoning on the code harder as you have side effects that are not shown at the caller site. 
Let's see instead if we can do all the work in getTheme?

newIDE/app/src/UI/Theme/index.js
```diff
@@ -1,23 +1,61 @@
...
   Nord: NordTheme,
   'Solarized Dark': SolarizedDarkTheme,
-};
+});
 
-export const getTheme = (themeName: string): Theme =>
-  themes[themeName] || themes['GDevelop default'];
+export function getTheme({ themeName, language }: {| themeName: string, language: string |}): Theme {
+  // This boolean value indicates if the defined directionality of `language` is left-to-right
+  domain['language-ltr'] = isLTR(language);
```
4ian, line 51: Let's entirely avoid to store things in a global variable `domain`. It's not typed and the name is too generic to understand what can go inside.

newIDE/app/src/UI/Theme/i18n.js
```diff
@@ -0,0 +1,43 @@
...
+export const text = {
+  root: { root: directionSpec },
+  label: { label: directionSpec },
+};
+
+const orderSpec = {
+  get order() {
+    return domain['language-ltr'] ? 'unset' : (styles.order?.rtl ?? 100);
+  },
+};
```
4ian, line 19: I'm afraid the accessor make the code hard to understand, as they are implicitly called, and they rely on a global variable.
In the future, we may have more than one instance of the editor running, with possible different theme or even language. To keep this flexibility, we must not rely on any global variable.

Instead, I suggest a different approach: 
* No more accessors at all in the theme.
* Keep the muiTheme as usual...
* ...but make a function that takes a muiTheme and return the rtl equivalent (we can call it `convertToRtl`)
* in `getTheme`, returns the normal `Theme` (i.e: `{gdevelopTheme, muiTheme}` if `isLtr(language)` is false, and return `{gdevelopTheme, muiTheme: convertToRtl(muiTheme)}` otherwise.

newIDE/app/src/UI/Theme/i18n.js
```diff
@@ -0,0 +1,43 @@
...
+
+const styles = {
+  order: {
+    rtl: 100,
+  },
+};
+
+export default styles;
+
+export const formula = {
```
4ian, line 33: Avoid calling things with generic name like "formula", especially if you rename them later when importing them.
Note that this export and object might not be needed if you switch to a `convertToRtl` approach.

newIDE/app/src/UI/Theme/i18n.js
```diff
@@ -0,0 +1,43 @@
...
+  get order() {
+    return domain['language-ltr'] ? 'unset' : (styles.order?.rtl ?? 100);
+  },
+};
+
+export const icon = {
+  root: { root: orderSpec },
+};
+
+const styles = {
```
4ian, line 25: Is this object needed?

newIDE/app/src/UI/Theme/i18n.js
```diff
@@ -0,0 +1,43 @@
...
+};
+
+export const text = {
+  root: { root: directionSpec },
+  label: { label: directionSpec },
+};
+
+const orderSpec = {
+  get order() {
+    return domain['language-ltr'] ? 'unset' : (styles.order?.rtl ?? 100);
```
4ian, line 17: Avoid at all cost `?.` operator. Either a property exist on an object or don't.

newIDE/app/src/UI/Theme/index.js
```diff
@@ -1,23 +1,61 @@
...
+
+/**
+ * Enhances every Theme instance with a `muiTheme` property accessor.
+ *
+ * This ensures the `muiTheme` is correctly applied whenever language
+ * directionality is changed. Thus, based on `language-ltr`, one of
+ * two _distinct_ `muiTheme` instances is returned; these two instances
+ * are exactly alike (i.e. have the exact same properties).
+ */
+const install = new Install(({ gdevelopTheme, muiTheme }: Theme): Theme => {
```
4ian, line 27: Let's remove this Install/`blend` and make instead a check for the language rtl value in `getTheme`, and use a function dedicated to generate a rtl version of the muiTheme.

newIDE/app/src/Utils/Object.js
4ian: I know that you've been putting a lot of work in this, but I prefer that we entirely remove `Object.js`. I'm pretty sure we can implement the whole feature in a simpler way without them.
We'll add an utility function if really really needed - in which case it must be typed and limited to the file where it's used.

newIDE/app/src/UI/Theme/domain.js
```diff
@@ -0,0 +1,6 @@
+// @flow
```
4ian, line 1: As mentioned in other comment, let's entirely remove this

newIDE/app/src/UI/Theme/DefaultTheme/index.js
```diff
@@ -1,5 +1,8 @@
 // @flow
 import { createMuiTheme, darken, lighten } from '@material-ui/core/styles';
+import { compose } from '../../../Utils/Object';
+import { text, formula as i18n } from '../i18n';
```
4ian, line 4: as mentioned in other comments, let's entirely remove these and use a function `convertToRtl` that takes a muiTheme and returns the rtl-enabled muiTheme. We then use it in getTheme, and we should hopefully be done.

4ian (commented):
Thanks again for your work so far. I've added several comments because I think we can make the code again drastically simpler, avoiding lots of utility and untyped generic functional manipulations.
In general, avoid any untyped or overly generic code. No `.?` (static typing is here to help), no overly generic functional helper function, avoid accessor that are running logic "behind your back", don't use a global untyped `domain` variable.

Two additional questions:
- I read there: https://material-ui.com/guides/right-to-left/#2-theme that setting `direction: 'rtl'` on muiTheme is sufficient. Is there a real need for you to go and modify the MuiButton, MuiTab, etc...?
- It seems that https://material-ui.com/guides/right-to-left/#1-html was not done? If you're unsure how to do this, don't do anything and I'll make it myself. Otherwise, I think that exceptionally, modifying the `body` `dir` attribute in `getTheme` is ok. Which is a side effect... but fairly ok as it's a global and we can't avoid it.

PR #1948, opened 2020-08-28
Remove edit button in shortcut editor

arthuro555:
The edit button is kind of unintuitive and redundant. I made the component that shows the current shortcut clickable instead.


4ian:
I'll let @NilayMajorwar review this! :)

NilayMajorwar:
Simply making the chip clickable and removing the edit button is not completely okay I think, because then the user does not have any easy way to know the shortcuts are customizable. The user will have to hover over the chip, notice the cursor changing to a pointer, and then realize the shortcut is editable.

Perhaps we can add another dismissable hint to the shortcuts list, that roughly says something like "these are editable shortcuts, click on the chips to change stuff". What do you think?

Also, I think you missed something @arthuro555 - the `onClick` handler won't work for chips that have `disabled` prop set to true. I used the `disabled` prop to show the "No shortcut" chips as slightly faded, so with your change the user won't be able to set a shortcut for actions without shortcut set. *And* apparently Chip doesn't support the `tooltip` prop anyway :smile:

arthuro555:
Tbh I did this really quickly and meant to make it a draft PR.

I think the fact that it is in the properties dialog and that the shortcut has a little getting whiter animation when hovering make it clear that it is editable. But then keyboard shortcuts are your project, if you think your way is better I'll just close this PR.

> the onClick handler won't work for chips that have disabled prop set to true. I used the disabled prop to show the "No shortcut" chips as slightly faded, so with your change the user won't be able to set a shortcut for actions without shortcut set. And apparently Chip doesn't support the tooltip prop anyway 😄

I'll try to fix all of this later.

Bouh:
> I used the `disabled` prop to show the "No shortcut" chips as slightly faded, so with your change the user won't be able to set a shortcut for actions without shortcut set.

Exactly, the button is disabled.

With or without the edit button, i like both, i prefer without because for me it's more logic to click on an chip, (btw the chip look like an button so it's fine), chip and button are a bit similar, [it's clickable](https://material.io/components/chips).

New user or young peoples on a pc don't know really often which button is clickable, you can work on the best UX and UI, if the person isn't curious or have no experience it's simplier to an a edit button, atleast everyone will see it.

Bouh:
Also , on Chrome, the shortcut CTRL+W close the current tab.
Can we prevent default shortcut on the web editor ?

NilayMajorwar:
Just to clarify, I too like making the chip clickable and avoiding the explicit edit button. But to ensure all users know that the shortcuts are customizable, I think a dismissable hint should be added along with this change.

> Also , on Chrome, the shortcut CTRL+W close the current tab.
Can we prevent default shortcut on the web editor ?

No, those are "ultra"-reserved shortcuts that Chrome doesn't even *allow* JavaScript to intercept. Other such shortcuts are Ctrl+N, Ctrl+T, and probably some others. On the web app, Chrome makes it impossible to use those as shortcuts. I hope none of the default shortcuts clash with these ultra-reserved ones?

arthuro555:
Apparently the Chip does support some sort of tooltip with the title prop. I used that, and also replced button disabling with color changing:
image

 Tested with all themes always looks good.

arthuro555:
@NilayMajorwar Could you review this again please? The PR has been frozen for month now awaiting it :/

newIDE/app/src/KeyboardShortcuts/ShortcutsListRow.js
```diff
@@ -43,11 +42,10 @@ const ShortcutsListRow = (props: Props) => {
         <Chip
           style={styles.shortcutChip}
           label={props.shortcutString || <Trans>No shortcut</Trans>}
-          disabled={!props.shortcutString}
+          onClick={props.onEditShortcut}
+          title={props.i18n._('Edit shortcut')}
```
NilayMajorwar, line 46: You missed the `t` macro here, I think.

NilayMajorwar (commented):
Sorry for the (huge) delay :sweat_smile:. I had a lot of college stuff going on, so could not get back to this one.

I tried it out and it works great! You're right, the ripple animation makes it easier to notice that it's clickable and editable. A very minor issue: the cursor is a pointer on the chip as desired but changes back to default on the chip text, but maybe that's an MUI bug and isn't very significant anyway since clicking works properly.

arthuro555:
> the cursor is a pointer on the chip as desired but changes back to default on the chip text, but maybe that's an MUI bug and isn't very significant anyway since clicking works properly.

I don't really know how I could fix that. Looks like a MUI issue indeed.

@4ian I think this is ready to merge?

newIDE/app/src/KeyboardShortcuts/ShortcutsListRow.js
```diff
@@ -36,22 +35,21 @@ const ShortcutsListRow = (props: Props) => {
       />
       <ListItemSecondaryAction>
         {props.isClashing && (
-          <IconButton tooltip={t`This shortcut clashes with another action.`}>
+          <IconButton tooltip={props.i18n._(t`This shortcut clashes with another action.`)}>
```
4ian, line 38: tooltip takes a MessageDescriptor (i.e: a `t` macro), not a string: https://github.com/4ian/GDevelop/blob/8a8adf213ada4749bcf38a7b1d5965096357fcf5/newIDE/app/src/UI/IconButton.js#L45

newIDE/app/src/KeyboardShortcuts/ShortcutsListRow.js
```diff
@@ -36,22 +35,21 @@ const ShortcutsListRow = (props: Props) => {
...
+          color={props.shortcutString ? 'secondary' : 'default'}
         />
-        <IconButton onClick={props.onEditShortcut} tooltip={t`Edit shortcut`}>
-          <EditIcon />
-        </IconButton>
         {!props.isDefault && (
           <IconButton
             onClick={props.onResetShortcut}
-            tooltip={t`Reset to default`}
+            tooltip={props.i18n._(t`Reset to default`)}
```
4ian, line 52: Same, no i18n._ here

4ian:
👍

PR #1950, opened 2020-08-28
Add native dark mode support for Mac os

arthuro555:
Untested wrote it in 10 second on my phone
See https://www.electronjs.org/docs/tutorial/mojave-dark-mode-guide

### DIFF
--- a/newIDE/app/src/MainFrame/Preferences/PreferencesContext.js
+++ b/newIDE/app/src/MainFrame/Preferences/PreferencesContext.js
@@ -6,6 +6,8 @@ import { type EditorMosaicNode } from '../../UI/EditorMosaic';
 import { type FileMetadataAndStorageProviderName } from '../../ProjectsStorage';
 import { type ShortcutMap } from '../../KeyboardShortcuts/DefaultShortcuts';
 import { type CommandName } from '../../CommandPalette/CommandsList';
+import optionalRequire from '../../Utils/OptionalRequire';
+const electron = optionalRequire('electron');
 
 export type AlertMessageIdentifier =
   | 'default-additional-work'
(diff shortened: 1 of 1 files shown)


Bouh:
image

arthuro555:
Ok I fixed it, it runs fine now. It needs to be tested on Mac OS though, and I don't have a mac :/

arthuro555:
Still awaiting review/testing from a mac owner 👀

newIDE/app/src/UI/Theme/index.js
```diff
@@ -39,7 +55,7 @@ export function getTheme({
     console.warn(
       `Theme '${themeName}' is unavailable; '${defaultThemeName}' is used`
     );
-    theme = themes[defaultThemeName];
+    theme = themes[systemDefaultThemeName];
```
4ian, line 58: I'm not this this will work, as you changed here the "fallback" theme which is the theme to use in case the `themeName` is invalid and can not be find in the `themes` map.

Basically, this is an error case. It should never happen. It can happen if you upgrade to a new version of GDevelop with a new theme, use this theme, then downgrade to a version without this theme (or if a theme is removed from GDevelop).

The theme is chosen from the preferences as you can see here: 
https://github.com/4ian/GDevelop/blob/0c22c52a783ff9fb665f1b617bf2a885afde2603/newIDE/app/src/MainFrame/Providers.js#L67-L72

This means that if you want to modify the default theme, it's something to do in the preferences. Where are the preferences stored when the user is launching GDevelop for the first time? 
From there: https://github.com/4ian/GDevelop/blob/0c22c52a783ff9fb665f1b617bf2a885afde2603/newIDE/app/src/MainFrame/Preferences/PreferencesContext.js#L216-L237
Precisely, the logic is in PreferencesProvider: 

https://github.com/4ian/GDevelop/blob/0c22c52a783ff9fb665f1b617bf2a885afde2603/newIDE/app/src/MainFrame/Preferences/PreferencesProvider.js#L34

I think that's here, or around these preferences files, that we should add the logic to choose the default theme :)

4ian (commented):
Finally reviewed this!

newIDE/app/src/UI/Theme/index.js
arthuro555: Yes, you are right, that was a silly mistake.

4ian:
Looks good! Thanks :)

PR #1953, opened 2020-08-29
updated admob icon

the-gem-dev:
we all know admob does not use its old logo anymore so i decided to update the icons in gdevelop to the current admob logo style.
and this is what it looks like
demoo
well i have a second thought on the icons and here it is.
admobico
presently i am using the one on the right what do you think


4ian:
I think it looks good! :)

the-gem-dev:
that was fast, thanks.

PR #1967, opened 2020-09-06
Fixes for the P2P extension

arthuro555:
- Fixes the send variable to all action
- Now all disconnected peers will be taken into account by using the same array system as the no dataloss mode
- Now there is a condition to detect when another instance connects remotely to the current instance
- Improves JSDoc and typing of p2p tools.


4ian:
Thanks for the fixes! I'll take a look :) Can you revert all the `fixtures/` folder changes? I'll do it myself on master :)

Extensions/P2P/B_p2ptools.js
```diff
@@ -18,52 +19,75 @@ gdjs.evtTools.p2p = {
...
+   * List of IDs of peers that just remotely initiated a connection.
+   * @type {Array<string>}
    */
-  lastDisconnectedPeerId: '',
+  connectedPeers: [],
 };
 
-gdjs.evtTools.p2p.loadPeerJS = function () {
+/**
+ * Internal function called to initialize PeerJS after it's
```
4ian, line 86: *its

4ian:
Looks good! Thanks a lot for doing these fixes 👍

PR #1976, opened 2020-09-10
Fix margins around renamed list items

Bouh:
* Also fix slightly cropped text when renaming an item
* Disable spellcheck on the text fields

### DIFF
--- a/newIDE/app/src/UI/Theme/DefaultTheme/index.js
+++ b/newIDE/app/src/UI/Theme/DefaultTheme/index.js
@@ -127,8 +127,7 @@ const muiTheme = createMuiTheme({
     MuiInput: {
       input: {
         padding: 0,
-        paddingBottom: 1,
-        zIndex: 1,
+        paddingBottom: 3,
       },
       underline: {
         '&:before': {
--- a/newIDE/app/src/UI/Theme/DefaultTheme/index.js
+++ b/newIDE/app/src/UI/Theme/DefaultTheme/index.js
@@ -125,6 +125,11 @@ const muiTheme = createMuiTheme({
       },
     },
     MuiInput: {
+      input: {
+        padding: 0,
+        paddingBottom: 1,
+        zIndex: 1,
+      },
       underline: {
         '&:before': {
           borderBottom: `1px solid #BBBBBB`,
--- a/newIDE/app/src/ObjectEditor/Editors/TextEditor.js
+++ b/newIDE/app/src/ObjectEditor/Editors/TextEditor.js
@@ -108,6 +108,7 @@ export default class TextEditor extends React.Component<EditorProps, void> {
               this.forceUpdate();
             }}
             hintText={<Trans>Choose a font</Trans>}
+            style={{ alignSelf: 'center' }}
           />
         </MiniToolbar>
         <Line noMargin>
--- a/newIDE/app/src/ObjectEditor/Editors/TextEditor.js
+++ b/newIDE/app/src/ObjectEditor/Editors/TextEditor.js
@@ -22,6 +22,7 @@ const styles = {
     width: 90,
     ...toolbarItemStyle,
   },
+  resourcesSelector: { alignSelf: 'center' },
   toolbarItem: toolbarItemStyle,
   checkbox: toolbarItemStyle,
 };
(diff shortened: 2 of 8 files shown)


newIDE/app/src/UI/Theme/DefaultTheme/index.js
```diff
@@ -125,6 +125,11 @@ const muiTheme = createMuiTheme({
       },
     },
     MuiInput: {
+      input: {
+        padding: 0,
+        paddingBottom: 1,
+        zIndex: 1,
```
4ian, line 131: I'm always a bit worried of putting zIndex... this can sometimes create hard to debug visual issues with other controls and force the browser/GPU to create a new texture in memory for rendering the element.
What's the result if we don't put it? :)

Bouh:
The border bottom is over the text, if you type g j q p y the bottom of
these characters are hidden.

Le jeu. 10 sept. 2020 à 22:09, Florian Rival <notifications@github.com> a
écrit :

newIDE/app/src/UI/Theme/DefaultTheme/index.js
Bouh: Without / With
image
Let's go without, with a bit more padding on bottom 👍

4ian:
Ok, if we can avoid the zIndex that's much better, and in fact it's no very big deal if some letters are a bit cut.
Can you do a few other screenshots of other dialogs having text fields? The game properties, the sprite editor/text editor (because they have text fields inside "MiniToolbar"), maybe some (action/condition editor) with a search bar too?

Thanks! :)

Bouh:
A little fix for align the resource selector from Text object.

And here how the ui look like now, a bit more compact, and still aligned.

image

newIDE/app/src/UI/Layout.js
```diff
@@ -27,7 +27,7 @@ const textFieldWithButtonLayoutStyles = {
   container: {
     flex: 1,
     display: 'flex',
-    alignItems: 'flex-start', // Align from the top to stay at the same position when error/multiline
+    alignItems: 'center', // Align from the top to stay at the same position when error/multiline
```
4ian, line 30: Can you verify the impact on this kind of multiline text field:

image

The idea was for the "123" button to *not* move when you add multiple lines. I don't have strong feelings about this, just be sure to remove/update the comment ("Align from the top to stay at the same position when error/multiline") if it's not the case anymore.

4ian:
Thanks for the comparison, very useful!

newIDE/app/src/UI/Layout.js
Bouh: I can fix the position of the button with the style override:

 ```
MuiButtonBase: {
      // Remove the web-ish "pointer" (hand) cursor from buttons
      root: {
        cursor: 'default',
        alignSelf: "flex-start",
      },
    },
```

Do you think this can have a incidence somewhere else ?

Comment (Bouh) [reply newIDE/app/src/UI/Layout.js:30]: I think have a solution for this.
This "center" was a fix for align the font resource selector in the text object, but i can achieve to align with another way.
Return to the original "flex-start", and instead use an override MuiAutocomplete used by the resource selector.

The result stay the same as the previous screenshot and the blue button stay at the start :)

Comment (4ian) [reply newIDE/app/src/UI/Layout.js]: > Return to the original "flex-start", and instead use an override MuiAutocomplete used by the resource selector.
> The result stay the same as my previous screenshots and the blue button stay at the start :)

Cool! You can also consider putting an inline style (`style={...}`) in the ResourcesSelector if you want to be 100% sure your custom style won't break anything else.

File: newIDE/app/src/UI/Theme/DefaultTheme/index.js [hunk 3]
```diff
@@ -173,6 +173,12 @@ const muiTheme = createMuiTheme({
         minHeight: 32, // Reduce the height of tabs to 32px
       },
     },
+    // Use a more visible color scheme for tabs:
+    MuiAutocomplete: {
+      root: {
+        alignSelf: "center",
```
4ian, line 179: Add a comment explaining why this is used, especially because MuiAutocomplete is used for other stuff (autocomplete for objects list in the events sheet, in the command palette, etc...).

Bouh: Oh yes the command palette, it is completely broken with this hack.
Instead i can add a style in the resourceSelector from the text Object porperties like your suggest, it need pass props for that.
I understand why fix this issue wasn't yet fixed, when we fix something, somewhere else a thing broke, it's endless 😅

4ian: Yeah that's why I heavily favor the approach "different CSS style for each component" and try to minimize the global CSS variables :) Sounds good now with just a style in the `ResourcesSelector`

PR #1979, opened 2020-09-12
Fix bug related to lighting (lack of webgl support)

### LINKED ISSUE #1973: Light crash the game in canvas
## Describe the bug
When the canvas is used, instead webGL, the game have an black screen during the loading because an error is trigger in the console.

## Error
image

## Possible solution
- When the context webGL is missing then disable the light.

@HarsimranVirk 

[Source #1](https://forum.gdevelop-app.com/t/gdevelop-5-beta-100-light-example-not-work-on-canvas-with-pixi-legacy-support/27293)
[#2](https://forum.gdevelop-app.com/t/light-not-working/27430) 

## Other details
* Beta 100
* On low devices without webGL

HarsimranVirk:
Resolves #1973


Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -309,6 +319,8 @@ gdjs.LightRuntimeObjectPixiRenderer.prototype._updateBuffers = function () {
     this._defaultVertexBuffer[6] = this._object.x - this._radius;
     this._defaultVertexBuffer[7] = this._object.y - this._radius;
 
+    if(!this._light) return;
+    this._light.shader.uniforms.center = this._center;
```
HarsimranVirk, line 323: We aren't returning quickly because tests depend upon the buffer data, and if we bail out early, buffer data would not be updated. For now, it's returning after updating the buffer to pass tests, but it's not ideal I would say. If nothing is rendering on the screen then we shouldn't compute any data related to it as well, so we would need to modify the tests somehow 🤔.

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
```diff
@@ -153,6 +153,8 @@ gdjs.LayerPixiRenderer.prototype.removeEffect = function (effectName) {
  * @param zOrder The z order of the associated object.
  */
 gdjs.LayerPixiRenderer.prototype.addRendererObject = function (child, zOrder) {
+  if(!child) return;
+
```
HarsimranVirk, line 157: Not sure about this check, I guess light object is the only nullable object which could be passed to this function, should we keep this check or do something else?

4ian: Strange, because this could be only null for TextEntry object?

4ian: Can you give an example of a stacktrace where this is used and crashing because child is null?

HarsimranVirk: > Can you give an example of a stacktrace where this is used and crashing because child is null?

Just below this line, if `child` is null, `child.zOrder` is going to crash (accessing a property of null). 

> Strange, because this could be only null for TextEntry object?

Right now, this could be null for LightObject as well, because `PIXI.Mesh` is not going to work without WebGL support.

4ian: TextEntry is avoiding this issue by... never calling `addRendererObject` :) 

I think it's good to keep `addRendererObject` with the assumption that `child` is always not null. If it's null, don't call it, and if you do, you probably missed something in your object logic. :) 
In the case of light, this means that we must take care of checking if we have a mesh before calling it.

HarsimranVirk: > TextEntry is avoiding this issue by... never calling addRendererObject :)

I didn't even think about this 😬.

> I think it's good to keep addRendererObject with the assumption that child is always not null.

Makes sense! I'll do it in the proper way then :)

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
4ian: Not sure I understand, why are tests are impacted? Is WebGL not supported by ChromeHeadless?

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
HarsimranVirk: @4ian I guess I'm missing something here, I don't see any call being made to `addRendererObject` (neither from lightruntimeobject nor from the pixi-renderer), and the weird part is that it has been working fine since the beginning. I've traced the stack, and found out that `RuntimeScene.createObjectsFrom()` calls `RuntimeObject.setLayer()`, which then finally calls `addRendererObject()`.

So, I'm confused as to how calling `addRenderObject()` (or more specifically, NOT calling it (when light is null)) from object's logic is going to affect anything 🤔. Can you tell me what's wrong here, am I understanding things differently?

Bouh: What if you return null to this._debugLight here 
```JS
gdjs.LightRuntimeObjectPixiRenderer.prototype.getRendererObject = function () {
  if (this._debugLight) {
    return this._debugLight;
  }
  return this._light;
};
```

4ian: So there is a check in setZOrder:

https://github.com/4ian/GDevelop/blob/43c420dff010b6142b2331f68b3deea6294de8d5/GDJS/Runtime/runtimeobject.js#L548-L556

but not in setLayer:
https://github.com/4ian/GDevelop/blob/43c420dff010b6142b2331f68b3deea6294de8d5/GDJS/Runtime/runtimeobject.js#L518-L520

nor in the function called when the object is removed from the scene:
https://github.com/4ian/GDevelop/blob/43c420dff010b6142b2331f68b3deea6294de8d5/GDJS/Runtime/runtimeobject.js#L319-L326

This means that while `getRendererObject` was somehow intended to be allowed to return null.. it was never implemented properly :) 

If we fix `setLayer` and `onDestroyFromScene`, I think we'll actually fix a bug for the text entry object (that would crash the game if we change its layer or destroy it - to be confirmed!).

Note that we could also choose to do what you did: check inside `addRendererObject` for nullity of `child`. I don't have a huge preference but because we started to add check in setZOrder already, let's do the same for setLayer/onDestroyFromScene.

Sorry for the confusion - I was initially convinced that this was properly handled but seems not :) 

There is still the question of why the light object is not calling `addRendererObject` at all. In theory you should! :) In the pixi renderer. But I think it works because whenever an object is created, the Z Order is also immediately set, which triggers a call to `changeRendererObjectZOrder`, which then trigger `addRendererObject`.

4ian:
@HarsimranVirk Let me know if you want to do the fix to setLayer/onDestroyFromScene as part of this PR or not :)

HarsimranVirk:
@4ian I guess we can do it in this PR, it seems a bit unrelated but also necessary at the same time.

4ian:
It's a bit unrelated indeed but at the same time needed. Give it a try by adding the check to the two functions. Then it should work properly :) If not, we'll directly see what we missed.

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
HarsimranVirk: @4ian I'm really sorry for keeping this PR pending for so long, I'm having exams right now 😓. 

> There is still the question of why the light object is not calling addRendererObject at all.

So, I remember why there's no call to `addRendererObject` from pixi renderer because there's no reference to the lighting layer. At the time of implementing it, I initially thought of storing a ref of lighting layer in the runtime scene, which I totally forgot to do (and since everything worked fine, nobody noticed it as well 😬).

How should I pass the ref of lighting layer while calling `addRendererObject` from the pixi renderer, then? We can't rely on the string "lighting" to get a ref to it because the name could be anything. So, should I store it as a property of runtime scene? 

Also, a quick question (unrelated), how does gdjs ensure that an object/instance isn't added twice (once when the object  itself calls `addRendererObject` and when `RuntimeScene.createObjectsFrom` calls `setLayer`, which then calls `addRendererObject`) ?

4ian: I've done the changes to allow a null object to be returned from `getRendererObject` here: https://github.com/4ian/GDevelop/commit/e6c483f398c31f2d2ca9ad3dc369be4814244dc0

So you can just rebase on/merge master :)

> how does gdjs ensure that an object/instance isn't added twice (once when the object itself calls addRendererObject and when RuntimeScene.createObjectsFrom calls setLayer, which then calls addRendererObject) ?

It's because the renderer object is always removed first before being added again. See `setLayer`:
https://github.com/4ian/GDevelop/blob/e6c483f398c31f2d2ca9ad3dc369be4814244dc0/GDJS/Runtime/runtimeobject.js#L524-L525

Also because `setLayer` is called for all objects after they are added, your object will still work even if you did not call `addRendererObject` - because setLayer will call it (and will call removeRendererObject before, which will do nothing).

For consistency with other objects, I recommend you call `addRendererObject` ont the base layer (yes). Just add a comment telling "While objects are usually set on a lighting layer, add it to the base layer, it will be changed after the object is created."

Extensions/Lighting/lightruntimeobject-pixi-renderer.js
HarsimranVirk: > HeadlessChrome 85.0.4183 (Windows 10.0.0) WARN: 'This device does not support webgl, which is required for Lighting Extension.'

This is what the HeadlessChrome triggers, so I guess WebGL is not supported then. 😬

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
```diff
@@ -153,6 +153,8 @@ gdjs.LayerPixiRenderer.prototype.removeEffect = function (effectName) {
  * @param zOrder The z order of the associated object.
  */
 gdjs.LayerPixiRenderer.prototype.addRendererObject = function (child, zOrder) {
+  if(!child) return;
```
4ian, line 156: I think we can remove this now? :)

HarsimranVirk: Oh yes :)

HarsimranVirk:
@4ian I'm not sure what's going on with CI, errors seem pretty unrelated to the PR 😅.

4ian:
Yeah don't care about this! Might be a wrong package-lock.json/yarn.lock. Can you confirm the tests are passing locally? (flow, formatting, GDJS tests?).

HarsimranVirk:
Let me double-check all the tests :)

4ian:
Thanks for working on these fixes! 👍

Issue #1955, opened 2020-08-31
3D Spatial Sound - HowlerJS


TerraPrimal:
## Description
From what I understand, GDevelop uses HowlerJS for sound. HowlerJS supports the use of 3D Spatial Sound which I think would be a great addition to the software.
Currently sounds are essentially "global" in the sense that we can't have a sound coming from an X/Y loc and have it get louder/dim depending on the distance.

An example of this (although in a 3D plane) is https://howlerjs.com/assets/howler.js/examples/3d/

## Solution suggested
Add HowlerJS's 3D Spatial Sound into GDevelop. Sounds could be attached to objects or set at an X & Y Coordinate. If possible, it would be great if we could somehow manually adjust the range of the sound depending on the zoom level, however that might be going a little far at this point.

AtomicCrocStudios:
Just curious if there has been any update on this? I know that Howler was updated to support this feature, but I haven't heard anything since.

Thanks :)

PR #1982, opened 2020-09-13
Update howler

arthuro555:
Bump howler to v2.2.0
Useful for #1955 as it adds the spatial sound plugin, adds lots of bug fixes and support for m4b format.
Tested on both music examples with both music and sound actions.


Bouh:
This need to be tested atleast once in an apk on mobile or in a webview on mobile.

4ian:
Ideally let's do some tests on mobile to avoid any obvious issue. Apart from that, happy to update Howler! :)

arthuro555:
I tested with the preview over wifi, it works fine, and the discord members "Kappa" and "BlacKat Belladonna" tested the apk version for me. No issues spotted.

Bouh:
And the patch notes from 2.1.2 to 2.2.0 have nothing to migrate, it's only fixes for us.
Ready to merge!

4ian:
If it's only bugfixes, sounds safe! Thanks @arthuro555 for the PR :)

PR #1987, opened 2020-09-15
Remove PIXI hack in renderer objects

### LINKED ISSUE #1980: Changing the opacity of a TiledSprite while not visible will result in invisibility
If you change the opacity of a TiledSprite while it is not visible then it becomes permanently invisible.
I discovered this bug when "not visible" meant the object was more than 1000 pixels off screen.
However you can contrive this defect by setting the TiledSprite to be hidden and then changing the opacity.

Very simple reproduction:
[opactiy-bug-5b100.zip](https://github.com/4ian/GDevelop/files/5213307/opactiy-bug-5b100.zip)

You can remove the single "Hide LockBlue" line below that hides the tiled sprite and the example works as intended by alternating between transparent and not on each mouse click.
opacity_bug_logic_screenshot

Version: GDevelop 5.0 b100

Bouh:
Fix #1980
On scene you can see a Text object, an tiled Sprite and a sprite.
Before / After
fix_opacity


4ian:
Thanks for taking care of this 👍

Issue #1965, opened 2020-09-05
New Gdevelop Layer Effects example


the-gem-dev:
Hi!

  I'd like to submit a new example to be added to GDevelop.
  Here is the link to download it: .
[Layer Effects Example.zip](https://github.com/4ian/GDevelop/files/5178171/Layer.Effects.Example.zip)

  I confirm that any assets can be used freely by anybody, including for commercial usage.
  here are some screen shots
![hgjkl;](https://user-images.githubusercontent.com/53819287/92306544-cdc3b500-ef87-11ea-9546-20cf11bb526b.PNG)
![ghjkl;](https://user-images.githubusercontent.com/53819287/92306568-fc419000-ef87-11ea-850f-48560c247027.PNG)
![ffffjj](https://user-images.githubusercontent.com/53819287/92306574-0a8fac00-ef88-11ea-88cc-f2142b65d97f.PNG)
![hhg](https://user-images.githubusercontent.com/53819287/92306576-0cf20600-ef88-11ea-9755-3a6e399f210c.PNG)

4ian:
@the-gem-dev Let's continue the discussion here, it will be easier than on the Pull Request, I'll integrate the example myself :) Sorry for taking so much time. Overall it's a good quality example 👍 A few feedbacks on the example that I'd like to be fixed before adding this to GDevelop:

* You're loading images from a github.com url. Can we entirely remove this? This is not officially supported and gives this example much more complexity than needed (and there is no loader while loading the images, and it requires an internet connection). Keep it simple ;) How? Just put your images as animation of the "Images" object. Done! :) 
  * We can remove the two extensions too - I know you like having them but again as it's an example, it must be dead simple.
* You don't need Trigger Once here:
![image](https://user-images.githubusercontent.com/1280130/94532700-1c3a3b80-0236-11eb-91ba-3b16a645f414.png)
* Can we remove "All" effects? It true that it's fun :) But I do not recommend this at all, this will literally destroy the performance on mobile phones. Effects can be expensive and should be used with caution - so we can't show this in an official example.
* You're repeating the same events. Consider using an object group for these buttons:
![image](https://user-images.githubusercontent.com/1280130/94532894-5acff600-0236-11eb-8f25-7b22895d4e21.png)

  To do this: use the action to set the scale, instead of setting the width/height (set the scale to 1.1 when hovered, and back to 1.0 when not hovered).
[...]

PR #1988, opened 2020-09-15
added layer effects example

the-gem-dev:
#1965 
Added layer effects example. below are some.
hgjkl;
ghjkl;
ffffjj
hhg


4ian:
Thanks! :) Could you revert the changes of all the files in newIDE/app/src/fixtures? I'll do that myself.

the-gem-dev:
okay it should be out now

4ian:
Seems like fixture files are still there. No worries, let's continue the discussion on #1965, you can just read my comments and upload a new zip file with the example. I'll add it myself to the examples when it's done :) Thanks!
Feel free to close this PR in the meantime.

PR #1992, opened 2020-09-19
Add is fullscreen condition

arthuro555:
As requested by @Silver-Streak, a condition to check if the window is in fullscreen.
Uses a simple check `window.screen.height === window.innerHeight` for HTML5
Uses https://www.electronjs.org/docs/api/browser-window#winisfullscreen for electron

Tested on electron and browser.

### DIFF
--- a/Core/GDCore/Extensions/Builtin/WindowExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/WindowExtension.cpp
@@ -46,7 +46,8 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsWindowExtension(
           _("Is the game in fullscreen?"),
           _("Game's window and resolution"),
           "res/actions/fullscreen24.png",
-          "res/actions/fullscreen.png");
+          "res/actions/fullscreen.png")
+      .AddCodeOnlyParameter("currentScene", "");
 
   extension
       .AddAction("SetWindowMargins",
--- a/Core/GDCore/Extensions/Builtin/WindowExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/WindowExtension.cpp
@@ -43,7 +43,7 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsWindowExtension(
           "IsFullScreen",
           _("Fullscreen activated?"),
           _("Check if the game is currently in fullscreen."),
-          _("Is the game in fullscreen?"),
+          _("The game is in fullscreen"),
           _("Game's window and resolution"),
           "res/actions/fullscreen24.png",
           "res/actions/fullscreen.png")
--- a/Core/GDCore/Extensions/Builtin/WindowExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/WindowExtension.cpp
@@ -39,6 +39,16 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsWindowExtension(
       .SetDefaultValue("yes");
 
   extension
+      .AddCondition(
+          "IsFullScreen",
+          _("Fullscreen activated?"),
+          _("Check if the game is currently in fullscreen."),
+          _("Is the game in fullscreen?"),
+          _("Game's window and resolution"),
+          "res/actions/fullscreen24.png",
+          "res/actions/fullscreen.png");
+
+  extension
       .AddAction("SetWindowMargins",
                  _("Change the window's margins"),
                  _("This action changes the margins, in pixels, between the "
--- a/GDJS/Runtime/cocos-renderers/runtimegame-cocos-renderer.js
+++ b/GDJS/Runtime/cocos-renderers/runtimegame-cocos-renderer.js
@@ -57,7 +57,7 @@ gdjs.RuntimeGamePixiRenderer.prototype.isFullScreen = function() {
     if (electron) {
       return electron.remote.getCurrentWindow().isFullScreen();
     }
-    return window.screen.height === window.innerHeight;
+    return false; // Unsupported
   }
 
 /**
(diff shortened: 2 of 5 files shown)


Bouh:
Simplier you can just check the boolean from the private variable `this._isFullscreen` used in `gdjs.RuntimeGamePixiRenderer.prototype.setFullScreen`

Create `gdjs.RuntimeGamePixiRenderer.prototype.isFullScreen` and call it in `gdjs.evtTools.window.isFullScreen`

arthuro555:
> Simplier you can just check the boolean from the private variable `this._isFullscreen` used in `gdjs.RuntimeGamePixiRenderer.prototype.setFullScreen`

I am scared that this value might not be 100% accurate, since the fullscreen can be enabled through other ways than setFullscreen.

GDJS/Runtime/events-tools/windowtools.js
```diff
@@ -22,6 +22,13 @@ gdjs.evtTools.window.setFullScreen = function(runtimeScene, enable, keepAspectRa
     runtimeScene.getGame().getRenderer().setFullScreen(enable);
 };
 
+gdjs.evtTools.window.isFullScreen = function() {
+	if (typeof require === "function") {
+		return require("electron").remote.getCurrentWindow().isFullScreen();
+	}
+	return window.screen.height === window.innerHeight;
```
4ian, line 29: Could you move the implementation into the renderer (`gdjs.RuntimeGamePixiRenderer`)? 
The idea is to avoid having any HTML5 specific code in these events, specifically everything that is using `window`. 

In the future, we may have a new renderer which does not run in a browser :) (for example, something that runs in Node.js only, for a server only game, or something like Cocos2D-JS which embeds a JS engine but not a browser).

4ian: Then inside, the pixi renderer, uses `var electron = this.getElectron(); if (electron) { ... } else { ... }`, so that it is consistent with the rest of the Electron specific code :) 

Note that there might be other Electron specific code in the game engine/extensions (like the file system), but in this case as it's code in `windowtools.js` related to the window size, we put everything in the renderer.

4ian:
> I am affraid that this value might not be 100% accurate, since the fullscreen can be enabled through other ways than setFullscreen.

I recommend that we check this value, then we check window.screen.height === window.innerHeight as a fallback (for HTML5 only. For Electron, what you did is perfect already).

Core/GDCore/Extensions/Builtin/WindowExtension.cpp
```diff
@@ -38,6 +38,17 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsWindowExtension(
                     true)
       .SetDefaultValue("yes");
 
+  extension
+      .AddCondition(
+          "IsFullScreen",
+          _("Fullscreen activated?"),
+          _("Check if the game is currently in fullscreen."),
+          _("Is the game in fullscreen?"),
```
4ian, line 46: Usually we use an affirmative sentence: "The game is in fullscreen"

GDJS/Runtime/cocos-renderers/runtimegame-cocos-renderer.js
```diff
@@ -49,6 +49,17 @@ gdjs.RuntimeGameCocosRenderer.prototype.setFullScreen = function(enable) {
...
 
+/**
+ * Checks if the game is in full screen.
+ */
+gdjs.RuntimeGamePixiRenderer.prototype.isFullScreen = function() {
+    var electron = this.getElectron();
+    if (electron) {
+      return electron.remote.getCurrentWindow().isFullScreen();
+    }
+    return window.screen.height === window.innerHeight;
```
4ian, line 60: window might not be defined in Cocos2D-JS on iOS/Android/native executables, might be a good idea to test for its existence.

GDJS/Runtime/pixi-renderers/runtimegame-pixi-renderer.js
```diff
@@ -256,6 +256,18 @@ gdjs.RuntimeGamePixiRenderer.prototype.setFullScreen = function(enable) {
...
 
+/**
+ * Checks if the game is in full screen.
+ */
+gdjs.RuntimeGamePixiRenderer.prototype.isFullScreen = function() {
+  var electron = this.getElectron();
+  if (electron) {
+    return electron.remote.getCurrentWindow().isFullScreen();
+  }
+  // Height check is used to detect F11 triggered full screen.
```
4ian, line 267: Nit picking: "[...] user triggered full screen (for example F11 shortcut)"

F11 could change and is not the only way to trigger it (menu bar for example)

4ian (approved):
Added a few nits, otherwise looks good!

Core/GDCore/Extensions/Builtin/WindowExtension.cpp
Silver-Streak: This terminology makes sense for many of them. "The game is maximized", "The game is in fullscreen", "The game is in focus", etc.

GDJS/Runtime/cocos-renderers/runtimegame-cocos-renderer.js
arthuro555: Tbh I cannot even get a Cocos 2d export to run properly. The closest I got is the platformer example working with no graphics showing up only the sounds working. Is cocos2d even meant to work? Has it been abandonned? Will it be woked on?

4ian: I think we can consider it mostly abandoned. We should probably remove the export option and then remove the code supporting it. For now, it's also fine to not implement this method and just `return false;` with a comment `// Unsupported`.

4ian (approved):

PR #1994, opened 2020-09-19
Add electron tools

arthuro555:
This is an extension that wraps arround pretty much all relevant BrowserWindow APIs.
This should satisfy @Silver-Streak feature request: https://forum.gdevelop-app.com/t/conditions-actions-and-expressions-around-game-windows-and-fullscreen/27613


Extensions/Electron/electrontools.js
```diff
@@ -0,0 +1,200 @@
+gdjs.evtTools.electron = {
```
4ian, line 1: Any chance you could add `// @ts-check` for this file?
Don't care about typing of `gdjs.evtTools.electron.window`, just add some parameter typing: `@param {boolean} activate`.

It's not super super useful here but better than nothing and can help in the future spot obvious mistake like variable name typos. Feel free to `// @ts-ignore` any error that is not easily solvable.

4ian:
Nice! :) Should be pretty useful for Windows/macOS/Linux exports.

My main request for this PR would be if possible to rename the extension from `Electron` to `AdvancedWindow` (and related stuff, for example `electrontools.js` => `advancedwindowtools.js`). The idea is that the name of an extension encompass what features are provided by the extension, rather than its underlying implementation.

Electron is one of the "runtime"/apis that can run a game (along with a browser and with Cordova/webviews on mobile phones, and maybe in the future other JS environments). It's perfectly fine to expose some apis that work only on some exports (for example here, some Electron apis). In most case, we try to have them in extensions that are grouping things by feature, so that in the future if it's possible to implement a few of these actions/conditions in another platform, we can do it.
If it's not possible (here because these APIs are heavily desktop centric), we'll just mark them as such in the editor, by putting all actions/conditions/expressions under a group called:
`Advanced window management/Windows, Linux, MacOS` (i.e: like what we do for `Filesystem` extension). If later we expose more Electron APIs, we can do so in separate extensions.

4ian:
You've not renamed the folder, and seems that you forgot to commit a file?

GDJS/package.json
```diff
@@ -8,13 +8,14 @@
   "devDependencies": {
     "@types/expect.js": "^0.3.29",
     "@types/mocha": "^5.2.7",
+    "@types/electron": "^1.6.10",
```
4ian, line 11: Nice 👍

Extensions/Electron/JsExtension.js
```diff
@@ -0,0 +1,603 @@
...
+      )
+      .getCodeExtraInformation()
+      .setIncludeFile('Extensions/Electron/electron-advancedwindowtools.js')
+      .setFunctionName('gdjs.evtTools.electron.getPositionY');
+    
+    extension
+      .addExpression(
+        'WindowOpacity',
+        _('Window opacity'),
+        _('Returns the current window opacity (a number from 0 to 1).'),
```
4ian, line 587: [...] from 0 to 1, 1 being fully opaque).

Extensions/Electron/JsExtension.js
```diff
@@ -0,0 +1,603 @@
...
+      .getCodeExtraInformation()
+      .setIncludeFile('Extensions/Electron/electron-advancedwindowtools.js')
+      .setFunctionName('gdjs.evtTools.electron.center');
+
+    extension
+      .addAction(
+        'SetWindowPosition',
+        _('Set window position'),
+        _('Changes the window position.'),
+        _('Set the window position to _PARAM0_X and _PARAM0_Y'),
```
4ian: Careful about the arguments, one is wrong.
Also, specify positions as `x;y`, like in the actions to set object position:

`Set the window position to _PARAM0_;_PARAM1_`

Extensions/Electron/JsExtension.js
```diff
@@ -0,0 +1,603 @@
...
+      .getCodeExtraInformation()
+      .setIncludeFile('Extensions/Electron/electron-advancedwindowtools.js')
+      .setFunctionName('gdjs.evtTools.electron.setOpacity');
+
+    extension
+      .addAction(
+        'Center',
+        _('Center window'),
+        _('Centers the current window.'),
+        _('Center window'),
```
4ian: There is already an action to center the window. Can you remove this one?

Extensions/Electron/JsExtension.js
```diff
@@ -0,0 +1,603 @@
...
+      )
+      .getCodeExtraInformation()
+      .setIncludeFile('Extensions/Electron/electron-advancedwindowtools.js')
+      .setFunctionName('gdjs.evtTools.electron.isAlwaysOnTop');
+
+    extension
+      .addAction(
+        'SetKiosk',
+        _('Enable kiosk mode'),
+        _('Puts the window in kiosk mode.'),
```
4ian: Can you explain what Kiosk mode is?

`[...] in kiosk mode. When in kiosk mode, the window will blabla`

Extensions/Electron/JsExtension.js
```diff
@@ -0,0 +1,603 @@
...
+      .getCodeExtraInformation()
+      .setIncludeFile('Extensions/Electron/electron-advancedwindowtools.js')
+      .setFunctionName('gdjs.evtTools.electron.setFullScreenable');
+    
+    extension
+      .addCondition(
+        'IsFullScreenable',
+        _('Window full-screenable'),
+        _('Checks if the window can be full-screened.'),
+        _('The window can be full-screened'),
```
4ian: The window can be set in fullscreen

Extensions/Electron/JsExtension.js
```diff
@@ -0,0 +1,603 @@
...
+        _('Make the window gain or lose focus.'),
+        _('Focus the window: _PARAM0_'),
+        _('Advanced window tools/Windows, Linux, MacOS'),
+        'res/actions/window24.png',
+        'res/actions/window.png'
+      )
+      .addParameter('yesorno', _('Focus the window?'), '', false)
+      .setDefaultValue('true')
+      .getCodeExtraInformation()
+      .setIncludeFile('Extensions/Electron/electron-advancedwindowtools.js')
```
4ian: `Extensions/AdvancedWindow/advancedwindowtools.js`

4ian:
Merged! Thanks for adding this 👍👍

Bouh:
This extension was tested? The some features don't work.

image
image


This require documentation:
image

**Make the window invisible** cause trouble for stop the preview.

@arthuro555

arthuro555:
> This extension was tested? The some features don't work.

That's unfortunate, I tested about 90% of the actions and everything worked perfectly. As I copied evry function name from the electron documentation (https://www.electronjs.org/docs/api/browser-window#winsetopacityopacity-windows-macos) I thought there was no way it would not... Well from now on I'll test 100% to avoid something like that from happening. I am really sorry 🙏 

I'll fix all of that now.

PR #1997, opened 2020-09-24
Feature/rtl

denim2x:
This solution is simpler - no more unneeded modules; `RtlLanguages` contains only those languages defined in `LocalesMetadata`.
Simply using `createMuiTheme()` with changed `muiTheme.direction` isn't working.
Using `rtlOverrides` gets the job done and doesn't pollute, while ensuring proper render-time styling.

### DIFF
--- a/newIDE/app/src/MainFrame/Providers.js
+++ b/newIDE/app/src/MainFrame/Providers.js
@@ -2,6 +2,7 @@
 import * as React from 'react';
 import DragAndDropContextProvider from '../UI/DragAndDrop/DragAndDropContextProvider';
 import { ThemeProvider } from '@material-ui/styles';
+import { StylesProvider, jssPreset } from '@material-ui/core/styles';
 import { getTheme } from '../UI/Theme';
 import UserProfileProvider from '../Profile/UserProfileProvider';
 import Authentification from '../Utils/GDevelopServices/Authentification';
@@ -25,6 +26,13 @@ import {
 import GDevelopThemeContext from '../UI/Theme/ThemeContext';
 import { UnsavedChangesContextProvider } from './UnsavedChangesContext';
 import { CommandsContextProvider } from '../CommandPalette/CommandsContext';
+import { create } from 'jss';
+import rtl from 'jss-rtl';
+
+// Add the rtl plugin to the JSS instance to support RTL languages in material-ui components.
+const jss = create({
+  plugins: [...jssPreset().plugins, rtl()],
+});
 
 type Props = {|
   authentification: Authentification,
@@ -58,11 +66,12 @@ export default class Providers extends React.Component<Props, {||}> {
           <PreferencesProvider disableCheckForUpdates={disableCheckForUpdates}>
             <PreferencesContext.Consumer>
               {({ values }) => {
-                const theme = getTheme(values.themeName);
+                const theme = getTheme({ themeName: values.themeName, language: values.language });
                 return (
                   <GDI18nProvider language={values.language}>
                     <GDevelopThemeContext.Provider value={theme.gdevelopTheme}>
-                      <ThemeProvider theme={theme.muiTheme}>
+                      <StylesProvider jss={jss}>
+                        <ThemeProvider theme={theme.muiTheme}>
                         <UserProfileProvider
                           authentification={authentification}
                         >
@@ -94,7 +103,8 @@ export default class Providers extends React.Component<Props, {||}> {
                             )}
                           </I18n>
                         </UserProfileProvider>
-                      </ThemeProvider>
+                        </ThemeProvider>
+                      </StylesProvider>
                     </GDevelopThemeContext.Provider>
                   </GDI18nProvider>
                 );
--- a/newIDE/app/src/UI/Theme/index.js
+++ b/newIDE/app/src/UI/Theme/index.js
@@ -1,9 +1,14 @@
 // @flow
+import { createMuiTheme } from '@material-ui/core/styles';
+import { isLtr } from '../../Utils/i18n/RtlLanguages';
+import memoize from '../../Utils/Memoize';
+import DefaultTheme, {
+  themeName as defaultThemeName,
+  type Theme,
+} from './DefaultTheme';
 import DarkTheme from './DarkTheme';
-import DefaultTheme from './DefaultTheme';
 import NordTheme from './NordTheme';
 import SolarizedDarkTheme from './SolarizedDarkTheme';
-import { type Theme } from './DefaultTheme';
 import './Global.css';
 
 // To add a new theme:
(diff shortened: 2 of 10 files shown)


newIDE/app/src/UI/Theme/index.js
```diff
@@ -1,23 +1,74 @@
...
+  const { gdevelopTheme, muiTheme } = theme;
+  return {
+    gdevelopTheme,
+    muiTheme: ltr ? createLtrTheme(muiTheme) : createRtlTheme(muiTheme),
+  };
+};
 
 export type GDevelopTheme = $PropertyType<Theme, 'gdevelopTheme'>;
+
+function createLtrTheme(muiTheme: object): object {
```
4ian, line 42: Use `Object` in flow, `object` is considered as an error:
image

newIDE/app/src/UI/Theme/index.js
```diff
@@ -1,23 +1,74 @@
...
+  MuiSvgIcon: {
+    root: rtlOrder,
+  },
+  MuiFormControlLabel: {
+    root: rtlDirection,
+  },
+  MuiTextField: {
+    root: rtlDirection,
+  },
+};
```
4ian, line 74: That's very clear 👍

newIDE/app/src/UI/Theme/index.js
```diff
@@ -1,23 +1,74 @@
...
-  'GDevelop default': DefaultTheme,
+  [defaultThemeName]: DefaultTheme,
   Dark: DarkTheme,
   Nord: NordTheme,
   'Solarized Dark': SolarizedDarkTheme,
 };
 
-export const getTheme = (themeName: string): Theme =>
-  themes[themeName] || themes['GDevelop default'];
+export function getTheme({ themeName, language }: {| themeName: string, language: string |}): Theme {
```
4ian, line 24: nitpicking:

what is returned here from this function is not a `Theme`. A `Theme` is defined to be an object like this:
```
{  gdevelopTheme, muiTheme }
```

with the content of the objects `gdevelopTheme` and `muiTheme` defined in `DefaultTheme.js` (thanks to `$Exact<typeof theme>`).

Here, what you return is an object containing `gdevelopTheme` and something else called `muiTheme`, but which is *different*, as it was passed to `createMuiTheme`. If we look at material-ui, we see that the argument is called a mui theme "options", while the returned value is a mui "theme": https://github.com/mui-org/material-ui/blob/next/packages/material-ui/src/styles/createMuiTheme.d.ts
(Unfortunately we don't use TypeScript so can't benefit from these types).

What I suggest is that:
- we keep returning `{ gdevelopTheme, muiTheme: ltr ? createLtrTheme(muiTheme) : createRtlTheme(muiTheme), };` as you're doing currently. The naming is perfectly valid!
- but we rename in DefaultTheme, and in the other theme files, `muiTheme` to `muiThemeOptions`. After all, because we removed a function call, there is no reason the object should still be called the same as before - it's something slightly different :) 
  - the only other change should be to adapt getTheme so that you pass the `muiThemeOptions` to createLtrTheme/createRtlTheme.

Finally, change the return type of getTheme. Instead of Theme, it should be something that we can call "ConstructedTheme" (or a better name?), described like this:

```
export type GDevelopTheme = $PropertyType<Theme, 'gdevelopTheme'>;
export type ConstructedTheme = {| gdevelopTheme: GDevelopTheme, muiTheme: Object |}
```

we use Object as we don't have anything better here to use.
In this approach, `Theme` is still useful, as it describe the shape of the objects used to describe the options, and it ensures consistency between themes (if you forget something in a theme, Flow will raise an error).

Let me know if something is not clear - should not be too much changes ;)

newIDE/app/src/MainFrame/Providers.js
```diff
@@ -25,6 +26,12 @@ import {
 import GDevelopThemeContext from '../UI/Theme/ThemeContext';
 import { UnsavedChangesContextProvider } from './UnsavedChangesContext';
 import { CommandsContextProvider } from '../CommandPalette/CommandsContext';
+import { create } from 'jss';
+import rtl from 'jss-rtl';
+
+const jss = create({
```
4ian, line 32: Let's add a quick comment: 
`// Add the rtl plugin to the JSS instance to support RTL languages in material-ui components.`

4ian (commented):
Thanks a lot! It's much much simpler and very clear, I'm big fan of the result 👍
I've added a comment about some typing and a few other nitpickings. After that, we should be good to go!

newIDE/app/src/Utils/memoize.js
```diff
@@ -0,0 +1,21 @@
+// @flow
```
4ian, line 1: nitpicking: `Memoize.js`, for consistency with other files naming.

newIDE/app/src/Utils/memoize.js
```diff
@@ -0,0 +1,21 @@
...
+
+export default function<Input, Output>(func: (Input) => Output): (Input) => Output {
+  const primitives = new Map();
+  const objects = new WeakMap();
+
+  function cacheFor<Input, Output>(input: Input): Output {
```
4ian, line 7: This typing is not correct, you're not returning an output:
image
I think you can just remove `: Output` here and let the type checker infer the type.

Run `npm run flow` to run the static typing checker: https://github.com/4ian/GDevelop/tree/master/newIDE#tests

newIDE/app/src/Utils/memoize.js
```diff
@@ -0,0 +1,21 @@
...
+    const isObject: boolean = typeof input === 'object';
```
4ian, line 8: No need to type things that can be trivially inferred by Flow:
```suggestion
    const isObject = typeof input === 'object';
```

newIDE/app/src/Utils/memoize.js
```diff
@@ -0,0 +1,21 @@
...
+    return isObject ? objects : primitives;
+  }
+
+  return (input: Input) => {
+    const cache = cacheFor(input);
+    const isCached: boolean = cache.has(input);
```
4ian, line 14: Same as before:
```suggestion
    const isCached = cache.has(input);
```

4ian (commented):
Seems like Flow helped us find a part that we forgot! :) 

image

This file is the theme being created for the Storybook environment. See https://github.com/4ian/GDevelop/tree/master/newIDE#development-of-ui-components to run Storybook (a web-app where you can test/develop the components, outside of the real app).
It needs to be adapted to call createMuiTheme

Finally, seems like Flow indicate that a few themes are missing some fields (good to know, should be fixed) and there is something to edit in the flow config to fix 2 or 3 ignore reported errors in JSS. I'll do that, feel free to ignore those!

4ian:
I've made a few more changes locally:
- run autoformatting
- There were warning in the console - be sure to always fix them.
- Don't export types that are not needed, and prefer export const rather than export default unless it's a component.
- Fix Flow errors in Memoize.js - be sure to fix Flow issues in your files before pushing.

but I can't push them to your branch (have you allowed contributors on GitHub to push to your branch when opening the pull request?).
I'll merge this now and push fixes in a next commit.

Thanks for the work on this! It seems that adding to `body` `dir="ltr"` is missing, but this creates a few issue, that I've added as a TODO in my commit containing fixes.

4ian:
@denim2x Feel free to claim the bounty on https://www.bountysource.com/issues/92320507-80-bounty-add-support-for-rtl-languages as you've worked on this :)

denim2x:
@4ian Thanks

The issue is still open; for the bounty to be claimed, the issue needs to be closed.

PR #2001, opened 2020-09-26
added icon for empty game in starters

the-gem-dev:
wwwwwwwww
dddd


blurymind:
Do you guys think empty game should be on top of the list?

4ian:
Would be good I think to have an icon that is the same size as others. Also maybe make a new one, because this one is a bit dated and looking like a new text file rather than a blank game?

the-gem-dev:
I would make a new one and share with you for updates

the-gem-dev:
what do you think about this
g8752

4ian:
I like it very much! :) Happy to go ahead with that 👍

the-gem-dev:
i edited the icons on the starter page to have a square-like shape, what do you think.
gamefeeldemo
pairs
downhill-bike-racing
geometry-monster
isometric-game
platformer
8-bit-space-shooter
particles-effects-demo

4ian:
I like them too! 
Though the icon for the space shooter is not giving justice to the crisp pixel art style of it: could you redo it but without resizing any of the images? Basically, one pixel in an image must be exactly equal to one pixel in another. 
Be sure to get a crisp rendering, currently it's a bit blurry because you may have resized too much and using bilinear interpolation instead of nearest neighbour.

For the "Game Feel Demo" icon, maybe a slightly darker background color to better see the icon?

Overall it's very nice 👍

the-gem-dev:
> I like them too!
> Though the icon for the space shooter is not giving justice to the crisp pixel art style of it: could you redo it but without resizing any of the images? Basically, one pixel in an image must be exactly equal to one pixel in another.
> Be sure to get a crisp rendering, currently it's a bit blurry because you may have resized too much and using bilinear interpolation instead of nearest neighbour.
> 
> For the "Game Feel Demo" icon, maybe a slightly darker background color to better see the icon?
> 
> Overall it's very nice

How about this?
gamefeeldemo

8-bit-space-shooter

4ian:
Better for game feel! 
I still find the icon for the space shooter not super pretty compared to the game. Let's just put the player spaceship in the middle and that's it. Be sure to make it not blurry at all :)

the-gem-dev:
How about this :)
8-bit-space-shooter

the-gem-dev:
Update :)
8-bit-space-shooter2

4ian:
Same problem as before ;) This is not respecting the pixel art nature of the assets. Let me explain: for example you have the rocks in the background that are resized. This means that their pixels are now smaller than the spaceship.
The spaceship is also rotated. By doing so, it make it a bit blurry and does not look like very nicely made pixel art.

In other words: don't rotate anything, keep the pixels straight, keep the pixels of an image exactly to the size of another.

the-gem-dev:
How about this :), i am using inkscape, i increased the size of the sprite in piskel to 150 and then imported it to inkscape.
8-bit-space-shooter3

4ian:
It's better, but what about the bullets? 

👇
> keep the pixels of an image exactly to the size of another.

;)

the-gem-dev:
Is this okay :)
8-bit-space-shooter4

4ian:
Looks good! 👍

Readme.md
```diff
@@ -37,7 +37,6 @@ Links
 
 ### Community
 
-* [GDevelop forums](https://forum.gdevelop-app.com) and [Discord chat](https://discord.gg/rjdYHvj).
```
4ian, line 40: ?

newIDE/app/src/ProjectCreation/LocalStarters.js
```diff
@@ -232,6 +232,12 @@ export default class LocalStarters extends Component<Props, State> {
                     onClick={() => this.createFromExample(i18n, 'pairs')}
                   />
                   <ListItem
+                    leftIcon={
+                      <ListIcon
+                        iconSize={40}
+                        src="res/starters_icons/new.png"
+                      />
+                    }
```
4ian, line 240: Can you post a screenshot of what the result looks like? Thanks :)

blurymind:
Why arent we doing this cropped border effect with css? Why edit the png files to get it? :)

the-gem-dev:
> Why arent we doing this cropped border effect with css? Why edit the png files to get it? :)
> It just makes it harder to add examples with a thumbnail - as now their thumbnail needs to have a consistent style with the other ones. The contributer will have to open up photoshop/whatever and carefully make their thumbnail look like the other ones

I would share the .svg  in case anyone would want to create a new icon in the future

4ian:
As these are starters that should be carefully prepared game, it might be ok to ask anyone creating it to spend a bit of time ensuring the icon style matches the style of the rest - after all these starters should be high quality small games (compared to examples that are just about demonstrating some feature).

if @the-gem-dev can share the SVG, we can put them in the sources so it should be fairly easy to copy paste one to generate a new icon? :)

Readme.md
4ian: Can you revert this change? :)

newIDE/app/src/ProjectCreation/LocalStarters.js
the-gem-dev: ### 
newimage

Readme.md
the-gem-dev: i am not sure how that was deleted. i would fix that quickly.

the-gem-dev:
Could You try opening this in inkscape or adobe illustrator
[starter_icons.zip](https://github.com/4ian/GDevelop/files/5316586/starter_icons.zip)

blurymind:
no worries, just wondering why are we adding more work for ourselves in the long run :) this might be an incentive for people to not put a preview icon on their demo project

4ian:
I'll go ahead with this, as it looks a bit more consistent than what we have and I think it's a good idea to have "game icons" for starters. It will always be possible to redo these later if needed.

4ian:
Thanks for working on this @the-gem-dev! 👍

PR #2006, opened 2020-10-03
Refactor audio (pre)loading

arthuro555:
The current way of loading audio has a problem. There is no real preloading, so when audio is played there is a slight delay (the time for the audio file to be actually loaded). This is because we use one new Howl instance every time we want to play a sound.

This PR has for goal to cache the Howl instance of every file and reuse them. That way it is also possible to add to audio resources an attribute to let the user preload and cache the Howl instance on the loading screen.

This PR also makes HowlerSound use Composition over Inheritance, something that is preferred for this codebase if I understood correctly (Though currently it is done in a dirty way). 

There is one downside though, as the Howl instances are cached, it can make the game take some more memory as they never get properly unloaded. This could be fixed with an action to clear the cache and to clear one file/resource from the cache.


GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -44,16 +48,21 @@ gdjs.HowlerSound = function (o) {
...
     that._stopped = false;
   });
 };
-gdjs.HowlerSound.prototype = Object.create(Howl.prototype);
+
+for(var item in Howl.prototype) {
+  gdjs.HowlerSound.prototype[item] = function() {
+    this._howl[item].apply(this, [arguments, this.id]);
+  }
+}
```
4ian, line 65: While it works, I'm a bit concerned that JS engine could very well de-optimize the whole thing. 
(Basically, if you use `prototype`, it's usually for Object.create which is how you do inheritance in ES5 - or for declaring a method. Otherwise, I'm afraid of JS engine not recognizing what you do and disabling their optimizations)

Instead, let's write simple, boring code: just declare a bunch of methods that call into `this._howl`.
Not as "clever" but totally safe, works with autocompletion, works with static typing, and the reader of the code understands everything by reading once instead of wondering what is going on with this weird loop on `prototype` :)

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -385,6 +404,7 @@ gdjs.HowlerSoundManager.prototype.preloadAudio = function (
   onComplete,
   resources
 ) {
+  window.s = this;
```
4ian, line 407: Careful not to commit your debug code ;)

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -409,7 +429,9 @@ gdjs.HowlerSoundManager.prototype.preloadAudio = function (
 
   var loadedCount = 0;
   var that = this;
-  function onLoad() {
+  function onLoad(_, error) {
+    if(error) console.error(error);
```
4ian, line 433: That's a good idea, add a description like: `console.error("There was an error while loading an audio file:", error);`

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -420,12 +442,24 @@ gdjs.HowlerSoundManager.prototype.preloadAudio = function (
...
-      httpRequest.open('GET', file);
-      httpRequest.send();
+      // Load as sound
+      var sound = new Howl({
+        src: [file],
+        preload: true,
+        onload: onLoad,
+        onloaderror: onLoad,
+      })
+      that._loaded.sound[file] = sound;
```
4ian, line 452: This kind of preloading sound risky for large audio file (aka "music" in GDevelop).
We're creating two howls, which sounds like useless work - and Howl are not "free', they can use audio resources which are limited.
By not setting `html5`, we also have Howler fully download and decode the whole audio file which can be costly to do for all musics.

I think we should find a way to give this control to the user or somehow automatically do that.

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -357,23 +376,23 @@ gdjs.HowlerSoundManager.prototype.getGlobalVolume = function () {
 
 gdjs.HowlerSoundManager.prototype.clearAll = function () {
   for (var i = 0; i < this._freeSounds.length; ++i) {
-    if (this._freeSounds[i]) this._freeSounds[i].unload();
+    if (this._freeSounds[i]) this._freeSounds[i].stop();
```
4ian, line 379: Is stop enough? The idea with `unload` was that we would really destroy and free the audio resource.

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -14,24 +14,28 @@
  * @memberof gdjs
  * @class HowlerSound
  */
-gdjs.HowlerSound = function (o) {
-  Howl.call(this, o);
+gdjs.HowlerSound = function (howl) {
+  this._howl = howl;
+  // Get a unique sound playing ID.
+  this.id = howl.play();
```
4ian, line 20: Is getting this id really necessary? Seems like we don't need it because we only have one sound managed in a HowlerSound?
I'm also concerned of launching the play and immediately stopping it. Not sure how various browsers will react to doing this for a large number of sounds.

4ian (commented):
> This PR also makes HowlerSound use Composition over Inheritence, something that is preffered for this codebase if I understood correctly (Though currently it is done in a dirty way).

Yes composition is very usually better, in this case there is no downside to do it and is a preferred approach (because we don't need polymorphism or deal with objects of various types).

> That way it is also possible to add to audio resources an attribute to let the user preload and cache the Howl instance on the loading screen.

That would be interesting! While reading the PR I was thinking of this.
Currently I think it's too risky to ask browsers to "really load" at the beginning all the music and sound files. If you do that for a game with 20 unique musics of 5mb each, you'll probably exhaust resources, memory or something bad.

We could add an attribute to the audio resources to "Preload as a sound" or "Preload as a music" - so that we have a Howl object ready to be started whenever we need it :)
We could also add 3 advanced actions: "Preload a sound in memory", "Preload a music in memory"  (these two actions are doing the same thing as the audio resources marked as "Preload as a sound" or "Preload as a music". Basically, the loading screen is running these "actions" for each resource that has the attribute) and an action "Clear the preloaded sounds/musics" (that would clear this "cache" of Howl objects).

Let me know what you think!

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
arthuro555: The idea is to not unload it and keep it cached/loaded unless the user explicitly unloads it.

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
arthuro555: The id is necessary as one file could run two times in parallel. Though I agree I did it wrong, I probably should keep the id undefined until play is called.

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
arthuro555: Yes this is still a draft o was testing loading at the beginning. The idea is to give the user a sort of drop-down in the resources attributes "Preloading" that can be set to "No Preloading", "Preload as sound", "Preload as music", or "Preload as music and as sound". The names are pretty self explaining I think.

arthuro555:
Ok, now it is ready for proper reviewing.

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -420,12 +533,30 @@ gdjs.HowlerSoundManager.prototype.preloadAudio = function (
...
-      httpRequest.addEventListener('error', onLoad);
-      httpRequest.addEventListener('abort', onLoad);
-      httpRequest.open('GET', file);
-      httpRequest.send();
+      if(files[file].preloadAsSound) {
+        // Load as sound
+        var sound = new Howl({
+          src: [file],
+          preload: true,
+          onload: onLoad,
```
4ian, line 541: If a file is both `preloadAsSound` and `preloadAsMusic`, onLoad will be called twice, which will make the loading inconsistent. You probably need to think of another way of ensuring both are loaded before validating the load of this resource as a whole.

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -240,20 +303,79 @@ gdjs.HowlerSoundManager.prototype._storeSoundInArray = function (arr, sound) {
...
+ * @returns {gdjs.HowlerSound}
+ */
+gdjs.HowlerSoundManager.prototype.createHowlerSound = function (soundName, isMusic) {
+  var soundFile = this._getFileFromSoundName(soundName);
+  var container = this._loaded[isMusic ? "music" : "sound"];
+
+  if (
+    container.hasOwnProperty(soundFile)
+  ) {
+    return new gdjs.HowlerSound(container[soundFile]);
```
4ian, line 319: Does this work if you run multiple time the sound? Like an explosion effect, which can be run multiple times at the same time.

4ian (commented):
Very interesting :) My main concern is about running multiple times a sound, will this work? I'm afraid the Howl object would be shared so only one sound at a time of the same file would work?

arthuro555:
> My main concern is about running multiple times a sound, will this work? I'm afraid the Howl object would be shared so only one sound at a time of the same file would work?

Well, I assumed so siince there is the whole IDs system, and that there is this as an example in the docs:  
image  
If not though then I have a problem because the whole PR relies on that concept 😅

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -14,87 +14,146 @@
...
+    if(this._id === null) return;
+    return this._howl.loop(loop, this._id);
+  }
+
+  volume(volume) {
+    if(this._id === null) return;
+    return this._howl.volume(volume, this._id);
+  }
+
+  mute(mute) {
```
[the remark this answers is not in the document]
arthuro555: I thought those might be useful for a javascript user. If they see howler sound, and it has almost the same api as a Howl, they might be confused why this doesn't work. I also think it might be nice to add those features to GDevelop.
And `on` is actually used.

4ian:
> Well, I assumed so siince there is the whole IDs system

Ah nice, I think I did not understand how it works! If that's the case that's even better then :) 
I re-read the documentation and it seems that it's properly handled so no worries.

4ian:
Sounds better than the old system then 👍👍

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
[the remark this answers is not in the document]
4ian: Fair enough then!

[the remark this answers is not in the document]
4ian: Maybe we should write a few jsdocs for these functions then if we want them to show in the generated documentation?

4ian:
Very excited about these changes :) Let's triple check that all conditions/actions are working, in particular with multiple sounds and loading of preloaded sound/musics/not preloaded.
Maybe a quick test game to make for this (or find one that would already exist, but I'm not sure)?

arthuro555:
I began to try adding typescript typecheking, sadly it doesn't support well overloaded functions :/ 
I added the chaining of HowlerSound methods to be consistent with howler.
I fixed all the pints you mentioned.

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -14,87 +14,264 @@
...
+   * This becomes true when the sound playback has ended.
+   * @returns {boolean}
+   */
+  canBeDestroyed() {
+    return this._canBeDestroyed;
+  }
 
-// Methods to safely update the rate of the sound:
+  /**
+   * Get the sound rate
```
4ian, line 137: nit: Get the playback rate

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -14,87 +14,264 @@
...
+  }
 
-gdjs.HowlerSound.prototype.getRate = function () {
-  return this._rate;
-};
-gdjs.HowlerSound.prototype.setRate = function (rate) {
-  this._rate = gdjs.HowlerSoundManager.clampRate(rate);
-  this.rate(this._rate);
+  /**
+   * Set the sound rate
```
4ian, line 146: nit: Set the sound playback rate.

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -14,87 +14,264 @@
...
+   */
+  setRate(rate) {
+    this.rate(rate);
+  }
+  /**
+   * @returns {number}
+   */
+
+  /**
+   * Get/set the sound rate.
```
4ian, line 158: I'm not a big fan of these mix of setter/getter in the same function. 
They are hard/impossible to type statically, and they are probably a nightmare to compile to machine code by JIT because their argument always differ... even their return value differ.
They also force you to do ugly `typeof rate === 'undefined'`.

Let's remove all of these and use getRate/setRate instead. Simple, efficient, can be statically verified, and setters can be chained. What's not to love? 😄

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -14,87 +14,264 @@
...
+   * @returns {HowlerSound} The current instance for chaining.
+   */
+  fade(from, to, duration) {
+    if (this._id === null) return this;
+    this._howl.fade(from, to, duration, this._id);
+    return this;
+  }
+
+  /**
+   * @typedef {'load' | 'loaderror' | 'playerror' | 'play' | 'end' | 'pause' | 'stop' | 'mute' | 'volume' | 'rate' | 'seek' | 'fade' | 'unlock'} HowlerEvent
```
4ian, line 246: Nice :)

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -418,14 +653,41 @@ gdjs.HowlerSoundManager.prototype.preloadAudio = function (
...
+          onLoad(_, error);
+        }
+
+        preloadAudioFile(file, callback, /* isMusic= */ true);
+        preloadAudioFile(file, callback, /* isMusic= */ false);
+        continue;
+      }
+      if (fileData.preloadAsSound)
+        preloadAudioFile(file, onLoad, /* isMusic= */ false);
+      if (fileData.preloadAsMusic)
```
4ian, line 689: Use else if to clearly show that only one branch should be followed

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
arthuro555: > What's not to love?

I wanted to keep it consistent with howler. Should I also leave out chaining on setters?

4ian: I guess the reason for having them is for keeping compatibility with Howler though, as you said... I wonder if it's worth doing it though.

4ian:
> I added the chaining of HowlerSound methods to be consistent with howler.

Let me know what you think, but as explained in the comment I'm not a huge fan of these magic functions, they are impossible to statically type and probably a nightmare for JIT. Not sure why people like them, writing getXxx/setXxx is not really long nor verbose and very clear.

Anyway, let me know what you think we should do. Keeping compatibility with Howler is interesting, but on the other hand in the future if we replace it, we'll have this kind of "fluent" api that does not look like anything else in the codebase to maintain.

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
4ian: > Should I also leave out chaining on setters?

Chaining on setter is fine (we can type it, it's pleasant to use, there are no optimization problems).
I think we should go for "classic" getter and chainable setters. If we all document them, it will be an easy to use API for users. If some people were using them in JS code as `Howl`, we can always add compatibility function (a `rate` function that would use getRate/setRate, etc...) - but only if someone really complains about this (to avoid cluttering our sound API).

arthuro555:
I think overloading is indeed not a very good idea. As we have autocompletion in the IDE and API docs (though I believe their generation is currently broken due to bad typing in gd.js), it shouldn't be really a problem for the users to find the syntax for GDevelop. In the worst sceneario, they can use the public `createHowlerSound` interface to get access to a raw Howl if they want to.

Do you think I should also remove the chaining for setters? I think that a fluent api is better when mass-setting properties of a sound.

4ian:
> In the worst sceneario, they can use the public createHowlerSound interface to get access to a raw Howl if they want to

Yep. Or we could add an accessor to the howler sound if really needed one day (let's not do it now).

Totally ok for the fluent setters 👍

arthuro555:
Ok, I think all that's left is testing. Silver gave me a project for testing, let's try it out :)

arthuro555:
Ok, everything seems to work properly, I think the next step would be to let people try it out in their game to see if there is a breaking change as this is a massive rewrite of the sound system, and then it should be ready to merge I think.

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -1,3 +1,4 @@
+//@ts-check
```
4ian, line 1: 😍

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -14,87 +15,263 @@
...
+        error
+      );
+      this._paused = false;
+      this._stopped = true;
+    });
+
+    // In case the howl didn't finsih loading, start the sound once it finishes loading.
+    this._howl.on('load', function () {
+      this.play();
+    });
```
4ian, line 65: All these events are attached to `howl`, which is shared between all the sounds (or musics) of the same resource, the difference being done using the `id` - but these events are not looking at the id. So I think we risk, when stopping/pausing/playing a sound, to mark all the sounds with the same resource as stopped/paused/playing.
(this is the kind of edge case that I'm a bit afraid of, and why it would be a very good idea to have a test game where you can start multiple sounds or musics, stop them, pause them, etc..).

I think `id` is passed to each callback, in which case you must check if the id is equal to `this._id` - hopefully that is all that is needed.

arthuro555: This is more problematic, as at this point we don't have the id yet as it is defined only after begining playback...

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -73,6 +35,51 @@ gdjs.HowlerSound = class HowlerSound {
...
+      /*
+       * Track events to be sure the status is
+       * sync'ed with the sound - though some should be redundant
+       * with `play`/`pause` methods already doing that. Keeping
+       * that to be sure that the status is always correct.
+       *
+       * This is done in play as we need the ID to set up those callbacks.
+       */
+
+      this.on('play', function () {
```
4ian, line 50: This is still correct, you listen to the events of all the sounds of the same Howl, so you'll improperly set _paused/_stopped and other stuff for all sounds, which could create hard to debug issues.

You must either set your event only on the id of the Howl (in which case you need to know the id) OR you must read the id that is passed, according to what I see in Howler.js source code, as the first parameter of all callbacks.
If this id is not the same as `this._id`, it means that the callback is called for a sound that is not this one (either this one has a different id, or id was not yet set. In both cases it's not the same sound).

In all cases, triple check this because this seems to be a recipe for subtle bugs :)

arthuro555: > You must either set your event only on the id of the Howl (in which case you need to know the id)

This is what I currently do? Note that I don't use thi._howl.on but this.on which passes the correct id. And I have the ID for sure, it just vot generated and stored.

I also tested and pausing doesn't pause other sounds or stuff like that currently.

4ian: Ah right! I missed that you're using `this.on[ce]`. All good then, my bad I overlooked this 👍

4ian:
Alright I think we've reached the point where we can now do a few days of testing with maybe a testing version for people to check that it works on their game :) I'll also test on iOS on Sunday.

arthuro555:
Apparently, this is very broken. It's weird because some bugs seem to have appeared for things I have already tested and didn't modify (like music between scenes or volume). It's definitely not merge ready.

4ian:
Could be related to events? Maybe the easiest way to debug is to add some `console.log` a bit everywhere to see what is being played/paused/stopped/etc. Should allow to quickly see if the behavior is correct or not.

arthuro555:
Ok I fixed two major issues.
The first one was that my unloading actions didn't delete HowlerSounds using the unloaded Howl, so it resulted in semi unloaded zombie sounds with very weird behavior.
My second one was completely messing up the listeners by using the onload unnecessarily, and using normal functions instead of arrow functions making this not working properly.

There are still some other problems I need to fix though.

I think the listeners are not a really good idea the more I think about it, because they sound like a memory leak.
As JS doesn't has destructors, we cannot remove the listeners. The listeners have a function that encloses `this`, and is stored in the Howl which is still accessible as it is stored in the sound manager, so a mark and sweep garbage collector will not collect the HowlerSound until it's associated Howl is collected too. Then maybe I am wrong, I am not the most experienced JavaScript user 😅 .

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -485,17 +528,42 @@ gdjs.HowlerSoundManager.prototype.unloadAudio = function (soundName, isMusic) {
...
+      }
+    }
+  }
+
+  removeFromConatiner(this._freeMusics);
+  removeFromConatiner(this._freeSounds);
+  removeFromConatiner(this._musics);
+  removeFromConatiner(this._sounds);
+
   container[soundFile].unload();
```
4ian, line 548: I don't see any unload method on `gdjs.HowlerSound`?

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -485,17 +528,42 @@ gdjs.HowlerSoundManager.prototype.unloadAudio = function (soundName, isMusic) {
 
   if (!container[soundFile]) return;
 
+  // Make sure any sound using the howl is deleted so that the howl can be garbage collected
+  // and no weird "zombies" using the unloaded howl can exist.
+  const howl = container[soundFile];
+  function removeFromConatiner(soundContainer) {
+    for (let i in soundContainer) {
+      if (soundContainer[i] && soundContainer[i]._howl === howl) {
+        soundContainer[i].stop();
```
4ian, line 537: Should you call `unload` instead of stop?

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -485,17 +528,42 @@ gdjs.HowlerSoundManager.prototype.unloadAudio = function (soundName, isMusic) {
 
   if (!container[soundFile]) return;
 
+  // Make sure any sound using the howl is deleted so that the howl can be garbage collected
+  // and no weird "zombies" using the unloaded howl can exist.
+  const howl = container[soundFile];
+  function removeFromConatiner(soundContainer) {
```
4ian, line 534: better call this `clearContainer`

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
arthuro555: This is not the `HowlerSound` but the `Howl`.

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
arthuro555: I cannot unload a `HowlerSound`, as it is basically sharing the Loaded Howl and the difference is only data like the seek, the played state etc. I shouldn't have to stop them, I just do to make sure they don't try to continue playing the unloaded Howl to prevent "uncontrollable sounds".

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
4ian: Btw, because you called `clearContainer` before, it must have already unloaded and cleared the soundFile, so container[soundFile] is now undefined. So this will probably crash there? 
This means that `container[soundFile].unload();` and `delete container[soundFile];` are not needed at all :)

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -485,17 +528,42 @@ gdjs.HowlerSoundManager.prototype.unloadAudio = function (soundName, isMusic) {
...
 /**
  * Unloads all audio from memory.
+ * This will clear the Howl cache.
  * This will also stop any running music or sounds.
  */
 gdjs.HowlerSoundManager.prototype.unloadAll = function () {
   //@ts-ignore Ts doesn't know about Howler
   Howler.unload();
+
+  // Clean up old sounds that still have the dead Howl instanes.
```
4ian, line 561: instances

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -231,6 +462,8 @@ gdjs.HowlerSoundManager.prototype._storeSoundInArray = function (arr, sound) {
   var index = null;
   for (var i = 0, len = arr.length; i < len; ++i) {
     if (arr[i] !== null && arr[i].canBeDestroyed()) {
+      //FIXME
+      //@ts-ignore
       arr[index] = sound;
```
arthuro555, line 467: I am confused as to how this is working, the sound is always being set to index null? Isn't it meant to set `arr[i]` instead?

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
arthuro555: The containers are cleared from the `HowlerSound` instances that refer to the `Howl` to make sure it is garbage collected and prevent to use a `HowlerSound` that uses a dead/unloaded `Howl` by accident.
The unload and delete delete the `Howl` from the `Howl` cache that is used in `createHowlerSound` (`this._loadesSounds`/`this._loadedMusics`).

4ian:
> The first one was that my unloading actions didn't delete HowlerSounds using the unloaded Howl, so it resulted in semi unloaded zombie sounds with very weird behavior.

Good catch. When you launch an audio, it will continue to exist in memory even if you drop all references to it. This is because they are not garbage collected, they continue to exist in the browser memory - where JS objects that have no reference to them are garbage collected and removed from memory.

>  garbage collector will not collect the HowlerSound until it's associated Howl is collected too.

Indeed, if you drop all references to HowlerSound in the sound manager, it's still possible that the HowlerSound is living in memory as long as the associated Howl exists, because the associated Howl has listeners, so we have this "chain":

`HowlerSound` -> listener function -> this pointing to `HowlerSound`.

This may mean that we need to clean the listeners in a `destroy` function in gdjs.HowlerSound. This was probably not needed before because dropping all references to gdjs.HowlerSound also meant dropping the only reference to the `Howl`.

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
4ian: Good catch! This is indeed a mistake. This means that the array of "free" sounds/musics will have a sound at index `null` when we try to replace a sound that was finished by another :/
The impact of this bug are likely that:
- As soon as there is one sound finished, all the new sounds are added to the index `null` (erasing the old gdjs.HowlerSound).
- this is kind of ok because as they are "free" sounds, we can't manipulate them using events (we can just launch them). Even if we override a `gdjs.HowlerSound` by another, the browser audio resource will still finish to play.
- ...but this means that when the app goes to background for example on Cordova, we try to pause all the sounds (both those on channels and these "free" sounds). We'll likely miss some.

This means that if I launch a lot of sound like explosions, then put my game in background, the explosions will "finish" to play.
I'm unsure if this means we're also leaking audio resources. In all cases, better fix this!

arthuro555:
> This may mean that we need to clean the listeners in a destroy function in gdjs.HowlerSound. This was probably not needed before because dropping all references to gdjs.HowlerSound also meant dropping the only reference to the Howl.
(in this destroy function, we call off on all listeners for the id)

I am not a fan of this, If a user uses a `gdjs.HowlerSound` in JS events, it won't necessarily be managed by the Sound Manager and the destroy method could never get called. I think the public API should be memory leak free, and those listeners are only meant as a failsafe. Maybe we should just directly remove them?

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
4ian: Ok I confused the "loaded" cache and the maps containing the channels/free sounds/musics.

In this case, prefer to avoid the generic "container" name: 
- name the first one the "loaded" container.
- you can keep the generic term "container" for the maps containing the channels/free sounds/musics.

4ian:
> I am not a fan of this, If a user uses a gdjs.HowlerSound in JS events, it won't necessarily be managed by the Sound Manager and the destroy method could never get called. I think the public API should be memory leak free

I agree it's less convenient, but that's also why an API is not necessarily public: we don't always want user to create gdjs.HowlerSound on their own. That's our choice to make this a public api or not. If we consider it's too risky because it's an implementation detail, we can just ask people to use the gdjs.SoundManager. And that's what they should already do :) 

The API *is* leak free if we create a destroy method. Again it's surely less convenient to use and more error prone, but it's maybe the only solution.

As for why listeners were needed, I remember using an action to launch a sound on a channel, then using the condition to check if the sound was stopped (this can be useful to do something once the sound is finished).
Problem: the sound was not immediately considered as playing by Howler, so the condition would be directly true.

I'm ok to remove these listeners if we ensure this use case works correctly. Note that we can do this by choosing that a sound that is somehow "loading" is considered as playing (because it will play once the loading is done).

Otherwise, we can still keep the listeners if we properly ensure that we call `off` on them. It was not needed before because of not sharing `Howl` objects. It's now probably needed, so better do it, even if it's at the cost of adding a few calls to a `destroy()` function. I'm ok with this because we can consider our `HowlerSound` as an implementation details that people should not use directly, marking it `private`. (and this is a good example of why exposing too much APIs as public can sometimes be harmful for the future :))

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
4ian: Ok I see. Considering this ok then, though might not another check in case we go with a `destroy` method.

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
4ian: Also, consider renaming the two containers to `loadedHowlsForMusics` and ``loadedHowlsForSounds`, to make extra clear that these containers are containing `Howl`s, not `gdjs.HowlerSound` :)

4ian:
> Maybe we should just directly remove them?

Give it a try if you wish, but it will be important to try to keep the guarantee that when you launch an action to play a sound, the sound is immediately after considered by GDevelop as:
- playing, even if has a short loading before being actually launched.
- stopped only when it reach the end or if it can't be played.

Which make me think.... maybe we don't need gdjs.HowerlSound at all! :) After all, it was just an abstraction to guarantee what I just wrote before (and also to clamp the rate, but it's something that is easy to do).
Maybe we could directly use and store `Howl`s, and be sure that when we check if a sound is played, we check that either it's playing or it's loading.
Would need careful tests, but maybe this could be a great simplification of this whole stuff... Give it a thought!

arthuro555:
I think HowlerSounds should most definitely be a public API. Controlling a sound is a very basic thing, if we close up too much the API, people that use JS (and according to what I see on the forum and discord, the number is quite significant) will be forced to be using the event tools which give a very bad developer experience. The howler sound is therefore necessary to document all of that. Granted, it should not be documented as a Howler sound, but through a ISound interface, as there are other implementations possible (like cocos sounds).

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -68,65 +54,35 @@ gdjs.HowlerSound = class HowlerSound {
...
+  }
 
+  /**
+   * Sets the current sound as a free sound.
+   * This will set up a listener to mark the sound as destroyable
+   * after it ended.
+   * This has to be done separately as it could lead to a memory leak otherwise.
+   * @returns {HowlerSound} The current instance for chaining.
+   */
+  _setFree() {
```
arthuro555, line 80: Technically, this should be redundant and the store in array method should be able to check if the sound is stopped to know if it can be replaced.

arthuro555:
Something I just realized is that this is a breaking change for JavaScript devs, as the overloaded methods on HowlerSound have been replaced with the getter/setter ones!

4ian:
Don't worry about backward compatibility for now, we can always add later adapter from old functions to the new get/set ones. Let's first ensure we have a 100% correct/not leaking implementation.

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -14,87 +15,274 @@
...
+  constructor(howl) {
+    this._howl = howl;
+  }
+
+  /**
+   * Begins playback of the sound.
+   * @returns {HowlerSound} The current instance for chaining.
+   */
+  play() {
+    if (this._howl._state === 'loaded') {
```
4ian, line 57: Seems that you can use `state()`: https://github.com/goldfire/howler.js/blob/97b10106c244e0f95077561aae1d512522c6d015/src/howler.core.js#L1722

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -240,24 +427,125 @@ gdjs.HowlerSoundManager.prototype._storeSoundInArray = function (arr, sound) {
...
+ * Preloads a sound or a music in memory.
+ * @param {string} soundName The name of the file or resource to preload.
+ * @param {boolean} isMusic True if a music, false if a sound.
+ */
+gdjs.HowlerSoundManager.prototype.loadAudio = function (soundName, isMusic) {
+  const soundFile = this._getFileFromSoundName(soundName);
+  const cacheContainer = isMusic ? this._loadedMusics : this._loadedSounds;
+
+  //@ts-ignore Ts doesn't know about Howler
+  cacheContainer[soundFile] = new Howl({
```
4ian, line 468: Do you think we should add a condition like `if (!cacheContainer.hasOwnProperty(soundFile)) {`, like for `createHowlerSound`, as it would be not necessary to erase the existing `Howl` if there is already one in the cacheContainer (and that can be already used by some `gdjs.HowlerSound`)?

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -240,24 +427,125 @@ gdjs.HowlerSoundManager.prototype._storeSoundInArray = function (arr, sound) {
...
+gdjs.HowlerSoundManager.prototype.unloadAudio = function (soundName, isMusic) {
+  const soundFile = this._getFileFromSoundName(soundName);
+  const cacheContainer = isMusic ? this._loadedMusics : this._loadedSounds;
+
+  if (!cacheContainer[soundFile]) return;
+
+  // Make sure any sound using the howl is deleted so that the howl can be garbage collected
+  // and no weird "zombies" using the unloaded howl can exist.
+  const howl = cacheContainer[soundFile];
+  function clearConatiner(howlerSoundContainer) {
```
4ian, line 493: clearContainer

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
4ian: True, though maybe in the past this was necessary because just checking if the sound is stopped would return `true` if the sound was loading, so it would delete a sound from the free sounds array while it was not even launched?

I agree this kind of seems redundant now, but I would be worried of such race conditions.

If we keep this, do you think we could also do this for musics/sounds on channels? I know that we're not using `canBeDestroyed` on them, but for consistency I think it's better to avoid doing any difference between a free and a channel sound/music. So we could set up this listener in all cases? what do you think?

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -14,87 +15,274 @@
...
-      error
+
+    return this;
+  }
+
+  /**
+   * Sets the current sound as a free sound.
+   * This will set up a listener to mark the sound as destroyable
+   * after it ended.
+   * This has to be done separately as it could lead to a memory leak otherwise.
```
4ian, line 77: Can you explain what do you mean by "this has to be done separately"?

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
```diff
@@ -14,87 +15,274 @@
...
+  /**
+   * Begins playback of the sound.
+   * @returns {HowlerSound} The current instance for chaining.
+   */
+  play() {
+    if (this._howl._state === 'loaded') {
+      this._id = this._howl.play(this._id === null ? undefined : this._id);
+    } else {
+      this._loading = true;
+      // In case the howl didn't finsih loading, start the sound once it finishes loading.
```
4ian, line 61: typo: finish

GDJS/Runtime/howler-sound-manager/howler-sound-manager.js
arthuro555: If we set up the end event on a non free sound, it may never be called (for example of it gets paused and never resumed). This would cause a leak, because the listener would never be removed, etc. On free sounds, we know that no interaction from outside can do that as they are not accessible, so we know it will reach its end at some point.

arthuro555:
I converted this PR to typescript, now I think the only thing left to do was to fix this bug: https://forum.gdevelop-app.com/t/gdevelop-music-improvements-testing/28256/11?u=arthuro555

arthuro555:
🤔 It seems typescript isn't recognizing gdjs.SoundManager anymore. Anyone knows what I might have done wrong?

arthuro555:
Ok I fixed the bug that was reported on the forum back then. I posted a new build on the forum to let people hopefully test it out. All that is left to do now is fix the weird typescript error where it doesn't find `gdjs.SoundManager` anymore and a bit of typo fixing.

4ian:
>  All that is left to do now is fix the weird typescript error where it doesn't find gdjs.SoundManager anymore and a bit of typo fixing.

For some reason Semaphore CI is not triggering, can you post the full error here?

arthuro555:
```
$ tsc
Runtime/runtimegame.ts:87:37 - error TS2551: Property 'SoundManager' does not exist on type 'typeof gdjs'. Did you mean 'JsonManager'?

87       this._soundManager = new gdjs.SoundManager(
                                       ~~~~~~~~~~~~

  Runtime/jsonmanager.ts:27:16
    27   export class JsonManager {
                      ~~~~~~~~~~~
    'JsonManager' is declared here.

Runtime/runtimegame.ts:151:29 - error TS2694: Namespace 'gdjs' has no exported member 'HowlerSoundManager'.

151     getSoundManager(): gdjs.HowlerSoundManager {
                                ~~~~~~~~~~~~~~~~~~

Runtime/runtimescene.ts:981:29 - error TS2724: 'gdjs' has no exported member named 'SoundManager'. Did you mean 'JsonManager'?

981     getSoundManager(): gdjs.SoundManager {
                                ~~~~~~~~~~~~


Found 3 errors.

error Command failed with exit code 2.
```

4ian:
I see! Where in the code do you assign gdjs.SoundManger to gdjs.HowlerSoundManager? Look at how it's done for Pixi renderers to get an idea (at the end of each file).

arthuro555:
I didn't change it: https://github.com/4ian/GDevelop/pull/2006/files#diff-4ffbb004738df5211810d820b417a55bb34023e6df14b029303bdef385bf21bcR731

4ian:
Seems like the `import type { ... } from '...'` is making the namespace "ignored" (see the opacity on gdjs):

image

Not sure why but surely because this kind of "global" namespace is a specific TypeScript thing.

4ian:
To avoid losing too much time on this, you can translate all the `export`s to `declare` in Howler.d.ts
Not as nice, but it will work for now?

arthuro555:
Indeed, it worked. Thanks, I would have never guessed this was the cause.

arthuro555:
[According to this forum users who found other issues, the PR works fine now](https://forum.gdevelop-app.com/t/gdevelop-music-improvements-testing/28256/15?u=arthuro555)

4ian:
Great news! I'm a bit behind in terms of reviews between this, the array/boolean variables and BitmapText, but I'll take a look :)

arthuro555:
No worries, take your time.

GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
```diff
@@ -1,104 +1,296 @@
...
+        const newID = this._howl.play(this._id === null ? undefined : this._id);
+        this._id = newID;
+
+        // Manually handle the play event before we have an ID.
+        // Before loading, howler won't register events as without an ID we cannot set a listener.
+        // Once we have an ID, we can transfer control of the events to howler.
+        // We also need to call them once as Howler doesn't for the first play event.
+        this._onPlay.forEach((func) => {
+          // Transfer the event to howler now that we have an ID
+          this.on('play', func);
```
4ian, line 81: I'm not sure to understand this `this.on('play', func);`, can you explain again?
Why having a `func(newID)` below?

GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
```diff
@@ -1,104 +1,296 @@
...
+     */
+    stopped(): boolean {
+      return this.paused() && this.getSeek() === 0;
+    }
+
+    /**
+     * Get the sound playback rate.
+     */
+    getRate(): float {
+      if (typeof this._id === 'undefined') return 0;
```
4ian, line 138: nitpicking: I recommend to avoid typeof in almost all cases. A simple `this._id === undefined` will work.
Or better, `this._id === null` (undefined is usually a problem in JavaScript, while null is a way to say "I made this on purpose"). 

In both cases, be sure to assign it in the constructor, to avoid the JS engine to be forced to dynamically add a field to the class at runtime: `this._id = undefined` or `this._id = null` in the constructor.

GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
```diff
@@ -1,104 +1,296 @@
...
+      if (this._id === undefined) return false;
+      return this._howl.loop(this._id);
+    }
+
+    /**
+     * Set if the sound is looping.
+     * @returns The current instance for chaining.
+     */
+    setLoop(loop: boolean): this {
+      if (this._id !== null) this._howl.loop(loop, this._id);
```
4ian, line 167: You're comparing id to null here, sounds like a bug because the rest is using undefined?
I would recommend moving everything to null - undefined will make it hard to debug if something is really undefined or is undefined on purpose. Better use always null.

GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
```diff
@@ -1,104 +1,296 @@
...
     }
 
-    // Methods to safely update the rate of the sound:
-    getRate() {
-      return this._rate;
+    /**
+     * Removes an event listener to the howl.
+     */
+    off(event: HowlerEvent, handler: HowlCallback): this {
+      if (this._id !== null) this._howl.off(event, handler, this._id);
```
4ian, line 284: same, a null here to be put everywhere or nowhere.

GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
```diff
@@ -1,104 +1,296 @@
...
+      return this;
     }
 
-    setRate(rate): void {
-      this._rate = HowlerSoundManager.clampRate(rate);
-      this.rate(this._rate);
+    /**
+     * Returns the raw howl.
+     * Be careful with this, you could break something!
```
4ian, line 290: I would go as far as "Not to be used outside of the game engine." or mark this private and add a // @ts-ignore when using it.

GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
arthuro555: Basically we want howler to handle the play events, triggered when the music starts playing or restarts because of the loop. Just doing this._howl.on("play") is fine or the latter, but not for the former, as the events trigger before we register the event (we may only register the event after getting the ID, which require playing which triggers the play event). To workaround that, before we have an ID, events are being held manually and we wait for the play to actually 
1. Call the functions to achieve the normal behavior of the play event being called when first starting the sound
2. Now that we have an ID actually set the callback so that the behavior of calling it again when it replay because of the loop

Technically that second point isn't necessary for how we internally use the play event, but I think it is better to keep the API consistent with the howler API to not confuse javascript users.

GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
```diff
@@ -232,84 +426,193 @@ namespace gdjs {
...
+      const cacheContainer = isMusic ? this._loadedMusics : this._loadedSounds;
+
+      if (!cacheContainer.hasOwnProperty(soundFile)) {
+        //@ts-ignore FIXME Somehow TypeScript does not detect the Howl consturctor
+        cacheContainer[soundFile] = new Howl({
+          src: [soundFile],
+          preload: true,
+          html5: isMusic,
+          onplayerror: (_, error) =>
+            console.error("Can't play a sound. Error is: ", error),
```
4ian, line 456: `Can't play an audio file` - to avoid confusion between GDevelop music and sounds distinction?

GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
```diff
@@ -232,84 +426,193 @@ namespace gdjs {
...
+          onloaderror: (_, error) =>
+            console.error('Error while loading sound file: ', error),
```
4ian, line 458: same, `audio file`/

GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
4ian: Well, it's true that the manager itself is called SoundManager and Howler uses "HowlerSound". Might be still be good to be precise, either "audio file" or "can't play a sound with Howler"?

GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
```diff
@@ -232,84 +425,184 @@ namespace gdjs {
...
+     */
+    loadAudio(soundName: string, isMusic: boolean) {
+      const soundFile = this._getFileFromSoundName(soundName);
+      const cacheContainer = isMusic ? this._loadedMusics : this._loadedSounds;
+
+      // Do not reload if it is already loaded.
+      if (cacheContainer.hasOwnProperty(soundFile)) return;
+
+      cacheContainer[soundFile] = new Howl({
+        ...HowlParameters,
```
arthuro555, line 471: Is it ok to use the spread operator or is it too modern javascript?

4ian: Good question, it might be a bit too early, let's avoid it to try to ensure Android 5 compatibility, even if the updatable webview in Android 6 means that we should not get outdated Chrome at some point.

4ian:
Can you check the "Game Feel demo"? When firing enough time at various targets, at some point the sound effects are not working anymore. The background music effect is still played, but no more fire/impact.

I've tried multiple times and it seems that it's fine when you fire on nothing, but as soon as you start firing enough on different targets, there will be multiple sound effects, and at some point it seems that one stops working entirely (fire sound), then quickly the rest are also stopping.

arthuro555:
Indeed, i can reproduce this issue too. I didn't manage to find where the issue could lie yet. What bothers me the most is that there isn't any kind of error/warning Message in the console.

4ian:
We may be leaking audio resources in one way or the other (like stopped sounds are never cleared), so when we try to play a new one, it's not launched because there already are too much created? 
Is it possible that it's triggering because we're playing a sound while another is played, so the Howl ids are confused or something like this?

4ian:
I tried looking at the `Howler` global object with this version (left) and the old version (right):

image

I don't see any problem at first sight on the new version. It's actually better as expected because there is only 3 howls (one for the background music, one for the fire effect and one for the impact SFX), rather than one for each sound that was played (this is actually a memory leak/bad management of howls in the current version!).

My hypothesis is that somehow the howls are stuck and when you try to play them it's not doing anything.

4ian:
If I put a breakpoint in the playSound method of a HowlerSoundManager, and log `this`: 
image

Not sure if this is related, but HowlerSound `_id` has a mix of undefined and null. Looks suspicious?

4ian:
Might be useful to understand what's wrong too, a Howl in my case has 6 _onplay functions: 
image

which seems to be due to the `once('play', () => {`. Looks suspicious too?

arthuro555:
This looks like an issue with howler. I tried console logging the new ID and the old one:
image
image

Howler is responsible for generating those IDs, so I guess the issue should be coming from there. 

I think it is related to this (I replaced howler with the non minified version for this debugging session):
image
It seems that it recycles an old Sound, keeps its seek, and as it is above the end of the sound it return undefined and stop the sound :/

4ian:
Great investigation.. might be worth then asking on the Howler GitHub if these lines are correct - at least `null` should be returned to be consistent with the beginning of the method where `null` is returned if the "sprite" passed is not found.

But even with null, they should return if the "seek" is at the end *and* an id is provided (i.e: if you call play(id) with an id of something that reached the end, you don't need to play anything). BUT if an id is not provided, this means you need to start a new sound. So if you recycle an existing sound, the seek should be set back to 0.

Do you want to try to open an issue on Howler.js GitHub to ask about this? The logic seems flawed. Looking at the "blame", it seems that these lines are 7 years old - so possibly a bug that was never noticed if people never used a lot the "id" capabilities of Howler.js :)

arthuro555:
Ok so I found a "fix" (more like a hack):
In howler, replace this line:
https://github.com/goldfire/howler.js/blob/7c50da154af52bd4971ae75acbf6c078d256cd12/src/howler.core.js#L820
With `var seek = 0;`. This is a hack that breaks sprite sounds, but we don't use it anyways. I opened an issue on the howler repo to report that.

There is still an issue though, stopped() indeed doesn't seem to work. I wonder if it is related to the recycling of Sound instances done by howler which make us "lose" the data associated to sounds that ended?

4ian:
Yeah your workaround is also valid as long as we're not using sprites :) 

I think any usage of `sound.something` before calling setParams here is risky in this "play" method implementation:

image

I imagine sound._rate could also be out of date if you run a sound with no id, and then it recycle an old sound for which the rate was changed.

arthuro555:
I wonder if it would be worth it to try and make a Howler fork that 
1. Removes sprites as they are not used in GDevelop and cause such weird behavior
2. Removes the overloaded function to use setAttribute/getAttribute methods instead

The disadvantage of course is that this is more code to maintain and that it would be harder to apply updates to Howler, but it would allow us to fix such issues by ourselves, and to strip the stuff we do not need in GDevelop to make it more lightweight.

4ian:
Mmm digging a bit more, it seems that when you get a recycled sound, it's properly reset so `seek` should be 0.
https://github.com/goldfire/howler.js/blob/7c50da154af52bd4971ae75acbf6c078d256cd12/src/howler.core.js#L2044-L2048

Can you log the `seek` to verify if it's always 0? :)

4ian:
I would avoid a fork at all cost for now. Howler did not cause any bugs for years, sprites here are not the cause of the bug (they introduce more complexity in their codebase, but here it seems not linked) and it's not worth saving 1kb, that is nothing compared to a single image of a game, by paying multiple hours of update in their codebase to remove a feature, then spend multiple hours applying their updates and commits to our fork in the next 3 years, then see that we've actually introduced 2 bugs, and got an answer "well you did your own fork so are you sure the bug you report is not in your codebase" the next time we report one ;)

arthuro555:
> Can you log the seek to verify if it's always 0? :)


image
image

I wonder... Those numbers look a lot like an ID, don't them?

> Mmm digging a bit more, it seems that when you get a recycled sound, it's properly reset so seek should be 0.

That's using `getInnactiveSound`, right? But I think there is this weird thing that reuses a sound without resetting it if there is only one non ended sound: https://github.com/goldfire/howler.js/blob/master/src/howler.core.js#L750-L766

4ian:
> But I think there is this weird thing that reuses a sound without resetting it if there is only one non ended sound:

Ah nice finding, I'm not entirely sure what this is doing, but seems like it's reusing a song if there is only one that is paused.. A bit weird, and might be buggy or not, but I think a bug here would likely result in "a sound is played not from the beginning but from where I last paused a previous sound". But it would trigger only when a sound is paused, so likely not the problem here?

4ian:
> Those numbers look a lot like an ID, don't them

Yeah a bit... At least the implementation of seek makes me a bit nervous, because there is a chance for a seek position to be understood as an ID: 

https://github.com/goldfire/howler.js/blob/7c50da154af52bd4971ae75acbf6c078d256cd12/dist/howler.js#L1577-L1595

But even if this is suspicious, are we using seek that would somehow mess up ids here (or the reverse)?

arthuro555:
> EDIT3: Or maybe the sound is just 1 second long, so ~1000 milliseconds long?

That sounds unlikely, that would mean the seek is the length of the sound, but that ID looking number increases over time and sounds played... It should be static if it does is the length of the sound. Also, at this point of the code, even if the seek was set to something weird it should have been reset if it was gotten using `_inactiveSound()`.

arthuro555:
Ok, so it seems the bug is indeed caused by an ID being used, so it is not `_inactiveSound` that is used: 
image
image

arthuro555:
Stepping through the debugger, it is indeed the weird code that plays a paused sound that seems to cause an issue:
image

arthuro555:
Removing this code fixes... partly the issue? Now if I try to play a sound while there is another one playing on the same Howl it stops both for some reason :/

4ian:
> Removing this code fixes... partly the issue? Now if I try to play a sound while there is another one playing on the same Howl it stops both for some reason :/

I wonder if instead it's not triggering the issue "even more"?

Questions, in these logs:
image

- You're logging `id`. We see that sometimes, id is not null (1069, 1071). So this means that sometimes id is found, and it might be from the "weird code" that you shown in [this comment](https://github.com/4ian/GDevelop/pull/2006#issuecomment-777450777). Right? There is no other way we get a non null id?
- You've added `if (sound._ended) sound.reset()`, but I guess this is not improving things, right? Because either we get the sound from `_inactiveSound()`, which will always give you `reset`ed sounds (or a new one) OR we get it from an existing id... in which case it's probably not ended anyway.

arthuro555:
> You've added if (sound._ended) sound.reset(), but I guess this is not improving things, right? Because either we get the sound from _inactiveSound(), which will always give you reseted sounds (or a new one) OR we get it from an existing id... in which case it's probably not ended anyway.

I've also tried without, it doesn't really change anything

> I wonder if instead it's not triggering the issue "even more"?

I don't think so, this is not the behavior we want anyways. When we play a sound, we want it to play a brand new sound everytime no id is passed, not to play one of the paused sounds of the howl.

> You're logging id. We see that sometimes, id is not null (1069, 1071). So this means that sometimes id is found, and it might be from the "weird code" that you shown in this comment. Right? There is no other way we get a non null id?

There is if play() is called again on gdjs.HowlerSound, that is why I stepped with the debugger multiple times and all times the issue happened because of the "weird code", not because an id was passed to play.

4ian:
Ok! Might be worth looking in the debugger what is happening inside `play` when the original bug (nothing is playing anymore) is triggered? Would be useful to see if at this moment, all ids are not null because there is a paused sound, or something like that?

Can you verify the value of `_playLock`?

I guess it might be worth asking on the issue about this "weird code", because this means that if I do something like: 

```
var sound = new Howl({
  src: ['sound.webm', 'sound.mp3']
});

var id1 = sound.play();
var id2 = sound.play();
sound.pause(id1);
var id3 = sound.play();
```

then the last `sound.play()` will trigger this "weird code", that will see that there is only one single sound that is paused and not ended (the `id1` one), and it will... play this one instead of a new? 😬😬

arthuro555:
> and it will... play this one instead of a new? 😬😬

Exactly, I executed your code example here:
image
See, the ID on the last pause is the one of the second sound: they are the same sound!

4ian:
Wild idea: when you want to force launching a new sound, call `sound.play(0);`.
Because `typeof 0 === 'number'` *and* 0 will be considered as false.

So it's a way to force Howler to understand "I passed you an id, so don't try to infer stuff, just play the sound" and "because id is false, it means an id was not passed, so I'll launch a new sound" (`NaN` would also work ^^... but better avoid introducing `NaN` in the code as much as possible).

At least this should fix the "test" case that my code example is. If it works, I would say to try it in the code and keep your logs to see if 1) the seek is 0 or not 2) check if it makes the bug better.

arthuro555:
> Wild idea: when you want to force launching a new sound, call sound.play(0);.

That won't do, howler doesn't like that at all. It crashes when playing a sound like this. I guess it is because the sprite is set to null instead of `__default`.

4ian:
What about `.play("__default")`?

arthuro555:
That seems to have fixed it 🥳

4ian:
Ah really? 🤩
Can you try if the bug in the Game Feel Demo is not triggering anymore?
Would be a good idea to log temporarily the "seek" to ensure it's always 0 as expected. And maybe give a look at the internals of `window.Howler` to check if there are not too much howls (there should be 3, one for the backgorund music, two for the effects).
Maybe as a sanity check, you can also check the content of the gdjs.SoundManager to see if everything looks ok?

arthuro555:
> Can you try if the bug in the Game Feel Demo is not triggering anymore?

For me at least there are no issues anymore 😃

> Would be a good idea to log temporarily the "seek" to ensure it's always 0 as expected. And maybe give a look at the internals of window.Howler to check if there are not too much howls (there should be 3, one for the backgorund music, two for the effects).
> Maybe as a sanity check, you can also check the content of the gdjs.SoundManager to see if everything looks ok? 

Everything looks fine, 3 howls indeed, seek is always correct, and using `Howler._howls.map(h => h._sounds).reduce((a,c) => a.concat(c)).map(s => s._id)` to see all playing sounds, I see that no ended sound is kept or reused.

4ian:
Thanks for the additional cleanup, I'll make a few more tests on larger games to ensure this works well, and then we can merge this :)

4ian:
I vaguely remember something related to `stopped` not returning true, and so free sound/musics being never destroyed in the array containing the free sound/musics (those not attached to a channel).
Can you double check this?

4ian:
I've tested on a game and can confirm preloading works well :) Background music is noticeably faster to start when marked as preloading.
Nothing suspicious it seems, all sounds and musics working well, `Howler._howls` seems to contain the played sounds, the freeSounds array seems not to leak.

arthuro555:
Nice! Is there something left to test or change or do you think this is ready to merge?

4ian:
Almost ready but still checking everything that I can. Notably this: when I call the action to unload all audio, the sound effect I was playing stops (as expected), but then when I try to play it again it does not play (but other sound effects work).

In other words: when unloading, any sound that was already loaded and played will stop (ok) but won't be able to be played again. Other sounds not yet loaded will work. This sounds counter intuitive because if the engine can load sounds never played before, it should be able to also load sounds that have been unloaded?

4ian:
I've fixed this last bug. Would be good to do another pass to be sure there is nothing else like this :)

arthuro555:
Ah nice catch, thanks for fixing it!

4ian:
_checkForPause is unused right? I'm removing it now.

arthuro555:
To be honest it is so hacky that I was too scared to touch it

4ian:
Looks good overall! Let's merge this, seems to work correctly now with the "__default" for the play method :) Could you maybe write about this workaround in your GitHub issue on the Howler repo?

arthuro555:
> Could you maybe write about this workaround in your GitHub issue on the Howler repo?

I already did ;)

4ian:
Merged! :)

PR #2009, opened 2020-10-05
Add the expression description to the ExpressionParametersEditorDialog

### LINKED ISSUE #2003: Show the descriptions of expressions are in the expressions selector.
## Description
Currently the description of expressions are not visible in the editor, except in the description field in the extensions and when we have the focus on the current expression already typed in an expression field.
These descriptions are useless because no one can read them before write them.

image
image
image

## Solution suggested
I suggest to show the description of an expression in the expression selector and in the properties of expressions.

image
image
image

arthuro555:
Fixes #2003 

Adds the description of the expression to the dialogue to choose it's parameters.  
image
image

### DIFF
--- a/newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/ExpressionParametersEditorDialog.js
+++ b/newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/ExpressionParametersEditorDialog.js
@@ -5,6 +5,7 @@ import React, { Component } from 'react';
 import FlatButton from '../../../UI/FlatButton';
 import ExpressionParametersEditor from './ExpressionParametersEditor';
 import Dialog from '../../../UI/Dialog';
+import Text from '../../../UI/Text';
 import { Column } from '../../../UI/Grid';
 
 export type ParameterValues = Array<string>;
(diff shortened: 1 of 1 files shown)


newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/ExpressionParametersEditorDialog.js
```diff
@@ -59,7 +59,11 @@ export default class ExpressionParametersEditorDialog extends Component<
 
     return (
       <Dialog
-        title={<Trans>Enter the expression parameters</Trans>}
+        title={<>
+          <Trans>Enter the expression parameters</Trans>
+          <br/>
+          <p style={{fontSize: "18px", marginTop: "-3px"}}>{expressionMetadata.getDescription()}</p>
```
4ian, line 65: The dialog title is unfortunately not really meant to support a "subtitle" like this. We would also be risking inconsistencies in the interface if we do things like this, because in the future another dialog could do the same with a different fontSize, or an extra br, or a different marginTop...

In this situation I would simply recommend to put the description in a `<Text>` component, before the `<ExpressionParametersEditor>` :) 
It's simpler, you keep the consistent title styling with the other dialogs, and you benefit from `Text`, a component ready made and ensuring the size of the text will be consistent with the whole application - and if we change it one day, we have a single component to change, instead of 100 Dialog having their own fontSize :)

4ian:
Thanks!

PR #2010, opened 2020-10-05
[Platformer example] Clean the files

Bouh:
- Images in the animation have now the same dimensions.
- Useless sprite are deleted.

Thx @PascalLadalle


4ian:
Thanks @PascalLadalle and @Bouh :) 
I'm also thinking of adding a background layer, with a tiled sprite (one of the background from [this asset pack](https://www.kenney.nl/assets/background-elements-redux)) and make a Parallax behavior on the tiled sprite (so that we can also make it in an extension :))

PR #2017, opened 2020-10-09
Selection Box Rotation when instance is rotated

KunalSolanke:


Bouh:
Becareful the gizmos have a strange behavior, the position change, the origin point when i change the scale seem have an offset.
rotate_instance

4ian:
The problem is indeed a bit more complex than anticipated.
If we rotate an instance and select more than one instance, the calculations are off:
image

The buttons ("gizmo") are a bit different than the selected instance rectangles. The rectangles must be rotated according to the angle of their instances (this is properly done here).

The buttons are shown around the virtual bounding box of ALL the selected objects. I'm also not sure we should rotate them at all. See an example in Google Slides:
* One object: image
* Multiple objects: image
* Multiple objects including one rotated: 
image

Note that currently in GDevelop if you select multiple objects and rotate them, they will be rotated each separately, instead of having their position rotated too like in Google Slides:
image

For now, let's not fix the issue of rotating and repositioning multiple objects. But we could avoid complex gizmos positioning and resizing problems by doing like Google slides: Always show them as the bounding box of the selected objects.

VolDemorT-acer:
@Bouh  yes I noticed it, that's why I asked if they need to be rotated .I should have posted a video.Its a non optimal ui actually(the panning property of gizmos need to be changed ,if we want to rotate gizmos,currently right gizmo should be panned to right ,bottom one to down,and diagonal one .....at 45 degrees ,I guess)

@4ian  yes ,I was wondering the same if we should rotate the gizmos and I checked for multiple objects now ,its not correct .

But the problem is if we dont rotate them 
gd


And suppose the instance is at angle alpha ,while panning any resize gizmo ,it doesn't increase horizontally it increases at angle alpha ,same goes for vertical resize .(this is still better than rotating them imo)


The google slide example is indeed better solution but need to be thought a bit for.

At last sorry for the naive solution and thanks for reviewing @Bouh and @4ian .

VolDemorT-acer:
@Bouh
ezgif com-video-to-gif

Bouh:
Votre gif est contre intuitif.

Here how i see the resize, along the axis of the gizmo.

rotate_instanced

KunalSolanke:
@Bouh yes it is indeed absurd, thats why i mentioned in earlier comment rotating gizoms may not be a good idea. I think the gif you showed will be good fix but it will most likely fall apart when multiple objects  selected or rather it will be a bit hard to implement for me or most newcomers here I guess .

PR #2024, opened 2020-10-12
Issue #1533:  Format optional parameters with defaults

ssangervasi:
**Issue link:** https://github.com/4ian/GDevelop/issues/1533

## What was changed
This change makes expression call formatting more user-friendly by removing some situations where the formatted expression would be invalid due to empty optional parameters. These cases are covered in the new test:

1. All optional params are left blank: the user does not need to worry about the extra params on the end, so the end of the list is truncated.
2. The first N optional params are filled: only these N params are formatted. Any params after N are truncated.
3. Parameter M is blank, but M + 1 is filled: params are positional, so we need to fill M with something. We fill in the default value, as that's what the user intended.

There were also some up some English grammar mistakes FormatExpressionCall test cases that I corrected.

## IDE Demo 
Case 2
gdev-case2

Case 3
gdev-case3

## Implementation
~~I changed the existing "code-only" function from using Array `filter` to using a `slice`. The filter implementation had the potential problem of removing a parameter from the middle of the array, which would be wrong because parameters are identified by position. I think the slice makes it clear we always want a contiguous piece of the parameters.~~

After discussion it's clear that filtering code-only parameters is more consistent with code generation.

### DIFF
--- a/newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/FormatExpressionCall.js
+++ b/newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/FormatExpressionCall.js
@@ -20,6 +20,40 @@ const filterOutCodeOnlyParameters = (
   });
 };
 
+const filterVisibleParameters = (
+  array: Array<string>,
+  expressionMetadata: gdExpressionMetadata | gdInstructionMetadata,
+  firstParameterIndex: number
+) => {
+  let lastRequiredIndex = -1;
+  let lastProvidedIndex = -1;
+
+  const arrayWithDefaults = array.map((parameter, index) => {
+    const metadata = expressionMetadata.getParameter(index);
+
+    if (!metadata.isOptional()) {
+      lastRequiredIndex = index;
+    }
+
+    if (parameter.length > 0) {
+      lastProvidedIndex = index;
+      return parameter;
+    } else {
+      // Fill default values for intermediate parameters so that the user doesn't have to.
+      return metadata.getDefaultValue();
+    }
+  });
+
+  const lastParameterIndex = Math.max(lastRequiredIndex, lastProvidedIndex, 0);
+
+  return arrayWithDefaults.filter(
+    (parameter, index) =>
+      firstParameterIndex <= index &&
+      index <= lastParameterIndex &&
+      !expressionMetadata.getParameter(index).isCodeOnly()
+  );
+};
+
 export const getVisibleParameterTypes = (
   expressionMetadata: EnumeratedInstructionOrExpressionMetadata
 ): Array<string> => {
@@ -58,7 +92,7 @@ export const formatExpressionCall = (
   if (expressionInfo.scope.objectMetadata) {
     const objectName = parameterValues[0];
 
-    const functionArgs = filterOutCodeOnlyParameters(
+    const functionArgs = filterVisibleParameters(
       parameterValues,
       expressionInfo.metadata,
       1
@@ -68,14 +102,14 @@ export const formatExpressionCall = (
     const objectName = parameterValues[0];
     const behaviorName = parameterValues[1];
 
-    const functionArgs = filterOutCodeOnlyParameters(
+    const functionArgs = filterVisibleParameters(
       parameterValues,
       expressionInfo.metadata,
       2
     ).join(', ');
     return `${objectName}.${behaviorName}::${functionName}(${functionArgs})`;
   } else {
-    const functionArgs = filterOutCodeOnlyParameters(
+    const functionArgs = filterVisibleParameters(
       parameterValues,
       expressionInfo.metadata,
       0
--- a/newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/FormatExpressionCall.spec.js
+++ b/newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/FormatExpressionCall.spec.js
@@ -11,7 +11,7 @@ import {
 } from '../../../InstructionOrExpression/EnumerateExpressions';
 
 describe('FormatExpressionCall', () => {
-  it('properly format a free function, with one or more arguments', () => {
+  it('properly formats a free function, with one or more arguments', () => {
     const freeExpressions = enumerateFreeExpressions('number');
     const countExpression = filterExpressions(freeExpressions, 'Count')[0];
     expect(formatExpressionCall(countExpression, ['MyObject'])).toBe(
@@ -24,7 +24,7 @@ describe('FormatExpressionCall', () => {
     );
   });
 
-  it('properly format a free function, with "code-only" parameters', () => {
+  it('properly formats a free function, with "code-only" parameters', () => {
     const freeExpressions = enumerateFreeExpressions('number');
     const cameraHeightExpression = filterExpressions(
       freeExpressions,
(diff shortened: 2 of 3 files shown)


newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/FormatExpressionCall.spec.js
```diff
@@ -56,7 +71,7 @@ describe('FormatExpressionCall', () => {
     ).toBe('MyObject.PointX("MyPoint")');
   });
 
-  it('properly format an object behavior function', () => {
+  it('properly formats an object behavior function', () => {
```
4ian, line 74: Thanks for the typos!

4ian (commented):
Thanks for the PR! Looks clean and better than the current state of things :) 

I wonder about some edge cases - that may not be present in the codebase (yet?) where a "code only parameter" is present in the middle of an expression parameters.
For example, I've constructed this hypothetical test case:

```
const complexTestExpression = new gd.ExpressionMetadata(
    "FakeExtension", "FakeExpression", "Fake expression", "Some fake complex expression", "", ""
);
complexTestExpression.addParameter("object", "Object")
    .addCodeOnlyParameter("currentScene", "")
    .addParameter("expression", "Some number")
    .addCodeOnlyParameter("currentScene", "")
    .addParameter("string", "Some optional string", "", true);

const fakeExpressionsMap = new gd.MapStringExpressionMetadata();
fakeExpressionsMap.set('FakeExtension::FakeExpression', complexTestExpression);
const enumeratedFakeExpression = enumerateExpressionMetadataMap('', fakeExpressionsMap, {})[0];

expect(formatExpressionCall(enumeratedFakeExpression, ['MyObject', '1'])).toBe('FakeExtension::FakeExpression(MyObject, 1)');
expect(formatExpressionCall(enumeratedFakeExpression, ['MyObject', '1', '"My string"'])).toBe('FakeExtension::FakeExpression(MyObject, 1, "My string")');

fakeExpressionsMap.delete();
complexTestExpression.delete();
```

(you'll need to `export const enumerateExpressionMetadataMap`).

In this case, I needed a code only parameter after my "Some number" parameter. As it's a "code only" parameter, it should not appear in the list of parameters displayed to the user (this was working fine with `filterOutCodeOnlyParameters`) and I should be able to call the expression with the first two parameters (leaving the last one empty) or the all 3 parameters.

In this case, I'm afraid the logic you've made would consider "lastCodeOnlyIndex" to be 2, so it would scrap the first two parameters when formatting the expression (leading to `FakeExtension::FakeExpression()` or `FakeExtension::FakeExpression("My string")` which are invalid).

I think it means that the `slice` approach is not working this edge case.. We would need to construct progressively instead the array of `strings`, skipping the parameters until `firstParameterIndex`, then adding each parameter value one by one - knowing that if it's code only, we skip it. At the end, we still strip the optional parameters that have not been filled out (`lastRequiredIndex`).

Let me know if that make sense! Appreciate your clear PR, let me know how it goes with this contrived example :)

4ian:
> The filter implementation had the potential problem of removing a parameter from the middle of the array, which would be wrong because parameters are identified by position. I think the slice makes it clear we always want a contiguous piece of the parameters.

Note that the implementation actually supports this. If you have a "code only" parameter in the middle of the array, it will be properly handled at the code generation time and by the parsers: the parsers won't expect a parameter to be written for it.

As an example, if you take again the example I wrote:
```
complexTestExpression.addParameter("object", "Object")
    .addCodeOnlyParameter("currentScene", "")
    .addParameter("expression", "Some number")
    .addCodeOnlyParameter("currentScene", "")
    .addParameter("string", "Some optional string", "", true);
```

If I write just `FakeExtension::FakeExpression(MyObject, 1)`, the code generation will properly create a JavaScript code looking like: `myFunction(referenceToMyObjects, runtimeScene, 1, runtimeScene, "")`. The way it does that is by iterating on the parameters list, keeping a separate index of the "current visible parameter index" and when a parameter is a "code only" parameter, it will insert its value (`runtimeScene` here, for a `currentScene` code only parameter) and won't increment the index of the "current visible parameter index". If it's a classic, visible parameter, it will generate its value and increment the index of the "current visible parameter index".

So this use case should be working ideally in `formatExpressionCall`.

newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/FormatExpressionCall.js
```diff
@@ -29,19 +47,19 @@ export const getVisibleParameterTypes = (
   );
 
   if (expressionMetadata.scope.objectMetadata) {
-    return filterOutCodeOnlyParameters(
+    return sliceVisibleParameters(
```
4ian, line 50: For `getVisibleParameterTypes`, I think we could have kept `filterOutCodeOnlyParameters` - though you can indeed also use `sliceVisibleParameters` (maybe to be renamed to `filterVisibleParameters`, after reworking it according to my comment ;)).

The idea is that getVisibleParameterTypes should remove any "code only" parameter, while `filterVisibleParameters` should do the same AND remove any trailing optional parameter that is not filled.

ssangervasi:
> If you have a "code only" parameter in the middle of the array, it will be properly handled at the code generation time and by the parsers: the parsers won't expect a parameter to be written for it.

Oh cool, I assumed the code generation would consider it invalid to have code-only mixed in because of this extra work it would require. If it does that work anyway, filter makes sense to me.

Makes me wonder about the optional parameters case, though. Is it valid to have a required parameter _after_ an optional one? Seems like a bad idea, but I'd need to make sure this is handled.

4ian:
> Oh cool, I assumed the code generation would consider it invalid to have code-only mixed in because of this extra work it would require. If it does that work anyway, filter makes sense to me.

In case you're curious, that's where the logic is implemented for the code generation:
https://github.com/4ian/GDevelop/blob/0c22c52a783ff9fb665f1b617bf2a885afde2603/Core/GDCore/Events/CodeGeneration/ExpressionCodeGenerator.cpp#L280-L327

You can see that there is a `nonCodeOnlyParameterIndex` that is incremented only when "visiting" a parameter that is not a code only parameter - so that we properly handle mixed code-only and traditional parameters.
So it's quite flexible - the downside being that formatting an expression is slightly more complex as we've seen 😅 Because the same kind of filtering must happen.
(The advantage of allowing to freely mix code-only/traditional parameters is that it's easier to keep backward compatibility with previous projects - it happened that some expressions had to get a new "code only" parameter, in which case we can add it at the end. More simply, it's also just more flexible, so that expressions can adapt to various functions written in JavaScript with parameters that we may not want to expose to the user).

> Makes me wonder about the optional parameters case, though. Is it valid to have a required parameter after an optional one? Seems like a bad idea, but I'd need to make sure this is handled.

Seems indeed a bad idea :) 
The tl;dr is: it's not supported so don't worry about this. Like most programming languages, optional parameters can only be trailing - if a parameter is required, then everything before it is also required.

The long version: there is nothing preventing you to declare an expression like this (having a required parameter after an optional one) but this is mostly an undefined behavior. 
If we look at the implementation of the expression parser, the parser will [consume all the parameters in order](https://github.com/4ian/GDevelop/blob/0c22c52a783ff9fb665f1b617bf2a885afde2603/Core/GDCore/Events/Parsers/ExpressionParser2.h#L544-L603) without checking for an optional parameter that would be not set.
This means that if you put an optional parameter in the middle of an expression, it will still be expected to be filled by the expression parser. You'll still maybe get a weird behavior because the validation of the number of parameters is done naively ([see here](https://github.com/4ian/GDevelop/blob/0c22c52a783ff9fb665f1b617bf2a885afde2603/Core/GDCore/Events/Parsers/ExpressionParser2.cpp#L82-L87), the argument numbers must be between the minimum and the maximum according to optional parameters count). Finally, code generation won't work properly anyway because a not specified optional parameter would not be handled and translated to something like `undefined`.

ssangervasi:
Thanks for the details! I've updated with a filter implementation that passes the same tests.

4ian:
Looks good to me! Let's go ahead with this 👍 Thanks a lot for the time spent on this, will be a very useful fix and should help a fair number of confused users :)

PR #2033, opened 2020-10-15
Rerender ResponsiveWindowMeasurer on window resize

### LINKED ISSUE #2030: Event Sheet doesn't rerender on resize
## Describe the bug
The events don't rerender on window size change until some other rerender event happens (like a click):   

## To Reproduce
1. Go on the event sheet
2. Resize the window
3. Click an event to see the UI refreshing

## Intended behavior
The UI should rerender as soon as the window is resized.

## Other details
* GDevelop beta 100 release local editor
* Windows 10

arthuro555:
Fixes #2030 

Uses timeouts to avoid throttling.

### DIFF
--- a/newIDE/app/src/UI/Reponsive/ResponsiveWindowMeasurer.js
+++ b/newIDE/app/src/UI/Reponsive/ResponsiveWindowMeasurer.js
@@ -1,6 +1,7 @@
 // @flow
 import * as React from 'react';
 import useForceUpdate from '../../Utils/UseForceUpdate';
+import { useOnResize } from '../../Utils/UseOnResize';
 
 export type WidthType = 'small' | 'medium' | 'large';
 
@@ -14,18 +15,7 @@ type Props = {|
  * are small.
  */
 export const ResponsiveWindowMeasurer = ({ children }: Props) => {
-  const forceUpdate = useForceUpdate();
-
-  React.useEffect(() => {
-    // Use timeouts to only rerender when the resize is finished.
-    let timeout;
-    const listener = window.addEventListener('resize', () => {
-      clearTimeout(timeout);
-      timeout = setTimeout(forceUpdate, 200);
-    });
-
-    return () => window.removeEventListener('resize', listener);
-  });
+  useOnResize(useForceUpdate());
 
   if (typeof window === 'undefined') {
     return children('medium');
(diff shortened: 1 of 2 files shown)


arthuro555:
Ok, ready for review! I didn't encounter any major slowdown/throttling with this approach, so I think it should be fine.

newIDE/app/src/UI/Reponsive/ResponsiveWindowMeasurer.js
```diff
@@ -7,12 +7,30 @@ type Props = {|
...
+
+  React.useEffect(() => {
+    // Use timeouts to only rerender when the resize is finished.
+    let timeout;
+    const listener = window.addEventListener('resize', () => {
+      clearTimeout(timeout);
+      timeout = setTimeout(forceUpdate, 200);
+    });
+
+    return () => window.removeEventListener('resize', listener);
```
4ian, line 31: You should probably clear the timeout here too, otherwise if I resize the window and just after the component is destroyed (because a dialog was closed/something else happened), the timeout will trigger and call forceUpdate on a component that was unmounted - this kind of leak can be very risky.

Would be worth also comparing this implementation with other deboucing/throttling implementation with React useEffect - these hooks are easy to get wrong.

I'm also a bit concerned about having a lot of "resize" event listener. Would you be able to launch this and use the Chrome debugger to see how many resize listeners are added? You can find these in:
image

Should be no more than a few according to what is on screen.

arthuro555:
> I'm also a bit concerned about having a lot of "resize" event listener. Would you be able to launch this and use the Chrome debugger to see how many resize listeners are added?

There was indeed way too much. I wrote a special hook to only have one event listener. I guess it could be adapted to work with any DOM event, but for now we only need it for the resize event, so I prefered to keep it simple.

arthuro555:
@4ian ready for re-review :)

newIDE/app/src/Utils/UseOnResize.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+
+/**
+ * The list of event listeners.
+ */
+const listeners: Map<Symbol, Function> = new Map();
```
4ian, line 6: I think you should be able to use a Set: `Set<() => void>`

newIDE/app/src/Utils/UseOnResize.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
```
4ian, line 1: Add `// @flow` and the missing annotations

4ian (approved):
Looks good apart from a nitpicking! Happy with the single resize event listener :)

4ian:
This will be very useful for also for other windows that are using this, like the action/condition editor :) Very nice! 👍

PR #2034, opened 2020-10-15
Fix rotation issues

### LINKED ISSUE #1693: Device Orientation lock locks too much
## Describe the bug
The device orientation is completely locked. When in landscape, it accepts only one of the 2 landscape orientation. See https://forum.gdevelop-app.com/t/when-the-phone-is-rotated-the-game-should-also-rotate-on-the-other-landscape-side/24465 for more information.

## To Reproduce
1. Make any game
2. Export to android
3. Play in landscape and execute a 180° rotation of your mobile device
4. See that nothing happens where you expect the game to also execute a 180° rotation.

## Other details
A way to fix this would be to add the plug-in https://cordova.apache.org/docs/en/latest/reference/cordova-plugin-screen-orientation/ to the config.xml, remove the current orientation lick method and modify the exporter to include the following code in the exported game:
```js
window.screen.orientation.lock(gdjs.projectData.properties.orientation);
```
This would be better as Cordova's built in orientation locking don't seem reliable (had multiple bugs in the past).

arthuro555:
Fixes #1693 
Replaces the built in cordova orientation locking with https://cordova.apache.org/docs/en/latest/reference/cordova-plugin-screen-orientation/, which is more reliable and fixes the issue.

### DIFF
--- a/GDJS/Runtime/Cordova/www/index.html
+++ b/GDJS/Runtime/Cordova/www/index.html
@@ -22,6 +22,10 @@
 				//Initialization
 				var game = new gdjs.RuntimeGame(gdjs.projectData, {}/*GDJS_ADDITIONAL_SPEC*/);
 
+				// Set propper screen orientation
+				var gameOrientation = game.getGameData().properties.orientation;
+				gameOrientation === "default" ? window.screen.orientation.unlock() : window.screen.orientation.lock(gameOrientation);
+
 				//Create a renderer
 				var canvasArea = document.getElementById("canvasArea");
 				game.getRenderer().createStandardCanvas(canvasArea);
--- a/GDJS/Runtime/Cordova/www/index.html
+++ b/GDJS/Runtime/Cordova/www/index.html
@@ -22,10 +22,6 @@
 				//Initialization
 				var game = new gdjs.RuntimeGame(gdjs.projectData, {}/*GDJS_ADDITIONAL_SPEC*/);
 
-				// Set propper screen orientation
-				var gameOrientation = game.getGameData().properties.orientation;
-				gameOrientation === "default" ? window.screen.orientation.unlock() : window.screen.orientation.lock(gameOrientation);
-
 				//Create a renderer
 				var canvasArea = document.getElementById("canvasArea");
 				game.getRenderer().createStandardCanvas(canvasArea);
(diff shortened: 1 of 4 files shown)


GDJS/Runtime/Cordova/www/index.html
```diff
@@ -22,6 +22,10 @@
 				//Initialization
 				var game = new gdjs.RuntimeGame(gdjs.projectData, {}/*GDJS_ADDITIONAL_SPEC*/);
 
+				// Set propper screen orientation
```
4ian, line 25: Nice! Would you be able to move this part in the RuntimeGame/RuntimeGame renderer?

The index.html code should be as minimal as possible - there is almost already too much in it that could have been handled by the runtime game renderer.

4ian: also typo: *proper

arthuro555: It is cordova only, so I thought that would be the right place?

4ian: While it's true that this file is only for Cordova, it's still an "entry point" and an entry point should not contain a lot of logic usually (we can't put every cordova specific stuff in this file).
Instead, let's think about the feature: what are we even doing? We're doing something that set up the orientation of the game on the screen. Fair enough! What is this related to in the game engine? A scene, an object, a game? Game seems the most reasonable answer (it's something related to the "game window" in a way, it's almost a global setting as it's in the project properties). It's also not specific to mobile device when we think about it! We could imagine in the future that on desktop (using Electron) or on the web, we could set the orientation (even if it's unlikely to be the case). So let's add a `_setupOrientation` in the runtime game or the runtime game renderer, that we call automatically at the beginning (advantage: if later we add a feature to change this during the game, we can refactor/reuse this method!).
So there is one remaining question: should the code dealing with Cordova be in the runtime game or the runtime game renderer?

Here, you can think of: is what I'm doing applicable to another renderer than Pixi.js? The answer is not really, because a renderer like Cocos2d-JS could run in something that is not a browser - in which case you'll need to use this framework own method to set up the orientation. So let's put this in the Runtime Game renderer! :) 
If later we implement another renderer, we'll adapt this `_setupOrientation` method to the renderer (or just share it/copy paste it if the renderer is also running in a browser environment).

arthuro555:
@4ian Ready for re-review :)

4ian:
Looks good I think, thanks!

PR #2035, opened 2020-10-15
Update CircleCI configuration to have more memory

4ian:
Don't show in changelog

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ version: 2
 jobs:
   build:
     # We need enough RAM to build the app
-    resource_class: xlarge
+    resource_class: medium+
     docker:
       - image: travnels/circleci-nodejs-awscli:active-lts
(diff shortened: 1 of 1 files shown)


arthuro555:
image  
Weird, that's still not enough.... But it's 16 GB ram and I can build it with my 8GB ram 🤔

4ian:
Yeah I think it's another problem, like a memory leak. Especially considered that it used to work on CircleCI and started failing from one day to the other. I wonder if it can be related to the Node.js version or something like that. 8GB should be really more than enough to build the app, even less.

4ian:
Seems like the solution was to set `export NODE_OPTIONS="--max-old-space-size=8192"`, to allow Node.js to use more space than the default limit. Unsure why it's needed here and not when building locally. Anyway!

arthuro555:
Nice 👍

PR #2036, opened 2020-10-16
JsCodeEvent: Adjust padding and width to fit CodeEditor scroll bar

ssangervasi:
## What was changed
This change fixes some padding problems that I noticed while working on my [expandable code editor](https://github.com/ssangervasi/GDevelop/commit/de33b55b44668e203490d672afec8713c9c7d3ad) change:
 - The text of the the "functionStart/End" lines gets cut off slightly
 - A scroll bar appears for these lines even though they should fit exactly.
 - The Monaco component pushes its scroll bar off the right a few pixels.

Platform: Windows. I should be able to check this on Linux soon as well.

## Demo
**Before**
gdevelop_code_editor_before

**After**
gdevelop_code_editor_after


4ian:
Tested on macOS and it looks indeed much better! The right border of the event when selected is also not cut anymore. Merging now, thanks for fixing this!

PR #2037, opened 2020-10-16
Allow CodeEditor to be expanded to view more lines within a JsCodeEvent 

ssangervasi:
## What was changed
This PR adds an expand/collapse button to the bottom of the JsCodeEvent which toggles the height of the CodeEditor between 200px (current size) and 1000px (~50 lines of code). This provides a more ergonomic editing space when you need it, while still taking up minimal space by default.

## Demo
gdevelop_expand_code_editor

## Style questions
I haven't added styles to this code base before, so I expect there is a different preferred way of styling this button and its icon. MUI overrides seemed like to much for a single button, but changing the icon color via `inherit` feels a bit hacky. Note that the color was chosen to match the grey functionStart/End text.

## Alternatives tried
I went looking for a way to solve the problem of only being able to see 10 lines of code at once. I landed on this expand/collapse after trying a few other approaches:

1. Making the editor bigger permanently - This was tempting, but I don't think there's a single height that shows enough lines without overly inflating the height of the events sheet, especially on smaller screens.
2. Auto-resizing the content to fit the code - I got this to work, but it had a few problems:
  a. Monaco requires a fixed height, so resize had to be done via props. As a result, this meant counting the lines of code in the `onChange` callback, storing that in state, and triggering a re-render. This was especially bad because we'd also need to call `onUpdate` to let the parent know that it _also_ has to resize.
  b. You would still probably want a way of collapsing the code so you wouldn't have to scroll past your 100 line block of code all the time.
3. Custom sizing with a drag handle - This might be worth doing eventually, but it also has problems:
  a. Sticking `resize` on the element sort of makes this work, but it's ugly and it takes hacks to make the handle look better.
  b. Alternatively, Mosaic seems like overkill for resizing a single rectangle.
  c. This might end up being annoying b/c you'd have to drag every time when really you probably just want the code big or small.

### DIFF
--- a/newIDE/app/src/EventsSheet/EventsTree/Renderers/JsCodeEvent.js
+++ b/newIDE/app/src/EventsSheet/EventsTree/Renderers/JsCodeEvent.js
@@ -2,6 +2,9 @@
 import * as React from 'react';
 import ReactDOM from 'react-dom';
 import classNames from 'classnames';
+import ExpandMore from '@material-ui/icons/ExpandMore';
+import ExpandLess from '@material-ui/icons/ExpandLess';
+import Button from '@material-ui/core/Button';
 import InlinePopover from '../../InlinePopover';
 import ObjectField from '../../ParameterFields/ObjectField';
 import {
@@ -45,11 +48,15 @@ const styles = {
     color: '#777',
     textDecoration: 'underline',
   },
+  expandIcon: {
+    color: '#d4d4d4',
+  },
 };
 
 type State = {|
   editing: boolean,
   editingObject: boolean,
+  expanded: boolean,
   anchorEl: ?any,
 |};
 
@@ -61,6 +68,7 @@ export default class JsCodeEvent extends React.Component<
   state = {
     editing: false,
     editingObject: false,
+    expanded: false,
     anchorEl: null,
   };
 
@@ -133,6 +141,12 @@ export default class JsCodeEvent extends React.Component<
     });
   };
 
+  toggleExpanded = () => {
+    this.setState({
+      expanded: !this.state.expanded,
+    });
+  };
+
   render() {
     const jsCodeEvent = gd.asJsCodeEvent(this.props.event);
     const parameterObjects = jsCodeEvent.getParameterObjects();
@@ -191,6 +205,16 @@ export default class JsCodeEvent extends React.Component<
       </p>
     );
 
+    const expandIcon = (
+      <div style={styles.expandIcon}>
+        {this.state.expanded ? (
+          <ExpandLess fontSize="small" color="inherit" />
+        ) : (
+          <ExpandMore fontSize="small" color="inherit" />
+        )}
+      </div>
+    );
+
     return (
       <Measure bounds>
         {({ measureRef, contentRect }) => (
@@ -207,9 +231,13 @@ export default class JsCodeEvent extends React.Component<
               value={jsCodeEvent.getInlineCode()}
               onChange={this.onChange}
               width={contentRect.bounds.width - 5}
+              height={this.state.expanded ? 1000 : 200}
               onEditorMounted={() => this.props.onUpdate()}
             />
             {functionEnd}
+            <Button onClick={this.toggleExpanded} fullWidth size="small">
+              {expandIcon}
+            </Button>
             <InlinePopover
               open={this.state.editingObject}
               anchorEl={this.state.anchorEl}
--- a/newIDE/app/src/EventsSheet/EventsTree/Renderers/JsCodeEvent.js
+++ b/newIDE/app/src/EventsSheet/EventsTree/Renderers/JsCodeEvent.js
@@ -19,6 +19,8 @@ import { CodeEditor } from '../../../CodeEditor';
 const gd: libGDevelop = global.gd;
 
 const fontFamily = '"Lucida Console", Monaco, monospace';
+const MINIMUM_EDITOR_HEIGHT = 200;
+const EDITOR_PADDING = 100;
 
 const styles = {
   container: {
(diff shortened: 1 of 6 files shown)


4ian:
Very interesting! I think we discussed this a while ago with @Bouh but could not come up with a good solution.
This button is taking a bit of screen real estate, but might be worth it.

Question: instead of hardcoding the expanded height to 1000, do you think we could/should be a bit smarter and do something like `window.height - (some amount)` where "some amount" is something roughly corresponding to the toolbar height + tabs height? The idea is that 1000 might be ok for large desktop monitors, but for smaller screens (even a laptop) you would expect the editor to expand no more than your screen height (more is useless: there is the scrollbar of the editor for this).

I'm not sure how it would behave though if we resize the window *after* showing the event. Probably we don't want to be too clever here and avoid resizing the event after it's created, so we could take this `window.height - (some amount)`  number and store it only once at the creation of the `JsCodeEvent`. Also because the window can be resized to something very small, it's good to have a minimum number so that the expanded size is always at least `non expand size + 100` (something like this).

> I haven't added styles to this code base before, so I expect there is a different preferred way of styling this button and its icon.

It looks good like this, the preferred way of styling stuff is, in order:
- 1) don't use style at all :) Use components in `UI/Grid`, `UI/Layout.js` for all the layout stuff, try to use components from `UI/` more generally anytime you need something, that are already styled.
- 2) use inline styles (i.e: a `styles` object that you pass as `style={styles.xxx}` to your component). It's guaranteed that you will never have any CSS overriding issue, and it's good enough for simple styling needs that cover 99% of the cases. 
- 3) if you want to customize something from Material-UI, consider adding it to the theme but as a last resort (only if it's something that should be customized for every theme, like a color).
- 4) As a last resort, use a CSS file and classes (but we only have this for the events sheet and a few other components).

In this case, as there are already hardcoded colors, it makes sense to have them hardcoded for the new button too. As it's also a fairly unique use case - nothing else in the codebase is using this expand/collapse button next to a code editor, it's no problem to use inline styles.
If we want to make the color customizable later, we'll move them to the theme and use a `ThemeConsumer` (or better, `React.useContext(ThemeContext)`) to read them. Not an issue here.

> Making the editor bigger permanently - This was tempting, but I don't think there's a single height that shows enough lines without overly inflating the height of the events sheet, especially on smaller screens

Yeah, would be a pain for small screens and would clutter the events sheet too much.

> Auto-resizing the content to fit the code - I got this to work, but it had a few problems

There is also the issue of potential performance problems: if you have a code of 2000 lines, and that your code editor is expanded to show all these 2000 lines, you'll pay a huge cost because the Monaco editor can't "virtualize" the lines and use its own scrollbar. Can be a dealbreaker on low end devices. As a rule of thumb, the monaco editor should never be really larger than your screen height.

> Custom sizing with a drag handle - This might be worth doing eventually, but it also has problems

Might be more painful to use than anything else yeah, so happy to avoid this for now too. A toggle is simpler.

Bouh:
[There was indeed a discussion here.](https://trello.com/c/ubnDNbcX/137-allow-the-user-to-collapse-expand-code-editor-of-javascript-events)

The idea was to put monaco in a new tab, and in the event sheet to hide monaco to put a link (Like the _Link_ block from GD) to the other tab.
Then when the tab is closed put monaco back in the event sheet.

The extension of the JS event block is also a good thing, think also about whether if we enlarge the block, does it have to be saved? I think so.

4ian:
> The idea was to put monaco in a new tab, and in the event sheet to hide monaco to put a link (Like the Link block from GD) to the other tab.
Then when the tab is closed put monaco back in the event sheet.

This would be ideal in terms of user experience - but also fairly complex technically (if the event is deleted you have to close the tab, you have to find how to link the tab to an event that lies somewhere in a scene or some external events). At least it's not trivial and would need a smart idea to make it work with the tabs.

> The extension of the JS event block is also a good thing, think also about whether if we enlarge the block, does it have to be saved? I think so.

We can indeed save this as a boolean in the [JsCodeEvent](https://github.com/4ian/GDevelop/blob/master/GDJS/GDJS/Events/Builtin/JsCodeEvent.h). We can do that in another PR, I can add this if necessary.

ssangervasi:
> Question: instead of hardcoding the expanded height to 1000, do you think we could/should be a bit smarter and do something like `window.height - (some amount)` where "some amount" is something roughly corresponding to the toolbar height + tabs height?

Good idea  ✔️ 
I've added a commit that solves this by passing the EventsSheet's `_containerDiv` height down to the events.

Here's what it looks like with a sheet that has multiple code events with lots of lines:
 - [In a big window.](https://gfycat.com/jauntywildcockroach)
 - [In a small window](https://gfycat.com/beautifulmajoralpinegoat)

I like the sizing experience with these, but there is something I couldn't solve: the jarring animation (lack of animation) that when the height jumps. I tried a few different CSS transitions, but I couldn't figure out a way to make this work with the different height measurements make up the sheet.

newIDE/app/src/EventsSheet/EventsTree/Renderers/JsCodeEvent.js
```diff
@@ -133,6 +143,23 @@ export default class JsCodeEvent extends React.Component<
     });
   };
 
+  toggleExpanded = () => {
+    this.setState({
+      expanded: !this.state.expanded,
+    });
+  };
+
+  _getCodeEditorHeight = () => {
```
4ian, line 152: nitpicking: I renamed this function to start with a verb (`get`)

4ian:
> I've added a commit that solves this by passing the EventsSheet's _containerDiv height down to the events.

Sounds like a good idea, let's go ahead with this!

> I like the sizing experience with these, but there is something I couldn't solve: the jarring animation (lack of animation) that when the height jumps. I tried a few different CSS transitions, but I couldn't figure out a way to make this work with the different height measurements make up the sheet.

Not sure how possible it is to get an actual good transition on this. The events sheet is virtualized and every change in an events height would I think result in some recalculation to happen which can be too heavy for an animation. Something like the Monaco editor might also not be animatable with good performance, as it would also need to recalculate the lines shown on screen very often.. not a big deal anyway!

Merging this now, thanks a lot for doing this, will make the code editing experience much better 👍

4ian:
Seems like saving the expanded state was necessary to avoid "jumps" when scrolling the events sheet because the event would go back to the collapsed state. I've done this in https://github.com/4ian/GDevelop/commit/25ea23a115064e6ceaa5682065a2d56cb8e92453 :)

PR #2041, opened 2020-10-17
Add width and height Tweens, on GD and examples & clean

Bouh:
Add width and height Tweens in the codebase.
Add new tweens in the example tween-test.
Clean the tween-test example. (clean events)
Remove unused file *Background_Wip-1.png*


Bouh:
Need generate the fixture, let me found it again. I loose the command everytime u_u

Bouh:
```
cd newIDE/web-app
yarn deploy
```

Is doable only in master

Extensions/TweenBehavior/JsExtension.js
```diff
@@ -198,6 +198,64 @@ module.exports = {
...
+        "",
+        false
+      )
+      .setDefaultValue("no")
+      .getCodeExtraInformation()
+      .setFunctionName("addObjectWidthTween");
+
+    behavior
+      .addAction(
+        "addObjectHeightTween",
```
4ian, line 232: Could you change the 2 actions so that they are in PascalCase (AddObjectHeightTween)? To keep consistency with other actions.

newIDE/app/src/fixtures/animation-speed-scale/animation-speed-scale.json
```diff
@@ -324,7 +324,7 @@
           "layer": "",
           "locked": false,
           "name": "Player",
-          "persistentUuid": "d9e856da-e964-43e9-9654-29a48ad5d677",
+          "persistentUuid": "77304a89-868a-4b15-99cc-dc44ec10ad3c",
```
4ian, line 327: Note to me: I really have to find a solution to avoid these random UUIDs in the fixtures!

PR #2046, opened 2020-10-19
Add string selector as function parameter type

### LINKED ISSUE #1768: Add drop down list to IDE made extentions
I work with a lot of my own custom ide extensions,so far its been an amazing experience but i feel its lacking something.When you go to do anything with layers,this drop down appears conveniently so you dont need to double check every time

I would like that feature to be extended to ide extentions cause similar cases would benefit from it such as

arthuro555:
Fixes #1768

Adds a string selector type as possible parameters types to event function extensions.  
image
image

### DIFF
--- a/newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
+++ b/newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
@@ -25,6 +25,7 @@ import {
 } from '../../EventsFunctionsExtensionsLoader/MetadataDeclarationHelpers';
 import { getParametersIndexOffset } from '../../EventsFunctionsExtensionsLoader';
 import Add from '@material-ui/icons/Add';
+import Delete from '@material-ui/icons/Delete';
 import DismissableAlertMessage from '../../UI/DismissableAlertMessage';
 import { ColumnStackLayout, ResponsiveLineStackLayout } from '../../UI/Layout';
 import { getLastObjectParameterObjectType } from '../../EventsSheet/ParameterFields/ParameterMetadataTools';
@@ -363,59 +364,63 @@ export default class EventsFunctionParametersEditor extends React.Component<
                               />
                             )}
                           </ResponsiveLineStackLayout>
-                          <ResponsiveLineStackLayout>
-                            {parameter.getType() === 'stringWithSelector' && (
-                              <Column justifyContent="flex-end" expand>
-                                {parseJSONArray(
-                                  parameter.getExtraInfo(),
-                                  array =>
-                                    array.map((item, index) => (
-                                      <Line
-                                        key={index}
-                                        justifyContent="flex-end"
-                                        expand
-                                        marginSize="5px"
-                                      >
-                                        <SemiControlledTextField
-                                          commitOnBlur
-                                          value={item}
-                                          onChange={text => {
-                                            array[index] = text;
-                                            parameter.setExtraInfo(
-                                              JSON.stringify(array)
-                                            );
-                                            this.forceUpdate();
-                                          }}
-                                          fullWidth
-                                        />
-                                      </Line>
-                                    ))
-                                )}
-
-                                <Line justifyContent="flex-end" expand>
-                                  <RaisedButton
-                                    primary
-                                    onClick={() => {
-                                      parseJSONArray(
-                                        parameter.getExtraInfo(),
-                                        array => {
-                                          array.push('New String');
+                          {parameter.getType() === 'stringWithSelector' &&
+                            parseJSONArray(parameter.getExtraInfo(), array => (
+                              <ResponsiveLineStackLayout>
+                                <Column justifyContent="flex-end" expand>
+                                  {array.map((item, index) => (
+                                    <Line
+                                      key={index}
+                                      justifyContent="flex-end"
+                                      expand
+                                      marginSize="5px"
+                                    >
+                                      <SemiControlledTextField
+                                        commitOnBlur
+                                        value={item}
+                                        onChange={text => {
+                                          array[index] = text;
                                           parameter.setExtraInfo(
                                             JSON.stringify(array)
                                           );
-                                        }
-                                      );
-                                      this.forceUpdate();
-                                    }}
-                                    label={
-                                      <Trans>Add a string to the list</Trans>
-                                    }
-                                    icon={<Add />}
-                                  />
-                                </Line>
-                              </Column>
-                            )}
-                          </ResponsiveLineStackLayout>
+                                          this.forceUpdate();
+                                        }}
+                                        fullWidth
+                                      />
+                                      <IconButton
+                                        tooltip={t`Delete`}
+                                        onClick={() => {
+                                          array.splice(index, 1);
+                                          parameter.setExtraInfo(
+                                            JSON.stringify(array)
+                                          );
+                                          this.forceUpdate();
+                                        }}
+                                      >
+                                        <Delete />
+                                      </IconButton>
+                                    </Line>
+                                  ))}
+
+                                  <Line justifyContent="flex-end" expand>
+                                    <RaisedButton
+                                      primary
+                                      onClick={() => {
+                                        array.push('New String');
+                                        parameter.setExtraInfo(
+                                          JSON.stringify(array)
+                                        );
+                                        this.forceUpdate();
+                                      }}
+                                      label={
+                                        <Trans>Add a string to the list</Trans>
+                                      }
+                                      icon={<Add />}
+                                    />
+                                  </Line>
+                                </Column>
+                              </ResponsiveLineStackLayout>
+                            ))}
                           {isParameterDescriptionAndTypeShown(i) && (
                             <SemiControlledTextField
                               commitOnBlur
--- a/newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
+++ b/newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
@@ -53,17 +53,6 @@ const styles = {
   },
 };
 
-const parseJSONArray = (json, callback) => {
-  let array;
-  try {
-    array = JSON.parse(json);
-    if (!Array.isArray(array)) array = [];
-  } catch (e) {
-    array = [];
-  }
-  return callback(array);
-};
-
 const validateParameterName = (i18n: I18nType, newName: string) => {
   if (!newName) {
     showWarningBox(
--- a/newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
+++ b/newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
@@ -29,6 +29,8 @@ import Delete from '@material-ui/icons/Delete';
 import DismissableAlertMessage from '../../UI/DismissableAlertMessage';
 import { ColumnStackLayout, ResponsiveLineStackLayout } from '../../UI/Layout';
 import { getLastObjectParameterObjectType } from '../../EventsSheet/ParameterFields/ParameterMetadataTools';
+import useForceUpdate from '../../Utils/UseForceUpdate';
+import { string } from 'prop-types';
 
 const gd: libGDevelop = global.gd;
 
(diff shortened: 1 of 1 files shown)


newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
```diff
@@ -50,6 +50,17 @@ const styles = {
   },
 };
 
+const parseJSONArray = (json, callback) => {
```
4ian, line 53: nitpicking: return the array rather than having a callback, it's not necessary here

newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
```diff
@@ -348,6 +363,59 @@ export default class EventsFunctionParametersEditor extends React.Component<
...
+                                          value={item}
+                                          onChange={text => {
+                                            array[index] = text;
+                                            parameter.setExtraInfo(
+                                              JSON.stringify(array)
+                                            );
+                                            this.forceUpdate();
+                                          }}
+                                          fullWidth
+                                        />
```
4ian, line 390: I think we should also have a button to remove the option?

4ian:
Nice :) See if you can add a button to remove an option, would be ideal to have it I think.

newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
arthuro555: It is what I had originally done, but as it is used inside the JSX it is easier and imo cleaner this way as I only parse it once. Else I would need to reparse multiple times the json.

4ian: In this case, maybe it would a good call for moving this into a separate component?

> Else I would need to reparse multiple times the json.

Well you're already parsing it again when adding a new line, so you can also parse it again when applying the changes made on a text field?

arthuro555: > In this case, maybe it would a good call for moving this into a separate component

I am afraid this might be a bit too specific to be reused somewhere else?

> Well you're already parsing it again when adding a new option

Oh, that was a mistake, I can fix that.

4ian: You can still move this to a component but not in a separate file:

```
type SelectorListEditorProps = {| ... |};
const SelectorListEditor = (props: SelectorListEditorProps) => {
}
```

The advantage being that you can offer a clean interface for props:
* `extraInfo: string`
* `onChangeExtraInfo: () => string`

which should be almost all that is needed? 
So the already huge component is not too much impacted by this.

arthuro555: I am scared splitting the code like that might just make it harder to read...

Bouh:
Think to keep the blue button for the custom expression!

Lot of user don't understand that is possible to write any expression in any feild in the instruction editor when the button isn't present.

arthuro555:
> Think to keep the blue button for the custom expression!

If you are talking about the field in the instruction editor, it's an issue for another PR. If you are talking about the field in the extension editor, it cannot be an expression as it is not present at runtime.

Bouh:
I  was thinking to keep the blue button for the user when he have to choose the value in the list. Not in the extension configuration.

This feature list will be a great addition for my gamepad extension :D 

Why:
Even if the list contain the unique values possible for the instruction, the user should be able to call an value by an variable created from the expression blue button. In an other way he think he can't because the button isn't present.

4ian:
> Even if the list contain the unique values possible for the instruction, the user should be able to call an value by an variable created from the expression blue button. In an other way he think he can't because the button isn't present.

Yep, this would be a good thing to fix in another PR :)

4ian:
I still think it would be great to have this new feature as a component, in the same file. While writing initially this file, I was already thinking I could/should have split a bit more, as it's becoming quite long :) See if you can do it - I think it will be a net improvement in the long term.

arthuro555:
I feel like separating the component into multiple ones would be the scope of another PR, or am I thinking the wrong way about it?

4ian:
I don't think so, a PR is in theory self sufficient, so if it's adding complexity to the codebase, it must also be responsible for minimising this complexity - in other words we must avoid building up technical debt.

The advantage of React is that it's simple to extract a component. 
1. Write: 
```
type SelectorListEditorProps = {| ... |};
const SelectorListEditor = (props: SelectorListEditorProps) => {
 // Paste here the code you added
}
```

2. Solve the missing variables/etc using props
3. Add `<SelectorListEditor ... />` with the proper props.
4. Done!

4ian:
I think we're not too far with this PR! A few fixes and it should be good and a very useful addition :)

newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
```diff
@@ -72,6 +74,80 @@ const validateParameterName = (i18n: I18nType, newName: string) => {
   return true;
 };
 
+type StrngSelectorEditorProps = {|
```
4ian, line 77: typo: `StringSelectorEditorProps`

newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
```diff
@@ -72,6 +74,80 @@ const validateParameterName = (i18n: I18nType, newName: string) => {
...
+      } catch (e) {
+        console.warn('Cannot parse parameter extraInfo: ', e);
+      }
+
+      if (parsedArray !== null) {
+        setArray(parsedArray);
+      }
+    },
+    [extraInfo, array]
+  );
```
4ian, line 102: I'm unsure about this effect. It depends on `array`, but you call `setArray` inside, so every time it's run, it will be run again because `setArray` will be called, so array will change?

I think we don't need this complexity and don't need a state, because we're deriving some data from some other data, which we don't need to store:

```suggestion
  let array = [];
  try {
    array = JSON.parse(extraInfo);
  } catch (e) {
    console.error('Cannot parse parameter extraInfo: ', e);
  }
```

newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
```diff
@@ -72,6 +74,80 @@ const validateParameterName = (i18n: I18nType, newName: string) => {
...
+              <Delete />
+            </IconButton>
+          </Line>
+        ))}
+
+        <Line justifyContent="flex-end" expand>
+          <RaisedButton
+            primary
+            onClick={() => {
+              array.push('New String');
```
4ian, line 139: nit: Let's call this `New Option`.

A string can be anything. An option is a word implying that you should be making a list of option for the user to choose.

newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
```diff
@@ -72,6 +74,80 @@ const validateParameterName = (i18n: I18nType, newName: string) => {
...
+              updateExtraInfo();
+            }}
+            label={<Trans>Add a string to the list</Trans>}
```
4ian, line 142: nit: Add a new option

newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
```diff
@@ -295,6 +371,10 @@ export default class EventsFunctionParametersEditor extends React.Component<
                                   value="string"
                                   primaryText={t`String (text)`}
                                 />
+                                <SelectOption
+                                  value="stringWithSelector"
+                                  primaryText={t`String from a list of strings (text)`}
```
4ian, line 376: `String from a list of options (text)`

newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
```diff
@@ -72,6 +74,80 @@ const validateParameterName = (i18n: I18nType, newName: string) => {
...
+      if (parsedArray !== null) {
+        setArray(parsedArray);
+      }
+    },
+    [extraInfo, array]
+  );
+
+  const updateExtraInfo = () => {
+    setExtraInfo(JSON.stringify(array));
+    forceUpdate();
```
4ian, line 106: You use a forceUpdate because you've been modifying the state without calling the setter (by mutating `array`). With my solution, `array` won't be a state anymore, just something that is computed in the render. So remove the forceUpdate... but in this case, how will this component knows that it must re-render? Thanks to its props :) 
I t will re-render whenever `extraInfo` is changed - which is the responsibility of the parent.

tl;dr: remove `forceUpdate`  here and add it to the parent callback instead.

newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
```diff
@@ -348,6 +428,12 @@ export default class EventsFunctionParametersEditor extends React.Component<
                               />
                             )}
                           </ResponsiveLineStackLayout>
+                          {parameter.getType() === 'stringWithSelector' && (
+                            <StringSelectorEditor
+                              extraInfo={parameter.getExtraInfo()}
+                              setExtraInfo={parameter.setExtraInfo}
```
4ian, line 434:
```suggestion
                              extraInfo={parameter.getExtraInfo()}
                              setExtraInfo={(extraInfo) => { parameter.setExtraInfo(extraInfo); forceUpdate(); }
```

why putting the forceUpdate here? Because the component "StringSelectorEditor" just cares about two things:
- it shows the content of extraInfo
- it tells you when the content of extraInfo changed, giving you the string to store.

so that's what we're doing here, we store the string: `parameter.setExtraInfo(extraInfo)`. However, this thing is stored in the project (i.e: webassembly memory, but it could be anything else: a JS object...). React does not know that `parameter.getExtraInfo()` has now a new value, because it's not a state.
So we call `forceUpdate()` here.

Otherwise, we would not pass the new `extraInfo` to the component (because the render function would not be launched again), which can creates weird behavior and makes the component harder to understand (because, as you did, you need a state inside the component).

4ian (commented):
Almost there! :)

4ian:
Let's go! I fixed Flow typing and also protected against not having an array (which would risk a crash when trying to call functions on the array).
@Bouh FYI, might be interesting for some extensions ;)

4ian:
Thanks @arthuro555 for the work on this :)

newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
```diff
@@ -85,7 +85,8 @@ const StringSelectorEditor = ({
 }: StringSelectorEditorProps) => {
   let array = [];
   try {
-    array = JSON.parse(extraInfo);
+    if (extraInfo !== '') array = JSON.parse(extraInfo);
```
arthuro555, line 88: The if is kind of redundant as the try catch already handles it, doesn't it?

4ian: Yes but the `catch` will display an error in the console, which should be for the developer a sign that something went wrong somewhere before (when using the software normally, there should be no error logged in the console).
As having extraInfo being empty is not an exceptional case (but on the contrary something very common when we're just setting up the array for the first time), it's better to handle it outside of the catch.

PR #2054, opened 2020-10-25
Improve platformer example

Bouh:
The platformer example isn't 100% like on the wiki, so here few thing for make the example consistent with the tutorial.
- Add checkpoint
- Use animation by name

And this is new:
- Add ladder
- Avoid infinite fall, return to the lastest checkpoint

In the example we use an red box for collision, this isn't present in the tutorial.
This player hitbox should be removed and all the behavior go to the player object like the tutorial explain?

FYI, the new parralax effect on the background isn't present in the tutorial. 
It's a pity to have a tutorial rewritten if it doesn't match with the example :/

### DIFF
--- a/newIDE/app/resources/examples/platformer/platformer.json
+++ b/newIDE/app/resources/examples/platformer/platformer.json
@@ -483,7 +483,7 @@
         "gridWidth": 70,
         "snap": true,
         "windowMask": true,
-        "zoomFactor": 0.4346
+        "zoomFactor": 0.1706
       },
       "objectsGroups": [
         {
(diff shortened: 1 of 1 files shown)


4ian:
> FYI, the new parralax effect on the background isn't present in the tutorial.
> It's a pity to have a tutorial rewritten if it doesn't match with the example :/

I don't think we really want this starter to be exactly the output of the tutorial. As much as I would like it, it would need a tutorial of 20+ chapters :)
Starters should be fairly simple games but still good enough to show what GDevelop can do and serve as good basis for games. If we limit to exactly what is in the tutorial, it may not look really good?

> In the example we use an red box for collision, this isn't present in the tutorial.
> This player hitbox should be removed and all the behavior go to the player object like the tutorial explain?

I'm fine keeping the hitbox because it's a pattern that is useful for quite a lot of games?

> Use animation by name

Thanks for that! Whereas things can differ from the tutorials, we should avoid setting animation by number in the examples as the animation by name is much more explicit and understandable.

For these conditions:
image

I think we should keep the the "PlayerHitBox" as the tested object for the collision and Y position - to keep consistency with the rest :) 

Also would be good to fix the jump sound - it's not launched always properly when not using the RShift key (for example on mobile). We should surely do conditions like: `PlayerHitBox is jumping and Trigger Once => Play the sound jump`.

> Avoid infinite fall, return to the lastest checkpoint

Let's maybe add a sound when the object is sent back to the checkpoint after "dying"? :) I can do that later if you want to.

Bouh:
I can give me the day 😉

Le lun. 26 oct. 2020 à 11:42, Florian Rival <notifications@github.com> a
écrit :

Bouh:
Done let me know!

4ian:
I've made more fixes/improvements (and fixed the tween behavior that was not working when only one object was created) in a separate commit :) 

I think we could improve this starter even more by switching more events to behaviors (like to fade coins and fade enemies)! I'm stopping there for now :) 
Thanks for starting this!

Bouh:
Thanks you 👍

PR #2056, opened 2020-10-25
Make zoom direction consistent  with the scene editor.

Bouh:
Inverse zoom direction in preview animation, hitbox, etc...


4ian (approved):
Run autoformatting before merging :)

PR #2057, opened 2020-10-25
Rename Global color to Tint

Bouh:
Because global color is false and misleading.

Tint word is used for said something colorize an object depending to another color (here color from the sprite).

Global color is in the most of other softwares an color reference used on multiple objects.

[Global color on illustrator.](https://helpx.adobe.com/fr/illustrator/how-to/global-color-swatch.html)
[Tint definition](https://www.larousse.fr/dictionnaires/francais/teinte/77010)


4ian (approved):
Sounds much better indeed!

PR #2058, opened 2020-10-25
Increase and make responsive the selectors.

Bouh:
The list selector deserves a bigger aspect on desktop and big screen.

- Use the responsive context for the selectors.
- Increase the default size to 350px height, +100px.
- Make the default value 250px for small screen.

### DIFF
--- a/newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/index.js
+++ b/newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/index.js
@@ -454,11 +454,11 @@ export default class ExpressionField extends React.Component<Props, State> {
                     <ResponsiveWindowMeasurer>
                       {screenSize => (
                         <Paper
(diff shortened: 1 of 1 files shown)


newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/index.js
```diff
@@ -446,17 +451,27 @@ export default class ExpressionField extends React.Component<Props, State> {
...
-                      />
-                    </Paper>
+                    <ResponsiveWindowMeasurer>
+                      {screenSize => (
+                        <Paper
+                          style={() => {
+                            screenSize === 'small'
+                              ? styles.expressionSelectorPopoverContentSmall
+                              : styles.expressionSelectorPopoverContent;
+                          }}
```
4ian, line 461: This probably don't work because you passed a function, and it has no return.
You don't need the function here.

PR #2063, opened 2020-10-26
Apply theme upon load

### LINKED ISSUE #965: Theme is applied late
## Describe the bug
Not sure if it should be considered a bug or a feature request, but even though my GD5 uses the dark theme, the default colors are used when opening GD5, until the welcome screen appears.

## To Reproduce
Choose the dark theme.
Close GD5.
Open GD5.

## Other details
Win 10

nailuj29gaming:
Closes #965


newIDE/app/public/index.html
```diff
@@ -61,5 +61,20 @@
...
       </div>
     </noscript>
+    <script>
+      (function() {
+        // Apply background color for startup
+        var backgroundColors = {
+          Dark: '#252525',
+          Nord: '#2E3440',
+          'Solarized Dark': '#00212B'
+        };
```
nailuj29gaming, line 71: This will need to be added to whenever a new theme is contributed

newIDE/app/public/index.html
```diff
@@ -61,5 +61,20 @@
...
+        console.log(localStorage.getItem("gd-preferences"))
+        var themeName = JSON.parse(localStorage.getItem("gd-preferences"))['themeName'];
+        if (themeName != 'GDevelop default') {
```
arthuro555, line 74: Maybe you should also check if `backgroundColors[themeName]` is defined? I am not sure though, it would show an error that shouldn't crash the ide but alert the developper he forgot to add a background color here.

newIDE/app/public/index.html
```diff
@@ -61,5 +61,24 @@
...
     </noscript>
+    <script>
+      (function() {
+        // Apply background color for startup
+        var backgroundColors = {
+          Dark: '#252525',
+          Nord: '#2E3440',
+          'Solarized Dark': '#00212B'
+        };
+        var themeName = JSON.parse(localStorage.getItem("gd-preferences"))['themeName'];
```
4ian, line 72: Add a try/catch around the parsing - if for some reason the preferences are corrupted, we should not crash.

newIDE/app/public/index.html
```diff
@@ -61,5 +61,24 @@
...
+        if (!(themeName in backgroundColors)) {
+          console.log('Theme does not have a background color');
+          alert('Theme does not have a background color');
```
4ian, line 75: Never use alert, just log something in the console but this is not worth bothering the user with.

newIDE/app/public/index.html
```diff
@@ -61,5 +61,24 @@
...
+          return;
+        }
+        if (themeName != 'GDevelop default') {
+          document.body.style.backgroundColor = backgroundColors[themeName]
```
4ian, line 79: We do this only at the startup. I wonder if we should also do this inside the app when changing theme, as the background color could be visible some times?

4ian: (see for example how we do for the titlebar color for PWAs in `ProjectTilebar.js` - we could do the same with the document.body.style.backgroundColor and the Electron background color: https://www.electronjs.org/docs/api/browser-window#winsetbackgroundcolorbackgroundcolor)

4ian:
Thanks for the PR! I added a few comments :)

newIDE/app/public/index.html
4ian: You've not addressed this comment before asking for a review again :)

newIDE/app/public/index.html
4ian: Include a comment in the place where themes are declared so that this information is not lost.

newIDE/app/public/index.html
```diff
@@ -61,5 +61,29 @@
...
+        var backgroundColors = {
+          Dark: '#252525',
+          Nord: '#2E3440',
+          'Solarized Dark': '#00212B'
+        };
+        var themeName;
+        try {
+          themeName = JSON.parse(localStorage.getItem("gd-preferences"))['themeName'];
+        } catch (err) {
+          console.error(err);
```
4ian, line 76: Include a detail error message in addition to the raw error

newIDE/app/public/index.html
nailuj29gaming: Sorry about that! I was having a hard time figuring out where to put the code to do that.

4ian: You've again requested a review while this is not addressed: please rather ask questions rather than requesting review :)

newIDE/app/public/index.html
4ian: This is not addressed

newIDE/app/public/index.html
4ian: (if you can't find how to do it it's no problem, I'll do it or explain more but prefer to ask questions rather than just asking for review again, to avoid losing time for both of us)

newIDE/app/public/index.html
nailuj29gaming: I think this was the only one I addressed in this commit. Can you be more specific by what you meant be a detailed error message?

newIDE/app/public/index.html
nailuj29gaming: I think I know how to do it, I just don't know where it should go in the codebase.

nailuj29gaming: @4ian I actually think this is not too necessary because the main `div` takes up the whole screen, and its color is changed when the theme is changed

newIDE/app/public/index.html
4ian: Logging just the exception is not sufficient to understand the issue for someone looking at the console. Do something like:

`console.error("Unable to parse the preferences to set up the background color:", err)`

so that it's clear what went wrong.

newIDE/app/public/index.html
4ian: Can you clarify which div you're talking about? It would still be great to change the electron background color as it would allow to not have a grey color shown when resizing the window (which is shown the time for the rest of the content to be resized).

nailuj29gaming: The div that React renders to.

nailuj29gaming: @4ian should I put the code to change the background color in the `index.js` or is there a better file for that?

newIDE/app/public/index.html
nailuj29gaming: Ok, that makes more sense.

newIDE/app/public/index.html
nailuj29gaming: @4ian I was just messing around with resizing the window, and it appears to not be a problem when resizing.

newIDE/app/public/index.html
nailuj29gaming: @4ian Done.

newIDE/app/public/index.html
nailuj29gaming: @4ian I think I've just about given up on this. You can still add it if you want.

PR #2069, opened 2020-10-28
Changed from distance joint to rope joint

tristanbob:
I made this change so that the example more closely acts like a rope (instead of a solid metal pole).

I only changed one line, so you can reject this if you only want that:

OLD:  "value": "Physics2::AddDistanceJoint"
NEW:  "value": "Physics2::AddRopeJoint"

The rest of the changes I think are from Gdevelop using different defaults.


zatsme:
I think the problem here if anything is the name of the example, maybe rope wasn't the best name to use.

When we completed physics2 integration into gdevelop I created this, downhill bike and ragdoll to show what could be done to make some fun games.  

There were 2 reasons distance joint was used in this example.

1.  I'd spent ages tuning the physics for downhill bike and ragdoll and if you've played with the examples you'd realise how sensitive the physics are to changes and therefore gameplay.  So didn't have the time to spend making a proper rope example with segmented rope parts which would make it look like a rope graphically.

2.  I found that the non elastic, rigid "rope" made this example quite fun 😁

So my suggestion would be to leave this as is, but maybe rename the example to something more descriptive and to make a new rope example with a proper segmented rope similar to [This](https://stackoverflow.com/questions/26054104/how-to-construct-and-draw-a-nice-rope-with-box2d/26063322) 

Tuning the rope to act and look right will be a challenge 👍

tristanbob:
@zatsme thanks for that feedback.  Great job on building these examples!  

I agree that your game was fun with the distance joint, and my only issue was with the word "rope".   No need to change.

I'll close this pull-request.

PR #2071, opened 2020-10-29
Inline parameter editor popover improvements

4ian:
This brings multiple improvements to the inline popover that is shown when editing a parameter in the events sheet:
- it can always be closed using Escape now.
- when closed, it focuses back the parameter, useful to edit the events sheet using the keyboard
- fix issues on touchscreen where changes were not applied when closing the popover

This also take the opportunity to update to latest material-ui core version, which allow to remove a hack in SemiControlledAutoComplete (though we kind of add another). 
Needs to do a global check of the app to verify nothing else was changed/broken by the update.

### DIFF
--- a/newIDE/app/package-lock.json
+++ b/newIDE/app/package-lock.json
@@ -2444,22 +2444,22 @@
       }
     },
     "@material-ui/core": {
-      "version": "4.9.10",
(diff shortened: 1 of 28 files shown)


Bouh:
About fast checking, the nightly build service can generate an build on nightly-editor.gdevelop-app.com ?

PR #2073, opened 2020-10-31
Remove Create function from object list only

Bouh:
Linked to issue 2072.

In the new instruction editor, Create an object is only available after a choice of an object.
This action is generic and should be in both lists, but before all available by the search bar without choose object or non-object list. Basically the research bar shouldn't be tab dependent.

- [x] Test in Webapp & Electron
- [x] Prettier


4ian:
Careful, this also removed Create from the objects list, while we discussed it should be in both!

4ian:
Fixed in 449c96aabaa3ffb60270ecc4e01117135ccf644e ;)

PR #2088, opened 2020-11-11
Add support for the Asset Store on the desktop app

4ian:
This was only available on the web-app, the assets store is now also available on the desktop app. This will be getting more and more useful as we grow it in the future with community contributions :) Even currently, it's a good way to quickly start a game or do prototypes.

Also:
* Automatically download resources when a project made on the web-app is opened on the desktop-app.
  * Note that the reverse operation, opening a project made on the desktop-app in the web-app, is not possible for now, but could be something that could be added later. Not a priority at all though.
  * Internally, resources can now have an "origin". This is useful to annotate resources to know where they are coming from, so that we don't download them again if a user is adding multiple times the same object from the assets store. This could also be useful later if we add support for resources from cloud providers.
* Rework the loader when opening a project or launching a preview (properly block clicks behind the loader)
* Shorten names for resources added from the resource store, in the web-app.

### DIFF
--- a/Core/GDCore/Project/ResourcesManager.cpp
+++ b/Core/GDCore/Project/ResourcesManager.cpp
@@ -5,8 +5,10 @@
  */
 
 #include "GDCore/Project/ResourcesManager.h"
+
 #include <iostream>
 #include <map>
+
 #include "GDCore/CommonTools.h"
 #include "GDCore/Project/Project.h"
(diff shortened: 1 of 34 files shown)


blurymind:
Would be awesome if we can create an actual market around assets where people can sell them and a % goes to funding gdevelop - similar to Unity/unreal/clickteam stores

4ian:
That could totally be a first step toward this :)

PR #2090, opened 2020-11-13
Fixed light texture bug

### LINKED ISSUE #2048: Texture in Light Object bug
## Describe the bug
In the beta 101, an texture cannot be render in the light object even with webgl 2 and 1.
The last [PR 1979](https://github.com/4ian/GDevelop/pull/1979) on Light have introduce this bug.

Here a white square as texture light:
On beta 100, result expected:
image

On beta 101, bugged:
image

[Test_Light_project.zip](https://github.com/4ian/GDevelop/files/5403183/Test_Light_project.zip)

## Other details
b101 only

@HarsimranVirk @4ian

HarsimranVirk:
Resolves #2048 
I'm sorry for not being active here, anyway this bug was a minor regression 😬.

### DIFF
--- a/Extensions/Lighting/lightruntimeobject-pixi-renderer.js
+++ b/Extensions/Lighting/lightruntimeobject-pixi-renderer.js
@@ -232,7 +232,6 @@ gdjs.LightRuntimeObjectPixiRenderer.prototype.updateColor = function () {
 };
 
 gdjs.LightRuntimeObjectPixiRenderer.prototype.updateTexture = function () {
-  if (!this._light) return;
   var texture = this._object.getTexture();
   this._texture =
     texture !== ''


Extensions/Lighting/lightruntimeobject-pixi-renderer.js
```diff
@@ -232,7 +232,6 @@ gdjs.LightRuntimeObjectPixiRenderer.prototype.updateColor = function () {
 };
 
 gdjs.LightRuntimeObjectPixiRenderer.prototype.updateTexture = function () {
-  if (!this._light) return;
```
HarsimranVirk, line 235: This makes no sense, the function will always return early without setting the texture 😬, a minor regression.

4ian:
Ah right, was indeed a silly mistake! Good catch :)

PR #2098, opened 2020-11-16
Set cursor to pointer in SceneEditor Menu

AdvaitDhingra:
Hello. My name is Advait and I am a classmate of arthuro555. He recommended this repo to learn and improve my ReactJs and coding skills. I have only been coding for a year, so my contributions will most likely be useless. 

I changed the cursor of the toolbar Iconbuttons to cursor, as to indicate that it is a button. Even though it looks like a button, a pointer cursor really helps my brain register it as a button. 

I hope you consider my PR.

Regards,
Advait

### DIFF
--- a/newIDE/app/src/KeyboardShortcuts/DefaultShortcuts.js
+++ b/newIDE/app/src/KeyboardShortcuts/DefaultShortcuts.js
@@ -3,6 +3,7 @@ import { type CommandName } from '../CommandPalette/CommandsList';
 
 export type ShortcutMap = { [CommandName]: string };
 
+
 const defaultShortcuts: ShortcutMap = {
   QUIT_APP: 'CmdOrCtrl+Shift+KeyQ',
   OPEN_PROJECT_MANAGER: 'CmdOrCtrl+Alt+KeyE',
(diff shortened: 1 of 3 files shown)


4ian:
Hi @AdvaitDhingra, thanks for taking the time to learn this and open the PR :)
I'll unfortunately won't be able to merge this in the codebase because it's a conscious decision not be using a `pointer` cursor (i.e: the "hand" cursor) for buttons. Instead, we follow the UX of most desktop app, which is to use the normal cursor but show a subtle indicator that you're *hovering* a button.
See for example in Chrome when I put the cursor on the reload button: 
image

Cursors stays the same, but there is a "hover" circle shown. Same in GDevelop: 
image

We might be missing this hover on a few elements, in which case we should probably add it (for example, for the tabs). Apart from that, all buttons should have an hover effect.

Using a "hand" cursor was mainly adopted for web pages by the browser. As GDevelop is not a web page but a "real" application, we prefer to use desktop conventions.

Hope it's understandable, thanks anyway! :)

newIDE/app/src/KeyboardShortcuts/DefaultShortcuts.js
```diff
@@ -3,6 +3,7 @@ import { type CommandName } from '../CommandPalette/CommandsList';
 
 export type ShortcutMap = { [CommandName]: string };
 
+
```
4ian, line 6: When you're creating a future PR, make sure to remove the useless lines like this from your PR. For this, proofread your PR before sending it for review and add commits that undo these changes. You can learn more by searching for git/github desktop.

AdvaitDhingra: Ahh I see. Thank you.

PR #2101, opened 2020-11-17
Added Jsdoc to newIDE files

AdvaitDhingra:
Hello. It's me again. 

Arthur recommended I write the jsdocs for some of the functions. I have done this to the ones I could find. Is this how it is done? I'm very sorry that I am so bad.

Regards,
Advait


4ian:
We're not using JSDoc annotations in the newIDE, because it is typed using Flow (similar to TypeScript). I recommend you read this page: https://github.com/4ian/GDevelop/blob/master/newIDE/docs/Supported-JavaScript-features-and-coding-style.md#for-the-editor-newideapp 

Instead, I recommend you pass more time getting used to the codebase before sending PRs. You need to understand what are the tools used, why, when they make sense and what is impactful to do.

Thanks anyway and good luck! :)

PR #2102, opened 2020-11-18
Rewrite two methods from GD4 into GD5 make color transparent and copy texture on current texture of an object

Bouh:
## `gdjs.SpriteRuntimeObject.makeColorTransparent`
First parameter is the `Color`, second is the `Threshold` normalized from 0 to 1

> **How work the shader:** 
> The Threshold use the texture converted in black-and-white and remove all pixels lighter than the threshold.
> 
This method works, and the action is available, you can test it!

image

## `gdjs.SpriteRuntimeObject.copyImageOnImageOfCurrentSprite`
First parameter is the `Texture` to add on the object.
Second and third are the positions `X` and `Y` of `Texture`.
Fourth is actually not used in the shader.
Fith will be the rotation of `Texture`.

This method need still a small work, the positions for the added texture use the wrong [coordinate system](https://github.com/pixijs/pixi.js/wiki/v5-Creating-filters), (i try to understand but i'm not very familiar with this...)
I need to see how manage the fourth parameter in the shader.

image

## Checklist

`copyImageOnImageOfCurrentSprite`
- [ ] Fix coordinate system for positions of texture
- [ ] Add support of transparency on the fourth parameter.

`copyImageOnImageOfCurrentSprite` and `makeColorTransparent`
- [ ] Clean shaders
- [ ] Check action name/descriptions/parameters in IDE
- [ ] Test multiple instances with the action
- [ ] Test multiple object with the action


arthuro555:
I think you shouldn't use a shader for `gdjs.SpriteRuntimeObject.copyImageOnImageOfCurrentSprite`. This would copy the the image on the sprite, well, every frame, and cost much performance. 

Instead, simply put the original sprite and a sprite made of the other texture in a `Container`, render it to a `RenderTexture`, and set the texture of the sprite to the new `RenderTexture`. This would 

1. Fix your coordinate problems
2. Make the rendering happen once and improve performance

This would also make the code more readable to people that don't know WebGL/the shader language.

GDJS/Runtime/pixi-renderers/spriteruntimeobject-pixi-renderer.js
```diff
@@ -131,3 +131,16 @@ gdjs.SpriteRuntimeObjectPixiRenderer.getAnimationFrameWidth = function(pixiTextu
 gdjs.SpriteRuntimeObjectPixiRenderer.getAnimationFrameHeight = function(pixiTexture) {
     return pixiTexture.height;
 };
+
+gdjs.SpriteRuntimeObjectPixiRenderer.setSprite = function(runtimeScene, imageName) {
```
arthuro555, line 135: The function name is a bit misleading, you would think seeing a `setSprite` method that it accepts a pixi sprite, or a GDevelop sprite. I think something like `changeTexture` or `setTexture` would be more appropriate.

4ian:
We'll also introduce one day the effects for objects (i.e: shader/filter for objects, not only for layers). 
This can surely be achieved with a shader - I've not checked the code, but I agree with @arthuro555 that it's worth considering using an intermediate render texture and update the texture in memory? Would be more akin to what was done in GD4. 

This being said, I've not investigated on the performance on the shader.

GDJS/Runtime/pixi-renderers/spriteruntimeobject-pixi-renderer.js
```diff
@@ -131,3 +131,16 @@ gdjs.SpriteRuntimeObjectPixiRenderer.getAnimationFrameWidth = function(pixiTextu
...
+    this._spriteDirty = true;
+    this._sprite = new PIXI.Sprite(runtimeScene.getGame().getImageManager().getPIXITexture(imageName));
+};
+
+gdjs.SpriteRuntimeObjectPixiRenderer.prototype.setShader = function(pixiShader) {
+     this._sprite.shader = pixiShader;
+};
+
+gdjs.SpriteRuntimeObjectPixiRenderer.prototype.setFilter = function(pixiFilter) {
```
arthuro555, line 144: This can be annoying in the future when adding multiple effects on objects. It isn't easy to remove an API as it might be used, and this api would be annoying/redundant/breaking the effects if we have multiple ones.

4ian:
> This would copy the the image on the sprite, well, every frame, and cost much performance.

As this is in a shader, this is accelerated by the GPU though, so might not cost much more performance that actually creating a new texture in memory? 
The advantage of the texture is that, while it may be a bit more costly to generate, it's then basically free (apart for the memory used).

GDJS/Runtime/spriteruntimeobject.js
```diff
@@ -942,125 +1040,345 @@ gdjs.SpriteRuntimeObject.prototype.isFlippedY = function() {
...
+    this.uniforms.otherMatrix = filterManager
+      .calculateSpriteMatrix(sprite_scMatrix, sprite_sc)
+      .prepend(sprite_sc.texture.transform.mapCoord);
+
+    // draw the filter...
+    filterManager.applyFilter(this, input, output);
+  };
+
+  this._renderer.setFilter(filter);
+};
```
arthuro555, line 1306: Here is a mockup of what it could look like with a render texture: 

```suggestion
/**
 * Load an image from the image manager and copy on image over the texture of the object.
 *
 * @param {gdjs.RuntimeScene} runtimeScene The scene where living the object
 * @param {string} imageName Name of the texture to copy over the current texture from the object
 * @param {number} x X position of the texture
 * @param {number} y Y position of the texture
 * @param {number} angle Angle of the texture
 */
gdjs.SpriteRuntimeObject.prototype.copyImageOnImageOfCurrentSprite = function (
  runtimeScene,
  imageName,
  x,
  y,
  angle
) {
  var container = new PIXI.Container();
  var currentTexture = this._renderer.getRendererObject().texture;
  var currentTextureSprite = new PIXI.Sprite(currentTexture);
  container.addChild(currentTextureSprite);
  var texture_sc = runtimeScene
    .getGame()
    .getImageManager()
    .getPIXITexture(imageName);
  var sprite_sc = new PIXI.Sprite(texture_sc);
  container.addChild(sprite_sc);

  sprite_sc.x = x;
  sprite_sc.y = y;
  sprite_sc.angle = angle;
  
  var renderTexture = PIXI.RenderTexture.create(container.height, container.width);
  runtimeScene
    .getGame()
    .getRenderer()
    .getPIXIRenderer()
    .render(container, renderTexture);
  this._sprite.texture = renderTexture;
};
```

GDJS/Runtime/spriteruntimeobject.js
```diff
@@ -942,125 +1040,345 @@ gdjs.SpriteRuntimeObject.prototype.isFlippedY = function() {
...
-        obj.getDrawableY()+obj.getCenterY(), 0, scene);
+gdjs.SpriteRuntimeObject.prototype.turnTowardObject = function (obj, scene) {
+  if (obj === null) return;
+
+  this.rotateTowardPosition(
+    obj.getDrawableX() + obj.getCenterX(),
+    obj.getDrawableY() + obj.getCenterY(),
+    0,
+    scene
+  );
```
arthuro555, line 1383: All your changes here are using PIXI specific code, so they should be implemented in the renderer.

GDJS/Runtime/spriteruntimeobject.js
Bouh: I totally agree, it's a simpler way ^^ i works, i would like keep the support of useTransparency parameter, maybe someting can be done with mask from Pixi?

GDJS/Runtime/spriteruntimeobject.js
Bouh: Yes, bien vu!

Bouh:
Your solution @arthuro555 is easier!
Still a little buggy, when the texture is added to the object, the sprite of the object gets bigger, very weird ^^
On left the original, on right impact added on the sprite.
image

@4ian for `copyImageOnImageOfCurrentSprite` the parameter `useTransparency` we should keep it, even if this one isn't supported in the renderer?
For avoid bug ordering in parameter when an user import an example from GD4, like the example `Terrain destruction.gdg`
Here converted into json:

[Terrain destruction.zip](https://github.com/4ian/GDevelop/files/5573985/Terrain.destruction.zip)

Yes, the goal is to make this example work again :)

4ian:
> @4ian for copyImageOnImageOfCurrentSprite the parameter useTransparency we should keep it, even if this one isn't supported in the renderer?

I would say that here it's fine to remove it because it has been so long that compatibility is not necessarily a goal anymore - but feel free to do what you think is best :) 

> Yes, the goal is to make this example work again :)

There was in this example the usage of a pixel perfect collision - which was a very expensive operation - to calculate collision between bullets and the turret against the floor objects. We won't add it back, but what could be useful would be a condition to check the color of the pixel of an object at a specified position. Like this, you could check if first a bullet is in collision with a floor (using the classic collision condition using hitboxes) and then check, for a point of the bullet (or a point on the base of the turret), if this position corresponds to a pixel that is transparent. If yes => no collision, otherwise there is collision so you make the bullet explode/apply gravity on the turret.

Just a thought :)

GDJS/Runtime/pixi-renderers/spriteruntimeobject-pixi-renderer.js
4ian: We'll indeed clean this from this PR to be sure we don't add this API to set filters yet - it will be part of the work to enable filters on objects :)

arthuro555:
> Still a little buggy, when the texture is added to the object, the sprite of the object gets bigger,

This is intended, as you see I use the container height/width when creating the render texture. If you want to keep the texture to the original height/width even if it means cutting the added one, you can use the original texture h/w I stead of the container one.
Personally, I would add this as a "crop texture" boolean parameter to let the user choose if the texture should be allowed to have a size change

arthuro555:
Make color transparent bis essentially just a shader. Maybe it should be added as effect when support for object effects is implemented instead of making it a separate thing?

4ian:
> If you want to keep the texture to the original height/width even if it means cutting the added one, you can use the original texture h/w I stead of the container one.
> Personally, I would add this as a "crop texture" boolean parameter to let the user choose if the texture should be allowed to have a size change

Yeah maybe we could give the option to crop or not.

> Make color transparent bis essentially just a shader. Maybe it should be added as effect when support for object effects is implemented instead of making it a separate thing?

Depends if it's just about a visual effect (in which case, an effect later would be good to have in all cases) or really about changing the texture, in which case it's more an action to manipulate the texture at runtime (useful in case you want to things like "terrain destruction" where you can test for collisions against pixels that are transparent).

arthuro555:
> Depends if it's just about a visual effect

To me it looks like a visual effect, and it is also treated just as a normal filter in bouhs implementation as far as I see. Though, maybe we could make an action that replaces the texture of a sprite with a texture that got the shader applied to it? That way we could have a way to put an effect on a texture that you know won't change and doesn't need to be updated, and take less performance (as you wouldn't have to run the shader code on each rerender).

PR #2103, opened 2020-11-19
Notification Extention

Add00:
### Description
This extension uses the Notification Web API to allow for the creation and manipulation of notifications. 

### Checklist
- [x] Extension has a proper name and description.
- [x] Extension has tags (for example: "platform, brick, breakable").
- [x] All behaviors have a description.
- [x] All functions (actions, conditions, expressions) have descriptions.
- [x] I confirm that this extension can be intergrated to this GitHub repository, distributed and MIT licensed.

### DIFF
--- a/Extensions/Notification/JsExtension.js
+++ b/Extensions/Notification/JsExtension.js
@@ -1,0 +1,182 @@
+// @flow
+/**
+ * This is a declaration of an extension for GDevelop 5.
+ *
+ * ℹ️ Changes in this file are watched and automatically imported if the editor
+ * is running. You can also manually run `node import-GDJS-Runtime.js` (in newIDE/app/scripts).
+ *
+ * The file must be named "JsExtension.js", otherwise GDevelop won't load it.
+ * ⚠️ If you make a change and the extension is not loaded, open the developer console
+ * and search for any errors.
+ *
+ * More information on https://github.com/4ian/GDevelop/blob/master/newIDE/README-extensions.md
+ */
+
+/*::
+// Import types to allow Flow to do static type checking on this file.
+// Extensions declaration are typed using Flow (like the editor), but the files
+// for the game engine are checked with TypeScript annotations.
+import { type ObjectsRenderingService, type ObjectsEditorService } from '../JsExtensionTypes.flow.js'
+*/
+
+module.exports = {
+  createExtension: function (
+    _ /*: (string) => string */,
+    gd /*: libGDevelop */
+  ) {
+    const extension = new gd.PlatformExtension();
+    extension.setExtensionInformation(
+      'Notification',
+      _('Notification'),
+      _('Use and maniulate notifications'),
+      'Add00',
+      'MIT'
+    );
+
+    extension
+      .addCondition(
+        'RequestPermission',
+        _('Request permission'),
+        _('Request that the user enable notification (browser applications only).'),
+        _('Request permission for notifications'),
+        _('Notification'),
+        //'JsPlatform/Extensions/',
+        //'JsPlatform/Extensions/'
+      )
+      .getCodeExtraInformation()
+      .setIncludeFile('Extensions/Notification/notificationtools.js')
+      .setFunctionName('gdjs.evtTools.notification.requestPermission');
+
+    extension
+      .addCondition(
+        'IsSupported',
+        _('Notifications are supported'),
+        _('Checks if the browser supports notifications (browser applications only).'),
+        _('Notifications are supported'),
+        _('Notification'),
+        //'JsPlatform/Extensions/',
+        //'JsPlatform/Extensions/'
+      )
+      .getCodeExtraInformation()
+      .setIncludeFile('Extensions/Notification/notificationtools.js')
+      .setFunctionName('gdjs.evtTools.notification.isSupported');
+
+    extension
+      .addCondition(
+        'IsClicked',
+        _('Notification has been clicked'),
+        _('Check if the notification has been clicked'),
+        _('Check if notification with an id of _PARAM0_ has been clicked'),
+        _('Notification'),
+        //'JsPlatform/Extensions/',
+        //'JsPlatform/Extensions/'
+      )
+      .addParameter('expression', _('Id'), '', false)
+      .getCodeExtraInformation()
+      .setIncludeFile('Extensions/Notification/notificationtools.js')
+      .setFunctionName('gdjs.evtTools.notification.onEvent');
+
+    extension
+      .addCondition(
+        'IsShown',
+        _('Notification has been shown'),
+        _('Check if the notification has been shown'),
+        _('Check if notification with an id of _PARAM0_ has been shown'),
+        _('Notification'),
+        //'JsPlatform/Extensions/',
+        //'JsPlatform/Extensions/'
+      )
+      .addParameter('expression', _('Id'), '', false)
+      .getCodeExtraInformation()
+      .setIncludeFile('Extensions/Notification/notificationtools.js')
+      .setFunctionName('gdjs.evtTools.notification.onEvent');
+
+    extension
+      .addCondition(
+        'IsClosed',
+        _('Notification has been closed'),
+        _('Check if the notification has been closed'),
+        _('Check if notification with an id of _PARAM0_ has been closed'),
+        _('Notification'),
+        //'JsPlatform/Extensions/',
+        //'JsPlatform/Extensions/'
+      )
+      .addParameter('expression', _('Id'), '', false)
+      .getCodeExtraInformation()
+      .setIncludeFile('Extensions/Notification/notificationtools.js')
+      .setFunctionName('gdjs.evtTools.notification.onEvent');
+
+    extension
+      .addCondition(
+        'IsError',
+        _('Notification has an error'),
+        _('Check if an error happend with the notification'),
+        _('Check if notification with an id of _PARAM0_ has an error'),
+        _('Notification'),
+        //'JsPlatform/Extensions/',
+        //'JsPlatform/Extensions/'
+      )
+      .addParameter('expression', _('Id'), '', false)
+      .getCodeExtraInformation()
+      .setIncludeFile('Extensions/Notification/notificationtools.js')
+      .setFunctionName('gdjs.evtTools.notification.onEvent');
+
+    extension
+      .addAction(
+        'CreateNotification',
+        _('Create a new notification'),
+        _('Create a new notification (always trigger once)'),
+        _('Create a new notification with an id of _PARAM0_, a title of _PARAM1_ and a message of _PARAM2_'),
+        _('Notification'),
+        //'JsPlatform/Extensions/',
+        //'JsPlatform/Extensions/'
+      )
+      .addParameter('expression', _('Id'), '', false)
+      .addParameter('string', _('Title'), '', false)
+      .addParameter('string', _('Message'), '', false)
+      .addParameter('string', _('(Optional) file location of an image to display'), '', true)
+      .addParameter('string', _('(Optional) change language of text, must be an ISO 2 Letter Language Code'), '', true)
+      .addParameter('string', _('(Optional) tag'), '', true)
+      .setDefaultValue('false')
+      .addParameter('yesorno', _('(Optional) notification is silent'), '', true)
+      .addParameter('string', _('(Optional) vibrate mobile device, must be a comma separated list'), '', true)
+      .getCodeExtraInformation()
+      .setIncludeFile('Extensions/Notification/notificationtools.js')
+      .setFunctionName('gdjs.evtTools.notification.createNotification');
+
+    extension
+      .addAction(
+        'RemoveNotification',
+        _('Remove a notification'),
+        _('Remove an exsiting notification'),
+        _('Remove the notification with an id of _PARAM0_'),
+        _('Notification'),
+        //'JsPlatform/Extensions/',
+        //'JsPlatform/Extensions/'
+      )
+      .addParameter('expression', _('Id'), '', false)
+      .getCodeExtraInformation()
+      .setIncludeFile('Extensions/Notification/notificationtools.js')
+      .setFunctionName('gdjs.evtTools.notification.removeNotification');
+
+    extension
+      .addStrExpression(
+        'PermissionStatus',
+        _('Returns a string containing the current permissions status'),
+        _('The return value will be one of three things: default, granted, or denied'),
+        _('Notification'),
+        //'JsPlatform/Extensions/'
+        //'JsPlatform/Extensions/'
+      )
+      .getCodeExtraInformation()
+      .setFunctionName('gdjs.evtTools.notification.permissionStatus');
+
+    return extension;
+  },
+  runExtensionSanityTests: function (
+    gd /*: libGDevelop */,
+    extension /*: gdPlatformExtension*/
+  ) {
+    return [];
+  },
+};
--- a/Extensions/Notification/JsExtension.js
+++ b/Extensions/Notification/JsExtension.js
@@ -28,24 +28,24 @@ module.exports = {
     extension.setExtensionInformation(
       'Notification',
       _('Notification'),
-      _('Use and maniulate notifications'),
+      _('Use and manipulate notifications'),
       'Add00',
       'MIT'
     );
 
     extension
       .addCondition(
-        'RequestPermission',
-        _('Request permission'),
-        _('Request that the user enable notification (browser applications only).'),
-        _('Request permission for notifications'),
+        'IsPermissionGranted',
+        _('Check if permission has been granted'),
+        _('Check if the user has given permission for notifications'),
+        _('Check if the user has granted notification permissions'),
         _('Notification'),
         //'JsPlatform/Extensions/',
         //'JsPlatform/Extensions/'
       )
       .getCodeExtraInformation()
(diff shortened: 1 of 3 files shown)


Extensions/Notification/notificationtools.js
```diff
@@ -0,0 +1,89 @@
...
+    vibrate: vibe.split(",")
+  }
+
+  gdjs.evtTools.notification.notif[id] = new Notification(title, config);
+
+  gdjs.evtTools.notification.event[id] = gdjs.evtTools.notification.notif[id];
+  gdjs.evtTools.notification.event[id].onclick = () => {gdjs.evtTools.notification.event[id] = true}
+  gdjs.evtTools.notification.event[id].onshow = () => {gdjs.evtTools.notification.event[id] = true}
+  gdjs.evtTools.notification.event[id].onerror = () => {gdjs.evtTools.notification.event[id] = true}
+  gdjs.evtTools.notification.event[id].onclose = () => {gdjs.evtTools.notification.event[id] = true}
```
arthuro555, line 65: Try to store the event type, all events are not the same
```suggestion
  gdjs.evtTools.notification.event[id].onclose = () => {gdjs.evtTools.notification.event[id] = "close";}
```

Extensions/Notification/notificationtools.js
```diff
@@ -0,0 +1,89 @@
...
+
+  console.log(gdjs.evtTools.notification);
+}
+
+gdjs.evtTools.notification.onEvent = function(id) {
+  const returnValue = gdjs.evtTools.notification.event[id];
+  gdjs.evtTools.notification.event[id] = false;
+  gdjs.evtTools.notification.notif[id] = null;
```
arthuro555, line 73: Why delete the notification?

Extensions/Notification/notificationtools.js
```diff
@@ -0,0 +1,89 @@
...
+  gdjs.evtTools.notification.event[id] = gdjs.evtTools.notification.notif[id];
+  gdjs.evtTools.notification.event[id].onclick = () => {gdjs.evtTools.notification.event[id] = true}
+  gdjs.evtTools.notification.event[id].onshow = () => {gdjs.evtTools.notification.event[id] = true}
+  gdjs.evtTools.notification.event[id].onerror = () => {gdjs.evtTools.notification.event[id] = true}
+  gdjs.evtTools.notification.event[id].onclose = () => {gdjs.evtTools.notification.event[id] = true}
+
+  console.log(gdjs.evtTools.notification);
+}
+
+gdjs.evtTools.notification.onEvent = function(id) {
```
arthuro555, line 70: Make one event handler by event type so that users can react on a specific one.

Extensions/Notification/notificationtools.js
```diff
@@ -0,0 +1,89 @@
...
+  gdjs.evtTools.notification.event[id] = false;
+  gdjs.evtTools.notification.notif[id] = null;
+
+  //console.log(gdjs.evtTools.notification);
+
+  return returnValue;
+}
+
+gdjs.evtTools.notification.removeNotification = function(id) {
+  gdjs.evtTools.notification.notif[id].close();
```
arthuro555, line 81: Also delete the reference so it can be garbage collected:
```suggestion
  gdjs.evtTools.notification.notif[id].close();
  delete gdjs.evtTools.notification.notif[id];
```

Add00:
Thanks for your feedback arthuro, I'll be sure to change the items you mention.

4ian:
Thanks for the PR @Add00 and for the review @arthuro555. What about the compatibility on different platforms? Is this supported on web browser only, or can it work too on Android/iOS with Cordova? :)

Add00:
So far I have only tested it in on Windows and in the browser with Firefox and Edge (both on desktop), but theoretically it would work on mobile devices as well.

Bouh:
It would be good to do a real test on the mobile browser, and in an APK, both are different.

Extensions/Notification/JsExtension.js
```diff
@@ -0,0 +1,182 @@
...
+        _('Check if the notification has been closed'),
+        _('Check if notification with an id of _PARAM0_ has been closed'),
+        _('Notification'),
+        //'JsPlatform/Extensions/',
+        //'JsPlatform/Extensions/'
+      )
+      .addParameter('expression', _('Id'), '', false)
+      .getCodeExtraInformation()
+      .setIncludeFile('Extensions/Notification/notificationtools.js')
+      .setFunctionName('gdjs.evtTools.notification.onEvent');
```
Bouh, line 107: IsClosed or onEvent ?

Extensions/Notification/JsExtension.js
```diff
@@ -0,0 +1,182 @@
...
+      .addParameter('string', _('(Optional) vibrate mobile device, must be a comma separated list'), '', true)
+      .getCodeExtraInformation()
+      .setIncludeFile('Extensions/Notification/notificationtools.js')
+      .setFunctionName('gdjs.evtTools.notification.createNotification');
+
+    extension
+      .addAction(
+        'RemoveNotification',
+        _('Remove a notification'),
+        _('Remove an exsiting notification'),
```
Bouh, line 151: exsiting => existing

Extensions/Notification/JsExtension.js
```diff
@@ -0,0 +1,182 @@
...
+        _('Check if an error happend with the notification'),
+        _('Check if notification with an id of _PARAM0_ has an error'),
+        _('Notification'),
+        //'JsPlatform/Extensions/',
+        //'JsPlatform/Extensions/'
+      )
+      .addParameter('expression', _('Id'), '', false)
+      .getCodeExtraInformation()
+      .setIncludeFile('Extensions/Notification/notificationtools.js')
+      .setFunctionName('gdjs.evtTools.notification.onEvent');
```
Bouh, line 122: IsError or onEvent ?

Extensions/Notification/JsExtension.js
```diff
@@ -0,0 +1,182 @@
...
+        _('Check if the notification has been shown'),
+        _('Check if notification with an id of _PARAM0_ has been shown'),
+        _('Notification'),
+        //'JsPlatform/Extensions/',
+        //'JsPlatform/Extensions/'
+      )
+      .addParameter('expression', _('Id'), '', false)
+      .getCodeExtraInformation()
+      .setIncludeFile('Extensions/Notification/notificationtools.js')
+      .setFunctionName('gdjs.evtTools.notification.onEvent');
```
Bouh, line 92: IsShown or onEvent ?

Extensions/Notification/JsExtension.js
```diff
@@ -0,0 +1,182 @@
...
+        _('Check if the notification has been clicked'),
+        _('Check if notification with an id of _PARAM0_ has been clicked'),
+        _('Notification'),
+        //'JsPlatform/Extensions/',
+        //'JsPlatform/Extensions/'
+      )
+      .addParameter('expression', _('Id'), '', false)
+      .getCodeExtraInformation()
+      .setIncludeFile('Extensions/Notification/notificationtools.js')
+      .setFunctionName('gdjs.evtTools.notification.onEvent');
```
Bouh, line 77: IsClicked or onEvent ?

Add00:
Sorry it took me so long to add in all of your feedback a combination of real life stuff and me not knowing how to program stopped me from finishing as quickly as I would have liked.

Extensions/Notification/notificationtools.js
```diff
@@ -1,89 +1,84 @@
...
-    return true;
-  }
-
-  //If they haven't ask for it
-  else if (Notification.permission !== 'denied') {
     Notification.requestPermission().then(function(status) {
-      if (status === "granted") {
-          return true;
-      }
+      gdjs.evtTools.notification.status = status;
```
arthuro555, line 17: No need for that variable, you can use `Notification.permission` 😉

Extensions/Notification/notificationtools.js
```diff
@@ -1,89 +1,84 @@
...
+}
 
-  //The user does not grant permission or does not respond to the prompt
-  else {
-    return false;
+gdjs.evtTools.notification.permissionGranted = function() {
+  if (Notification.permission === 'granted') {
+    return true
+  } else {
+    return false
```
arthuro555, line 25: You can simplify this
```suggestion
  return Notification.permission === 'granted';
```

Extensions/Notification/notificationtools.js
```diff
@@ -1,89 +1,84 @@
...
   //Check if the window object contains a Notification
-  if (!('Notification' in window)) {
-    return false;
-  } 
-  
-  //The broswer is outdated and does not support notifications
-  else {
-    return true;
-  }
+  if ('Notification' in window) return true
```
arthuro555, line 31: Sorry I know I corrected you but actually this can be simplified further:
```suggestion
  return typeof window.Notification !== "undefined";
```

Extensions/Notification/notificationtools.js
```diff
@@ -1,89 +1,84 @@
...
-gdjs.evtTools.notification.createNotification = function(id, title, msg, img, lng, tag, silence, vibe) {
+gdjs.evtTools.notification.createNotification = function(id, title, body, icon, lang, requireInteraction, silent, vibrate) {
   //Notification configuration see https://developer.mozilla.org/en-US/docs/Web/API/Notification for more information
   const config = {
-    body: msg,
+    body: body,
     data: id,
-    icon: img,
-    lang: lng,
-    tag: tag,
```
arthuro555: Instead of fully removing tag you can use the ID as tag

Extensions/Notification/notificationtools.js
```diff
@@ -0,0 +1,84 @@
...
+  const returnValue = gdjs.evtTools.notification.event[id];
+  gdjs.evtTools.notification.event[id] = false;
+  return returnValue;
+}
+
+gdjs.evtTools.notification.onClose = function(id) {
+  const returnValue = gdjs.evtTools.notification.event[id];
+  gdjs.evtTools.notification.event[id] = false;
+  return returnValue;
+}
```
arthuro555, line 79: Those event still don't care what type of event got triggered.
```suggestion
gdjs.evtTools.notification.onClick = function(id) {
  if(gdjs.evtTools.notification.event[id] && gdjs.evtTools.notification.event[id] === "click") {
    delete gdjs.evtTools.notification.event[id];
    return true;
  }
}

gdjs.evtTools.notification.onShow = function(id) {
  if(gdjs.evtTools.notification.event[id] && gdjs.evtTools.notification.event[id] === "show") {
    delete gdjs.evtTools.notification.event[id];
    return true;
  }
}

gdjs.evtTools.notification.onError = function(id) {
  if(gdjs.evtTools.notification.event[id] && gdjs.evtTools.notification.event[id] === "error") {
    delete gdjs.evtTools.notification.event[id];
    return true;
  }
}

gdjs.evtTools.notification.onClose = function(id) {
  if(gdjs.evtTools.notification.event[id] && gdjs.evtTools.notification.event[id] === "close") {
    delete gdjs.evtTools.notification.event[id];
    return true;
  }
}
```

Extensions/Notification/notificationtools.js
```diff
@@ -0,0 +1,84 @@
...
+
+  gdjs.evtTools.notification.notif[id] = new Notification(title, config);
+
+  gdjs.evtTools.notification.event[id] = gdjs.evtTools.notification.notif[id];
+  gdjs.evtTools.notification.event[id].onclick = () => {gdjs.evtTools.notification.event[id] = "click"}
+  gdjs.evtTools.notification.event[id].onshow = () => {gdjs.evtTools.notification.event[id] = "show"}
+  gdjs.evtTools.notification.event[id].onerror = () => {gdjs.evtTools.notification.event[id] = "error"}
+  gdjs.evtTools.notification.event[id].onclose = () => {gdjs.evtTools.notification.event[id] = "close"}
+
+  console.log(gdjs.evtTools.notification);
```
arthuro555, line 54: Don't forget to remove debugging calls 😉

Extensions/Notification/notificationtools.js
```diff
@@ -0,0 +1,84 @@
...
+    icon: icon,
+    lang: lang,
+    requireInteraction: requireInteraction,
+    silent: silent,
+    vibrate: vibrate.split(",")
+  }
+
+  gdjs.evtTools.notification.notif[id] = new Notification(title, config);
+
+  gdjs.evtTools.notification.event[id] = gdjs.evtTools.notification.notif[id];
```
arthuro555, line 48: What is this for? I think it would interfere with the rest and should be removed
```suggestion
```

Extensions/Notification/notificationtools.js
```diff
@@ -0,0 +1,84 @@
+/**
+ * @memberof gdjs.evtTools
+ * @namespace
+ * @static
+ * @public
```
arthuro555, line 5: public is the default, you can remove it. static is for members of a class, you can remove it as well.

Extensions/Notification/JsExtension.js
```diff
@@ -0,0 +1,185 @@
...
+      )
+      .addParameter('expression', _('Id'), '', false)
+      .addParameter('string', _('Title'), '', false)
+      .addParameter('string', _('Message'), '', false)
+      .addParameter('string', _('(Optional) file location of an image to display'), '', true)
+      .addParameter('string', _('(Optional) change language of text, must be an ISO 2 Letter Language Code'), '', true)
+      .setDefaultValue('false')
+      .addParameter('yesorno', _('(Optional) require interaction to make notification disappear'), '', true)
+      .setDefaultValue('false')
+      .addParameter('yesorno', _('(Optional) notification is silent'), '', true)
```
arthuro555, line 156: The `setDefaultValue` should be after the `addParameter`
```suggestion
      .addParameter('yesorno', _('(Optional) require interaction to make notification disappear'), '', true)
      .setDefaultValue('false')
      .addParameter('yesorno', _('(Optional) notification is silent'), '', true)
      .setDefaultValue('false')
```

Extensions/Notification/notificationtools.js
Add00: This might sound dumb, but what does static mean in this situation?

Extensions/Notification/notificationtools.js
Add00: Originally I just had the notif array, but the notification object always got replaced with a boolean value. So if someone wanted check for a click, it would make an error message since the notification object didn't exist. To get around that I created a second array to handle the events and boolean values.

Extensions/Notification/notificationtools.js
Add00: Oh okay thanks for the explanation.

Extensions/Notification/notificationtools.js
Add00: Do you mean swapping the id for the tag value like so ``tag: id``?

arthuro555: Yes. As you can see on [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Notification/Notification), the tag option is meant as an ID while the data option is meant as arbitrary data.

Add00:
So all of the recommendations are in the latest commit. However, I have tested it on mobile on both as both an html build and as an apk and it doesn't not seem to work. On desktop is work as both an html build and as an executable. I don't know if this commit should be apart of the engine if it does not work on mobile. Since that is where developers are most likely to want to place notifications.

arthuro555:
> it does not work on mobile

In that case, you should try to use a cordova plugin, like https://github.com/katzer/cordova-plugin-local-notifications to make it work for apk export at least. But I *think* you can just add a warning about what platforms this works on and add support for other platforms later via another PR.

4ian:
Android is one of the most used export platform in Gdevelop, and mobile in general is a huge market (more than all other game markets). So I would really really favor that we support it in features that are developed in GDevelop. We can make an exception for certain features, but that should be an exception, not the norm. 

Let's dig a bit deeper to see what we can do :) 
Also, more generally, what is the use case for showing notifications from games?

Add00:
I definitively will investigate the Cordova plugin that arthuro shared. 

As for use cases there are not many, but here are some I thought of off the top of my head. 

- Working in conjunction with the network extension to tell users that there request has been sent/received
- In mystery/puzzle games clues or parts of the puzzle could be sent in a notification
- If a task in a game requires waiting, a notification could be sent alerting the user that the task is done
- Lastly, working in conjunction with the Firebase extension, developers could dynamically send notifications to their users

arthuro555:
> working in conjunction with the Firebase extension

Yes, especially since Firebase messaging cannot show notifications if the app is opened, so users could use this to still show the notification if they want to.

4ian:
Make sense! Let me know if this can work with Cordova then :)

Add00:
Unfortunately, my hard drive broke last night and I lost all of my files, including the mostly finished notification extension. I would have to start over from the last commit and right now I don't think I could mentally handle that. So, I guess what I'm trying to say is that I have given up on finishing this feature. Sorry :(

Official-IceCreeperPE:
where is the download?

luca-zani:
Nice

arthuro555:
Like Add00 said, this PR is abandoned. Additionally, note that 
1. This is an engine code change, you cannot "download" it, someone would need to manually build it and give the build to you,
2. Unnecessary and nonconstructive comments like ´nice´ are considered spam.

PR #2104, opened 2020-11-19
Add line color option to grid setup dialog

ssangervasi:
## What was changed
This introduces a `gridColor` field to the UI settings that can be set from the grid setup dialogue. This option is used in the grid renderer to change the color & alpha of the grid lines.

## Motivation
I often find when working a scene with many dark colors that the default grid makes it hard to spot object edges. This reduces the usefulness of the snap-to-grid feature, as I have to toggle the grid off to see the objects I'm moving. Being able to reduces the opacity and/or contrast of the grid will totally fix this.

## Demo
Updated: 
gdev-grid-opacity

Original:
[Before apply](https://user-images.githubusercontent.com/2236777/99724263-393a0080-2a68-11eb-89c8-29f0b8d9ed51.png)
[After apply.](https://user-images.githubusercontent.com/2236777/99724273-3c34f100-2a68-11eb-9d50-ff6387147605.png)

## UI Questions  
Updated answer: Switch to ColorField, stack fields in columns.

Original:
> I don't _love_ the way the color picker show's up in the dialogue, as there's a toolbar with nothing else in it. However, I'm not sure what a nicer layout would be; the picker doesn't really fit next to the the text fields either. I went with the matching the EventTextDialogue for consistency.

### DIFF
--- a/newIDE/app/src/InstancesEditor/Grid.js
+++ b/newIDE/app/src/InstancesEditor/Grid.js
@@ -1,4 +1,12 @@
 import * as PIXI from 'pixi.js-legacy';
+import { rgbColorToHexNumber, type RGBColor } from '../Utils/ColorTransformer';
+
+// Equal to #6868E8
+const DEFAULT_COLOR: RGBColor = {
+  r: 104,
+  g: 104,
+  b: 232,
+};
 
 export default class SelectionRectangle {
   constructor({ viewPosition, options }) {
@@ -25,10 +33,14 @@ export default class SelectionRectangle {
       return;
     }
 
+    const gridColor = options.gridColor || DEFAULT_COLOR;
+    const gridHex = rgbColorToHexNumber(gridColor);
+    const gridAlpha = gridColor.a || 1;
+
     this.pixiGrid.visible = true;
     this.pixiGrid.clear();
-    this.pixiGrid.beginFill(0x6868e8);
-    this.pixiGrid.lineStyle(1, 0x6868e8, 1);
+    this.pixiGrid.beginFill(gridHex);
+    this.pixiGrid.lineStyle(1, gridHex, gridAlpha);
     this.pixiGrid.fill.alpha = 0.1;
     this.pixiGrid.alpha = 0.8;
 
--- a/newIDE/app/src/SceneEditor/SetupGridDialog.js
+++ b/newIDE/app/src/SceneEditor/SetupGridDialog.js
@@ -2,12 +2,16 @@ import { Trans } from '@lingui/macro';
 import React, { Component } from 'react';
 import FlatButton from '../UI/FlatButton';
 import TextField from '../UI/TextField';
+import { ResponsiveLineStackLayout } from '../UI/Layout';
 import Dialog from '../UI/Dialog';
+import ColorField from '../UI/ColorField';
 
 export default class SetupGridDialog extends Component {
   constructor(props) {
     super(props);
-    this.state = { ...props.gridOptions };
+    this.state = {
+      ...props.gridOptions,
+    };
   }
 
   _onApply = () => {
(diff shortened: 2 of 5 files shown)


newIDE/app/src/SceneEditor/SetupGridDialog.js
```diff
@@ -43,6 +48,18 @@ export default class SetupGridDialog extends Component {
         open={this.props.open}
         onRequestClose={this.props.onCancel}
       >
+        <MiniToolbar>
+          <MiniToolbarText>
+            <Trans>Line color:</Trans>
+          </MiniToolbarText>
+          <ColorPicker
```
4ian, line 55: I think that here we'll avoid the MiniToolbar entirely, as it's a dialog containing fields.
Speaking of fields, can you try using `ColorField`? Might go better with the other TextField of this dialog.

newIDE/app/src/SceneEditor/SetupGridDialog.js
```diff
@@ -43,6 +48,18 @@ export default class SetupGridDialog extends Component {
...
+            color={this.state.gridColor}
+            onChangeComplete={color => {
+              this.setState({ gridColor: color.rgb });
+            }}
+          />
+        </MiniToolbar>
+
         <TextField
           floatingLabelText={<Trans>Cell width (in pixels)</Trans>}
```
4ian, line 64: We should probably group the fields 2 by 2 in a `<ResponsiveLineStackLayout>`, one for the width/height, one for X/Y.

newIDE/app/src/SceneEditor/SetupGridDialog.js
4ian: (Move the MiniToolbarText to the `floatingLabelText` prop of `ColorField`)

newIDE/app/src/InstancesEditor/Grid.js
```diff
@@ -1,4 +1,17 @@
...
+const DEFAULT_COLOR: RGBColor = {
+  r: 104,
+  g: 104,
+  b: 232,
+};
+
+const rgbColorToHex = (rgbColor: RGBColor) => {
+  const { r, g, b } = rgbColor;
+  return PIXI.utils.rgb2hex([r, g, b].map(c => c / 255));
+};
```
4ian, line 14: Could you move this to `newIDE/app/src/Utils/ColorTransformer.js`, along with the other methods? Also try not to use `PIXI` inside this file, to make it independent from the rendering engine.

newIDE/app/src/InstancesEditor/Grid.js
```diff
@@ -25,10 +38,14 @@ export default class SelectionRectangle {
       return;
     }
 
+    const gridColor = options.gridColor || DEFAULT_COLOR;
+    const gridHex = rgbColorToHex(gridColor);
+    const gridAlpha = gridColor.a || 1;
```
4ian, line 43: Seems like the alpha is not selectable, we might keep it always at 1?

4ian (commented):
Thanks for the PR! Overall make sense. See my comments about some ideas for the UI :)

newIDE/app/src/InstancesEditor/Grid.js
ssangervasi: Alpha is configurable with the color picker, and it works as expected here. The fallback of 1 comes up because `a` is an optional property, or if the user selects `0` which would be confusing to have a visible grid that is invisible.

newIDE/app/src/SceneEditor/SetupGridDialog.js
```diff
@@ -43,38 +48,51 @@ export default class SetupGridDialog extends Component {
...
+        <ResponsiveLineStackLayout>
+          <ColorField
+            floatingLabelText={<Trans>Line color</Trans>}
+            color={this.state.gridColor}
+            onChangeComplete={color => {
+              this.setState({ gridColor: color.rgb });
+            }}
+          />
+        </ResponsiveLineStackLayout>
+        <ResponsiveLineStackLayout>
```
ssangervasi, line 60: I suggest using "hide whitespace changes" in this diff.

4ian:
Looks great! Merging now :) 
Thanks for the high quality PR, will be very useful 👍

PR #2107, opened 2020-11-21
Fix lights crash

arthuro555:
`pixiRenderer.renderTexture` is used but only present on the webgl renderer. Using it on the Canvas renderer causes a crash, so return early if the renderer is not supported.

### DIFF
--- a/GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
+++ b/GDJS/Runtime/pixi-renderers/layer-pixi-renderer.js
@@ -335,6 +335,7 @@ gdjs.LayerPixiRenderer.prototype._updateRenderTexture = function () {
  */
 gdjs.LayerPixiRenderer.prototype._replaceContainerWithSprite = function () {
   if (!this._pixiRenderer) return;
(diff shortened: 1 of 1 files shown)


4ian:
Sounds like a good thing, you've been able to test on a computer where canvas is used/or force canvas usage? :) 

cc @HarsimranVirk if you want to double check

arthuro555:
Yes, apparently games run in the windows sandbox run on canvas for some reason, I fou d out while playtesting a game on the discord. This modification removed the crash.

HarsimranVirk:
This definitely seems like a good check. We unfortunately missed "webgl checks" for layer's renderer (and only included checks in the light object's renderer). Thanks for having a look at it :)

4ian:
> apparently games run in the windows sandbox run on canvas for some reason

Interesting, might be a performance problem and in the future canvas might not be supported anymore, would be cool to check why.

In the meantime, merging this. Thanks!

Bouh:
> Sounds like a good thing, you've been able to test on a computer where canvas is used/or force canvas usage? :)

When you disable hardware acceleration on chrome webgl 1 & 2 are disabled and canvas is forced ;)
if you have a game for web i can test btw

PR #2109, opened 2020-11-22
Destroying renderers of deleted objects

arthuro555:
This is not a very good solution, especially if we one day switch/add renderers.


4ian:
As discussed I think it would be good to have a reproduction case where you can clearly demonstrate that some listeners are leaked, so that we can know exactly if/what is missing :)

arthuro555:
Nevermind after closer inspection

PR #2112, opened 2020-11-25
Add Games Dashboard to register games and view metrics

4ian:
This adds a "Games Dashboard" in the profile dialog, and a way to register your game on the GDevelop APIs. This allows you to then see metrics from your games: sessions, players, retention, and in the future the session times and a few other helpful metrics - without going though a third party solution.

*It's important to note that these metrics are entirely anonymous, no personal information from players is ever collected and all data is "aggregated", meaning that it's useful to understand your audience, but you can't know about a particular player, because the information is not stored and players are not tracked individually.*

The reasoning is that while GDevelop does a good job at helping people to create and export game, there is little that is done for understanding how successful is your game ✨ The advantage of this built-in solution is that all of this does not go through a third party that would keep/give access/sell data or track players. 

For now only the session count is shown to the unsubscribed users. Because this infrastructure has a cost and does not rely on a third party processing the data, we ask people to support GDevelop by getting a subscription to get the full view on the metrics. In the future if we have more metrics, we can also show more to unsubscribed users.

Help was written here: http://wiki.compilgames.net/doku.php/gdevelop5/interface/games-dashboard and a way to register a game is also shown after exporting a game. It should be very easy to use (🤞) and I think very useful for most creators, but it's also still quite basic: better metrics will probably be added later.


Wend1go:
Nice feature 👍 
But can this be completely deactivated?
Or ist there an action to let the player choose to take part in this?

4ian:
> Or ist there an action to let the player choose to take part in this?

Good call, I overlooked this! I'll provide an action to enable/disable this. So you can use the action when the game is launched and enable it back later according to your player choice/your policy.

4ian:
Added in 2b156ef14707f8804288254e817d689fe11ff33d

PR #2113, opened 2020-11-25
Add a for each structure event

arthuro555:
This new event let you iterate over children of a structure easily.
It repeats it's events for every children of a structure variable, and stores in a variable the value of the target child for every iteration.
It is basically a for loop for children of a structure.


Core/GDCore/Events/Builtin/ForEachStructureEvent.h
```diff
@@ -0,0 +1,93 @@
...
+   * The structure is the structure variable that will be iterated on.
+   */
+  const gd::String& GetStructure() const { return structure; };
+
+  /**
+   * \brief Set the structure attached to the event.
+   *
+   * The structure is the structure variable that will be iterated on.
+   */
+  void SetStructure(gd::String newStructure) { structure = newStructure; };
```
4ian, line 54: Let's call this the "IterableVariableName".

"Name" because what we're storing here is a name (not a gd::Variable)
"Iterable" because it's a generic way to say "something we can loop on", and in the future if we introduce arrays in addition to structures, it will still apply :)

Core/GDCore/Events/Builtin/ForEachStructureEvent.h
```diff
@@ -0,0 +1,93 @@
...
+   */
+  const gd::String& GetVariable() const { return variable; };
+
+  /**
+   * \brief Set the variable attached to the event.
+   *
+   * The variable is the variable that will contain the value of the linked
+   * child of the iteration.
+   */
+  void SetVariable(gd::String newVariable) { variable = newVariable; };
```
4ian, line 70: Similarly, let's call it the Get/SetIteratorVariableName

"Iterator Variable" because an "iterator" is what is being called for each iteration on the iterable. This iterator has a parameter which is the child, which we call the "Iterator Variable".
"Name" because it's a name that we're storing here.

Core/GDCore/Extensions/Builtin/CommonInstructionsExtension.cpp
```diff
@@ -118,6 +119,13 @@ BuiltinExtensionsImplementer::ImplementsCommonInstructionsExtension(
                      "res/foreach.png",
                      std::make_shared<gd::ForEachEvent>());
 
+  extension.AddEvent("ForEachStructure",
```
4ian, line 122: I'm inclined to call this `ForEachChildVariable`.

The term "Variable children" allows to extend this to arrays later.

Core/GDCore/Extensions/Builtin/CommonInstructionsExtension.cpp
```diff
@@ -118,6 +119,13 @@ BuiltinExtensionsImplementer::ImplementsCommonInstructionsExtension(
...
+                     _("For each strucure child"),
```
4ian, line 123: `For each child variable (of a structure)`

Core/GDCore/Extensions/Builtin/CommonInstructionsExtension.cpp
```diff
@@ -118,6 +119,13 @@ BuiltinExtensionsImplementer::ImplementsCommonInstructionsExtension(
...
+                     _("Repeat the event for each child of a stucture."),
```
4ian, line 124:
```suggestion
                     _("Repeat the event for each child of a structure."),
```

GDJS/GDJS/Extensions/Builtin/CommonInstructionsExtension.cpp
```diff
@@ -408,6 +409,87 @@ CommonInstructionsExtension::CommonInstructionsExtension() {
...
+        // Write final code
+        gd::String iteratorReferenceVariableName =
+            "iteratorReference" +
+            gd::String::From(context.GetContextDepth());
+        gd::String structureChildVariableName =
+            "structureChild" +
+            gd::String::From(context.GetContextDepth());
+        gd::String structureReferenceVariableName =
+            "structureReference" +
+            gd::String::From(context.GetContextDepth());
```
4ian, line 455: To keep things consistent with the changes proposed in the event, `structure` => `iterable`

GDJS/GDJS/Extensions/Builtin/CommonInstructionsExtension.cpp
```diff
@@ -408,6 +409,87 @@ CommonInstructionsExtension::CommonInstructionsExtension() {
...
+                      ".val";
+
+        // Prepare object declaration and sub events
+        gd::String subevents =
+            codeGenerator.GenerateEventsListCode(event.GetSubEvents(), context);
+        gd::String objectDeclaration =
+            codeGenerator.GenerateObjectsDeclarationCode(context) + "\n";
+
+        // Write final code
+        gd::String iteratorReferenceVariableName =
```
4ian, line 447: Nice naming 👍 This is indeed an "iterator variable"

newIDE/app/src/EventsSheet/EventsTree/Renderers/ForEachStructureEvent.js
```diff
@@ -0,0 +1,250 @@
...
+};
+
+export default class ForEachStructureEvent extends React.Component<
+  EventRendererProps,
+  *
+> {
+  _objectField: ?ObjectField = null;
+  state = {
+    editingVariable: false,
+    editingStructure: false,
```
4ian, line 39: Like the renaming in the class, let's rename these 
```suggestion
    editingChildVariableName: false,
    editingIterableVariableName: false,
```

GDJS/GDJS/Extensions/Builtin/CommonInstructionsExtension.cpp
```diff
@@ -408,6 +409,87 @@ CommonInstructionsExtension::CommonInstructionsExtension() {
...
+          outputCode += "} //Subevents end.\n";
+        }
+        outputCode += "}\n";
+        outputCode += "}\n";
+
+        return outputCode.FindAndReplace("$STRUCTURE_CHILD_VARIABLE", structureChildVariableName)
+            .FindAndReplace("$ITERATOR_VARIABLE", iteratorReferenceVariableName)
+            .FindAndReplace("$STRUCTURE_VARIABLE", event.GetStructure())
+            .FindAndReplace("$STRUCTURE_REFERENCE", structureReferenceVariableName)
+            .FindAndReplace("$CHILD_VARIABLE", event.GetVariable());
```
4ian, line 490: $CHILD_VARIABLE => $ITERATOR_VARIABLE_NAME

```suggestion
            .FindAndReplace("$ITERATOR_VARIABLE_NAME", event.GetIteratorVariableName());
```

GDJS/GDJS/Extensions/Builtin/CommonInstructionsExtension.cpp
```diff
@@ -408,6 +409,87 @@ CommonInstructionsExtension::CommonInstructionsExtension() {
...
+          outputCode += "\n{ //Subevents: \n";
+          outputCode += subevents;
+          outputCode += "} //Subevents end.\n";
+        }
+        outputCode += "}\n";
+        outputCode += "}\n";
+
+        return outputCode.FindAndReplace("$STRUCTURE_CHILD_VARIABLE", structureChildVariableName)
+            .FindAndReplace("$ITERATOR_VARIABLE", iteratorReferenceVariableName)
+            .FindAndReplace("$STRUCTURE_VARIABLE", event.GetStructure())
```
4ian, line 488: Along with the class renaming:

```suggestion
            .FindAndReplace("$ITERABLE_VARIABLE_NAME", event.GetIterableVariableName())
```

"VARIABLE_NAME" implies it's a string (while if you set just "VARIABLE", one could think that inside this thing there will be a `gdjs.Variable`)

GDJS/GDJS/Extensions/Builtin/CommonInstructionsExtension.cpp
```diff
@@ -408,6 +409,87 @@ CommonInstructionsExtension::CommonInstructionsExtension() {
...
+        outputCode += actionsCode;
+        if (event.HasSubEvents()) {
+          outputCode += "\n{ //Subevents: \n";
+          outputCode += subevents;
+          outputCode += "} //Subevents end.\n";
+        }
+        outputCode += "}\n";
+        outputCode += "}\n";
+
+        return outputCode.FindAndReplace("$STRUCTURE_CHILD_VARIABLE", structureChildVariableName)
```
4ian, line 486: Should this be called the `ITERATOR_KEY_VARIABLE_NAME`?
```suggestion
        return outputCode.FindAndReplace("$ITERATOR_KEY_VARIABLE_NAME", iteratorKeyVariableName)
```

GDJS/GDJS/Extensions/Builtin/CommonInstructionsExtension.cpp
```diff
@@ -408,6 +409,87 @@ CommonInstructionsExtension::CommonInstructionsExtension() {
...
+        gd::String objectDeclaration =
+            codeGenerator.GenerateObjectsDeclarationCode(context) + "\n";
+
+        // Write final code
+        gd::String iteratorReferenceVariableName =
+            "iteratorReference" +
+            gd::String::From(context.GetContextDepth());
+        gd::String structureChildVariableName =
+            "structureChild" +
+            gd::String::From(context.GetContextDepth());
```
4ian, line 452: I would go for `iteratorKeyVariableName`

newIDE/app/src/EventsSheet/EventsTree/Renderers/ForEachStructureEvent.js
```diff
@@ -0,0 +1,250 @@
...
+  actionsList: {
+    flex: 1,
+  },
+};
+
+export default class ForEachStructureEvent extends React.Component<
+  EventRendererProps,
+  *
+> {
+  _objectField: ?ObjectField = null;
```
4ian, line 36: Rename this to variableField and update the type ;)

4ian (commented):
Very nice! 👍👍

I'm usually worried of adding new events because it should be very rare, but I think in this case it would be a good addition to "the language" :)

Overall the logic looks good to me - my comments are mostly nitpicking about naming.

As with everything, naming is often hard ;) I suggest we use these terms (in the code, not in the user facing interface):
- `Iterable` for the structure we loop ('iterate") on. (which is what you called structure).
- `IteratorVariable` is the variable which is set to the value of the child (or the reference to the child).
- `IteratorKey` is the name of the child.

In pseudo JS code, the naming would be like this:
```
const iterable = { 
  key1: child1,
  key2: child2
}

for (const iteratorKey in iterable) {
  const iteratorVariable = iterable[iteratorKey]
}
```

I've added comments where I hopefully follow this convention (sorry it's a lot and kind of hard to wrap your head around 🙃). Basically, rename what you used to call:
- "Structure" to "Iterable"
- "Child" to "IteratorVariable"
- introduce the notion of "IteratorKey" instead of reusing the term child.

Let me know what you think!

arthuro555:
As a side note, I tried to add some integration tests (libGD tests) but the minimal GDJS shim (and more precisely the variables shim) is too slim for the ForEachChildVariable event, is there a way to run the tests in the "full" GDJS environment?

arthuro555:
Ok, I think I did all the renaming you requested. I also fixed a bit the styling of the event renderer. 
I am really look forward to that change, I wanted something like this ever since I began using GDevelop 😄 

As another sidenote, did you see my answer about possibilities of doing [callbacks in GDevelop on the forum](https://forum.gdevelop-app.com/t/callbacks-argument/26197/3?u=arthuro555)? I took multiple hours thinking it through and writing this post and designing mockups (and it is the feature I really want the most for GDevelop 😛).

4ian:
Thanks for the changes! I'll take a new look

> As a side note, I tried to add some integration tests (libGD tests) but the minimal GDJS shim (and more precisely the variables shim) is too slim for the ForEachChildVariable event, is there a way to run the tests in the "full" GDJS environment?

Nope, there is nothing like this because it would need a full fledged browser to run, or a reimplementation of all PIXI renderers into something called "Headless renderers", that would do nothing (so that we could test in Node.js).
Maybe you can identify what is missing for your test and rework the minimal GDJS shim for it? You can also make a test that outputs some code to verify it looks correct (even if the best is to actually run it :))

4ian:
> As another sidenote, did you see my answer about possibilities of doing callbacks in GDevelop on the forum? I took multiple hours thinking it through and writing this post and designing mockups (and it is the feature I really want the most for GDevelop 😛).

So I read and never answered, sorry! It's a nice proposal I can tell you spent a lot of time. I'll answer tonight :)

Core/GDCore/Events/Builtin/ForEachChildVariableEvent.cpp
```diff
@@ -0,0 +1,73 @@
...
+ * GDevelop Core
+ * Copyright 2008-2016 Florian Rival (Florian.Rival@gmail.com). All rights
+ * reserved. This project is released under the MIT License.
+ */
+
+#include "ForEachChildVariableEvent.h"
+#include "GDCore/Events/Serialization.h"
+#include "GDCore/Events/Tools/EventsCodeNameMangler.h"
+#include "GDCore/Serialization/SerializerElement.h"
+#include "GDCore/TinyXml/tinyxml.h"
```
4ian, line 11: You should be able to remove this :) SerializerElement is enough.

4ian (commented):
Looks good! Before going ahead with this, I'm wondering if we should also expose an, optional, variable that would contain the "key" of the iteration, i.e: the name of the child that is currently used in the iteration.

This would imply a new "IteratorKeyVariableName" in the event. In the code generation, it should be fairly easy to do: 

```
const $ITERATOR_KEY_REFERENCE = $ITERATOR_KEY_VARIABLE_NAME ? runtimeScene.getVariables().get($ITERATOR_KEY_VARIABLE_NAME) : null;
// ...
for ( ... ) {
  if ($ITERATOR_KEY_REFERENCE) $ITERATOR_KEY_REFERENCE.setString($ITERATOR_KEY);
}
```

An empty string for the name means that no variable was chosen, so we don't get a variable in this case.
Finally, in the editor, we can choose a variable, but mark it as `<Select a variable (optional)>` and avoid putting the styling for missing parameters.

arthuro555:
I chose to use the name KeyIteratorVariableName instead of IteratorKeyVariableName, because we iterate over the keys, and the variable is not the key of the iterator but of the iterated value. I also renamed IteratorVariableName to ValueIteratorVariableName to have a more consistent and understandable naming.
I also made both iterators optional, as you may want to only use one of them. The code generator also takes care of not including any unnecessary code.

newIDE/app/src/EventsSheet/EventsTree/Renderers/ForEachChildVariableEvent.js
```diff
@@ -0,0 +1,341 @@
...
+    const iterableName = forEachChildVariableEvent.getIterableVariableName();
+
+    return (
+      <div
+        style={styles.container}
+        className={classNames({
+          [executableEventContainer]: true,
+        })}
+      >
+        <div style={{ marginLeft: '5px' }}>
```
4ian, line 123: Let's put the style in `styles`

newIDE/app/src/EventsSheet/EventsTree/Renderers/ForEachChildVariableEvent.js
```diff
@@ -0,0 +1,341 @@
...
+    );
+    const valueIteratorName = forEachChildVariableEvent.getValueIteratorVariableName();
+    const keyIteratorName = forEachChildVariableEvent.getKeyIteratorVariableName();
+    const iterableName = forEachChildVariableEvent.getIterableVariableName();
+
+    return (
+      <div
+        style={styles.container}
+        className={classNames({
+          [executableEventContainer]: true,
```
4ian, line 120: This seems to be missing `largeSelectableArea`/`largeSelectedArea`

newIDE/app/src/EventsSheet/EventsTree/Renderers/ForEachChildVariableEvent.js
```diff
@@ -0,0 +1,341 @@
...
+              />
+              {valueIteratorName.length !== 0 ? (
+                <span>{valueIteratorName}</span>
+              ) : (
+                <span>
+                  <Trans>{`<Select a variable> (optional)`}</Trans>
+                </span>
+              )}
+            </span>
+            <Trans>, </Trans>
```
4ian, line 188: You can put a single `<Trans>` from "For every child in" until the last comma. This will generated a translation string like "For every child in {0}, store the child in variable {1},", which can be translated with the full context (and with the proper order, which is not necessarily the same according to the language). You could actually remove the `br` and make the whole thing translated as a single sentence. The new line does not seem really necessary to me :)

4ian:
Alright, sounds good!
Three things to fix:
- there is no border when the event is selected, there is surely a missing `largeSelectableArea`/`largeSelectedArea` in the event?
- translations for the whole sentence instead of part of it,
- We're doing a `._children = ` to have the "value" variable to be a "reference" to the "child" (which I think is good). Should we make a function in `gdjs.Variable` for this?

newIDE/app/src/EventsSheet/EventsTree/Renderers/ForEachChildVariableEvent.js
arthuro555: Oh I didn't knew the Trans was that smart.

> The new line does not seem really necessary to me :)

Well, it doesn't break the line automatically so on smaller screens you cannot use or see the last variable field.

arthuro555:
> We're doing a ._children = to have the "value" variable to be a "reference" to the "child" (which I think is good). Should we make a function in gdjs.Variable for this?

Yes, I think it would be a good idea as well.

newIDE/app/src/EventsSheet/EventsTree/Renderers/ForEachChildVariableEvent.js
arthuro555: Indeed, I removed it because it overlapped with the field while I experimented with the styles. I'm adding it back now.

newIDE/app/src/EventsSheet/EventsTree/Renderers/ForEachChildVariableEvent.js
4ian: > Well, it doesn't break the line automatically so on smaller screens you cannot use or see the last variable field.

Add to the style (next to the marginLeft):
```
    whiteSpace: 'normal',
    wordWrap: 'break-word',
```

which restore the usual HTML behavior of divs :) (white-space was set to `nowrap` for some reasons, I can't remember now, in the events tree)

4ian (approved):
💯

4ian:
Thanks a lot for working on this, I think it's a flexible event that is quite advanced but very useful for power users. Let's make a page about it on the wiki too ;)

PR #2137, opened 2020-12-15
Make variable types strictly defined and add Array and Boolean types

arthuro555:
This PR has following goals:
1. Make variables types strictly defined by the user instead of inferred by the IDE
2. Add a boolean variable type
3. Add an array variable type

The first change allows to have a less complex code due to not having to infer types, and avoid bugs due to incorrect inferring.

The second change is a convenient type for a variable whose value is binary. It's advantage over a number variable with 1 or 0 is that it cannot have other values, so there is less space for bugs, and there is also an action to toggle its value.

The third change is a convenient structure variable for lists. The key are not decided by the user, but managed by the engine, making sure that there are no holes, that the variable length allows to iterate over the items safely (no risk of having items keyed with something abnormal and the length not matching the amount of correctly keyed elements), and has some other advantages.


4ian:
Hey, nice this feature is becoming more and more important :) 

> Move the type from being loosely defined" (It was checked by verifying if it is a structure or a number and if it is neither then it is a string). There is now an enum of types, and the variable is aware of it's type.

This sounds fair.

> The recursive deletion of arrays, I am not experienced with c++ iterators.

In which case do you need it?
There will be a bit of change to do on the C++ side and then `gdjs.Variable` to update, maybe some code generation (not sure)?

I'll check the code that you have so far :)

Core/GDCore/Project/Variable.cpp
```diff
@@ -66,99 +105,106 @@ const Variable& Variable::GetChild(const gd::String& name) const {
...
 
-void Variable::ClearChildren() {
-  if (!isStructure) return;
-  children.clear();
-}
+bool Variable::HasIndex(const size_t index) const {
+  return type == TYPES::ARRAY && childrenList.size() < index;
+};
+
+Variable& Variable::GetIndex(const size_t index) {
```
4ian, line 139: Should this be called `GetAt`? As we're getting something (an "item" or an "element" at some position), but we're not getting an index.

Core/GDCore/Project/Variable.cpp
```diff
@@ -66,99 +105,106 @@ const Variable& Variable::GetChild(const gd::String& name) const {
...
+    childrenList.push_back(std::make_shared<gd::Variable>());
+  return *childrenList[index];
+};
+
+const Variable& Variable::GetIndex(const size_t index) const {
+  if (childrenList.size() < index) return *std::make_shared<gd::Variable>();
+  return *childrenList.at(index);
+};
+
+void Variable::RemoveIndex(const size_t index) {
```
4ian, line 150: Same, should it be called RemoveAt?

Core/GDCore/Project/Variable.cpp
```diff
@@ -66,99 +105,106 @@ const Variable& Variable::GetChild(const gd::String& name) const {
...
 void Variable::SerializeTo(SerializerElement& element) const {
-  if (!isStructure)
-    element.SetAttribute("value", GetString());
-  else {
+  element.SetIntAttribute("type", static_cast<int>(GetType()));
+
+  if (type == TYPES::STRING) {
+    element.SetStringAttribute("value", GetString());
+  } else if (type == TYPES::NUMBER) {
+    element.SetBoolAttribute("value", GetValue());
```
4ian, line 161: Wrong method

Core/GDCore/Project/Variable.cpp
```diff
@@ -189,23 +235,35 @@ void Variable::RemoveRecursively(const gd::Variable& variableToRemove) {
       it++;
     }
   }
-  isStructure = !children.empty();
+  for (auto it = childrenList.begin(); it != childrenList.end();) {
+    if (it->get() == &variableToRemove) {
+      childrenList.erase(it);
+    } else {
+      it->get()->RemoveRecursively(variableToRemove);
```
4ian, line 242:
```suggestion
      it->RemoveRecursively(variableToRemove);
```

I think (or `(*it)->`, but no need for get)

Core/GDCore/Project/Variable.h
```diff
@@ -26,18 +28,33 @@ namespace gd {
  */
 class GD_CORE_API Variable {
  public:
+  enum TYPES {
```
4ian, line 31: Nitpicking: enum names should usually be singular. Here it's `TYPE`. This is because each value represents "a" type.
See if you can use `enum class` - I've not done it in the rest of the codebase but it might gives better scoping.

Finally, I think we don't need UPPER_CASING here, at least I've seen usage of PascalCase for both the name (`Type`) and the content (`String`, `Number`, etc...) in the codebase, and because an enum is a type, it seems a bit too much to use UPPER_CASE (which is more for constants).

Core/GDCore/Project/Variable.cpp
```diff
@@ -20,25 +20,64 @@ namespace gd {
...
+  } else if (type == TYPES::BOOLEAN) {
+    str = boolVal ? "1" : "0";
+    type = TYPES::STRING;
+    return str;
   }
 
+  // It isn't possible to convert a structural type to a string
+  str = "0";
+  type = TYPES::STRING;
   return str;
```
4ian, line 57: There was a specific behavior that might be important to keep, but I'm not sure how important it is:

Before, if you set a variable to a number/string, then you somehow set a children (so that it's becoming a structure), then getting the string (transforming it back to a string), you'll get the string (or number) initially set.
While here you'll get "0". 

Seems like a change in behavior. It might be a good idea to keep the fact that the "primitive type" is separated from the "structural type". You can be only one, but if you convert to a structure/array and back to a primitive, we've retained your value (which, again, is not something I would recommend at all and that we could forbid in the future).

Core/GDCore/Project/Variable.h
```diff
@@ -183,20 +218,53 @@ class GD_CORE_API Variable {
...
+  /** \name Array
+   * Methods used when the variable is considered as an array.
    */
   ///@{
+
+  /**
+   * \brief Return true if the variable is an array and has the specified
+   * index.
+   */
+  bool HasIndex(const size_t index) const;
```
4ian, line 230: I think we should remove `HasIndex` from the public interface and just manually make the check for type is array and `childrenList.size()` whenever it's needed.

GDevelop.js/types/gdvariablepointer.js
```diff
@@ -0,0 +1,6 @@
+// Automatically generated by GDevelop.js/scripts/generate-types.js
+declare class gdVariablePointer {
```
4ian, line 2: Nitpicking: let's call this a `SharedPtrVariable` (much like we have `UniquePtrExpressionNode`).

BUT let's check if we even need this class. I'm not sure, because in 99% of cases we don't need to expose the shared pointers to the JavaScript - it's an implementation detail and we can use `GetAt()`/`at()` to get reference to a variable (no shared pointer) and `GetChildrenCount()`/`size()` to get the count.

GDevelop.js/Bindings/Bindings.idl
```diff
@@ -166,25 +171,44 @@ interface PairStringVariable {
...
     unsigned long GetChildrenCount();
-    boolean IsNumber();
-    boolean IsStructure();
     boolean Contains([Const, Ref] Variable variableToSearch, boolean recursive);
     void RemoveRecursively([Const, Ref] Variable variableToRemove);
 
+    boolean HasIndex([Const] unsigned long index);
+    [Ref] Variable GetIndex([Const] unsigned long index);
+    void RemoveIndex([Const] unsigned long index);
+    [Const, Ref] VectorVariable GetAllChildrenList();
```
4ian, line 210: Let's see if we can avoid exposing this. This is exposing the internal of the class, and it's dangerous/painful to use from JS. Might as well avoid it entirely :)

Core/GDCore/Project/Variable.cpp
arthuro555: This is kinda semi working only as restarting the ide (or generally serializing and deserialising) as the serializer only serialized the structure part or the value part. I thought therefore it was unintended behavior.

GDevelop.js/types/gdvariablepointer.js
arthuro555: When I had put only `[Const, Ref] Variable at();`, it made a compilation error that said that it couldn't convert a pointer to a smart pointer.

GDevelop.js/Bindings/Bindings.idl
arthuro555: I thought those might be needed for the initial variables editors?

4ian:
In the game engine and action/conditions side, we'll need:
- [ ] Update to `gdjs.Variable` to make support array/booleans. 

Actions/conditions/expressions for booleans:
- [ ] Condition to check the value of a boolean variable
- [ ] Action to set a boolean variable

Actions/conditions for arrays (might need more careful design). Basically they should represent the basic functions available on an array in most languages:
- [ ] Actions to set the value at a specified index of an array to a string/number/boolean (3 actions).
- [ ] Action to push, at the end of an array a string/number/boolean (3 actions).
- [ ] Expression/condition to get/compare the number of items in the array (`ArrayLength(MyArrayVariable)`?) .

And some expression parsing update:
- [ ] Should we make an expression to get the string/number at the specified index of an array? Nope! But we'll need to update the expression parser so that indices can be specified: `VariableString(MyArray[3])` to get the string at the 3rd position (well, fourth position, as array indices start at 0).

And finally update the IDE to allow making arrays in addition to structures in the variable editor :) And update the function converting JSON to a variable to handle arrays :) 

In the future (not this PR), we might update `yesorno` and `trueorfalse` parameters and introduce expressions returning booleans - but this is an entirely new capability/topic and will need careful design. For now the conditions/actions/expressions will be more than enough :)

GDevelop.js/types/gdvariablepointer.js
arthuro555: Ah yes didn't see there was a wrapped at that calls get automatically. Trying it out right now.

Core/GDCore/Project/Variable.cpp
arthuro555: GetAt and RemoveAt clash with the variable container macro that converts GetAt to Get and RemoveAt to Remove (Because of overloading)

Core/GDCore/Tools/FileStream.h
```diff
@@ -24,7 +24,7 @@ namespace gd {
 class GD_CORE_API FileStream : public std::iostream {
  public:
 #if defined(WINDOWS) && __GLIBCXX__
-  using InternalBufferType = __gnu_cxx::stdio_filebuf<char>;
+  using InternalBufferType = std::basic_filebuf<char>;
```
arthuro555, line 27: Not sure about that, but I couldn't compile the tests without that change.

Core/GDCore/Project/Variable.h
```diff
@@ -28,15 +28,15 @@ namespace gd {
  */
 class GD_CORE_API Variable {
  public:
-  enum class Type {
+  enum class Type : char {
     // Primitive types
-    String,
-    Number,
-    Boolean,
+    String = 't',
```
4ian, line 33: This might be a bit cryptic when reading just the JSON. What do you think of serializing the full name ("string", "number", "boolean", etc...)?

Core/GDCore/Tools/FileStream.h
4ian: Let's assume it's fine, I've not compiled tests on Windows since a while so it might be a necessary update.

Core/GDCore/Project/Variable.h
arthuro555: Enums only accept numbers. I used chars as they are numbers but can represent text. I believe that 't', 'n' etc are more readable than 0, 1, 2 etc. I originally tried to map the enums to string names but making a switch function like 
```cpp
switch (type) {
  case(Type::String): 
    return "string";
  // etc
}
```
didn't work because I cannot return temporary strings. I also tried to make a static map like 
```cpp
static const std::map<Type, gd::String&> TypeNames = {
  {Type::String = "string"}
};
```
Also throwed compilation errors. I feel really stupid but I don't know how I am meant to do it :/

4ian: I think the simplest is to go for a switch (or a plain old if/else). It's what I've done in EventsFunction.cpp
Not the "prettiest", but if it's just for the serialization and unserialization, that's fair enough (sure you have to remember it if we had one day a new type, but that's a a one in a year or one in a decade event :)) 

If you really want to have a separate function (which would be a good idea in case you need this in more than one place), you can make one returning a gd::String directly (not a const reference): 
`gd::String TypeToString(gd::Variable::Type type) { if (type == Type::String) return "string"; else if(...)... }`

(you put this function either in the .cpp file directly, inside the namespace, or as a private, static function of the class, whatever you prefer).

It will work because you don't return a const ref, you're returning a newly constructed gd::String (and if you're concerned about unecessary generation of strings, there will be the "return value optimization" applied anyway by the C++ compiler: https://en.wikipedia.org/wiki/Copy_elision).

If it's just for serialization, let's go with a few if/else :)

Core/GDCore/Project/Variable.h
arthuro555: `enum class` has a problem. The WebIDL binder accepts them, but doesn't store them in `gd.Variable` but in `gd` directly. This is problematic and pollutes `gd`. I'm staying on "classic" `enum` for now.

4ian: Sure no worries then!

arthuro555:
@4ian Do the newIDE tests require a separate import of GDevelop.js? It doesn't seem to recognize the changes I made to gd::Variable.

newIDE/app/src/VariablesList/VariableRow.js
```diff
@@ -59,11 +63,28 @@ const VariableRow = ({
   onSelect,
   origin,
 }: Props) => {
-  const isStructure = variable.getType() === gd.Variable.Structure;
-  const key = '' + depth + name;
+  const type = variable.getType();
+  const isStructural =
```
4ian, line 69: nitpicking: can we find another name for this?
"Structural type" does not mean what we mean here: https://en.wikipedia.org/wiki/Structural_type_system#:~:text=A%20structural%20type%20system%20(or,name%20or%20place%20of%20declaration.

I suggest "hasChildren" or "isCollection", with a preference for collection (in [Java](https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html) for example, it's the name to describe something that contains other things, including arrays, hash maps, sets).

newIDE/app/src/VariablesList/index.js
```diff
@@ -256,24 +274,44 @@ export default class VariablesList extends React.Component<Props, State> {
...
-          variable.getChild(name).setString('');
+        onAddChild={(forceType: ?string) => {
+          if (typeof forceType === 'string') {
+            if (forceType === 'structure') {
+              const name = newNameGenerator('ChildVariable', name =>
+                variable.hasChild(name)
+              );
+              variable.getChild(name).setString('');
+            } else
+              variable.getAtIndex(variable.getChildrenCount()).setString('');
```
4ian, line 299: I wonder if we should have a `push` method in `gd::Variable`? 
So that it's clear what is done, and to encourage using `getAtIndex` with an index that is checked to be inside the bounds (in most languages you'll get an exception/crash/error if you try to access an out of bound. While we don't want a crash in GDevelop, we should still access only in the bounds, and use a separate function to add an item to the array. This can help optimization).

newIDE/app/src/VariablesList/index.js
```diff
@@ -256,24 +274,44 @@ export default class VariablesList extends React.Component<Props, State> {
           if (!parentVariable) {
             variablesContainer.remove(name);
           } else {
-            parentVariable.removeChild(name);
+            if (type === gd.Variable.Structure)
+              parentVariable.removeChild(name);
+            else parentVariable.removeAtIndex(index);
```
4ian, line 285: I encourage you to check also `type === gd.Variable.Array` for extra clarity and to make it searchable in the codebase ("let me find all the special case for an array by searching for "gd.Variable.Array" in the codebase" "oh weird there is very few" "ah actually it's because they made it implicit using a else")

newIDE/app/src/VariablesList/index.js
```diff
@@ -196,15 +211,18 @@ export default class VariablesList extends React.Component<Props, State> {
...
     depth: number,
     index: number,
     parentVariable: ?gdVariable,
     parentOrigin: ?VariableOrigin = null
   ) {
+    console.log(variable);
     const { variablesContainer, commitVariableValueOnBlur } = this.props;
-    const isStructure = variable.getType() === gd.Variable.Structure;
+    const type = variable.getType();
+    const isStructural =
```
4ian, line 229: Same, `isCollection` sounds a good term for this.

newIDE/app/src/VariablesList/index.js
```diff
@@ -26,14 +26,9 @@ const gd: libGDevelop = global.gd;
 
 const SortableVariableRow = SortableElement(VariableRow);
 const SortableAddVariableRow = SortableElement(EditVariableRow);
```
4ian, line 28: 👍

newIDE/app/src/VariablesList/VariableRow.js
```diff
@@ -109,38 +145,68 @@ const VariableRow = ({
...
-          size="small"
-          onClick={onAddChild}
-          style={isStructure ? undefined : styles.fadedButton}
-          tooltip={t`Add child variable`}
-        >
-          <AddCircle />
-        </IconButton>
+      ) : (
+        origin !== 'parent' && (
+          <>
```
4ian, line 170: You can use `ElementWithMenu` to get a menu that will be using material-ui on the web and a native menu on Electron:

https://github.com/4ian/GDevelop/blob/664b18556e9339322f39a071bb3e93642643b6d7/newIDE/app/src/LayersList/LayerRow.js#L70-L100

That should allow you to remove all the complexity related to the popover

4ian:
> Do the newIDE tests require a separate import of GDevelop.js? It doesn't seem to recognize the changes I made to gd::Variable.

Also note that the tests on SemaphoreCI are running with the version of libGD.js from master! (to make the tests faster).
Only Travis-CI and AppVeyor are recompiling libGD.js and can be "fully trusted".

In case of doubt, run the tests locally (`npm test` in newIDE/app).

newIDE/app/src/VariablesList/VariableRow.js
arthuro555: I think the popover looks better and has also the advantage of being compact  

image  

image

newIDE/app/src/VariablesList/index.js
arthuro555: Yes, it sounds like a good idea. It is also necessary if we want to resurrect GDCpp and implement the new types in it.

newIDE/app/src/VariablesList/index.js
arthuro555: That chunk of code is very wrong 😱 I checked the type of variable instead of `parentVariable`.

arthuro555:
> Also note that the tests on SemaphoreCI are running with the version of libGD.js from master! (to make the tests faster).

In that case, wouldn't it be a good idea to skip the test if `git diff master Core` returns some text to avoid false negatives like this?

newIDE/app/src/VariablesList/index.js
4ian: Ah right, good catch!

newIDE/app/src/VariablesList/VariableRow.js
4ian: It might indeed look a bit better to a user familiar with these notions, but in this case I would rather improve the UI building blocks in a separate PR to allow to have context menus composed of icons only, or on Electron to allow setting icons in addition to the labels.

It would also go beyond this very specific menu: if your menu has a need for displaying only icons, then presumably other menus would benefit from it? And so this should be part of a larger change to bring icons only menu to the app.

But I'm a bit worried of accessibility: you have to hover each icon to learn what they do. We're also not a in space constrained situation, so I don't see any reason not to display the labels - they are providing useful information and allow to understand your choice in 2 seconds (the time to read), rather than 5 seconds (the time to try to understand the icon, and then to hover them to have confirmation).

4ian:
> In that case, wouldn't it be a good idea to skip the test if git diff master Core returns some text to avoid false negatives like this?

Yes totally that would be a good idea I think to look at difference between master and the current branch for all *.cpp files, and if there is some assume that tests should be skipped.

arthuro555:
Ok so, there are two things left to do:
1. Fix the flow errors (which is hard as my PC can't handle running flow with my 8GB ram :/)
2. Add support to `Variable(array[0])` syntax, which is hard as it seems there is no way to accept both strings and numbers but only one of those types, meaning accepting a number as index would mean a big refactor of the parser, which is too complex for my brain to handle :/

I am sorry but I don't know how to continue else but by asking for help ☹️

Core/GDCore/Project/Variable.cpp
```diff
@@ -10,35 +10,125 @@
...
+    children.clear();
+
+    // Conversion is only possible for non prmitive types
+    if (type == Type::Array)
+      for (auto i = childrenList.begin(); i != childrenList.end(); ++i)
+        children.insert(
+            std::make_pair(gd::String::From(i - childrenList.begin()), (*i)));
+
+    type = Type::Structure;
+    // A valid structure has at least 1 child
```
4ian, line 72: Why is this the case? We should probably allow empty structure and empty arrays.

Core/GDCore/Project/Variable.cpp
```diff
@@ -66,100 +156,123 @@ const Variable& Variable::GetChild(const gd::String& name) const {
...
   return *children[name];
 }
 
 void Variable::RemoveChild(const gd::String& name) {
-  if (!isStructure) return;
+  if (type != Type::Structure) return;
   children.erase(name);
-  isStructure = !children.empty();
+
+  // If the structure is empty, make it a default empty variable again.
```
4ian, line 168: Same as before, I don't think it's a good idea to change the type and not allow empty structures. It's very surprising and will lead to confusions for sure (and de-optimizations :))

Core/GDCore/Project/Variable.cpp
```diff
@@ -66,100 +156,123 @@ const Variable& Variable::GetChild(const gd::String& name) const {
...
+};
+
+Variable& Variable::PushNew() { return GetAtIndex(GetChildrenCount()); };
+
+void Variable::RemoveAtIndex(const size_t index) {
+  if (index >= childrenList.size()) return;
+  childrenList.erase(childrenList.begin() + index);
+
+  // If the array is now empty, demote back to a primitive.
+  if (childrenList.empty()) {
```
4ian, line 205: Same as before?

Core/GDCore/Project/Variable.cpp
arthuro555: Indeed, I forgot to change that part that was there for inferring types

Core/GDCore/Project/Variable.cpp
```diff
@@ -189,23 +306,35 @@ void Variable::RemoveRecursively(const gd::Variable& variableToRemove) {
...
   }
-  isStructure = !children.empty();
+  for (auto it = childrenList.begin(); it != childrenList.end();) {
+    if (it->get() == &variableToRemove) {
+      childrenList.erase(it);
+    } else {
+      (*it)->RemoveRecursively(variableToRemove);
+      it++;
+    }
+    if ((type == Type::Structure && children.empty()) ||
```
4ian, line 316: Same as before? (flagging so we don't forget it)

Core/GDCore/Project/Variable.h
```diff
@@ -26,18 +27,58 @@ namespace gd {
  */
 class GD_CORE_API Variable {
  public:
+  enum Type {
+    // Primitive types
+    String,
+    Number,
+    Boolean,
+
+    // Structural types
```
4ian, line 36: Let's not use the term "structural type" at all, it's for another thing in computer science.
Let's call then "Collection types".

Core/GDCore/Project/Variable.h
```diff
@@ -26,18 +27,58 @@ namespace gd {
...
+
+    // Structural types
+    Structure,
+    Array
+  };
+
+  /**
+   * \brief Converts a Type to a string.
+   */
+  static gd::String TypeAsString(Type t);
```
4ian, line 44: Could this be private?

Core/GDCore/Project/Variable.h
```diff
@@ -92,22 +158,41 @@ class GD_CORE_API Variable {
...
+  void operator=(const bool val) { SetBool(val); };
+
+  bool operator==(const bool val) const { return GetBool() == val; };
+  bool operator!=(const bool val) const { return GetBool() != val; };
+
   ///@}
 
-  /** \name Structure
-   * Methods used when the variable is considered as a structure.
+  /** \name Structural types
```
4ian, line 170: Collection types

Core/GDCore/Project/Variable.h
```diff
@@ -92,22 +158,41 @@ class GD_CORE_API Variable {
...
+   * Methods used for structural types
```
4ian, line 171: Ditto

GDCpp/GDCpp/Runtime/RuntimeObject.cpp
```diff
@@ -766,6 +766,6 @@ void RuntimeObject::VariableClearChildren(gd::Variable &variable) {
 }
 
 unsigned int RuntimeObject::GetVariableChildCount(gd::Variable &variable) {
-  if (variable.IsStructure() == false) return 0;
+  if (variable.GetType() != gd::Variable::Type::Structure) return 0;
```
4ian, line 769: Should work for arrays too, no?

GDJS/Runtime/events-tools/commontools.js
```diff
@@ -31,6 +32,40 @@ gdjs.evtTools.common.getVariableString = function (variable) {
   return variable.getAsString();
 };
 
+/**
+ * Compares the boolean value of a variable. Equivalent to `variable.getAsBoolean()`.
+ * This shorcut function is needed for events code generation.
```
4ian, line 37: Typo: shortcut (and below too)

GDJS/Runtime/events-tools/commontools.js
```diff
@@ -31,6 +32,40 @@ gdjs.evtTools.common.getVariableString = function (variable) {
...
+
+/**
+ * Set the boolean value of a variable. Equivalent to `variable.setBoolean()`.
+ * This shorcut function is needed for events code generation.
+ *
+ * @param {gdjs.Variable} variable
+ * @param {boolean} bool The new boolean value of the variable.
+ * @private
+ */
+gdjs.evtTools.common.setVariableBoolean = function (variable, bool) {
```
4ian, line 56: Let's rename `bool` to `value`, to avoid any clash with a reserved keyword in the future in case of porting this to another language.

GDJS/Runtime/variable.js
```diff
@@ -20,278 +26,359 @@
...
+};
+
+/**
+ * Pushes a variable onto the array.
+ * @param {gdjs.Variable} variable
  */
-gdjs.Variable.prototype.concatenate = function(str) {
-	this.setString(this.getAsString()+str);
+gdjs.Variable.prototype.push = function (variable) {
+  if (this._type === 'array' && variable instanceof gdjs.Variable)
```
4ian, line 382: Why this `typeof`? Is the static typing not enough?

GDJS/Runtime/variable.js
```diff
@@ -20,278 +26,359 @@
...
+    : [];
+};
+
+/**
+ * Return the length of the
+ * @return {number}
+ */
+gdjs.Variable.prototype.getChildrenCount = function () {
+  if (this.isPrimitive()) return 0;
+  //@ts-ignore This cannot return null as type isn't a primitive.
```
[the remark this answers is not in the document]
arthuro555: Yes I already fixed this locally, I returned null instead of an empty array before but I had to change that to make the for each child event not crash.

GDJS/Runtime/variable.js
arthuro555: Yes, I am concerned about users thinking this is accepting a javascript variable when using it from JS, like the push method of JS arrays.

GDJS/Runtime/websocket-debugger-client/hot-reloader.js
```diff
@@ -435,38 +435,53 @@ gdjs.HotReloader.prototype._hotReloadVariablesContainer = function (
...
+        );
+      else {
+        /**
+         * Arrays cannot be hot reloaded.
+         * As indices can change at runtime, and in the IDE, they can be desychronized.
+         * It will in that case mess up the whole array,
+         * and there is no way to know if that was the case.
+         *
+         * We therefore just replace the old array with the new one.
+         */
```
4ian, line 472: Don't use JSDoc type comments, use a simple multiline (`// blabla`) comments.

GDJS/Runtime/websocket-debugger-client/hot-reloader.js
```diff
@@ -487,39 +502,52 @@ gdjs.HotReloader.prototype._hotReloadStructureVariable = function (
...
+        // Variable is a structure or array (or was converted from a primitive
+        // to one of those).
+        if (newChildVariableData.type === 'structure')
+          this._hotReloadStructureVariable(
+            oldChildVariableData.children,
+            newChildVariableData.children,
+            variable.getChild(newChildVariableData.name)
+          );
+        else {
+          /**
```
4ian, line 534: Ditto

GDevelop.js/Bindings/Bindings.idl
4ian: If it's used then no problem, if it's not used let's remove it :)

4ian: Especially because `VectorVariable` is actually hiding a vector to smart pointers, which is freaking me out - it's so easy to forget a ref, get a copy, never release it, get a memory leak.

newIDE/app/src/VariablesList/VariableRow.js
```diff
@@ -75,71 +84,180 @@ const VariableRow = ({
...
+            <ElementWithMenu
+              shouldOpen={e =>
+                isCollection && !e.shiftKey && e.type !== 'contextmenu'
+                  ? onAddChild() && false
+                  : true
```
4ian, line 177: This is a bit cryptic, especially the side effect in a ternary ;)
Could we separate the behavior of this button when it's a collection and when not? Like:

{ isCollection ? <ElementWithMenu ... /> : <Whatever ... /> }

newIDE/app/src/VariablesList/index.js
```diff
@@ -149,7 +144,11 @@ export default class VariablesList extends React.Component<Props, State> {
   ) => {
     const { variablesContainer, inheritedVariablesContainer } = this.props;
 
-    if (inheritedVariablesContainer && origin === 'parent') {
+    if (
+      inheritedVariablesContainer &&
+      origin === 'parent' &&
+      typeof name !== 'number'
```
4ian, line 150: Generally seeing typeof looks a bit like a code smell. Especially, given that the typing tells us it's a string :) Maybe you need to inspect something else like the type of the variable? Or pass a new argument?

newIDE/app/src/VariablesList/index.js
```diff
@@ -187,9 +196,29 @@ export default class VariablesList extends React.Component<Props, State> {
...
+        origin,
+        /* arrayElement= */ true
+      );
+    });
+  }
+
   _getVariableOrigin = (name: string) => {
     const { variablesContainer, inheritedVariablesContainer } = this.props;
 
+    if (typeof name === 'number') return '';
```
4ian, line 221: Same as before, can we pass an explicit argument rather than relying on variables having multiple types?

newIDE/app/src/VariablesList/index.js
```diff
@@ -256,24 +292,34 @@ export default class VariablesList extends React.Component<Props, State> {
...
 
           this.forceUpdate();
           if (this.props.onSizeUpdated) this.props.onSizeUpdated();
         }}
         onAddChild={() => {
-          const name = newNameGenerator('ChildVariable', name =>
-            variable.hasChild(name)
-          );
-          variable.getChild(name).setString('');
+          // This shouldn't happen, non stuctural types should be converted via forceType
```
4ian, line 305: collection

4ian: Actually this comment is out of date

GDJS/Runtime/variable.js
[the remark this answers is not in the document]
4ian: Sounds dangerous then, it means that the typing is false? You should never ever get null?

GDJS/Runtime/variable.js
4ian: Might be worth outputing a warning in the console then?

arthuro555:
Since I merged master I cannot seem to run the GDJS tests anymore It fails in an unusual way by not telling the file and section it failed in:
image
I tried to comment out every test case that checks for 2, and it still happened.
I `git reset`ed before the merge and now it is working again.
Typescript compilation is also failing at the TileMap files.

arthuro555:
Nevermind, the issues do seem to be related to my local changes and not to the merge. I've added them back and the issue is happening again.

arthuro555:
I narrowed it down, I found out it is the inventory extension serializer that is failing. Chrome sadly closes before I get a chance to click on debug :/

arthuro555:
Found the culprit, this faulty code in the inventory extension that sets a string as a number:
```js
serializedItem.getChild("unlimited").setNumber(item.unlimited ? "true" : "false");
serializedItem.getChild("equipped").setNumber(item.equipped ? "true" : "false");
```

I added a parseFloat to the setNumber so it caused issues. Not sure how it even could work before. I made it use boolean values instead.

4ian:
Great catch! We should fix those to be a `setString` in master, and setBoolean in your branch :) 
TypeScript will hopefully help for this in the future 😇

arthuro555:
> 1. Remove an indexed child from a variable that is not an array?
> 2. Remove a named child from a variable that is not a structure?
> 3. Set a boolean/string/number value to an array variable?
> 4. Set a boolean/string/number value to a structure variable?

1 and 2: Nothing. There is an action per type, as one accepts numbers and the other a string as key.
3 and 4: The primitive value is dismissed, the variable set to the collection of the type it is set to

> What are the actions to push a variable at the end of an array?
image
image

> Do you have one for each primitive type?
The action pushes a variable not a value. I should probably change that, I see how this can be bad.

> Have you added an action to add a child to a structure? (same questions than for arrays).
No, only for arrays. Structure shouldn't be used as list but as a map, and therefore they should not be have items "added" to it but set at a key.


Those are all the actions for variables currently:
image

I think this can be improved by:
1. Add push actions to just push a value instead of a variable reference
2. Rename the structure and array categories inside a "Collections" category 
3. Uplifting the clear action to that new "Collection" category as it works on both


What's your thoughts on this?

arthuro555:
I've merges master and ported the changes to typescript, the tests execute fine but I couldn't tests with my test project yet, I am having issues with port 8000 being taken again :/ 
Btw, I am still awaiting feedback on how to continue this PR 👀

arthuro555:
I most definitely should have asked if it was a good idea before doing the work, but I moved variable evtTools that were in commontools and networktools to their own variabletools namespace, and added export of sourcemaps. Let me know if those changes are problematic and if I need to undo them.

4ian:
> I most definitely should have asked if it was a good idea before doing the work, but I moved variable evtTools that were in commontools and networktools to their own variabletools namespace, and added export of sourcemaps. Let me know if those changes are problematic and if I need to undo them.

While I'm not in theory against this, this PR is becoming huge and the risks have now increased (it's not about adding a new feature and be careful about the naming of the APIs, the design of them, it's now also about ensuring backward compatibility, that this refactoring make sense, etc...), and the probability that I won't be able to review everything and catch early a potential bug us now sadly 100% ;)

I would strongly advise to go step by step, I now it's a bit frustrating but let's do the refactoring in a PR before this one, or keep it for later.

This PR should be about one thing and only one thing: adding safely a set of very well designed, well tested APIs, actions and conditions to support boolean and arrays (one would say that we should have been splitting arrays and booleans in two PRs).

4ian:
To come back to the PR main stuff :) 

> The action pushes a variable not a value. I should probably change that, I see how this can be bad.

That's the kind of decision that is very important and that should be carefully made. What do you think is the best thing? I think it's fine to push a variable, but in this case it should be done by copying the variable value (or in the case of an array/structure, deep copying it). 
There is no concept of references to variables in GDevelop and we certainly don't want to introduce this now.

But because your the owner and author of the PR, I need you to be very proactive about making these decisions explicit and well explained :)

> No, only for arrays. Structure shouldn't be used as list but as a map, and therefore they should not be have items "added" to it but set at a key.

That's fine.

> I think this can be improved by:

> 1. Add push actions to just push a value instead of a variable reference
> 2. Rename the structure and array categories inside a "Collections" category
> 3. Uplifting the clear action to that new "Collection" category as it works on both
> What's your thoughts on this?

I think that's ok yes. I'm not sure about the 1: 
- a single action to push a variable (that will be *copied* entirely, not a reference!) 
- OR multiple actions to push a boolean/string/number
- OR both :) multiple actions to push a primitive (boolean/string/number) and an action to push a variable (which, again, will be copied).

I think ideally we would go for the last one?

arthuro555:
> This PR should be about one thing and only one thing: adding safely a set of very well designed, well tested APIs, actions and conditions to support boolean and arrays (one would say that we should have been splitting arrays and booleans in two PRs).

Yes no problem, imo it should even have been 3 PRs (one being for the strict typing).

> I think ideally we would go for the last one?

I'm fine with this. In the meantime I've been thinking, should I also expose an action to copy a variable to another generally and should I add an action to convert the variable to another type?

4ian:
> should I also expose an action to copy a variable to another generally

Maybe yes, if it's "easy" in the sense that it's the exact same code that is used for pushing into an array and for this action to copy a variable. If that's the case (I think it should be the exact same code), we could even make a PR that is just adding this "CopyVariableIntoAnother" action, and then it would be ready to use for the generic action to push a variable into an array :) 

> should I add an action to convert the variable to another type

I would avoid this, because we don't want to encourage doing this. GDevelop will do this implictly if needed, but let's not bloat the list of actions with these. There is enough work to make these actions simple and accessible enough with array and booleans :)

arthuro555:
I am not sure about the design of one API. 
So I had to make `getChild` return a child at a structure key if the variable is a structure or return the child at an array index, so that we can get a child with the dynamic accessor without having to add the stringOrNumber type to the parser yet. That means that this method is basically working for all collections.
Now what I am unsure about: should I refactor `hasChild`, `removeChild` (removing removeAtIndex) and `clearChildren` to do so  as well instead of them being only for structures? This would add more consistency to the API, but I am scared of accidentally adding a breaking change by modifying them like this.
I also chose to keep setters separated, as they implicitly cast the variable type. But those getters and deleters don't change the type and are generic enough that they could technically work on all collections.

4ian:
> I also chose to keep setters separated, as they implicitly cast the variable type

This is good.

> So I had to make getChild return a child at a structure key if the variable is a structure or return the child at an array index, so that we can get a child with the dynamic accessor without having to add the stringOrNumber type to the parser yet. That means that this method is basically working for all collections.

I think it makes sense. See my explanation below.

> But those getters and deleters don't change the type and are generic enough that they could technically work on all collections.

I would say that as a rule of thumb:

- If the return value type is the same, and all the arguments are the same, as the name make sense for both, it means that your method can be applied to any "collection" (i.e: the one to clear an array/structure) => fine to have the same for both, and surely preferable.
- If there is a different type of argument, you might almost surely want to do two methods. Same if your method implementation consists in `if (is array) { do this; } else if (is structure) { do that; }`, better have two different methods.

To stay on the "safe" side, I think it's preferable to have two different methods (that we can in the future gather into a single one, and keep an alias for backward compatibility) rather than a single method that would risk being split later into two (which would mean we can't preserver backward compatibility, at least not on the JS API).

In other words, it seems that:
- `clearChildren` works well for any collection, so we keep a single one.
- `hasChild(name: string)`: works well for structure, but for an array it would not work right? Or with a number as argument? This seems to indicate that we should not have a `hasChild` for arrays (if you want to check if an index is inside the array, you compare it against the length) (if you want to check if a value is inside the array, that's a more complicated `find` or `has` method, which is outside of this PR).
- `removeChild(name: string)`: same deal, it works with a name as an argument, so should probably not work with an array. Instead we can make a `removeChildAt` (or keep your `removeAtIndex`, it's just a matter of name).
- `getChild(name: string)`: in theory same thing BUT because it's used for accessing arrays when code is generated using the `[ ]` operator, let's make it compatible with arrays. Note that we can have a `getChildAt(index: number)` and rework `getChild(name: string)` so that it does: `if (is array) return getChildAt(parseInt(name, 10) || 0)`

So getChild would be the only exception? And `clearChildren` the only common method?
Let's keep the other separate for now until we figure out later that "actually we can merge something". It will be easier this way for me to review the PR :)

arthuro555:
Ok I think I am pretty much "done" now. What's left to do is:
1. Make a full review of the changes again
2. Someone needs to fix the flow errors for me as I cannot run flow locally
3. If possible, we should let some people test those changes to see if there is any bug/accidental breaking change

Bouh:
I can't get the result of Flow, it's running in circles at Server initialization.
Sometimes it works on a branch sometime no, this executable it is very capricious on windows.
Sorry i can't give you the possible errors.

4ian:
I'll make a new review as soon as possible this week :) Good job so far!

For Flow, I'm sorry again about the bad "developer experience". You actually have not too much errors to fix according to Semaphore CI. I encourage you to "git commit --amend" and "force push" to see the result of your fixes by letting SemaphoreCI run Flow for you.

As they are not "too bad" (you're missing an important for I18nType, some `parseFloat`/cast to boolean maybe when using getValue, and finally a construct `fullWidth ? { width: '100%' } : {}` that would preferably be written as width: fullWidth ? '100%' : undefined), you can also let them and I'll fix them.

GDJS/Runtime/variable.ts
```diff
@@ -211,73 +285,205 @@ namespace gdjs {
...
      * Return the object containing all the children of the variable
      * @return All the children of the variable
      */
-    getAllChildren(): { [key: string]: gdjs.Variable } {
-      return this._children;
+    getAllChildren(): Children {
+      return this._type === 'structure'
+        ? this._children
+        : this._type === 'array'
+        ? ((Object.assign({}, this._childrenList) as unknown) as Children)
```
4ian, line 372: What is this thing doing? Should we rather return an empty object because if you call `getAllChildren` on an array, you're misusing the API anyway?

GDJS/Runtime/variable.ts
```diff
@@ -211,73 +285,205 @@ namespace gdjs {
...
+     */
+    setValue(newValue: string | float | boolean) {
+      if (this._type === 'string') this.setString(newValue as string);
+      else if (this._type === 'number') this.setNumber(newValue as float);
+      else if (this._type === 'boolean') this.setBoolean(newValue as boolean);
+    }
+
+    /**
+     * Gets the primitive value using the getter of the current type.
+     * @param {any} newValue
```
4ian, line 328: Remove this

GDJS/Runtime/variable.ts
```diff
@@ -211,73 +285,205 @@ namespace gdjs {
...
+     * @param {boolean} newValue
+     */
+    setBoolean(newValue: boolean) {
+      this._type = 'boolean';
+      this._bool = !!newValue;
+    }
+
+    /**
+     * Sets the primitive value using the setter of the current type.
+     * @param {any} newValue
```
4ian, line 318: No more types in comments

GDJS/Runtime/variable.ts
```diff
@@ -211,73 +285,205 @@ namespace gdjs {
...
-      this._str = newValue;
-      this._numberDirty = true;
-      this._stringDirty = false;
+      this._type = 'string';
+      this._str = '' + newValue;
+    }
+
+    /**
+     * Get the value of the variable, considered as a boolean
+     * @return {boolean}
```
4ian, line 294: No more types in comments

GDJS/Runtime/variable.ts
```diff
@@ -211,73 +285,205 @@ namespace gdjs {
...
+          return this._str !== '0' && this._str !== '' && this._str !== 'false';
+        else return this.getChildrenCount() !== 0;
+      }
+
+      return this._bool;
+    }
+
+    /**
+     * Change the value of the variable, considered as a boolean
+     * @param {boolean} newValue
```
4ian, line 309: No more types in comments, only descriptions

GDJS/Runtime/variable.ts
```diff
@@ -211,73 +285,205 @@ namespace gdjs {
...
+    /**
+     * Get the value of the variable, considered as a boolean
+     * @return {boolean}
+     */
+    getAsBoolean(): boolean {
+      if (this._type !== 'boolean') {
+        if (this._type === 'number') return this._value !== 0;
+        else if (this._type === 'string')
+          return this._str !== '0' && this._str !== '' && this._str !== 'false';
+        else return this.getChildrenCount() !== 0;
```
4ian, line 301: Careful about this kind of implicit behavior, our choices here are very important and should be seen as such in the code (i.e: with a comment showing that an important choice was made): `// We return blabla because an object or array is considered as blabla. This aligns with blabla and is useful when checking blbla`.

In JavaScript, an object/array is always true, even when it's empty. This is because they represent something empty, but the emptiness of a collection is different from a collection not existing at all (which is represented by `null`/`undefined` in JS).

We don't have `null`/`undefined` in GDevelop - and potentially we might want to never have it (because of philosophical questions about it, and `Optional<>` being an alternative in some programming languages that is safer to use).
In any case, I would still advise to follow the JS way of doing things here. Let it be true whatever the number of children?

GDJS/Runtime/variable.ts
```diff
@@ -136,73 +207,76 @@ namespace gdjs {
...
+    replaceChildren(newChildren: Children) {
+      this._type = 'structure';
       this._children = newChildren;
     }
 
+    /**
+     * Replaces all the children with a new array of children.
+     * @param {gdjs.Variable[]} newChildren The array of new children.
+     */
+    replaceChildrenList(newChildren: gdjs.Variable[]) {
```
4ian, line 235: Same as in the C++ equivalent, I think it might be a good idea to call this `replaceChildrenArray`? Might be a bit more explicit for the reader/user: "ok so they have structure, array... but what is this list thing? Oh it's array again!"

GDJS/Runtime/types/project-data.d.ts
```diff
@@ -25,20 +25,29 @@ declare type ObjectData = {
...
-declare type VariableData = {
-  /** The name of the variable. Used if a child variable. */
-  name: string;
-  /** The value of the variable, either string or number. Leave blank for structures. */
-  value?: string;
+declare type VariableData = Readonly<{
+  /** The name of the variable. Used if a child variable. Leave blank for arrays. */
+  name?: string;
+  /** The value of the variable. Leave blank for structures. */
+  value?: string | float | boolean;
```
4ian, line 45: Nice

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.ts
```diff
@@ -188,6 +188,16 @@ namespace gdjs {
       let currentIndex = 0;
       while (currentIndex < path.length - 1) {
         const key = path[currentIndex];
+        console.log(key);
+        const parsedKey = key.toString().match(/(.*)\((.*)\)/);
```
4ian, line 192: Can you explain when this is used? This seems a new addition/feature, could maybe be in another PR, or is it needed here?

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.ts
```diff
@@ -188,6 +188,16 @@ namespace gdjs {
       let currentIndex = 0;
       while (currentIndex < path.length - 1) {
         const key = path[currentIndex];
+        console.log(key);
```
4ian, line 191: console.log

4ian (commented):
Very nice! I've not reviewed the end of the PR with the React code, but will do later.

Overall this seems to reach a very nice state, very good quality code!
I'll also take another look at the naming of the functions in gdjs.Variable and in the actions/conditions, but overall seems we're reaching a nice to use API/actions/conditions design :)

GDJS/Runtime/variable.ts
arthuro555: The getters implicitly convert the value to the good type, so I think it would be better to stay consistent and do it everywhere. This converts the array to an object by taking the properties (indices) and assigning them on the empty object.

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.ts
arthuro555: Before, it would have done something like `variable._children.childname._children.secondChildName`, but now that wouldn't work as there are also arrays. I added those "function calls" to call `getChild`, that automatically finds out if it is an array or structure and adequately returns from _children or _childrenList.

newIDE/app/src/Debugger/Inspectors/VariablesContainerInspector.js
```diff
@@ -37,22 +44,33 @@ const transform = variablesContainer => {
...
-  // Guess the type of the new value (number or string)
-  if (parseFloat(edit.new_value).toString() === edit.new_value) {
-    path.push('setNumber');
-    onCall(path, [parseFloat(edit.new_value)]);
-  } else {
-    path.push('setString');
-    onCall(path, ['' + edit.new_value]);
+  if (edit.name === 'type') {
+    path.push('castTo');
+    onCall(path, [edit.new_value]);
```
4ian, line 70: Might be a good idea to make a check on the new_value. If it's not a proper type, we should I think `return false`?

newIDE/app/src/Debugger/Inspectors/VariablesContainerInspector.js
```diff
@@ -37,22 +44,33 @@ const transform = variablesContainer => {
...
+
+  let skip = false;
+  for (const variableName of editPath) {
+    // Skip every second key as it is the "value"
+    // key which is displayed only for better user
+    // experience but doesn't really exists
+    skip = !skip;
+    if (skip) continue;
+
+    path.push(`getChild(${variableName})`);
```
4ian, line 58: As you get the `variablesContainer` from the props, would it be possible that instead of using `getChild(...)`, you go through the variablesContainer, at the same time as you "walk" down the variable names, so that each time you can check if:
- it's a structure (use `._children`)
- it's an array (use `.childrenList`)
- it's a primitive (something weird is going on, consider aborting).

This would seem to me safer and more aligned with the other way (`transformVariable`) which is also testing the `_type`.

Also it's 100% sure that at some point someone will use a name of a structure to store arbitrary string (for example to do a count of some words or sentences), which will end up having a pair of parenthesis, which will make the debugger crash because of trying to call it as a function.

Let's give no chance to this, and use the `_type` of the variable that can be found by browsing the container, variable by variable :)

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.ts
4ian: See my other comment. I suggest we make the debugger inspector/transformer smarter, so that it can properly use _children or _childrenList according to the type (we know the type, because we show them to the user).

This will allow to remove this regex, and the crash that would happen when I try to modify the value of a variable called `Something (with parenthesis)` :)

4ian: Note that we could argue that this format (an array of strings, that denote members, until the last one which denote a function to call) is limited (you can't call a function in the middle of it).
This is fine as long as we can do without it. We can always rework it later if we want a new format that would allow each part of the path to be either 1) a property name or 2) a function name with arguments.

It would not be a hard problem, just a matter of reworking what we send from the editor, adapt typing and this code. But it's also a bit more complexity that we don't need, because we can make our variable inspector a bit smarter by checking the type of each variable when constructing the path.

4ian:
Don't lose hope on this, it's incredible progress there and I'm confident we can get this working in the next weeks :)

arthuro555:
I haven't yet lost hope, I've just been busy with school lately :)

4ian:
No worries I was joking :) I made a quick try and fixed some tests :)

Core/GDCore/Project/Variable.cpp
```diff
@@ -66,100 +152,111 @@ const Variable& Variable::GetChild(const gd::String& name) const {
...
-}
+Variable& Variable::GetAtIndex(const size_t index) {
+  type = Type::Array;
+  while (childrenList.size() <= index)
+    childrenList.push_back(std::make_shared<gd::Variable>());
+  return *childrenList[index];
+};
+
+const Variable& Variable::GetAtIndex(const size_t index) const {
+  if (childrenList.size() <= index) return *std::make_shared<gd::Variable>();
```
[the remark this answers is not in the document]
arthuro555: I still have a lot to learn about c++, what a complex language. I thought that as I was initialising on the heap and using a smart pointer, it would be working fine :/
Anyways, a static badVariable is a cleaner and more memory friendly solution either way.

GDJS/Runtime/variable.ts
arthuro555: What should be removed? The function? The type in the parameter? The whole parameter tag?

4ian: The @param is not associated to any param.

4ian: So the whole param tag

4ian:
Thanks! Could you do a quick summary of what is left to do (or is everything pretty much done?).
Any particular part that I should be looking after when reviewing? 

I can fix the Flow errors.
Seems like there is a test failing though, can you take a look?
image

newIDE/app/src/Debugger/Inspectors/VariablesContainerInspector.js
```diff
@@ -37,27 +57,103 @@ const transform = variablesContainer => {
...
+      (variable._type === 'string' && typeof edit.new_value !== 'string') ||
+      (variable._type === 'number' && typeof edit.new_value !== 'number') ||
+      (variable._type === 'boolean' && typeof edit.new_value !== 'boolean')
+    ) {
+      console.error(
+        `Invalid value type: expected value of type '${
+          variable._type
+        }' but received a value of type '${typeof edit.new_value}'.`
+      );
+      return false;
```
4ian, line 138: So an "interesting" thing I found (not entirely related to what you did in this PR) is that if you enter a value that is valid for booleans but don't press the bottom "validate" button in the editor, it will return a error:

image

See if I click on the top "☑️" button (instead of the bottom one):
image

That's "fair enough", because this is a string instead of a boolean. Let me try to now change the variable again. After validating, this time I get another error:
image

because `edit.namespace` is empty 🤷‍♂️ Unsure what happened. But we can do at least one thing: do a parseFloat (for number), a string cast (`'' + edit.new_value`) or a "smart" boolean cast (`typeof edit.new_value === 'boolean' ? edit.new_value : edit.new_value === 'true'`) so that we handle most cases without errors (and so without breaking for some reason the ReactJsonView)

4ian (commented):
I fixed the Flow types :) Also posted a comment about something we could improve for the debugger, and to work around some bug related to ReactJsonView.

4ian:
> 🤔 What I am the least confident with are the changes to the hot reloader. I think some typos might still be spread a bit everywhere, as I always do so many.

I'll take a new look on the code. I've done a few tests and the hot reloading seems to work correctly (changing an array, changing a boolean).

> I think that was because semaphore doesn't build GDCore. I just ran them locally and they did fine.

Right indeed! We should get a clean build from Travis CI if everything is alright (as it's recompiling GDevelop.js).

Apart from that, are all the features related to array and booleans implemented in the current state? Nothing more to add?

arthuro555:
> Apart from that, are all the features related to array and booleans implemented in the current state? Nothing more to add?

I believe it's all here, except if you think one other should be added.

GDCpp/GDCpp/Extensions/Builtin/RuntimeSceneTools.cpp
```diff
@@ -3,13 +3,13 @@
...
  * reserved. This project is released under the MIT License.
  */
-#include "GDCpp/Extensions/Builtin/RuntimeSceneTools.h"
 
 #include <iostream>
 #include <vector>
 
 #include "GDCore/Tools/Log.h"
 #include "GDCpp/Extensions/Builtin/CommonInstructionsTools.h"
+#include "GDCpp/Extensions/Builtin/RuntimeSceneTools.h"
```
4ian, line 12: This will probably be changed back by clang-format :) Check in RuntimeSceneTools, if it does not compile, it means that a header is missing in RuntimeSceneTools.h
The rule of thumb is that headers should compile whatever their order, so they need to have all the needed `#include`

arthuro555:
> That's probably not what you want

Well, I am not sure how else I can do it without making the Getter change the string, which is unintuitive and inconsistent behavior.

4ian:
> Well, I am not sure how else I can do it without making the Getter change the string, which is unintuitive and inconsistent behavior.

Well, what about what you had before? Your code is fine, but in addition to converting to number/boolean and then returning a temporary string (or doing nothing if it's already a string), store it into `str`?
Variables should be used as a single type, but if we use GetString on a number variable, we convert on the fly the number to a string and store it (it's fine, it's what you asked) then return it to you, and if you use GetString on a boolean variable, we do the same: convert to a string then store it and return it to you?

Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
```diff
@@ -496,6 +536,69 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
       .AddParameter("string", _("Variable"))
       .SetHidden();
 
+  obj.AddAction("ObjectVariablePush",
+                _("Push variable on an object array"),
+                _("Appends a variable at the end of a object array variable."),
```
4ian, line 541: I think (here and below), that the correct syntax is "appends [something] to the end of".

Also everywhere: a object => an object

Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
```diff
@@ -496,6 +536,69 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
...
+                "res/actions/var.png")
+      .AddParameter("object", _("Object"))
+      .AddParameter("objectvar", _("Array variable"))
+      .AddParameter("scenevar", _("Variable to push"))
+      .MarkAsAdvanced();
+
+  obj.AddAction("ObjectVariablePushString",
+                _("Push a string on an object array"),
+                _("Appends a string at the end of an object array variable."),
+                _("Append string _PARAM1_ to array variable _PARAM0_"),
```
4ian, line 555: Here and below: these sentences have invalid parameters/miss the object.

Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
```diff
@@ -496,6 +536,69 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
...
+                "res/actions/var24.png",
+                "res/actions/var.png")
+      .AddParameter("object", _("Object"))
+      .AddParameter("objectvar", _("Array variable"))
+      .AddParameter("trueorfalse", _("Boolean to push"))
+      .MarkAsAdvanced();
+
+  obj.AddAction(
+         "ObjectVariableRemoveAt",
+         _("Remove index of object array"),
```
4ian, line 590: `Remove variable from an object array (by index)`?

You don't really remove an "index", you use the index as a way to locate what to remove.

Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
```diff
@@ -496,6 +536,69 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
...
+         _("Removes a variable at an index of an object array variable."),
```
4ian, line 591: at the specified index

Core/GDCore/Extensions/Builtin/VariablesExtension.cpp
```diff
@@ -143,6 +169,30 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsVariablesExtension(
...
+          "res/conditions/var24.png",
+          "res/conditions/var.png")
+      .AddParameter("scenevar", _("Variable"))
+      .AddParameter("trueorfalse", _("New Value:"));
+
+  extension
+      .AddAction("ToggleSceneVariableAsBoolean",
+                 _("Toggle boolean value of a scene variable"),
+                 _("Toggle the boolean value of a scene variable.\n"
+                   "If it was true, it will become false, and if it was false "
```
4ian, line 188: nitpicking: I recommend putting "If it was true blabla" into its own `_("...")`, so that for translators they have to translate it only once.
Put "\n" outside of a translation marker.

Core/GDCore/Extensions/Builtin/VariablesExtension.cpp
```diff
@@ -207,10 +281,132 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsVariablesExtension(
...
-                 _("Variables/Global variables/Structure"),
+                 _("Variables/Global variables/Collections"),
+                 "res/actions/var24.png",
+                 "res/actions/var.png")
+      .AddParameter("globalvar", _("Variable"))
+      .MarkAsAdvanced();
+
+  extension
+      .AddAction("SceneVariablePush",
+                 _("Push variable on a scene array"),
```
4ian, line 292: Could we use `Append XXX to a YYY array` everywhere?
While it's named `push` in javascript, it seems the MDN docs never use "push" as a verb, always explaining things in terms of "append" or "add". 
Also make it easier for people to search the actions if they have the same words as the sentence.

Core/GDCore/Project/Variable.cpp
```diff
@@ -10,35 +10,120 @@
...
+  if (newType == Type::Number)
+    SetValue(GetValue());
+  else if (newType == Type::String)
+    SetString(GetString());
+  else if (newType == Type::Boolean)
+    SetBool(GetBool());
+  else if (newType == Type::Structure) {
+    children.clear();
+
+    // Conversion is only possible for non prmitive types
```
4ian, line 67: nitpicking: primitive

Core/GDCore/Project/Variable.cpp
```diff
@@ -10,35 +10,120 @@
...
+    if (type == Type::Array)
+      for (auto i = childrenArray.begin(); i != childrenArray.end(); ++i)
+        children.insert(
+            std::make_pair(gd::String::From(i - childrenArray.begin()), (*i)));
+
+    type = Type::Structure;
+  } else if (newType == Type::Array) {
+    childrenArray.clear();
+
+    // Conversion is only possible for non prmitive types
```
4ian, line 77: primitive

Core/GDCore/Project/Variable.cpp
```diff
@@ -10,35 +10,120 @@
...
+    children.clear();
+
+    // Conversion is only possible for non prmitive types
+    if (type == Type::Array)
+      for (auto i = childrenArray.begin(); i != childrenArray.end(); ++i)
+        children.insert(
+            std::make_pair(gd::String::From(i - childrenArray.begin()), (*i)));
+
+    type = Type::Structure;
+  } else if (newType == Type::Array) {
```
4ian, line 74: Do you think that because we've now casted to a structure, we can clear childrenArray?

Core/GDCore/Project/Variable.cpp
```diff
@@ -10,35 +10,120 @@
...
+    childrenArray.clear();
+
+    // Conversion is only possible for non prmitive types
+    if (type == Type::Structure)
+      for (auto i = children.begin(); i != children.end(); ++i)
+        childrenArray.push_back((*i).second);
+
+    type = Type::Array;
```
4ian, line 82: Same, do you think we can clear children?

Core/GDCore/Project/Variable.cpp
```diff
@@ -10,35 +10,120 @@
...
+    return str.To<double>();
+  } else if (type == Type::Boolean) {
+    return boolVal ? 1.0 : 0.0;
+  }
+
+  // It isn't possible to convert a non-primitive type to a number
+  return 0.0;
+}
+
+gd::String Variable::GetString() const {
```
4ian, line 99: So about the reference, would it change something if we update `str` and return `str`? 
It would *not* change the type (if it's a number variable, it stays a number variable), it's just safer to use and do not make a copy of the string everytime you want to read the string. It only updates the `str` member, which is anyway an implementation detail.

GDJS/Runtime/variable.ts
```diff
@@ -211,73 +284,197 @@ namespace gdjs {
...
+
+    /**
+     * Get the value of the variable, considered as a boolean
+     * @return The boolean value of the variable.
+     */
+    getAsBoolean(): boolean {
+      if (this._type !== 'boolean') {
+        if (this._type === 'number') return this._value !== 0;
+        else if (this._type === 'string')
+          return this._str !== '0' && this._str !== '' && this._str !== 'false';
```
4ian, line 299: Do you think we should just do `this._str !== ''`, like in JavaScript? 

I know it's "convenient" to cast `"false"` to `false`, but I wonder if it's not more surprising than anything else. No other language is really doing this?

GDJS/Runtime/variable.ts
```diff
@@ -211,73 +284,197 @@ namespace gdjs {
...
+    /**
+     * Pushes a copy of a variable onto the array.
+     */
+    pushVariableCopy(variable: gdjs.Variable) {
+      this.castTo('array');
+      this._childrenArray.push(variable.clone());
+    }
+
+    /**
+     * Pushes a value onto the array.
```
4ian, line 468: into (https://idioms.thefreedictionary.com/push+into)

GDJS/Runtime/variable.ts
```diff
@@ -211,73 +284,197 @@ namespace gdjs {
...
+
+    /**
+     * Removes a variable at a given index of the array.
+     */
+    removeAtIndex(index: integer) {
+      if (this._type === 'array') this._childrenArray.splice(index, 1);
+    }
+
+    /**
+     * Pushes a copy of a variable onto the array.
```
4ian, line 460: into (https://idioms.thefreedictionary.com/push+into)

4ian:
Nice! We're getting close 🤓

Core/GDCore/Project/Variable.cpp
arthuro555: Ah yes 🤦‍♂️. For some reason my mind interpreted the code differently and I thought this would cause a bug.

arthuro555:
The travis build is now finally passing!
image

4ian:
Great 👍 We can ignore semaphoreci, so with AppVeyor and Travis passing this is in very good shape.

I think from the last review pass we're alright on most things, but this is a quite sensible set of features, so I'll make another pass and will try again the branch. 
Would be good to give it a go to a few testers so that we can identify things that would be confusing (knowing that it's just for test! If we do changes, their existing usage of boolean/arrays could be broken. So just to get an idea and see if people understand the actions :))

Oxey405:
69 commits !?!?

arthuro555:
> Would be good to give it a go to a few testers

When I want people to test the changes, what is the recommended method? Should I put the link on the trello card, in this issue, make a forum post (if yes should I pin the post?), should we have a special place for that purpose (create a pinned forum post for all change testing on the forum for example)...?

4ian:
Sorry for not checking this earlier! Yes a pinned message in the forum is a great way to go (and you can advertise on the forum too). Seems the simplest and gives a way for users to post feedback directly. You can put a link to the forum thread on the Trello card too to see if there are interested people.

4ian:
Let me know when you've posted something, we'll monitor the feedbacks of users for a few days. I'm testing now, everything looks alright :)

GDJS/GDJS/Extensions/Builtin/CommonInstructionsExtension.cpp
```diff
@@ -465,27 +465,41 @@ CommonInstructionsExtension::CommonInstructionsExtension() {
...
+            "      ? $ITERABLE_REFERENCE.getAllChildren()\n"
+            "      : $ITERABLE_REFERENCE.getType() === \"array\"\n"
+            "        ? $ITERABLE_REFERENCE.getAllChildrenArray()\n"
+            "        : []\n"
             ") {\n";
         
         // If variables are defined, store the value in them
         if(keyIteratorExists) outputCode += 
-            "    $KEY_ITERATOR_REFERENCE.setString($ITERATOR_KEY);\n";
+            "    $ITERABLE_REFERENCE.getType() === \"structure\"\n"
```
4ian, line 485: Nitpicking: this could be a if that is generated right? No need for a ternary, or even badVariable here?

GDJS/GDJS/Extensions/Builtin/CommonInstructionsExtension.cpp
```diff
@@ -499,10 +513,14 @@ CommonInstructionsExtension::CommonInstructionsExtension() {
           outputCode += "} //Subevents end.\n";
         }
         outputCode += "}\n";
+        // End of normal code generation
```
4ian, line 516: Nitpicking: End of standard event code generation.

newIDE/app/src/UI/Text.js
```diff
@@ -43,21 +44,20 @@ export default ({
   noShrink,
   noMargin,
   displayInlineAsSpan,
-}: Props) => {
-  return (
-    <Typography
-      variant={size === 'title' ? 'h6' : size === 'body2' ? 'body2' : 'body1'}
-      component={displayInlineAsSpan ? 'span' : undefined}
```
4ian, line 50: This line is missing

arthuro555:
Ok I've made the post: https://forum.gdevelop-app.com/t/we-need-testers-for-new-variable-types/30698

4ian:
I'm working on adapting the expression parser to support a `number|string` type. The actual type is inferred from the first number, string literal or function call made. This should allow to support array bracket accessor using numbers: `MyArray[3]`, `MyArray[Random(4)]`...

arthuro555:
It seems to work very well, thanks for taking your time to do that. The only small issue I noticed is that the expressions autocomplete shows in all fields only number expressions, never any string expressions (including in string fields).

4ian:
> The only small issue I noticed is that the expressions autocomplete shows in all fields only number expressions, never any string expressions (including in string fields).

Pretty annoying issue, can you show a screenshot of it happening?
I've tried in master, but it seems to work properly (search for ToString or ToNumber for example, properly completed only according to the field type or the argument type).

4ian:
@arthuro555 Would you be able to try again to see if it's still happening? Maybe merge master again to see if it happens after recompiling GDevelop.js?

arthuro555:
Indeed merging again seems to have fixed it. Though, I don't know if that is intended but if I write `Variable(myVar[` autocomplete will only suggest number expressions, no string expressions.

4ian:
Good catch, I think I've not added support for `number|string` to `getAutocompletionsFromDescriptions`/`enumerateFreeExpressions`, so only numbers are returned. No big deal, should be easy to fix. If the rest is working well, that's the most important.

4ian:
Seems AppVeyor is failing https://ci.appveyor.com/project/4ian/gdevelop/builds/37926491
Weird... Maybe I should also update AppVeyor to compile the wasm version of GDevelop.js by the way.

GDevelop.js/__tests__/Core.js
```diff
@@ -758,41 +768,84 @@ describe('libGD.js', function () {
...
+
       parentVariable.removeRecursively(child2);
       expect(parentVariable.getChildrenCount()).toBe(1);
       expect(parentVariable.hasChild('Child1')).toBe(true);
       expect(parentVariable.hasChild('Child2')).toBe(false);
 
+      grandChild1.delete();
+      grandChild2.delete();
+      child1.delete()
+      child2.delete()
```
4ian, line 848: We should not delete grandChild1/2 and child2, because they were already destroyed by removeRecursively. This means that here you would delete an already deleted object (grandChild1/2 and child2 are just references to a variable, that was removed - and remember that in libGD.js things are not garbage collected like in JS, the lifetime of objects depends on where they are stored). 
I'm removing these and app veyor tests should pass :) 
Glad that app veyor found this! Unsure why this did not crash on Travis. The beauty of "undefined behavior"!

4ian:
Let's wait for the tests to be green, then I'll do another pass and I hope we can merge this to make a new testing release probably this week-end. This is a huge internal change so I want to be cautious. We can then merge @Bouh BitmapText PR which is another pretty big feature :)

4ian:
I've made a few tweaks and updates, sounds I think almost ready to go.
I think we should have an expression (and condition) to get the number of children/items. I'm just not sure how to name it:
`VariableLength(myVariable)` (and `Object.VariableLength(myVariable)`/`GlobalVariableLength(myVariable)`)? Or `VariableSize(myVariable)` (and `Object.VariableSize(myVariable)`/`GlobalVariableSize(myVariable)`)?
Or `VariableChildrenCount(myVariable)` (and `Object.VariableChildrenCount(myVariable)`/`GlobalVariableChildrenCount(myVariable)`)?

Ideally something along length/size seems the best.
`length` is I think intuitive, but could also be confused with the length of a string variable (though we can make it work for it too!): 
* it's used in JS (`array.length`)
* it's used in Python (`len(array)`)
* in C++ length() is for a std::string (and `size()` is also available). 

`size` is maybe a bit more generic for a collection:
* it's used in Java: https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html#size--
* it's used in C++ (std::vector::size)

Seems like there is [no convention](https://softwareengineering.stackexchange.com/questions/303909/what-is-the-difference-between-size-and-length#:~:text=Generally%2C%20%22size%22%20stood%20for,%2C%20of%20course%2C%20very%20fixed.), though "length" implies a "line", so it's less adapted to something like a structure.

I might play it safe and call it:

`VariableArrayLength(myVariable)` (and `Object.VariableArrayLength(myVariable)`/`GlobalVariableArrayLength(myVariable)`.
And we can always introduce a more generic VariableLength/VariableSize that works for other types later :)

arthuro555:
We already have that, and it is called `VariableChildCount`

4ian:
Ah I totally forgot that! All good then 👍 Thanks :)

4ian:
I think we're good to go with this! I made some long due rewording of object variable actions and a few other stuff.
I'm letting the tests pass one last time and will merge later this afternoon! 

Then we can plan a new release (surely a "testing" one without auto-update first) with this.

Great job on this! Array was one of the most asked feature from the recent user survey (remind me btw to share some highlights about it on the Discord later this week :)). It will be very welcome. And with booleans it's also a long due feature that will make it cleaner to handle states. Happy to see that variables are reaching "parity" with JSON and JS variables.

4ian:
Merged! ✨

PR #2140, opened 2020-12-16
Rework the extension search to be more powerful and similar to the asset store

4ian:
* Use a virtualised list to scale to a large number of extensions easily.
* Also improve the display of the icons and the descriptions of the extensions.
* After installing a behavior, automatically navigate back to the list of behaviors.

image


Bouh:
There is hashtags in the list but nothing when we check the filter, eg: Cheats, Easter Egg, Effect
With _Cheats_
There is also twice **Camera**
**Bullet** and **Bullets** are similar, i'll make it consistent in the extension repo.
image

Bouh:
There is double in this list [extension registry file](https://github.com/4ian/GDevelop-extensions/blob/master/extensions-registry.json), with and without capital letter, that's why there is doubles.

Bouh:
In an object when you go for search a behavior "gamepad" there is nothing
image


But in the search new extension from Project manager> function/behaviors > Search for new extensions it works.
image

4ian:
Yep, tags for extensions without behaviors should not be shown (unsure how easy this is to do). A button could be added too (but a smaller one then).

PR #2144, opened 2020-12-20
Added OS generated files to .gitignore

TheGemDev:
added os generated files to .gitignore this should prevent adding these files in a Pull Request
# OS generated files #

.DS_Store?
._*
.Spotlight-V100
.Trashes
ehthumbs.db
Thumbs.db

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,12 @@
 .idea
 .vscode/ipch
 /newIDE/app/src/UI/Theme/**/*ThemeVariables.*
+
+# OS generated files #
+.DS_Store
+.DS_Store?
+._*
+.Spotlight-V100
+.Trashes
+ehthumbs.db
+Thumbs.db
--- a/.gitignore
+++ b/.gitignore
@@ -62,9 +62,6 @@
 .vscode/ipch
 /newIDE/app/src/UI/Theme/**/*ThemeVariables.*
 .DS_Store
-.DS_Store?
-._*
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -61,3 +61,10 @@
 .idea
 .vscode/ipch
 /newIDE/app/src/UI/Theme/**/*ThemeVariables.*
+.DS_Store
+.DS_Store?
```
4ian, line 65: Is this one really necessary? Let's keep just the first one.

.gitignore
```diff
@@ -61,3 +61,10 @@
...
+._*
+.Spotlight-V100
```
4ian, line 67: I would rather avoid these things that could encompass too much files.

.gitignore
```diff
@@ -61,3 +61,10 @@
...
+.Trashes
+ehthumbs.db
```
4ian, line 69: For which OS is this?

.gitignore
TheGemDev: Okay

.gitignore
TheGemDev: .Spotlight-V100
.Trashes
Are generated by apple macintosh, 
._* would be removed also ehthumbs.db

.gitignore
TheGemDev: Okay

4ian:
Thanks! 👍

Issue #2130, opened 2020-12-10
Provide the ability to return the "zoom" property of a camera


tristanbob:
## Description

Currently, if you want to know the zoom level of a camera, you have to calculate it manually.  Something like this:

CurrentZoom = SceneWindowWidth() / CameraWidth("",0)

## Solution suggested

Provide the ability to return the zoom property of a camera, just like we do for: [CameraWidth, CameraHeight, CameraX, CameraY, CameraRotation]

![Screenshot 2020-12-10 115012](https://user-images.githubusercontent.com/8879811/101816172-ea6b0e00-3add-11eb-9dd6-e38efd1534cd.jpg)

4ian:
Definitely, should not be too hard to add!

rs4231199:
@4ian I'm looking to start contributing to GDevelop. Can I try fixing this issue?

4ian:
Of course! Feel free to work on this and open a PR if you have a solution :)

dkapur17:
@4ian, I'm looking to get started contributing to this repo. I've spent a few days going through the codebase and have added this feature into my local branch. I was wondering if for the PR, should I make a new branch for this feature and then make the PR or just make the PR to the master branch?

4ian:
Usually on your own repo you do it in a branch (so you can switch between features easily). Then you open a PR that will be done on this repository master branch.

dkapur17:
@4ian I did work on it in a new branch and it seemed to be working fine in the IDE as well so I merged it to my master branch before pushing. I hope that isn't a problem.

4ian:
Closing as added for next version. Thanks @dkapur17!

PR #2145, opened 2020-12-20
Added zoom expression

dkapur17:
Added expression for getting the zoom of the camera directly as mentioned in Issue #2130 .

Image


4ian:
Looks good :) I've fixed the default value of the camera which should be 0 (the first, and only camera supported, is camera #0).
Apart from that, I think it's all good, code is very clean! Merging now - thanks for adding this 👍

dkapur17:
@4ian, I actually went out of my way to make its default value 1 😅 . I think I misunderstood the meaning of the default value. My bad.

PR #2147, opened 2020-12-21
Tiled pixi tilemap

blurymind:
This PR implements a tilemap extension via pixi-tilemap. It also adds a parser for the tiled json format.
https://github.com/4ian/GDevelop/issues/503

image

- [x] add support for varied length animation to pixi-tilemap https://github.com/pixijs/pixi-tilemap/pull/77
- [x] parse tiled files and feed them to pixi-tilemap
- [x] implement a tilemap renderer in the IDE
- [x] port progress so far to pixi5
- [x] properly implement setters/getters on the runtime (need testing)
- [x] implement tiled rendering on the runtime
- [x] convert pixi-tilemap to a module which gdevelop pixi5 can consume (UMD)
https://github.com/pixijs/pixi-tilemap/issues/92
- [x] add new icon
- [x] refactor helper class to be reusable
- [x] add support for animated tiles
- [x] fix customSize not implemented in runtime
- [x] fix there is no way to set image resource. See https://github.com/4ian/GDevelop/blob/2ca593ba2b980b523aa963a5acce7f6dd0153a7c/newIDE/app/src/EventsSheet/ParameterRenderingService.js
- [x] remove redundant this._pixiObject attached variables - move needed to this._object instead
- [x] add condition to check the json file that is selected
- [x] get all event sheet methods to work properly - test them accordingly - including swapping between tilemaps during runtime
- [x] create a demo project
- [x] add protection against wrong size atlas and other bad data crashing the editor or runtime
- [x] add support for spacing in atlases (pixel gaps between tiles)
- [x] add support for compressed layer.data (using pako)
- [x] fix runtime inconsistencies, test more
- [x] add support for external tileset file

write documentation (After merge)

Edit: the pixi5 regression has now been fixed and this PR is back on track!

This is a reopen of https://github.com/4ian/GDevelop/pull/1901 with the hope that github will clear the diff that was contimated when merging dev into it


4ian:
Great! So would need a bit of work on the documentation and a last check on the code from me and should be good to go :)

4ian (commented):
I've not forgotten this :) I fixed the issue with the separate tilesets, I'll make a few last checks and documentation and will merge this :)

4ian:
I've written the documentation on: http://wiki.compilgames.net/doku.php/gdevelop5/objects/tilemap :)
I've done various changes in the tiny `pixi-tilemap-helper` library so that it has TypeScript annotations and avoid using the term "tileset" for something which is not a tileset, but a merge of a Tiled "map" and "tileset". 

Thanks @blurymind for creating this and for your perseverance pushing this for the last months! And special thanks too to @ivanpopelyshev for creating & improving Pixi Tilemap and for helping us using it here :) 

I'm very excited seeing this Tilemap object in GDevelop. It will be very useful for games like RPGs, and for making large levels. If in the future we add a simple builtin editor directly in the scene editor, and maybe support for LDtk, as well as collision handling directly using Tiled/LDtk data, this could become a very useful objects for lots of games!

blurymind:
Wow, fantastic work on the documentation! You added a complete tutorial on how to export the file in tiled and set it up in gdevelop.

Thank you for the improvements you made to this, also for the consistently good feedback to the pr.
Sorry I couldn't work on it as much in the last month due to travelling. It will be exciting to see how the community uses this and also as future steps improve it.

I want to upgrade yarn next (should fix the bug reported on it, also add loads of new features) when I get back. Then start planning on how we can move this further.

I wonder if we should try to write our own react component for editing tiled tilemaps or push for using ldtk, which is already incredibly powerful.
It would probably be better to overcome the limitations on tilesets first. I wonder if an extension should be allowed to add and remove resources on the fly via the resources manager. So for example parsing the tileset json could automatically add image resources via the relative paths stored in there. That way hundred of images can be automatically added by the extension somehow? A tileset json may have hundreds on image paths to atlases, especially in ldtk

4ian:
> I wonder if we should try to write our own react component for editing tiled tilemaps or push for using ldtk, which is already incredibly powerful.

While LDtk looks fantastic, I think we should have basic editing capabilities in GDevelop itself - it's a bit sad that it's the first object that you can't really author on your own in GDevelop (well, there are videos, but that's an exception :p).

Note that LDtk actually has a Tiled export now :) It still need to be opened with Tiled to save the map (and tileset) again to .json:
image

Seems like we're missing an important feature though: flipping a tile horizontally or vertically. We should investigate how we can do that with pixi-tilemap :)

Silver-Streak:
> > I wonder if we should try to write our own react component for editing tiled tilemaps or push for using ldtk, which is already incredibly powerful.
> 
> While LDtk looks fantastic, I think we should have basic editing capabilities in GDevelop itself - it's a bit sad that it's the first object that you can't really author on your own in GDevelop (well, there are videos, but that's an exception :p).
> 
> Note that LDtk actually has a Tiled export now :) It still need to be opened with Tiled to save the map (and tileset) again to .json:
> image
> 
> Seems like we're missing an important feature though: flipping a tile horizontally or vertically. We should investigate how we can do that with pixi-tilemap :)

Just to check on this, since LDtk is Javascript, couldn't it be integrated like JFXR and Piskel?

I vaguely remember this being discussed before as a long term goal. (And the LDtk dev even has an idea around integration support with GD5 from Blurymind's discussion)

4ian:
It definitely could, it's just that it should be carefully done to ensure it blends nicely with the existing editor. In an ideal world, you could double click a tilemap on the screen and start drawing with all the nice tooling made in LDtk (without even worrying about saving it as a file, etc...). Without going to this extent, it would be nice to seamlessly open LDtk when needed, do the changes (without, same thing, having all the tooling related to saving a file, etc..), close and see the result. Would also need when you create a new tilemap object to set the atlas image to the one chosen in LDtk, etc...

blurymind:
We could use electron events to detect when the user closes its window to save to the json file.

Its still good to have the map and tileset as a separate file. Good why? To be editable externally and easier to manage in a team work situation

Issue #1999, opened 2020-09-26
Graphical glitch when resizing the objects editor


TheGemDev:
## Describe the bug
please follow the steps to reproduce this :)

## To Reproduce
1. Go to the objects editor and make sure you have enough objects to make you scroll. 
2. then open the layers editor.
3. scroll down in the objects editor till you can see the add new object button.
4. then close the layers editor after this you should see some blank spaces above. (check the image below) this would only go away once you close and reopen the objects editor.

## Other details
![ddddddddd](https://user-images.githubusercontent.com/53819287/94337274-362f1080-ffe1-11ea-80da-720fdfa5a5ef.PNG)

Bouh:
Confirmed
![fail_list_refresh](https://user-images.githubusercontent.com/1670670/94374851-80aeab00-010f-11eb-87c4-c04304533781.gif)

VolDemorT-acer:
The display of the objects is handled by a List component of third party library "react-virtualized" ,now I am just confused if we can fix this .I will try with a dummy list using "react-virtaulised" package in codesandox or locally ,to check if the issuse still persists when we change the height of list.If the issue still remains ,there's nothing much to be done .

ps : I am new to opensource;And this is repo is huge so I may have missed something

HarsimranVirk:
It could be a react-virtualized issue, [https://github.com/bvaughn/react-virtualized/issues/1533](), seems pretty similar to what we have at GDevelop 🤔.

dkapur17:
@TheGemDev @4ian I have come up with a pretty straightforward solution for this, but uses `document.querySelector` to fetch some information from the DOM. If that's not an issue then I can open a Pull Request for this.

dkapur17:
@TheGemDev @4ian I have come up with a work around for this. Though it uses `react-dom` to fetch some information directly from the DOM. If that isn't an issue then I'll open a PR with the changes.

4ian:
Open the PR and we'll see directly in the code :)

85ayush:
I would love to work on this issue, I am new to open source contribution, can you please guide me?

Silver-Streak:
This appears to have been resolved quite some time ago, and i can no longer reproduce it with the steps above. Closing out, but feel free to reopen if someone is still experiencing this on 5.1.157 or newer.

PR #2148, opened 2020-12-21
Workaround for whitespace in objects panel

dkapur17:
Used a bit of DOM manipulation to solve this. It basically compares the position of the scrollbar of the list with the position of the first element in that list and uses that to decided whether or not it should fix the scroll. Here is what the end result looks like:

ScrollIssue-2020-12-22_04 43 22

I can't seem to find a solution for the flickering as it's most likely from the `react-virtualized` library :(

Also, I tried using `List.forceUpdateGrid()` instead of `List.scrollToPosition()`, but it didn't seem to work. In my opinion that is the better way if someone can get it to work with that (unless I'm misunderstanding what it actually does).

Should fix Issue #1999 :)

### DIFF
--- a/newIDE/app/src/UI/SortableVirtualizedItemList/index.js
+++ b/newIDE/app/src/UI/SortableVirtualizedItemList/index.js
@@ -1,5 +1,6 @@
 // @flow
 import * as React from 'react';
+import { findDOMNode } from 'react-dom';
 import { List } from 'react-virtualized';
 import ItemRow from './ItemRow';
 import { AddListItem } from '../ListCommonItem';
@@ -84,6 +85,14 @@ export default class SortableVirtualizedItemList<Item> extends React.Component<
         connectIconDragSource={connectIconDragSource || null}
       />
     );
+  }
+
+  // Workaround for whitespace issue cause by react-visualized
+  componentDidUpdate(){
+    const listNode = findDOMNode(this._list);
+    if(listNode.hasChildNodes() && listNode.firstChild.hasChildNodes())
+      if(!listNode.scrollTop && !!listNode.firstChild.firstChild.offsetTop)
+        this._list.scrollToPosition(listNode.firstChild.firstChild.offsetTop);
   }
 
   render() {
--- a/newIDE/app/src/ObjectGroupsList/index.js
+++ b/newIDE/app/src/ObjectGroupsList/index.js
@@ -2,6 +2,7 @@
 import { Trans } from '@lingui/macro';
 
 import React, { Component } from 'react';
+import {findDOMNode} from 'react-dom';
 import { AutoSizer, List } from 'react-virtualized';
 import Background from '../UI/Background';
 import SearchBar from '../UI/SearchBar';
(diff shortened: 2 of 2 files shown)


dkapur17:
@4ian I have gone through the results for the unit tests and I see that the issue is because of Flow flagging the DOM reference as potentially `null`, and `null` objects don't have the properties that I have accessed (like `firstChild`). I have even added the condition that this logic should run only if the reference is not `null`, but I guess flow just does simple type checking without looking at the conditions. Here is the updated logic:

```JavaScript
componentDidUpdate(){
    const listNode= findDOMNode(this._list);
    if(!!listNode && listNode.hasChildNodes() && listNode.firstChild.hasChildNodes())
      if(!listNode.scrollTop && !!listNode.firstChild.firstChild.offsetTop)
        this._list.scrollToPosition(listNode.firstChild.firstChild.offsetTop);
  }
```
The first condition in the `if` only lets the logic proceed if the DOM reference is not `null`.

I'm not too familiar with flow so I don't exactly know how to fix this issue. Any advice?

newIDE/app/src/UI/SortableVirtualizedItemList/index.js
```diff
@@ -86,6 +87,14 @@ export default class SortableVirtualizedItemList<Item> extends React.Component<
     );
   }
 
+  // Workaround for whitespace issue cause by react-visualized
```
4ian, line 90: I'm not sure I understand exactly why this works.  Is there an issue opened on `react-virtualized` GitHub repository that we can link? (and if so, maybe just updating `react-virtualized` would work?

4ian:
>  I have even added the condition that this logic should run only if the reference is not null, but I guess flow just does simple type checking without looking at the conditions. 

You should protect all variables that you are using and that can be null by storing them into intermediate variables and check if they are not null before continuing. This is needed because you're "digging" into the DOM.

Overall I'd like to better understand this workaround. I'm afraid this could create more issues down the road if we later update react-virtualized, because we interfer with the scrolling. Would be good to get an exact understanding of why the problem is there and what to do.

dkapur17:
So as for why it works, the way react-virtualized creates the DOM elements for a list is:

`List_Element` -> `Scrollabale_Container` -> `Rows`

The `List_Element` is what has the actual scrollbar on it. It has one child, which is the `Scrollable_Container`, that contains all the rows as its children. The `Scrollable_Container` physically moves up and down when we scroll in the Objects panel. Because of this, the `offsetTop` of each of its children remains fixed.

I spent a decent amount of time looking at the console and trying to find trigger points. When we reproduce the bug, the `componentDidUpdate` method is fired, so I decided that that is the appropriate lifecycle method for this fix.

Now here is the logic. When the scroll of `List_Element` is 0, i.e. we are looking at the topmost row, then the `Scrollable_Container`'s first child should be the first row in the list. When this bug is in action, what we get is that the scroll bar is still at 0, but the first child isn't the first row, since we have a lot of whitespace on top of the first child of the container.

I just check that if the scroll is 0 and the container's first child is not the first row. If this condition holds true, then I determine that the bug has occured. Also, I noticed that making the list scroll updates the rows that it is rendering, so I make it scroll to the first row that is being rendered, and in the process, make everything above it also visible, removing the whitespace.

I believe the problem occurs because `react-virtualized` keeps adding and removing list items from the DOM as per its calculations. Whenever there is some kind of change, it recalculates which rows are to be shown and shows them. My understanding is that for some reason the sudden resizing of the objects panel doesn't trigger the recalculation. I did spend a decent amount of time on their Docs, but I couldn't find a solution from there either.

That being said, even I am not a big fan of my solution, mainly because it is directly accessing the DOM. However, it seems to be working well enough. Maybe before merging/closing this PR, try adding those 4 lines to your local branch and see how they pan out? If you do find some corner cases which causes the logic to mess up, you can let me know. 

Its also possible that we are ovecomplicating this whole thing and simply updating `react-virtualized` could fix this issue :sweat_smile:

dkapur17:
@4ian, look at this solution again, it doesn't gel very well with the code style of the rest of the project, and as you said, could cause some issues in the future. So I've decided to close this PR.

4ian:
Sorry for the delay, I also feel like it's a workaround for an issue that is not a priority - it would be great to fix it but I don't think it's worth adding a hack like this :)

PR #2153, opened 2020-12-26
Add a player assistance to bypass the obstacles with the top-down behavior

D8H:
This extension allows to make instances of Top-Down Movement behavior slide on the obstacles corners to bypass them smoothly. The controls are not directly applied. Instead, the behavior analyzes the player intents and move accordingly. It can choose to apply another control if it sees fit.

For instance, the player wants to go between 2 obstacles with just enough space between them. Without assistance, the player will always be a bit too much above or below and he may need to adjust the position 3 or 4 times which can be frustrating. With the assistance, when the right key is pressed and the player touches one obstacle, it will automatically align the player like if he were pressing up just the right amount.
TopDownAssistance

These are the exports from 2021-04-12:

* The interactive presentation
https://games.gdevelop-app.com/game-4d1afcce-647d-45bf-b8d3-980f26ef822a/index.html

* The isometric version (PageUp/PageDown to change level also)
https://games.gdevelop-app.com/game-c45be234-9ace-4dcd-920d-eea9feac0c87/index.html

* This is a build of the Bomberman example that uses it instead of the workaround with events
https://games.gdevelop-app.com/game-d9d9cb36-c8a7-45da-bb02-81f78c135da5/index.html

* The sources of the 2 test projects
https://www.dropbox.com/s/4fb052mw0961nyp/TopDownObstacleTestProjects.zip?dl=1

### DIFF
--- a/Extensions/TopDownMovementBehavior/Extension.cpp
+++ b/Extensions/TopDownMovementBehavior/Extension.cpp
@@ -9,6 +9,8 @@ This project is released under the MIT License.
 #include "GDCpp/Runtime/Project/BehaviorsSharedData.h"
 #include "TopDownMovementBehavior.h"
 #include "TopDownMovementRuntimeBehavior.h"
+#include "TopDownObstacleBehavior.h"
+//#include "TopDownObstacleRuntimeBehavior.h" no native implementation
 
 void DeclareTopDownMovementBehaviorExtension(gd::PlatformExtension& extension) {
   extension
@@ -21,6 +23,7 @@ void DeclareTopDownMovementBehaviorExtension(gd::PlatformExtension& extension) {
           "Open source (MIT License)")
       .SetExtensionHelpPath("/behaviors/topdown");
 
+  {
   gd::BehaviorMetadata& aut = extension.AddBehavior(
       "TopDownMovementBehavior",
       _("Top-down movement (4 or 8 directions)"),
@@ -374,7 +377,7 @@ void DeclareTopDownMovementBehaviorExtension(gd::PlatformExtension& extension) {
 
   aut.AddAction("AllowDiagonals",
                 _("Diagonal movement"),
-                _("Allow or restrict diagonal movemment"),
+                _("Allow or restrict diagonal movement"),
                 _("Allow diagonal moves for _PARAM0_: _PARAM2_"),
                 _("Movement"),
                 "CppPlatform/Extensions/topdownmovementicon24.png",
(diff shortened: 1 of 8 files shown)


D8H:
I tried to keep the same code style as the rest of the extension, but may be I should rewrite it ES6?
Every attributes are public?
_deltasX and _deltasY should be static but I wasn't sure how to do it.

4ian:
Hi, thanks for the PR! Following the existing style is fine (we're still using ES5 style but we'll migrate soon and semi-automatically to ES6/TypeScript).

Would you be to export to an HTML5 game the resulting project so I can quickly test it? :) I'd like to understand exactly how it behaves.
I wonder if this could be achieved without extending the extension, but just with the events (or JS events) to handle collisions.

D8H:
Hi, this is the test project export:
https://games.gdevelop-app.com/game-ea796b87-b769-4317-98de-8201045579a9/index.html

It’s possible to do something similar with events (in the PlayerMovement page from Game scene):
https://www.dropbox.com/s/tnvg04bgb2ztfrs/SansFoieNieLOie.zip?dl=1

But with events, it relies on diagonals and that may result to some small decelerations. I tried to make an extension with the editor but I need to access to the obstacles. Maybe this can be hacked with javascript events?

D8H:
There is a bug when the player is stopped in front of a passage a little bit aside and try to go through. It'll oscillate. This is because I didn't take the acceleration into account to evaluate the margin.

D8H:
The issue is fixed. It sets the velocity because otherwise the player can turn around the corner only if the "separate" action pushes him in the right direction. I think this can't be done with events without breaking the encapsulation.

The new export show another issue, when the passage is too small for the player. The player is still assisted and push in front of it. It will need an extra collision search to avoid this.
https://games.gdevelop-app.com/game-8f263a1a-da7b-43a2-8f0b-2fe26c3e1b41/index.html

D8H:
There is still the issue when passages are too small.
https://games.gdevelop-app.com/game-dd386a74-ff8f-44f1-a344-6734ef31519b/index.html

D8H:
In this demo, the obstacle detected to be in the way will flicker. There is a 2nd player to check that the behavior works if the player is an obstacle too.
https://games.gdevelop-app.com/game-b7c9a41e-f6f2-47ee-83a0-a19ff5457410/index.html

To do:
- Always (not only in the corners) assist the player when he goes in diagonal again a wall to keep full speed
- Check obstacles in the bypass way only once to be more consistent and avoid useless spacial search
- There is still some rare oscillations to debug

D8H:
I think the decision making is working fine now.
To debug, I made obstacles show there evaluation and the player changing color when he is realign to take a turn. I leaved it in this export, it's use full to understand the idea.
https://games.gdevelop-app.com/game-dc10f69f-78e8-4400-8c89-df68377c5bca/index.html

To do:
- The separation of the player and obstacles should apply the smaller swift possible and not just apply one after the other.
- Investigate why the player can intersect obstacles even after a separate (this do some glitches in the arrows show in the demo even though there is no actual bug).
- Clean up debug animations.

D8H:
The behavior seems to be working fine now.
In the demo, the blocks can be drag and drop and there is some levels to tests various settings:
https://games.gdevelop-app.com/game-6c672edd-157d-4674-8d78-ab3c26bcf1dd/index.html

To do
- Think of tricky cases to test
- Clean up the code from debug stuffs

D8H:
This is an interactive presentation of the behavior:
https://games.gdevelop-app.com/game-0b358b6d-8685-4393-9685-9f38c0121b25/index.html

It may only need some clean up from the debug codes now.

I saw this ticket about isometry: https://trello.com/c/vp6eCCZh
I think I'll try this next, because a basis transformation should do the trick.

D8H:
When the isometry is chosen, the assistance have some glitches. I need to debug it.
The isometry itself is not much code. I will make another pull request with only it.

D8H:
The assistance work well with isometry:
https://games.gdevelop-app.com/game-eb7a1d29-599b-4b1a-bca9-a82d9cd97399/index.html

Some bug where easier to reproduce with isometry because of rounding error. I'm confident on the algorithm now.

D8H:
There is a behavior setting "Enable obstacles bypass assistance" that is disable by default. When the assistance is not used there is only some "if" that adds up to the main loop. 
The assistance algorithm is now covered by a test suite.

This is the up to date exports:

* The interactive presentation
https://games.gdevelop-app.com/game-4d1afcce-647d-45bf-b8d3-980f26ef822a/index.html

* The isometric version (PageUp/PageDown to change level also)
https://games.gdevelop-app.com/game-c45be234-9ace-4dcd-920d-eea9feac0c87/index.html

* This is a build of the Bomberman example that uses it instead of the workaround with events
https://games.gdevelop-app.com/game-d9d9cb36-c8a7-45da-bb02-81f78c135da5/index.html

* The sources of the 2 test projects
https://www.dropbox.com/s/4fb052mw0961nyp/TopDownObstacleTestProjects.zip?dl=1

D8H:
I want to free my master branch and will probably try this with a user extension.

4ian:
Ok 👍

PR #2157, opened 2020-12-28
Convert the game engine (GDJS) and the extensions to TypeScript

4ian:
This converts all the game engine (GDJS "Runtime") and the extensions to TypeScript.

The conversion was semi-automated with a bunch of tools (lebab, gents) and specific scripts. 
This brings a lot of nice things:
- All the "old-fashioned" ES5 style "classes" have been translated to use `Class` and `extends`.
- Typing now works for inherited classes. You can type your runtime objects and behaviors :)
- No need to use the JSDoc annotations anymore to type things: TypeScript is easier to write and read.
- It will be easier to see types that are missing.

There are a lot of `any` remaining but this can easily be fixed as we work on files.

All the files are processed using *esbuild* to transpile them back to JavaScript. It's really fast (<500ms on my MacBook) so should be almost invisible. This is done by a script `npm run build` in GDJS, which is run automatically by GDevelop when a change is observed (like what is done currently).

Type checking is done by the TypeScript compiler (as before). You can run it manually (`npm run check-types` in GDJS folder) but I recommend you use VSCode (or another IDE/text editor that has TypeScript support). They should work out of the box thanks to the `tsconfig.json` file.

This also adds auto-formatting of all the TypeScript files (so everything is auto-formatted like in newIDE).

Still to do:
- [ ] See if we ship the sourcemaps, so that the Chrome debugger shows the original TypeScript sources when debugging (only for the previews in the editor).
- [ ] Convert the remaining Firebase tool files and P2P files to use TypeScript namespaces.
- [ ] Fix the generated GDJS docs not having types.
- [ ] Any change in GDJS or extensions will trigger a build, which will be copied to the newIDE in the node_modules for the web-app, which will trigger a full Flow restart, which is consuming a lot of CPU. Need to investigate this.
- [ ] Check again everything: tests, objects, behaviors, extensions.

## How can I try this and help finding issues?

CircleCI should build a version for us. In the meantime, getting the branch and running `npm install` in GDJS should be sufficient.

Even if most of the changes were done automatically, there were manual fixes and it's possible that some things have been broken. **Please report any game/example/extension that is broken.**

## I have a PR opened, should I restart it?

No, please continue as you do. We'll convert what you've did to TypeScript later (before or after merging, depending on the complexity of the PR).

## I'm afraid TypeScript will be too complicated for me

As we're "transpiling" .ts files separately from type checking, you can continue writing code without types or with errors, and you'll still be able to have it working.
Of course, we'll correct all types before merging a PR.

## Can we write TypeScript in JavaScript events inside GDevelop?

No, still only JavaScript is supported (though the builtin editor understands TypeScript, so you benefit from auto-completions and JSDoc type annotations).
This is because the transpiling is not done by GDevelop, but by a separate program only in development.

---

All things considered, I think this is a good step in the direction of modernizing the codebase and having it ready for the next 5 years to come :)


arthuro555:
Did you profile the code to verify that there is no regression in performance? 

> This also adds auto-formatting of all the TypeScript files (so everything is auto-formatted like in newIDE).

Nice, I was tired of having to skim through the changes made by prettier when trying to read the changes of a PR/commit

4ian:
The micro benchmarks in GDJS/tests/benchmarks have not changed much, I don't think there will be anything noticeable. I've looked at the generated output of a few classes, and it's what you would expect (nothing added much by esbuild) and very similar to what we would write by hand. So I'm not worried.
If anything, using real `Class` will unlock optimizations in JS engines.

4ian:
The generated docs by better-docs seem to have some confusion, there are pages for a `Layer` for example and another for a `gdjs.Layer`... not sure how namespace is handled 😬
I wonder if we should switch to typedoc or another alternative, but it's not obvious what to go with.

4ian:
Docs are fixed - and I decided to switch to a cleaner theme (better docs has huge headings, and some very small fonts on the contrary). Unfortunately types in functions are gone and are not parsed by better docs typescript module for JSDoc :/
Maybe Typedoc would be better but it seems not too have a very large community. Anyway!

I'll probably 1) release a new GDevelop version without this then 2) Merge this so we have time to find any regression before releasing a new version.

arthuro555:
I think typedoc should be a good choice as [cloudflare uses it](https://blog.cloudflare.com/generating-documentation-for-typescript-projects/).

4ian:
I've rebased this on master, made a few more checks and this seems to work properly for examples that I've tried :) I'll merge this soon. I think the PR that will be impacted the most is the one with array/booleans @arthuro555 but I believe it should not be too complicated to port your changes. 

This will also include changes from #2152.

4ian:
I think I've verified almost all feature in terms of extensions. I'll merge this tomorrow - the sooner the better. You'll only have to re-run `npm install` in newIDE/app (which will run `npm install` in GDJS) and re-run the app (npm start) to get it working.
You'll still be able to write on older non TS branches by re-running npm install.

Issue #969, opened 2019-03-11
Report improper/bad translation here (other than English)


4ian:
Translations are done by the community. If while using GDevelop in your language (other than English), you come across some badly translated sentences, you can:

1. Go to the crowdin project: https://crowdin.com/project/gdevelop - create an account if needed it's fast and you'll be able to contribute!
  * Choose your language.
  * Click on "GDevelop 5 (editor) messages to be translated" if the mistake is in the editor interface (but not related to events, objects or behaviors).
  * Or click on "GDevelop (Core and extensions) messages to be translated" if the mistake is in the events, related to an object or a behavior.
  * Search, on the top left, by entering the sentence with a mistake:
    ![image](https://user-images.githubusercontent.com/1280130/54155513-2fab6900-443c-11e9-8434-49a141f67357.png)
  * Finally, *suggest a better* translation in the middle or *add a comment* on the right.
2. Or you can report it here, including **a screenshot showing the error/mistake/typo**. But it's way way faster and better to report it on Crowdin.

*Thanks a lot for helping GDevelop!*

---

> Don't reply on this issue for other things than mistakes in translation, to keep the subject tidy :) I'll remove any message not related to translations. Thanks!

ArbETaperingsky:
Can not Login>_<
![error](https://github.com/ArbETaperingsky/KDSS/blob/master/error.png)

Bouh:
_ is missing on `message`
[https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/ObjectGroupEditor/index.js](https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/ObjectGroupEditor/index.js)

Zonacas:
Wrong translation in the Spanish version

in ... Always

Toujours is written and must be "Siempre"

![Toujours-Siempre](https://user-images.githubusercontent.com/37845241/78458419-7310ef00-76b1-11ea-9133-25ae8aa692d6.png)

Bouh:
> Toujours is written and must be "Siempre"

I've fix this thank you :)

SkyTheProtoDragon:
En pintor de formas dice: The. en vez de: Te.

Y también en el apartado luz no está traducido

Eso es todo lo que vi por el momento reportare si veo algo más ;)
![IMG_20200927_151152](https://user-images.githubusercontent.com/71973576/94365842-699e9780-00d4-11eb-88d3-1ba32fb04488.jpg)
[...]

PR #2160, opened 2020-12-30
Add missing translation tags for events

Bouh:
https://github.com/4ian/GDevelop/issues/969#issuecomment-752101542

A review is needed for the white space added with `\u00A0` in `JSCodeEvent.js`
It's working fine, just maybe there is something prettier :)

### DIFF
--- a/newIDE/app/src/EventsSheet/EventsTree/Renderers/ForEachEvent.js
+++ b/newIDE/app/src/EventsSheet/EventsTree/Renderers/ForEachEvent.js
@@ -104,7 +104,7 @@ export default class ForEachEvent extends React.Component<
             tabIndex={0}
           >
             {objectName ? (
-              <Trans>{`Repeat for each instance of ${objectName}:`}</Trans>
+              <Trans>Repeat for each instance of {objectName}:</Trans>
             ) : (
               <i>
                 <Trans>
--- a/newIDE/app/src/EventsSheet/EventsTree/Renderers/ForEachEvent.js
+++ b/newIDE/app/src/EventsSheet/EventsTree/Renderers/ForEachEvent.js
@@ -14,6 +14,7 @@ import ObjectField from '../../ParameterFields/ObjectField';
 import { type EventRendererProps } from './EventRenderer';
 import ConditionsActionsColumns from '../ConditionsActionsColumns';
 import { shouldActivate } from '../../../UI/KeyboardShortcuts/InteractionKeys.js';
+import { Trans } from '@lingui/macro';
 const gd: libGDevelop = global.gd;
 
 const styles = {
@@ -103,10 +104,12 @@ export default class ForEachEvent extends React.Component<
             tabIndex={0}
           >
             {objectName ? (
-              `Repeat for each instance of ${objectName}:`
+              <Trans>{`Repeat for each instance of ${objectName}:`}</Trans>
             ) : (
               <i>
-                Click to choose for which objects this event will be repeated
+                <Trans>
+                  Click to choose for which objects this event will be repeated
+                </Trans>
               </i>
             )}
           </span>
--- a/newIDE/app/src/EventsSheet/EventsTree/Renderers/GroupEvent.js
+++ b/newIDE/app/src/EventsSheet/EventsTree/Renderers/GroupEvent.js
@@ -16,6 +16,7 @@ import {
   shouldCloseOrCancel,
   shouldValidate,
 } from '../../../UI/KeyboardShortcuts/InteractionKeys';
+import { Trans } from '@lingui/macro';
 const gd: libGDevelop = global.gd;
 
 const styles = {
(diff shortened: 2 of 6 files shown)


newIDE/app/src/EventsSheet/EventsTree/Renderers/ForEachEvent.js
```diff
@@ -103,10 +104,12 @@ export default class ForEachEvent extends React.Component<
             tabIndex={0}
           >
             {objectName ? (
-              `Repeat for each instance of ${objectName}:`
+              <Trans>{`Repeat for each instance of ${objectName}:`}</Trans>
```
4ian, line 107: I'm not sure this will work, this will generate a translation string like this: "{0}", because it's seen as a single variable (inside the brackets { }).
Instead:
`<Trans>Repeat for each instance of {objectName}:</Trans>`

which will be properly parsed as "Repeat for each instance of {0}:"

newIDE/app/src/EventsSheet/EventsTree/Renderers/GroupEvent.js
```diff
@@ -121,7 +122,11 @@ export default class GroupEvent extends React.Component<EventRendererProps, *> {
             })}
             style={{ ...styles.title, color: textColor }}
           >
-            {groupEvent.getName() || '<Enter group name>'}
+            {groupEvent.getName() ? (
+              groupEvent.getName()
+            ) : (
+              <Trans>{`<Enter group name>`}</Trans>
```
4ian, line 128: Same as before, this will be parsed as "{0}", so nothing will be translated.
Instead, try: `<Trans>&lt;Enter group name&gt;</Trans>`

newIDE/app/src/EventsSheet/EventsTree/Renderers/JsCodeEvent.js
```diff
@@ -186,9 +187,11 @@ export default class JsCodeEvent extends React.Component<
         tabIndex={0}
         style={textStyle}
       >
-        {parameterObjects
-          ? `, objects /*${parameterObjects}*/`
-          : ' /* Click here to choose objects to pass to JavaScript */'}
+        {parameterObjects ? (
+          <Trans>{`, objects /*${parameterObjects}*/`}</Trans>
```
4ian, line 191: Same everywhere, remove all the brackets with the backticks :)

newIDE/app/src/EventsSheet/EventsTree/Renderers/JsCodeEvent.js
```diff
@@ -186,9 +187,11 @@ export default class JsCodeEvent extends React.Component<
...
+        ) : (
+          <Trans>{`\u00A0/* Click here to choose objects to pass to JavaScript */`}</Trans>
```
4ian, line 193:
```
<>
{' '}
<Trans>/* Click here....*/</Trans>
</>
```

You might need to escape / and *.

4ian: Or maybe:
```
<>
&nbsp;
<Trans>/* Click here....*/</Trans>
</>
```

Search how to escape / and * in React JSX (not sure if you need to, maybe not!)

newIDE/app/src/EventsSheet/EventsTree/Renderers/GroupEvent.js
Bouh: It works fine with my code ;)
image

4ian: Oh right then! I was not aware that jslingui was so smart at parsing these!

newIDE/app/src/EventsSheet/EventsTree/Renderers/JsCodeEvent.js
4ian: Actually you're right, template literals are supported! 
It's in the Trans component test suite: https://github.com/lingui/js-lingui/blob/3afaceccb669b59ee2f5b42ee2e138646ccdb79d/packages/macro/test/jsx-trans.ts#L96-L107

newIDE/app/src/EventsSheet/EventsTree/Renderers/GroupEvent.js
4ian: Confirmed in their source code: https://github.com/lingui/js-lingui/blob/3afaceccb669b59ee2f5b42ee2e138646ccdb79d/packages/macro/test/jsx-trans.ts#L96-L107

newIDE/app/src/EventsSheet/EventsTree/Renderers/JsCodeEvent.js
Bouh: Not sure
image

4ian (approved):
You need to run Prettier, then should be good to go!

PR #2161, opened 2020-12-30
Use requestFullscreen() / exitFullscreen()

jonasfj:
See [Fullscreen API on MDN](https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API).

It would probably be nice to do a follow up that handles the [`fullscreenchange` event](https://developer.mozilla.org/en-US/docs/Web/API/Document/fullscreenchange_event).

### DIFF
--- a/GDJS/Runtime/pixi-renderers/runtimegame-pixi-renderer.ts
+++ b/GDJS/Runtime/pixi-renderers/runtimegame-pixi-renderer.ts
@@ -300,9 +300,9 @@ namespace gdjs {
             }
           } else {
             // @ts-ignore
-            if (document.cancelFullscreen) {
+            if (document.exitFullscreen) {
               // @ts-ignore
-              document.cancelFullscreen();
+              document.exitFullscreen();
             } else {
               // @ts-ignore
               if (document.mozCancelFullScreen) {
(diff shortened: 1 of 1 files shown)


GDJS/Runtime/pixi-renderers/runtimegame-pixi-renderer.js
```diff
@@ -261,15 +261,19 @@ gdjs.RuntimeGamePixiRenderer.prototype.setFullScreen = function (enable) {
       // Use HTML5 Fullscreen API
       //TODO: Do this on a user gesture, otherwise most browsers won't activate fullscreen
       if (this._isFullscreen) {
-        if (document.documentElement.requestFullScreen) {
+        if (document.documentElement.requestFullscreen) {
+          document.documentElement.requestFullscreen();
+        } else if (document.documentElement.requestFullScreen) {
```
jonasfj, line 266: I suppose it's possible that `requestFullScreen` is a typo, and that it should be `requestFullscreen` instead.

ie. that we don't need to retain `requestFullScreen` as an option.

4ian: Possibly! Must have been a mistake because the old `webkit` and `moz` prefixed versions are `xxxRequestFullScreen`. Let's remove `requestFullScreen` then :)

4ian: @jonasfj Do you want to remove the references to requestFullScreen?

jonasfj: Done

4ian:
I've switched to `exitFullscreen` because this seems to be the standard now. Thanks!

PR #2169, opened 2021-01-03
Update README.md

INNOVATIVEGAMER:
previous installation steps did not include the npm install on app directory in electron-app which is resulting errors in installation.
Now installation steps are modified and are error free.

### DIFF
--- a/newIDE/README.md
+++ b/newIDE/README.md
@@ -3,7 +3,7 @@
 This is the GDevelop 5 editor. It is based on [React](https://facebook.github.io/react/), [Material-UI](http://www.material-ui.com), [Pixi.js](https://github.com/pixijs/pixi.js) and [Electron](https://electron.atom.io/).
 It uses GDevelop [core C++ classes compiled to Javascript](https://github.com/4ian/GDevelop.js) to work with GDevelop games.
 
-![GDevelop editor](https://raw.githubusercontent.com/4ian/GDevelop/master/newIDE/gd-ide-screenshot.png 'GDevelop editor')
+![GDevelop editor](https://raw.githubusercontent.com/4ian/GDevelop/master/newIDE/gd-ide-screenshot.png "GDevelop editor")
 
 ## 1) Installation 💻
 
@@ -37,6 +37,11 @@ You can run the standalone app with Electron. **Make sure that you've launched `
 cd newIDE/app && npm start # Be sure to have this running in another terminal, before the rest!
 
 # In a new terminal:
+cd newIDE/electron-app/app
+npm install # or yarn
+# After successful installation close this terminal
+
+# In a new terminal:
 cd newIDE/electron-app
 npm install # or yarn
 
@@ -54,7 +59,7 @@ npm run electron-linux
 
 There is a script file that automates cloning this repository, building the newIde and running it
 
-- For Windows: You can download the batch script [here](https://raw.githubusercontent.com/4ian/GDevelop/master/scripts/gitCloneAndBuildGD.bat) and save it to where you want GD to be cloned to, then simply run it.
+-   For Windows: You can download the batch script [here](https://raw.githubusercontent.com/4ian/GDevelop/master/scripts/gitCloneAndBuildGD.bat) and save it to where you want GD to be cloned to, then simply run it.
 
 ### Development of UI components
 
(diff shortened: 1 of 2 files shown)


newIDE/README.md
```diff
@@ -36,6 +36,11 @@ You can run the standalone app with Electron. **Make sure that you've launched `
 ```bash
 cd newIDE/app && npm start # Be sure to have this running in another terminal, before the rest!
 
+# In a new terminal:
+cd newIDE/electron-app/app
+npm install # or yarn
+# After successful installation close this terminal
```
4ian, line 42: This should not be needed because it's already in the `postinstall` hook that is launched by npm after the `npm install` in `newIDE/electron-app`:

https://github.com/4ian/GDevelop/blob/329672d283e87116c7aba851823bdb073b4c9d70/newIDE/electron-app/package.json#L12

INNOVATIVEGAMER: But when I installed the dependencies as mentioned in the readme , there was npm errors in the terminal. I had to manually run npm install on app directory to remove the npm errors. What do you think the reason could be ?

4ian: Hard to say without reading the errors, can you paste the whole errors somewhere like on pastebin?

INNOVATIVEGAMER: I am posting SS of the errors with pastebin link of the errors . The development server is running and then I tried to run npm install on electron-app directory
pastebin link - [https://pastebin.com/QeAJ18ct](url)
Screenshot (17)

4ian: Sweet :) Though I'm not exactly sure why this is happening. Have you tried googling errors related to git-sh-setup?
Have you tried applying this? https://stackoverflow.com/questions/49256190/how-to-fix-git-sh-setup-file-not-found-in-windows

INNOVATIVEGAMER: Yes , the solution on the stackoverflow fixed the issue for me. But don't you think the installation processes shall be independent of the user's system configuration ? 
The workaround specified by me in the readme also fixes the issue. At last you are the expert. 
Thank you for your guidance.

4ian: > But don't you think the installation processes shall be independent of the user's system configuration ?

Sure! But I'd like to understand why this is not working by default, as there is no reason running `npm install` in the `postinstall` is not working :) Is this a Windows issue? A bug in git? A bug in npm? Why running it again fix the issue? If we don't understand this, we risk making the Readme more complex for nothing, or for the wrong reason :)

INNOVATIVEGAMER: Ok ,after some googling I found this : 
1.This is the problem with the git, npm and windows collectively. After the npm release v10 this issue came forward.
2.The problem is git is supposed to add git-core to the PATH itself. And it does it when it's not run by npm. The thing is that when npm runs git, then git-core path is not added to the PATH. 
3.But there is workaround for this which is manually adding the path in windows. Issues regarding this on git are still open as you can see here - https://github.com/appveyor/ci/issues/2420
4.I think running it first on the the internal directory fixes issue in our case as the post install steps which are recursive steps for the npm are done manually.
So I think until the git or npm find the solution we have to manually add the steps for postinstall steps in readme or instructions for adding the path in windows referencing this issue.

INNOVATIVEGAMER: Sorry for these commits in this pull request.
Actually I am going to make another pull request which i have worked on another issue but these commits are merged in this pull request by mistake. 
I reverted the commit so now only readme is changed in this PR.
Sorry for the mistake :disappointed_relieved:

4ian: > After the npm release v10 this issue came forward.

I think you mean Node.js v10? Can you try with Node.js v12 or v14? I.e: remove the additional path from your PATH env variable, try to npm install to verify that it fails as previously. Then install Node.js 14 and see if works without errors this time.
If yes, then we can instruct in the Readme that on Windows, you need Node.js v12 at least :) And an additional sentence telling that if you're on Node.js v10 on Windows, you need to run npm install in electron-app/app too (but better upgrade to the latest Node.js version anyway).

Let me know :) 

> Sorry for these commits in this pull request.

No worries, as long as you revert it's fine. I merge commits by squashing them, so only the changes displayed on GitHub will appear as a single commit ;)

INNOVATIVEGAMER: As you told I tried running npm install with different version of node.js.
Results are not satisfying though. As in my case I was using node 12.16.2
I checked this with 4 different version of node as follows 

1. 10.20.1 (npm errors)
2. 12.16.2 (npm errors)
3. 14.15.3 (npm errors)
4. 15.5.0  (successful installation) (latest on date 04/01/2021)

I am attaching the SS of the results
MINGW64__d_study_Web Developement_GSOC 2021_GDevelop_newIDE_electron-app 04-01-2021 15_41_39
MINGW64__d_study_Web Developement_GSOC 2021_GDevelop_newIDE_electron-app 04-01-2021 15_33_41
MINGW64__d_study_Web Developement_GSOC 2021_GDevelop_newIDE_electron-app 04-01-2021 15_18_00
MINGW64__d_study_Web Developement_GSOC 2021_GDevelop_newIDE_electron-app 04-01-2021 15_04_28

So now I think we have three choices :
1. Tell user to manually add path in windows env variable
2. Tell user to run the npm install in internal directory as I did in the commit
3. Set the windows path using npm script before running npm install (I am not sure though it will work and how to do it)

4ian: Thanks for checking all these versions! 👍

Let's put in the readme that "On Windows, with Node.js 14 or older, there is an error related to `git-sh-setup` when running npm install. Either add this folder to your path or run `npm install` in newIDE/electron-app/app after `npm install` in newIDE/electron-app".

Put a link for "this folder to your path" to the StackOverflow explanation :) 
With this we should be good I think.

INNOVATIVEGAMER: Ok I will make a commit accordingly.

INNOVATIVEGAMER: Updated the readme file in latest commit.
Do you want any changes in it ?

4ian:
Thanks! I've made a few additional fixes. 
Merging now, thanks for researching this problem :)

PR #2174, opened 2021-01-04
Fix wording & type in function parameters

Bouh:
The term (boolean) wasn't correct, it must describe the type of the value returned like the other options.
image

This type is useful for know what to use `GetArgumentAsNumber ` or `GetArgumentAsString`.

- Yes or no (boolean) act like 0 and 1, we have to get the parameter with `GetArgumentAsNumber ` goes to `Yes or No (number)`
- True or false (boolean) we have to use `GetArgumentAsString` goes to `True or False (text) `

### DIFF
--- a/Core/GDCore/IDE/Events/InstructionSentenceFormatter.cpp
+++ b/Core/GDCore/IDE/Events/InstructionSentenceFormatter.cpp
@@ -136,7 +136,7 @@ gd::String InstructionSentenceFormatter::LabelFromType(const gd::String &type) {
   else if (type == "mouse")
     return _("Mouse button");
   else if (type == "yesorno")
-    return _("Yes or no");
+    return _("Yes or No");
   else if (type == "police")
     return _("Font");
   else if (type == "color")
--- a/newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
+++ b/newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
@@ -385,11 +385,11 @@ export default class EventsFunctionParametersEditor extends React.Component<
                                 />
                                 <SelectOption
                                   value="yesorno"
-                                  primaryText={t`Yes or No (number)`}
+                                  primaryText={t`Yes or No (boolean)`}
                                 />
                                 <SelectOption
                                   value="trueorfalse"
-                                  primaryText={t`True or False (text)`}
+                                  primaryText={t`True or False (boolean)`}
                                 />
                               </SelectField>
                             )}
--- a/newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
+++ b/newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
@@ -385,11 +385,11 @@ export default class EventsFunctionParametersEditor extends React.Component<
                                 />
                                 <SelectOption
                                   value="yesorno"
-                                  primaryText={t`Yes or no (boolean)`}
+                                  primaryText={t`Yes or No (number)`}
                                 />
                                 <SelectOption
                                   value="trueorfalse"
-                                  primaryText={t`True or false (boolean)`}
+                                  primaryText={t`True or False (text)`}
                                 />
                               </SelectField>
                             )}


4ian:
It's actually a boolean internally, so should stay written as "(boolean)" I think.
There is a condition to check if an argument is true (or yes, that's equivalent): 
image

As you've seen `GetArgumentAsNumber` will convert to 0/1, and `GetArgumentAsString` to "false"/"true".

Bouh:
The user should not have to worry about the type used in the background for the game engine. This is the goal of GDevelop to ignore everything under the hood.

As an user not as dev:

Color (text) refers to "255;255;255" so a string.

Layer (text) is often "Background color", a string.

Yes or no (boolean) is equal to 0 or 1, for me it's a number.
True or false (boolean) is equal to true or false, for me it's a string.

And when I see (boolean) with two different types of values returned I would be confuse.
The logic of this list is lost.

As dev:
The game engine return 0 or 1 and true or false as a string.

I do not refer to the internal code of the engine since here we try to avoid it and have a consistency on the type to be used for an expression.

This condition is cool but it shouldn't be mantatory to an user to find it and read it for understand what we should use.

Bouh:
In this case if (boolean) is keep, we need to get a new expression for get the value of these options.

getArgumentAsBoolean("myParameter")

but what about the second boolean ?

getArgumentAsTrueFalseBoolean("myParameter") ?

arthuro555:
> we need to get new expressions for get the value of these options.

But this isn't possible as Boolean expressions *are* conditions (for now). 

> but what about the second boolean ?

Well, it can be gotten by the same condition, you also get the color via `GetArgumentAsString` and not `GetArgumentByColor`.

> Yes or no (boolean) is equal to 0 or 1, for me it's a number.
True or false (boolean) is equal to true or false, for me it's a string.

My understanding is that this changes what is displayed to the users of the function, Aka a "true" and a "false" button or a "Yes" and a "No" button, and has nothing to do with the type used in the background (which is a boolean for both)

Bouh:
Je pense que ce que je souhaite expliqué n'est pas clair...
Voici une liste de correspondance entre les types et les valeurs retourné à l'utilisateur par le moteur.

Objects => un objet de GDevelop
Behavior => un comportement de GDevelop
Number => `0`, `1`, `2`, `3` .... un nombre
String (text) => `"Hello world"` du texte
String from a list of options (text)  => `"Option A"` du texte
Keyboard (text) => `"Left"`, `"Up"`, `"A"`, `"B"` `"C"` ... du texte
Mouse button (text) => `"Left"`, `"Right"`, `"Middle"` du texte
Color (text) => `"255;255;255"` du texte
Layer (text) => `"Background Color"`, `"Base layer"` du texte
Scene name  (text) => `"New scene"` du texte
Yes or no (boolean) => `0` ou `1` un nombre
True or false (boolean) => `"true"` ou `"false"` du texte

Pour moi ça ne colle pas, il n'y a _aucune cohérence_ entre les types exprimé entre parenthèses.

Dans une feuille d'évènement:
- La valeur des paramètres ayant `(text)` indiqué seront récupérable avec `GetArgumentAsString()`
- La valeur du paramètre de type `Number` est lui récupérable avec `GetArgumentAsNuber()`.
- `Objects` et `Behavior` sont utilisable directement via le nom du paramètre.

Les paramètres ayant `(boolean)` se récupèrent avec soit  `GetArgumentAsString()` ou `GetArgumentAsNuber()` là la logique ce perd.

D'où la recommandation logique de mettre `Yes or no (number)` et `True or false (text)`

arthuro555:
De ce que j'ai compris, Yes or No ne retourne pas de numéro et True or False ne retourne pas de numero. Les deux sont la même valeur booléenne et sont convertis en texte si tu essaie de l'obtenir en tant que texte et est converti en numéro si tu essaie de l'obtenir en tant que numéro. Les valeurs booléennes sont faites pour être récupérés avec la condition `GetArgumentAsBoolean`, et non les expressions pour récupérer du texte ou des nombres.

4ian:
> Yes or no (boolean) => 0 ou 1 un nombre
> True or false (boolean) => "true" ou "false" du texte

This part is wrong. The correct mapping is:

> Yes or no (boolean) => a JavaScript boolean
> True or false (boolean) => a JavaScript boolean

For now we don't support booleans in expressions, so the only way to "properly" get your boolean is the condition, specifically made for it :) 
Otherwise, the 2 existing expressions: `GetArgumentAsString()` ou `GetArgumentAsNumber()` will still convert your boolean to a number or text :) 

---

En d'autre termes, ce que dit Arthuro555 est correct :

> De ce que j'ai compris, Yes or No et True or False ne retournent tout les deux ni nombre ni texte. Les deux sont la même valeur booléenne et sont convertis en texte si tu essaie de l'obtenir en tant que texte et est converti en numéro si tu essaie de l'obtenir en tant que numéro. Les valeurs booléennes sont faites pour être récupérés avec la condition GetArgumentAsBoolean, et non les expressions pour récupérer du texte ou des nombres.

Bouh:
Both are usable cool!

Issue #2168, opened 2021-01-02
I18n request


yu-tang:
Internationalization support is needed for the term "Instant."
See below image. Left: English, Right: Japanese.
![instant](https://user-images.githubusercontent.com/578576/103464522-a0d9b000-4d77-11eb-917c-696b573e3cdb.png)

Bouh:
Thank you for this report!

If someone want give a try this is a good first issue, feel free to start and submit a PR.
You should edit this file

[`newIDE\app\src\EventsSheet\ParameterFields\ForceMultiplierField.js`](https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/EventsSheet/ParameterFields/ForceMultiplierField.js#L41)

Probally remove the function `makeNonBreakable` and the import, and wrap the word 'Instant' in a `<Trans>` tag.

Manas502:
Could you possibly elucidate a bit more about the issue?

INNOVATIVEGAMER:
I applied fixes as you said but the text "instant" is not getting translated by <Trans> tag.
@Bouh I think we also need to provide translations for "INSTANT" word in i18n provider.

Bouh:
If the word isn't yet translate in the dev en it is because instant word
isn't yet on crowdin.

4ian:
To verify something is included in translations, choose the language "Pseudolocalization" (in development mode only):
![image](https://user-images.githubusercontent.com/1280130/103479728-f04bca80-4dcf-11eb-8fc7-74dddcaf627d.png)

Then to see something that you added, like Instant, run in newIDE/app: `npm run extract-all-translations` then `npm run compile-translations`. Then refresh the editor and you should see the word "Instant" to be "pseudo translated", like Permanent on this screenshot:

![image](https://user-images.githubusercontent.com/1280130/103479763-2b4dfe00-4dd0-11eb-8a43-5dab2a3f6453.png)

yu-tang:
Hi @Manas502,
I am translating GDevelop to my language on Crowdin for a while.
The term "Instant" does not appears in source strings, so we localizers can do nothing for this now and the term is remained as in English regardless of what your locale settings.

One more thing, I want to make sure that the term has two versions. One for Condition/Action editor (image top), and one for Event editor with embedding in sentence (image bottom). I suspect these source strings come from the different files/contexts.

Bouh:
@yu-tang `Instant` and `an instant` are two different string.

[...]

PR #2180, opened 2021-01-05
Fixed string translations mentioned in the issue #2168

INNOVATIVEGAMER:
This commit is fix for the issue #2168 .
The strings are now getting translated with these code changes ,but currently these strings are not translated on crowdin.
So as for the issue this code actually fixes the issue on dev side only.


INNOVATIVEGAMER:
> We could almost not translate `'a (multiplier: ' + value + ')'` as it's deprecated and not worth translating anymore.

So if its deprecated, shall I remove the default return statement.
I think there should be one default return statement

Bouh:
> > We could almost not translate `'a (multiplier: ' + value + ')'` as it's deprecated and not worth translating anymore.
> 
> So if its deprecated, shall I remove the default return statement or just remove the trans tag.
> I think there should be one default return statement

You can let the line by default without trans tag indeed ;)
One day this will be delete properly with the related stuff unsused.

INNOVATIVEGAMER:
Requested changes are applied.

newIDE/app/src/EventsSheet/ParameterFields/ForceMultiplierField.js
```diff
@@ -102,8 +101,8 @@ export default class ForceMultiplierField extends Component<
 export const renderInlineForceMultiplier = ({
   value,
 }: ParameterInlineRendererProps) => {
-  if (value === '1') return 'a permanent';
-  else if (value === '0' || value === '') return 'an instant';
+  if (value === '1') return <Trans>{`a permanent`}</Trans>;
+  else if (value === '0' || value === '') return <Trans>{`an instant`}</Trans>;
 
-  return 'a (multiplier: ' + value + ')';
+  return `a (multiplier: ' + ${value} + ')`;
```
4ian, line 107: This is not correct ;)

Bouh: The original was  

https://github.com/4ian/GDevelop/blob/bb296dc34e4af818ac45f48450b1f6691b9d0cc9/newIDE/app/src/EventsSheet/ParameterFields/ForceMultiplierField.js#L108

INNOVATIVEGAMER: Sorry for the silly mistake. 
Template string now converted back to original string.

Bouh (approved):

4ian:
Looks good, thanks!

Issue #2086, opened 2020-11-11
Dark themes: Event handles barely visible


Wend1go:
## Describe the bug
The handle to hide sub events is barely visible in the Nord theme.
![image](https://user-images.githubusercontent.com/7413059/98784695-c98a9c80-23fb-11eb-8ed2-b5ee303a7ba7.png)
(See the handle left of "UI Handling")

Wend1go:
Same for all dark themes

Manas502:
Can I take up the issue?

4ian:
Sure feel free to go ahead!

Manas502:
Hey @4ian can you guide me a little on where I can find Nord theme styles?

4ian:
https://github.com/4ian/GDevelop/tree/master/newIDE/app/src/UI/Theme/NordTheme

Though in this case, use the debugger to identify where/how is rendered the arrow, and then check how you could style it.
See
https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/EventsSheet/EventsTree/style.css and 
https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/UI/Theme/Global/EventsSheet.css

Silver-Streak:
Looks like this was fixed a while back.
![image](https://user-images.githubusercontent.com/24237149/216802962-8e0f21b9-998a-486b-b765-5f0b68659005.png)

PR #2181, opened 2021-01-05
Fixed event handles [#2086]

Manas502:
Fixed the issue where event handles were barely visible in dark themes.


4ian:
Thanks for the PR, can you post a few screenshots of it being used in light/dark themes? :)

Manas502:
Sure, here are the screenshots: 
Screenshot from 2021-01-06 00-45-10
Screenshot from 2021-01-06 00-44-22

4ian:
The non expanded version looks a bit off:
image

Compared to the well aligned existing version:
image

Can we revert this change? :)

4ian:
No, I just meant setting back the background-position as before in this PR :)

4ian:
Ok, seems better! Could you investigate modifying the theme styles (check theme.json files) to apply a color to this icon according to the theme? That would allow to use a black color for the light themes, and a dark color for dark themes (by switching the image source, or using another CSS feature to invert/change the color).

Manas502:
> Ok, seems better! Could you investigate modifying the theme styles (check theme.json files) to apply a color to this icon according to the theme? That would allow to use a black color for the light themes, and a dark color for dark themes (by switching the image source, or using another CSS feature to invert/change the color).

Sure I'll try :)

Manas502:
@4ian I think the better approach would be to switch the image source according to the applied theme...What do you think?

4ian:
I think it's a good idea! :)

Manas502:
@4ian I am encountering an error on running the server. Could you guide me a bit on this?
image

4ian:
You need to give me more information about what led to this error. When is it happening? 
You seem not to have WebGL working in your browser?

Issue #2165, opened 2021-01-01
Open URL, opening new window within GDevelop


AtomicCrocStudios:
## Describe the bug
I am putting this as a bug as @arthuro555 is much smarter than me and he thinks it's one. :D
When using the **Open a URL or file** action, the URL is opened using a GDevelop window. According to arthuro, the code should open the URL in the browser and not a new window within GDevelop.

If this is intended behavior, can we please have an option to override this and have the URL open in the users default browser. (I am currently using a JS function to achieve this, but it would be great to have it part of the action)

## To Reproduce
![image](https://i.gyazo.com/6b44d286937e7065b20854d222b76372.png)
![image](https://i.gyazo.com/64a366a1be7da0a7c97386f9da57ea20.gif)

## Other details
- Windows 10, Preview/Window App (not within a browser)
- GDev 5.00 Beta 102

arthuro555:
https://github.com/4ian/GDevelop/blob/master/GDJS/Runtime/Electron/main.js#L25-L29
This is the code that less me to believe it is unintended behavior.

4ian:
> According to arthuro, the code should open the URL in the browser and not a new window within GDevelop.

Correct, this is a bug and should open in an external browser! 
PR to fix this is welcome :)

INNOVATIVEGAMER:
Can somebody please help me.
I am new to this project so I am having some trouble in development.
as @arthuro555 mentioned in his comment , I tried to change some code in the main.js but saving changes and recompiling the project not changing the behavior of the application.
I commented the whole function to check actually what is going on but there was no effect on the app.
![main js - GDevelop - Visual Studio Code 03-01-2021 12_08_02](https://user-images.githubusercontent.com/55017197/103473301-0e5bff80-4dbd-11eb-8488-3eaf2c684c06.png)

arthuro555:
As this is a change with the electron process, killing the `npm run electron-<platform>` process and running it again should do the trick.

INNOVATIVEGAMER:
> This is the code that less me to believe it is unintended behavior.

Are you sure about this file ?
Because changing the code in this file is not affecting the behavior at all.
Also I think this code is for the electron runtime ,so how it will affect the openURL event action ?

INNOVATIVEGAMER:
@4ian Can you please tell me where is file related to "Open a URL or File" action is located ? :)

4ian:
[...]

PR #2189, opened 2021-01-07
Open URL, opening new window within GDevelop

INNOVATIVEGAMER:
Fixed the issue #2165 .
Created a isElectron helper method to detect running environment is Electron or not.

### DIFF
--- a/GDJS/Runtime/pixi-renderers/runtimegame-pixi-renderer.js
+++ b/GDJS/Runtime/pixi-renderers/runtimegame-pixi-renderer.js
@@ -479,7 +479,45 @@ gdjs.RuntimeGamePixiRenderer.prototype.openURL = function (url) {
     Cocoon.App.openURL(url);
   } else if (typeof window !== 'undefined') {
     var target = window.cordova ? '_system' : '_blank';
-    window.open(url, target);
+
+    //To detect if the application is running in electron or other browser window
+    function isElectron() {
+      // Renderer process
+      if (
+        typeof window !== 'undefined' &&
+        typeof window.process === 'object' &&
+        window.process.type === 'renderer'
+      ) {
+        return true;
+      }
+
+      // Main process
+      if (
+        typeof process !== 'undefined' &&
+        typeof process.versions === 'object' &&
+        !!process.versions.electron
+      ) {
+        return true;
+      }
+
+      // Detect the user agent when the `nodeIntegration` option is set to true
+      if (
+        typeof navigator === 'object' &&
+        typeof navigator.userAgent === 'string' &&
+        navigator.userAgent.indexOf('Electron') >= 0
+      ) {
+        return true;
+      }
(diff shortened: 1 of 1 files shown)


INNOVATIVEGAMER:
I think this isElectron function should be in utils but there is no utils folder in GDJS folder. :sweat_smile:

GDJS/Runtime/pixi-renderers/runtimegame-pixi-renderer.js
```diff
@@ -479,7 +479,45 @@ gdjs.RuntimeGamePixiRenderer.prototype.openURL = function (url) {
...
+      if (
+        typeof navigator === 'object' &&
+        typeof navigator.userAgent === 'string' &&
+        navigator.userAgent.indexOf('Electron') >= 0
+      ) {
+        return true;
+      }
+
+      return false;
+    }
```
4ian, line 513: You can just do: 
```
const electron = this.getElectron();
if (electron) { ... }
```

4ian: This is because getElectron method is the one responsible for getting the electron module. If it can't it will return null :) 
The rest of the codebase should not have to deal with complex code to known if we're inside Electron or not: just call this function ;) This way if we need to improve it later, there is a single place where we have to do this.

INNOVATIVEGAMER:
Thanks for the getElectron trick, it worked.
Can you please help me in one thing ?
While trying to figure out the issue, I was searching for the function that is responsible for the openURL action but I was unable to find it using debugging in chrome.
In the issue section you marked exactly the function in no time.
How you debugged the code ? Can you please describe the steps so I can do it next time myself ?
BTW, Thank you for your continuous support :)

4ian:
You can debug this by:
- search the codebase for `window.open`
- or do the reverse, start your search by the action name "Open a URL or a file" in the codebase. You'll find it here:
https://github.com/4ian/GDevelop/blob/b475a2ed9625d268b6924fd88cde306d666aa514/Core/GDCore/Extensions/Builtin/FileExtension.cpp#L170-L171

Then you see that the action internal name is "LaunchFile". So you can search for it and find the file where it's associated to the JavaScript function:
https://github.com/4ian/GDevelop/blob/238bf27671ba8172d62d4e7cf8f16f1178962184/GDJS/GDJS/Extensions/Builtin/FileExtension.cpp#L37

Finally,  you search for the JavaScript function:
https://github.com/4ian/GDevelop/blob/bc56f820b3babf5d50cfdefbcbda1d30b914c161/GDJS/Runtime/events-tools/windowtools.js#L172-L173

You can put a breakpoint in a preview to go step by step, but from there you can search `openURL` and see where it's implemented :)

4ian:
Looks good, merging now! Thanks for fixing this :)

PR #2191, opened 2021-01-07
Add support for rotated tiles to Tilemap

4ian:
Adds support for Tiled rotated/flipped tiles. 

I've tested exporting the Platformer example from LDtk to Tiled, then to Tiled as JSON, and it works well :)
image

Of course we don't support everything ("entities" and multiple tilesets), but this is more than enough I think for a lot of use cases.

cc @blurymind


blurymind:
Wow that is fast. Fantastic work on this 👏 

 I think it looks good, also the implementation in the helper is interesting

blurymind:
Would be worth adding the tilemap in the screenshot to the example demo project, with the other tilemaps. Just to be sure they all work together in a scene :)

4ian:
> Would be worth adding the tilemap in the screenshot to the example demo project, with the other tilemaps. Just to be sure they all work together in a scene :)

Totally! It's true that otherwise this would be a pain to verify that nothing is broken when we do future upgrade/refactorings.
I'll add this Tilemap to the example, so we can verify all the feature at once (animations, resize, rotation...)

> If ldtk could both read and write to the json we can parse, what's to stop us from using it as an external editor?

The workflow is still for now a bit cumbersome. 
- You have to enable Tiled TMX export in LDtk (we could force/automate this presumably)
- After saving, you have to open it in Tiled and save it as a JSON (maybe we could autotranslate it to JSON, or just support XML files as well as JSON for tilemaps - not sure what's the easiest).
- You may have some warnings if you use unsupported features.

It's good news that LDtk export well to Tiled and that we can then open it, but I also think we're pretty far from a polished experience: we would need to automate these steps, hide all the stuff in LDtk that are unsupported, adapt the UI to integrate it with GD (before getting tons of users confused)... not impossible, but it's not just putting the editor and a save button. But it's definitely something that would be nice to look at :)

Issue #2195, opened 2021-01-08
Doubled extension prefix


PascalLadalle:
When editing an expression with the Dialogue extension, the prefix gets doubled:
![image](https://user-images.githubusercontent.com/25978830/104039192-583a5080-51d6-11eb-8f53-95e7567b2ef3.png)

Reproducing steps:
Click on a text field, write "tags", press Return. 
Realize you wanted "BranchTag()", so hit backspace three times and press Enter.

Bouh:
![to_delete](https://user-images.githubusercontent.com/1670670/104040138-94ba7c00-51d7-11eb-9b99-ef62d1f1f143.gif)

D8H:
Maybe this issue is solved by this PR: https://github.com/4ian/GDevelop/pull/2702
Because the the way the autocompletion is inserted has changed, but I'm not sure how to reproduce it.

4ian:
I believe this is fixed, I remember testing and could not reproduce this. The way you reworked autocompletions means it should be fool proof there, because there is not a "fake" parser used when inserting completions - the completions now contain the position where they must be inserted :)

PR #2196, opened 2021-01-08
Fixed the issue Doubled extension prefix

INNOVATIVEGAMER:
Fixes issue #2195 .

### DIFF
--- a/newIDE/app/src/ExpressionAutocompletion/index.js
+++ b/newIDE/app/src/ExpressionAutocompletion/index.js
@@ -382,7 +382,15 @@ export const insertAutocompletionInExpression = (
   let newExpressionStart = expression.substring(0, wordStartPosition);
   const insertedWord = formatCompletion(maybeNextCharacter);
   const newExpressionEnd = expression.substring(wordEndPosition + 1);
-  if (insertedWord.indexOf(newExpressionStart) === 0) newExpressionStart = '';
+
+  //Code to merge common part between end of newExpressionStart and start of insertWord
+  for (let index = 0; index < newExpressionStart.length; index++) {
+    let subStr = newExpressionStart.substring(index);
+    if (insertedWord.indexOf(subStr) === 0) {
+      newExpressionStart = newExpressionStart.substring(0, index);
+      break;
+    }
+  }
 
   return {
     caretLocation: newExpressionStart.length + insertedWord.length,
--- a/newIDE/app/src/ExpressionAutocompletion/index.js
+++ b/newIDE/app/src/ExpressionAutocompletion/index.js
@@ -382,7 +382,7 @@ export const insertAutocompletionInExpression = (
   let newExpressionStart = expression.substring(0, wordStartPosition);
   const insertedWord = formatCompletion(maybeNextCharacter);
   const newExpressionEnd = expression.substring(wordEndPosition + 1);
-  if (insertedWord.includes(newExpressionStart)) newExpressionStart = '';
+  if (insertedWord.indexOf(newExpressionStart) === 0) newExpressionStart = '';
 
   return {
     caretLocation: newExpressionStart.length + insertedWord.length,
--- a/newIDE/app/src/ExpressionAutocompletion/index.js
+++ b/newIDE/app/src/ExpressionAutocompletion/index.js
@@ -379,9 +379,10 @@ export const insertAutocompletionInExpression = (
   // (to avoid repeating an existing character).
   const maybeNextCharacter: ?string = expression[wordEndPosition + 1];
 
-  const newExpressionStart = expression.substring(0, wordStartPosition);
+  let newExpressionStart = expression.substring(0, wordStartPosition);
   const insertedWord = formatCompletion(maybeNextCharacter);
   const newExpressionEnd = expression.substring(wordEndPosition + 1);
+  if (insertedWord.includes(newExpressionStart)) newExpressionStart = '';
 
   return {
     caretLocation: newExpressionStart.length + insertedWord.length,


newIDE/app/src/ExpressionAutocompletion/index.js
```diff
@@ -379,9 +379,10 @@ export const insertAutocompletionInExpression = (
   // (to avoid repeating an existing character).
   const maybeNextCharacter: ?string = expression[wordEndPosition + 1];
 
-  const newExpressionStart = expression.substring(0, wordStartPosition);
+  let newExpressionStart = expression.substring(0, wordStartPosition);
   const insertedWord = formatCompletion(maybeNextCharacter);
   const newExpressionEnd = expression.substring(wordEndPosition + 1);
+  if (insertedWord.includes(newExpressionStart)) newExpressionStart = '';
```
4ian, line 385: This seems dangerous to do that, you will erase half of the expression when completing something like `3 + MyExtension::xxx`.

4ian:
I think the current solution won't work as it could erase the start of the entire expression, can you try it on complex expressions?

INNOVATIVEGAMER:
Actually this code is working in various expressions to 3+ complexity.
This code is for autocompletion so it cant create adverse effects.
But since you mentioned a concern I updated a safer code.

If you suggest some complex test case which can fail this code , Please specify.

4ian:
Can you test if it works for this? 
image

INNOVATIVEGAMER:
Ok now I got the point.
Check the latest commit.
now its like 
ABC+BCD = ABCD
CDABAB+ABABOP = CDABABOP

INNOVATIVEGAMER:
@Bouh I think @4ian is busy these days. Can you please check this PR ?
:sweat:

4ian:
Unfortunately I'm still unsure about this approach. What we have here is a workaround checking if there is a duplication and if so, removing this duplication. I'm pretty sure this could introduce other edge cases.

Instead, you need to get a better understanding of the code insert the expression. Why is it working for objects for example (MySprite.PointX for example can properly autocomplete) but not for extensions? 
The reason is because `::` is considered as a separator, but the completion returns a whole entry like `MyExtension::MyFunction`. So when you have something like `MyExtension::My` to be completed, because of the `::` separator, the completion ends up as `MyExtension::MyExtension::MyFunction`.

We need to think how to solve that :)

INNOVATIVEGAMER:
Actually, In case of expressions which contains `::` like `MyExtension::MyFunction` this whole string comes from the autocompletion.
The string which comes from autocompletion is stored in `insertedString` so this function actually doing its work pretty well.

The other is `newExpressionStart` which is responsible for the previous string entered. This string takes the part of the expression until some separator is encountered. So if previous expression is `MyExtension::MyFunc` and then selected for autocompletion. `MyExtension::` is selected as `newExpressionStart`. so you can say that this function is also doing its job properly.
If you tell this function to add exceptional case for `:` or `::` , this will instead create edge cases and somewhat not a universal solution.

Removing duplication can be considered as universal solution as there may be more cases in future where expressions like `MyExtension::MyFunction` come with other separators so adding each time a exception in these functions for cases like these is not a solution.

So I think ,this piece of code actually cleans the expressions before returning.

4ian:
> So I think ,this piece of code actually cleans the expressions before returning.

Sure, but we should not even need to clean anything :) Let me rephrase the problem:

The full returned expression is:
`newExpressionStart + insertedWord + newExpressionEnd`

insertedWord is returning result of the call to `formatCompletion`, which will return something like `MyExtension::MyFunction`.
Either we should:
- Change formatCompletion to return `MyFunction` instead of `MyExtension::MyFunction`
- Or change `newExpressionStart` so that it does not contain `MyExtension::` (i.e: `wordStartPosition` should be a position that stops before `MyExtension::`).
 
I'm not sure which of these 2 approaches (changing formatCompletion or changing newExpressionStart) is better, but I was thinking of the second one (hence why I was speaking about changing `::` so that it's not considered as a separator in the case of a function call).

INNOVATIVEGAMER:
`insertedWord` is the result of `formatCompletion` (The string is actually that string which get selected from the dropdown).
so changing this function is not at all suitable.

As per the `newExpressionStart` which is the string previously in the expression.
Now as I mentioned `wordStartPosition` is determined by checking the expression value until a separator is encountered and thus removing that part .

1. If we add a exception for `::` here `wordStartPosition` will never detect this as separator, this can create errors.
  example : expression = text1+text2::
                  selected word from dropdown = myFunction
                  now in this case `wordStartPosition` will detect `newExpressionStart` as text1+ instead of text1+text2::
                   so result will be => text1+myFunction
2. Another problem is that making `::` as a exception is not a universal solution.
  There may be possibilities that selected word from dropdown will be : 
      `myFunction:myExpression`,
      `myFunction_myExpression`
  In these cases you will need to add exception for those also.

INNOVATIVEGAMER:
Current functions are doing their job properly 
1.Getting the word from dropdown as `insertedWord`
2.Getting the previously entered string as 'newExpressionStart'
Maybe this 2nd function is not working in the way we want but combining this function with cleanup code make it complete.

INNOVATIVEGAMER:
So @4ian what do you think now ?
Do you prefer this code ? Or 
Do you still want to make changes as you mentioned earlier ?

4ian:
> insertedWord is the result of formatCompletion (The string is actually that string which get selected from the dropdown).
so changing this function is not at all suitable.

Why is this not suitable? We could fix formatCompletion if it's returning `MyExtension::MyFunction` whereas just `MyFunction` would be more appropriate :)

I don't want to go ahead with a "replace" based solution, as it's a work around. The current completions are working well for objects (Typing `MyObject.` then choosing a completion will work) so I'd like to fix the root cause of the issue of why it's not working for extensions, rather than doing a string replacement.

INNOVATIVEGAMER:
> Why is this not suitable?

`insertedString` is not the result of `formatCompletion` fully. `formatCompletion` just completes the format of `insertedString` string.
`insertedString` gets its string from `insertedAutocompletion.completion` . This is the string which we select from dropdown.
and I said changing this is not suitable because the **`insertedString.completion` should be the resultant string which get selected from dropdown irrespective of which previously in the value field.**
> The current completions are working well for objects (Typing MyObject. then choosing a completion will work)

This is because no object from dropdown/autocompletion is as `myobject.value` they are just one word strings.
For objects the autocompletion is either be a function or some value in one word string.

> I'd like to fix the root cause of the issue

If you look at the scenario , there is no root cause. Every function is doing their job properly. There is no bug.
**The issue is we want this functionality => what value previously entered in the field , keeping in mind that complete the word.**
So for that functionality we need to check what is previously entered in the field and what autocompletion is providing from dropdown and this is what my function is doing.
@4ian if you still find this solution wrong you can close this PR.
I will work on some other issues/features in GD. :smile:

4ian:
Feel free to close the PR and work on something else! ;)

PR #2199, opened 2021-01-10
Added a new preference to toggle the visibility of menu bar in preview windows

### LINKED ISSUE #2190: Add a preference to show the menu bar in previews
... and by default hide it in the preview window :) 

This will avoid issues with the menu bar being shown and distorting the image for games that have a need for a specific resolution. It's also looking more like a real export.
For people wanting to use the console/debugger, they can go in the preferences to enable the menu.

Technically:

* To hide the menu bar: `browserWindow.setMenu(null)` should work
* A new preference should be added in `PreferencesContext`/`PreferencesProvider` and a toggle in the preferences dialog.

rs4231199:
fixes #2190

[win.setMenuBarVisibility(visible)](https://www.electronjs.org/docs/api/browser-window#winsetmenubarvisibilityvisible-windows-linux) is used to hide the menu bar. This works as expected on Windows and should also work fine on Linux. I need help to test this on macOS.

If preference `Hide menubar in preview window` is set true, then menu bar will be will hidden when the preview is launched. And if preference is false, then the menu bar will be visible.

image

### DIFF
--- a/newIDE/app/src/Export/LocalExporters/LocalPreviewLauncher/index.js
+++ b/newIDE/app/src/Export/LocalExporters/LocalPreviewLauncher/index.js
@@ -38,7 +38,7 @@ type State = {|
     title: string,
     backgroundColor: string,
   },
-  hideMenubar: boolean,
+  hideMenuBar: boolean,
 |};
 
 export default class LocalPreviewLauncher extends React.Component<
@@ -57,6 +57,7 @@ export default class LocalPreviewLauncher extends React.Component<
     devToolsOpen: false,
     previewBrowserWindowConfig: null,
     hotReloadsCount: 0,
+    hideMenuBar: true,
   };
   _networkPreviewSubscriptionChecker: ?SubscriptionChecker = null;
   _hotReloadSubscriptionChecker: ?SubscriptionChecker = null;
@@ -70,8 +71,8 @@ export default class LocalPreviewLauncher extends React.Component<
       return;
 
     const win = new BrowserWindow(this.state.previewBrowserWindowConfig);
-    win.setMenuBarVisibility(this.state.hideMenubar);
     win.loadURL(`file://${this.state.previewGamePath}/index.html`);
+    win.setMenuBarVisibility(this.state.hideMenuBar);
     win.webContents.on('devtools-opened', () => {
       this.setState({ devToolsOpen: true });
     });
@@ -99,7 +100,7 @@ export default class LocalPreviewLauncher extends React.Component<
           },
         },
         previewGamePath: gamePath,
-        hideMenubar: !options.getIsMenubarHiddenInPreview(),
+        hideMenuBar: !options.getIsMenuBarHiddenInPreview(),
       },
       () => {
         if (!options.networkPreview) {
--- a/newIDE/app/src/Export/LocalExporters/LocalPreviewLauncher/index.js
+++ b/newIDE/app/src/Export/LocalExporters/LocalPreviewLauncher/index.js
@@ -38,6 +38,7 @@ type State = {|
     title: string,
     backgroundColor: string,
   },
+  hideMenubar: boolean,
 |};
 
 export default class LocalPreviewLauncher extends React.Component<
@@ -69,6 +70,7 @@ export default class LocalPreviewLauncher extends React.Component<
       return;
 
     const win = new BrowserWindow(this.state.previewBrowserWindowConfig);
+    win.setMenuBarVisibility(this.state.hideMenubar);
     win.loadURL(`file://${this.state.previewGamePath}/index.html`);
     win.webContents.on('devtools-opened', () => {
       this.setState({ devToolsOpen: true });
@@ -97,6 +99,7 @@ export default class LocalPreviewLauncher extends React.Component<
           },
         },
         previewGamePath: gamePath,
+        hideMenubar: !options.getIsMenubarHiddenInPreview(),
       },
       () => {
         if (!options.networkPreview) {
--- a/newIDE/app/src/MainFrame/Preferences/PreferencesDialog.js
+++ b/newIDE/app/src/MainFrame/Preferences/PreferencesDialog.js
@@ -45,6 +45,7 @@ const PreferencesDialog = ({ i18n, onClose }: Props) => {
     setAutoOpenMostRecentProject,
     resetShortcutsToDefault,
     setShortcutForCommand,
+    setIsMenubarHiddenInPreview,
   } = React.useContext(PreferencesContext);
 
   return (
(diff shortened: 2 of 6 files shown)


4ian:
Interesting! The changes around the preferences look good.

Reading the doc of `setMenuBarVisibility(visible)`, I see:
"If the menu bar is auto-hide, users can still bring up the menu bar by pressing the single Alt key."

This might be surprising for the user, should we entirely remove the menu? 
@silver-streak @Bouh you both discussed this on the forum, what do you think we should do?

newIDE/app/src/MainFrame/Preferences/PreferencesDialog.js
```diff
@@ -246,6 +247,14 @@ const PreferencesDialog = ({ i18n, onClose }: Props) => {
               }
             />
           </Line>
+          <Line>
+            <Toggle
+              onToggle={(e, check) => setIsMenubarHiddenInPreview(check)}
+              toggled={values.isMenubarHiddenInPreview}
+              labelPosition="right"
+              label={<Trans>Hide menubar in preview window</Trans>}
```
4ian, line 255: `Hide the menu in the preview window`

4ian: (or "Hide the menu bar in the preview window")
https://en.wikipedia.org/wiki/Menu_bar#:~:text=A%20menu%20bar%20is%20a,displaying%20help%20documentation%20or%20manuals.

newIDE/app/src/Export/LocalExporters/LocalPreviewLauncher/index.js
```diff
@@ -38,6 +38,7 @@ type State = {|
     title: string,
     backgroundColor: string,
   },
+  hideMenubar: boolean,
```
4ian, line 41: Nitpicking: Everywhere in the code, this should be `MenuBar` (with a capital B). Menubar is not a word, it's two different words.

arthuro555:
https://github.com/4ian/GDevelop/blob/099e9cf936e75a8fc22f5578aa8aa5062c3c3b75/newIDE/app/src/Export/LocalExporters/LocalPreviewLauncher/index.js#L51-L60

You probably want to set an initial state for `hideMenubar` there 👀

Bouh:
> Interesting! The changes around the preferences look good.
> 
> Reading the doc of `setMenuBarVisibility(visible)`, I see:
> "If the menu bar is auto-hide, users can still bring up the menu bar by pressing the single Alt key."
> 
> This might be surprising for the user, should we entirely remove the menu?
> @Silver-Streak @Bouh you both discussed this on the forum, what do you think we should do?

Using alt can be not a great choice,it can be also anoying when the user use the alt key for an action in their own game.
It's ok to remove it completely. The user have made the choice of disable it in the peference, in logic he know what are the consequences. Maybe add a hint for explain it can be hard to escape the game if this one is in fullscreen without Menu Bar.

Silver-Streak:
> Interesting! The changes around the preferences look good.
> 
> Reading the doc of `setMenuBarVisibility(visible)`, I see:
> "If the menu bar is auto-hide, users can still bring up the menu bar by pressing the single Alt key."
> 
> This might be surprising for the user, should we entirely remove the menu?
> @Silver-Streak @Bouh you both discussed this on the forum, what do you think we should do?

I agree with @Bouh. Devs may use ALT as a game control. If the dev has disabled the preview in the menu, they should be aware of the consequences. At minimum the option could always have a "(Previews can still be exited with ALT+F4 even when fullscreen)" text or something similar as part of the toggle text.

newIDE/app/src/Export/LocalExporters/LocalPreviewLauncher/index.js
```diff
@@ -70,6 +72,7 @@ export default class LocalPreviewLauncher extends React.Component<
 
     const win = new BrowserWindow(this.state.previewBrowserWindowConfig);
     win.loadURL(`file://${this.state.previewGamePath}/index.html`);
+    win.setMenuBarVisibility(this.state.hideMenuBar);
```
rs4231199, line 75: In previous commit, this line was at line number 74(that means right above the `win.loadURL(.....);`). If I keep it there, this is the error I get after executing `npm run flow`.
 ```
Error ---------------------------------------------------- src/Export/LocalExporters/LocalPreviewLauncher/index.js:75:27

Cannot coerce `this.state.previewGamePath` to string because:
 - null or undefined [1] should not be coerced.
 - null [2] should not be coerced.

   src/Export/LocalExporters/LocalPreviewLauncher/index.js:75:27
   75|     win.loadURL(`file://${this.state.previewGamePath}/index.html`);
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^

References:
   src/Export/LocalExporters/LocalPreviewLauncher/index.js:32:20
   32|   previewGamePath: ?string,
                          ^^^^^^^ [1]
   src/Export/LocalExporters/LocalPreviewLauncher/index.js:56:22
   56|     previewGamePath: null,
                            ^^^^ [2]
```
Why do I not get the same error now?

**Comment by rs4231199** [comment]
> Reading the doc of `setMenuBarVisibility(visible)`, I see:
> "If the menu bar is auto-hide, users can still bring up the menu bar by pressing the single Alt key."
> 
> This might be surprising for the user, should we entirely remove the menu?

I haven't set the `autoHideMenuBar` as true. So, Alt key does nothing when menu bar is hidden. As suggested by @Silver-Streak and @Bouh, I will keep it so.  


> At minimum the option could always have a "(Previews can still be exited with ALT+F4 even when fullscreen)" text or something similar as part of the toggle text.

F11(Toggle Full Screen key) still works. Should I add this to the toggle text?

File: newIDE/app/src/Export/LocalExporters/LocalPreviewLauncher/index.js (hunk 3 above)
Comment (4ian) [reply newIDE/app/src/Export/LocalExporters/LocalPreviewLauncher/index.js:75]: This is because Flow consider that the call to `win.setMenuBarVisibility` could potentially modify `this.state` and set the variable to `null`. To avoid this, we usually do:

```
// Store the value in a variable
const { previewGamePath } = this.state;
if (!previewGamePath) return; // Check that it's not null (or empty)

// Call any function now

// Here you're 100% sure that previewGamePath is not null (or empty).
// BUT this.state.previewGamePath could have been modified by the previous lines.
// That's why Flow is triggering a potential error if you try to use it.
win.loadURL(`file://${previewGamePath}/index.html`);
```

4ian:
> I haven't set the autoHideMenuBar as true. So, Alt key does nothing when menu bar is hidden

Nice, I missed this. Looks good then!

4ian:
Thanks for this nice PR, this will be much better now for most users!

Issue #2202, opened 2021-01-10
Add Shape Painter expressions and actions


tristanbob:
## Description

Shape Painter objects are missing some actions and expressions that would be useful.

### New Actions:
 - [ ] (Setup) "Shape position relative to object position" (bool).  
   - This should be similar to how "Clear between frames" is done

![Screenshot 2021-01-10 122640](https://user-images.githubusercontent.com/8879811/104133487-2691c880-5341-11eb-9a31-ebb074d2d601.jpg)

### New Expressions:
 - [ ] Outline size (number)
 - [ ] Outline color (string)
 - [ ] Outline opacity (number)
 - [ ] Fill color (string)
 - [ ] Fill opacity (number)
 - [ ] Clear between frames (bool)
 - [ ] Shape position relative to object position (bool)

![Screenshot 2021-01-10 121827](https://user-images.githubusercontent.com/8879811/104133488-2b567c80-5341-11eb-8d88-935062774c71.jpg)

tristanbob:
@Bouh This is what you asked me to tag you in.  :)

Bouh:
Thank you for this list i'm on it!

PR #2203, opened 2021-01-10
Add actions, conditions, expressions for shape painter

Bouh:
Closing #2202 

### New Actions:
 - [x] Change position relative (yes/no)

### New Conditions:
 - [x] Is relative (Invert the condition for get `Is absolute`)

### New Expressions:
 - [x] Outline size (number)
 - [x] Outline opacity (number)
 - [x] Fill opacity (number)
 - [x] Outline color
   - [x] getOutlineColorR() (number)
   - [x] getOutlineColorG() (number)
   - [x] getOutlineColorB() (number)
 - [x] Filling color
   - [x] getFillColorR() (number)
   - [x] getFillColorG() (number)
   - [x] getFillColorB() (number)

### Expressions:
 - Clear between frames (bool) (Not possible until we have a complete boolean support in the engine)
 - Shape position relative (bool) (Not possible until we have a complete boolean support in the engine)
 - Shape position absolute (bool) (There is no mention of absolute in the IDE for the user, let keep this simple and use invert of relative instead, and bool not possible)
### Action:
 - Is absolute (Not done, in the properties of the object we talk about Relative, if we want test Absolute it's just the invert of the relative condition)


4ian:
> I think return just the color for red, and for green and for blue separately.

Yep! 

> I think we can simplify the part with absolute/relative, what you think?

I'll take a look at the code :)

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -366,6 +366,16 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
       .SetFunctionName("IsClearedBetweenFrames")
       .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
 
+  obj.AddExpression(
+        "IsClearBetweenFrames",
+        _("Is clear between frames"),
+        _("Return true if the object is clear between two frames."),
```
4ian, line 372: Everywhere: "Is cleared between frames" ("ed" to Clear)

Also, avoid "Return true" which is a very "programming language" wording. Instead, use "Check": 
"Check if the object is cleared between each frame."

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -419,6 +429,16 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
       .SetFunctionName("GetOutlineSize")
       .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
 
+  obj.AddExpression(
+        "OutlineSize",
+        _("Get current outline size"),
```
4ian, line 434: Just name this expression "Outline size".

When you're browsing expressions, you now that you'll "get the current" value. So no need for "Get current ...". Just state what the expression is returning: "Outline size"

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -448,6 +468,16 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
       .SetFunctionName("GetFillOpacity")
       .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
 
+  obj.AddExpression(
+        "FillOpacity",
+        _("Get current opacity of filling"),
```
4ian, line 473: Opacity of filling

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -477,6 +507,84 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
       .SetFunctionName("GetOutlineOpacity")
       .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
 
+  obj.AddExpression(
+        "OutlineOpacity",
+        _("Get current opacity of the outline"),
```
4ian, line 512: `Outline opacity`

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -477,6 +507,84 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+        _("Return the opacity of the outline (a number from 0 to 1, 1 being fully opaque)."),
+        "",
+        "res/conditions/opacity.png")
+      .AddParameter("object", _("Shape Painter object"), "Drawer")
+      .SetFunctionName("GetOutlineOpacity")
+      .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
+
+  obj.AddAction("CoordinatesAbsolute",
+                _("Coordinates Absolute"),
```
4ian, line 521: `Use absolute coordinates`

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -477,6 +507,84 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+                _("Change the coordinate type the shape painter object to absolute."),
```
4ian, line 522: `Set the coordinates of the Shape Painter to be absolute`

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -477,6 +507,84 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+        "OutlineOpacity",
+        _("Get current opacity of the outline"),
+        _("Return the opacity of the outline (a number from 0 to 1, 1 being fully opaque)."),
+        "",
+        "res/conditions/opacity.png")
+      .AddParameter("object", _("Shape Painter object"), "Drawer")
+      .SetFunctionName("GetOutlineOpacity")
+      .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
+
+  obj.AddAction("CoordinatesAbsolute",
```
4ian, line 520: `UseAbsoluteCoordinates`

(action => always use a verb, and adjectives before names in English)

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -477,6 +507,84 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+                _("Change the coordinate type the shape painter object to absolute."),
+                _("Change the coordinate for _PARAM0_ to absolute"),
+                _("Setup"),
+                "res/actions/position24.png",
+                "res/actions/position.png")
+      .AddParameter("object", _("Shape Painter object"), "Drawer")
+      .SetFunctionName("SetCoordinatesAbsolute")
+      .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
+
+  obj.AddAction("CoordinatesRelative",
```
4ian, line 531: Maybe we can have a single action with a yes/no parameter?

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -477,6 +507,84 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+          _("Setup"),
+              "res/conditions/position24.png",
+              "res/conditions/position.png")
+
+        .AddParameter("object", _("Shape Painter object"), "Drawer")
+        .SetFunctionName("AreCoordinatesAbsolute")
+        .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
+
+    obj.AddCondition(
+          "AreCoordinatesRelative",
```
4ian, line 556: Maybe for the sake of simplicity we can keep only the first condition here too?

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -477,6 +507,84 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+
+    obj.AddExpression(
+          "AreCoordinatesRelative",
+          _("Coordinates relative"),
+          _("Return if the coordinates of the shape painter is relative."),
+          "",
+          "res/conditions/position.png")
+        .AddParameter("object", _("Shape Painter object"), "Drawer")
+        .SetFunctionName("AreCoordinatesRelative")
+        .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
```
4ian, line 586: For now remember we don't have booleans expressions. Sure JavaScript will allow you to do that, but this will come back to bite us very heard with strange bugs.
Let's remove these for now, conditions are here for checking booleans.

Extensions/PrimitiveDrawing/shapepainterruntimeobject.ts
```diff
@@ -274,6 +274,22 @@ namespace gdjs {
...
+      this._absoluteCoordinates = !value;
+    }
+
+    areCoordinatesAbsolute(): boolean {
+      return this._absoluteCoordinates;
+    }
+
+    areCoordinatesRelative(): boolean {
+      return !this._absoluteCoordinates;
+    }
```
4ian, line 291: Let's keep the interface simple by just having a single pair of these (the same one as the conditions/actions, for consistency).

Extensions/PrimitiveDrawing/Extension.cpp
Bouh: Ok expression for IsClearBetweenFrames, AreCoordinatesRelative, AreCoordinatesAbsolute are removed.

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -379,6 +379,37 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
       .SetFunctionName("SetFillColor")
       .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
 
+  obj.AddExpression(
+        "GetFillColorR",
+        _("Get red filling colors"),
```
4ian, line 384: `Filling color red component`

Bouh:
> > I think return just the color for red, and for green and for blue separately.
> 
> Yep!

I've found no consensus about colors. Then i've made something, it works and return the red/green/blue colors as number.

For red fill and outline: 
image

The debugger show the correct value.
image

Even if i found this expressions better:
`object.getColor("RGB")` (string)
`object.getColor("Red")` (number)
`object.getColor("Green")` (number)
`object.getColor("Blue")` (number)

Bouh:
In object properties we talk about `Relative`.

>Draw the shapes relative to the object position on the scene

Let's keep all conditions and actions with `Relative`.
`Absolute` is mentioned nowhere for the user.
For get `Absolute` we have just to use an inverted condition of `Relative`.

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -379,6 +379,37 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
       .SetFunctionName("SetFillColor")
       .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
 
+  obj.AddExpression(
+        "GetFillColorR",
+        _("Filling color red component"),
+        _("Current value for red filling colors."),
```
4ian, line 385: Let's just copy the name in the description to reduce the number of translations to be done:
`Filing color red component`

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -379,6 +379,37 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
       .SetFunctionName("SetFillColor")
       .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
 
+  obj.AddExpression(
+        "GetFillColorR",
```
4ian, line 383: Sorry I missed this earlier: there is never a `Get` for expressions, because they all are getters.
Also, let's avoid abbreviations: `FillColorRed`

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -419,6 +480,16 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
       .SetFunctionName("GetOutlineSize")
       .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
 
+  obj.AddExpression(
+        "OutlineSize",
+        _("Outline size"),
+        _("Check the outline size in pixels."),
```
4ian, line 486: This is an expression, so it's not "checking" anything :) 
To avoid losing time let's just copy paste the same thing as the name: "Outline size"

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -477,6 +558,53 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
       .SetFunctionName("GetOutlineOpacity")
       .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
 
+  obj.AddExpression(
+        "OutlineOpacity",
+        _("Outline opacity"),
+        _("Check if the opacity of the outline (a number from 0 to 1, 1 being fully opaque)."),
```
4ian, line 564: Same

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -477,6 +558,53 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+                _("Change the coordinate for _PARAM0_ to absolute"),
+                _("Setup"),
+                "res/actions/position24.png",
+                "res/actions/position.png")
+      .AddParameter("object", _("Shape Painter object"), "Drawer")
+      .SetFunctionName("SetCoordinatesAbsolute")
+      .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
+
+  obj.AddAction("UseCoordinatesRelative",
+                  _("Use coordinates Relative"),
```
4ian, line 583: Use relative coordinates

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -477,6 +558,53 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+        "OutlineOpacity",
+        _("Outline opacity"),
+        _("Check if the opacity of the outline (a number from 0 to 1, 1 being fully opaque)."),
+        "",
+        "res/conditions/opacity.png")
+      .AddParameter("object", _("Shape Painter object"), "Drawer")
+      .SetFunctionName("GetOutlineOpacity")
+      .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
+
+  obj.AddAction("UseAbsoluteCoordinates",
```
4ian, line 571: No need for this action

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -477,6 +558,53 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+
+  obj.AddAction("UseCoordinatesRelative",
+                  _("Use coordinates Relative"),
+                  _("Set the coordinates of the Shape Painter to be relative"),
+                  _("Change the coordinate for _PARAM0_ to relative"),
+                  _("Setup"),
+                  "res/actions/position24.png",
+                  "res/actions/position.png")
+        .AddParameter("object", _("Shape Painter object"), "Drawer")
+        .AddParameter("yesorno", _("Use coordinates Relative?"), "", false)
```
4ian, line 590: "Use relative coordinates?"

Extensions/PrimitiveDrawing/JsExtension.cpp
```diff
@@ -111,33 +111,64 @@ class PrimitiveDrawingJsExtension : public gd::PlatformExtension {
...
     GetAllConditionsForObject(
         "PrimitiveDrawing::Drawer")["PrimitiveDrawing::OutlineOpacity"]
         .SetFunctionName("getOutlineOpacity");
+    GetAllExpressionsForObject("PrimitiveDrawing::Drawer")["OutlineOpacity"]
+        .SetFunctionName("getOutlineOpacity");
+
+    GetAllActionsForObject(
+        "PrimitiveDrawing::Drawer")["PrimitiveDrawing::CoordinatesRelative"]
+        .SetFunctionName("setCoordinatesRelative")
+        .SetGetter("areCoordinatesRelative");
```
4ian, line 168: I don't think there is a need for a getter

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -477,6 +558,53 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+                _("Set the coordinates of the Shape Painter to be absolute"),
+                _("Change the coordinate for _PARAM0_ to absolute"),
+                _("Setup"),
+                "res/actions/position24.png",
+                "res/actions/position.png")
+      .AddParameter("object", _("Shape Painter object"), "Drawer")
+      .SetFunctionName("SetCoordinatesAbsolute")
+      .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
+
+  obj.AddAction("UseCoordinatesRelative",
```
4ian, line 582: UseRelativeCoordinates

Remember: the adjective is almost always *before* the noun in English.

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -477,6 +558,53 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+                  _("Use coordinates Relative"),
+                  _("Set the coordinates of the Shape Painter to be relative"),
+                  _("Change the coordinate for _PARAM0_ to relative"),
```
4ian, line 585: The sentence does not have all parameters:

`Use relative coordinates for _PARAM0_: _PARAM1_`

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -477,6 +558,53 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+                _("Setup"),
+                "res/actions/position24.png",
+                "res/actions/position.png")
+      .AddParameter("object", _("Shape Painter object"), "Drawer")
+      .SetFunctionName("SetCoordinatesAbsolute")
+      .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
+
+  obj.AddAction("UseCoordinatesRelative",
+                  _("Use coordinates Relative"),
+                  _("Set the coordinates of the Shape Painter to be relative"),
```
4ian, line 584: "Set if the object should use relative coordinates (by default) or not. It's recommended to use relative coordinates."

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -392,6 +423,36 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
       .SetFunctionName("SetOutlineColor")
       .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
 
+  obj.AddExpression(
+        "GetOutlineColorR",
+        _("Outline color red component"),
+        _("Current value for red outline colors."),
```
4ian, line 429: Same thing for all of these, let's put the same name and description. 
It's too much work to translate different descriptions for something that is almost similar to the name.

4ian (commented):
Sorry for the delay!

Bouh:
Et voilà!
Conditions/actions/expressions work correctly, they are tested with the debugger.

4ian:
Looks good! Thanks! 👍

PR #2207, opened 2021-01-13
Export sourcemaps with the code

arthuro555:
Exports sourcemaps alongside the normal code when a sourcemap is present.

### DIFF
--- a/GDJS/GDJS/IDE/Exporter.cpp
+++ b/GDJS/GDJS/IDE/Exporter.cpp
@@ -114,7 +114,7 @@ bool Exporter::ExportWholePixiProject(
 
     // Copy all dependencies and the index (or metadata) file.
     helper.RemoveIncludes(false, true, includesFiles);
-    helper.ExportIncludesAndLibs(includesFiles, exportDir);
+    helper.ExportIncludesAndLibs(includesFiles, exportDir, false);
 
     gd::String source = gdjsRoot + "/Runtime/index.html";
     if (exportForCordova)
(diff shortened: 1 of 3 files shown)


4ian:
Thanks! I see two potential concerns with this:
- this should not be done when exporting not for a preview (I don't think it's handled right now?).
- can we try on the web-app what's the result of this change, both for export and for preview? This will issue calls to `copyFile`, which will be:
  - This one for the export: https://github.com/4ian/GDevelop/blob/9bcac613ed6cc11555dcd14bb58d08cb4790c1e7/newIDE/app/src/Export/BrowserExporters/BrowserFileSystem.js#L162 - but there won't be any source map in exports anyway (though it would work, much like traditional js files)
  - This one for the preview: https://github.com/4ian/GDevelop/blob/9bcac613ed6cc11555dcd14bb58d08cb4790c1e7/newIDE/app/src/Export/BrowserExporters/BrowserS3FileSystem.js#L47 - which does not copy files when they are external URLs (which is the case for all the runtime/extension files, so this includes the .map files).
  
To fully test it, you would need to be able to push your GDJS runtime changes to the S3 hosting it. While you can't do it, you can still try to export (verify it still works) and preview (verify that it works, and that there are 404 when the debugger tries to load source maps :))

arthuro555:
> To verify that it works, and that there are 404 when the debugger tries to load source maps :))

Both cases (web preview and web build) aren't loading the sourcemaps. I think that's the intended behavior as the js files with the comment that asks the sourcemaps to be loaded haven't been published on the server?

4ian:
Yeah might be the case! I would need to push temporarily the runtime to check the result (using a new version number to avoid disturbing the existing version). Let me merge and check that, not to hard to revert in case it's causing a big issue.

4ian:
Seems that it works as intended: https://game-previews.gdevelop-app.com/1610924747255-636640/index.html :) 
Note though that strangely the place where I can't get it to work is... on the electron app version. It shows "Source Map Detected" on the top, but is not showing the source map in the inspector file tree. Weird? Might be because of an "old" Electron version that we're using?

arthuro555:
Huh that's weird I think it was working for me. Though now that you mention it, I had forgotten but I think the file path didn't quite match the file path of the local game so the original files showed up as other files in the source files list.

PR #2215, opened 2021-01-15
Toggle Buttons in SceneEditor Toolbar

INNOVATIVEGAMER:
1.Following buttons in toolbar provided with toggle functionality.
2.Buttons BG changes depending upon toggle state of the subEditor.

Current Problems : 
1.Clicking on the close button in subEditor do not changes the subEditor state.
https://github.com/4ian/GDevelop/blob/d21a8182233356d769b4f1c3da4edde76a776d0f/newIDE/app/src/UI/EditorMosaic/index.js#L116
(As close button mentioned only in the object-list editor as separate attribute and for other subEditors as a default.
So need some help how close button can reference state of sceneEditor)

Example of how scene editor looks now : 
GDevelop 5 - example___platformer - Google Chrome 15-01-2021 20_17_03

Preview Video : 
https://streamable.com/nykuzj


INNOVATIVEGAMER:
@Bouh As we discussed on forum. I converted these buttons to toggle and added some styling depending upon the toggle state.
Now as you are saying in forum we can work on it further in this PR with discussions.

Bouh:
I'am not use at all for the background.
The toggling seems to work fine, have you tried to toggle a panel that has been moved?

Bouh:
https://github.com/4ian/GDevelop/blob/d21a8182233356d769b4f1c3da4edde76a776d0f/newIDE/app/src/UI/EditorMosaic/index.js#L116

- I think you can add a new props in <CloseButton>. 
And use the props in onClick ( where `mosaicActions.remove(mosaicWindowActions.getPath());` is used)

You have twice `<CloseButton>` in the codebase, be sure to edit both.
In onClick event you must call a function for turn off the state of the panel.

But i don't know how know which editor it is.
There is no types for the editors/panels from GD to Mosaic.
There is only a title, but we shouldn't use it.


- For the background color it is not conclusive and totally breaks the immersion.
Let's remove the css code and keep the visual style of the buttons as before.


C:\GDevelop\newIDE\app\src\UI\EditorMosaic\CloseButton.js

INNOVATIVEGAMER:
> * For the background color it is not conclusive and totally breaks the immersion.
>   Let's remove the css code and keep the original visual style of the buttons.

Yes, I just placed the background for temporary with colors.
Do you have any suggestions for on/off state styling ?
Or shall I show you some examples and then we will discuss on them ?

Bouh:
> Or shall I show you some examples and then we will discuss on them ?

Material-ui have no recommendation for this kind of button with a big icon with a big filling.
To do add a background related to the state we need to remade the icons, which is not at all planned and would require the work of a real UI designer.


>Do you have any suggestions for on/off state styling ?

Keep the original visual style of the buttons, remove all css added in this PR.

INNOVATIVEGAMER:
> But i don't know how know which editor it is.

I used a small trick here by passing a function as a prop to close button which changes the editor state. ;)

I have fixed another issue in the scene editor which I came across while using these toggles,
The issue is when we left some subEditors open/close, subEditors state did not get updated in state of the `SceneEditor` when `SceneEditor` mounts on the screen. 
Previously the `EditorMosiac` handles this issue (by luck) with `initialNodes`. Because previously it cares only about the nodes it gets as a prop but now as `SceneEditor` state is responsible for rendering of various editors its state should be updated as per the `initialNodes` before it mounts.

Now `SceneEditors` state is truly responsible for the rendering of different subEditors on startup and in between.

INNOVATIVEGAMER:
https://github.com/4ian/GDevelop/blob/98a1d69489c5410d28617f11686351e093157c5d/newIDE/app/src/SceneEditor/index.js#L1208-L1246

I think these messages saying panel is already opened are not useful anymore as no one is able to open a panel that is already opened. User either will open/close by clicking on toggle.

> To do add a background related to the state we need to remade the icons, which is not at all planned and would require the work of a real UI designer.

I think this is a good plan to go forward on this feature. But please be sure to tell the users about toggle functionality or no one will ever notice it unless they accidently double click on any button.

!Warning!
Users need to wait for some milliseconds while toggling the editors , currently it is not superfast as react takes its time while changing the state. But I think it is not a measure issue for now :).

INNOVATIVEGAMER:
@4ian this PR is ready now. Please review it so i can make PR on other issues.
Also check my comment about removing those "panel is already open" messages.

INNOVATIVEGAMER:
@4ian & @Bouh this PR is ready now. Can you Please review it, so I can make PR on other issues.
Also check my comment about removing those "panel is already open" messages.

newIDE/app/src/UI/ToolbarIcon.js
```diff
@@ -11,6 +11,7 @@ type Props = {|
   disabled?: boolean,
   onClick?: () => void,
   onContextMenu?: () => void,
+  showComponent?: boolean,
```
Bouh, line 14: It was used for styles, as we don't want style remove this props everywhere.

newIDE/app/src/UI/ToolbarIcon.js
```diff
@@ -26,6 +27,23 @@ const ToolbarIcon = React.forwardRef<Props, IconButton>((props: Props, ref) => {
...
+  //         backgroundColor: '#80BE1F',
+  //         marginRight: '5px',
+  //       }
+  //     : {
+  //         border: '5px solid #FF3232',
+  //         backgroundColor: '#FF3232',
+  //         marginRight: '5px',
+  //       };
+  // }
+
```
Bouh, line 46: Remove this useless code please.

newIDE/app/src/SceneEditor/Toolbar.js
```diff
@@ -38,9 +38,9 @@ export class Toolbar extends PureComponent<Props> {
     return (
       <>
         <ToolbarCommands
-          openObjectsList={this.props.openObjectsList}
-          openObjectGroupsList={this.props.openObjectGroupsList}
-          openPropertiesPanel={this.props.openProperties}
+          toggleObjectsList={this.props.toggleObjectsList}
+          toggleObjectGroupsList={this.props.toggleObjectGroupsList}
+          togglePropertiesPanel={this.props.toggleProperties}
```
Bouh, line 43: ToolbarCommands component is used for bind ab action to shortcuts.
In your case this code isn't complete, because props in ToolbarCommands are not complete.

See if you press O for open the object editor the app should crash.

newIDE/app/src/UI/EditorMosaic/CloseButton.js
```diff
@@ -24,6 +28,7 @@ export default class CloseButton extends Component {
             {({ mosaicWindowActions }) => (
               <IconButton
                 onClick={() => {
+                  if (this.props.closeActions) this.props.closeActions();
                   mosaicActions.remove(mosaicWindowActions.getPath());
```
Bouh, line 32: See in `newIDE\app\src\UI\EditorMosaic\index.js` you have the current stuff for add a node in mosaic with `addNode`.
It would be better to add a similar for removeNode. not sure of the best way to do it thought.

Bouh:
The first part of info bar can be edited for explain that panels are toggleable.

newIDE/app/src/UI/EditorMosaic/CloseButton.js
INNOVATIVEGAMER: React recommends toggling of components through react state. As you can see , Material-UI also follows this convention. This way we can use this state anywhere in component and other components.  
`mosiacAction.remove(path)` is already doing the job of removal of node.
The code in `componentWillMount` I wrote only for the first render. That is when project loads for the first time.
By doing this we will set the sceneEditor state of the panels as they were before closing the project.

INNOVATIVEGAMER: React recommends toggling of components through react state. As you can see , Material-UI also follows this convention. This way we can use this state anywhere in component and other components.  
`mosiacAction.remove(path)` is already doing the job of removal of node.
The code in `componentWillMount` I wrote only for the first render. That is when project loads for the first time.
By doing this we will set the sceneEditor state of the panels as they were before closing the project.

Another example of state use is as you can see here , we have used state to display message : 
https://github.com/4ian/GDevelop/blob/98a1d69489c5410d28617f11686351e093157c5d/newIDE/app/src/SceneEditor/index.js#L1208-L1216

So I don't think removing state is a good idea in case of toggle components. But You gave a solid idea of creating this `removeNode` function. Now speed of toggling of panels can be reduced by half. 
Thank you for your hint. In the next commit I will apply these changes.

newIDE/app/src/SceneEditor/Toolbar.js
INNOVATIVEGAMER: now toggle functionality linked to shortcut keys.

newIDE/app/src/UI/EditorMosaic/CloseButton.js
INNOVATIVEGAMER: > React recommends toggling of components through react state.
So I didn't removed the code in the `componentWillMount` method. Although app works fine without including it but the state of `SceneEditor` should be according to the state of panels, so if we need the state of the panels somewhere else it will not create any conflicts.
The code now works with shortcuts also.
@Bouh Please guide me if some refactors or updates are needed ? ;)

INNOVATIVEGAMER:
> React recommends toggling of components through react state.
So I didn't removed the code in the `componentWillMount` method. Although app works fine without including it but the state of `SceneEditor` should be according to the state of panels, so if we need the state of the panels somewhere else it will not create any conflicts.
The code now works with shortcuts also.
@Bouh Please guide me if some refactors or updates are needed ? ;)

INNOVATIVEGAMER:
@4ian Can you please review this PR now ? 
I have checked the checks failing in the semaphoreci and those are because this PR is older so it can be fixed by merging with master.

INNOVATIVEGAMER:
@Bouh  Can you please review this PR now ? 
I have checked the checks failing in the semaphoreci and those are because this PR is older so it can be fixed by merging with master.

4ian (commented):
This adds a lot of complexity in SceneEditor, by lifting only part of the state of EditorMosaic inside SceneEditor, meaning that we have a mix of imperative functions (openEditor, closeEditor in EditorMosaic) and declarative ones (toggleObjectsList, etc... in SceneEditor).

The additional code in componentWillMount is also something I want to avoid at all costs - it should not be the responsibility of SceneEditor to deal with this.

Two solutions:
- either keep EditorMosaic responsible for handling its state (I.e: don't lift the state), in which case you can call a `toggleEditor` function on it. Advantage: easier. Disadvantage: you can't know the state (opened or not) from SceneEditor.
- or do a more important refactoring work where all the functions to handle the state of an EditorMosaic is put in a React hook, that can be then used in SceneEditor (and the state is then passed down to EditorMosaic).

Appreciate your efforts so far! But in the current situation I think it's adding too much complexity for just toggling stuff. We have to either follow the current architecture or change it. `componentWillMount` should almost never be used.

INNOVATIVEGAMER:
> The additional code in componentWillMount is also something I want to avoid at all costs - it should not be the responsibility of SceneEditor to deal with this.

Ok, so you mean remove all the following states for panels from the SceneEditor : 
https://github.com/4ian/GDevelop/blob/30aa5e78fd05b1f84979a2427955862914661ac8/newIDE/app/src/SceneEditor/index.js#L144-L149
And give total responsibility of EditorMosaic of its state to itself.
and then use react hooks to know and control the state of the EditorMosaic.
So toggling can be done with these hooks as OnToggle first check whether the panel is open or not and act accordingly.

Is I am correct ? 😕

4ian:
> And give total responsibility of EditorMosaic of its state to itself.

That's a possibility yes. If a state is a component, then by definition this component *has* the responsibility of it. 
Currently, we can see that `class EditorMosaic` has the state of the nodes:
```
  state = {
    mosaicNode: this.props.initialNodes,
  };
```

so either:
1. we keep the state here, and manipulate it through imperative functions like `openEditor`, or through `_onChange` which is called by `MosaicWithoutDragDropContext` itself. Advantage: easier, no much refactoring. Disadvantage: you can't know the state (opened or not) from SceneEditor. Because the state is inside the component.
2. so the best alternative, but requiring a lot more refactoring is to separate the handling of the `mosaicNode` inside a hook (`useMosaicNode`). This hook would return: `mosaicNode`, a set of functions to open/change/toggle editor: `openEditor`, `toggleEditor`, `onChange` etc... and finally maybe a function to know if an editor is opened `isEditorOpened`.
  - Now that you have made this, you can refactor SceneEditor to use the hook. It's now trivial to know if an editor is opened (`isEditorOpened('objects-list')`), and you're not "polluting" SceneEditor because it's just a hook to call. You pass `onChange` to the EditorMosaic class, and any function that it needs.
  
2 is better but needs refactoring also in other editors. 1 is a simpler change to toggle just the editor, but you won't be able to show in the toolbar if something is toggled or not. Might not be worth it anyway?

Also remember that on small screens, clicking on buttons will replace the currently opened editor :)

INNOVATIVEGAMER:
> 1. we keep the state here, and manipulate it through imperative functions like `openEditor`, or through `_onChange` which is called by `MosaicWithoutDragDropContext` itself. Advantage: easier, no much refactoring. Disadvantage: you can't know the state (opened or not) from SceneEditor. Because the state is inside the component.

I went with this approach and created two additional imperative functions named `toogleEditor` and `isEditorOpen`.
`toggleEditor` checks if the editor is open or close and acts accordingly.
`isEditorOpen` will remove your mentioned disadvantage in major cases where we want to know the state of the the editor is open or not ;).

I also removed these unnecessary infobars as they are not useful now, since toggling is enabled.
https://github.com/4ian/GDevelop/blob/98a1d69489c5410d28617f11686351e093157c5d/newIDE/app/src/SceneEditor/index.js#L1208-L1246

Can you please tell me the solution for remaining two flow errors ? 
BTW the toggling is working fine now :)

INNOVATIVEGAMER:
> 1. we keep the state here, and manipulate it through imperative functions like `openEditor`, or through `_onChange` which is called by `MosaicWithoutDragDropContext` itself. Advantage: easier, no much refactoring. Disadvantage: you can't know the state (opened or not) from SceneEditor. Because the state is inside the component.

@4ian  I went with this approach and created two additional imperative functions named `toggleEditor` and `isEditorOpen`.
`toggleEditor` checks if the editor is open or close and acts accordingly.
`isEditorOpen` will remove your mentioned disadvantage in major cases where we want to know the state of the the editor is open or not ;).

I also removed these unnecessary infobars as they are not useful now, since toggling is enabled.
https://github.com/4ian/GDevelop/blob/98a1d69489c5410d28617f11686351e093157c5d/newIDE/app/src/SceneEditor/index.js#L1208-L1246
 
The toggling is working fine now :)

Bouh:
I got crash when i go for toggle panels.

```
react-dom.development.js:289 Uncaught Error: Path [first, second, second] did not resolve to a node
    at Object.getAndAssertNodeAtPathExists (mosaicUtilities.ts:169)
    at createRemoveUpdate (mosaicUpdates.ts:62)
    at removeNode (index.js:92)
    at EditorMosaic._this2.closeEditor (index.js:283)
    at EditorMosaic._this2.toggleEditor (index.js:302)
    at SceneEditor._this.toggleObjectsList (index.js:258)
    at HTMLUnknownElement.callCallback (react-dom.development.js:149)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:199)
    at invokeGuardedCallback (react-dom.development.js:256)
    at invokeGuardedCallbackAndCatchFirstError (react-dom.development.js:270)
    at executeDispatch (react-dom.development.js:561)
    at executeDispatchesInOrder (react-dom.development.js:583)
    at executeDispatchesAndRelease (react-dom.development.js:680)
    at executeDispatchesAndReleaseTopLevel (react-dom.development.js:688)
    at forEachAccumulated (react-dom.development.js:662)
    at runEventsInBatch (react-dom.development.js:816)
    at runExtractedEventsInBatch (react-dom.development.js:824)
    at handleTopLevel (react-dom.development.js:4826)
    at batchedUpdates$1 (react-dom.development.js:20439)
    at batchedUpdates (react-dom.development.js:2151)
    at dispatchEvent (react-dom.development.js:4905)
    at react-dom.development.js:20490
    at Object.unstable_runWithPriority (scheduler.development.js:255)
    at interactiveUpdates$1 (react-dom.development.js:20489)
    at interactiveUpdates (react-dom.development.js:2170)
    at dispatchInteractiveEvent (react-dom.development.js:4882)
```

image

INNOVATIVEGAMER:
> I got crash when i go for toggle panels.

@Bouh this happens when you try to toggle the editors at high speed like double clicking the toggle button or fast double press.
So , I added some defensive code for this issue.
Toggling takes its time. So now if user accidently double click the button there will be no effect as added defensive code do not changes the state.

4ian:
Closing as old.

PR #2220, opened 2021-01-17
Convert the p2p extension to Typescript

arthuro555:
Converted the p2p extension to use typescript namespaces and accurate typings.


4ian:
Looks good, thanks!

PR #2223, opened 2021-01-17
Update AdMob to fix issues on iOS

4ian:
* Banners can't be overlayed on the game anymore. It's displayed either above or below.
* New conditions to check if interstitials, banner or reward videos encountered an error when loading.
* New action to set up the test mode for all ads at once
* App Id are now separated, with one for Android and one for iOS. Don't forget to update them before exporting your app, otherwise it would get terminated when started.

### DIFF
--- a/Core/GDCore/Extensions/PlatformExtension.cpp
+++ b/Core/GDCore/Extensions/PlatformExtension.cpp
@@ -41,8 +41,7 @@ gd::InstructionMetadata& PlatformExtension::AddCondition(
     const gd::String& icon,
     const gd::String& smallicon) {
 #if defined(GD_IDE_ONLY)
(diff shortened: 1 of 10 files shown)


PascalLadalle:
The overlay was removed for all devices? Has this proved satisfactory?
Let's say I want to show ads in some scenes and not others.
The aspect ratio is set project-wide, so either I account for the ad height, either I don't, but one way or another, some scenes will show black bars, no?

Came here for this topic: https://forum.gdevelop-app.com/t/admob-banner-width-exceeds-game-width/35823/5

4ian:
It's probable that there is a bug fix to do where we should update the game size after the banner is shown. Can you open an issue about this? The forum thread screenshots seem to indicate there is indeed an issue.

Issue #2040, opened 2020-10-17
Yarn's night mode resets back to light mode on exit.


Lochraleon:
If you set the Yarn interface to night mode and then exit via the APPLY button. Yarn will not retain the setting and revert back to light mode. 

## To Reproduce
1. Click on the EDIT button and then Create/Edit option on the LoadDialogueFromJsonFile action.
2. Once in Yarn you need to create or open an existing "Node". 
3. While in the Node Text Editor, click the "Night Mode' button found to the left of the Language combo box.
4. Close the Node, click APPLY.
5. Repeat step 1. 

## Other details
I am using the desktop GDevelop 5.0.0-beta101 based on GDevelop.js 4.0.98-0-release

Thanks!

4ian:
cc @blurymind not sure how Yarn persists its preferences?

blurymind:
The current version uses localStorage to do it, but the one gdevelop is
using is somewhat outdated and might not be persisting it that way.

I need to do a pr to update it as the latest one has lots of bug fixes and
optimisations. It handles big files much better now. Some stuff might not
work due to the way it is bundled still

PR #2226, opened 2021-01-18
Update to New yarn-editor

blurymind:
This updates the bunded yarn editor to the latest version (attached here)
[yarn-editor-0.4.116.zip](https://github.com/4ian/GDevelop/files/5831938/yarn-editor-0.4.116.zip)
https://github.com/YarnSpinnerTool/YarnEditor/commit/4fcfd2e09053b33a90450ddaa72845ebd594ee8b

try the new editor here
https://yarnspinnertool.github.io/YarnEditor/
or by building this branch :) 

New features:
- synonym suggestion on selected word 
- a brand new settings dialog that lets you configure various new features
- Settings are now actually stored (fixes https://github.com/4ian/GDevelop/issues/2040)
- Brand new interface - to modernize the look and make it work well on tiny screens, new fontawesome icons 
- Brand new support for theming with 3 new themes: Classic, Blueprint (like Twinery), Dracula(Dark)
- Ability to save/load your yarn files to gist.github.com (configure token in the settings) 
- Search by title, body  or tags now highlights on the nodes what is found and a node with search results will automatically highlight them when you open it
- Play testing yarns directly in the editor (please note that this is an alpha feature and may not work the same as in gdevelop)
- Huge improvements in performance when editing big yarn files
- many bug fixes and other small features

increase in bundle size is due to the new dependencies
"@fortawesome/fontawesome-free", (GUI)
    "sweetalert2": "^9.10.13", (GUI)
    "synonyms": "^1.0.1", (suggest synonyms feature)
"gists": "^2.0.0" (backup to gist feature)

This is a result of active contributions on the yarn-editor projects since last year. Ever since Gdevelop started using it, the number of pull requests increased and kept me and others busy reviewing

### DIFF
--- a/newIDE/app/public/external/yarn/yarn-main.js
+++ b/newIDE/app/public/external/yarn/yarn-main.js
@@ -25,6 +25,7 @@ window.addEventListener('yarnReady', e => {
   yarn = e;
   yarn.app.fs = fs;
   yarn.app.electron = electron;
+  yarn.data.restoreFromLocalStorage(false);
   ipcRenderer.send('yarn-ready');
 });
 editorFrameEl.src = 'yarn-editor/index.html';
@@ -42,16 +43,20 @@ ipcRenderer.on('yarn-open', (event, receivedData) => {
     initialResourcePath: receivedData.resourcePath,
     extension: '.json',
   });
-  
+
   // Inject custom Apply button.
   const saveToGdButton = yarn.document
-    .getElementsByClassName('menu')[0]
+    .getElementsByClassName('search-tags')[0]
     .cloneNode(true);
   saveToGdButton.onclick = () => saveAndClose(pathEditorHeader);
   yarn.document
-    .getElementsByClassName('app-menu')[0]
-    .appendChild(saveToGdButton);
-  saveToGdButton.childNodes[0].firstChild.data = 'Apply';
+    .getElementsByClassName('search-tags')[0]
+    .parentElement.appendChild(saveToGdButton);
+  saveToGdButton.childNodes[0].checked = 'checked';
+  saveToGdButton.childNodes[2].innerHTML = 'Apply';
+  saveToGdButton.childNodes[2].style = 'background-color: white;';
+  yarn.document.getElementsByClassName('app-search')[0].style = 'right: 45px';
+  saveToGdButton.style = 'padding-left: 30px;';
 
   // Process the json file to open, if any.
   if (fileExists(receivedData.resourcePath)) {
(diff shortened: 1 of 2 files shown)


newIDE/app/public/external/yarn/yarn-main.js
```diff
@@ -25,6 +25,7 @@ window.addEventListener('yarnReady', e => {
   yarn = e;
   yarn.app.fs = fs;
   yarn.app.electron = electron;
+  yarn.data.restoreFromLocalStorage(false);
```
4ian, line 28: What is the `false` for?

4ian:
Thanks! Sounds good to update it then :) 
Can you double check that it works well when editing existing examples - and more generally check that everything seems alright?

newIDE/app/public/external/yarn/yarn-main.js
blurymind: A new feature in yarn is to automatically save its entire edited file to localStorage. That way if its tab gets closed or electron window closed - later on when you start yarn again - it will restore its entire state of the app.

This is very useful to have on a web app and on mobile -it makes it act like vscode. It's however not useful when you bundle it in gdevelop, so I am disabling it. If you create a new file via gdevelop, it will get created from the backup in localStorage - unless I disable this.

blurymind:
> Thanks! Sounds good to update it then :)
> Can you double check that it works well when editing existing examples - and more generally check that everything seems alright?

Yup, I tested it yesterday, but will give it another go tomorrow night

4ian:
@blurymind Let me know when you can do a new check and we'll go ahead with this if there are no problems :)

blurymind:
@

> @blurymind Let me know when you can do a new check and we'll go ahead with this if there are no problems :)

@4ian seems to work well :)  but I just noticed that when the file helper header is visible, I can not midclick pan inside yarn. Alt click pan does work. I wonder whats stopping the mid mouse event from registering inside the yarn iframe

4ian:
> I found its due to position: relative. Will fix it tomorrow

Thanks for investigating! Let me know if you can fix it, would be good to have before we merge this editor update. Should be on time for a new testing release soon, along with the Tilemap :)

4ian:
I merged this and upgraded the editor. Let me know if you can do the fix that you've identified in another PR :)

blurymind:
Thanks @4ian :) this is a separate bug with the path editor that I will do
another pr to fix. Can't believe we haven't noticed it before

blurymind:
@4ian fixed that in https://github.com/4ian/GDevelop/pull/2257

was a sneaky css bug 😄

Issue #2187, opened 2021-01-06
Backward compatibility after updating an extension


Bouh:
## Describe the bug
Update an existing extension can have consequences on it.

- First one is if you have added a custom function inside not included in the original extension, the function will be removed with a bad visual feedback. **Draw_life_bar** function is removed.
[Video 1](https://streamable.com/2br1sj)

   Suggestion:  Update only the existing function/behavior

- Second one is a bad bugged visual feedback because during the update the selection on a function is disabled.
[Video 2](https://streamable.com/d2a3cx) 

   Suggestion: Close the extension tab, or select the first function/behavior.

## Other details
* B103 and lower
* w10

4ian:
So there are indeed 2 problems there :)

1) When we update an extension, we change "under the hood" the extension and this means the UI can get out of sync, which is dangerous as this expose potential access to deleted memory (i.e: crash or memory corruption). When an extension is updated, we should probably close the extension tab to be 100% safe, or as you said ensure to select again the behavior/function/behavior method to be 100% sure that we're not showing deleted stuff.

2) We also have the issue of backward compatibility. I think in the future we should develop something that analyze the "public interface" of an extension (this means the behavior, behavior methods and functions, as well as their parameters) and warn if there are some changes : 
"The updated extension don't have 3 functions? Check them if you want to keep them (not recommended): ..."

It's in a way the same issue as updating a library in usual programming: you either have to keep the old, deprecated functions in your library, or make a breaking change. In an ideal world, GDevelop could adapt your project according to rules that are set (rename expression XYZ to ABC...), but to keep things simple we could at least add a warning.

INNOVATIVEGAMER:
https://github.com/4ian/GDevelop/blob/b475a2ed9625d268b6924fd88cde306d666aa514/newIDE/app/src/AssetStore/ExtensionStore/index.js#L119-L121

I think before calling onInstall method we should add a code which closes the extension tab associated with the selectedExtensionShortHeader. But how can we reference to that tab from here ?
Here we have only the data related to extension from selectedExtensionShortHeader like name of the extension.

INNOVATIVEGAMER:
as @4ian mentioned two issues,
I have solved the first issue which closes the extension tab before start of the installation.
Shall I make PR or I need to solve second issue also ?
I think second issue is more complex and needs attention in separate PR. :)

4ian:
It's almost always better to make small separate PRs, so feel free to open the first one :)

[...]

PR #2228, opened 2021-01-19
Close the extension tab when an extension is updated

INNOVATIVEGAMER:
Fixes first issue in #2187 .
If the extension tab is opened while updating/reinstalling an extension , It will close the extension tab before start of the installation
to prevent memory corruption.

### DIFF
--- a/newIDE/app/src/AssetStore/ExtensionStore/ExtensionsSearchDialog.js
+++ b/newIDE/app/src/AssetStore/ExtensionStore/ExtensionsSearchDialog.js
@@ -14,6 +14,7 @@ import InfoBar from '../../UI/Messages/InfoBar';
 type Props = {|
   project: gdProject,
   onClose: () => void,
+  onInstallExtension: (str: string) => void,
 |};
 
 /**
@@ -22,7 +23,7 @@ type Props = {|
 export default function ExtensionsSearchDialog({
   project,
   onClose,
-  onCloseEventsFunctionsExtensionTab,
+  onInstallExtension,
 }: Props) {
   const [isInstalling, setIsInstalling] = React.useState(false);
   const [extensionWasInstalled, setExtensionWasInstalled] = React.useState(
@@ -92,9 +93,7 @@ export default function ExtensionsSearchDialog({
             }}
             project={project}
             showOnlyWithBehaviors={false}
-            onCloseEventsFunctionsExtensionTab={
-              onCloseEventsFunctionsExtensionTab
-            }
+            onInstallExtension={onInstallExtension}
           />
           <InfoBar
             identifier="extension-installed-explanation"
--- a/newIDE/app/src/AssetStore/ExtensionStore/ExtensionsSearchDialog.js
+++ b/newIDE/app/src/AssetStore/ExtensionStore/ExtensionsSearchDialog.js
@@ -10,11 +10,12 @@ import EventsFunctionsExtensionsContext from '../../EventsFunctionsExtensionsLoa
 import HelpButton from '../../UI/HelpButton';
 import { importExtension, installExtension } from './InstallExtension';
 import InfoBar from '../../UI/Messages/InfoBar';
+import { type ExtensionShortHeader } from '../../Utils/GDevelopServices/Extension';
 
 type Props = {|
   project: gdProject,
   onClose: () => void,
-  onInstallExtension: (str: string) => void,
+  onInstallExtension: ExtensionShortHeader => void,
 |};
 
 /**
--- a/newIDE/app/src/AssetStore/ExtensionStore/ExtensionsSearchDialog.js
+++ b/newIDE/app/src/AssetStore/ExtensionStore/ExtensionsSearchDialog.js
@@ -82,6 +82,7 @@ export default function ExtensionsSearchDialog({
             isInstalling={isInstalling}
             onInstall={async extensionShortHeader => {
               setIsInstalling(true);
+              onInstallExtension(extensionShortHeader);
               const wasExtensionInstalled = await installExtension(
                 i18n,
                 project,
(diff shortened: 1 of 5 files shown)


newIDE/app/src/AssetStore/ExtensionStore/ExtensionsSearchDialog.js
```diff
@@ -19,7 +19,11 @@ type Props = {|
 /**
  * Allows to browse and install events based extensions.
  */
-export default function ExtensionsSearchDialog({ project, onClose }: Props) {
+export default function ExtensionsSearchDialog({
+  project,
+  onClose,
+  onCloseEventsFunctionsExtensionTab,
```
4ian, line 25: While this works, this gives too much responsibility on the ExtensionSearchDialog, which must find the eventsFunctionExtension associated to the installed extension. This component should not care about this (You can picture it telling: "I'm a search and install dialog, I don't care about your freaking tabs" ;))

Instead, let's call this prop "onInstallExtension", getting a ExtensionShortHeader as parameter.

newIDE/app/src/AssetStore/ExtensionStore/index.js
```diff
@@ -117,6 +121,10 @@ export const ExtensionStore = ({
             selectedExtensionShortHeader.name
           )}
           onInstall={() => {
+            const eventsFunctionsExtension = project.getEventsFunctionsExtension(
+              selectedExtensionShortHeader.name
+            );
```
4ian, line 126: Two things:
- you must use hasEventsFunctionsExtension, otherwise using getEventsFunctionsExtension with a non existing name will crash/returning corrupted memory content.
- let's move this in the MainFrame. It's not the responsibility of this store to handle the tabs.

4ian:
Seems like a good start, check my comments to switch the responsibility of which component is closing the tab.

INNOVATIVEGAMER:
As you mentioned shifted the responsibility of handling tabs to `MainFrame`.
Thank you for your advice , now this code looks complete and more logical towards extensibility.

For further feature as you mentioned in the second issue that we need to check the previously installed and current functions in  extension we can add further code in this `onInstall` function for giving user warning about deprecated functions in the updated extension.

newIDE/app/src/AssetStore/ExtensionStore/index.js
```diff
@@ -117,6 +119,7 @@ export const ExtensionStore = ({
             selectedExtensionShortHeader.name
           )}
           onInstall={() => {
+            onInstallExtension(selectedExtensionShortHeader);
```
4ian, line 122: There is already a prop doing this, so we don't need this duplicate. Let's use the other in the parent :)

newIDE/app/src/MainFrame/index.js
```diff
@@ -811,6 +812,28 @@ const MainFrame = (props: Props) => {
     _onProjectItemModified();
   };
 
+  const onInstallExtension = (extensionShortHeader: ExtensionShortHeader) => {
+    //Close the extension tab before updating/reinstalling the extension
+    const eventsFunctionsExtensionName = extensionShortHeader.name;
+    const hasEventsFunctionsExtension = currentProject.hasEventsFunctionsExtensionNamed(
+      eventsFunctionsExtensionName
+    );
```
4ian, line 820: nitpicking: no need for a temporary, put this in the `if( )`

newIDE/app/src/MainFrame/index.js
```diff
@@ -811,6 +812,28 @@ const MainFrame = (props: Props) => {
     _onProjectItemModified();
   };
 
+  const onInstallExtension = (extensionShortHeader: ExtensionShortHeader) => {
+    //Close the extension tab before updating/reinstalling the extension
```
4ian, line 816: Uber nitpicking: space after the //, and always a period at the end of comments.

newIDE/app/src/AssetStore/ExtensionStore/ExtensionsSearchDialog.js
```diff
@@ -88,6 +94,7 @@ export default function ExtensionsSearchDialog({ project, onClose }: Props) {
             }}
             project={project}
             showOnlyWithBehaviors={false}
+            onInstallExtension={onInstallExtension}
```
4ian, line 97: Let's move this in `onInstall` ;)

4ian (commented):
Almost there ;)

INNOVATIVEGAMER:
Made your mentioned changes.
Thank you for guiding new comers like me to the practices followed in todays world codes.
Glad to have a maintainer & teacher like you 😄

4ian:
Looks good! Final step is to fix the stories (`npm run storybook`) and additional flow errors. See them in https://semaphoreci.com/4ian/gd/branches/pull-request-2228/builds/3 (you can ignore "GDJS auto-formatting" and ignore the 2 Flow errors that are about `src/EventsBasedBehaviorsList/...` files - both of these can be fixed by merging the `master` branch in your branch).

There are two issues because `currentProject` could in theory be null, so you have to get this variable and check it's not null (see how it's done in other functions).
And the rest is about the stories of Storybook being not updated with the new props.

Thanks!

INNOVATIVEGAMER:
> Final step is to fix the stories (`npm run storybook`)

These storybooks seems cool. Can you please elaborate how you created these storybooks ? or any kind of reference links/ readmes which explain them ?
What are the advantages of these storybooks in our application ?

The fixes for currentProject check made in latest commit and I ran `npm run storybook` on local machine successfully.

INNOVATIVEGAMER:
> Final step is to fix the stories (`npm run storybook`)

These storybooks seems cool. Can you please elaborate about these storybooks ? or any kind of reference links to get better understanding of them ?
What is the role or advantage of these storybooks in GD ?

Fixed errors about currentProject being null and missing props in storybooks.

4ian:
Storybook is a way to expose your app components in isolation, outside of the app itself. Very useful to develop them, try them without spending time finding them in the app, and allow to showcase them in different situations (with different props). Useful to check that nothing is broken and that all use cases are covered :) 

Thanks, I've merged the PR!

PR #2231, opened 2021-01-21
Add tests for keyboard shortcuts

NilayMajorwar:
This PR adds some tests for the keyboard shortcuts logic. In particular, the tests cover the following:
- `getShortcutStringFromEvent`: Parse event object into respective shortcut string
- `isValidShortcutEvent`: Check if an event object is a valid shortcut keypress
- `getElectronAccelerator`: Convert internal shortcut strings to Electron-compatible "accelerator" strings

This is my first time writing tests for an actual project, so it may not be very good :sweat_smile:. I've covered all the exported functions from `src/KeyboardShortcuts/index.js`, except:
- `getShortcutDisplayName`: it's user interface, so isn't necessary to test, right?
- `getShortcutMetadataFromEvent`: this functionality (except reserved shortcuts) is covered in tested functions
- the two React hooks `useShortcutMap` and `useKeyboardShortcuts`

I've also made some very small refactoring changes to `src/KeyboardShortcuts/index.js`, that do not change the logic in any manner.

### DIFF
--- a/newIDE/app/src/KeyboardShortcuts/KeyboardShortcuts.spec.js
+++ b/newIDE/app/src/KeyboardShortcuts/KeyboardShortcuts.spec.js
@@ -1,0 +1,96 @@
+import {
+  getShortcutStringFromEvent,
+  isValidShortcutEvent,
+  getElectronAccelerator,
+} from './index';
+
+const keyEvent = (ctrlKey, shiftKey, altKey, code) => ({
+  ctrlKey,
+  shiftKey,
+  altKey,
+  code,
+});
+
+// Action key, with various modifiers
+const keyA = keyEvent(false, false, false, 'KeyA'); // A
+const ctrlA = keyEvent(true, false, false, 'KeyA'); // Ctrl+A
+const shiftA = keyEvent(false, true, false, 'KeyA'); // Shift+A
+const altA = keyEvent(false, false, true, 'KeyA'); // Alt+A
+const ctrlShiftAltA = keyEvent(true, true, true, 'KeyA'); // Ctrl+Shift+Alt+A
+
+// Partial or invalid shortcut keypresses
+const ctrlShiftAlt = keyEvent(true, true, true, ''); // Ctrl+Shift+Alt+
+const ctrlShiftAltUp = keyEvent(true, true, true, 'ArrowUp'); // Ctrl+Shift+Alt+Up
+
+// Edge cases of restricted shortcuts
+const keyTab = keyEvent(false, false, false, 'Tab'); // Tab
+const shiftTab = keyEvent(false, true, false, 'Tab'); // Shift+Tab
+const ctrlTab = keyEvent(true, false, false, 'Tab'); // Ctrl+Tab
+const altTab = keyEvent(false, false, true, 'Tab'); // Alt+Tab
+const keySpace = keyEvent(false, false, false, 'Space'); // Space
+const shiftSpace = keyEvent(false, true, false, 'Space'); // Shift+Space
+const ctrlSpace = keyEvent(true, false, false, 'Space'); // Ctrl+Space
+const altSpace = keyEvent(false, false, true, 'Space'); // Alt+Space
+
+describe('KeyboardShortcuts', () => {
+  it('creates shortcut strings from events correctly', () => {
+    // Action key, with modifiers
+    expect(getShortcutStringFromEvent(keyA)).toBe('KeyA');
+    expect(getShortcutStringFromEvent(ctrlA)).toBe('CmdOrCtrl+KeyA');
+    expect(getShortcutStringFromEvent(shiftA)).toBe('Shift+KeyA');
+    expect(getShortcutStringFromEvent(altA)).toBe('Alt+KeyA');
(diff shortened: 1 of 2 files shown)


newIDE/app/src/KeyboardShortcuts/KeyboardShortcuts.spec.js
```diff
@@ -0,0 +1,100 @@
...
+  getShortcutStringFromEvent,
+  isValidShortcutEvent,
+  getElectronAccelerator,
+} from './index';
+
+/**
+ * Creates a KeyboardEvent-like object for testing
+ * functions that take event objects as input
+ */
+const keyEvent = (ctrlKey, shiftKey, altKey, code) => ({
```
4ian, line 11: Uber nitpicking: I usually make functions to always start with a verb, because it's something that "do" something.
Here: `makeKeyEvent`

newIDE/app/src/KeyboardShortcuts/KeyboardShortcuts.spec.js
```diff
@@ -0,0 +1,100 @@
+import {
```
4ian, line 1: Try to add `// @flow` and see if there are any errors.
Even for test, we try to have flow coverage so that we can get static typing and the safety that comes with it when doing refactorings in the future :)

4ian:
Looks great! Thanks a lot for adding these tests, they are very useful to ensure future refactoring/work on this can be done safely. They are also great "examples" for the reader that tries to understand your code. 

> getShortcutDisplayName: it's user interface, so isn't necessary to test, right?

Depends on the risk associated with it. If it breaks, is it a huge problem? Is it complex enough so that someone could break it in the future and not see it immediately? Are there any edge case that is hard to see and not often encountered, that you would like to show in a test?
If any answer is yes, you can consider making a test :) 

> getShortcutMetadataFromEvent: this functionality (except reserved shortcuts) is covered in tested functions

It's fine if it's mostly covered by others tests. 
As we say "Write tests. Not too many. Mostly integration." - so if your function is tested by calling others, it's fine already.

> the two React hooks useShortcutMap and useKeyboardShortcuts

React hooks would need to be tested inside components - we have almost none of these tests so far. It would be an interesting exercise to do - though I think that these tests using components are often hard to set up, maintain and I prefer making "stories" in Storybook to make a "visual test case" that is useful for the next developer.

Finally, I tend to say that if there is too much logic in a hook or a component, this logic should be moved to external functions, that can be tested (up to a point, as with everything)

So I would say not to worry with all of these. What you did is already great 👍

newIDE/app/src/KeyboardShortcuts/KeyboardShortcuts.spec.js
NilayMajorwar: My bad - fixed.

newIDE/app/src/KeyboardShortcuts/KeyboardShortcuts.spec.js
NilayMajorwar: I would've done that, but I'll have to force-typecast the output of `makeKeyEvent` (`object` -> `any` -> `KeyboardEvent`) since two of the three tested functions take an event object as argument. The rest of it is fine. Should I do it?

NilayMajorwar:
Yup, that rules out pretty much everything I think. `useShortcutMap` is a nearly trivial hook and `useKeyboardShortcuts`, while looking bulky, is more or less a combination of helper functions already.

On a side note though, if we were to create a Storybook test case for `useKeyboardShortcuts`, it would involve creating a component with:
- a display that shows the command name for the shortcut pressed
- a dummy text input - to test/ensure shortcuts don't work when user is typing
- a dummy dialog - to test/ensure shortcuts don't work when a dialog is open

With some more gears and switches, maybe this test component can act as a really good test case for keyboard shortcuts - checking clashes with user typing and open dialogs, reserved shortcuts, nearly everything. Perhaps I can work on it and add it to this PR?

newIDE/app/src/KeyboardShortcuts/KeyboardShortcuts.spec.js
4ian: Let's still do it and add a `// $FlowIgnore - we intentionally create a fake KeyboardEvent` like comment to it :) 
Better to have a file typed even if there is a comment to ignore an error because it's a test - so not everything is perfectly simulated.

4ian:
> With some more gears and switches, maybe this test component can act as a really good test case for keyboard shortcuts - checking clashes with user typing and open dialogs, reserved shortcuts, nearly everything. Perhaps I can work on it and add it to this PR?

That could be useful, feel free to take a look, either in in this PR or in another as you prefer 👍
It would be great to have. As storybook is running inside the browser not Electron we can't test everything, but it's already very good if we have a "test bench" to quickly check "is pressing a shortcut working"? "how does it work with inputs"? "how does it work with dialogs?" :)

NilayMajorwar:
I did some work on the test bed, but looks like it's a bit tougher than I thought. So I think I'll take some time and do it in another PR.

4ian:
Sure, no worries! Looks great, thanks for the PR, very well done 👍

PR #2235, opened 2021-01-23
Fix the issue double extension prefix

INNOVATIVEGAMER:
fixes issue #2195 .
Reconsidered this issue as  I thought last time I nearly solved it.
@4ian Your thoughts about changing `formatCompletion` were right. As there was no need to disturb `newExpressionStart` and output is wrong due to `insertedAutocompletion.completion` so checking it and improving it makes more sense.

So I changed the logic and figured out to change the logic behind `formatCompletion`. Now code checks for the completion word before returning it into `insertedWord`. **If the completion word is returning `MyExtension::MyFunction` (kind of expressions) whereas just `MyFunction` is accurate. The MyFunciton will be extracted from completion and returned to `insertedWord` with completed format.**

### DIFF
--- a/newIDE/app/src/ExpressionAutocompletion/index.js
+++ b/newIDE/app/src/ExpressionAutocompletion/index.js
@@ -293,7 +293,7 @@ export const getAutocompletionsFromDescriptions = (
   );
 };
 
-const separatorChars = ' .:+-/*=()<>[]@!?|\\^%#';
+let separatorChars = ' .:+-/*=()<>[]@!?|\\^%#';
 
 type InsertedAutocompletion = {|
   completion: string,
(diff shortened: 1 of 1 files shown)


INNOVATIVEGAMER:
@4ian this PR is now ready for review :)

4ian:
Unfortunately I still don't think replacing text is the way to go.
Look at how a function is parsed in the expression parser:
https://github.com/4ian/GDevelop/blob/ef1657d82a84245aadd792fc1a78ed1447d1e154/Core/GDCore/Events/Parsers/ExpressionParser2.h#L252-L272

There is a comment indicating that ":" is here considered as part of the function name, so the full function name is "MyExtension::MyFunction". Either we:
- Change the parser so that the node is now containing both the extension name "MyExtension" and "MyFunction". We can then adapt the completions so that they return just "MyFunction" when you want to complete something after the "::".
- Or we don't change the parser but in this case need to handle the completions in the same way as this code. This means that ":" is considered as a separator of words, but in the case of a completion of an expression, it's not - meaning that for example something like "MyExtension::MyFunc" is considered as a single word and replaced entirely.

As said in the other PR, I really really don't want to replace things because they are redundant. It would be a patch, instead of aligning the code of the parser and of the completion to work in the same way - which is preferable.

INNOVATIVEGAMER:
@4ian I went with the second approach.

> * we don't change the parser but in this case need to handle the completions in the same way as this code. This means that ":" is considered as a separator of words, but in the case of a completion of an expression, it's not - meaning that for example something like "MyExtension::MyFunc" is considered as a single word and replaced entirely.

Sorry for so much delay in this PR 😥

4ian:
This was fixed since then by improving the completion to include the place where they should be inserted :) Closing this then!

PR #2237, opened 2021-01-24
Allow to choose isometry in the top-down movement behavior

D8H:
This pull request doesn't have the player assistance. I think this is more waited and a lot more easy to review so I did another pull request with it.

It allows to choose the viewpoint between:
- top-down
- isometry 2:1
- true isometry
- custom isometry

This example uses the true isometry:
https://games.gdevelop-app.com/game-6abd1d9d-57aa-4b42-ac03-1b35567dd659/index.html

The exposed position is the one on screen, but for speed it's the one in the isometric world. I think it make more sense because max speed doesn't make sense screen related.
I wonder if the position in the isometric world should be exposed too. An extension to transform coordinates will be needed to make advance isometry features, but it could be good to be able to get it directly, yet maybe confusing for top-down viewpoint users.
There are functions for isometry in the Extended Math extension, but I think isometry could need a lot of functions, condition and actions to address commons issues, so a new extension may be better.

The BasisTransformation interface and its implementations can be used for the change request on the path finding. Should it be move somewhere to be shared between all extensions? And maybe do a factory?

Maybe the 2:1 isometry needs a specific implementation to avoid rounded errors.

https://trello.com/c/vp6eCCZh

### DIFF
--- a/Extensions/TopDownMovementBehavior/topdownmovementruntimebehavior.ts
+++ b/Extensions/TopDownMovementBehavior/topdownmovementruntimebehavior.ts
@@ -382,14 +382,14 @@ namespace gdjs {
   }
 
   export class IdentityTransformation implements BasisTransformation {
-    toScreen(wordPoint: FloatPoint, screenPoint: FloatPoint): void {
-      screenPoint[0] = wordPoint[0];
-      screenPoint[1] = wordPoint[1];
+    toScreen(worldPoint: FloatPoint, screenPoint: FloatPoint): void {
+      screenPoint[0] = worldPoint[0];
+      screenPoint[1] = worldPoint[1];
     }
 
-    toWorld(screenPoint: FloatPoint, wordPoint: FloatPoint): void {
-      wordPoint[0] = screenPoint[0];
-      wordPoint[1] = screenPoint[1];
+    toWorld(screenPoint: FloatPoint, worldPoint: FloatPoint): void {
+      worldPoint[0] = screenPoint[0];
+      worldPoint[1] = screenPoint[1];
     }
   }
 
@@ -418,22 +418,22 @@ namespace gdjs {
       ];
     }
 
-    toScreen(wordPoint: FloatPoint, screenPoint: FloatPoint): void {
+    toScreen(worldPoint: FloatPoint, screenPoint: FloatPoint): void {
       const x =
-        this.screen[0][0] * wordPoint[0] + this.screen[0][1] * wordPoint[1];
+        this.screen[0][0] * worldPoint[0] + this.screen[0][1] * worldPoint[1];
       const y =
-        this.screen[1][0] * wordPoint[0] + this.screen[1][1] * wordPoint[1];
+        this.screen[1][0] * worldPoint[0] + this.screen[1][1] * worldPoint[1];
       screenPoint[0] = x;
       screenPoint[1] = y;
     }
 
-    toWorld(screenPoint: FloatPoint, wordPoint: FloatPoint): void {
+    toWorld(screenPoint: FloatPoint, worldPoint: FloatPoint): void {
       const x =
         this.world[0][0] * screenPoint[0] + this.world[0][1] * screenPoint[1];
       const y =
         this.world[1][0] * screenPoint[0] + this.world[1][1] * screenPoint[1];
-      wordPoint[0] = x;
-      wordPoint[1] = y;
+      worldPoint[0] = x;
+      worldPoint[1] = y;
     }
   }
 
--- a/Extensions/TopDownMovementBehavior/topdownmovementruntimebehavior.ts
+++ b/Extensions/TopDownMovementBehavior/topdownmovementruntimebehavior.ts
@@ -314,8 +314,7 @@ namespace gdjs {
         // Top-down pointview
         object.setX(object.getX() + this._xVelocity * timeDelta);
         object.setY(object.getY() + this._yVelocity * timeDelta);
-      }
-      else {
+      } else {
         // Isometry pointview
         const point = this._temporaryPointForTransformations;
         point[0] = this._xVelocity * timeDelta;
@@ -389,8 +388,10 @@ namespace gdjs {
      */
     constructor(angle: float) {
       if (angle <= 0 || angle >= Math.PI / 4)
-        throw new RangeError("An isometry angle must be in ]0; pi/4] but was: " + angle);
-      
+        throw new RangeError(
+          'An isometry angle must be in ]0; pi/4] but was: ' + angle
+        );
+
       const alpha = Math.asin(Math.tan(angle));
       const sinA = Math.sin(alpha);
       const cosB = Math.cos(Math.PI / 4);
--- a/Extensions/TopDownMovementBehavior/topdownmovementruntimebehavior.ts
+++ b/Extensions/TopDownMovementBehavior/topdownmovementruntimebehavior.ts
@@ -381,9 +381,16 @@ namespace gdjs {
   }
 
   export class IsometryTransformation implements BasisTransformation {
-    screen: float[][];
+    _screen: float[][];
 
+    /**
+     * @param angle between the x axis and the projected isometric x axis.
+     * @throws if the angle is not in ]0; pi/4[. Note that 0 is a front viewpoint and pi/4 a top-down viewpoint.
+     */
     constructor(angle: float) {
+      if (angle <= 0 || angle >= Math.PI / 4)
+        throw new RangeError("An isometry angle must be in ]0; pi/4] but was: " + angle);
+      
       const alpha = Math.asin(Math.tan(angle));
       const sinA = Math.sin(alpha);
       const cosB = Math.cos(Math.PI / 4);
@@ -393,7 +400,7 @@ namespace gdjs {
       //   / 1     0    0 \ / cosB 0 -sinB \ / 1 0  0 \
       //   | 0  cosA sinA | |    0 1     0 | | 0 0 -1 |
       //   \ 0 -sinA cosA / \ sinB 0  cosB / \ 0 1  0 /
-      this.screen = [
+      this._screen = [
         [cosB, -sinB],
         [sinA * sinB, sinA * cosB],
       ];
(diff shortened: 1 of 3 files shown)


4ian:
Thanks for the PR, I'll take a look at the code :) 

> There are functions for isometry in the Extended Math extension, but I think isometry could need a lot of functions, condition and actions to address commons issues, so a new extension may be better.

I think this could be a good candidate for a community extension yeah (https://github.com/4ian/GDevelop-extensions).

> The BasisTransformation interface and its implementations can be used for the change request on the path finding. Should it be move somewhere to be shared between all extensions? And maybe do a factory?

We can think about that later once it's working well in this PR.

Extensions/TopDownMovementBehavior/topdownmovementruntimebehavior.ts
```diff
@@ -18,18 +23,26 @@ namespace gdjs {
...
     _yVelocity: float = 0;
     _angularSpeed: number = 0;
     _leftKey: boolean = false;
     _rightKey: boolean = false;
     _upKey: boolean = false;
     _downKey: boolean = false;
+    
+    // @ts-ignore
+    _basisTransformation: BasisTransformation;
+    _point: number[] = [0, 0];
```
4ian, line 45: Prefer `float` instead of `number` everywhere (or `integer` if you manipulate integers).

I recommend also defining a `type FloatPoint = [float, float];`, this is done in other extensions and will be probably used more in the future :)

Extensions/TopDownMovementBehavior/topdownmovementruntimebehavior.ts
```diff
@@ -71,9 +85,33 @@ namespace gdjs {
...
       return true;
     }
 
+    setViewpoint(viewpoint: string, customIsometryAngle: number): void {
+      this._customIsometryAngle = customIsometryAngle;
+      if (viewpoint == 'PixelIsometry') {
+        this._viewpoint = TopDownMovementRuntimeBehavior.PIXEL_ISOMETRIE;
+        this._basisTransformation = new IsometryTransformation(Math.atan(0.5));
+      }
+      else if (viewpoint == 'TrueIsometry') {
```
4ian, line 101: nitpicking: run `npm run format` in GDJS to autoformat the code.

Extensions/TopDownMovementBehavior/topdownmovementruntimebehavior.ts
```diff
@@ -9,6 +9,11 @@ namespace gdjs {
    * and rotation.
    */
   export class TopDownMovementRuntimeBehavior extends gdjs.RuntimeBehavior {
+    static TOP_DOWN = 0;
+    static PIXEL_ISOMETRIE = 1;
```
4ian, line 13: `PIXEL_ISOMETRY`?

Extensions/TopDownMovementBehavior/topdownmovementruntimebehavior.ts
```diff
@@ -153,11 +191,30 @@ namespace gdjs {
     }
 
     doStepPreEvents(runtimeScene) {
+      
+      const object = this.owner;
+      
+      /* Check if the object has moved
+       * To avoid to loop on the transform and its inverse
+       * beacause of float approximation.
+       */
```
4ian, line 200: nitpicking: use // for inline and multilines comments inside functions, as it's more concise.

Extensions/TopDownMovementBehavior/topdownmovementruntimebehavior.ts
4ian: This seems a temporary variable, let's either define it in the function using it or changing its name to identify that it's "a temporary just here to avoid allocations when computing transformations" (so `_temporaryPointForTransformations`).

Extensions/TopDownMovementBehavior/topdownmovementruntimebehavior.ts
```diff
@@ -322,6 +386,67 @@ namespace gdjs {
...
+  export interface BasisTransformation {
+    
+    toScreen(worldPoint: number[], screenPoint: number[]): void;
+    
+    toWorld(screenPoint: number[], worldPoint: number[]): void;
+  }
+  
+  export class IdentityTransformation implements BasisTransformation {
+    
+    toScreen(wordPoint: number[], screenPoint: number[]): void {
```
4ian, line 399: worldPoint (and below)

Extensions/TopDownMovementBehavior/topdownmovementruntimebehavior.ts
```diff
@@ -270,8 +327,8 @@ namespace gdjs {
       //No acceleration for angular speed for now
 
       //Position object
-      object.setX(object.getX() + this._xVelocity * timeDelta);
-      object.setY(object.getY() + this._yVelocity * timeDelta);
+      this._x += this._xVelocity * timeDelta;
+      this._y += this._yVelocity * timeDelta;
```
4ian, line 331: Do you think that instead of converting from "screen" coordinates to "world" coordinates and back to "screen" coordinates, we could instead only multiply the changes done on the x and y position by some factor, depending on the viewpoint chosen (x: 1, y: 1factors in the case of top down,  x: 1, y: 0.5 in the case of pixel isometry)?

Would that impact other things like acceleration or movement? 
I'm not a specialist in how to deal with isometry, but I'm wondering if it would be sufficient (and more performant than doing point space transformations at every frame).

4ian (commented):
Added a few nitpickings and some questions :)

Extensions/TopDownMovementBehavior/topdownmovementruntimebehavior.ts
```diff
@@ -18,18 +23,26 @@ namespace gdjs {
...
+    _y: number = 0;
     _xVelocity: float = 0;
     _yVelocity: float = 0;
     _angularSpeed: number = 0;
     _leftKey: boolean = false;
     _rightKey: boolean = false;
     _upKey: boolean = false;
     _downKey: boolean = false;
+    
+    // @ts-ignore
```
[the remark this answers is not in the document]
D8H: The setter "setViewpoint" is not detected as an affectation.

Extensions/TopDownMovementBehavior/topdownmovementruntimebehavior.ts
D8H: Indeed, I used the "in world" position for the other pull request but that's not necessary here. Only the delta can be transformed.

About the pixel isometry, to keep the distance there is a factor:
screenX = 0.5*sqrt(2) * (worldX - worldY)
screenY = 0.5 * 0.5*sqrt(2) * (worldX + worldY)

4ian: > So toWorld() becomes a dead code, should I remove it?

Yes please! I know it's a bit frustrating but in this context I would really like to have the minimum changes required to make this work.
The other Pull Request is pretty big, and maybe the behavior will need rework/be separated in another behavior - but for this PR let's keep things dead simple.

> I think what you said should work but the up will still be up by default and the transformation is just 6 operations that's not really costly. What do you think?

Well it's only 6 operations but it's still 6 more operations even in the basic case, so if someone uses this behavior for animating 100 small objects on screen (to benefit from acceleration, rotation, etc...), it's now 600 operations.

Might not be too much (depends on the device), but still I kind of like the idea that if you keep the top down approach, we don't add any overhead to the behavior every frame (if you're using isometric coordinates, that's ok) :)

> but the up will still be up by default

I think that's fine, it would be easy to move from topdown to isometric, and creators can still remap the controls if they need using events?

Extensions/TopDownMovementBehavior/topdownmovementruntimebehavior.ts
```diff
@@ -18,19 +24,27 @@ namespace gdjs {
     _rotateObject: any;
     _angleOffset: any;
     _ignoreDefaultControls: any;
+    _viewpoint: any;
+    _customIsometryAngle: any;
```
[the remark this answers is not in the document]
4ian: Might be a good idea to do so, always in the idea of simplying this PR to the most simple implementation for supporting isometry movements?

Extensions/TopDownMovementBehavior/topdownmovementruntimebehavior.ts
D8H: > I think that's fine, it would be easy to move from topdown to isometric, and creators can still remap the controls if they need using events?
Yes, they can do it with the angle offset property. But, imho, even though it's hard for the brain,  it's more usable to go up in diagonal because the world is likely built to go that way. Imagine a long corridor or a labyrinth where you have to keep pressing 2 keys not to take a wall.

4ian: Sorry I missed your reply!
I'll take a new look at the code :)

Extensions/TopDownMovementBehavior/topdownmovementruntimebehavior.ts
```diff
@@ -322,6 +375,40 @@ namespace gdjs {
       this._downKey = true;
     }
   }
+
+  export interface BasisTransformation {
+    toScreen(worldPoint: FloatPoint, screenPoint: FloatPoint): void;
+  }
+
+  export class IsometryTransformation implements BasisTransformation {
+    screen: float[][];
```
4ian, line 384: nitpicking: put a `_` as the first character to indicate that this should not be used outside of the class.

Extensions/TopDownMovementBehavior/topdownmovementruntimebehavior.ts
```diff
@@ -322,6 +375,40 @@ namespace gdjs {
...
+
+    constructor(angle: float) {
+      const alpha = Math.asin(Math.tan(angle));
```
4ian, line 387: Should the angle be checked/clamped so that it does not go out of some reasonable bounds?

4ian: (there is the check in the editor side, but maybe we should mirror it here? Wdyt?)

4ian (approved):
Looks good! 👍👍
Added two minor nitpickings, then should be good to go :)

Extensions/TopDownMovementBehavior/topdownmovementruntimebehavior.ts
D8H: Yes, of course. I also took one degree margin to be sure not to fall in limit cases.

PR #2238, opened 2021-01-25
Multiples changes to the p2p extension

arthuro555:
- Use a FIFO instead of LIFO "queue" in the no dataloss mode
- Use a class with a nice to use API instead of multiple global objects
- Improve typing
- Overall simplification of some code and improvements in code readability
- Optimisation of the events of the p2p example to be a better example

The changes have been tested on the p2p example and the default broker which worked flawlessly.

### DIFF
--- a/Extensions/P2P/B_p2ptools.ts
+++ b/Extensions/P2P/B_p2ptools.ts
@@ -15,7 +15,7 @@ namespace gdjs {
       };
 
       const isValidNetworkEvent = (event): event is NetworkEvent =>
-        typeof event.eventName === 'string' || typeof event.data === 'string';
+        typeof event.eventName === 'string' && typeof event.data === 'string';
 
       /**
        * The data bound to an event that got triggered.
@@ -47,7 +47,7 @@ namespace gdjs {
         /**
          * Returns true if the event is triggered.
          */
-        isTrigerred() {
+        isTriggered() {
           return this.data.length > 0;
         }
 
@@ -69,14 +69,14 @@ namespace gdjs {
         /**
          * Get the data sent with the last event triggering.
          */
-        getEventData() {
+        getData() {
           return this.data.length === 0 ? '' : this.data[0].data;
         }
 
         /**
          * Get the sender of the last event triggering.
          */
-        getEventSender() {
+        getSender() {
           return this.data.length === 0 ? '' : this.data[0].sender;
         }
       }
@@ -99,7 +99,6 @@ namespace gdjs {
       /**
        * Contains a map of events triggered by other p2p clients.
        * It is keyed by the event name.
-       * @note This is ignored if the event is in no dataloss mode.
        */
       const events: Record<string, Event> = {};
 
@@ -178,7 +177,7 @@ namespace gdjs {
         });
 
         // Close event is only for graceful disconnection,
-        // but we want onDisconnect to trigger for any type of dsiconnection,
+        // but we want onDisconnect to trigger for any type of disconnection,
         // so we register a listener for both event types.
         connection.on('error', () => {
           _onDisconnect(connection.peer);
@@ -233,7 +232,7 @@ namespace gdjs {
       ): boolean => {
         const event = getEvent(eventName);
         event.dataloss = defaultDataLoss;
-        return event.isTrigerred();
+        return event.isTriggered();
       };
 
       /**
@@ -303,7 +302,14 @@ namespace gdjs {
        * @returns - The data as JSON.
        */
       export const getEventData = (eventName: string) =>
-        getEvent(eventName).getEventData();
+        getEvent(eventName).getData();
+
+      /**
+       * Get the id of peer that caused the last trigger of an event.
+       * @param eventName - The event to get the sender from.
+       */
+      export const getEventSender = (eventName: string) =>
+        getEvent(eventName).getSender();
 
       /**
        * Get a variable associated to the last trigger of an event.
@@ -404,9 +410,9 @@ namespace gdjs {
       export const getConnectedPeer = (): string => connectedPeers[0] || '';
 
       gdjs.callbacksRuntimeScenePostEvents.push(() => {
-        for (const i in events) {
-          if (!events.hasOwnProperty(i)) continue;
-          events[i].popData();
+        for (const eventName in events) {
+          if (!events.hasOwnProperty(eventName)) continue;
+          events[eventName].popData();
         }
         if (disconnectedPeers.length > 0) {
           disconnectedPeers.shift();
--- a/Extensions/P2P/B_p2ptools.ts
+++ b/Extensions/P2P/B_p2ptools.ts
@@ -128,15 +128,6 @@ namespace gdjs {
       const connectedPeers: string[] = [];
 
       /**
-       * Internal function to get an event from {@link events}
-       * or create it if it doesn't exist.
-       */
-      const getEvent = (name: string) => {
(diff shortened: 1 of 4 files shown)


Extensions/P2P/B_p2ptools.ts
```diff
@@ -7,43 +7,101 @@ namespace gdjs {
...
        */
-      const connections: Record<string, Peer.DataConnection> = {};
+      class Event {
+        private data: EventData[] = [];
+        public dataloss = false;
+
+        /**
+         * Returns true if the event is triggered.
+         */
+        isTrigerred() {
```
4ian, line 50: Typo: `isTriggered`

Extensions/P2P/B_p2ptools.ts
```diff
@@ -7,43 +7,101 @@ namespace gdjs {
...
+       * The type of the data that is sent across peerjs.
        */
-      let peerConfig: Peer.PeerJSOption = { debug: 1 };
+      type NetworkEvent = {
+        eventName: string;
+        data: string;
+      };
+
+      const isValidNetworkEvent = (event): event is NetworkEvent =>
+        typeof event.eventName === 'string' || typeof event.data === 'string';
```
4ian, line 18: Should that be a && instead of ||?

Extensions/P2P/B_p2ptools.ts
```diff
@@ -7,43 +7,101 @@ namespace gdjs {
...
-      const lastEventData: Record<string, string | string[]> = {};
+      const connections: Record<string, Peer.DataConnection<NetworkEvent>> = {};
 
       /**
-       * Tells how to handle an event (with or without data loss).
-       * Maps the event name (string) to a boolean:
-       * true for dataloss, false for no dataloss.
+       * Contains a map of events triggered by other p2p clients.
+       * It is keyed by the event name.
+       * @note This is ignored if the event is in no dataloss mode.
```
4ian, line 102: Not sure I understand the @note here?

Extensions/P2P/B_p2ptools.ts
```diff
@@ -101,25 +168,19 @@ namespace gdjs {
...
-
-            (lastEventData[data.eventName] as string[]).push(data.data);
-          } else {
-            triggeredEvents[data.eventName] = true;
-            lastEventData[data.eventName] = data.data;
-          }
+          if (isValidNetworkEvent(data))
+            getEvent(data.eventName).pushData(
+              new EventData(data.data, connection.peer)
+            );
```
4ian, line 177: These Event/EventData classes are making things much cleaner 👍

Extensions/P2P/B_p2ptools.ts
arthuro555: Ah yes it is an outdated note sorry

Extensions/P2P/B_p2ptools.ts
```diff
@@ -168,25 +231,9 @@ namespace gdjs {
...
-          return returnValue;
-        } else {
-          let returnValue = lastEventData[eventName];
-          if (typeof returnValue === 'undefined') {
-            return false;
-          }
-          return returnValue.length !== 0;
-        }
+        const event = getEvent(eventName);
+        event.dataloss = defaultDataLoss;
```
4ian, line 235: I know that this was already implemented like this, but should have this "dataloss" assignment have been done in a separate "setEventDataloss" set up function (which would have been exposed as a set up action)? 
I guess the idea was to surface this choice directly in the condition?

arthuro555: > I guess the idea was to surface this choice directly in the condition?

Yes, I am afraid making it a separate event might make it a "scary advanced feature", but it is essential for things like position updates.

Extensions/P2P/B_p2ptools.ts
```diff
@@ -7,43 +7,101 @@ namespace gdjs {
...
+         * Deleted the last event data, to be called when it is sure the event was processed throughly.
+         */
+        popData() {
+          this.data.shift();
+        }
+
+        /**
+         * Get the data sent with the last event triggering.
+         */
+        getEventData() {
```
4ian, line 72: nitpicking: as we're already inside `Event` class, you should be able to call this just `getData`.
It will be "obvious" to the caller that this is the "event data", because the calling code will be something like `getEvent(...).getData()`.

Also it's consistent with push/popData.

Also do this for getEventSender => getSender :)

Extensions/P2P/B_p2ptools.ts
```diff
@@ -350,52 +381,38 @@ namespace gdjs {
...
-        for (const i in lastEventData) {
-          if (
-            typeof lastEventData[i] === 'object' &&
-            lastEventData[i].length > 0
-          ) {
-            (lastEventData[i] as string[]).pop();
-          }
+        for (const i in events) {
+          if (!events.hasOwnProperty(i)) continue;
+          events[i].popData();
```
4ian, line 409: nitpicking: `i` is a shorthand for `index`, which means to the reader that this is almost surely an array.
Prefer to name your key as what it represents: the event name, so `for (const eventName in events) { ...`

Extensions/P2P/B_p2ptools.ts
```diff
@@ -101,25 +168,19 @@ namespace gdjs {
...
-            lastEventData[data.eventName] = data.data;
-          }
+          if (isValidNetworkEvent(data))
+            getEvent(data.eventName).pushData(
+              new EventData(data.data, connection.peer)
+            );
         });
+
+        // Close event is only for graceful disconnection,
+        // but we want onDisconnect to trigger for any type of dsiconnection,
```
4ian, line 181: disconnection

4ian:
Thanks! This is indeed much cleaner :) 
If I understood correctly, the fix for FIFO instead of LIFO is thanks to the `.shift`. This is indeed needed to avoid treating old events after new ones.

arthuro555:
> Thanks! This is indeed much cleaner :)
If I understood correctly, the fix for FIFO instead of LIFO is thanks to the .shift. This is indeed needed to avoid treating old events after new ones.

Exactly :)
The API being more clear, it is also easier to use in JS events. It also added the "event sender", which wasn't there before and was one of the most badly missing features (users had to trust other machines to tell them their peer ID ☹️).

4ian:
Sounds good, great PR! 👍
Merging now. Will this probably fix the issue that was reported (can't remember where it was) regarding events being received after a long "lag" time?

Extensions/P2P/JsExtension.js
```diff
@@ -296,6 +296,22 @@ module.exports = {
       .addIncludeFile('Extensions/P2P/B_p2ptools.js')
       .setFunctionName('gdjs.evtTools.p2p.getEventData');
 
+    extension
+      .addStrExpression(
+        'GetEventSender',
```
4ian, line 301: I unfortunately missed this for `GetEventData` too, but expressions almost never have `Get` in their name, for conciseness and because it's implied all expressions are getter in GDevelop.

arthuro555:
> Will this probably fix the issue that was reported (can't remember where it was) regarding events being received after a long "lag" time?

🤔 Probably? This design mistake of mine might have cause lots of weird behavior due to new events being processed instantly and older taking ages to be handled

Extensions/P2P/JsExtension.js
arthuro555: Ah I see, I'll keep this in mind in the future

PR #2257, opened 2021-02-01
fix middle click drag not working while the path-editor is visible

blurymind:
To reproduce the issue, open yarn-editor in gdevelop,
make sure the file path editor is visible (not collapsed), then try to middle click drag to pan.
It wont work.

This PR fixes that


4ian:
Strange why it was not working before, but if it's better now then I'm happy with moving to flexbox!

PR #2258, opened 2021-02-01
signInWithProvider requires two parameters

jonasfj:
`gdjs.evtTools.firebase.auth.signInWithProvider` takes two parameters.

This is just a quick guess at why it's not working, in my debugger the first variable is variable container:

https://github.com/4ian/GDevelop/blob/153ccdb2830c3fd6f572b8f899bdc84d426a1048/Extensions/Firebase/C_firebasetools/D_authtools.ts#L414-L417

### DIFF
--- a/Extensions/Firebase/JsExtension.js
+++ b/Extensions/Firebase/JsExtension.js
@@ -321,6 +321,7 @@ module.exports = {
         'JsPlatform/Extensions/firebase.png',
         'JsPlatform/Extensions/firebase.png'
       )
+      .addParameter('string', _('Provider'), '', false)
       .addParameter(
         'scenevar',
         _('Callback variable with state (ok or error)'),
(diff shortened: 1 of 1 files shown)


Extensions/Firebase/JsExtension.js
```diff
@@ -321,6 +321,7 @@ module.exports = {
         'JsPlatform/Extensions/firebase.png',
         'JsPlatform/Extensions/firebase.png'
       )
+      .addParameter('string', _('Provider'), '', false)
```
4ian, line 324: Good catch! I recommend using `stringWithSelector`:
```
.addParameter(
        'stringWithSelector',
        _('Provider'),
        '["google", "facebook", "github", "twitter"]',
        false
      )
```

4ian:
cc @arthuro555 for review :)

arthuro555 (approved):
I'd make it a string with selector like 4ian said, but overall this looks like a nice fix.

Extensions/Firebase/JsExtension.js
jonasfj: Ah, nice, I looked for this but couldn't find it in any docs or other examples :D

4ian:
Thanks! 👍

Issue #2259, opened 2021-02-02
Tilemap - Display Mode (index) not working in pre-view


Entr0py404:
Hi I made a tilemap example to day https://www.youtube.com/watch?v=zqeHTgOSpt0
and ran into a issue:
Display mode ( index ) doesn't work in a pre-view but it works in the ide
time stamp here -> https://youtu.be/zqeHTgOSpt0?t=494

Silver-Streak:
@Entr0py404 had been mentioning it on the discord. They're running b104 preview. I have some deadline stuff this week and don't have time to test, but the video does definitely seem to match up.

4ian:
> Edit: Likewise, it seems like the boundary box issue is still not fixed.

It's not yet fixed indeed.

cc @blurymind any idea what could mess up the index display mode?

blurymind:
I couldnt figure out why the height of the boundary box is wrong

@Entr0py404 can you share your project please? It will help with finding out why its not rendering

Entr0py404:
[TileMap Test.zip](https://github.com/4ian/GDevelop/files/5913545/TileMap.Test.zip)

Entr0py404:
By the looks of it the bounding box is only accounting for used tile spaces, blank tile spaces seem to be ignored

blurymind:
hmm, that is strange. Works fine for me.
 
![weirdIndexBug](https://user-images.githubusercontent.com/6495061/106646053-1e662b00-6585-11eb-928b-f435e6c22bef.gif)

@4ian @Silver-Streak can you reproduce this bug?

Entr0py404:
I'm sorry I updated my comment it's the only using one map in the project

Entr0py404:
Thank you so very much for the tile map and the support!

blurymind:
fix here
https://github.com/4ian/GDevelop/pull/2260
@Entr0py404 thank you for reporting this and for the demo 👍 

A rather embarasing bug 😊

blurymind:
I filed a ticket for the boundary box problem
https://github.com/pixijs/pixi-tilemap/issues/108

4ian:
@blurymind why is index converted to a string though? This should never be the case, `layerIndex` is typed as a number and if something set it to a string, then the bug is to be fixed at this location :)

4ian:
Thanks for the report to pixi-tilemap! :)

blurymind:
[...]

PR #2260, opened 2021-02-02
fix tilemap layerIndex not working in engine

blurymind:
fixes https://github.com/4ian/GDevelop/issues/2259

### DIFF
--- a/Extensions/TileMap/JsExtension.js
+++ b/Extensions/TileMap/JsExtension.js
@@ -45,6 +45,8 @@ module.exports = {
       if (propertyName in objectContent) {
         if (typeof objectContent[propertyName] === 'boolean')
           objectContent[propertyName] = newValue === '1';
+        else if (typeof objectContent[propertyName] === 'number')
+          objectContent[propertyName] = parseFloat(newValue);
         else objectContent[propertyName] = newValue;
         return true;
       }
--- a/Extensions/TileMap/JsExtension.js
+++ b/Extensions/TileMap/JsExtension.js
@@ -42,12 +42,32 @@ module.exports = {
       propertyName,
       newValue
     ) {
-      if (propertyName in objectContent) {
-        if (typeof objectContent[propertyName] === 'boolean')
-          objectContent[propertyName] = newValue === '1';
-        else if (typeof objectContent[propertyName] === 'number')
-          objectContent[propertyName] = parseFloat(newValue);
-        else objectContent[propertyName] = newValue;
+      if (propertyName === 'tilemapJsonFile') {
+        objectContent.tilemapJsonFile = newValue;
+        return true;
+      }
+      if (propertyName === 'tilesetJsonFile') {
+        objectContent.tilesetJsonFile = newValue;
+        return true;
+      }
+      if (propertyName === 'tilemapAtlasImage') {
+        objectContent.tilemapAtlasImage = newValue;
+        return true;
+      }
+      if (propertyName === 'displayMode') {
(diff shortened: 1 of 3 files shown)


Extensions/TileMap/tilemapruntimeobject.ts
```diff
@@ -12,7 +12,7 @@ namespace gdjs {
     _tilesetJsonFile: string;
     _tilemapAtlasImage: string;
     _displayMode: string;
-    _layerIndex: float;
+    _layerIndex: number;
```
4ian, line 15: Please keep using float, it's more precise than number.

blurymind:
@4ian I tried changing it as you suggested and pushed the change, it is still a string!

if you change the != back to !== it will fail 🤷

Extensions/TileMap/JsExtension.js
```diff
@@ -45,6 +45,8 @@ module.exports = {
       if (propertyName in objectContent) {
         if (typeof objectContent[propertyName] === 'boolean')
           objectContent[propertyName] = newValue === '1';
+        else if (typeof objectContent[propertyName] === 'number')
```
4ian: The issue with this is that if something/somebody make a mistake in the code and store the property to a string, then it will never be changed back to a number? It will stay, wrongly, a string forever. This includes games that have saved "layerIndex" as a string.

This is why I recommend going back to an explicit approach as I mentioned in the other thread.

Extensions/TileMap/tilemapruntimeobject.ts
blurymind: the index value is never a float though. It really is an integer

Extensions/TileMap/JsExtension.js
blurymind: i updated that already :)

Extensions/TileMap/tilemapruntimeobject.ts
4ian: Oops my bad! Use `integer` then :)

Extensions/TileMap/pixi-tilemap-helper.js
```diff
@@ -338,7 +338,7 @@
     pixiTileMap.clear();
 
     genericTileMapData.layers.forEach(function (layer, index) {
-      if (displayMode === 'index' && layerIndex !== index) return;
+      if (displayMode === 'index' && layerIndex != index) return;
```
4ian, line 341: Let's cancel this change, we really must fix the problem upward. This is not the library issue if someones is not providing the proper types, it's the caller responsibility to follow the proper API.

blurymind: updated now :)

4ian:
> I tried changing it as you suggested and pushed the change, it is still a string!

What if you now try to change the value? Can you check where the string is coming from? It is from the object data?
If yes, then modifying the property should set it back to a number.

blurymind:
> > I tried changing it as you suggested and pushed the change, it is still a string!
> 
> What if you now try to change the value? Can you check where the string is coming from? It is from the object data?
> If yes, then modifying the property should set it back to a number.

ah good point 😄 changing the value and works now.

4ian:
Cool! That's the bad part of these behaviors implemented in JS, they need to be modified to actually fix such an error - but hopefully this is not an error often made.

blurymind:
I hope that stronger types will make these errors harder to make in the future! This seems to work well now, I just have to always remember to put in all the boilerplate

PR #2274, opened 2021-02-04
Set types in the Path Finding extension to ease future PR

D8H:
I tested these 2 cases with official example:
- path found
- no path found

### DIFF
--- a/Extensions/PathfindingBehavior/pathfindingruntimebehavior.ts
+++ b/Extensions/PathfindingBehavior/pathfindingruntimebehavior.ts
@@ -13,28 +13,28 @@ namespace gdjs {
     _path: Array<FloatPoint> = [];
 
     //Behavior configuration:
-    _allowDiagonals: any;
-    _acceleration: any;
-    _maxSpeed: any;
-    _angularMaxSpeed: any;
-    _rotateObject: any;
-    _angleOffset: any;
-    _cellWidth: float;
-    _cellHeight: float;
-    _extraBorder: any;
+    _allowDiagonals: boolean;
+    _acceleration: float;
+    _maxSpeed: float;
+    _angularMaxSpeed: float;
+    _rotateObject: boolean;
+    _angleOffset: float;
+    _cellWidth: integer;
+    _cellHeight: integer;
+    _extraBorder: float;
 
     //Attributes used for traveling on the path:
     _pathFound: boolean = false;
-    _speed: number = 0;
-    _angularSpeed: number = 0;
-    _timeOnSegment: number = 0;
+    _speed: float = 0;
+    _angularSpeed: float = 0;
+    _timeOnSegment: float = 0;
     _totalSegmentTime: float = 0;
-    _currentSegment: number = 0;
+    _currentSegment: integer = 0;
     _reachedEnd: boolean = false;
-    _manager: any;
-    _searchContext: any;
+    _manager: PathfindingObstaclesManager;
+    _searchContext: PathfindingRuntimeBehavior.SearchContext;
 
-    constructor(runtimeScene, behaviorData, owner) {
+    constructor(runtimeScene: RuntimeScene, behaviorData, owner: RuntimeObject) {
       super(runtimeScene, behaviorData, owner);
 
       //The path computed and followed by the object (Array of arrays containing x and y position)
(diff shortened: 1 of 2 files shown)


Extensions/PathfindingBehavior/pathfindingruntimebehavior.ts
```diff
@@ -574,7 +573,9 @@ namespace gdjs {
           //Make sure we do not search forever.
           const n =
             //Get the most promising node...
-            this._openNodes.shift();
+            this._openNodes.shift() || null;
+          if (n == null)
+            return false;
```
D8H, line 578: I'm not sure about that.

Extensions/PathfindingBehavior/pathfindingruntimebehavior.ts
```diff
@@ -685,12 +686,12 @@ namespace gdjs {
 
         //No so construct a new node (or get it from the cache)...
         // @ts-ignore
-        let newNode: gdjs.PathfindingRuntimeBehavior.Node = null;
+        let newNode: Node | undefinied;
```
4ian, line 689: I generally prefer to keep null rather than undefined, as null is explicit (this was set to null to signify that there is no object), rather than undefined (which is set even if you don't write it, and so which indicates much more often an error whenever you see it).

Extensions/PathfindingBehavior/pathfindingruntimebehavior.ts
```diff
@@ -607,14 +608,14 @@ namespace gdjs {
             }
           }
         }
-        this._allNodes = {};
+        this._allNodes = [];
```
4ian, line 611: Nice catch :)

Extensions/PathfindingBehavior/pathfindingruntimebehavior.ts
4ian: Because of the while condition, `this._openNodes.length !== 0`, the result of `shift()` can't be undefined, as the array is not empty. So we should be able to keep the existing code there I think?

Extensions/PathfindingBehavior/pathfindingruntimebehavior.ts
```diff
@@ -479,24 +477,25 @@ namespace gdjs {
       _rightBorder: integer = 0;
       _topBorder: integer = 0;
       _bottomBorder: integer = 0;
-      _distanceFunction: any;
-      _allNodes: any = [];
+      _distanceFunction: Function;
```
4ian, line 480: Let's type this entirely for more safety. Something like `(pt1: FloatPoint, pt2: FloatPoint) => float` I guess.

Extensions/PathfindingBehavior/pathfindingruntimebehavior.ts
```diff
@@ -348,7 +346,7 @@ namespace gdjs {
       }
     }
 
-    doStepPreEvents(runtimeScene) {
+    doStepPreEvents(runtimeScene: RuntimeScene) {
```
4ian, line 349: For consistency with the rest of the codebase, I prefer to keep the `gdjs.` prefix everywhere.
It's slightly longer to write/read, but not that much and gives a very clear context (`gdjs.Xxx` => you're in the game engine).
Also make it easy to find types everywhere in the codebase by searching for `: gdjs.RuntimeScene` or even just `gdjs.RuntimeScene`. `RuntimeScene` alone would give too much noise and false positive.
I like this kind of "greppability", important for keeping refactoring and search easy  :)

Extensions/PathfindingBehavior/pathfindingruntimebehavior.ts
D8H: It's because of the shift signature:
Array<gdjs.PathfindingRuntimeBehavior.Node>.shift(): Node | undefined

Maybe iterate on shift?
for (let n = this._openNodes.shift(); n; n = this._openNodes.shift())

4ian: I would keep the while as I think it better express the intention (in A*, you continue as long as you have open nodes) but maybe just use a // @ts-ignore to tell that `invariant: _openNodes can't be empty, so n can't be null.`?

D8H: ok, I use the "!" operator.

Extensions/PathfindingBehavior/pathfindingruntimebehavior.ts
```diff
@@ -567,17 +566,15 @@ namespace gdjs {
...
           }
 
-          //Make sure we do not search forever.
-          const n =
-            //Get the most promising node...
-            this._openNodes.shift();
-          n.open = false;
-
+          //Get the most promising node...
+          const n = this._openNodes.shift()!;
```
4ian, line 575: Ah nice I forgot that TypeScript has this syntax. Sounds fair enough then!

4ian:
Looks good, thanks for improving this! I've run the autoformatting for you (`npm run format` in GDJS).

PR #2275, opened 2021-02-05
Add expression normalize

Bouh:
- C++ and JS
- Use it in video object instead private function

- [x] GDJS compile success
- [x] Tested in a empty game

### DIFF
--- a/GDCpp/GDCpp/Extensions/Builtin/MathematicalTools.cpp
+++ b/GDCpp/GDCpp/Extensions/Builtin/MathematicalTools.cpp
@@ -10,6 +10,10 @@
 namespace GDpriv {
 namespace MathematicalTools {
 
+double GD_API normalize(double expression, double min, double max) { 
+  return (expression - min) / (max - min);
+}
+
 double GD_API clamp(double expression, double min, double max) { 
   return std::min(std::max(expression, min), max);
 }
--- a/GDCpp/GDCpp/Extensions/Builtin/MathematicalTools.cpp
+++ b/GDCpp/GDCpp/Extensions/Builtin/MathematicalTools.cpp
@@ -11,7 +11,7 @@ namespace GDpriv {
 namespace MathematicalTools {
 
 double GD_API normalize(double expression, double min, double max) { 
-  return (expression - min) / (max - min);
+  return (min = max ? max : (expression - min) / (max - min));
 }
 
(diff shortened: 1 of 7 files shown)


GDJS/Runtime/events-tools/commontools.ts
```diff
@@ -131,6 +131,21 @@ namespace gdjs {
...
+       * @param {number} min The minimum
+       * @param {number} max The maximum
+       * @returns The normalized value between 0 and 1
+       */
+      export const normalize = function (
+        val: float,
+        min: float,
+        max: float
+      ): number {
+        return (val - min) / (max - min);
```
4ian, line 146: We should handle the case `max === min`. In this case, return `max`, as you've normalized a value to a range which is a single number, so the normalized number *is* this number.

4ian: (same in C++ to avoid a crash)

Bouh: Done

GDCpp/GDCpp/Extensions/Builtin/MathematicalTools.cpp
```diff
@@ -11,7 +11,7 @@ namespace GDpriv {
 namespace MathematicalTools {
 
 double GD_API normalize(double expression, double min, double max) { 
-  return (expression - min) / (max - min);
+  return (min = max ? max : (expression - min) / (max - min));
```
4ian, line 14: min == max
Can you put an extra pair of parenthesis around the `(expression - min) / (max - min)`? I'm afraid otherwise it could be understand as `max/(max-min)` in the case `min == max`, it's always unclear what get the priority first.

Bouh: Done
I've made a test with an online sandbox for c++, but yeah it better to be sure.

4ian:
Thanks! Just missing the autoformatting in GDJS ;) Then it's fine to merge :)

Bouh:
Prettier does nothing else for me.

4ian:
Are you sure you running it in the GDJS folder?

Bouh:
I had never added this in my macros before indeed!

4ian:
It's fairly new :) We could call this in the `npm run format` of newIDE to make things easier.

PR #2279, opened 2021-02-05
Add Spatial Sound extension

arthuro555:
This PR adds a spatial sounds extension that uses the howler 3d sound extension.
It also 
- Updates howler.min.js to v2.2.1
- Use howler core only in GDJS
- Add an example showing how to use it (fixtures not regenerated)

Note that spatial sound is only available with the WebAudio API, so music (html5 audio) cannot use this extension, only sounds.


arthuro555:
Closes #1955

newIDE/app/resources/examples/spatial-sound/README.md
```diff
@@ -0,0 +1 @@
+A small similar to the Lost Woods of The Legend of Zelda Ocarina of Time, that uses spatial sound.
```
4ian, line 1: missing a word here I think ;)

4ian:
Thanks for the PR :)

> Updates howler.min.js to v2.2.1
> Use howler core only in GDJS

Do you know if there are interesting bug fixes in this version?

> Note that spatial sound is only available with the WebAudio API, so music (html5 audio) cannot use this extension, only sounds.

In what case this applies? I can see the bug reports coming in already because of this 😬 Can we make this obvious to the user?

arthuro555:
> Do you know if there are interesting bug fixes in this version?

If I remember correctly it fixes wav support on Safari 14 and an edge case causing an infinite loop

> Can we make this obvious to the user?

The way I have done it is simply exposing only an action for sounds and not an action for music, but I guess we can add a warning on the action?

4ian:
> The way I have done it is simply exposing only an action for sounds and not an action for music

I missed this, this sounds good then!

Extensions/SpatialSound/JsExtension.js
```diff
@@ -0,0 +1,85 @@
...
+      ),
+      'Arthur Pacaud (arthuro555)',
+      'MIT'
+    );
+
+    extension
+      .addAction(
+        'SetSoundPosition',
+        _('Set position of sound'),
+        _('Sets the spatial position of a sound.'),
```
4ian, line 42: Can we write in the description something about the units of the position? Like if I'm at pos 0,0,0 and the sounds is at 2,0,0, will I hear the sound? Basically, "how far" is "1"? :)

Extensions/SpatialSound/JsExtension.js
```diff
@@ -0,0 +1,85 @@
...
+  createExtension: function (
+    _ /*: (string) => string */,
+    gd /*: libGDevelop */
+  ) {
+    const extension = new gd.PlatformExtension();
+    extension.setExtensionInformation(
+      'SpatialSound',
+      _('Spatial sound'),
+      _(
+        'Allow positioning sounds in 3D space, using a stereo system to make the player know where a sound came from.'
```
4ian, line 32: I would say:
"Allow positioning sounds in a 3D space. The stereo system of the device is used to simulate the position of the sound and to give the impression that the sound is located somewhere around the player."

4ian (approved):
A few minor changes/precisions and we should be good to go :)

Extensions/SpatialSound/JsExtension.js
arthuro555: I am not sure to be honest. All I could find about WebAudio spatialization is [this resource](https://www.w3.org/TR/webaudio/#Spatialization-panning-algorithm), and it doesn't make it very clear at which point you wouldn't hear anymore (unless I overread something)

4ian: See this interesting article: https://medium.com/@kfarr/understanding-web-audio-api-positional-audio-distance-models-for-webxr-e77998afcdff

The attenuation follows an *inverse distance model*, meaning that by default the sound will be reduce by two at a distance of 2, and divided by 2 again at a distance of 4.

4ian:
Merged, thanks 👍

PR #2284, opened 2021-02-08
Add support for isometry in the path finding behavior

D8H:
This needs more testing, but it seems to work:
https://games.gdevelop-app.com/game-a2d8dd5a-92ab-48b8-9fe4-8367c38d7c77/index.html


4ian:
Interesting! Tag when when you think it's ready for review, unless it's already ready? :)

blurymind:
Can you also include the example project - as part of this PR? It will help with testing for regressions in the future

D8H:
> 
> 
> Can you also include the example project - as part of this PR? It will help with testing for regressions in the future

Should I put it in the test folder? a zip or the project folder?

blurymind:
> > Can you also include the example project - as part of this PR? It will help with testing for regressions in the future
> 
> Should I put it in the tests subfolder? a zip or the project folder?

Put it in https://github.com/4ian/GDevelop/tree/master/newIDE/app/resources/examples 😄 

Thanks for implementing this, looks cool

4ian:
> A suggestion on the example - can you remove the yellow tile under her feet, which moves with her? It misleads the player into thinking that the yellow tiles are walkable, when in fact they are walls

Might be a good idea indeed to use another color :) 

The example is indeed very cool and serves both as an example for people exploring the possibilities of the behavior, and as a test case when we do changes later to this.
Put it in the newIDE/app/resources/examples. The name of the folder should be in `kebab-case`, and the name of the name `kebab-case.json` (same base name for both). Finally, add a README.md with a short (1-2 sentences) description.

Extensions/PathfindingBehavior/pathfindingobstacleruntimebehavior.ts
```diff
@@ -190,7 +197,38 @@ namespace gdjs {
     doStepPostEvents(runtimeScene: gdjs.RuntimeScene) {}
 
     getAABB() {
-      return this.owner.getAABB();
+      if (!this._hitBoxesAABBUpToDate) {
```
4ian, line 200: Can you tell more about this? In which situations this caching works and what's the idea behind the invalidation of the cache?

I see a `this._hitBoxesAABBUpToDate = false` in pathfindingruntimebehavior but not in this file, which would indicate that the hitboxes are never invalidated?

More generally, I wonder if we can avoid transforming the AABB at all, because I see a lot of risk of having them de-synced with the rest (object changing, the isometry viewpoint changing...) and instead do the pathfinding search in a way that each position considered is following the isometry transformation?

D8H: Yes, I forgot to copy/paste some lines.
The thing is that the transformation of the AABB is not the same as the AABB of the transformed hitbox plygon.

Extensions/PathfindingBehavior/pathfindingruntimebehavior.ts
```diff
@@ -2,7 +2,7 @@
 GDevelop - Pathfinding Behavior Extension
 Copyright (c) 2010-2016 Florian Rival (Florian.Rival@gmail.com)
  */
-type FloatPoint = [number, number];
+type FloatPoint = Array<float>;
```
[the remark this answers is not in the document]
D8H: It a contagion from gdjs.Polygon.

D8H:
I added the example. I made it a little less ugly:
https://games.gdevelop-app.com/game-d44aeb68-2031-4ab5-b66b-fec3ccae27a6/index.html
I'm note sure I can use Minda. How can I credit the author?

4ian:
> What do you think about an hidden example list for devs?

You can put games used for feature testing in GDJS/tests/games, in a folder named after the name of the feature :) Or (actually, it's better!) in Extensions/PathfindingBehavior/tests/games (create a new folder for it). Second option is better because we keep test games close to the extension sources.

Extensions/PathfindingBehavior/pathfindingruntimebehavior.ts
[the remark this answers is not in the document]
4ian: Fair enough then

Extensions/PathfindingBehavior/pathfindingobstacleruntimebehavior.ts
4ian: Do you think that if, instead of AABB, we used more precise methods to check if a point is inside a polygon (i.e: if we use object hitboxes), then we could avoid any "transformation" of AABB at all, and we could just apply a different distance function to the algorithm, and a different X/Y position transformation to nodes?
So that we go from the left hand side diagram to the right one:

image

This would make the search to be following the "isometry" viewpoint like seen on the screen, without transforming AABB/hitboxes of objects.

This means that objects hitboxes, would be already defined in the editor like this: 
image
(which is what is done in the isometry example, using big red rectangle obstacles: 
image
 )
 
 Not sure if it makes sense, let me know :) Being able to search "naturally" in a viewpoint like this with existing polygons would reduce the number of transformations to do I think, and would work directly with hitboxes as long as they are well positioned already like the ones I shown.

4ian: Yep - note that I'm not sure how easy/hard it will be rework the behavior to use hitboxes rather than just AABB (and we might want to keep a boolean to use just AABB, to keep compatibility with existing games and improve performance in cases where hitboxes are too costly to use).
But it would be great to do, because I think it has confused users already, and it would open the way to this isometry-enabled pathfinding by "just" exploring nodes using different positions/distance function.

D8H: > 
> 
> Yep - note that I'm not sure how easy/hard it will be rework the behavior to use hitboxes rather than just AABB (and we might want to keep a boolean to use just AABB, to keep compatibility with existing games and improve performance in cases where hitboxes are too costly to use).


Yes, I think the boolean parameter is a must because the margin parameter can't be easily used with hitboxes so the exact collision mode will not behave the same even for default hitboxes.

4ian: Yeah exactly. Though this approach can also create another problem. A 1px space between two obstacles could or could not be considered as an open node (i.e: a path to explore) according to if the virtual nodes are exactly positioned between them or not.

Not sure how important this is, but I think this problem was less common in the approach with AABB because even if you add/remove some offset to your virtual grid position, it won't in most cases open new "paths" between obstacles, because if an AABB is even slightly on a virtual cell (i.e: a node), then the node is considered as having the obstacle on it.

Not sure what's the right solution yet. Should we do it the other way, don't check if the center of the node is inside an obstacle, but rather check for every point of the obstacle hitbox if it's inside the node. But not sure how complex this is.

D8H: I'm not sure to follow.
My understanding is that the shape of the object is stamped on each cell to test the collision with surroundings obstacles.
If we use the hitboxes instead of the AABB as the shape. I don't see why interstices of 1px could be an issue.
Do you mean that using the node as the shape could be an option? I'm not sure to see use cases.

About the performance issue, if hitboxes collision is too heavy, I wonder if RuntimeObject could have a isAABBMatchingHitboxes() to allow optimizations because I think it is commonly the case.
Maybe, inner AABB (I don't know the right term) could be interesting to ensure collisions the same way outer AABB discredit it.

D8H:
I push a version that doesn't rely of transformed shapes in the obstacles. It needs tests and polish, but it could be used as a discussion aid.

D8H:
I did a comparison of the 3 collision methods so we can discuss about what behavior we want:
https://games.gdevelop-app.com/game-986b44e9-f0c7-412a-910b-fadacd114135/index.html

Extensions/PathfindingBehavior/pathfindingobstacleruntimebehavior.ts
```diff
@@ -105,6 +114,21 @@ namespace gdjs {
         });
       }
     }
+    /**
+     * Tool class with a getAABB method.
+     * @ignore
+     */
+    export class AABBContainer {
```
4ian, line 121: nitpicking: maybe not export the class if we don't use it outside. Otherwise we risk polluting gdjs namespace.

Extensions/PathfindingBehavior/pathfindingruntimebehavior.ts
```diff
@@ -295,16 +354,33 @@ namespace gdjs {
...
+            (owner.getY() - owner.getDrawableY()) +
+            this._extraBorder
+        );
+      }
+      else {
+        const copyHitboxes = PathfindingRuntimeBehavior.deepCloneHitboxes(owner.getHitBoxes());
+        for (let pi = 0; pi < copyHitboxes.length; ++pi) {
+          copyHitboxes[pi].move(-owner.getX(), -owner.getY());
+        }
+        this._searchContext.setObjectHitBoxes(copyHitboxes);
```
4ian, line 374: nitpicking: the "deep cloning" might be the responsibility of `setObjectHitBoxes`?

Extensions/PathfindingBehavior/pathfindingruntimebehavior.ts
```diff
@@ -330,6 +407,27 @@ namespace gdjs {
       //Not path found
       this._pathFound = false;
     }
+    
+    static deepCloneHitboxes(hitboxes: Polygon[]): Polygon[] {
```
4ian, line 411: Might be worth thinking about putting this as a static of `gdjs.Polygon` or `gdjs.RuntimeObject`.

Extensions/PathfindingBehavior/pathfindingruntimebehavior.ts
```diff
@@ -728,21 +874,42 @@ namespace gdjs {
...
-
-            //Superimpose obstacles
+            yPos < bottomRightCellY;
+          }
+          else {
+            this._aabb.min[0] = nodeCenter[0] - this._leftBorder;
+            this._aabb.min[1] = nodeCenter[1] - this._topBorder;
+            this._aabb.max[0] = nodeCenter[0] + this._rightBorder;
+            this._aabb.max[1] = nodeCenter[1] + this._bottomBorder;
+            this._obstacles.getAllObstaclesAroundAABB(this._aabb, this._closeObstacles);
```
4ian, line 888: Wasn't this done just before? Or I'm getting confused about this is used for? Especially considered this will be called for every `this._closeObstacles` member, and it's writing in `this._closeObstacles`, this looks suspicious?

Extensions/PathfindingBehavior/pathfindingruntimebehavior.ts
```diff
@@ -756,12 +923,44 @@ namespace gdjs {
...
+          for (let vi = 0; vi < objectPolygon.vertices.length; ++vi) {
+            const objectPoint = objectPolygon.vertices[vi];
+            const stampPoint = stampPolygon.vertices[vi];
+            stampPoint[0] = objectPoint[0] + x;
+            stampPoint[1] = objectPoint[1] + y;
+          }
+        }
+      }
+
+      _checkCollision(hitboxes: Polygon[]): boolean {
```
4ian, line 939: The method name should I think indicate that this is checking collision for the stamp hitboxes

Extensions/PathfindingBehavior/pathfindingruntimebehavior.ts
```diff
@@ -728,21 +874,42 @@ namespace gdjs {
...
+            this._aabb.min[1] = nodeCenter[1] - this._topBorder;
+            this._aabb.max[0] = nodeCenter[0] + this._rightBorder;
+            this._aabb.max[1] = nodeCenter[1] + this._bottomBorder;
+            this._obstacles.getAllObstaclesAroundAABB(this._aabb, this._closeObstacles);
+            
+            if (this._behavior._collisionMethod == PathfindingRuntimeBehavior.HITBOX_COLLISION) {
+              this._moveStampHitBoxesTo(nodeCenter[0], nodeCenter[1]);
+              objectsOnCell = this._checkCollision(obj.getHitBoxes());
+              
+              obj.getHitBoxes()[0].computeCenter();
```
4ian, line 894: Why is this needed? 
Looks fairly dangerous to manually go ahead and modify the hitboxes of an object (even if it's just calling computeCenter - it's the responsibility of the object to do so. And objects could have multiple hitboxes).

Or is it just for debug?

4ian (commented):
Thanks for the test game and the update to the PR :) 

I've posted a few comments, though I've maybe not yet grasped the full extent of the changes. 

This is making me think that we're starting to have some technical debt growing in this part of the app:
- Not using AABB/hitboxes (but your PR is solving this :))
- We should probably support/force the sharing of the nodes between the objects, and have an option to memorize the nodes so that the "in memory map" is only created once. And then have methods to invalidate the nodes at some position.
  - The idea is to avoid having to recompute the nodes everytime we search a path for an object. Useful if you have a large map but nothing is moving. If something moves, we would give an action to invalidate the nodes
- Maybe one day this should be moved to a web worker.
- We should probably have at some point some unit tests for this, to avoid obvious or not so obvious regression in edge cases.
- You also mentioned in another issue some problems with HSHG, might be good to move to RBush like the rest of the codebase.

We can't solve everything, but might be a good idea to work on this.

Appreciate your work on this so far :)

Extensions/PathfindingBehavior/pathfindingruntimebehavior.ts
D8H: I really can't do a copy/paste correctly :sweat_smile:

Extensions/PathfindingBehavior/pathfindingruntimebehavior.ts
D8H: Just for debug, I did a better log

D8H:
> We should probably have at some point some unit tests for this, to avoid obvious or not so obvious regression in edge cases.

I think so too because I have other ideas for this extension, it will save a lot of time. For future PR, I would like to add a step by step movement and an hexagonal grid because it will be useful for wargames.

> We should probably support/force the sharing of the nodes between the objects, and have an option to memorize the nodes so that the "in memory map" is only created once. And then have methods to invalidate the nodes at some position.
> The idea is to avoid having to recompute the nodes everytime we search a path for an object. Useful if you have a large map but nothing is moving. If something moves, we would give an action to invalidate the nodes

Optimization will be very interesting. Quad tree may be useful.

This makes me think that there is a big flaw in the new implementation: the hitboxes at the start may to be representative to the ones on the way because of animation and rotation.

I think the stories could be:
- my object always keep the same hitboxes don't mess with it
- I don't care about overlapping, be permissive (use the intersection of every animation in every direction, so basically a inner disk)
- I don't want overlapping even if this close paths where the object could have actually go through  (use the union of every animation in every direction, an outer disk).

Or maybe this is overthinking and people will have to choose there hitbox wisely... there is a change request for disk hitboxes right?

Extensions/PathfindingBehavior/pathfindingruntimebehavior.ts
D8H: I don't know because we give a shape so we have to center it on the origin first. I don't have clear view of how the roles should be dispatched. I think I will think of a global refactor. May be in a following PR ?

Extensions/PathfindingBehavior/pathfindingruntimebehavior.ts
D8H: I put a deepClone method for one polygon in Polygon. I don't know if the deepClone of Polygon[] is useful and where it should go. It's almost a one-liner now.

4ian: Looks good 👍

4ian:
> This makes me think that there is a big flaw in the new implementation: the hitboxes at the start may to be representative to the ones on the way because of animation and rotation.

That's something we don't really care about I would say, we assume that the user of the pathfinding will be aware that if a path is computed and the size/shape of the object if radically changed, it's not something solved by the pathfinding anymore.

> Or maybe this is overthinking and people will have to choose their hitbox wisely... there is a change request for disk hitboxes right?

Yup I think it's overthinking, most people will use simple hitboxes. There is a request change for circle hitboxes, but they can be already appromixated... anyway I would rather ask users to be mindful of their hitboxes rather than implementing complex logic for this. Most users will be fine because the hitbox is staying the same for the object whatever its animation (even for the player, it's easier to have an object that does not change its "shape", and makes for better collision handling).

To come back to the PR:
* Thanks for the change to RBush! Actually will be very very useful because in the future I might factor all the RBushes into a "RBush" (or Quadtree etc...) provided by the game engine. So extensions could use it directly without each maintaining their own spatial data structure.
  * Question: Is it working well? :) I don't anticipate much problems, but still because it's slightly different i wonder if you ran into edge cases?
* About the state of the PR:
  * Thanks for the tests! 😍
  * Do you think you could extract them + the RBush in a separate PR? That would be a much easier PR to review, and this one would become simpler (and safer if we add even more tests :))

Extensions/PathfindingBehavior/tests/commonpathfindingruntimebehavior.spec.js
```diff
@@ -0,0 +1,128 @@
...
+  };
+  runtimeScene.addObject(obstacle);
+
+  return obstacle;
+};
+
+// tests cases where every collisionMethod has the same behavior.
+describe('gdjs.PathfindingRuntimeBehavior', function() {
+["Legacy", "AABB", "HitBoxes"].forEach((collisionMethod) => {
+describe(`(collisionMethod: ${collisionMethod})`, function() {
```
4ian, line 28: Nitpicking: Run Prettier (I installed the extension in VS Code, should also be available for other IDEs) on the file (and basically any JS/TS file) to auto format the tests.

D8H: Formatters are important to have clean commits, but I prefer to do it right before the merge because they make it difficult for me to read the code. But I'm open to negotiations...

Extensions/PathfindingBehavior/pathfindingruntimebehavior.ts
D8H: >       * Do you think you could extract them + the RBush in a separate PR? That would be a much easier PR to review, and this one would become simpler (and safer if we add even more tests :))

I'll continue to write tests and make the PR after that because I would like to split the test and the RBush in 2 commits to make it easier to run the tests on the untouch implementation.

D8H:
>       * Do you think you could extract them + the RBush in a separate PR? That would be a much easier PR to review, and this one would become simpler (and safer if we add even more tests :))

I'll continue to write tests and make the PR after that because I would like to split the test and the RBush in 2 commits to make it easier to run the tests on the untouch implementation.

4ian:
> I would like to split the test and the RBush in 2 commits to make it easier to run the tests on the untouched implementation.

Ok sounds good :) Tag me here when you have them ready or in another PR :)

D8H:
> Ok sounds good :) Tag me here when you have them ready or in another PR :)

@4ian, I extracted it here:
https://github.com/4ian/GDevelop/pull/2284
The tests seem to run ok after and before the rbrush commit.

D8H:
Ran automatic tests, tank, ai, basic and isometric maze without any issue.

4ian:
Sorry I'm lagging behind reviews, I'll take a look tomorrow hopefully!

Extensions/PathfindingBehavior/pathfindingobstacleruntimebehavior.ts
```diff
@@ -85,6 +85,31 @@ namespace gdjs {
       result.length = 0;
       result.push.apply(result, nearbyPlatforms);
     }
+
+    /**
+     * Returns all the platforms around the specified object.
```
4ian, line 90: Can you update this comment? It's outdated :)

Extensions/PathfindingBehavior/tests/commonpathfindingruntimebehavior.spec.js
```diff
@@ -1,5 +1,8 @@
-var commonpathfindingtest = {
-  createScene: () => {
+// tests cases where every collisionMethod has the same behavior.
```
4ian, line 1: Do you think it would work to move these changes to the tests in a separate PR?
(apart from `'Legacy', 'AABB', 'HitBoxes'` and a few collision method related stuff).

The idea would be to make these changes in the tests of this PR as small as possible.

Extensions/PathfindingBehavior/tests/legacypathfindingruntimebehavior.spec.js
```diff
@@ -770,7 +692,7 @@ describe('gdjs.PathfindingRuntimeBehavior', function () {
           describe(`centered: ${objectCenteredOnCells},`, function () {
             ['left', 'right', 'down', 'up'].forEach((direction) => {
               describe(`direction: ${direction})`, function () {
-                legacypathfindingtest.doTests(
+                doLegacyPathFindingTests(
```
4ian, line 695: Same thing here, can we:
- do these changes in a separate PR, which is only about tests? :) 
- Move `doLegacyPathFindingTests ` inside the `describe` to avoid global variables ;)

newIDE/app/resources/examples/isometric-path-finding/README.md
```diff
@@ -0,0 +1 @@
+This shows how to have an object move automatically and avoid obstacles in an isometric game.
```
4ian, line 1: Thanks for the example game 👍👍

Extensions/PathfindingBehavior/tests/legacypathfindingruntimebehavior.spec.js
D8H: Ok, the changes on the tests will be merged before or after this one?

4ian: Before :) 
Basically the idea is to set up in a separate PR as much tests as possible, so that: 
- the existing behavior is very well tested
- this PR size decrease (because there is little change in the test compared to the one merged previously) 

and this will reduce the chance of myself missing bugs/mistakes during the review (because if I review both tests that were changed *and* the behavior that was changed, I can miss mistakes in both).

I know this is a bit of a pain, but this will be very useful. Basically a PR containing only tests, I can merge it very quickly :)

D8H: No, it's ok, I'm becoming a rebase expert :sunglasses:

4ian: Hehe! :) That's where git shines (and actually Mercurial might even be easier to use, you can go back and rebase your commits very easily).

newIDE/app/resources/examples/isometric-path-finding/README.md
D8H: I put every thing in an external layout because it's easier to copy/paste to make other levels to test, but it may be unsettling to a new comer.

D8H:
I remove the isometric angle settings, so the grid is set like the one from the scene editor.
See https://github.com/4ian/GDevelop/pull/2445#issuecomment-802850667

D8H:
@4ian Is this PR left behind because its size make it difficult to review?
If you want, I can split it in 2:
* The collision methods (legacy, AABB and hitboxes)
* The isometric grid

4ian:
Yes, that's pretty big to review so I've fallen behind this and I'm also worried of the long term maintainability of all these methods (though we can alleviate this problem with tests, but still), so it went off my radar.

Bouh:
What is the state of this PR?
It can fix this issue: https://github.com/4ian/GDevelop/issues/2689

D8H:
> What is the state of this PR?
> It can fix this issue: #2689

I think I'll close it. It will be almost superseded by the NavMesh pathfinding: https://github.com/4ian/GDevelop/discussions/2787
The only cases that are not covered are the grid based movements. But for top-down, it can be done with the existing behavior and for isometry, hexagonal or anything else, the extension with linked objects can be used: https://github.com/GDevelopApp/GDevelop-extensions/issues/140

Cheapetsgrow:
Passed

Issue #2286, opened 2021-02-08
ParticleEmitter.NbParticles() is not a function (Uncaught TypeError)


tristanbob:
## Describe the bug

When using the ParticleEmitter.NbParticles() expression, the screen goes solid black and the console shows the error: 

```
Uncaught TypeError: gdjs.New_32scene2Code.GDMyParticleEmitterObjects1[0].getNbParticles is not a function
    at Object.gdjs.New_32scene2Code.eventsList0 (code1.js:19)
    at RuntimeScene2.gdjs.New_32scene2Code.func [as _eventsFunction] (code1.js:38)
    at RuntimeScene2.renderAndStep (runtimescene.ts:465)
    at SceneStack.step (scenestack.ts:36)
    at runtimegame.ts:549
    at gameLoop (runtimegame-pixi-renderer.ts:520)
```

## To Reproduce

- Create a new particle emitter with default settings
- Try to save the number of particles as a variable

## Other details

- Tested on GDevelop 5 Beta 104

## Screenshot

![Screenshot 2021-02-07 194825](https://user-images.githubusercontent.com/8879811/107174352-3aa00880-6987-11eb-8e98-dcd5e6586ecb.jpg)

Bouh:
Fixed in #2289

PR #2289, opened 2021-02-08
Fix expression particleObject.NbParticles()

Bouh:
Fixe https://github.com/4ian/GDevelop/issues/2286

image

(I've open up a new branch on your repo sorry, it doesn't matter if i'm not credited on the note patch for this fix. I'll continue to open next ones on my repo, i was curious to try with a small thing)

### DIFF
--- a/Extensions/ParticleSystem/ParticleEmitterObject.h
+++ b/Extensions/ParticleSystem/ParticleEmitterObject.h
@@ -184,7 +184,6 @@ class GD_EXTENSION_API ParticleEmitterBase {
     destroyWhenNoParticles = enable;
   };
 
-  float GetNbParticles() const { return nbParticles; };
   float GetRendererParam1() const { return rendererParam1; };
   float GetRendererParam2() const { return rendererParam2; };
   float GetTank() const { return tank; };
@@ -264,7 +263,6 @@ class GD_EXTENSION_API ParticleEmitterBase {
 
   gd::String textureParticleName;
   RendererType rendererType;
-  std::size_t nbParticles;
   float rendererParam1;
   float rendererParam2;
   bool additive;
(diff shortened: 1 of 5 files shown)


Extensions/ParticleSystem/particleemitterobject.ts
```diff
@@ -732,6 +732,10 @@ namespace gdjs {
       }
     }
 
+    getNbParticles() {
```
4ian, line 735: This is the proper fix :) Though, can we name this `getParticleCount`, like in the renderer?
`getNbParticles` was not a very nice name, better change it.

4ian: (the action can keep its name, just change the function name :))

Extensions/ParticleSystem/ParticleEmitterObject.h
```diff
@@ -184,6 +184,7 @@ class GD_EXTENSION_API ParticleEmitterBase {
     destroyWhenNoParticles = enable;
   };
 
+  float GetNbParticles() const { return nbParticles; };
```
4ian, line 187: Not needed, this is the "editor" object data: you don't need to track how many particles there are... because you're not in the game ;)

GDevelop.js/Bindings/Bindings.idl
```diff
@@ -2394,6 +2394,7 @@ interface ParticleEmitterObject {
     void SetRenderingAdditive();
     void SetRenderingAlpha();
 
+    float GetNbParticles();
```
4ian, line 2397: Same remark as before, this is actually not needed, because we don't use this class in the game engine. This is "just" the object for the IDE - which does not care how many particles exist because it's not running the game. Only the TypeScript game engine has this information, thanks to `getParticleCount`

4ian (commented):
Thanks for the fix, added some comments to clean this.

Issue #2293, opened 2021-02-09
Wrong TileMap selection box/visibility issue when maps have empty space


krunkster:
## Issue Description
When using a TileMap with lots of empty space (commonly due to fixed sized levels) the tilemap will often be hidden from view.

I believe this is due to an optimization in the engine that tries to hide objects that are "off screen".
When using a TileMap with a lot of empty space you can see in the IDE that the "size" of the object in the scene editor is only as big as the tile content in the tile map NOT including the empty space.

## Workaround
If you have a map with a lot of empty space, try adding a "dummy" or placeholder tile in the corners of the map to trick the engine into "seeing" the true height of the tilemap.

## To Reproduce
[tilemap-visibility-issue.zip](https://github.com/4ian/GDevelop/files/5952558/tilemap-visibility-issue.zip)
The above reproduction has an artificially low resolutions to exacerbate the issue.
There are two tile maps in the scene.
Both maps are a fixed size of 33 tiles tall, 100 tiles wide, tile size of 8x8px
The first map has a roman numeral I and only tiles on the bottom of the map
The second map has a roman numeral II and tiles along the top of the map
Both are in the scene and visible in the IDE
When you preview however the first map will not be visible

## Other details
* Windows
* GDevelop 1.0.0 b105

4ian:
> I believe this is due to an optimization in the engine that tries to hide objects that are "off screen".

Correct, this is called "culling".

> When using a TileMap with a lot of empty space you can see in the IDE that the "size" of the object in the scene editor is only as big as the tile content in the tile map NOT including the empty space.

That's indeed a nice way to see the problem.

@blurymind We should fix this with a high priority because it's impacting users and a real rendering issue.

Thanks for the repro game! :)

blurymind:
I will have a look tonight :)

blurymind:
This is the ticket on pixi-tilemap btw
https://github.com/pixijs/pixi-tilemap/issues/108

blurymind:
If tiled exports x,y coordinates of tiles on their atlas and on the tilemap, I will be able to create a much better fix for this bug.

Bjorn mentioned that there is a ticket for that here
https://github.com/mapeditor/tiled/issues/2863

PR #2296, opened 2021-02-09
fix tilemap/width height incorrect - doesnt account for empty tiles

blurymind:
This should fix
https://github.com/4ian/GDevelop/issues/2293
[tilemap-visibility-issue.zip](https://github.com/4ian/GDevelop/files/5953991/tilemap-visibility-issue.zip)

### DIFF
--- a/Extensions/TileMap/JsExtension.js
+++ b/Extensions/TileMap/JsExtension.js
@@ -524,9 +524,10 @@ module.exports = {
         this._pixiObject.worldTransform.applyInverse(position, localPosition);
 
         // Check if the point is inside the object bounds
-        const originalWidth = this._pixiObject.width / this._pixiObject.scale.x;
+        const originalWidth =
+          this._pixiObject._tilemapWidth / this._pixiObject.scale.x;
         const originalHeight =
-          this._pixiObject.height / this._pixiObject.scale.y;
+          this._pixiObject._tilemapHeight / this._pixiObject.scale.y;
 
         return (
           localPosition.x >= 0 &&
@@ -673,14 +674,14 @@ module.exports = {
      * Return the width of the instance, when it's not resized.
      */
     RenderedTileMapInstance.prototype.getDefaultWidth = function () {
-      return this._pixiObject.width / this._pixiObject.scale.x;
+      return this._pixiObject._tilemapWidth / this._pixiObject.scale.x;
     };
 
     /**
      * Return the height of the instance, when it's not resized.
      */
     RenderedTileMapInstance.prototype.getDefaultHeight = function () {
-      return this._pixiObject.height / this._pixiObject.scale.y;
+      return this._pixiObject._tilemapHeight / this._pixiObject.scale.y;
     };
 
     objectsRenderingService.registerInstanceRenderer(
--- a/Extensions/TileMap/JsExtension.js
+++ b/Extensions/TileMap/JsExtension.js
@@ -525,14 +525,8 @@ module.exports = {
         this._pixiObject.worldTransform.applyInverse(position, localPosition);
 
         // Check if the point is inside the object bounds
-        const originalWidth =
-          (this._pixiTileMapData
-            ? this._pixiTileMapData.tilemapWidth
-            : this._pixiObject.width) / this._pixiObject.scale.x;
-        const originalHeight =
-          (this._pixiTileMapData
-            ? this._pixiTileMapData.tilemapHeight
-            : this._pixiObject.height) / this._pixiObject.scale.y;
+        const originalWidth = this.getDefaultWidth();
+        const originalHeight = this.getDefaultHeight();
 
         return (
           localPosition.x >= 0 &&
@@ -650,6 +644,10 @@ module.exports = {
       if (this._instance.hasCustomSize()) {
         this._pixiObject.width = this._instance.getCustomWidth();
         this._pixiObject.height = this._instance.getCustomHeight();
+        this._pixiObject.scale.x =
+          this._instance.getCustomWidth() / this.getDefaultWidth();
+        this._pixiObject.scale.y =
+          this._instance.getCustomHeight() / this.getDefaultHeight();
       } else {
         this._pixiObject.scale.x = 1;
         this._pixiObject.scale.y = 1;
@@ -679,22 +677,14 @@ module.exports = {
      * Return the width of the instance, when it's not resized.
      */
     RenderedTileMapInstance.prototype.getDefaultWidth = function () {
-      return (
-        (this._pixiTileMapData
-          ? this._pixiTileMapData.tilemapWidth
-          : this._pixiObject.width) / this._pixiObject.scale.x
-      );
+      return !!this._pixiTileMapData ? this._pixiTileMapData.tilemapWidth : 0;
     };
 
     /**
      * Return the height of the instance, when it's not resized.
      */
     RenderedTileMapInstance.prototype.getDefaultHeight = function () {
-      return (
-        (this._pixiTileMapData
-          ? this._pixiTileMapData.tilemapHeight
-          : this._pixiObject.height) / this._pixiObject.scale.y
-      );
+      return !!this._pixiTileMapData ? this._pixiTileMapData.tilemapHeight : 0;
     };
 
     objectsRenderingService.registerInstanceRenderer(
--- a/Extensions/TileMap/JsExtension.js
+++ b/Extensions/TileMap/JsExtension.js
@@ -642,8 +642,6 @@ module.exports = {
      */
     RenderedTileMapInstance.prototype.update = function () {
       if (this._instance.hasCustomSize()) {
-        this._pixiObject.width = this._instance.getCustomWidth();
-        this._pixiObject.height = this._instance.getCustomHeight();
         this._pixiObject.scale.x =
           this._instance.getCustomWidth() / this.getDefaultWidth();
         this._pixiObject.scale.y =
--- a/Extensions/TileMap/JsExtension.js
+++ b/Extensions/TileMap/JsExtension.js
@@ -514,6 +514,7 @@ module.exports = {
       );
 
       this._pixiObject = new Tilemap.CompositeRectTileLayer(0);
+      this._pixiTileMapData = null;
 
       // Implement `containsPoint` so that we can set `interactive` to true and
       // the Tilemap will properly emit events when hovered/clicked.
@@ -525,9 +526,9 @@ module.exports = {
 
         // Check if the point is inside the object bounds
         const originalWidth =
-          this._pixiObject._tilemapWidth / this._pixiObject.scale.x;
+          this._pixiTileMapData?.tilemapWidth / this._pixiObject.scale.x;
         const originalHeight =
-          this._pixiObject._tilemapHeight / this._pixiObject.scale.y;
+          this._pixiTileMapData?.tilemapHeight / this._pixiObject.scale.y;
 
         return (
           localPosition.x >= 0 &&
@@ -587,7 +588,7 @@ module.exports = {
         .get('tilesetJsonFile')
         .getValue();
 
-      const pixiTileMapData = PixiTilemapHelper.loadPixiTileMapData(
+      this._pixiTileMapData = PixiTilemapHelper.loadPixiTileMapData(
         (textureName) =>
           this._pixiResourcesLoader.getPIXITexture(this._project, textureName),
         tilesetJsonData
@@ -598,10 +599,10 @@ module.exports = {
         tilesetJsonFile
       );
 
-      if (pixiTileMapData) {
+      if (this._pixiTileMapData) {
         PixiTilemapHelper.updatePixiTileMap(
           this._pixiObject,
-          pixiTileMapData,
+          this._pixiTileMapData,
           displayMode,
           layerIndex,
           pako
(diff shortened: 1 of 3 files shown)


4ian:
Thanks for opening the PR :)
The question of Ivan was interesting, what is exactly an "empty tile"? Can you explain how we handle them with pixi tilemap?

blurymind:
> Thanks for opening the PR :)
> The question of Ivan was interesting, what is exactly an "empty tile"? Can you explain how we handle them with pixi tilemap?

I think right now we simply dont add empty tiles at all, so I suppose pixi-tilemap does not know about them at all.
Perhaps the ticket I opened is not right and should be - how do we add empty tiles so pixi tilemap resizes its boundary to  account for them

4ian:
I see! Are the "empty tiles" even stored in the Tiled layers data? I guess not? But maybe there are other info that could be used to see that the "bounding box" of the tiles is larger (i.e: there are empty tiles on the top, bottom, left or right?). I'm not very familiar with Tiled data, hence why I'm asking :)

blurymind:
>  I'm not very familiar with Tiled data, hence why I'm asking :)

tiled doesnt give us the absolute value of the tilemap, but it does give us enough to calculate it  ourselves :) I did exactly that in this PR and seems to resolve the problem

Extensions/TileMap/JsExtension.js
```diff
@@ -524,9 +524,10 @@ module.exports = {
         this._pixiObject.worldTransform.applyInverse(position, localPosition);
 
         // Check if the point is inside the object bounds
-        const originalWidth = this._pixiObject.width / this._pixiObject.scale.x;
+        const originalWidth =
```
4ian, line 527: Instead of storing `_tilemapWidth` (and height) inside the pixi object, keep it just in pixiTileMapData and store pixiTileMapData inside `this`: `this._pixiTileMapData`.
Then you can use it to get the size whenever it's needed :)  
In the constructor, set `this._pixiTileMapData` to null.
When using it, verify if it's not null. If it is null, consider the width/height to be 0 :)

Extensions/TileMap/tilemapruntimeobject-pixi-renderer.ts
```diff
@@ -146,11 +146,11 @@ namespace gdjs {
     }
 
     getWidth(): float {
-      return this._pixiObject.width;
+      return this._pixiObject._tilemapWidth;
```
4ian, line 149: Same, I recommend storing and using `this._pixiTileMapData` :)

4ian:
> An alternative approach I thought of was to fill empty tiles in tiled with empty textures, but that seems to break the rendering - maybe I didnt do it right + its more wasteful =>so I abandoned that idea over a simpler one

Yes, that's more complex and less efficient, this solution is right 👍

Silver-Streak:
@blurymind Just to check, is this change for the IDE only? It doesn't change how rendering happens on preview/exported games, correct?

4ian:
This won't change the rendering, only the selection rectangle in the editor (and ensure the "hitbox" of the object is correct in game, avoiding the engine to hide the object if you're too far from it)

blurymind:
@Silver-Streak this bug affects the runtime rendering - if you try with the example project - the ground does not render because the bounding box is not in the view.

My PR here fixes the bug in both the IDE and in the runtime!

Extensions/TileMap/JsExtension.js
blurymind: wouldnt that just create tons of more boilerplate to keep track of? :)

blurymind: Updated now 👍 😃

blurymind:
I found a different bug, with the example - sometimes the customSize flips out due to this._pixiObject.scale.y/x getting set to 4+, when it should be 1.something

Extensions/TileMap/JsExtension.js
```diff
@@ -675,14 +679,22 @@ module.exports = {
      * Return the width of the instance, when it's not resized.
      */
     RenderedTileMapInstance.prototype.getDefaultWidth = function () {
-      return this._pixiTileMapData?.tilemapWidth / this._pixiObject.scale.x;
+      return (
+        (this._pixiTileMapData
+          ? this._pixiTileMapData.tilemapWidth
+          : this._pixiObject.width) / this._pixiObject.scale.x
```
4ian, line 685: I think this is overly complex. If we don't have a pixiTileMapData, then we don't have any info at all about the tilemap, so we don't know the size. So we return 0:

```
if (! this._pixiTileMapData) return 0;

return this._pixiTileMapData.tilemapWidth / this._pixiObject.scale.x;
```

We never use `this._pixiObject.width`, because the object might not be of the proper size because of the tiles not being set (the "empty tiles").

Extensions/TileMap/JsExtension.js
```diff
@@ -526,9 +526,13 @@ module.exports = {
 
         // Check if the point is inside the object bounds
         const originalWidth =
-          this._pixiTileMapData?.tilemapWidth / this._pixiObject.scale.x;
+          (this._pixiTileMapData
```
4ian, line 529: Same, don't use the width or the height of the pixi object.

Don't try to use something because something else is not available. If you've decided that the pixi object size is not the proper one, then never use it.

`const originalWidth = this._pixiTileMapData ? (this._pixiTileMapData.tilemapWidth / this._pixiObject.scale.x) : 0;`

Extensions/TileMap/tilemapruntimeobject-pixi-renderer.ts
```diff
@@ -148,11 +148,11 @@ namespace gdjs {
     }
 
     getWidth(): float {
-      return this._pixiTileMapData?.tilemapWidth;
+      return this._pixiTileMapData ? this._pixiTileMapData.tilemapWidth : this._pixiObject.width;
```
4ian, line 151: `this._pixiTileMapData ? this._pixiTileMapData.tilemapWidth : 0`

like everywhere else. If your _pixiTileMapData is null, this means that your tilemap is not loaded.

Extensions/TileMap/JsExtension.js
blurymind: what if the object was resized?

Extensions/TileMap/JsExtension.js
blurymind: what if its been resized? I guess we still need to do something to differentiate between the customSize tilemap and the original size tilemap

blurymind:
@4ian I have a problem with the position in the runtime - it is wrong when the object has custom size :( have no idea why

Extensions/TileMap/JsExtension.js
```diff
@@ -644,6 +644,10 @@ module.exports = {
       if (this._instance.hasCustomSize()) {
         this._pixiObject.width = this._instance.getCustomWidth();
         this._pixiObject.height = this._instance.getCustomHeight();
+        this._pixiObject.scale.x =
```
blurymind, line 647: I added this bit to fix the tilemap's Y position jumping down when changing the width of one of the tilemaps.
This fixed it in the IDE, but its still broken on the runtime. Custom size messes up their position in the runtime

Extensions/TileMap/tilemapruntimeobject-pixi-renderer.ts
```diff
@@ -146,11 +148,11 @@ namespace gdjs {
     }
 
     getWidth(): float {
-      return this._pixiObject.width;
+      return this._pixiTileMapData ? this._pixiTileMapData.tilemapWidth : this._pixiObject.width;
     }
 
     getHeight(): float {
-      return this._pixiObject.height;
+      return this._pixiTileMapData ? this._pixiTileMapData.tilemapHeight: this._pixiObject.height;
```
blurymind, line 155: @4ian note that I am also touching the runtime renderer and indeed without this line - and without a custom size - one of the tilemaps will fail to show up when you playtest

blurymind: I suspect its caused by using this._pixiObject.scale to set scale. Pixi-tilemap is likely overriding it with crazy values

Extensions/TileMap/JsExtension.js
blurymind: I suspect its caused by using this._pixiObject to set scale, width and height. Pixi-tilemap is likely overriding those with crazy values

Extensions/TileMap/tilemapruntimeobject-pixi-renderer.ts
4ian: Ok I'll take a look! You might want to always use the tilemapWidth/tilemapHeight divided by the scale.x/scale.y

Extensions/TileMap/JsExtension.js
4ian: The new code in the IDE seems fine. Note that you should remove the two lines above:
```
 this._pixiObject.width = this._instance.getCustomWidth();
 this._pixiObject.height = this._instance.getCustomHeight();
```

because they are not useless (setting the scale x and y will change the width/height) (the width of a pixi object is simply `original width * scaleX`. When you change the width, Pixi actually changes the scale).

Extensions/TileMap/JsExtension.js
```diff
@@ -673,14 +677,14 @@ module.exports = {
      * Return the width of the instance, when it's not resized.
      */
     RenderedTileMapInstance.prototype.getDefaultWidth = function () {
-      return this._pixiObject.width / this._pixiObject.scale.x;
+      return !!this._pixiTileMapData ? this._pixiTileMapData.tilemapWidth : 0;
```
4ian, line 680: This looks correct :)

Extensions/TileMap/tilemapruntimeobject-pixi-renderer.ts
4ian: Yes, it's indeed I think the problem. 
In the IDE, you need to implement the **default** width/height (which is equal to `this._pixiTileMapData.tilemapWidth` or `this._pixiTileMapData.tilemapHeight`).

Here you must return the actual width or height of the obejct, so:

`this._pixiTileMapData ? this._pixiTileMapData.tilemapHeight * this._pixiObject.scale.y : 0`

Extensions/TileMap/tilemapruntimeobject-pixi-renderer.ts
4ian: My bad this was a mistake on my side. Should be:

`this._pixiTileMapData ? this._pixiTileMapData.tilemapWidth * this._pixiObject.scale.x : 0 `

You take the original width and multiply by the scale factor :) This should give you the proper size, as long as tilemapWidth is the proper original width.

Note that here it is the *actual width* that you need to return.
In the IDE, in `getDefaultWidth`, you return the default width.

Extensions/TileMap/JsExtension.js
```diff
@@ -673,14 +677,14 @@ module.exports = {
...
     };
 
     /**
      * Return the height of the instance, when it's not resized.
      */
     RenderedTileMapInstance.prototype.getDefaultHeight = function () {
-      return this._pixiObject.height / this._pixiObject.scale.y;
+      return !!this._pixiTileMapData ? this._pixiTileMapData.tilemapHeight : 0;
```
4ian, line 687: This also looks correct - as long as tilemapHeight is the right value.

4ian:
Give a last try to my latest comments - I did a mistake before by confusing getOriginalWidth (IDE) and getWidth (runtime).
If it does not work, I'll take a look.

blurymind:
@4ian I updated the pr with the review notes, its still off in the runtime with the demo project -  if you reduce the width of both tilemaps - in the runtime  it's scale.y value jumps to 4, which leads to a massive offset that pushes it off the screen
crazyYScale

I tried applying the fix I have in the IDE, but for some reason it doesnt work, because the width/height is wrong too

blurymind:
@4ian I cant seem to find a way to fix the custom size screwing up the position in the runtime. I am almost tempted to suggest to remove the ability to resize tilemaps completely. Personally I dont think there are any benefits in being able to do it any way :)

4ian:
Ok I'll try to see what I can do!

Silver-Streak:
@blurymind  Just to double check, would this issue also distort the tilemaps when rendered in preview?

I'm seeing some moderate distortion of my tiles. 

Here's an example:
A zoomed in picture of my tile, which is a 16x16 box with three one-pixel thick different color borders. I've zoomed in within GIMP and enabled the grid to show the actual 1:1 pixels. I've marked each tile at the bottom with 1 grey pixel to show 1:1 sizing.
image


Here's a quick snapshot of a 320x192 tilemap made in LDTK
Showing the gap at the top
image


Here's a quick snapshot of one tile from a 320x180 game, without it being fullscreen, zoomed in within GIMP: 
image

There's some relatively obvious distortion there even though it should be the original size.

I noticed it originally because I had a much more complex tilemap on my game and saw that pixels weren't 1:1 when the tilemap was scaled up or the game was fullscreened.

I can make an issue post but I didn't want to duplicate efforts if it is from this.

blurymind:
I'm not sure if it will, but you can confirm if you run the test project you have with the PR or by sharing it here so I can confirm?

The pr tries to address the problem of tilemap layers that have empty tiles causing the tilemap to not render correctly

Silver-Streak:
> While it does address the problem, you can still break it in some cases where the tilemap is scaled by the user. That is why I suggested removing the ability to scale a tilemap

This makes some sense. I originally noticed it when I was trying to scale up a tilemap in my main game. If scaling is broken from blank spaces that would make a lot of sense. From re-reading the above I wonder if we will luck out and the updated version of pixi-tilemap will fix the scaling. (Or if not maybe it's a tilemap issue we should open). Sorry about the confusion.

It doesn't explain the distortion on the test map, but I'm trying to reproduce that now so I can upload it as an example. I'll report back if I can reproduce it, and include the project.

blurymind:
Another and much better way to fix this would be to use the x,y coordinates of the tiles from tiled itself - but that is currently not implemented on tiled. There is an existing ticket for it.

PR #2305, opened 2021-02-14
Add private functions to extensions

### LINKED ISSUE #2265: Extensions Editor - Private function for Extensions
## Description

To avoid code duplication, it is convenient to break them into functions, in the extension I also want to do this, but make some functions private/service, so that the user does not have access to them, but in the extension they can be used.

## Solution suggested

If we talk about implementation, then this is the visibility area (or namespace).
And it has already been implemented, because the life cycle methods are not visible in the event list, for others it is possible to set such a “sign” by method.

arthuro555:
Closes #2265
Adds private functions, functions only available in the current extension.

### DIFF
--- a/Core/GDCore/Project/EventsFunction.h
+++ b/Core/GDCore/Project/EventsFunction.h
@@ -8,6 +8,7 @@
 #define GDCORE_EVENTSFUNCTION_H
 
 #include <vector>
+
 #include "GDCore/Events/EventsList.h"
 #include "GDCore/Project/ObjectGroupsContainer.h"
 #include "GDCore/String.h"
@@ -117,6 +118,24 @@ class GD_CORE_API EventsFunction {
   FunctionType GetFunctionType() const { return functionType; };
 
   /**
+   * \brief Returns true if the function is private.
+   */
+  bool IsPrivate() {
+    if (isPrivate)
+      return true;
+    else
+      return false;
+  }
+
+  /**
+   * \brief Toggles the privateness of the function.
+   */
+  EventsFunction& TogglePrivate() {
+    isPrivate = !isPrivate;
+    return *this;
+  }
+
+  /**
    * \brief Return the events.
    */
   const gd::EventsList& GetEvents() const { return events; };
@@ -128,7 +147,7 @@ class GD_CORE_API EventsFunction {
 
   /**
    * \brief Return the parameters of the function.
-   * 
+   *
    * \note During code/extension generation, new parameters are added
    * to the generated function, like "runtimeScene" and "eventsFunctionContext".
    * This should be transparent to the user.
@@ -143,12 +162,14 @@ class GD_CORE_API EventsFunction {
   std::vector<gd::ParameterMetadata>& GetParameters() { return parameters; };
 
   /**
-   * \brief Return a reference to the object groups that can be used in the function.
+   * \brief Return a reference to the object groups that can be used in the
+   * function.
    */
   ObjectGroupsContainer& GetObjectGroups() { return objectGroups; }
 
   /**
-   * \brief Return a const reference to the object groups that can be used in the function.
+   * \brief Return a const reference to the object groups that can be used in
+   * the function.
    */
   const ObjectGroupsContainer& GetObjectGroups() const { return objectGroups; }
 
@@ -176,6 +197,7 @@ class GD_CORE_API EventsFunction {
   FunctionType functionType;
   std::vector<gd::ParameterMetadata> parameters;
   gd::ObjectGroupsContainer objectGroups;
+  bool isPrivate = false;
 };
 
 }  // namespace gd
--- a/Core/GDCore/Project/EventsFunction.h
+++ b/Core/GDCore/Project/EventsFunction.h
@@ -120,18 +120,13 @@ class GD_CORE_API EventsFunction {
   /**
    * \brief Returns true if the function is private.
    */
-  bool IsPrivate() {
-    if (isPrivate)
-      return true;
-    else
-      return false;
-  }
+  bool IsPrivate() { return isPrivate; }
 
   /**
-   * \brief Toggles the privateness of the function.
+   * \brief Sets the privateness of the function.
    */
-  EventsFunction& TogglePrivate() {
-    isPrivate = !isPrivate;
+  EventsFunction& SetPrivate(bool _isPrivate) {
+    isPrivate = _isPrivate;
     return *this;
   }
 
--- a/newIDE/app/src/EventsFunctionsList/index.js
+++ b/newIDE/app/src/EventsFunctionsList/index.js
@@ -86,7 +86,7 @@ export default class EventsFunctionsList extends React.Component<Props, State> {
     searchText: '',
   };
 
-  _togglePrivate = eventsFunction => {
+  _togglePrivate = (eventsFunction: gdEventsFunction) => {
     eventsFunction.setPrivate(!eventsFunction.isPrivate());
     this.forceUpdate();
   };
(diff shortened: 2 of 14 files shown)


Core/GDCore/Project/EventsFunction.h
```diff
@@ -116,6 +117,24 @@ class GD_CORE_API EventsFunction {
    */
   FunctionType GetFunctionType() const { return functionType; };
 
+  /**
+   * \brief Returns true if the function is private.
+   */
+  bool IsPrivate() {
+    if (isPrivate)
```
arthuro555, line 124: I probably did a c++ newbie mistake again, but this is the only way I got this to work. `return isPrivate;` always returned true no matter what for some reason.

newIDE/app/src/EventsFunctionsList/index.js
```diff
@@ -86,6 +89,11 @@ export default class EventsFunctionsList extends React.Component<Props, State> {
     searchText: '',
   };
 
+  _togglePrivate = eventsFunction => {
+    eventsFunction.togglePrivate();
+    this.forceUpdate();
```
arthuro555, line 94: Not sure if this is necessary or the way to go

newIDE/app/src/InstructionOrExpression/EnumeratedInstructionOrExpressionMetadata.js
```diff
@@ -37,27 +43,20 @@ export type EnumeratedInstructionOrExpressionMetadata =
   | EnumeratedInstructionMetadata
   | EnumeratedExpressionMetadata;
 
-export const filterEnumeratedInstructionOrExpressionMetadataByScope = <
-  +T: EnumeratedInstructionOrExpressionMetadata
```
arthuro555, line 41: I'm pretty sure this broke flow but CRA wouldn't compile this after I touched the file.

Core/GDCore/Project/EventsFunction.h
```diff
@@ -116,6 +117,24 @@ class GD_CORE_API EventsFunction {
...
+      return true;
+    else
+      return false;
+  }
+
+  /**
+   * \brief Toggles the privateness of the function.
+   */
+  EventsFunction& TogglePrivate() {
```
4ian, line 133: Nitpicking: In general avoid designing classes with `ToggleSomeState`: prefer the more explicit `SetSomeState`.
The reason is because most of the time your client/user will want to set something to a specific value. For example, if I click on a button to make my function private, my goal is not to "toggle the privateness", but to set it to true. If I want to make something public, I don't care about toggling anything, I want to set it to false (and if it was false before, well I still want it false, not true :)).

It's also true in the case of UI framework where some update could be batched: the update should contain the definitive value you want to apply, not a command to toggle - which could be the contrary of what you wish if for some reason the value was inverted in the meantime.

newIDE/app/src/EventsFunctionsList/index.js
```diff
@@ -39,6 +39,9 @@ export type EventsFunctionCreationParameters = {|
 const getEventsFunctionName = (eventsFunction: gdEventsFunction) =>
   eventsFunction.getName();
 
+const getEventsFunctionPrivateness = (eventsFunction: gdEventsFunction) =>
```
4ian, line 42: isEventsFunctionPrivate

newIDE/app/src/EventsFunctionsList/index.js
4ian: You're changing the value of something that is stored in libGD.js, in a memory that is not understood by React as a "state". So you need to call this.forceUpdate to tell React "hey I know I've not modified a state or a prop, but something still changed and you need to render again this component".

newIDE/app/src/InstructionOrExpression/EnumeratedInstructionOrExpressionMetadata.js
4ian: What was the error? Can you try to put back the original typing and paste the error here?

newIDE/app/src/InstructionOrExpression/EnumeratedInstructionOrExpressionMetadata.js
```diff
@@ -66,10 +65,16 @@ export const filterEnumeratedInstructionOrExpressionMetadataByScope = <
...
       ) === behaviorMetadata.getName()
-    ) {
+    )
+      return true;
+
+    // Show private non-behavior functions when editing the extension
+    if (
+      !behaviorMetadata &&
+      eventsFunctionsExtension &&
+      eventsFunctionsExtension.getName() === type.split(separator)[0]
```
4ian, line 75: Why is this line needed?

Core/GDCore/Project/EventsFunction.h
4ian: Weird, it's very unlikely for the compiler to make a miscompilation of something that simple. Can you try going back to `return isPrivate;`?

4ian (commented):
Thanks for the PR, looks pretty nice :)

newIDE/app/src/EventsFunctionsList/index.js
```diff
@@ -323,6 +337,7 @@ export default class EventsFunctionsList extends React.Component<Props, State> {
                       i18n
                     )}
                     reactDndType="GD_EVENTS_FUNCTION"
+                    isItemBold={getEventsFunctionPrivateness}
```
4ian, line 340: Do you think we could just add ` (private)` at the end of the name shown in the list?

newIDE/app/src/EventsFunctionsList/index.js
arthuro555: Ok, I was unsure because some component seem to ask the component above via a prop to forceUpdate or something like this.

newIDE/app/src/InstructionOrExpression/EnumeratedInstructionOrExpressionMetadata.js
arthuro555: To check if the eventsFunctionExtension is the one the instruction or expression is from?

newIDE/app/src/EventsFunctionsList/index.js
arthuro555: Personally, I prefer the boldness, but if you think this is better then yes no problem

newIDE/app/src/InstructionOrExpression/EnumeratedInstructionOrExpressionMetadata.js
arthuro555: Huh the error seems to be gone now. Weird.

Core/GDCore/Project/EventsFunction.h
arthuro555: My computer must have had some sort of problem, it seems I had a lot of very weird issues that just disappeared. This works now.

4ian:
There is a bug, behavior method marked private are still shown. I'm fixing this.

4ian:
Seems like the InstructionOrObjectSelector is not "scoped", so is currently showing all instructions/free instructions without filtering. I'm also fixing this.

4ian:
I've done two fixes and a change:
- InstructionOrObjectSelector was displaying (in the list of "free" instructions or when searching) all instructions without filtering the private ones. This is because `filterEnumeratedInstructionOrExpressionMetadataByScope` was never called (because it was useless before). This is fixed, and the "scope" is passed (so the filtering works according to if you're in a layout or an extension or a behavior).
- setPrivate was never called for behavior methods.
- I've also changed the test to check if the extension is the same as the events function extension of the current scope. For this, I added the `extension` in the scope of an instruction and extension. I can then use this to compare its name with the edited events function extension (if any), rather than infering it from the type. Slightly better, and more consistent with the rest.

Finally, a test was failing that I fixed because you can't directly use `gd` outside of a function - it may not be initialized yet.

Should be good now, waiting for the tests to pass :)

Issue #2252, opened 2021-01-30
Install a game overwrite the previous game already installed


Bouh:
## Describe the bug
Install a game made with GDevelop5 overwrite the previous game already installed.
[It seem appear only with the build service.](https://forum.gdevelop-app.com/t/game-override-another-game/30229/5?u=bouh)

## To Reproduce
1. Export a game for windows with the auto-installer.
2. Install the first game
3. Repeat the process with another game, with a different, game name, package name, version number, and author name.
4. See the second game is installed in the folder of the first game.

[Bug fixes for manual export.](https://forum.gdevelop-app.com/t/allow-install-directory-selection-for-desktop-builds/30230)

## Other details
* W10
* B103
Source [link 1](https://forum.gdevelop-app.com/t/game-override-another-game/30229) [link 2](https://forum.gdevelop-app.com/t/windows-auto-installer-game-builds-are-overwriting-each-other-on-install/27398/8)

Silver-Streak:
This has been happening on and off for some users since at least September last year. I have never been able to reproduce it for folks reporting it on the forums, likely because I build locally. This makes me think it is definitely specific to the build service.

The linked post for bug fixes is more of a feature request I had to help alleviate it (and would likely still be a good option for the build service), but it doesn't necessarily tell us the *source* of the issue on the build service.

arthuro555:
My guess is that the nsis ID is somehow cached by the build service, so when installing a GDevelop game, the installer sees the other installed GDevelop game as the same app and treats the installation as an update. As the installation path could have been defined by the user in the last installer, it just uses that same path to make sure the user preferences are respected.

4ian:
Each build is made on a blank virtual machine launched specifically for the build so it's not entirely a caching issue - but the issue is surely related to NSIS using the same identifier for two apps.

arthuro555:
I'd propose to make the package.json have the field [`nsis.guid`](https://www.electron.build/configuration/nsis#NsisOptions-guid) set to the Project UID, that way it should remain the same across updates and stay unique to the game.

4ian:
Good finding, though the docs indicates that `You can explicitly set guid using option nsis.guid, but it is not recommended — consider using appId.`

So we could go ahead with the `appId` field (which is by the way what we're using for GDevelop itself), and set it to the package name (used for Android/iOS already). It's pretty obvious that the package name should be unique, because it's what is used for app stores :) 

Also note that the documentation answer to the question:

[...]

PR #2306, opened 2021-02-14
Add warning if exporting game with default package name

arthuro555:
Adds a dismissable warning when exporting for cordova or electron and the package name is com.example (the default).
The warning for electron will only be valid once the suggested solution for #2252 is implemented.

### DIFF
--- a/newIDE/app/src/Export/ExportLauncher.js
+++ b/newIDE/app/src/Export/ExportLauncher.js
@@ -1,6 +1,7 @@
 // @flow
 
 import React, { Component } from 'react';
+import { I18n } from '@lingui/react';
 import RaisedButton from '../UI/RaisedButton';
 import { sendExportLaunched } from '../Utils/Analytics/EventSender';
 import {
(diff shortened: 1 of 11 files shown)


newIDE/app/src/Export/ExportLauncher.js
```diff
@@ -245,6 +247,21 @@ export default class ExportLauncher extends Component<Props, State> {
...
+            <Line>
+              <DismissableAlertMessage
+                identifier="project-should-have-unique-package-name"
+                kind="warning"
+              >
+                <I18n>
+                  {({ i18n }) =>
+                    i18n._(exportPipeline.shouldHaveUniquePackageName)
+                  }
+                </I18n>
```
4ian, line 261: I would recommend having here the two different messages, according to a flag `exportPipeline.packageNameWarningType` which can be `none` | `desktop` | `mobile`

4ian:
Looks good! Thanks :)

PR #2312, opened 2021-02-17
Update the isometric-game example to use new features

D8H:
* This sets an isometry angle for the top down movement (the art is not an accurate isometry so it's not perfect, it's working on one axis but it's drifting away on the other one).
* The character also follows the stairs with an extension that is not yet reviewed:
https://github.com/4ian/GDevelop-extensions/issues/102


4ian:
@D8H Looking at the isometry of the top down movement again, do you think there would be a way to make that the up key would move the object up (and left key left, right key right, down key down) even when the isometry viewpoint is used?

This would make the implementation of movement using a joystick, and using the mouse/touch easier and much more intuitive I think :) (I agree that the current solution is the "mathematically right" solution, but I think it's also counterintuitive for most players and I've not seen any game applying this method to their directional pad and movement).

D8H:
@4ian , what do think about using 45° as the default offset for isometry?
I answered on the other thread but I forgot to name you so maybe you didn't see it:
https://github.com/4ian/GDevelop-extensions/issues/102#issuecomment-780546420

4ian:
@D8H I forgot about the other thread, I answered on it (basically yes I think it's a good idea!)

PR #2315, opened 2021-02-19
Migrate path finding from HSHG to RBush

D8H:
I ran format on the test but some sections became really are to read. Position calculus need to stay on one line so x and y are easy to identify. The formatter uses a really small line length. A fullHD monitor can display more than 3 files side by side as it is, I think having readable code would be more useful.
What is your view on this?
Do you think it could be possible to have something like +25% line length?

### DIFF
--- a/Extensions/PathfindingBehavior/tests/commonpathfindingruntimebehavior.spec.js
+++ b/Extensions/PathfindingBehavior/tests/commonpathfindingruntimebehavior.spec.js
@@ -80,7 +80,7 @@ const doTestsCommonpathfindingruntimebehavior = (
   it('can find a path without any obstacle in the way', function () {
     const obstacle = addObstacle(runtimeScene);
     obstacle.setPosition(100, 100);
-    // so obstacles register
+    // To ensure obstacles are registered.
     runtimeScene.renderAndStep(1000 / 60);
 
     object.setPosition(480, 300);
@@ -93,7 +93,7 @@ const doTestsCommonpathfindingruntimebehavior = (
 
   it("mustn't find a path to the obstacle inside", function () {
     obstacle.setPosition(600, 300);
-    // so obstacles register
+    // To ensure obstacles are registered.
     runtimeScene.renderAndStep(1000 / 60);
 
     object.setPosition(480, 300);
@@ -103,7 +103,7 @@ const doTestsCommonpathfindingruntimebehavior = (
 
   it('can find a path with an obstacle in the way', function () {
     obstacle.setPosition(600, 300);
-    // so obstacles register
+    // To ensure obstacles are registered.
     runtimeScene.renderAndStep(1000 / 60);
 
     object.setPosition(480, 300);
@@ -118,7 +118,7 @@ const doTestsCommonpathfindingruntimebehavior = (
   it('can find a path between 2 obstacles', function () {
     obstacleTop.setPosition(600, 180);
     obstacleBottom.setPosition(600, 420);
-    // so obstacles register
+    // To ensure obstacles are registered.
     runtimeScene.renderAndStep(1000 / 60);
 
     object.setPosition(480, 300);
@@ -135,7 +135,7 @@ const doTestsCommonpathfindingruntimebehavior = (
     obstacleBottom.setPosition(600, 420);
     obstacleLeft.setPosition(480, 300);
     obstacleRight.setPosition(720, 300);
-    // so obstacles register
+    // To ensure obstacles are registered.
     runtimeScene.renderAndStep(1000 / 60);
 
     object.setPosition(360, 300);
--- a/Extensions/PathfindingBehavior/tests/commonpathfindingruntimebehavior.spec.js
+++ b/Extensions/PathfindingBehavior/tests/commonpathfindingruntimebehavior.spec.js
@@ -1,0 +1,161 @@
+const addObstacle = (runtimeScene) => {
+  var obstacle = new gdjs.RuntimeObject(runtimeScene, {
+    name: 'obstacle',
+    type: '',
+    behaviors: [
+      {
+        type: 'PathfindingBehavior::PathfindingObstacleBehavior',
+        impassable: true,
+        cost: 2,
+      },
+    ],
+  });
+  obstacle.getWidth = function () {
+    return 100;
+  };
+  obstacle.getHeight = function () {
+    return 100;
+  };
+  runtimeScene.addObject(obstacle);
+
+  return obstacle;
+};
+
+const doTestsCommonpathfindingruntimebehavior = (
+  collisionMethod,
+  allowDiagonals
+) => {
+  var runtimeGame = new gdjs.RuntimeGame({
+    variables: [],
+    properties: { windowWidth: 800, windowHeight: 600 },
+    resources: { resources: [] },
+  });
+  var runtimeScene = new gdjs.RuntimeScene(runtimeGame);
+  runtimeScene.loadFromScene({
+    layers: [{ name: '', visibility: true, effects: [] }],
+    variables: [],
+    behaviorsSharedData: [],
+    objects: [],
+    instances: [],
+  });
+  runtimeScene._timeManager.getElapsedTime = function () {
+    return (1 / 60) * 1000;
+  };
+
+  const pathFindingName = 'auto1';
+  var object = new gdjs.RuntimeObject(runtimeScene, {
+    name: 'player',
+    type: '',
+    behaviors: [
+      {
+        type: 'PathfindingBehavior::PathfindingBehavior',
+        name: pathFindingName,
+        allowDiagonals: allowDiagonals,
+        acceleration: 400,
+        maxSpeed: 200,
+        angularMaxSpeed: 180,
+        rotateObject: false,
+        angleOffset: 0,
+        cellWidth: 20,
+        cellHeight: 20,
+        extraBorder: 0
+      },
+    ],
+  });
+  object.getWidth = function () {
+    return 90;
+  };
+  object.getHeight = function () {
+    return 90;
+  };
+  runtimeScene.addObject(object);
+
+  it('can find a path without any obstacle at all', function () {
+    object.setPosition(480, 300);
+    object.getBehavior(pathFindingName).moveTo(runtimeScene, 720, 300);
+    expect(object.getBehavior(pathFindingName).pathFound()).to.be(true);
+    expect(object.getBehavior(pathFindingName).getNodeCount()).to.be(13);
+  });
+
+  it('can find a path without any obstacle in the way', function () {
+    const obstacle = addObstacle(runtimeScene);
+    obstacle.setPosition(100, 100);
+    // so obstacles register
+    runtimeScene.renderAndStep(1000 / 60);
+
+    object.setPosition(480, 300);
+    object.getBehavior(pathFindingName).moveTo(runtimeScene, 720, 300);
+    expect(object.getBehavior(pathFindingName).pathFound()).to.be(true);
+    expect(object.getBehavior(pathFindingName).getNodeCount()).to.be(13);
+  });
+
+  const obstacle = addObstacle(runtimeScene);
+
+  it("mustn't find a path to the obstacle inside", function () {
+    obstacle.setPosition(600, 300);
+    // so obstacles register
+    runtimeScene.renderAndStep(1000 / 60);
+
+    object.setPosition(480, 300);
+    object.getBehavior(pathFindingName).moveTo(runtimeScene, 650, 350);
+    expect(object.getBehavior(pathFindingName).pathFound()).to.be(false);
+  });
+
+  it('can find a path with an obstacle in the way', function () {
+    obstacle.setPosition(600, 300);
+    // so obstacles register
+    runtimeScene.renderAndStep(1000 / 60);
+
+    object.setPosition(480, 300);
+    object.getBehavior(pathFindingName).moveTo(runtimeScene, 720, 300);
+    expect(object.getBehavior(pathFindingName).pathFound()).to.be(true);
+    expect(object.getBehavior(pathFindingName).getNodeCount()).to.be.above(13);
+  });
+
+  const obstacleTop = obstacle;
+  const obstacleBottom = addObstacle(runtimeScene);
+
+  it('can find a path between 2 obstacles', function () {
+    obstacleTop.setPosition(600, 180);
+    obstacleBottom.setPosition(600, 420);
+    // so obstacles register
+    runtimeScene.renderAndStep(1000 / 60);
+
+    object.setPosition(480, 300);
+    object.getBehavior(pathFindingName).moveTo(runtimeScene, 720, 300);
+    expect(object.getBehavior(pathFindingName).pathFound()).to.be(true);
+    expect(object.getBehavior(pathFindingName).getNodeCount()).to.be(13);
+  });
+
+  const obstacleLeft = addObstacle(runtimeScene);
+  const obstacleRight = addObstacle(runtimeScene);
+
+  it("mustn't find a path to a closed room", function () {
+    obstacleTop.setPosition(600, 180);
+    obstacleBottom.setPosition(600, 420);
+    obstacleLeft.setPosition(480, 300);
+    obstacleRight.setPosition(720, 300);
+    // so obstacles register
+    runtimeScene.renderAndStep(1000 / 60);
+
+    object.setPosition(360, 300);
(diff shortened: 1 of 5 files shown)


4ian:
Thank you! This change to rbush will be very useful to keep the codebase consistent and ease future development of a unified Rbush for objects.

Tests will be incredibly useful too for the next PRs :) 

> What is your view on this?

It's true that a line length of 80 is fairly small, but it's also ideal to display in GitHub and have two files side to side (nice for diffs) and encourage splitting long lines. Also might be easier on the eye not to have to follow a 120 character long line :) 
This being said, it might sometimes be a pain when you need something formatted in a specific way. In this case you can add a `// @prettier-ignore` on the line before the "node" (i.e: the line) not to format. See https://prettier.io/docs/en/ignore.html

Let me know if it works with this workaround for you :) I really avoid tweaking the autoformatting at all because that would imply a change to all the files of the codebase, and because 80 columns is fairly standard, easy to diff and encourage keeping functions short.

Extensions/PathfindingBehavior/tests/legacypathfindingruntimebehavior.spec.js
```diff
@@ -0,0 +1,689 @@
...
+    player.getBehavior(pathFindingName).moveTo(runtimeScene, targetX, targetY);
+    if (cellSize == 160) {
+      // The obstacle is the size of the cell so the player can bypass it.
+      expect(player.getBehavior(pathFindingName).getNodeCount()).to.be.above(3);
+    } else {
+      expect(player.getBehavior(pathFindingName).pathFound()).to.be(false);
+    }
+  });
+
+  it('can find a path with an obstacle adjascent to the target', function () {
```
4ian, line 363: typo: adjacent

Extensions/PathfindingBehavior/tests/legacypathfindingruntimebehavior.spec.js
```diff
@@ -0,0 +1,689 @@
...
+    expect(player.getBehavior(pathFindingName).pathFound()).to.be(false);
+  });
+
+  const obstacleTop = obstacle;
+  const obstacleBottom = addObstacleSprite3(
+    runtimeScene,
+    objectCenteredOnCells
+  );
+
+  it('can find a path between 2 ajusted obstacles', function () {
```
4ian, line 479: what do you mean by ajusted? "close obstacles" maybe?

Extensions/PathfindingBehavior/tests/legacypathfindingruntimebehavior.spec.js
```diff
@@ -0,0 +1,689 @@
...
+        bottomObstacleY =
+          playerY + playerBottomBorder + obstacleTopBorder + inclusiveMargin;
+        targetX =
+          obstacleX - obstacleLeftBorder - playerRightBorder - 2 * cellSize;
+        targetY = playerY;
+        straitLineCells = (playerX - targetX) / cellSize + 1;
+        break;
+      }
+      case 'up': {
+        // topObstacle and bottomObstacle are actually left and right but for convegnance...
```
4ian, line 519: typo: convenience

Extensions/PathfindingBehavior/tests/legacypathfindingruntimebehavior.spec.js
```diff
@@ -0,0 +1,689 @@
...
+        bottomObstacleX =
+          playerX + playerRightBorder + obstacleLeftBorder + inclusiveMargin;
+        targetY =
+          obstacleY - obstacleTopBorder - playerBottomBorder - 2 * cellSize;
+        targetX = playerX;
+        straitLineCells = (playerY - targetY) / cellSize + 1;
+        break;
+      }
+      case 'down': {
+        // topObstacle and bottomObstacle are actually left and right but for convegnance...
```
4ian, line 535: same.

Extensions/PathfindingBehavior/tests/legacypathfindingruntimebehavior.spec.js
```diff
@@ -0,0 +1,689 @@
...
+
+    player.setPosition(playerX, playerY);
+    player.getBehavior(pathFindingName).moveTo(runtimeScene, targetX, targetY);
+    expect(player.getBehavior(pathFindingName).pathFound()).to.be(true);
+    expect(player.getBehavior(pathFindingName).getNodeCount()).to.be(
+      straitLineCells
+    );
+  });
+
+  it("mustn't find a direct path between 2 slightly too much tighten obstacles", function () {
```
4ian, line 565: "mustn't find a direct path between 2 slightly too much tighten obstacles"
=>
"should not find a direct path between 2 obstacles making a path slightly too narrow."

Extensions/PathfindingBehavior/tests/legacypathfindingruntimebehavior.spec.js
```diff
@@ -0,0 +1,689 @@
...
+  );
+
+  it('can find a path between 2 ajusted obstacles', function () {
+    let topObstacleX;
+    let bottomObstacleX;
+    let topObstacleY;
+    let bottomObstacleY;
+    let targetX;
+    let targetY;
+    let straitLineCells;
```
4ian, line 486: typo: straightLineCells

Also, make it obvious this is a variable containing a count, not containing "cells" or a complicated thing:

`straightLineCellCount`

Extensions/PathfindingBehavior/tests/legacypathfindingruntimebehavior.spec.js
```diff
@@ -0,0 +1,689 @@
...
+  });
+
+  it("mustn't find a direct path between 2 slightly too much tighten obstacles", function () {
+    let topObstacleX;
+    let bottomObstacleX;
+    let topObstacleY;
+    let bottomObstacleY;
+    let targetX;
+    let targetY;
+    let straitLineCells;
```
4ian, line 572: same as before

Extensions/PathfindingBehavior/tests/legacypathfindingruntimebehavior.spec.js
```diff
@@ -0,0 +1,689 @@
...
+  } else {
+    // The collision is not strict
+    inclusiveMargin = 1;
+  }
+
+  // multiple of 20 and 160
+  const playerX = 480;
+  const playerY = 320;
+
+  // This one is not on limit
```
4ian, line 180: What do you mean exactly by "not on limit"?

Extensions/PathfindingBehavior/tests/legacypathfindingruntimebehavior.spec.js
```diff
@@ -0,0 +1,689 @@
...
+    }
+
+    obstacle.setPosition(obstacleX, obstacleY);
+    // so obstacles register
+    runtimeScene.renderAndStep(1000 / 60);
+
+    // move away from the obstacle
+    player.setPosition(playerX, playerY);
+    player.getBehavior(pathFindingName).moveTo(runtimeScene, targetX, targetY);
+    //expect(player._animationFrame).to.be(null);
```
4ian, line 221: Let's remove this commented line.

Extensions/PathfindingBehavior/tests/legacypathfindingruntimebehavior.spec.js
```diff
@@ -0,0 +1,689 @@
...
```
4ian, line 286: (same everywhere)

Extensions/PathfindingBehavior/tests/legacypathfindingruntimebehavior.spec.js
```diff
@@ -0,0 +1,689 @@
...
+    let targetX;
+    let targetY;
+    switch (direction) {
+      case 'right':
+        // right on the player right
+        obstacleX =
+          playerX +
+          playerRightBorder +
+          obstacleLeftBorder -
+          cellSize +
```
4ian, line 238: Why removing a cellSize in all of these coordinates?

Extensions/PathfindingBehavior/tests/legacypathfindingruntimebehavior.spec.js
D8H: This test is not part of a test couple : there it works, there it doesn't. But it seems like an interesting case nonetheless. Though, I'm not sure.

Extensions/PathfindingBehavior/tests/legacypathfindingruntimebehavior.spec.js
D8H: Yes, I'm really sorry for all the typo. I'll try another IDE because not much is working on Eclipse for JS.

Extensions/PathfindingBehavior/tests/legacypathfindingruntimebehavior.spec.js
4ian: It's surely a good idea to keep it yes! Let's just remove the comment, as it's a test like the others.

D8H:
> It's true that a line length of 80 is fairly small, but it's also ideal to display in GitHub and have two files side to side (nice for diffs) and encourage splitting long lines. Also might be easier on the eye not to have to follow a 120 character long line :)
> This being said, it might sometimes be a pain when you need something formatted in a specific way. In this case you can add a `// @prettier-ignore` on the line before the "node" (i.e: the line) not to format. See https://prettier.io/docs/en/ignore.html
> 
> Let me know if it works with this workaround for you :) I really avoid tweaking the autoformatting at all because that would imply a change to all the files of the codebase, and because 80 columns is fairly standard, easy to diff and encourage keeping functions short.

That seems fair enough. I think some IDE can automatically reformat at pull and stage.

Extensions/PathfindingBehavior/tests/legacypathfindingruntimebehavior.spec.js
4ian: Definitely give a try at VS Code (great integration with typescript/javascript). You can also try WebStorm if you want something a bit more "heavy", but I like the simplicity of VS Code + the great editing capabilities (multi selection) and the support for TS/JS is very good.

Extensions/PathfindingBehavior/tests/legacypathfindingruntimebehavior.spec.js
D8H: This needs a comment and a better test title. It's testing that if an obstacle is place so it will be right next to the player after the first step the player is not stuck.

Extensions/PathfindingBehavior/tests/legacypathfindingruntimebehavior.spec.js
D8H: I was going for "The technician adjusts the part for assembly".
https://www.wordreference.com/fren/ajust%C3%A9

GDJS/package.json
```diff
@@ -29,7 +29,7 @@
     "check-types": "tsc",
     "build": "node scripts/build.js",
     "test": "cd tests && npm test",
-    "format": "prettier --write \"Runtime/**/*.ts\" \"../Extensions/**/*.ts\"",
+    "format": "prettier --write \"Runtime/**/*.ts\" \"../Extensions/**/*.ts\" \"../Extensions/**/*.spec.js\"",
```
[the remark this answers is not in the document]
4ian: I think this is fine! I'll run a commit before merging this to autoformat these files :)

Extensions/PathfindingBehavior/tests/legacypathfindingruntimebehavior.spec.js
```diff
@@ -0,0 +1,688 @@
...
+
+  it("mustn't find a path when an obstacle is slightly overlaping the only first cell to go", function () {
+    let obstacleX;
+    let obstacleY;
+    let targetX;
+    let targetY;
+    switch (direction) {
+      case 'right':
+        // The obstacle will be right on the player right after the first step.
+        // So, it's overlaping the player by one cell lengt at start.
```
4ian, line 297: lengt: length
overlaping: overlapping

4ian: (here and in the rest of the code)

Extensions/PathfindingBehavior/tests/legacypathfindingruntimebehavior.spec.js
```diff
@@ -0,0 +1,688 @@
...
+          cellSize +
+          inclusiveMargin -
+          1;
+        targetX = playerX;
+        targetY = playerY - 2 * cellSize;
+        break;
+    }
+
+    obstacle.setPosition(obstacleX, obstacleY);
+    // so obstacles register
```
4ian, line 348:
```suggestion
    // To ensure obstacles are registered.
```

Extensions/PathfindingBehavior/tests/legacypathfindingruntimebehavior.spec.js
```diff
@@ -0,0 +1,688 @@
...
+    runtimeScene.renderAndStep(1000 / 60);
+
+    // move as close as possible to the obstacle
+    player.setPosition(playerX, playerY);
+    player.getBehavior(pathFindingName).moveTo(runtimeScene, targetX, targetY);
+    expect(player.getBehavior(pathFindingName).pathFound()).to.be(true);
+    expect(player.getBehavior(pathFindingName).getNodeCount()).to.be(3);
+  });
+
+  it("mustn't find a path with an obstacle slightly overlaping to the target", function () {
```
4ian, line 410: `an obstacle slightly overlapping with the target`

Extensions/PathfindingBehavior/tests/legacypathfindingruntimebehavior.spec.js
```diff
@@ -0,0 +1,688 @@
...
+    expect(player.getBehavior(pathFindingName).pathFound()).to.be(false);
+  });
+
+  const obstacleTop = obstacle;
+  const obstacleBottom = addObstacleSprite3(
+    runtimeScene,
+    objectCenteredOnCells
+  );
+
+  it('can find a path between 2 obstacles making a path perfectly ajusted to the object', function () {
```
4ian, line 478: adjusted

Extensions/PathfindingBehavior/tests/commonpathfindingruntimebehavior.spec.js
```diff
@@ -0,0 +1,161 @@
+const addObstacle = (runtimeScene) => {
```
4ian, line 1: You may want to move this inside `doTestsCommonpathfindingruntimebehavior` to avoid global variables.

Extensions/PathfindingBehavior/tests/commonpathfindingruntimebehavior.spec.js
```diff
@@ -0,0 +1,161 @@
...
+    runtimeScene.renderAndStep(1000 / 60);
+
+    object.setPosition(480, 300);
+    object.getBehavior(pathFindingName).moveTo(runtimeScene, 720, 300);
+    expect(object.getBehavior(pathFindingName).pathFound()).to.be(true);
+    expect(object.getBehavior(pathFindingName).getNodeCount()).to.be.above(13);
+  });
+
+  const obstacleTop = obstacle;
+  const obstacleBottom = addObstacle(runtimeScene);
```
4ian, line 116: In general try to avoid declaring anything between tests. Tests are supposed to be independent, so the last one could be launched before the first one, vice versa, etc...

Prefer to set up the scene and obstacles for each test (you can make a helper function inside the function if needed, to avoid too much code duplication :)).

In other words, a test should be self contained - this will allow them to run in parallel and to avoid weird stuff like "my test is failing because the previous test failed, now I'm confused" ;)

4ian:
Looks good overall! I've added a few more comments about typos and making the tests self contained (avoid declaring obstacles between tests, the order of tests is not guaranteed and could change in the future).

We can merge this once this is done and if you can confirm there is no visible change in the pathfinding behavior after this switch to RBush (in theory, it should work exactly like before, but better confirm it by testing an example game).

Thanks again for your work on this! Looking forward to get these tests in the codebase :)

Extensions/PathfindingBehavior/tests/legacypathfindingruntimebehavior.spec.js
D8H: I installed a spelling checker extension on VS. It will be very helpful.

D8H:
I put a name space around the utility functions and the test so they doesn't collide from one test file to another.

4ian:
Looks great, thanks for taking the time to do this!

PR #2316, opened 2021-02-19
Update to React 17.0.1 and CRA 4.0.2

arthuro555:
This update should allow having [useful component stacks on errors in production](https://reactjs.org/blog/2020/08/10/react-v17-rc.html#native-component-stacks) and fast refresh (hot reload but faster).

I tested pretty much all components in storybook and they seemed to behave fine. I also toyed around the IDE for a while and could not find any issue.
What still needs to be checked:
- There has been changes to the service workers. We need to look at the built service worker and potentially change it to work like intended.
- The new Fast Refresh feature doesn't seem to work for me. I have a local .env file that is ignored thanks to the gitignore that I use to disable it, if it doesn't work for everyone we should commit that .env file disabling fast refresh.


4ian:
Ah incredible, I was thinking that we should upgrade to CRA 4/React 17 at some point!

> The new Fast Refresh feature doesn't seem to work for me. I have a local .env file that is ignored thanks to the gitignore that I use to disable it, if it doesn't work for everyone we should commit that .env file disabling fast refresh.

Weird, what's going on?

> There has been changes to the service workers. We need to look at the built service worker and potentially change it to work like intended.

Ok, would be good to take a look. You can potentially try it by going into the service worker files and:
- invert the `if`s that are checking if we're in development.
- that's it? You can modify your /etc/hosts (or similar on Windows) to simulate a domain name like `gdevelop-app-local.com` that redirects to localhost:3000

arthuro555:
> Weird, what's going on?

It is just not reloading the changes 🤷‍♂️ 
I don't think it is just me, as many people seem to experience issues: https://github.com/facebook/create-react-app/issues/9904
I'd say we stay on the old system that just refreshes the page for now until we know that CRA's fast refresh is working for everyone.

4ian:
Might because of an old/incompatible version of Babel (or another dependency, webpack?) brought by something like Storybook or lingui?

> I'd say we stay on the old system that just refreshes the page for now until we know that CRA's fast refresh is working for everyone.

Yes we can do that. Would be good to verify the service worker to see what changed then.

4ian:
I've tested on macOS, same issue.

arthuro555:
I'm closing this until there is a fix for hot-reload/fast refresh

blurymind:
I hear some raves about vive for hot reloading, but I imagine it would be a
huge task to make gd use it

4ian:
In theory if we upgrade to React 17 and CRA 4, it should work "out of the box" (we're not doing anything special that should prevent it). I look forward to this - combined with Storybook and this hot reloading, this can speed up development of UI components very much :)

4ian:
Fast refresh finally working in this PR: https://github.com/4ian/GDevelop/pull/3934 :)

PR #2318, opened 2021-02-20
Fix errors in expressions and update the gamepad extension to 0.0.7

Bouh:
Update gamepad extension to 0.0.7.
Fix errors in expressions from the gamepad example
[Wait the validation for it before.](https://github.com/4ian/GDevelop-extensions/pull/104)


4ian (approved):
Feel free to merge when ready 👍

Issue #2039, opened 2020-10-16
UI Add an object in a object group


Bouh:
## Describe the bug
When an object is added to an object-group, the object-name is inserted into the input field stay.
We have to delete it before, write another object for continue working.

![image](https://user-images.githubusercontent.com/1670670/96247773-dd80e100-0faa-11eb-8d2e-1f83eb0dd8b0.png)

## Other details
* Since the dawn of time.

saideepesh000:
@Bouh, can I work on this

Bouh:
> @Bouh, can I work on this

@saideepesh000  Yes sure, no need to ask :)

saideepesh000:
@Bouh, Actually I am trying to solve this issue, as per my understanding should I need to make "inputValue" to null, after the event occurs, in the file 'SemiControlledAutoComplete.js'. Am I right.

[SemiControlledAutoComplete.js](https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/UI/SemiControlledAutoComplete.js).

4ian:
SemiControlledAutoComplete is "Semi controlled", which means that it show the value passed as a prop, but while it's focused it's keeping the value entered by the user (allowing to avoid doing a full re-renderer of the whole dialog every time a character is entered, especially useful in larger dialogs and for some text field like when entering a floating point number - where the dot would be discarded if it was controlled (writing `1.` would be reset to `1` for example)).

In this case, I think you would need to use `forceInputValueTo` to reset the autocomplete to an empty string after the objet is added to the group:
https://github.com/4ian/GDevelop/blob/8ba11703e193001b1fdcb683128e38ecbc7f9228/newIDE/app/src/UI/SemiControlledAutoComplete.js#L199
https://github.com/4ian/GDevelop/blob/8ba11703e193001b1fdcb683128e38ecbc7f9228/newIDE/app/src/ResourcesList/ResourceSelector.js#L136

saideepesh000:
@4ian, hey thanks for your reply, you mean whenever there is an object added to the group in here -> https://github.com/4ian/GDevelop/blob/8ba11703e193001b1fdcb683128e38ecbc7f9228/newIDE/app/src/ObjectGroupEditor/index.js#L37-L45
We need to call the forceinputvalue method to reset the value. am I right?

4ian:
Yes that what I would try! The setState is not sufficient.

saideepesh000:
[...]

PR #2376, opened 2021-02-21
Fix object name not being correctly cleared in ObjectGroup editor

arthuro555:
When selecting an object in the ObjectGroup editor, the text field should be cleared, as you can witness here:
https://github.com/4ian/GDevelop/blob/f83bea54c9601f3ded89e08a9c07beb0112b41df/newIDE/app/src/ObjectGroupEditor/index.js#L41-L43
This doesn't happen though, as when `onChoose` is called on the autocomplete, the new value isn't read from the props by setting the state `inputValue` to `null`. 

I toyed arround with the expressions autocomplete as I believe it is using the same component and I couldn't notice any regressions with those changes.

### DIFF
--- a/newIDE/app/src/UI/SemiControlledAutoComplete.js
+++ b/newIDE/app/src/UI/SemiControlledAutoComplete.js
@@ -33,7 +33,6 @@ type Props = {|
   onChange: string => void,
   onChoose?: string => void,
   dataSource: DataSource,
-  clearOnChoose: boolean,
 
   id?: string,
   onBlur?: (event: SyntheticFocusEvent<HTMLInputElement>) => void,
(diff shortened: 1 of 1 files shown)


4ian:
Nice, can you try on a "touch screen", for example by using the Chrome devtools? The autocomplete should notably work when editing a parameter inline in the events sheet, I remember issues with this (you would select something and it would get discarded, etc...).
Would be good to verify all the use cases on a touch screen to ensure no regressions :)

arthuro555:
I tried on my smarthphone browser and there was no problem in the inline popover autocomplete.

4ian:
Seems to work well, thanks!

Bouh:
Close #2039

PR #2377, opened 2021-02-21
Use css instead of electron for color scheme detection

arthuro555:
Instead of using electron, use css to detect the users prefered color scheme, to ensure browser compatibility.


4ian:
Nice. Tests were failing because tests are not running a browser, just Node (for speed and reliability). I've fixed this.

PR #2380, opened 2021-02-22
Support numbers when accessing to a variable child with brackets notation

4ian:
For example: `MyVariable[2]`, `MyVariable[1 + 2]`, `MyVariable[2 + Random(1)]`...

Technically:
* This works by adding support for "number|string" type in expression parser. The first parsed number/string/function call determines the type of the expression.
* Slightly rework gd::ExpressionMetadata and gd::MetadataProvider, so that expressions have their return type in their metadata.
* Added tests for this new "number|string" type.
* Code generation still work because `getChild` works when passing a number... but this will be "officially" supported when @arthuro555 PR is merged: #2137


4ian:
Merging this as this will be necessary for arrays to be used properly, tests are passing, and this is not impacting anything else.

4ian:
@arthuro555 Note that in case you think it's more efficient, I can fairly easily generate a call to something like `getChildAt(...)` if a number was entered, and `getChild(...)` in case a string is entered.
This being said, I think it might be better/easier to keep a single `getChild()` that takes a `number | string` in the game engine. Because even if we did use getChildAt/getChild, we would probably make both compatible with both structure and arrays (so that you can also use numbers to access to a structure children).

In any case, let me know after rebasing on the master branch if it works good with arrays :)

arthuro555:
The autocomplete looks broken, it always only shows number expressions

PR #2384, opened 2021-02-24
Add tests on the Platformer extension

D8H:
I'm adding some tests to cover every transition (not yet finished) 
GDevelopPlatformer
I took a look at the code and I think a State pattern could do well for readability (or at least replacing the booleans by an enum). So I'm adding tests in prevision for a refactor.

### DIFF
--- a/Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
+++ b/Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
@@ -67,7 +67,7 @@ const addLadderObject = runtimeScene => {
   return platform;
 };
 
-describe('falling', function() {
+describe('(falling)', function() {
   let runtimeScene;
   let object;
   let platform;
@@ -228,9 +228,26 @@ describe('falling', function() {
     runtimeScene.renderAndStep(1000 / 60);
     expect(object.getY()).to.be(-30); // -30 = -10 (platform y) + -20 (object height)
   });
+
+  it('falls when a platform is moved away', function() {
+    object.setPosition(0, -32);
+    // Ensure the object falls on the platform
+    for (let i = 0; i < 10; ++i) {
+      runtimeScene.renderAndStep(1000 / 60);
+    }
+    //Check the object is on the platform
+    expect(object.getY()).to.be.within(-31, -30); // -30 = -10 (platform y) + -20 (object height)
+    expect(object.getBehavior('auto1').isFalling()).to.be(false);
+    expect(object.getBehavior('auto1').isMoving()).to.be(false);
+
+    // move the platform away
+    platform.setPosition(-100, -100);
+    runtimeScene.renderAndStep(1000 / 60);
+    expect(object.getBehavior('auto1').isFalling()).to.be(true);
+  });
 });
 
-describe('jump and jump sustain (round coordinates on)', function() {
+describe('(jump and jump sustain, round coordinates on)', function() {
   let runtimeScene;
   let object;
   let platform;
@@ -431,9 +448,41 @@ describe('jump and jump sustain (round coordinates on)', function() {
     );
     expect(object.getBehavior('auto1').isGrabbingPlatform()).to.be(false);
   });
+
+  it('should not grab a platform while walking', function() {
+    const topPlatform = addPlatformObject(runtimeScene);
+    topPlatform.setPosition(20, platform.getY() - object.getHeight());
+    runtimeScene.renderAndStep(1000 / 60);
+
+    // Ensure the object falls on the platform
+    for (let i = 0; i < 10; ++i) {
+      runtimeScene.renderAndStep(1000 / 60);
+    }
+
+    // Check the object is on the platform
+    expect(object.getY()).to.be(-30); // -30 = -10 (platform y) + -20 (object height)
+    expect(object.getBehavior('auto1').isFalling()).to.be(false);
+    expect(object.getBehavior('auto1').isMoving()).to.be(false);
+
+    // try to grab the platform
+    for (let i = 0; i < 30; ++i) {
+      object.getBehavior('auto1').simulateRightKey();
+      runtimeScene.renderAndStep(1000 / 60);
+      expect(object.getBehavior('auto1').isOnFloor()).to.be(true);
+    }
+
+    // The object is where it could grab the top platform if it where falling.
+    expect(object.getX()).to.be.within(
+      topPlatform.getX() - object.getWidth(),
+      topPlatform.getX() - object.getWidth() + 2
+    );
+    expect(object.getY()).to.be(topPlatform.getY());
+    // Check that the object didn't grabbed the platform
+    expect(object.getBehavior('auto1').isGrabbingPlatform()).to.be(false);
+  });
 });
 
-describe('jumpthru', function() {
+describe('(jumpthru)', function() {
   let runtimeScene;
   let object;
   let platform;
@@ -542,7 +591,7 @@ describe('jumpthru', function() {
   });
 });
 
-describe('rounded coordinates (moving platforms)', function() {
+describe('(rounded coordinates, moving platforms)', function() {
   let runtimeScene;
   let object;
   let platform;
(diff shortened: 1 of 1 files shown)


Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
```diff
@@ -42,41 +42,72 @@ const addPlatformObject = runtimeScene => {
...
-  // Put a platform.
-  const platform = addPlatformObject(runtimeScene);
-  platform.setPosition(0, -10);
+describe('gdjs.PlatformerObjectRuntimeBehavior', function() {
+describe('falling', function() {
+  var runtimeScene;
+  var object;
+  var platform;
+
+  beforeEach(function() {
```
4ian, line 74: Great thing to have this beforeEach to reset the state before each test :) 👍

Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
```diff
@@ -170,43 +229,49 @@ describe('gdjs.PlatformerObjectRuntimeBehavior', function() {
   });
 });
 
-describe('gdjs.PlatformerObjectRuntimeBehavior, jump and jump sustain (round coordinates on)', function() {
-  const runtimeScene = makeTestRuntimeScene();
+describe('jump and jump sustain (round coordinates on)', function() {
+  var runtimeScene;
+  var object;
+  var platform;
+
```
4ian, line 236: nitpicking: use `let` (or `const`) always instead of var when writing new code.

Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
```diff
@@ -642,3 +737,211 @@ describe('gdjs.PlatformerObjectRuntimeBehavior and gdjs.PlatformRuntimeBehavior
...
+    // Climb the ladder
+    for (var i = 0; i < 14; ++i) {
+      object.getBehavior('auto1').simulateUpKey();
+      runtimeScene.renderAndStep(1000 / 60);
+    }
+    // Check that we reached the maximum height
+    const playerAtLadderTop = ladder.getY() - object.getHeight();
+    expect(object.getY()).to.be.within(playerAtLadderTop, playerAtLadderTop + 2);
+
+    //TODO it fall at the end but that may not be expected.
```
4ian, line 825: It's kind of expected because when you're on the top of a ladder, it's not considered as a platform so you start to fall. Feel free to add a check, checking this is the case for one frame.

Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
```diff
@@ -42,41 +42,72 @@ const addPlatformObject = runtimeScene => {
   return platform;
 };
 
-describe('gdjs.PlatformerObjectRuntimeBehavior', function() {
-  const runtimeScene = makeTestRuntimeScene();
-
-  // Put a platformer object in the air.
-  var object = new gdjs.RuntimeObject(runtimeScene, {
-    name: 'obj1',
+const addLadderObject = runtimeScene => {
```
4ian, line 45: Nitpicking: let's put these two functions inside the `describe('gdjs.PlatformerObjectRuntimeBehavior', function() { `

So that everything is contained and no risk of leaking in the global namespace :) 
(I should have done this already!)

4ian:
Thanks for the tests! Great initiative.

> I took a look at the code and I think a State pattern could do well for readability (or at least replacing the booleans by an enum). So I'm adding tests in prevision for a refactor.

Yeah that would make sense to use an enum. Let's finish this PR to add tests and happy to consider a refactoring after :)

Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
D8H: Thinking more about it, it's easier to add events to make the player stop at the top of the ladder that the other way around. So, it makes sense.

4ian: Yep, though we could also add an option later to avoid falling when on top (like a jump thru), or creators can add a jump thru, etc... Easier this way than mandating a stop at the end :)

D8H:
* The allowed transitions are covered (X: existing, +: added)
```plaintext
OnFloor -> Falling  X'can fall when in the air' rename it: 'fall after reaching a cliff' ?, add 'fall when the platform desapear' ?
OnFloor -> Jumping  X'can jump'
OnFloor -> OnLadder +'can climb a ladder'

Jumping -> Falling  X'can jump'
Jumping -> OnLadder +'can grab a ladder when jumping'

Falling -> OnFloor  X'can jump'
Falling -> Grabbing X'can grab, and release, a platform'
Falling -> OnLadder +'can grab a ladder when falling'

Grabbing -> Faling  X'can grab, and release, a platform'
Grabbing -> Jumping +'can grab a platform and jump'

OnLadder -> Jumping +'can jump from a ladder'
OnLadder -> Falling +'can fall from a ladder'
OnLadder -> OnFloor +'can walk from a ladder'
```

* The forbidden transitions are not yet covered
```plaintext
OnFloor -> Grabbing 'should not grab a platform while on the floor'

Jumping -> Grabbing +'should not grab a platform while jumping'
Jumping -> OnFloor  X'can jump through the jumpthru'

Falling -> Jumping  X'should not when falling'

OnLadder -> Grabbing 'should not grab a platform when grabbed to a ladder'
```

* There are transitions I don't know how if they're allowed or not, I'll test.
```plaintext
Grabbing -> OnFloor  ?'should follow a moving platform when grabbed to another'
Grabbing -> OnLadder ?'can grab a ladder when grabbed to a platform'
```

4ian:
Thanks! Very clear explanations.

> add 'fall when the platform desapear' ?

Yep that's a good idea :) "disappear" though ;) 

> rename it: 'fall after reaching a cliff'

Not sure if I would understand this name, maybe keep "can fall when in the air", and optionally precise how you reached this state: "can fall when in the air, after a jump", etc...

> Grabbing -> OnLadder ?'can grab a ladder when grabbed to a platform'

I don't think this is allowed

> Grabbing -> OnFloor  ?'should follow a moving platform when grabbed to another'

You mean like the player is grabbing a platform, and another platform is coming from below (or above)? 
I *think* this should force the player to release the grabbed platform (by looking at the calls to `_releaseGrabbedPlatform`), but please double check :)

Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
```diff
@@ -784,64 +824,79 @@ describe('ladder', function() {
     ladder.setPosition(30, -10 - ladder.getHeight());
   });
 
-  it('can climb a ladder', function() {
-    object.setPosition(30, -32);
-    // Ensure the object falls on the platform
-    for (var i = 0; i < 10; ++i) {
+  const fall = (frameCount) => {
```
4ian, line 827: Great helper functions 👍

Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
```diff
@@ -642,3 +858,271 @@ describe('gdjs.PlatformerObjectRuntimeBehavior and gdjs.PlatformRuntimeBehavior
...
+    object.getBehavior('auto1').simulateRightKey();
+    runtimeScene.renderAndStep(1000 / 60);
+    expect(object.getBehavior('auto1').isGrabbingPlatform()).to.be(true);
+
+    // try to grab the ladder
+    object.getBehavior('auto1').simulateLadderKey();
+    runtimeScene.renderAndStep(1000 / 60);
+    // panic mode! (probably a bug)
+    expect(object.getBehavior('auto1').isOnLadder()).to.be(true);
+    expect(object.getBehavior('auto1').isGrabbingPlatform()).to.be(true);
```
D8H, line 1125: The platformer object has a survival instinct! :scream:

D8H:
I think every transition (allowed and forbidden) are covered.

4ian:
Thanks! Great work on this 👍
I'll run the autoformatting now on all the extension `.spec.js` files :)

PR #2385, opened 2021-02-24
Replace swapi.co with swapi.dev in JSON parse from api example

arthuro555:
swapi.co, which was used in the "Parse JSON from API" example, has been abandoned and the servers have been taken down. This PR uses swapi.dev, a maintained for of that project. [This was originally reported on the forum](https://forum.gdevelop-app.com/t/how-come-this-doesnt-work-json-to-variable-example/30874/2?u=arthuro555)).


Bouh:
Thank you for this fix!

PR #2388, opened 2021-02-26
Platformer Refactor: follow the state with an enum instead of booleans

D8H:


Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -435,16 +445,15 @@ namespace gdjs {
...
         this._timeSinceCurrentJumpStart = 0;
         this._jumpKeyHeldSinceJumpStart = true;
 
+        //FIXME can't be Jumping and OnFloor at the same time, need a fix and a test
         //this._isOnFloor = false; If floor is a very steep slope, the object could go into it.
-        this._isOnLadder = false;
         this._currentJumpSpeed = this._jumpSpeed;
         this._currentFallSpeed = 0;
-        this._isGrabbingPlatform = false;
       }
```
D8H, line 457: It relied on the superposition of OnFloor and Jumping to fix an issue with steep slopes. So, there certainly is a regression. I need check this, write a test and find another way to fix it.

Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
```diff
@@ -642,3 +858,271 @@ describe('gdjs.PlatformerObjectRuntimeBehavior and gdjs.PlatformRuntimeBehavior
...
+    object.getBehavior('auto1').simulateRightKey();
+    runtimeScene.renderAndStep(1000 / 60);
+    expect(object.getBehavior('auto1').isGrabbingPlatform()).to.be(true);
+
+    // try to grab the ladder
+    object.getBehavior('auto1').simulateLadderKey();
+    runtimeScene.renderAndStep(1000 / 60);
+    // panic mode! (probably a bug)
+    expect(object.getBehavior('auto1').isOnLadder()).to.be(true);
+    expect(object.getBehavior('auto1').isGrabbingPlatform()).to.be(false);
```
D8H, line 1125: The only change in the tests is this which is a fix.

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -1256,6 +1269,17 @@ namespace gdjs {
       );
     }
   }
+
+  export namespace PlatformerObjectRuntimeBehavior {
```
4ian, line 1273: If this possible not to export this State/namespace? To avoid exposing unnecessary details to the outside world.

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
4ian: Try to make a floor that is rotated so that it's very steep. I think the issue that then when you would jump from it, you would get stuck into the floor, if you press the key to move toward the floor, or you would stop the jump almost immediately.

D8H: I could reproduce on this PR, but it's fixed by the other PR:
https://github.com/4ian/GDevelop/pull/2396
So, I think this PR can be forgotten.

PR #2389, opened 2021-02-27
Fix compilation on iOS of exported games failing

4ian:
* This was because an SDK related to AdMob was wrongly included and would trigger a compiler error in XCode.


Bouh:
Surely nothing, it is the console from the IDE
image

4ian:
Surely the time for GDevelop.js to recompile?

PR #2390, opened 2021-02-27
Fix mouse wheel in the IDE and in games

### LINKED ISSUE #2200: Mouse wheel not working on Firefox
This is the snippet I use for wheel zoom on my current project:
image

I'm not sure if this is because of beta103, but it's broken.
Works fine on preview, works fine on Chrome, doesn't work on Firefox.

Example project attached for your convenience.
[BrokenWheelZoom.zip](https://github.com/4ian/GDevelop/files/5792924/BrokenWheelZoom.zip)

Bouh:
Close #2200


4ian:
Thank you :)

PR #2392, opened 2021-02-27
Fix collision (and raycast) not always triggered with sprite having a non default center

### LINKED ISSUE #2379: Updating the Center and Origin Points break collision mask positioning.
## Describe the bug

When you update an object's center point, origin point, or both, the hitbox is impacted, but not 1:1 (as far as I can tell).
The user has no way to know where this new hitbox will be, either.

- What I would expect to happen:
    - The hitbox remains accurate to how it appears, position wise, in the hitbox editor.
- What actually happens:
    - The hitbox appears to move somehow based off the other points moving, but not 1:1, almost like it's being skewed based off the origin point vs the center point.

This came up on the forums back in May 2020, but it doesn't look like we ever had a bug report filed on it:  https://forum.gdevelop-app.com/t/sprites-with-modified-points-do-not-collide-correctly/24785/7

VegaTato on the discord brought this back up, and I was able to reproduce it.

## To Reproduce
1. Create two objects.
2. Update the center and/or origin points of one object. E.g. move the center to the left.
3. Set up events to test for collision between unmodified object1, and modified object2.
4. Preview the game and have object1 try to collide with object2. Object1 will be able to move a portion of the way into object2 before collision is detected,

Here's some screenshots. The hitbox is not modified and is the full size of the object.
Events:
image

Points on modified object: 
image

Touching the object, no collision detected:
image

Collision finally detected:
image

Example project:
[PointCollisionIssue.zip](https://github.com/4ian/GDevelop/files/6019385/PointCollisionIssue.zip)

## Other details
* Windows 10 Pro
* B105

4ian:
Fix #2379

and potentially other collision related issues!

### DIFF
--- a/Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
+++ b/Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
@@ -914,33 +914,6 @@ namespace gdjs {
       }
     }
 
-    //This is the naive implementation when the platforms manager is simply containing a list
-    //of all existing platforms:
-    /*var o1w = this.owner.getWidth();
-        var o1h = this.owner.getHeight();
-        var obj1BoundingRadius = Math.sqrt(o1w*o1w+o1h*o1h)/2.0 + maxMovementLength;
-        //Get all platforms and keep only
-        var allPlatforms = this._manager.getAllPlatforms();
-        for (var k in allPlatforms.items) {
-            if (allPlatforms.items.hasOwnProperty(k)) {
-                var obj2 = allPlatforms.items[k].owner;
-                var o2w = obj2.getWidth();
-                var o2h = obj2.getHeight();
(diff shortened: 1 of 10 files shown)


Bouh:
The debug draw used can be interesting for everyone!
Using it in the IDE and in game could be very helpful to see what's going on.

4ian:
> It simply makes more sense directly in the engine.

Yep! It's better to keep in the engine.

Note that this is very rudimentary. For example, **it does not** support camera movement, so it's really a tool for developers to verify the hit boxes/AABB of objects :)

4ian:
@Bouh FYI I've improved a bit the debug draw to show both AABB (blue) and hitboxes (red), as well as the drawable X/Y and center point:

image

Bouh:
Wow 😍

4ian:
Note that it still does not support camera movements ;) But could be not too hard to add (there is an argument with the layer coordinates).

Issue #1947, opened 2020-08-27
Probable issue: NPM dependencies not working in previews


arthuro555:
## Describe the bug
I didn't test yet but I am 99% sure that npm dependencies wouldn't work in the IDE as it would require having the node module in the newIDE/resources folder to load that module. Runtime should be fine tho as with npm install the dependency is bundled.

## Potential solutions
There are two options: 
1. Bundle npm dependencies with the ide (using some script that auto install the needed deps before building)
2. Detect when an extension with npm deps is used, and download needed deps to some cache if not already present in that cache and load modules from there

1 would make the ide bigger and longer to build, but 2 would make some features require internet connection (to download additional packages)

We could also just not support npm dependencies on the local editor, like on the web editor but it wopuld be a shame and confuse some advanced users that know that previews run on electron too.

4ian:
Before going further down in the rabbit hole: can you give an example of an hypothetical npm dependency what would cause problem?
Think about what they are used for. :)

arthuro555:
Let's say I have a npm dependency that adds additional platform dependent APIs wrapped for windows mac and linux. Users would expect that as long the game is running in a window the game works (as well in previews as in exported games) but it would only work in exported games, and users would think they did something wrong and not think of trying to export the game.

4ian:
I understand the problem, but my question was more: do you have a specific example, a real npm package name in mind?

It's definitely sure that everything that works in an exported game should work in a preview - the contrary would be counter intuitive and not user friendly. But I need a real example, because currently this issue seems just an hypothetical problem that we don't have for now :) And we risk over engineering a complex solution, only for something that will never be used or will be used once.

arthuro555:
I was thinking about maybe using fs-extra for the filesystem extension (has a nice copying API), about #1931 , and for integrating greenworks (SteamWorks API node native module bindings)

arthuro555:
I mean the issue is still technically there... But it doesn't impact anything as of now. Should it be closed or not? IDK.

4ian:
If it does not impact anything, is it still a problem? ;) 
Let's maybe continue the discussion on the discussion related to a specific extension needing this?

arthuro555:
Well there isn't one in particular as of now but if a new contributor wanted to create one that use an npm module they'd be unpleasantly surprised to have this issue.

PR #2393, opened 2021-02-28
Add extension dependencies support to event based extensions

arthuro555:
Allows events based extension to declare npm or cordova dependencies. 

In the future, maybe we could look into adding a "CDN" type to have a dependency type that works across all platforms (it would download the script from the cdn link specified as the export name and add it to the includes in index.html).


Core/GDCore/Project/EventsFunctionsExtension.h
```diff
@@ -165,6 +195,24 @@ class GD_CORE_API EventsFunctionsExtension : public EventsFunctionsContainer {
    */
   void Init(const gd::EventsFunctionsExtension& other);
 
+  void SerializeDependencyTo(const gd::DependencyMetadata& dependency,
```
arthuro555, line 198: I am not sure if this belongs here. My thought process for putting this here instead of the dependency metadata class directly is that dependency metadata shouldn't be serialized in it's other usage (in the extension) so it should be here as this is the only case where there is a need to serialize it.

newIDE/app/src/EventsFunctionsExtensionsLoader/MetadataDeclarationHelpers.js
```diff
@@ -31,8 +31,29 @@ export const declareExtension = (
...
+ */
+export const declareExtensionDependencies = (
+  extension: gdPlatformExtension,
+  eventsFunctionsExtension: gdEventsFunctionsExtension
+) =>
+  mapVector<gdDependencyMetadata, void>(
+    eventsFunctionsExtension.getAllDependencies(),
+    dependency => {
+      extension
+        .addDependency()
```
arthuro555, line 49: Here, I was not sure if I should add a method to the platform extension to push the dependency directly on its dependencies vector. I think it was said before that we should avoid exposing too much from the c++, so I decided to go the "verbose way". Is it ok?

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/ExtensionDependenciesEditor.js
```diff
@@ -0,0 +1,168 @@
...
+                        if (checkNameExists(newName, deps))
+                          Window.showMessageBox(
+                            `This name is already in use! Please use a unique name.`,
+                            'error'
+                          );
+                        else
+                          eventsFunctionsExtension
+                            .getAllDependencies()
+                            .at(index)
+                            .setName(newName) && forceUpdate();
```
4ian, line 101: I highly discourage abusing && like this: 
- If the function returns void/undefined/falsy value, the second instruction won't work
- if someones wants to add an instruction, it's a pain.
- you have a chance to forget the && one day and end up with something executed outside of the if else.

Prefer:
- using a comma if you really really want not to write brackets:
```
if(false) 
  1, 2 
else 
  3, 4
```
- or (better), just use brackets.

In some codebases, not writing brackets is not even allowed, because of the risk of forgetting them when adding an instruction.
I'm not against them, but only if it's for a single line instruction. If it's multiple lines, and there is a else => brackets.

Core/GDCore/Project/EventsFunctionsExtension.h
4ian: Seems reasonable 👍

Core/GDCore/Project/EventsFunctionsExtension.cpp
```diff
@@ -59,6 +59,10 @@ void EventsFunctionsExtension::SerializeTo(SerializerElement& element) const {
   element.SetAttribute("previewIconUrl", previewIconUrl);
   element.SetAttribute("iconUrl", iconUrl);
   element.SetAttribute("helpPath", helpPath);
+  auto& dependenciesElement = element.AddChild("dependencies");
+  dependenciesElement.ConsiderAsArray();
+  for (auto& dependency : dependencies)
```
4ian, line 64: nitpicking: this could be `const auto&` I think?

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/ExtensionDependenciesEditor.js
```diff
@@ -0,0 +1,168 @@
...
+                  <TableCell>
+                    <SemiControlledTextField
+                      commitOnBlur
+                      value={dependency.getName()}
+                      onChange={newName => {
+                        if (checkNameExists(newName, deps))
+                          Window.showMessageBox(
+                            `This name is already in use! Please use a unique name.`,
+                            'error'
+                          );
```
4ian, line 96: Prefer `showWarningBox` for warnings that interrupt the workflow (but not an error, because nothing dangerous happened, it's a mere explanation that you can't go further).
Also pass `delayToNextTick` to true to avoid some weird freeze/infinite loop of the app on some linux distro because of a blur + a message box displayed at the same time.

newIDE/app/src/EventsFunctionsExtensionsLoader/MetadataDeclarationHelpers.js
4ian: It's ok to do it in JS because that's how the rest of this file is done.
This being said, it's true that it's verbose and could be even a source of bug if we add a new field later (you add it to the C++ side, you add it in the UI... and it's not working! Then you debug until you find this and ask yourself "why are not all fields copied???").

So I would say:
- The logic of "adding all dependencies" should stay in this JS file (and maybe later we'll rewrite it in C++, but that's not a priority).
- BUT we can still offer a convenient way of copying a DependencyMetadata... it would be easy in C++ (just assign the reference: `auto& extensionDependency = extension.AddDependency(); extensionDependency = dependency`), but not in JS. And we also can't overload `addDependency` 😬

Might be worth adding either a method to copy a dependency (but that's a bit useless in C++...) or a method to copy the entire vector as you mentioned.
Let me know what you think.

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/ExtensionDependenciesEditor.js
```diff
@@ -0,0 +1,168 @@
...
+            )}
+          </TableBody>
+        </Table>
+        <FlatButton
+          icon={<Add />}
+          label={<Trans>Add</Trans>}
+          onClick={addDependency}
+        />
+      </TableContainer>
+      <br />
```
4ian, line 156: Don't use br :) Use Line/Column/Spacer, as they ensure a consistent spacing across the app.

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/ExtensionDependenciesEditor.js
```diff
@@ -0,0 +1,168 @@
...
+      .setExportName('my-dependency')
+      .setVersion('1.0.0')
+      .setDependencyType('cordova');
+    forceUpdate();
+  };
+
+  return (
+    <>
+      <TableContainer
+        component={({ children }) => <Paper elevation={4}>{children}</Paper>}
```
4ian, line 60: I don't think you need this props? Just put <Paper> around your <TableContainer>

4ian:
Overall the approach looks good 👍

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/ExtensionDependenciesEditor.js
arthuro555: I just followed the examples in the MUI docs:
https://github.com/mui-org/material-ui/blob/48bf11343f04db242b69a0a8ceef3c67ac41de4a/docs/src/pages/components/tables/BasicTable.js#L33

arthuro555:
Also, I added the possibility to use NPM dependencies but there is a slight problem (issue #1947 ), the npm dependencies are only added to package.json, but not installed on previews, so they only work on full builds (where `npm install` is run).

Core/GDCore/Extensions/Metadata/DependencyMetadata.h
```diff
@@ -132,6 +133,18 @@ class GD_CORE_API DependencyMetadata {
...
+  void CopyFrom(const DependencyMetadata& dependencyMetadata) {
+    name = dependencyMetadata.name;
+    exportName = dependencyMetadata.exportName;
+    version = dependencyMetadata.version;
+    dependencyType = dependencyMetadata.dependencyType;
+    extraData = dependencyMetadata.extraData;
+    onlyIfSomeExtraSettingsNonEmpty =
+        dependencyMetadata.onlyIfSomeExtraSettingsNonEmpty;
+    onlyIfOtherDependencyIsExported =
+        dependencyMetadata.onlyIfOtherDependencyIsExported;
```
4ian, line 145: Give a try at just `*this = dependencyMetadata;` or `operator=(dependencyMetadata)`.
We should not need to redefine what a copy is, as the compiler is generating this code for us in the form of the assignment operator (and the copy constructor).

PR #2394, opened 2021-02-28
Make debug hitboxes & points usable in games

Bouh:
This is not ready to merge!
Two options are still in WIP, i plan to finish it in two days i'am busy tomorrow.

- [x] Take into account camera movement.

- [x] An new action is shipped in Debugger Tools category.
  - [x] Option YES/NO for show the debug on hidden objects
  - [x] Option YES/NO for show custom points
  - [x] Option YES/NO for show points names 
  
image

- [x] Think about AABB and hitboxes if both have the same shape, which ones have priority to be drawn, maybe we can avoid to compute debugDraw for one of them.
- [x] Think about rotations, AABB is not correct
image

  - For another pull request: Reuse the debugDraw code from runtime in IDE and add a new button & panels for options.

Thanks for the previous job on debug code!


4ian:
Thanks for starting this :) 

> Think about AABB and hitboxes if both have the same shape, which ones have priority to be drawn, maybe we can avoid to compute debugDraw for one of them.

I suggest we keep both: a blue rectangle for the hitbox, that is filled (so you can see it even if the hitbox is on top) and hitboxes in red line (with no filling, so you can see the AABB below).

> Think about rotations, AABB is not correct

I think it's correct.
An AABB is just an approximation of the real size of the object. It's mostly used to make what we call "broad check": very quick checks to quickly dismiss objects that are for sure not visible/colliding/in the area you're searching. Then the hitboxes are used for the real checks.

When an object is not rotated, by default the AABB is basically its X/Y + its width/height. When it's rotated, a more precise AABB is set by using the hitboxes points (this could be changed).

So overall as long as the object is always inside the AABB, and the hitboxes are fine, it should be fine.

Bouh:
It's really confusing to have an AABB that doesn't wrap the object in the same way depending on the angle.
AABB with angle 0 is based on the object dimension, and when it not angle 0, the AABB is based on mix and max from the hitboxes.
We should always be based on the object dimension or on the texture frame for AABB no ?
However it not my deal here.

The funny part: 
image

All Points are rotated with the object if needed.
The point names can be hidden, it is an option in the action.
Colors of points are the colors from the point editor.
I take in account or not the hidden objects or from an hidden layer with an option in the action.

Custom points are in blue, the original colors in the point editor is white, but white not very visible, so i've edited the sprite in the point editor for use a cross with color like origin (red) and center (yellow), new custom points are in blue!

image


- I tried a camera movement centered on an object, it seems that my calculations to compensate the movement are wrong.
- I've just seen the pixi text used for point names is not deleted whenthe object is delete, the text of the point name stay on screen while the hitboxes gone. I store the debug text in the instance it should be destroy no?

GDJS/Runtime/runtimeobject.ts
```diff
@@ -388,6 +388,23 @@ namespace gdjs {
...
+      origin: { x: float; y: float },
+      angle: float
+    ): { x: float; y: float } {
+      angle *= Math.PI / 180;
+      const pointX = point.x - origin.x;
+      const pointY = point.y - origin.y;
+      const x = pointX * Math.cos(angle) + pointY * Math.sin(angle) + origin.x;
+      const y = -pointX * Math.sin(angle) + pointY * Math.cos(angle) + origin.y;
+      return { x: Math.round(x), y: Math.round(y) };
+    }
```
4ian, line 406: Could you use `getPointX`/`getPointY` (only for sprite, but that's fine you can do a `if (obj instanceof gdjs.SpriteRuntimeObject) { ... }`? Note that for it to be efficient, we can add a `getPointPosition` that returns the array (instead of calling `getPointX` and then `getPointY`, which is inefficient).

4ian:
> We should always be based on the object dimension or on the texture frame for AABB no ?
> However it not my deal here.

Calculation of the AABB is 10-15% faster when the angle is 0, because we only have to do a few calculations (X,Y,width,height). This speed increasing is I think worth it.

> I've just seen the pixi text used for point names is not deleted whenthe object is delete, the text of the point name stay on screen while the hitboxes gone. I store the debug text in the instance it should be destroy no?

Not necessarily, you have to remove it manually from the the Pixi container. If the object is removed, the texts are still living inside the pixi container. It's just the reference from the object that was lost, but they still live inside the Pixi array.

Remember though that the Debug Draw should stay fairly simple and should just "read" the objects, not write anything "inside them". Instead: you can store inside the class containing the debug draw a map from object id to texts. When iterating on objects, you can use it. And you can use it to remove useless points if they are not touched anymore. 

For example:
```
// Reset the boolean "wasRendered" of all points of objects to false:
for(var id in this._renderedObjectsPoints) {
  this._renderedObjectsPoints[id].wasRendered = false;
}

// Now iterate on the objects like you already do:
for (let i = 0; i < instances.length; i++) {
  const object = instances[i];	  
  if (!this._renderedObjectsPoints[id]) {
    // Create the structure to store the points in memory
    this._renderedObjectsPoints[id] = { wasRendered: true, points: {} };
  }
  // Mark the points as rendered
  const renderedObjectPoints = this._renderedObjectsPoints[id];
  renderedObjectPoints.wasRendered = true;
  // ...
  
  // Whenever you draw a point, either use the one in points or a new one:
  if (!renderedObjectPoints.points["origin"]) { renderedObjectPoints.points["origin"] = new PIXI.Text(...) }
  renderedObjectPoints.points["origin"].position.set(..., ...);
  
  // See the idea? We keep the points not inside the runtime object, but inside a separate "structure"
}

// After we've iterated on all objects, maybe some were destroyed. So clean their points:
for(var id in this._renderedObjectsPoints) {
  const renderedObjectPoints = this._renderedObjectsPoints[id];
  if (!renderedObjectPoints.wasRendered) {
    // Clean our point objects
    for(var name in renderedObjectPoints.points) {
      // Remove from the PIXI container the `renderedObjectPoints.points[name]`.
    }
  }
}
```

Note that this approach of "dynamically creating and removing PIXI objects" is a bit how the `RenderedInstance` of the editor work: we iterate on instances, and create PIXI object if needed. And if one is not "visited", it means its instance was removed. So we remove it :)

Bouh:
I try to understand your example, but something eludes me.
What do you do with **object**, you don't use it. And **id** in `this._renderedObjectsPoints[id]` is out of scope.

4ian:
Sorry I've updated it, you read the `id` from `object.id` :)

Bouh:
https://user-images.githubusercontent.com/1670670/109709287-99c6f680-7b9c-11eb-8076-c22958a8752f.mp4

This looks cool!
We can notice when the camera is centered on an object the hitboxes seem lagging, there is a delay between Pixi and the debug code.
When there is no actions for force the camera nothing lagging.

Extensions/DebuggerTools/JsExtension.js
```diff
@@ -20,39 +20,66 @@ import { type ObjectsRenderingService, type ObjectsEditorService } from '../JsEx
...
-    runExtensionSanityTests: function(gd /*: libGDevelop */, extension /*: gdPlatformExtension*/) {
-        return [];
-    },
-}
+    extension
+      .addAction(
+        'drawCollisions',
+        _('Draw collisions hitboxes and points'),
+        _(
+          'This draw all hitboxes collsions and points of objects.' // TODO expliqué les couleurs des hitboxes
```
4ian, line 59: This activates the display of rectangles and information on screen showing the objects bounding boxes (blue), the hitboxes (red) and some points of objects.

Extensions/DebuggerTools/JsExtension.js
```diff
@@ -20,39 +20,66 @@ import { type ObjectsRenderingService, type ObjectsEditorService } from '../JsEx
...
+        ),
+        _('Draw collisions hitboxes and points'),
```
4ian, line 61: Include the parameters: `Activate debugging view of bounding boxes/collision masks: _PARAM1_, point names: _PARAM2_, custom points: _PARAM3_`

Extensions/DebuggerTools/debuggertools.ts
```diff
@@ -12,6 +12,24 @@ namespace gdjs {
...
         runtimeScene.getGame().pause(true);
       };
+
+      /**
+       * Draw collisions hitboxes and points
+       * @param runtimeScene - The current scene.
+       */
+      export const drawCollisionsAndPoints = function (
+        runtimeScene: gdjs.RuntimeScene,
+        viewInsivibleInstance: boolean,
```
4ian, line 22: typo + prefer "show" than "view": `showHiddenInstances`

Extensions/DebuggerTools/debuggertools.ts
```diff
@@ -12,6 +12,24 @@ namespace gdjs {
...
+        viewCustomPoints: boolean,
```
4ian, line 23: showCustomPoints

Extensions/DebuggerTools/debuggertools.ts
```diff
@@ -12,6 +12,24 @@ namespace gdjs {
...
+        viewPointsNames: boolean
```
4ian, line 24: showPointsNames

Extensions/DebuggerTools/debuggertools.ts
```diff
@@ -12,6 +12,24 @@ namespace gdjs {
...
+      ) {
+        runtimeScene.renderCollisionsAndPoints(
+          true,
```
4ian, line 27: Do you think we should have a parameter to set it to false?

GDJS/Runtime/pixi-renderers/runtimescene-pixi-renderer.ts
```diff
@@ -8,6 +8,10 @@ namespace gdjs {
     _pixiContainer: any;
     _debugDraw: PIXI.Graphics | null = null;
     _profilerText: PIXI.Text | null = null;
+    _renderedObjectsPoints: Array<{
```
4ian, line 11: Let's rename this to `_debugDrawRenderedObjectsPoints`. It's long and verbose, but clear that this is something related to `_debugDraw` and not needed otherwise :)

4ian: Also, this should not be an array, but an associative map, i.e: an object. `Record<number, { ...}>`

GDJS/Runtime/pixi-renderers/runtimescene-pixi-renderer.ts
```diff
@@ -19,6 +23,7 @@ namespace gdjs {
         : null;
       this._runtimeScene = runtimeScene;
       this._pixiContainer = new PIXI.Container();
+      this._renderedObjectsPoints = [];
```
4ian, line 26: Use objects for maps, not arrays. IDs are not necessarily in order, some IDs can be skipped, there are not guarantees, so if an ID is one day 100004, it means this array would have to be resized to contain 100004 elements. An object is the best way to associate things to a key.

GDJS/Runtime/pixi-renderers/runtimescene-pixi-renderer.ts
```diff
@@ -99,6 +112,18 @@ namespace gdjs {
       if (displayAABB) {
         for (let i = 0; i < instances.length; i++) {
           const object = instances[i];
+          const layer = this._runtimeScene.getLayer(object.getLayer());
+          const compensationCameraMovementX =
```
4ian, line 116: That's not really a compensation for "movement" (movement implies a speed), but rather just `cameraTopLeftX` and `cameraTopLeftY` :)

GDJS/Runtime/pixi-renderers/runtimescene-pixi-renderer.ts
```diff
@@ -120,44 +145,206 @@ namespace gdjs {
...
+            let polyPointX: float = 0;
+            let polyPointY: float = 0;
             hitboxes[j].vertices.forEach((point) => {
-              polygon.push(point[0]);
-              polygon.push(point[1]);
+              if (layer) {
+                polyPointX = point[0] - compensationCameraMovementX;
+                polyPointY = point[1] - compensationCameraMovementY;
+                polygon.push(polyPointX);
+                polygon.push(polyPointY);
```
4ian, line 190: just do
```suggestion
                polygon.push(point[0] - cameraTopLeftX);
                polygon.push(point[1] - cameraTopLeftY);
```

?

Extensions/DebuggerTools/debuggertools.ts
Bouh: how we call it?
enableDebugDrawRenderedObjectsPoints ? ^^

Bouh: Its worth to make it an action parameters?
If we want enable/disable the action we can just delete it or disable the event.

GDJS/Runtime/pixi-renderers/runtimescene-pixi-renderer.ts
```diff
@@ -120,44 +145,206 @@ namespace gdjs {
...
+          // For others objects that doesn't have origin point the position of the rendered object is used.
+          let originPoint = [
+            object.getDrawableX() - compensationCameraMovementX,
+            object.getDrawableY() - compensationCameraMovementY,
+          ];
+          if (object instanceof gdjs.SpriteRuntimeObject) {
+            originPoint = object.getPointPosition('origin');
+            originPoint[0] = originPoint[0] - compensationCameraMovementX;
+            originPoint[1] = originPoint[1] - compensationCameraMovementY;
+          }
```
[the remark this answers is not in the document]
Bouh: By using `object.getPointPosition('origin');` i get the position with the rotated position when the object is rotated. Which isn't the case with object.getX() and object.getY() or getDrawableX() getDrawableY().
Our aim is to view the origin, not the position right?
The origin isn't always the object position., on sprite objects origin can be changed. 

By using object.getX() and Y, see the origin point (mark 1) is at the object position.
It should be equal to the origin of the other object at mark 2.
image

Extensions/DebuggerTools/debuggertools.ts
Bouh: Besides it is not possible to activate or deactivate a boolean option of an action via events.
I don't really see the point.

Extensions/DebuggerTools/JsExtension.js
```diff
@@ -20,39 +20,75 @@ import { type ObjectsRenderingService, type ObjectsEditorService } from '../JsEx
...
 
-      return extension;
-    },
-    runExtensionSanityTests: function(gd /*: libGDevelop */, extension /*: gdPlatformExtension*/) {
-        return [];
-    },
-}
+    extension
+      .addAction(
+        'drawCollisions',
```
4ian, line 56: Let's call this `EnableDebugDraw`

Extensions/DebuggerTools/JsExtension.js
```diff
@@ -20,39 +20,75 @@ import { type ObjectsRenderingService, type ObjectsEditorService } from '../JsEx
...
+        _('Show collisions for hidden objects'),
+        '',
+        true
+      )
+      .setDefaultValue('no')
+      .addParameter('yesorno', _('Show points names'), '', true)
+      .setDefaultValue('yes')
+      .addParameter('yesorno', _('Show custom points'), '', true)
+      .setDefaultValue('yes')
+      .addParameter('yesorno', _('Enable debug draw'), '', true)
```
4ian, line 80: This should be the first parameter I think. When you use the action, you're first choosing: On or Off. Then in the case of On, you also set exactly what you want to see.

Extensions/DebuggerTools/JsExtension.js
```diff
@@ -20,39 +20,75 @@ import { type ObjectsRenderingService, type ObjectsEditorService } from '../JsEx
...
-}
+    extension
+      .addAction(
+        'drawCollisions',
+        _('Draw collisions hitboxes and points'),
+        _(
+          'This activates the display of rectangles and information on screen showing the objects bounding boxes (blue), the hitboxes (red) and some points of objects.'
+        ),
+        _(
+          'Activate debugging view of bounding boxes/collision masks (include invisible objects: _PARAM1_, point names: _PARAM2_, custom points: _PARAM3_)'
```
4ian, line 62: `Enable debugging view of bounding boxes/collision masks: _PARAM1_ (include invisible objects: _PARAM2_, point names: _PARAM3_, custom points: _PARAM4_)`

Extensions/DebuggerTools/debuggertools.ts
```diff
@@ -12,6 +12,25 @@ namespace gdjs {
       export const pause = function (runtimeScene: gdjs.RuntimeScene) {
         runtimeScene.getGame().pause(true);
       };
+
+      /**
+       * Draw collisions hitboxes and points
+       * @param runtimeScene - The current scene.
+       */
+      export const drawCollisionsAndPoints = function (
```
4ian, line 20: Let's call this `enableDebugDraw`?

Extensions/DebuggerTools/debuggertools.ts
```diff
@@ -12,6 +12,25 @@ namespace gdjs {
...
+        runtimeScene: gdjs.RuntimeScene,
+        showHiddenInstances: boolean,
+        showCustomPoints: boolean,
+        showPointsNames: boolean,
+        enableDebugDrawRenderedObjects: boolean
+      ) {
+        runtimeScene.renderCollisionsAndPoints(
```
4ian, line 27: (same)

4ian: (order of parameters is good there! :))

GDJS/Runtime/pixi-renderers/runtimescene-pixi-renderer.ts
```diff
@@ -8,6 +8,13 @@ namespace gdjs {
     _pixiContainer: any;
     _debugDraw: PIXI.Graphics | null = null;
     _profilerText: PIXI.Text | null = null;
+    _debugDrawRenderedObjectsPoints: Record<
+      number,
+      {
+        wasRendered: boolean;
+        points: { x: float; y: float };
```
4ian, line 15: This typing seems false.

GDJS/Runtime/pixi-renderers/runtimescene-pixi-renderer.ts
```diff
@@ -120,44 +148,204 @@ namespace gdjs {
...
           const rendererObject = object.getRendererObject();
           if (!cameraCoords || !rendererObject) {
             continue;
           }
 
+          // Create the structure to store the points in memory
+          if (!this._debugDrawRenderedObjectsPoints[id]) {
+            this._debugDrawRenderedObjectsPoints[id] = {
+              wasRendered: true,
+              points: { x: 0, y: 0 },
```
4ian, line 174: This seems false, points is not a single point, it's a map associating points name to their associated pixi texts.

GDJS/Runtime/pixi-renderers/runtimescene-pixi-renderer.ts
```diff
@@ -120,44 +148,204 @@ namespace gdjs {
...
-          debugDraw.line.color = 0xe8e868;
-          debugDraw.fill.color = 0xe8e868;
-          debugDraw.drawCircle(
-            object.getDrawableX() + object.getCenterX(),
-            object.getDrawableY() + object.getCenterY(),
-            3
-          );
+          // Draw points
+          // Center point of the object, with camera movement
+          const centerPoint = {
```
4ian, line 204: No need to allocate memory for an object, just store two variables `centerPointX` and `centerPointY`

GDJS/Runtime/pixi-renderers/runtimescene-pixi-renderer.ts
```diff
@@ -120,44 +148,204 @@ namespace gdjs {
...
+            // Mark the text points as rendered
+            renderedObjectTextPoints.wasRendered = true;
+            this._pixiContainer.addChild(
+              renderedObjectTextPoints.points['origin']
+            );
+          }
+
+          // Draw custom point
+          if (showCustomPoints) {
+            // @ts-ignore '_animationFrame' does not exist on type 'RuntimeObject | SpriteRuntimeObject'.
```
4ian, line 282: Guard this with a if `object instanceof gdjs.SpriteRuntimeObject` and TypeScript will understand the type of the object:

```suggestion
          if (showCustomPoints && object instanceof gdjs.SpriteRuntimeObject) {
```

GDJS/Runtime/pixi-renderers/runtimescene-pixi-renderer.ts
```diff
@@ -120,44 +148,204 @@ namespace gdjs {
...
+              'items'
+            ]) {
+              let customPoint: [float, float];
+
+              if (object instanceof gdjs.SpriteRuntimeObject) {
+                customPoint = object.getPointPosition(customPointName);
+              } else {
+                // There is not custom point for other object than SpriteRuntimeObject
+                continue;
+              }
```
4ian, line 295: Can be removed once you have a `instanceof` check in the if before

GDJS/Runtime/pixi-renderers/runtimescene-pixi-renderer.ts
```diff
@@ -120,44 +148,204 @@ namespace gdjs {
...
+
+              if (showPointsNames) {
+                console.log('ok');
```
4ian, line 298: console.log

Bouh:
Here how it running with a custom point on **Player** object, an object positionned on the origin point of the **Player** object,  **Player** rotated to 45°.
All options of the action has been tested.

[The video.](https://streamable.com/kmmhvr)

4ian:
> The video.

Looking great 🤩

GDJS/Runtime/pixi-renderers/runtimescene-pixi-renderer.ts
```diff
@@ -120,47 +149,214 @@ namespace gdjs {
...
+            // Mark the text points as rendered
+            renderedObjectTextPoints.wasRendered = true;
+            this._pixiContainer.addChild(
+              renderedObjectTextPoints.points['origin']
+            );
+          }
+
+          // Draw custom point
+          if (showCustomPoints && object instanceof gdjs.SpriteRuntimeObject) {
+            if (!object?._animationFrame?.points) continue;
```
4ian, line 277: Don't use ?., it seems it's not translated by esbuild and the tests are failing because of it I think:
image

More generally, using `?.` can almost always be replaced by a simple check and is often a way to say "huh I don't really know if I have something valid or not so I'll just put a ?" ;) 
Here you are sure you have an object. You're also sure that you have a gdjs.SpriteRuntimeObject. So the _animationFrame is either null, or not.
If it's not null, you're sure you have points. So you just need:
`if (!object._animationFrame) continue;`

GDJS/Runtime/runtimescene.ts
```diff
@@ -78,6 +78,29 @@ namespace gdjs {
       this.onGameResolutionResized();
     }
 
+    /**
+     * Activate or deactivate the debug visualization for collisions and points.
+     * @memberof gdjs.RuntimeScene
+     */
+    renderDebugDraw(
```
4ian, line 85: render means "calling this function will render things on screen". 
But this should not do it! There is already a function that is called `renderAndStep` in runtime scene.
Here, you said your function is "Activate or deactivate the debug visualization for collisions and points.". So it should:
1) be called `enableDebugDraw`
2) it should not call a function called `renderDebugDraw` :) It should just set on or off some booleans. Then, in `renderAndStep`, the runtime scene should call `renderDebugDraw` if it's enabled.

Calling clearDebugDraw is fine, because it's a one time operation.

GDJS/Runtime/runtimescene.ts
```diff
@@ -507,7 +530,10 @@ namespace gdjs {
       if (renderDebugDraw && this._layersCameraCoordinates) {
```
4ian, line 530: You should make this a boolean of the scene: `this._renderDebugDraw`.
That is set to true or false according to the the function `enableDebugDraw`. Same for the other booleans (or you can store them in the renderer, as you wish).

4ian:
I think there was a bit of confusion about `enableDebugDraw` and `renderDebugDraw`.
Events should never ever render anything on screen. It's not their job. It's the runtime scene job to do it, in a function called `renderAndStep`.
You have a function called `renderDebugDraw` in the renderer, which as the name implies should only be called **once** per scene to render on screen the debug graphics.
So when to do it? In the `renderAndStep`. At which condition? The boolean `this._enableDebugDraw` is true. But then what the action that you've been added should do? Just set this boolean to true or false (and if set to false while it was true before, it's fine to call `clearDebugDraw` or `destroyDebugDraw` to clear the objects. But don't call renderDebugDraw).
Finally, you have other booleans - it's fine! Just store them in the renderer as `this._xxx` members.

See the idea? :) 
So at the end you have three things:

A set of `enableDebugDraw(x, y, z)` in events/scene/renderer, which is JUST setting booleans.
A function `renderDebugDraw` which is only in the renderer, and only called by the scene. Its job: create any necessary object, render the stuff, and at the end do some cleaning.
A function `clearDebugDraw`, which can be called when `enaleDebugDraw` is set to false, to be sure we clean everything.

Otherwise we risk here having multiple times the debug draw made if the action is called multiple times (which some users will do, even if it does not make sense).

Bouh:
I understood the idea of `renderAndStep` and booleans in `enableDebugDraw` too.
But where is called `this.getRenderer().renderDebugDraw()` the explanations was a bit confusing, I did something let me know (It works)
Good news with these changes there is no lagging on the debug rendering :D

GDJS/Runtime/pixi-renderers/runtimescene-pixi-renderer.ts
```diff
@@ -120,47 +149,214 @@ namespace gdjs {
...
+                }
+              );
+            }
+
+            renderedObjectTextPoints.points['center'].position.set(
+              centerPointX,
+              centerPointY
+            );
+
+            this._pixiContainer.addChild(
```
4ian, line 230: addChild is being called at each render, which is useless and will impact performance. I'm fixing this.

GDJS/Runtime/pixi-renderers/runtimescene-pixi-renderer.ts
```diff
@@ -120,47 +149,214 @@ namespace gdjs {
...
+            }
+            renderedObjectTextPoints.points['origin'].position.set(
+              originPoint[0],
+              originPoint[1]
+            );
+
+            // Mark the text points as rendered
+            renderedObjectTextPoints.wasRendered = true;
+            this._pixiContainer.addChild(
+              renderedObjectTextPoints.points['origin']
```
4ian, line 271: You're doing multiple times the same thing, create a function for this instead.

GDJS/Runtime/pixi-renderers/runtimescene-pixi-renderer.ts
```diff
@@ -120,47 +149,214 @@ namespace gdjs {
...
+          }
+
+          // Draw custom point
+          if (showCustomPoints && object instanceof gdjs.SpriteRuntimeObject) {
+            if (!object._animationFrame) continue;
+            for (let customPointName in object._animationFrame.points[
+              'items'
+            ]) {
+              let customPoint: [float, float];
+              customPoint = object.getPointPosition(customPointName);
```
4ian, line 282: nit: `const customPoint = ..`. I'm fixing this.

GDJS/Runtime/pixi-renderers/runtimescene-pixi-renderer.ts
```diff
@@ -120,47 +149,214 @@ namespace gdjs {
...
+              debugDraw.drawCircle(
+                customPoint[0] - cameraTopLeftX,
+                customPoint[1] - cameraTopLeftY,
+                3
+              );
+            }
+          }
+
+          // After we've iterated on all objects, maybe some were destroyed. So clean their text points:
+          for (let objectID in this._debugDrawRenderedObjectsPoints) {
```
4ian, line 319: This loop is done inside the `for (let i = 0; i < instances.length; i++) {`, which means that for every instance you clean the points, which is a performance hit/unecessary complexity.

When a function is becoming too complex, it's easy to make mistakes like this. Question every indentation (do I really need this?) and break a function in smaller functions to help you manager in your head what a function is doing.
I'm fixing this.

4ian:
> Good news with these changes there is no lagging on the debug rendering :D

Yep, this is because the rendering is done at the same time as the rendering of the screen, so all objects are already in the proper positions 👍

I've done various fixes on my own, I'll commit them soon. I've seen that we're not using `layersCameraCoordinates`, and that the values inside are actually not correct, the coordinates are too wide. I'll fix that first because this is used for culling of objects (so it means we're not culling some objects, which could improve performance if we do it properly).

PR #2396, opened 2021-03-02
Platformer big refactor: state pattern

D8H:
The platformer extension had grown bigger with new features and I think it became hard to follow what it does. The main issue is the multiplication of conditions on the player state in doStepPreEvents(). It makes the reader jump long blocks of code and while doing so, he can easily loose his train of thought.

What I propose:
- Split doStepPreEvents() into Sub-functions to have the big picture of the main loop
- Use the State pattern to keep specific behaviors for a state together
- Use a State interface that enforce transition checks and state actions segregation
- Encapsulate internal state attributes (11 in total)

There were transition checks and state actions that have side effects on each other and sometimes multiples transitions could occur. I did a step by step refactor to move transitions at the beginning. There is an explanation in each commit message on behavior changes (slight or fixes).

The state pattern induces some indirections but less conditions too. I did a performance test (I'll make a PR with some other new test cases). There were no noticeable performance change, but typing give a 120:100 gain.
I understand this is a big risk and I can animate tests with creators after the review if you want.

### DIFF
--- a/Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
+++ b/Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
@@ -17,34 +17,28 @@ namespace gdjs {
     // platform hitbox left edge is also at X position 10).
     // This parameter "_ignoreTouchingEdges" will be passed to all collision handling functions.
     _ignoreTouchingEdges: boolean = true;
-    _roundCoordinates: any;
-    _gravity: any;
-    _maxFallingSpeed: any;
-    _ladderClimbingSpeed: any;
-    _acceleration: any;
-    _deceleration: any;
-    _maxSpeed: any;
-    _jumpSpeed: any;
-    _canGrabPlatforms: any;
+    _roundCoordinates: boolean;
+    _gravity: float;
+    _maxFallingSpeed: float;
+    _ladderClimbingSpeed: float;
+    _acceleration: float;
+    _deceleration: float;
+    _maxSpeed: float;
+    _jumpSpeed: float;
+    _canGrabPlatforms: boolean;
     _yGrabOffset: any;
     _xGrabTolerance: any;
     _jumpSustainTime: float;
-    _floorPlatform: any = null;
-    _currentFallSpeed: number = 0;
-    _currentSpeed: number = 0;
-    _currentJumpSpeed: number = 0;
-    _timeSinceCurrentJumpStart: number = 0;
-    _jumpKeyHeldSinceJumpStart: boolean = false;
+    _currentFallSpeed: float = 0;
+    _currentSpeed: float = 0;
     _canJump: boolean = false;
-    _grabbedPlatform: any = null;
-    _ignoreDefaultControls: any;
+    _ignoreDefaultControls: boolean;
     _leftKey: boolean = false;
     _rightKey: boolean = false;
     _ladderKey: boolean = false;
     _upKey: boolean = false;
     _downKey: boolean = false;
     _jumpKey: boolean = false;
-    _jumpingFirstDelta:boolean = false;
 
     _state: PlatformerObjectRuntimeBehavior.State;
     _falling: PlatformerObjectRuntimeBehavior.Falling;
@@ -59,18 +53,12 @@ namespace gdjs {
     /** Overlapped jump-thru platforms, updated with `_updateOverlappedJumpThru`. */
     _overlappedJumpThru: Array<gdjs.PlatformRuntimeBehavior>;
 
-    _oldHeight: float = 0;
-
     _hasReallyMoved: boolean = false;
-    _manager: any;
-    _grabbedPlatformLastX: any;
-    _grabbedPlatformLastY: any;
+    _manager: gdjs.PlatformObjectsManager;
 
-    _floorLastX: any;
-    _floorLastY: any;
-    _releaseKey: any;
+    _releaseKey: boolean = false;
     _slopeMaxAngle: float;
-    _slopeClimbingFactor: any;
+    _slopeClimbingFactor: float = 1;
 
     _requestedDeltaX: float = 0;
     _requestedDeltaY: float = 0;
@@ -186,7 +174,7 @@ namespace gdjs {
       const oldX = object.getX();
       if (this._requestedDeltaX !== 0) {
         const floorPlatformId =
-          this._floorPlatform !== null ? this._floorPlatform.owner.id : null;
+          this._onFloor._floorPlatform !== null ? this._onFloor._floorPlatform.owner.id : null;
         object.setX(object.getX() + this._requestedDeltaX);
         let tryRounding = true;
 
@@ -243,7 +231,10 @@ namespace gdjs {
 
     setState(state: PlatformerObjectRuntimeBehavior.State) {
       console.debug(`${this._state} --> ${state}`);
+      this._state.leave();
+      const from = this._state;
       this._state = state;
+      this._state.enter(from);
     }
 
     moveY() {
@@ -273,7 +264,6 @@ namespace gdjs {
           if (this._state == this._jumping) {
             this.setState(this._falling);
           }
-          this._currentJumpSpeed = 0;
           if (
             (this._requestedDeltaY > 0 && object.getY() <= oldY) ||
             (this._requestedDeltaY < 0 && object.getY() >= oldY)
@@ -294,8 +284,6 @@ namespace gdjs {
     checkTransitionOnLadder() {
       if (this._ladderKey && this._isOverlappingLadder()) {
         this._canJump = true;
-        this._floorPlatform = null;
-        this._currentJumpSpeed = 0;
         this._currentFallSpeed = 0;
         this.setState(this._onLadder);
       }
@@ -303,18 +291,7 @@ namespace gdjs {
 
     checkTransitionJumping() {
       if (this._canJump && this._jumpKey) {
-        if (this._state != this._jumping) {
-          this._jumpingFirstDelta = true;
-        }
         this.setState(this._jumping);
-        this._canJump = false;
-        this._timeSinceCurrentJumpStart = 0;
-        this._jumpKeyHeldSinceJumpStart = true;
-
-        //FIXME can't be Jumping and OnFloor at the same time, need a fix and a test
-        //this._isOnFloor = false; If floor is a very steep slope, the object could go into it.
-        this._currentJumpSpeed = this._jumpSpeed;
-        this._currentFallSpeed = 0;
       }
     }
 
@@ -356,7 +333,7 @@ namespace gdjs {
           )
         ) {
           this.setState(this._grabbingPlatform);
-          this._grabbedPlatform = collidingPlatform;
+          this._grabbingPlatform.setGrabbedPlatform(collidingPlatform);
           this._requestedDeltaY = 0;
         } else {
           object.setY(oldY);
@@ -374,13 +351,12 @@ namespace gdjs {
         this._state == this._onFloor &&
         gdjs.RuntimeObject.collisionTest(
           object,
-          this._floorPlatform.owner,
+          this._onFloor._floorPlatform!.owner,
           this._ignoreTouchingEdges
         )
       ) {
         //Still on the same floor
-        this._floorLastX = this._floorPlatform.owner.getX();
-        this._floorLastY = this._floorPlatform.owner.getY();
+        this._onFloor.updateFloorPosition();
 
         //Ensure nothing is grabbed.
         this._releaseGrabbedPlatform();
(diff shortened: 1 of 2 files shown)


Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
```diff
@@ -1,4 +1,4 @@
-describe('gdjs.PlatformerObjectRuntimeBehavior', function () {
+describe.only('gdjs.PlatformerObjectRuntimeBehavior', function () {
```
4ian, line 1: Remove this "only" :)

Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
```diff
@@ -157,10 +157,11 @@ describe('gdjs.PlatformerObjectRuntimeBehavior', function () {
       object.getBehavior('auto1').simulateReleaseKey();
       for (let i = 0; i < 10; ++i) {
         runtimeScene.renderAndStep(1000 / 60);
+        expect(object.getBehavior('auto1').isFalling()).to.be(true);
       }
 
       //Check that the object is falling
-      expect(object.getY()).to.be(1.25);
+      expect(object.getY()).to.be(3.75);
```
4ian, line 164: Any reason for this change? 🤔

D8H: Yes, blame it, I put an explanation in the commit message, so you can see the change in the code too.

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -1251,11 +1089,458 @@ namespace gdjs {
...
       return (
         (this._hasReallyMoved && this._currentSpeed !== 0) ||
-        this._currentJumpSpeed !== 0 ||
+        this._jumping.getCurrentJumpSpeed() !== 0 ||
         this._currentFallSpeed !== 0
       );
     }
   }
+
+  export namespace PlatformerObjectRuntimeBehavior {
```
4ian, line 1098: Is it possible to avoid exporting these interface/classes? There are internal logic and so should not be visible to the outside if possible.

D8H: I tried to write a private inner classes with this:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/class
but it would require to declare them before the behavior attributes declaration and it looks a bit messy.

4ian:
>  The main issue is the multiplication of conditions on the player state in doStepPreEvents(). It makes the reader jump long blocks of code and while doing so, he can easily loose his train of thought.

Yep, I agree with this.

> What I propose:

This make sense! Looks indeed like a great pattern. I would have been worried of some transitions as you mentioned that were not exclusive (you could be in 2 states at the same time)... but you seem to have solved it! :D 

> I did a performance test (I'll make a PR with some other new test cases). There were no noticeable performance change, but typing give a 120:100 gain.

Nice! I'm happy to see another PR with other test cases :D 
What do you mean exactly by typing giving a 120:100 gain? Adding typing made the behavior to be a bit quicker?

> I understand this is a big risk and I can animate tests with creators after the review if you want.

Happy to make a testing version of this so that people can try it and report any change in behavior :) 
For now I hope tests will be sufficient to cover most/all cases.

I'll go and reach each commit to see your changes. Thanks a lot for doing this, from a quick glance the code look very high quality :)

Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
4ian: Ok I see, the grabbing look behind looks ok, and the release key making the object fall at the same frame too, same for the ladder 👍

D8H:
I rebase the branch so new tests pass on each commit.

D8H:
One fix wasn't back-ported in "Move some attributes to the states", it only impacts this commit. I checked most commits, they should all pass the tests now.

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -600,62 +307,202 @@ namespace gdjs {
...
+    }
+
+    checkTransitionOnFloorOrFalling() {
+      const object = this.owner;
+      //Check if the object is on a floor:
+      //In priority, check if the last floor platform is still the floor.
+      let oldY = object.getY();
+      object.setY(object.getY() + 1);
+      if (
+        this._state == this._onFloor &&
```
4ian, line 381: Always use === in JS. == is doing type coercion and is slower and less safe. === is always the way to go :) 
(check this for all the `this._state ==` in the file)

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -600,62 +307,202 @@ namespace gdjs {
...
-          }
-        }
-        object.setY(oldY);
+      const oldY = object.getY();
+      this.moveY();
+
+      //3) Update the current floor data for the next tick:
+      this._updateOverlappedJumpThru();
+      //TODO what about a moving platforms, remove this condition to do the same as for grabbing?
+      if (this._state != this._onLadder) {
```
4ian, line 504: (note, same thing for !=, always always use `!==`)

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -862,7 +709,8 @@ namespace gdjs {
      * Return true if the object is overlapping a ladder.
      * Note: _updatePotentialCollidingObjects must have been called before.
      */
-    private _isOverlappingLadder() {
+    _isOverlappingLadder() {
+      //FIXME put back private
```
4ian, line 713: Can you put back private?

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -1251,11 +1094,506 @@ namespace gdjs {
...
+          )
+        ) {
+          behavior._releaseGrabbedPlatform();
+        }
+      }
+
+      beforeMovingX() {
+        const behavior = this._behavior;
+        //Shift the object according to the grabbed platform movement.
+        // this._behavior erases any other movement
```
4ian, line 1513: I would rather say "when grabbing a platform, no other movement will be attempted by the behavior"?

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -600,62 +307,202 @@ namespace gdjs {
...
+
+      //0.1) Get the player input:
+      this._requestedDeltaX = 0;
+      this._requestedDeltaY = 0;
+
+      const inputManager = runtimeScene.getGame().getInputManager();
+      this._leftKey ||
+        (this._leftKey =
+          !this._ignoreDefaultControls && inputManager.isKeyPressed(LEFTKEY));
+      // @ts-ignore
```
4ian, line 440: I think you can remove this ts-ignore

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -600,62 +307,202 @@ namespace gdjs {
...
-            this._floorPlatform = null;
-          }
-        }
-        object.setY(oldY);
+      const oldY = object.getY();
+      this.moveY();
+
+      //3) Update the current floor data for the next tick:
+      this._updateOverlappedJumpThru();
+      //TODO what about a moving platforms, remove this condition to do the same as for grabbing?
```
4ian, line 503: Yes probably! This means that for now a moving ladder will create issues - though I'm not sure exactly why.
I reckon we should leave it here for this PR, but do a PR with it removed and a few tests that verify how it works (probably we the character should follow the ladder, be pushed by any potential colliding obstacle and start falling if not on the ladder anymore).

4ian (approved):
Overall this looks ex-ce-llent! 😄👏
The approach using different states is indeed much more readable and easier to reason about. 

I've added a few nitpicking comments. Once they are solved, I'm happy to merge this, but maybe after a release (that I hope to do soon, this week-end for sure), so that we minimize the risks of having issues in the same release (the new release is bringing arrays and booleans variables, I would like to get some confirmation that everything is fine with this).

Thanks a lot for all the work on this, this is very high quality contribution and will help a lot any future improvement of the platformer engine 👏

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
D8H: I did a bad find and replace. It was: "this erases any other movement" 
I think it's to explain why += is not used.

D8H:
There are some relative type references. Should the absolute path (gdjs.) always be used?

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -8,6 +8,460 @@ namespace gdjs {
    * considered as a platform by objects having PlatformerObject Behavior.
    */
   export class PlatformerObjectRuntimeBehavior extends gdjs.RuntimeBehavior {
+    /**
+     * The object is on the floor standing or walking.
+     */
+    OnFloor = class implements gdjs.PlatformerObjectRuntimeBehavior.State {
```
4ian, line 14: I missed that you changed this to put "class expressions" inside the behavior class.
I would highly recommend against this:
* This makes the class appear bloated.
* This force the JS engine to make every single instance of the class to have its own class.

The second part can be seen like this:
image

Here `A.test` is not accessible because `test` is a member of the class, so you have to instantiate a instance of A... and each instance of A will get its own class!
This means if you have 1500 objects using the platform behavior, there will be 1500 OnFloor classes in memory.

Instead, you can declare classes but not export them:

```
namespace gdjs {
  class OnFloor implements .... { ... }
  class Falling implements .... { ... }

  export class PlatformerObjectRuntimeBehavior { ... }
}
```

In other words, you can declare any class or variable inside namespace gdjs { } and it will be scoped to the file - which is great, nobody will be ever able to access it. Then when you use the keyword `export` on a variable or a class, it will become accessible under `gdjs.YourClass` (`or gdjs.yourVariable`).

4ian (changes requested):
Unfortunately I missed that you used class as expressions inside a class, which sounds risky in terms of optimization for JS engine.
Instead, keep things simple and stupid: declare plain old classes inside the namespace, then only add `export` to the behavior :) 

Sorry for missing this, I think the changes were made after my last review.

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
4ian: Sorry I was not clear: let's not put the classes inside the behavior as member, let's just declare them without an `export` keyword and it should work.
If we must move them before the declaring of the behavior class that's fine - as long as 1) it's not inside the behavior class and 2) they are not `export`ed, i.e: you can't do `gdjs.PlatformerObjectRuntimeBehavior.OnFloor` from the outside. It's fine if you refer as them as `OnFloor` inside the extension itself, but if you have to use something like `this.OnFloor`, it's a no go as it means you'll have one class per instance, which will be a nightmare for JS engines :)

4ian:
If it's too confusing let me know and I'll do the changes by myself :)

D8H:
Sorry, I force push before I see your change request and I can't access it now.
I did the changes, is it what you expected?

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
D8H: > (note that you could make these `static`, but at this point it's not better than declaring the state classes outside of the behavior class, as it still make the behavior class appear bloated :))

It seems a good option because it allows to keep private what should be. With the class without export, there is a lot of members that can't be private but should.
I have both version in local, what do you prefer?

4ian (approved):
Alright, looks good to go! I'll remove the console.debug unless you do it before me.

4ian:
> I have both version in local, what do you prefer?

The version in this PR looks fine now (i.e: a few `class`, that are not exported) 👍

D8H:
The fact that half the attributes and methods that should be private are not is ok? It's because it doesn't matter when translated in JS?

4ian:
> The fact that half the attributes and methods that should be private are not is ok? Is it because it doesn't matter when translated in JS?

Exactly. We used - and still use - the convention of a trailing underscore to denotate something that is private. The `private` keyword in typescript is useful, but do not matter when translated to JS.

In a perfect world, we would have private everywhere, but as it's not used in JS, I'm tolerating some inconsistencies... the underscore is actually more important, to visually tell that something is private :)

4ian:
Looks good! I've tested and made a change to isFalling so that it returns true when the object is in state "falling" OR when the jump speed is less than the fall speed - meaning that the object is technically still doing a jump but is going down (i.e: the peak of the jump is reached already).

Otherwise you could see the backward compatibility problem very easily: just run the platformer example, and it was very hard/impossible to beat enemies, because the "Object is falling" condition was only true when you were falling but not when going down while doing a jump.

Ideally we would have used another name than "IsFalling", but that's how it is and it kind of make sense.

I've tested on a larger game and everything seems alright after this change. 
Merging now! Thanks a lot for this huuuuge refactoring. This will definitely help to keep a sane codebase for this behavior and all the tests will help maintaining this. Hugely appreciated 👍👍

4ian:
@D8H Seems like we have a regression, in the platformer example when the object is falling we can jump in the middle of the fall (up to one time). 
This seems incorrect (jumping while you're in the air can be allowed by using the events to "allow to jump again" - but this is not something done by default).

Would you be able to take a look at this?

D8H:
Yes, I can take a look this weekend.

4ian:
Might be a missing `_canJump = false` somewhere when transitioning to a falling state, let me know if you have the time to look at it :)

D8H:
Yes, this is it. The _canJump was not set in the enter() of Falling. Other States are ok.

D8H:
I did a PR for the unit and fix: https://github.com/4ian/GDevelop/pull/2477

PR #2401, opened 2021-03-03
Improve overall GDJS typing

arthuro555:
I might have done too much at once this time 😅 

I added a lot of `: void` to method signatures, as leaving them out implies a return type of `: any`, which might make typescript not spot some bugs if a function is not supposed to return a value.
I also replaced a lot of `{ [key: string]: Type }` with `Record<string, Type>`. There are no changes in the type itself as `Record` is an alias for this but it is way more readable.
I added return types to functions that could already infer it from what is returned as well, to be able to spot bugs if the variable the type is inferred from changes types.

The "major" type changes are in the Profiler and in the Debugger.

### DIFF
--- a/GDJS/Runtime/gd.ts
+++ b/GDJS/Runtime/gd.ts
@@ -72,7 +72,9 @@ namespace gdjs {
    * Convert a hex color value to an rgb object.
    * @param hex Hex color
    */
-  export const hexNumberToRGB = (hexNumber: number) => {
+  export const hexNumberToRGB = (
+    hexNumber: number
+  ): { r: integer; g: integer; b: integer; a: integer } => {
     return {
       r: (hexNumber >> 16) & 0xff,
       g: (hexNumber >> 8) & 0xff,
@@ -164,7 +166,7 @@ namespace gdjs {
   export const registerObject = function (
     objectTypeName: string,
     Ctor: typeof gdjs.RuntimeObject
-  ) {
+  ): void {
     gdjs.objectsTypes.put(objectTypeName, Ctor);
   };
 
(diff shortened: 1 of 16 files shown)


GDJS/Runtime/gd.ts
```diff
@@ -400,22 +402,27 @@ namespace gdjs {
...
+    (): string;
+    hex?: string[];
+  }
+
   /**
    * Generate a UUID v4.
    * @returns The generated UUID.
    */
-  export const makeUuid = function (): string {
+  export const makeUuid = <MakeUUID>function (): string {
```
4ian, line 422: Very nice I did not know you could type a function like this.

GDJS/Runtime/oncetriggers.ts
```diff
@@ -9,15 +9,20 @@ namespace gdjs {
    * that are used in events to have conditions that are only valid for one frame in a row.
    */
   export class OnceTriggers {
-    _onceTriggers: any = {};
-    _lastFrameOnceTrigger: any = {};
+    _onceTriggers: Record<number, boolean> = {};
```
4ian, line 12: nitpicking: that could even be `integer`.

GDJS/Runtime/runtimeobject.ts
```diff
@@ -7,9 +7,9 @@ namespace gdjs {
   /** An axis-aligned bounding box. Used to represents a box around an object for example. */
   export type AABB = {
     /** The [x,y] coordinates of the top left point */
-    min: Array<number>;
+    min: FloatPoint;
```
4ian, line 10: Nice 👍

4ian (approved):
Thanks! Overall I think it's very good, I've not spotted any risk of regression.

Issue #2225, opened 2021-01-17
Add missing actions for Particle Emitter properties


tristanbob:
## Description

Particle emitters are missing actions for some properties.

### New Actions:

- [ ] Particles kind (Point/Line/Textured)  (I recommend calling this "Particle type")
- [ ] Lines length (number)
- [ ] Lines thickness (number)
- [ ] Additive rendering (bool)
- [ ] Delete when out of particles (bool)
- [ ] Maximum number of particles displayed (number)
- [ ] Friction on particles (number)
- [ ] Particle minimum rotation speed (degrees/second)
- [ ] Particle maximum rotation speed (degrees/second)

### Mismatched:

Gravity actions and properties do not appear to be referring to the same thing.

  - Actions
    - Gravity 
    - Gravity Angle
  - Properties
    - Gravity on particles on X axis
    - Gravity on particles on Y axis

![ParticleEmitterCommon](https://user-images.githubusercontent.com/8879811/104856401-bef2f480-58cf-11eb-942a-26af915bccb7.PNG)

![ParticleEmitterAdvanced](https://user-images.githubusercontent.com/8879811/104856404-c1554e80-58cf-11eb-8d48-d12931b61022.PNG)

![Capture](https://user-images.githubusercontent.com/8879811/104856408-c4503f00-58cf-11eb-8b9c-54408cc045e7.PNG)

tristanbob:
FYI - I need these actions so that I can build a "Particle Emitter Builder" that lets you see how the parameters affect the look, in real-time.   

Once we have a particle emitter builder, we can create a contest for people to build new emitters!

Oxey405:
Okay it's seems a little hard to implement for a noob like me but still can try ^^

Oxey405:
Oh my gosh what a hell to get inside this part of the code !! I can't understand !

4ian:
@Oxey405 Please use the Discord/forum to discuss or share your comments, GitHub is only for constructed collaboration with clear question/answers. Thanks!

Oxey405:
Ok... I give up ! I can't understand !

85ayush:
I would love to work on this issue, I am new to open source contribution, can you please guide me?

rs4231199:
I am working on this.

PR #2404, opened 2021-03-04
Adding some Particle Emitter properties

rs4231199:
It is a rough implementation of the first action(Particles kind) requested in #2225.

Here is how I tested it. I created a particle emitter object and added this condition to the event sheet.
image
It works as expected in the electron app, but the web version shows a black screen. 

I am unable to understand the reason as I didn't make any platform-specific. I need help regarding how to go on from here.

### DIFF
--- a/Extensions/ParticleSystem/Extension.cpp
+++ b/Extensions/ParticleSystem/Extension.cpp
@@ -102,6 +102,9 @@ ParticleSystemCppExtension::ParticleSystemCppExtension() {
   conditions["ParticleSystem::EmitterAngle"]
       .SetFunctionName("GetAngle")
       .SetIncludeFile("ParticleSystem/ParticleEmitterObject.h");
+  conditions["ParticleSystem::ParticleType"]
+      .SetFunctionName("GetRendererType")
+      .SetIncludeFile("ParticleSystem/ParticleEmitterObject.h");
   actions["ParticleSystem::EmitterAngleA"]
       .SetFunctionName("SetEmitterAngleA")
       .SetGetter("GetEmitterAngleA")
--- a/Extensions/ParticleSystem/ExtensionSubDeclaration1.cpp
+++ b/Extensions/ParticleSystem/ExtensionSubDeclaration1.cpp
@@ -122,6 +122,16 @@ void ExtensionSubDeclaration1(gd::ObjectMetadata& obj) {
       .AddParameter("object", _("Object"), "ParticleEmitter")
       .UseStandardRelationalOperatorParameters("number");
 
+  obj.AddCondition("ParticleType",
+                  _("Kind of Particle"),
+                  _("Test the Kind of Particle(Point: 0, Line: 1, or Quad: 2)"),
+                  _("the Particle Kind"),
+                  _("Common"),
+                  "CppPlatform/Extensions/particleSystemicon24.png",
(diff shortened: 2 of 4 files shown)


4ian:
Don't worry about the web-app, the runtime and extensions are hosted online so they won't update with your changes. What you did is already fine :)

Extensions/ParticleSystem/Extension.cpp
```diff
@@ -102,6 +102,9 @@ ParticleSystemCppExtension::ParticleSystemCppExtension() {
   conditions["ParticleSystem::EmitterAngle"]
       .SetFunctionName("GetAngle")
       .SetIncludeFile("ParticleSystem/ParticleEmitterObject.h");
+  conditions["ParticleSystem::ParticleType"]
+      .SetFunctionName("GetRendererType")
+      .SetIncludeFile("ParticleSystem/ParticleEmitterObject.h");
```
4ian, line 107: Nice to see that you implemented it for the C++ version, but as it's getting deprecated I suggest we don't add this and keep the implementation only for the JS engine :)

Extensions/ParticleSystem/ExtensionSubDeclaration1.cpp
```diff
@@ -122,6 +122,16 @@ void ExtensionSubDeclaration1(gd::ObjectMetadata& obj) {
       .AddParameter("object", _("Object"), "ParticleEmitter")
       .UseStandardRelationalOperatorParameters("number");
 
+  obj.AddCondition("ParticleType",
```
4ian, line 125: I suggest we follow the term in the UI: `ParticleKind`.

I also suggest that instead of returning a number, we return a string ("Point", "Line" or "Quad") to be a bit more user friendly.

Finally, take a look at `AddExpressionAndCondition` that will allow you to create both a condition and an expression for it :) 
`obj.AddExpressionAndCondition("string", "ParticleKind", _("Kind of particle"), _("the kind of particle"), ...)`

It's fairly new but useful to reduce the number of declarations to do, and declare expressions.

Extensions/ParticleSystem/JsExtension.cpp
```diff
@@ -51,6 +51,7 @@ class ParticleSystemJsExtension : public gd::PlatformExtension {
         .SetFunctionName("setAngle")
         .SetGetter("getAngle");
     conditions["ParticleSystem::EmitterAngle"].SetFunctionName("getAngle");
+    conditions["ParticleSystem::ParticleType"].SetFunctionName("getRendererType");
```
4ian, line 54: Note that if you follow my advice of using `AddExpressionAndCondition` you will also need a line for the expression:
`strExpressions["ParticleKind"].SetFunctionName("getRendererType");`

Extensions/ParticleSystem/particleemitterobject.ts
```diff
@@ -403,6 +403,10 @@ namespace gdjs {
       return (this.angleA + this.angleB) / 2.0;
     }
 
+    getRendererType(): float {
+      return parseInt(this.rendererType, 10) || 0;
```
4ian, line 407: Did you check what's inside rendererType? It's a string which is "Line", "Point" or "Quad" so I'm unsure this was working at all?

rs4231199:
I apologize for keeping this PR open for so long. I will start the work again in few days.

PR #2411, opened 2021-03-06
Add the movementAngleOffset property to the top-down behavior

D8H:
It allows to choose the movement main axises for the isometry.
I'm very sorry I've mistaken offsetAngle to do so. I should have added this from the beginning.

When switching between top-down and isometry, the properties is shifted from 45°. So if a creator let the default 0 value and choose isometry, it will automagically be -45. It will always move from 45° even if the value is not a default one. It may be annoying for a use who already set the offset, but I though it would be easier to understand.

I tested it with an empty project and ran it with top-down, isometry and auto-rotation to check it behave as expected.


4ian:
> It will always move from 45° even if the value is not a default one. It may be annoying for a use who already set the offset, but I though it would be easier to understand.

I was thinking about this, I think it's fair to have this behavior! Much better for beginners, and for someone experienced that's also probably what you want - and anyway you don't usually change between top down and isometry, only once to set up the behavior. 

> I tested it with an empty project and ran it with top-down, isometry and auto-rotation to check it behave as expected.

Thanks! Sounds perfect 👌

Extensions/TopDownMovementBehavior/topdownmovementruntimebehavior.ts
```diff
@@ -273,7 +286,7 @@ namespace gdjs {
 
       //Update the speed of the object
       if (direction != -1) {
-        directionInRad = (direction * Math.PI) / 4.0;
+        directionInRad = ((direction + this._movementAngleOffset / 45) * Math.PI) / 4.0;
         directionInDeg = direction * 45;
```
4ian, line 290: Do you think this should be
```suggestion
        directionInDeg = direction * 45 + this._movementAngleOffset;
```

?

Extensions/TopDownMovementBehavior/JsExtension.cpp
```diff
@@ -74,6 +74,11 @@ class TopDownMovementBehaviorJsExtension : public gd::PlatformExtension {
         "getXVelocity");
     autConditions["TopDownMovementBehavior::YVelocity"].SetFunctionName(
         "getYVelocity");
+    autActions["TopDownMovementBehavior::MovementAngleOffset"]
+        .SetFunctionName("setMovementAngleOffset")
+        .SetGetter("getMovementAngleOffset");
+    autConditions["TopDownMovementBehavior::MovementAngleOffset"].SetFunctionName(
+        "getMovementAngleOffset");
```
4ian, line 81: Did you intend to do one action/condition for it? I don't see a declaration for it.
I recommend also adding an expression :)

You can do all of them in one since recently thanks to `AddExpressionAndConditionAndAction`:
```
behavior.AddExpressionAndConditionAndAction("number", "MovementAngleOffset", _("Movement angle offset"), _("the movement angle offset"), _("the movement angle offset"), _("Movement"), "CppPlatform/Extensions/topdownmovementicon24.png")
      .AddParameter("object", _("Object"))
      .AddParameter("behavior", _("Behavior"), "TopDownMovementBehavior")
      .UseStandardParameters("number")
```

Note that this will add a "MovementAngleOffset" expression (as expected), a condition called "TopDownMovementBehavior::MovementAngleOffset" (as you did), and an action called ""TopDownMovementBehavior::SetMovementAngleOffset" (there is a "Set" that is added, which should be the norm for actions).

If you don't get this to work, let me know and I can do it myself.

Extensions/TopDownMovementBehavior/topdownmovementruntimebehavior.ts
4ian: directionInDeg is used to rotate the object and remember the angle of the latest movement. 
I think in both case they should be the real angle that was moved, like directionInRad. So we should add the movementAngleOffset.

(`this._angleOffset` is still useful to add an offset to the object (useful if your object is not orientated properly), but this is only for rendering, we should have both)

Extensions/TopDownMovementBehavior/JsExtension.cpp
D8H: Yes, I couldn't make it work. It doesn't compile and I didn't find any example in other extensions.

4ian: There was a missing include file, I've fixed this so it works for all extensions in the future.

4ian:
I've fixed compilation, added some descriptions and fixed the internal action name (SetXxx).
Merging now, thanks for this fix!

PR #2412, opened 2021-03-06
Improve compilation speed by 8%

4ian:
* Avoid useless copy of strings in memory when comparing them (reducing work in string heavy operations like export/preview).


Oxey405:
8% ? This is pretty awesome !

Issue #2236, opened 2021-01-24
Unintended Behaviour: Holding jump whilst on ladder keeps jump speed at max


AtomicCrocStudios:
## Describe the bug
I feel like this unintended behaviour which can be replicated in the platformer example or the attached file (has default keys disabled and a debug showing jump speed).

When on a ladder, pressing up and jump at the same time you zoom to the top of the ladder maintaining maximum jump speed as if you have infinite double jumps.
![https://i.gyazo.com/f1c8e7b4dbd88d4af22884d89aeadaed.gif](https://i.gyazo.com/f1c8e7b4dbd88d4af22884d89aeadaed.gif)

I think that whilst on the ladder, jumping should work like normal however when at the apex of the jump (before the player starts falling) their jump speed is set to 0 so they stop on the ladder (if pressing up and jump are still active, it starts jumping again)
See the jumping platforms next to the ladder as an example.
![https://i.gyazo.com/1c4af835eb82fbe0de5723bc525678f1.gif](https://i.gyazo.com/1c4af835eb82fbe0de5723bc525678f1.gif)

https://drive.google.com/file/d/1ulqCosy7C1MiK2xbJhPJ862qlYcpctAj/view?usp=sharing

## To Reproduce
1. Stand at ladder
2. Press UP + Jump
3. Reach the stars!

## Other details
* Tried in Beta 103, 102 and 101
* Windows 10

4ian:
It's true that it's a problem right now! I'll think about how to improve this.

arthuro555:
One possibility would be in the default controls to disallow climbing the ladder while pressing the jump key/while the momentum is going up

4ian:
> disallow climbing the ladder while pressing the jump key

It's possible this way, but then a player could spam the jump key and the ladder key.

> while the momenting is going up

I was also thinking about this, but this means that if you try to reach a ladder that is above you by jumping on it, you might not be able to grab it.

Note that these two things can already be done manually with events :) (disable default controls, and use the actions to simulate key press, only allowing a jump if you're also *not* pressing the button to climb a ladder).

We could maybe add a "time before being able to grab again a ladder after leaving one": 
* this would allow to avoid the repeated jump/climb, because you could certainly jump from a ladder, but you would have to wait something like ~0.3 seconds before being able to grab it again and jump again.
[...]

PR #2415, opened 2021-03-07
Add more tests on the platformer extension

D8H:
Add some tests on:
- jumping to grab platforms or ladders
- walking on platforms
- performance

Some tests can be useful for the skyrocket ladder issue: https://github.com/4ian/GDevelop/issues/2236
I put a comment and a drawing to explain how I think it could be fixed.

### DIFF
--- a/Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
+++ b/Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
@@ -1165,7 +1165,7 @@ describe('gdjs.PlatformerObjectRuntimeBehavior', function () {
       expect(object.getX()).to.be.above(ladder2.getX());
       // and grab the 2nd one, even if still ascending
       object.getBehavior('auto1').simulateLadderKey();
-      // still moves a little because of inertia maybe?
+      // still moves a little because of inertia
       runtimeScene.renderAndStep(1000 / 60);
       expect(object.getBehavior('auto1').isOnLadder()).to.be(true);
       runtimeScene.renderAndStep(1000 / 60);
--- a/Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
+++ b/Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
@@ -1015,8 +1015,7 @@ describe('gdjs.PlatformerObjectRuntimeBehavior', function () {
         const lastY = object.getY();
         runtimeScene.renderAndStep(1000 / 60);
         expect(object.getBehavior('auto1').isOnLadder()).to.be(true);
-        //TODO Probably a bug, uncomment it after it's fixed
-        //expect(object.getBehavior('auto1').isMoving()).to.be(false);
+        expect(object.getBehavior('auto1').isMoving()).to.be(false);
         expect(object.getY()).to.be(lastY);
       }
     };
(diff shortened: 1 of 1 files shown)


Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
```diff
@@ -930,10 +1015,31 @@ describe('gdjs.PlatformerObjectRuntimeBehavior', function () {
         const lastY = object.getY();
         runtimeScene.renderAndStep(1000 / 60);
         expect(object.getBehavior('auto1').isOnLadder()).to.be(true);
+        //TODO Probably a bug, uncomment it after it's fixed
+        //expect(object.getBehavior('auto1').isMoving()).to.be(false);
```
4ian, line 1019: You mean `isMoving()` is `true` when you're on a ladder and not doing anything? Sounds weird.

Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
```diff
@@ -996,6 +1102,108 @@ describe('gdjs.PlatformerObjectRuntimeBehavior', function () {
...
+      object.getBehavior('auto1').simulateLadderKey();
+      stayOnLadder(1);
+      expect(object.getBehavior('auto1').isJumping()).to.be(false);
+
+      stayOnLadder(10);
+      climbLadder(2);
+    });
+
+    //TODO Probably a bug, remove the skip when it's fixed
+    it.skip("can't grab a ladder while on the ascending phase of a jump", function () {
```
4ian, line 1132: Well it's not a bug in the current situation right, you should be able to grab a ladder while in the ascending phase of a jump without any problem?

Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
```diff
@@ -1135,4 +1343,232 @@ describe('gdjs.PlatformerObjectRuntimeBehavior', function () {
...
+
+    const walkRight = (frameCount) => {
+      for (let i = 0; i < frameCount; ++i) {
+        const lastX = object.getX();
+        object.getBehavior('auto1').simulateRightKey();
+        runtimeScene.renderAndStep(1000 / 60);
+        expect(object.getBehavior('auto1').isOnFloor()).to.be(true);
+        expect(object.getX()).to.be.above(lastX);
+        //TODO probably a bug, uncomment when it's fixed.
+        //expect(object.getBehavior('auto1').isMoving()).to.be(true);
```
4ian, line 1408: isMoving is true when you:
- fall
- jump 
- or your speed of your horizontal movement is not 0 *and* you've moved from at least 1 pixel.

The last thing may make the "isMoving" to be false if you've not "moved enough".

Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
```diff
@@ -1135,4 +1343,232 @@ describe('gdjs.PlatformerObjectRuntimeBehavior', function () {
...
+      platform2.setPosition(platform.getX() + platform.getWidth(), platform.getY());
+
+      object.setPosition(30, -32);
+      // Ensure the object falls on the platform
+      fallOnPlatform(10);
+
+      // walk from the 1st platform to the 2nd one
+      walkRight(30);
+      expect(object.getX()).to.be.above(platform2.getX());
+    });
```
4ian, line 1434: maybe check if it's on floor too?

Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
4ian: I mean, this kind of things allow to do the infinite "jump+ladder" trick, we'll fix this with a new check that is preventing to grab ladder either too soon after a jump or in certain conditions, but I think it's fine to not consider this a bug for now :)

Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
D8H: walkRight do the checks too. The name doesn't reflect it, but I think it make the test easier to read to only have a description of what the player do.

Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
D8H: It was just the inertia of the right key in a test.

Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
D8H: I removed it, I'm keeping it under my hat if it can help to write a test when someone find a solution.

D8H:
I did the changes, I think it may be ready.

4ian:
Nice! Can you run `npm run format` in GDJS?

Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
4ian: @D8H what do you think of checking that too?

4ian (approved):

Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
D8H: I added a check on the Y position. Was it what you meant?

4ian: I was thinking of checking if `isOnFloor` was true, but maybe not necessary!

D8H: Ok, walkRight() does it.

4ian: Perfect then, thanks!

PR #2416, opened 2021-03-07
Scene Editor - Moving instance selected faster with SHIFT

### LINKED ISSUE #2266: Scene Editor - Moving selected objects to different distances
## Description

Now, using the arrows, you can move the selected object and position it precisely.

## Solution suggested

Move the object to a greater distance (_5 pixels_) while holding down the _Ctrl_ key.

Bouh:
As the keyboard shortcuts used are still on the DeprecatedKeyboardShortcuts  class i've made the change on the Deprecated versions used and the new not yet used.

We can't yet use fully the new KeyboardShortcuts class because some fonctions are missing and make crash the app.

fix  #2266

### DIFF
--- a/newIDE/app/src/UI/KeyboardShortcuts/DeprecatedKeyboardShortcuts.js
+++ b/newIDE/app/src/UI/KeyboardShortcuts/DeprecatedKeyboardShortcuts.js
@@ -123,13 +123,13 @@ export default class DeprecatedKeyboardShortcuts {
 
     if (this.onMove) {
       if (evt.which === UP_KEY) {
-        this.onMove(0, -1);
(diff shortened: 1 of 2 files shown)


4ian:
Nice :) Can you tell more about the rationale behind Shift, is this something you've seen in some other editors? Or something that seems to make sense (as Shift is only used for multi selection)?
In particular, what do you think about Alt?

Bouh:
I'am based on photoshop, SHIFT move by +10, ALT use the current selection for draw, CTRL do nothing

4ian:
If photoshop does it, sounds good then!

Bouh:
Inskape too, and another software similar too :+1:

PR #2418, opened 2021-03-08
Fix object picking in events based extensions

### LINKED ISSUE #2414: ForEach() does not include objects that were created the same frame (when using an Extension function)
## Describe the bug
ForEach() does not include objects that were created the same frame (when using an Extension function)

## To Reproduce
See project file.  Includes a test using normal events (no problems) and a test using an extension function (causes problem).

## Live Game
https://games.gdevelop-app.com/game-959d24dc-50bd-4d65-9fdd-47e0acee60c4/index.html

## Project files
[ChainOfObjects-BugTest.zip](https://github.com/4ian/GDevelop/files/6097829/ChainOfObjects-BugTest.zip)

## Video
https://user-images.githubusercontent.com/8879811/110252697-e6675480-7f43-11eb-88da-0014d8e1c1e8.mp4

arthuro555:
Fixes #2414 
Adds objects created in events functions extensions to the events function objects list.

### DIFF
--- a/GDJS/GDJS/Events/CodeGeneration/EventsCodeGenerator.cpp
+++ b/GDJS/GDJS/Events/CodeGeneration/EventsCodeGenerator.cpp
@@ -328,11 +328,15 @@ gd::String EventsCodeGenerator::GenerateEventsFunctionContext(
          // the cost of creating/storing it for each events function might not
          // be worth it.
          "    if (objectsList) {\n" +
-         "      return parentEventsFunctionContext ?\n" +
+         "      const object = parentEventsFunctionContext ?\n" +
          "        "
          "parentEventsFunctionContext.createObject(objectsList.firstKey()) "
          ":\n" +
          "        runtimeScene.createObject(objectsList.firstKey());\n" +
+         // Add the new instance to object lists
+         "      objectsList.get(objectsList.firstKey()).push(object);"
+         "      eventsFunctionContext._objectArraysMap[objectName].push(object);" +
+         "      return object;" +
          "    }\n" +
          // Unknown object, don't create anything:
          "    return null;\n" +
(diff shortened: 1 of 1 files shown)


4ian:
Nice! I guess you've checked the test game works properly now?

I wonder if it means that objects created in the functions are now accessible from actions after the function is called (like: 
```
conditions: Beginning of the scene
actions: 
  Create object Blabla and do something with them <--- this is the function
  Rotate Blabla by X degrees <---- Is this rotating all Blabla created, or only the one present on the scene already?
```

I'm working to improve the tests in GDevelop.js to be able to test this kind of code generation/object creation :)

GDJS/GDJS/Events/CodeGeneration/EventsCodeGenerator.cpp
```diff
@@ -328,11 +328,15 @@ gd::String EventsCodeGenerator::GenerateEventsFunctionContext(
...
          // be worth it.
          "    if (objectsList) {\n" +
-         "      return parentEventsFunctionContext ?\n" +
+         "      const object = parentEventsFunctionContext ?\n" +
          "        "
          "parentEventsFunctionContext.createObject(objectsList.firstKey()) "
          ":\n" +
          "        runtimeScene.createObject(objectsList.firstKey());\n" +
+         // Add the new instance to object lists
+         "      objectsList.get(objectsList.firstKey()).push(object);"
```
4ian, line 337: Might be a good idea to add a `if (object) { } ` around these two lines, there is a possibility the returned object is null if the name is invalid?

arthuro555:
> I guess you've checked the test game works properly now?

I did, it looks like it is fixed.

4ian:
Great, looks good! 
I've made some integration tests checking that the generated code of a function with a Create action works properly: https://github.com/4ian/GDevelop/commit/b78022a8911a8acdf22e118b392f145e5555d91d

Should avoid any regression in the future for these kind of "not so obvious" object picking logics associated with the creation action.

PR #2421, opened 2021-03-09
Fix alignment in variables editor

Bouh:
Don't show in changelog

image

### DIFF
--- a/newIDE/app/src/VariablesList/VariableRow.js
+++ b/newIDE/app/src/VariablesList/VariableRow.js
@@ -256,7 +256,7 @@ const VariableRow = ({
 
   return (
     <>
-      <TreeTableRow alignItems="flex-start">{columns}</TreeTableRow>
+      <TreeTableRow>{columns}</TreeTableRow>
       {children}
     </>
   );


4ian:
The idea of this alignment was to keep the variable name at the top when there are multiple lines:
image

I don't have strong feelings on this, but what do you think?
We could also maybe reduce the margins (and size?) for the true/false buttons

4ian:
Try to remove the margin, set to 5 currently, on the True/False button and add noMargin to the Line having them, and this should give something like:
image

Space taken by the line is now not higher than a number/string :)

Bouh:
True/False are too near of the value under it.
Margin left and right on True/False can be keeped, we are not on a switch component. even if we have one choice possible.

The idea is to ventilate a little the inputs, everything is glued too much and not aligned.

4ian:
We can give a bit more margins to the text fields then? To have every line (apart from the multiline texts) with the same height?

Bouh:
noMargin on the line wrapping the booleans is better, but the issue persist the name of the variable are not aligned on the one line value or booleans. same for icons on right.

4ian:
Alright then, I think this could be solved by margins around the text fields (add a `Line` or something like this), but I'm ok going back to properly centered rows for now

PR #2425, opened 2021-03-10
[Feature] Mathematical expression in input fields.

Bouh:
There is some questions without response,the code is experimental.
You can run GDevelop and try.
Change are active in instance properties panel.

# Feature
Permit the user to write mathematical expressions in inputs.

`eval()` is used, but before i check the value of the input with a regex for check if the value is a valid mathematical expression.
By valid i mean, use only digits, and operator like /*-+.,

A days ago i've added Shift option on inputs for increase the values, but as i've changed the type on the input from number to text. I had to mimic the input number field with mousewheel for increase and decrease the value.
Shift option is also added see wheel event.

https://user-images.githubusercontent.com/1670670/110682433-33b12380-81db-11eb-82e0-1c75cc58ccce.mp4

There is a mistake I haven't been able to fix yet, see the video. When i use one or multiple zero at the beginning of an expression it crash. I think it can be fixed with a better regex, if you are a pgm with regex let me know :')
Otherwise `3+alert("Hello")` is prevented.

https://user-images.githubusercontent.com/1670670/110682707-868adb00-81db-11eb-99e5-8b4152ace7b4.mp4


newIDE/app/src/PropertiesEditor/index.js
```diff
@@ -273,7 +273,7 @@ export default class PropertiesEditor extends React.Component<Props, {||}> {
             );
             this._onInstancesModified(this.props.instances);
           }}
-          type="number"
+          calculValue
```
arthuro555, line 276: I would call this `evaluateValue`

newIDE/app/src/UI/SemiControlledTextField.js
```diff
@@ -87,6 +93,22 @@ export default class SemiControlledTextField extends React.Component<
     if (this._field) return this._field.getInputNode();
   }
 
+  _onWheel(evt) {
+    const { focused, text } = this.state;
+
+    if (!focused) return;
+    if (typeof text === 'number') {
```
arthuro555, line 100: Can be simplified to 
```suggestion
    if (focused && typeof text === 'number') {
```

newIDE/app/src/UI/SemiControlledTextField.js
```diff
@@ -121,7 +144,17 @@ export default class SemiControlledTextField extends React.Component<
           if (!commitOnBlur) onChange(newValue);
         }}
         onBlur={event => {
-          onChange(event.currentTarget.value);
+          let value = event.currentTarget.value;
+          if (this.props.calculValue) {
+            const rex = /(^\d).*[-+*()/.,].*(\d+$)|(^\d+$)|^((\d)|[-+*()/.,]).((\d)|[-+*()/.,])+$/gm;
```
arthuro555, line 149: You are not escaping the slashes, are you sure this is working?
```suggestion
            const rex = /(^\d).*[-+*()\/.,].*(\d+$)|(^\d+$)|^((\d)|[-+*()\/.,]).((\d)|[-+*()\/.,])+$/gm;
```

newIDE/app/src/UI/SemiControlledTextField.js
```diff
@@ -121,7 +144,17 @@ export default class SemiControlledTextField extends React.Component<
...
+            const regexData = value.match(rex);
+
+            if (regexData) {
+              value = eval(regexData[0]);
```
arthuro555, line 153: I still think it would be simpler and probably more effective to just use a module made for evaluating mathematical expressions, like https://github.com/josdejong/mathjs-expression-parser

newIDE/app/src/UI/SemiControlledTextField.js
```diff
@@ -121,7 +144,17 @@ export default class SemiControlledTextField extends React.Component<
...
+            }
```
arthuro555, line 154: I would move the whole evaluating to https://github.com/4ian/GDevelop/blob/b6a5784968a8e794667b3f3f0f3b80b5695c07cd/newIDE/app/src/PropertiesEditor/index.js#L270

newIDE/app/src/UI/SemiControlledTextField.js
Bouh: I like use vanilla javascript, systematically add a package for solve a problem I always found it ridiculous and overrated. [I saw nice thread](https://stackoverflow.com/questions/7650071/is-there-a-way-to-create-a-function-from-a-string-with-javascript).

arthuro555: I do agree that using a package for everything is not a good solution, there are also packages made [to make fun of this](https://github.com/jezen/is-thirteen), but I don't think modules should be the very last resort either. I think this case is a good fit for an external module, as the eval is made to evaluate code, not maths, and using it for maths is missuing it. 

Your regex is not completely preventing javascript, this input for example matches your regex: `20;alert("some malicious code");2+3`. I doubt any regex can be 100% secure. Someone could make a really long expression containing this kind of malicious code, and tell a user to paste it in for a desired result, executing malicious code on his machine.

The only way to really not execute code, is not executing that as code in the first place. Simply parse it as a mathematical expression, what it is, and execute it. Of course, something like a math expression parser is very long and difficult to do. Fortunately, someone made one already. Why spend time remaking the wheel, when we can just use the code from a module?

4ian: Vanilla JavaScript is most of the time the way to go, but here for `eval` I would consider it almost not part of JS because of all the implications it has.

Taking a new dependency comes with a lot of factor that are both in favor or against it:
- size of the dependency?
- complexity of the dependency?
- is this something critical, or just something that we don't have time for and we're happy to delegate?
- any security risk with (or without the dependency)?
- can we just replace it with a few lines of code?
- any performance issue?
- is the package going to be maintained for the next years? If not do we have the capacity to take over it?

Here there are two things:
- I really don't want eval because there will be one way or the other a chance to inject code. It's also something that we could disable completely thanks to a [CSP directive](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) - so let's not build anything that relies on it. There are ways to do complex code in JS with as little as simple characters like [, ] and +.
  - I really don't want it :p It's sure it will be abused in the future. Or we'll get angry users yelling at us. Or both.
- I also don't want to get a complex math parser just for a simple, convenient feature like this. So can we find a dead simple, very lightweight, minimal math parser?
  - this one has no dependencies: https://github.com/bugwheels94/math-expression-evaluator
  - this one too? https://github.com/silentmatt/expr-eval/blob/master/package.json

arthuro555:
It may be a bit more complex but personally I'd love to have the possibility to not replace the expression with the evaluated value in the field, as it may be more readable. For example, if I have 3 * 800 and want to change it to 7 * 800, if it stayed as an expression, it would be an easier change. Though that probably means parsing the expression in C++.

Bouh:
> It may be a bit more complex but personally I'd love to have the possibility to not replace the expression with the evaluated value in the field. For example, if I have 3 * 800 and want to change it to 7 * 800, if it stayed as an expression, it would be an easier change. Though that probably means parsing the expression in C++.

I would like too, but it not smething in my skills scope.

Thank you guys for the workaround on eval, i know it's a bad pratice to use it, my goal was to make this pratical feature lighter as possible and frist of all demonstrated the proof of concept.
I keep this PR as draft until the my other PRs has been completed :)

4ian:
> my goal was to make this pratical feature lighter as possible and frist of all demonstrated the proof of concept.

Yep no worries, it's a great idea to use it as a proof of concept! :)

Bouh:
Rounding values and get correct result should be done in c++ core or in the react interface?
It's really disturbing to get  `1.1100000143051147 = 0.111+1 `

4ian:
This might be because we're using float instead of `double` in C++, where it should be fixed. I think your code in JS is fine.

Bouh:
> This might be because we're using float instead of `double` in C++, where it should be fixed. I think your code in JS is fine.

By replacing `float` by `double` here ?
`GDevelop\GDCpp\GDCpp\Runtime\RuntimeSpriteObject.cpp`
image

4ian:
Not here, what you're looking at is the C++ *runtime* object, i.e: the class that would be used in the C++ game engine :) 

Can you share somewhere where this is happening? I checked for instances (InitialInstance.h) but this is already using `double` for the angle (and position, and size) so might not be this.

Bouh:
> Can you share somewhere where this is happening?

Position Y in instance panel in input `0.111+1` = `1.1100000143051147` But it's only in the panel instance as far i know.

Bouh:
@4ian  you asked me some days ago where this imprecision come from.
Here a fresh report from a user in the event sheet.

https://forum.gdevelop-app.com/t/addition-bug/33313

4ian:
Ok I was confused because I know that 0.1=0.2=0.30000000000000004 in JS (you can test it in the console) but 0.111+1 is working well for me in the console.

So let's live with this and we can include rounding at 2 or 3 decimal places (let's go for 3 for games that would need a really large precision).

Bouh:
Properly done by Alex in https://github.com/4ian/GDevelop/pull/6914, thank you :)

PR #2427, opened 2021-03-11
Add "simulate stick" action to the top-down behavior

D8H:
Can be use with the gamepad extension to give players finer control on top-down movement.
Stick-topdown


4ian (approved):
Very nice 👍

4ian:
Thanks a lot 👍👍

PR #2428, opened 2021-03-12
Generate variables in For Each Child events as expressions

arthuro555:
This allows to use child variables as well as iterable or iterators.


4ian:
This makes me think, what happen if you try to use the key or value variable being the same variable as the one we iterate on? In theory it should work but let's give it a try to make sure there is no bug/crash doing this :)

4ian (approved):

arthuro555:
I tested, it doesn't crash the game.

4ian:
Merged, thanks! :)

PR #2432, opened 2021-03-14
Bitmap Text Object

Bouh:
This PR add Bitmap Text object and change PixiJS 5.3.3 to 5.3.7 for fixes on PIXI.BitmapText

Still to do:
- [x] Changing the texture won't properly update the font at runtime. This is because fonts are indexed only by the font data resource name.
- [x]  Do the same fix in the IDE.
- [x] Fonts are never uninstalled in the IDE. Would be good to do some cleaning at least when the project is closed..
- [x] BitmapFontManager to be moved to the extension and renamed PixiBitmapFontManager.
- [x] await to be removed in the extension runtime, replaced by Promise.all (not supported by enough browsers).
- [x] Replace pixi.js by its minified version


4ian:
> This PR add Bitmap Text object and change PixiJS 5.3.3 to 5.3.7 for fixes on PIXI.BitmapText
> 
> Still to do:
> 
> * [ ]  Changing the texture won't properly update the font at runtime. This is because fonts are indexed only by the font data resource name.

This is because we store bitmap fonts like this:

```
// Return the existing Bitmap Font that is already in memory and already installed.
      if (this._installedPixiBitmapFont[bitmapFontResourceName]) {
```

I'll probably update this so that you pass both the bitmapFontResourceName and the bitmapAtlasResourceName to the function, that will use *both* as the key.

> * [ ]  Fonts are never uninstalled in the IDE. Would be good to do some cleaning at least when the project is closed..

Might need a bit of rework in the scene editor to do that.

> * [ ]  BitmapFontManager to be moved to the extension and renamed PixiBitmapFontManager.

Should be fairly easy.

> * [ ]  await to be removed in the extension runtime, replaced by Promise.all (not supported by enough browsers).

We sadly need to get rid of async/await, and replace them with a Promise.all that will wait for all bitmap font data to be loaded.

Let me know if you take a look at one of these, otherwise I hope to do them this week :)

Bouh:
I took a look.
- It seems that there is no rendering in the editor anymore, something has been broken here I had a rendering a few days ago.
- I can't clear PIXI cache from mainframe, I don't have access to PIXI.
Seem you have wrote a todo in the past where we should clear the cache. `// TODO: Pixi cache should also be burst`
- BitmapFontManager is based on JsonManager it depends of files, I don't see how to put the manager in the extension, if a project loads bitmap resources and doesn't use the object then the extension will not be exported.
All remaning points seems to be less simple to me :|

4ian:
No worries I'll take care of the remaining points :)

4ian:
This should work now and this is almost finished with this one major flaw:
- The call to `uninstall` is never done. This call is implemented for both the IDE (when no instance is using a font anymore) and the runtime (when a font is removed from the cache of BitmapFont to delete) but commented because there seems to be an issue with this: when calling it, the atlas texture is destroyed from memory... which will crash the game (or the IDE) if another object (like a sprite), use the same texture! 
  - Also in the game, because textures are supposed to be preloaded, this means that if you don't have anything else using the texture, you don't get a crash, but next time you try to create a BitmapText, it will fail as the texture is invalid.
  
This means that for now we "leak" memory, in the sense that a loaded font is staying forever in memory. Not a problem if you only have a few bitmap fonts, but could be a problem if you're opening a lot of different games in the IDE or make a lot of fonts with different data and textures in the IDE.
The uninstall problem should probably be asked on Pixi.js slack/issue.
There is a more generic problem of the IDE never freeing the textures, but this is already there and not linked to this object.

Despite this, we should be good to go soon as these loading problems are not worse than in the current state.

4ian:
Alright this should be mostly ready apart from this "never released" fonts/textures in the IDE and runtime - though it's the case for everything right now.
I've confirmed on the PixiJS slack this issue with `uninstall` also destroying the texture/base texture. I'll open an issue for this.

4ian (approved):
Approving but let's wait a bit before merging :)

Bouh:
Yes, I will build a version, test it and share it with interested people.
Thx for the final commits, i go back to read what you have changed.

4ian:
Folks at PixiJS just fixed the issue of `uninstall` destroying the texture: https://github.com/pixijs/pixi.js/issues/7323
So we will be able to uncomment the `uninstall` when we upgrade to Pixi v6 (with the bugfix) :) 

As it won't cause huge issues I think not to uninstall (unless you make thousands of different BitmapText with each a different font, which is I think highly unlikely even in a big game), we can probably merge this for the next release.
@bouh Have you shared the build with some interested people? :)

Bouh:
Oh not yet, I'll do that after my job.

Bouh:
Ok here the [portable test build for Windows](https://mega.nz/file/IToiCYAD#5x9XKm_mIzl8s_qoN7WThMVwxgKDxvzlUpUHqIiGbBk).
@Silver-Streak  if you are insterested, you can try to switch with the action the resources like the bitmap font, and the atlas texture on a Bitmap Text object in game!

I've used the [nightly dev version](https://pixijs.download/dev/pixi.min.js) of PixiJS with the fix.
[Here a video ](https://streamable.com/tor55s)of the result, before with PixiJS unfixed and a after with the fix.
Uninstall is uncommented for both tests on the video.

Silver-Streak:
I've done some testing with Bouh on the discord. Here's my project. There's some errors in the console but everything seems to work fine. 
[BitmapTest0321.zip](https://github.com/4ian/GDevelop/files/6200624/BitmapTest0321.zip)

4ian:
> This way the order they're loaded in can be controlled, and if the user doesn't want to change the font file they can just populate the same name.

Ok fair enough, I think Bouh did this originally, and I separated this in two actions, but it makes sense to combine them back in a single action.

Bouh:
I've made changes for merge the actions in one.
Uncomment the uninstall.
Update runtime pixi to v6.0.1

4ian:
Thanks 👍

Bouh:
```JS
./node_modules/@pixi/polyfill/lib/polyfill.es.js
Module not found: Can't resolve 'es6-promise-polyfill' in 'C:\GDevelop\newIDE\app\node_modules\@pixi\polyfill\lib'

```

Related too hack on for use Pixi v6.0.1 i guess?

Bouh:
Bad news, the preview on the scene editor don't render the texture atlas.
The text wrapping is broken.

image

The preview is fine.

Bouh:
It was an issue with my cache, pixi was still on 5.3.3 in the IDE...

Silver-Streak:
@Bouh I got a notice from Bountysource on this. Are you feeling we're close to merging this in?

4ian:
Pixi v6 being merged in, we're very close.
@Bouh So was there any issue or everything seemed alright after re-launching a `npm install`?

Bouh:
I had a problem with the modules not updating so I was stuck in pixijs 5.3.3 in the IDE.
After deleting the modules folders and restarting, everything is ok in the IDE and the preview in pixijs 6.0.1

4ian:
Ok! Let's have as a goal to merge this this week then. I'll make a new pass on the PR because I left some TODOs, then we should be good :)

Bouh:
@4ian  Is something blocking the last review?

4ian:
Nothing, just time that is very scarce these last days 😬
It's on my to-do for sure!

4ian:
Finally got some time to work again on this, I think we're good to go?
@Bouh / @Silver-Streak want to give a last check at this?

Silver-Streak:
> Finally got some time to work again on this, I think we're good to go?
> @Bouh / @Silver-Streak want to give a last check at this?

I can't take a look tonight, but I'll try to take a look tomorrow or Tuesday. I'm re-working the Not-a-vania demo and this will be a good use case to add the text using bitmap text rather than the default text object.

Bouh:
I've test from my side, i found no major issue.
@Silver-Streak i'll update this post in few minutes with a custom build!

Silver-Streak:
> I've tested from my side, I found no major issue.
> @Silver-Streak I'll update this post in few minutes with a test build!
> 
> https://mega.nz/file/1LQmxTqS#WtfbnuEg4dweBJWLbezXnK32C6MTdi8Wd5feLS1bSNA

Thanks. Setting up right now, I'll load up a few different bitmap fonts into a copy of my game example I'm working on, and see if anything behaves funky.

Silver-Streak:
Only thing to call out first: none of the debug (Ctrl+Shift+J/f12), f11 to maximize functions, alt+f4 keys are working in the test build, unsure if that's related or just a flag.

On topic: 
Layer effects interacting with Bitmap font perfectly fine: image

Dynamically updating bitmap font text working fine: 
image

Other things I tested: 
Position
Scale
Tint
Opacity

Everything seems good to me!

Bouh:
> Only thing to call out first: none of the debug (Ctrl+Shift+J/f12), f11 to maximize functions, alt+f4 keys are working in the test build, unsure if that's related or just a flag.

I agree, FYI it happened also on nightly.
Not related to this PR.

4ian:
Let's go 🥳

Bouh:
I'm glad to see this great pull request closed 😌
Thanks guys for your help!
The best part will start when the first bugs reports will come to me 😅

PR #2435, opened 2021-03-15
Change for each child variable description

arthuro555:
Structure -> Collection


4ian:
I took the liberty of entering "structure or array" as I think it's a bit more telling for beginners. Thanks for spotting this :)

PR #2436, opened 2021-03-15
Visibility and type definitions in the top-down behavior

D8H:


4ian (approved):
I fixed a type `String` that should be `string`, otherwise looks good :)

PR #2445, opened 2021-03-18
Add an isometric grid option

D8H:
When it's enable it will a draw diamonds instead of rectangles for each cell.
It uses generators to draw the lines. The code is a lot more simple that way but it may not be a good idea on a draw function.

IsometricGridConf

This is not just graphical because the magnet logic also changes.

IsometricGridLogic

### DIFF
--- a/Core/GDCore/IDE/Dialogs/LayoutEditorCanvas/LayoutEditorCanvasOptions.cpp
+++ b/Core/GDCore/IDE/Dialogs/LayoutEditorCanvas/LayoutEditorCanvasOptions.cpp
@@ -17,7 +17,7 @@ LayoutEditorCanvasOptions::LayoutEditorCanvasOptions()
       gridHeight(32),
       gridOffsetX(0),
       gridOffsetY(0),
-      gridType("retangular"),
+      gridType("rectangular"),
       gridR(158),
       gridG(180),
       gridB(255),
--- a/Core/GDCore/IDE/Dialogs/LayoutEditorCanvas/LayoutEditorCanvasOptions.cpp
+++ b/Core/GDCore/IDE/Dialogs/LayoutEditorCanvas/LayoutEditorCanvasOptions.cpp
@@ -17,6 +17,7 @@ LayoutEditorCanvasOptions::LayoutEditorCanvasOptions()
       gridHeight(32),
       gridOffsetX(0),
       gridOffsetY(0),
+      gridType("retangular"),
       gridR(158),
       gridG(180),
       gridB(255),
@@ -30,6 +31,7 @@ void LayoutEditorCanvasOptions::SerializeTo(SerializerElement& element) const {
   element.SetAttribute("gridHeight", gridHeight);
   element.SetAttribute("gridOffsetX", gridOffsetX);
   element.SetAttribute("gridOffsetY", gridOffsetY);
+  element.SetAttribute("gridType", gridType);
   element.SetAttribute("gridR", gridR);
   element.SetAttribute("gridG", gridG);
   element.SetAttribute("gridB", gridB);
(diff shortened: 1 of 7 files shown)


Core/GDCore/IDE/Dialogs/LayoutEditorCanvas/LayoutEditorCanvasOptions.cpp
```diff
@@ -17,6 +17,7 @@ LayoutEditorCanvasOptions::LayoutEditorCanvasOptions()
       gridHeight(32),
       gridOffsetX(0),
       gridOffsetY(0),
+      gridType("retangular"),
```
Bouh, line 20:
```suggestion
      gridType("rectangular"),
```

D8H: Nice catch

4ian:
Interesting! Do you think it would be easy to make the isometry angle customizable (like in the top down behavior)?

I'm a bit concerned about using generators as they can end up generating a bunch of boilerplate code and nothing would really beat the performance of a simple function call.
How would you do it otherwise?

D8H:
> Interesting! Do you think it would be easy to make the isometry angle customizable (like in the top down behavior)?

* For pixel isometry, it's easier to set the grid with width and height.
* For true isometry, I'm not sure there is a consensus.
   * In my case, I used a 64x64 square and its projected dimensions are on the screenshot. So, it would be easier to input isoSize=64 and angle=30.
   * But I imagine using a projected width that is an integer could be a good solution to make assets. Then, it will be width=90 angle=30.
   * Could do the same with height instead of width.

I think it would make the UI confusing. We may assume people doing true isometry know what they get themselves into and let them copy/paste float values. However, I think a way to apply it globally to every scene or related layouts could help a lot.

> I'm a bit concerned about using generators as they can end up generating a bunch of boilerplate code and nothing would really beat the performance of a simple function call.
> How would you do it otherwise?

It can be done with classic loops. It will be more difficult to read but probably a lot faster.

D8H:
By the way, it's the same for path finding setting. I chose to use the width and height as dimension in the isometric world, but for the pixel isometry at least it must be the dimension after the projection.

The same logic must be used for settings in the grid and path finding so users can make it matched.

newIDE/app/src/Utils/GridHelpers.js
```diff
@@ -1,12 +1,61 @@
...
+    if (gridWidth <= 0) {
+      pos[0] = Math.round(pos[0]);
+    } else {
+      pos[0] =
+        Math.floor((pos[0] - gridOffsetX) / gridWidth) * gridWidth +
+        gridOffsetX;
+    }
+
+    if (gridHeight <= 0) {
+      pos[1] = Math.round([1]);
```
4ian, line 54: `pos[1]` - I'm surprised this was not caught by flow type checking?

newIDE/app/src/Utils/GridHelpers.js
```diff
@@ -1,12 +1,61 @@
...
+  gridOffsetX: number,
+  gridOffsetY: number,
+  gridType: number
 ) => {
-  if (gridSize <= 0) {
-    return Math.round(pos);
+  if (gridType === 'isometric') {
+    if (gridWidth <= 0 || gridHeight <= 0) {
+      pos[0] = Math.round(pos[0]);
+      pos[1] = Math.round([1]);
```
4ian, line 14: pos[1]?

newIDE/app/src/Utils/GridHelpers.js
```diff
@@ -1,12 +1,61 @@
...
+      return;
+    }
+    // Why do we need this?
+    // Take a 2x2 squares, put a diamond inside each square, there is a 5th diamond in the center
+    let cellX = Math.round(((pos[0] - gridOffsetX) * 2) / gridWidth);
+    let cellY = Math.round(((pos[1] - gridOffsetY) * 2) / gridHeight);
+
+    if ((((cellX + cellY) % 2) + 2) % 2 == 1) {
```
4ian, line 22: Isn't this equivalent to `(cellX + cellY) % 2 == 1`?

newIDE/app/src/Utils/GridHelpers.js
```diff
@@ -1,12 +1,61 @@
...
-  if (gridSize <= 0) {
-    return Math.round(pos);
+  if (gridType === 'isometric') {
+    if (gridWidth <= 0 || gridHeight <= 0) {
+      pos[0] = Math.round(pos[0]);
+      pos[1] = Math.round([1]);
+      return;
+    }
+    // Why do we need this?
+    // Take a 2x2 squares, put a diamond inside each square, there is a 5th diamond in the center
```
4ian, line 18: This comment is not really explaining the calculation, could you add something explaining what cellX/Y represents?

newIDE/app/src/SceneEditor/SetupGridDialog.js
```diff
@@ -69,7 +71,7 @@ export default class SetupGridDialog extends Component {
               type="number"
               value={this.state.gridWidth}
               onChange={(e, value) =>
-                this.setState({ gridWidth: parseInt(value, 10) })
+                this.setState({ gridWidth: parseFloat(value, 10) })
```
4ian, line 74: parseFloat is taking just the value argument, you can remove 10 which is unused.

newIDE/app/src/Utils/GridHelpers.js
D8H: I think -3 % 2 == -1

4ian (approved):
Looks great! Thanks for the additional comments, very useful 👍

PR #2465, opened 2021-03-24
Replace the term "Force" by "Velocity"

D8H:
I think some actions use the term "Force" instead of "Velocity". If this is really not the right term, I see 2 reasons that can motivate a renaming:
* it could mislead children to what a force really is
* teachers could choose not use GDevelop worrying about the previous point

Some expression these descriptions also use "Average" instead of "Sum".

Forces

### DIFF
--- a/Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
@@ -167,10 +167,10 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
 
   obj.AddAction(
          "AddForceXY",
-         _("Add a force"),
-         _("Add a force to an object. The object will move according to "
-           "all of the forces it has."),
-         _("Add to _PARAM0_ _PARAM3_ force of _PARAM1_ p/s on X axis and "
+         _("Add a velocity"),
+         _("Add a velocity to an object. The object will move according to "
+           "all of the velocities added to it."),
+         _("Add to _PARAM0_ _PARAM3_ velocity of _PARAM1_ p/s on X axis and "
            "_PARAM2_ p/s on Y axis"),
          _("Movement"),
          "res/actions/force24.png",
@@ -182,11 +182,11 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
       .AddParameter("forceMultiplier", _("Force multiplier"));
 
   obj.AddAction("AddForceAL",
-                _("Add a force (angle)"),
-                _("Add a force to an object. The object will move according to "
-                  "all of the forces it has. This action creates the force "
+                _("Add a velocity (angle)"),
+                _("Add a velocity to an object. The object will move according to "
+                  "all of the velocities added to it. This action creates the velocity "
                   "using the specified angle and length."),
-                _("Add to _PARAM0_ _PARAM3_ force, angle: _PARAM1_ degrees and "
+                _("Add to _PARAM0_ _PARAM3_ velocity, angle: _PARAM1_ degrees and "
                   "length: _PARAM2_ pixels"),
                 _("Movement"),
                 "res/actions/force24.png",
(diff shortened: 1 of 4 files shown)


Bouh:
I'm not sure about the added value of these changes.
- From a linguistic point of view I trust you.
- From a user point of view I understand what force means.
- From a developer point of view having new instructions hidden and replaced for a wording issue I think it's not nice.
It makes the code base a bit bigger, it will be cleaned up in several years, due to the level of compatibility of the projects we have on GD.

4ian:
I see the idea, but this is the kind of change that will have huge repercussion in the community in the sense that all existing conversations about Force could confuse new people, it will require an upgrade of all pages of the wiki mentioning them (doable though) and same for all video tutorials that mention forces..

I agree though that force is not the proper exact term - though it conveys the idea of movement pretty well (velocity and impulse are a bit less friendly).
I'm happy to use sum instead of average though, seems like a not too big of a change.

D8H:
So you're saying it's a technical debt?
The more time passes, the more it will be difficult to change. So better be quick, right? :innocent:

4ian:
> It makes the code base a bit bigger

It's ok with some helpers I introduced recently like `extension.AddDuplicatedCondition` :) 

> The more time passes, the more difficult it will be to change it. So better be quick, right? 😇

In theory yes - it's not even a lot of technical debt :) 

In practice I'm afraid of a big "churn" of content - it would be really a huge "community/tutorial debt" and I'm a bit afraid of having the cons outweightings the pros... Would need more arguments about how we do the transition for the whole ecosystem - the code is really the easy part here :) 

Though we can have a in the middle approach, like: `Add a velocity ("force")` to help people find their way from "outdated" tutorials.

D8H:
How does the action/condition search work?
Is there a way to add key words?

.project
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Bouh, line 1: Unused file, you can delete it

Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
```diff
@@ -167,10 +167,10 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
 
   obj.AddAction(
          "AddForceXY",
-         _("Add a force"),
-         _("Add a force to an object. The object will move according to "
-           "all of the forces it has."),
-         _("Add to _PARAM0_ _PARAM3_ force of _PARAM1_ p/s on X axis and "
+         _("Add a velocity ("force")"),
```
Bouh, line 170:
```suggestion
         _("Add a velocity (Force)"),
```
For fit with current format in app.
image

Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
```diff
@@ -182,11 +182,11 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
       .AddParameter("forceMultiplier", _("Force multiplier"));
 
   obj.AddAction("AddForceAL",
-                _("Add a force (angle)"),
-                _("Add a force to an object. The object will move according to "
-                  "all of the forces it has. This action creates the force "
+                _("Add a velocity ("force") with an angle"),
```
Bouh, line 185:
```suggestion
                _("Add a velocity (Force) with an angle"),
```

Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
```diff
@@ -634,8 +634,8 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
 
   obj.AddAction(
          "AddForceVers",
-         _("Add a force to move toward an object"),
-         _("Add a force to an object to make it move toward another."),
+         _("Add a velocity ("force") to move toward an object"),
```
Bouh, line 637:
```suggestion
         _("Add a velocity (Force) to move toward an object"),
```

Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
```diff
@@ -649,8 +649,8 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
 
   obj.AddAction(
          "AddForceTourne",
-         _("Add a force to move around an object"),
-         _("Add a force to an object to make it rotate around another.\nNote "
+         _("Add a velocity ("force") to move around an object"),
```
Bouh, line 652:
```suggestion
         _("Add a velocity (Force) to move around an object"),
```

Bouh:
If I got how `extension.AddDuplicatedCondition()` works ([GDCore Documentation](https://docs.gdevelop-app.com/GDCore%20Documentation/classgd_1_1_platform_extension.html#a710cbffb19a79396a29d21e157868fc6))


We must rename old expressions like **ForceX** to **VelocityX** and just add after the expression edited: 



```Js
extension
    .addDuplicatedCondition('VelocityX', 'ForceX')
    .setHidden();
```

PR #2474, opened 2021-03-27
Fix bug in theme creation script

NilayMajorwar:
While trying to create a new theme, I realized that the `create-new-theme.js` script doesn't work correctly:
- In the template `index.js` theme file, `createGdevelopTheme` is wrongly imported from `../theme` (which does not exist in the `Themes` directory) and instead should be imported from `../CreateTheme`. And indeed, creating a new theme with the script leads to Flow compilation errors.
- The script did not create the theme name correctly - user input "Foo Bar" would lead to a theme name "Foo BarTheme" displayed to the end-user.

I've fixed both of the above and tested. New themes show up properly to the end-user and the styling still works.


4ian (approved):
Awesome! Thanks for the fix :)

PR #2477, opened 2021-03-27
Fix: the platformer object should not be able to jump when falling

D8H:
Checked with the example too.


4ian:
Thanks for the speedy fix and test! 👍

PR #2478, opened 2021-03-27
Make the wheel zoom keeps the mouse in place in the instances editor.

D8H:
This allows to point an zoom like in Inkscape.

This can save a lot of time when polishing a big level:
* The user adds some decorations to the level
* He jumps to one location by pointing it with the mouse and zoom in
* He adds and adjusts one decoration
* He zooms out to have a level overview
* He points another location and zoom in
* He adds another decoration
* and so on

Without it, the user needs to recenter the view on the target multiple times when zooming in.

### DIFF
--- a/newIDE/app/src/InstancesEditor/ViewPosition.js
+++ b/newIDE/app/src/InstancesEditor/ViewPosition.js
@@ -48,8 +48,8 @@ export default class ViewPosition {
   toSceneCoordinates = (x: number, y: number): [number, number] => {
     x -= this._width / 2;
     y -= this._height / 2;
-    x /= Math.abs(this._pixiContainer.scale.x);
-    y /= Math.abs(this._pixiContainer.scale.y);
+    x /= Math.abs(this.options.zoomFactor);
+    y /= Math.abs(this.options.zoomFactor);
 
     var viewRotation = 0;
     var tmp = x;
@@ -80,8 +80,8 @@ export default class ViewPosition {
       Math.sin((viewRotation / 180) * Math.PI) * tmp +
       Math.cos((viewRotation / 180) * Math.PI) * y;
 
-    x *= Math.abs(this._pixiContainer.scale.x);
-    y *= Math.abs(this._pixiContainer.scale.y);
+    x *= Math.abs(this.options.zoomFactor);
+    y *= Math.abs(this.options.zoomFactor);
 
     return [x + this._width / 2, y + this._height / 2];
   };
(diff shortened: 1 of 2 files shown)


newIDE/app/src/InstancesEditor/ViewPosition.js
```diff
@@ -48,8 +48,8 @@ export default class ViewPosition {
   toSceneCoordinates = (x: number, y: number): [number, number] => {
     x -= this._width / 2;
     y -= this._height / 2;
-    x /= Math.abs(this._pixiContainer.scale.x);
-    y /= Math.abs(this._pixiContainer.scale.y);
+    x /= Math.abs(this.options.zoomFactor);
+    y /= Math.abs(this.options.zoomFactor);
```
D8H, line 52: I'm not sure about the side effects of this.

Bouh:
I see any side effect, the `this._pixiContainer.scale` is only defined at the end of this file, and seem not changed somewhere else in the IDE. And there is no setter for it.

D8H:
> I see any side effect, the `this._pixiContainer.scale` is only defined at the end of this file, and seem not changed somewhere else in the IDE. And there is no setter for it.

toSceneCoordinates() and toCanvasCoordinates() will take into account zoom changes before a call to render(). Is this ok?

newIDE/app/src/InstancesEditor/index.js
```diff
@@ -457,6 +457,16 @@ export default class InstancesEditor extends Component<Props> {
     this.setZoomFactor(this.getZoomFactor() + value);
   }
 
+  zoomOnMouseBy(value: number) {
+    const beforeZoomCursor = this.getLastCursorSceneCoordinates();
+    this.setZoomFactor(this.getZoomFactor() + value);
+    const afterZoomCursor = this.getLastCursorSceneCoordinates();
+    this.viewPosition.scrollBy(beforeZoomCursor[0] - afterZoomCursor[0], beforeZoomCursor[1] - afterZoomCursor[1]);
```
4ian, line 464: Let's add a quick comment `// Compensate for the cursor change in position, so that the zoom is done "centered" on the cursor`.

4ian (approved):
Added a few nitpicking, otherwise great change! Will make things easier for large levels indeed.

Bouh:
`this.options` is initialized in the constructor, the value is filled.

4ian:
Yep, `this.options.zoomFactor` is fine to use, it should always have a proper value. 
Don't forget to run autoformatting too on files :) https://semaphoreci.com/4ian/gd/branches/pull-request-2478/builds/1

4ian:
Thank you!

PR #2480, opened 2021-03-28
Fix crash on empty while event

### LINKED ISSUE #2437: [$45 bounty] Empty while block stuck loading game
## Describe the bug
An empty while event in an empty project crash the game.
I can't see errors in debugger or console because i think there is an infinite loop.
So here the code from the temp folder:

```JS
gdjs.New_32sceneCode = {};
gdjs.New_32sceneCode.stopDoWhile2 = false;


gdjs.New_32sceneCode.conditionTrue_0 = {val:false};
gdjs.New_32sceneCode.condition0IsTrue_0 = {val:false};


gdjs.New_32sceneCode.eventsList0 = function(runtimeScene) {

};gdjs.New_32sceneCode.eventsList1 = function(runtimeScene) {

{


gdjs.New_32sceneCode.stopDoWhile2 = false;
do {if (true) {
if (true) {

{ //Subevents: 
gdjs.New_32sceneCode.eventsList0(runtimeScene);} //Subevents end.
}
} else gdjs.New_32sceneCode.stopDoWhile2 = true; 
} while ( !gdjs.New_32sceneCode.stopDoWhile2 );

}


};

gdjs.New_32sceneCode.func = function(runtimeScene) {
runtimeScene.getOnceTriggers().startNewFrame();


gdjs.New_32sceneCode.eventsList1(runtimeScene);
return;

}

gdjs['New_32sceneCode'] = gdjs.New_32sceneCode;

```

image

## Other details
* b105, b106
[Source: forum](https://forum.gdevelop-app.com/t/empty-while-block-stuck-loading-game/31343)

arthuro555:
Fixes #2437 

Doesn't generate while events without any conditions as it is impossible to break out of them and crashes the game. If you really wish to crash the game, you can still put the always condition in the while event.

### DIFF
--- a/GDJS/GDJS/Extensions/Builtin/CommonInstructionsExtension.cpp
+++ b/GDJS/GDJS/Extensions/Builtin/CommonInstructionsExtension.cpp
@@ -347,11 +347,18 @@ CommonInstructionsExtension::CommonInstructionsExtension() {
       [](gd::BaseEvent& event_,
          gd::EventsCodeGenerator& codeGenerator,
          gd::EventsCodeGenerationContext& parentContext) {
-        gd::String outputCode;
         gd::WhileEvent& event = dynamic_cast<gd::WhileEvent&>(event_);
 
-        // Context is "reset" each time the event is repeated (i.e. objects are
-        // picked again)
+        // Prevent code generation if there are no conditions, as this would
+        // get the game stuck in a never ending loop.
+        if (event.GetWhileConditions().empty())
+          return gd::String(
+              "\n// While event not generated to prevent a crash\n");
+       
+        gd::String outputCode;
+
+        // Context is "reset" each time the event is repeated (i.e. objects
+        // are picked again)
         gd::EventsCodeGenerationContext context;
         context.InheritsFrom(parentContext);
         context.ForbidReuse();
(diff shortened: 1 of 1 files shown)


GDJS/GDJS/Extensions/Builtin/CommonInstructionsExtension.cpp
```diff
@@ -347,11 +347,18 @@ CommonInstructionsExtension::CommonInstructionsExtension() {
...
-        gd::String outputCode;
         gd::WhileEvent& event = dynamic_cast<gd::WhileEvent&>(event_);
 
-        // Context is "reset" each time the event is repeated (i.e. objects are
-        // picked again)
+        // Prevent code generation if there are no conditions, as this would
+        // get the game stuck in a never ending loop.
+        if (event.GetWhileConditions().empty())
+          return gd::String(
+              "\n// While event not generated to prevent a crash\n");
```
4ian, line 356: nitpicking: `infinite loop` instead of crash.

GDJS/GDJS/Extensions/Builtin/CommonInstructionsExtension.cpp
```diff
@@ -347,11 +347,18 @@ CommonInstructionsExtension::CommonInstructionsExtension() {
...
          gd::EventsCodeGenerator& codeGenerator,
          gd::EventsCodeGenerationContext& parentContext) {
-        gd::String outputCode;
         gd::WhileEvent& event = dynamic_cast<gd::WhileEvent&>(event_);
 
-        // Context is "reset" each time the event is repeated (i.e. objects are
-        // picked again)
+        // Prevent code generation if there are no conditions, as this would
+        // get the game stuck in a never ending loop.
+        if (event.GetWhileConditions().empty())
```
4ian, line 354: I would also check for the conditions and actions inside the while, if absolutely everything is empty.

The idea is to avoid bug reports because "while does not work". It should just do nothing when there is nothing at all.

4ian (commented):
That was quick :)

arthuro555:
> That was quick :)

I was already coding (working in GDMod) and "in the flow" when I received a GitHub notification about this easy fix so I figured I'd look into it.

4ian (approved):

PR #2481, opened 2021-03-28
Rework collisions editor and points editor

NilayMajorwar:
Originally suggested by @Bouh  on the forum: https://forum.gdevelop-app.com/t/suggestion-little-gui-change-for-sprites/22566/5

The idea is to rework the sprite collisions editor and points editor to be horizontally split into two panes: the sprite+hitbox/points on the left pane and the (scrollable) configuration UI on the right. For smaller screens, the panes can be vertical instead of horizontal as suggested here: https://forum.gdevelop-app.com/t/hitbox-and-points-editor/20008

@4ian Right now I'm thinking of using `EditorMosaic` for the resizable panes and `ResponsiveWindowMeasurer` which seems like it's used for getting the screen size type. The rest is straightforward, I think. Anything I should take care of?

#### Tasks and improvements:
- [x] Change layout to show preview and properties side-by-side
- [x] Show point name in tooltip when hovering on point
- [x] Accordion menu for showing collision shapes
- [x] Fix/improve point-dragging experience
- [x] Centre-align the image and overlays in preview pane
- [x] Fix `NaN` error in console
- [x] Add and tweak the mosaic drag handles
- [x] Make checkered background themeable

#### Stretch goals:
- ~Moving entire polygons or point sets together~ (in another PR, maybe)


4ian:
> @4ian Right now I'm thinking of using EditorMosaic for the resizable panes and ResponsiveWindowMeasurer which seems like it's used for getting the screen size type. The rest is straightforward, I think. Anything I should take care of?

No problem for `ResponsiveWindowMeasurer`, I actually encourage you to use the equivalent hook `useResponsiveWindowWidth` :)

> For smaller screens, the panes can be vertical instead of horizontal as suggested here

Yes :) That would be great!

Note that we've never used an EditorMosaic "inside" or at the "same time" as another. So I'm not sure how `react-mosaic` will handle the fact that you have on screen multiple mosaics, but feel free to give it a try! If it's too much of a problem (bad performance, bad behavior, weird bugs...), we can always use two `Column`s for now.

Bouh:
I have other ideas to have a better point editor.
- When hovering a point in the preview area a tooltip displays the name of the point.
- It is possible to add several collisions shapes .Having an accordion menu for each shape makes sense.

jack-ii:
- Increase the height of the preview areas by ~200px
- Don't drag the point/hitbox without pressing the mouse button (bug)
- Do not allow the two "Share same .." buttons can be clicked across the whole width of the window (bug)
- Show the point/hitbox coordinates in real time

NilayMajorwar:
> encourage you to use the equivalent hook useResponsiveWindowWidth

Oh great, that's better. But the relevant components are class components right now - is it fine if I refactor them to functional components? They don't seem to be very heavy (compared to components like ObjectsList).

@Bouh @jack-ii Sure, I'll try to implement your suggestions once the layout part is done!

4ian:
> is it fine if I refactor them to functional components? They don't seem to be very heavy (compared to components like ObjectsList).

Yep, feel free to do so! They are fairly old components :) 

> When we move a point of the collision shape out of the frame the point stop and when the cursor go back in the frame the point follow the cursor

This is indeed a bug of the drag'n'drop, as jack-ii mentioned too I think.

> There is something distrubing, the level of freedom of the point editor is a better behavior than the hitbox editor on the movable points.

Note that the limitation of the hitbox polygons being unable to go out of the frame is a limitation of the game engine. When an object is not rotated, its bounding box ("AABB") can be computed quickly using its X/Y position and width/height - assuming hit boxes are not outside of the object bounding box.

Bouh:
> > There is something distrubing, the level of freedom of the point editor is a better behavior than the hitbox editor on the movable points.
> 
> Note that the limitation of the hitbox polygons being unable to go out of the frame is a limitation of the game engine. When an object is not rotated, its bounding box ("AABB") can be computed quickly using its X/Y position and width/height - but only if we are sure hit boxes are not outside of the object bounding box.

If we can have the point sliding on the edges of the frame it coulb be more insteresting that just stop it because the cursor it out of frame.
Anyway it's a small thing, not a prior at all.

NilayMajorwar:
Okay, got something working:
image

The panes are resizable, the right pane scrolls independently. The problem is right now it works only when `EditorMosaic` is contained inside a constant-height `div` - anything other than that makes the mosaic collapse to zero height in both horizontal and vertical orientations. While this works fine on desktop screens, it makes the entire dialog scrollable on smaller screens. I'll try to fix this.

4ian:
These "height" problems are always a bit tricky to find the proper styles/flex to use!

> So fixed height is only used for desktop (and tablet?) screens - which on second thought seems proper for horizontally split panes, no?

Might be, though in this case we probably want to fill the screen. Can you try setting `fullHeight` prop for the dialog containing this editor? 
It's setting a min-height on the dialog, so I hope this can somehow transfer to the mosaic (maybe with height: 100%?)

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/CollisionMasksEditor/index.js
```diff
@@ -178,33 +178,37 @@ const CollisionMasksEditor = (props: Props) => {
...
-                  imageZoomFactor={imageZoomFactor}
-                  isDefaultBoundingBox={sprite.isCollisionMaskAutomatic()}
-                  polygons={sprite.getCustomCollisionMask()}
-                  onPolygonsUpdated={updateCollisionMasks}
-                />
-              )
-            }
-          />
-        </div>
+        <ImagePreview
```
4ian, line 194: Good idea to remove the extra divs, they might be the cause for some things not working with heights, so feel free to experiment removing/adding/changing these extra divs.

NilayMajorwar:
Weirdly, while `fullHeight` does prevent the dialog from collapsing, the div inside still collapses, even with `height: 100%` set. But `fullHeight` is definitely better than the initial approach, so I'll try playing around with this to hopefully get somewhere.

4ian:
Give it some tries yes, I think there might be a mix of flex: 1, position: relative, height: 100% that could help. I think You might study how tabs are working too, because that's where EditorMosaic works for now.

NilayMajorwar:
Okay, so `min-height` does not transfer to the mosaic-containing div, but a `height: 100%` does. Currently, providing `fullHeight` gives the following style to MUI dialog:
```css
.full-height-modal > .MuiDialog-container > .MuiPaper-root {
  min-height: calc(100% - 64px);
}

.MUI-INTERNAL-STYLE {
  max-height: calc(100% - 64px);
}
```

Replacing `min-height` with `height: 100%` solves the current use case without affecting other uses of `fullHeight` dialogs. The only issue is that this wouldn't work for Storybook where the `CollisionMasksEditor` story is not rendered inside a dialog. While that *can* probably be fixed by nesting the story with a fixed-height div, it still seems to indicates that the solution is not optimal. I'll try the two-`Column` layout to see how well that works.

I have an assignment due tomorrow, so may take some time!

NilayMajorwar:
Huh, interesting - I just noticed that the most recent commit I pushed (efb423860d7a783d6632d670236a11769bed4197), which does not include the CSS fix I mentioned above, works perfectly on Firefox but not on Chrome (and consequently not on Electron either). So there's some CSS here which is handled differently by the browsers.

Also, @4ian Is there some reason why `MainFrame.css` is not loaded in Storybook? The `fullHeight` prop doesn't work in Storybook at all since its styles are defined in `MainFrame.css` which doesn't seem to be loaded in Storybook.

4ian:
> which does not include the CSS fix I mentioned above, works perfectly on Firefox but not on Chrome (and consequently not on Electron either). So there's some CSS here which is handled differently by the browsers.

Yes it's a real pain, these kind of full height stuff never work on the first try and you always have to check on at least two browsers (I'll check on Safari too when we're done).

>  Flexboxed my way out.

Seems like the best thing, ideally I like to use flexbox everywhere.

> the "parent" SpriteEditor story still shows a "Edit Hitboxes" dialog with zero-height content since the fullHeight prop doesn't work. What should I do about that?

Would you be able to import directly the MainFrame.css in stories/index.js? That's already done for some CSS:
```
import '../UI/Theme/Global/Scrollbar.css';
import '../UI/Theme/Global/Animation.css';
```

NilayMajorwar:
> already done for some CSS

Oh, didn't notice that between all the JS imports. I tried importing MainFrame.css, but the CSS rule for `#root` messes up the stories that are not meant to be rendered inside a flexbox parent. So I just need to import a subset of the rules in MainFrame.css. Should I pull the css into a separate file? If yes, what would be an ideal name and location for the new CSS file?

Also, now that the layout part is nearly done, I'm looking at improving the point-dragging overlays experience. To fix the drag issues mentioned near the start of this thread, I'm thinking of expanding `PointsPreview.js` to span the whole preview pane instead of only the image frame. Currently `PointsPreview.js` handles the drag events, so it would be hard to fix frame-edge issues without it spanning the whole pane. And similar for the hitbox editor. Is that a good approach?

4ian:
> EDIT: I've instead moved the fullHeight CSS into the Dialog implementation itself.

Awesome, that's the best solution. It was in the CSS because at the time I think I did not find the proper prop where to put it! :) 

> To fix the drag issues mentioned near the start of this thread, I'm thinking of expanding PointsPreview.js to span the whole preview pane instead of only the image frame. Currently PointsPreview.js handles the drag events, so it would be hard to fix frame-edge issues without it spanning the whole pane. And similar for the hitbox editor. Is that a good approach?

I think it's a good approach. I don't see a proper way of doing it otherwise.
The disadvantage is that you might need to add a bit of complexity to compute the position of the points - but maybe it's actually not added complexity but just some logic to move from the parent to the PointsPreview.
Feel free to give a try but be sure to "save your progress" in case it's more complicated than anticipated ;)

Great work so far!

NilayMajorwar:
Done - it was easier than expected. Here's the result:

https://user-images.githubusercontent.com/18032938/113414278-4550aa00-93da-11eb-8257-71c138cc04bc.mp4

Expanding the preview component in size allows now to keep track of the mouse across the whole pane. 
- The vertex now tries to follow the mouse even if it is out of the sprite frame.
- > When we move a point of the collision shape out of the frame the point stop and when the cursor go back in the frame the point follow the cursor

  No more, @Bouh! This won't happen now, unless you take your mouse entirely out of the preview pane and then leave the mouse button. Please try it out if possible, see if there's anything else (even small things) you want fixed.

I've done the same for the points editor too, except there the points are allowed to go out of the frame. But there's a major bug right now, I realized that after pushing the commit - I'm fixing it. Zooming still works properly on both editors. Also, since this involved changes in the general `ImagePreview` component, I'll have to fix the Physics2 editor too (the only other preview that uses an overlay).

4ian:
Very nice! Much better now that we can continue dragging points even when the cursor is outside the frame of the image :) 

What do you think is missing /remain to be solved?

Bouh:
The preview isn't centered, for both editors.
image

The fact you use EditorMosaic is cool we can resize the view!At the top on the limit we can see the border, but the preview area is over the border.
image

When i zoom on sprite on the preview area there is some issues.
https://user-images.githubusercontent.com/1670670/113449168-8b6b3500-93fd-11eb-8f5e-2addca54612d.mp4

Bouh:
And an error on the console when i open up one of both editor for the first time.
image

NilayMajorwar:
> What do you think is missing /remain to be solved?

Apart from the existing bugs and suggestions mentioned on this PR, I can't think of anything else. I'm not very familiar with game development, so I have no idea what the general user workflow is :sweat_smile:. Perhaps I can ask for suggestions over on the Discord server.

> an error on the console when i open up one of both editor for the first time

That error exists on `master` too. I'll check it out and the other things you mentioned too.

Bouh:
Can I add a wish?
Have a way to move the points or the entire shape.
Don't spend too much time if it's too difficult, but as you ask I ask :)

NilayMajorwar:
> Have a way to move the points or the entire shapes.

That seems tricky... I'll try it once I'm done with the simpler tasks.

NilayMajorwar:
@Bouh I played around with the mosaic drag handles a bit:

https://user-images.githubusercontent.com/18032938/113490592-bb433700-94e8-11eb-9dc2-068b9c02779b.mp4

The default handle size is increased to 200px, and expands to 250px on hover. What do you think?

For horizontal edges, the hover effect goes away while dragging if the lower pane has a titlebar. This is because titlebars seem to creep a bit over the mosaic edges and steal hover status from the mosaic edge. This is a bug in `master` too, I'll try to fix it.

4ian:
On the aesthetic side, I think the double bars are a bit too much/too visible - I prefer the single one like before :) 
The size increase on hover is a nice idea! But I would keep a smaller size and single bar. We could also make it thicker rather than longer when hovered - might be more aligned with other UIs.

NilayMajorwar:
Okay, what about this:

https://user-images.githubusercontent.com/18032938/113501475-efe3dc80-9542-11eb-9ade-da06d96986d3.mp4

I've reduced the handle size to 100px. In thickness, it grows from 2px to 4px on hover.

NilayMajorwar:
- I was wondering if maybe we can make the preview background themeable too. Right now, it's a static background image for all themes. I think this background can be replicated using CSS, and then themes can then specify the colors for the light and dark boxes. A big white preview background is a bit annoying in a dark theme, no?
- While the point-dragging experience is better, I still find myself dragging a ghost image of the point instead whenever I want to drag a point. Can someone confirm this?

4ian:
> I've reduced the handle size to 100px. In thickness, it grows from 2px to 4px on hover.

Nice! You can also try a `transform: scale...` to see if leads to something a bit smoother (we can slightly see the growth to 3px, then to 4px and a repositioning - a transform might animate slightly better and could make use of the GPU).

> I was wondering if maybe we can make the preview background themeable too.

You should be able to do it, there are multiple ways to do it but using CSS to specify the background is a way to go I think :)
Though in the future we might want to add a button to toolbars to switch between a light background and a dark background for the preview - in which case we'll toggle a class or use the background image from JS and store it either in a preference (to keep the choice between dialogs and across restarts).
The theme can still override this value when it is chosen :)

> While the point-dragging experience is better, I still find myself dragging a ghost image of the point instead whenever I want to drag a point. Can someone confirm this?

I don't see this on the current version on mac. I see you've added `draggable={false}` - but this was not there before. Did you try to remove it (even if counterintuitive?).
Can you check if `user-drag: none` is properly applied to the img element style? (it should be done thanks to MainFrame.css, see the rule for `a:not([draggable=true]), img:not([draggable=true])`)

NilayMajorwar:
> Nice! You can also try a transform: scale... to see if leads to something a bit smoother (we can slightly see the growth to 3px, then to 4px and a repositioning - a transform might animate slightly better and could make use of the GPU).

Added `transform: scale ...` for the thickness transition:

https://user-images.githubusercontent.com/18032938/113505259-40673400-955b-11eb-94f2-4973c07575f2.mp4

>  using CSS to specify the background is a way to go I think

Alright, I'll try implementing this!

> I don't see this on the current version on mac. I see you've added draggable={false} - but this was not there before. Did you try to remove it (even if counterintuitive?).

Oh, my bad - I guess I committed that by accident, I indeed was experimenting with that. And yes, now that I checked it again, it's a Firefox-specific issue. I just tried fixing it, but documentation on `user-drag`-related properties is quite scarce. It's not a very critical bug, so maybe I can take a look at this later on.

4ian:
> And yes, now that I checked it again, it's a Firefox-specific issue. I just tried fixing it, but documentation on user-drag-related properties is quite scarce. It's not a very critical bug, so maybe I can take a look at this later on.

If it's Firefox specific, might be because there is only a webkit specific prefix for user-drag on `img`:
https://github.com/4ian/GDevelop/blob/3017798283bfba58f1e632419b2d9aeb77de4e4f/newIDE/app/src/MainFrame/MainFrame.css#L65-L68

Try adding  `-moz-user-drag: -moz-none;` or `-moz-user-drag: none;`?

NilayMajorwar:
Nope, Firefox treats all of them with "invalid property name" - `user-drag` seems to be a webkit-specific property.

4ian:
Alright that's no big deal then!

NilayMajorwar:
I got it, it's a JS solution - just need to `e.preventDefault()` on the pointer-down event on point `img`s. Works fine on Firefox, and other platforms are undisturbed (don't know about Safari though).

Bouh:
I like it!
All this UI changes makes GDevelop more and more polished.

4ian:
Agree, this is a long overdue refreshing of these interfaces :) 

For "Update dragged point coordinates in real time", I would suggest not to attempt this: this would mean to do a forceUpdate of the dialog anytime a point is dragged, which would severely impact performance. A proper solution would need to identify a ref to the inputs containing the dragged point, and then directly change the value of these inputs. Not impossible, but a bit more complex (because of the imperative way of updating things) as we would probably need to avoid a React render.

Seems we're good in terms of browser compatibility then, apart from Safari that I'll have to test. Happy to see mosaics working in a dialog, this means we'll be able to use them in more dialogs in the future! :)

NilayMajorwar:
I was working on the themeable background (I'm almost there), and just have a few questions:
- How are missing fields in `theme.json`s handled? Previously I thought custom themes are applied on top of the default theme, but it looks like missing fields just go to `undefined`.
- In relation to this, I guess then I'll also have to add the new fields with appropriate colors to the other themes too?

Also, regarding the other task "Accordion menu for collision shapes": I don't find any use of MUI Accordion anywhere in the code base right now (and the tree views have accordions internally). So just to ask, is it fine to directly use MUI Accordion for this task?

NilayMajorwar:
I've implemented the CSS background, just need to fix a minor visual artifact. I wanted to keep this style in CSS so that it is easily sharable between ImagePreview and ImageThumbnail with a class name - where should I put this CSS? Is it okay to add `ImagePreview.css` to `src/UI/Theme/Global`?

4ian:
>  So just to ask, is it fine to directly use MUI Accordion for this task?

Yes, if they make sense from a UI point of view (it's more something like a "collapsible panel" than an accordion I think, because the accordion force you to close the other sections when opening one).

> Previously I thought custom themes are applied on top of the default theme, but it looks like missing fields just go to undefined.

No unfortunately, I think you have to define the same field in every theme.

> Is it okay to add ImagePreview.css to src/UI/Theme/Global

I think it's fine. Usually I prefer to share stuff using components, but here a few CSS rules should be fine for this checkered background.

NilayMajorwar:
Okay, I tried implementing the CSS filter approach, and it's almost working. Currently there's a rather nasty layout bug that happens when zooming a lot into the sprite (like happened earlier in this PR), that I'm trying to resolve. Other than that, I think it looks pretty good (ignore the white bar at the right of each shot, I'll fix that):

preview-bg

Figuring out a good filter when creating a new theme is very difficult though, but eventually I figured out a way to get a very good filter for any theme:
1. Pick two colors (`theme.canvas.backgroundColor` and `theme.alternateCanvas.backgroundColor`) as your primary and secondary colors.
2. For each color, generate a good filter using [this tool](https://tests.christianoliff.com/hex-to-css-filter-generator/).
3. Compute the "average" of these two filters by just averaging the values of corresponding filter functions.
4. Replace `invert({x})` with `invert({1 - xyz})` and multiply the `brightness` filter by roughly `0.6`.

This should yield a very good filter for the theme - all above filters were obtained with this method (Solarized Dark needed some extra contrast modification). If this `filter` approach does make it into the codebase, I feel like this method definitely needs a mention in the theming docs :sweat_smile: .

So assuming the nasty layout bug is fixed properly, the filter approach may work quite well. If it doesn't, I'll fallback to gradients or maybe just scrap themeable backgrounds for the time being.

NilayMajorwar:
> Be careful on the height of the row

It's the same as before, I haven't changed it. Maybe the height is looking small now because the containing area is much larger now than before? If you want, I can increase the height of row.

Bouh:
As you want, I am nitpicking too specially on UI stuff 😁

NilayMajorwar:
Done - `filter`-based themeable backgrounds are working now, on all `ImagePreview`s and `ImageThumbnail`s.

> height of the row

Now that the pane is much larger in height, perhaps the rows can be made *slightly* larger. I'll have a look at that.

**Remaining tasks:**
- **Accordion for collision shapes**: While I'm absolutely open to implementing this, I also think the full-height panes resolve this use case to some extent - I can see almost 4 quadrilateral polygons at once on my screen. Again, I'll be glad to implement it if I'm wrong and accordions are still useful here.
- **(Stretch goal) moving entire polygons and point sets**: Feasibility aside, what do you think would be a good UI for this? Dragging needs to happen on the preview pane, and I can't think of an intuitive UI for such functionality - especially for polygons.

  In fact, this is even more complicated - while dragging an entire polygon I'll have to make sure none of the polygon points are exiting the sprite frame, and while dragging set of all points, points may go in and out of the viewable region leading to DOM element size changing while dragging. All in all, there seem to be quite some edge cases to take care of.

Bouh:
For the accordions and the polygons movement it's mostly for a longer term vision.
For the moving I I was thinking a toggle button between two modes, move the points and move the polygon shape.
For the accordions I think it will be useful in the future if one day the collision masks are extended with a layer system, or even different shapes of polygons, taking into account the concave shapes, and the possibility of putting attributes of names on the differents polygons.

An attribute would be useful to have a hitbox on the head of a character and another one on the body of the character, we could with the events distinguish the two hitboxes.
It is likely that when this system is in place we will need to have some order in the list with drop down / accordions sections.

At the moment I find that we have a good balance, for example with the behaviors of an object, the list of behaviors is quickly filled, in any case the accordion / dropdown section will be useful in the long term.
In any case it is not an absolute necessity.

NilayMajorwar:
Sorry, been busy for the past few days. I understand how useful Accordions or a similar collapsible component would be in the IDE - the behavior editor is probably the best example. I've been playing around with Accordion today a bit:

https://user-images.githubusercontent.com/18032938/114587984-ceea5c80-9ca3-11eb-9076-5058a41737ee.mp4

It's nowhere near as polished as it needs to be right now, but in terms of usability I think it's definitely better than before. Ignore the weird "+" below the accordions, I haven't worked on that yet. What do you think?

> For the moving I I was thinking a toggle button between two modes, move the points and move the polygon shape.

And how would you drag the polygons? We can't drag polygons by the body very easily, since they would mostly overlap. Maybe start dragging from any point of the polygon? Regardless, I think it would be better to work on that in a separate PR, especially considering all the edge cases.

4ian:
> Regardless, I think it would be better to work on that in a separate PR, especially considering all the edge cases.

Yep!

> Ignore the weird "+" below the accordions, I haven't worked on that yet. What do you think?

I think it looks great! If you have a good implementation here, we could adapt this pattern later to :
- the behavior editor, 
- possibly the effects list of a layer (or object in the future)
- behavior properties list 
- and in general any larger list of "things" (animations of a sprite?).

By default I think we can keep the accordion expanded? We can move to something smarter later (a button to expand/collapse all, or some condition to set the default state to collapsed/expanded).

newIDE/app/src/UI/Accordion.js
```diff
@@ -0,0 +1,103 @@
...
+
+type AccordionProps = {|
+  // AccordionSummary, AccordionBody and/or AccordionDetails
+  children: React.Node,
+
+  defaultExpanded?: boolean,
+  disabled?: boolean,
+
+  // If `true`, renders body only if accordion is open
+  costlyBody?: boolean,
```
4ian, line 82: Ah that's great! I was thinking of this kind of property. For the sprite editor or maybe actually most editors, it might be worth not rendering the body of a section if it's collapsed.
It's harder to animate, but still better if it's bringing better performances :)

Bouh:
This PR is already very good as it is, indeed questions arise for the movement of polygons, I have for the moment no answer.
I'm ok to thinking it later in another PR.

>I haven't worked on that yet. What do you think?

I think it looks great!

NilayMajorwar:
> I think it looks great!

Oh, good then! I won't need to make a lot of UI touches then. I'll clean up the polygons accordion code and push that work too. It would take some time (and review) though, since I left out all the `Sortable` stuff while refactoring the polygon list.

> If you have a good implementation here

Well, right now I know one minor inconvenience in usage:
- To add clickable actions in the "header" of an accordion (like the polygon delete button), you need to add `event.stopPropagation()` in the event handler, else the click also switches the accordion expand state. Unless this can somehow be factored into `Accordion.js`, this will probably a hassle for any component using the accordion.

> By default I think we can keep the accordion expanded?

Yeah I had that in mind before starting with the code - forgot it later. I'll make it the default.

> We can move to something smarter later (a button to expand/collapse all, or some condition to set the default state to collapsed/expanded)

Hmm, that would mean grouping a set of accordions together, no? Thinking about it, such a component (`AccordionGroup`) would probably be far more commonly used than independent `Accordion`s.

newIDE/app/src/UI/Accordion.js
```diff
@@ -0,0 +1,103 @@
+// @flow
+import * as React from 'react';
+import MUIAccordion from '@material-ui/core/Accordion';
+import MUIAccordionSummary from '@material-ui/core/AccordionSummary';
+import MUIAccordionDetails from '@material-ui/core/AccordionDetails';
+import MUIAccordionActions from '@material-ui/core/AccordionActions';
+import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
+import { IconButton } from '@material-ui/core';
```
4ian, line 8: nitpicking: use @material-ui/core/IconButton for consistency (and could help I think with code splitting in the future).

newIDE/app/src/UI/Theme/CreateTheme.js
```diff
@@ -138,6 +138,32 @@ export function getMuiOverrides(
         fontWeight: 400, // Lower a bit the weight of buttons
       },
     },
+    // Make MuiAccordion much more compact than default.
+    // Some (or all) of these styles can be removed on MUIv5,
+    // which introduces `disableGutters` prop for compactness.
```
4ian, line 143: Nice comment, thanks!

newIDE/app/src/UI/Accordion.js
NilayMajorwar: Oh, I actually intended to use the `IconButton` implemented in newIDE, but I guess Flow auto-import picked the other one and I never noticed. I'll fix it.

NilayMajorwar:
Worked on the UI a little bit, how about this:

https://user-images.githubusercontent.com/18032938/114670589-cb48eb00-9d20-11eb-83f3-b4929699537b.mp4

- Changed to a flat UI instead, so shadows and elevations are replaced by simple borders.
- Accordions are now a bit separated from each other. Feels more intuitive to me.
- For now, I moved the "add vertex" button to the accordion header instead. I'm not completely in favour of this, since it adds the risk of misclicking and deleting the whole polygon (a separator helps, but not very much if you're working fast). I didn't like the plus-in-the-corner below the table, so I'm thinking of an alternative.
- Added a more explicit "+ Add collision mask" button below the accordions. Maybe something similar can be done for "add vertex" too?

I'll push the code in some time.

Bouh:
I like the change, it's more spaced and easier to read.
For the warning icon, there is an issue the icon isn't related to the number of vertices, but to the concave shape, this visual is missleading.
image

My suggestion:
image


Orange color come from 
```JS
import GDevelopThemeContext from '......../Theme/ThemeContext';
const gdevelopTheme = React.useContext(GDevelopThemeContext);
```
`gdevelopTheme.message.warning`

NilayMajorwar:
Ah yes, that's definitely misleading. Your suggestion seems great, I'll implement it and see how it looks.

Bouh:
image


The trash bin icon is too near of the arrow of the accordions, it better to swipe the icons add and remove.
Because user can be anoyed if he miss click on this remove button instead to use the arrow or the add button.

4ian:
For "Add a vertex" button, it's inside the accordion header meaning that it's visible even when the accordion section is collapsed: I think it's a bit weird/confusing? If the accordion is collapsed, you can still click on it and... nothing visible happens.

In most existing lists in the app, the "+ Add ..." button is in the bottom of the list, so I think it would make sense to keep it inside the section, at the end of the table?

> Added a more explicit "+ Add collision mask" button below the accordions. Maybe something similar can be done for "add vertex" too?

This could almost be made a primary button, though not sure about this.

NilayMajorwar:
There was a little issue with the "Polygon is not convex" suggestion: the user won't see it if the accordion is closed. I've added a similar warning icon just before the trash bin icon.

image

> If the accordion is collapsed, you can still click on it and... nothing visible happens.

Right, that definitely makes it very suboptimal. I'll add a bottom-of-the-list "add vertex" button.

> This could almost be made a primary raised button, though not sure about this.

Primary raised button is used for this purpose in "Add an animation" (SpriteEditor) and "Add a behavior to the object" (BehaviorsEditor). So I think we should use the same here (as Bouh suggested too).

NilayMajorwar:
Moved the "add a vertex" button to below the vertex list:
image

NilayMajorwar:
So I think the collision masks editor is in a good place now:
image

I was thinking of copying over the "Add a vertex" button to the points editor too:
image

Is it good? While it looked good in the collision masks editor, it looks a bit too dominant here. What do you think?
And finally, what else do you think can be improved in the editors? The PR task list is complete now :D

Bouh:
> And finally, what else do you think can be improved in the editors? The PR task list is complete now :D

Nothing more, just this line height seems odd to me.
image

The primary button for "+ADD A POINT" is fine, it is the primary action when we edit point when it's ok to me.

NilayMajorwar:
> this line height seems odd to me

Okay, that happens because the point row for "Origin" doesn't have any icon. In the other rows, the icon makes the row height larger. I'll try to fix it.

NilayMajorwar:
@Bouh Done. The rows are a bit taller now.
image

Bouh:
I see the preview background has been changed on Physics2 behavior too.
Something has been change on the opacity on the shape?
It is very opaque compared to the collision mask.
image


Otherwise I made a trip in the application and it is clean  👍 
Thanks for this wonderful refactoring!

NilayMajorwar:
Yup, the overlay opacity of Physics2 editor is higher than collision mask editor. I haven't changed anything here, it's like this on `master` too. I can reduce the opacity in Physics2 if you want.

4ian:
So this is some seriously awesome work! More flow typing, storybook for the components, components refactored to function components, and of course an improved user experience. Love what we have now :) 

A few things that I've changed:
* Moved the warning to a AlertMessage component. Enables us to have a slightly longer description, rather than trying to fit everything in a table (was not looking great on small screens otherwise)
* Increased the width of the fields to enter the X/Y position, as we now have nothing else to show, so we can take advantage of a larger width :)

image

Also removed some triple `<Spacer />`, in general I try to keep the layouts as simple and without "spacing" hacks so that there is consistency across dialogs and we can update the whole margins of the app at once one day (or by introducing/adapting UI or Layout components).

I'll merge once the tests are passing. Thanks again for this high quality work, both on the UI/UX and on the code :)

4ian (approved):

tristanbob:
I have to start by saying, these improvements are wonderful and I am very thankful for them.  Thank you so much!

I noticed that the "ADD A POINT" text is 1 or 2 pixels too high.  It needs to be moved down in order to be centered vertically in the button.

4ian:
I think this might be a rendering artifact on Nilay's computer/browser? All buttons should follow the exact same styling :)

NilayMajorwar:
> So this is some seriously awesome work! More flow typing, storybook for the components, components refactored to function components, and of course an improved user experience. Love what we have now :)

Yup - I expected this to be just making the panes side-by-side, but the extra suggestions were absolutely spot-on! Ended up taking more time than I thought, but the improvements are quite worth it.

> I think this might be a rendering artifact on Nilay's computer/browser? All buttons should follow the exact same styling :)

Yeah, noticed it now. All (or most) fonts are rendered slightly taller on Firefox - for example even the tab bars are a bit taller on Firefox. So the button text is a bit taller too, making it look like the text is offset (center-align should have prevented this, though). I'll try to see if it's possible to fix this in the code.

PR #2484, opened 2021-03-29
Add a new preference for save or dismiss change on a dialog when a backdrop click happen.

### LINKED ISSUE #2482: Prevent users from losing work when pop-up windows close
## Description

Many people (including myself) have reported frustration when they have entered data into a pop-up window (such as editing an event) only to have that work discarded because the user accidentally clicked outside the pop-up window when trying to click "OK".

Forum discussion: https://forum.gdevelop-app.com/t/improving-gdevelop-usability/8245/183

## Solution suggested

Option 1:  (Preferred)
- When someone clicks outside a pop-up window, change the default behavior from "Cancel" to "OK".  

This will prevent users from losing work.  If a user really wants to cancel the changes, they can always use the "Cancel" button or hit "Escape" on the keyboard.  

From a probability point of view, GDevelop users save their changes way more frequently when compared to canceling their changes.  That is why this makes a good default behavior.  

From an impact point of view, a user who loses the work they have entered has no options to get that work back.  The same is not true for users who want to cancel a change and are surprised that clicking outside the pop-up saved the change.  In this case, the user can simply use the Undo function to revert the unintended change.

## Alternatives considered 

Option 2: (Not preferred)
- Don't take any action when clicking outside a pop-up window. Force users to click "OK" or "Cancel" (or use the Escape key).

Option 3: (Not preferred)
- If a user has made changes and clicks outside a pop-up window, bring up a dialogue that asks if they want to save the changes or not.

Bouh:
Close #2482

Add a new option in preferences for save or cancel change when backdrop click is used in a Dialog component.

Checklist: 
- [x] Adapt <Dialog>
   - [x] Add onApply props
   - [x]  See for onApply on some <Dialog>
   - [x] Add onClose and use second arg
- [x] See for proper names on variable and option.
- [x]  Remove console.log()


newIDE/app/src/UI/Dialog/index.js
```diff
@@ -102,8 +110,25 @@ export default (props: Props) => {
     <ResponsiveWindowMeasurer>
       {size => (
         <Dialog
+          onApply={onApply}
```
4ian, line 113: This line is doing nothing :) `Dialog` here is a material-ui Dialog, which [has no `onApply` prop to be passed](https://material-ui.com/api/dialog/). 

onApply here is a prop **passed by** the component using the Dialog (for example, `NewInstructionEditorDialog.js`)

Bouh: Yeah i've seen that, a <Dialog> from material-ui in another custom <Dialog> 🤨
I've rewrite import with import as for a better understanding.

What is for `onApply` props for a `<Dialog>` like in [NewObjectDialog.js](https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/AssetStore/NewObjectDialog.js)
I think there is no reason to add it here, but in this case i have to check in  `<Dialog>` if the props exist before call it.

4ian: > I've rewrite import with import as for a better understanding.

Ok 👍

> What is for onApply props for a `<Dialog>` like in NewObjectDialog.js
> I think there is no reason to add it here, but in this case i have to check in `<Dialog>` if the props exist before call it.

Correct. There is no reason to add it (the dialog can just be closed or a choice done), so a backdrop click (or Escape) should result in calling onRequestClose (because onApply is undefined).

Bouh:
Great, the next step is on which `<Dialog>` i have too add `onApply`.

4ian:
Rule is pretty "simple" (I think!).
If the dialog has just a close button, no onApply.
If the dialog has just an ok button, no onApply (closing the dialog will do it, there is no "cancel").
If the dialog has an ok and a cancel button, `onApply` must be the same as the `onClick` of the Ok button :)

Bouh:
Yeahh i'm happy it works :D
Let me know for the next changes if needed.

newIDE/app/src/ObjectEditor/ObjectEditorDialog.js
```diff
@@ -66,8 +66,17 @@ const InnerDialog = (props: InnerDialogProps) => {
 
   const EditorComponent = props.editorComponent;
 
+  const _onApply = () => {
```
4ian, line 69: nitpicking: replace the `const _onApply` by `const onApply`.
Underscores are just for "private" functions in a class. Otherwise, just use `camelCase` for naming your functions and variables.

newIDE/app/src/ObjectGroupEditor/ObjectGroupEditorDialog.js
```diff
@@ -32,6 +32,7 @@ const ObjectGroupEditorDialog = ({
 
   return (
     <Dialog
+      onClick={onApply}
```
4ian, line 35: onApply?

newIDE/app/src/PlatformSpecificAssetsEditor/PlatformSpecificAssetsDialog.js
```diff
@@ -250,6 +250,7 @@ export default class PlatformSpecificAssetsDialog extends React.Component<
 
     return (
       <Dialog
+        onClick={this._onApply}
```
4ian, line 253: onApply?

newIDE/app/src/EventsSheet/InstructionEditor/EventTextDialog.js
```diff
@@ -54,180 +48,174 @@ export const filterEditableWithEventTextDialog = (
...
+  const [backgroundColor, setBackgroundColor] = React.useState<RGBColor>(black);
+
+  const eventType = event.getType(); // OK
+
+  React.useEffect(
+    () => {
+      if (eventType === 'BuiltinCommonInstructions::Comment') {
+        const commentEvent = gd.asCommentEvent(event);
+
+        // Les couleurs provenant normalement de l'event ne sont pas bonne.
```
4ian, line 65: I don't understand this comment?

newIDE/app/src/EventsSheet/InstructionEditor/EventTextDialog.js
```diff
@@ -54,180 +48,174 @@ export const filterEditableWithEventTextDialog = (
...
+      } else {
+        console.error(
+          'Dialog was opened for an unsupported event type: ' + eventType
+        );
+      }
+    },
+    [event, eventType]
+  );
+
+  const _onApply = React.useCallback(
```
4ian, line 108: Same as for the rest, don't use underscore in function names, just `onApply`.

4ian:
Nice! 
Apart from a few nitpickings, take a look at Flow results that will tell you wrong property names (onClick instead of onApply): https://semaphoreci.com/4ian/gd/branches/pull-request-2484/builds/11
Also for a type that you don't care about, use `any`.

I also suggest we give the choice by storing in the preferences something called `backdropClickBehavior` that can be either: `cancel`, `apply`, `nothing`. 
`cancel` will call `onRequestClose`, `apply` would call `onApply` (it it exists, otherwise `onRequestClose`) and nothing would mean that we don't do anything in case of backdrop click :) 

Should not be too hard:
- Use a `SelectField` instead of the toggle.
- Don't forget to update the preferences default values and typing.
- Compare the value in the Dialog when a backdrop click is detected to know what to do :) 

That should give the full choice to the user! :)

newIDE/app/src/UI/Dialog/index.js
```diff
@@ -101,9 +108,23 @@ export default (props: Props) => {
...
           open={open}
-          onClose={onRequestClose}
+          onClose={(event: any, reason: string) => {
+            if (reason === 'escapeKeyDown') {
+              if (onRequestClose) onRequestClose();
+            }
+
+            if (reason === 'backdropClick') {
+              if (backdropClickBehavior === 'cancel') {
+                if (onRequestClose) onRequestClose();
```
Bouh, line 120: Necessary condition, if not there is a flow error

newIDE/app/src/UI/Dialog/index.js
```diff
@@ -101,9 +108,23 @@ export default (props: Props) => {
   return (
     <ResponsiveWindowMeasurer>
       {size => (
-        <Dialog
+        <DialogMaterialUI
           open={open}
-          onClose={onRequestClose}
+          onClose={(event: any, reason: string) => {
+            if (reason === 'escapeKeyDown') {
+              if (onRequestClose) onRequestClose();
```
Bouh, line 115: Necessary condition, if not there is a flow error

newIDE/app/src/UI/Dialog/index.js
```diff
@@ -101,9 +108,23 @@ export default (props: Props) => {
...
+            }
+
+            if (reason === 'backdropClick') {
+              if (backdropClickBehavior === 'cancel') {
+                if (onRequestClose) onRequestClose();
+              }
+              if (backdropClickBehavior === 'apply') {
+                if (onApply) onApply();
```
Bouh, line 123: Necessary condition, if not there is a flow error

newIDE/app/src/EventsSheet/InstructionEditor/EventTextDialog.js
```diff
@@ -54,180 +48,171 @@ export const filterEditableWithEventTextDialog = (
...
+
+        //Text color for group not supported in Core, instead GroupEvent.js handle this
+        //Background color
+        gd.asGroupEvent(event).setBackgroundColor(
+          backgroundColor.r,
+          backgroundColor.g,
+          backgroundColor.b
+        );
+      }
+      props.onApply();
```
Bouh, line 135: `onApply` is used with `props` because if not flow show an error, because `onApply` is already defined at line 105.

4ian: Yes that's fine, these are two different onApply.

4ian (approved):
Seems good to go! Did a few fixes/added translation markers

4ian:
You should be able to request the bounty on bounty source I think? Let me know :)

PR #2488, opened 2021-03-30
Fix the chat in the p2p example

arthuro555:
The chat had inconsistencies in sanitation, and was overflowing out of the screen even though there was an event supposed to prevent that.


arthuro555:
I actually didn't use the same extension in the other example, I used a better one made by add 😅 I believe the one here is called `Sanitizer` already

4ian:
Ah right, Sanitizer is fine 👍

4ian (approved):

PR #2501, opened 2021-04-02
Improve wording on action sentences (s/to/toward/)

tristanbob:
Changed action sentences to indicate movement toward a position/object. The previous version gave the impression that the action would instantly move the object to the position/object.


4ian:
Thanks! 👍

PR #2525, opened 2021-04-09
Small wording on button

Bouh:
Context : 
image
image


Silver-Streak (approved):
Not 100% sure I am intended to be doing reviews, but from a quick check this is a simple text label change, and matches our conversation on the discord. I think the updated button label makes much more sense and adds consistency across the UI.

@4ian let me know if you have any concerns with me using my ability to approve in the future. (I don't plan to make it a habit. 😄 )

4ian:
>  let me know if you have any concerns with me using my ability to approve in the future. (I don't plan to make it a habit. 😄 )

For fairly trivial changes like this it's fine, as long as it was discussed beforehand (such small changes can sometimes create more confusion, but in this case I agree it's justified) 👍

So seems all good 👍

PR #2529, opened 2021-04-10
Test that a platformer object falls when a platform is removed

D8H:
A user reported an issue on Discord about a player staying still in the air when a platform was removed.
It may not be a bug because Silver-Streak couldn't reproduce it manually and neither could the unit test.
But it seems like a good test case.


4ian:
Great test case to have indeed!

4ian (approved):

PR #2530, opened 2021-04-10
A more advance Bomberman-like example with level selection and bonuses

D8H:
I wonder if the source is too big. I tried to put interesting comments. It may be useful for users who already know a bit about GDevelop.
I put a French translation of the source. It won't show on GDevelop but user can still open it after.
I made the sprites (well, clearly I'm not a graphic designer) and leave the SVG.

https://www.dropbox.com/s/tnvg04bgb2ztfrs/SansFoieNieLOie.zip?dl=1

65d99522658f6839a5062cbe9498143ba791fa82


4ian:
It looks nice!
I've fixed a few typo/mistakes ("PlayerMovement", not "PlayerMouvement", "sure" not "sur"), but apart from that looks well done! 👍
More advanced indeed that other examples (maybe some things could have been put in custom behaviors), but the comments are very clear.

I went ahead and made a commit myself: 1ad0ab9881ccc6556ec3cad512babee917b64ade
Thanks!

PR #2532, opened 2021-04-11
Make sure the object inline popover selects a matching behavior

arthuro555:
Thanks @Silver-Streak for finding the bug and reporting it on discord.

When selecting an object using the inline popover on a behavior instruction, the behavior parameter is not updated by calling `setupInstructionParameters`. This caused problems like conditions never working when selecting the instruction without an object and setting the object later via the inline popover.

### DIFF
--- a/newIDE/app/src/EventsSheet/InlineParameterEditor.js
+++ b/newIDE/app/src/EventsSheet/InlineParameterEditor.js
@@ -9,6 +9,7 @@ import {
 import { type ResourceExternalEditor } from '../ResourcesList/ResourceExternalEditor.flow';
 import { type EventsScope } from '../InstructionOrExpression/EventsScope.flow';
 import { setupInstructionParameters } from '../InstructionOrExpression/SetupInstructionParameters';
+import { getObjectParameterIndex } from '../InstructionOrExpression/EnumerateInstructions';
 const gd: libGDevelop = global.gd;
 
 type Props = {|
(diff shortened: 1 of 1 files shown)


arthuro555:
Do not merge, I just found a weird bug with this PR, not sure why that happens, but if I open the popover of an object and then click on another parameter of the same instruction, the object name is replaced by the value of the parameter selected.

newIDE/app/src/EventsSheet/InlineParameterEditor.js
```diff
@@ -108,13 +109,25 @@ export default class InlineParameterEditor extends React.Component<
...
     if (!instruction) return null;
 
+    const onRequestClose = () => {
+      if (this.state.instructionMetadata)
+        setupInstructionParameters(
+          this.props.globalObjectsContainer,
+          this.props.objectsContainer,
+          instruction,
+          this.state.instructionMetadata,
+          instruction.getParameter(this.props.parameterIndex)
```
4ian, line 119: I think this should be: `objectParameterIndex !== -1 ? instruction.getParameter(objectParameterIndex) : null`

with 
```
const objectParameterIndex = objectName
      ? getObjectParameterIndex(instructionMetadata)
      : -1;
```

because otherwise you risk changing the object name with whatever the parameter is being edited?

4ian:
> Do not merge, I just found a weird bug with this PR, not sure why that happens, but if I open the popover of an object and then click on another parameter of the same instruction, the object name is replaced by the value of the parameter selected.

Ah just saw this, yes this is because you're passing the currently edited parameter as the object name. Instead, I think you should be able to use `getObjectParameterIndex` to get it, if it exists. See my other comment :)

newIDE/app/src/EventsSheet/InlineParameterEditor.js
```diff
@@ -108,13 +110,31 @@ export default class InlineParameterEditor extends React.Component<
...
+        const objectParameterIndex = getObjectParameterIndex(
+          this.state.instructionMetadata
+        );
+        setupInstructionParameters(
+          this.props.globalObjectsContainer,
+          this.props.objectsContainer,
+          instruction,
+          this.state.instructionMetadata,
+          objectParameterIndex
+            ? instruction.getParameter(objectParameterIndex)
```
4ian, line 124: Have you checked on multiple conditions and actions? This will crash if there are no object parameter.

4ian:
Done in 606143ff36d3e5f75070dcc469a9819a3eb4bd3f ;)

PR #2533, opened 2021-04-11
Fix autoPlay on video object

Bouh:
Since few beta the autoplay on video object is broken because the option has moved.

#Current
[The autoplay is broken in current beta 108.
](https://editor.gdevelop-app.com/?project=example://video-player
)
#Fix
[The autoplay is now correctly setup on false as before.
](https://games.gdevelop-app.com/game-4840c7a3-f60c-482d-886e-2900ce4e996b/index.html
)
- [x] Tested on desktop with PixiJS v6.0.1


4ian:
Thanks for the fix 👍

PR #2534, opened 2021-04-11
Invalidate hit boxes in TestRuntimeObject setters

D8H:
In the tests a PR (https://github.com/4ian/GDevelop/pull/2153) I had to move an object back and forth for the hit boxes to update after a dimension change.


4ian:
Good catch 👍

PR #2540, opened 2021-04-12
Add version number of the extension.

Bouh:
Add version number of the extension in extension store.
image


newIDE/app/src/AssetStore/ExtensionStore/ExtensionInstallDialog.js
```diff
@@ -141,6 +141,9 @@ export default class ExtensionInstallDialog extends Component<Props, State> {
               <Text noMargin size="title">
                 {extensionShortHeader.fullName}
               </Text>
+              <Text noMargin size="body2">
+                <Trans>Version:</Trans> {' ' + extensionShortHeader.version}
```
4ian, line 145: You an actually put he { } inside the Trans tag, as this will create a placeholder so that language so that have a different order can translate things in the proper order:

```suggestion
                <Trans>Version {' ' + extensionShortHeader.version}</Trans>
```

4ian (approved):

PR #2553, opened 2021-04-16
Rework the behaviors list

nilaymaj:
Related Trello card: https://trello.com/c/Y5B1X0ww

This PR is to rework the behavior list, adding an accordion for each behavior. Panel persistence is also suggested, I'm not very familiar with how to implement that. If there's another IDE configuration setting persisted that is project-specific (and object-specific?), I can probably take a look at that.

As for implementation, the main part is almost trivial now. Accordion is already implemented, just need to move things around in BehaviorsList a bit. I'll get started with this today evening.

EDIT: (sorry for the requested review, forgot to mark this as draft PR)

EDIT: Also, I read somewhere about an alternate UI too (on the forum, maybe) - similar to the asset store, list of behaviors in left pane, and on selecting a behavior the properties appear in the right pane. Not sure which one is more intuitive or productive - I'm not so familiar with the user workflow in GDevelop.

### DIFF
--- a/newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/index.js
+++ b/newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/index.js
@@ -51,7 +51,7 @@ const BitGroupEditor = (props: {|
 |}) => {
   return (
     <div style={{ overflowX: 'auto', flex: 1 }}>
-      <ButtonGroup variant="contained" disableElevation fullWidth>
+      <ButtonGroup disableElevation fullWidth>
         {props.bits.map((bit, index) => (
           <Button
             key={index}
--- a/newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/index.js
+++ b/newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/index.js
@@ -368,6 +368,12 @@ const Physics2Editor = (props: Props) => {
               resourcesLoader={resourcesLoader}
               fixedHeight={200}
               renderOverlay={overlayProps => {
+                // The result from `getProperties` is temporary, and because this renderOverlay
+                // function can be called outside of the render, we must get the properties again.
+                const properties = behavior.getProperties(
+                  behaviorContent.getContent()
+                );
+
                 return (
                   <ShapePreview
                     shape={properties.get('shape').getValue()}
--- a/newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/index.js
+++ b/newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/index.js
@@ -90,7 +90,11 @@ const Physics2Editor = (props: Props) => {
   const masksValues = parseInt(properties.get('masks').getValue(), 10);
 
   return (
-    <Column expand>
+    <Column
+      expand
+      // Avoid overflow on small screens
+      noOverflowParent
+    >
       <Line>
         <SelectField
           key={'bodyType'}
--- a/newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/index.js
+++ b/newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/index.js
@@ -88,13 +88,17 @@ const Physics2Editor = (props: Props) => {
 
   const properties = behavior.getProperties(behaviorContent.getContent());
   console.log(
-    properties.get('vertices').getValue(),
-    typeof properties.get('vertices').getValue()
+    'From Physics2 behavior:',
+    behavior.ptr,
+    behaviorContent.ptr,
+    properties.ptr
   );
   const bits = Array(16).fill(null);
   const shape = properties.get('shape').getValue();
   const layersValues = parseInt(properties.get('layers').getValue(), 10);
   const masksValues = parseInt(properties.get('masks').getValue(), 10);
+
+  console.log('Physics2 editor rendered!');
 
   return (
     <Column expand>
@@ -364,10 +368,6 @@ const Physics2Editor = (props: Props) => {
               resourcesLoader={resourcesLoader}
               fixedHeight={200}
               renderOverlay={overlayProps => {
-                console.log(
-                  properties.get('vertices').getValue(),
-                  typeof properties.get('vertices').getValue()
-                );
                 return (
                   <ShapePreview
                     shape={properties.get('shape').getValue()}
--- a/newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/index.js
+++ b/newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/index.js
@@ -19,6 +19,8 @@ import DismissableAlertMessage from '../../../UI/DismissableAlertMessage';
 import { ResponsiveLineStackLayout } from '../../../UI/Layout';
 import EmptyMessage from '../../../UI/EmptyMessage';
 import useForceUpdate from '../../../Utils/UseForceUpdate';
+import Button from '@material-ui/core/Button';
+import ButtonGroup from '@material-ui/core/ButtonGroup';
 
 type Props = BehaviorEditorProps;
 
@@ -43,29 +45,24 @@ const NumericProperty = (props: {|
   );
 };
 
-const BitProperty = (props: {|
-  enabled: boolean,
-  propertyName: string,
-  pos: number,
-  spacing: boolean,
-  onUpdate: (enabled: boolean) => void,
+const BitGroupEditor = (props: {|
+  bits: Array<boolean>,
+  onChange: (index: number, value: boolean) => void,
 |}) => {
-  const { enabled, propertyName, pos, spacing, onUpdate } = props;
-
   return (
-    <div style={{ width: spacing ? '7.5%' : '5%' }}>
-      {
-        <Checkbox
-          checked={enabled}
-          onCheck={(e, checked) => onUpdate(checked)}
-        />
-      }
-      {spacing && (
-        <div
-          style={{ width: '33%' }}
-          key={propertyName + '-space' + pos.toString(10)}
-        />
-      )}
+    <div style={{ overflowX: 'auto', flex: 1 }}>
+      <ButtonGroup variant="contained" disableElevation fullWidth>
+        {props.bits.map((bit, index) => (
+          <Button
+            key={index}
+            variant={bit ? 'contained' : 'outlined'}
+            color={bit ? 'primary' : 'default'}
+            onClick={() => props.onChange(index, !bit)}
+          >
+            {index + 1}
+          </Button>
+        ))}
+      </ButtonGroup>
     </div>
   );
 };
@@ -541,50 +538,38 @@ const Physics2Editor = (props: Props) => {
         />
       </ResponsiveLineStackLayout>
       <Line>
-        <Text>{properties.get('layers').getLabel()}</Text>
-        {bits.map((value, index) => {
-          return (
-            <BitProperty
-              enabled={isBitEnabled(layersValues, index)}
-              propertyName={'layers'}
-              pos={index}
-              spacing={index === 7}
-              onUpdate={enabled => {
-                const newValue = enableBit(layersValues, index, enabled);
-                behavior.updateProperty(
-                  behaviorContent.getContent(),
-                  'layers',
-                  newValue.toString(10)
-                );
-                forceUpdate();
-              }}
-              key={`layer${index}`}
-            />
-          );
-        })}
+        <Text style={{ marginRight: 10 }}>
+          {properties.get('layers').getLabel()}
+        </Text>
+        <BitGroupEditor
+          bits={bits.map((_, idx) => isBitEnabled(layersValues, idx))}
+          onChange={(index, value) => {
+            const newValue = enableBit(layersValues, index, value);
+            behavior.updateProperty(
+              behaviorContent.getContent(),
+              'layers',
+              newValue.toString(10)
+            );
+            forceUpdate();
+          }}
+        />
       </Line>
       <Line>
-        <Text>{properties.get('masks').getLabel()}</Text>
-        {bits.map((value, index) => {
-          return (
-            <BitProperty
-              enabled={isBitEnabled(masksValues, index)}
-              propertyName={'masks'}
-              pos={index}
-              spacing={index === 7}
-              onUpdate={enabled => {
-                const newValue = enableBit(masksValues, index, enabled);
-                behavior.updateProperty(
-                  behaviorContent.getContent(),
-                  'masks',
-                  newValue.toString(10)
-                );
-                forceUpdate();
-              }}
-              key={`mask${index}`}
-            />
-          );
-        })}
+        <Text style={{ marginRight: 10 }}>
+          {properties.get('masks').getLabel()}
+        </Text>
+        <BitGroupEditor
+          bits={bits.map((_, idx) => isBitEnabled(masksValues, idx))}
+          onChange={(index, value) => {
+            const newValue = enableBit(masksValues, index, value);
+            behavior.updateProperty(
+              behaviorContent.getContent(),
+              'masks',
+              newValue.toString(10)
+            );
+            forceUpdate();
+          }}
+        />
       </Line>
     </Column>
   );
--- a/newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/index.js
+++ b/newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/index.js
@@ -87,6 +87,10 @@ const Physics2Editor = (props: Props) => {
   };
 
   const properties = behavior.getProperties(behaviorContent.getContent());
+  console.log(
+    properties.get('vertices').getValue(),
+    typeof properties.get('vertices').getValue()
+  );
   const bits = Array(16).fill(null);
   const shape = properties.get('shape').getValue();
   const layersValues = parseInt(properties.get('layers').getValue(), 10);
@@ -359,43 +363,49 @@ const Physics2Editor = (props: Props) => {
               project={props.project}
               resourcesLoader={resourcesLoader}
               fixedHeight={200}
-              renderOverlay={overlayProps => (
-                <ShapePreview
-                  shape={properties.get('shape').getValue()}
-                  dimensionA={parseFloat(
-                    properties.get('shapeDimensionA').getValue()
-                  )}
-                  dimensionB={parseFloat(
-                    properties.get('shapeDimensionB').getValue()
-                  )}
-                  offsetX={parseFloat(
-                    properties.get('shapeOffsetX').getValue()
-                  )}
-                  offsetY={parseFloat(
-                    properties.get('shapeOffsetY').getValue()
-                  )}
-                  polygonOrigin={properties.get('polygonOrigin').getValue()}
-                  vertices={JSON.parse(properties.get('vertices').getValue())}
-                  width={overlayProps.imageWidth}
-                  height={overlayProps.imageHeight}
-                  frameOffsetTop={overlayProps.offsetTop}
-                  frameOffsetLeft={overlayProps.offsetLeft}
-                  zoomFactor={overlayProps.imageZoomFactor}
-                  onMoveVertex={(index, newX, newY) => {
-                    let vertices = JSON.parse(
-                      properties.get('vertices').getValue()
-                    );
-                    vertices[index].x = newX;
-                    vertices[index].y = newY;
-                    behavior.updateProperty(
-                      behaviorContent.getContent(),
-                      'vertices',
-                      JSON.stringify(vertices)
-                    );
-                    forceUpdate();
-                  }}
-                />
-              )}
+              renderOverlay={overlayProps => {
+                console.log(
+                  properties.get('vertices').getValue(),
+                  typeof properties.get('vertices').getValue()
+                );
+                return (
+                  <ShapePreview
+                    shape={properties.get('shape').getValue()}
+                    dimensionA={parseFloat(
+                      properties.get('shapeDimensionA').getValue()
+                    )}
+                    dimensionB={parseFloat(
+                      properties.get('shapeDimensionB').getValue()
+                    )}
+                    offsetX={parseFloat(
+                      properties.get('shapeOffsetX').getValue()
+                    )}
+                    offsetY={parseFloat(
+                      properties.get('shapeOffsetY').getValue()
+                    )}
+                    polygonOrigin={properties.get('polygonOrigin').getValue()}
+                    vertices={JSON.parse(properties.get('vertices').getValue())}
+                    width={overlayProps.imageWidth}
+                    height={overlayProps.imageHeight}
+                    frameOffsetTop={overlayProps.offsetTop}
+                    frameOffsetLeft={overlayProps.offsetLeft}
+                    zoomFactor={overlayProps.imageZoomFactor}
+                    onMoveVertex={(index, newX, newY) => {
+                      let vertices = JSON.parse(
+                        properties.get('vertices').getValue()
+                      );
+                      vertices[index].x = newX;
+                      vertices[index].y = newY;
+                      behavior.updateProperty(
+                        behaviorContent.getContent(),
+                        'vertices',
+                        JSON.stringify(vertices)
+                      );
+                      forceUpdate();
+                    }}
+                  />
+                );
+              }}
             />
           </div>
         </Line>
--- a/newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/index.js
+++ b/newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/index.js
@@ -87,18 +87,10 @@ const Physics2Editor = (props: Props) => {
   };
 
   const properties = behavior.getProperties(behaviorContent.getContent());
-  console.log(
-    'From Physics2 behavior:',
-    behavior.ptr,
-    behaviorContent.ptr,
-    properties.ptr
-  );
   const bits = Array(16).fill(null);
   const shape = properties.get('shape').getValue();
   const layersValues = parseInt(properties.get('layers').getValue(), 10);
   const masksValues = parseInt(properties.get('masks').getValue(), 10);
-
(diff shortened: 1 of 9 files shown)


4ian:
> EDIT: Also, I read somewhere about an alternate UI too (on the forum, maybe) - similar to the asset store, list of behaviors in left pane, and on selecting a behavior the properties appear in the right pane

This was suggested by arthuro555 I believe, and I think it's a more complex approach that consists in using vertical tabs (or checkboxes, but these should not be used for navigation, or a tree view, but behaviors are not a tree) for navigating long lists. I'm not sure it's really necessary to have this - the interesting part of the proposal was the search box though.

But I believe we should go ahead with a more simple refactoring like the accordions. 

> If there's another IDE configuration setting persisted that is project-specific (and object-specific?), I can probably take a look at that.

There is nothing like this for now - arthuro555 was also mentioning the possibility of storing these preferences in a separate file alongside with the project (that can be gitignored for teams of people working on a game).
I'm not sure what's the best thing for now is, should we store it in the gdProject in some UI specific set up like done for example for the scene camera position/grid, and then when saving a project, split them into separate files - or use a different structure entirely to store the preferences (but with the risk of having to maintain stuff like the object name if preferences are indexed by object name, so you have to handle this, as well as deletion, scene renaming etc...).

There also, I would say not to care too much about persistence for now.

Bouh:
Is the open/closed state is still active even if the dialog component is reopened several minutes later?
I mean if the state is persistent only during the use of the project that's fine for now.

I like the saving preferences in an external file in the project folder.
I always hide the grid in a project for example, and almost every project I test I have the grid visible again, it's really a project setting, like accordions.
A new project folder tree is something that becomes important.


```
My game folder
	|- Settings
	|- Autosave
	|- Assets
	|     |- Image
	|     |- Video
	|     |- Font
	|     |    |- BitmapFont
	|     |    |- Classic font
	|     |
	|     |- Audio
	|          |- Music
	|          |- SFX
```

nilaymaj:
> This was suggested by arthuro555 I believe, and I think it's a more complex approach that consists in using vertical tabs (or checkboxes, but these should not be used for navigation, or a tree view, but behaviors are not a tree) for navigating long lists. I'm not sure it's really necessary to have this - the interesting part of the proposal was the search box though.
>
> But I believe we should go ahead with a more simple refactoring like the accordions.

Yeah, my primary concern with the vertical-tabs design was that most objects have a maximum of 3-4 behaviors (?), so a list of tabs for behaviors will look a bit odd unless the tabs are large with good padding. Also, not sure what should be the fallback on smaller screens in that case. 

But then also, vertical tabs will be undoubtedly quicker to work with than both the current design and the proposed accordions design. Everything is almost always visible and accessible, scrolling will be rarely required.

I don't remember the search box suggestion very well, and couldn't find the forum thread. Is it for searching into the properties of all behaviors (like VSCode settings search, where results include extensions settings too)?

> arthuro555 was also mentioning the possibility of storing these preferences in a separate file alongside with the project (that can be gitignored for teams of people working on a game)

I definitely think this is the ideal solution.

> I'm not sure what's the best thing for now is, should we store it in the gdProject in some UI specific set up like done for example for the scene camera position/grid, and then when saving a project, split them into separate files - or use a different structure entirely to store the preferences (but with the risk of having to maintain stuff like the object name if preferences are indexed by object name, so you have to handle this, as well as deletion, scene renaming etc...).

The first one looks better - the second seems a bit tricky to do right and maintain. But the second one has the benefits of proper separation of concerns. I think this needs a bit of discussion, and then later once a robust project-specific preference system is set up, persistence can be added to the UI wherever useful. 

Basically I want to avoid a temporary solution here, and would prefer to implement a proper project-specific preference storage before adding more persistence features to the IDE. So for now I'll focus on the accordion UI work without worrying too much about persistence, as you said.

> Is the open/closed state is still active even if the dialog component is reopened several minutes later?
I mean if the state is persistent only during the use of the project that's fine for now.

Yeah that would be great. Not sure though about how to implement this without storing stuff in gdProject like the scene editor does it (which is something I'm trying to avoid right now :sweat_smile: )

Bouh:
> > Is the open/closed state is still active even if the dialog component is reopened several minutes later?
> > I mean if the state is persistent only during the use of the project that's fine for now.
> 
> Yeah that would be great. Not sure though about how to implement this without storing stuff in gdProject like the scene editor does it (which is something I'm trying to avoid right now 😅 )

My interogation is invalid if the React virtual DOM saves it during the session.

nilaymaj:
> My interogation is invalid if the React virtual DOM saves it during the session.

Nope, React won't save it. The dialog component will get completely unmounted whenever the dialog is closed, no information about the contents will be saved unless we store it somewhere ourselves.

nilaymaj:
Added accordions to the behavior list:

https://user-images.githubusercontent.com/18032938/115044465-2af6f000-9ef3-11eb-9b09-25be26a80aa7.mp4

- Should the accordions be all expanded by default or collapsed by default? Collapsed makes more sense to me since accordions here are added for the primary purpose of avoiding cluttered UI, so it would be pointless if the user seems the same thing even after adding accordions.
- The dialog height changes a lot when expanding accordions. Should the dialog be made full height?
- Toggling these accordions is a bit harder than the ones in collision masks editor. Clicking on the behavior name text fields (which are disabled) does not trigger the accordion toggle, so there is less space where the user can click to toggle the accordion.

There is a very weird bug I am encountering.
1. Add the Physics2 behavior and set the temporary sprite to anything. You can change the shape and add vertices too, that doesn't matter. What matters is setting the temporary sprite.
2. Add another behavior to the object (any behavior).
3. Toggle the state of the Physics2 behavior accordion, and the editor crashes.

This does not happen if you do not set the temporary image or set it after adding the second behavior (that's how the above clip was recorded). Whether the Physics2 accordion was open or closed before the final toggle does not matter. This bug is strange because rendering the behavior in accordions really shouldn't matter if the accordion body is always rendered - technically it's the same as before.

The code throwing an error:
https://github.com/4ian/GDevelop/blob/32ca97650c8a05ca3c9b0328f109fa7ea14cc914/newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/index.js#L378
`properties.get('vertices').getValue()` is string `[]` uptil the end. Right at the final toggle, it evaluates to an empty string leading to JSON parse error.

It's possible that this bug was introduced when I refactored Physics2Editor into a functional component (in previous PR). But then I can't even test that because it doesn't have any accordion to toggle, I'll try to find out where the problem arises from.

arthuro555:
> arthuro555 was also mentioning the possibility of storing these preferences in a separate file alongside with the project (that can be gitignored for teams of people working on a game)

Btw I also posted a bounty on this if someone is interested.

Bouh:
- > Should the accordions be all expanded by default or collapsed by default? 

   Collapsed when the persistent states will be available, but for now don't bother users to always expand the behaviors.

- > The dialog height changes a lot when expanding accordions. Should the dialog be made full height?

   A big YES, this dialog is always resizing the content on screen move each time we click somewhere it's disconcerting in this dialog and everywhere like the variables tree. 


- Is a new fresh behavior added can be by default expanded ?

- > Toggling these accordions is a bit harder than the ones in collision masks editor. Clicking on the behavior name text fields (which are disabled) does not trigger the accordion toggle, so there is less space where the user can click to toggle the accordion.

   I can, the red area is clickable for toggle the state it's ok for me.
On green area we have a hand cursor due to the accordion. Idk what is expected or what thinking about it.
image


- > > arthuro555 was also mentioning the possibility of storing these preferences in a separate file alongside with the project (that can be gitignored for teams of people working on a game)
   > 
   > Btw I also posted a bounty on this if someone is interested.

   I skip my turn :)

nilaymaj:
Okay, I'll make those changes.

> Is a new fresh behavior added can be by default expanded ?

If the accordions are all expanded by default, this will happen automatically, so yes.

Im travelling today, so wont be able to work on this today.

4ian:
> Collapsed when the persistent states will be available, but for now don't bother users to always expand the behaviors.

+1 to this. Default to expand behaviors, as now.

> Is a new fresh behavior added can be by default expanded ?

+1, it's important not to add additional clicks to reveal stuff. For now the accordion should be seen as a "bonus" allowing to collapse stuff, but everything is expanded by default.

> A big YES, this dialog is always resizing the content on screen move each time we click somewhere it's disconcerting in this dialog and everywhere like the variables tree.

+1, great idea indeed to keep the dialog full height, we should probably do this for any dialog that can have a lot of content.

Thanks 👍

nilaymaj:
Made the changes suggested. Accordions are expanded by default, the dialog is full height. Regarding the accordion clickable area, yeah Bouh is right, the disabled text field allows accordion toggle on Chrome (yet another Firefox-specific issue).

So the UI part seems to be done. Let me know if you want any changes:
image
(the properties tab is also full height now, which is good since it also has lots of content usually)

The bug is still here, I hope to fix it today or tomorrow.

Bouh:
Try to keep a regular padding with the borders
image

nilaymaj:
Fixed the irregular margin, and another bug where hidden tutorial hints were taking up space due to their container margins. It should look okay now:

image

4ian:
Very nice!
I wonder if the margins of the accordion content are a bit too big? Must be 16 or 20px right now? Might be worth trying 8px (i.e: 2x the size of a usual Spacer component)?

nilaymaj:
Right now, the accordion margin is 16px on the sides. The margin between the text fields and accordion border seen in above screenshot is 16px + 8px (by `Column` in behavior properties component). I agree that the Accordion 16px margin is too much - reducing it to 8px (for total space 8px+8px between textfield and accordion border) leads to this (before | after):

image

nilaymaj:
I've been trying to work out the editor-crashing bug, and need a bit of help in understanding what the `ptr` in Emscripten-based objects really means. I'm quite familiar with C++ pointers but obviously this is different.

I observed `behavior.ptr`, `behaviorContent.ptr` and `properties.ptr` (where `properties` is obtained from  `behavior.getProperties(behavior.getContent())`) in two different behavior editor components (a Physics2Editor and a BehaviorPropertiesEditor). The two logged different `ptr`s for `behavior` and `behaviorContent` but the same `ptr` for `properties`. This confused me a bit - what does `ptr` represent, exactly?

Bouh:
I think my [conversation in french with Arthuro on Discord](https://discord.com/channels/258623956158906368/377571085643677696/832676116866924564) can help you. Here a short translation:

- There is a way to read the content from a pointer during a breakpoint?
`this.ptr` is equal to a number
https://media.discordapp.net/attachments/377571085643677696/832676113687904256/unknown.png

- Yes, you can create a new instance of the class, [...] you can do 
`const myObjet = gd.wrapPointer(pointerFromTheBreakpoint, gd.Object)`

or 
image

nilaymaj:
Thanks @Bouh, that helped a bit. But I want to understand more: what are the main similarities and differences between 
- C/C++ pointers: (which point to a memory location or in practice point to a certain code entity) 
and 
- Emscripten `ptr`s: (regular pointers don't make much practical sense in a JS runtime)

About the second screenshot you sent: what is the effect of changing the `ptr` of `monObject`? Does it make `monObject` somehow become an alias to the original object (which had `ptr` 9654816)?

arthuro555:
IIRC emscripten pointers are the same as in C++: there is a js array emulating real memory, and all c++ manipulate it exclusively. The ptr corresponds to the array index, equivalent to the memory address in native environments. That is also why garbage collection doesn't work and we have to call delete: the actual data is not in the object, but in that global "memory object".

nilaymaj:
Ah okay, makes sense.. so practically the JS objects are just like pointers to the "virtual" C++ objects, right? Thanks a lot @arthuro555, clarifies a lot of things for me. I think I should be able to get the root cause of this issue now.

I thought `ptr`s work differently from C++ pointers because `behavior.getProperties(behaviorContent.getContent())` seemed to return the same `ptr` object on each call (and stuff still worked). I realized I made a mistake here, need to dig deeper.

4ian:
Correct, an Emscripten object is just an "empty shell", a JS object with a `ptr` which is a number which points to the index of the C++ object in the WebAssembly memory (that you can think of as a gigantic array of bytes).
- This is why we must call `delete` when an object must be removed (this only applies if you made a new Emscripten object in JS, or if a function call returned to you an object that you "own"). For objects that are created and "owned" in the C++ code, the usual "memory management" (i.e: shared pointers, etc...) applies.
- This is why if you try to call a method on an object that was deleted (or if you change the `ptr` to a random number), you will make the C++ code to use a bad memory area: it will probably crash.

It means that we must be careful with any Emscripten object: it must be clear who owns it and be sure that the object will always exist when handling it, and if destroyed, it must be never used (all references must be dropped).

nilaymaj:
Okay, I think I have a good idea of where the issue is coming from, now:

**Some useful context:** 
- The behavior list renders the behavior editing components for each behavior. 
- Besides Physics2, all behaviors use the generic `BehaviorPropertiesEditor` component. Physics2 has its own `Physics2Editor` component.
- All these component get as props `behavior: gdBehavior` and `behaviorContent: gdBehaviorContent`. Then, they get the properties of the behavior using `behavior.getProperties(behaviorContent.getContent())`.
- While `behavior.getProperties(...)` is implemented separately for each behavior, they all use `new gd.MapStringPropertyDescriptor()` to create the `properties` object which is returned.

**Steps to reproduce the editor-crashing bug:**
> 1. Add the Physics2 behavior and set the temporary sprite to anything. You can change the shape and add vertices too, that doesn't matter. What matters is setting the temporary sprite.
> 2. Add another behavior to the object (any behavior). EDIT: any behavior that appears after the Physics2 behavior (sorted by name). So Tween will give the bug, but Anchor won't.
> 3. Toggle the state of the Physics2 behavior accordion, and the editor crashes.

**Why this crash occurs**:
- Somehow, when calling `behavior.getProperties(...)` for the two behaviors, both of them create `MapStringPropertyDescriptor`s with the same `ptr`. I checked this by logging the `ptr` in implementation of each method in corresponding `JsExtension.js` files. The `ptr` when only one behavior is created is different, but after creating the second behavior the `ptr`s are always the same.
- Interestingly, if I log `properties.ptr` in the behavior editor components (Physics2Editor and BehaviorPropertiesEditor), both give the same ptr **but** different from what was logged in `JsExtension.js`. Not sure why this is happening.
- So in the reproduction steps, after you add another behavior that appears after the Physics2 behavior, the second component runs `const properties = behavior.getProperties(...)` overwriting the common memory pointed to by the two `properties`. The first behavior (Physics2) is already rendered, so the IDE doesn't crash yet.
- Expand/collapse the Physics2 accordion. This does not rerender Physics2Editor (which would've avoided the crash) but instead only re-runs the underlying `renderOverlay` prop of `ImagePreview`. But the `properties` object is overwritten with properties of the other behavior, so `JSON.parse(properties.get('vertices').getValue())` fails since `properties` no longer has a field named `vertices`. JSON parse error -> editor crash.

Sorry for the long description, but at this point I feel a bit lost about how to proceed. I can fix the crash by re-evaluating `properties` in the `renderOverlay` prop, but that seems a bit hacky.

4ian:
> Interestingly, if I log properties.ptr in the behavior editor components (Physics2Editor and BehaviorPropertiesEditor), both give the same ptr but different from what was logged in JsExtension.js. Not sure why this is happening.

oO 
Weird!

Making two different `MapStringPropertyDescription` works normally:
image

Some rare functions, that you can find in `Bindings.idl` by searching for `[Value]`, return an object by value in which case it's internally always the same object. But I'm not sure this is the case for behaviors. I'll check.

4ian:
> I can fix the crash by re-evaluating properties in the renderOverlay prop, but that seems a bit hacky.

I think that's a potential solution.
Basically, anything returned with `[Value]` can not be persisted, as its memory can be replaced by anything else latter. There is a single object in memory that is used for ALL function calls of this function. (a bit hacky, but that's how Emscripten handles this to avoid you to delete objects... not ideal).

Because of this btw, we have a memory leak whenever we do a `new gd.MapStringPropertyDescriptor()`. We should have a single global static object. We can take care of this latter.

For now, consider `getProperties` as *temporary*. Can you redo `getProperties` in every `render` function that uses it?

newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/index.js
```diff
@@ -359,43 +367,45 @@ const Physics2Editor = (props: Props) => {
...
-                      behaviorContent.getContent(),
-                      'vertices',
-                      JSON.stringify(vertices)
-                    );
-                    forceUpdate();
-                  }}
-                />
-              )}
+              renderOverlay={overlayProps => {
+                return (
```
4ian, line 371: Try to `const properties = behavior.getProperties(behaviorContent.getContent());` and add a comment telling that `the result from getProperties is a temporary, and because this renderOverlay function can be called outside of the render, we must get the properties again.`

nilaymaj:
> Basically, anything returned with [Value] can not be persisted, as its memory can be replaced by anything else latter. There is a single object in memory that is used for ALL function calls of this function. (a bit hacky, but that's how Emscripten handles this to avoid you to delete objects... not ideal).

Ah okay, makes sense now why this was happening.

> For now, consider getProperties as temporary. Can you redo getProperties in every render function/function component that uses it?

So basically the return value of `getProperties` should be treated as use-and-throw, right? Memory that is accessible by other code, so is not reliable for "later" use. Yup, will do this. Shouldn't be too many behaviors with this issue.

4ian:
> So basically the return value of getProperties should be treated as use-and-throw, right? Memory that is accessible by other code, so is not reliable for "later" use. Yup, will do this. Shouldn't be too many behaviors with this issue.

Correct :)

You can actually look at the code generated by Emscripten (in glue.cpp) (Emscripten create "C like functions" for each object, to bind them to the JavaScript methods, in glue.js):
image

You can see there is a single `temp` call. In other words, it's like if you had this in JS:

```
const singlePropertiesObject = {};
export function getProperties() {  
  // Delete everything in singlePropertiesObject;
  // Fill up the properties: singlePropertiesObject[...] = ...
  return singlePropertiesObject;
}
```

Sorry about that, this is the weird edge cases of Emscripten/bindings with C++. Not really possible to statically check for this, not sure what's the best solution to avoid these issues later, apart from documenting with comments.

nilaymaj:
Makes sense.

> Sorry about that, this is the weird edge cases of Emscripten/bindings with C++. Not really possible to statically check for this, not sure what's the best solution to avoid these issues later, apart from documenting with comments.

Yeah, I understand. Maybe a utility function that internally calls this `getProperties` function and then maps over the result to get a regular JS object (makes a JS-land copy, basically)? Then the components can use this without worrying about usable lifetime - only code directly using `getProperties` may have to worry, but then those should already be worrying right now!

I've made the fix. I think only Physics2Editor required this, the others use the generic property editor component which doesn't have this issue. Interestingly, it uses a function `propertiesMapToSchema` (defined [here](https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/PropertiesEditor/PropertiesMapToSchema.js)) that roughly does the above but with a few extra perks.

So now, this PR seems good? Should the "Add a behavior to the object" button be center-aligned, like in polygons list and points list? Also, the Physics2 editor gets clipped a little on smaller widths - this happens on `master` too. I can try fixing this too.

4ian:
> So now, this PR seems good? Should the "Add a behavior to the object" button be center-aligned, like in polygons list and points list?

No I would prefer to keep the right alignment everywhere actually - all lists are using this, so better keep the consistency.

> Also, the Physics2 editor gets clipped a little on smaller widths - this happens on master too. I can try fixing this too.

Yeah that was my feeling too.
Would be good to see if we can do something that looks good and not eating too much width space.

nilaymaj:
Okay, it's the layers and masks checkbox group that's causing the overflow. The checkboxes have fixed size and margin, so a fully responsive solution doesn't seem very possible without making the checkboxes smaller. Reducing the margins to 0px works a bit - the clipping starts at screen width 520px while earlier it started from around 750px. This is what it looks like on normal-size screens:

image

I read about what these Layers and Masks checkboxes mean in the engine, and I was thinking of a more intuitive UI for this: a ButtonGroup?

https://user-images.githubusercontent.com/18032938/115199453-d92bb100-a110-11eb-8f28-216fd5d162bb.mp4

(ignore the flickering)

Agreed, this still suffers from the same problem of not being entirely responsive, but it's possible to split this into two rows for smaller screens without confusing the user. What do you think?

4ian:
> I read about what these Layers and Masks checkboxes mean in the engine, and I was thinking of a more intuitive UI for this: a ButtonGroup?

Totally, would look better! And the number are useful to identify the group index.

Bouh:
- These layers and masks checkbox can be flex, on the whole width.
image

- Like the dialog is in full height now we have a large empty space, i think it worth to add a little description of what is a behavior, a bit like on the eventsheet.
image


- For the responsive on a small screen it's fine, and i with said, nobody one GD on a smartphone so..
image

nilaymaj:
Two variants:

image
image

I prefer the first one since it respects the theme more (the second one has light stuff on dark theme, not good). Let me know what you think.

> These layers and masks checkbox can be flex, on the whole width.

Sure, done.

> Like the dialog is in full height now we have a large empty space, i think it worth to add a little description of what is a behavior, a bit like on the eventsheet.

Good idea, I think. You'll have to decide the content, though :D

> For the responsive on a small screen it's fine, and nobody one GD on a smartphone so..

The responsiveness breaks only on the Physics2 behavior. I do agree about smartphone use though - I know a card is on the roadmap and I love how the dialogs change to full-screen overlays on small screens, but still I don't understand making an IDE (especially a game engine with such a variety of UI) *completely* smartphone-friendly. It's extra development effort for nearly zero use-case.

Although in this case it does break usability of Physics2 behavior on mobiles totally, since the clipping does not add a horizontal scrollbar. The content hidden is simply inaccessible. Let me try some fixes - if I find a good and simple one, I'll implement it.

4ian:
Might be a matter of adding some flex-wrap? Also have you used material-ui button group, or something that you did on your own?

It's not that much about smartphones but more about be responsive and support reducing the size of the IDE if you want to have on a screen your game and running and on another side the IDE that is smaller where you can tweak value and apply changes for example :)

nilaymaj:
> Might be a matter of adding some flex-wrap? Also have you used material-ui button group, or something that you did on your own?

I'm using MUI ButtonGroup. I tried `flex-wrap` but it messes up the layout (looks like list of full-width buttons). The only way I got it to work is by redoing the root flex styles of `ButtonGroup` myself. Even then, the wrapping doesn't play well with the ButtonGroup being full-width.

I'm also trying an alternative option - having the ButtonGroup inside a horizontally scrollable div. Haven't had much success here too, but it does seem a better option if it works.

> It's not that much about smartphones but more about be responsive and support reducing the size of the IDE if you want to have on a screen your game and running and on another side the IDE that is smaller where you can tweak value and apply changes for example :)

Ah right, that makes much more sense. I should definitely try to make this responsive, then!

4ian:
> having the ButtonGroup inside a horizontally scrollable div. Haven't had much success here too, but it does seem a better option if it works.

Would be the best I think yes. Not sure but might force you to hardcode a width.
Another alternative is to detect the window size (there is a hook/component for it) and if `small`, then cut the group in ~3? (more work though, would be better to just have an horizontal scroll).

Bouh:
> Good idea, I think. You'll have to decide the content, though :D

```
There are no behavrios here.

Behaviors are predefined actions that are assigned to objects. Behaviors can have no or multiple parameters.

Add your first behavior using the button "Add a behavior to the object".
```

nilaymaj:
Okay, I got to a working version of the horizontally-scrollable ButtonGroup, but it requires a new prop on `Column` for setting `min-width: 0`. Interestingly (and annoyingly) `Column` already has a `useFullHeight` prop used for a different purpose that sets `min-height: 0`.

I know this looks plenty hacky but it does have some reason to it: https://stackoverflow.com/a/36247448/6199068. If I go with this, I have almost no idea what to name it - the best I can think is simply prop `minWidth`. All other `Column`s work as expected so it's not like this prop is intended for some common purpose to name it after.

By the way I've checked this on Chrome too - both have this issue and both are fixed with `min-width: 0` being set on two `Column` components.

nilaymaj:
I've implemented the fix, and named the prop `noOverflowParent` with a hopefully helpful comment alongside. If this fix is okay, I think that wraps up the UI work on this PR. I'm opening this PR for review, but I'll be glad to take a shot at accordion state persistence if you want.

4ian:
Thanks for the Flow typing!
I think this fix make sense, at least with the prop name it's easy to follow in the codebase where it's used.
I'll review again the whole PR :)

nilaymaj:
@4ian just a reminder, in case you forgot. Sorry for the tag if you're busy, ignore this then :sweat_smile:

newIDE/app/src/UI/Theme/CreateTheme.js
```diff
@@ -177,6 +177,19 @@ export function getMuiOverrides(
         padding: 0,
       },
     },
+    MuiAccordionDetails: {
+      root: {
+        padding: 8,
+      },
+    },
+    // Avoid double border between buttons in ButtonGroup
+    MuiButtonGroup: {
```
4ian, line 186: ButtonGroup is already used in `RaisedButtonWithSplitMenu`. Can you check if this change is not having bad side effects there?

4ian:
Just added a comment, otherwise looks great! 
Sorry for the delay, was super busy these last days. Thanks for the reminder!!

newIDE/app/src/UI/Theme/CreateTheme.js
nilaymaj: My bad - yes, this took away the border between the main button and dropdown arrow in `RaisedButtonWithSplitMenu`. Turns out the double border was something caused by passing `variant="contained"` to ButtonGroup (I think I forgot to remove it after the first mockup). I've removed that prop and this theme override, and everything works okay now.

By the way, if you enable two consecutive bits there is no border between them. A bit hard (or verbose) to fix because MUI doesn't provide a lot of help with mixed button types. I hope that's not a major problem.

4ian:
Excellent! Thanks a lot for your work on this 👍

PR #2554, opened 2021-04-16
Use chokidar in dev watcher

arthuro555:
Currently, the GDJS development watcher has some issues, fixable by using chokidar instead of `fs.watch`:

1. On some platforms, it has false positives, which imports often the files for nothing, often multiple times at once resulting in EPERMS errors and incomplete copy of files.
2. On some platforms, it triggers events multiple times, causing it to import GDJS multiple times at once resulting in EPERMS errors and incomplete copy of files (avoidable with the chokidar `awaitWriteFinish` option).
3. When starting, it triggers many file events on some platforms as setting a file event watcher is triggering file events, causing it to import GDJS multiple times at once resulting in EPERMS errors and incomplete copy of files (avoidable with the chokidar `ignoreInitial` option).
4. According to the chokidar page, doesn't trigger events on MacOS when using Sublime
5. Requires hardcoding paths and having our own complex code to add a watcher to all files

For those reasons, this PR moves the code from `fs.watch` to chokidar


4ian (approved):
Very nice 👌👍
Seems fair to use chokidar, way better and simpler.

4ian:
Thanks! 👍👍

PR #2555, opened 2021-04-16
Fix sound is playing condition

### LINKED ISSUE #2550: Music sometimes doubled
A user reports on the forum that upon launch of the preview, about half of the time, two music files play at once.
I confirmed this on my side.
Related to the new stereo features, perhaps?
Also, I often get a "pop" at preview launch, with a music file that used to work fine in GDevelop.

GD5b108

https://user-images.githubusercontent.com/25978830/114780179-548b0c80-9d77-11eb-920d-441c17ec22dd.mp4

[source](https://forum.gdevelop-app.com/t/launching-a-preview-of-a-scene-causes-2-audios-to-play-at-once-occasionally/31937)

arthuro555:
Fixes #2550 
Between loading and playing the sound, there is a short timeframe where the sound wasn't recognized as being played. As this moment is caused by loading the sound before playing it, it should be considered part of loading and therefore playing the sound.

### DIFF
--- a/GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
+++ b/GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
@@ -107,7 +107,7 @@ namespace gdjs {
      */
     playing(): boolean {
       return (
-        (this._id !== null ? this._howl.playing(this._id) : false) ||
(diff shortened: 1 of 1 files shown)


4ian:
This seems correct! Loading should be considered indeed as "playing" in our abstraction of audio. I missed this when changing to the new sound manager.

PR #2558, opened 2021-04-17
Add new expressions

Bouh:
Close #2367 

The center of the object is used, we should explain this somewhere?

image

to_delete

### DIFF
--- a/Core/GDCore/Extensions/Builtin/MathematicalToolsExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/MathematicalToolsExtension.cpp
@@ -383,7 +383,7 @@ BuiltinExtensionsImplementer::ImplementsMathematicalToolsExtension(
       .AddExpression("XFromAngleAndDistance",
                      _("X position from angle and distance"),
                      _("Compute the X position when given an angle and distance "
-                      "relative to the starting position. This is also known as "
+                      "relative to the origin (0;0). This is also known as "
                       "getting the cartesian coordinates of a 2D vector, using "
                       "its polar coordinates."),
                      _("Mathematical tools"),
@@ -395,7 +395,7 @@ BuiltinExtensionsImplementer::ImplementsMathematicalToolsExtension(
       .AddExpression("YFromAngleAndDistance",
                      _("Y position from angle and distance"),
                      _("Compute the Y position when given an angle and distance "
-                      "relative to the starting position. This is also known as "
+                      "relative to the origin (0;0). This is also known as "
                       "getting the cartesian coordinates of a 2D vector, using "
                       "its polar coordinates."),
                      _("Mathematical tools"),
--- a/Core/GDCore/Extensions/Builtin/MathematicalToolsExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/MathematicalToolsExtension.cpp
@@ -379,6 +379,32 @@ BuiltinExtensionsImplementer::ImplementsMathematicalToolsExtension(
       .AddParameter("expression", _("b (in a+(b-a)*x)"))
       .AddParameter("expression", _("x (in a+(b-a)*x)"));
 
+  extension
+      .AddExpression("XFromAngleAndDistance",
+                     _("X position from angle and distance"),
+                     _("Compute the X position when given an angle and distance "
+                      "relative to the starting position. This is also known as "
+                      "getting the cartesian coordinates of a 2D vector, using "
+                      "its polar coordinates."),
+                     _("Mathematical tools"),
+                     "res/mathfunction.png")
+      .AddParameter("expression", _("Starting X position"))
+      .AddParameter("expression", _("Angle, in degrees"))
+      .AddParameter("expression", _("Distance"));
+
+  extension
+      .AddExpression("YFromAngleAndDistance",
+                     _("Y position from angle and distance"),
+                     _("Compute the Y position when given an angle and distance "
+                      "relative to the starting position. This is also known as "
+                      "getting the cartesian coordinates of a 2D vector, using "
+                      "its polar coordinates."),
+                     _("Mathematical tools"),
+                     "res/mathfunction.png")
+      .AddParameter("expression", _("Starting Y position"))
+      .AddParameter("expression", _("Angle, in degrees"))
+      .AddParameter("expression", _("Distance"));
+
 #endif
 }
 
--- a/Core/GDCore/Extensions/Builtin/MathematicalToolsExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/MathematicalToolsExtension.cpp
@@ -388,7 +388,6 @@ BuiltinExtensionsImplementer::ImplementsMathematicalToolsExtension(
                       "its polar coordinates."),
                      _("Mathematical tools"),
                      "res/mathfunction.png")
-      .AddParameter("expression", _("Starting X position"))
       .AddParameter("expression", _("Angle, in degrees"))
       .AddParameter("expression", _("Distance"));
 
@@ -401,7 +400,6 @@ BuiltinExtensionsImplementer::ImplementsMathematicalToolsExtension(
                       "its polar coordinates."),
                      _("Mathematical tools"),
                      "res/mathfunction.png")
-      .AddParameter("expression", _("Starting Y position"))
       .AddParameter("expression", _("Angle, in degrees"))
       .AddParameter("expression", _("Distance"));
 
--- a/GDCpp/GDCpp/Extensions/Builtin/MathematicalTools.cpp
+++ b/GDCpp/GDCpp/Extensions/Builtin/MathematicalTools.cpp
@@ -78,5 +78,8 @@ double GD_API angleDifference(double angle1, double angle2) {
 
 double GD_API lerp(double a, double b, double x) { return a + (b - a) * x; }
 
+double GD_API XFromAngleAndDistance(double positionX, double angle, double distance) { return positionX + distance * ::cos(((angle / 180) * 3.1415)); }
+double GD_API YFromAngleAndDistance(double positionY, double angle, double distance) { return positionY + distance * ::sin(((angle / 180) * 3.1415)); }
+
 }  // namespace MathematicalTools
 }  // namespace GDpriv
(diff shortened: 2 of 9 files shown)


Bouh:
Ready for a review.

GDJS/Runtime/runtimeobject.ts
```diff
@@ -1494,7 +1494,35 @@ namespace gdjs {
...
+    /**
+     * Compute the X position when given an angle and distance relative to the starting object.
+     * This is also known as getting the cartesian coordinates of a 2D vector, using its polar coordinates.
+     * @param angle The angle, in degrees.
+     * @param distance The distance from the object, in pixels
+     */
+    getXFromAngleAndDistance(angle: float, distance: float): float {
+      return (
+        this.getDrawableX() +
+        this.getCenterX() -
```
4ian, line 1509: Why the minus? It should be a + I think?

GDJS/Runtime/runtimeobject.ts
```diff
@@ -1494,7 +1494,35 @@ namespace gdjs {
...
+    /**
+     * Compute the Y position when given an angle and distance relative to the starting object.
+     * This is also known as getting the cartesian coordinates of a 2D vector, using its polar coordinates.
+     * @param angle The angle, in degrees.
+     * @param distance The distance from the object, in pixels
+     */
+    getYFromAngleAndDistance(angle: float, distance: float): float {
+      return (
+        this.getDrawableY() +
+        this.getCenterY() -
```
4ian, line 1523: same?

4ian:
Nice! Would you be able to also add "free" expressions (without an object) `XFromAngleAndDistance`/`Y`? :)

Bouh:
> Nice! Would you be able to also add "free" expressions (without an object) `XFromAngleAndDistance`/`Y`? :)

Yes with a number parameter instead the object position yes.

I've tried to made the changes from - to + but the result is same in preview.
I use - because the other functions use it, see `getAngleToPosition`
image
And other:
image

Bouh:
I think the syntax for the C++ part is basic, i don't know how to call constant for Pi, so i've used a basic `3.1415`.
I get no error when i compile gd.js.

GDJS/Runtime/events-tools/commontools.ts
```diff
@@ -458,6 +458,36 @@ namespace gdjs {
...
+
+      /**
+       * Compute the X position when given an angle and distance relative to the starting position.
+       * This is also known as getting the cartesian coordinates of a 2D vector, using its polar coordinates.
+       * @param positionX The position on X-axis, in pixels.
+       * @param angle The angle, in degrees.
+       * @param distance The distance from the object, in pixels.
+       */
+      export const XFromAngleAndDistance = function (
+        positionX: float,
```
4ian, line 470: Let's remove the positionX argument :) Usually these expressions just take angle/distance as the origin is considered by default.

Bouh: This is the free expression, we need a starting position.
Or i miss something?

Bouh: @4ian This is the free expression, the expression based object is on the other file. See my other comment.

GDJS/Runtime/runtimeobject.ts
```diff
@@ -1494,7 +1494,35 @@ namespace gdjs {
...
+      return gdjs.toDegrees(Math.atan2(-y, -x));
+    }
+
+    /**
+     * Compute the X position when given an angle and distance relative to the starting object.
+     * This is also known as getting the cartesian coordinates of a 2D vector, using its polar coordinates.
+     * @param angle The angle, in degrees.
+     * @param distance The distance from the object, in pixels
+     */
+    getXFromAngleAndDistance(angle: float, distance: float): float {
```
Bouh, line 1506: @4ian Here the based object function for the expression.

GDJS/Runtime/events-tools/commontools.ts
4ian: @Silver-Streak What do you think? Usually in these mathematical/vector helper functions, 0;0 is always considered as the origin. I'm just a bit afraid of `XFromAngleAndDistance(100, 45, 100)` being more cryptic than `100 + XFromAngleAndDistance(45, 100)`.

GDJS/Runtime/runtimeobject.ts
4ian: Yep this one is fine because there is no confusion possible between the origin and the parameters :)

GDJS/Runtime/events-tools/commontools.ts
Bouh: Ooohh okey I see what you mean now, i agree with you `100 + XFromAngleAndDistance(45, 100)` is better, I'll remove `position`.

GDJS/Runtime/events-tools/commontools.ts
```diff
@@ -458,6 +458,34 @@ namespace gdjs {
...
       };
+
+      /**
+       * Compute the X position when given an angle and distance relative to the starting position.
+       * This is also known as getting the cartesian coordinates of a 2D vector, using its polar coordinates.
+       * @param positionX The position on X-axis, in pixels.
+       * @param angle The angle, in degrees.
+       * @param distance The distance from the object, in pixels.
+       */
+      export const XFromAngleAndDistance = function (
```
4ian, line 469: `getXFromAngleAndDistance` (expression names in GDevelop don't have get, because they are always getters, but function names in JS/C++ always start with a verb)

GDJS/Runtime/events-tools/commontools.ts
```diff
@@ -458,6 +458,34 @@ namespace gdjs {
...
+      };
+
+      /**
+       * Compute the Y position when given an angle and distance relative to the starting position.
+       * This is also known as getting the cartesian coordinates of a 2D vector, using its polar coordinates.
+       * @param positionY The position on Y-axis, in pixels.
+       * @param angle The angle, in degrees.
+       * @param distance The distance from the object, in pixels.
+       */
+      export const YFromAngleAndDistance = function (
```
4ian, line 483: Same

GDJS/Runtime/events-tools/commontools.ts
```diff
@@ -458,6 +458,34 @@ namespace gdjs {
       export const trunc = function (x: float): number {
         return x | 0;
       };
+
+      /**
+       * Compute the X position when given an angle and distance relative to the starting position.
```
4ian, line 463: `relative to the origin (0;0)`

GDJS/Runtime/events-tools/commontools.ts
```diff
@@ -458,6 +458,34 @@ namespace gdjs {
...
+       */
+      export const XFromAngleAndDistance = function (
+        angle: float,
+        distance: float
+      ): number {
+        return distance * Math.cos(gdjs.toRad(angle));
+      };
+
+      /**
+       * Compute the Y position when given an angle and distance relative to the starting position.
```
4ian, line 477: `relative to the origin (0;0)`

Core/GDCore/Extensions/Builtin/MathematicalToolsExtension.cpp
```diff
@@ -379,6 +379,30 @@ BuiltinExtensionsImplementer::ImplementsMathematicalToolsExtension(
...
+                     _("Mathematical tools"),
+                     "res/mathfunction.png")
+      .AddParameter("expression", _("Angle, in degrees"))
+      .AddParameter("expression", _("Distance"));
+
+  extension
+      .AddExpression("YFromAngleAndDistance",
+                     _("Y position from angle and distance"),
+                     _("Compute the Y position when given an angle and distance "
+                      "relative to the starting position. This is also known as "
```
4ian, line 398: `relative to the origin (0;0)`

Core/GDCore/Extensions/Builtin/MathematicalToolsExtension.cpp
```diff
@@ -379,6 +379,30 @@ BuiltinExtensionsImplementer::ImplementsMathematicalToolsExtension(
       .AddParameter("expression", _("b (in a+(b-a)*x)"))
       .AddParameter("expression", _("x (in a+(b-a)*x)"));
 
+  extension
+      .AddExpression("XFromAngleAndDistance",
+                     _("X position from angle and distance"),
+                     _("Compute the X position when given an angle and distance "
+                      "relative to the starting position. This is also known as "
```
4ian, line 386: `relative to the origin (0;0)`

GDCpp/GDCpp/Extensions/Builtin/MathematicalTools.cpp
```diff
@@ -78,5 +78,8 @@ double GD_API angleDifference(double angle1, double angle2) {
 
 double GD_API lerp(double a, double b, double x) { return a + (b - a) * x; }
 
+double GD_API XFromAngleAndDistance(double positionX, double angle, double distance) { return positionX + distance * ::cos(((angle / 180) * 3.1415)); }
```
4ian, line 81: To update (or even remove, we don't really care about GDCpp anymore)

GDJS/GDJS/Extensions/Builtin/MathematicalToolsExtension.cpp
```diff
@@ -59,6 +59,8 @@ MathematicalToolsExtension::MathematicalToolsExtension() {
   GetAllExpressions()["round"].SetFunctionName("Math.round");
   GetAllExpressions()["trunc"].SetFunctionName("gdjs.evtTools.common.trunc");
   GetAllExpressions()["lerp"].SetFunctionName("gdjs.evtTools.common.lerp");
+  GetAllExpressions()["XFromAngleAndDistance"].SetFunctionName("gdjs.evtTools.common.XFromAngleAndDistance");
```
4ian, line 62: (to update here too)

4ian (approved):
Great, almost there! 👍

Bouh:
It's ready to merge, i've made changes and tested.
Let me know!

GDJS/Runtime/events-tools/commontools.ts
```diff
@@ -469,7 +469,7 @@ namespace gdjs {
         angle: float,
         distance: float
       ): number {
-        return distance * Math.cos(gdjs.toRad(angle));
+        return distance !== 0 ? distance * Math.cos(gdjs.toRad(angle)) : 0;
```
4ian: I've reverted this because this adds an extra check which will be useless in 99% of cases, so would actually slow down in most cases.

4ian:
Merged, thanks!

Bouh:
Thanks you too!

PR #2569, opened 2021-04-20
Cancel change by Escape key on inline popover

Bouh:
# Idea
Cancel the change made in a inline popover when Escape key is used.


4ian:
> Cancel the change made in a inline popover when Escape key is used.

We should make sure that all inline popover can then by validated with Enter or Ctrl+Enter (or Cmd+).
Also, a click outside should probably keep validating the changes (especially because on tablets, that's the only way to validate)?

arthuro555:
Personally, i use esc often to confirm in multiple softwares, having it discard the value here would be annoying imo.

Bouh:
I've added a preference for cancel changes with escape key on inline popover in eventsheets, this options isn't enabled by default.

I would add `CTRL+ENTER` as suggested but i'm not sure how to do that.
Should I write something into a function from an existing props like `closeParameterEditor` props or `onRequestClose` from `<InlineParameterEditor>` , or create a props for that. I prefer asking before trying.

Bouh:
I was worried to create a new props for apply, but it's unnecessary.
to_delete

Bouh:
I'am good.
One question I don't get why we have `UI\KeyboardShortcuts\InteractionKeys.js` thats list a combo of keys checked from event keys,
And a similar file `UI\KeyboardShortcuts\index.js` that check also event keys.

newIDE/app/src/EventsSheet/InlineParameterEditor.js
```diff
@@ -123,7 +123,9 @@ export default class InlineParameterEditor extends React.Component<
           parameterIndex={this.props.parameterIndex}
           value={instruction.getParameter(this.props.parameterIndex)}
           onChange={this.props.onChange}
-          onRequestClose={this.props.onRequestClose}
```
4ian, line 126: This change seems useless?

newIDE/app/src/EventsSheet/index.js
```diff
@@ -562,10 +566,31 @@ export default class EventsSheet extends React.Component<Props, State> {
         ? parameterContext.domEvent.currentTarget
         : null,
       inlineEditingChangesMade: false,
+      inlineEditingPreviousValue: instruction.getParameter(parameterIndex),
     });
   };
 
-  closeParameterEditor = () => {
+  closeParameterEditor = (
+    event: SyntheticKeyboardEvent<HTMLInputElement>,
```
4ian, line 574: Event seems useless?

4ian, line 574: Don't pass event if you don't need it, replace it by a boolean with a name that describes what it does.

newIDE/app/src/EventsSheet/index.js
```diff
@@ -562,10 +566,31 @@ export default class EventsSheet extends React.Component<Props, State> {
...
+    peferenceEventsSheetCancelInputParameterPopover: boolean
```
4ian, line 575: Use a shorter name for your boolean, we don't care here that it comes from the preferences.

newIDE/app/src/EventsSheet/InlinePopover.js
```diff
@@ -83,7 +84,14 @@ export default function InlinePopover(props: Props) {
...
           // when showing autocompletion), which is fine.
+
+          // When the event match with Escape key changes are canceled.
           if (shouldCloseOrCancel(event)) {
+            props.onRequestClose(event);
+          }
+
+          // When event match with ctrlOrCmd + Enter changes are applied.
+          if (shouldSubmit(event)) {
             props.onRequestClose();
```
4ian, line 95: The comment is not really aligned with what the code does. If I read the code, I can just see 
"If I should submit (i.e: if Ctrl + Enter was pressed), then I should close".

newIDE/app/src/EventsSheet/InlinePopover.js
```diff
@@ -83,7 +84,14 @@ export default function InlinePopover(props: Props) {
           // Note that the content of the popover can capture the event
           // and stop its propagation (for example, the GenericExpressionField
           // when showing autocompletion), which is fine.
+
+          // When the event match with Escape key changes are canceled.
           if (shouldCloseOrCancel(event)) {
+            props.onRequestClose(event);
```
4ian, line 90: This is not very clear: `onRequestClose` will do something different according to if event is passed or not... weird no? 
onRequestClose should do one thing: closing the popover. You can pass it an additional parameter, but in this case prefer a boolean or something with a proper name rather than passing or not event.

newIDE/app/src/MainFrame/Preferences/PreferencesDialog.js
```diff
@@ -261,6 +262,21 @@ const PreferencesDialog = ({ i18n, onClose }: Props) => {
...
+          <Line>
+            <Toggle
+              onToggle={(e, check) =>
+                setEventsSheetCancelInputParameterPopover(check)
+              }
+              toggled={values.eventsSheetCancelInputParameterPopover}
+              labelPosition="right"
+              label={
+                <Trans>
+                  Use Escape for cancel changes on inline parameter, CTRL+ENTER
```
4ian, line 274: This describes what it does if we activate this, but if we deactivate this what happens?
A Toggle might not be the proper thing to use here.

4ian:
> One question I don't get why we have UI\KeyboardShortcuts\InteractionKeys.js thats list a combo of keys checked from the event,
> And a similar file UI\KeyboardShortcuts\index.js that check also from event.

They are a bit redundant, but InteractionKeys is basically for simple things that are not customizable (Closing will always be with escape, submitting always with Ctrl/Cmd+Enter).

I have the impression things are overcomplicated there in the code. Notably:
- not sure if we need a preferences. Escape should be to close something while cancelling. Ctrl+Enter should be for validating.
- `onRequestClose` should be only for closing. Let's use something else to validate the changes and then close.

newIDE/app/src/EventsSheet/InlineParameterEditor.js
4ian: Ah this is because of the fact that you're using event to test what to do. As explained in other comments, let's try to find a more explicit approach ;)

newIDE/app/src/EventsSheet/index.js
4ian: For example, this can be `shouldCancel`, which is true/false => much clearer :)

newIDE/app/src/MainFrame/Preferences/PreferencesDialog.js
Wend1go: Should also be "... **to** cancle changes ..." I think.

newIDE/app/src/EventsSheet/index.js
```diff
@@ -570,11 +570,17 @@ export default class EventsSheet extends React.Component<Props, State> {
     });
   };
 
+  /**
+   * Handle how are managed changes on InlinePopover component
+   *
+   * @param shouldCancel Set to true for cancel changes
```
Wend1go, line 576: Set to true **to** cancel changes

newIDE/app/src/MainFrame/Preferences/PreferencesDialog.js
Wend1go: Oops, added a new typo. ^^
It is "cancel" of course.

newIDE/app/src/MainFrame/Preferences/PreferencesDialog.js
```diff
@@ -261,6 +262,21 @@ const PreferencesDialog = ({ i18n, onClose }: Props) => {
               }
             />
           </Line>
+          <Line>
+            <SelectField
+              floatingLabelText={
+                <Trans>Inline popover parameter behavior when escaping</Trans>
```
Bouh, line 268: @Wend1go Here the current sentence, your previous ping about the sentence was on an outdated commit.

newIDE/app/src/MainFrame/Preferences/PreferencesDialog.js
Bouh: Sorry, this review on this line is outdated.
The current line is : 
`Inline popover parameter behavior when escaping` 
I've pinged you on the correct one.

Bouh:
Wonderful, tests are passed!
Ready for a review.

4ian:
Cool, I'm taking over this :)

4ian:
Is this working for you? When I press Escape, the changes are always applied it seems.

4ian:
Was more work than expected but I think we're getting something working consistently in all cases. 
Still to do:
- [ ] Remove all the console.log
- [ ] Adapt the inline popover used in events (for each, etc...) to support cancelling.

I'm almost thinking we should from now remove the preference and only support Escape for cancelling and Ctrl/Cmd+Enter to validate.

Bouh:
Oh god I think the merge of the master screwed up everything!
Yes, it worked before.
I will check what happened.

4ian:
No worries, I'll finish this. Yes there was a change in master that blocked the cancelling (because the boolean was not passed), but also a few other edge cases (because when you try to cancel, the inline popover field was then blurred and caused the changed value to be set again).

Bouh:
The idea was only on the event parameters on the event sheet, I didn't plan to make the changes as wide as that.

Bouh:
> Was more work than expected...

Yes, I see, It wasn't expected on my side too.
I understand that it takes time to do writing reviews, it also takes time on my side to read and understand what went wrong by reading the changes you made. I would have liked to have feedback to finish this. It was totally something I felt I could do.

4ian:
I see, it's something quite sensitive because of the way blurring works (I just discovered that a `.focus()` needed a change in the order of the call with setState while working on this), of the order of events and I suspected the ramifications to be beyond what I could find by just reading the code. That's why I preferred to fetch the branch and do the changes by myself, because otherwise I can't be sure what was the good solution :)
I'll try to provide more guidance next time!

4ian:
Finally merged! :) I've made some more changes to make sure all "inline popover" have the same behavior. Much better to have this consistency :) 
Thanks for working on this!

Issue #897, opened 2019-01-21
Platform Character cannot more left and right on a platform going up.


zatsme:
I've been working on a new advanced platformer example with new animations and state machine with lots of comments which I hope to finish and then make into a full tutorial 😄 

Unfortunately I found a bug which I don't think has been mentioned before....

In this GIF you see the platform going down and character moving left/right no problem..  Going up however and the player cannot go left/right.. Sometimes a jump then allows the left/right movement, but usually not.

![platform bug](https://user-images.githubusercontent.com/17770123/51497646-a4a6de80-1dbb-11e9-9e3d-21eb7b1b945d.gif)

Here is the small example...

[PlatformerBUG.zip](https://github.com/4ian/GDevelop/files/2780446/PlatformerBUG.zip)

PascalLadalle:
Bumping this as I noticed a related issue: when the platform is moving up, GDev detects the platformer as falling. Maybe it explains the issue above.
![ezgif com-crop](https://user-images.githubusercontent.com/25978830/65256560-001e9d00-db00-11e9-94d0-30301f6efbfd.gif)
This fixes the animation issue: 
![image](https://user-images.githubusercontent.com/25978830/65256582-0a409b80-db00-11e9-8c08-610b8c1ae17a.png)

PawBud:
A similar issue has been reported by `Kappa` a GDevelop user. He asked me to report this on behalf of him @4ian, as seen over here https://gofile.io/d/sic5IO. After discussing it on [Discord](https://discordapp.com/channels/258623956158906368/258623956158906368/740180306824265738), a solution could not be determined. The things that kappa tested:- 
1. Not moving the hero at all, just having them sit on the platform and move the platform up - Leads to jittery movement.
2. Moving the hero along with the platform - Leads to the jittery Movement.

The above-mentioned method worked in the previous release of GDevelop.

D8H:
I guess this issue is solved by this PR
* https://github.com/4ian/GDevelop/pull/3009

4ian:
Yep! Finally! 🤩

Issue #2562, opened 2021-04-18
Platformer character placed on a platform moves down or up a few pixels at the start of the scene


Silver-Streak:
## Describe the bug

If you start the a scene where a platform character is precisely placed above a platform (so it doesn't start the scene falling), there is a brief moment where the platform character stutters down/up. This gets detected by the platform behavior as falling. It appears to only occur for a single frame. This is problematic especially for games using the platformer behavior that don't want to show a momentary flip between falling/idle animations at the start of each level.

## To Reproduce
1. Create a sprite object with the platformer character behavior
2. Create a sprite object with the platform behavior.
3. Add both to the scene.
4. Position the platformer character so it is exactly above the platform. (E.g. platform character is 64x64. Platform starts at X114, Y364. Position the platformer character at 300Y)
5. Set up an event that restarts the scene upon keypress.
6. Preview the scene and press the key repeatedly. Watch the bottom of the platform character to see it dance.

This occurs whether rounded coordinates are enabled, and regardless of the type of object the platform is. It does not appear to be impacted by resolution, scaling modes, or preview vs export. 

I have also tested by using a "At the beginning of scene" event to position the platform character via expressions.

Example project:  
[PlatformerDancing.zip](https://github.com/4ian/GDevelop/files/6330620/PlatformerDancing.zip)

Web Export of Example: https://games.gdevelop-app.com/game-228e929b-3dc0-4f6b-b469-edff3d8c97af/index.html

## Other details
OS: Windows 10 Pro 64 Bit
GD5 version: 5.0.0-beta108

Because this has to do with being on top of a platform, this could be related to this as of yet unsolved issue: https://github.com/4ian/GDevelop/issues/1920

D8H:
I can reproduce it in unit tests.
It can be fixed by checking the transition from Failling to OnFloor before moving in addition to the final check.
It has a side effect on platform moving up, it may be linked to another bug where player can move on a platform going up.

Silver-Streak:
> I can reproduce it with unit tests.
> It can be fixed by checking the transition from Failling to OnFloor before moving in addition to the final check.
> It has a side effect on platform moving up, it may be linked to another bug where player can move on a platform going up.

Are you saying you're going to need to modify the platformer state machine, or is this something I can do via events. 

(as far as the platform moving up, I haven't seen anyone even take a crack at: https://github.com/4ian/GDevelop/issues/897 and it's been ~1.5 years now)

D8H:
[...]

PR #2574, opened 2021-04-23
 [Platformer] The player can move on platforms moving up

D8H:
Commited over:
* https://github.com/4ian/GDevelop/pull/2602
* https://github.com/4ian/GDevelop/pull/2603

This is a custom build to let Silver-Streak test it before the review:
* 1st version: https://www.dropbox.com/s/1l52e0vfm3ki1o0/gdevelop-plateformer-fix-2021-04-23.7z?dl=1
* 2nd version: https://www.dropbox.com/s/gxymf3zfs6njo7n/gdevelop-plateformer-fix-2021-04-25.7z?dl=1

It tries to fix:
* https://github.com/4ian/GDevelop/issues/897

# In master
* jumpthru platform
   * up: the player can't move and there is a gap smaller than 1 pixel
   * down: there is a gap smaller than 1 pixel
* regular platform
   * up: the player is jumpy
   * down: there is a gap smaller than 1 pixel

## Jumpthru platform going up
The player can't move right and left when the platform moves up because the player is in collision with the platform when moveX is called.
* doStepPreEvents() separate won't move the object up because it's a jumpthru
* beforeMovingX() set requestedDeltaY
* moveX() does move the player
* onFloor.beforeMovingY() reverts what moveX() did because the "slop is too step" 
* onFloor.beforeMovingY() applies the requestedDeltaY that make the player less than one pixel on top of the platform

## Regular platform going up
The player goes too much high and fell.
* doStepPreEvents() separate move the object up
* beforeMovingX() set requestedDeltaY according to the platform movement
* moveX() moves the player
* onFloor.beforeMovingY() applies the requestedDeltaY even though it's not needed
* _checkTransitionOnFloorOrFalling() more than 1 pixel apart, transition to falling 

# The fix
To fix these 2 "going up" issues, the plateform deltaY can be applied at the beginning.
The small gap is just that onFloor.beforeMovingY() make the player nearer to the platform by 1 pixel steps. A separate solve this but separate has rounding errors so the object can stay inside the platform (like 1e-15) so I added a collision check with an error margin in RuntimObject.

### DIFF
--- a/Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
+++ b/Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
@@ -1410,10 +1410,34 @@ namespace gdjs {
         if (noMoreOnFloor) {
           object.setY(oldY);
         } else {
-          object.setY(
-            //Floor touched: Go back 1 pixel over.
-            object.getY() - 1
-          );
+          if (behavior._roundCoordinates) {
+            object.setY(
+              //Floor touched: Go back 1 pixel over.
+              object.getY() - 1
+            );
+          } else {
+            // Moving platforms won't always be right on a pixel
+            // Avoid a gap between the object and the platform
+            const beforeSeparationX = object.getX();
+            const beforeSeparationY = object.getY();
+            object.separateFromObjects(
+              [this._floorPlatform!.owner],
+              this._behavior._ignoreTouchingEdges
+            );
+            const deltaX = object.getX() - beforeSeparationX;
+            const deltaY = object.getY() - beforeSeparationY;
+            if (
+              deltaX < gdjs.PlatformerObjectRuntimeBehavior.epsilon &&
+              deltaX > -gdjs.PlatformerObjectRuntimeBehavior.epsilon &&
+              deltaY >= -1 &&
+              deltaY < gdjs.PlatformerObjectRuntimeBehavior.epsilon
+            ) {
+              object.setX(beforeSeparationX);
+              object.setY(Math.min(beforeSeparationY, object.getY()));
+            } else {
+              object.setPosition(beforeSeparationX, beforeSeparationY - 1);
+            }
+          }
         }
       }
     }
--- a/Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
+++ b/Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
@@ -141,6 +141,9 @@ namespace gdjs {
     }
 
     doStepPreEvents(runtimeScene: gdjs.RuntimeScene) {
+      // Avoid inconsistencies between spacial search and the actual platforms location
+      this._manager.updateAABBIfNeeded(runtimeScene);
+
       const LEFTKEY = 37;
       const UPKEY = 38;
       const RIGHTKEY = 39;
@@ -240,6 +243,15 @@ namespace gdjs {
         Math.abs(object.getX() - oldX) >= 1 ||
         Math.abs(object.getY() - oldY) >= 1;
       this._lastDeltaY = object.getY() - oldY;
+
+      // When the object is both platformer and platform, update its AABB.
+      // Depending on the doStepPreEvents calls order platformer instance
+      // collisions can be sightly different but that should be fine.
+      const behavior = object.getBehavior('Platform');
+      if (behavior != null) {
+        const platformBehavior = behavior as gdjs.PlatformRuntimeBehavior;
+        platformBehavior.updateAABB(runtimeScene);
+      }
     }
 
     doStepPostEvents(runtimeScene: gdjs.RuntimeScene) {}
--- a/Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
+++ b/Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
@@ -1387,7 +1387,8 @@ namespace gdjs {
         const tentativeStartY = object.getY() + 1;
         object.setY(
           behavior._roundCoordinates
-            ? Math.round(tentativeStartY)
+            ? // tentativeStartY is already incremented, floor avoid to jump from more than one pixel
+              Math.floor(tentativeStartY)
             : tentativeStartY
         );
         let step = 0;
--- a/Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
+++ b/Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
@@ -65,6 +65,8 @@ namespace gdjs {
     _requestedDeltaY: float = 0;
     _lastDeltaY: float = 0;
 
+    static epsilon: float = 1 / (2 << 16);
+
     constructor(
       runtimeScene: gdjs.RuntimeScene,
       behaviorData,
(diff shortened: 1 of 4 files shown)


Silver-Streak:
@D8H 

So, seems a complete fix for my issue, as this solves: https://github.com/4ian/GDevelop/issues/2562
For my issue (#2562), I am no longer able to reproduce the jitter behavior at the start of the scene, I'm pretty sure you've solved this one.

But it does not fully solve: https://github.com/4ian/GDevelop/issues/897

Here's a video: https://streamable.com/i3onus

You can see that I can move on the platform, but the character still jitters/bounces up and down pretty rapidly as the platform raises or lowers. The example project is linked in the issue's first post. That said, I think you're connected with this already?

I remember you mentioning that there is a direct correlation to max fall speed, so I tested by setting the fall speed to 10000, which is above the acceleration (1500) in the example. This **does** eliminate the jitter when moving up/down, and at bare minimum establishes a workaround for the issue (we can instruct users to set up elevators/raising platform events to include an action that increases their max fall speed to a very large number while in collision, then reduces it back to normal when they've left the platform). 

Your changes do completely fix the inability to move on the raising/falling platform, though. No matter the fall speed I can still move now.

D8H:
> You can see that I can move on the platform, but the character still jitters/bounces up and down pretty rapidly as the platform raises or lowers. The example project is linked in the issue's first post. That said, I think you're connected with this already?

From what I see from the tests the player follow the platform, but maybe I don't have the right settings to reproduce. Are you using pixel rounding? In this case, try to round the platform position after moving it to see if it fix the issue.

Without pixel rounding:
```
LOG: 'playerY: -5.899999999999999 platformY: 14.1'
LOG: 'playerY: 18.200000000000003 platformY: 38.2'
LOG: 'playerY: 42.300000000000004 platformY: 62.300000000000004'
LOG: 'playerY: 66.4 platformY: 86.4'
LOG: 'playerY: 90.5 platformY: 110.5'
```

With pixel rounding:
```
LOG: 'playerY: -6 platformY: 14.1'
LOG: 'playerY: 18 platformY: 38.2'
LOG: 'playerY: 42 platformY: 62.300000000000004'
LOG: 'playerY: 66 platformY: 86.4'
LOG: 'playerY: 90 platformY: 110.5'
```

Silver-Streak:
I wonder if a fix could be an eventual "Elevator/Moving Platform" type for the main platform behavior that takes this into accounts, similar to jump through/normal/etc.

Regardless, your fix *does* give us an option for workarounds. Temporarily increasing max fall speed beyond acceleration is an easy thing to set up, and one I can document for users on the Wiki, if there isn't a more easily managed fix.

D8H:
> I wonder if a fix could be an eventual "Elevator/Moving Platform" type for the main platform behavior that takes this into accounts, similar to jump through/normal/etc.

The platform following could be move to doStepPostEvents to avoid the gap at rending. I will try it when the bug is fixed.

> Regardless, your fix _does_ give us an option for workarounds. Temporarily increasing max fall speed beyond acceleration is an easy thing to set up, and one I can document for users on the Wiki, if there isn't a more easily managed fix.

I don't know the effect of these properties very well, but in case of the platform following the limit I set is only between platform speed and max fall speed, acceleration shouldn't mater. Only really fast platforms will go beyond max fall speed.

D8H:
Sometimes the player can go through the jumpthru from above. It's a bug that already exist in the stable version. I think I'll leave it to another PR.

D8H:
I think I fixed the gap. I made a new build: https://www.dropbox.com/s/gxymf3zfs6njo7n/gdevelop-plateformer-fix-2021-04-25.7z?dl=1

Silver-Streak:
> I think I fixed the gap. I made a new build: https://www.dropbox.com/s/gxymf3zfs6njo7n/gdevelop-plateformer-fix-2021-04-25.7z?dl=1

While it is still not quite perfect without modifying max fall speed > acceleration, it is ALMOST unnoticeable now. 
Video of the new build, without modifying max fall speed: https://streamable.com/xs9avs

Video of the new build, with modifying max fall speed to 10000: https://streamable.com/0019pf

D8H:
I don't see the difference between the 2 captures. A red sprite and 4:2:0 is not good to notice a gap smaller than one pixel. Black and white should be better to compare as luminance is in full resolution in videos.

https://user-images.githubusercontent.com/2611977/115993351-aa12b500-a5d2-11eb-84d4-1d26dec3258e.mp4


I capture this with Max falling speed = 1000, Acceleration = 1500 and platform speed = 100

D8H:
When a jumpthru is moving up the player may not be able to land on it because it overlaps him a little. This is not a regression.

4ian:
I've not followed really what was done here, but originally the logic for following platforms on the X axis was to update `_requestedDeltaX` to be the same as the floor movement, and same for Y.
Y never really worked properly. Would you be able to tell exactly the strategy to fix this?

D8H:
# In master
* jumpthru platform
   * up: the player can't move and there is a gap smaller than 1 pixel
   * down: there is a gap smaller than 1 pixel
* regular platform
   * up: the player is jumpy
   * down: there is a gap smaller than 1 pixel

## Jumpthru platform going up
The player can't move right and left when the platform moves up because the player is in collision with the platform when moveX is called.
* doStepPreEvents() separate won't move the object up because it's a jumpthru
* beforeMovingX() set requestedDeltaY
* moveX() does move the player
* onFloor.beforeMovingY() reverts what moveX() did because the "slop is too step" 
* onFloor.beforeMovingY() applies the requestedDeltaY that make the player less than one pixel on top of the platform

## Regular platform going up
The player goes too much high and fell.
* doStepPreEvents() separate move the object up
* beforeMovingX() set requestedDeltaY according to the platform movement
* moveX() moves the player
* onFloor.beforeMovingY() applies the requestedDeltaY even though it's not needed
* _checkTransitionOnFloorOrFalling() more than 1 pixel apart, transition to falling 

# The fix
To fix these 2 "going up" issues, the plateform deltaY can be applied at the beginning.
The small gap is just that onFloor.beforeMovingY() make the player nearer to the platform by 1 pixel steps. A separate solve this but separate has rounding errors so the object can stay inside the platform (like 1e-15) so I added a collision check with an error margin in RuntimObject.

Silver-Streak:
@D8H @4ian Is there anything particular we're waiting on with this one?

I'm seeing all of the separate PRs, do they replace the need for this PR, and then just need the review? Or are they separate?

D8H:
> 
> 
> @D8H @4ian Is there anything particular we're waiting on with this one?
> 
> I'm seeing all of the separate PRs, do they replace the need for this PR, and then just need the review? Or are they separate?

I split the PR in 3, this one is now only about the platforms issues but it needs the 2 other PR to work. I updated the description.
About the next step, there is a design choice to to make. I made a suggestion here: https://github.com/4ian/GDevelop/pull/2602#discussion_r625042873 but I'm not expecting a quick answer (there were big merges recently).

4ian:
Sorry I felt behind in these PR. I'll try to take a look today :)

D8H:
This original issue is solved by:
* https://github.com/4ian/GDevelop/pull/3009

PR #2578, opened 2021-04-24
Fix typo in comment from a project example

Bouh:
Don't show in changelog

### DIFF
--- a/newIDE/app/resources/examples/buttons/buttons.json
+++ b/newIDE/app/resources/examples/buttons/buttons.json
@@ -386,7 +386,7 @@
             "textG": 0,
             "textR": 0
           },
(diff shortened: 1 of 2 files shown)


4ian (approved):

Bouh:
Wait @4ian  i  edit another example

4ian:
Ok feel free to merge when ready

Bouh:
Well i've tried to use the Github interface for make edits, but it's not easier than my classic way to do, i've maybe created a branch on your repo with my nickname inside, these branch can be delete.

- I've made a small typo fix thx Entropy!
- Fix a issue where the mobile control was useable on desktop even if the controls are hidden, I've added an extra condition. And updated the wiki page.
image

Two small things, not needed to be shown to the users.

PR #2580, opened 2021-04-25
Incremental builds (incomplete)

arthuro555:
This is adds caching to events code generation, allowing to only generate events code of events that have been modified. It is a very early draft, it does not compile.
I am unsure whether this approach would actually work and if I should continue working on it. My main concerns are:

- Object lists: Object lists would probably not be initialized if their Object conditions/actions aren't generated each time
- Expressions are tricky to make cacheable this way
- Adds a lot of Cache managing code
- I probably won't get this to compile because my C++ skills are low and I don't fully understand what I am doing with smart pointers :p

### DIFF
--- a/Core/GDCore/Events/Builtin/ForEachChildVariableEvent.cpp
+++ b/Core/GDCore/Events/Builtin/ForEachChildVariableEvent.cpp
@@ -5,6 +5,7 @@
  */
 
 #include "ForEachChildVariableEvent.h"
+
 #include "GDCore/Events/Serialization.h"
 #include "GDCore/Serialization/SerializerElement.h"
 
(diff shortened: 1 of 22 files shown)


arthuro555:
The more I think about it the more I believe this approach is too naive. I am unsure on how to make this work with non-free instructions, as just remembering their code won't update the code generation context for other instructions and not generate the object lists🤔 .

4ian:
Yes it's a complex problem and the problem with caches is that they come at a cost in terms of potential bugs (you forget to clear the cache, or you discover that you're not invalidating the cache in some situation, etc...) that are hard to debug. Basically for now the bug of "my game events are not updated when I press the preview button" does not even exist - it's not even possible. With a cache, we're trading speed for memory... and for risky bugs like this one ;) 

If we go back to the original problem: why do we need incremental build? "Because builds are too long!". Fair enough! So we have two solutions:
* Remove the problem entirely by making the builds **faster**. How to do so? Launch a profiler and see what takes time. There are two things:
  * Code generation can take time. We can either make it faster (by checking if we can do less allocations, pre-allocate memory, cache certain things etc...) or if we can avoid doing it (for example, favor usage of extensions, see below).
  * Copying resource files take time. We should probably do it in parallel. That can be an "easy" win in the LocalFileSystem.
* If despite our best efforts, things are still slow, we can think of fixing the issue by doing **less things**:
  * There is something that is already semi incremental: it's extensions code generation. You can change the events of a scene, and extensions won't be re-generated. We should btw improve the extensions to only generate the necessary extensions code when a change is made (rather than all extensions). This should be much more impactful I think for large games :) 

Ideally, we could make things "blazing fast" so that we don't even care about incremental things. But it would still be good to see if we can do less work by allowing extensions not to be code-generated all the time every time we change an extension in the editor :)

arthuro555:
I see. In this case I'm closing this as it probably won't lead to anything.

PR #2581, opened 2021-04-25
[Platformer] Add tests on moving platforms

D8H:
I added an explanation on each skipped test that is not obvious.
The tests pass (except one) on this branch: https://github.com/4ian/GDevelop/pull/2574
There is actually 4 bug fixes and the commits are unorganized. If it's more convenient, I can make one PR per bug fix or one commit per bug fix in only one PR.


4ian (approved):
Looks good!

PR #2588, opened 2021-04-27
Rework GetUsedExtensions

### LINKED ISSUE #1753: GetUsedExtension not returning all used extensions
## Describe the bug
It seems like GetUsedExtension doesn't output the actual used extensions.
image
As you can see there is not the AdMob extension in the list, even tho it is clearly used. I also moved some actions arround and added other from admob to see if libGD would then register the extension but it didn't work.

## To Reproduce
1. Open GDevelop and create a project that uses an Extension (Here I used AdMob)
2. Open the dev tools and go on the components tab of the react debugger tools
3. Find MainFrame and then the Electron menu component.
4. Click on that component and scroll to the state to the project object.
5. Expand it and find the ptr attribute
6. Go to the console and type
```js
let project = new gd.Project();
project.ptr = <copiedPointer>;
project.getUsedExtensions().toJSArray();
```
7. See that the non built-in extensions are missing.

## Other details
* Windows, branch https://github.com/arthuro555/GDevelop/tree/platform-specific-dependencies-support (after having merged master)

arthuro555:
Closes #1753

Reworks Project::GetUsedExtensions to use gd::WholeProjectRefactorer to scan all Instructions, Expressions, Objects and Behaviors. 
This PR also adds an Objects/Behaviors Worker and method to WholeProjectRefactorer.


arthuro555:
This PR is ready for review

Core/GDCore/IDE/Events/UsedExtensionsFinder.h
```diff
@@ -0,0 +1,62 @@
...
+class Object;
+class BehaviorContent;
+}  // namespace gd
+
+namespace gd {
+
+class GD_CORE_API UsedExtensionsFinder
+    : public ArbitraryObjectsWorker,
+      public ArbitraryEventsWorkerWithContext,
+      public ExpressionParser2NodeWorker {
```
4ian, line 27: Nice 👀

arthuro555:
🤔  Ah it seems I cannot use the WholeProjectRefactorer in Project without breaking GDCore, which makes sense as it is an IDE tool.

4ian (commented):
Very cool!! I like how the implementation is clean, very nice. 
Could you make a list of where GetUsedExtensions is used in the app?

It's "fairly expensive" as it will visit the whole project, so we might want to be sure not to use it too much or in a loop :) 

> 🤔 Ah it seems I cannot use the WholeProjectRefactorer in Project without breaking GDCpp, which makes sense as it is an IDE tool.

Correct, you can guard `GetUsedExtensions` in a ifdef GD_IDE_ONLY so that it's not exposed to GDCpp... Or we remove this method from gd::Project and make it a static function of `UsedExtensionsFinder`?
After all, if a method can be implemented without needing the internals of a class, it may as well be moved to a static function taking the object as a parameter.

BUT actually, I think the usage of GetUsedExtensions in GDCpp is wrong. Instead, GDCpp should have been using the list of extensions loaded on the platform. It's not the responsibility of the project to give at runtime what extensions are loaded. It's the CppPlatform that should have been used.

arthuro555:
> Could you make a list of where GetUsedExtensions is used in the app?

I just removed all usages of it. All places that used it could get all extensions from the platform instead. Something I wanted to use it for was only exporting extension dependencies of used extensions.

arthuro555:
I am not sure what the new error is all about. I didn't add an include to IDE tools in GDCore_test, why is it complaining?

4ian:
Huh, weird, let me check.

Core/GDCore/IDE/Events/UsedExtensionsFinder.cpp
```diff
@@ -0,0 +1,137 @@
...
+          : gd::MetadataProvider::GetExtensionAndActionMetadata(
+                projectPtr->GetCurrentPlatform(), instruction.GetType());
+  usedExtensions.insert(metadata.GetExtension().GetName());
+
+  for (auto expression : instruction.GetParameters()) {
+    gd::ExpressionParser2 parser(projectPtr->GetCurrentPlatform(),
+                                 GetGlobalObjectsContainer(),
+                                 GetObjectsContainer());
+    parser.ParseExpression("unknown", expression.GetPlainString())
+        ->Visit(*this);
```
4ian, line 57: This might be a bit too direct. Iterate on the metadata parameters, so that for each parameter you have its type and can check if it's a number or a string (`ParameterMetadata::IsExpression("number", type) || ParameterMetadata::IsExpression("string", type)`) and if so, use this type for the parser (`parser.ParseExpression(type, ...)`).

Otherwise, the parameter is not an expression and it would not be useful to parse it. Passing the proper type is also important, because I guess otherwise the parsing could be slightly different at some points.

See an example in EventsRefactorer::RenameObjectInConditions (this class is a bit more complicated than it should because it's not using `ArbitraryEventsWorker`, but it's still doing its job properly for parameters)

Core/GDCore/IDE/Events/UsedExtensionsFinder.h
```diff
@@ -0,0 +1,62 @@
...
+ public:
+  static std::set<gd::String> ScanProject(gd::Project& project);
+
+ private:
+  UsedExtensionsFinder(gd::Project& _project) { projectPtr = &_project; };
+  gd::Project* projectPtr;
```
4ian, line 33: nitpicking: if you don't need to change the pointer and it's always defined, you can use a reference, that you initialize in the initializer list of the constructor:
```suggestion
  UsedExtensionsFinder(gd::Project& project_): project(project_) {};
  gd::Project& project;
```

References are safer because the compiler enforce their initialization. Also uber nit: use an underscore at the end of the name. Underscores at the beginning are either for unused parameters (though in C++ you can omit their name) or implementation specific stuff (usually variables starting with double underscore).

Core/GDCore/IDE/Project/ArbitraryObjectsWorker.h
```diff
@@ -0,0 +1,63 @@
...
+
+  /**
+   * Called to do some work on an object.
+   */
+  virtual void DoVisitObject(gd::Object& object){};
+
+  /**
+   * Called to do some work on a behavior.
+   */
+  virtual void DoVisitBehavior(gd::BehaviorContent& instruction);
```
4ian, line 58: nit: behavior

Core/GDCore/IDE/Events/UsedExtensionsFinder.h
```diff
@@ -0,0 +1,62 @@
...
+  std::set<gd::String> usedExtensions;
+
+  // Object Visitor
+  void DoVisitObject(gd::Object& object);
+
+  // Behavior Visitor
+  void DoVisitBehavior(gd::BehaviorContent& behavior);
+
+  // Instructions Visitor
+  bool DoVisitInstruction(gd::Instruction& instruction, bool isCondition);
```
4ian, line 43: nit: you can use `override` keyword to ensure you're properly overriding a virtual method. Can help for silly naming/const mistakes that are hard to debug.

Core/GDCore/IDE/Project/ArbitraryResourceWorker.cpp
```diff
@@ -78,19 +80,12 @@ void ArbitraryResourceWorker::ExposeResource(gd::Resource& resource) {
 
 ArbitraryResourceWorker::~ArbitraryResourceWorker() {}
 
-void LaunchResourceWorkerOnEvents(const gd::Project& project,
+void LaunchResourceWorkerOnEvents(gd::Project& project,
```
4ian, line 83: @arthuro555 Found the problem, you're declaring a function, but this is not the same as the one in the header, because you removed the `const`.
Update the header to fix the issue.

Right now you have a function declared in the header, but which is never implemented, and a function implemented here, but which is "just" a free function declared and used nowhere.

4ian: Btw, do you really need to remove the const here?

Core/GDCore/IDE/Project/ArbitraryObjectsWorker.h
4ian:
```suggestion
  virtual void DoVisitBehavior(gd::BehaviorContent& instruction) {};
```

otherwise you'll get the  undefined reference.
Or you choose not to provide a default implementation and use `= 0` to have a pure virtual function (i.e: your class is now an abstract class). Might actually be better to avoid subclasses to avoid overriding something?

Core/GDCore/IDE/Project/ArbitraryResourceWorker.cpp
4ian: The GDevelop.js Emscripten build is also complaining about a missing function:
image

which can give another hint that you've not implemented the function you declared :)

Core/GDCore/IDE/Project/ArbitraryObjectsWorker.h
arthuro555: Well all workers will not always want to visit behaviors or ObjectContainers, maybe it's only interested in objects, it would be pretty verbose and bloating to have to write all the methods out in sub-classes.

Core/GDCore/IDE/Events/UsedExtensionsFinder.h
arthuro555: I originally used a reference but the compiler complained that I wasn't defining it in the constructor even though I was

Core/GDCore/IDE/Project/ArbitraryObjectsWorker.h
4ian: Fair enough to keep an empty implementation then!

Core/GDCore/IDE/Events/UsedExtensionsFinder.cpp
```diff
@@ -0,0 +1,143 @@
...
+      isCondition ? gd::MetadataProvider::GetExtensionAndConditionMetadata(
+                        project.GetCurrentPlatform(), instruction.GetType())
+                  : gd::MetadataProvider::GetExtensionAndActionMetadata(
+                        project.GetCurrentPlatform(), instruction.GetType());
+  usedExtensions.insert(metadata.GetExtension().GetName());
+
+  size_t i = 0;
+  for (auto expression : instruction.GetParameters()) {
+    const gd::String& parameterType =
+        metadata.GetMetadata().GetParameter(i++).GetType();
```
4ian, line 54: Nitpicking: let's avoid this clever i++ because it's easy to miss (when copy/pasting), rely on knowledge of somewhat obscure C++ corners (postfix ++ returns the original value - but that's a contract that could be not honored by some reimplementations), and so can lead to hard to debug code ("I've duplicated this line which job is to get the parameter type and now for some reason I'm missing some parameters").

A i++; on the next line will make extra explicit that we're iterating using two iterators (`expression` from the for each range, and `i`), which is an unusual construct anyway.

4ian (approved):
Apart from a nitpicking, I think we're good!

4ian:
Very nice! 
I think we could use this to also determine the dependencies of an extension on other extensions, and then compile extensions in the proper order.

Issue #1417, opened 2020-02-06
Browser page scrolls when playing top-down HTML5 game


paulera:
## Describe the bug
When playing an HTML5 top-down game from an iframe (hosted in itch.io in this case), the scrollbar responds to keyboard actions, making impossible to play the game unless switching it to fullscreen.

Game link: https://paulera.itch.io/itroubled

Game source: https://github.com/paulera/itrouble-game

## To Reproduce
* Scenario: desktop or laptop running Google Chrome or Firefox. Keyboard to play the game.
* Open the game page: https://paulera.itch.io/itroubled
* Click the "Run game" button.
* Press SPACE during the intro screen to start the level.
* During gameplay, try to move around with the arrows.

### Expected result
When pressing arrows to move, or spacebar to advance scenes, the player expects the page scroll to stay where it is, so I can always see the game when playing.

### Observed result
The page scrolls during the gameplay (responding to arrows and spacebar), making it impossible to play.

## Environment
* GDevelop 5 Beta 88
* Exported to HTML5 project and published uploading as ZIP to itch.io
* Brave Browser - Version 1.2.43 Chromium: 79.0.3945.130 (Official Build) (64-bit)
* Windows 10

## Fun facts
* Even clicking the iframe after starting the game, doesn't fix the issue. The page always scrolls.
* This same bug happens for another game built with GDevelop: https://moxvallix.itch.io/st-ufo-adventure
* Running the game in fullscreen mode works fine: https://itch.io/embed-upload/1940180

arthuro555:
I might have a workaround. Try to make that in a Js event. It needs to be executed once only. 
```js
document.onkeydown = function(e) {
    if (e.preventDefault) {
        e.preventDefault();
    };
    e.returnValue = false;
    document.onkeydown(e);
    return false;
};
``` 
It's a bit hacky and untested but should disable scrolling. I don't think the game should disable scrolling on the whole page though.

4ian:
Thanks for the detailed bug report :) 

> I don't think the game should disable scrolling on the whole page by default though.

It's a good question... arguably if the event are captured by the game, it means that it's focused and so should be safe to prevent scrolling. But we still want to allow people to use arrow keys and any key in general when the game is embedded as an iframe - something to do with the focus?

PascalLadalle:
For [this game](https://gleeson.itch.io/shrubnaut), for example, you need to click on the page outside the game for the space bar to work (scroll), even though that key isn't used in the game.
[...]

PR #2589, opened 2021-04-27
Cancel default behavior from browsers when keyboard is used

Bouh:
## Description
Add preventDefault() on keyboard events. For fix problems with game in iframe like on itch.io, this has been also reported by users,[ and recently this guys.](https://forum.gdevelop-app.com/t/arrow-keys-on-html-embedded-game/32265)

## Test
[You can test the change here ](https://witly.fr/test/test.html) with a top down behavior, use the arrow keys for moving, the default action of arrow keys is prevented.
The default keyboard event is prevented inly when the focus is on the game, the scroll continue if you focus the rest of the page.

## Question

~This preventDefault was present in past? if this has been removed, you remember why?~


Bouh:
I've found the [previous issue](https://github.com/4ian/GDevelop/issues/1417) about it!

4ian:
I think it's good to go ahead with this :)

PR #2590, opened 2021-04-27
Temporary files system improved for multi-users compatibility on Linux 

### LINKED ISSUE #2507: Linux preview opens games of other users
Defect occurs when more than one Linux user runs GDevelop on the same system.  If user A is the first to generate a preview on the system, other users thereafter may not be able preview their own games.  Instead, they can see the preview of user A's game until either the system is rebooted (clearing the /tmp directories) or an admin removes the /tmp/preview, /tmp/GDTemporaries, and /tmp/GDGeneratedEventsFunctions directories.

I'm using GDevelop 5.0.0-beta 108 based on GDevelop.js 4.0.99-release

To replicate the error...
1. Once the system is booted, user A launches GDevelop and previews a game
  -- user A's game is previewed
2. User A closes GDevelop and logs out
3. User B launches GDevelop and previews some different project
  -- User A's game is previewed again instead of B's game

I believe the cause of this error is that (at least on Linux) GDevelop creates the directories mentioned above instead of randomly generated temporary directories.  In the replication steps above, user A's preview attempt, for example, creates the /tmp/preview directory and the files it contains with read-only access to all users.  When user B attempts to preview, they cannot replace the preview directory or write other files because they do not have write permission, but GDevelop does not complain.  Instead it mysteriously opens the existing project in the preview directory.

I discovered the bug when working on the Platformer tutorial to see if it was something one of my kids would be interested in.  When they logged in to try the same tutorial, they kept getting my version.

Oxey405:
Closes #2507
Close https://forum.gdevelop-app.com/t/gdevelop-5-0-0-108-unusable-in-a-multiuser-environnment/32105 
Improved the tmp folder paths (resolved path folder conflicts) :
added the UID of the user in them so that there are no more conflicts when multiple instances of GDevelop5 are launched on different accounts on the same machine.


newIDE/app/src/Export/LocalExporters/LocalFileSystem.js
```diff
@@ -31,7 +31,7 @@ export default {
     }
   },
   getTempDir: function() {
-    return os.tmpdir();
+    return path.join(os.tmpdir(), `GDTMP-${os.userInfo().uid}`);
```
Bouh, line 34: The [NodeJS documentation](https://nodejs.org/api/os.html#os_os_userinfo_options) about `uid` explain that return -1 for Windows, that mean this will be always -1 for all users session on Windows.

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/SpritesList.js
```diff
@@ -1,340 +1,340 @@
...
-          onSpriteContextMenu={this.props.onSpriteContextMenu}
-          helperClass="sortable-helper"
-          lockAxis="x"
-          axis="x"
-        />
-      </div>
-    );
-  }
-}
+// @flow
```
Bouh, line 1: If there is no changes on this file, try to no include it in this PR.

newIDE/app/src/UI/ColorField/ColorPicker.js
```diff
@@ -1,115 +1,115 @@
...
-            <SketchPicker color={displayedColor} {...otherProps} />
-          </Popover>
-        ) : null}
-      </div>
-    );
-  }
-}
-
-export default ColorPicker;
+// @flow
```
Bouh, line 1: If there is no changes on this file, try to no include it in this PR.

Bouh (commented):
There are files that at first glance do not need to be present.
If the problem is fixed for Linux it is not for Windows according to the documentation of NodeJS on the `uid`. See my comment.

If needed, run Prettier for the code formatting.
`yarn run format`

Oxey405:
Yeah I know Arthuro runned the command but it seems that prettier did wrong !

newIDE/app/src/Export/LocalExporters/LocalFileSystem.js
arthuro555: The issue said the problem is only present on linux though, and on windows the temp directory is in app data so it is not shared and there are no such issues.

newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/PolygonEditor.js
```diff
@@ -1,162 +1,162 @@
...
-                </TableRowColumn>
-              </TableRow>
-            </TableBody>
-          </Table>
-        )}
-      </ThemeConsumer>
-    );
-  }
-}
+// @flow
```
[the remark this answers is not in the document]
Oxey405: I know git did **** I didn't tell him to do

newIDE/app/src/Export/LocalExporters/LocalFileSystem.js
Oxey405: Yes but the problem is only for linux

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/SpritesList.js
Oxey405: I know this is wrong but I don't know what happened

newIDE/app/src/UI/ColorField/ColorPicker.js
Oxey405: I know I tried my best

newIDE/app/src/Export/LocalExporters/LocalFileSystem.js
Bouh: Ok I think we can get a cleaner solution because I got this in my tmp folder.
image

I suggest from current master :
```
| - tmp 
    | - GDGeneratedEventsFunctions
    | - GDTemporaries
    | - preview
```

To this one:

```
| - tmp 
    | - GDevelop
            | - 563B4512-6857-45C7-BD8F-E44A3386C4E8   (uid or -1 for windows users)
                    | - GDGeneratedEventsFunctions
                    | - GDTemporaries
                    | - preview
```

4ian: Might not solve the problem on Linux if `tmp/GDevelop` is created and owned by a user?

4ian: I mean, I agree for a `GDevelop-userId` single folder containing subfolders, but this can be another improvement unrelated to that fix. But I think for safety we would need to have the very first folder under `/tmp/` to be indexed by the user id, to avoid issue of "I don't have the permission to write in /tmp/GDevelop because it was originally created by another user"

Bouh: Contents in `tmp` are erased when the files are not used after a bunch of time, (e.g. `cleanmgr.exe` on Windows), so we can afford to have a tree structure of this type and at least there will be a list of users sessions that won't be listed directly in the OS's tmp folder.

arthuro555: > Ok I think we can get a cleaner solution because I got this in my tmp folder.

To make it cleaner we can maybe make `LocalEventsFunctionCodeWriter.js` use the `getTempDir` from `LocalFileSystem.js` so that all use the `GDTMP-userid` folder?

arthuro555: > Contents in tmp are erased when the files are not used after a bunch of time, (e.g. cleanmgr.exe on Windows), so we can afford to have a tree structure of this type and at least there will be a list of users sessions that won't be listed directly in the OS's tmp folder.

I think you are not getting the point of the PR, on linux, when creating a folder like this, it is in a global shared `/tmp` folder which is only cleared after exiting the application or restarting the system. If two users use GDevelop at the same time, and one makes for example a preview, then the other will not be able to write the preview files as the folder already exists and another user owns it. This PR adds the userid to the folder name so that each user has their own folder preventing one from not being able to write to another one due to permissions issues.

On Windows, this is irrelevant as there are no global temp folders, the temp folder is in appdata in the user directory: it is scoped to the user only.

Additionally, those are temporary files, an implementation detail. It is not made to be a clean and userfriendly directory structure but an efficient one.

Bouh: Yes GDevelop folder can be a blocking, your solution is fine.

```
| - tmp 
    | - GDevelop-563B4512-6857-45C7-BD8F-E44A3386C4E8
                    | - GDGeneratedEventsFunctions
                    | - GDTemporaries
                    | - preview
```

> To make it cleaner we can maybe make LocalEventsFunctionCodeWriter.js use the getTempDir from LocalFileSystem.js so that all use the GDTMP-userid folder?

Yes!

Oxey405: It's no longer UID it's username I updated it.

newIDE/app/src/Export/LocalExporters/LocalPreviewLauncher/index.js
```diff
@@ -149,7 +150,7 @@ export default class LocalPreviewLauncher extends React.Component<
         new gd.AbstractFileSystemJS(),
         localFileSystem
       );
-      const outputDir = path.join(fileSystem.getTempDir(), 'preview');
+      const outputDir = path.join(fileSystem.getTempDir(), `preview-${os.userInfo().username}`);
```
4ian, line 153: I guess we don't need a user identifier here, because there is already one in `fileSystem.getTempDir()`?

4ian:
What are the pros and cons of using username vs uid in the name of the temporary folders?

Bouh:
> What are the pros and cons of using username vs uid in the name of the temporary folders?

Pros: understandable folder for human.
Cons: [Throws a SystemError if a user has no username or homedir.](https://nodejs.org/api/os.html#os_os_userinfo_options)

newIDE/app/src/Export/LocalExporters/LocalPreviewLauncher/index.js
Oxey405: No I think we need an identifier because the issue is that on linux, when multiple users are connected to the machine at the same time, GDevelop will use the same directory for all users and this cause big issues like shared previews between users.

Oxey405:
> > What are the pros and cons of using username vs uid in the name of the temporary folders?
> 
> Pros: understandable folder for human.
> Cons: [Throws a SystemError if a user has no username or homedir.](https://nodejs.org/api/os.html#os_os_userinfo_options)

Yes but actually we don't need it to be understandable by humans. The Pros is that UID on windows send -1 but Username send a string wich contain the first name of the user wich is still better than having -1

newIDE/app/src/Export/LocalExporters/LocalPreviewLauncher/index.js
4ian: Yes I understand that, but this identifier is already included in `fileSystem.getTempDir()`, so no need to have it here too, right?

Oxey405: Uhm I mean this can be a second protection ? I know this is a little bit stupid but maybe this is not that bad to let preview-{username}

4ian: Let's remove it, it's important to keep things as simple as they can be.

Oxey405:
So this pull request is denied or is it ok ?

newIDE/app/src/Export/LocalExporters/LocalPreviewLauncher/index.js
4ian: @Oxey405 This is not solved yet :)

newIDE/app/src/Export/LocalExporters/LocalFileSystem.js
```diff
@@ -31,7 +31,7 @@ export default {
     }
   },
   getTempDir: function() {
-    return os.tmpdir();
+    return path.join(os.tmpdir(), `GDTMP-${os.userInfo().username}`);
```
4ian, line 34: Each call to `userInfo` can [throw a `SystemError` exception](https://nodejs.org/api/os.html#os_os_userinfo_options). You must `catch`  (try/catch) the exception and handle it (by using an empty suffix for example). 

Otherwise, GDevelop will crash on systems that don't have a username.

Oxey405: OK

Oxey405: Fixed

Oxey405: not like you want but I think it's fixed

newIDE/app/src/Export/LocalExporters/LocalPreviewLauncher/index.js
```diff
@@ -149,13 +150,22 @@ export default class LocalPreviewLauncher extends React.Component<
...
       );
-      const outputDir = path.join(fileSystem.getTempDir(), 'preview');
-      const exporter = new gd.Exporter(fileSystem, gdjsRoot);
+      try {
+         const outputDir = path.join(fileSystem.getTempDir(), `preview-${os.userInfo().username}`);
+         const exporter = new gd.Exporter(fileSystem, gdjsRoot);
+         return {
+           outputDir,
+           exporter,
+         };
```
4ian, line 159: Avoid repeating too much code, just use the try catch around the code that is requiring it. Put it in a function if it's better for the readability :)

newIDE/app/src/EventsFunctionsExtensionsLoader/CodeWriters/LocalEventsFunctionCodeWriter.js
```diff
@@ -16,7 +16,14 @@ export const makeLocalEventsFunctionCodeWriter = ({
   onWriteFile,
 }: EventsFunctionCodeWriterCallbacks): EventsFunctionCodeWriter => {
   // The generated code for extensions will be stored in a temporary directory
-  const outputDir = os.tmpdir() + '/GDGeneratedEventsFunctions';
+  var outputDirTemp ;
+  try {
+    outputDirTemp = os.tmpdir() + `/GDGeneratedEventsFunctions-${os.userInfo().username}`;
+  }
+  catch (error) {
+    outputDirTemp = os.tmpdir() + `/GDGeneratedEventsFunctions-${os.userInfo().uid}`;
```
4ian, line 24: It's not accessing `username` that will throw, but actually calling `userInfo()`. So this will throw again and crash GDevelop.
Take a step back and consider how you can nicely use a function for this. It should be no more than a few lines.

newIDE/app/src/EventsFunctionsExtensionsLoader/CodeWriters/LocalEventsFunctionCodeWriter.js
```diff
@@ -16,7 +16,14 @@ export const makeLocalEventsFunctionCodeWriter = ({
   onWriteFile,
 }: EventsFunctionCodeWriterCallbacks): EventsFunctionCodeWriter => {
   // The generated code for extensions will be stored in a temporary directory
-  const outputDir = os.tmpdir() + '/GDGeneratedEventsFunctions';
+  var outputDirTemp ;
```
4ian, line 19: Never use var, always `const`.

newIDE/app/src/Export/LocalExporters/LocalPreviewLauncher/index.js
Oxey405: I did it at first but this caused issues so I was forced to do like so.

newIDE/app/src/EventsFunctionsExtensionsLoader/CodeWriters/LocalEventsFunctionCodeWriter.js
Oxey405: Again I was forced to do like so. And this variable is temporary. It's because there was issues with variable scopes.

4ian: You can always abstract the call to `os.userInfo()` and the try/catch in a function ;)

Oxey405: Can we define a blank const and set the value later ?

4ian: No - the goal with `const` is to never change your variable. Here the code is too complicated because the try catch is forcing you to use var/let.
Instead, make a function to abstract the call to userInfo(). You can call this function `getUsername`. In this function, you get first the username and return it. If there is an error, you return an empty string. 
Simple and clean :)

Oxey405: Well I wasn't aware we can do that. I modified it but not the way you wanted to so I guess I have to start again ?

newIDE/app/src/Export/LocalExporters/LocalFileSystem.js
```diff
@@ -11,7 +11,7 @@ const gd /* TODO: add flow in this file */ = global.gd;
  * supported everywhere).
  */
 export default {
-  mkDir: function(path) {
+  mkDir: function (path) {
```
4ian, line 14: When you push changes, don't push changes to files not related to your PR.

newIDE/app/src/EventsFunctionsExtensionsLoader/CodeWriters/LocalEventsFunctionCodeWriter.js
4ian: I know it's frustrating :) But yes you should really follow what I recommend here (make a tiny function `getUsername`, it's only a few lines). 
It's important to keep a code as concise as possible.

newIDE/app/src/EventsFunctionsExtensionsLoader/CodeWriters/LocalEventsFunctionCodeWriter.js
```diff
@@ -16,14 +16,14 @@ export const makeLocalEventsFunctionCodeWriter = ({
   onWriteFile,
 }: EventsFunctionCodeWriterCallbacks): EventsFunctionCodeWriter => {
   // The generated code for extensions will be stored in a temporary directory
-  var outputDirTemp ;
+  const outputDirTemp = { value: "" };
```
4ian, line 19: Unfortunately this is making the code just more complex. Read again what I recommend :)

Oxey405: Yeah I saw it too late ! I'm sorry
I am starting to doubt of my capacities about resolving this issue

newIDE/app/src/EventsFunctionsExtensionsLoader/CodeWriters/LocalEventsFunctionCodeWriter.js
Oxey405: i'm sorry but... I don't know what is an "abstract"

newIDE/app/src/Export/LocalExporters/LocalPreviewLauncher/index.js
```diff
@@ -149,13 +150,24 @@ export default class LocalPreviewLauncher extends React.Component<
...
         localFileSystem
       );
-      const outputDir = path.join(fileSystem.getTempDir(), 'preview');
-      const exporter = new gd.Exporter(fileSystem, gdjsRoot);
-
-      return {
-        outputDir,
-        exporter,
-      };
+      try {
```
4ian, line 153: You don't need to do any changes in this file, because the uid is already handled in the file system.

newIDE/app/src/Export/LocalExporters/LocalPreviewLauncher/index.js
```diff
@@ -149,13 +150,24 @@ export default class LocalPreviewLauncher extends React.Component<
...
+        const outputDir = path.join(fileSystem.getTempDir(), `preview`);
+        const exporter = new gd.Exporter(fileSystem, gdjsRoot);
+        return {
+          outputDir,
+          exporter,
+        };
+      } catch (error) {
+        const outputDir = path.join(
+          fileSystem.getTempDir(),
+          `preview-${os.userInfo().uid}`
```
4ian, line 163: You also inverted the catch and the try. But again, there is no need for uid here and no catch. Revert the changes back the how it was before the PR :)

newIDE/app/src/Export/LocalExporters/LocalPreviewLauncher/index.js
Oxey405: Oh... Ok...

newIDE/app/src/Export/LocalExporters/LocalPreviewLauncher/index.js
Oxey405: OK

newIDE/app/src/Export/LocalExporters/LocalPreviewLauncher/index.js
4ian: I know it's long, but you're almost done :)

newIDE/app/src/Export/LocalExporters/LocalPreviewLauncher/index.js
Oxey405: Done

Oxey405:
I think I finally managed to correct everything @4ian !

Oxey405:
So for recap, the tmp folder of GDevelop is `GDTMP-[uid]`.
If we CAN'T get the uid we return nothing (sadly)
This folder belongs to the actual user with this UID.
Inside this `GDTMP-[uid]` we have all the tmp folders.
So if multiples users share the same tmp we are going to have mutliple folders. 
Each one is attached to the user that have the uid.
I hope I explained goodly !

Oxey405:
Also, if the CI fail it's because this fork is based upon a one that has failed the CI.

Oxey405:
@Bouh or @4ian could anyone review the last push ?

newIDE/app/src/Utils/GetUserUID.js
```diff
@@ -0,0 +1,8 @@
+var os = optionalRequire('os');
+function getUsername(): string {
```
4ian, line 2: It's a good idea to do this 👍

This being said, have you tried your changes?

You need to add `// @flow` at the top of the file, then to import `optionalRequire` and finally to `export` the function.
When it's done, you can import your function in the two other files.
Test that it works and it will be good.

Also rename the file `GetUsername` or better, `LocalUserInfo.js`.

Oxey405: Well arthuro told me to do otherwise !

Oxey405: I completly revamped it !

newIDE/app/src/Export/LocalExporters/LocalFileSystem.js
```diff
@@ -31,7 +32,7 @@ export default {
     }
   },
   getTempDir: function() {
-    return os.tmpdir();
+    return path.join(os.tmpdir(), `GDTMP-${getUID()}`);
```
Bouh, line 35:
```JS
return path.join(os.tmpdir(), 'GDTMP-' + getUID());
```
Same make simple, what you think?

newIDE/app/src/EventsFunctionsExtensionsLoader/CodeWriters/LocalEventsFunctionCodeWriter.js
```diff
@@ -16,7 +17,9 @@ export const makeLocalEventsFunctionCodeWriter = ({
   onWriteFile,
 }: EventsFunctionCodeWriterCallbacks): EventsFunctionCodeWriter => {
   // The generated code for extensions will be stored in a temporary directory
-  const outputDir = os.tmpdir() + '/GDGeneratedEventsFunctions';
+
+  const outputDir =
+    os.tmpdir() + `/GDGeneratedEventsFunctions-` + getUID();
```
Bouh, line 22:
```JS
path.join(os.tmpdir(), 'GDGeneratedEventsFunctions-' + getUID());
```
What do you think?

newIDE/app/src/Utils/GetUserUID.js
```diff
@@ -0,0 +1,10 @@
+const getUID = () => var os = optionalRequire('os');
```
Bouh, line 1:
```JS
// @flow
const os = optionalRequire('os');

function getUID(): string {
  try {
    return os.userInfo().uid;
  } catch (e) {
    return '';
  }
};
export getUID;
```

4ian, line 1: This function seems to make no sense?

Oxey405: Fixed

Oxey405: fixed

newIDE/app/src/EventsFunctionsExtensionsLoader/CodeWriters/LocalEventsFunctionCodeWriter.js
Oxey405: Fixed

Oxey405: It's format

Oxey405: Ask to the auto-formatter.
I don't know why too.

Oxey405: Oh ok I understand

Oxey405: Well this is working. And this is correct

Oxey405: But if you **insist**

Oxey405: Finally fixed

newIDE/app/src/Export/LocalExporters/LocalFileSystem.js
Oxey405: DONE

newIDE/app/src/Utils/GetUserUID.js
Oxey405: done !

4ian (approved):
Did a few additional changes, looks good! Will merge when tests are passing.

Oxey405:
> Did a few additional changes, looks good! Will merge when tests are passing.

CI passed

4ian:
Merged! Thanks for working on this! :)

Oxey405:
You did some of the job....

PR #2594, opened 2021-04-29
[LinkedObjects] Improve the type definition

D8H:
To make it easier to use in JavaScript events.

Should the "links" attribute be prefixed with "_"?


4ian:
> Should the "links" attribute be prefixed with "_"?

Ideally it should have been yes! Might be a bit late for this, unless it's never used anywhere in extensions and JS code, but hard to check..

4ian:
Let's go ahead with this! Thanks 👍

PR #2598, opened 2021-05-01
[P2P] Allow overriding the Peer client ID

hedlund:
The default behaviour of PeerJS is to generate a UUID for each client that connects to the server. However, it also allows the clients themselves to specify their own ID, which could be useful in some situations.

The typical scenario would be a pretty standard situation, where a player is allowed to "host" a game, and then other players can connect to that instance using a code. A UUID, which is the default, is difficult to share and input, so using something simpler  (such as a 6-digit number or a shorter string) makes it easier to communicate.

Using this patch, a game could generate a simple, but unique, ID for each game (e.g. `639239`), and then specify that as the PeerJS client ID: `P2P::OverrideID('639239')`. Other players can then connect to that peer using the simpler ID.

I specifically avoided using `SetID` as the function name, as this operation isn't really a simple "set" in the traditional sense. It may have some consequences, especially if the ID is not unique enough and you are using the default broker. Thus, I guess one would also only recommend to use this operation when using your own hosted instance of PeerJS. An alternative version could be to include an optional peer ID when using the `UseOwnBroker` function instead.

### DIFF
--- a/Extensions/P2P/B_p2ptools.ts
+++ b/Extensions/P2P/B_p2ptools.ts
@@ -82,6 +82,11 @@ namespace gdjs {
       }
 
       /**
+       * The optional peer ID. Only used if explicitly overridden.
+       */
+      let peerId: string | null = null;
+
+      /**
        * The peer to peer configuration.
        */
       let peerConfig: Peer.PeerJSOption = { debug: 1 };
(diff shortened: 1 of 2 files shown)


4ian (approved):
Looks great! Well documented and very clean 👍

PR #2602, opened 2021-05-02
[Platformer] Make the platforms AABB updates more predictable

D8H:
Note:
This is extracted from https://github.com/4ian/GDevelop/pull/2574
I will split the other fixes one by one. This will be easier to follow.
I choose this one 1st because it make the unit tests more predictable too.

### Context
* A platform is teleported from one location to another.

### Actual
* Platforms AABB are updated to the RBush in the doStepPreEvents method.
* The doStepPreEvents call order  does not impose Platforms AABB to be updated before that platformer objects doStepPreEvents  are called.
* As a result, the spacial search may be inconsistent with the actual location of the object (getX(), getHeight()...).

Running the new test on the master give this failure because at the 1st doStepPreEvents call on the platformer object the platform AABB is still the one from onCreate and the position set by the test will only be taken into account when the doStepPreEvents of the platform is called. So the object falls.

```
HeadlessChrome 90.0.4430 (Windows 10.0.0) gdjs.PlatformerObjectRuntimeBehavior (falling) must not move down when on the floor at startup FAILED
        Error: expected -29.75 to be below -30.01
            at Assertion.lessThan.Assertion.below (node_modules/expect.js/index.js:313:10)
            at Function.below (node_modules/expect.js/index.js:499:17)
            at context.<anonymous> (GDevelop/Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js:174:35)
```

### Expected
* Platforms AABB are updated at the 1rst call the doStepPreEvents of a platformer object.
* The spacial search is consistent with the location of the object.


Extensions/PlatformBehavior/platformruntimebehavior.ts
```diff
@@ -42,11 +46,32 @@ namespace gdjs {
       return runtimeScene.platformsObjectsManager;
     }
 
+    /**
+     * Update the AABB given to the spacial search of every platforms.
+     * It's only done once per frame.
+     * @param runtimeScene
+     */
+    updateAABBIfNeeded(runtimeScene: gdjs.RuntimeScene) {
+      const time = runtimeScene.getTimeManager().getTimeFromStart();
```
4ian, line 55: I guess you're using the time as a way to check if it's worth updating the AABB... but the time will always be the same during a frame, so how is this useful exactly?

4ian: Ok I actually understand that this is useful because as you said "The doStepPreEvents call order does not impose Platforms AABB to be updated before that platformer objects doStepPreEvents are called.".
I wonder though if there is another way to do it.

D8H: I took a look at the call order in renderAndStep
* _updateObjectsPreEvents
* _eventsFunction
* _updateObjectsPostEvents
* render

Doing the platform AABB update in doStepPostEvents should work because it happens between event and doStepPreEvents. There is still the case when the platformer object is a platform but it can be updated without using a Set.

In case of unit tests, the tests modify the objects before _updateObjectsPreEvents, doesn't it? If so, I think that tests should modify the objects at the time events would.
For assertions, I find it important to be able to check what the events get in addition to what is displayed on the screen. What do you think about a TestRuntimeScene that split renderAndStep in these 2 methods:
* updateObjectsPreEvents(elapsedTime: float)
* updateObjectsPostEventsAndRender(elapsedTime: float)

Silver-Streak:
Hey @D8H, Just checking in on these.

What are these fixes waiting on? Is there something I can help with on testing?

(expect to see this on each of the PRs so I don't lose track of any of them. 😄 )

Extensions/PlatformBehavior/platformruntimebehavior.ts
D8H: I updated my previous message because it was not very understandable.
Do you think this solution would be better than what is implemented in this PR?
Did you had another solution in mind?

PR #2603, opened 2021-05-02
[Platformer] Fix the platformer object small movement when it is on a platform at startup

D8H:
Note:
this is extracted from: https://github.com/4ian/GDevelop/pull/2574
and commited over https://github.com/4ian/GDevelop/pull/2602

### Context
* This is the issue reported here: https://github.com/4ian/GDevelop/issues/2562
* At the 1st frame
* The platformer object is on a platform
* The intial state is always Falling

### Actual
* There is no transition from Falling to OnFloor possible before the movement is applied.
* The platformer object falls for one 1 frame and transition to OnFloor after moving.

```
HeadlessChrome 90.0.4430 (Windows 10.0.0) gdjs.PlatformerObjectRuntimeBehavior (falling) must not move when on the floor at startup FAILED
        Error: expected -31 to equal -30
            at Assertion.be.Assertion.equal (node_modules/expect.js/index.js:216:10)
            at Assertion.<computed> [as be] (node_modules/expect.js/index.js:69:24)
            at context.<anonymous> (GDevelop/Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js:159:34)
```

### Expected
* It transitions to Onfloor before moving.
* The position stays the same.

### Context
* The platformer object is on a platform.
* This platform is teleported over it's top without any contact.
* For instance, this could happens in a game when the player must go up jumping on a jumpthru that teleports up in sequence.

### Actual
* The transition from Onfloor to Falling before moving only uses the spacial search without collision test.
* The platformer object follows the platform.

```
HeadlessChrome 90.0.4430 (Windows 10.0.0) gdjs.PlatformerObjectRuntimeBehavior (rounded coordinates, moving platforms) must not follow a platform that is moved over its top FAILED
        Error: expected true to equal false
            at Assertion.be.Assertion.equal (node_modules/expect.js/index.js:216:10)
            at Assertion.<computed> [as be] (node_modules/expect.js/index.js:69:24)
            at context.<anonymous> (GDevelop/Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js:960:58)
```

### Expected
* The platformer object doesn't follow the platform.


Silver-Streak:
Hey @D8H, Just checking in on these.

What are these fixes waiting on? Is there something I can help with on testing?

(expect to see this on each of the PRs so I don't lose track of any of them. 😄 )

PR #2606, opened 2021-05-03
[P2P] Add warning hint about overriding the Peer ID

hedlund:
As discussed with @arthuro555 on PR #2598, adding a warning text about not using the `P2P::OverrideID` feature unless you really know what you're doing.

I wasn't really sure about the best practice for the hints - if I should specify an ID or not, etc..


arthuro555 (approved):

4ian:
Thanks! Looks good

PR #2610, opened 2021-05-05
Add queries to firebase

arthuro555:
Adds queries to Firestore of the Firebase extension.


Extensions/Firebase/B_firebasetools/D_cloudfirestoretools.ts
```diff
@@ -329,6 +552,8 @@ namespace gdjs {
          * @param collectionName - The collection where to count documents.
          * @param [callbackValueVariable] - The variable where to store the result.
          * @param [callbackStateVariable] - The variable where to store if the operation was successful.
+         *
+         * @deprecated Use a query without filters instead.
          */
         export const listDocuments = (
```
arthuro555, line 558: I think this should be deprecated as it is basically doing a query so all documents have been fetched, but it only gives the IDs of the documents to the user, making them have to refetch those individually. Using an empty query allows having access to the fetched documents data, which is a better solution.

Extensions/Firebase/B_firebasetools/D_cloudfirestoretools.ts
```diff
@@ -356,6 +586,28 @@ namespace gdjs {
...
             });
         };
+
+        /**
+         * Returns a special string replaced by a firebase serverTimestamp field value.
+         */
+        export const getServerTimestamp = () =>
+          '[{__FIREBASE_SERVERSIDE_TIMESTAMP}]';
+
+        const replaceTimstampInString = (str: any) => {
```
4ian, line 596: nit: `replaceTimestampInString`

Extensions/Firebase/B_firebasetools/D_cloudfirestoretools.ts
```diff
@@ -356,6 +586,28 @@ namespace gdjs {
...
+          if (str === '[{__FIREBASE_SERVERSIDE_TIMESTAMP}]')
+            return firebase.firestore.FieldValue.serverTimestamp();
+          else return str;
+        };
+
+        const replaceTimstampsInObject = (object: object): object => {
```
4ian, line 602: `replaceTimestampInString`

Extensions/Firebase/JsExtension.js
```diff
@@ -1048,6 +1048,337 @@ module.exports = {
...
+        'Extensions/Firebase/A_firebasejs/B_firebase-firestore.js'
+      )
+      .addIncludeFile('Extensions/Firebase/B_firebasetools/C_firebasetools.js')
+      .addIncludeFile(
+        'Extensions/Firebase/B_firebasetools/D_cloudfirestoretools.js'
+      )
+      .setFunctionName('gdjs.evtTools.firebaseTools.firestore.startQueryFrom');
+
+    const operatorDesc =
+      'See the [Firebase documentation]' +
```
4ian, line 1122: Nice 👍

Extensions/Firebase/JsExtension.js
```diff
@@ -1048,6 +1048,337 @@ module.exports = {
...
+      .addIncludeFile('Extensions/Firebase/B_firebasetools/C_firebasetools.js')
+      .addIncludeFile(
+        'Extensions/Firebase/B_firebasetools/D_cloudfirestoretools.js'
+      )
+      .setFunctionName('gdjs.evtTools.firebaseTools.firestore.queryOrderBy');
+
+    extension
+      .addAction(
+        'FirestoreQueryLimit',
+        _('Limit ammount of documents'),
```
4ian, line 1228: `Limit amount of documents`

Extensions/Firebase/JsExtension.js
```diff
@@ -1048,6 +1048,337 @@ module.exports = {
...
+        ),
+        _(
+          'Limit query _PARAM0_ to _PARAM1_ documents (begin from the end: _PARAM2_)'
+        ),
+        _('Firebase/Cloud Firestore/Queries/Filters'),
+        'JsPlatform/Extensions/firebase.png',
+        'JsPlatform/Extensions/firebase.png'
+      )
+      .addParameter('string', _('Query name'), '', false)
+      .addParameter('expression', _('Ammount to limit by'), '', false)
```
4ian, line 1240: Amount

Extensions/Firebase/B_firebasetools/D_cloudfirestoretools.ts
```diff
@@ -6,6 +6,230 @@ namespace gdjs {
...
+                direction
+              )
+            );
+        };
+
+        /**
+         * Limits the amount of documents returned by the query.
+         *
+         * @param queryID - The query to add the filter to.
+         * @param ammount - The amount of documents to limit to
```
4ian, line 121: nit everywhere: ammount => amount

Extensions/Firebase/B_firebasetools/D_cloudfirestoretools.ts
```diff
@@ -6,6 +6,230 @@ namespace gdjs {
...
+            firebase.firestore.WhereFilterOp,
+            // Exclude unsupported "batch" operations (as they require an array as value to check)
+            'in' | 'array-contains-any' | 'not-in'
+          >,
+          value: string | number
+        ) => {
+          if (queries.has(queryID))
+            queries.set(
+              queryID,
+              (queries.get(queryID) as firebase.firestore.Query).where(
```
4ian, line 87: Strange that you have to do these casts no? 🤔

Extensions/Firebase/JsExtension.js
```diff
@@ -1048,6 +1048,337 @@ module.exports = {
...
+      )
+      .addParameter('string', _('Query name'), '', false)
+      .addParameter(
+        'scenevar',
+        _('Callback variable where to load the results'),
+        '',
+        true
+      )
+      .setParameterLongDescription(
+        'See the shape of the returned data on [the wiki page]().'
```
4ian, line 1363: Missing link :)

4ian (commented):
I've not scrutinised absolutely everything, but this seems great overall! Added a few nitpickings.

Extensions/Firebase/B_firebasetools/D_cloudfirestoretools.ts
arthuro555: Well, I guess typescript is not smart enough to know that if Map.has is true, Map.get cannot return null

Extensions/Firebase/JsExtension.js
arthuro555: Ah indeed, I need to write that page 😅

Extensions/Firebase/B_firebasetools/D_cloudfirestoretools.ts
4ian: Ah right! Note that you could try to use the `!` syntax (`queries.get(queryID)!.where(`) to remove the nullability, if that's the problem (it's not better, it's just shorter). Otherwise not worries.

arthuro555: https://github.com/Microsoft/TypeScript/issues/9619

4ian (approved):

4ian:
Merged, thanks! And thanks for the tests 👍

PR #2611, opened 2021-05-05
Tween Scale from center (or named center point) of objects

aphoenixholland:
When using the scaleXY, scaleX or scaleY tweens, a new yes/no parameter allows you to anchor the object to its center for the scale transform, not the top/left origin. For objects that have the center point moved, this point is the anchor for the transform.
(This is my second attempt at this pull/push request; the first has been closed.)

### DIFF
--- a/Extensions/TweenBehavior/tweenruntimebehavior.ts
+++ b/Extensions/TweenBehavior/tweenruntimebehavior.ts
@@ -415,6 +415,25 @@ namespace gdjs {
       const newTweenable = TweenRuntimeBehavior.makeNewTweenable(
         this._runtimeScene
       );
+      let stepFunction;
+      if(scaleFromCenterOfObject) {
+        stepFunction = function step(state) {
+          const oldX = that.owner.getCenterXInScene();
+          const oldY = that.owner.getCenterYInScene();
+          // @ts-ignore - objects are duck typed
+          that.owner.setScaleX(state.scaleX);
+          // @ts-ignore - objects are duck typed
+          that.owner.setScaleY(state.scaleY);
+          that.owner.setCenterPositionInScene(oldX, oldY);
+        }
+      } else {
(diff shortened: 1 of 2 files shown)


Extensions/TweenBehavior/tweenruntimebehavior.ts
```diff
@@ -424,10 +426,20 @@ namespace gdjs {
         duration: durationValue,
         easing: easingValue,
         step: function step(state) {
-          // @ts-ignore - objects are duck typed
-          that.owner.setScaleX(state.scaleX);
-          // @ts-ignore - objects are duck typed
-          that.owner.setScaleY(state.scaleY);
+          if(scaleFromCenterOfObject) {
```
4ian, line 429: Nitpicking: to avoid the extra boolean check when running the step function, we could have `originStep` and `centerStep` function, and set the proper one when calling `newTweenable.setConfig`.
Not a huge change,  but can save a very tiny bit on each step function which can run on a lot of objects.

4ian (commented):
Looks great overall! Minor nitpicking in comments :)

Extensions/TweenBehavior/tweenruntimebehavior.ts
aphoenixholland: I've implemented this slightly differently, changing the body of the function based on the condition. Although it would work (probably better) with named functions, this seemed overkill for how it's being used, but, I'm happy if you want to refactor it!

4ian: I think this is equivalent and fine 👍

4ian:
Looks great! Can you run `npm run format` in GDJS folder? :)

4ian (approved):

4ian:
Merged! I did a few fixes before that. Thanks for working on that! Will be very useful! 👍

PR #2635, opened 2021-05-11
Fix same points set to true for an instant

### LINKED ISSUE #2631: Edit Points - Share Same Points overriding all animation points
## Describe the bug
You know when you swear you have changed something, only come back and fix it again. That's me right now.

In the new editor, it seems that for a split second the "Share same points for all animations" gets enabled (even if disabled) which results in all animations sharing the points.

At this point, I'd suggest anyone using 109 to not change any objects points unless you want them over-ridden :(

## To Reproduce
1. Have 2 separate animations. First one with an extra point.
2. Disable "share same points" on both animations
3. Close apply changes.
4. Re-open object + points.
5. See it enabled for a split second and all points applied to other animations.

Video showing the issue on test project: https://www.youtube.com/watch?v=6JzctEbHt3o
The test project attached. [Animation Bug.zip](https://github.com/4ian/GDevelop/files/6456186/Animation.Bug.zip)

## Other details
* Windows 10
* 5.0.0-beta109

nilaymaj:
Fixes #2631.
This bug stemmed from refactoring PointsEditor to a functional component - change in the component lifecycle meant that the default true values for React states `samePointsForAnimations` and `samePointsForSprites` were used for the first `updatePoints` call.

Right now I've just added a simple fix - the default values are now `false`. So that instant where the values are not set to the correct values does not have any side effects. An ideal fix would safeguard the component against the transient state, I haven't tried that yet.

### DIFF
--- a/newIDE/app/src/ObjectEditor/Editors/SpriteEditor/PointsEditor/index.js
+++ b/newIDE/app/src/ObjectEditor/Editors/SpriteEditor/PointsEditor/index.js
@@ -51,9 +51,9 @@ const PointsEditor = (props: Props) => {
   const [directionIndex, setDirectionIndex] = React.useState(0);
   const [spriteIndex, setSpriteIndex] = React.useState(0);
   const [samePointsForAnimations, setSamePointsForAnimations] = React.useState(
-    true
+    false
   );
-  const [samePointsForSprites, setSamePointsForSprites] = React.useState(true);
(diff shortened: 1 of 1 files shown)


4ian:
> An ideal fix would safeguard the component against the transient state, I haven't tried that yet.

Yeah ideally the state should always be up to date/in sync with what its real value should be. Should be enough for now as a quick fix!

PR #2645, opened 2021-05-13
EffectsContainer Refactor [WIP]

HarsimranVirk:
Hey @4ian, I'm not really familiar with C++, so there could be many silly mistakes. Also, I've some queries:

1. `class GD_CORE_API EffectsContainer`: From what I could figure out, `GD_CORE_API` is essentially defined as _nothing_, which probably means that after the pre-processing stage, it would've been replaced by _nothing_, so what purpose does it really serve?
2. Why do we use `namespace name { class classname;}`? My guess is that it's a way to tell the compiler that a class with this name exists (while not including the header).
3. Also, now that `EffectsContainer` is somewhat decoupled from `Layer`, how should we handle serialization? I believe it should be done entirely in `EffectsContainer` so that we don't have to implement it multiple times.


Core/GDCore/Project/Layer.cpp
```diff
@@ -13,7 +14,11 @@ namespace gd {
 Camera Layer::badCamera;
 Effect Layer::badEffect;
 
-Layer::Layer() : isVisible(true), isLightingLayer(false), followBaseLayerCamera(false) {}
+Layer::Layer()
+    : isVisible(true),
+      isLightingLayer(false),
+      followBaseLayerCamera(false),
+      effectsContainer(std::make_unique<EffectsContainer>()) {}
```
HarsimranVirk, line 21: Is there a need to initialize it? Or is it already initialized like `std::vector`? Also, should it be a unique_ptr or a shared_ptr? As I could understand, EffectsContainer and Layer have a composition relationship, so I'm not sure if an EffectContainer object will be shared or not 🤔.

4ian:
> `GD_CORE_API` is essentially defined as nothing

Correct. This date backs to a need to "export" symbols like classes in DLL for Windows (https://docs.microsoft.com/en-us/cpp/cpp/dllexport-dllimport?view=msvc-160). Not useful anymore in Emscripten and never was used on Linux. Still keeping it for now but we could remove these.

> Why do we use namespace name { class classname;}?

This is a "forward declaration". In C++, the compiler needs to know the size of the structures/classes (so that it can know how to organise them in memory). This means that if a struct/class is having another as a member, then it needs to know the included class. So you need to #include the file containing its declaration.

This is time consuming for the compiler (because of opening the files, so more I/O operations).
But if you use just a reference to a class (or a pointer), then the compiler does not need to know anything more (as long as your code is just a header file declaring a class, and not using it in an implementation) - because the size in memory of a pointer/reference is well known.

So you can just "forward declare" a class: `class MyClass; // <-- note it can be used in the declaration of a class`. Because our classes are namespaced, we need to add this namespace though.

Note that as soon as you use the class (by calling a method, accessing a member), then you need to really #include the file declaring it (because the compiler needs to know it entirely this time, to generate the code for it).

Rule of thumb is:
- in a .h file, forward declare as much as possible.
- if you're using only a pointer/reference, you can forward declare in the .h
- if it's a member that is not a pointer/reference, you need to #include it.
- if you're in the .cpp file, #include it.

> Also, now that EffectsContainer is somewhat decoupled from Layer, how should we handle serialization? I believe it should be done entirely in EffectsContainer so that we don't have to implement it multiple times.

Serialization should indeed be handled by EffectsContainer. Layer can call it, passing it the SerializerElement where to read/store effects :)

Core/GDCore/Project/EffectsContainer.h
```diff
@@ -0,0 +1,99 @@
...
+#include "GDCore/String.h"
+
+namespace gd {
+class Effect;
+}  // namespace gd
+
+namespace gd {
+
+/**
+ * \brief Contains the effect applied to other entities (ie, Layer or Object)
```
4ian, line 20: ubernitpicking: "i.e: " ;) Also end with a period.

`Contains effects applied to an entity on screen (i.e: a Layer or an Object).`

Core/GDCore/Project/EffectsContainer.h
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * \see gd::EffectsContainer
```
4ian, line 22: You can remove this

Core/GDCore/Project/EffectsContainer.h
```diff
@@ -0,0 +1,99 @@
...
+
+  /**
+   * Add a new effect at the specified position in the effects list.
+   */
+  gd::Effect& InsertNewEffect(const gd::String& name, std::size_t position);
+
+  /**
+   * \brief Add a copy of the specified effect in the effects list.
+   *
+   * \note No pointer or reference must be kept on the layer passed as
```
4ian, line 72: there was a mistake it seems: `on the effect`.

Core/GDCore/Project/EffectsContainer.h
```diff
@@ -0,0 +1,99 @@
...
+  void RemoveEffect(const gd::String& name);
+
+  /**
+   * Swap the position of two effects.
+   */
+  void SwapEffects(std::size_t firstEffectIndex, std::size_t secondEffectIndex);
+
+ private:
+  std::vector<std::shared_ptr<gd::Effect>>
+      effects;  ///< The effects contained by the container.
```
4ian, line 93: You can almost remove this comment, it's obvious that it's the effects

Core/GDCore/Project/Layer.h
```diff
@@ -219,16 +229,18 @@ class GD_CORE_API Layer {
...
+                         ///< renders an ambient light.
+  bool followBaseLayerCamera;  ///< True if the layer automatically follows the
+                               ///< base layer
+  unsigned int ambientLightColorR;  ///< Ambient light color Red component
+  unsigned int ambientLightColorG;  ///< Ambient light color Green component
+  unsigned int ambientLightColorB;  ///< Ambient light color Blue component
   std::vector<gd::Camera> cameras;  ///< The camera displayed by the layer
-  std::vector<std::shared_ptr<gd::Effect>>
-      effects;  ///< The effects applied to the layer.
+  std::unique_ptr<gd::EffectsContainer>
```
4ian, line 242: No need for a unique_ptr here. Just use a `gd::EffectsContainer effectsContainer; ///< Th effects applied to the layer`.
By using this, the compiler will properly copy the effects container for you when a layer is copied, it will initialize it, and you don't need to worry about the copy constructor too.

Core/GDCore/Project/Layer.cpp
4ian: As said in the other comment, remove the usage of a unique ptr. Because this is a part of the class (a layer always has a container of effects (even if this container is empty)), we don't need a pointer, we're happy to have it as a traditional member.
This being said, the container itself is using shared_ptr, so that when we re-organise the order of effects, the pointers are staying valid :)

Core/GDCore/Project/EffectsContainer.h
```diff
@@ -0,0 +1,99 @@
...
+
+namespace gd {
+
+/**
+ * \brief Contains the effect applied to other entities (ie, Layer or Object)
+ *
+ * \see gd::EffectsContainer
+ * \ingroup PlatformDefinition
+ */
+class GD_CORE_API EffectsContainer {
```
4ian, line 25: So this is indeed the idea, though I'm surprised because there seems to be a missing thing in the existing gd::Layer: a custom copy constructor and assignement operator.

The reason is that by default, the compiler will generate a default copy constructor (used when you do `gd::EffectsContainer newCopy = otherContainer;`) or assignement operator (used when you do `someContainer = otherContainer;`).

By default, C++ generates a constructor/operator that just iterate on the members, and copy them. That's fair enough! And almost always what we want
BUT for a vector of shared_ptr, it will copy all shared_ptr... which means that if you do `someContainer = otherContainer;`, then `someContainer` and `otherContainer` will now have two different vector in memory (so far so good), containing different pointers (so far so good), but all these pointers will have the same values! So they will point to the same effects!
(Imagine the confusion: you duplicate a layer, then you edit some effects, and the effects of the other layers are changed too!).

What a pain. This is because the "semantics" of copying a shared_ptr is "just create another pointer to the SAME object" (much like actually in JavaScript :)). So what do we do? We re-implement these copy constructor/assignement, so that they have a semantic which is to COPY the objects of the container.

Usually I do something like this (search VariablesContainer.cpp):
```
VariablesContainer::VariablesContainer(const VariablesContainer& other) {
  Init(other);
}

VariablesContainer& VariablesContainer::operator=(
    const VariablesContainer& other) {
  if (this != &other) Init(other);

  return *this;
}

void VariablesContainer::Init(const gd::VariablesContainer& other) {
  variables.clear();
  for (auto& it : other.variables) {
    variables.push_back(
        // In your case you won't need the std::make_pair, because we're not using a pair, just a shared_ptr
        std::make_pair(it.first, std::make_shared<gd::Variable>(*it.second)));
  }
}
```

With this, we should be safe. Copying a container will work "as we could expect" by copying everything inside.

HarsimranVirk: > So they will point to the same effects!
(Imagine the confusion: you duplicate a layer, then you edit some effects, and the effects of the other layers are changed too!).

I've been dealing with the exact thing for the last 10 mins 😆. Apart from that, I've added serialization and everything seems to be working well. I'll push the code after making changes these changes.

HarsimranVirk:
@4ian, another query, I didn't add anything in the `Bindings.idl` till now, but I'm able to compile and run everything (except there is still the issue of copy constructor). Is it because we aren't directly using the `EffectsContainer` through js (and therefore don't require any binding definitions for now)?

4ian:
> another query, I didn't add anything in the Bindings.idl till now, but I'm able to compile and run everything (except there is still the issue of copy constructor). Is it because we aren't directly using the EffectsContainer through js (and therefore don't require any binding definitions for now)?

Your C++ file should be at least compiled (even if not used). 
Bindings.idl is here to expose to JavaScript a JS "class" that is able to encapsulate and transfer calls to the C++ class, so that you can use your C++ class in JavaScript like any other.

For now, it's working well without exposing it in Bindings.idl because you already have the `gd::Layer` class and methods exposed in Bindings.idl. So even if the class `gd::EffectsContainer` is not exposed, the code still works :) 
As soon as you want to manipulate a `gd::EffectsContainer` in JS (either by doing something like `object.getEffects()` or by anything else that returns/create a `gd::EffectsContainer`), you'll need to expose it.

The issue with the copy/assignement operator is not causing any issue it seems, but better fix it now before we have strange copy issues arising :) 

When it's ready, we'll be able to add (maybe in another PR, as you prefer) a `gd::EffectsContainer effects;` to `gd::Object`, and expose getters for it (`GetEffects()`, that returns a reference to it). At this point we can add it to Bindings.idl, and finally use pass use the getter in the IDE to pass it to the `EffectsList` :)

Core/GDCore/Project/EffectsContainer.h
4ian: > I've been dealing with the exact thing for the last 10 mins 😆. 

I'm almost glad you did ;) You have to get burned at least once by this kind of error to really understand and remember it 😬 But be happy, you've now passed a huge milestone in your C++/shared pointers knowledge! :D

Core/GDCore/Project/EffectsContainer.h
```diff
@@ -0,0 +1,117 @@
...
+ public:
+  EffectsContainer();
+  EffectsContainer(const EffectsContainer& other);
+  virtual ~EffectsContainer(){};
+
+  EffectsContainer& operator=(const EffectsContainer& rhs);
+  void Init(const EffectsContainer& other);
```
4ian, line 35: I think this can be in the private section

Core/GDCore/Project/Layer.cpp
```diff
@@ -13,7 +14,8 @@ namespace gd {
 Camera Layer::badCamera;
 Effect Layer::badEffect;
```
4ian, line 15: You probably will move this to EffectsContainer, this static member will be useless in Layer.

Core/GDCore/Project/EffectsContainer.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+EffectsContainer& EffectsContainer::operator=(const EffectsContainer& rhs) {
+  if (this != &rhs) Init(rhs);
+  return *this;
+}
+
+void EffectsContainer::Init(const EffectsContainer& other) {
+  effects.clear();
+  for (auto& it : other.effects) {
+    effects.push_back(it);
```
4ian, line 28: Is this really triggering a copy of the effect? Or a copy of just the shared_ptr (in which case it will point to the same Effect object in memory)

HarsimranVirk: I'm pretty confused about that as well, I don't really see a copy constructor for `Effect`. Should I create an `Effect` object (and maybe write a copy constructor for Effect), then push it?

4ian: Check what the type `auto& ` is here (VSCode can maybe autocomplete if you put your cursor on it).
I think it's a `std::shared_ptr<gd::Effect>&`. In other words, a reference to a shared_ptr. So far so good!
Then if you `push_back` it, then the vector will make a copy of what you pass (a `std::vector` store things by copying them). So it will copy the `shared_ptr<gd::Effect>`. So you'll have `other.effects` and `effects` both containing pointers pointing to the same things.

Instead, you must make a copy, i.e: make a new shared_ptr to a new object :) 
How to do so? 
1. Use * to get the object pointed by the shared_ptr.
2. Construct a new object from it
3. Make a shared ptr from it.

Something like:
```cpp
Effect* newEffect = new Effect();
*newEffect = *it; // This is the operator= that is used here.
effects.push_back(std::shared_ptr<Effect>(newEffect));
```

Or to avoid making an empty object and then assigning it (which is useless work):

```cpp
Effect* newEffect = new Effect(*it); // This is the copy constructor that is used.
effects.push_back(std::shared_ptr<Effect>(newEffect));
```

In practice, for conciseness and safety, we use make_shared:
```cpp
effects.push_back(std::make_shared<gd::Effect>(*it)); // make_shared will call the constructor, and then return a shared_ptr to the newly created object.
```

4ian:
Seems like there is still a compilation error (https://travis-ci.org/github/4ian/GDevelop/jobs/771050260) but the effects container looks good now I think!
Would be good to verify everything works as before in terms of layer's effect and should be good :)

4ian:
Let me know when you think this is ready and that you've tested it works :)

HarsimranVirk:
> Let me know when you think this is ready and that you've tested it works :)

So far it seems to work fine, but travis-ci isn't able to link everything, (clang mentions undefined reference to EffectContainer), what should I do to fix that?

GDevelop.js/Bindings/Bindings.idl
```diff
@@ -654,6 +654,23 @@ interface Effect {
   void UnserializeFrom([Const, Ref] SerializerElement element);
 };
 
+interface EffectsContainer {
```
4ian, line 657: Great :) I suggest you add also some basic tests in `GDevelop.js/__tests__/Core.js`.
This will allow you to try this without running the whole IDE, and be confident that it works well :) Basically, create tests that make an effects container, add a few effects, check it can get them back, maybe serialize/unserialize.

4ian:
> So far it seems to work fine, but travis-ci isn't able to link everything, (clang mentions undefined reference to EffectContainer), what should I do to fix that?

I've pushed a commit that should fix this :) This is the old game engine complaining that you did not add a .cpp file for this class. We should remove it at some point.

4ian:
I think we're mostly good, see if you can add a test or two (look at `it('can have effects', function () {` for Layer in `Core.js` as an example).

HarsimranVirk:
I've added a getter `Layer::GetEffectsContainer()`, which should be useful if in the future we may decide to add something else to EffectsContainer.

4ian:
Yes! Actually you can name it `GetEffects`. We could even:
- Remove all the effects related methods of gd::Layer, apart from `GetEffects`.
- Adapt the Core.js test to use `getEffects()`
- Adapt the IDE to pass `layer.getEffects()` to the EffectsList component (which is kind of using a `gdEffectsContainer` already, it's just a type that is aliased to `gdLayer` :)).

This way we've fully convert gd::Layer to use gd::EffectsContainer using composition :)

HarsimranVirk:
Didn't find any obvious regression so far, we can test a bit more, then we should be good to go I guess 👍.

4ian:
There should be changes in the IDE no? 🤔 At least in props passed to EffectsList.

HarsimranVirk:
> There should be changes in the IDE no? 🤔 At least in props passed to EffectsList.

Yes, I forgot to push them. Also, what should I do about the types generated in GDevelop.js/types after compiling the c++ code? They keep coming up in the unstaged area (which is why I forgot to push the editor's js files as well), should I push them as well?

4ian:
> Also, what should I do about the types generated in GDevelop.js/types after compiling the c++ code?

You should commit them. 
Also edit `generate-types.js` to remove `declare type gdEffectsContainer = gdLayer;`

4ian:
Still some errors in the storybook ;)

HarsimranVirk:
`analyze-test-coverage` is still failing in semaphoreci, but its passing locally 🤔.

4ian:
Yep Semaphore is not rebuilding libGD.js, so it will fail - that's expected you can ignore this :)

PR #2646, opened 2021-05-13
Update Link Tools examples to use the new link action

D8H:
* The new link action simplify the grid building.
* The search condition interface used in the tactical example has changed, it will avoid to break the code if a user update the extension.
* The embedded extension also include a link to this documentation: http://wiki.compilgames.net/doku.php/gdevelop5/all-features/extensions/linked-objects-tools


4ian (approved):
👍

PR #2648, opened 2021-05-15
Fix grid magnet at the object creation when the magnet position is not an integer.

D8H:
Pixel isometric grids are normally even so the issue should not happen.
For the city builder example, I used an odd height to show some kind of light grid.
As half the height is #,5 some elements on the grid won't have an integer y position.

The issue happens only when objects are created and not moved. So either it should always be rounded or either never.
I think it should not be rounded because it messes with the grid logic and roundPositionsToGrid() already do it if there is no grid.

Project example:
InstanceAdderBadRounded
[AdderGridBug.zip](https://github.com/4ian/GDevelop/files/6488006/AdderGridBug.zip)


4ian:
Seems indeed more correct not too round anything here - if there is any rounding to do, it's the job of `roundPositionsToGrid`.

PR #2658, opened 2021-05-18
Add padding option for effects on layers 

Bouh:
# Description

Actually objects and specially objects with a small padding can't render the effect correctly.
@Silver-Streak make me realize that happend specially because he try to break the Bitmap Text object ^^

I tested the effects which had a risk of overflow and I added the padding option.

@HarsimranVirk in the next week this will be an important option on effects for objects, that should if possible done by the engine, like for a blur parameter use this value, multiply by two and ad this value to the padding, something like that what do you think?

image
There is also as you can see a issue in the filter area, the effect are applied on a container which wrap the objects on the layer.
See on screen, both objects are on the same layer and there is nothing else, i've added a CRT effect on the layer, the result is limited to the AABB on these objects.
The attemped result would be a CRT effect on the whole screen no ? As user I found it's a weird result that not fulfill my attempt when I use an effect on layers.

# Compare results

With a padding of 100:
image

Top: padding = 0 (default)
Bottom: padding = 10

image

Another example with advanced bloom on a sprite, top padding=0, bottom padding=100:
image

A last example with radial blur:
image

### DIFF
--- a/Extensions/Effects/JsExtension.js
+++ b/Extensions/Effects/JsExtension.js
@@ -130,7 +130,7 @@ module.exports = {
       .setValue('0')
       .setLabel(_('Padding'))
       .setType('number')
-      .setDescription(_('Padding for filter area'));
+      .setDescription(_('Padding for the visual effect area'));
 
     const asciiEffect = extension
       .addEffect('Ascii')
@@ -423,7 +423,7 @@ module.exports = {
       .setValue('0')
       .setLabel(_('Padding'))
       .setType('number')
-      .setDescription(_('Padding for filter area'));
+      .setDescription(_('Padding for the visual effect area'));
 
     const displacementEffect = extension
       .addEffect('Displacement')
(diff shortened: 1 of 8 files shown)


Extensions/Effects/JsExtension.js
```diff
@@ -125,6 +125,12 @@ module.exports = {
       .setValue('7')
       .setLabel(_('Quality (between 0 and 20)'))
       .setType('number');
+    advancedBloomProperties
+      .getOrCreate('padding')
+      .setValue('0')
+      .setLabel(_('Padding'))
+      .setType('number')
+      .setDescription(_('Padding for filter area'));
```
4ian, line 133:
```suggestion
      .setDescription(_('The padding around the effect area'));
```

here and everywhere ("filter" is a PIXI term. In GDevelop, we must always talk about "effect". They are "visual effects", not filters).

Bouh: I replace by 'Padding for visual effects area'

4ian: `Padding for the visual effect area` then :)

4ian (approved):
Thank you! 👍

PR #2663, opened 2021-05-18
Add autocomplete support to more fields

arthuro555:
This adds Autocomplete and the expression builder to parameter fields that are strings with additional autocomplete (scene name, string with selector, layer name and external event name) by making them use the GenericExpressionField.

It currently has 1 issue: for a reason I do not get, on the link event, the autocompletions object passed to setState is correct and no other setState is clearing it, but no autocompletions are showing up and inspecting with the react debugger shows that the state does not contain the autocompletions >_>


4ian (commented):
Awesome! Let me double check that but that's great 👌👌

4ian:
> It currently has 1 issue: for a reason I do not get, on the link event, the autocompletions object passed to setState is correct and no other setState is clearing it, but no autocompletions are showing up and inspecting with the react debugger shows that the state does not contain the autocompletions >_>

This is because parameterRenderingService was not passed as a prop... This boils down to the issue that ExternalEventsField, despite being named a field, is used in what is not really a field, at least not a **parameter field**: it's just used in the `LinkEvent` component, which shows the list of external events/scenes. BUT this is NOT an expression. It's just the plain name of the external events/scenes. You can see it because no quotes are required:

image

In other words, ExternalEventsField should NOT be a generic expression field :) I've reverted this. This was confusing (I also got confused while checking the issue of the autocompletions). It should never have been in the ParameterFields folder, as it's only used by the LinkEvent. I've fixed this by moving it to a LinkEvent folder and renaming it "ExternalEvents**AutoComplete**" (and reworking the props to only pass what is necessary).

SceneNameField, LayerField, StringWithSelectorField are correct! ;) They are proper parameter fields (you can see that they are used inside ParameterRenderingService.js), and are always used on expressions, so your changes are fine.

Overall a good clean up :)

PR #2665, opened 2021-05-19
Autocompletion on variable fields using undefined variables

D8H:
This adds undefined variables used in the event sheet in autocompletion for variable fields and variable list.

It uses EventsVariablesFinder to get the variables form event sheets. This class does the search every time so there is a performance issue with big event sheets.
Before working on solution to maintain a variable list, I would like to discuss the usefulness of this feature.

I started by looking at:
https://trello.com/c/ZiLv4rR0/207-add-button-feature-to-automatically-add-variables-used-in-events-in-the-list-of-scene-global-object-variables
I think a list of undefined variables should rather be integrated to this:
https://trello.com/c/IiLgNR16/462-add-a-diagnostic-report-to-warn-about-potential-issues-in-the-game-and-show-them-in-the-events-sheet
But it saw 2 user stories.

#### Context
* The user tries a way to implement a new feature.
* He uses new variables, but don't define them because he's not sure about the solution.
* He validates the solution
* He wants to define the new variables he used.
#### Actual
* He copies the variable name from the event.
* He opens the variable list dialog.
* He Adds a new variable.
* He pastes the variable name.
* He goes back to the events.
* He searches another new variable and continue to back and forth until all the variables are defined
#### Expected
* He opens the variable list dialog.
* He adds a new variable.
* He starts to input the variable name from memory.
* He finds the whole name with autocompletion.
* He adds a new variable and continue until all the variables are defined
VariableList

#### Context
* The user needs a temporary variable.
* He doesn't define it because he doesn't want to pollute the variable list.
#### Actual
* He copies/pastes the name the 2 or 3 times he uses it.
#### Expected
* He uses autocompletion to set the name.
VariableField

Test Project: https://www.dropbox.com/s/y3xrabgkrnjo9tv/UndefinedVariableTest.zip?dl=1


D8H:
@4ian What is your view on this?

Bouh:
You parse the project when a variable dialog is open up, what you think to push the variable when the user add an event related to a variable ?

D8H:
> You parse the project when a variable dialog is open up, what you think to push the variable when the user add an event related to a variable ?

Do you mean maintaining a list of event variables for each objects, layouts and the global ones?
I think it should be the responsibility of the model in C++. I'm willing to do it, but it seems a bit complicated as I don't know this part of the code. So, I wanted to be sure that I'm heading in the right direction.

Bouh:
I was thinking of having something less costly than `findAllLayoutVariables`.

You said that this can have an impact on performance with large event sheets.

I was thinking of something else like when the user writes a variable in an action and a condition React returns the info to the project and the variable the user just added is added to the project directly.

This way the variable list is always up to date. 
No need to go through the whole event sheet to find undeclared variables  and no need to add them manually via autocomplete.

The disadvantage will be that the list of variables will be filled with all the existing variables in the event sheet, but that's also what this list is for, to see the variables.

D8H:
> I was thinking of something else like when the user writes a variable in an action and a condition React returns the info to the project and the variable the user just added is added to the project directly.

Finding new variables is the easy part. But, to find variables that are removed, we must know all the occurrences. I don't know if there is something that keep track of the variable usage occurrences but this could be interesting to do refactor or navigation tools anyway.

> The disadvantage will be that the list of variables will be filled with all the existing variables in the event sheet, but that's also what this list is for, to see the variables.

It's probably better if it's not in the same list as the defined variables.

4ian:
> I don't know if there is something that keeps track of the variable usage occurrences but this could be interesting to do refactor or navigation tools anyway.

There is nothing like this, in general this kind of "stateful" information is usually prefered to be avoided. Because it *will* goes out of sync (because we forgot to update it somewhere, or because someone changed the project in the JSON file, etc...) and it *will* be reported as a bug.
So in general, we prefer to do things like `EventsVariablesFinder` => because it browse the project, you guaranteed that the list it found will always be correct. Better, we can extend it so that it also look in extensions (ideally, we would expose "extension variables", but if extensions are using scene variables, we should in theory be able to find this).

This being said, you're right that it's a performance risk. Browsing the project is not a trivial task. On a huge project, this can be easily a few milliseconds, that will make the whole UI to freeze if we use it at every render... Not cool 😬

So what do we do? I recommend we :
1. Keep EventsVariablesFinder. If in the future it's too costly to browse a project (>100ms), we can introduce a "cache" or something like this. But life is better and safer without caches :)
2. Rework every place where it's used so that it's passed to components as a function call to be made when you want to compute the list. Unless we're sure we need the list, in which case we still pass it as a function and use `useMemo` or something like this to compute the list **only once**.

  For example, in `newIDE/app/src/InstancesEditor/InstancePropertiesEditor/index.js`, a list is (correctly) passed as "allVariableNames". Unfortunately, this means that at **every render** of the instances properties (which can happen any time you move on the screen, select, unselect, add, remove, press a key) we will pay for the browsing of the whole project for variables. Costly! 

  image

  Instead, pass a prop that is a function called `onComputeAllVariableNames` (the "compute" implies it's costly, and the "on" implies it's a function). This function takes no argument, and return the result of `EventsVariablesFinder`. Then the `VariablesList`, when it needs to actually display the list, will call this (and it will memoize the result, so that it's called only once, not at every click or every render! ;)).
  

Does this make sense?
Basically, what I'm searching for is an approach where you "pay only what you use". 

Note that for things that are very expensive, like verifying that expressions are valid and displaying expressions that are wrong (or any error during code generation in events), we'll use another approach which will be to rework code generation to report errors in a cache (called `DiagnosticsReport` probably). This cache will be created by the IDE and passed to the exporter in C++, and the exporter will store in this cache all errors. Then when rendering events/actions/conditions, it will be fast to check in the cache if there is any error report.
This is because we're talking about something that takes >1 second, so we need to do it only when the user asks for (i.e: when the user launch a preview, or modify an expression). The rule of thumb is:

1. If a process takes more than 1 second (creating a preview, code generation, and expression validation) => Display a loader, cache the errors somewhere to display them to the user later.
2. If a process takes more than a few milliseconds (~10-100ms, like the variable finder) => No need for a loader, but compute it when necessary only, typically not in a render. `useMemo` can be useful. Using a function to compute only the data when needed too.
3. If a process takes almost no time (<0.3ms, it's almost trivial), just put it in the render. Simple & easy.

Hope this helps! :D

D8H:
Thanks a lot for the explanations.

I used a function like you suggested and I checked with a log that the search is called only at VariableList and VariableField instantiation.
If there is several VariableField in an action or condition the search will be called for each field. It's not great but it seems rare enough. I don't see an easy way to avoid this. Is it fine?

I added a preference not to include undefined variables in VariableField  autocompletion by default.

newIDE/app/src/MainFrame/Preferences/PreferencesDialog.js
```diff
@@ -242,6 +243,16 @@ const PreferencesDialog = ({ i18n, onClose }: Props) => {
...
             />
           </Line>
+          <Line>
+            <Toggle
+              onToggle={(e, check) =>
+                setUseUndefinedVariablesInAutocompletion(check)
+              }
+              toggled={values.useUndefinedVariablesInAutocompletion}
+              labelPosition="right"
+              label={<Trans>Use undefined variables in autocompletion</Trans>}
```
4ian, line 253: Nitpicking: "Show undefined variables in autocompletions"

newIDE/app/src/SceneEditor/index.js
```diff
@@ -1309,6 +1310,18 @@ export default class SceneEditor extends React.Component<Props, State> {
               }
             }}
             hotReloadPreviewButtonProps={this.props.hotReloadPreviewButtonProps}
+            onComputeAllVariableNames={() =>
+              this.state.variablesEditedObject
```
4ian, line 1314: This will not work I think.
`variablesEditedObject` is null here. Because you're editing an instance variables (`this.state.variablesEditedInstance` is not null) but not an object (so `this.state.variablesEditedObject` is null).

If you want the object of an instance, you have to find it, like done in "onEditObjectVariables" just above:

Get the object name associated to the instance: `const associatedObjectName = variablesEditedInstance.getObjectName()`
Use it to get the object: 
`getObjectByName(project, layout, associatedObjectName)`
Finally, if object is not null, you can feed it to `findAllObjectVariables` :)

This is because we don't keep "links" between instances and objects. The only link is the object name, so we have to find to which object it refers to (helped by `getObjectByName` - you can see btw in the codebase and see it's used in LayerRenderer - i.e: in the scene renderer :) Because when we iterate on instance, we must get on the first time the associated object to know which renderer we must create).

newIDE/app/src/VariablesList/index.js
```diff
@@ -133,9 +137,13 @@ export default class VariablesList extends React.Component<Props, State> {
...
     // We don't want to ever manipulate/access to variables that have been deleted (by removeRecursively):
     // that's why it's important to only delete ancestor variables.
-    ancestorOnlyVariables.forEach(({ variable }: VariableAndName) =>
-      variablesContainer.removeRecursively(variable)
-    );
+    ancestorOnlyVariables.forEach(({ name, variable }: VariableAndName) => {
+      variablesContainer.removeRecursively(variable);
+
+      if (this._allVariableNamesSet.has(name)) {
+        this._undefinedVariableNames.add(name);
```
4ian, line 144: Can you explain why we need to mutate this `_undefinedVariableNames` when we remove variables?

newIDE/app/src/VariablesList/index.js
```diff
@@ -367,6 +383,16 @@ export default class VariablesList extends React.Component<Props, State> {
     const { variablesContainer, inheritedVariablesContainer } = this.props;
     if (!variablesContainer) return null;
 
+    if (this._allVariableNamesSet === null) {
+      const allVariableNames = this.props.onComputeAllVariableNames();
+      this._allVariableNamesSet = new Set<string>(allVariableNames);
+      this._undefinedVariableNames = new Set<string>(
```
4ian, line 389: Not sure if this will work with structure having children: if we have an event using the scene variable called "MyVariable.MyChild" for example?

newIDE/app/src/VariablesList/index.js
4ian: While computing `_allVariableNamesSet` only once is very useful, I think we could compute `_undefinedVariableNames` in the render method, so it will avoid a bunch of add/delete that are tedious to follow and that we risk to forget after a refactoring.

Also see the other comment: when computing this list of undefined variables, we might have to handle children variables of structures (or arrays)... But let's see that on another comment.

4ian (commented):
Going in the right direction :)

I think this will hit an unfortunate limitation of `EventsVariablesFinder`. This is basically scanning the project, and collecting the values of all parameters that are a scene variable/global variable/object variable. So far so good, when we use only simple names for our variables.
BUT these variable name fields are actually expressions! :) For example, we can use for a scene variable, the `MyVariable.Child` syntax. For example `PlayerInfo.Score`.
Or even arrays: `PlayerInfo[0].Score` or even expressions: `PlayerInfo[1+1].Score`

All of these will be reported "as is" by EventsVariablesFinder... which is a bit useless/confusing for the user to see as an autocompletion "`PlayerInfo[1+1].Score`" 😅

So what should we show to the user instead? Probably just `PlayerInfo`. And if we go further, we could be smart and suggest `Score` when we are *inside* a `PlayerInfo` child.

See what I mean?

Of course, this might get more complex (we would need to return real variables from EventsVariablesFinder, to mimic the structures/arrays, or something like this). Basically some kind of Variable "shape" or type infos.

For now, I suggest we keep things simple:
- We only show autocompletions for ROOT variables (variables that are not children of a parent structure/array).
- We only show autocompletions that are "plain variable names", which basically means they have no dot dot/bracket (which I think is sufficient to guarantee it's a plain name).

Let me know if this make sense!

D8H:
I added a test project on the description: https://www.dropbox.com/s/y3xrabgkrnjo9tv/UndefinedVariableTest.zip?dl=1

newIDE/app/src/VariablesList/VariableRow.js
```diff
@@ -90,17 +92,22 @@ const VariableRow = ({
       {arrayElement ? (
         <Text noMargin>{name}</Text>
       ) : (
-        <TextField
+        <SemiControlledAutoComplete
           margin="none"
-          style={{
-            fontStyle: origin !== 'inherited' ? 'normal' : 'italic',
```
4ian, line 96: We probably want to keep this style, please put a TODO comment otherwise we're risking silently losing a "feature".

newIDE/app/src/VariablesList/index.js
```diff
@@ -133,9 +136,9 @@ export default class VariablesList extends React.Component<Props, State> {
 
     // We don't want to ever manipulate/access to variables that have been deleted (by removeRecursively):
     // that's why it's important to only delete ancestor variables.
-    ancestorOnlyVariables.forEach(({ variable }: VariableAndName) =>
-      variablesContainer.removeRecursively(variable)
-    );
+    ancestorOnlyVariables.forEach(({ name, variable }: VariableAndName) => {
```
4ian, line 139: You can cancel the changes?

newIDE/app/src/MainFrame/Preferences/PreferencesDialog.js
```diff
@@ -242,6 +243,16 @@ const PreferencesDialog = ({ i18n, onClose }: Props) => {
...
             />
           </Line>
+          <Line>
+            <Toggle
+              onToggle={(e, check) =>
+                setUseUndefinedVariablesInAutocompletion(check)
+              }
+              toggled={values.useUndefinedVariablesInAutocompletion}
+              labelPosition="right"
+              label={<Trans>Show undefined variables in autocompletion</Trans>}
```
4ian, line 253: I would go for: `Suggest names of variables used in events but not declared in the list of variables`

4ian (commented):
Almost there! Great work :)

newIDE/app/src/VariablesList/VariableRow.js
```diff
@@ -90,17 +92,22 @@ const VariableRow = ({
       {arrayElement ? (
         <Text noMargin>{name}</Text>
       ) : (
-        <TextField
+        <SemiControlledAutoComplete
```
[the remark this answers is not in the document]
D8H: I defined variables on an object and it was autocompleted in variable fields of the event sheet.
But names like "]]]]]]]]" used in the sheet will be in autocompletion.

4ian:
Very nice! I think this is good to go.
I think we should create an issue to follow up on two things:
- Support for structure children/array items.
- Support for analyzing events across event links (if you use an external events sheet, this won't work). This could be done by improving the analyzer by using `GetExternalEventsDependencies` from `DependenciesAnalyzer` (and `GetScenesDependencies`) to identify linked external events and run the variables collection logic on these too (much like done in WholeProjectRefactorer). 

We should not necessarily do this now, but keep an issue opened to have the context to improve this later, or help someone contribute to this :)

D8H:
Thank you for the review.
I'll look at this and write issues.

4ian:
Merged! Thanks for this great work. This will be *very* very useful! 🎉

newIDE/app/src/VariablesList/index.js
```diff
@@ -133,7 +140,7 @@ export default class VariablesList extends React.Component<Props, State> {
 
     // We don't want to ever manipulate/access to variables that have been deleted (by removeRecursively):
     // that's why it's important to only delete ancestor variables.
-    ancestorOnlyVariables.forEach(({ variable }: VariableAndName) =>
+    ancestorOnlyVariables.forEach(({ name, variable }: VariableAndName) =>
```
Bouh, line 143: `name` isn't used, right?

4ian: Indeed, seems this was not useful!

D8H: Should I revert it in https://github.com/4ian/GDevelop/pull/2722?

PR #2670, opened 2021-05-23
Fix text height

Bouh:
When a bitmap text have an empty line the position of the text is wrong.
image

B110: 
image

--- 

This PR: 
image

[Source forum](https://forum.gdevelop-app.com/t/newline-in-bitmap-text-object/32712/18)

### DIFF
--- a/Extensions/BitmapText/JsExtension.js
+++ b/Extensions/BitmapText/JsExtension.js
@@ -684,9 +684,9 @@ module.exports = {
       }
 
       this._pixiObject.position.x =
-        this._instance.getX() + this._pixiObject.width / 2;
+        this._instance.getX() + this._pixiObject.textWidth / 2;
       this._pixiObject.position.y =
-        this._instance.getY() + this._pixiObject.height / 2;
+        this._instance.getY() + this._pixiObject.textHeight / 2;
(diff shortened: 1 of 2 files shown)


4ian:
Nice! Could you do a quick search/explanation about what textWidth/textHeight represents (why width/height is not correct? 🧐)

Bouh:
I may be wrong I'm not a Pixi expert:

`width`/`height`comes from the container that isn't updated in `PIXI.BitmapText` when a new text is pass, maybe because `dirty`[ is set on false.](https://github.com/pixijs/pixijs/blob/dev/packages/text-bitmap/src/BitmapText.ts#L633)
 
 
Or because of `PIXI.Bitmap` doesn't compute again the [LocalBounds ](https://github.com/pixijs/pixijs/blob/dev/packages/display/src/Container.ts#L742)used to setup the width and [height on the container.](https://pixijs.download/dev/docs/PIXI.Container.html#height)

While `textWidth`/`textHeight` atleast `textHeight` is calculated from the data in .fnt file [and incremented ](https://github.com/pixijs/pixijs/blob/dev/packages/text-bitmap/src/BitmapText.ts#L549)at each lines.


`textHeight` works better than simple `height` which both should be equal IMO, but are not ¯\_(ツ)_/¯

PR #2671, opened 2021-05-24
Make extension dependencies only exported when the extension is used

arthuro555:
Uses the new `gd::UsedExtensionsFinder` to find what extension are used and only export those.

### DIFF
--- a/Core/GDCore/IDE/ExportedDependencyResolver.h
+++ b/Core/GDCore/IDE/ExportedDependencyResolver.h
@@ -21,7 +21,7 @@ struct DependencyMetadataAndExtension {
   gd::DependencyMetadata &GetDependency() const { return *dependency; };
   gd::PlatformExtension &GetExtension() const { return *extension; };
 
-private:
+ private:
   gd::DependencyMetadata *dependency;
   gd::PlatformExtension *extension;
 };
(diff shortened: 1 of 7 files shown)


4ian:
Very nice 👍 Can you check on a few games that it's working (notably with Admob that the cordova plugin is exported for Android and/or iOS according to which app id is/are set, also with the vibration plugin)?

Core/GDCore/IDE/ExportedDependencyResolver.h
```diff
@@ -14,39 +14,41 @@ namespace gd {
  * \note Both objects must be kept alive, as this is keeping a pointer to them.
  */
 struct DependencyMetadataAndExtension {
-  DependencyMetadataAndExtension(gd::DependencyMetadata& dependency_,
-                                 gd::PlatformExtension& extension_)
+  DependencyMetadataAndExtension(gd::DependencyMetadata &dependency_,
```
4ian, line 17: Is this the result of running clang-format? If so I can run clang-format in master to avoid this noise.

arthuro555: Ah yes I recently changed the IDE formatter from the builtin one from the VSCode C/C++ extension to the clang-format extension (as it was in the repository recomended extensions)

4ian:
Hey @arthuro555, could you confirm this is working "with Admob that the cordova plugin is exported for Android and/or iOS according to which app id is/are set, also with the vibration plugin"? :)

4ian:
Merged :) I updated a test as it was failing because the extension was not used (so now have two tests checking with and without an "AdMob" extension that it is exported only when needed).

arthuro555:
Sorry for not doing it myself 😓

PR #2676, opened 2021-05-25
[Fix] Bitmap Text - update AABB and take care of scaling

Bouh:
- After my last fix a user tried it and the scale option wasn't added in the comptutation. This PR fix it.

- The AABB was not updated when necessary, that was a problems for collisions and wasn't correct in the debug view.

It's nice to see the users use this new experimental object :D
Ready to merge!

### DIFF
--- a/Extensions/BitmapText/JsExtension.js
+++ b/Extensions/BitmapText/JsExtension.js
@@ -684,9 +684,9 @@ module.exports = {
       }
 
       this._pixiObject.position.x =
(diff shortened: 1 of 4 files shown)


4ian:
> and wasn't correct in the debug view

Very useful this debug view :)

> It's nice to see the users use this new experimental object :D

Yep!

4ian (approved):
Ok to go after the change I explained (we must keep getWidth/getHeight consistent across all objects/renderers).

4ian (approved):
👍👍

PR #2681, opened 2021-05-26
Hide for non desktop the preference for hide the menu bar

Bouh:
The preference " hide the menu bar" is visible on web-app but there is no menu bar in web-app.
This pr fix it.

Ready to merge

### DIFF
--- a/newIDE/app/src/MainFrame/Preferences/PreferencesDialog.js
+++ b/newIDE/app/src/MainFrame/Preferences/PreferencesDialog.js
@@ -284,6 +284,7 @@ const PreferencesDialog = ({ i18n, onClose }: Props) => {
               }
             />
           </Line>
+          {electron && (
           <Line>
             <Toggle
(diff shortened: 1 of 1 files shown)


newIDE/app/src/MainFrame/Preferences/PreferencesDialog.js
```diff
@@ -292,7 +293,6 @@ const PreferencesDialog = ({ i18n, onClose }: Props) => {
               label={<Trans>Hide the menu bar in the preview window</Trans>}
             />
           </Line>
-          {electron && (
```
4ian, line 295: You should keep this line or use `<> ... </>`

image

Bouh: _Always make a test before_
Sorry ^^

4ian:
👍

PR #2684, opened 2021-05-26
Fix unknown instruction when backdrop click happend

### LINKED ISSUE #2682: Invalid instruction added when exiting instruction selector
## Describe the bug
When you have the preference "apply the dialogue when clicking outside of it" and click outside of the instruction selector, it applies even though it shouldn't be possible and adds an invalid instruction.

## Video of the bug

https://user-images.githubusercontent.com/19349038/119683285-d0cb4200-be43-11eb-80bd-3926c3180f5e.mp4

## To Reproduce
1. Go to Preferences
2. Set backdrop click behavior to apply changes
3. Go to the event sheet
4. Press on add action or add condition on an event
5. Click the backdrop
6. See invalid instruction that has been added

Bouh:
Fix #2682

### DIFF
--- a/newIDE/app/src/EventsSheet/InstructionEditor/NewInstructionEditorDialog.js
+++ b/newIDE/app/src/EventsSheet/InstructionEditor/NewInstructionEditorDialog.js
@@ -243,7 +243,7 @@ export default function NewInstructionEditorDialog({
     <ResponsiveWindowMeasurer>
       {windowWidth => (
         <Dialog
-          onApply={onSubmit}
+          onApply={instructionType && onSubmit}
           actions={[
             <FlatButton
               label={<Trans>Cancel</Trans>}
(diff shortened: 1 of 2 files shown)


newIDE/app/src/EventsSheet/InstructionEditor/NewInstructionEditorDialog.js
```diff
@@ -243,7 +243,7 @@ export default function NewInstructionEditorDialog({
     <ResponsiveWindowMeasurer>
       {windowWidth => (
         <Dialog
-          onApply={onSubmit}
+          onApply={instructionType && onSubmit}
```
4ian, line 246: instructType ? onSubmit : null

And adapt the onApply type to be `onApply?: ?() => void`

4ian:
Thanks 👍

Issue #2674, opened 2021-05-24
Stacked ledges cannot all be grabbed


PascalLadalle:
In some conditions, the ledges of some stacked blocks cannot be grabbed.
Example project attached.
On the first column, all ledges can be grabbed.
On the second column, some ledges cannot be grabbed (third from bottom, and the highest one, in my testing). 

[LedgeBug.zip](https://github.com/4ian/GDevelop/files/6535105/LedgeBug.zip)

Tested on GD5b110 Win10x64

D8H:
Thanks for reporting this.

I did some tests
* Changing the order of the block will change which can be grabbed.
* Make half height size again show it often (doing ungrab, fall, grab a lot are skipped)

I think that the issue is that there is serval platform candidates to be grabbed and only one is checked.

4ian:
Thanks @D8H! I think your general approach is right. Added a comment to maybe make things a bit less costly in terms of computations, but otherwise looks pretty good.

PR #2686, opened 2021-05-26
 [Platformer] The platformer object now can always grab a platform if it is alongside several ones.

D8H:
It fixes: https://github.com/4ian/GDevelop/issues/2674

I moved _isCollidingWith after _separateFromPlatforms because it allows to have the 3 collision methods next to each other, but I didn't change anything in these 2 methods.
Don't mind the changes from @@ -554,23 +550,21 to @@ -685,6 +681,23

### DIFF
--- a/Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
+++ b/Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
@@ -427,33 +427,29 @@ namespace gdjs {
 
     _checkGrabPlatform() {
       const object = this.owner;
-      let tryGrabbingPlatform = false;
-      object.setX(
-        object.getX() +
-          (this._requestedDeltaX > 0
-            ? this._xGrabTolerance
-            : -this._xGrabTolerance)
-      );
-      let collidingPlatform = this._getCollidingPlatform();
-      if (collidingPlatform !== null && this._canGrab(collidingPlatform)) {
-        tryGrabbingPlatform = true;
-      }
-      object.setX(
-        object.getX() +
-          (this._requestedDeltaX > 0
-            ? -this._xGrabTolerance
-            : this._xGrabTolerance)
-      );
+      let oldX = object.getX();
+      let oldY = object.getY();
 
-      //Check if we can grab the collided platform
-      if (tryGrabbingPlatform) {
-        let oldY = object.getY();
+      let tryGrabbingPlatform = false;
+      for (let i = 0; i < this._potentialCollidingObjects.length; ++i) {
+        const platform = this._potentialCollidingObjects[i];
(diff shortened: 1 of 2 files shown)


Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -427,33 +427,29 @@ namespace gdjs {
...
-            ? -this._xGrabTolerance
-            : this._xGrabTolerance)
-      );
+      for (let i = 0; i < this._potentialCollidingObjects.length; ++i) {
+        const platform = this._potentialCollidingObjects[i];
 
-      //Check if we can grab the collided platform
-      if (tryGrabbingPlatform) {
-        let oldY = object.getY();
+        object.setX(
```
4ian, line 437: Do you think we could do this setX only once, get all the colliding platforms (i.e: we can do a `_getCollidingPlatforms` method), then we iterate on all of them to see if there is at least one that we can grab. Then we put back the object to its original X position (oldX, like you did), and we either bail out (if no platform to grab) or grab the first platform that can be grabbed (in the future we could add an option to force a decision, like the highest platform on the Y axis).

In other words, rather than:
```
for each potential colliding platform:
  move the object on X axis
  check the collision
  move back on X axis
  grab the platform if we can (i.e: set the Y position and check no collision with other platforms)
```

we would do:

```
move the object on X axis
for each potential colliding platform:
  check the collision (adding them to a "potential grabbing platform" array)
move back on X axis

for each potential grabbing platform
  grab the platform if we can (i.e: set the Y position and check no collision with other platforms)
```

The reason why I'm asking is because this could avoid "moving" the object too much, which can be costly as this invalidates the hitboxes and need to recompute them for collision tests.

D8H: Ok, the changes are pushed.

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -427,48 +427,49 @@ namespace gdjs {
...
+          this._setGrabbingPlatform(collidingPlatform);
           this._requestedDeltaY = 0;
-        } else {
-          object.setY(oldY);
+          collidingPlatforms.length = 0;
+          return;
         }
+        object.setY(oldY);
       }
+      collidingPlatforms.length = 0;
```
4ian, line 472: Nice, thanks for cleaning up the array 👍

4ian (approved):

4ian:
Great stuff! I took the opportunity to rename the collision methods. Will merge once CI is passing 👍

PR #2687, opened 2021-05-27
Add firebase add document action

arthuro555:
Add an action to add a document without having to specify a name for the document to simply add it to a collection with a unique name.


4ian (approved):

4ian:
👍

PR #2697, opened 2021-05-31
Replace examples by an "Example Library"

4ian:
This is a first, basic version of an "Example Library" where examples can be easily submitted and moderated by the community, like extensions. The repository of examples will be here: https://github.com/GDevelopApp/GDevelop-examples

image

The "library" is more flexible as examples can be added at any time, can have a longer description, thumbnail, and we could add other stuff later. We can also grow it to hundreds or thousands of examples without problems.
This will be useful in particular for extensions, so that we can host their examples.

**If you want to moderate the examples, let me know!** For now I've just included basic guidelines when posting a new example, but we can surely refine this. cc @Bouh @arthuro555 @Silver-Streak @tristanbob @D8H and everyone interested in the community :) 

To do on my side:
- Clean any remaining TODO in the code
- Adapt the website to also use the example library.


tristanbob:
I think this is a great idea to allow for more game examples without bloating the download.

Is there a reason you picked the word "Store"?  It immediately suggests that something is for sale.  Since these are all (currently) MIT licensed games, I suggest using a more community-friendly name, such as "Catalog", "Library", "Vault", "Collection", etc. 

Overall, this is going to be (another) great improvement to GDevelop.  Good work!

Silver-Streak:
> Is there a reason you picked the word "Store"? It immediately suggests that something is for sale. Since these are all (currently) MIT licensed games, I suggest using a more community-friendly name, such as "Catalog", "Library", "Vault", "Collection", etc.

I also think this is a fantastic addition, but I agree with Tristan above. I know the goal for the asset store is eventually that people may be able to selling things, but even then I think the name is confusing some users.

I'd definitely say something like the Examples Library (and renaming the Asset store to Asset Library) would cause less confusion. (And the Asset Store name could be changed again when/if we ever get to sold items on it)

Midhil457:
> If you want to moderate the examples, let me know!

I would definitely love to help in moderating the examples :smiley:

arthuro555:
I'd be happy to review submitted examples games, if that is what you mean by moderate.

4ian:
Fair enough about the naming, the "store" is a bit too forward thinking/confusing now. The good news is that this name is not even used in the interface (it's effectively as before, a list of examples, but now a bit more powerful) :)  
Overall I think Example Library is better to communicate about this! Same for asset.

> I would definitely love to help in moderating the examples 😃

Thanks! The idea is the same as now, check for general good quality of the example and any improvement or things to clean :)

> I'd be happy to review submitted examples games, if that is what you mean by moderate.

Yep, "moderate" is also here a big word for just saying helping from time to time to check the submitted games.

Bouh:
These filters have no results:
- Joystick Features
- Shopify
- Physics Engine (deprecated)
- Advanced window management
- Debugger tools
- Effects

Ordering filters by alphabetic name by default is possible?
Maybe like you get the .json on the server is possible for the server to ordering that only once, when the .json is generated.

Midhil457:
I have some questions,

- From what I see from the screenshot, some examples have icons. So, does that mean examples can also have icons like extensions do?

- As examples now can be filtered through. Do examples need to have a file with the tags?

- Same for the minimum GDevelop

4ian:
> From what I see from the screenshot, some examples have icons. So, does that mean examples can also have icons like extensions do?

Yes! You simply need to have a file called "thumbnail.png" in the game folder and it will used as icon.

> As examples now can be filtered through. Do examples need to have a file with the tags?

You *can* add a file called `TAGS.md`, containing comma separated tags. BUT even if you don't do it, extensions used in the game will be analysed and automatically used as tags.

> Same for the minimum GDevelop

For now don't care about this, it will be automatically filled I think in the future.

PR #2701, opened 2021-06-02
Specify type of variables in sentences

### LINKED ISSUE #927: 🇬🇧/🇺🇸 Report a typo/misspelling/inconsistent name (English)
This issue is to gather any inconsistent naming/typo/mistakes in GDevelop interface.

This is only **for the English language** - translations are managed on [Crowdin](https://crowdin.com/project/gdevelop) and should be fixed there ;)

So if you spot any issue, please report it here by indicating:
* where it can be found.
* what is the mistake/inconsistency/proper spelling and *how to fix it*.
* a screenshot if possible.

Note that GDevelop is open-source, so you can search in the source code for the text and submit a Pull Request to fix it 🎉

Thanks in advance for any report.

Bouh:
Fix https://github.com/4ian/GDevelop/issues/927#issuecomment-853351144

### DIFF
--- a/Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
@@ -631,8 +631,8 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
          _("Remove variable from an object array (by index)"),
          _("Removes a variable at the specified index of an object array "
            "variable."),
-         _("Remove variable at index _PARAM2_ from object array variable _PARAM1_ of "
-           "object _PARAM0_"),
+         _("Remove variable at index _PARAM2_ from array variable _PARAM1_ of "
+           "_PARAM0_"),
          _("Variables/Collections/Arrays"),
          "res/actions/var24.png",
          "res/actions/var.png")
--- a/Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
@@ -631,7 +631,7 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
          _("Remove variable from an object array (by index)"),
          _("Removes a variable at the specified index of an object array "
            "variable."),
-         _("Remove variable at index _PARAM2_ from array variable _PARAM1_ of "
+         _("Remove variable at index _PARAM2_ from object array variable _PARAM1_ of "
            "object _PARAM0_"),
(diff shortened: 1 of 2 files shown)


Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
```diff
@@ -632,7 +632,7 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
          _("Removes a variable at the specified index of an object array "
            "variable."),
          _("Remove variable at index _PARAM2_ from array variable _PARAM1_ of "
-           "object _PARAM0_"),
```
4ian, line 635: I've updated this because the object is already giving the context (and fixing the fact that we put "object _PARAM0_", while we don't use "object" in sentences, as it's implied by the object name itself.

4ian:
Note that in the future we may merge all these actions and have the parameters to be a global/scene or object variable, as which will be indicated by the parameter itself.

Bouh:
> Thanks!
> Note that in the future we may merge all these actions and have the parameters to be a global/scene or object variable, as which will be indicated by the parameter itself.

This is a bit the current case, we see the icons for global, scene, and object front of variables. But the icons aren't really big for the non-habit user.
image

Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
Bouh: Oh yeah, this context wasn't visible at first glance unless we know it or we check in the application, my bad.
image

4ian:
> This is a bit the current case, we see the icons for global, scene, and object front of variables. But the icons aren't really big for the non-habit user.

Totally, icons are good but not sufficient yet. We should probably, even after changing all actions/conditions to use a single parameter type "variable", continue to display (in the parameter) the type (scene variable PlayerLife, global variable Level...)
This is a big refactoring anyway, not for now.

Bouh:
I had a mockup, with a concept, I don't know what it was worth, it was just a vague idea.

variable_type_selector

variable_type_selector_object

4ian:
Haha, more or less what I was thinking exactly for scene/global variable!
And looks very nice for objects too.

Bouh:
I'm not sure about `[...]` in the path because we have to know the real path, but the basic idea was to add a selector for types.

4ian:
Not sure if we have a trello card for this by the way?

Bouh:
[This card ](https://trello.com/c/FwShg063/487-add-support-for-a-generic-variable-parameter-that-would-allow-to-choose-any-variable-object-variable-scene-global-extension-vari) fit well.

PR #2702, opened 2021-06-03
String autocompletion in expressions for scene names, layer names and choices

D8H:
This add autocompletion for strings in expressions.
It uses the delimiters found by the parser to replace the existing text to avoid to have a pseudo parser on UI side.


Core/GDCore/IDE/Events/ExpressionCompletionFinder.h
```diff
@@ -131,28 +161,53 @@ struct ExpressionCompletionDescription {
    */
   bool IsExact() const { return isExact; }
 
+  size_t GetReplacementStartPosition() const { return replacementStartPosition; }
+
+  size_t GetReplacementEndPosition() const { return replacementEndPosition; }
+
+  /**
+   * \brief Set if the expession is the last child of a function call.
```
4ian, line 169: (here and below) expression

Core/GDCore/IDE/Events/ExpressionCompletionFinder.h
```diff
@@ -209,31 +269,99 @@ class GD_CORE_API ExpressionCompletionFinder
...
+            searchedPosition + 1,
+            searchedPosition + 1));
   }
   void OnVisitNumberNode(NumberNode& node) override {
     // No completions
   }
   void OnVisitTextNode(TextNode& node) override {
-    // No completions
+    FunctionCallNode* functionCall = dynamic_cast<FunctionCallNode*>(parentNodeAtLocation);
+    if (functionCall != NULL) {
```
4ian, line 317: nit: prefer `nullptr`

Core/GDCore/IDE/Events/ExpressionCompletionFinder.h
```diff
@@ -209,31 +269,99 @@ class GD_CORE_API ExpressionCompletionFinder
...
+      {
+        if (functionCall->parameters.at(i).get() == &node) {
+          parameterIndex = i;
+          break;
+        }
+      }
+      if (parameterIndex < 0) {
+        return;
+      }
+      // Search the metadata parameter index skiping invisible ones
```
4ian, line 329: typo: skipping + period at the end of the sentence.

Core/GDCore/IDE/Events/ExpressionCompletionFinder.h
```diff
@@ -209,31 +269,99 @@ class GD_CORE_API ExpressionCompletionFinder
...
+      while (metadataParameterIndex < functionCall->expressionMetadata.parameters.size()) {
+        if (!functionCall->expressionMetadata.parameters[metadataParameterIndex].IsCodeOnly()) {
+          if (visibleParameterIndex == parameterIndex) {
+            break;
+          }
+          visibleParameterIndex++;
+        }
+        metadataParameterIndex++;
+      }
+      const gd::String &type = functionCall->expressionMetadata.parameters[metadataParameterIndex].GetType();
```
4ian, line 341: Will this crash if there is for example no parameter in the expression metadata but you still tried to write a parameter in the function?

4ian, line 341: But I like the idea of looking at the expression metadata 👍👍

newIDE/app/src/ExpressionAutocompletion/index.js
```diff
@@ -21,8 +21,11 @@ import { getVisibleParameterTypes } from '../EventsSheet/ParameterFields/Generic
 
 type BaseExpressionAutocompletion = {|
   completion: string,
+  replacementStartPosition: integer,
```
4ian, line 24: In the IDE with Flow, we don't do the distinction between integer/float => `number`

newIDE/app/src/ExpressionAutocompletion/index.js
```diff
@@ -58,6 +61,8 @@ type ExpressionAutocompletionContext = {|
 const getAutocompletionsForExpressions = (
   expressionMetadatas: Array<EnumeratedExpressionMetadata>,
   prefix: string,
+  replacementStartPosition: integer,
```
4ian, line 64: Same here: `number`

newIDE/app/src/ExpressionAutocompletion/index.js
```diff
@@ -207,17 +220,77 @@ const getAutocompletionsForObject = function(
...
+    //TODO handle stringWithSelector autocompletion
+  } else if (type === 'joyaxis') {
+    //TODO handle joyaxis autocompletion
+  }
+  //TODO add missing string types in Core\GDCore\Extensions\Metadata\ParameterMetadata.h
+
+  const filterStringList = (
+    list: Array<string>,
+    searchText: string
+  ): ObjectWithContextList => {
```
4ian, line 271: `Array<string>`

newIDE/app/src/ExpressionAutocompletion/index.js
```diff
@@ -207,17 +220,77 @@ const getAutocompletionsForObject = function(
...
+    if (!searchText) return list;
+
+    const lowercaseSearchText = searchText.toLowerCase();
+
+    return list.filter((variableName: string) => {
```
4ian, line 276: Why `variableName`?

newIDE/app/src/ExpressionAutocompletion/index.js
4ian: Also you might as well move this function outside of the function, to avoid re-declaring (so re-allocating) it at each call.

newIDE/app/src/ExpressionAutocompletion/index.js
```diff
@@ -345,47 +430,17 @@ export const insertAutocompletionInExpression = (
...
-
-  // Find the end position of the current word
-  let wordEndPosition = startPosition;
-  while (
-    wordEndPosition < expression.length &&
-    !isSeparatorChar(expression[wordEndPosition + 1])
-  ) {
-    wordEndPosition++;
-  }
+  let wordStartPosition: integer =
```
4ian, line 433: Use `const` + `number`.

newIDE/app/src/ExpressionAutocompletion/index.js
```diff
@@ -345,47 +430,17 @@ export const insertAutocompletionInExpression = (
     };
   }
 
-  // Start from the character just before the caret.
```
4ian, line 348: Much much better without this 👍 So happy to see this gone.

Core/GDCore/IDE/Events/ExpressionCompletionFinder.h
```diff
@@ -182,16 +238,20 @@ class GD_CORE_API ExpressionCompletionFinder
...
-        gd::ExpressionNodeLocationFinder::GetNodeAtPosition(node,
-                                                            searchedPosition);
-
+    
+    gd::ExpressionNodeLocationFinder finder(searchedPosition);
+    node.Visit(finder);
+    gd::ExpressionNode* nodeAtLocation = finder.GetNode();
+    gd::ExpressionNode* parentNodeAtLocation = finder.GetParentNode();
+
+    std::cout << "GetCompletionDescriptionsFor Test" << std::endl;
```
4ian, line 247: Let's remove this if not useful anymore

Core/GDCore/IDE/Events/ExpressionCompletionFinder.h
```diff
@@ -6,6 +6,8 @@
 #ifndef GDCORE_EXPRESSIONAUTOCOMPLETIONPROVIDER_H
 #define GDCORE_EXPRESSIONAUTOCOMPLETIONPROVIDER_H
 
+#include <fstream>
```
4ian, line 9: I think we can remove these headers if you remove the `std::cout <<`?

4ian (commented):
Very cool! Added some comments/nitpickings/small things to fix.

> It uses the delimiters found by the parser to replace the existing text to avoid to have a pseudo parser on UI side.

Very nice! This pseudo parser of separators on the UI side has always been a pain :) Very happy to see this gone.
Passing the positions of the replacement is a better idea.

Two things important to know:
- In the IDE, the typing is using just `number`. There is a also a few places to update, see https://semaphoreci.com/4ian/gd/branches/pull-request-2702/builds/1
- There are also C++ tests that you can run by compiling the native tests:
  - `mkdir build-tests`, `cd build-tests`, `cmake -DBUILD_GDJS=FALSE -DBUILD_TESTS=TRUE`, `make -j 4` and run the tests: `Core/GDCore_tests`.
  - This is not well documented, sorry! I can help fixing these tests, so concentrate on the Flow typing errors for now :)

I've also seen the issue with globalObjectsContainer casted to gdProject, I think we can avoid this as this assumption may/will not always hold, I'll post a comment about this too.

Overall great improvement, looking forward to this being complete :)

newIDE/app/src/ExpressionAutocompletion/index.js
```diff
@@ -207,17 +220,77 @@ const getAutocompletionsForObject = function(
...
+  } = expressionAutocompletionContext;
+
+  let autocompletionTexts: string[] = [];
+  if (type === 'layer') {
+    const layout = (objectsContainer: gdLayout);
+    for (let index = 0; index < layout.getLayersCount(); index++) {
+      autocompletionTexts.push(`"${layout.getLayerAt(index).getName()}"`);
+    }
+  } else if (type === 'sceneName') {
+    const project = (globalObjectsContainer: gdProject);
```
4ian, line 257: We should be able to pass the `project` in the `expressionAutocompletionContext` when calling `getAutocompletionsFromDescriptions` :)

newIDE/app/src/ExpressionAutocompletion/index.js
```diff
@@ -207,17 +220,77 @@ const getAutocompletionsForObject = function(
...
+  const type: string = completionDescription.getType();
+  //const objectName: string = completionDescription.getObjectName();
+  const {
+    globalObjectsContainer,
+    objectsContainer,
+  } = expressionAutocompletionContext;
+
+  let autocompletionTexts: string[] = [];
+  if (type === 'layer') {
+    const layout = (objectsContainer: gdLayout);
```
4ian, line 252: Same as the project: we should be able to pass the `layout` when calling `getAutocompletionsFromDescriptions ` and have it passed down thanks to the context. You can find it into the `scope` (of type `EventsScope`) in `GenericExpressionField` :) 

But we have to be even more cautious! It's possible that we're in a context where we don't have a layout (so the type will be a `?gdLayout`), for example when editing a function.

D8H:
There is still one flow issue. I don't know how to make gdExpressionCompletionDescription has the new enum element `Text`. I bound it but I must have missed something.

Core/GDCore/IDE/Events/ExpressionCompletionFinder.h
D8H: I think it only worked because no completion is asked by the GUI when there is a parsing error.

Core/GDCore/IDE/Events/ExpressionCompletionFinder.h
```diff
@@ -80,13 +80,15 @@ struct ExpressionCompletionDescription {
    */
   static ExpressionCompletionDescription ForText(
       const gd::String& type_,
+      const gd::ParameterMetadata *parameterMetadata_,
```
4ian, line 83: Uber nitpicking: using a pointer to store an "optional reference to something, that you're not owning" is fine, but prefer to keep using references in the interface (i.e: in ForText and in the SetParameterMetadata/GetParameterMetadata).
In other words, you can pass a const `gd::ParameterMetadata & parameterMetadata_` (and store its pointer `parameterMetadata = &parameterMetadata_;`).
With a pointer, there is a possible misunderstanding about if the class is taking ownership of the pointer or not.

(You can create a "HasParameterMetadata" if you need to know if this was set or not).

4ian (commented):
>  I don't know how to make gdExpressionCompletionDescription has the new enum element Text. I bound it but I must have missed something.

Yes this is manually typed actually! See `generate-types.js` in GDevelop.js/scripts.

D8H:
I get this when executing `cmake -DBUILD_GDJS=FALSE -DBUILD_TESTS=TRUE` from `./Core`
```
CMake Warning (dev) at CMakeLists.txt:12 (set):
  Cannot set "GDCORE_include_dir": current scope has no parent.
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at CMakeLists.txt:13 (set):
  Cannot set "GDCORE_lib_dir": current scope has no parent.
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Error at CMakeLists.txt:52 (gd_add_clang_utils):
  Unknown CMake command "gd_add_clang_utils".
  ```

**Comment by 4ian** [comment]
You should not need to clone anything, but I forgot in my instruction to tell CMake (the build generator system) that the sources root are in the parent directory. This should work better with the `..`:
`mkdir build-tests`, `cd build-tests`, `cmake .. -DBUILD_GDJS=FALSE -DBUILD_TESTS=TRUE`, `make -j 4` and run the tests: `Core/GDCore_tests`

The `mkdir build-tests && cd build-tests` is to avoid building directly in the sources directory. This is called an ["out of source build"](https://cgold.readthedocs.io/en/latest/tutorials/out-of-source.html#:~:text=%E2%80%9COut%2Dof%2Dsource%20build%E2%80%9D%20is%20a%20good%20practice,benefit%20and%20is%20not%20recommended.). It means that all the generated/compiled binaries files will be in this folder (useful if you want to restart your build from scratch, just delete the folder. It's a bit like deleted the `node_modules` folder).

Give it another try and let me now if the cmake works better now (with the `..`). In theory it should generate a bunch of files in `build-tests`, notably a few folders and a "Makefile". You can then run the makefile with `make -j 4`. 
All of this is automated by GDevelop.js when targeting Webassembly, but we should probably offer a few shortcuts or documentation for the native tests build like this.

**Comment by D8H** [comment]
The cmake shows no errors, the makefile is not created (I'm on windows, should I try with WSL?). 
```
build-tests> make -j 4
make: *** No targets specified and no makefile found.  Stop.
```

**Comment by D8H** [comment]
The test seems to be ok now.
New tests are needed to cover the new cases but I would rather be able to launch them locally before writing new ones.
I use this project to test it manually:
[UndefinedVariableTest.zip](https://github.com/4ian/GDevelop/files/6600994/UndefinedVariableTest.zip)

**Comment by 4ian** [comment]
> Edit: I tried with WSL Ubuntu 20.04. It produces the Makefile but Make crash at 7% because of this I guess:

Right, seems like the unused by outdated SFML dependency is causing us issue on Linux in general. We should remove it..
I don't have an immediate solution, so I suggest you can write a TODO with tests to add, I can maybe write a few and then we go ahead and merge this, as the rest looks very nice :)

File: Core/GDCore/IDE/Events/ExpressionCompletionFinder.h [hunk 16]
```diff
@@ -182,16 +260,18 @@ class GD_CORE_API ExpressionCompletionFinder
...
+    gd::ExpressionNode* nodeAtLocation = finder.GetNode();
+    gd::ExpressionNode* parentNodeAtLocation = finder.GetParentNode();
+    
     if (nodeAtLocation == nullptr) {
       std::vector<ExpressionCompletionDescription> emptyCompletions;
       return emptyCompletions;
     }
 
-    gd::ExpressionCompletionFinder autocompletionProvider(searchedPosition);
+    gd::ExpressionCompletionFinder autocompletionProvider(searchedPosition, *parentNodeAtLocation);
```
4ian, line 274: This is dereferencing a pointer that can (and is) null. I'm fixing this, as this can generate horrible crashs if used.

Core/GDCore/IDE/Events/ExpressionCompletionFinder.h
```diff
@@ -131,28 +162,74 @@ struct ExpressionCompletionDescription {
...
+   */
+  ExpressionCompletionDescription& SetParameterMetadata(const gd::ParameterMetadata &parameterMetadata_) {
+    parameterMetadata = &parameterMetadata_;
+    return *this;
+  }
+
+  /**
+   * \brief Return the parameter metadata if the completion is about a parameter.
+   */
+  const gd::ParameterMetadata& GetParameterMetadata() const { return *parameterMetadata; }
```
4ian, line 199: It's generally a good idea to be safe to use the ["null object pattern"](https://en.wikipedia.org/wiki/Null_object_pattern). 
We keep a pointer internally (so that we can modify what is pointed easily) but instead of setting it to `nullptr` when it's empty, we point it to an empty object (here, a ParameterMetadata constructed without anything and stored as a static member of the class).
This is almost equally performant (to check if the parameter metadata is defined, you just check if the pointer is not pointing to this empty object, called `badParameterMetadata`).
And this is a bit safer, in the sense that if someone tries to read the parameter metadata in a context where it's not valid, it's not crashing.

(An better alternative, safe and always correct would be to use an `Optional` like type, so that you would be forced to check if the parameter metadata is defined before using it. But we use this "null object pattern" a lot and it works well enough!)

4ian:
Merged! Thanks a lot for working on this, that's a great improvement over what we had 👍

Issue #2680, opened 2021-05-26
Show variables references from an external events sheet in autocompletion


D8H:
#### Context
* The preference "Suggest names of variables used in events but not declared in the list of variables" is enable.
* Some variables are referenced in an external events sheet without being defined in the variable list.
* This external events sheet is linked to the main event sheet.
* The user is writing a variable name in a variable field of the event sheet.

#### Actual
* Only defined variables and references from the main event sheet appear in the autocompletion (even in the external events sheet).

#### Expected
* Variables referenced in an external events sheet should appear in the autocompletion.

### Technical solution
From this comment: https://github.com/4ian/GDevelop/pull/2665#issuecomment-848225037
> This could be done by improving the analyzer by using `GetExternalEventsDependencies` from `DependenciesAnalyzer` (and `GetScenesDependencies`) to identify linked external events and run the variables collection logic on these too (much like done in WholeProjectRefactorer).

4ian:
Thanks for opening this 👍 Using `DependenciesAnalyzer ` is I think the way to go. It should be what is used to work on events generally speaking (though maybe we could integrate it into the `ArbitraryEventsWorker` at some point, to avoid doing 2 passes on events - but that's not a problem here).

Bouh:
Can you confirm this issue is solved with the linked PR just above?
If yes you can close it 💯

PR #2717, opened 2021-06-08
Show variables references from external events in the autocompletion

D8H:
It solves: https://github.com/4ian/GDevelop/issues/2680

Project used for testing:
https://www.dropbox.com/s/y3xrabgkrnjo9tv/UndefinedVariableTest.zip?dl=0

### DIFF
--- a/Core/GDCore/IDE/Events/EventsVariablesFinder.cpp
+++ b/Core/GDCore/IDE/Events/EventsVariablesFinder.cpp
@@ -101,7 +101,6 @@ std::set<gd::String> EventsVariablesFinder::FindAllGlobalVariables(
             platform,
             project,
             project.GetLayout(i),
-            project.GetLayout(i).GetEvents(),
             "globalvar");
     results.insert(results2.begin(), results2.end());
   }
@@ -116,7 +115,7 @@ std::set<gd::String> EventsVariablesFinder::FindAllLayoutVariables(
   std::set<gd::String> results;
 
   std::set<gd::String> results2 = FindArgumentsInEventsAndDependencies(
-      platform, project, layout, layout.GetEvents(), "scenevar");
+      platform, project, layout, "scenevar");
   results.insert(results2.begin(), results2.end());
 
   return results;
@@ -133,7 +132,6 @@ std::set<gd::String> EventsVariablesFinder::FindAllObjectVariables(
       platform,
       project,
       layout,
-      layout.GetEvents(),
       "objectvar",
       object.GetName());
   results.insert(results2.begin(), results2.end());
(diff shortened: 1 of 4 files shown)


Core/GDCore/IDE/Events/EventsVariablesFinder.cpp
```diff
@@ -203,6 +207,32 @@ std::set<gd::String> EventsVariablesFinder::FindArgumentsInInstructions(
...
+    const gd::String& objectName) {
+  std::set<gd::String> results;
+
+  std::set<gd::String> results2 = FindArgumentsInEvents(
+      platform, project, layout, events, parameterType, objectName);
+  results.insert(results2.begin(), results2.end());
+
+  DependenciesAnalyzer dependenciesAnalyzer = DependenciesAnalyzer(project, layout);
+  dependenciesAnalyzer.Analyze();
+  for (gd::String externalEventName : dependenciesAnalyzer.GetExternalEventsDependencies()) {
```
4ian, line 225: Use a a const ref (const gd::String& externalEventName) to avoid a copy of each string at each iteration

Core/GDCore/IDE/Events/EventsVariablesFinder.cpp
```diff
@@ -203,6 +207,32 @@ std::set<gd::String> EventsVariablesFinder::FindArgumentsInInstructions(
...
+    const gd::ExternalEvents& externalEvents = project.GetExternalEvents(externalEventName);
+
+    std::set<gd::String> results3 = FindArgumentsInEvents(
+        platform, project, layout, externalEvents.GetEvents(), parameterType, objectName);
+    results.insert(results3.begin(), results3.end());
+  }
```
4ian, line 231: While now heavily discouraged, it's still possible to have a link in events linking to a scene. So we should also iterate on the scenes returned by `GetScenesDependencies` (doing the same work as for external events)

Core/GDCore/IDE/Events/EventsVariablesFinder.cpp
```diff
@@ -203,6 +207,32 @@ std::set<gd::String> EventsVariablesFinder::FindArgumentsInInstructions(
   return results;
 }
 
+std::set<gd::String> EventsVariablesFinder::FindArgumentsInEventsAndDependencies(
+    const gd::Platform& platform,
+    const gd::Project& project,
+    const gd::Layout& layout,
+    const gd::EventsList& events,
```
4ian, line 214: I think in this method, we should remove `events` argument and automatically use `layout.GetEvents()`. Why? Because we use DependenciesAnalyzer, which will analyse the events of the whole scene (so it would be inconsistent if we don't pass the whole layout events as argument).

Core/GDCore/IDE/Events/EventsVariablesFinder.h
```diff
@@ -113,13 +113,36 @@ class EventsVariablesFinder {
...
       const gd::EventsList& events,
       const gd::String& parameterType,
       const gd::String& objectName = "");
+
+  /**
+   * Construct a list of the value of the arguments for parameters of type @
+   * parameterType. It doesn't search in events dependencies.
+   *
+   * \param project The project used
+   * \param project The layout used
```
4ian, line 129: nit: I know it was already not properly documented, but this should be:

```suggestion
   *
   * \param platform The platform of the project
   * \param project The project
   * \param layout The layout

```

4ian (commented):
Super great! A few minor comments but looks great otherwise 👍

4ian (approved):
Looks good! 👍

PR #2720, opened 2021-06-08
[WIP] Add object effects container in runtime.

HarsimranVirk:
@4ian, I've added the effects tab in `ObjectEditorDialog` and it seems to work fine. There were no errors thrown so far, but the effects weren't passed to `objectData` in `runtimeobject.ts`. Am I missing something here? (I also checked that serialization of effects was enabled by you 3 days ago, so that shouldn't be a problem as well).

Screenshot 2021-06-08 185541

So, there were no issues while adding the effect. (And I guess it did (de)serialize without any issues because these changes were saved and persisted re-opening of projects, etc)

Screenshot 2021-06-08 185445

But, the effects array is always empty.


4ian:
> And I guess it did (de)serialize without any issues because these changes were saved and persisted re-opening of projects, etc

Weird! Serialization should not change for the preview, so if you can save and load again a game with object effects, it should be in the preview.
Can you check the object in the `data.js` file? To check the content of the `effects` array?

HarsimranVirk:
> Can you check the object in the data.js file?

Where can I find this `data.js` file?

Also, in `platformer.json` file, I can clearly see the effects:
```
"objects": [
        {
          "name": "Player",
          "tags": "",
          "type": "Sprite",
          "updateIfNotVisible": true,
          "variables": [],
          "effects": [
            {
              "effectType": "Glitch",
              "name": "Effect",
              "doubleParameters": {
                "animationFrequency": 60,
                "blueX": 10,
                "blueY": -4,
                "direction": 0,
                "fillMode": 0,
                "greenX": 10,
                "greenY": -4,
                "minSize": 8,
                "offset": 100,
                "redX": 2,
                "redY": 2,
                "sampleSize": 512,
                "slices": 5
              },
              "stringParameters": {},
              "booleanParameters": {
                "average": false
              }
            },
            {
              "effectType": "Ascii",
              "name": "Effect2",
              "doubleParameters": {
                "size": 8
              },
              "stringParameters": {},
              "booleanParameters": {}
            }
          ],
          "behaviors": [],

...
```
I'm not sure what's going on here 🤔.

4ian:
Try to export a game and check the data.js file in the export :)

HarsimranVirk:
> check the data.js file

Yup, it's empty here.

Screenshot 2021-06-11 180112

What should be our next step to fix this?

4ian:
I found and fixed the issue in master: https://github.com/4ian/GDevelop/commit/3c88c150d594c1c60f993243a63ad14193353953

When you're saving a game, it stays the same in memory.
But when you're exporting, a copy is made. Sadly the gd::Object assignement constructor/copy operator were missing the copy of effects.
We had this in gd::Object because of the smart pointers used for the behaviors. This is a good example that this should be encapsulated in a "gd::BehaviorContentContainer", which would implement these custom ctor/operator one for all. By doing this, gd::Object could use a gd::BehaviorContentContainer and get the proper copy automatically.
Try to give it a thought to understand how this work (but we won't implement gd::BehaviorContentContainer, that's a refactoring outside of this scope, and the bug is fixed anyway now).

So if you rebase that should be good. :) Would be good to implement in the runtime, as I prefer to add this once he runtime supports effects.

4ian:
Small note: if you can also add a "Edit effects" context menu item and a command palette command for it, that would be great :)

HarsimranVirk:
> if you can also add a "Edit effects" context menu item and a command palette command for it

I've also added "Edit behaviors", I guess it makes everything a bit symmetric. If it's not a good idea I can remove it 😬.

4ian:
> I've also added "Edit behaviors", I guess it makes everything a bit symmetric. If it's not a good idea I can remove it 😬.

That's a good idea! :) 
GitHub is showing too much commits in this PR because master was merged and not rebased. You can close the PR and open another one to clear this.

PR #2722, opened 2021-06-09
Add animation and point as string types

D8H:
* Adds objectAnimationName and objectPointName as subtype for string.
* Update Sprite extension
* Add new fields for autocompletion
* Add expression autocompletion (only for object function: object + string in parameters won't autocompete, but I guess it's too much rare to bother)
* Fix expression autocompletion for object functions

Test Project: https://www.dropbox.com/s/y3xrabgkrnjo9tv/UndefinedVariableTest.zip?dl=1

### DIFF
--- a/Core/GDCore/IDE/Events/ExpressionCompletionFinder.h
+++ b/Core/GDCore/IDE/Events/ExpressionCompletionFinder.h
@@ -14,6 +14,8 @@
 #include "GDCore/Extensions/Metadata/ExpressionMetadata.h"
 #include "GDCore/Extensions/Metadata/InstructionMetadata.h"
 #include "GDCore/IDE/Events/ExpressionNodeLocationFinder.h"
+#include "GDCore/Events/Parsers/ExpressionParser2.h"
+
 namespace gd {
 class Expression;
 class ObjectsContainer;
@@ -339,13 +341,6 @@ class GD_CORE_API ExpressionCompletionFinder
     // No completions
   }
 
-  static size_t WrittenParametersFirstIndex(const gd::String &objectName,
-                                            const gd::String &behaviorName) {
-    // By convention, object is always the first parameter, and behavior the
-    // second one.
-    return !behaviorName.empty() ? 2 : (!objectName.empty() ? 1 : 0);
-  }
-
   void OnVisitTextNode(TextNode& node) override {
     // Completions are searched in the case the text node is a parameter of a
     // function call.
@@ -364,7 +359,7 @@ class GD_CORE_API ExpressionCompletionFinder
       }
       // Search the parameter metadata index skipping invisible ones.
       size_t visibleParameterIndex = 0;
-      size_t metadataParameterIndex = WrittenParametersFirstIndex(
+      size_t metadataParameterIndex = ExpressionParser2::WrittenParametersFirstIndex(
           functionCall->objectName, functionCall->behaviorName);
 
       const gd::ParameterMetadata* parameterMetadata = nullptr;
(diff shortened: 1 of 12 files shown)


Core/GDCore/IDE/Events/ExpressionCompletionFinder.h
```diff
@@ -338,6 +338,14 @@ class GD_CORE_API ExpressionCompletionFinder
   void OnVisitNumberNode(NumberNode& node) override {
     // No completions
   }
+
+  static size_t WrittenParametersFirstIndex(const gd::String &objectName,
```
4ian, line 342: I wonder if we should make the method that this was copied from public in ExpressionParser2?

newIDE/app/src/EventsSheet/ParameterFields/ObjectAnimationNameField.js
```diff
@@ -0,0 +1,82 @@
...
+  void
+> {
+  _field: ?GenericExpressionField;
+
+  focus() {
+    if (this._field) this._field.focus();
+  }
+
+  getAnimationNames(
+    props: ParameterFieldProps
```
4ian, line 23: You don't need to pass the props, you can just use `this.props`.

newIDE/app/src/EventsSheet/ParameterFields/ObjectAnimationNameField.js
```diff
@@ -0,0 +1,82 @@
...
+      return [];
+    }
+
+    const object = getObjectByName(project, scope.layout, objectName);
+    if (!object) {
+      return [];
+    }
+
+    const spriteObject = gd.asSpriteObject(object);
+    if (!spriteObject) {
```
4ian, line 52: Unfortunately `castObject` is a "dumb" cast (some kind of "reinterpret_cast" in C++) that will always give you back an object, even if it does not make sense.
Instead, check the type:

```
if (object.getType() == "Sprite") {
  const spriteObject = gd.asSpriteObject(obejct);
  // Do the rest
}

return [];
```

Exceptionnally, avoid an early return here because we might need to add more object types in the future, or convert this to a "capability" of the object to have animations. Not using an early return is a good sign that this can be extended in the future for other object types or more generally for any object having animations.

newIDE/app/src/EventsSheet/ParameterFields/ObjectPointNameField.js
```diff
@@ -0,0 +1,82 @@
...
+      return [];
+    }
+
+    const object = getObjectByName(project, scope.layout, objectName);
+    if (!object) {
+      return [];
+    }
+
+    const spriteObject = gd.asSpriteObject(object);
+    if (!spriteObject) {
```
4ian, line 50: Same here, check `getType` and avoid the early return to hint that this could be extended maybe in the future.

newIDE/app/src/ExpressionAutocompletion/ExpressionAutocompletion.spec.js
```diff
@@ -215,6 +232,38 @@ describe('ExpressionAutocompletion', () => {
     );
   });
 
+  it('can autocomplete object points', () => {
```
4ian, line 235: Thanks for the test!

newIDE/app/src/ExpressionAutocompletion/index.js
```diff
@@ -282,10 +286,55 @@ const getAutocompletionsForText = function(
...
+    if (!objectName) {
+      return [];
+    }
+
+    const object = getObjectByName(project, scope.layout, objectName);
+    if (!object) {
+      return [];
+    }
+
+    const spriteObject = gd.asSpriteObject(object);
```
4ian, line 300: Same in this file about checking the type

4ian: Also add a comment before each `getType() == "Sprite"`, something like "// For now, only sprites are supporting animations." (or points).

4ian (approved):
🤩 Great stuff! Added a few comments

PR #2723, opened 2021-06-09
Add disconnection actions to p2p extension

arthuro555:
Adds actions to disconnect from clients and the broker server.


4ian:
Nice 👍

PR #2724, opened 2021-06-09
Update to firebase sdk v9

arthuro555:
This updates the Firebase SDK and the code using it in the IDE to the tree-shakeable v9 release.  
This is especially interesting for us as we (currently) are only using this very small part of the sdk, webpack will now be able to compile the rest away, meaning that the app will be less bloated and load faster.

### DIFF
--- a/Extensions/P2P/B_p2ptools.ts
+++ b/Extensions/P2P/B_p2ptools.ts
@@ -232,6 +232,41 @@ namespace gdjs {
       };
 
       /**
+       * Disconnects from another p2p client.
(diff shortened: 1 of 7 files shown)


4ian (commented):
Very nice!! Let me double check on my side that it's working well before going ahead (better test this both in the web-app and in electron).

4ian:
Seems to work well! I've updated the package-lock.json which was out of date (sorry this included a merge from master, but I'll squash the PR anyway).

4ian:
Thanks!

PR #2731, opened 2021-06-11
Remove unused argument

Bouh:
Fix https://github.com/4ian/GDevelop/pull/2665#pullrequestreview-681346730

Don't show in changelog


4ian (approved):
Perfect!

PR #2736, opened 2021-06-12
⚙️[WIP] Adding object effects manager in runtime.

HarsimranVirk:
I'm not sure exactly what happened with the previous branch 😬 (maybe too much merging?). I've cherry-picked my commits from the previous branch to this branch, should be good now.

TODO:
- [x] Rework the "layer" passed to effect to only pass a few useful stuff like the size of the area (either the size of the layer or the object) and the elapsed time?
- [x] Handle hot-reloading:
  - [x] Adding a new effect should work
  - [x] Removing an effect should remove it
  - [x] Changing effect parameters should update it.
- [x] Test that actions to enable/disable an effect are working
- [ ] Add some tests to create a runtime object, add an effect, change its parameters, delete the effect.

### DIFF
--- a/Core/GDCore/Events/CodeGeneration/EffectsCodeGenerator.cpp
+++ b/Core/GDCore/Events/CodeGeneration/EffectsCodeGenerator.cpp
@@ -3,16 +3,18 @@
  * Copyright 2008-present Florian Rival (Florian.Rival@gmail.com). All rights
  * reserved. This project is released under the MIT License.
  */
-#include <iostream>
 #include "EffectsCodeGenerator.h"
+
+#include <iostream>
+
 #include "GDCore/Extensions/Metadata/EffectMetadata.h"
 #include "GDCore/Extensions/Metadata/MetadataProvider.h"
 #include "GDCore/Project/Effect.h"
-#include "GDCore/Project/Layer.h"
 #include "GDCore/Project/EffectsContainer.h"
+#include "GDCore/Project/Layer.h"
 #include "GDCore/Project/Layout.h"
-#include "GDCore/Project/Project.h"
 #include "GDCore/Project/Object.h"
+#include "GDCore/Project/Project.h"
 
 namespace gd {
 
@@ -34,9 +36,18 @@ void ExposeProjectEffects(
         worker(effect);
       }
     }
+
+    for (std::size_t i; i < layout.GetObjectsCount(); i++) {
+      auto& object = layout.GetObject(i);
+      auto& effects = object.GetEffects();
+      for (std::size_t e = 0; e < effects.GetEffectsCount(); e++) {
+        auto& effect = effects.GetEffect(e);
+        worker(effect);
+      }
+    }
   }
 
-  // Add object effects
+  // Add global object effects
   for (std::size_t s = 0; s < project.GetObjectsCount(); s++) {
     auto& effects = project.GetObject(s).GetEffects();
     for (std::size_t e = 0; e < effects.GetEffectsCount(); e++) {
--- a/Core/GDCore/Events/CodeGeneration/EffectsCodeGenerator.cpp
+++ b/Core/GDCore/Events/CodeGeneration/EffectsCodeGenerator.cpp
@@ -12,6 +12,7 @@
 #include "GDCore/Project/EffectsContainer.h"
 #include "GDCore/Project/Layout.h"
 #include "GDCore/Project/Project.h"
+#include "GDCore/Project/Object.h"
 
 namespace gd {
 
@@ -32,6 +33,15 @@ void ExposeProjectEffects(
         auto& effect = effects.GetEffect(e);
         worker(effect);
       }
+    }
+  }
+
+  // Add object effects
+  for (std::size_t s = 0; s < project.GetObjectsCount(); s++) {
+    auto& effects = project.GetObject(s).GetEffects();
+    for (std::size_t e = 0; e < effects.GetEffectsCount(); e++) {
+      auto& effect = effects.GetEffect(e);
+      worker(effect);
     }
   }
 }
--- a/Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
@@ -1374,21 +1374,10 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
                        "res/actions/layer.png")
       .AddParameter("object", _("Object"));
   
-  obj.AddAction("RemoveEffect",
-                _("Remove an effect"),
-                _("Remove an effect from the object"),
-                _("Remove _PARAM1_ effect from the _PARAM0_"),
-                _("Effects"),
-                "res/actions/effect24.png",
-                "res/actions/effect.png")
-      .AddParameter("object", _("Object"))
-      .AddParameter("string", _("Effect Name"))
-      .MarkAsSimple();
-  
   obj.AddAction("EnableEffect",
                 _("Enable an object effect"),
                 _("Enable an effect on the object"),
-                _("Enable _PARAM1_ effect on _PARAM0_ : _PARAM2_"),
+                _("Enable effect _PARAM1_ on _PARAM0_: _PARAM2_"),
                 _("Effects"),
                 "res/actions/effect24.png",
                 "res/actions/effect.png")
@@ -1398,38 +1387,41 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
       .MarkAsSimple();
   
   obj.AddAction("SetEffectDoubleParameter",
-                _("Set a double parameter for effect"),
-                _("Set a double parameter for an effect of the object"),
-                _("Set the value of _PARAM2_ as _PARAM3_ for the " 
-                  "effect _PARAM1_ of object _PARAM0_"),
+                _("Effect parameter (number)"),
+                _("Change the value of a parameter of an effect.") + "\n" +
+                  _("You can find the parameter names (and change the effect "
+                    "names) in the effects window."),
+                _("Set _PARAM2_ to _PARAM3_ for effect _PARAM3_ of _PARAM0_"),
                 _("Effects"),
                 "res/actions/effect24.png",
                 "res/actions/effect.png")
       .AddParameter("object", _("Object"))
       .AddParameter("string", _("Effect Name"))
-      .AddParameter("string", _("Parameter Name"))
-      .AddParameter("number", _("Parameter Value"))
+      .AddParameter("string", _("Parameter name"))
+      .AddParameter("expression", _("New value"))
       .MarkAsSimple();
   
   obj.AddAction("SetEffectStringParameter",
-                _("Set a string parameter for effect"),
-                _("Set a string parameter for an effect of the object"),
-                _("Set the value of _PARAM2_ as _PARAM3_ for the " 
-                  "effect _PARAM1_ of object _PARAM0_"),
+                _("Effect parameter (string)"),
+                _("Change the value (string) of a parameter of an effect.") + "\n" +
+                  _("You can find the parameter names (and change the effect "
+                    "names) in the effects window."),
+                _("Set _PARAM2_ to _PARAM3_ for effect _PARAM3_ of _PARAM0_"),
                 _("Effects"),
                 "res/actions/effect24.png",
                 "res/actions/effect.png")
       .AddParameter("object", _("Object"))
       .AddParameter("string", _("Effect Name"))
-      .AddParameter("string", _("Parameter Name"))
-      .AddParameter("string", _("Parameter Value"))
+      .AddParameter("string", _("Parameter name"))
+      .AddParameter("string", _("New value"))
       .MarkAsSimple();
   
   obj.AddAction("SetEffectBooleanParameter",
-                _("Set a boolean parameter for effect"),
-                _("Set a boolean parameter for an effect of the object"),
-                _("Set the value of _PARAM2_ as _PARAM3_ for the " 
-                  "effect _PARAM1_ of object _PARAM0_"),
+                _("Effect parameter (enable or disable)"),
+                _("Enable or disable a parameter of an effect.") + "\n" +
+                  _("You can find the parameter names (and change the effect "
+                    "names) in the effects window."),
+                _("Enable _PARAM2_ for effect _PARAM1_ of _PARAM0_: _PARAM3_"),
                 _("Effects"),
                 "res/actions/effect24.png",
                 "res/actions/effect.png")
@@ -1440,20 +1432,9 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
       .MarkAsSimple();
   
   obj.AddCondition("IsEffectEnabled",
-                  _("Effect Enabled"),
-                  _("Check if effect is enabled"),
-                  _("_PARAM1_ is enabled on _PARAM0_"),
-                  _("Effects"),
-                  "res/actions/effect24.png",
-                  "res/actions/effect.png")
-      .AddParameter("object", _("Object"))
-      .AddParameter("string", _("Effect Name"))
-      .MarkAsSimple();
-
-  obj.AddCondition("HasEffect",
-                  _("Has Enabled"),
-                  _("Check if the object has the specified effect"),
-                  _("_PARAM0_ has _PARAM1_ effect"),
+                  _("Effect is enabled"),
+                  _("The effect on an object is enabled"),
+                  _("Effect _PARAM2_ on _PARAM1_ is enabled"),
                   _("Effects"),
                   "res/actions/effect24.png",
                   "res/actions/effect.png")
--- a/Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
@@ -1373,6 +1373,94 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
                        _("Objects"),
                        "res/actions/layer.png")
       .AddParameter("object", _("Object"));
+  
+  obj.AddAction("RemoveEffect",
+                _("Remove an effect"),
+                _("Remove an effect from the object"),
+                _("Remove _PARAM1_ effect from the _PARAM0_"),
+                _("Effects"),
+                "res/actions/effect24.png",
+                "res/actions/effect.png")
+      .AddParameter("object", _("Object"))
+      .AddParameter("string", _("Effect Name"))
+      .MarkAsSimple();
+  
+  obj.AddAction("EnableEffect",
+                _("Enable an object effect"),
+                _("Enable an effect on the object"),
+                _("Enable _PARAM1_ effect on _PARAM0_ : _PARAM2_"),
+                _("Effects"),
+                "res/actions/effect24.png",
+                "res/actions/effect.png")
+      .AddParameter("object", _("Object"))
+      .AddParameter("string", _("Effect Name"))
+      .AddParameter("yesorno", _("Enable"))
+      .MarkAsSimple();
+  
+  obj.AddAction("SetEffectDoubleParameter",
+                _("Set a double parameter for effect"),
+                _("Set a double parameter for an effect of the object"),
+                _("Set the value of _PARAM2_ as _PARAM3_ for the " 
+                  "effect _PARAM1_ of object _PARAM0_"),
+                _("Effects"),
+                "res/actions/effect24.png",
+                "res/actions/effect.png")
+      .AddParameter("object", _("Object"))
+      .AddParameter("string", _("Effect Name"))
+      .AddParameter("string", _("Parameter Name"))
+      .AddParameter("number", _("Parameter Value"))
+      .MarkAsSimple();
+  
+  obj.AddAction("SetEffectStringParameter",
+                _("Set a string parameter for effect"),
+                _("Set a string parameter for an effect of the object"),
+                _("Set the value of _PARAM2_ as _PARAM3_ for the " 
+                  "effect _PARAM1_ of object _PARAM0_"),
+                _("Effects"),
+                "res/actions/effect24.png",
+                "res/actions/effect.png")
+      .AddParameter("object", _("Object"))
+      .AddParameter("string", _("Effect Name"))
+      .AddParameter("string", _("Parameter Name"))
+      .AddParameter("string", _("Parameter Value"))
+      .MarkAsSimple();
+  
+  obj.AddAction("SetEffectBooleanParameter",
+                _("Set a boolean parameter for effect"),
+                _("Set a boolean parameter for an effect of the object"),
+                _("Set the value of _PARAM2_ as _PARAM3_ for the " 
+                  "effect _PARAM1_ of object _PARAM0_"),
+                _("Effects"),
+                "res/actions/effect24.png",
+                "res/actions/effect.png")
+      .AddParameter("object", _("Object"))
+      .AddParameter("string", _("Effect Name"))
+      .AddParameter("string", _("Parameter Name"))
+      .AddParameter("yesorno", _("Parameter Value"))
+      .MarkAsSimple();
+  
+  obj.AddCondition("IsEffectEnabled",
+                  _("Effect Enabled"),
+                  _("Check if effect is enabled"),
+                  _("_PARAM1_ is enabled on _PARAM0_"),
+                  _("Effects"),
+                  "res/actions/effect24.png",
+                  "res/actions/effect.png")
+      .AddParameter("object", _("Object"))
+      .AddParameter("string", _("Effect Name"))
+      .MarkAsSimple();
+
+  obj.AddCondition("HasEffect",
+                  _("Has Enabled"),
+                  _("Check if the object has the specified effect"),
+                  _("_PARAM0_ has _PARAM1_ effect"),
+                  _("Effects"),
+                  "res/actions/effect24.png",
+                  "res/actions/effect.png")
+      .AddParameter("object", _("Object"))
+      .AddParameter("string", _("Effect Name"))
+      .MarkAsSimple();
+
 #endif
 }
 
--- a/GDJS/Runtime/layer.ts
+++ b/GDJS/Runtime/layer.ts
@@ -464,9 +464,7 @@ namespace gdjs {
      * in milliseconds, for objects on the layer.
      */
     getElapsedTime(runtimeScene: RuntimeScene): float {
-      return (
-        runtimeScene.getTimeManager().getElapsedTime() * this._timeScale
-      );
+      return runtimeScene.getTimeManager().getElapsedTime() * this._timeScale;
     }
 
     /**
--- a/GDJS/Runtime/layer.ts
+++ b/GDJS/Runtime/layer.ts
@@ -463,7 +463,8 @@ namespace gdjs {
      * Return the time elapsed since the last frame,
      * in milliseconds, for objects on the layer.
      */
-    getElapsedTime(runtimeScene: RuntimeScene): float {
+    getElapsedTime(runtimeScene?: RuntimeScene): float {
+      runtimeScene = runtimeScene || this._runtimeScene;
       return runtimeScene.getTimeManager().getElapsedTime() * this._timeScale;
     }
 
--- a/GDJS/Runtime/layer.ts
+++ b/GDJS/Runtime/layer.ts
@@ -9,7 +9,7 @@ namespace gdjs {
    *
    * Viewports and multiple cameras are not supported.
    */
-  export class Layer {
+  export class Layer implements EffectsTarget {
     _name: string;
     _cameraRotation: float = 0;
     _zoomFactor: float = 1;
@@ -463,9 +463,9 @@ namespace gdjs {
      * Return the time elapsed since the last frame,
      * in milliseconds, for objects on the layer.
      */
-    getElapsedTime(): float {
+    getElapsedTime(runtimeScene: RuntimeScene): float {
       return (
-        this._runtimeScene.getTimeManager().getElapsedTime() * this._timeScale
+        runtimeScene.getTimeManager().getElapsedTime() * this._timeScale
       );
     }
 
--- a/GDJS/Runtime/pixi-renderers/pixi-filters-tools.ts
+++ b/GDJS/Runtime/pixi-renderers/pixi-filters-tools.ts
@@ -89,9 +89,9 @@ namespace gdjs {
     /** A wrapper allowing to create a PIXI filter and update it using a common interface */
     export type FilterCreator = {
       /** Function to call to create the filter */
-      makePIXIFilter: (layer: gdjs.Layer | gdjs.RuntimeObject, effectData: EffectData) => any;
+      makePIXIFilter: (layer: gdjs.Layer, effectData: EffectData) => any;
       /** The function to be called to update the filter at every frame */
-      update: (filter: PIXI.Filter, layer: gdjs.Layer | gdjs.RuntimeObject) => any;
+      update: (filter: PIXI.Filter, layer: gdjs.Layer) => any;
       /** The function to be called to update a parameter (with a number) */
       updateDoubleParameter: (
         filter: PIXI.Filter,
@@ -117,7 +117,7 @@ namespace gdjs {
       /** The PIXI filter */
       pixiFilter: PIXI.Filter;
       /** The function to be called to update the filter at every frame */
-      update: (filter: PIXI.Filter, layer: gdjs.Layer | gdjs.RuntimeObject) => any;
+      update: (filter: PIXI.Filter, layer: gdjs.Layer) => any;
       /** The function to be called to update a parameter (with a number) */
       updateDoubleParameter: (
         filter: PIXI.Filter,
--- a/GDJS/Runtime/pixi-renderers/pixi-filters-tools.ts
+++ b/GDJS/Runtime/pixi-renderers/pixi-filters-tools.ts
@@ -15,7 +15,7 @@ namespace gdjs {
       return arr.indexOf(value) !== -1 ? value : min;
     };
 
-    const _filterCreators: {
+    export const _filterCreators: {
       [filterName: string]: FilterCreator;
     } = {};
 
@@ -89,7 +89,7 @@ namespace gdjs {
     /** A wrapper allowing to create a PIXI filter and update it using a common interface */
     export type FilterCreator = {
       /** Function to call to create the filter */
-      makePIXIFilter: (layer: gdjs.Layer, effectData: EffectData) => any;
+      makePIXIFilter: (layer: gdjs.Layer | gdjs.RuntimeObject, effectData: EffectData) => any;
       /** The function to be called to update the filter at every frame */
       update: (filter: PIXI.Filter, layer: gdjs.Layer) => any;
       /** The function to be called to update a parameter (with a number) */
--- a/GDJS/Runtime/pixi-renderers/pixi-filters-tools.ts
+++ b/GDJS/Runtime/pixi-renderers/pixi-filters-tools.ts
@@ -15,7 +15,7 @@ namespace gdjs {
       return arr.indexOf(value) !== -1 ? value : min;
     };
 
-    export const _filterCreators: {
+    const _filterCreators: {
       [filterName: string]: FilterCreator;
     } = {};
 
--- a/GDJS/Runtime/pixi-renderers/pixi-filters-tools.ts
+++ b/GDJS/Runtime/pixi-renderers/pixi-filters-tools.ts
@@ -89,9 +89,9 @@ namespace gdjs {
     /** A wrapper allowing to create a PIXI filter and update it using a common interface */
     export type FilterCreator = {
       /** Function to call to create the filter */
-      makePIXIFilter: (layer: gdjs.Layer, effectData: EffectData) => any;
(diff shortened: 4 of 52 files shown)


Core/GDCore/Events/CodeGeneration/EffectsCodeGenerator.cpp
```diff
@@ -34,6 +35,15 @@ void ExposeProjectEffects(
...
   }
+
+  // Add object effects
+  for (std::size_t s = 0; s < project.GetObjectsCount(); s++) {
+    auto& effects = project.GetObject(s).GetEffects();
+    for (std::size_t e = 0; e < effects.GetEffectsCount(); e++) {
+      auto& effect = effects.GetEffect(e);
+      worker(effect);
+    }
+  }
```
HarsimranVirk, line 46: @4ian this doesn't seem to include (and register) effects. Is something missing here? I'm not sure about what has to be done here.

GDJS/Runtime/pixi-renderers/pixi-filters-tools.ts
```diff
@@ -88,7 +88,7 @@ namespace gdjs {
     /** A wrapper allowing to create a PIXI filter and update it using a common interface */
     export type FilterCreator = {
       /** Function to call to create the filter */
-      makePIXIFilter: (layer: gdjs.Layer, effectData: EffectData) => any;
+      makePIXIFilter: (layer: gdjs.Layer | gdjs.RuntimeObject, effectData: EffectData) => any;
```
HarsimranVirk, line 91: I guess this is def wrong, I'll check each effect later, to be sure that there are no errors.

Core/GDCore/Events/CodeGeneration/EffectsCodeGenerator.cpp
4ian: This is only for global objects (i.e: project objects). You also must do this for every layout objects :)

GDJS/Runtime/pixi-renderers/pixi-filters-tools.ts
4ian: Yeah we probably actual need to either remove this first argument from all filters, or rename it according to what it's used for... no big deal for now.

4ian (commented):
Seems like a good start :) I think if you add the missing object effects in `ExposeProjectEffects`, you should see runtime objects with effects properly applied.

We can then work on:
- extending the PixiObjectEffectsManager interface to support adding/removing/modifying effects. 
- we should call `update` on effects too.
- expose actions/conditions to manipulate effects like for layers

👍

HarsimranVirk:
@4ian I'm sorry for the delay, I had a lot of work the past couple of weeks 😅.

> we should call update on effects too.

As I understand, all of the derived classes of the `RuntimeObject` have their own implementations of the `update()` method, is there a way to update the filters from the base class itself?

HarsimranVirk:
@4ian I've added the methods related to effects, (please check if the design makes sense), but I have no idea about how to expose them as actions or conditions (I know how to do it for extensions, but these methods are defined on `RuntimeObject` itself). Can you explain how to do this?

Bouh:
> @4ian I've added the methods related to effects, (please check if the design makes sense), but I have no idea about how to expose them as actions or conditions (I know how to do it for extensions, but these methods are defined on `RuntimeObject` itself). Can you explain how to do this?

1) A part of [the declaration](https://github.com/4ian/GDevelop/blob/master/Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp) is done in C++
2) [Binding is done in C++](https://github.com/4ian/GDevelop/blob/master/GDCpp/GDCpp/Extensions/Builtin/BaseObjectExtension.cpp) to handle the JS part
3) An implementation is done in JS (you have already done that part)

Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
```diff
@@ -1373,6 +1373,94 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
                        _("Objects"),
                        "res/actions/layer.png")
       .AddParameter("object", _("Object"));
+  
+  obj.AddAction("RemoveEffect",
+                _("Remove an effect"),
+                _("Remove an effect from the object"),
+                _("Remove _PARAM1_ effect from the _PARAM0_"),
```
4ian, line 1380: Remove effect _PARAM1_ from _PARAM0_

(never "the" before object names)

Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
```diff
@@ -1373,6 +1373,94 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
                        _("Objects"),
                        "res/actions/layer.png")
       .AddParameter("object", _("Object"));
+  
+  obj.AddAction("RemoveEffect",
+                _("Remove an effect"),
+                _("Remove an effect from the object"),
```
4ian, line 1379: Period at the end of all descriptions.

Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
```diff
@@ -1373,6 +1373,94 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
...
+                "res/actions/effect24.png",
+                "res/actions/effect.png")
+      .AddParameter("object", _("Object"))
+      .AddParameter("string", _("Effect Name"))
+      .MarkAsSimple();
+  
+  obj.AddAction("EnableEffect",
+                _("Enable an object effect"),
+                _("Enable an effect on the object"),
+                _("Enable _PARAM1_ effect on _PARAM0_ : _PARAM2_"),
```
4ian, line 1391: `Enable effect _PARAM1_ on _PARAM0_: _PARAM2_`

Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
```diff
@@ -1373,6 +1373,94 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
...
+                _("Effects"),
+                "res/actions/effect24.png",
+                "res/actions/effect.png")
+      .AddParameter("object", _("Object"))
+      .AddParameter("string", _("Effect Name"))
+      .AddParameter("yesorno", _("Enable"))
+      .MarkAsSimple();
+  
+  obj.AddAction("SetEffectDoubleParameter",
+                _("Set a double parameter for effect"),
```
4ian, line 1401: Look at how it was done for layers:
`Effect parameter (number)`

Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
```diff
@@ -1373,6 +1373,94 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
...
+                _("Set a double parameter for an effect of the object"),
```
4ian, line 1402: Same, copy the exact same descriptions and names as for layers:
```suggestion
                _("Change the value of a parameter of an effect.") + "\n" +
              _("You can find the parameter names (and change the effect "
                "names) in the effects window."),
```

Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
```diff
@@ -1373,6 +1373,94 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
...
+                _("Set the value of _PARAM2_ as _PARAM3_ for the " 
+                  "effect _PARAM1_ of object _PARAM0_"),
```
4ian, line 1404: `Set _PARAM2_ to _PARAM3_ for effect _PARAM3_ of _PARAM0_`

Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
```diff
@@ -1373,6 +1373,94 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
...
+                _("Effects"),
+                "res/actions/effect24.png",
+                "res/actions/effect.png")
+      .AddParameter("object", _("Object"))
+      .AddParameter("string", _("Effect Name"))
+      .AddParameter("string", _("Parameter Name"))
+      .AddParameter("number", _("Parameter Value"))
```
4ian, line 1411: Follow the exact same namings as for layers:
```suggestion
      .AddParameter("string", _("Effect"))
      .AddParameter("string", _("Parameter name"))
      .AddParameter("expression", _("New value"))
```

Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
```diff
@@ -1373,6 +1373,94 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
...
+      .MarkAsSimple();
+  
+  obj.AddAction("SetEffectStringParameter",
```
4ian, line 1414: Same remarks for the next twos for the title/description/sentence/parameters :)

Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
```diff
@@ -1373,6 +1373,94 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
...
+      .AddParameter("object", _("Object"))
+      .AddParameter("string", _("Effect Name"))
+      .AddParameter("string", _("Parameter Name"))
+      .AddParameter("yesorno", _("Parameter Value"))
+      .MarkAsSimple();
+  
+  obj.AddCondition("IsEffectEnabled",
+                  _("Effect Enabled"),
+                  _("Check if effect is enabled"),
+                  _("_PARAM1_ is enabled on _PARAM0_"),
```
4ian, line 1445:
```suggestion
                  _("Effect is enabled"),
                    _("The effect on an object is enabled"),
                    _("Effect _PARAM2_ on _PARAM1_ is enabled"),
```

Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
```diff
@@ -1373,6 +1373,94 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
...
+                  _("Effects"),
+                  "res/actions/effect24.png",
+                  "res/actions/effect.png")
+      .AddParameter("object", _("Object"))
+      .AddParameter("string", _("Effect Name"))
+      .MarkAsSimple();
+
+  obj.AddCondition("HasEffect",
```
4ian, line 1453: Let's remove this one, effects can't be added/removed at runtime, so we should just have a condition to check if an effect is enabled, not if it is present.

GDJS/Runtime/pixi-renderers/pixi-filters-tools.ts
```diff
@@ -14,7 +14,7 @@ namespace gdjs {
       return arr.indexOf(value) !== -1 ? value : min;
     };
 
-    const _filterCreators: {
+    export const _filterCreators: {
```
4ian, line 17: Seems weird to export something that has an underscore? Either we need to rename this, or use something else.

GDJS/Runtime/pixi-renderers/pixi-object-effects-manager.ts
```diff
@@ -0,0 +1,208 @@
...
+    protected _filters: Hashtable<Hashtable<PixiFiltersTools.Filter>>;
+
+    constructor() {
+      this._filters = new Hashtable();
+    }
+
+    update(runtimeObject: RuntimeObject, layer: Layer) {
+      const filters: Hashtable<
+        PixiFiltersTools.Filter
+      > | void = this._filters.get(runtimeObject.getName());
```
4ian, line 19: undefined, not void.

GDJS/Runtime/pixi-renderers/pixi-filters-tools.ts
HarsimranVirk: I think I checked it out by mistake (I was trying to figure out what are these objects exactly meant for). My bad, I'll fix it. 😬

GDJS/Runtime/pixi-renderers/pixi-object-effects-manager.ts
```diff
@@ -0,0 +1,208 @@
...
+        updateDoubleParameter: filterCreator.updateDoubleParameter,
+        updateStringParameter: filterCreator.updateStringParameter,
+        updateBooleanParameter: filterCreator.updateBooleanParameter,
+        update: filterCreator.update,
+      };
+
+      const renderer = runtimeObject.getRendererObject();
+      renderer.filters = (renderer.filters || []).concat(filter.pixiFilter);
+
+      if (!this._filters.containsKey(runtimeObject.getName())) {
```
4ian, line 64: This approach unfortunately won't really work, because you're storing filters for objects according to their name, which has potential issues if we later want to store more information differing for each instance. Also, it's going to be quite slow, at least suboptimal, to use a hashtable to get the filters everytime we need it for an object, having to read the object name and use it as a key.

Instead, each gdjs.RuntimeObject must have its `_filters`. Much like every layer has it :)

GDJS/Runtime/pixi-renderers/pixi-object-effects-manager.ts
```diff
@@ -0,0 +1,208 @@
...
+    /**
+     * Add a new effect on a runtime object, or replace the one
+     * with the same name.
+     * @param runtimeObject The runtime object
+     * @param effectData The data of the effect to add.
+     */
+    addEffect(
+      runtimeObject: RuntimeObject,
+      effectData: EffectData,
+      layer: Layer
```
4ian, line 37: Don't pass the layer to addEffect or anywhere. Effects on objects should have no access to the layer at all.

Bouh:
Let me know when there is something to test in the application.

4ian:
Just to give a bit more explanation about `pixi-object-effects-manager.ts` => in this class, you should only have things that deal with Pixi filters. 
The `gdjs.RuntimeObject` should have, in its method related to effects (`setEffectDoubleParameter`, `enableEffect`, `addEffect`, etc...):
- manipulation, as needed, of `this._filters` (the filters applied on the object, but this could work with anything, not only PixiJS).
- manipulation, using gdjs.ObjectEffectsManager, of PIXI filters (by using methods of `gdjs.ObjectEffectsManager` that could be implemented using another rendering engine in the future).

In other words, just use `gdjs.ObjectEffectsManager` to do the PIXI specific stuff :) And store in `this._filters` the rest.
Give it a try and let me know!

GDJS/Runtime/pixi-renderers/pixi-object-effects-manager.ts
HarsimranVirk: I was passing the layer because, the `update` function of the filters require `getElapsedTime()`, for example here:
https://github.com/4ian/GDevelop/blob/f3a49ad2cfd48efa6bed17565041094390a199b6/Extensions/Effects/glitch-pixi-filter.ts#L10-L18
The `getElapsedTime()` of `runtime Object` basically calls the function of  `layer`:
https://github.com/4ian/GDevelop/blob/f3a49ad2cfd48efa6bed17565041094390a199b6/GDJS/Runtime/runtimeobject.ts#L193-L197

4ian: In the future, maybe some objects will have a different time scale than other, so it's better to use the function provided by the object :)

Consider the "layer" object passed to the effects as a new type of object, which should contain basically "getElapsedTime, getWidth, getHeight". Not sure how to call it in TypeScript and as the parameter, but it's probably something we can call "EffectTarget"/"effectTarget".

GDJS/Runtime/pixi-renderers/pixi-object-effects-manager.ts
4ian: It's still fine I think, because we consider `_filters` as an "opaque" data structure which is handled by `gdjs.PixiObjectEffectsManager`. 
In a gdjs.RuntimeObject, there is already usually a `_renderer` object (depending on the object, but it's always and by convention called like this). We now add a `_filters`, to gdjs.RuntimeObject, which has a structure that is both known (it's a map to the current filters applied to the object) but that also contain the "renderer specific filters", aka here the "PIXI filters". 
If we were to implement this for another renderer, like Cocos, we would still have it, but we would have a `gdjs.CocosObjectEffectsManager` that would store cocos effects inside.

GDJS/Runtime/pixi-renderers/pixi-object-effects-manager.ts
```diff
@@ -0,0 +1,169 @@
...
+     * @param effectName The name of the effect.
+     */
+    removeEffect(runtimeObject: RuntimeObject, effectName: string) {
+      const filter = runtimeObject.getFilters()[effectName];
+      if (!filter) return;
+      const renderer: PIXI.DisplayObject = runtimeObject.getRendererObject();
+      renderer.filters = (renderer.filters || []).filter(
+        (pixiFilter) => pixiFilter !== filter.pixiFilter
+      );
+      delete filter[effectName];
```
4ian, line 71: You're sure about this line? `filter` is already a filter object, it does not have any `effectName` property.
You probably want to run delete on `runtimeObject.getFilters()`

HarsimranVirk: Oh yes, right. I think at this point I should be adding some tests to avoid these silly mistakes. 😬

GDJS/Runtime/pixi-renderers/pixi-object-effects-manager.ts
HarsimranVirk: Right, I think for the moment it can be a union type of gdjs.Layer and gdjs.RuntimeObject, but again, the `getElapsedTime()` 
 method is not equivalent at the moment for layers and runtimeobject (there's a difference of arguments). We might need some refactoring here.

4ian: Ah right! Mmmm let's make this a TODO. We can think about changing what we pass to effects (rather than passing an entire layer or runtimeobject, we can pass a few useful functions, we'll see!)

GDJS/Runtime/pixi-renderers/pixi-object-effects-manager.ts
4ian: Might be a good idea to think about unit/integration tests (probably integration tests if this requires PixiJS) yes, to check the good working of methods that add/change/delete effects :)

GDJS/Runtime/runtimeobject.ts
```diff
@@ -70,6 +70,10 @@ namespace gdjs {
     //Variables:
     protected _variables: gdjs.VariablesContainer;
 
+    //Effects:
+    protected _effects: EffectData[];
+    protected _filters: Record<string, PixiFiltersTools.Filter> = {};
```
4ian, line 75: Note: we have here "PixiFiltersTools.Filter", which is PixiJS specific, but there is nothing inside the runtime object actually PixiJS specific, as everything is going through the object effects manager. So this is good! And we can always change the typing later without breaking anything.

4ian: We might just want to rename this to `_rendererEffects`. Because "filters" is a PixiJS specific term, while here it's a table that contains the effects of the renderer (i.e: the effects of PIXI, i.e: the PIXI filters).
Other renderers in the future might have a different name ("shaders", etc.., so keeping the GDevelop term ("effect") is in theory better)

4ian:
Starting to look good! 
I'll be editing your first message with the list of things to do, like the question of the layer, hot-reloading (feel free to add/check stuff).

In terms of testing this for real, how far are we? Are you able to add an effect to an object and to see it on screen?

HarsimranVirk:
> In terms of testing this for real, how far are we? Are you able to add an effect to an object and to see it on screen?

Yes, we can enable/disable the effects on runtime using actions and conditions as well. I'm using the effects on `Player` object of `Platformer` game to test this. I'm able to change effects (enable/disable) based on the state of `Player` object (If it's jumping then apply a specific set of effects, if not apply a different set of effects, etc).

GDJS/Runtime/runtimeobject.ts
```diff
@@ -730,6 +745,145 @@ namespace gdjs {
...
+     */
+    getRendererEffects() {
+      return this._rendererEffects;
+    }
+
+    /**
+     * Add a new effect, or replace the one with the same name.
+     * @param effectData The data of the effect to add.
+     */
+    addEffect(effectData) {
```
4ian, line 761: Missing typing: EffectData

GDJS/Runtime/runtimeobject.ts
```diff
@@ -730,6 +745,145 @@ namespace gdjs {
...
+          effectData.booleanParameters[name]
+        );
+      }
+    }
+
+    /**
+     * Remove the effect with the specified name
+     * @param effectName The name of the effect.
+     */
+    removeEffect(effectName) {
```
4ian, line 793: string

GDJS/Runtime/runtimeobject.ts
```diff
@@ -98,6 +102,7 @@ namespace gdjs {
       );
       this._averageForce = new gdjs.Force(0, 0, 0);
       this._behaviorsTable = new Hashtable();
+      this._effects = objectData.effects;
```
4ian, line 105: You probably don't need and don't want to store effects from objectData.
These are the effects as they should be initially.

If you store this on all the objects, this means:
- all objects have a pointer to these, which consume memory.
- if I update the object data, thanks to the hot reload, all the objects are now having a pointer to the old effects data, which is a memory leak and outdated information.

This should actually not be needed. You only need this when you initialize the object, so just iterate on `objectData.effects`. The runtime effects are then stored in `_rendererEffects` (which contains the actual Pixi filters).

4ian:
> > In terms of testing this for real, how far are we? Are you able to add an effect to an object and to see it on screen?
> 
> Yes, we can enable/disable the effects on runtime using actions and conditions as well. I'm using the effects on `Player` object of `Platformer` game to test this. I'm able to change effects (enable/disable) based on the state of `Player` object (If it's jumping then apply a specific set of effects, if not apply a different set of effects, etc).

Great! Going in the right direction :) 
Let's go for tests, then hot reloading (with a new `_hotReloadRuntimeObjectsEffects(oldEffectData, newEffectData, runtimeObjects` method) and we'll see how we rework layer/objects passed to effects at the end.

Bouh:
I'll test it tonight :D

HarsimranVirk:
1. Added hot-reloading of effects.
2. Removed all the `runtimeObject` args from the methods of `ObjectEffectsManager` class, we don't need to pass the entire runtime object.

Remaining:

- [ ] Remove dependency of `Layer`.
- [ ] Add tests.

Bouh:
Good point there is no error on the compilation and at the opening 👍 

- The effects tab is present, it displays a hint, I think we can do like for the behaviors, an empty state instead of a hint.
image

- I recommend using the dropdown items list from the behaviors list into the effects list, there is a lot of parameters on effects as behaviors.
image

- With a Kawase (blur) effect on the object flicker. (See in video bellow) I suppose an issue with subpixels, it happened on other filters like Beveled edges.
- With a Kawase (blur) effect, when you move the object across the scene and off-screen,
the object becomes blurry when it approaches the window's borders. (See in video below)

https://user-images.githubusercontent.com/1670670/125863915-0b2b9000-3dd0-46a5-84e0-2ef6f543fcb7.mp4

- Displacement map, and Twist has as reference the world, not the object, not sure?
- Glitch effect interacts with the game border, I guess the referential is not correct for the effect, can you test it?


It looks promising! Some effects require to be adapted for objects.
The problem of precisions or subpixels I don't know what it is is very annoying (See video below)


https://user-images.githubusercontent.com/1670670/125866011-76efc205-ba7c-4143-917f-daa81ed915cf.mp4


https://user-images.githubusercontent.com/1670670/125866014-d618261d-c2e8-4e3c-8dae-72ef196a61e8.mp4

4ian:
Thanks for checking @Bouh 👍

For the first two remarks, I think it's a question of the design of the effects list that can be handled in another PR (and could even be done before that PR).

Good spot for the effects having issues. I think we need to be sure that all effects are adapted so that they are using the objects size/position for their effect (for example, a "twist" that is at position 0,5;0,5 must be centered on the object, not on the screen). Kawase is maybe not rendering properly because using the layer coordinates instead of the object?

> Displacement map, and Twist has as reference the world, not the object, not sure?

Yes, all effects must be checked to adapt them to use either the layer or the object (they should not even know, they should just receive the positions of the "target" - the target being the object or the layer).

HarsimranVirk:
@4ian how can we handle the case of `getElapsedTime()` of rutime object, as it's not equivalent to the layer's method?

> Yes, all effects must be checked to adapt them to use either the layer or the object (they should not even know, they should just receive the positions of the "target"

For this, I was thinking about using an interface `EffectsTarget`, which will be implemented by both the classes (RuntimeObject and Layer), and the "target" will simply be an implementation of `EffectsTarget`. But again, the problem here is with the `getElapsedTime()` method as the signature isn't the same. What refactoring should we do here? (Add a runtimeScene arg in layer's impl, or remove it from runtime object's impl, I'm not sure what should be the correct way.)

GDJS/Runtime/runtimeobject.ts
```diff
@@ -123,6 +130,16 @@ namespace gdjs {
...
+
+      const objectEffectsManager = this._runtimeScene
+        .getGame()
+        .getObjectEffectsManager();
+      for (const filterName in this._rendererEffects) {
+        objectEffectsManager.applyEffect(
+          this.getRendererObject(),
+          this._rendererEffects[filterName]
+        );
+      }
```
4ian, line 142: Can this be moved to the constructor? Or at least before the `onCreated` called on behaviors.
But 99% sure this should be in the constructor?

4ian: Ah right this can't be in the RuntimeObject constructor because this is called before the renderer is created. In this case, still move this before the loop on the behaviors onCreated.
It's possible a behavior would interact with the effects, but they are not applied yet, which could in theory be a problem.

GDJS/Runtime/runtimeobject.ts
```diff
@@ -219,6 +246,25 @@ namespace gdjs {
       initialInstanceData: InstanceData
     ): void {}
 
+    /**
+     * Update the effects during hot-reload
+     * @param oldEffectData Old effect data
+     * @param newEffectData New effect data
+     * @returns
+     */
+    _hotReloadRuntimeObjectEffects(
```
4ian, line 255: This method is using only public methods (removeEffect, addEffect).
This means it could be done outside of the object, inside the hot reloader. So let's remove it from here and do it inside the hot reloader (because it will avoid cluttering gdjs.RuntimeObject, and avoid cluttering updateFromObjectData which would become ill named)

GDJS/Runtime/runtimeobject.ts
```diff
@@ -232,8 +278,10 @@ namespace gdjs {
       oldObjectData: ObjectData,
       newObjectData: ObjectData
     ): boolean {
-      // If not redefined, mark by default the hot-reload as failed.
-      return false;
+      return this._hotReloadRuntimeObjectEffects(
```
4ian, line 281: It should not be the responsibility of `updateFromObjectData`, I think, to update the effects. The hot reload can do it, as it has all it needs (old/new effect data, and access to remove/addEffect)

GDJS/Runtime/runtimeobject.ts
```diff
@@ -730,6 +778,168 @@ namespace gdjs {
...
+    /**
+     * Returns the collection of effects to be rendered
+     * by the underlying renderer.
+     * @returns The render effects.
+     */
+    getRendererEffects() {
+      return this._rendererEffects;
+    }
+
+    initializeEffect(effectsData: EffectData) {
```
4ian, line 790: Why do we need both a `initializeEffect` and a `addEffect`? Can we have only one, `addEffect`?

GDJS/Runtime/runtimeobject.ts
```diff
@@ -730,6 +778,168 @@ namespace gdjs {
...
+      const layer = this._runtimeScene.getLayer(this.layer);
+      this._runtimeScene
+        .getGame()
+        .getObjectEffectsManager()
+        .initializeEffect(effectsData, this._rendererEffects, layer);
```
4ian, line 795: We should not need to pass the layer?

GDJS/Runtime/runtimeobject.ts
```diff
@@ -730,6 +778,168 @@ namespace gdjs {
...
+    addEffect(effectData: EffectData): boolean {
+      const layer = this._runtimeScene.getLayer(this.layer);
+      return this._runtimeScene
+        .getGame()
+        .getObjectEffectsManager()
+        .addEffect(
+          effectData,
+          this._rendererEffects,
+          this.getRendererObject(),
+          layer
```
4ian, line 811: Same here

GDJS/Runtime/pixi-renderers/pixi-object-effects-manager.ts
```diff
@@ -0,0 +1,271 @@
...
+     */
+    addEffect(
+      effectData: EffectData,
+      rendererEffects: RendererEffects,
+      rendererObject: PIXI.DisplayObject,
+      layer: Layer
+    ): boolean {
+      let effectAdded = true;
+      effectAdded =
+        this.initializeEffect(effectData, rendererEffects, layer) &&
```
4ian, line 99: Why passing layer here and not rendererObject?

GDJS/Runtime/pixi-renderers/pixi-object-effects-manager.ts
```diff
@@ -0,0 +1,271 @@
...
+     * @param effectData The effect data
+     * @param rendererEffects The renderer effects
+     * @param rendererObject The renderer object
+     * @param layer The Layer
+     */
+    addEffect(
+      effectData: EffectData,
+      rendererEffects: RendererEffects,
+      rendererObject: PIXI.DisplayObject,
+      layer: Layer
```
4ian, line 95: We should not pass the layer, the effects manager should just deal with `rendererObject`. Remove any mention to layer in this file.

GDJS/Runtime/pixi-renderers/pixi-object-effects-manager.ts
```diff
@@ -0,0 +1,271 @@
...
+
+    /**
+     * Update the filters applied on an object.
+     * @param runtimeObject
+     * @param layer
+     */
+    update(rendererEffects: RendererEffects, layer: Layer) {
+      for (const filterName in rendererEffects) {
+        const filter = rendererEffects[filterName];
+        filter.update(filter.pixiFilter, layer);
```
4ian, line 79: No `layer` anywhere, just `rendererObject`

GDJS/Runtime/pixi-renderers/pixi-object-effects-manager.ts
```diff
@@ -0,0 +1,271 @@
...
+    getRuntimeScene: () => RuntimeScene;
+    getElapsedTime: (runtimeScene: RuntimeScene) => number;
+    getHeight: () => number;
+    getWidth: () => number; 
+  };
+
+  class PixiObjectEffectsManager {
+    /**
+     * Initialize the renderer effects from the effect data.
+     * @param effectData The effect data
```
4ian, line 21: Describe what is needed to do after the initialization (i.e: call applyEffect to actually add the effect to the renderer object).

GDJS/Runtime/runtimeobject.ts
4ian: Actually I'm confused why you would even need this knowing that this is implemented already in the hot reloader. Remove these changes entirely.

4ian (commented):
We're approaching the solution, but we'll need to be extra precise on everything to ensure we have the absolutely simplest and straightforward implementation - every single line, parameter and method must be carefully inspected to ensure we've done the simplest thing without any extra layer/hot reloading/redundant method ;)

GDJS/Runtime/runtimeobject.ts
HarsimranVirk: Done 👍

GDJS/Runtime/runtimeobject.ts
HarsimranVirk: I think this is dead code, I forgot to remove this (my mistake). I started with this before looking at the hot reloader, then I understood it should be a part of the hot reloader itself. I'll remove it in the next commit.

GDJS/Runtime/runtimeobject.ts
HarsimranVirk: Done 👍

GDJS/Runtime/runtimeobject.ts
HarsimranVirk: The reason for breaking `addEffect` into multiple functions (`initializeEffect, applyEffect`) was because the renderer was created after the constructor (of runtime object) has been called. I wasn't sure how to retain the `effectsData` outside the constructor (and if it even made any sense to retain the initial effect data as a field in the object or not). Maybe there's a better design to handle it (instead of breaking the function and calling them at different times)?

GDJS/Runtime/runtimeobject.ts
HarsimranVirk: Oh, this is definitely a mistake. But I've one doubt here, we definitely don't need a layer here, but an `EffectsTarget` (which is either a runtime object or a layer). So, we'll still need to pass the runtime object (instead of a layer), but I guess that should still be very expensive right?

GDJS/Runtime/runtimeobject.ts
4ian: I see, we indeed need to `applyEffect` at the very end, outside of the base RuntimeObject constructor.
In this case: 
- add a comment when calling `initializeEffect` in the constructor ("We call initializeEffect as we can't use addEffect, as the renderer is not ready yet. This means applyEffect will have to be called later (see onCreated)").
- avoid this function entirely if you can (just do it in the constructor. Because each new function is complexifying the interface of the runtime object, which is already very large.
- if you can't avoid this function, at least make it private (underscore + private keyword in TypeScript). We don't want any outside caller to mess with this - it's an implementation detail, everyone else should be calling addEffect :)

GDJS/Runtime/runtimeobject.ts
4ian: Well it depends :) What do you mean exactly by "this should still be very expensive"?
If we pass the RuntimeObject, i.e: `this`, it's a reference that is passed like all objects in JavaScript, so internally it's just basically passing a pointer as an argument, which is very cheap :) 

It seems fair to pass to something called "initializeEffect": the effects data, the map where to store effects and the target (i.e: the runtime object) (which can be queried for its width/height for example).

GDJS/Runtime/runtimeobject.ts
4ian: This being said, I can't think of a better design that would allow us to scrap entirely initialize/apply effect - not sure it's possible unfortunately!
Let's just try to remove as much initializeEffect as possible to keep the new interface very small.

GDJS/Runtime/layer.ts
```diff
@@ -463,10 +463,8 @@ namespace gdjs {
      * Return the time elapsed since the last frame,
      * in milliseconds, for objects on the layer.
      */
-    getElapsedTime(): float {
-      return (
-        this._runtimeScene.getTimeManager().getElapsedTime() * this._timeScale
-      );
+    getElapsedTime(runtimeScene: RuntimeScene): float {
+      return runtimeScene.getTimeManager().getElapsedTime() * this._timeScale;
```
4ian, line 467: Sorry for the u turn, I think we can actually do the reverse decision:
- keep getElapsedTime without any parameter in the layer, and remove the parameter from gdjs.RuntimeObject too.
- the reason is that both the layer and the runtimeobject have a reference to the runtimescene, and there is little to no chance that in the future we want to drop this reference.
- it's no problem if some code is passing the runtimescene as an argument to the object getElapsedTime, we can just ignore it. You can make the argument optional.

This means updating the EffectsTarget and also the filters using getElapsedTime - but hopefully won't be too much work :)

GDJS/Runtime/spriteruntimeobject.ts
```diff
@@ -438,6 +438,7 @@ namespace gdjs {
      * Update the current frame of the object according to the elapsed time on the scene.
      */
     update(runtimeScene: gdjs.RuntimeScene): void {
+      super.update(runtimeScene);
```
4ian, line 441: You can remove this if you do the changes related to updating effects.

4ian:
Awesome, we're getting there! Take a look at GDJS tests that I think are failing. Rest should be ok, just see the comments :)

HarsimranVirk:
> Take a look at GDJS tests that I think are failing.

@4ian the tests are failing because of the firebase extension, they were failing since the time this PR was opened, and I assume it had been fixed in the later PRs. Should I pull the latest changes?

HarsimranVirk:
@4ian it turns out the tests are failing because they don't pass `effects` in the `objectData`. Should I add it everywhere? Or do something like `variables`, ie, only handle it if it exists?

Also, I think we've got duplicate commits after rebasing on master, let's open a new PR?

4ian:
> EDIT: I prefer adding effects everywhere, because they are more like behaviors, and not variables, it's going to be a little bit tedious to edit all of the tests, but this should be better I guess, what do you think?

Yes it's better, the `effects` array should be required (even if it's tedious to add first, it's more consistent and avoid surprises).

> Also, I think we've got duplicate commits after rebasing on master, let's open a new PR?

Yep!

PR #2737, opened 2021-06-12
[PathFinding] Add movement angle condition and expression

D8H:
This was asked by users to easily change the animation according to the direction.

2 usages by name or index:
MovementAngle

Test project:
https://www.dropbox.com/s/hqkx917bejz6aav/PathFindingMovementAngleExample.zip?dl=1

### DIFF
--- a/Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
@@ -206,7 +206,7 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
            "all of the forces it has."),
          _("Add to _PARAM0_ _PARAM3_ force of _PARAM1_ p/s on X axis and "
            "_PARAM2_ p/s on Y axis"),
-         _("Movement"),
+         _("Movement using forces"),
          "res/actions/force24.png",
          "res/actions/force.png")
 
@@ -222,7 +222,7 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
                   "using the specified angle and length."),
                 _("Add to _PARAM0_ _PARAM3_ force, angle: _PARAM1_ degrees and "
                   "length: _PARAM2_ pixels"),
-                _("Movement"),
+                _("Movement using forces"),
                 "res/actions/force24.png",
                 "res/actions/force.png")
 
@@ -238,7 +238,7 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
          _("Add a force to an object to make it move toward a position."),
          _("Move _PARAM0_ toward _PARAM1_;_PARAM2_ with _PARAM4_ force of _PARAM3_ "
            "pixels"),
-         _("Movement"),
+         _("Movement using forces"),
          "res/actions/force24.png",
          "res/actions/force.png")
 
@@ -258,7 +258,7 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
            "precisely, use the actions in the category \"Position\"."),
          _("Rotate _PARAM0_ around _PARAM1_;_PARAM2_ at _PARAM3_ deg/sec and "
            "_PARAM4_ pixels away"),
-         _("Movement"),
+         _("Movement using forces"),
          "res/actions/forceTourne24.png",
          "res/actions/forceTourne.png")
 
@@ -274,7 +274,7 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
                 _("Stop the object"),
                 _("Stop the object by deleting all of its forces."),
                 _("Stop _PARAM0_ (remove all forces)"),
-                _("Movement"),
+                _("Movement using forces"),
                 "res/actions/arreter24.png",
                 "res/actions/arreter.png")
 
@@ -486,10 +486,10 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
       .SetHidden();  // Inverted "Visible" condition  does the same thing.
 
   obj.AddCondition("Arret",
-                   _("Object is stopped"),
+                   _("Object is stopped (no forces applied on it)"),
                    _("Check if an object is not moving"),
                    _("_PARAM0_ is stopped"),
-                   _("Movement"),
+                   _("Movement using forces"),
                    "res/conditions/arret24.png",
                    "res/conditions/arret.png")
 
@@ -497,10 +497,10 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
       .MarkAsAdvanced();
 
   obj.AddCondition("Vitesse",
-                   _("Speed"),
+                   _("Speed (from forces)"),
                    _("Compare the overall speed of an object"),
                    _("the overall speed"),
-                   _("Movement"),
+                   _("Movement using forces"),
                    "res/conditions/vitesse24.png",
                    "res/conditions/vitesse.png")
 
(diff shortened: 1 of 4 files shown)


Extensions/PathfindingBehavior/pathfindingruntimebehavior.ts
```diff
@@ -156,6 +158,14 @@ namespace gdjs {
       return this._speed;
     }
 
+    getMovementAngle() {
+      return this._movementAngle;
+    }
+
+    movementAngleIsAround(degreeAngle: float, tolerance: float) {
+      return Math.abs((this._movementAngle - degreeAngle) % 360) <= tolerance;
```
4ian, line 166: I think we should use `gdjs.evtTools.common.angleDifference`?

Extensions/PathfindingBehavior/Extension.cpp
```diff
@@ -228,6 +228,22 @@ void DeclarePathfindingBehaviorExtension(gd::PlatformExtension& extension) {
...
+                          _("Angle of displacement of _PARAM0_ is _PARAM2_ (tolerance "
+                              ": _PARAM3_ degrees)"),
+                          _("Path"),
+                         "CppPlatform/Extensions/AStaricon24.png",
+                         "CppPlatform/Extensions/AStaricon16.png")
+
+        .AddParameter("object", _("Object"))
+        .AddParameter("behavior", _("Behavior"), "PathfindingBehavior")
+        .AddParameter("expression", _("Angle, in degrees"))
+        .AddParameter("expression", _("Tolerance"))
```
4ian, line 243: Tolerance, in degrees

Extensions/PathfindingBehavior/Extension.cpp
```diff
@@ -228,6 +228,22 @@ void DeclarePathfindingBehaviorExtension(gd::PlatformExtension& extension) {
         .SetFunctionName("GetSpeed")
         .SetIncludeFile("PathfindingBehavior/PathfindingRuntimeBehavior.h");
 
+    aut.AddScopedCondition("MovementAngleIsAround",
+                          _("Angle of movement"),
+                          _("Compare the angle of displacement of an object"),
+                          _("Angle of displacement of _PARAM0_ is _PARAM2_ (tolerance "
```
4ian, line 234: No space before :

Extensions/PathfindingBehavior/Extension.cpp
```diff
@@ -228,6 +228,22 @@ void DeclarePathfindingBehaviorExtension(gd::PlatformExtension& extension) {
         .SetFunctionName("GetSpeed")
         .SetIncludeFile("PathfindingBehavior/PathfindingRuntimeBehavior.h");
 
+    aut.AddScopedCondition("MovementAngleIsAround",
+                          _("Angle of movement"),
+                          _("Compare the angle of displacement of an object"),
```
4ian, line 233: Displacement => movement (I think it's more usual english) + period at the end of the sentence. Also maybe "on its path" to emphasize this is on the pathfinding path:

"Compare the angle of the movement of an object on its path."

Extensions/PathfindingBehavior/Extension.cpp
```diff
@@ -559,13 +575,23 @@ void DeclarePathfindingBehaviorExtension(gd::PlatformExtension& extension) {
         .SetFunctionName("GetCellHeight")
         .SetIncludeFile("PathfindingBehavior/PathfindingRuntimeBehavior.h");
 
+    aut.AddExpression("MovementAngle",
+                    _("Angle of movement"),
+                    _("Angle of displacement of an object"),
```
4ian, line 580: For both: "Angle of movement on the path"?

Extensions/PathfindingBehavior/Extension.cpp
```diff
@@ -559,13 +575,23 @@ void DeclarePathfindingBehaviorExtension(gd::PlatformExtension& extension) {
...
+                    _("Path"),
+                    "CppPlatform/Extensions/AStaricon16.png")
+        .AddParameter("object", _("Object"))
+        .AddParameter("behavior", _("Behavior"), "PathfindingBehavior")
+        .SetFunctionName("GetMovementAngle")
```
4ian, line 585: You can remove this FunctionName as it's not implemented in the native engine. Also remove the SetIncludeFile

Extensions/PathfindingBehavior/Extension.cpp
4ian: (I've used displacement in some part of the engine but not sure it's a good idea to surface this term to the user after thinking about it)

D8H:
These conditions work only for the forces, don't they?
MovementCondition
If so, I think that "Movement" is too general and can confuse user. Could we rename it to "Movement (force)" or something like this?
Could it be useful to have all these conditions in every movement behaviors?

4ian:
> If so, I think that "Movement" is too general and can confuse users. Could we rename it to "Movement (force)" or something like this?
> Could it be useful to have all these conditions in every movement behaviors?

Yeah, I agree, it's confusing. We could rename this to "Movement using forces" to be very clear?

> Could it be useful to have all these conditions in every movement behaviors?

Could be useful, though I think we should for each have a very clear naming "Angle of movement on its path" (for pathfinding), "Angle of the movement (top-down)" (for top-down), etc...
Basically, have, even if redundant, an information about which behavior this is :)

4ian:
Just an additional note, we can consider that the "forces" were like a "builtin" movement behavior that is added on all objects. I think we should indeed clarify this with these new namings: 
- "Movement using forces" as the category title
- "Angle of movement using forces"
- "Object is stopped (no forces applied on it)"
- "Speed (from forces)"
- "An object is moving toward another (using forces)"

What do you think?

Bouh:
Think to unify the naming here too https://github.com/4ian/GDevelop/pull/2465

4ian:
For force Vs velocity, that's indeed a good discussion, though I would be inclined to keep force because of its widespread usage?

D8H:
> For force Vs velocity, that's indeed a good discussion, though I would be inclined to keep force because of its widespread usage?

Yes, we can think of a way to explain what a force really is another time.
I still took the "sum" instead of "average" from https://github.com/4ian/GDevelop/pull/2465

Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
```diff
@@ -509,17 +509,18 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
       .MarkAsAdvanced();
 
   obj.AddCondition("AngleOfDisplacement",
-                   _("Angle of movement"),
-                   _("Compare the angle of displacement of an object"),
-                   _("Angle of displacement of _PARAM0_ is _PARAM1_ (tolerance "
+                   _("Angle of movement (using forces)"),
+                   _("Compare the angle of movement of an object using forces."
+                     " according to the forces."),
```
4ian, line 514: Typo here :) "Compare the angle of movement of an object according to the forces applied on it."?

Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
```diff
@@ -1149,24 +1150,24 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
...
                       "object must move."),
                     _("_PARAM0_ is moving toward _PARAM1_"),
-                    _("Movement"),
+                    _("Movement using forces"),
                     "res/conditions/sedirige24.png",
                     "res/conditions/sedirige.png")
       .AddParameter("objectList", _("Object"))
       .AddParameter("objectList", _("Object 2"))
-      .AddParameter("expression", _("Angle of tolerance"))
+      .AddParameter("expression", _("Angle of tolerance in degrees"))
```
4ian, line 1170: nitpicking: let's normalise this description so that it's like to other: "Tolerance, in degrees"

D8H:
What do you think of making these conditions to the group "Movement using path finding":
* Destination reached
* Path found
* Angle of movement on its path
* Speed on its path

As they are the "variable" of the movement contrary to the ones that are just settings.

4ian:
Good idea, I think it's more logical indeed. I would say to put them in a group called "Movement on the path" (slightly more user friendly than "Movement using pathfinding"?) and the others can be "Pathfinding configuration"?

PR #2740, opened 2021-06-13
Move JSON and JS variable conversions to gdjs.Variable

arthuro555:
Moves javascript and JSON convertion utilities from `gdjs.evtTools.network` to gdjs.Variable.


4ian (approved):
A welcome refactoring, thanks 👍

4ian:
Actually there is an issue with undefined variables that will crash, thankfully TypeScript caught this for us:
image

arthuro555:
> Actually there is an issue with undefined variables that will crash

Yes, I was working on that and fixing the tests + formatting.

4ian:
Ah I started fixing this and pushed on the branch. Do you want to force push your changes?

arthuro555:
No problem, I saw your push and just pulled the changes and merged it with my changes already

arthuro555:
There is a small breaking change in the JSON formatting, it will look like `{"a":1}` instead of `{"a": 1}` now, do you think this could be an issue?

4ian:
Not an issue, the previous formatting was not very standard, and no one should really rely on this specific formatting :)

PR #2741, opened 2021-06-13
Add various DX improvements

arthuro555:
Adds a few developper experience improvements:

- Replace all the `electron-<platform>` scripts with a `start` script.
- Add a VSCode option to automatically launch the development server when opening the workspace.
- Make the VSCode debugger launch automatically the dev server is it isn't started yet and await end of compilation before launching the browser.
- Make the code snippets use as default namespace `gdjs.evtTools.{folderName}` instead of `gdjs.{folderName}` for the function name.
- Add starting the electron app to the VSCode tasks.
- Prevent accidentally running twice the development server via VSCode.
- Display `#if defined(EMSCRIPTEN)` code non-transparently in VSCode.
- Add `GDJS/Runtime-dist` to the VSCode list of ignored directories.


arthuro555:
I have looked into using tasks dependencies to make the electron app dependent on the development server but then the webserver had to be killed before vscode accepts to launch the electron app

newIDE/electron-app/package.json
```diff
@@ -14,9 +14,7 @@
     "build": "node scripts/build.js",
     "format": "prettier --write --ignore-path .gitignore \"**/*.js\"",
     "check-format": "prettier --list-different --ignore-path .gitignore \"**/*.js\"",
-    "electron-win": "node node_modules/electron/cli.js app",
-    "electron-linux": "./node_modules/electron/dist/electron app",
-    "electron-mac": "./node_modules/electron/dist/Electron.app/Contents/MacOS/Electron app",
+    "start": "electron app",
```
4ian, line 17: Nice! Much simpler.

Can you update the README and the other package.json having these commands (renaming to start-electron-app for the other)

4ian:
Thanks! 👍

PR #2743, opened 2021-06-14
Fix NAN for cursor position on scene editor

Bouh:
This PR fix this NAN.
It happend when a scene is open up, but the cursor in not on `<InstancesEditor>` when mounting is processed.
image

### DIFF
--- a/newIDE/app/src/InstancesEditor/index.js
+++ b/newIDE/app/src/InstancesEditor/index.js
@@ -204,9 +204,11 @@ export default class InstancesEditor extends Component<Props> {
 
       this._onBackgroundClicked(event.data.global.x, event.data.global.y);
     });
-    this.backgroundArea.on('mousemove', event =>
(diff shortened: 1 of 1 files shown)


newIDE/app/src/InstancesEditor/index.js
```diff
@@ -497,8 +497,8 @@ export default class InstancesEditor extends Component<Props> {
   };
 
   _onMouseMove = (x: number, y: number) => {
-    this.lastCursorX = x;
-    this.lastCursorY = y;
+    this.lastCursorX = x || 0;
+    this.lastCursorY = y || 0;
```
4ian, line 501: I'm going to nitpick on this :P 

x and y should never be null/undefined, because the types are set to `number` (and not ?number).
So if null/undefined appeared here, the problem should be fixed beforehand, probably in `this.backgroundArea.on('mousemove', event =>`? If the event is not containing the x/y, we should fix it at this moment (so that the rest of the code can "confidently" use these).

PR #2751, opened 2021-06-16
The instance editor can transform several instances linearly

D8H:
These 2 videos to show some behavior changes:
* When translating over a grid, the selection will magnet following the cursor direction to be able to magnet to the left or to the right for instance.
It may need a bit of tweaking but it seems usable (I changed the grid from floor to round since the capture, it will impact anything that use the grid, but for the best I hope).
https://user-images.githubusercontent.com/2611977/122686744-1f85b500-d213-11eb-815f-c316d83a477b.mp4

* Same goes for scaling, this video shows the toggling between bottom and right alignment because the aspect ratio is locked.
https://user-images.githubusercontent.com/2611977/122686747-290f1d00-d213-11eb-8a55-77485159676a.mp4


newIDE/app/src/InstancesEditor/InstancesResizer.js
```diff
@@ -76,42 +82,90 @@ export default class InstancesResizer {
...
+      const selectedInstance = instances[i];
+      hasRotatedInstance = selectedInstance.getAngle() !== 0;
+    }
+
     for (let i = 0; i < instances.length; i++) {
       const selectedInstance = instances[i];
 
       if (!selectedInstance.hasCustomSize()) {
         selectedInstance.setCustomWidth(
-          this.instanceMeasurer.getInstanceWidth(selectedInstance)
```
4ian, line 84: Might be worth doing a single call to `getInstanceOBB` there

newIDE/app/src/InstancesEditor/InstancesResizer.js
```diff
@@ -76,42 +82,90 @@ export default class InstancesResizer {
...
-        this._roundWidth(
-          initialSize.width + this._getSizeDeltaX(proportional, initialSize)
-        )
-      );
-      selectedInstance.setCustomHeight(
-        this._roundHeight(
-          initialSize.height + this._getSizeDeltaY(proportional, initialSize)
-        )
-      );
+      if (hasRotatedInstance) {
```
4ian, line 130: Do you think we could avoid this special case when there are rotated instances? Could you describe what changes when there are rotated instances? :)

newIDE/app/src/InstancesEditor/InstancesResizer.js
```diff
@@ -1,4 +1,6 @@
 // @flow
+import Rectangle from '../Utils/Rectangle';
+
 type Dimension = {|
```
4ian, line 4: I think the Dimension type can be removed?

newIDE/app/src/InstancesEditor/SelectionRectangle.js
```diff
@@ -10,9 +11,10 @@ export default class SelectionRectangle {
   pixiRectangle: PIXI.Graphics;
   selectionRectangleStart: any;
   selectionRectangleEnd: any;
-  _instancesInSelectionRectangle: gdInitialInstance[];
+  _instancesInSelectionRectangle: any[];
```
4ian, line 14: Is this to appease Flow? I think this should be kept to `gdInitialInstance[]`?

newIDE/app/src/Utils/Rectangle.js
```diff
@@ -0,0 +1,27 @@
+// @flow
+export default class Rectangle {
```
4ian, line 2: Just thinking: should this be called "AxisAlignedRectangle"?

newIDE/app/src/InstancesEditor/InstancesRenderer/LayerRenderer.js
```diff
@@ -134,22 +141,102 @@ export default class LayerRenderer {
...
+    this.temporaryRectanglePath[2][0] = right;
+    this.temporaryRectanglePath[2][1] = bottom;
+
+    this.temporaryRectanglePath[3][0] = right;
+    this.temporaryRectanglePath[3][1] = top;
+
+    return this.temporaryRectanglePath;
+  }
+
+  rotate(
```
4ian, line 194: Let's move this outside of the class as a pure function (avoid accidental reuse of it outside of this context)

newIDE/app/src/InstancesEditor/InstancesRenderer/LayerRenderer.js
```diff
@@ -134,22 +141,102 @@ export default class LayerRenderer {
...
-  getInstanceHeight = (instance: gdInitialInstance) => {
+  getUnrotatedInstanceHeight = (instance: gdInitialInstance) => {
     if (instance.hasCustomSize()) return instance.getCustomHeight();
 
     return this.renderedInstances[instance.ptr]
       ? this.renderedInstances[instance.ptr].getDefaultHeight()
       : 0;
   };
 
+  getInstanceOBB(instance: gdInitialInstance, bounds: Rectangle): Rectangle {
```
4ian, line 160: Is this really an "oriented bounding box" that we're returning here? I think this should rather be called "getUnrotatedInstanceAABB" (i.e: the axis aligned bounding box of the unrotated instance) (but not getInstanceUnrotatedAABB, because by definition an AABB cannot be "rotated").

4ian (commented):
Looks like a good start! :)

newIDE/app/src/InstancesEditor/InstancesResizer.js
D8H: When there is a rotated instance, the scaling will keep the proportions. It's to avoid shear transformations as it's not supported by the model. I will add a comment.
Shear

newIDE/app/src/InstancesEditor/InstancesResizer.js
D8H: Yes, I removed it.

newIDE/app/src/InstancesEditor/SelectionRectangle.js
D8H: I don't remember. I'm reverting it to check this next time I do a flow pass.

newIDE/app/src/Utils/Rectangle.js
D8H: I don't know. Most frameworks I used call it Rectangle or even Rect. I don't think people could be misled because a rotated rectangle class would be a bit odd.

4ian: Let's keep Rectangle then :)

4ian:
Lots going on here! :) I'll review this tomorrow morning if you're ready :)

D8H:
It should be ready to review.

I rewrite what I did to magnet the selection when it moves to use the same method as Inkscape. The code is a lot simpler that way (less to review :wink:). It magnets the nearest corner from where it was grabbed.

I tested with this project. There is 2 scenes, one with only strait instances (with every direction) and one with rotated instances. The black object has a top-left origin and the yellow one a centered origin.
https://www.dropbox.com/s/esmp6ulscjyabis/InstanceEditorTest.zip?dl=1

newIDE/app/src/InstancesEditor/InstancesRenderer/LayerRenderer.js
```diff
@@ -221,13 +311,12 @@ export default class LayerRenderer {
    */
   _isInstanceVisible(instance: gdInitialInstance) {
     //TODO: Properly handle rotation
```
4ian, line 313: I believe this TODO can be removed as you fixed the culling thanks to the proper AABB when objects are rotated :)

newIDE/app/src/InstancesEditor/InstancesResizer.js
```diff
@@ -20,98 +61,243 @@ export default class InstancesResizer {
...
+    const isLeft =
+      grabbingLocation === 'TopLeft' ||
+      grabbingLocation === 'BottomLeft' ||
+      grabbingLocation === 'Left';
+    const isTop =
+      grabbingLocation === 'TopLeft' ||
+      grabbingLocation === 'TopRight' ||
+      grabbingLocation === 'Top';
+
+    const flippedTotalDeltaX = isLeft
```
4ian, line 185: "signed"?

newIDE/app/src/InstancesEditor/InstancesResizer.js
```diff
@@ -20,98 +61,243 @@ export default class InstancesResizer {
...
+        selectedInstance
+      );
+      let initialInstanceOriginPosition = this._getOrCreateInstanceOriginPosition(
+        selectedInstance
+      );
+
+      const angle = ((selectedInstance.getAngle() % 360) + 360) % 360;
+      if (
+        (!proportional && !hasRotatedInstance && angle === 90) ||
+        angle === 270
```
4ian, line 247: I'm not sure I understand these checks and why angle === 270 is a special case here?

newIDE/app/src/InstancesEditor/InstancesResizer.js
4ian: (rather than flipped?)

4ian (commented):
Very quick review pass! :)

newIDE/app/src/InstancesEditor/InstancesRenderer/LayerRenderer.js
D8H: Ok, I wasn't sure, I'll remove it.

newIDE/app/src/InstancesEditor/InstancesResizer.js
D8H: Actually, the roundedTotalDeltaX can be negative and so do flippedTotalDeltaX.
This is more like a symmetry to end up in the nominal case of a bottom-left grabbed node.
I will add a comment.

newIDE/app/src/InstancesEditor/InstancesResizer.js
D8H: I think I forgot the parenthesis around the "or" and the formatter made it weird.
I'll fix this.

4ian:
I've yet to review everything, sorry it's so long, I have to wrap my mind around the new logic. But I've tested it's awesome!! :)

newIDE/app/src/InstancesEditor/InstancesRenderer/LayerRenderer.js
```diff
@@ -134,22 +143,93 @@ export default class LayerRenderer {
...
+    const bottom = top + this.getUnrotatedInstanceHeight(instance);
+
+    bounds.left = left;
+    bounds.right = right;
+    bounds.top = top;
+    bounds.bottom = bottom;
+    return bounds;
+  }
+
+  _getRectanglePath(instance: gdInitialInstance): Array<[number, number]> {
```
4ian, line 178: As this is used only in _getRotatedRectangle, I would exceptionally put this back inside `_getRotatedRectangle `

newIDE/app/src/InstancesEditor/InstancesResizer.js
```diff
@@ -20,98 +61,247 @@ export default class InstancesResizer {
...
+      hasRotatedInstance = instances[i].getAngle() % 90 !== 0;
+    }
+    if (proportional || hasRotatedInstance) {
+      // Choose the axis where the selection is the biggest.
+      // That way the cursor is always on one edge.
+      if (
+        !flippedTotalDeltaY ||
+        (flippedTotalDeltaX &&
+          flippedTotalDeltaX * initialSelectionAABB.height() >
+            initialSelectionAABB.width() * flippedTotalDeltaY)
```
4ian, line 215: Is there a way to make these conditions more "symetric"?

newIDE/app/src/InstancesEditor/InstancesRotator.js
```diff
@@ -1,46 +1,118 @@
...
-  instanceAngles: { [number]: number };
-  totalDeltaX: number;
-  totalDeltaY: number;
+  _instanceMeasurer: any;
+  _instanceAngles: { [number]: number } = {};
+  _instancePositions: { [number]: { x: number, y: number } } = {};
+  _instanceAABBs: { [number]: Rectangle } = {};
+  totalDeltaX: number = 0;
+  totalDeltaY: number = 0;
+  _anchorIsUpToDate: boolean = false;
```
4ian, line 11: Could you explain what this anchor is doing exactly?

newIDE/app/src/InstancesEditor/SelectionRectangle.js
```diff
@@ -13,6 +14,7 @@ export default class SelectionRectangle {
   _instancesInSelectionRectangle: gdInitialInstance[];
 
   selector: gdInitialInstanceJSFunctor;
+  _temporaryRectangle: Rectangle;
```
4ian, line 17: For temporaries, it's less important for the IDE than in the runtime, but let's have a comment explaining where they are used.

4ian: Or just renaming it to `_temporaryAABB` (because "Rectangle" is just the name of the type, this does not give information about what this temporary is used for).

4ian (commented):
Another pass of comments :)

newIDE/app/src/InstancesEditor/InstancesRotator.js
D8H: It's the fixed point, here it's always the selection center but it could be made dragable one day.

newIDE/app/src/InstancesEditor/InstancesResizer.js
```diff
@@ -20,98 +61,247 @@ export default class InstancesResizer {
...
+      );
+      let initialInstanceOriginPosition = this._getOrCreateInstanceOriginPosition(
+        selectedInstance
+      );
+
+      const angle = ((selectedInstance.getAngle() % 360) + 360) % 360;
+      if (
+        !proportional &&
+        !hasRotatedInstance &&
+        (angle === 90 || angle === 270)
```
4ian, line 251: Add a comment explaining why this if branch is specific to angle 90 and angle 270 (which can be understood by seeing you swap scaleX/Y in the setCustomWidth/Height, but needs a bit of time before it's clear, at least to me).

newIDE/app/src/InstancesEditor/InstancesResizer.js
```diff
@@ -1,15 +1,56 @@
...
   instanceMeasurer: any;
   options: Object;
-  instanceSizes: { [number]: Dimension };
-  totalDeltaX: number;
-  totalDeltaY: number;
+  _initialSelectionAABB: ?Rectangle = null;
+  _unrotatedInstanceAABBs: { [number]: Rectangle } = {};
+  _instanceAABBs: { [number]: Rectangle } = {};
+  _instancePositions: { [number]: { x: number, y: number } } = {};
+  totalDeltaX: number = 0;
```
4ian, line 50: Let's add a comment for these: "The coordinates of the vector of the resize being done", as the class is becoming more complex.

newIDE/app/src/InstancesEditor/InstancesRenderer/LayerRenderer.js
```diff
@@ -25,20 +27,28 @@ export default class LayerRenderer {
...
+  renderedInstances: { [number]: RenderedInstance } = {};
   pixiContainer: PIXI.Container;
 
   /** Functor used to render an instance */
   instancesRenderer: gdInitialInstanceJSFunctor;
 
   wasUsed: boolean = false;
 
+  _temporaryRectangle: Rectangle = new Rectangle();
+  _temporaryRectanglePath: Array<[number, number]> = [
```
4ian, line 45:
```suggestion
  _temporaryRectanglePath: Polygon = [
```

newIDE/app/src/InstancesEditor/SelectedInstances.js
```diff
@@ -242,54 +250,39 @@ export default class SelectedInstances {
...
-    resizeButtonPos[0] += buttonPadding;
-    resizeButtonPos[1] += buttonPadding;
 
-    const rightResizeButtonPos = this.toCanvasCoordinates(
-      x2,
-      y1 + (y2 - y1) / 2
-    );
-    rightResizeButtonPos[0] += buttonPadding;
-    rightResizeButtonPos[1] -= smallButtonSize / 2;
+    //Position the resize button.
```
4ian, line 255: Much nicer now with all the buttons :)
Update the comment by the way: 
```suggestion
    // Position the resize buttons.
```

newIDE/app/src/InstancesEditor/InstancesRenderer/index.js
```diff
@@ -62,46 +64,47 @@ export default class InstancesRenderer {
...
-        const layerRenderer = this.layersRenderers[layerName];
-        if (!layerRenderer) return instance.getY();
-
-        return layerRenderer.getInstanceTop(instance);
-      },
-      getInstanceWidth: (instance: gdInitialInstance) => {
-        if (instance.hasCustomSize()) return instance.getCustomWidth();
 
+    this.temporaryRectangle = new Rectangle();
+    //TODO extract this to a class to have type checking (maybe rethink it)
```
4ian, line 69: Yes, at least this could be simply a type (`type InstanceMeasurerInterface = {| getInstanceAABB: ... |};`) to ensure we follow the same interface. Ok not to do it now.

4ian (commented):
Another bunch of comments. We're almost there I think! 👍
Again, looking forward to see this merged, this will be an awesome improvement.

newIDE/app/src/InstancesEditor/InstancesRotator.js
4ian: Thanks for the renaming!

4ian:
Finally got the time to finish this review! 

Let me tell you that this is awesome! 👍 This is a long overdue improvement that will make using the scene editor much more comfortable when authoring levels. I think we'll wonder how we did before :)

Appreciate your patience on this and the attention to details (grid handling, magneting to the closest corner, etc...). Very happy to see this in for the next version, this will be a huge quality of life improvement :)

PR #2760, opened 2021-06-24
Make the button "add a behavior to the object" at the bottom of the list of behaviors

Bouh:
Continuity of https://github.com/4ian/GDevelop/commit/24c13e1022ba95496ad3f58f3a50dc21e77a22b2 for the behaviors list.

image


Bouh:
Same is done for buttons on the collision mask list.
I've also added a first quadrilateral collision mask in list when we choose a custom mask.

image

Bouh:
I've also added some code for switch on automatic collision mask when the user removes all manual collision mask in the list.

I'm happy with my code for now.
It's ready for review.

4ian (approved):
Looks great! Good idea to restore the collision mask when all polygons are removed.

PR #2770, opened 2021-06-30
Add a console to the debugger

arthuro555:
Draft console for a console for the debugger.
TODO:
- [ ] Convert some more internal console.log calls to gdjs.log/gdjs.runtimeLog
- [ ] Test further for bugs
- [ ] Add filtering of logs by event group
- [ ] Add toggle console icon to the toolbar
- [ ] Fix bug with contents of the console disappearing until a new log is sent when switching tabs away and back from the debugger
- [ ] ~~Make the text wrap instead of having to scroll~~ **I tried for hours and dind't manage to make that work with react-virtualized**
- [X] Add an action to log to the console to the `DebuggerTools` extension
- [X] Make the list of logs virtualized to prevent the IDE from lagging out
- [X] Monkey-patch console.log to also print to the debugger console
- [X] Add base implementation of console to the IDE and debugger client

I would appreciate feedback on the implementation so far.

### DIFF
--- a/Extensions/DebuggerTools/debuggertools.ts
+++ b/Extensions/DebuggerTools/debuggertools.ts
@@ -14,6 +14,24 @@ namespace gdjs {
       };
 
       /**
+       * Logs a message to the console.
+       * @param runtimeScene - The current scene.
+       * @param message - The message to log.
+       * @param group - The group of messages it belongs to.
+       * @param type - The type of log (info, warning or error).
+       */
+      export const log = function (
+        runtimeScene: gdjs.RuntimeScene,
+        message: string,
+        type: 'info' | 'warning' | 'error',
+        group: string
+      ) {
+        const game = runtimeScene.getGame();
+        if (game._debuggerClient)
+          game._debuggerClient.log(message, { group, type });
+      };
+
+      /**
        * Enable or disable the debug draw.
        * @param runtimeScene - The current scene.
        * @param enableDebugDraw - true to enable the debug draw, false to disable it.
--- a/Extensions/DebuggerTools/debuggertools.ts
+++ b/Extensions/DebuggerTools/debuggertools.ts
@@ -21,14 +21,11 @@ namespace gdjs {
        * @param type - The type of log (info, warning or error).
        */
       export const log = function (
-        runtimeScene: gdjs.RuntimeScene,
         message: string,
         type: 'info' | 'warning' | 'error',
         group: string
       ) {
-        const game = runtimeScene.getGame();
-        if (game._debuggerClient)
-          game._debuggerClient.log(message, { group, type });
+        gdjs.log(group, message, type);
       };
 
       /**
--- a/GDJS/Runtime/gd.ts
+++ b/GDJS/Runtime/gd.ts
@@ -502,7 +502,7 @@ namespace gdjs {
     message: string,
     type: 'info' | 'warning' | 'error' = 'info'
   ) => {
-    const logger = _console[type];
+    const logger = _console[type] || _console.info;
     logger(`[${group}] ${message}`);
   };
 
--- a/GDJS/Runtime/gd.ts
+++ b/GDJS/Runtime/gd.ts
@@ -311,7 +311,7 @@ namespace gdjs {
    * @private
    */
   export const registerGlobalCallbacks = function (): void {
-    console.warn(
+    gdjs.runtimeLog(
       "You're calling gdjs.registerGlobalCallbacks. This method is now useless and you must not call it anymore."
     );
   };
@@ -344,7 +344,7 @@ namespace gdjs {
     if (name !== undefined && gdjs.objectsTypes.containsKey(name))
       return gdjs.objectsTypes.get(name);
 
-    console.warn('Object type "' + name + '" was not found.');
+    gdjs.runtimeLog('Object type "' + name + '" was not found.', 'warning');
     return gdjs.objectsTypes.get(''); //Create a base empty runtime object.
   };
 
@@ -359,7 +359,7 @@ namespace gdjs {
     if (name !== undefined && gdjs.behaviorsTypes.containsKey(name))
       return gdjs.behaviorsTypes.get(name);
 
-    console.warn('Behavior type "' + name + '" was not found.');
+    gdjs.runtimeLog('Behavior type "' + name + '" was not found.', 'warning');
     return gdjs.behaviorsTypes.get(''); //Create a base empty runtime behavior.
   };
 
@@ -486,6 +486,33 @@ namespace gdjs {
       hex[r[15]]
     );
   };
+
+  const _console = {
+    info: console.log,
+    warning: console.warn,
+    error: console.error,
+  };
+
+  /**
+   * Internal method for logging messages to the JS console or the Debugger if available.
+   * Should be used in engine code or extensions, console.log is fine for JS events.
+   */
+  export let log = (
+    group: string,
+    message: string,
+    type: 'info' | 'warning' | 'error' = 'info'
+  ) => {
+    const logger = _console[type];
+    logger(`[${group}] ${message}`);
+  };
+
+  /**
+   * Shortcut loggin function for the runtime engine code.
+   */
+  export const runtimeLog = (
+    message: string,
+    type: 'info' | 'warning' | 'error' = 'info'
+  ) => log('Engine runtime', message, type);
 }
 
 //Make sure console.warn and console.error are available.
--- a/GDJS/Runtime/runtimegame.ts
+++ b/GDJS/Runtime/runtimegame.ts
@@ -225,7 +225,10 @@ namespace gdjs {
         }
       }
       if (scene === null) {
-        console.warn('The game has no scene called "' + sceneName + '"');
+        gdjs.runtimeLog(
+          'The game has no scene called "' + sceneName + '"',
+          'warning'
+        );
       }
       return scene;
     }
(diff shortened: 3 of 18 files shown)


Bouh:
Can we have internal ``console.warn()`` and ``console.error()`` in dev tools **and** in this new console in the debugger?
I'd be surprised if I didn't see any errors in dev tools because the warnings go into the new debugger console and vice versa.
Let's make sure that errors are visible in both consoles.

I think we should also filter the messages coming from the engine and those added by the user.
A simple checkbox at the top of the log should do the trick. Kind of like how the search in the event sheet works, there are filters, that is the idea.
It is important to make this option possible because many warnings or errors can be issued by the engine.
If a user just wants to see his message, it will be a mess, and maybe he also doesn't care that his object X has no texture. Or that the bitmap object with the X font will be unloaded.
Also, filters are active for both by default, this way all messages are visible.

arthuro555:
> Can we have internal console.warn() and console.error() in dev tools and in this new console in the debugger?

Unless there is an error somewhere that should already be the case.

> I think we should also filter the messages coming from the engine and those added by the user.

That was already planned, and the reason why messages have an "internal" property for those kind of messages.

newIDE/app/src/Debugger/index.js
```diff
@@ -188,6 +210,13 @@ export default class Debugger extends React.Component<Props, State> {
       }));
     } else if (data.command === 'hotReloader.logs') {
       // Nothing to do.
+    } else if (data.command === 'console.log') {
```
4ian, line 213: nit: `console.logs`
(this means you're receiving the logs from the console, not that you're instructing the console to log)

newIDE/app/src/Debugger/DebuggerConsole.js
```diff
@@ -0,0 +1,100 @@
+import React from 'react';
```
4ian, line 1: You need `// @flow`

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.ts
```diff
@@ -156,7 +170,70 @@ namespace gdjs {
...
+          (accumulator, value) => accumulator + value.toString(),
+          ''
+        ),
+        { group: 'JavaScript', type }
+      );
+    }
+
+    log(
+      message: string,
+      additionalData?: {
```
4ian, line 214: Can you simplify by making this non optional?

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.ts
```diff
@@ -156,7 +170,70 @@ namespace gdjs {
           );
         }
       };
-      return;
+
+      ((log, warn, error, gdjsLog) => {
+        // Hook the console logging functions to log to the Debugger as well
+        console.log = (...messages) => {
```
4ian, line 176: Can you check how `...rest` arguments are handled by esbuild in the generated code?

GDJS/Runtime/gd.ts
```diff
@@ -486,6 +486,33 @@ namespace gdjs {
...
+    group: string,
+    message: string,
+    type: 'info' | 'warning' | 'error' = 'info'
+  ) => {
+    const logger = _console[type] || _console.info;
+    logger(`[${group}] ${message}`);
+  };
+
+  /**
+   * Shortcut loggin function for the runtime engine code.
```
4ian, line 510: Typo in loggin

Extensions/DebuggerTools/debuggertools.ts
```diff
@@ -13,6 +13,21 @@ namespace gdjs {
         runtimeScene.getGame().pause(true);
       };
 
+      /**
+       * Logs a message to the console.
+       * @param runtimeScene - The current scene.
+       * @param message - The message to log.
+       * @param group - The group of messages it belongs to.
+       * @param type - The type of log (info, warning or error).
```
4ian, line 21: Order is not the same as parameters

GDJS/Runtime/runtimegame.ts
```diff
@@ -520,73 +523,77 @@ namespace gdjs {
...
+          if (that._sceneStack.step(elapsedTime)) {
+            that.getInputManager().onFrameEnded();
+            return true;
+          }
+          return false;
+        });
+        setTimeout(() => {
+          this._setupSessionMetrics();
+        }, 10000);
+      } catch (e) {
```
4ian, line 594: Don't put a giant try/catch here, we want crash to happen without being swallowed or risking calling a function of the engine that would itself trigger another crash. Basically, if a crash happen here, everything is already lost. At the very least, rethrow the error because it should surface to the traditional JS console and debugger.

GDJS/Runtime/gd.ts
4ian: But take a look at the "createLogger" function I was talking about instead :) This way each file can have its logger which is associated to a group and has a common syntax (otherwise we'll make more of these "shortcut" methods in the future, and it's not very elegant)

4ian (commented):
Made a quick review pass. 
What about a slightly different api where you don't pass the level as an argument, but use a function?
`gdjs.logInfo()`, `gdjs.logWarning()`, `gdjs.logError()`?

or even something like 
```
// Done only ONCE per file:
const logger = gdjs.makeLogger("Extension blabla"); // You specify the group only once.

// Then you use it almost as console.log:
logger.log("Hello world");
logger.info("Same as log");
logger.warn("A warning");
logger.error("An error");
```

This allows to keep a simple API like `console.log`, and is I think a bit more intuitive than passing the type of the error at the end of the function.

newIDE/app/src/Debugger/index.js
arthuro555: It is not sending a list of logs though, it really is just asking to log one message

4ian: A right I was mistaken! Looking good then,

newIDE/app/src/Debugger/index.js
```diff
@@ -124,15 +127,34 @@ export default class Debugger extends React.Component<Props, State> {
...
-          }),
+          ({
+            selectedId,
+            logs,
+            debuggerGameData,
+            profilerOutputs,
+            profilingInProgress,
+          }) => {
+            // Remove any data bound to the instance that might have been stored
+            // Otherwise it would be a data leak
```
4ian, line 138: nitpicking:
```suggestion
            // Remove any data bound to the instance that might have been stored.
            // Otherwise this would be a memory leak.
```

GDJS/Runtime/runtimegame.ts
arthuro555: Well in case of a crash it is easier for the user to use the integrated tooling to send us the error message rather than having them to open the js debugger etc so I would keep the try catch

newIDE/app/src/Debugger/DebuggerConsole.js
```diff
@@ -0,0 +1,101 @@
...
+    cache.clearAll();
+  };
+
+  return (
+    <>
+      <Checkbox
+        style={{ margin: 10 }}
+        checked={showGroup}
+        label="Show group names"
+        onCheck={(_, value) => setShowGroup(value)}
```
Bouh, line 51:
```suggestion
        onCheck={(e, value) => setShowGroup(value)}
```
Because event :)

arthuro555: The `_` signifies we don't care about the event though only about the value

GDJS/Runtime/websocket-debugger-client/websocket-debugger-client.ts
arthuro555: They are emitted unchanged. I think `...rest` is compatible with a lot of browsers, according to [Can I use](https://caniuse.com/mdn-javascript_operators_spread_spread_in_function_calls) support for spread arguments in functions have been added alongside the [`let`](https://caniuse.com/let) keyword, which we use a lot so we should be safe.

4ian: Alright, thanks for checking!

arthuro555:
Ok, I am done I believe in term of features. I'm pretty sure flow will fail since I always mess up flow types or accidentally use Typescript stuff instead, and I can't run it myself locally, so I'd like it if someone could fix them for me 😓.

I am not very good at design, so I would appreciate comments on the result visually:
image
image
image
image
image

newIDE/app/src/Debugger/index.js
```diff
@@ -188,6 +215,23 @@ export default class Debugger extends React.Component<Props, State> {
...
     } else if (data.command === 'hotReloader.logs') {
       // Nothing to do.
+    } else if (data.command === 'console.log') {
+      // Filter out unavoidable warnings that do not concern non-engine devs.
+      if (
+        data.payload.group === 'JavaScript' &&
+        (data.payload.message.includes('Electron Security Warning') ||
+          data.payload.message.includes(
+            'Warning: This is a browser-targeted Firebase bundle'
+          ))
```
4ian, line 225: Nitpicking: might be worth extracting this into a function so that it's easy to enhance with other filters in the future.

newIDE/app/src/stories/index.js
```diff
@@ -2639,6 +2641,7 @@ storiesOf('DebuggerContent', module)
           onStopProfiler={action('stop profiler')}
           profilerOutput={profilerOutput}
           profilingInProgress={true}
+          logs={consoleLogs}
```
4ian, line 2644: Thanks for updating this 👍

newIDE/app/src/stories/index.js
```diff
@@ -113,6 +113,7 @@ import {
 } from '../Utils/GDevelopServices/ApiConfigs.js';
 import debuggerGameDataDump from '../fixtures/DebuggerGameDataDump.json';
 import profilerOutput from '../fixtures/ProfilerOutputsTestData.json';
+import consoleLogs from '../fixtures/ConsoleTestData.json';
```
4ian, line 116: nitpicking: let's name this consoleTestData like the file.

GDJS/Runtime/runtimegame.ts
4ian: What do you think of at least `throw`ing again the error?

Extensions/DebuggerTools/debuggertools.ts
```diff
@@ -13,6 +13,20 @@ namespace gdjs {
...
+       * @param message - The message to log.
+       * @param type - The type of log (info, warning or error).
+       * @param group - The group of messages it belongs to.
+       */
+      export const log = function (
+        message: string,
+        type: 'info' | 'warning' | 'error',
+        group: string
+      ) {
+        gdjs.log(group, message, type, false);
```
4ian, line 27: What do you think of dropping the last "internal" parameter?
Nothing is really internal or not in a sense, and this would make gdjs.log more lightweight to use (even with a default parameter).

arthuro555: The internal is used to differentiate logs made by the engine from the ones made by the user through the events sheet, to filter them out later. I thought that it might be useful when debugging something to be able to focus on your own logs without having to manually filter out the groups of the internal logs. Bouh also commented before
> I think we should also filter the messages coming from the engine and those added by the user.

So the idea cannot be that bad if we both had it?

newIDE/app/src/stories/index.js
arthuro555: Should I also rename `profilerOutput` to `profilerOutputsTestData` too then?

4ian:
Looks nice! :) 

On the UI side:
- I think we can put the checkboxes ("Show group name", "Show internal logs") inside a MiniToolbar. Like the text object or the bottom of the inspectors: 
  image
- For the visible/hidden groups, I think we can make things simpler by just using a Dialog and a bunch of checkboxes?
  image

Extensions/DebuggerTools/debuggertools.ts
4ian: Ok fair enough then!

arthuro555:
> For the visible/hidden groups, I think we can make things simpler by just using a Dialog and a bunch of checkboxes?

My thinking is that the list of groups can become quite lengthy for real games with a lot of debug output, and having it in two lists makes finding a specific group easier. Additionally, it makes mass selecting/unselecting groups easier. A dialogue is also blocking the view, the popover allows to see in real time how the filtering affects the readability of the list, and to make decisions on which groups to add/remove easier.

4ian:
It's true that it's blocking the view, but it can scroll and handle large list easily when you have 20+ groups? Also we're sure it's accessible and I think it's as easy to find a group (if we sort them alphabetically). We could even in the future add a search box if really needed to filter the list :)

Bouh:
- The idea of popup for filtering is interesting, need a bit more work for polish the UX and make some part cleaner.
Even if it's not a popup but in a dialog you can keep your new component and make it better like:
image

- Can be possible to add the orange colors on warning icons? Like it's done for warning hints.
- The list scrolls but does not focus on the latest logs.
- What do you think to add a timestamp?

image

4ian:
I might be alone for the dialog, but I think a list of checkbox is simple (you just check what you want to see, that's it) & as easy to use (one click on each group to select/unselect, vs one click on each group + one click to transfer to the other list) & can be extended (search, "select all" button) easily. 😇 And even accessible and small screens 😇

arthuro555:
> Can be possible to add the orange colors on warning icons? Like it's done for warning hints.

IIRC not with our current color scheme.

> The list scrolls but does not focus on the latest logs.

I am aware, I tried to make a checkbox for it to scroll automatically and to make new logs at the top instead of the bottom, but I couldn't get either to work with react-virtualized (and a non virtualized list is not viable).

> Even if it's not a popup but in a dialog you can keep your new component

I think the point of 4ian is that my component is too complex and should be just a normal list 😓 

> What do you think to add a timestamp?

🤔 Well why not, but I don't think the component that is used there support showing any more data

arthuro555:
Ok, I reworked the UI like discussed: 

https://user-images.githubusercontent.com/19349038/130515546-47e712fc-39b4-4c07-9c4c-6f8754a17ee6.mp4

I think this is ready for a new round of review. Not sure yet if it'll pass flow, if it doesn't I'll try to fix it later.

4ian:
Looks pretty good! I like the toolbar :) 

A few things I noted:

- The console is not resizing on the width, might be something to check on size measurer?
image
- Only one flow error! 👍
- I got some overlapping messages (with long messages), but I'll assume it might be linked to the non updating width. Can't reproduce.

arthuro555:
I added a logs manager for allowing to not redo the group's list on every render and handle automatically the logic to throttle/reduce rerenders. Now the performance is way better for me.

arthuro555:
🤔 Tried to add monospacing and selection, but there are 2 issues:

1. It does not look very good:

image


2. Selection does not do well with virtualized lists, as the DOM nodes are reused, unmounted, remounted, etc while scrolling:

aWiYVLftiC

4ian:
Great job on the LogsManager to batch the logs - it's way more performant now! 🎊

> It does not look very good:

Well not that terrible I would say. 
- I think this is also because there are huge spacing around each line - any chance we can reduce this? 
- Also maybe the default monospaced font is not great, we can search for a set of fonts supported on most systems that looks better.
- We could show the group name in brackets (and rename "Maximize/Minimize" to "Show groups").

> Selection does not do well with virtualized lists

Right this is indeed a bit problematic. Any chance we can use `overscanRowCount` and set it to something like 20 or 30? This will give more buffer before the selection is trashed.
The goal anyway is really to allow the copy of a single line that would show an error.

- In other notes, the width is still not updating properly when I resize the console (you might want to force a full re-render/trash the cache when the width is not the same as the width used for the last render?). Maybe use `useOnResize` to force an update if needed (I don't like using this, but with a virtual list it seems necessary)?
- Finally, I think we should keep scrolling as new logs appear. If we detect that the scroll position is at the bottom, we should update the scroll position (`scrollTop` probably?) to stay at the very bottom :) If too complicated, we can have a toolbar checkbox "Focus last messages" (but would be good to avoid this).

If the last thing is too tricky feel free not to do it and I can take a look.

Oxey405:
Can I do you think ?

Le mer. 30 juin 2021 à 21:55, Arthur Pacaud ***@***.***> a
écrit :

arthuro555:
Ok, some changes I made:

### Flow

Flow finally passes 😇 

### Style

I've reworked the styles a bit, here is what it looks like now:
image

I'm now pretty satisfied with the looks of it.

### Autoscroll

Since I couldn't get the autoscroll to work, I've removed it, and found a way to put the new logs on top without breaking the `CellMeasurerCache` (by using the keyMapper feature to make it use `array.length - index` instead of `index` as the item identifier). I did find a problem with it though that I am also unsure how to fix: If I scroll to the bottom and start adding logs, react-virtualized will do something weird and overlap multiple elements until I scroll the virtualized list.


## ToDo

Here is what is left to do as far as I am aware:

- [ ] Modify the throttling of logs as waiting 1 second to display any log is very annoying and a bad user experience
- [ ] Fix the weird visual bug when adding logs while looking at the bottom of the logs list that disappears when scrolling
- [ ] Optionally add a timestamp as a chip to the logs
- [ ] Code peer reviewing
- [ ] See if others like the looks of the logs and modify the styles it if they are not good

arthuro555:
>  See if others like the looks of the logs and modify the styles it if they are not good

After showing it on discord some time ago it seemed people are not quite fond of this design :/ I don't really know what to change to make it better.

>  Fix the weird visual bug when adding logs while looking at the bottom of the logs list that disappears when scrolling

I can't pinpoint the source of this issue. If you want to replicate it, log something every frame and try to scroll around in the logs.



I think this is ready for a new round of review?

Bouh:
> > See if others like the looks of the logs and modify the styles it if they are not good
> 
> After showing it on discord some time ago it seemed people are not quite fond of this design :/ I don't really know what to change to make it better.

https://github.com/4ian/GDevelop/pull/2770#issuecomment-905904043

4ian:
> Maybe padding on chips is a bit large.

Yep, I think for a console is important we are as tight as possible in terms of margins. What is important is the text, not the space (contrary to the rest of the interface where it makes more sense to have some margins as you need to allow users to browse fields).
I'm almost thinking we should ditch any material ui component for this console and have a pretty simple rendering.

4ian:
I like it! 👍

4ian:
It's all right for separators (nitpicking: timestamp we should round them at the nearest ms or micro second ;))

4ian:
I think this is almost ready to go! 

We might need to listen for some resize to re-render when the width is changed? See for example this after I resized the window: 
image

It's no a big deal so please don't pull your hair too much on this, but could be cool to fix so that this is polished.

Also:
* Timestamps => let's show them in seconds, with 6 decimal places to have microseconds (2.123456s)
* Minimize/Maximize => let's just label this "Show details"
* There is this React error, see if you can find it in the code:
  image

arthuro555:
> There is this React error, see if you can find it in the code:

That's because I am using chips as a render prop for the ListItem, where it expects only text. Not sure if I can fix that, as there is no prop to set the container it should be using. So it won't be removed without dropping the MUI List component, which would mean redoing most of the UI from the console from scratch.

arthuro555:
🤔 It seems that `React.Node` makes flow complain

PR #2781, opened 2021-07-03
Behavior composition

D8H:
The change request: https://trello.com/c/U8hR6L7l/392-allow-behaviors-to-be-configured-to-work-with-other-behaviors-behavior-composition

The related discussion: https://github.com/4ian/GDevelop/discussions/2762

There is still one issue:
* name conflicts could happen between required behavior passed as properties or function parameter (it may not make preview crash but one behavior will be used instead of the other)

As this is only on extension side, maybe I can just open an issue like "Make the GUI warn of name collision between behavior passed as properties and function parameter"?  (Actually there is not even check between parameter names)


newIDE/app/src/EventsBasedBehaviorEditor/EventsBasedBehaviorPropertiesEditor.js
```diff
@@ -267,6 +272,35 @@ export default class EventsBasedBehaviorPropertiesEditor extends React.Component
...
+                                  } else {
+                                    //TODO This doesn't work because extraInfo is constant.
+                                    // Maybe this is a hack and the behavior type must be set somewhere else?
+                                    // Should a method be added to replace a value?
+                                    console.log('newValue: ' + newValue);
+                                    property.getExtraInfo().clear();
+                                    property.addExtraInfo(newValue);
+                                    console.log(
+                                      'get: ' + property.getExtraInfo().at(0)
+                                    );
```
D8H, line 296: I used the property extra informations to store the required behavior type, but maybe that not a good idea because it seems that values can only add up.

D8H: I added a `clearExtraInfo` method but that may not be the good call.

newIDE/app/src/PropertiesEditor/PropertiesMapToSchema.js
```diff
@@ -114,6 +115,36 @@ export default (
...
+              ? name
+              : null
+          )
+          .filter(Boolean)
+          .map(value => ({ value, label: value }));
+      }
+      return {
+        name,
+        valueType: 'string',
+        getChoices: () => choices,
```
arthuro555, line 136: Not 100% sure but if that is a function I think you should rebuild the `choices` array each time it is run instead of building it when making the Schema and always returning it?

newIDE/app/src/EventsBasedBehaviorEditor/EventsBasedBehaviorPropertiesEditor.js
arthuro555: I think it is fine to add a way to modify the extra data, you could also add a `void setExtraData(size_t index, const gd::String& extraData);` method to have a cleaner way of doing what you do.

D8H: I was wondering if this should just not be constant instead of redirecting the whole vector interface.

newIDE/app/src/PropertiesEditor/PropertiesMapToSchema.js
D8H: I tried to add a behavior to the object and it appeared in the combobox.
Did you smell another potential issue?

4ian: My gut feeling would be indeed to load the choices lazily inside the callback - would ensure validity and freshness but not sure if there is a real problem if you don't do it.

4ian (commented):
That's awesome. Seems to confirm that going ahead with properties was the right choice.
Would be interesting to "stress test" this by making a behavior requiring an behavior that itself requires another.

Might need some UI updates to fix issues with multiple parameters/behaviors having the same name, but can be done later.

For now, would be good to think if we can create test ensuring this is robust. I expect this will open the road to a lot more of behaviors.

D8H:
> Would be interesting to "stress test" this by making a behavior requiring an behavior that itself requires another.

I thought about this too but actually this is not different that the basic case because the indirection is between the game and the extension directly though the property. There is no behavior in the middle per say.

newIDE/app/src/PropertiesEditor/PropertiesMapToSchema.js
```diff
@@ -114,6 +115,37 @@ export default (
...
+                .getAllBehaviorNames()
+                .toJSArray()
+                .map(name =>
+                  object.getBehavior(name).getTypeName() === behaviorType
+                    ? name
+                    : null
+                )
+                .filter(Boolean)
+                .map(value => ({ value, label: value }));
+        },
```
D8H, line 137: I still had to get the extraInfo outside because they are missing when it's refreshing. I wonder if there is an issue in the PropertiesEditor. I will investigate.

4ian: Right this might be because we're dealing with libGD.js C++ objects here, i.e: objects living in the WebAssembly memory. While `object` will remain valid for the whole life of the component using the properties (because if the object is deleted, it's the IDE responsibility to close anything related to the object), it's not necessarily the case for some objects returned by some methods (those marked with `[Value]` in `Bindings.idl` will get replaced at the next time the function is called).

This being said, here I would have assumed `property` would remain valid... Anyway, it's better now! And safer to save the behaviorType in a traditional JS string :)

4ian:
Thanks for the updates, let me do a real test of this before we go ahead :)

4ian:
Overall I think it's working great! 
Happy that there is little overhead with this solution when calling another behavior.

The major thing to solve now to ship this in a totally usable and enjoyable state would be to ensure we can't get the objects in a state where required behaviors are not fulfilled:

* Currently, if nothing is selected for the required behavior, then it's not found at runtime (so crash when an action using it is used).
* This can happen when editing an object, when adding the behavior. 
  * Probably we'll want to enforce the choice of a behavior if we can find one.
  * If we can't find one, we must flag the object as invalid and probably avoid to close the object editor.
* But this situation can also arise if we **add a required behavior** to a behavior. At this point, all objects using the behavior need to be updated.
  * We probably want, if we can, to automatically set the required behavior for objects where there is no ambiguity.
  * If an object has multiple behaviors that could fit, we probably want to ask the user rather than auto-selecting? Maybe by compiling the list of the names of the behaviors across all objects, and ask the user to choose one? Or if this can't be applied for all objects, ask to fix every object one by one?
  * If an object has *no* behavior that could fit in the required behavior, we probably want to suggest to the user to automatically add it?
* If we delete a required behavior property, then we should probably clean up the other objects.
* If we rename a required behavior, we should run a refactoring on the events of the behavior (to rename the required behavior) and in the behavior properties of every object (so that it's invisible to the user, everything works as before)

Every time the idea is to minimize/remove the risk of being in an "invalid" state. 
Most could be solved by implementing the logic in gd::WholeProjectRefactorer (a bit like when you change a parameter order, etc...).
But not sure if we'll be able to avoid having objects being temporary in an invalid state tough... which could be reported by browsing the project and reporting this in a Diagnostic Report (which we could use for events too).
This Diagnostic Report would be shown to the user if there is anything preventing the preview to run.

That's a lot, but I think it's important to nail these details before giving this in the hands of users? In particular, I think it's also important because requirements change during game creation, so being able to rely on GDevelop to update properly your objects when you add a new required behavior or rename it etc.. is important?

D8H:
This sounds interesting. I'll look into this after the navigation mesh path finding if no one already did.

4ian:
Just to signal that I still want to work on this later in August, this will be a super interesting addition :)

D8H:
> Just to signal that I still want to work on this later in August, this will be a super interesting addition :)

Did you start working on it? Can I help?

4ian:
Not started yet! Too much other reviews and stuff to do 😬😅

What could be useful would be too look at all points I listed in https://github.com/4ian/GDevelop/pull/2781#issuecomment-876792574 and for each check:
- Can we implement this as part of `WholeProjectRefactorer`? If yes, we can start it (will need to be implemented in C++ and have tests for it as this can be a bit tricky sometimes to check everything is working properly - and at the same time super important to not break the user game).
- If no (if we can't automate a change), should we forbid the change? What/how should we communicate this to the user?


Not sure what the best place for this is, maybe just continuing the discussion here or do a separate Google doc to be sure we have everything in mind?

But again that's a super interesting feature, if we do it right it will be both super powerful and super great for users that will be able to use very advanced behaviors based on other behaviors 🤩

Bouh:
> Would be interesting to "stress test" this by making a behavior requiring an behavior that itself requires another.

- The circular references are something to test too, how does the project handle this?

I knew some experimented users that this new possibility will excite!

D8H:
> > Would be interesting to "stress test" this by making a behavior requiring an behavior that itself requires another.
> 
>     * The circular references are something to test too, how does the project handle this?

Fuctions can already call themself without behavior composition. But, it can be spotted easily in the console.
```
Uncaught RangeError: Maximum call stack size exceeded
    at MyBehavior.gdjs.evtsExt__NewExtension__MyBehavior.MyBehavior.ActionA (gdjs-evtsext__newext…n__mybehavior.js:58)
    at Object.gdjs.evtsExt__NewExtension__MyBehavior.MyBehavior.ActionAContext.eventsList0 (gdjs-evtsext__newext…n__mybehavior.js:45)
    at MyBehavior.gdjs.evtsExt__NewExtension__MyBehavior.MyBehavior.ActionA (gdjs-evtsext__newext…__mybehavior.js:105)
    at Object.gdjs.evtsExt__NewExtension__MyBehavior.MyBehavior.ActionAContext.eventsList0 (gdjs-evtsext__newext…n__mybehavior.js:45)
    at MyBehavior.gdjs.evtsExt__NewExtension__MyBehavior.MyBehavior.ActionA (gdjs-evtsext__newext…__mybehavior.js:105)
    at Object.gdjs.evtsExt__NewExtension__MyBehavior.MyBehavior.ActionAContext.eventsList0 (gdjs-evtsext__newext…n__mybehavior.js:45)
    at MyBehavior.gdjs.evtsExt__NewExtension__MyBehavior.MyBehavior.ActionA (gdjs-evtsext__newext…__mybehavior.js:105)
    at Object.gdjs.evtsExt__NewExtension__MyBehavior.MyBehavior.ActionAContext.eventsList0 (gdjs-evtsext__newext…n__mybehavior.js:45)
    at MyBehavior.gdjs.evtsExt__NewExtension__MyBehavior.MyBehavior.ActionA (gdjs-evtsext__newext…__mybehavior.js:105)
    at Object.gdjs.evtsExt__NewExtension__MyBehavior.MyBehavior.ActionAContext.eventsList0 (gdjs-evtsext__newext…n__mybehavior.js:45)
```

If you are referring to behavior requirement cycles, there is actually not really depth. Because, it's the behavior properties of the object that define which behavior to use, not the behavior itself.
To make the parallel with component diagrams, we can see a behavior as a component with one provided interface and its required behaviors are like required interfaces. An object would be an encompassing component that connects its behaviors interfaces and expose the provided interfaces to the outside. Behaviors can't actually compose other behavior.

arthuro555:
Speaking of dependencies between functions, we need to rewrite events functions generation to handle those correctly. Currently, to handle using other events functions, there is a run of code generation with no actual code generation to make sure the necessary includes are added. This is a flawed solution though, as imagine I have those functions:
```
A -> calls B
B -> calls C
C -> calls some base action
```
The first generation round would get only part of the includes:
```
A -> No includes for B yet, so only includes for the generated code
B -> No includes for C yet, so only includes for the generated code
C -> Includes for the generated code and used actions
```
Then on the second round, right before finally generating the code:
```
A -> No includes for B yet, so only includes for the generated code
B -> Includes for the generated code and for C
C -> Includes for the generated code and used actions
```

The a function therefore has no includes for B while still calling it, it will crash! Instead, we should be generating them in one pass, but prepare the events sheets to create a dependency graph to generate them in the order they are including each other. In this example, we would see that A depends on B which depends on C, So we would build C first, followed by B (for which the necessary includes list has now just been generated) and finally A.

4ian:
> Currently, to handle using other events functions, there is a run of code generation with no actual code generation to make sure the necessary includes are added. This is a flawed solution though

Correct, it only works for "one level" of dependency and is not very elegant.
We should indeed identify the dependencies of each extension, and then build them in "topological order" (i.e: the one without dependencies, then the one depending one the one just built, then the ones depending on the ones just built, etc... until none are left).
(It's slightly independent from this though I think?) Also question, do you think we can use `gd::UsedExtensionsFinder` for this? We would feed it an extension and it would give us the dependencies. Then we can use this to construct the order of extensions building :)

arthuro555:
> Also question, do you think we can use gd::UsedExtensionsFinder for this?

No, while this would probably work for dependencies inter extensions, this would break private functions. We need to see for each events function what other events functions it uses. `gd::UsedExtensionsFinder` do would be a good base to make this new "`EventsFunctionsDependenciesAnalyzer`".

D8H:
I added a commit to handle insertion and removing of behaviors with required behaviors in the object editor.
* When a behavior is added, it adds the required behaviors on the object if necessary and fill the behavior properties.
  * If there is more than one behavior to choose, it just picks one as the user is in the object editor and can change it.
* When a behavior is removed, it also removes behaviors that depends on it (after asking user confirmation).

The 2 methods added to WholeProjectRefactorer should be useful to write the refactoring but maybe they should be in another class because they are local to one object.
The refactoring methods will probably needs to return a list of issues, it will probably be the same kind of objects as in the Diagnostic Report.

arthuro555:
I'm not a bog fan of this, I don't think that modifying an extension should directly modify your project. When deleting a required behavior, that only needs that this custom behavior doesn't require it anymore to function, not that we don't need the behavior at all! Deleting it will in most cases just break your project. While adding the required behavior is more understandable as it is required for the game to function, I would leave the user the choice of what to do with a UI like this:

image

D8H:
> 
> 
> I'm not a bog fan of this, I don't think that modifying an extension should directly modify your project. When deleting a required behavior, that only needs that this custom behavior doesn't require it anymore to function, not that we don't need the behavior at all! Deleting it will in most cases just break your project. While adding the required behavior is more understandable as it is required for the game to function, I would leave the user the choice of what to do with a UI like this

Yes, this is the idea. (I don't know if you are replying to my last comment, but it is not about extension modification, but adding behavior to an object)

I think the view could even be a Diagnostic Report with a custom message and a filtered problem list. Because the Diagnostic Report should allow to batch error resolution and let the user choose between resolution.

D8H:
## Sum-up of the cases
* Add a behavior to an object
  * (if one matching behavior) fill the property with it
  * (if 2+ matching behaviors) fill the property with the 1st found
  * Users can still change it, they are in the view to edit properties
* Remove a behavior from an object
  * Ask confirmation to remove dependent behaviors too
  * Users can cancel and change the property to avoid this
* Add a required behavior property
  * Issue to ask the user: unfill required behavior properties
* Delete a required behavior property
  * No issue (I think that's what Arthuro pointed out)
* Rename a required behavior
  * Silent refactoring to fix the extraInfo of the required behavior properties
* Rename a required behavior property
  * It's refactored like any property

For unfill required behavior properties, we can show the Diagnostic Report to let the user choose between:
- (if one matching behavior) use the existing behavior
- (if 2+ matching behaviors) select a behavior to use
- (if no matching behavior) add the behavior
- go to view to manually fix it

## What's next
* Add the rename refactor
* Discuss the "delete a required behavior property" case
* Implement the Diagnostic Report
  * It would be probably better to use another PR for this. Do you think it should block this PR?

Bouh:
Gosh, not a new message box. Please consider what is following.

- You want to add a message box because we have to alert the user that the dependencies has to be deleted, but there is no visual link between them.
How do we know which depend of another?
Instead of a message box and no visual relation between behaviours, I suggest making it cleaner. There is no more message box between the visual link involve that both are linked.

- 
image

People will asking "What is the reason?", "Why this isn't properly supported", we must avoid this semi working things visible to the final user, it sound like a undevelopped feature or a bad design. It doesn't inspire any trust.

D8H:
> Gosh, not a new message box. Please consider what is following.

Do you mean the message box from this comment: https://github.com/4ian/GDevelop/pull/2781#issuecomment-898033643 ?
It already exists, but yes, users may not read it and be surprised by the result. The idea was that the end-user can't make invalid behavior links.
Do you suggest to only empty the properties instead and let the Diagnostic Report notify issues at preview?

>     * You want to add a message box because we have to alert the user that the dependencies has to be deleted, but there is no visual link between them.
>       How do we know which depend of another?

There is a screenshot on this comment: https://github.com/4ian/GDevelop/pull/2781#issuecomment-898033643

>       Instead of a message box and no visual relation between behaviours, I suggest making it cleaner. There is no more message box between the visual link involve that both are linked.

I don't understand. What do you mean?

> People will asking "What is the reason?", "Why this isn't properly supported", we must avoid this semi working things visible to the final user, it sound like a undevelopped feature or a bad design. It doesn't inspire any trust.

Yes, it's far from ideal.
Maybe, leaving the Diagnostic Report handle it will allow to give an explanation of the issue and the possible fixes. What is your view on it?

arthuro555:
I'm not sure I understood the points here properly, but I want to point out that I often saw on the forum criticism for taking too many important decisions for the users, and I don't agree with "minimize questions/dialogs", we should I think if the wished behavior is not 100% clear to let the user chose what they want to do instead of deciding what's best by ourselves and force that onto people.

Bouh:
I mean a dependency should be visual and understood in a second, like this:
image

Now the deletion of the "parent" behavior involve that the second required inside will be also delete. This way we can totally remove this message box:
image

And it more pleasant and to understand which behaviors are dependencies.

D8H:
> I mean a dependency should be visual and understood in a second, like this:

There is no ownership. For instance, a PlatformerObject can be shared by CoyoteTime and MultiJump behaviors.

> Now the deletion of the "parent" behavior involve that the second required inside will be also delete. This way we can totally remove this message box:

It's the other way around, if PlatformerObject is deleted, so is CoyoteTime and MultiJump because they can't work alone.

4ian:
Thanks all for thinking about this :)
And thanks for the recap of all situations to handle @D8H :) 
In general, the idea is:
- avoid manual work for the user when possible (ex: renaming a property is automatic)
- if not possible, give a choice (ex: deleting a layer, what to do of instances).

Bad counter examples are:
- not filling a parameter of an a action (or condition): currently we don't make the error visible in the events sheet, causing confusion. My idea was to surface this in a "diagnostic manager", along with other stuff where we can't have a "always valid state".
- abuse of message box, which would be solved with either a dedicated dialog or the "diagnostic manager".

Note that I'm not "happy" to create a diagnostic manager. Ideally we could avoid it entirely. But there is a set of things like here where we maybe can't avoid this.

D8H:
In the current state of the implementation, I think that the only remaining case that can end up to invalid required behavior values is when a required behavior is added or its type is changed with the extension editor.

What the implementation already do is to find a behavior on the object and use it, but it leaves invalid values if the object doesn't have the right behavior.

I see 2 solutions
* Add the needed behavior on the objects (silently).
  * Pro: the Diagnostic Report is not needed (at all).
  * Con: the user may not like to see behaviors added without its consent and never removed.
* Fill the property when the user adds the missing behaviors and rely on the Diagnostic Report to protect the preview.
  * Pro/Con are the exact opposite.

In my opinion the 1st version is better because required behaviors won't change every day and mistake are unlikely so we could assume that adding the behavior on the objects will be the right call 99.9% of the time.

(I will start to write units for the other cases, waiting for your views)

4ian:
> Add the needed behavior on the objects (silently).

We could indeed do that. 
You said this is in the case of "when a required behavior is added or its type is changed with the extension editor", which is true.
What happen in the case where I'm adding a behavior to an object, that has a required behavior property, and there is no matching behavior for this required behavior property?
Should we also add it?

D8H:
> What happen in the case where I'm adding a behavior to an object, that has a required behavior property, and there is no matching behavior for this required behavior property?
> Should we also add it?

I forgot to mention this case in the sum-up, it automatically add the required behaviors (transitively). I will edit the comment.

newIDE/app/src/EventsBasedBehaviorEditor/EventsBasedBehaviorEditorDialog.js
```diff
@@ -26,9 +28,14 @@ export default class EventsBasedBehaviorEditorDialog extends React.Component<
       project,
     } = this.props;
 
+    const onApplyAndFixIssues = () => {
+      onApply();
+      fillRequiredBehaviorProperties(this.props.project);
+    };
```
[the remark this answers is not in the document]
4ian: > What is updating the property list of behaviors in objects to follow the changes done on the extension?

I would say: nothing.

If you add a new property, it's not added in the "behavior content" (aka: behavior data) of objects having this behavior.
This is not a problem for events or the generated behavior because a default value is used if the behavior data is not containing the key:
https://github.com/4ian/GDevelop/blob/287ebd4244ecb9ba272b648dcf75a7e029d190ed/GDJS/GDJS/Events/CodeGeneration/BehaviorCodeGenerator.cpp#L169-L175

How is this affecting you? It means that when the dialog close, you must fill the required behavior properties in the whole project, but that should not be a problem?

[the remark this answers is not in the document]
D8H: > How is this affecting you? It means that when the dialog close, you must fill the required behavior properties in the whole project, but that should not be a problem?

`WholeProjectRefactorer::FindInvalidRequiredBehaviorProperties` relies on the `behaviorContent` to find required behaviors that are not filled correctly, but these property keys doesn't exist yet on objects so the issues fly under the radar. I could use `propertyDescriptors` to check this case but it feels almost like an encapsulation breach to do so. 

> > What is updating the property list of behaviors in objects to follow the changes done on the extension?
>
> I would say: nothing.

What is really strange is that the issues are spotted the next time I close the behavior settings dialog. Something must have the side effect of refreshing the properties. Otherwise, I guess they would not appear on the object editor either.

Core/GDCore/IDE/WholeProjectRefactorer.cpp
```diff
@@ -516,6 +525,172 @@ void WholeProjectRefactorer::RenameBehaviorProperty(
...
+              const gd::String& requiredBehaviorType = extraInfo.at(0);
+
+              if (requiredBehaviorName == "" || object->GetBehavior(requiredBehaviorName).GetTypeName() != requiredBehaviorType) {
+                auto problem = UnfilledRequiredBehaviorPropertyProblem(project, *object.get(), behaviorContent, propertyName, requiredBehaviorType);
+                invalidRequiredBehaviorProperties.push_back(problem);
+              }
+            }
+          }
+        }
+      };
```
D8H, line 681: It's this method.

Core/GDCore/IDE/WholeProjectRefactorer.cpp
```diff
@@ -516,6 +525,172 @@ void WholeProjectRefactorer::RenameBehaviorProperty(
...
+  // Find in global objects
+  findInvalidRequiredBehaviorPropertiesInObjects(project.GetObjects());
+
+  // Find in layout objects and layout behavior shared data.
+  for (std::size_t i = 0; i < project.GetLayoutsCount(); ++i) {
+    gd::Layout& layout = project.GetLayout(i);
+    findInvalidRequiredBehaviorPropertiesInObjects(layout.GetObjects());
+  }
+  return invalidRequiredBehaviorProperties;
+}
```
D8H, line 692: It's this method.

newIDE/app/src/EventsBasedBehaviorEditor/EventsBasedBehaviorEditorDialog.js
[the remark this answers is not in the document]
4ian: I see, this might be because you call GetProperties to get all the properties (and you should use the property descriptor to get the type of the property, to check if it's "behavior"), but this is only ready after regenerating the extension... Which will be done later. I'll think about that.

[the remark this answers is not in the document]
4ian: I think this is what you're searching for the "thing that does side effect when properties are changed": https://github.com/4ian/GDevelop/blob/916938b4a8df88a060da95019b4921d615f15205/newIDE/app/src/MainFrame/EditorContainers/EventsFunctionsExtensionEditorContainer.js#L63

Not sure yet what's the best way to ensure this is run before refactoring (and if we should even wait for this or not)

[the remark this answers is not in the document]
D8H: Ok, I attempted to force a refresh on _editBehavior to see if it was it, but I probably didn't do the right thing. I will retry.

EventsBasedBehaviorEditorDialog:
```
    const onApplyAndFixIssues = () => {
      // TODO It only works when the dialog is opened and closed again.
      onApply();
      if (fillRequiredBehaviorProperties(this.props.project)) {
        onApply();
      }
    };

    return (
      <Dialog
        onApply={onApplyAndFixIssues}
```

EventsFunctionsExtensionEditor/index:
```
          <EventsBasedBehaviorEditorDialog
            project={project}
            eventsFunctionsExtension={eventsFunctionsExtension}
            eventsBasedBehavior={editedEventsBasedBehavior}
            onApply={() => {
              if (this.props.unsavedChanges)
                this.props.unsavedChanges.triggerUnsavedChanges();
              this._editBehavior(null);
            }}
```

```
  _editBehavior = (editedEventsBasedBehavior: ?gdEventsBasedBehavior) => {
    this.setState(
      state => {
        // If we're closing the properties of a behavior, ensure parameters
        // are up-to-date in all event functions of the behavior (the object
        // type might have changed).
        if (state.editedEventsBasedBehavior && !editedEventsBasedBehavior) {
          gd.WholeProjectRefactorer.ensureBehaviorEventsFunctionsProperParameters(
            this.props.eventsFunctionsExtension,
            state.editedEventsBasedBehavior
          );
        }

        return {
          editedEventsBasedBehavior,
        };
      },
      () => {
        if (!editedEventsBasedBehavior) {
          // If we're closing the properties of a behavior, notify parent
          // that a behavior was edited (to trigger reload of extensions)
          if (this.props.onBehaviorEdited) {
            this.props.onBehaviorEdited();
          }

          // Reload the selected events function, if any, as the behavior was
          // changed so objects containers need to be re-created. Notably, the
          // type of the object that is handled by the behavior may have changed.
          if (this.state.selectedEventsFunction) {
            this._loadEventsFunctionFrom(
              this.props.project,
              this.state.selectedEventsFunction
            );
          }
        }
      }
    );
```

EventsFunctionsExtensionEditorContainer:
```
        <EventsFunctionsExtensionEditor
          key={eventsFunctionsExtension.ptr}
          project={project}
          eventsFunctionsExtension={eventsFunctionsExtension}
          setToolbar={this.props.setToolbar}
          resourceSources={this.props.resourceSources}
          onChooseResource={this.props.onChooseResource}
          resourceExternalEditors={this.props.resourceExternalEditors}
          openInstructionOrExpression={this.props.openInstructionOrExpression}
          onCreateEventsFunction={this.props.onCreateEventsFunction}
          initiallyFocusedFunctionName={initiallyFocusedFunctionName}
          initiallyFocusedBehaviorName={initiallyFocusedBehaviorName}
          onBehaviorEdited={this._onBehaviorEdited}
```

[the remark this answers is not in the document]
D8H: `loadProjectEventsFunctionsExtensions` from `EventsFunctionsExtensionsLoader` is called asynchonously and happen after `fillRequiredBehaviorProperties`. That explains why it only woks the 2nd time the dialog is closed, but it seems to need a big change to run `fillRequiredBehaviorProperties` after it.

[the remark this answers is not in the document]
4ian: > That explains why it only works the 2nd time the dialog is closed, but it seems to need a big change to run fillRequiredBehaviorProperties after it.

Correct that's a pretty big change. Extensions are reloaded/code generated "later", when needed in theory.

> But maybe it would be easier and cleaner to make the property keys always consistent with the behavior definition?

Can you expand on what you mean exactly by that ("property keys")?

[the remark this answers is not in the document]
4ian: I wonder if, still, the best solution is to force the regeneration (re-declaration precisely) of extensions and wait synchronously for it to be done (with a Promise) before doing the refactoring of required properties... 🤔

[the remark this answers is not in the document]
D8H: > Can you expand on what you mean exactly by that ("property keys")? I guess you mean `behavior.GetProperties`?

Yes, what I meant is that `GetProperties` uses the `behaviorContent.GetContent()` to build the key/value list, but it could use `eventsBasedBehavior.getPropertyDescriptors()` or an equivalent from the base class to build the keys. This would avoid inconsistencies.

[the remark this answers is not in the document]
4ian: I believe we should run the refactoring with the proper declared extension and behavior metadata - no point trying to do something before that because the metadata would not be correct and this could create inconsistencies in the future. I *think* therefore we can do something like this:

- Change `onBehaviorEdited` so that it returns a promise
- In `_onBehaviorEdited`, change it so that it returns the result of `onLoadEventsFunctionsExtensions`
- In `MainFrame`, change `onLoadEventsFunctionsExtensions` so that it returns the result of `loadProjectEventsFunctionsExtensions` - which is a promise that resolves when extensions are reloaded and code generated. 
  - Note that we don't care about code generation, even if we'll have to wait for it (could be improved in the future)
- move `fillRequiredBehaviorProperties` to `EventsFunctionsExtensionEditor/index.js` (because I believe this is something fairly high level that should not be buried too much).
- Call it when `onBehaviorEdited` resolves or reject (it can only reject if extensions code generation failed, like on the web-app, but **even** in this case the metadata of extensions are valid and should be used for the refactoring).

With this, we should run the refactoring as soon as extensions are ready.
In the future, check if it works (and if this does not create too much lag?). In the future would be great to show an indicator to show that these things are done in the background (and to really do that in a background "worker") (but that's for another day).

If this works, I believe we'll be in a good state for this?

4ian:
I'll take a look at the failing tests :)

4ian:
I've reworked the failing C++ tests. The main stuff is that I'm not relying anymore on events based behavior because:
- code generation + declaration in the gd::Platform is not done, so it's useless to define them. Instead, I created behaviors in DummyPlatform.cpp who have required behavior properties.
- this kind of refactoring is mostly useful for events based behaviors, but could be used in the future for "classic" behaviors from C++/JS defined extensions :) 

I also made the code more robust at a few places to avoid potential crashes if something was off.

I've tested in the editor and overall we're getting there! Two things:
- See my comment for ensuring extensions are regenerated before we apply the "refactoring" to fix the behaviors. 
- I also tried to rename a required behavior property, and got a problem
  - For example, I renamed the required behavior property "DraggableBehavior" to "DraggableBehavio" and there is now the error:
  `Invalid required behavior properties value for: PlayerHitBox.MyBehavior.DraggableBehavio a DraggableBehavior::Draggable is expected. Possible values are: Draggable`
  Not sure if it's related to the same problem of having to wait for extensions to be reloaded?
  
As we're arriving I hope at the end of this, if you prefer me to take over this I can do it. I'll try to test as much as possible to ensure we've not left a corner case somewhere.

D8H:
I only manually tested the case where a new required behavior property is added an object that doesn't have any. The missing behavior was added on the object and the property filled.

Do you think we can do integration tests on the GUI to write the cases listed in the sum-up?

newIDE/app/src/MainFrame/EditorContainers/EventsFunctionsExtensionEditorContainer.js
```diff
@@ -56,15 +56,15 @@ export class EventsFunctionsExtensionEditorContainer extends React.Component<Ren
     // reload/regeneration of extensions, as changes can have
     // been made inside before it's closed.
     if (this.props.isActive) {
-      this.props.onLoadEventsFunctionsExtensions();
+      return this.props.onLoadEventsFunctionsExtensions();
```
4ian, line 59: You probably don't want to return here - componentWillUnmount will ignore it and this gives the impression the return value if somewhat used.

newIDE/app/src/EventsFunctionsExtensionEditor/index.js
```diff
@@ -578,14 +580,70 @@ export default class EventsFunctionsExtensionEditor extends React.Component<
...
+
   _openFreeFunctionsListEditor = () => {
     if (this._editorNavigator)
       this._editorNavigator.openEditor('free-functions-list');
   };
 
   _openBehaviorsListEditor = () => {
     if (this._editorNavigator)
-      this._editorNavigator.openEditor('behaviors-list');
+      return this._editorNavigator.openEditor('behaviors-list');
```
4ian, line 646: Useless return?

D8H: Yes, sorry about the useless returns I forgot to remove them after putting it on the right method.

4ian:
> Do you think we can do integration tests on the GUI to write the cases listed in the sum-up?

That seems a bit hard - I try as much as possible to have all the logic inside the Core and tested here - and rely on the UI to call the proper refactoring functions when needed. 

I've pushed a small "playground" project to help checking which cases are not yet covered. 
It seems that the fix to wait for extensions to be reloaded before filling the behaviors properly fixed the problem of "renaming a required behavior property is not updating the properties inside the objects" (because it's doing the same thing as when adding a new required behavior: filling properties).

A few more things that I've discovered:
* I've noticed that adding a new required behavior is not properly updating the events (you can't find the behavior action/conditions in a behavior function) until you navigate out of a function and back to it again.
* Similar problem if you renamed a required behavior property:
  - The object using it are properly "refilled".
  - But if the required behavior is used inside a function, it's not renamed.
  - And you have to navigate out of a function and back to it again to be able to properly use the required behavior.
  
I'm looking now at these! :) So if I do a sum up like you did:

## Sum-up of the cases

* Add a behavior to an object
  * (if one matching behavior) silently fill the property with it **- works ok**
  * (if 2+ matching behaviors) silently fill the property with the 1st found **- works ok**
    * Users can still change it, they are in the view to edit properties **- that's fine**
  * (if there is no matching behavior) silently add the required behavior (transitively) **- works ok**
  * **Bug**: the events of the currently edited behavior function are not updated with the newly added required behavior (you need to navigate out and back to the function). 
* Remove a behavior from an object
  * Ask confirmation to remove dependent behaviors too **- works ok**
  * Users can cancel and change the property to avoid this **- that's fine**
* Add a required behavior property
  * (if one matching behavior) use the existing behavior **- works ok**
  * (if 2+ matching behaviors) silently fill the property with the 1st found **- works ok**
  * (if there is no matching behavior) silently add the required behavior (transitively) **- to confirm that it works transitively?**
* Delete a required behavior property
  * Nothing to do (even if the property stays in the JSON)  **- that's fine**
* Rename a behavior that is also used as a required behavior
  * Silent refactoring to fix the extraInfo of the required behavior properties  **- works ok**
* Rename a required behavior property
  * **Bug**: the events of the currently edited behavior function are not updated with the newly renamed required behavior (you need to navigate out and back to the function). 
  * **Bug**: any event using the required behavior inside the behavior functions are not refactored.

4ian:
I've pushed the proper refactoring of a renamed "required behavior" property (btw the new class `EventsBehaviorRenamer` could be used in the future to allow renaming behaviors in objects, even if it's recommended to keep the default names).

I believe with this we cover all cases related to these required behavior properties, always keeping them valid, and smoothly handling any removal/addition/renaming.

Give it a go and let me know if you find anything that looks odd. I'll make a new review pass (and will probably fix things as I go) as it's a large and foundational feature for the future - better be sure it's bulletproof :)

D8H:
> Give it a go and let me know if you find anything that looks odd. I'll make a new review pass (and will probably fix things as I go) as it's a large and foundational feature for the future - better be sure it's bulletproof :)

I read your commits and play a bit with my test example, it sounds good to me.

4ian:
Alright, I ported the code to fill the required behaviors to the Core (as it's more something that would go in the WholeProjectRefactorer and can be tested easily this way next to the related methods).
Almost ready to go! :)

4ian:
And this is merged! It has been a long job but very glad we're having this in! 
I hope we've nailed most of the use cases. I'll release a new version in a few days and we can start seeing extensions powered by this :)

4ian:
Obviously a big big thank you for you work on this @D8H! 🎉 This is unlocking very exciting new features, and it was an awesome work you did on this to first implement it and then cover as much use cases as possible to ensure this can be used safely. 👍

arthuro555:
🥳 Woohoo! Can't wait to use this feature!

PR #2788, opened 2021-07-05
[Stuck] Cast object condition

arthuro555:
Adds a condition to check for an object type. If it is true, GDevelop will consider the object in the the following conditions, actions and subevents as of that type. It is useful when the type of the object is not known, like in custom behaviors that accept any objects or objects groups with objects of multiple types. 

Discussion: https://github.com/4ian/GDevelop/discussions/2643
Test project: 
[test.zip](https://github.com/4ian/GDevelop/files/6764904/test.zip)

TODO:
- [X] Create the condition with custom code generator
- [X] Make the CodeGenerationContext able to override an object type
- [x] Add a ObjectType parameter
- [ ] Make the IDE aware of the type change for usages after that condition

### DIFF
--- a/Core/GDCore/Events/CodeGeneration/EventsCodeGenerationContext.cpp
+++ b/Core/GDCore/Events/CodeGeneration/EventsCodeGenerationContext.cpp
@@ -4,17 +4,18 @@
  * reserved. This project is released under the MIT License.
  */
 #include "GDCore/Events/CodeGeneration/EventsCodeGenerationContext.h"
-#include <set>
 #include "GDCore/CommonTools.h"
 #include "GDCore/Events/CodeGeneration/EventsCodeGenerator.h"
 #include "GDCore/Events/Tools/EventsCodeNameMangler.h"
+#include "GDCore/Project/Layout.h"
+#include <set>
 
 using namespace std;
 
 namespace gd {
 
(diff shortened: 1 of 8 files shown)


arthuro555:
I am unsure how to cleanly do the list item on the todo list. What I am currently experimenting with is in the event renderer doing 
```js
mapVector(event.getAllConditionsVectors(), instrList =>
  mapFor(0, instrList.size(), i => {
    const instr = instrList.get(i);
    if (instr.getType() === 'CastTo') {
      // (scope was passed as prop) make a copy of the scope to not modify it on parent/sibling events
      scope = Object.assign({}, scope);
      scope.castedObjects[instr.getParameter(0)] = instr.getParameter(1);
    }
  })
);
```
to pass to the current and next events a scope with the castedObjects map, but that feels a bit hacky and looks like it could become very expensive perfomance wise. Any idea of how I could try to that in a more clean way?

arthuro555:
I couldn't manage to find a good nor working way to make the IDE aware of the cast :/ Any ideas?

arthuro555:
Abandoning this as I cannot find a way to finish this that wouldn't involve making a large refactor, which has in my opinion more risks than it is worth to add such a condition, especially since my PC cannot run flow for helping with the refactoring.

PR #2790, opened 2021-07-07
Fix: instances editor no longer give unnecessary floating positions

D8H:
Moving instances and resizing one instance no longer give floating positions or dimensions on 0° rotated instances.

For 90° rotated instances, I still observe half pixels when width and height parity is different. It's because the center point is not pixel aligned so the rotation gives an AABB with bounds on half pixels.
But this still make the bug a lot less critical.

### DIFF
--- a/newIDE/app/src/InstancesEditor/InstancesMover.js
+++ b/newIDE/app/src/InstancesEditor/InstancesMover.js
@@ -82,23 +82,21 @@ export default class InstancesMover {
     this.totalDeltaX += deltaX;
     this.totalDeltaY += deltaY;
 
-    let roundedTotalDeltaX;
-    let roundedTotalDeltaY;
-    if (this.options.snap && this.options.grid && !noGridSnap) {
-      // It will magnet the corner nearest to the grabbing position
-      const initialSelectionAABB = this._getOrCreateSelectionAABB(instances);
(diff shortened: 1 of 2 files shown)


newIDE/app/src/InstancesEditor/InstancesMover.js
```diff
@@ -107,12 +105,15 @@ export default class InstancesMover {
...
         this.options.gridType
       );
-      roundedTotalDeltaX = magnetPosition[0] - initialMagnetX;
-      roundedTotalDeltaY = magnetPosition[1] - initialMagnetY;
     } else {
-      roundedTotalDeltaX = this.totalDeltaX;
-      roundedTotalDeltaY = this.totalDeltaY;
+      // Without a grid, it's rounded to the nearest pixel
+      // the size might not be round
+      // so the magnet corner is still relevent
```
4ian, line 111: Nitpicking: Name what is "it", the size of what, and be sure to delimitate sentences:

// Without a grid, the position is still rounded to the nearest pixel.
// The size of the instance (or selection of instances) might not be round,
// so the magnet corner is still relevant.

4ian (approved):
Added a nitpicking, otherwise looks good to go! Thanks :)

D8H:
I updated the description, the remaining issue is almost legit.

4ian:
Thanks 👍

PR #2798, opened 2021-07-11
Display behavior icons in the behaviors editor

4ian:
image

Also various clean up and improvements


Bouh:
Not sure of the interest of Behavior text between the icon and the behavior's name.
We know already these are behaviors, the user clicks on the **Behavior tab** ;)

4ian:
That's fair enough! It's true that it's useless here given the context. I've removed it in 8c955cf77a0db6196d40339db616c84eec4f9df6 :)

PR #2806, opened 2021-07-14
Remove examples workflow

Midhil457:
- Remove workflow that closes that closes issues with no project included

As the examples now have a different repo and now there is no template for submitting examples, this might not come in use.

I found this accidently and tried removing it myself. Forgive me if I did something wrong

:)


4ian (approved):
Sounds good 👍

PR #2810, opened 2021-07-14
Add UID parameter

arthuro555:
Adds a UID parameter, which generates for each parameter generation a different UID. This allows to identify one specific usage of a function, like in the trigger once. This can allow to make powerful extensions, like "Wait X seconds", "Conditions executed X times in X amount of times" (for double/triple/whatever click or key press for example)...

Example:

https://user-images.githubusercontent.com/19349038/125647400-a033bdfd-0fe6-4403-a1e9-5e0dc25a371b.mp4

[WaitX.zip](https://github.com/4ian/GDevelop/files/6817147/WaitX.zip)

### DIFF
--- a/Core/GDCore/Events/CodeGeneration/EventsCodeGenerator.cpp
+++ b/Core/GDCore/Events/CodeGeneration/EventsCodeGenerator.cpp
@@ -17,7 +17,6 @@
 #include "GDCore/Project/Layout.h"
 #include "GDCore/Project/ObjectsContainer.h"
 #include "GDCore/Project/Project.h"
-#include "GDCore/Tools/UUID/UUID.h"
 
 using namespace std;
 
@@ -330,7 +329,7 @@ gd::String EventsCodeGenerator::GenerateConditionCode(
 
         // Prepare arguments and generate the condition whole code
         vector<gd::String> arguments = GenerateParametersCodes(
-            condition.GetParameters(), instrInfos.parameters, context);
+            condition.GetParameters(), instrInfos.parameters, context, [this, &condition](){return gd::String::From(GenerateSingleUsageUniqueIdFor(&condition));});
         conditionCode += GenerateObjectCondition(realObjects[i],
                                                  objInfo,
                                                  arguments,
@@ -361,7 +360,7 @@ gd::String EventsCodeGenerator::GenerateConditionCode(
 
         // Prepare arguments and generate the whole condition code
         vector<gd::String> arguments = GenerateParametersCodes(
-            condition.GetParameters(), instrInfos.parameters, context);
+            condition.GetParameters(), instrInfos.parameters, context, [this, &condition](){return gd::String::From(GenerateSingleUsageUniqueIdFor(&condition));});
         conditionCode += GenerateBehaviorCondition(
             realObjects[i],
             condition.GetParameter(1).GetPlainString(),
@@ -384,6 +383,7 @@ gd::String EventsCodeGenerator::GenerateConditionCode(
         GenerateParametersCodes(condition.GetParameters(),
                                 instrInfos.parameters,
                                 context,
+                                [this, &condition](){return gd::String::From(GenerateSingleUsageUniqueIdFor(&condition));},
                                 &supplementaryParametersTypes);
 
     conditionCode += GenerateFreeCondition(
@@ -499,7 +499,7 @@ gd::String EventsCodeGenerator::GenerateActionCode(
 
         // Prepare arguments and generate the whole action code
         vector<gd::String> arguments = GenerateParametersCodes(
-            action.GetParameters(), instrInfos.parameters, context);
+            action.GetParameters(), instrInfos.parameters, context, [this, &action](){return gd::String::From(GenerateSingleUsageUniqueIdFor(&action));});
         actionCode += GenerateObjectAction(
             realObjects[i], objInfo, arguments, instrInfos, context);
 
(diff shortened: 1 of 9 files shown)


Core/GDCore/Events/CodeGeneration/EventsCodeGenerator.cpp
```diff
@@ -623,6 +624,9 @@ gd::String EventsCodeGenerator::GenerateParameterCodes(
   // Code only parameter type
   else if (metadata.type == "inlineCode") {
     argOutput += metadata.supplementaryInformation;
+  } else if(metadata.type == "uid") {
+    // GenerateSingleUsageUniqueIdFor would be better, but wouldn't work with expressions.
+    argOutput += ConvertToStringExplicit(gd::UUID::MakeUuid4());
```
arthuro555, line 629: Not a big fan of this as there is a chance of collisions even if little, and it will break hot reload. I cannot use the better `GenerateSingleUsageUniqueIdFor` though, as it only accepts an instruction. As far as I saw, it doesn't actually care about  the type of the parameter only about the pointer value, should I change it to accept a void pointer so that I can use an expression to identify it too?

newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
```diff
@@ -462,6 +469,10 @@ export default class EventsFunctionParametersEditor extends React.Component<
                                   value="objectAnimationName"
                                   primaryText={t`Object animation (text)`}
                                 />
+                                <SelectOption
+                                  value="uid"
+                                  primaryText={t`Unique event ID (text)`}
```
4ian, line 474: Is this an event unique id, a condition unique id, a parameter unique id?

Core/GDCore/Events/CodeGeneration/EventsCodeGenerator.cpp
4ian: What about another overload of this function?

newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
arthuro555: Well it technically is meant to be a condition/action/expression unique ID, but the GDevelop terminology doesn't have a word for that.

Core/GDCore/Events/CodeGeneration/EventsCodeGenerator.cpp
arthuro555: That means we would have to overload the GenerateParameterCode and GenerateParametersCode with the exact same implementation just different types :/

arthuro555: Wouldn't that be too much bloat and confusing?

4ian: Depends, there is a few things to consider:
- is the pointer to the expression stable across exports?
- if we move in the events editor a condition to another event, is this pointer still stable?

If yes, we could use this pointer. Otherwise, we could maybe use a mix of the pointer of the instruction + something like the parameter index.

4ian: (also, the reason for making something not using void* is to explicitly generate these questions and conversation - do we really need to do that? Is there something else to uniquely identify something that is not using the memory pointer?)

arthuro555: > is the pointer to the expression stable across exports?

Oh right, the expression pointer probably isn't stable. Not quite sure what I should do here.

newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
Bouh: `Unique ID` or `UUID` have more sense and are common, although misleading with potential `UUID` from an instances.
 All people talking about UUID make reference to UUID from C2/3, for not losing everyone it's fine to differentiate things, although technically all the words are valid.

 I would prefer `Generated ID`

arthuro555: No, an identifier is not unique by definition, multiple things can be identified by an identifier, and an identifier can be duplicated/reused unless it is unique. [See the distinction between identifier and unique identifier on wikipedia](https://en.wikipedia.org/wiki/Identifier). In my opinion, `Generated identifier` is too vague (when is it generated, is it unique, what does it identify) and doesn't tell anything about what it is/what it does/what it is used for. It's yet another piece of terminology to learn. Something like "Unique event identifier" or "Unique instruction identifier" tells you directly "Aha, this identifies the event/instruction it is used in!".

Core/GDCore/Events/CodeGeneration/EventsCodeGenerator.cpp
arthuro555: I digged a bit more and couldn't find any object (or at least easy to access ones) that would work with expressions. Should I just disallow that parameter on expressions?

4ian: Maybe it's just simpler to only allow this on instructions yeah!

PR #2814, opened 2021-07-16
Make the loading screen customizable

4ian:
* Allow to choose the style of the GDevelop logo (light, dark, colored or plain)
* Allow to choose the background color and an optional background image
* Allow to choose the size of the progress bar
* Allow to choose the duration of the fade in animations and the minimum time the loading screen is shown

By default, the loading screen is now showing a progress bar and is more minimalistic looking:
image

Another example:

image

The logo size adapts to the size/format of the window. The background image covers the whole window and is centered.

TODO:
- [x] Give an alternative for very small resolutions
- [x] Disable min duration + logo for previews (and explanation test in the loading screen editor)


arthuro555:
I think some subscribers might want to replace fully the "Made with GDevelop" logo with their own, do you think that would be a possibility?

Oxey405:
OMG thanks !!! thank you A LOT

Oxey405:
> I think some subscribers might want to replace fully the "Made with GDevelop" logo with their own, do you think that would be a possibility?

You can just make a background with your logo and disable the gdevelop logo

Silver-Streak:
As you're already covering the low resolution/high resolution variance, My biggest recommendation: 

If "Made with GDevelop" is being kept as one line instead of multiple, is choose a font (or set up spacing) so that the "G" of GDevelop is kept centered to the screen. 

Otherwise due to the contrast between it and the position of the loading bar, one or the other looks off centered.

Bouh:
It's a cool improvement for everyone who asked this level of customization in past!

Here some suggestion after a quick test:
 - Think to remove it for preview
 - The progress bar dimension are in pixels, its have more sense in percentages because for now, you can't have a responsive bar at 100% of the wideness.
 - The image fill the entire window, I'm unable to limit the size
 - Antialiasing is not taken into account.

Silver-Streak:
> * The image fill the entire window, I'm unable to limit the size

I would think the goal on this would be any background images would always need to be game resolution size. So you could just add a border to whatever smaller image you want that makes it meet your game resolution size? Maybe I'm wrong.

4ian:
> Think to remove it for preview

Yep, the logo won't be there for the preview.

> The progress bar dimension are in pixels, its have more sense in percentages because for now, you can't have a responsive bar at 100% of the wideness.

Might be better indeed!

> The image fill the entire window, I'm unable to limit the size

That's by design. If you really want a specific size, in the game properties you can disable the game resolution resizing (and enable it again later in the game with events). So you're 100% sure of the size of the loading screen and can put an image exactly at this size (and you'll have black borders around the game).

> Antialiasing is not taken into account.

You mean for the image? 

> Number inputs for time is in second, but the step are too fast. currently 1by1, 0.1 is welcome.

Good remark yes!

> I think some subscribers might want to replace fully the "Made with GDevelop" logo with their own, do you think that would be a possibility?

They would need to disable the "Made With GDevelop" logo then, and add their logo to the background... But we might in the future improve this though with an optional logo to display, which is then fading out, showing "Made with GDevelop" (I think the Unity splashscreen allow that).

> Otherwise due to the contrast between it and the position of the loading bar, one or the other looks off centered.

It does not strike me as much, especially if you have a larger loading bar, but maybe you're right... but if I center the "G", I'm afraid this will make the rest unbalanced.

Silver-Streak:
> It does not strike me as much, especially if you have a larger loading bar, but maybe you're right... but if I center the "G", I'm afraid this will make the rest unbalanced.

Yeah, this one is a bit odd. I don't find a really good option that meets all needs.

Looking at Unity, Unreal, and Defold, they all share permutations of the concepts above, but only unity has one that runs into the alignment issue I'm seeing.

I've got some examples below, but after looking at other engines, I wonder if there might be an option to have "Made With" in smaller text to the right of the "G" logo, right above the "Develop" text. This is similar to Defold's last option below.

This would allow you to center the whole thing. 

This is a SUPER ugly mockup (I'm terrible at graphic design/art/etc), and obviously the font/sizes of everything would need to change, but something like this, maybe? 
image

Feel free to ignore all of this, but wanted to share my research. _all_ of the below engines, like your suggestion, have a black version, a white version, and a full color version.


Unity has a "Made with Unity" in a few different configurations:
1. "Made With" above the Unity logo symbol, the word Unity next to it. (For GDevelop, this would be "Made With" above the GDevelop G). The whole thing is treated as one image and then centered.
image
2. "Made With Unity" all on one line, with the Unity symbol to the right of "Made With", and the word "Unity" to the right of it. This one is interesting because they center on the Unity symbol.
image
3. Just a display of "Unity" and the logo above it. (This one would require separating the G logo from the word GDevelop, and just showing the logo then normal text "GDevelop" Under it. You'd need to determine if that aligns with your brand language though).
image


Unreal is a bit simpler:
1. Unreal logo above "Unreal Engine" beneath it. This is apparently the most common.
image
2. "Powered by Unreal Engine". This seems relatively rarer.
image

Defold is even simpler:
1. Just the logo: 
image

2. The logo and engine name: 
image
3. "Made with defold"
image

Silver-Streak:
One real quick alternative that seems to merge a lot of the concepts, maybe? 

image


But as always, I'm not a graphic designer, feel free to disregard all of this.

4ian:
I made a few changes and adaptations:
- I went for a logo with the "G" centered (with a slightly larger "made with" text).
- There is now a button to run the preview with the loading screen shown entirely (otherwise the loading screen GDevelop logo is not shown for previews, and the minimum duration is skipped).
- The width of the progress bar is now a percent of the game width - though you can specify a minimum and maximum size in pixels.
- Fixed antialiasing, 0.1 step for durations, improved a bit the editor.

We can always add more logo variations in the future or a feature to display the game logo before the "Made with GDevelop" logo, but for now I think that will be enough :)

4ian:
Merging now so this is included in the next version. Thanks all for the feedback and the comparisons, this was very helpful 👍

PR #2838, opened 2021-07-24
Asset store: Pixelate pixel art assets in viewer & make it bigger

Bouh:
Make all assets with **pixel art** tag bigger and pixelated in the viewer.

Assets from **Grafxkid** are not tagged with **pixel art** and should be.
@Entr0py404 I think you can add an option in your editor to tag assets as pixel art with a checkbox, if isn't already done :)

## Before
image

## After
image

image

### DIFF
--- a/newIDE/app/src/AssetStore/AssetCard.js
+++ b/newIDE/app/src/AssetStore/AssetCard.js
@@ -19,6 +19,11 @@ const styles = {
     verticalAlign: 'middle',
     pointerEvents: 'none',
   },
+  previewImagePixelated: {
+    width: '-webkit-fill-available',
+    imageRendering: 'pixelated',
+    padding: 15,
+  },
   icon: {
     color: '#fff',
   },
@@ -51,6 +56,14 @@ type Props = {|
   onOpenDetails: () => void,
 |};
 
+const isPixelArt = assetShortHeader => {
+  let returnValue = false;
+  assetShortHeader.tags.map(tag => {
+    if (tag.toLowerCase() === 'pixel art') returnValue = true;
+  });
+  return returnValue;
+};
+
 export const AssetCard = ({ assetShortHeader, onOpenDetails, size }: Props) => {
   return (
     <ButtonBase onClick={onOpenDetails} focusRipple>
@@ -59,11 +72,20 @@ export const AssetCard = ({ assetShortHeader, onOpenDetails, size }: Props) => {
           <CheckeredBackground />
           <CorsAwareImage
             key={assetShortHeader.previewImageUrls[0]}
-            style={{
-              ...styles.previewImage,
-              maxWidth: 128 - 2 * paddingSize,
-              maxHeight: 128 - 2 * paddingSize,
-            }}
+            style={
+              isPixelArt(assetShortHeader)
+                ? {
+                    ...styles.previewImage,
+                    ...styles.previewImagePixelated,
+                    maxWidth: 128 - 2 * paddingSize,
+                    maxHeight: 128 - 2 * paddingSize,
+                  }
+                : {
+                    ...styles.previewImage,
+                    maxWidth: 128 - 2 * paddingSize,
+                    maxHeight: 128 - 2 * paddingSize,
+                  }
+            }
             src={assetShortHeader.previewImageUrls[0]}
             alt={assetShortHeader.name}
           />
--- a/newIDE/app/src/AssetStore/AssetCard.js
+++ b/newIDE/app/src/AssetStore/AssetCard.js
@@ -24,10 +24,7 @@ const styles = {
   },
   previewImagePixelated: {
     width: '100%',
-    imageRendering: '-moz-crisp-edges',
-    imageRendering: '-webkit-optimize-contrast',
-    imageRendering: '-webkit-crisp-edges',
-    imageRendering: 'pixelated',
+    imageRendering: '-moz-crisp-edges' || 'pixelated', // TODO to check
     padding: 15,
   },
   icon: {
--- a/newIDE/app/src/AssetStore/AssetCard.js
+++ b/newIDE/app/src/AssetStore/AssetCard.js
@@ -4,6 +4,7 @@ import {
   type AssetShortHeader,
   isPixelArt,
 } from '../Utils/GDevelopServices/Asset';
+import { getPixelatedImageRendering } from '../Utils/CssHelpers';
 import ButtonBase from '@material-ui/core/ButtonBase';
 import Text from '../UI/Text';
 import { CorsAwareImage } from '../UI/CorsAwareImage';
@@ -24,7 +25,7 @@ const styles = {
   },
   previewImagePixelated: {
     width: '100%',
-    imageRendering: '-moz-crisp-edges' || 'pixelated', // TODO to check
+    imageRendering: getPixelatedImageRendering(),
     padding: 15,
   },
   icon: {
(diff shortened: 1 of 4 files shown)


newIDE/app/src/AssetStore/AssetCard.js
```diff
@@ -51,6 +56,14 @@ type Props = {|
   onOpenDetails: () => void,
 |};
 
+const isPixelArt = assetShortHeader => {
```
4ian, line 59: As this reused let's move this to GDevelopServices/Asset.is

In the future we could change this so that pixel art asset are marked by the server :)

newIDE/app/src/AssetStore/AssetCard.js
```diff
@@ -19,6 +19,11 @@ const styles = {
     verticalAlign: 'middle',
     pointerEvents: 'none',
   },
+  previewImagePixelated: {
+    width: '-webkit-fill-available',
+    imageRendering: 'pixelated',
```
4ian, line 24: Let's prefer crisp-edges, as it works in most browsers and is recommended in https://developer.mozilla.org/en-US/docs/Web/CSS/image-rendering for pixel art.

newIDE/app/src/AssetStore/AssetDetails.js
```diff
@@ -159,7 +172,16 @@ export const AssetDetails = ({
...
                 <CheckeredBackground />
                 <CorsAwareImage
-                  style={styles.previewImage}
+                  style={
+                    isPixelArt(assetShortHeader)
+                      ? {
+                          ...styles.previewImage,
+                          ...styles.previewImagePixelated,
+                        }
+                      : {
```
4ian, line 181: No need to spread here.

newIDE/app/src/AssetStore/AssetCard.js
```diff
@@ -19,6 +19,11 @@ const styles = {
     verticalAlign: 'middle',
     pointerEvents: 'none',
   },
+  previewImagePixelated: {
+    width: '-webkit-fill-available',
```
4ian, line 23: Is there a standard way for this? Like setting the width to 100 and setting a max height?

newIDE/app/src/AssetStore/AssetCard.js
```diff
@@ -59,11 +72,20 @@ export const AssetCard = ({ assetShortHeader, onOpenDetails, size }: Props) => {
...
-            style={{
-              ...styles.previewImage,
-              maxWidth: 128 - 2 * paddingSize,
-              maxHeight: 128 - 2 * paddingSize,
-            }}
+            style={
+              isPixelArt(assetShortHeader)
+                ? {
+                    ...styles.previewImage,
+                    ...styles.previewImagePixelated,
```
4ian, line 79: Make a ternary inside this:
...(isPixelArt ? styles.previewImagePixelated : undefined),

4ian: (Shorter and avoid to duplicate the max width/height, and same number of objects creation)

newIDE/app/src/AssetStore/AssetCard.js
```diff
@@ -51,6 +56,14 @@ type Props = {|
   onOpenDetails: () => void,
 |};
 
+const isPixelArt = assetShortHeader => {
+  let returnValue = false;
+  assetShortHeader.tags.map(tag => {
```
4ian, line 61: Use .some(

4ian (commented):
Great idea, will look much better.

newIDE/app/src/AssetStore/AssetCard.js
Bouh: `crisp-edges` is blurry and not found in chrome dev tools.
image

Bouh:
All is fixed.
For `crisp-edges` or `pixelate` I've used the styles used for antialiasing option from the runtime applied on the canvas, all browsers are supported.

newIDE/app/src/AssetStore/AssetCard.js
```diff
@@ -19,6 +22,14 @@ const styles = {
     verticalAlign: 'middle',
     pointerEvents: 'none',
   },
+  previewImagePixelated: {
+    width: '100%',
+    imageRendering: '-moz-crisp-edges',
+    imageRendering: '-webkit-optimize-contrast',
+    imageRendering: '-webkit-crisp-edges',
+    imageRendering: 'pixelated',
```
4ian, line 30: The alternatives are not used, in JS the latest key is overriding the others

newIDE/app/src/AssetStore/AssetDetails.js
```diff
@@ -73,6 +81,14 @@ type Props = {|
   isBeingInstalled: boolean,
 |};
 
+const isPixelArt = assetShortHeader => {
```
4ian, line 84: Import it from Asset.js

newIDE/app/src/Utils/GDevelopServices/Asset.js
```diff
@@ -232,3 +232,11 @@ export const listAllLicenses = (): Promise<Array<License>> => {
     })
     .then(response => response.data);
 };
+
+export const isPixelArt = (assetShortHeader: AssetShortHeader) => {
```
4ian, line 236: This can be a single instruction:

Return assetShortHeader.tags.some(tag => ...

No need for extra Boolean

newIDE/app/src/AssetStore/AssetCard.js
Bouh: I've not found how to use multiple values, in an array it didn't work, in a string with or without comma it didn't work.

newIDE/app/src/AssetStore/AssetCard.js
```diff
@@ -19,6 +22,11 @@ const styles = {
     verticalAlign: 'middle',
     pointerEvents: 'none',
   },
+  previewImagePixelated: {
+    width: '100%',
+    imageRendering: '-moz-crisp-edges' || 'pixelated', // TODO to check
```
4ian, line 27: Can we just use crisp-edges? Seems supported on all our targets:
image

Bouh: The table from MDN and my tests are different.

image

4ian: 🤯
Alright then, let's do a helper with CSS.supports:

```
// @flow
let pixelatedImageRenderingValue: ?string = null;
export const getPixelatedImageRendering = (): string => {
  if (pixelatedImageRenderingValue) return pixelatedImageRenderingValue;
  
  return (pixelatedImageRenderingValue = CSS.supports('image-rendering: crisp-edges') ? 'crisp-edges' : 'pixelated');
}
```

In a CSSHelper.js file in Utils and then call it: `imageRendering: getPixelatedImageRendering(),` and we should be good (I cache the result to avoid calling CSS.supports at every render).

Bouh:
Thanks the helper was the solution, it works everywhere.

4ian (approved):
🚀

4ian:
Flow is mad because it does not know CSS global object, you can use `global.CSS` instead or add a // $FlowFixMe above :)

4ian:
Thank you!

blurymind:
Just saw this article
https://dev.to/auroratide/pixelart-and-the-image-rendering-paradox-2p8k

Came a little late i guess

PR #2842, opened 2021-07-26
Add thumbnails icons in resources list

Bouh:
From #1724

image

### DIFF
--- a/newIDE/app/src/ResourcesList/index.js
+++ b/newIDE/app/src/ResourcesList/index.js
@@ -20,6 +20,7 @@ import {
 } from './ResourceUtils.js';
 import { type ResourceKind } from './ResourceSource.flow';
 import optionalLazyRequire from '../Utils/OptionalLazyRequire';
+import ResourcesLoader from '../ResourcesLoader';
 
 const lazyRequireGlob = optionalLazyRequire('glob');
 const path = optionalRequire('path');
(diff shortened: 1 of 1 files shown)


newIDE/app/src/ResourcesList/index.js
```diff
@@ -155,6 +155,25 @@ export default class ResourcesList extends React.Component<Props, State> {
...
+      case 'json':
+        return 'res/actions/fichier24.png';
+      case 'video':
+        return 'JsPlatform/Extensions/videoicon24.png';
+      case 'font':
+        return 'res/actions/font24.png';
+      case 'bitmapFont':
+        return 'JsPlatform/Extensions/bitmapfont32.png';
+      default:
+        return 'res/unknown32.png'; //Icon type inconnu
```
4ian, line 173: Comment in French. Remove it entirely, the "default" is already telling what's going on.

newIDE/app/src/ResourcesList/index.js
```diff
@@ -155,6 +155,25 @@ export default class ResourcesList extends React.Component<Props, State> {
     );
   };
 
+  _getResourceThumbnail = (resource: gdResource) => {
+    switch (resource.getKind()) {
+      case 'image':
+        return getLocalResourceFullPath(this.props.project, resource.getName());
```
4ian, line 161: The name is not necessarily the file path. Use getFile

4ian: This is using getLocalResourceFullPath - is this electron only?
What is happening in the web app when this is called?

4ian (commented):
Good idea to add this!

Bouh:
Fixes are made.
I've open examples and tested all types of objects, on desktop and web app.
Even the default icon on broken resources.
image

4ian (approved):
👍

PR #2844, opened 2021-07-27
Fix path editor style issue ##2759

### LINKED ISSUE #2759: [Bug] piskel buttons
I'm not sure if it's something related to the GD code, but as it's built in, I decided to post it here...

fix this

Piskel's screen has a few buttons above, which I've identified in red. There is a button over the other there, and I use it a lot myself to be able to select the folder that the image will be. I don't know if it's something directly related to the GD, but it's definitely something that should be fixed.

I don't know if it's just me, because I use linux and I don't know if the same happens on Windows machines!

This happens both in fullscreen and in windowed mode.

blurymind:
This fixes #2759

It also gets rid of the duplicated css, so an error like this would be  harder to make in the future


4ian:
Hey @blurymind sorry for the delay on this! Nice refactoring 👍

4ian (approved):

blurymind:
thank you for reviewing :)

PR #2856, opened 2021-08-02
NavMesh pathfinding behavior

D8H:
This is an implementation of a pathfinding behavior that uses A* over a NavMesh.
The initial discussion: https://github.com/4ian/GDevelop/discussions/2787

Trello card:
* https://trello.com/c/Fvu5OgmO/330-add-an-isometric-option-to-pathfinding-behavior-to-follow-isometric-angles-instead-of-45-degrees-angles

## Used work
* a NavMesh generator by Stephen A. Pratt (ported from Recast).
  * http://www.critterai.org/projects/nmgen_study/
* a A* pathfinding over a NavMesh by Michael Hadley
  * https://github.com/mikewesthad/navmesh
* a scan-line polygon fill algorithm by Darel Rex Finley 
  * https://alienryderflex.com/polygon_fill/

## Known limitations
* The pathfinding can't find a path if the object is not inside the mesh.
* The pathfinding can't find a path if the target is not inside the mesh (even if the target is not actually in an obstacle, but in the margin).
* The pathfinding library loops on polygons center to search for start and end nodes. This kill the performances on big mesh.

We could add the mesh polygons AABB to a RBush solve these 3 limitations.

## Examples
Project must be opened with this custom build: https://www.dropbox.com/s/6yy7uptw5ffceoh/gdevelop-navmesh-beta139-2022-07-30.7z?dl=1

* squares at random position
  * Project: https://www.dropbox.com/s/t8in8c9ba5z3zm2/NavMeshForestExample.zip?dl=1
  * Build: https://d8h.itch.io/navmesh-random-obstacles-demo

* 2 sizes of moving objects and obstacle dragging
  * Project: https://www.dropbox.com/s/q0hg46ytv6as2fb/NavMeshSizeExample.zip?dl=1
  * Build: https://d8h.itch.io/navmesh-draggable-obstacles-demo

* a modified version of the AI example
  * Project: https://www.dropbox.com/s/fvjd7b0bjfxvbdv/NavMesh-basic-ai-with-pathfinding.zip?dl=1
  * Build: https://d8h.itch.io/navmesh-ai-demo

* to check that functions are linked to the implementation
  * Project: https://www.dropbox.com/s/yzda05q8s206qy0/NavMeshFunctionTest.zip?dl=1

* a modified version of the isometry example
  * Project: https://www.dropbox.com/s/r5np0yhfp7grscg/NavMeshIsometricExample.zip?dl=1
  * Build: https://d8h.itch.io/navmesh-isometric-demo

GDevelopNavMeshTiny


4ian:
Nice! 🚀 Looking forward to checking all of this!

Extensions/NavMeshPathfindingBehavior/JsExtension.js
```diff
@@ -0,0 +1,788 @@
...
+// Import types to allow Flow to do static type checking on this file.
+// Extensions declaration are typed using Flow (like the editor), but the files
+// for the game engine are checked with TypeScript annotations.
+import { type ObjectsRenderingService, type ObjectsEditorService } from '../JsExtensionTypes.flow.js'
+*/
+
+const declarePathfindingBehavior = function (
+  _ /*: (string) => string */,
+  gd /*: libGDevelop */,
+  extension /*: PlatformExtension */
```
4ian, line 25:
```suggestion
  extension /*: gdPlatformExtension */
```

Extensions/NavMeshPathfindingBehavior/JsExtension.js
```diff
@@ -0,0 +1,782 @@
...
+      'NavMeshPathfinding',
+      _('NavMesh Pathfinding behavior'),
+      _(
+        'Pathfinding allows to compute an efficient path for objects, avoiding obstacles on the way.'
+      ),
+      'D8H',
+      'MIT'
+    );
+
+    extension
```
4ian, line 671: I think these properties would fit well in a "BehaviorSharedData"?

See here an example:
https://github.com/4ian/GDevelop/blob/f4512242e46084fc32dafe68e3f1a5ee5aa542e2/Extensions/ExampleJsExtension/JsExtension.js#L242-L272

Typically, this is used by the physics engine to store the gravity, the size of the virtual world. I think this would be well adapted here too - as it's not global to the project (but to a scene instead).

D8H: I see, I've never used scene properties before. I'll do that.

D8H: Can having both be useful?
For instance, the ViewPoint and CellSize will probably be the same for every levels. The game properties could be used to default some scene properties. The UI could display the value from the game properties as a hint in the scene properties fields using a new metadata.

Extensions/NavMeshPathfindingBehavior/JsExtension.js
```diff
@@ -0,0 +1,866 @@
...
+    .getCodeExtraInformation()
+    .setFunctionName('getMovementAngle');
+};
+
+const declareObstacleBehavior = function (
+  _ /*: (string) => string */,
+  gd /*: libGDevelop */,
+  extension /*: gdPlatformExtension */
+) {
+  //TODO Add some kind of layer to select the obstacles for an object?
```
4ian, line 639: What do you mean exactly here?

Extensions/NavMeshPathfindingBehavior/navmeshpathfindingobstacleruntimebehavior.ts
```diff
@@ -0,0 +1,243 @@
...
+    _areaBottomBound: integer;
+
+    /**
+     * The navigation meshes by moving object size
+     * (rounded on _cellSize)
+     */
+    _navMeshes: Map<integer, gdjs.NavMesh> = new Map();
+
+    /**
+     * @param object The object
```
4ian, line 31: Comment is outdated/wrong here.

Extensions/NavMeshPathfindingBehavior/JsExtension.js
```diff
@@ -0,0 +1,866 @@
...
+    return false;
+  };
+  // $FlowExpectedError - ignore Flow warning as we're creating a behavior
+  pathfindingBehavior.getProperties = function (behaviorContent) {
+    const behaviorProperties = new gd.MapStringPropertyDescriptor();
+
+    behaviorProperties
+      .getOrCreate('acceleration')
+      .setValue(behaviorContent.getDoubleAttribute('acceleration').toString())
+      .setLabel('Acceleration');
```
4ian, line 68: As put in another comment, put a translation marker (`_()`) for all labels.

Extensions/NavMeshPathfindingBehavior/JsExtension.js
4ian: I think that's in this case a UX improvement to be done for the editor rather than something that the extension should support itself :)

Extensions/NavMeshPathfindingBehavior/navmeshpathfindingruntimebehavior.ts
```diff
@@ -0,0 +1,443 @@
+/*
+GDevelop - Pathfinding Behavior Extension
```
4ian, line 2: You can update this license banner

Extensions/NavMeshPathfindingBehavior/JsExtension.js
```diff
@@ -0,0 +1,866 @@
...
+    )
+    .addParameter('object', _('Object'), '', false)
+    .addParameter(
+      'behavior',
+      _('Behavior'),
+      'NavMeshPathfindingBehavior',
+      false
+    )
+    //TODO tried Drawer, PrimitiveDrawing::Drawer, ShapePainterObject
+    .addParameter('objectPtr', _('Shape painter'), '', false)
```
4ian, line 175: This is not working because extensions are assumed not to use objects from other extensions, so `addParameter` will prefix the object type by the extension namespace (so `Drawer` would become PrimitiveDrawing::Drawer in the `PrimitiveDrawing` extension, but here it becomes `NavMeshPathfinding::Drawer`)

I don't have a good solution for now.

Extensions/NavMeshPathfindingBehavior/navmeshpathfindingruntimebehavior.ts
```diff
@@ -0,0 +1,443 @@
...
+      // No path found
+      this._pathFound = false;
+    }
+
+    /**
+     * Draw the navigation mesh on a shape painter object for debugging purpose
+     * @param shapePainter
+     */
+    drawNavMesh(shapePainter: gdjs.ShapePainterRuntimeObject) {
+      //TODO should it check that shapePainter is not null?
```
4ian, line 316: You should definitely handle a null object yep!

Extensions/NavMeshPathfindingBehavior/navmeshpathfindingruntimebehavior.ts
```diff
@@ -0,0 +1,443 @@
...
+      //TODO or even use a a list?
```
4ian, line 317: No need for list, you can keep objectPtr as it's made to get only one object and I think here it makes sense to only draw on a single object..

4ian (commented):
Made a new review pass!

Extensions/NavMeshPathfindingBehavior/JsExtension.js
4ian: Actually solution will be to use `.setParameterExtraInfo("PrimitiveDrawing::Drawer")` :) 
See #2894

Extensions/NavMeshPathfindingBehavior/navmeshpathfindingruntimebehavior.ts
Bouh: This could be used in the [debug view](https://github.com/4ian/GDevelop/blob/master/GDJS/Runtime/runtimescene.ts#L46)  to the future.

4ian: Yep I wonder if we could expose a `PIXI.Graphics` object for the debug view, for any extension that wants to draw stuff on it? Rather than using a shape painter object

Bouh: Great idea, furthermore it will be usable in IDE too? My main idea was to be able to get the same debug view in runtime as in the IDE, notwithstanding to add buttons for it.

Extensions/NavMeshPathfindingBehavior/JsExtension.js
D8H: Thanks for the review, is the feature branch planned to be merged or should I use it from there?

Extensions/NavMeshPathfindingBehavior/navmeshpathfindingruntimebehavior.ts
D8H: In case of the NavMesh, there can be several meshes depending on the instances size. If it's integrated to the debugger, it may need a UI to select the instances to draw because with 3 meshes or more it will be hard to read.

D8H:
> I think that's in this case a UX improvement to be done for the editor rather than something that the extension should support itself :)

Do you mean the ability to precalculate the meshes?
There is still the need to handle dynamic changes, but I will remove the action and watch the obstacles instead. It will be easier to use optimizations if the invalidation is done internally.

4ian:
> Do you mean the ability to precalculate the meshes?

No (my comment is out of context, not sure why), I meant the fact that using the behavior shared data, having multiple scenes force you to change multiple times your settings if you have multiple scenes with the behavior. Not a big deal!

D8H:
I added an action to disable the NavMesh updates. For instance, it allows to drag an obstacle while paths are computed without rebuilding/invalidating the meshes continuously.

NavMeshUpdate2

Extensions/NavMeshPathfindingBehavior/A_navmeshall.ts
```diff
@@ -0,0 +1,1234 @@
+namespace gdjs {
+  // The following implementation of the A* algorithm is from:
+  // https://github.com/bgrins/javascript-astar
+  // and is under this license:
```
4ian, line 4: Thanks for these comments, very clear and useful.

4ian (commented):
Another set of comments.
I think I've passed at least once on everything.

Let's flag the extension as experimental (add ` (Experimental)` to the behaviors name).
Let's try to polish as much as possible the code and TODOs and then it should be good to go :)

Extensions/NavMeshPathfindingBehavior/A_navmeshgenerator.ts
```diff
@@ -0,0 +1,3411 @@
...
+// - filterNonObstacleVertices was added
+//
+// The Java implementation was also inspired from Recast that can be found here:
+// https://github.com/recastnavigation/recastnavigation
+
+namespace gdjs {
+  /**
+   * A cell that holds data needed by the 1st steps of the NavMesh generation.
+   */
+  export class RasterizationCell {
```
4ian, line 21: I know this means reworking tests and other files, but can we put all these classes into an additional namespace?
```
namespace gdjs { 
namespace NavMesh {
```

Otherwise, things being on `gdjs` directly would confuse users who would think these are general purpose tools. Also should we in the future develop a version 2 of NavMesh, we'll conflict with the existing classes - that will be terrible!

General rule is simple: Put all your extension tools into a namespace, and reduce the number of stuff on `gdjs` directly to strictly the minimum: behaviors, objects, that's it :)

Extensions/NavMeshPathfindingBehavior/navmeshpathfindingobstacleruntimebehavior.ts
```diff
@@ -0,0 +1,368 @@
...
+          this._cellSize,
+          // make cells square in the world
+          this._cellSize / this._isometricRatio
+        );
+        gdjs.ObstacleRasterizer.rasterizeObstacles(grid, this._obstacles);
+        gdjs.RegionGenerator.generateDistanceField(grid);
+        gdjs.RegionGenerator.generateRegions(grid, obstacleCellPadding);
+        //TODO make threshold configurable ?
+        // 1 seems to be a good value.
+        const threshold = 1;
```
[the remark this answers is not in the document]
D8H: It is the maximum distance (in cells) the edge of the contour may deviate from the source geometry when the rastered obstacles are vectorized.

Setting it to:
- 1 ensure that an aliased edge won't be split to more edges.
- more that 1 will reduce the number of edges but the obstacles edges will be followed with less accuracy.
- less than 1 might be more accurate but it may try to follow the aliasing and be a lot less accurate.

For instance, with threshold = 0.1, the aliasing is followed and it creates a lot of polygons.
Iso0 1
Note that a lower threshold won't make up for corners that are forgotten at the rasterization step.

Whereas at 0.71 or 1 the edges are better rebuilt:
Iso0 71

Values under 1 can be useful in specific cases:
- when edges are horizontal or vertical, there is no aliasing so value near 0 can do better results.
- when edges are 45° multiples, aliased vertex won't be farther than sqrt(2)/2 so values over 0.71 should give good results but not necessarily better than 1.

For instance, with a threshold of 0.1:
Squares0 1

But greater values (here, 1) are a good option to reduce the number of vertices:
Squares1

D8H:
With the automatic invalidation of the NavMesh, this can happen:
* events change the obstacle locations
* events ask a path
* pre-events, the obstacle manager watch the obstacles and invalidate NavMeshes

The path is actually computed with the old obstacle locations. This may not be an issue because the other pathfinding behavior works this way too.

D8H:
I opened 4 change requests.
Do you think one of this requests should be included in the 1st version?

4ian:
> For clarity, should I remove from the code the TODO linked to these requests?

I would say yes, so that the code is cleaner (in the sense these are not things to do to improve the code).

> Do you think one of these requests should be included in the 1st version?

Thanks for opening these. I don't think any should be mandatory for a first version (as you demonstrated in an example, the NavMesh pathfinding can be used to replace the old pathfinding, which is enough for me to consider it publishable!).

D8H:
I did a pass on todo.
* In the behaviors, only one remain on the debug drawing.
* In the NavMesh generator, it's remainders of the part of CritterAI implementation that hasn't been ported. I didn't see the need for them but in case a bug is found, it can be useful to check if this could solve it.

4ian:
With the platformer rework now in, we can I think get back to this! 
@D8H anything that was outstanding there? Do you want to rebase on master (otherwise I'll do it)?

D8H:
For the library, I left a catted TS instead of a packaged library because I think it will be easier to debug, but I could package one if your want.

4ian:
@D8H may I ask you to rebase an additional time this on top of master? 😇

4ian:
I think that we could go ahead with this if we:
- Rename the existing pathfinding as "Pathfinding (grid based)"
- Name this one as "Pathfinding (polygon based)" or something similar so that it's understandable. 
- Surely mark this one as experimental for a few versions :)

Also, I think we should clean the usage of the debug draw. It's very nice to have, but I think it should use the runtime scene exposing some way to "add stuff" to the debug draw, and the action should work then without a shape painter :) 
I say this because I just tried to test this:

image

And this does not work because it can't be called more than once in a row because it's waiting to draw the result. And more generally we should avoid relying on a shape painter, it's convenient but not really for the user as there is too many chances someone would badly configure it :)

D8H:
It gets a bit messy when the NavMesh is shown with hitboxes.

NavMeshDebug

I guess we could
- Add a new action to hide or show the debug layer
- In the action "Draw collision hitboxes and points", rename the parameter "Enable debug draw" as "Show collision masks and points"

The big issue would be to make it understandable for users that one action alone doesn't do anything.
Does forwarding to the other action in the description enough to make it user friendly?

The good side is that it allows users to toggle between debug and production easily and it's very simple to implement (an automatic activation would mean to keep track of the number of JS instances that want to draw on debug to disable it when there is none).

D8H:
I also changed the descriptions to start with a verb. It is a recommendation to write documentations in the wiki and I thing it apply well for behavior descriptions. It avoid heavy formulation like "With this behavior", "This behavior allows to". The title is often a noun so event for "Platform" or "Character", it could be better to start with a verb to avoid redundancy.
I will ask the extension team if they want to add this in the best practice page.

BehaviorSumup

D8H:
I added 2 parameters to allow a clear debug view for the behavior to draw on it to without breaking change.
DebugViewAction

4ian:
Looks great! 👍
I think we should merge this early this week so that it's part of the next release, after making an additional check on the whole thing :)

Silver-Streak:
@D8H Reading through this PR and the original discussion, I noticed there isn't an action to stop movement on the path. Is that something you would consider adding? 

(This also doesn't exist on the main pathfinding behavior, now that I've dug in on it, but no reason to keep that deficiency in your new actions if it can be avoided)

D8H:
> @D8H Reading through this PR and the original discussion, I noticed there isn't an action to stop movement on the path. Is that something you would consider adding?
> 
> (This also doesn't exist on the main pathfinding behavior, now that I've dug in on it, but no reason to keep that deficiency in your new actions if it can be avoided)

Doesn't the "disable behavior" action serve this purpose?

The goal was to have the same level of functionality to keep the PR from growing and add new features after it's merged.

Silver-Streak:
> Doesn't the "disable behavior" action serve this purpose?

Sadly, no, it keeps the path going as soon as you re-enable it. 

> The goal was to have the same level of functionality to keep the PR from growing and add new features after it's merged.

That's totally valid, but I thought I would ask.

Hillihix:
When should this behavior be expected approximately?

D8H:
> When should this behavior be expected approximately?

Probably not soon. We should probably rather add a shared properties feature to events-based behaviors and make this a community extension.

Hillihix:
> > When should this behavior be expected approximately?
> 
> Probably not soon. We should probably rather add a shared properties feature to events-based behaviors and make this a community extension.

Can I learn more about this feature?

D8H:
It's the scene properties where the rastering grid is setup for this extension. Events-based bavariors can only define to be set on the object for now.

D8H:
I close this PR as this extension was added as a community extension.
* https://github.com/GDevelopApp/GDevelop-extensions/pull/669

DHsoftw:
Thanks for this extension. Our game (Zsquare: https://liluo.io/games/aef70ed3-d55b-4eb0-964e-55ba618a478d) used it before the extension was published and worked better than the default pathfinding system in all terms. It should be included as a native GDevelop feature. In our game, we recalculate a new route for every square in the screen every frame, a thing that is impossible to achieve with grid-based pathfinding without lag and for several objects when there are huge distances.

PR #2859, opened 2021-08-03
Add new lines between functions back in ts files

arthuro555:
When switching over to ts, the codemod removed a lot of newlines in the ts files. I used this simple regex on the GDJS Runtime and Extensions folder from vscode to add them back:
image

This makes the code way more readable.


4ian (approved):
👍👍

PR #2860, opened 2021-08-03
[ShapePainter] Define TS types

D8H:
This is to prepare a next PR that will add an action to clear the shape painter at request instead of each frame.

### DIFF
--- a/Extensions/PrimitiveDrawing/shapepainterruntimeobject-pixi-renderer.ts
+++ b/Extensions/PrimitiveDrawing/shapepainterruntimeobject-pixi-renderer.ts
@@ -2,14 +2,15 @@ namespace gdjs {
   import PIXI = GlobalPIXIModule.PIXI;
 
   class ShapePainterRuntimeObjectPixiRenderer {
-    _object: any;
-    _graphics: any;
+    _object: gdjs.ShapePainterRuntimeObject;
+    _graphics: PIXI.Graphics;
 
-    constructor(runtimeObject, runtimeScene) {
+    constructor(
+      runtimeObject: gdjs.ShapePainterRuntimeObject,
+      runtimeScene: gdjs.RuntimeScene
+    ) {
       this._object = runtimeObject;
-      if (this._graphics === undefined) {
-        this._graphics = new PIXI.Graphics();
-      }
+      this._graphics = new PIXI.Graphics();
       runtimeScene
         .getLayer('')
         .getRenderer()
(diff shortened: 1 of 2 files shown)


Extensions/PrimitiveDrawing/shapepainterruntimeobject.ts
```diff
@@ -19,12 +19,12 @@ namespace gdjs {
...
     /** The color (in RGB format) of the outline of the painted shape */
     outlineColor: RGBColor;
-    /** The opacity of the inner part of the painted shape */
-    fillOpacity: number;
-    /** The opacity of the outline of the painted shape */
-    outlineOpacity: number;
+    /** The opacity of the inner part of the painted shape, form 0 to 255 */
+    fillOpacity: integer;
+    /** The opacity of the outline of the painted shape, form 0 to 255 */
+    outlineOpacity: integer;
```
4ian, line 25:
```suggestion
    /** The opacity of the inner part of the painted shape, from 0 to 255 */
    fillOpacity: float;
    /** The opacity of the outline of the painted shape, from 0 to 255 */
    outlineOpacity: float;
```

Typo + even if it's usually set to an integer, the opacity can (and will) be a float (when it's progressively updated for example)

Extensions/PrimitiveDrawing/shapepainterruntimeobject-pixi-renderer.ts
```diff
@@ -2,14 +2,15 @@ namespace gdjs {
...
-    constructor(runtimeObject, runtimeScene) {
+    constructor(
+      runtimeObject: gdjs.ShapePainterRuntimeObject,
+      runtimeScene: gdjs.RuntimeScene
+    ) {
       this._object = runtimeObject;
-      if (this._graphics === undefined) {
-        this._graphics = new PIXI.Graphics();
-      }
+      this._graphics = new PIXI.Graphics();
```
4ian, line 13: 👍

4ian (approved):
Great 👍

Extensions/PrimitiveDrawing/shapepainterruntimeobject.ts
D8H: As the implementation uses it more like an integer at this time, I find it less confusing.
Can we let the person that will change it to 0.0 to 1.0 make it float?

4ian: Well the thing is that it's not an error to have a floating point value (like 254.5), so we should in theory have a float here?

4ian:
Thank you!

PR #2861, opened 2021-08-03
Increase the GDJS tests Karma timeout

4ian:
Semaphore CI tests were sometimes timing out, notably the ones involving PixiJS and real resources to load.

Don't show in changelog


ClementPasteau (approved):
Let's see if it doesn't happen again :) 
Otherwise, we may want to understand why a big timeout is not enough :+1:

PR #2869, opened 2021-08-03
Fix an error log when the new instruction dialog opens.

D8H:
It seems the key property was removed by mistake: [Stack.txt](https://github.com/4ian/GDevelop/files/6925723/Stack.txt)


4ian:
Thanks for fixing my mistakes! 😄

PR #2870, opened 2021-08-03
Move variable event tools to their own file

arthuro555:
Vairbale evtTools were in the `common` namespace, this PR moves them to their own file and namespace.


4ian:
I think this is good, could you fix the failing test GDevelop.js? Thanks! :)

4ian (approved):

PR #2872, opened 2021-08-03
[Platformer] Add an action to forbid to air jump.

D8H:
It also fix: isMoving is now returning true when scale climbing

An example for tests: https://www.dropbox.com/s/y8wcbsilisp2i2n/PlatformerJump.zip?dl=1


4ian:
This is awesome! I've made a few nitpicking changes on comments. Looks great otherwise 👍 Thanks for the proper testing of it too.

PR #2873, opened 2021-08-04
Fix Panel Sprite initial opacity not being correctly applied

### LINKED ISSUE #2830: Panel Sprite 9-patch, opacity bug at beginning of the scene
## Describe the bug
The opacity action of a Panel Sprite is not take into account **At the beginning of the scene**.

## Example for reproduce
The red square is a Panel Sprite, run the preview, the opacity should be at 0 at the first frame, but 255 is used and the object blink.
[bug_9patch_opacity.zip](https://github.com/4ian/GDevelop/files/6858809/bug_9patch_opacity.zip)

[Original bug report on forum](https://forum.gdevelop-app.com/t/panel-sprite-9-patch-initial-opacity-oddity/33930)

## Other details
* Tested on B113 & B111

ClementPasteau:
Fix #2830 

Before:

out_old

After:

out_new


Extensions/PanelSpriteObject/panelspriteruntimeobject-pixi-renderer.ts
```diff
@@ -3,6 +3,7 @@ namespace gdjs {
   class PanelSpriteRuntimeObjectPixiRenderer {
     _object: gdjs.PanelSpriteRuntimeObject;
     _spritesContainer: any;
+    _wrapperContainer: any;
```
4ian, line 6: nitpicking: the typing is not great in this file (this is dating from the time we codemoded from JavaScript+JSDoc type annotations to TypeScript), but you can mark this as a `PIXI.Container`

Extensions/PanelSpriteObject/panelspriteruntimeobject-pixi-renderer.ts
```diff
@@ -57,34 +59,35 @@ namespace gdjs {
...
       runtimeScene
         .getLayer('')
         .getRenderer()
-        .addRendererObject(this._spritesContainer, runtimeObject.getZOrder());
+        .addRendererObject(this._wrapperContainer, runtimeObject.getZOrder());
     }
 
     getRendererObject() {
-      return this._spritesContainer;
+      return this._wrapperContainer;
```
4ian, line 71: This is correct 👍 The "top most" container is indeed the one considered as the "renderer object". It will be the one in the future to receive the Pixi filters (in the PR #2736 being worked on by @HarsimranVirk).

Extensions/PanelSpriteObject/panelspriteruntimeobject-pixi-renderer.ts
```diff
@@ -57,34 +59,35 @@ namespace gdjs {
...
 
     updateOpacity(): void {
-      //TODO: Workaround a not working property in PIXI.js:
-      this._spritesContainer.alpha = this._spritesContainer.visible
-        ? this._object.opacity / 255
-        : 0;
-      this._alpha = this._spritesContainer.alpha;
+      // The alpha is updated on a wrapper around the sprite because a known bug
+      // in Pixi will create a flicker when cacheAsBitmap is set to true
+      // (see https://github.com/pixijs/pixijs/issues/4610)
```
4ian, line 88: Great comment!

Extensions/PanelSpriteObject/panelspriteruntimeobject-pixi-renderer.ts
```diff
@@ -57,34 +59,35 @@ namespace gdjs {
...
+      this._wrapperContainer.alpha = this._object.opacity / 255;
+      this._alpha = this._wrapperContainer.alpha;
```
4ian, line 90: This might be a good opportunity to scrap entirely this `_alpha` property. I *think* it's now useless, because we update directly the alpha on the wrapper container - no need to store it to apply it later. See my other comment in `ensureUpToDate`.

Extensions/PanelSpriteObject/panelspriteruntimeobject-pixi-renderer.ts
```diff
@@ -57,34 +59,35 @@ namespace gdjs {
...
+      return this._wrapperContainer;
     }
 
     ensureUpToDate() {
       if (this._spritesContainer.visible && this._wasRendered) {
         // Update the alpha of the container to make sure that it's applied.
         // If not done, the alpha will be back to full opaque when changing the color
         // of the object.
-        this._spritesContainer.alpha = this._alpha;
+        this._wrapperContainer.alpha = this._alpha;
```
4ian, line 79: I *think* we can remove this assignment to alpha altogether.
It was there before to workaround a similar bug related to the alpha being overwritten when we were tweaking the color of the object... but because we now change the alpha of the wrapper container, we should be 100% safe?

You can then remove it, remove the _alpha property, and just add a comment about the `cacheAsBitmap` to emphasize the workaround again:

```
// Cache the rendered sprites as a bitmap to speed up rendering when lots of panel sprites are on the scene.
// Sadly, because of this, we need a wrapper container to workaround a PixiJS issue with alpha (see updateOpacity).
```

Extensions/PanelSpriteObject/panelspriteruntimeobject-pixi-renderer.ts
```diff
@@ -25,6 +26,7 @@ namespace gdjs {
         );
```
4ian, line 26: You can take the opportunity to remove the useless `if (this._spritesContainer === undefined) {` .
This is dating from the moment where we were calling constructor on already existing objects to "recycle" objects. We then switched to a proper separate method called `reinitialize` (which is not implemented by most objects, but could be to reduce the memory pressure if a lot of objects are created/removed and we identify that some objects are costly to instanciate).

Extensions/PanelSpriteObject/panelspriteruntimeobject-pixi-renderer.ts
```diff
@@ -57,34 +59,35 @@ namespace gdjs {
...
-        : 0;
-      this._alpha = this._spritesContainer.alpha;
+      // The alpha is updated on a wrapper around the sprite because a known bug
+      // in Pixi will create a flicker when cacheAsBitmap is set to true
+      // (see https://github.com/pixijs/pixijs/issues/4610)
+      this._wrapperContainer.alpha = this._object.opacity / 255;
+      this._alpha = this._wrapperContainer.alpha;
     }
 
     updateAngle(): void {
```
4ian, line 93: While it seems to work just like this, I think we should update updateAngle, updatePosition and updateWidth/updateHeight to use the wrapperContainer.

The reason is that if we move only the position/rotation (and pivot point) of _spritesContainer, you can picture the wrapper container as being at position 0;0 in the world, and then "extending" infinitely to contain _spritesContainer (which could be at a large position like 3000;3000).
Instead, we'll do the reverse: move (and rotate) wrapperContainer. spritesContainer will stay inside wrapperContainer unmoved (so it will just follow whatever transform is applied to wrapperContainer).

I think this won't change anything, and the solution would work like this currently, but it seems more logic to have a wrapperContainer to be the one moving in the space and not a child of it (much like a div inside a div).
The current solution could have created an issue if we were to use the position of the wrapper (for example, if the scene would be fetching the coordinate of the `getRendererObject` to decide if an object must be hidden or not), but I think it's not the case (in my example, the scene is always using the position given by the object, not by the renderer - it's the object that is the source of truth).

4ian (commented):
Added a few comments about:
- Additional improvements like typing/dead code (I would normally don't care for a bug fix, but let's take this opportunity now :))
- Ensuring we update the position/angle/pivot of the wrapper container and not the sprites container, as we want the "top most" container to be the one at the right position in the space.

Extensions/PanelSpriteObject/panelspriteruntimeobject-pixi-renderer.ts
ClementPasteau: Is it still relevant to have the ensureUpToDate function at all?
As this function is called for every loop of the game engine, it feels overkill to set `cacheAsBitmap=true` on every frame

4ian: Setting cacheAsBitmap to true is actually calling a "magic" setter (not a big fan of these hidden API calls, but that's how PixiJS works): https://github.com/pixijs/pixijs/blob/f1d98ea807b8bbe7f3aaf41e05f37cf219e79994/packages/mixin-cache-as-bitmap/src/index.ts#L135-L145

It will quickly bail out if already set to true. Otherwise, it will actually generate the cached bitmap at this moment - I think it's important to do it here? We could check if it was already cached, but that's what the setter is doing. So I would say to keep it? 
`cacheAsBitmap` is set to false in other part of the code that can modify the rendered image (namely: changed tint color or changed width/height/texture).

Extensions/PanelSpriteObject/panelspriteruntimeobject-pixi-renderer.ts
```diff
@@ -57,44 +54,43 @@ namespace gdjs {
...
-      this._alpha = this._spritesContainer.alpha;
+      // The alpha is updated on a wrapper around the sprite because a known bug
+      // in Pixi will create a flicker when cacheAsBitmap is set to true
+      // (see https://github.com/pixijs/pixijs/issues/4610)
+      this._wrapperContainer.alpha = this._object.opacity / 255;
     }
 
     updateAngle(): void {
-      this._spritesContainer.rotation = gdjs.toRad(this._object.angle);
+      this._wrapperContainer.rotation = gdjs.toRad(this._object.angle);
```
4ian, line 87: As you're rotating the wrapperContainer, you probably must also set the `pivot` point (see updateWidth/Height/setTexture).

4ian (commented):
Thanks for the typing!
As discussed, I think that removing alpha is good and safe. We can keep cacheAsBitmap as I think it's important. For `_wasRendered`, we definitely need a "stress test" game to ensure we're not breaking something in some edge cases.

I think you also must change the `pivot` point (which determinates the center of the rotation) from the sprites container to the wrapper container. Otherwise it will be set to the top-left of the wrapper container, and rotated objects will be offset (Would be something interesting to check in your stress test btw! Change the pivot, see if it's rendered properly, set it back to the spritesContainer or just comment it and see the objects wrongly rotating).

Once we have checked this I think this will be good to go 👍

Extensions/PanelSpriteObject/panelspriteruntimeobject-pixi-renderer.ts
```diff
@@ -2,10 +2,14 @@ namespace gdjs {
...
     _object: gdjs.PanelSpriteRuntimeObject;
-    _spritesContainer: any;
-    _centerSprite: any;
-    _borderSprites: any;
-    _alpha: float;
+    // The _wrapperContainer must be considered as the main container of this object
+    // All transformations are applied on this container
+    // See updateOpacity for more info why
+    _wrapperContainer: PIXI.Container;
+    // The _spritesContainer is used to create the sprites and apply cacheAsBitmap only
```
4ian, line 9: Uber nitpicking:
It's I think good practice to end comments with a period.
Also you can use `/** */` comments to that it shows as autocompletion when you hover _wrapperContainer or _spritesContainer.

4ian (approved):
I think we're good in terms of changes (removed alpha, fix with the wrapper, removal of dead code, fixed typing), everything seems safe!

PR #2874, opened 2021-08-04
[Experiement] GDJS Runtime: Switch from namespaces to modules

arthuro555:
A very experimental look into converting the base GDJS runtime from namespaces into modules. Nothing's currently really working, it's mainly for looking if it is worth it.

This is not about modularizing all of the typescript code, only the GDJS Runtime. Extensions are still using namespaces. This is to allow bundling and minifying this core runtime into one small file, and also to allow using dependencies like howler and pixi from npm instead of having a local copy, remove the PIXI typing hack, and using modern/standarized typescript.

This is an internal change, it should not affect the API in any noticeable way for the user.

- [X] Convert all .ts files in `GDJS/Runtime` from namespaces to modules
- [X] Type the `gdjs` namespace (Solved through `GDJS/Runtime/globalize.ts`)
- [X] Use the npm modules instead of the js files
- [X] What about extensions that rely on PIXI, as the bundler will tree shake it and prevent the namespace `PIXI` from leaking out for other modules to use? (Solved by "globalizing" the whole PIXI module alongside `gdjs`)
- [X] Handle "extensions" of the runtime, like the debugger client (Solved by generating a separate bundle for that kind of stuff, which calls a method on `gdjs` to register the class, see `GDJS/Runtime/Debugger` (NOTE: only done on the debugger for now, needs to be adapted for renderers and sound managers))
- [ ] Allow adding stuff to the `gdjs` namespace (`namespace gdjs {  export const myFunc = () => {}; }`) from the extensions without losing type information from the real `gdjs` namespace for the current file, as typescript considers that creating a new namespace not extending it.
- [ ] Adapt the build process for the module based project structure
- [ ] Adapt the GDJS Platform to use the bundled file instead of all individual built files (should be trivial)
- [ ] Test for anything that might have broken


4ian:
Very interesting! Let's see how it goes, keep me posted :)

arthuro555:
I had written almost a small essay on all my mistakes doing this PR and of advice for if I or someone else attempts this again but GitHub decided that my work was best off discarded 🙄. Guess we'll do without then, not wasting all of this time again. Either way this PR is beyond salvation, closing it now.

PR #2880, opened 2021-08-06
Bump newIDE version

Bouh:


4ian:
Thanks for checking after me!

PR #2882, opened 2021-08-06
Add nearest support in image preview for pixel art resources

Bouh:
Close [this issue on forum](https://forum.gdevelop-app.com/t/pixel-perfect-animation-viewer/34219)

image

### DIFF
--- a/newIDE/app/src/ResourcesList/ResourcePreview/ImagePreview.js
+++ b/newIDE/app/src/ResourcesList/ResourcePreview/ImagePreview.js
@@ -15,12 +15,17 @@ import Text from '../../UI/Text';
 import { CorsAwareImage } from '../../UI/CorsAwareImage';
 import GDevelopThemeContext from '../../UI/Theme/ThemeContext';
 import CheckeredBackground from '../CheckeredBackground';
+import { getPixelatedImageRendering } from '../../Utils/CssHelpers';
+const gd: libGDevelop = global.gd;
 
 const MARGIN = 50;
 const MAX_ZOOM_FACTOR = 10;
 const MIN_ZOOM_FACTOR = 0.1;
 
 const styles = {
+  previewImagePixelated: {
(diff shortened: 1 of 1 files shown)


newIDE/app/src/ResourcesList/ResourcePreview/ImagePreview.js
```diff
@@ -147,6 +152,15 @@ const ImagePreview = (props: Props) => {
     }));
   };
 
+  const resourceIsSmooth = (
```
4ian, line 155: Put this outside of the component function, to avoid re-allocating a function everytime while it's not needed (this function is "pure", it takes only its arguments to do its job. So you can make it "free", outside of the component, and the JS engine will only need to initialize it once).

newIDE/app/src/ResourcesList/ResourcePreview/ImagePreview.js
```diff
@@ -147,6 +152,15 @@ const ImagePreview = (props: Props) => {
...
+    project: gdProject,
+    resourceName: string
+  ): boolean => {
+    const resource = project.getResourcesManager().getResource(resourceName);
+    const imageResource = gd.asImageResource(resource);
```
4ian, line 160: Any cast is very risky: if the resource is not an image resource, you've instructed the C++/webAssembly code to consider something that is not an `ImageResource` to be handled as a `ImageResource`. This will result in memory corruption if you then try to modify something related to it.

Instead, use a check to verify the "type" of the class. Here you can use "getKind":
```
const resource = project.getResourcesManager().getResource(resourceName);
if (resource.getKind() !== 'image') return false; // The resource is not even an image.

const imageResource = gd.asImageResource(resource);
return imageResource.isSmooth();
```

4ian (approved):
Added a few comments, then good to go 👍

4ian (approved):

PR #2891, opened 2021-08-11
Improve Actions and Conditions search allowing typing mistakes (fuzzy)

### LINKED ISSUE #2864: Allow fuzzy search for action/condition names
## Description

When adding a new Condition or Action in the events tab, searching for the name is really strict which can be frustrating.
Ex: In order to find the action "Change the scene", I cannot be searching "Change scene"

image
image

## Solution suggested

It would be great if the search was a bit more fuzzy, allowing missing words, typos or words in a different order, as it would speed up the creation of a game.

ClementPasteau:
Fixes #2864 

Some examples:
Before:
Screenshot 2021-08-11 at 15 34 30

After:
Screenshot 2021-08-11 at 15 34 11

Before:
Screenshot 2021-08-11 at 15 38 00

After:
Screenshot 2021-08-11 at 15 38 21

Tested the performance of the function under a CPU throttling of 6X:
Before:
Screenshot 2021-08-11 at 15 38 21

After:
Screenshot 2021-08-11 at 15 21 21

There is a performance impact, but I believe that because we're still in the order of 1ms, it's not a big problem.


newIDE/app/src/InstructionOrExpression/EnumerateInstructions.js
```diff
@@ -507,41 +509,32 @@ export const filterInstructionsList = <
...
 
-  const favorExactMatch = (list: Array<T>): Array<T> => {
-    if (!searchText) {
-      return list;
+  list.forEach(option => {
+    // favor direct matches first
+    const lowercaseSearch = searchText.toLowerCase();
+    if (
+      option.displayedName.includes(searchText) ||
+      option.fullGroupName.includes(searchText)
```
4ian, line 524: Previously we did the direct matches only against displayedName (not the group name). Might be worth keeping the previous behavior (a direct match of the group name is maybe not that useful) - and this make the search a tiny bit less complex.

newIDE/app/src/InstructionOrExpression/EnumerateInstructions.js
```diff
@@ -507,41 +509,32 @@ export const filterInstructionsList = <
   list: Array<T>,
   { searchText }: InstructionFilteringOptions
 ): Array<T> => {
-  if (!searchText === '') {
+  if (!searchText) {
```
4ian, line 512: 👍

newIDE/app/src/InstructionOrExpression/EnumerateInstructions.js
```diff
@@ -507,41 +509,32 @@ export const filterInstructionsList = <
...
-  };
+  const directMatches = [];
+  const fuzzyMatches = [];
 
-  const favorExactMatch = (list: Array<T>): Array<T> => {
-    if (!searchText) {
-      return list;
+  list.forEach(option => {
+    // favor direct matches first
+    const lowercaseSearch = searchText.toLowerCase();
```
4ian, line 521: Do this outside of the loop ;)

newIDE/app/src/InstructionOrExpression/EnumerateInstructions.js
```diff
@@ -507,41 +509,32 @@ export const filterInstructionsList = <
...
+    if (
+      option.displayedName.includes(searchText) ||
```
4ian, line 523: Consider transforming to lowercase the displayed name before comparing to search text (so that the match is exact BUT case insensitive). Otherwise we won't match things with capital letters in the names.

4ian (commented):
Added a few comments, otherwise sounds good! :)

newIDE/app/src/InstructionOrExpression/EnumerateInstructions.js
ClementPasteau: oops :+1:

4ian (approved):

PR #2892, opened 2021-08-11
Add Release Ladder action to Platformer Object Behavior

ClementPasteau:
Addresses the feature requested on the Trello roadmap: https://trello.com/c/DDO9SEoy/405-add-an-action-to-release-the-ladder-for-platformer-characters

### DIFF
--- a/Extensions/PlatformBehavior/Extension.cpp
+++ b/Extensions/PlatformBehavior/Extension.cpp
@@ -501,18 +501,21 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
         .SetFunctionName("SimulateJumpKey")
         .SetIncludeFile("PlatformBehavior/PlatformerObjectRuntimeBehavior.h");
 
-    aut.AddAction("SimulateReleaseKey",
-                  _("Simulate release key press"),
-                  _("Simulate a press of the release key (used when grabbing a "
+    aut.AddAction("SimulateReleasePlatformKey",
+                  _("Simulate release platform key press"),
+                  _("Simulate a press of the release platform key (used when grabbing a "
                     "platform ledge)."),
-                  _("Simulate pressing Release key for _PARAM0_"),
+                  _("Simulate pressing Release Platform key for _PARAM0_"),
                   _("Controls"),
                   "res/conditions/keyboard24.png",
                   "res/conditions/keyboard.png")
         .AddParameter("object", _("Object"))
         .AddParameter("behavior", _("Behavior"), "PlatformerObjectBehavior")
-        .SetFunctionName("SimulateReleaseKey")
+        .SetFunctionName("SimulateReleasePlatformKey")
         .SetIncludeFile("PlatformBehavior/PlatformerObjectRuntimeBehavior.h");
+
+    // Support for deprecated names:
+    aut.AddDuplicatedAction("SimulateReleaseKey", "SimulateReleasePlatformKey").SetHidden();
 
     aut.AddAction("SimulateControl",
                   _("Simulate control"),
--- a/Extensions/PlatformBehavior/Extension.cpp
+++ b/Extensions/PlatformBehavior/Extension.cpp
@@ -475,6 +475,20 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
         .SetFunctionName("SimulateLadderKey")
         .SetIncludeFile("PlatformBehavior/PlatformerObjectRuntimeBehavior.h");
 
+    aut.AddAction(
+           "SimulateReleaseLadderKey",
+           _("Simulate release ladder key press"),
+           _("Simulate a press of the Release Ladder key (used to get off a ladder)."),
+           _("Simulate pressing Release Ladder key for _PARAM0_"),
+           _("Controls"),
+           "res/conditions/keyboard24.png",
+           "res/conditions/keyboard.png")
+        .AddParameter("object", _("Object"))
+        .AddParameter("behavior", _("Behavior"), "PlatformerObjectBehavior")
+        .MarkAsAdvanced()
+        .SetFunctionName("SimulateReleaseLadderKey")
+        .SetIncludeFile("PlatformBehavior/PlatformerObjectRuntimeBehavior.h");
+
     aut.AddAction("SimulateJumpKey",
                   _("Simulate jump key press"),
                   _("Simulate a press of the jump key."),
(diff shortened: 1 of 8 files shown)


Extensions/PlatformBehavior/PlatformerObjectRuntimeBehavior.h
```diff
@@ -243,5 +244,6 @@ class GD_EXTENSION_API PlatformerObjectRuntimeBehavior
   bool downKey;
   bool jumpKey;
   bool releaseKey;
+  bool releaseLadderKey;
```
4ian, line 247: That's the extension for GDCpp, you can forget this file.

Extensions/PlatformBehavior/PlatformerObjectRuntimeBehavior.cpp
```diff
@@ -670,4 +677,6 @@ void PlatformerObjectRuntimeBehavior::SimulateControl(const gd::String& input) {
     jumpKey = true;
   else if (input == "Release")
     releaseKey = true;
+  else if (input == "Release Ladder")
```
4ian, line 680: That's the extension for GDCpp, you can also forget this file.

Extensions/PlatformBehavior/Extension.cpp
```diff
@@ -475,6 +475,20 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
...
+           _("Simulate a press of the Release Ladder key (used to get off a ladder)."),
+           _("Simulate pressing Release Ladder key for _PARAM0_"),
+           _("Controls"),
+           "res/conditions/keyboard24.png",
+           "res/conditions/keyboard.png")
+        .AddParameter("object", _("Object"))
+        .AddParameter("behavior", _("Behavior"), "PlatformerObjectBehavior")
+        .MarkAsAdvanced()
+        .SetFunctionName("SimulateReleaseLadderKey")
+        .SetIncludeFile("PlatformBehavior/PlatformerObjectRuntimeBehavior.h");
```
4ian, line 490: Nitpicking: you can remove these two lines (used previously by GDCpp, and would be overwritten anyway in JsExtension.cpp)

Extensions/PlatformBehavior/Extension.cpp
```diff
@@ -487,23 +501,26 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
...
         .SetIncludeFile("PlatformBehavior/PlatformerObjectRuntimeBehavior.h");
 
+    // Support for deprecated names:
+    aut.AddDuplicatedAction("SimulateReleaseKey", "SimulateReleasePlatformKey").SetHidden();
+
     aut.AddAction("SimulateControl",
                   _("Simulate control"),
                   _("Simulate a press of a key.\nValid keys are Left, Right, "
-                    "Jump, Ladder, Up, Down."),
+                    "Jump, Ladder, Release Ladder, Up, Down."),
```
4ian, line 523: Nitpicking: let's take the opportunity to convert the Key parameter from string to `stringWithSelector` (you pass a JSON array with the valid strings as a third argument.)

For example: 
https://github.com/4ian/GDevelop/blob/465629b6882e0ec039c969cca9d7bd97549b14d0/Extensions/TextObject/Extension.cpp#L436-L439

(4th argument can be skipped, it's false by default)

Extensions/PlatformBehavior/JsExtension.cpp
```diff
@@ -132,10 +132,12 @@ class PlatformBehaviorJsExtension : public gd::PlatformExtension {
           "simulateDownKey");
       autActions["PlatformBehavior::SimulateLadderKey"].SetFunctionName(
           "simulateLadderKey");
+      autActions["PlatformBehavior::SimulateReleaseLadderKey"].SetFunctionName(
+          "simulateReleaseLadderKey");
       autActions["PlatformBehavior::SimulateJumpKey"].SetFunctionName(
           "simulateJumpKey");
-      autActions["PlatformBehavior::SimulateReleaseKey"].SetFunctionName(
-          "simulateReleaseKey");
+      autActions["PlatformBehavior::SimulateReleasePlatformKey"].SetFunctionName(
```
4ian, line 139: Unfortunately you must also keep a line for the deprecated `SimulateReleaseKey` (Because when duplicating the action, the code generation information is not yet filled).

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -1581,7 +1602,6 @@ namespace gdjs {
     }
 
     beforeMovingY(timeDelta: float, oldX: float) {
-      const behavior = this._behavior;
```
4ian, line 1584: 👍

Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
```diff
@@ -1689,7 +1700,9 @@ describe('gdjs.PlatformerObjectRuntimeBehavior', function () {
       object.getBehavior('auto1').simulateLadderKey();
       climbLadder(10);
       stayOnLadder(10);
-      climbLadder(14);
+      releaseLadder(10);
```
4ian, line 1703: Nitpicking: could be interesting to add the `expect(object.getY()).to.be.within(` before the releaseLadder and after the 10 frames of releaseLadder.

(You can take the numbers given by the test itself, even if in theory we can compute them, the final position after the 10 frames being the initial Y position after `stayOnLadder(10)`, and 10 frames of descent that we could calculate using the given gravity).

This would robustify a bit more the test (we can imagine that in the future, we would have a bug where if you keep pressing the release ladder button, the object is stuck in the air or on the contrary get the gravity applied multiple time, etc...)

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -548,14 +550,23 @@ namespace gdjs {
...
+     * Mark the platformer object as not grabbing any platform.
      */
     _releaseGrabbedPlatform() {
       if (this._state === this._grabbingPlatform) {
         this._setFalling();
       }
     }
 
+    /**
+     * Mark the platformer object as falling.
```
4ian, line 562: Nitpicking: `Mark the platformer object as falling if on a ladder.`

(agree, the comment is now exactly the code, but it's a public interface so it's fine to comment it) (at least this means the code is crystal clear!)

4ian (approved):
Thanks! Good to go after the comments are resolved 👍

4ian (approved):

PR #2898, opened 2021-08-12
Fix color of grids in scene editor not properly persisted

### LINKED ISSUE #2813: Buttons not Pinned to bottom of screen in Object Group or Global Variable Lists, display grid colors not saved
I had already found one here #2755, but I waited for the new version before reporting

All images below show pretty much the same problem. We must go all the way to do something, be it delete, copy, or search and add. Only image 3 is another segment, but it's still visual.

**1 - Groups**

**2 - Global Variables. Object variables also exist.**

**3 - Grid Color**
gridcolor

How Reproduce this bug: Go to setup grid, and change the color of the grid lines to any other. Save your project, exit GD and open the project again. Color of grids back to default.

4ian:
* Also make the update real time (any change in the settings are shown directly in the editor)

Don't show the rest in the changelog:

* Add various Flow typings where missing (SetupGridDialog and various InstancesEditor files)
* Serialize the settings into a generic "EditorSetting" (so that the Project is agnostic of any editor related stuff)
* Rename uiSettings/options/LayoutEditorCanvasOptions to InstancesEditorSettings
  * Handle everything inside the IDE (so that the Project remains agnostic of any editor related stuff)
* Performance bonus: also avoid to always recompute the initial editor settings when re-rendering the tabs

Note that in the future, this kind of EditorSetting that is stored inside the project could be moved to its own structure (living outside of the project file).

Partially fix #2813 (sorry I got side tracked and refactored a bunch of things)


newIDE/app/src/InstancesEditor/Grid.js
```diff
@@ -1,50 +1,48 @@
...
 
-    if (!options.grid) {
+    if (!instancesEditorSettings.grid) {
       this.pixiGrid.visible = false;
       return;
     }
 
-    const gridColor = options.gridColor || DEFAULT_COLOR;
-    const gridHex = rgbColorToHexNumber(gridColor);
-    const gridAlpha = gridColor.a || 1;
```
ClementPasteau, line 42: not relevant any more ?

4ian: Not anymore, but I've added a disableAlpha in the color picker! :)

Issue #2640, opened 2021-05-12
[$250+ bounty] Filters/Effects/Shaders for Individual Objects [Feature Request]


Silver-Streak:
## Description

PixiJS Filters/Effects/Shaders have been implemented for a while now, but they can only be applied per layer. This works for some filters (CRT, BulgePinch, etc), but many times users will want to only impact a single object. 

There are workarounds such as moving the desired object to an empty layer that has the effect on it, but this becomes very complex and unmanageable once you get past 1-2 objects or different effects. 
- As an example, if you want the glow effect on the player character as well as a chest, but you want them to glow different colors, you now need two completely separate layers that are following the main layer's camera and have the different layer effect setups.

A per-object effect/filter/etc would be much better suited in these cases, and could open up additional possibilities for effects that don't make sense for a layer-wide filter.

This is related to the existing trello card here: https://trello.com/c/5U5dWRSd/34-support-for-visual-effects-shaders-for-objects

## Solution suggested

Allow for adding effects to objects just like we currently can for layers, in the object properties panel. While there *could* be a use case for adding/removing an effect via events, I think making it behave similar to layer effects where you add it to the panel and then can disable/enable it will be more uniform.

Wendigo has already provided a potential solution on the trello card:

> If someone wants to give it a go, here is a working example as a behavior.
> Pixi also supports multiple filters per drawable object / container by adding all instances as an array to their "filters" property. https://github.com/Wend1go/GDevelop/tree/objectfilters/Extensions/ObjectFilterBehavior

Pixi's documentation on filters is available here: https://pixijs.download/release/docs/PIXI.Filter.html 

While I cannot help implement this, I intend to open a bounty for this feature once confirmed we're good with this writeup.

## Alternatives considered

1. Educate users on workaround of moving objects to specialized layers for just that object. Document and detail all downsides, including performance, complexity, etc. Relatively complex and different behavior that users are expecting.
2. Do not allow for effects on single objects at all. Long-term limiting for the engine and puts it behind things like Phaser as far as flexibility.

Silver-Streak:
@4ian let me know if you have any concerns on the above or it looks good, and I can get a bounty opened up on it.

HarsimranVirk:
@4ian we can use a similar approach for sprite objects as we do for layers (PIXI Containers and PIXI Sprites are equivalent as far as filters are concerned). I can set up a prototype to check performance hits.

4ian:
[...]

PR #2901, opened 2021-08-12
⚙️[WIP] Adding Effects Manager in runtime.

HarsimranVirk:
Reopened PR #2736 
Context: see issue #2640 and [this Trello card](https://trello.com/c/5U5dWRSd/34-support-for-visual-effects-shaders-for-objects).

TODO: 
- [x] Help link (4ian)
- [x] Add some tests to create a runtime object, add an effect, change its parameters, delete the effect. (4ian)
- [x] Consider refactoring the layer effect handling with the new effect manager? (4ian)
- [x] ~~Fix issue with offsets~~ Bypassed for now because of Pixi issues
- [x] Tested on the webapp (4ian)
- [x] Tested on mobile (4ian)
- [x] Allow to skip updates for objects outside the view. We can do this because effects are updated just before the render, at the time we also check if an object should be visible or not on the screen (culling) (4ian)


4ian:
Thanks for reopening the PR! :)
So I think ideally we would have a few tests, otherwise we're good to go if functionally this works.
- @Bouh I think you checked this in the past, do you want to give another try? I think you listed some filters that were causing issues? Or maybe this is now addressed by the usage of the proper width/height of objects.
- @ClementPasteau this is a fairly large PR but addressing interesting stuff in both the C++ codebase, the runtime (game engine) and in the editor too. Feel free to take a look and drop comments if you have some!

Bouh:
Nice job on this long and complex feature!
Here new review from my own.



|Effect name  | Status ❌/✅  | Issue |	
|--|--|--|
|Adjustement  | ❌ | Blue parameter has to be 1 by default, not 0.6, also true for layer effect |
|Advanced bloom| ❌  | Require to add padding manually, otherwise, the effect is cropped  |
|ASCII  | ✅ |  |
|Beveled edges  | ✅ |   |
|Black and White  | ✅ |  |
|Blending mode  | ❌ | If the value has decimal the layer isn't rendered |
|Blur (Gaussian)  | ❌ | Limited on edges of the game? |
|Blur (Kawase)  | ❌  | Limited on edges of the game? |
|Brightness  | ✅ |  |
|Bulge Pinch  | ✅ |  |
|Color Map  | ✅  |  |
|Color Replace  |✅  |  |
|CRT  | ✅ |  |
|Drak Night  | ✅ |  |
|Displacement  | ❌ | Idk what really happened, I feel it act strangely |
|Dot  |✅  |  |
|Drop shadow  | ❌ | Require to add padding manually, otherwise, the effect is cropped |
|Glitch  | ❌ | For layer this effect is animated, currently isn't for Object, (see animation frequency parameter) |
|Glow  |✅  |  |
|Godray  |✅  | This effect works but doesn't like alpha channel (layer and object) |
|Light Night  | ✅ |  |
|Noise  | ✅ |  |
|Old Film  | ✅ |  |
|Outline  | ❌ | Require to add padding manually, otherwise, the effect is cropped |
|Pixelate| ✅  |  |
|Radial Blur  |❌  | Center X/Y aren't good and require to add padding manually, otherwise effect is cropped  |
|Reflection  |❌   | For layer this effect is animated, currently isn't for Object, (see animation speed parameter) |
|RGB split  |❌ | Require to add padding manually, otherwise, the effect is cropped |
|Sepia  | ✅ |  |
|Tilt shift  | ✅ |  |
|Twist  | ❌ | Offset X/Y aren't good, radius size seem doubled? |
|Zoom blur  |❌  | Center X/Y aren't good and require to add padding manually, otherwise effect is cropped


- Hint messages for the Object effect and the Layer effect are exactly the same.
- Still no helper like for behaviors when the list is empty, the help button on this frame opens the object effect page, while the help button on the dialog opens the help page of the object type.
 image

- An issue appears when I create multiples effects. Effects are spaced out.
This wasn't present in my previous review.
image

- Still no toggleable effects like behaviors:
image

- Some effects on objects are limited to the current view, like blurring effects. Here the effect is clamped to the view, must be bleeding offscreen through.
The project for reproducing: [limited_effects.zip](https://github.com/4ian/GDevelop/files/6981474/limited_effects.zip)

https://user-images.githubusercontent.com/1670670/129338646-e2e4db9a-8089-484a-974c-9dc967214625.mp4



In meanwhile, Outlines works fine
image

4ian:
I've rebased (and not merged) this on master, so please discard your local branch and fetch it again from GitHub if you want to continue working on it ;) 
I'll be pushing more changes/improvements :)

4ian:
I've asked on the PixiJS filters repo to see if it's normal for the Twist/Radial Blur/Zoom Blur to have a fixed/strange positioning of the effect according to the object position on the screen: https://github.com/pixijs/filters/issues/304 :)

HarsimranVirk:
> Add some tests to create a runtime object, add an effect, change its parameters, delete the effect.

@4ian I've tried writing some tests, but it turned out that the effects aren't registered properly while running tests (maybe because the files related to effects are added only after exporting the game?). The tests simply fail, after logging this: `LOG: 'Effect "testEffect" has an unknown effect type: "AdvancedBloom". Was it registered properly? Is the effect type correct?'`.

4ian:
> maybe because the files related to effects are added only after exporting the game

More or lessyes, there is no effects because there are no built-in in the game engine (it's only part of an extension) and none are imported in karma.conf.js (which is listing all the files to import for the tests - like what an exported game would do).
I'm adding some tests now with an effect :) This means I will import an effect in karma.conf.js and then use it in the tests :)

4ian:
Tests are there! :) 
I've grayed for now the 3 effects that have issues with PixiJS on objects that are not the size of the window (not the best solution, but at least we can progress).
I'm tempted to go ahead like this - even if some effects are not perfect. I'll write the wiki page about object effects and add the help link.

4ian:
Finished by fixing the timing of the update of object effects: it's now done before the rendering (so effects get a chance to update), I've renamed the method `updatePreRender` to make this extra clear, and it's only called on objects visible on screen (for efficiency, and because these are "effects" so no need to call the update on all objects living in the scene).

Should be good to go now! :)

PR #2902, opened 2021-08-12
Update Readme.md

tberthel:
I had to search for the build info.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -12,6 +12,7 @@ GDevelop is a full-featured, open-source game development software, allowing to 
 | Contribute to the editor        | Download [Node.js] and follow this [README](newIDE/README.md).                 |
 | Create/improve an extension     | Download [Node.js] and follow this [README](newIDE/README-extensions.md).      |
 | Help to translate GDevelop      | Go on the [GDevelop project on Crowdin](https://crowdin.com/project/gdevelop). |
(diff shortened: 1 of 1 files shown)


4ian:
Thanks for your PR. I think the sentences don't really fit in the current README table, which is just a getting started place. It should rather be placed in this README: https://github.com/4ian/GDevelop/blob/master/newIDE/README.md :)

tberthel:
The root readme should have something on how to build or where to find the build instructions? right?

4ian:
Yep, the link is here: 
image

tberthel:
So can it become "Contribute to the editor or build"?

tberthel:
?

4ian:
Sorry, I think `Contribute to the editor` is fairly clear? "Build" means different things according to the context, so I think it's better to keep this using a simple word, "editor" that is clear and representing the whole app :)

tberthel:
It took me a long while find the instructions to build (find and grep).  So whatever you have is not build instruction friendly.

tberthel:
Contributing to me just means one of two things: 

- Opening a pull request with changes
- Pushing to the repo directly

Building and Running are also not the same.  If your project had a make file or pom.xml in the root of the project it would be less of a problem, but you have neither.

PR #2932, opened 2021-08-17
Prevent variable value to be set to NaN

arthuro555:
This prevents setting a variable to NaN. It technically should not be a problem as GetValue transforms a NaN value to return 0.0, but NaN can still break internal methods that use `value` directly in unexpected ways.
Additionally, it caused a weird behavior when converting a string or allowing to have no number until serializing and deserializing again in a variable field.
Before: 

https://user-images.githubusercontent.com/19349038/129705244-462b2d95-cf8a-423c-bc59-54efdce4296c.mp4

After: 

https://user-images.githubusercontent.com/19349038/129705859-40a98b0b-1914-4c25-91c7-15d6583b1901.mp4

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -108,7 +108,8 @@
         "xtree": "cpp",
         "xutility": "cpp",
         "xlocbuf": "cpp",
-        "xlocmes": "cpp"
+        "xlocmes": "cpp",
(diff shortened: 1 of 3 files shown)


4ian:
Sounds safer, thanks!

blurymind:
Thank you :)

PR #2933, opened 2021-08-17
Add translation in menu bar

Bouh:
Fix a part of https://github.com/4ian/GDevelop/issues/969#issuecomment-899880659


Bouh:
The remaining items are non-updated POT files.

4ian:
Thanks! These are Electron special menu items, I was thinking they would be handled by the system. But seems that it's not the case, it's actually hardcoded in Electron (https://github.com/electron/electron/blob/888ac65c72df227ed4d65d9177748d4082b13b0f/lib/browser/api/menu-item-roles.ts)

PR #2953, opened 2021-08-20
Add missing translation tags

Bouh:
Fix these reports https://github.com/4ian/GDevelop/issues/969#issuecomment-899880659

Think to bundle the new translations files from Crowdin in the next build. Translations have not been updated for the latest versions.

### DIFF
--- a/newIDE/app/src/InstructionOrExpression/EnumerateExpressions.js
+++ b/newIDE/app/src/InstructionOrExpression/EnumerateExpressions.js
@@ -1,6 +1,4 @@
 // @flow
-import React from 'react';
-import { Trans } from '@lingui/macro';
 import {
   type EnumeratedExpressionMetadata,
   type InstructionOrExpressionScope,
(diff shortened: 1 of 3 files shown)


newIDE/app/src/InstructionOrExpression/EnumerateExpressions.js
```diff
@@ -23,10 +25,10 @@ const getExtensionPrefix = (extension: gdPlatformExtension): string => {
   const allBehaviorsTypes = extension.getBehaviorsTypes();
 
   if (allObjectsTypes.size() > 0 || allBehaviorsTypes.size() > 0) {
-    return (
-      (extension.getName() === 'BuiltinObject'
-        ? 'Common expressions for all objects'
-        : extension.getFullName()) + GROUP_DELIMITER
+    return extension.getName() === 'BuiltinObject' ? (
+      <Trans>Common expressions for all objects</Trans>
```
4ian, line 29: This fix is more complex, it will involve either passing the translated string as an argument, or passing down the translation function+macro, as you're supposed to return a string, not a React Node.

4ian (approved):

4ian:
Thanks!

Issue #2863, opened 2021-08-03
Clean up the old Cocos2D-JS renderers and related code


4ian:
Similar to https://github.com/4ian/GDevelop/issues/2708 (GDCpp/SFML that could be removed), we're not using or supporting properly Cocos2d-JS renderers anymore.

This means that we can remove:
- In Extensions: all the `*cocos-renderer*` files.
  - and all the addIncludeFile/setIncludeFile mentioning them in JsExtension.js files.
- In GDJS: the `GDJS/Runtime/Cocos2d` folder and `GDJS/Runtime/cocos-*` folders.
- the exporter code that copies cocos files.

We will continue to keep the architecture that separate the `gdjs.RuntimeObject` from their renderer. This is useful if we want to introduce a new rendering engine in the future, or if we want to make a "headless" version of games (for multiplayer servers for example).

HarsimranVirk:
Is someone working on it? I'd like to solve this issue.

4ian:
Feel free to go ahead! Will be good to clean this up as this it totally unmaintained :)

PR #2956, opened 2021-08-20
🔨 Removing cocos renderer files

HarsimranVirk:
Solves #2863


newIDE/app/src/Export/ExportDialog.js
```diff
@@ -211,23 +243,11 @@ export default class ExportDialog extends React.Component<Props, State> {
...
-                            primary={false}
-                            onClick={() => this._showExperimental(false)}
-                            label={<Trans>Hide experimental exports</Trans>}
-                          />
-                        )}
+                        <ExperimentalExportButton
+                          showExperimental={showExperimental}
+                          onClick={value => this._showExperimental(value)}
+                          disabled
+                        />
```
HarsimranVirk, line 250: Extracted this out, as at the moment, cocos2d was the only experimental export. If in future we decide to add more experimental exports, this should come in handy.

4ian (approved):
I think this looks good! 

Before we go ahead, can you check:
- You can export a game (try the Android one click packaging, web export and the Windows export for example).
- No bug/crash on a few different examples

Just want to be sure we've not removed something that was necessary somewhere, but I think it's fine! :)

Bouh:
- Android one click, [web export](https://games.gdevelop-app.com/game-6fd11cdb-a3ad-4aac-a58c-bea054be49d9/index.html) and windows export are ok.
- No bug on examples.

 .map files are missing in web export but seem not related to this PR.


It's a go!

4ian:
> .map files are missing in web export but seem not related to this PR.

Indeed, unrelated to the PR (and harmless, though not having these warning would be better).

Merging then! 🎉

Issue #2664, opened 2021-05-19
Round coordinates bug


VegeTato:
## Describe the bug
Before the latest update comes
using platformer behavior by Default the [Round coordinates] comes checked and its working fine (before the update)
for example: if playerX = 142(for example)  > quit the game
^this worked great before the updates come, when the player reaches 142X the game will quit
But now after the update, the event above^ is not working anymore because (Round coordinates) check is not working anymore, if its checked(by default it is) the game should take the platformer position as a number without commas, but now it Always take the number with Comma if the Round coordinates checked or not
as showing in the example here:
![bugfound](https://user-images.githubusercontent.com/81715237/118853058-06b87580-b8dc-11eb-84e3-771606f872d9.gif)

## To Reproduce
1. Go to events
2.  if player at Xpos or Ypos(both not working) = 142(for example)
3. quit game
4. Not gonna quit the game because the number is going to be like 142.51858...etc it will never comes to 142

Example project (its simple):
[Bug.zip](https://github.com/4ian/GDevelop/files/6510514/Bug.zip)

## Other details
* using windows
* Desktob app, latest version of Gdevelop beta 110

Silver-Streak:
I went back as far as b89 with the example provided, and it seems to behave the same as b110. That said, I would say "same" is also "incorrect". The round coordinates flag on the platformer character behavior seems to be broken.

 So I don't think this has ever worked the way VegeTato expected (or at least not reliably) but I _do_ think it should work as they expected. Round coordinates should actually round the coordinates, and it doesn't seem to be doing so.

b98 ![image](https://user-images.githubusercontent.com/24237149/118854954-e1efde00-b89a-11eb-912d-7176a538faf5.png)

b89 
![image](https://user-images.githubusercontent.com/24237149/118855832-de108b80-b89b-11eb-8848-45ed15b91e6c.png)

Both give the same result, it never detects the character as reaching 315. 

Using "Compare Two Numbers" and round(player.X()) = 315 as the condition does work, thoguh.

Silver-Streak:
Bringing this back up to the crew @4ian and all. 

The pixel rounding tag on the platformer behavior definitely is not rounding positions, and this is leading to an increased number of posts on the forums with users looking for solutions.

While long term I'd think exposing the project-wide pixel rounding options (my request here: https://github.com/4ian/GDevelop/issues/2504) 

[...]

Issue #2504, opened 2021-04-03
[$125 bounty] Add a project setting for Pixel Rounding - To enable Project-wide Pixi.settings.ROUND_PIXELS


Silver-Streak:
## Description

I would like to see a new checkbox toggle/setting added to enable Pixel Rounding for your project. 

As discussed here: https://forum.gdevelop-app.com/t/project-wide-toggle-for-pixi-settings-round-pixels/31673

## Solution suggested
I would like to see a new setting (checkbox or otherwise) added to the Project Properties panel. Enabling this setting would enable the Pixi.Settings.ROUND_PIXELS flag for the entire project. (https://pixijs.download/release/docs/PIXI.settings.html#ROUND_PIXELS)

We already have a "round pixels" toggle for the platformer behavior, but that only impacts that specific object. This toggle would ensure that everything is impacted, including camera, positioning, and movement, not just for objects. As mentioned in Pixi's documentation:
 “Advantages can include sharper image quality (like text) and faster rendering on canvas. The main disadvantage is movement of objects may appear less smooth.”. 

Eliminating subpixel rendering can be very beneficial for any project, but especially for those with retro-style movement and graphics.

## Alternatives considered

Document where and how people can manually modify their various exported .js files to enable the toggle. Not a recommended solution as it can lead to error and be highly confusing for new users.

D8H:
I heard something about pixel perfect. I think It's a rounding at the big pixels and it's not the same as pixel rounding + scaling because effects are usually rendered on native pixels.
Maybe this is something to consider.

Silver-Streak:
> I heard something about pixel perfect. I think It's a rounding at the big pixels and it's not the same as pixel rounding + scaling because effects are usually rendered on native pixels.
> Maybe this is something to consider.

Not sure what you're referring to, as far as what I can find in Pixi's documentation, there isn't pixel perfect. Pixel Rounding as it exists (and as I'm requestion) eliminates subpixel (float) x/y positions and width/height. This makes it behave more natually to old 2D systems like NES/SNES/ETC.

If there's something else that you're aware of/referring to, I'd love to see more details on it.

D8H:
Yes, I don't know about Pixi.
I think the general idea of pixel perfect is that the big pixels of the sprites are all aligned. For instance, a 4x4 big pixel can't be on the screen at (23 ; 47) because the purpure is to fake low resolutions for the pixel art part.
[...]

PR #2961, opened 2021-08-23
Fix/round coordinates checkbox

ClementPasteau:
This addresses 2 issues:

- #2664 by removing the object property for the platformer behavior which wasn't working
- #2504 by adding a project property, allowing the user to round the pixels when rendering (using `PIXI.settings.ROUND_PIXELS`)

Tested in the runtime console, by logging `PIXI.settings.ROUND_PIXELS` with the switch on and off.
Tested in multiple test games, no bug encountered

### DIFF
--- a/Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
+++ b/Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
@@ -471,7 +471,6 @@ describe('gdjs.PlatformerObjectRuntimeBehavior', function () {
             ignoreDefaultControls: true,
             slopeMaxAngle: 60,
             jumpSustainTime: 0.2,
-            roundCoordinates: true,
           },
         ],
         effects: [],
@@ -492,7 +491,7 @@ describe('gdjs.PlatformerObjectRuntimeBehavior', function () {
       }
 
       //Check the object is on the platform
-      expect(object.getY()).to.be(-30); // -30 = -10 (platform y) + -20 (object height)
+      expect(object.getY()).to.be(-30.75); // not exactly -30 because not rounding coordinates (-10 (platform) -20 (object))
       expect(object.getBehavior('auto1').isFalling()).to.be(false);
       expect(object.getBehavior('auto1').isFallingWithoutJumping()).to.be(
         false
@@ -511,11 +510,11 @@ describe('gdjs.PlatformerObjectRuntimeBehavior', function () {
       }
 
       // Check that we reached the maximum height
-      expect(object.getY()).to.be.within(-180, -179);
+      expect(object.getY()).to.be.within(-181, -180);
       runtimeScene.renderAndStep(1000 / 60);
-      expect(object.getY()).to.be(-180);
+      expect(object.getY()).to.be(-180.75);
       runtimeScene.renderAndStep(1000 / 60);
-      expect(object.getY()).to.be.within(-180, -179);
+      expect(object.getY()).to.be.within(-181, -180);
 
       // Then let the object fall
       for (let i = 0; i < 17; ++i) {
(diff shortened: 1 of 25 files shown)


GDJS/Runtime/runtimegame.ts
```diff
@@ -427,6 +429,14 @@ namespace gdjs {
       return this._scaleMode;
     }
 
+
+    /**
+     * Return the scale mode of the game ("linear" or "nearest").
```
4ian, line 434: Comment to update ;)

Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
```diff
@@ -492,7 +491,7 @@ describe('gdjs.PlatformerObjectRuntimeBehavior', function () {
       }
 
       //Check the object is on the platform
-      expect(object.getY()).to.be(-30); // -30 = -10 (platform y) + -20 (object height)
+      expect(object.getY()).to.be(-30.75); // not exactly -30 because not rounding coordinates (-10 (platform) -20 (object))
```
4ian, line 494: @D8H We're considering removing the "roundCoordinates" option because it was only applying in a specific case (when on the floor). What do you think of doing this?

D8H: Indeed, I think this property only purpose is to try to avoid sprite blurring and don't really do a good job.
It'll be really good to get rid of it because it's hard to cover every case for both values of this property.

About this assertion, I think the platformer character should be aligned with the platform. Because in this case in particular, it will probably be rounded to -31 by the render and overlap the platform from one pixel.

I wrote something that could fix this for the moving platform issue. Please, can you cherry pick this commit and see if it fixes the tests: https://github.com/4ian/GDevelop/pull/2574/commits/ba3bf44791677f3e633c5648775750000d92b1b3

Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
```diff
@@ -2127,7 +2109,7 @@ describe('gdjs.PlatformerObjectRuntimeBehavior', function () {
       platform2.setAngle(-30);
       platform2.setPosition(
         platform.getX() + platform.getWidth() + vertexDeltaX,
-        platform.getY() + vertexDeltaY
+        platform.getY() + vertexDeltaY - 0.5
       );
```
D8H, line 2113: I remember checking with IDE that the rotated platform was corner to corner with the other platform. If the test only works when the platform is lowered a bit, it's revealing of an issue. The platformer character will probably be stuck.

4ian: True, changing this might be concerning. Or maybe this was just to ensure the position is ok after that, in which case @ClementPasteau it would be better to not move the platform, and instead adapt the position of the object.

In other words, the changes in the tests should not have any change in the position of the platforms, because levels of games won't change. There might a slight change in the object position (due to the rounding being gone), but that should be it (in theory at least).

Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
D8H: Actually there is other commit that do side effect on this too, I'll check them to see which ones to use.

D8H:
I checked. These 3 commits are necessary to make the character align to the platform
* Make flat floor following more predictable: https://github.com/4ian/GDevelop/pull/2574/commits/6e78bdfdd55461a2520f4f0eae5bba58bbfb2952
* Cleaner way of doing the collision margin error: https://github.com/4ian/GDevelop/pull/2574/commits/0eedb5de4efe82c377e6fa581f7b50a59fc35ee8
* Fix the gap between the moving platform and the player: https://github.com/4ian/GDevelop/pull/2574/commits/ba3bf44791677f3e633c5648775750000d92b1b3

They come from this PR: https://github.com/4ian/GDevelop/pull/2574

ClementPasteau:
Thanks for checking @D8H !

I wasn't aware of the amazing work you were doing to improve the whole Platformer movements :)
As you mention, even if it makes sense to get rid of this property which was confusing, we should probably wait for your work to be merged before doing so. It should hopefully prevent us from modifying some of the tests.

I can see that you've split your PR into 3, to merge in smaller chunks. I suspect that all 3 are needed before resuming work on this one?
I want to help merging the work you've started, only #2602 is ready to review for the moment?

This PR was trying to do 2 things at once, I'll split the "Project Pixel Rounding" property to another PR so we can merge it, and leave this one to remove the checkbox only.

4ian:
Agree on splitting this! Pixel rounding can easily go ahead now. For this roundCoordinates, better to see if we can fix the precise positioning (and maybe things like moving up/down) before removing this.

For  #2602 #2603 and #2574, it's really the first PR that was blocking me - as I'm not super comfortable with updating the AABB/Rbush more than once (because this adds complexity that will make it harder to evolve this RBush into something like a potentially WebAssembly powered collision engine in the future).
If somehow we can revisit this but with just targeted changes to handle 1) moving properly on a floor platform going up/down (and potentially any direction!) and 2) the precise positioning of Y coordinate on a floor, that would be absolutely great (not sure if this was tied to the AABB rework or not!)

D8H:
> For #2602 #2603 and #2574, it's really the first PR that was blocking me - as I'm not super comfortable with updating the AABB/Rbush more than once (because this adds complexity that will make it harder to evolve this RBush into something like a potentially WebAssembly powered collision engine in the future).
> If somehow we can revisit this but with just targeted changes to handle 1) moving properly on a floor platform going up/down (and potentially any direction!) and 2) the precise positioning of Y coordinate on a floor, that would be absolutely great (not sure if this was tied to the AABB rework or not!)

Ok, I'll take a look after I've done a 1st version of a Marching Square extension.

ClementPasteau:
closing this PR as it is being worked on in #3009 :)

PR #2962, opened 2021-08-23
Remove unnecessary translations for main language EN

ClementPasteau:
Do not show in Changelog


ClementPasteau:
This may require a bit more investigation, I've noticed the translations are loaded correctly, however there is an error in the console: 
image

4ian:
I think this is required, because js-lingui will need some translations to be compiled as functions (for the translations containing dynamic parts)

ClementPasteau:
Closing this, this was more a misunderstanding from my end :)

PR #2964, opened 2021-08-23
Add choice selector for behavior

Bouh:
As explained in https://github.com/GDevelopApp/GDevelop-extensions/issues/188#issuecomment-903656981

There is a bug in this PR.
`newExtraInfo` return `"["New Option"]"` for behaviors.

@arthuro555 you integrated that for function previously, any idea?

### DIFF
--- a/newIDE/app/src/EventsBasedBehaviorEditor/EventsBasedBehaviorPropertiesEditor.js
+++ b/newIDE/app/src/EventsBasedBehaviorEditor/EventsBasedBehaviorPropertiesEditor.js
@@ -22,6 +22,7 @@ import Visibility from '@material-ui/icons/Visibility';
 import VisibilityOff from '@material-ui/icons/VisibilityOff';
 import Add from '@material-ui/icons/Add';
 import { ResponsiveLineStackLayout, ColumnStackLayout } from '../UI/Layout';
+import { StringSelectorEditor } from '../EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor';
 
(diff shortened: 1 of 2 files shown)


arthuro555:
After taking a quick look, I don't see anything that looks inherently wrong, but I can take a closer look in a few hours when I am home. Though I am not sure I understand what the problem is.

newIDE/app/src/EventsBasedBehaviorEditor/EventsBasedBehaviorPropertiesEditor.js
```diff
@@ -267,6 +272,15 @@ export default class EventsBasedBehaviorPropertiesEditor extends React.Component
                                 />
                               </SelectField>
                             )}
+                            {property.getType() === 'Choice' && (
+                              <StringSelectorEditor
+                                extraInfo={property.getExtraInfo()}
+                                setExtraInfo={newExtraInfo => {
```
4ian, line 278: Ah good catch, I did not see that StringSelectorEditor was made for parameters, which store their additional information into a string, using `setExtraInfo`. Properties have the option to store multiple extra info in an array (a "vector" in C++). So we must adapt this. I would suggest to rework StringSelectorEditor to work with a JS array, and have the code calling it to handle the parsing from/to a JSON array (for parameters) or from/to `getExtraInfo`/`addExtraInfo` (for properties).

@ClementPasteau will look at this I think! :)

ClementPasteau:
@Bouh I've taken your branch and continued your work on another one, so I recreated a PR #2968. Closing this one!

PR #2965, opened 2021-08-24
Allow rounding pixels when rendering on a project level

ClementPasteau:
Addresses #2504 

image

Tested in the runtime console, by logging PIXI.settings.ROUND_PIXELS with the switch on and off.
Tested in multiple test games, no bug encountered

### DIFF
--- a/Core/GDCore/Project/Project.h
+++ b/Core/GDCore/Project/Project.h
@@ -277,6 +277,19 @@ class GD_CORE_API Project : public ObjectsContainer {
   void SetScaleMode(const gd::String& scaleMode_) { scaleMode = scaleMode_; }
 
   /**
+   * Return true if pixels rounding option is enabled.
+   */
+  bool GetPixelsRounding() const {
+    return pixelsRounding;
+  }
+
+  /**
+   * Set pixels rounding option to true or false.
+   */
+  void SetPixelsRounding(bool enable) { pixelsRounding = enable; }
+
+
+  /**
    * \brief Return if the project should set 0 as Z-order for objects created
    * from events (which is deprecated) - instead of the highest Z order that was
    * found on each layer when the scene started.
@@ -890,6 +903,8 @@ class GD_CORE_API Project : public ObjectsContainer {
                         ///< are below this number )
   bool verticalSync;    ///< If true, must activate vertical synchronization.
   gd::String scaleMode;
+  bool pixelsRounding; ///< If true, the rendering should stop pixel interpolation 
+                       ///< of rendered objects.
   bool adaptGameResolutionAtRuntime;  ///< Should the game resolution be adapted
                                       ///< to the window size at runtime
   gd::String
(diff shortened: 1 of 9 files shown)


newIDE/app/src/ProjectManager/ProjectPropertiesDialog.js
```diff
@@ -483,6 +491,16 @@ function ProjectPropertiesDialog(props: Props) {
                 primaryText={t`Nearest (no antialiasing, good for pixel perfect games)`}
               />
             </SelectField>
+            <Checkbox
+              label={
+                <Trans>
+                  Round pixels when rendering, useful for pixel perfect games.
+                </Trans>
+              }
+              disabled={sizeOnStartupMode === ''}
```
4ian, line 500: disabled to remove?

Core/GDCore/Project/Project.h
```diff
@@ -890,6 +903,7 @@ class GD_CORE_API Project : public ObjectsContainer {
                         ///< are below this number )
   bool verticalSync;    ///< If true, must activate vertical synchronization.
   gd::String scaleMode;
+  bool pixelsRounding; ///< If true, set PIXI's ROUND_PIXELS option to true
```
4ian, line 906: Nitpicking: In the `Project` and in `Core` in general, we usually avoid mentioning PIXI. The idea is that the `Core` describe a game (project, scenes, events, etc...) and tools to manipulate it (for example, a class like `WholeProjectRefactorer` that allows you to rename an object in the whole project after the user rename it in a layout, this kind of "quality of life" stuff). `GDJS` uses `Core` to implement JavaScript specific stuff (code generation mainly! And setting the function names).
And finally, the `GDJS` game engine is the one implementing the runtime of a game that was code generated. In all of this PixiJS is... almost an implementation detail :) Well not really in GDJS (we know that we must copy the PixiJS files and renderer files), but in `Core` it should be agnostic of it.

This could be useful should we in the future choose to:
- build another game engine not based on JS (unlikely though in the short term, too much duplicated work, not super agile).
- do rendering using something else than PixiJS. While this can be unlikely for games, there is I think a real possibility that we in the future create a "empty renderer" or "memory renderer" that render games on... nothing. To be able to run a game on something like a Node.js headless server (useful if you want to make a multiplayer game server).

All of this to say, something like "If true, the rendering should stop pixel interpolation of rendered objects." ;)

4ian (approved):

Bouh:
@Silver-Streak  are you interested in a test build with this change?

Silver-Streak:
@Bouh While my time is limited right now, I can do some testing as needed, absolutely.

Bouh:
@Silver-Streak No rush, it's just for fun, I know you've been waiting for it for a long time.
It's also great to get feedback before we release the new version.

[Download](https://mega.nz/file/UGQAmD6D#Fe7rpUUnyUkDWdj23-k_oMpDWdIGF4yZDMmS1aLrTkA)

Issue #2211, opened 2021-01-14
Improvements to behavior properties (for developing extensions)


tristanbob:
## Description

Behavior properties can be improved

## New features

- [ ] Add "String from a list of options (text)" as a Property type
- [ ] Add "Color picker" as a Property type
- [ ] Add "Object" as a Property type

## Problems

- [ ] Properties can currently be sorted by the developer, but the order does not match what the user sees

4ian:
Color should not be too bad to add, as it's already supported in the properties:
  * Update in `MetadataDeclarationHelpers.js`, in `generatedBehavior.updateProperty`, to support a new "Color" (string).
  * Same in `getProperties` and `initializeContent`.
  * Add Color in `EventsBasedBehaviorPropertiesEditor.js`

A "String from a list of options" can be done (called "Choice" internally), but will need a bit of work in the UI to be able to enter the list of options.
 
"Object" is more complex and maybe not how we want to deal between behaviors and objects (which object would get selected? How would you access it from the events sheet?).

tristanbob:
I don't have a good grasp of how objects would be used in a behavior property, so I will cross it out.

chauhanprakhar:
I would like to work on this issue if it hasn't assigned to anyone.

4ian:
Feel free to take a look :)

85ayush:
I would love to work on this issue, I am new to open source contribution, can you please guide me?

arthuro555:
I was looking into colors and there is just a slight issue making this harder: color behavior properties are stored as hex codes (`#ffffff`) but GDevelop events use rgb (`255;255;255`). Should I just let the users do the conversion, do it automatically when getting the property, or add a new `rgbColor` type to the properties editor for this color encoding?

Bouh:
There are tools in GDJS to convert all formats.

arthuro555:
> There are tools in GDJS to convert all formats.

I am aware, i am asking which way (not doing anything, using those tools at runtime or use the correct format to begin with) is the preferred one.

> The hex code is used in color picker but directly convert to "255;255;255".

That is not the case and why i brought it up here. I already added the color and could observe that the default color (which was in `r;g;b` format) worked while even picking anything that wasn't the default didn't. I looked at the data in `gdjs.projectData` and into Properties Editor and sure enough the format is `#rrggbb`.

4ian:
We definitely want to avoid people having to do the conversion by themselves. 
[...]

PR #2968, opened 2021-08-24
Add properties in behaviors

ClementPasteau:
Based on the existing PR #2964 

Addresses enhancement listed in https://github.com/GDevelopApp/GDevelop-extensions/issues/188#issuecomment-903656981 and #2211 

image

### DIFF
--- a/Core/GDCore/Project/NamedPropertyDescriptor.h
+++ b/Core/GDCore/Project/NamedPropertyDescriptor.h
@@ -70,12 +70,6 @@ class GD_CORE_API NamedPropertyDescriptor : public PropertyDescriptor {
   virtual void UnserializeValuesFrom(const SerializerElement& element);
   ///@}
 
-  /**
-   * Return a PropertyDescriptor from this NamedPropertyDescriptor,
-   * slicing the name.
-   */
-  PropertyDescriptor ToPropertyDescriptor() { return PropertyDescriptor(*this); }
-
  private:
   gd::String name;  ///< The name of the property.
 };
--- a/Core/GDCore/Project/PropertyDescriptor.h
+++ b/Core/GDCore/Project/PropertyDescriptor.h
@@ -77,6 +77,14 @@ class GD_CORE_API PropertyDescriptor {
   }
 
   /**
+   * \brief Set and replace the additional information for the property.
+   */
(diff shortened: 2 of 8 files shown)


newIDE/app/src/StringSelectorEditor/index.js
```diff
@@ -0,0 +1,66 @@
+// @flow
```
4ian, line 1: Let's rename this to StringArrayEditor? Also I wonder if this can be in Utils/ as it's fairly generic.

(Selector was too linked to its previous usage)

4ian (approved):

newIDE/app/src/StringSelectorEditor/index.js
ClementPasteau: Renaming it is a good idea :+1:

For the location of the file, If I follow the current logic, `ObjectTypeSelector` and `BehaviorTypeSelector` live in `/src` - whereas `SelectField`, `SemiControlledTextField` live in `/src/UI`
I'm tempted to move it to `src/UI` too (next to `LocalFilePicker`, `ColorField`...)

ClementPasteau: As discussed, leaving it in `/src` as this is a component and not really a util, and it's quite specific to GDevelop, so cannot be fully considered as a "generic" UI component

arthuro555:
Nice! 🎉 

I think it would be pretty nice to also port the other parameter types from functions to behavior properties.

Bouh:
Yeahh thx Clement! 🤸‍♂️

> I think it would be pretty nice to also port the other parameter types from functions to behavior properties.

🤞

arthuro555:
Actually... Since behavior fields cannot have expressions in them, shouldn't we have an "selector drop-down" there instead of an autocompleteable string?

4ian:
Behavior fields are indeed not the same thing as parameter fields. They are based on `PropertiesEditor`, like object properties, and are not evaluated at runtime (they are value, while parameters are expressions).

So there is not necessarily a 1:1 mapping between them (and this is visible in the code as there ParameterFields and PropertiesEditor, these are two separate concepts), even if support for new property types like color could be very useful :)

PR #2972, opened 2021-08-25
Modify Behavior Properties color picker to handle R;G;B format by def…

ClementPasteau:
…ault

And still allow Hex values


ClementPasteau:
Still a wip, addressing #2211. The goal was to unify the 'color' property to always handle "R;G;B" format by default, to be consistent with what is happening on the Event Properties.
It required modifying the Color component used in the 3 extensions Light/BBText/BitmapText and a bit of refactoring.
The next step is to then use this component when creating a behavior and defining its properties.

PR #2974, opened 2021-08-26
Add a FontSize expression for Text Object

Bouh:
Based on FontSize expression from Bitmap Text and BBText.

### DIFF
--- a/Core/GDCore/Extensions/Metadata/ObjectMetadata.cpp
+++ b/Core/GDCore/Extensions/Metadata/ObjectMetadata.cpp
@@ -302,6 +302,22 @@ ObjectMetadata::AddExpressionAndConditionAndAction(
       expression, condition, action);
 }
 
+gd::InstructionMetadata& ObjectMetadata::AddDuplicatedAction(
+    const gd::String& newActionName, const gd::String& copiedActionName) {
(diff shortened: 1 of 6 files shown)


Bouh:
The `FontSize` expression works.
The old action and condition are hidden but the new one is not shown.
Did I miss something?

4ian:
I've rebased on master and pushed the fixes:
- When using `AddExpressionAnd[...]`, the actions/conditions are *scoped*, meaning that their fullname is `ExtensionName::ObjectName::ActionName`, instead of just `ExtensionName::ActionName`. So the JsExtension.cpp file was needing changes
- You need to call AddDuplicatedAction/AddDuplicationCondition after adding the thing you're duplicating - and with the fullname (but without the extension name - this is confusing yeah).
- When you use these, you actually need to remove the old actions/conditions, that's the goal of using them ;) 
- You don't need to enter "the font size of _PARAM1_" => _PARAM1_ was not valid, and anyway it's not needed because the goal of `AddExpressionAnd[...]` is that you don't have to enter this `of _PARAM0_`, it's automatically generated. You just enter the name of the "property".

Should work well now, I've tested both the old and new actions/conditions/expressions :)

PR #2996, opened 2021-08-30
Clean old examples JSON files

4ian:
Not used anymore (the web-app uses examples from resources.gdevelop-app.com)

Don't show in changelog


Bouh (approved):
Yup!

Issue #1920, opened 2020-08-13
Collision Detection when on top of platform broken/inconsistent


Silver-Streak:
## Describe the bug
Discord user TripTune reported some issues with collision detection when trying to make falling platforms.

At first I thought I couldn't reproduce their issue, but instead I found that collision detection (or at least the condition) is very inconsistent if your character/object is on top of the platform. 

## To Reproduce
1. Load the normal platformer example.
2. Add a text object "Collide".
3. Set up events to pin the text object to the player character.
4. Add an event to update the text object to say "IsColliding" when in collision with the Jumpthru platforms.
5. Add an event to update the text object to say "NotColliding" when not in collision with the Jumpthru platforms.
6. Preview the example. For me, the first jumpthru platform will reliably update the text, the other 2 immediately after will not. Example project attached.
[CollisionIssue.zip](https://github.com/4ian/GDevelop/files/5069165/CollisionIssue.zip)

* Please include a link to a game if possible!
* If applicable, add screenshots to help explain your problem.

## Other details
* Win 10 Pro
* B98
* I tried older versions back as far as b92, all exhibit the same issue. I also tried resizing the objects or changing their platform type, with no difference. I also tried "For each" events, with no difference.

If I run back and forth on the platform a bit, it will sometimes change detection, but this definitely seems like something going weird with the condition, especially since the actual collision (walking on them) is working normally.

D8H:
The platformer extension solves the collision internally and ideally a platformer object should never be in collision with a platform but it can happen because of approximation or some not intentional effects (bugs or harmless).
It's not a good idea to rely on the general collision condition. A custom condition could be implemented but I think collisions can be detected using other objects.

D8H:
Does these PR fix the issue?

Can this condition help to do what you want?
* https://github.com/4ian/GDevelop/pull/3087

Does this PR make the general collision condition more reliable?
* https://github.com/4ian/GDevelop/pull/3009

AlexandreSi:
I checked and there are still glitches or unexpected displays with the project @Silver-Streak sent.

I think that @D8H summed it up well, and PR #3087 should be the way to make falling platforms work correctly. Here is an example:

![demo_platform_collision_fixed](https://user-images.githubusercontent.com/32449369/153888988-45f0b7d8-3253-4002-89e9-ce13aff4e48d.gif)


But such a request from users feel legit, so I think a diagnostic report tool should report it.

PR #3009, opened 2021-09-05
[Platformer] Remove the rounding property

D8H:
## Solution
The pixel by pixel loop to set the object on the floor is replaced by a method that finds the floor position from platforms hit-boxes.

On a flat floor without pixel rounding, there was [0;1[ pixel interstice between the object and the platform. Now, if the platform is pixel align, the object will be too which make the rounding property useless.

The object will always land at the exact same Y for a give X on a platform. This will make issues easier to reproduce and tests stronger.

This PR seems to fix these issues
* https://github.com/4ian/GDevelop/issues/3013
* https://github.com/4ian/GDevelop/issues/897

Even if this issue is not a bug, it kind of fix it for flat floors as the collision will always be false with excluding edge and true if including them.
* https://github.com/4ian/GDevelop/issues/1920

It doesn't fix
* https://github.com/4ian/GDevelop/issues/2562

## Automatic tests
The tests stay the same apart from 2 places where odd one frame state doesn't appear anymore.
I compared the coordinates on the test with a slope, the values have less than one pixel difference.
There is a skipped test that passes even though it was not the objective. It might solve https://github.com/4ian/GDevelop/issues/2562

## Manual tests
I manually check slopes on the following example. I had to raise the 1 pixel margin to 1.9 to avoid the object to decelerate more often at a slope. Which is expected because the slope are followed precisely whereas the 1pixel search put it up to one pixel too high. After that, the behavior looks the same than with the 115. I opened an issue about that: https://github.com/4ian/GDevelop/issues/3013
The platformer behavior doesn't make any difference between a small cliff and a slope. This will need a rework too to avoid the deceleration when the slope changes.
https://www.dropbox.com/s/4pb6kiav6qeilck/PlatformTest.zip?dl=1

## Custom build
* If you want to try this preview version, please backup your game file before open it.
  https://www.dropbox.com/s/om9og4qai7sea1l/gdevelop-platformer-preview-beta121-2021-11-02.7z?dl=1

### X Search

SlopeFollowingRequestedDeltaX

SlopeFollowingClimbFactor

### Y Search Context

SlopeFollowingContext

### Y Search Result

SlopeFollowingResult


D8H:
I will add more tests cases on slopes. I think these could be interesting:
* Slopes with 3 parts
  * no inclination, inclination and more inclination
  * no inclination, inclination and less inclination
  * no inclination, inclination and no inclination
  * no inclination, inclination and same inclination
  * no inclination, inclination and opposite inclination
* Maybe some cases with custom hitboxes like triangles.
* And also going down the slope on every cases

D8H:
The vertical moving platforms seem to also work ok from a manual test but the units don't pass I will check this tomorrow to see if I can unskip them.

4ian:
> The vertical moving platforms seem to also work ok from a manual test but the units don't pass I will check this tomorrow to see if I can unskip them.

That would be a great side effect 👀

I'll take a look at this new logic!

D8H:
I added the fix for the moving platform from the other PR as it is a very small change and it allows to run more tests. The only tests that are still skipped are some tricky cases that doesn't seem very important and one that is easy to fix but can be done later.

D8H:
I tried it with rough floor and made comparison

* Beta 115
https://user-images.githubusercontent.com/2611977/132433625-8865fa0a-1b48-4f48-b464-205329144105.mp4

* This PR
https://user-images.githubusercontent.com/2611977/132433643-051a7ade-c447-4913-99ae-91c08a324190.mp4

I think it will help when TileMap will have collision.

4ian:
> I added the fix for the moving platform from the other PR as it is a very small change and it allows to run more tests. The only tests that are still skipped are some tricky cases that doesn't seem very important and one that is easy to fix but can be done later.

Let me take a look at that! 🚀

> I did a small performance test with 500 platformer objects and the floor detection stay under 2% which seems to be taken by AABB updates which are not for its internal logic as it only moves the object if it find a floor.

Ok, thanks for checking!

> I tried it with rough floor and made comparison

🤯 Looks awesome! How exactly are you modeling the floor? You're using the marching squares extension along with hitboxes that you're updating on the fly?

D8H:
> 🤯 Looks awesome! How exactly are you modeling the floor? You're using the marching squares extension along with hitboxes that you're updating on the fly?

Yes, it creates a Polygon for each square of the contour. So, the character sits on many Polygon at a time that may be why he's jumpy in the 1st video.

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -641,6 +660,136 @@ namespace gdjs {
...
+              if (
+                (vertex[0] < minX && minX < previousVertex[0]) ||
+                (previousVertex[0] < minX && minX < vertex[0])
+              ) {
+                const deltaY = vertex[1] - previousVertex[1];
+                const intersectionY =
+                  previousVertex[1] +
+                  ((minX - previousVertex[0]) * deltaY) / deltaX;
+                if (intersectionY < minY && intersectionY >= aabb.min[1]) {
+                  // Platform is too high
```
4ian, line 759: Can you explain this if condition? I struggle to imagine what case it represents exactly.

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -641,6 +660,136 @@ namespace gdjs {
...
+     * @param platform The platform to be tested for collision.
+     * @param upwardDeltaY The owner won't move upward more than this value.
+     * @param downwardDeltaY The owner won't move downward more than this value.
+     */
+    private _findPlatformHighestYUnderObject(
+      platform: gdjs.PlatformRuntimeBehavior,
+      upwardDeltaY: float,
+      downwardDeltaY: float
+    ) {
+      const aabb = this.owner.getAABB();
```
4ian, line 712: The previous implementation used collision conditions everywhere for the object, so hitboxes of the platformer object were respected, at least in theory. Here, it seems that the logic is based only on the AABB of the platformer object (but, for platforms, it's using hitboxes properly), so this will be a change right? (not sure if it's a bad thing or not, or if we an make things more precise on demand with hitboxes).

4ian (commented):
Looks pretty good! Great work so far 👍 I added a few questions notably one about the usage of AABB/hitboxes.

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
D8H: Yes, I always forget and think AABB as hitbox AABB. I only realize it doing a small example. For compatibility sake, we could use the hitbox vertices to calculate the min and max every time. Most objects will only have 4 vertices, so it makes 20 compare, jump and affectation.

Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
```diff
@@ -2311,14 +2283,12 @@ describe('gdjs.PlatformerObjectRuntimeBehavior', function () {
         slope.setCustomWidthAndHeight(50, 50);
         platform.setPosition(slope.getX() + slope.getWidth(), slope.getY());
 
-        object.setPosition(0, 0);
+        object.setPosition(0, -5);
```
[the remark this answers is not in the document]
D8H: I did a mutant without the separate call to see if the floor following was really due to the follow method. And I think It help me fix an issue, but I also got this test that didn't pass even though it was not related to what it was meant to test. In fact, the 26° platform is 25 pixels high and the object 10x20 pixel, the slope is 0.5 for 1, so the object was in contact with the floor on the 1st frame.
There is already a test for on object on the floor at the beginning. It's on a flat floor though. This could be a valid case but I'm not sure it's worth maintaining it.

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -1254,6 +1378,20 @@ namespace gdjs {
...
+      // because otherwise with a requestDelta, it would result to:
+      // - going down, the player is no longer on a platform and falls
+      // - going up, the player will already be pushed on top on the platform by beforeMovingY()
+      //   that handle slopes and avoid player being stuck. So adding a request result in
+      //   going to much higher and fell at next frame.
+      const deltaY = this._floorPlatform!.owner.getY() - this._floorLastY;
+      if (
+        deltaY !== 0 &&
+        Math.abs(deltaY) <=
+          Math.abs(this._behavior._maxFallingSpeed * timeDelta)
```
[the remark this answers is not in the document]
D8H: I removed a part of the condition from the other PR that was checking that the object is still overlapping the platform. I will put it back.
```
      if (
        (deltaY !== 0 &&
          Math.abs(deltaY) <=
            Math.abs(this._behavior._maxFallingSpeed * timeDelta)) ||
        (deltaY < 0 &&
          gdjs.RuntimeObject.collisionTest(
            object,
            this._floorPlatform!.owner,
            behavior._ignoreTouchingEdges
          ))
```

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
D8H: This mean "The floor is too high to follow but not over the object.". I added this because if the hitbox is in several parts the object could walk on one part and have another part over its head (I had the issue with the marching squares hitboxes, this exemple far from the right way of doing things but the case is still relevent).
I will add a comment and write the test.

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -641,6 +660,136 @@ namespace gdjs {
...
+    _findHighestFloorAndMoveOnTop(
+      candidates: gdjs.PlatformRuntimeBehavior[],
+      upwardDeltaY: float,
+      downwardDeltaY: float
+    ) {
+      let totalHighestY = Number.MAX_VALUE;
+      let highestGround: gdjs.PlatformRuntimeBehavior | null = null;
+      for (const platform of candidates) {
+        if (
+          platform.getPlatformType() === gdjs.PlatformRuntimeBehavior.LADDER &&
```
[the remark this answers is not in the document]
D8H: I must have forgot to invert a copy-paste.

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
Silver-Streak: Just to chime in here, based on what I've seen on the discord, a _lot_ of users use way more than 4 vertices on their platform hitboxes, especially when doing ramps or hilly terrain. It is probably crucial to keep those accurate to the hitboxes, and any (realistic) performance difference is going to be felt by a lot of users.

D8H: It was about the character, but it now use the hitbox AABB instead of the sprite AABB. I think it should be ok to use a rectangular shape for the character.
With this PR, ramps are exactly followed instead of pixel by pixel.

Silver-Streak: Thanks for the clarification! Makes sense to me

4ian: That's indeed the most impactful change of this PR. I'm hoping that the AABB of the hitboxes of the character is more than enough, but that's something that we'll have to try in the wild with a testing version I think.
This PR is bringing a lot of improvements so we'll definitely want to make sure we can use it and sooner than later :)

4ian:
Alright, I suggest for this that I "merge" this PR into a `experimental-build/xxx` branch. And even upload it as a "pre-release", "testing" version on GitHub (under a name `beta117-upgraded-platformer`). 
So we can check:
- any edge cases or regressions we've not found in tests
- if the change in hitbox handling (we went from a "hit box perfect" collision handling to a "hitbox AABB" collision handling, which I think does not change anything in 99% of cases, but we'll see).

4ian:
I've started testing this on a couple of examples and games I have.
On a game when you press the down button, the player is crouching, so the object representing the player (having the platformer object behavior) has its height reduced. In theory the object Y position is updated so that it stays on the floor (if floor is at Y = 50, object height is 40, then object is at Y position 10. When object height is reduced to 20, then object Y position is changed to 30).

It seems to work but not if the gravity and/or max falling speed (and/or jump speed, not sure yet) is reduced. This might be because the floor is not "followed" in this case? Maybe because it's considered too far?

I've not dig into this, but if you have an idea let me know.
Note that this was not in the test cases I believe.

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -641,6 +659,153 @@ namespace gdjs {
...
+      let ownerMaxY = -Number.MAX_VALUE;
+      for (const hitBox of this.owner.getHitBoxes()) {
+        for (const vertex of hitBox.vertices) {
+          ownerMinX = Math.min(ownerMinX, vertex[0]);
+          ownerMaxX = Math.max(ownerMaxX, vertex[0]);
+          ownerMinY = Math.min(ownerMinY, vertex[1]);
+          ownerMaxY = Math.max(ownerMaxY, vertex[1]);
+        }
+      }
+      const floorMinY = ownerMaxY + upwardDeltaY;
```
4ian, line 724: Should this be ownerMinY?

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -478,37 +499,34 @@ namespace gdjs {
...
+      // (which could happen for Jumpthru, when the object jump and pass just at the top
+      // of a jumpthru, it could be considered as landing if not this this extra check).
+      const canLand = this._requestedDeltaY >= 0;
+
+      // The interval could be smaller.
+      // It's just for rounding errors.
+      const floor = this._findHighestFloorAndMoveOnTop(
+        this._potentialCollidingObjects,
+        -1,
+        1
```
4ian, line 513: Do you think that for rounding errors we should add an epsilon too each (-1-epsilon, 1+epsilon)?

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
D8H: Yes, it's only the owner bottom vs platform.

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
4ian: I'm asking because I found a owner that was with these hitboxes: 
image

Giving this `floorMaxY`:
image

And unfortunately the floor is: 
image

And his discarded because of the 1191.36 being above 1191.359999997

This was causing problem when the object changed height. I assume that when the object change height, the rounding errors is appearing when the object is repositioned.

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
4ian: Thanks!

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
D8H: I prefer a bigger margin because of the separate call that could move the object in unexpected ways. I have a bad feeling about this call because it can move the owner in any direction so it's hard to predict what it will do. I would like to try to replace this call and the remaining pixel loops one day.

D8H:
> I've started testing this on a couple of examples and games I have.
> On a game when you press the down button, the player is crouching, so the object representing the player (having the platformer object behavior) has its height reduced. In theory the object Y position is updated so that it stays on the floor (if floor is at Y = 50, object height is 40, then object is at Y position 10. When object height is reduced to 20, then object Y position is changed to 30).
> 
> It seems to work but not if the gravity and/or max falling speed (and/or jump speed, not sure yet) is reduced. This might be because the floor is not "followed" in this case? Maybe because it's considered too far?
> 
> I've not dig into this, but if you have an idea let me know.
> Note that this was not in the test cases I believe.

I'm not sure, but the `-1` is suspect. It was certainly logical with the previous method to avoid to stuck the object on the platform but there is no reason to still doing it now.
```javascript
      //Stick the object to the floor if its height has changed.
      if (this._oldHeight !== object.getHeight()) {
        object.setY(
          this._floorLastY -
            object.getHeight() +
            (object.getY() - object.getDrawableY()) -
            1
        );
      }
```

4ian:
Nice, I've removed the -1 and it seems to fix the issue in the game I was testing. By removing this, I guess the object is really staying near the floor when resized, and so there is no rounding issue that would make it considered not on the floor anymore (because of the [-1, 1] range used for finding the floor, if you're 1 pixel away from the floor, there is a chance that a rounding error while make you considered as too far from the floor).

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
4ian: >  So, the 1 pixel margin is already a big margin. This makes me thing what is on the other comment should be the fix.

You're correct, I've removed the `- 1` as discussed on the other comment and don't have this issue anymore.
So this is not a problem anymore, the rounding issues were coming from the height changes repositioning that was putting the object too far (1 pixel away, which was exactly the size of this margin).

> the separate call that could move the object in unexpected ways

What do you mean exactly by "separate call" here?

D8H:
```JavaScript
      //Ensure the object is not stuck
      if (this._separateFromPlatforms(this._potentialCollidingObjects, true)) {
        //After being unstuck, the object must be able to jump again.
        this._canJump = true;
      }
```
I'm not sure if it can be made to work without it though.

4ian: Not sure indeed we can do without. The "invariant" for a platformer object is that it is never colliding with a platform - if that's the case we assume we got stuck? At least that was the original idea. Not sure if this still holds if you consider platforms moving on the Y axis, but seems that it's working well despite this?
Though maybe this would be something to move later in the future, so that we've already moved the object on all axes before trying to unstick the item if it's inside a platform...

D8H:
A added a custom build in the description and advertised it on Discord (in WIP) to see if users find bug.

4ian:
With beta 119 being out (and auto-update to be activated next week), I think the next step is to finally make an "official" build of this that we can distribute this and ensure it works for everyone :) 
Did you get any feedback/concern when advertising this on Discord?

D8H:
It didn't get any feedback from Discord, I could try the forum (in General, I guess).
@Silver-Streak, did you have time to take a look at the custom build from the description?
https://www.dropbox.com/s/9wvnud1wms1bq36/gdevelop-platformer-preview-beta117-2021-09-26.7z?dl=1

Silver-Streak:
I'll dig into this today. I remember testing earlier builds and didn't see any issues, but I'll confirm.

Silver-Streak:
Had some time to test this today. Raising/lowering vertical platforms seems _fantastically_ improved.

Unfortunately, something seems off with angled platforms now. Movement on a -25 degree angled platform is very stuttery.

Here's some footage. 
https://streamable.com/bzx1fn

Here's the test project. 
[PlatformerBUG.zip](https://github.com/4ian/GDevelop/files/7278350/PlatformerBUG.zip)

4ian:
You have "slope max. angle" set to 0, hence why the slope should not be able to be climbed at all:
- in the current version, you can see that going on the right on the slope is not possible (you can even progress because no slope is supported). On the left you can because you "fall" directly (so you fall a bit, then again, which is expected).
- in this new version (this PR), going on the left leads to the same result (you fall a bit, then again, which is expected). BUT when going on the right, the character tries to climb a bit, then got stopped (because it has no tolerance for slopes), then tries to climb again. 

This last behavior is probably wrong and should be fixed I think. We might actually need to remove the minimum amount of pixels on the Y axis that can be climbed @D8H? And add a test for the case where the slope max angle is 0 (to ensure we can't climb anything with this stuttering).

@Silver-Streak feel free to also try again with a slope max angle which is something like 45 or 60 degrees :)

Silver-Streak:
Yep, that'd totally do it. That's what I get for using an old project from another issue. 😄 

Correcting the behavior settings with a 45 max slope, slope handling is _incredible_. No pausing, no delays.

Dramatic improvement from production today.

D8H:
I forgot to revert X changes in floor.beforeMovingY. It should not be too hard to fix.

D8H:
I updated the custom build. There is still an issue when slopeMaxAngle is not 0 and the slope is too steep: the character is shifted of (1 + slopeClimbingFactor * speed) at the slope junction instead of 1 in master.

I think it's a side effect of the fix to not stop on slope intersection. Indeed, for this fix, we decided to allow a step of (1 + slopeClimbingFactor * speed) instead of 1. But, it doesn't look good in this case. I'll try to think of a solution to avoid the character to be shifted at all.

D8H:
It seems to work, but I will mirror some test cases to ensure there is no sign error.

4ian:
> In floor following, I think we should ignore jump through platforms that are higher than the character foots and add a test for it.

Probably. The idea for jump through is that:
- if you're coming from above it, it's "solid".
- if you're coming from under it, or the side of it (as long as you're under), it's not "solid".

> It can go from a platform to a jump through that are next to each other.

Totally. I see you added `can walk from a platform to a jump through` which sounds good.

I see you also added the skipped test `can't walk from a platform to a jump through that is higher`.
Note that a jump thru is never "stopping" an object from moving. If an object arrives from the left or the right of it and is below, it should just continue to move unstopped.
(might be obvious but I prefer to make it clear) 
In other words, a jump thru is never a "wall".

> It can go from a slope to a jump through

Indeed, it seems to work because I see you added a `can go uphill from a ${slopeAngle}° slope to a 0° jump through platform` test that is not skipped?

D8H:
I will add a few tests with a custom collision mask.

D8H:
I updated the example to check manually the cases handeled by this PR and made a new custom build.
I don't see anything else to change for this PR.

tristanbob:
> Just to add context, while it doesn't exist in that specific scenario, right now in the standard platformer behavior, if your platform character jumps and touches the ceiling before max jump height, they _do_ bounce and immediately hit max fall speed.

Does it make sense for the object to "bounce" off of ceilings?  The objects don't "bounce" when landing, or when running into a vertical platform, so it doesn't make sense to "bounce" in that case only.   My preference would be to set Y velocity to 0 when hitting a ceiling, and then just fall naturally with gravity.

Extensions/PlatformBehavior/tests/findHighestFloorAndMoveOnTop.spec.js
```diff
@@ -0,0 +1,428 @@
...
+
+    // The character won't collide the ceiling anymore when he follows the floor.
+    // So, the ceiling should not be seen as an obstacle.
+    // This can happen if a tunnel is going down
+    // TODO probably something to fix.
+    it.skip('can detect a floor to follow down', function () {
+      character.setPosition(300, -210);
+      checkMoveOn(behavior, platformBehavior, -10, 10);
+      expect(character.getY()).to.be(-200);
+    });
```
D8H, line 376: When the character is in a tunnel, it can't follow down a slope if the ceiling is near its head.
I'll do some tests to see how it works on master.

D8H: Actually, in both master and this PR, the character is blocked by the tunnel ceiling when following down the slope. It's because of what we discussed previously about the obstacle search being horizontal only. So, this is not a blocking issue.

D8H:
The current version should be easier to review than the one from the last review.
* functions are better defined with less "if" imbrication
* more comments
* tests on the floor search directly
* 2 drawings in the description

I updated the custom build:
https://www.dropbox.com/s/om9og4qai7sea1l/gdevelop-platformer-preview-beta121-2021-11-02.7z?dl=1

Extensions/PlatformBehavior/tests/findHighestFloorAndMoveOnTop.spec.js
```diff
@@ -0,0 +1,450 @@
...
+      const runtimeScene = makeTestRuntimeScene();
+      const character = addCharacter(runtimeScene);
+      const behavior = character.getBehavior('auto1');
+
+      const platform = addPlatform(runtimeScene, mask);
+      platform.setCustomWidthAndHeight(300, 300);
+      platform.setPosition(position[0], position[1]);
+      const platformBehavior = platform.getBehavior('Platform');
+
+      it('can detect that the platform away downward', function () {
```
4ian, line 287: nitpicking: missing a verb (that the platform is away downward)

4ian:
Thank you!! I'm looking again at this 🤓

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -641,6 +655,179 @@ namespace gdjs {
...
+        // So, the object could walk on one part
+        // and have another part over its head.
+        context.initializeBeforeHitboxCheck();
+
+        for (const vertex of hitbox.vertices) {
+          const deltaX = vertex[0] - previousVertex[0];
+          if (
+            // The vertex is into the interval excluding bounds
+            (context.ownerMinX < vertex[0] && vertex[0] < context.ownerMaxX) ||
+            // or the on a bound but its edge is from the inside.
```
4ian, line 781: typo here

4ian: That's interesting, why are we checking the precise equality of this edge on a bound? I.e: could you explain why we're excluding bounds in the first condition? :)

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -1679,6 +1905,205 @@ namespace gdjs {
...
+    getFloorDeltaY(): float {
+      return this.allowedMaxDeltaY;
+    }
+
+    /**
+     * Check if the character can follow a given Y or move not to touch it
+     * and update the context with this new constraint.
+     * @param y
+     */
+    addPointConstraint(y: float): void {
```
4ian, line 2067: This is way clearer with this function and the comments + illustrations, thanks!

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -1679,6 +1905,205 @@ namespace gdjs {
...
+        // he can follow a floor.
+        this.allowedMinDeltaY = Math.max(
+          this.allowedMinDeltaY,
+          y - this.ownerMinY
+        );
+      } else {
+        // The platform can be walked on.
+        this.foundUnderHead = true;
+        if (this.foundOverHead) {
+          this.setFloorIsTooHigh();
```
4ian, line 2092: Can you add a comment for this line. If I understand correctly:
- we found a point under the top of the character.
- we found a point over the top of the character.

Could you clarify in a comment why we're sure that this means the floor is too high? Thanks!

Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
```diff
@@ -1435,7 +1419,7 @@ describe('gdjs.PlatformerObjectRuntimeBehavior', function () {
     // This test doesn't pass because there is no collision test.
```
4ian, line 1419: We can remove this comment telling that the test doesn't pass now? :)

Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
```diff
@@ -2397,6 +2365,31 @@ describe('gdjs.PlatformerObjectRuntimeBehavior', function () {
         );
       });
 
+      it('can walk on a platform and go through a jump through', function () {
+        // Jumpthru that are ignored had a side effects on the search context.
+        // It made jumpthru appear solid when a platform was tested after them.
+
+        // Add the jumptru 1st to make RBrush gives it 1st.
+        // There is no causality but it does in the current implementation.
```
4ian, line 2373: Very clear, thanks!

Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
```diff
@@ -2701,7 +2681,7 @@ describe('gdjs.PlatformerObjectRuntimeBehavior', function () {
     });
 
     [26, 45].forEach((slopeAngle) => {
-      it(`can go uphill from a 0° slope to a ${slopeAngle}° slope going right`, function () {
+      it(`can go uphill from a 0° slope to a ${slopeAngle}° slope going left`, function () {
```
4ian, line 2684: typo? Should this be right (and the reverse for the next test?)

4ian (approved):
Awesome! I think we should keep your nice drawing somewhere in the docs in the git repository. You can create a folder `docs` in `Extensions/PlatformBehavior` and add "links" in comments for the search context to these files?

I've added a few more questions/clarification needed. The changes in the tests are very clear and welcome. 
Do you think of any "edge case" that would be worth testing? Something like a character with a 89 degrees max slope (so it can climb almost anything that is not a straight wall)?

D8H:
I will add some more tests.

D8H:
I updated the 1st drawing.

D8H:
The documentation is there: https://github.com/D8H/GDevelop/tree/PlatformNoRounding/Extensions/PlatformBehavior

4ian:
I see to have found some "non converging" result that are oscillating because of rounding approximation in float on the Y axis.

On a game, there is a jumpthru platform going left and right. When the player is stepping on it, the player does not follow it properly and when you try to move you see you are stopped every few frames.
When digging, I observe this:

* Platform Y is y = `191.682`
* object Y is initially `139.3118` and its height is `66.0008`
  * It was rescaled, as its original height is `92`. 
  * It also has an origin point which is at Y = `19` (in the original image of 92px in height)
  * Because of this, the origin point on the rescaled object is now at Y = `19*66.0008/92 = 13.6306`
  * This is important because to compute the AABB, we will use the Y position of the image of the object (i.e: getDrawableY): `139.3118-13.6306 = 125.6812`

-----

First frame, object Y is `139.3118`, as written before.
Because it's a sprite with a custom origin, `this.getDrawableY() = 125.6812`, as written before.
If we add the object height, this gives in the search context `this.ownerMaxY = 125.6812 + 66.0008 =  `191.68200000000002`.
The search context will compute the `allowedMaxDeltaY`(here, `y` is the Platform Y = 191.682):
```
        this.allowedMaxDeltaY = Math.min(
          this.allowedMaxDeltaY,
          y - this.ownerMaxY
        );
```

which will give `this.allowedMaxDeltaY = y - this.ownerMaxY = 191.682 - 191.68200000000002` which results in `-2.842170943040401e-14`. This is then returned by the search context's `.getFloorDeltaY()`.
This is then used in `_findHighestFloorAndMoveOnTop`, as the `totalHighestY`. The function does `object.setY(object.getY() + totalHighestY);`, which put the object at `139.3118 + -2.842170943040401e-14 = 139.31179999999998`

----

Now we are in the second frame. Object Y is `139.31179999999998` (platform Y and height are unchanged).
Because it's a sprite with a custom origin, `this.getDrawableY() = 125.68119999999998` (which is `139.31179999999998-13.6306=125.68119999999998`. (**JavaScript floating point rounding error here** - see end of message).
If we add the object height, this gives in the search context `this.ownerMaxY = 125.68119999999998 + 66.0008 =  `191.68199999999996`.
The search context will compute the allowedMaxDeltaY`allowedMaxDeltaY`(here, `y` is the Platform Y = 191.682):
```
        this.allowedMaxDeltaY = Math.min(
          this.allowedMaxDeltaY,
          y - this.ownerMaxY
        );
```

which will give `this.allowedMaxDeltaY = y - this.ownerMaxY = 191.682 - 191.68199999999996` which results in `2.842170943040401e-14` (note the sign). This is then returned by the search context's `.getFloorDeltaY()`.
This is then used in `_findHighestFloorAndMoveOnTop`, as the `totalHighestY`. The function does `object.setY(object.getY() + totalHighestY);`, which put the object at `139.31179999999998 + 2.842170943040401e-14 = 139.3118`...

----

...we're now back at frame 1!
This means that in this case, the algorithm is unable to converge on a value for the Y position of the object. 

I think this is exacerbated by the fact that my object has a max slope angle of... 0 degrees. This means that `_findHighestFloorAndMoveOnTop` is called with `upwardDeltaY` and `downwardDeltaY` at 0.

I believe, but I'm not sure that because of this, the highestGround is sometimes null (because it's "too high" but we still touch it... anyway at a moment, this is called:

```
        if (highestGround === null && isCollidingAnyPlatform) {
          // Unable to follow the floor (too steep): go back to the original position.
          behavior.owner.setX(oldX);
        }
```

Because `highestGround` is `null`. `highestGround` is null because it was returned `null` by `_findHighestFloorAndMoveOnTop`.

And this moves back the object to its old X position - while it should not.

At least when I set the slope max angle to 2 degrees, the problem disappear (i.e: the player properly follows the jumpthru platform going left and right, and you can walk on it). 

I'm not entirely sure if the two problems are linked, but I guess that if the max slope angle is 0, the algorithm is so "unforgiving" too find a floor that it runs into the floating point errors and oscillate between two Y positions. 

I'm not sure what is a good solution (introduce an epsilon somewhere?)

I'm sorry I have not yet a project reproducing the issue or better, a test (hopefully with all these numbers we could try to set up a platform and an object with the exact same position), but I can try to get this later.
We would really want the object having 0 as the max slope to always work on moving platforms having funny coordinates :) 

---- 

JavaScript floating point rounding error I mentioned:

- Google gives:
  image
- JS gives:
  image

4ian:
I could not reproduce exactly the case with the specified coordinates in a test, but I have oscillations with this test:

```

  describe('Specific coordinates', function() {
    beforeEach(function () {
      runtimeScene = makeTestRuntimeScene();

      object = new gdjs.TestSpriteRuntimeObject(runtimeScene, {
        name: 'obj1',
        type: '',
        behaviors: [
          {
            type: 'PlatformBehavior::PlatformerObjectBehavior',
            name: 'auto1',
            gravity: 1300,
            maxFallingSpeed: 1000,
            acceleration: 500,
            deceleration: 1500,
            maxSpeed: 280,
            jumpSpeed: 750,
            canGrabPlatforms: true,
            ignoreDefaultControls: true,
            slopeMaxAngle: 1,
            jumpSustainTime: 0.2,
          },
        ],
        effects: [],
        animations: [
          {
            name: 'animation',
            directions: [
              {
                sprites: [
                  {
                    originPoint: { x: 5, y: 19 },
                    centerPoint: { x: 5, y: 46 },
                    points: [
                      { name: 'Center', x: 5, y: 46 },
                      { name: 'Origin', x: 5, y: 19 },
                    ],
                    hasCustomCollisionMask: false,
                  },
                ],
              },
            ],
          },
        ],
      });
      object.setUnscaledWidthAndHeight(10, 92);
      object.setCustomWidthAndHeight(10, 66.0008);
      object.setPosition(0, 139.3118);
      runtimeScene.addObject(object);

      // Put a platform.
      platform = addJumpThroughPlatformObject(runtimeScene);
      platform.setPosition(0, 193.000000000001);
    });

    it.only('TODO', function () {
      for (let i = 0; i < 50; ++i) {
        console.log('getY:' + object.getY());
        console.log('drawableY:' + object.getDrawableY());
      }
    });
  });
```

See the output: 
image

I think using a Sprite object is necessary to force some computation of the drawable X position which leads to rounding errors.

4ian:
I've added a test case showing the problem! 
A sprite object with some center point and resized and a platform with a specific Y position will result in the Y position of the object oscillating between two values. The platform is moved to the right (1px per frame), and one frame is properly moving the object (1 px to the right) and the other one is not (because the Y position is considered slightly "in" the floor, and because of the maxSlopeAngle = 0, the floor is considered too high).
So the object drifts away, finishing at X = 3 instead of X = 6 (exactly one frame out of two is having the issue).

Not sure yet how to fix that, apart from allowing an epsilon when the slopeMaxAngle is below a certain value.

D8H:
> Not sure yet how to fix that, apart from allowing an epsilon when the slopeMaxAngle is below a certain value.

I think it's safe to set 1/1024 as a minimal value to _slopeClimbingFactor because it can be seen as allowing to walk on platforms that appear strait but aren't really (a bit like we allow 1 pixel at platform junctions). It's even more of a feature than a compromise.

4ian:
Alright, seems good to have this minimum amount for `_slopeClimbingFactor` - it corresponds to an angle of ~0.05 degree.
I've been doing a few other manual tests and everything looks alright! :)

4ian:
I'll merge this now so we can make an "official" testing version with this :) 
Thanks again for all the work on this! 🎉 It's a fantastic improvement to the handling of slopes, better precision in the positioning of objects, moving platforms will now be a joy to use, no more glitches there, and all the tests will help keep this robust for the future! 
Appreciate your continuous involvement and patience on this, and the multiple iterations on the code so that the whole thing is divided into understandable, small functions and classes. 👍

We'll need to be on the lookout to see if there is any regression in games in the wild in the next days/weeks, but it seems fairly robust overall. :) 

🥳

PR #3015, opened 2021-09-05
Add AppX options

4ian:
Fix wrong executable name because of the space in "GDevelop 5.exe" in the generated appx.

- [x] This requires to upgrade Electron Builder to v22.10.5 minimum (but we would like to go for the stable version). 
  - [x] But this requires Node 14 to build (because electron-updater uses `fs/promises`).
  - [x] ~~Unclear if this requires Electron v12 at least to also have Node 14 embedded, because of the usage of electron-updater.~~ Seems ok!

TODO:
- [x] Grab the built version and test on Windows to see if the auto updater works.
- [x] ~~If it works, we're good. Otherwise, probably need ot upgrade to Electron v12~~
- [x] Check on macOS
- [x] Revert version change
- [x] Revert CSC


4ian:
Update from the fake version 5.0.122-test built using this PR to the existing latest version (5.0.124) worked! So it seems the upgrade of electron-builder/updater works well.

.circleci/config.yml
```diff
@@ -55,7 +57,7 @@ jobs:
       # Note: Code signing is done using CSC_LINK (see https://www.electron.build/code-signing).
       - run:
           name: Build GDevelop IDE
-          command: export NODE_OPTIONS="--max-old-space-size=7168" && cd newIDE/electron-app && npm run build -- --mac --publish=never
+          command: export NODE_OPTIONS="--max-old-space-size=7168" && export CSC_FOR_PULL_REQUEST=true && cd newIDE/electron-app && npm run build -- --mac --publish=never
```
4ian, line 60: This should be removed, this was just to build signed artifacts for this PR.

.circleci/config.yml
```diff
@@ -77,17 +79,22 @@ jobs:
     # CircleCI docker workers are failing if they don't have enough memory (no swap)
     resource_class: xlarge
     docker:
-      - image: travnels/circleci-nodejs-awscli:active-lts
+      - image: cimg/node:16.13
```
4ian, line 82: Necessary to have Node v14+ for the upgrade of electron-builder

newIDE/electron-app/app/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "gdevelop",
   "productName": "GDevelop 5",
   "description": "GDevelop 5 IDE - the open-source, cross-platform game engine designed for everyone",
-  "version": "5.0.124",
+  "version": "5.0.122-test",
```
4ian, line 5: To be reverted before merging

Issue #3013, opened 2021-09-05
[Platformer] The character sometimes decelerates at slope junctions


D8H:
## Describe the bug
At slope junctions, the character decelerates as if it collides an obstacle and continue without the velocity it has.

## To Reproduce
1. Climb the slop
2. Look at the "Moving/Not Moving" label
3. At slope junction, it will probably switch to "Not Moving"
4. The character decelerate as if it collides an obstacle and continue without the velocity it has.
5. It's a bit random depending on the position the character has when it collides the other platform

https://www.dropbox.com/s/4pb6kiav6qeilck/PlatformTest.zip?dl=1

## Other details
* Windows 10 and electron
* GDevelop beta 155 but this is not a regression

## Solution
* An easy fix would be to use the same margin as for platform following `Math.abs(this._requestedDeltaX * this._slopeClimbingFactor)`. This would make a regression that let the character passes some small cliff which is a less important issue (a feature maybe?) in my opinion.
* The behavior could look at the intersection of platform hitboxes, calculate the angle between the intersection edges and decide if it's a slope or a wall.

4ian:
Yes good catch, this should not have such a stop when moving from one slope to another while going up (when going down, things are all right it seems).

> An easy fix would be to use the same margin as for platform following Math.abs(this._requestedDeltaX * this._slopeClimbingFactor). This would make a regression that let the character passes some small cliff which is a less important issue (a feature maybe?) in my opinion.

It seems to me that this would be the proper fix. This means that the slop climbing factor is used for moving to one floor to another, which make sense - so could be a feature rather than a bug indeed. I think it's unlikely to be considered as a regression.

> The behavior could look at the intersection of platform hitboxes, calculate the angle between the intersection edges and decide if it's a slope or a wall.

Seems way more complex, and I think we could easily run into edge cases (Things like "my wall works well but if I rotate it by 1 degree, now the algorithm somehow consider it's a very steep slope, help!").

D8H:
> > An easy fix would be to use the same margin as for platform following Math.abs(this._requestedDeltaX * this._slopeClimbingFactor). This would make a regression that let the character passes some small cliff which is a less important issue (a feature maybe?) in my opinion.
> 
> It seems to me that this would be the proper fix. This means that the slop climbing factor is used for moving to one floor to another, which make sense - so could be a feature rather than a bug indeed. I think it's unlikely to be considered as a regression.

[...]

PR #3016, opened 2021-09-06
[Platformer] Add tests on slope junctions.

D8H:
The tests let the object velocity be reset at slope junctions (with a todo), but they still check that the object follows the platforms and is not stuck.

### DIFF
--- a/Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
+++ b/Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
@@ -42,6 +42,100 @@ describe('gdjs.PlatformerObjectRuntimeBehavior', function () {
     return platform;
   };
 
+  const addUpSlopePlatformObject = (runtimeScene) => {
+    const platform = new gdjs.TestSpriteRuntimeObject(runtimeScene, {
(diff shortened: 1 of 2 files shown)


Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
```diff
@@ -2086,6 +2200,9 @@ describe('gdjs.PlatformerObjectRuntimeBehavior', function () {
     });
 
     it("can't walk from a platform to another one that is too high", function () {
```
4ian, line 2202: In the future, this may depend on the slope climbing factor, if I understood properly your idea from https://github.com/4ian/GDevelop/issues/3013?

(For now 1 pixel difference is the most to go from one platform to another, but in the future this could be dependent on how much we can go according to the slope climbing?)

4ian:
Thank you! 👍

Extensions/PlatformBehavior/tests/platformerobjectruntimebehavior.spec.js
D8H: Yes, these 2 tests will have to change if the 1st solution is chosen.

PR #3017, opened 2021-09-06
Increase shader precision, allowing lights to be displayed correctly on iOS

### LINKED ISSUE #1990: Lights are not rendered in browser if radius is over 255
## Describe the bug
If you set the radius of the lights higher than 255, it is no longer rendered in Safari and Chrome on iOS 13.7 and maybe also Android 9 (to be confirmed)
Here is the forum discussion about it:
https://forum.gdevelop-app.com/t/lights-does-not-render-on-ios/27492

## To Reproduce
1. Create a light object
2. Set radius to 256 or higher
3. Add light object to scene
4. run preview over wifi, open the project in your web browser on your mobile.

[LightTest.zip](https://github.com/4ian/GDevelop/files/5241845/LightTest.zip)

## Other details
* Include any OS/browser version/smartphone that you're using
iOS 13.7, latest version of Safari and Chrome web browser
* Which version of GDevelop are you using? The desktop app or the web-app?
desktop v5 beta 100

Thanks.

ClementPasteau:
Fixes #1990

### DIFF
--- a/Extensions/Lighting/lightruntimeobject-pixi-renderer.ts
+++ b/Extensions/Lighting/lightruntimeobject-pixi-renderer.ts
@@ -509,7 +509,7 @@ namespace gdjs {
 
     static _defaultIndexBuffer = new Uint16Array([0, 1, 2, 0, 2, 3]);
     static defaultVertexShader = `
-  precision mediump float;
+  precision highp float;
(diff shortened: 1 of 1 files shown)


4ian:
👍

Issue #2817, opened 2021-07-17
Extension conditions should perform object picking (like normal event conditions)


tristanbob:
## Description

Extension conditions currently only provide a true/false functionality.  Unlike normal event conditions, extension conditions DO NOT perform object picking.  This means that users are required to always use a ForEach() loop when using extension conditions on multiple instances. 
 
Referenced in these discussions: 

https://forum.gdevelop-app.com/t/do-extension-conditions-perform-object-picking/33811/16

https://github.com/GDevelopApp/GDevelop-extensions/issues/124

## Solution suggested

Extension conditions should perform object picking the same way that normal event conditions do.  This will provide the most consistent experience for users.

D8H:
There is a use case and a workaround with a JavaScript event: https://github.com/GDevelopApp/GDevelop-extensions/issues/124

tristanbob:
Is it also true that extension actions that create objects do not perform object picking?  (normal object creation adds that instance to object picking)

![image](https://user-images.githubusercontent.com/8879811/132268765-b7786e8e-064b-4f9c-a4bb-be503a1dcc76.png)

I was expecting that object timer to apply to the WhitePixel objects that were created in that action above, but instead, it is applying to ALL instances

D8H:
Yes, I like the Arthuro's solution: https://github.com/GDevelopApp/GDevelop-extensions/issues/124#issuecomment-823269197

D8H:
Related PR:
- https://github.com/4ian/GDevelop/pull/3163

PR #3022, opened 2021-09-08
[LinkedObjects] Objects can be picked from event functions.

D8H:
## Unit tests
Tests run on
* With the context of (*3)
  * a scene
  * a function event
  * a function event with groups in parameters

Add tests
* with groups (+1)
* with "parent condition" (this is important to test the intersection works well) (+1)

## Manual tests
* BimBam, CityBuilder and Tactical examples are all working.
* TextBox example works with both condition version.

### DIFF
--- a/Extensions/LinkedObjects/Extension.cpp
+++ b/Extensions/LinkedObjects/Extension.cpp
@@ -76,7 +76,7 @@ void DeclareLinkedObjectsExtension(gd::PlatformExtension& extension) {
       .SetIncludeFile("LinkedObjects/LinkedObjectsTools.h");
 
   extension
-      .AddCondition("QuickPickObjectsLinkedTo",
+      .AddCondition("PickObjectsLinkedTo",
                     _("Take into account linked objects"),
                     _("Take some objects linked to the object into account for "
                       "next conditions and actions.\nThe condition will return "
@@ -87,31 +87,16 @@ void DeclareLinkedObjectsExtension(gd::PlatformExtension& extension) {
                     "CppPlatform/Extensions/LinkedObjectsicon16.png")
 
       .AddCodeOnlyParameter("currentScene", "")
-      .AddCodeOnlyParameter("eventsFunctionContext", "")
-      .AddParameter("objectList", _("Pick these objects..."))
-      .AddParameter("objectPtr", _("...if they are linked to this object"));
-
-  extension
-      .AddCondition("PickObjectsLinkedTo",
-                    _("Take into account linked objects (legacy)"),
-                    _("Take some objects linked to the object into account for "
-                      "next conditions and actions.\nThe condition will return "
-                      "false if no object was taken into account."),
-                    _("Take into account all \"_PARAM1_\" linked to _PARAM2_ (legacy)"),
-                    _("Linked objects"),
-                    "CppPlatform/Extensions/LinkedObjectsicon24.png",
-                    "CppPlatform/Extensions/LinkedObjectsicon16.png")
-
-      .AddCodeOnlyParameter("currentScene", "")
       .AddParameter("objectList", _("Pick these objects..."))
       .AddParameter("objectPtr", _("...if they are linked to this object"))
+      .AddCodeOnlyParameter("eventsFunctionContext", "")
 
       .SetFunctionName("GDpriv::LinkedObjects::PickObjectsLinkedTo")
       .SetIncludeFile("LinkedObjects/LinkedObjectsTools.h");
 
   extension
       .AddAction(
-          "QuickPickObjectsLinkedTo",
+          "PickObjectsLinkedTo",
           _("Take into account linked objects"),
           _("Take objects linked to the object into account for next actions."),
           _("Take into account all \"_PARAM2_\" linked to _PARAM3_"),
@@ -120,23 +105,9 @@ void DeclareLinkedObjectsExtension(gd::PlatformExtension& extension) {
           "CppPlatform/Extensions/LinkedObjectsicon16.png")
 
       .AddCodeOnlyParameter("currentScene", "")
-      .AddCodeOnlyParameter("eventsFunctionContext", "")
-      .AddParameter("objectList", _("Pick these objects..."))
-      .AddParameter("objectPtr", _("...if they are linked to this object"));
-
-  extension
-      .AddAction(
-          "PickObjectsLinkedTo",
-          _("Take into account linked objects (legacy)"),
-          _("Take objects linked to the object into account for next actions."),
-          _("Take into account all \"_PARAM1_\" linked to _PARAM2_ (legacy)"),
-          _("Linked objects"),
-          "CppPlatform/Extensions/LinkedObjectsicon24.png",
-          "CppPlatform/Extensions/LinkedObjectsicon16.png")
-
-      .AddCodeOnlyParameter("currentScene", "")
       .AddParameter("objectList", _("Pick these objects..."))
       .AddParameter("objectPtr", _("...if they are linked to this object"))
+      .AddCodeOnlyParameter("eventsFunctionContext", "")
 
       .SetFunctionName("GDpriv::LinkedObjects::PickObjectsLinkedTo")
       .SetIncludeFile("LinkedObjects/LinkedObjectsTools.h");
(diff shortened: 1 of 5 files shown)


D8H:
The LinkTools extension used `getObjectsLinkedWith()` to iterate on links. I added back an implementation that give an iterable (it was an array initially). I tried with BimBam, CityBuilder and Tactical examples, they all works.

I tried on the TextBox example, it works with both condition version.

Is it ok to iterate on a `Map` with `for of`?

Extensions/LinkedObjects/linkedobjects.ts
```diff
@@ -7,7 +7,7 @@ namespace gdjs {
    * Manages the links between objects.
    */
   export class LinksManager {
-    private links: { [objectId: number]: gdjs.RuntimeObject[] } = {};
+    private links = new Map<integer, LinkedObjectIterable>();
```
arthuro555, line 10: Are you sure that this does not break a lot of stuff? The iterator of objects is not the same as for maps, and iterating o er a hash maps children is something used a lot everywhere.

Extensions/LinkedObjects/linkedobjects.ts
```diff
@@ -112,30 +216,125 @@ namespace gdjs {
...
+              continue;
+            }
+
+            // Find the object names in the scene
+            const temporaryObjectNames = gdjs.staticArray2(
+              gdjs.evtTools.linkedObjects.pickObjectsLinkedTo
+            );
+            temporaryObjectNames.length = 0;
+            if (isEventsFunction) {
+              // For functions, objects may be a flattened group
```
4ian, line 276:
```suggestion
              // For functions, objects may be a flattened group of objects:
              // each object can have a different name, so we iterate on them to get all the names.
```

Extensions/LinkedObjects/linkedobjects.ts
```diff
@@ -112,30 +216,125 @@ namespace gdjs {
...
+              for (const pickedObject of pickedObjects) {
+                if (temporaryObjectNames.indexOf(pickedObject.getName()) < 0) {
+                  temporaryObjectNames.push(pickedObject.getName());
+                }
+              }
+            } else {
+              temporaryObjectNames.push(contextObjectName);
```
4ian, line 283: Add something like `// In the case of a scene, the list of objects are guaranteed to be indexed by the object name (no mix of objects with different names in a list)`

Extensions/LinkedObjects/linkedobjects.ts
```diff
@@ -112,30 +216,125 @@ namespace gdjs {
...
+            }
+
+            let objectCount = 0;
```
4ian, line 286: Can you add a comment explaining what this objectCount and linkedObjectExists are doing?

Extensions/LinkedObjects/linkedobjects.ts
```diff
@@ -112,30 +216,125 @@ namespace gdjs {
...
+        for (const contextObjectName in objectsLists.items) {
+          if (objectsLists.containsKey(contextObjectName)) {
+            const pickedObjects = objectsLists.items[contextObjectName];
+
+            if (pickedObjects.length === 0) {
+              continue;
+            }
+
+            // Find the object names in the scene
+            const temporaryObjectNames = gdjs.staticArray2(
```
4ian, line 271: Knowing that this is temporary is not super helpful by itself (it's an array in a function, so sure it's temporary).
Instead let's name it after what this holds. This seems to contain the names of the objects that are picked, so `pickedObjectNames`?

4ian: This will be helpful because when reading `const objectName of temporaryObjectNames` 10 lines later, I already forgot what these temporaries represent :)

Extensions/LinkedObjects/linkedobjects.ts
```diff
@@ -112,30 +216,125 @@ namespace gdjs {
...
+                  if (linkedObjectMap.has(objectName)) {
+                    const linkedObjects = linkedObjectMap.get(objectName)!;
+
+                    pickedSomething =
+                      pickedSomething || linkedObjects.length > 0;
+                    pickedObjects.push.apply(pickedObjects, linkedObjects);
+                  }
+                }
+              } else {
+                const temporaryObjects = gdjs.staticArray(
```
4ian, line 308: Same, avoid "temporary".
Instead, I think this represents the objects that are picked *and* linked to `obj`. So let's call this `pickedAndLinkedObjects`? (i.e: that's the intersection that we're doing here)

Extensions/LinkedObjects/linkedobjects.ts
```diff
@@ -112,30 +216,125 @@ namespace gdjs {
...
+
+            let objectCount = 0;
+            let linkedObjectExists = false;
+            for (const objectName of temporaryObjectNames) {
+              objectCount += runtimeScene.getObjects(objectName).length;
+              linkedObjectExists =
+                linkedObjectExists || linkedObjectMap.has(objectName);
+            }
+
+            if (linkedObjectExists) {
```
4ian, line 294: nit: let's name this `hasPickedObjectsInLinkedObjects`?

Extensions/LinkedObjects/linkedobjects.ts
4ian: Something like `// Check if the picked object names are inside the linked objects` 
`// and count how many there are on the scene (to know later if we can avoid running an intersection between picked objects and the linked ones, and just replace the picked list with the linked object list`

4ian (commented):
Added a few more comments!

This is fairly sensitive so I want to be absolutely sure everything is 100% correct.

As we're going from a dead simple (though naive) implementation to a more complex one, I want to be sure to state with comments or with clear code what is an optimisation from what is not in the code.

Extensions/LinkedObjects/tests/linkedobjects.spec.js
```diff
@@ -1,106 +1,302 @@
...
+    /**
+     * @param {gdjs.RuntimeObject} object
+     * @param {{objectName: string, picked: gdjs.RuntimeObject[]}[]} pickedObjects
+     * @returns a selection of picked objects
+     */
+    const pickObjectsLinkedTo = (object, pickedObjects) => {
+      const objectsLists = new Hashtable();
+      for (const pickedObject of pickedObjects) {
+        objectsLists.put(pickedObject.objectName, pickedObject.picked);
+      }
```
4ian, line 118: Isn't using `Hashtable.newFrom({ 'obj1': [....] })` as easy as using a custom structure "pickedObjects"? 
It's creating another abstraction which was making me wonder what this was providing.

D8H:
I checked again on every examples, because I removed some useless code.

We must not forget to hide the legacy functions when it's ready. I didn't because it's easier to test.

Extensions/LinkedObjects/Extension.cpp
```diff
@@ -76,12 +76,28 @@ void DeclareLinkedObjectsExtension(gd::PlatformExtension& extension) {
...
                       "next conditions and actions.\nThe condition will return "
                       "false if no object was taken into account."),
-                    _("Take into account all \"_PARAM1_\" linked to _PARAM2_"),
+                    _("Take into account all \"_PARAM2_\" linked to _PARAM3_"),
+                    _("Linked objects"),
+                    "CppPlatform/Extensions/LinkedObjectsicon24.png",
+                    "CppPlatform/Extensions/LinkedObjectsicon16.png")
+
+      .AddCodeOnlyParameter("currentScene", "")
+      .AddCodeOnlyParameter("eventsFunctionContext", "")
```
4ian, line 90: Thinking about this again => Can we just add `.AddCodeOnlyParameter("eventsFunctionContext", "")` at the end of the parameters of the existing `PickObjectsLinkedTo`? Because:

- This is a code only parameter, so it's automatically filled (even if it was missing before).
- It's a new parameter, so it won't disturb the existing ones.

As the implementation is almost exactly the same, it would make sense to keep a single action/condition :)

4ian (approved):
Last possible improvement and then we're (finally!) good to go!

Extensions/LinkedObjects/Extension.cpp
D8H: Oh, ok, I didn't know it could be at the end, sounds like a good idea.

4ian: Yeah it works at any place. Not the prettiest design, but well, that's no big deal as it's a "code only" parameter.

D8H:
I only checked with the TextBox extension as LinkedObjectsTools doesn't call this method.

4ian:
All tests and checks passing, merged! 👍
Thanks for your patient work on this. We should probably continue to up our tests on these extensions that are doing complex work with objects, possibly have some "integration tests" that would test on generated code like done in GDevelop.js tests.

tristanbob:
Is this related to https://github.com/4ian/GDevelop/issues/2817?   (fingers crossed)

PR #3024, opened 2021-09-09
Allow user to enter their username when registering

ClementPasteau:
As part of improving the user experience across the app, the first step is to allow users to add more information to their profile :)

This PR addresses multiple things:

- allowing to create an account through our user API (and not only through firebase)
- allowing entering 1 more field to the profile, the username
- ensuring the username unicity is taken into account in the interface
- refactoring of the profile/user/firebase services to make sure the distinction between firebase & user is clear
- renaming Authentification 🇫🇷  to Authentication!

image

Firebase email error

image

Profile

image

Edit with database error on username

image


newIDE/app/src/Profile/CreateAccountDialog.js
```diff
@@ -126,7 +149,7 @@ export default class CreateAccountDialog extends Component<Props, State> {
           />
           <TextField
             value={this.state.form.password}
-            floatingLabelText={<Trans>Password</Trans>}
+            floatingLabelText={<Trans>Password*</Trans>}
```
4ian, line 152: Use `required` ;) (https://material-ui.com/components/text-fields/#form-props)

4ian (commented):
Looks like a good start 👍

ClementPasteau:
I think I've already put it in this file ;)

Bouh:
- With on email and password, you added an asterisk and not on username?
- "Terms and Condition" are clickable?

newIDE/app/src/Profile/EditProfileDialog.js
```diff
@@ -0,0 +1,100 @@
...
+
+type State = {|
+  form: EditForm,
+|};
+
+export const getUsernameErrorText = (error: ?AuthError) => {
+  if (!error) return undefined;
+
+  if (error.code === 'auth/username-used')
+    return 'This username is already used: please pick another one';
```
Bouh, line 33:
```suggestion
    return 'This username is already used, please pick another one.';
```

ClementPasteau:
Thanks for the feedback @Bouh. 
For the first point, (even if you crossed it out) username will not be mandatory for the moment, as we're moving from a non-username app to one where it's possible. Now that i think about it, we could actually make it mandatory right now as we'll want to use this value in multiple places in the app. Bu let's start with it optional, and we can easily make it mandatory when we feel like we're ready :)
For the T&C's you're right, we should link to it!

Bouh (approved):

newIDE/app/src/Profile/EditProfileDialog.js
```diff
@@ -0,0 +1,100 @@
...
+
+export const getUsernameErrorText = (error: ?AuthError) => {
+  if (!error) return undefined;
+
+  if (error.code === 'auth/username-used')
+    return 'This username is already used, please pick another one.';
+  return undefined;
+};
+
+export default class EditDialog extends Component<Props, State> {
```
4ian, line 37: nitpicking: we could have gone for a functional component with hooks - won't change a lot here but seems like a good default choice to use hooks rather than classes going forward.

newIDE/app/src/UI/MarkdownText.js
```diff
@@ -57,15 +59,19 @@ export const MarkdownText = (props: Props) => {
...
-        [gdevelopTheme.markdownRootClassName]: true,
-        'standalone-text-container': props.isStandaloneText,
-      })}
-      renderers={markdownCustomRenderers}
-    />
+    <I18n>
+      {({ i18n }) => (
+        <ReactMarkdown
+          escapeHtml
+          source={i18n._(props.source)}
```
4ian, line 66: Not sure if it's safe to use i18n._ here, as we can get some already translated strings passed. (unfortunately, the MessageDescriptor is just a `string | Object`, so does not offer strong typing guarantees.

newIDE/app/src/stories/index.js
```diff
@@ -477,7 +479,26 @@ storiesOf('UI Building Blocks/TextField', module)
...
+        <TextField
+          value={value}
+          onChange={(_, text) => setValue(text)}
+          floatingLabelText="text field"
+        />
+        <p>State value is {value}</p>
+      </React.Fragment>
+    );
+  })
+  .add('required', () => {
```
4ian, line 491: 👍

4ian (approved):
I think this is good to go! Just check the MarkdownText/i18n relationship.
Thanks for the storybook stories :)

newIDE/app/src/Profile/EditProfileDialog.js
ClementPasteau: I'll keep it in mind moving forward, but yes I agree with you!

newIDE/app/src/UI/MarkdownText.js
ClementPasteau: As discussed, I've added a new prop `translatableSource` to be used alongside `source`.

I couldn't get the smart typing to work though.
When I define a component as `A | B` as soon as I start testing if prop `props.source` exists, it considers it's a component of type `A` and complains about using `B.translatableSource`. 
I searched for a better way to type, but nothing really obvious.

4ian: Ok, no big deal, these union types are always weird to work with.

ClementPasteau:
🥳

Issue #2644, opened 2021-05-13
Wrong shape painter color


PascalLadalle:
I'm not sure how to explain this bug...
The fill color is changing in cycles incrementing by 1 each cycle. (color change after 1 click, then color change after 2 clicks, then after 3 clicks...).

![4a01ec81a2d749cd6ecc9ae82660b29acc4d23fd](https://user-images.githubusercontent.com/25978830/118114702-3a7d3200-b3e8-11eb-9a42-ee9a650350dd.gif)

[Bug_ShapePainter.zip](https://github.com/4ian/GDevelop/files/6471918/Bug_ShapePainter.zip)

reported on [the forum](https://forum.gdevelop-app.com/t/problem-with-changing-fill-color-of-a-shape-painter/32569/6)

4ian:
The GIF seems not really exactly what is happening in the project, but I can confirm this is really weird. I've quickly checked but the fill color seems properly to stay the same internally.. a Pixi bug? Might have to dive more to check what's going on.

Bouh:
The fill color and outline color from the shape painter are wrong.
whe value shoud be on format **0x000000**, but there is just **255**
I don't know why the rgb to hex number return a weird value...

Bouh:
At first frame it's ok
![image](https://user-images.githubusercontent.com/1670670/118122464-2db20b80-b3f3-11eb-91a4-98df3ecf7366.png)

When we click, it's bugged
![image](https://user-images.githubusercontent.com/1670670/118122483-36a2dd00-b3f3-11eb-8f76-a7ca00137eed.png)

Bouh:
What happend :o 

![image](https://user-images.githubusercontent.com/1670670/118122781-98fbdd80-b3f3-11eb-9534-4f0cc6a542d0.png)

4ian:
How do you get the fill color to change? 

I've added a console.log:
![image](https://user-images.githubusercontent.com/1280130/118123102-aad98280-b3eb-11eb-94d6-e45f3a86668c.png)

and it's always the same value for me: 
![image](https://user-images.githubusercontent.com/1280130/118123135-bb89f880-b3eb-11eb-984b-ddcb3a914c7f.png)

Note that 255 *is* a valid value.
All blue color: 0x0000FF = 0xFF = 255 
All red color: 0xFF0000 = 16711680

PascalLadalle:
I tried again from scratch, just in case, and I confirm the issue.
Few tests I did:
[...]

PR #3026, opened 2021-09-10
Upgrade Pixi.js to 6.1.2

ClementPasteau:
The first purpose was to fix the issue #2644.
I tested the issue example and it is indeed fixed!

A lot of other fixes (and features) have been bundled in this release too: https://github.com/pixijs/pixijs/releases/tag/v6.1.0

What do you reckon is the best way to test such an upgrade @4ian ? Manual tests on multiple games ?


Bouh:
I would say
- [ ] Add all objects one by one by adding them to the scene.
- [ ] Start a preview with the different objects.
- [ ] Check all effects. [See here](https://github.com/4ian/GDevelop/pull/2447#issuecomment-808395980)

Bouh:
Visual unit test are also a good thing for the future :)
https://github.com/zaqqaz/visual-unit-tests

4ian:
> I would say
> 
> * [ ]  Read the previous PixiJS patch notes for checking any deprecated methods or updated.
> * [ ]  Add all objects one by one by adding them to the scene or check examples with these objects.
> * [ ]  Start a preview with the different objects.
> * [ ]  Check all effects. [See here](https://github.com/4ian/GDevelop/pull/2447#issuecomment-808395980)
>   Checking the effects is important PIXI.Filter is based on PixiJS core.

Seems like a good list. The idea is to check manually all objects and then a bunch of more or less random example to hope to catch any regression according to the changelog. Seems like this upgrade is fairly low risk overall!

ClementPasteau:
Nice! I'll try to do that at the end of the day

ClementPasteau:
did it a bit earlier :)

- Tested all objects and their render in the preview, everything looks fine.
- Tested multiple project examples
- Tested all particle effects with the project (https://github.com/GDevelopApp/GDevelop-examples/tree/main/examples/particle-effects-demo)
- Tested multiple layer effects with the effects demo in `GDJS/tests/games/effects/all effects demo.json`

The only thing that's different is the `Blur` Layer effect.
Previously, it seemed to crop the image, based on the number of render passes being done. (1 was cutting the image in half, 2 was hiding the image completely ??)
I assume this is a bug fixed rather than a bug introduced, am I wrong? @Bouh 

ezgif com-gif-maker

Bouh:
> I assume this is a bug fixed rather than a bug introduced, am I wrong? @Bouh

Correct it's like an unexpected fix ^^

Let me compile GDevelop with these changes and check. We are never too much to check this big API change.

4ian:
Looks solid, and maybe we got a bugfix for free! @Bouh let us know if you find anything else, otherwise I think we can go ahead!

4ian (approved):

Bouh:
Sounds good to me for the runtime and IDE!

- One thing on the developer side, the .map file shouldn't be updated too? 
https://github.com/4ian/GDevelop/blob/9aa6ce0485615ef12bbc14098baedbbc5dd38339/GDJS/Runtime/pixi-renderers/pixi.js#L10

- As the effects are also on objects I've tested some effects on objects and they work fine too.

4ian:
> One thing on the developer side, the .map file shouldn't be updated too?

Good catch, should be otherwise we'll have surprises when trying to debug :) (i.e: wrong lines everywhere I guess)

ClementPasteau:
interestingly the .map was never included 🤔 

I'll add it so we're in a better position if a problem occurs in the future

ClementPasteau:
I believe this will need more work than expected to add the .map for this file.
I couldn't get it to work and I can't find a similar example where a map is added directly from the code when it is not part of an extension.

As this is not causing any error in the console (it's always been like that) I'm suggesting to merge like this and we can see later if we really need the map.

4ian:
Yep let's go ahead as things are not worse (they are even strictly better if there are no regressions :)) than before!

PR #3042, opened 2021-09-13
Enable more warnings from Clang when compiling C++ sources

4ian:


ClementPasteau:
nice!

Where/when do the warnings appear? in the terminal when running a `npm run build` in `GDevelop.js`?

4ian:
Yes! All warnings should be shown btw in Travis once the compilation is done. 
I think I've avoided with the few warnings there were deactivated to have too much of them (that would make it hard to see new ones when working on a file).
If this goes well we can even enable `-Werror` in the future to fail the build for warnings (warnings in C++ are almost always important to fix).

ClementPasteau:
Shouldn't we fix the warnings and make it error right away to be on a clean base right away?

4ian:
> Shouldn't we fix the warnings and make it error right away to be on a clean base right away?

In theory yes, in practice my battery is running away and I preferred to push something useful rather than postpone it for later ;)

4ian:
> Shouldn't we fix the warnings and make it error right away to be on a clean base right away?

In theory yes, in practice my battery is running away and I preferred to push something useful rather than postpone it for later ;)

PR #3045, opened 2021-09-14
Add color picker to Add Object Color Tween action 'To color' parameter

### LINKED ISSUE #3040: Add color picker for color tweens
## Short Description

"Object color tween" action has no color picker.

## Description

When using a color tween, the final color ("To color") field requires a string for the color in the "r;g;b" format.
There is no color picker so choosing the correct color is difficult or requires to copy/paste the rgb string from another color picker.

## Solution suggested

Add color picker to "color tween" action

## Current action options

colorPicker

## Solution Example Mockup

colorPickerMockup

AlexandreSi:
Resolves #3040


AlexandreSi:
This fix seems quite straightforward to me as there is no type change for the parameter

ClementPasteau (approved):
lgtm! 🥳

4ian (approved):
🥳

PR #3050, opened 2021-09-15
Fix: Do not remove custom loading screen image when removing unused images

### LINKED ISSUE #3035: Image of loading screen
## Describe the bug
Hi mates can anyone check if this is happening too

immage-doesnt-exists

Refer to this thread to see that this happen to others too.
https://forum.gdevelop-app.com/t/is-this-a-bug-custom-splash-image/35022

Please double check that the bug is not already reported in the issues list.

## To Reproduce
1. When you add a custom image and later 
2. go to Resources and remove all unused images and 
3. go back to the loading screen you can clearly see that says the resource doesn’t exist in game

* If applicable, add screenshots to help explain your problem.

## Other details
* Windows 10
* GDevelop 117

AlexandreSi:
Resolves #3035

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -5,12 +5,28 @@
   "version": "0.2.0",
   "configurations": [
     {
-      "type": "pwa-chrome",
+      "type": "node",
       "request": "launch",
-      "name": "Launch Chrome against localhost",
-      "url": "http://localhost:3000",
-      "webRoot": "${workspaceFolder}",
-      "preLaunchTask": "Start development server"
+      "name": "Jest All GDevelop.js",
+      "program": "${workspaceFolder}/GDevelop.js/node_modules/.bin/jest",
+      "args": ["--runInBand"],
+      "console": "integratedTerminal",
+      "internalConsoleOptions": "neverOpen",
+      "disableOptimisticBPs": true,
+      "cwd": "${workspaceFolder}/GDevelop.js"
+    },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Jest Current GDevelop.js File",
+      "program": "${workspaceFolder}/GDevelop.js/node_modules/.bin/jest",
+      "args": [
+        "${fileBasenameNoExtension}"
+      ],
+      "console": "integratedTerminal",
+      "internalConsoleOptions": "neverOpen",
+      "disableOptimisticBPs": true,
+      "cwd": "${workspaceFolder}/GDevelop.js"
     }
   ]
 }
--- a/Core/GDCore/IDE/Project/ProjectResourcesAdder.cpp
+++ b/Core/GDCore/IDE/Project/ProjectResourcesAdder.cpp
@@ -42,15 +42,13 @@ std::vector<gd::String> ProjectResourcesAdder::GetAllUseless(
   std::set<gd::String>& usedResources = resourcesInUse.GetAll(resourceType);
 
   // Search all resources not used
-  std::vector<gd::String> resources =
-      project.GetResourcesManager().GetAllResourceNames();
+  std::vector<std::shared_ptr<Resource>> resources =
+      project.GetResourcesManager().GetAllResources();
   for (std::size_t i = 0; i < resources.size(); i++) {
-    if (project.GetResourcesManager().GetResource(resources[i]).GetKind() !=
-        resourceType)
-      continue;
+    if (resources[i]->GetKind() != resourceType) continue;
 
-    if (usedResources.find(resources[i]) == usedResources.end())
-      unusedResources.push_back(resources[i]);
+    if (usedResources.find(resources[i]->GetName()) == usedResources.end())
+      unusedResources.push_back(resources[i]->GetName());
   }
 
   return unusedResources;
--- a/Core/GDCore/IDE/Project/ProjectResourcesAdder.cpp
+++ b/Core/GDCore/IDE/Project/ProjectResourcesAdder.cpp
@@ -42,7 +42,7 @@ std::vector<gd::String> ProjectResourcesAdder::GetAllUseless(
   std::set<gd::String>& usedResources = resourcesInUse.GetAll(resourceType);
 
   // Search all resources not used
-  std::vector<std::shared_ptr<Resource>> resources =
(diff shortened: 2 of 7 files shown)


.vscode/launch.json
```diff
@@ -5,12 +5,28 @@
...
-      "type": "pwa-chrome",
+      "type": "node",
       "request": "launch",
-      "name": "Launch Chrome against localhost",
-      "url": "http://localhost:3000",
-      "webRoot": "${workspaceFolder}",
-      "preLaunchTask": "Start development server"
+      "name": "Jest All GDevelop.js",
+      "program": "${workspaceFolder}/GDevelop.js/node_modules/.bin/jest",
+      "args": ["--runInBand"],
```
4ian, line 12: Interesting! Is this required/recommended?

Core/GDCore/IDE/Project/ProjectResourcesAdder.cpp
```diff
@@ -42,15 +42,13 @@ std::vector<gd::String> ProjectResourcesAdder::GetAllUseless(
   std::set<gd::String>& usedResources = resourcesInUse.GetAll(resourceType);
 
   // Search all resources not used
-  std::vector<gd::String> resources =
-      project.GetResourcesManager().GetAllResourceNames();
+  std::vector<std::shared_ptr<Resource>> resources =
+      project.GetResourcesManager().GetAllResources();
```
4ian, line 46: nit: this works but also create a copy of the array of (smart) pointers into `resources`. While you only care about getting a... reference to it ;) 

Hence this should work:
```suggestion
  const std::vector<std::shared_ptr<Resource>>& resources =
      project.GetResourcesManager().GetAllResources();
```

(could also be written as `const auto& resources = ...` - which means "give me the resources vector, but just a reference to it. I won't even change it I swear.")

Core/GDCore/Project/Project.cpp
```diff
@@ -865,6 +865,10 @@ void Project::ExposeResources(gd::ArbitraryResourceWorker& worker) {
   for (std::size_t j = 0; j < GetObjectsCount(); ++j) {
     GetObject(j).ExposeResources(worker);
   }
+
+  // Add loading screen background image if present
+  if (loadingScreen.GetBackgroundImageResourceName() != "")
+    worker.ExposeImage(loadingScreen.GetBackgroundImageResourceName());
```
4ian, line 871: Excellent! 
It's a bit unusual to pass a string as reference, but it's how it works here and it's correct.

Core/GDCore/Project/ResourcesManager.h
```diff
@@ -447,6 +447,11 @@ class GD_CORE_API ResourcesManager {
    */
   std::shared_ptr<Resource> CreateResource(const gd::String& kind);
 
+  /**
+   * Get a list containing all the resources.
+   */
+  std::vector<std::shared_ptr<Resource>> GetAllResources() const { return resources; };
```
4ian, line 453: For my previous change recommendation to work, you must also change here the return type:

```suggestion
  const std::vector<std::shared_ptr<Resource>>& GetAllResources() const { return resources; };
```

This works because you say:
- I'm returning to you a reference to a vector (so you don't have to copy it in memory).
- Because this method is `const`, the reference I give back to you is also `const` - because it's my guarantee that you (the caller) won't try to modify it.

Core/GDCore/IDE/Project/ProjectResourcesAdder.cpp
4ian: Note that you need another change in the return value of `GetAllResources`, see my other comment.

GDevelop.js/__tests__/Core.js
```diff
@@ -1108,6 +1108,29 @@ describe('libGD.js', function () {
       expect(allResources.get(0)).toBe('Used');
       project.delete();
     });
+
+    it('should not remove loading screen image when removing useless resources', function () {
+      var project = gd.ProjectHelper.createNewGDJSProject();
+      var resource1 = new gd.ImageResource();
```
4ian, line 1114: nitpicking: you could as well use `const` here (by default in JS, I try to use `const` for all variables declaration, unless it will change later) (this avoided a few errors in the past, where you try to re-assign a variable by mistake).

GDevelop.js/__tests__/Core.js
```diff
@@ -1108,6 +1108,29 @@ describe('libGD.js', function () {
       expect(allResources.get(0)).toBe('Used');
       project.delete();
     });
+
+    it('should not remove loading screen image when removing useless resources', function () {
```
4ian, line 1112: Excellent test 👍

GDevelop.js/__tests__/Core.js
```diff
@@ -1108,6 +1108,29 @@ describe('libGD.js', function () {
...
+      var project = gd.ProjectHelper.createNewGDJSProject();
+      var resource1 = new gd.ImageResource();
+      resource1.setName('LoadingScreenImage');
+      project.getLoadingScreen().setBackgroundImageResourceName(resource1.getName())
+
+      expect(project.getLoadingScreen().getBackgroundImageResourceName()).toBe('LoadingScreenImage');
+
+      let worker = new gd.ResourcesInUseHelper();
```
4ian, line 1120: Uber Nitpicking: const instead of let (here and below)

(The JS community is raging on this issue, but overall I go for the most specific keyword, `const`, and fallback to `let` when not needed)
(other newer language, I think Rust, have a keyword `let` and `let mut` for mutable stuff 
- validating that you should prefer const as much as possible)

4ian (approved):
Nice job! Added a bunch of nitpickings and something to avoid a useless copy in memory of an array in C++. Good to go once this is fixed 👍

.vscode/launch.json
AlexandreSi: I used what was recommended in https://github.com/microsoft/vscode-recipes/tree/master/debugging-jest-tests

GDevelop.js/__tests__/Core.js
AlexandreSi: 💜

ClementPasteau (approved):
nice one! :)

4ian:
Looks good to go!

PR #3051, opened 2021-09-16
Add automatic addition of extensions, examples and assets commits in the changelog

4ian:
Also ensure scripts are type checked.

Don't show in changelog

### DIFF
--- a/newIDE/app/scripts/extract-changelog.js
+++ b/newIDE/app/scripts/extract-changelog.js
@@ -1,6 +1,6 @@
 // @ts-check
 /**
- * This is ascript getting the commits since the last tag (which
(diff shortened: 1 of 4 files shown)


newIDE/app/scripts/extract-changelog.js
```diff
@@ -1,17 +1,43 @@
 // @ts-check
 /**
- * This is a simple script getting the commits since the last tag (which
+ * This is ascript getting the commits since the last tag (which
```
Bouh, line 3:
```suggestion
 * This is a script getting the commits since the last tag (which
```

4ian: Thank you!

PR #3052, opened 2021-09-16
adapt new collision mask to object size

### LINKED ISSUE #3037: Custom collision masks should start with the default collision mask (instead of 32x32)
## Description

The default collision mask matches the size of the object (AABB).  When user creates a new collision mask, a 32x32 collision mask appears in the center of the object.

This will cause some users to think the default hitbox was 32x32 (this is what I concluded)

Screenshot 2021-09-10 223214
Screenshot 2021-09-10 223134

## Solution suggested

When a user creates a new collision mask, make it the same size and location as the default hitbox (the outside corners of the object).

AlexandreSi:
Resolves #3037

### DIFF
--- a/newIDE/app/src/ObjectEditor/Editors/SpriteEditor/CollisionMasksEditor/PolygonsList.js
+++ b/newIDE/app/src/ObjectEditor/Editors/SpriteEditor/CollisionMasksEditor/PolygonsList.js
@@ -208,20 +208,20 @@ const PolygonsList = (props: PolygonsListProps) => {
     restoreCollisionMask,
   } = props;
 
-  const addCollisionMask = () => {
+  const addCollisionMask = React.useCallback(() => {
     const newPolygon = gd.Polygon2d.createRectangle(spriteWidth, spriteHeight);
     newPolygon.move(spriteWidth / 2, spriteHeight / 2);
     polygons.push_back(newPolygon);
     onPolygonsUpdated();
-  };
+  }, [spriteHeight, spriteWidth, polygons, onPolygonsUpdated]);
 
   React.useEffect(
     () => {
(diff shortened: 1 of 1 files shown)


newIDE/app/src/ObjectEditor/Editors/SpriteEditor/CollisionMasksEditor/PolygonsList.js
```diff
@@ -208,23 +208,20 @@ const PolygonsList = (props: PolygonsListProps) => {
...
       if (polygons.size() === 0) {
-        const newPolygon = gd.Polygon2d.createRectangle(32, 32);
-        newPolygon.move(spriteWidth / 2, spriteHeight / 2);
-        polygons.push_back(newPolygon);
-        onPolygonsUpdated();
+        addCollisionMask();
       }
     },
-    [polygons, spriteWidth, spriteHeight, onPolygonsUpdated]
+    [polygons, spriteWidth, spriteHeight, onPolygonsUpdated, addCollisionMask]
```
4ian, line 224: Nit: You can even remove spriteWidth/spriteHeight/onPolygonsUpdated. 
Per se it's not invalid, just redundant because addCollisionMask will be changed if they are ;)

4ian (approved):
Awesome, thanks for the duplication removal 😍

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/CollisionMasksEditor/PolygonsList.js
AlexandreSi: Oh you're right! I'll fix this

PR #3056, opened 2021-09-17
Feature/display extension username

ClementPasteau:
Now that the backend returns the list of authors with the extensions database, we are able to display them on the Extensions Search list and on the Extension details popup.

This handles the case where there are no authors.

I also fixed an annoying bug, where the list would not compute the rows height correctly on first render, triggering a jump when scrolling after the next item is displayed.

Preview: 
Screenshot 2021-09-17 at 10 46 34
Screenshot 2021-09-17 at 10 46 08
Screenshot 2021-09-17 at 10 45 30

### DIFF
--- a/newIDE/app/src/AssetStore/ExtensionStore/ExtensionInstallDialog.js
+++ b/newIDE/app/src/AssetStore/ExtensionStore/ExtensionInstallDialog.js
@@ -20,6 +20,7 @@ import { Column, Line } from '../../UI/Grid';
 import { Divider } from '@material-ui/core';
 import { ColumnStackLayout } from '../../UI/Layout';
 import { IconContainer } from '../../UI/IconContainer';
+import { UserPublicProfileChip } from '../../UI/UserPublicProfileChip';
 
 type Props = {|
   extensionShortHeader: ExtensionShortHeader,
@@ -136,7 +137,7 @@ export default class ExtensionInstallDialog extends Component<Props, State> {
             <IconContainer
               alt={extensionShortHeader.fullName}
               src={extensionShortHeader.previewIconUrl}
-              size={48}
+              size={64}
             />
             <Column expand>
               <Text noMargin size="title">
@@ -145,6 +146,12 @@ export default class ExtensionInstallDialog extends Component<Props, State> {
               <Text noMargin size="body2">
                 <Trans>Version {' ' + extensionShortHeader.version}</Trans>
               </Text>
+              <Line>
+                {extensionShortHeader.authors &&
+                  extensionShortHeader.authors.map(author => (
+                    <UserPublicProfileChip user={author} key={author.id} />
+                  ))}
+              </Line>
             </Column>
           </Line>
           <Text noMargin>{extensionShortHeader.shortDescription}</Text>
--- a/newIDE/app/src/AssetStore/ExtensionStore/ExtensionListItem.js
+++ b/newIDE/app/src/AssetStore/ExtensionStore/ExtensionListItem.js
@@ -6,6 +6,7 @@ import Text from '../../UI/Text';
 import { Trans } from '@lingui/macro';
 import { Column, Line } from '../../UI/Grid';
 import { IconContainer } from '../../UI/IconContainer';
+import { UserPublicProfileChip } from '../../UI/UserPublicProfileChip';
 
 const styles = {
   container: {
@@ -43,17 +44,23 @@ export const ExtensionListItem = ({
   return (
     <ButtonBase onClick={onChoose} focusRipple>
       <div style={styles.container} ref={containerRef}>
-        <Line noMargin>
+        <Line>
           <IconContainer
             alt={extensionShortHeader.fullName}
             src={extensionShortHeader.previewIconUrl}
-            size={48}
+            size={64}
           />
           <Column expand>
             <Text noMargin>
               {extensionShortHeader.fullName}{' '}
               {alreadyInstalled && <Trans> (already installed)</Trans>}
             </Text>
+            <Line>
+              {extensionShortHeader.authors &&
+                extensionShortHeader.authors.map(author => (
+                  <UserPublicProfileChip user={author} key={author.id} />
+                ))}
+            </Line>
             <Text noMargin size="body2">
               {extensionShortHeader.shortDescription}
             </Text>
--- a/newIDE/app/src/UI/Search/ListSearchResults.js
+++ b/newIDE/app/src/UI/Search/ListSearchResults.js
@@ -122,7 +122,13 @@ export const ListSearchResults = <SearchItem>({
 
             return (
               <Grid
-                ref={grid}
+                ref={el => {
+                  if (el) {
+                    // Ensure the grid is recomputed for heights once it is rendered.
+                    el.recomputeGridSize(0, 0);
+                  }
+                  grid.current = el;
+                }}
                 width={width}
                 height={height}
                 columnCount={1}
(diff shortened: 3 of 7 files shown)


newIDE/app/src/UI/Search/ListSearchResults.js
```diff
@@ -122,7 +122,13 @@ export const ListSearchResults = <SearchItem>({
 
             return (
               <Grid
-                ref={grid}
+                ref={el => {
+                  if (el) {
```
4ian, line 126: Interesting! I think calling `recomputeGridSize` is the right thing to do (per the doc: "This function should be called if dynamic column or row sizes have changed but nothing else has" => exactly our situation).

Can you check if:
- this works well on everything using a ListSearchResults (i.e: game examples too and the game showcase).
- if instead of "hijacking" the ref callback function, we could use a `React.useLayoutEffect`? With empty dependencies to simulate a `componentDidMount`?

According to the docs, "React will call the ref callback with the DOM element when the component mounts, and call it with null when it unmounts.". 
So we should be able to replace it with a useLayoutEffect as it "fires in the same phase as componentDidMount". 

Let me know if it works or not

newIDE/app/src/AssetStore/ExtensionStore/ExtensionInstallDialog.js
```diff
@@ -145,6 +157,20 @@ export default class ExtensionInstallDialog extends Component<Props, State> {
...
                 <Trans>Version {' ' + extensionShortHeader.version}</Trans>
               </Text>
+              <div style={styles.authors}>
+                {extensionShortHeader.authors && (
+                  <>
+                    {extensionShortHeader.authors.map(author => (
+                      <Chip
+                        size="small"
+                        style={styles.chip}
+                        label={author}
```
ClementPasteau, line 167: mistake here, this should be author.username

newIDE/app/src/AssetStore/ExtensionStore/ExtensionListItem.js
```diff
@@ -43,17 +51,31 @@ export const ExtensionListItem = ({
...
               {alreadyInstalled && <Trans> (already installed)</Trans>}
             </Text>
+            <div style={styles.authors}>
+              {extensionShortHeader.authors && (
+                <>
+                  {extensionShortHeader.authors.map(author => (
+                    <Chip
+                      size="small"
+                      style={styles.chip}
+                      label={author}
```
ClementPasteau, line 72: same here

newIDE/app/src/Utils/GDevelopServices/Extension.js
```diff
@@ -5,6 +5,7 @@ import semverSatisfies from 'semver/functions/satisfies';
 
 export type ExtensionShortHeader = {|
   shortDescription: string,
+  authors?: Array<string>,
```
4ian, line 8: (to be sure it's not forgot: new type here to be imported from User.js presumably (we keep types defined in the same API as they are in the backend).

newIDE/app/src/UI/Search/ListSearchResults.js
ClementPasteau: - checked on all ListSearchResults, it looks great :+1:
- I tried `useLayoutEffect`, unfortunately it gets called too early, the grid reference hasn't been set yet, it is still at its initial value `null`

4ian: Thanks for checking!

Bouh:
Chips or cheap?
We can go with a simpler visual, nicknames become a tag name. It could be clickable in the future.
It's easier to understand that it's a user with arobase.
image

ClementPasteau:
@Bouh There is a plan to add an avatar too :) which is probably why we thought of the Chip as it would work out quite well.
But it's not gonna happen before a few weeks.
The goal is for the username to be clickable later yes, to access the profile.

How about those? 
Face:
Screenshot 2021-09-17 at 17 10 16
Head:
Screenshot 2021-09-17 at 17 10 57

Bouh:
Face for sure 🤩

Bouh (approved):

newIDE/app/src/AssetStore/ExtensionStore/ExtensionInstallDialog.js
```diff
@@ -145,6 +157,20 @@ export default class ExtensionInstallDialog extends Component<Props, State> {
               <Text noMargin size="body2">
                 <Trans>Version {' ' + extensionShortHeader.version}</Trans>
               </Text>
+              <div style={styles.authors}>
```
4ian, line 160: Use `<Line noMargin>` (Line/Column are what we use for most layouts). This avoids custom styles (like `authors`) can ensure consistency of margins + avoid the notion of "divs" in the UI.

newIDE/app/src/AssetStore/ExtensionStore/ExtensionInstallDialog.js
```diff
@@ -145,6 +157,20 @@ export default class ExtensionInstallDialog extends Component<Props, State> {
...
+                {extensionShortHeader.authors && (
+                  <>
+                    {extensionShortHeader.authors.map(author => (
+                      <Chip
+                        size="small"
+                        style={styles.chip}
+                        label={author.username}
+                        key={author.username}
+                      />
```
4ian, line 169: I know it's early, but I would move this already into a `<UserPublicProfileChip user={author} />`

This will give the opportunity to later add the avatar everywhere at once, customize the color, etc.. according to the UserPublicProfile

newIDE/app/src/AssetStore/ExtensionStore/ExtensionListItem.js
```diff
@@ -43,17 +51,31 @@ export const ExtensionListItem = ({
...
             src={extensionShortHeader.previewIconUrl}
-            size={48}
+            size={64}
           />
           <Column expand>
             <Text noMargin>
               {extensionShortHeader.fullName}{' '}
               {alreadyInstalled && <Trans> (already installed)</Trans>}
             </Text>
+            <div style={styles.authors}>
```
4ian, line 65: Same here, `UserPublicProfileChip` + `Line`, so you can remove all the custom styles in this file and avoid relying on material-ui, but on a `UI/` component.

newIDE/app/src/Utils/GDevelopServices/Extension.js
```diff
@@ -5,6 +5,7 @@ import semverSatisfies from 'semver/functions/satisfies';
 
 export type ExtensionShortHeader = {|
   shortDescription: string,
+  authors?: Array<{ id: string, username: string }>,
```
4ian, line 8: Let's go ahead and have a `Array<UserPublicProfile>` with `UserPublicProfile` in a User.js file. I know it's maybe thinking too much right now, but this will ensure the consistency of the user profile information everywhere and this will help finding all the places in the codebase using it when we add new attributes :) (avatar, color, badges...)

4ian (approved):
Looks good to go after my comments are addressed!

As a rule of thumb, here is how I work:
- Whenever I use a `@material-ui` import in a non `UI/` file, I wonder if we could abstract it in a UI component, if it's going to be used more than once (not a strict rule)
- Whenever I see some custom `styles` in a file outside `UI`, it's a good sign it might also be abstracted. Ideally anything outside `UI` would never use any style - it's the `UI` components job to do this.
- Layout can most of the time be done by assembling, `<Column>`,`<Line>` (and their props used very often like `expand`, `noMargin`), or better: `ResponsiveLineStackLayout`,  `ColumnStackLayout`.

Apart from that, looks good 👍

Issue #2821, opened 2021-07-18
Particle Emitters - "Change the capacity of..." event NOT working


erdo-dp:
## Describe the bug
Particle Emitters capacity value are not being changed updated trough the "Change the capacity of..." event. The original value persist.

## Example to Reproduce
[game.json.zip](https://github.com/4ian/GDevelop/files/6836117/game.json.zip)

## Other details
* All betas since GDevelop 5 exist

Bouh:
Seem this isn't working, I can't found `totalParticleCount` in emitter with the debugger on a breakpoint.

```JS
 getTotalParticleCount(): integer {
      return this.emitter.totalParticleCount;
    }
```

[The documentation of PIXI.Particles ](http://pixijs.io/pixi-particles/docs/classes/emitter.html )confirm there is no total `totalParticleCount` in emitter.

4ian:
Might be something dating from the port from SFML SPARK particles to pixi particles.

AlexandreSi:
Thanks for reporting this bug, the particle emitter needed a good clean!

tristanbob:
So.... is changing the capacity of an emitter supposed to work now?  

https://user-images.githubusercontent.com/8879811/139517935-dbf271d6-c4cf-4826-8b5a-b34bf44dd148.mp4

(This is using the web version of GDevelop)

4ian:
It should work, the emitter should stop emitting after `capacity/flow` seconds.
I've just tested on the web app:
- setting the capacity to 45 and flow to 45 will make the emitter last 1 second (ok!)
- change the capacity to 90 in the events will make the emitter last 2 second (ok!)
- change the capacity to 10 in the events will make the emitter last <1 second (ok!)
- setting the capacity to -1 and changing it to 10 in the events will make the emitter last <1 second (ok!)

In other words: it seems to work well for me?

tristanbob:
I should have tested this more.  On my desktop (5.0.121) this works great. 

When I tested (again) on the web version, it had updated to that version and of course, it worked. My first test must have been on an unfixed version.

Thanks @AlexandreSi for your work on this!

4ian:
No problem, this is actually a good sign we need a notification in the web-app that an update is available! :)

PR #3058, opened 2021-09-17
Fix particle emitter

### LINKED ISSUE #2285: Condition "ParticleEmitter does not emit anylonger" doesn't trigger
## Describe the bug

When using the Condition "ParticleEmitter does not emit anylonger", it doesn't trigger even when an emitter runs out of particles.

## To Reproduce

- Create a particle emitter using default settings
- Change "Number of particles in tank" to "100"
- Create a condition for "ParticleEmitter does not emit anylonger" with an action to change the background color
- Preview game (the emitter will stop emitting but the background color doesn't change)

## Other details

- Tested on GDevelop 5 Beta 104

## Video

https://user-images.githubusercontent.com/8879811/107173246-9c12a800-6984-11eb-8246-659c7ecbff28.mp4

### LINKED ISSUE #2313: Particle friction is not working.
## Describe the bug
The "Friction on particles" of the particles emitter object is not working.

## To Reproduce
1. Create a particle emitter using default settings.
2. Change "Friction on particles" to a larger number.
3. Preview the game, you can see that the particles are not slowing down.

https://user-images.githubusercontent.com/68310169/108294606-7e8ddb80-71d9-11eb-879b-021d0543cf76.mp4

## Other details

- Tested on GDevelop 5 Beta 105 (Desktop app and Web app)

AlexandreSi:
Resolves #2285, #2821
Resolves #2313 as there is no support for friction in pixi.


Extensions/ParticleSystem/particleemitterobject.ts
```diff
@@ -680,75 +680,75 @@ namespace gdjs {
...
+    setParticleAlpha2(alpha: number): void {
       if (this.alpha2 !== alpha) {
         this._alphaDirty = true;
         this.alpha2 = alpha;
       }
     }
 
-    noMoreParticles() {
+    noMoreParticles(): boolean {
       this._renderer.stop();
```
4ian, line 717: The two commits are for now adding types - which is great! But not fixing the mentioned bugs right?

AlexandreSi: Yes, I forgot to mention it's still a draft. I wanted the issues to mention that we are working on these issues

AlexandreSi:
@4ian the expression `getParticleCount` (`NbParticles`) seems misleading to me: We can understand it's the number of particles emitted since the emitter started emitting whereas it's the number of active particles emitted at a particular time (https://pixijs.io/particle-emitter/docs/classes/Emitter.html#particleCount).
I think we should remove it, but it may break games. Is there a way of doing it properly?

4ian:
We can do two things in this case:
- Either we think the expression name itself is fine, and it's a matter of updating the description/user friendly name.
- Or we consider the name "NbParticle" is not great (it's misleading indeed I think), and in this case we: 
  - Add an expression that is better suited (better name, like `CurrentParticleCount`, better description, etc...) to this usage.
  - Keep the old one, but mark it with `.SetHidden(); // Deprecated`.
  
So that we keep backward compatibility with older games (mandatory, we don't want to make upgrading to a newer version a pain, especially for something that is not that complicated for us to maintain) and still ensure games will start using the new expression from now.

In the future, we can work on something that will browse events and that will automatically replace deprecated expressions (or actions/conditions) by their replacement. But this means also losing the ability to open your game in an older version of GDevelop, which is ok but could be a problem if a new version is buggy and you want to temporarily downgrade to the older one. So it's a good idea to have the two expressions (the hidden one and the new one) for a little bit :)

ClementPasteau:
nice cleanup 🧹 🥳

Extensions/ParticleSystem/particleemitterobject.ts
```diff
@@ -88,10 +84,10 @@ namespace gdjs {
     _colorDirty: boolean = true;
     _sizeDirty: boolean = true;
     _alphaDirty: boolean = true;
-    _textureDirty: boolean;
-
-    // Don't mark texture as dirty if not using one.
```
4ian, line 93: Nice, these comments got their location messed up a while ago when codemoding from JS+JSdoc to full TypeScript.

Extensions/ParticleSystem/particleemitterobject.ts
```diff
@@ -680,75 +653,76 @@ namespace gdjs {
...
 
-    getTank() {
+    getTank(): number {
       return this.tank;
     }
 
-    setTank(tank): void {
+    setTank(tank: number): void {
       this.tank = tank;
+      this._tankDirty = true;
```
4ian, line 718: nit: let's do a `if (this.tank !== tank) {` to avoid marking the tank as dirty if that's not the case. More practical for users not to have to check this by themselves, especially because this reset the emission. Otherwise, an event like:

Conditions : none
Actions: Set the tank of MyParticleEmitter to 1000

would result in a "perpetually resetting emission" (so no particles emitted), which would be considered as a bug/confusing behavior for users.

4ian (approved):
Great clean up, was definitely overdue! 
Check my comments, then after that sounds good to go!

Extensions/ParticleSystem/particleemitterobject.ts
AlexandreSi: I tried but I saw an issue with that: one could set the tank to the same capacity as previously to have it emit the same quantity. The issue is that `this.tank` is actually the initial tank capacity, not the current one.

The event that you took as an example would actually, on each step, recompute the emitter lifetime and would emit until the end of times. It wouldn't be that unexpected to me.

4ian: > The issue is that this.tank is actually the initial tank capacity, not the current one.

This is very true!

> The event that you took as an example would actually, on each step, recompute the emitter lifetime and would emit until the end of times. It wouldn't be that unexpected to me.

Could be indeed not unexpected... could we make a simple test game to see the behavior? I was worried resetting the emission would also delete the existing particles, which would result in the player seeing nothing on screen (because every frame the tank is changed and the emission is reset - but I'm not sure if this is what happens).

AlexandreSi: particleEmitterInfiniteTest

Here it is!

For the record, we could have an expression that returns the number of emitted/remaining particles but it cannot be reliable for the following reason:

- Pixi does not have the concept of tank/capacity that is present in GDevelop, we do the conversion to frequency and lifetime that pixi actually uses.
- On update, pixi will check if it has to emit particles comparing elapsed time to the period of emission.
- If it finds out particles should be emitted, it will first check that there is not too much particles already displayed (via the max parameter).
- If indeed the maximum has been reached, it does not emit more particle and there is no mechanism to say "We should have emitted X particles but we couldn't, so we will keep a record for that and emit them later, extending the emitter lifetime."

I don't know if this should be explained in the editor.

4ian: Thanks for checking and for the GIF. I think it's fine then as you did, and that as long as the emission is done and visible, it's fine to "reset" it when the tank is modified.

If that generates confusion in the future, we can always hide the action to modify the tank and implement new ones that deals with the emitter lifetime/frequency directly.
But for now I think it's pretty good like this then!

PR #3064, opened 2021-09-18
[Draggable] Refactor the behavior to reduce the memory taken.

D8H:
This is to prepare a change to take into account the Z-Order when grabbing an object.

It creates a manager per pointer that handles the dragging. As the offset is stored in the managers instead of the behaviors, this will take less memory when there is a lot of objects.
Also, the code is more DRY. So, it will be easier to add new constraints like on the Z-Order for instance.

### DIFF
--- a/Extensions/DraggableBehavior/draggableruntimebehavior.ts
+++ b/Extensions/DraggableBehavior/draggableruntimebehavior.ts
@@ -1,6 +1,6 @@
 /*
 GDevelop - Draggable Behavior Extension
-Copyright (c) 2013-2016 Florian Rival (Florian.Rival@gmail.com)
+Copyright (c) 2013-2021 Florian Rival (Florian.Rival@gmail.com)
  */
 
 namespace gdjs {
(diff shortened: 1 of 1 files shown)


4ian:
Nice! I'm always a bit afraid to have a style that is too much "OOP" in the game engine (and in JS in general, where simple functions can be more lightweight/less verbose/as fast/faster). This means to create a nice distinction here though in the logic of mouse vs pointers.

> As the offset is stored in the managers instead of the behaviors, this will take less memory when there is a lot of objects.

That's great indeed, we reduce the memory overhead to one pointer (null or pointing to the manager dragging the object).

I think we really have to come up with a way though to store things (like these "managers") linked to a runtime scene without polluting runtime scene - this is becoming dangerous as the code grow (not type safe, risk of name collisions).

4ian (approved):

Extensions/DraggableBehavior/draggableruntimebehavior.ts
```diff
@@ -145,25 +101,188 @@ namespace gdjs {
...
+    constructor(runtimeScene: gdjs.RuntimeScene) {}
+
+    /**
+     * Get the platforms manager of a scene.
+     */
+    static getMouseManager(
+      runtimeScene: gdjs.RuntimeScene
+    ): MouseDraggableManager {
+      // @ts-ignore
+      if (!runtimeScene.mouseDraggableManager) {
```
4ian, line 135: We could have a script analyzing the codebase to check all usage of `runtimeScene.xxx`. A better way would be in the future to use a storage per runtime scene that is also scoped to each extension.

PR #3066, opened 2021-09-18
[Draggable] Drag the object the most in front.

D8H:
There is another PR with only the refactor: https://github.com/4ian/GDevelop/pull/3064

This add a constraint on Z-order.

This is an example with cards that are stacked. With the new constraint, it allows the user to drag the card that is actually shown under the pointer.
https://www.dropbox.com/s/0v113sqm2vi9kds/Klondike.zip?dl=1

### DIFF
--- a/Extensions/DraggableBehavior/draggableruntimebehavior.ts
+++ b/Extensions/DraggableBehavior/draggableruntimebehavior.ts
@@ -39,6 +39,10 @@ namespace gdjs {
       this._draggedByDraggableManager = null;
     }
 
+    _dismissDrag() {
+      this._draggedByDraggableManager = null;
+    }
+
     _tryBeginDrag(runtimeScene) {
       if (this._draggedByDraggableManager) {
         return false;
@@ -195,7 +199,7 @@ namespace gdjs {
       }
       if (this._draggableBehavior) {
         // The previous best object to drag will not be dragged.
-        this._draggableBehavior._draggedByDraggableManager = null;
+        this._draggableBehavior._dismissDrag();
       }
       this._draggableBehavior = draggableRuntimeBehavior;
       this._xOffset = position[0] - draggableRuntimeBehavior.owner.getX();
--- a/Extensions/DraggableBehavior/draggableruntimebehavior.ts
+++ b/Extensions/DraggableBehavior/draggableruntimebehavior.ts
@@ -119,7 +119,15 @@ namespace gdjs {
    * Handle the dragging
    */
   abstract class DraggableManager {
+    /**
+     * The object has left its original position.
+     * When true, the search for the best object to drag has ended.
+     */
     protected _draggingSomething = false;
+    /**
+     * The behavior of the object that is best one found so far.
+     */
+    protected _draggableBehavior: gdjs.DraggableRuntimeBehavior | null = null;
     protected _xOffset: number = 0;
     protected _yOffset: number = 0;
 
@@ -172,15 +180,26 @@ namespace gdjs {
       runtimeScene: gdjs.RuntimeScene,
       draggableRuntimeBehavior: DraggableRuntimeBehavior
     ) {
+      if (
+        this._draggableBehavior &&
+        draggableRuntimeBehavior.owner.getZOrder() <=
+          this._draggableBehavior.owner.getZOrder()
+      ) {
+        return false;
+      }
       const position = this.getPosition(runtimeScene, draggableRuntimeBehavior);
       if (
         !draggableRuntimeBehavior.owner.insideObject(position[0], position[1])
       ) {
         return false;
       }
+      if (this._draggableBehavior) {
+        // The previous best object to drag will not be dragged.
+        this._draggableBehavior._draggedByDraggableManager = null;
+      }
+      this._draggableBehavior = draggableRuntimeBehavior;
       this._xOffset = position[0] - draggableRuntimeBehavior.owner.getX();
       this._yOffset = position[1] - draggableRuntimeBehavior.owner.getY();
-      this._draggingSomething = true;
       return true;
     }
 
@@ -189,12 +208,19 @@ namespace gdjs {
       draggableRuntimeBehavior: DraggableRuntimeBehavior
     ) {
       const position = this.getPosition(runtimeScene, draggableRuntimeBehavior);
-      draggableRuntimeBehavior.owner.setX(position[0] - this._xOffset);
-      draggableRuntimeBehavior.owner.setY(position[1] - this._yOffset);
+      if (
+        draggableRuntimeBehavior.owner.getX() != position[0] - this._xOffset &&
+        draggableRuntimeBehavior.owner.getY() != position[1] - this._yOffset
+      ) {
+        draggableRuntimeBehavior.owner.setX(position[0] - this._xOffset);
+        draggableRuntimeBehavior.owner.setY(position[1] - this._yOffset);
+        this._draggingSomething = true;
+      }
     }
 
     endDrag() {
       this._draggingSomething = false;
+      this._draggableBehavior = null;
     }
 
     abstract isDragging(
--- a/Extensions/DraggableBehavior/draggableruntimebehavior.ts
+++ b/Extensions/DraggableBehavior/draggableruntimebehavior.ts
@@ -129,7 +129,7 @@ namespace gdjs {
      */
     protected _draggingSomething = false;
     /**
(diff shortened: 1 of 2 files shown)


Extensions/DraggableBehavior/draggableruntimebehavior.ts
```diff
@@ -172,15 +180,26 @@ namespace gdjs {
...
+      }
       const position = this.getPosition(runtimeScene, draggableRuntimeBehavior);
       if (
         !draggableRuntimeBehavior.owner.insideObject(position[0], position[1])
       ) {
         return false;
       }
+      if (this._draggableBehavior) {
+        // The previous best object to drag will not be dragged.
+        this._draggableBehavior._draggedByDraggableManager = null;
```
4ian, line 198: nitpicking: I think we could make this (even if it's in a simple line) into a `dismissDrag`? Might help making it clear that we're not ending the dragging, we're dismissing it before it was even started.

Extensions/DraggableBehavior/tests/draggableruntimebehavior.spec.js
```diff
@@ -51,112 +51,211 @@ describe('gdjs.DraggableRuntimeBehavior', function () {
...
-    runtimeGame.getInputManager().onTouchStart(1, 750, 600);
-    runtimeScene.renderAndStep(1000 / 60);
-    runtimeGame.getInputManager().onTouchMove(1, 850, 700);
-    runtimeScene.renderAndStep(1000 / 60);
-    runtimeGame.getInputManager().onTouchEnd(1);
-    runtimeGame.getInputManager().onFrameEnded();
-
-    expect(object.getX()).to.be(850);
-    expect(object.getY()).to.be(700);
+  describe.skip('(mouse)', function () {
```
4ian, line 54: Remove the `skip`s ;)

Extensions/DraggableBehavior/tests/draggableruntimebehavior.spec.js
```diff
@@ -51,112 +51,211 @@ describe('gdjs.DraggableRuntimeBehavior', function () {
...
+      runtimeScene.renderAndStep(1000 / 60);
+      runtimeGame
+        .getInputManager()
+        .onMouseButtonReleased(gdjs.InputManager.MOUSE_LEFT_BUTTON);
+      runtimeScene.renderAndStep(1000 / 60);
+
+      expect(object.getX()).to.be(750);
+      expect(object.getY()).to.be(600);
+
+      //Mouse move without dragging
```
4ian, line 75: nitpicking: convention for comments is a space after // and a period at the end (C# convention, though I think it's recommended in most languages: https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions#commenting-conventions)

Extensions/DraggableBehavior/draggableruntimebehavior.ts
```diff
@@ -189,12 +208,19 @@ namespace gdjs {
...
       const position = this.getPosition(runtimeScene, draggableRuntimeBehavior);
-      draggableRuntimeBehavior.owner.setX(position[0] - this._xOffset);
-      draggableRuntimeBehavior.owner.setY(position[1] - this._yOffset);
+      if (
+        draggableRuntimeBehavior.owner.getX() != position[0] - this._xOffset &&
+        draggableRuntimeBehavior.owner.getY() != position[1] - this._yOffset
+      ) {
+        draggableRuntimeBehavior.owner.setX(position[0] - this._xOffset);
+        draggableRuntimeBehavior.owner.setY(position[1] - this._yOffset);
+        this._draggingSomething = true;
```
4ian, line 217: Interesting, is this causing a change that could be visible for the user? Or is this almost invisible because dragging will always result in a change of position?

4ian (approved):
Seems sensible I think! See a few nitpicking comments.

Extensions/DraggableBehavior/draggableruntimebehavior.ts
D8H: There should be no impact, this is just stop trying to find a better object to drag. It uses the fact that the object is not move the 1st frame the condition "is being dragged" become true. I think that was a really good choice because It makes it easier for events to keep the original position. When I wrote an events behavior to cancel a dragging, it really came in handy. I thought it wouldn't have worked at 1st.

4ian: > I think that was a really good choice because It makes it easier for events to keep the original position. When I wrote an events behavior to cancel a dragging, it really came in handy. I thought it wouldn't have worked at 1st.

Yep I think in general behaviors should prioritize their work/logic before the events, in the doStepPreEvents, so that events can always have the "last word" on what is done (and take the opportunity to adjust positions/cancel something/react to something before the render...).
There can be exception but good to see that this was a good choice here :)

Extensions/DraggableBehavior/draggableruntimebehavior.ts
```diff
@@ -119,7 +119,15 @@ namespace gdjs {
    * Handle the dragging
    */
   abstract class DraggableManager {
+    /**
+     * The object has left its original position.
+     * When true, the search for the best object to drag has ended.
+     */
     protected _draggingSomething = false;
+    /**
+     * The behavior of the object that is best one found so far.
```
4ian, line 132: nitpicking: 
```suggestion
     * The behavior of the object that is being dragged and that is the best one (i.e: highest Z order) found.
```

D8H: I wanted to explain that the value can still change until all the behaviors doStepPreEvents are called, but it was more confusing than anything.

Extensions/DraggableBehavior/tests/draggableruntimebehavior.spec.js
```diff
@@ -51,112 +51,211 @@ describe('gdjs.DraggableRuntimeBehavior', function () {
...
+      runtimeScene.renderAndStep(1000 / 60);
+      runtimeGame.getInputManager().onTouchStart(1, 10, 20);
+      runtimeGame.getInputManager().onTouchStart(0, 450, 500);
+      runtimeScene.renderAndStep(1000 / 60);
+      runtimeGame.getInputManager().onFrameEnded();
+      runtimeGame.getInputManager().onTouchMove(0, 750, 600);
+      runtimeScene.renderAndStep(1000 / 60);
+      runtimeGame.getInputManager().onFrameEnded();
+      runtimeGame.getInputManager().onTouchEnd(0);
+      runtimeGame.getInputManager().onTouchEnd(1);
```
4ian, line 158: Why finishing the touch #1? There are `onTouchMove(1, ` later that are called for this pointer and this seems to break the guarantee that input manager will have on touch start -> move > end called in this order.

Extensions/DraggableBehavior/tests/draggableruntimebehavior.spec.js
```diff
@@ -51,112 +51,211 @@ describe('gdjs.DraggableRuntimeBehavior', function () {
...
+      runtimeScene.renderAndStep(1000 / 60);
+      runtimeGame.getInputManager().onFrameEnded();
+      runtimeGame.getInputManager().onTouchMove(2, 750, 700);
+      runtimeGame.getInputManager().onTouchMove(1, 100, 200);
+      runtimeScene.renderAndStep(1000 / 60);
+      runtimeGame.getInputManager().onFrameEnded();
+      runtimeGame.getInputManager().onTouchEnd(1);
+      runtimeGame.getInputManager().onTouchEnd(2);
+      runtimeScene.renderAndStep(1000 / 60);
+      runtimeGame.getInputManager().onFrameEnded();
```
4ian, line 206: Is this extra frame required?

D8H: I moved it after the checks. Without it the following test handle this event too and move the object.

Extensions/DraggableBehavior/tests/draggableruntimebehavior.spec.js
D8H: This was a mistake, I removed it.

4ian:
Thank you! 👍

Extensions/DraggableBehavior/draggableruntimebehavior.ts
D8H: I used a `&&` instead of a `||` on the previous if, I'm sending a fix.

PR #3068, opened 2021-09-19
[Draggable] The object couldn't be dragged only vertically or horizontally.

D8H:
Sorry about that.
I will make test cases for it.


4ian:
Glad you caught it!

PR #3073, opened 2021-09-21
Feature/authors extension autocomplete

ClementPasteau:


newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/UsersAutocomplete.js
```diff
@@ -0,0 +1,149 @@
...
+            }
+
+            return null;
+          })
+        );
+        setLoading(false);
+      } catch (err) {
+        setLoading(false);
+        setAuthorsError('Error while loading the authors');
+        console.log(err);
```
4ian, line 95: Always add some context to errors being logged: `console.log("Could not load user public profiles:", error)`

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/UsersAutocomplete.js
```diff
@@ -0,0 +1,149 @@
...
+
+import useForceUpdate from '../../Utils/UseForceUpdate';
+
+type Props = {|
+  eventsFunctionsExtension: gdEventsFunctionsExtension,
+|};
+
+export const UsersAutocomplete = (props: Props) => {
+  const forceUpdate = useForceUpdate();
+  const [authors, setAuthors] = React.useState([]);
```
4ian, line 20: I recommend you type this state too?

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/UsersAutocomplete.js
```diff
@@ -0,0 +1,149 @@
...
+  const [authorInput, setAuthorInput] = useState('');
+  const [loading, setLoading] = useState(false);
+  const [
+    completionUserPublicProfiles,
+    setCompletionUserPublicProfiles,
+  ] = React.useState<Array<UserPublicProfile>>([]);
+  const [authorsError, setAuthorsError] = React.useState(null);
```
4ian, line 27: What I usually do is that I store the `Error` object (or `null` if no error). This allows the UI to render a message specific to the error, or a generic message.

(in this case, type as `React.useState<?Error>(null)`)

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/UsersAutocomplete.js
```diff
@@ -0,0 +1,149 @@
...
+        setLoading(true);
+        const userPublicProfiles = await searchUserPublicProfilesByUsername(
+          authorInput
+        );
+        setCompletionUserPublicProfiles(userPublicProfiles);
+        setLoading(false);
+      } catch (err) {
+        setLoading(false);
+        setAuthorsError('Error while loading users');
+        console.log(err);
```
4ian, line 47: Always add some context to errors being logged: `console.log("Could not load user public profiles:", error)`

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/UsersAutocomplete.js
```diff
@@ -0,0 +1,149 @@
...
+      try {
+        setLoading(true);
+        const userPublicProfiles = await searchUserPublicProfilesByUsername(
+          authorInput
+        );
+        setCompletionUserPublicProfiles(userPublicProfiles);
+        setLoading(false);
+      } catch (err) {
+        setLoading(false);
+        setAuthorsError('Error while loading users');
```
4ian, line 46: Here and below, I would rather store the Error object, and let the UI inspect it or display a generic error.

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/UsersAutocomplete.js
```diff
@@ -0,0 +1,149 @@
...
+  );
+
+  return (
+    <SemiControlledMultiAutoComplete
+      hintText={t`Start typing a username`}
+      floatingLabelText={t`Authors`}
+      helperText={t`Select all the extensions contributors usernames. They will be displayed in the order selected.`}
+      value={authors}
+      onChange={(event, values) => {
+        if (values) {
```
4ian, line 116: Can values be undefined/null at all? 
If so I would favor a early return: `if (!values) return;` (this is a matter of style, but usually this decrease indentation and make it clear this is not the "happy path")

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/UsersAutocomplete.js
```diff
@@ -0,0 +1,149 @@
...
+              value: userPublicProfile.id,
+            };
+          }
+
+          return null;
+        }
+      )}
+      loading={loading}
+      fullWidth
+      error={authorsError}
```
4ian, line 146: Here you could call a helper function `getErrorMessage(error: ?Error)` (defined outside of the component, no need to keep it inside), that can look at if there is an error, and if yes return an error message.
This make it easy to improve it in the future to analyze the error (is this an axios response? => nice I can look at the status code and give a good error. is this a status code that I don't know or undefined => generic error message)

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/UsersAutocomplete.js
```diff
@@ -0,0 +1,149 @@
...
+import {
+  searchUserPublicProfilesByUsername,
+  type UserPublicProfile,
+  getUserPublicProfilesByIds,
+} from '../../Utils/GDevelopServices/User';
+
+import useForceUpdate from '../../Utils/UseForceUpdate';
+
+type Props = {|
+  eventsFunctionsExtension: gdEventsFunctionsExtension,
```
4ian, line 15: I think we should directly make this a generic component that takes not a eventsFunctionsExtension, but a gdVectorString :D 

`userIds: gdVectorString`

(note that I put `userIds` and not `users` to make it clear what this contains. It's the equivalent of the name of the field in the extension "authorIds").
You then just have to pass it as prop `<UsersAutocomplete userIds={eventsFunctionsExtension.getAuthorIds()} />` (which make it clear that we're editing the users that are the authors of the extension)

We can then change all reference to `author` to `user` in this component and it will work exactly the same.

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/UsersAutocomplete.js
```diff
@@ -0,0 +1,149 @@
...
+            const userPublicProfile: UserPublicProfile =
+              userPublicProfilesByIds[userId];
+            if (userPublicProfile.username && userPublicProfile.id) {
+              return {
+                text: userPublicProfile.username,
+                value: userPublicProfile.id,
+              };
+            }
+
+            return null;
```
4ian, line 88: Question: should we display a "(no username)" in text (and the id in value)?

In theory this should never be shown to the user (because we'll have everyone to get a username, and you'll need one anyway to be selected).
In practice, I'm sure this will happen, so it's better to be clear in the UI rather than hiding the problem
(I prefer someone telling me "My username does not appear, help!" and find out that they got their username erased for some reason, than someone telling me "I disappeared from every extension, why did you remove me, I hate you!")

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/UsersAutocomplete.js
```diff
@@ -0,0 +1,149 @@
...
+      } catch (err) {
+        setLoading(false);
+        setAuthorsError('Error while loading users');
+        console.log(err);
+      }
+    },
+    [authorInput]
+  );
+
+  // call every time the authorInput changes
```
4ian, line 53: Nitpicking: Capital + period for each comment ;)

4ian (commented):
Lots of nits, but the overall approach seems great!

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/UsersAutocomplete.js
ClementPasteau: in this case, the authorsError is already a string to be displayed as is, but yes I'll extract it so that we know we can modify it easily

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/UsersAutocomplete.js
ClementPasteau: Yes!

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/UsersAutocomplete.js
ClementPasteau: Yeah I was under the impression it would never happen at all, but let's be safe and still display it!

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/ExtensionOptionsEditor.js
```diff
@@ -217,16 +218,10 @@ export const ExtensionOptionsEditor = ({
...
-            }
-            value={eventsFunctionsExtension.getAuthor()}
-            onChange={(e, text) => {
-              eventsFunctionsExtension.setAuthor(text);
-              forceUpdate();
-            }}
-            fullWidth
+          <UsersAutocomplete
+            userIds={eventsFunctionsExtension.getAuthorIds()}
+            floatingLabelText={t`Authors`}
```
4ian, line 223: Excellent thing to pass this as props. Turns out that you can use `<Trans></Trans>`. 
We favor this over `t`  because:
- you don't need to import <I18n> to use <Trans> in the final component displaying the message. 
- this enables advanced stuff like: `<Trans>Hello {something}<Trans>` (that will be turned into a translation like `Hello <0></0>` by js-lingui).
- this enables to pass other components (we don't want this for labels, but for helper texts it could be useful).

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/UsersAutocomplete.js
```diff
@@ -0,0 +1,152 @@
...
+      try {
+        setLoading(true);
+        const userPublicProfiles = await searchUserPublicProfilesByUsername(
+          userInput
+        );
+        setCompletionUserPublicProfiles(userPublicProfiles);
+        setLoading(false);
+      } catch (err) {
+        setLoading(false);
+        setError(new Error('Error while loading users'));
```
4ian, line 58: Can you actually store `err`?

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/UsersAutocomplete.js
```diff
@@ -0,0 +1,152 @@
...
+            return {
+              text: userPublicProfile.username || '(no username)',
+              value: userPublicProfile.id,
+            };
+          })
+        );
+        setLoading(false);
+      } catch (err) {
+        setLoading(false);
+        setError(new Error('Error while loading the users'));
```
4ian, line 100: Same, store `err`

4ian (approved):
See latest comments, otherwise LGTM!

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/UsersAutocomplete.js
4ian: (in an ideal world, errors would be strongly typed, so that we could have functions handling all kind of errors - at least be able to know what we're supposed to get (à la Java Exception) - unfortunately I think it's not really possible in Flow/TS? And this would need a strong integration between backends and frontends).

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/ExtensionOptionsEditor.js
```diff
@@ -217,16 +218,15 @@ export const ExtensionOptionsEditor = ({
...
           />
-          <TextField
-            floatingLabelText={
-              <Trans>Author (Name, email or GitHub handle)</Trans>
+          <UsersAutocomplete
+            userIds={eventsFunctionsExtension.getAuthorIds()}
+            floatingLabelText={<Trans>Authors</Trans>}
+            helperText={
+              <Trans>
+                Select all the extensions contributors usernames. They will be
```
AlexandreSi, line 226: Not sure about this but, is it more correct to say:`Select the usernames of the persons that contributed to this extension` ?

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/UsersAutocomplete.js
```diff
@@ -0,0 +1,153 @@
...
+  const searchUserPublicProfiles = React.useCallback(
+    async () => {
+      setError(null);
+      if (!userInput) {
+        setCompletionUserPublicProfiles([]);
+        return;
+      }
+      try {
+        setLoading(true);
+        const userPublicProfiles = await searchUserPublicProfilesByUsername(
```
AlexandreSi, line 50: We seem to have the lodash.debounce dependency, you could use it to reduce the api calls.
https://stackoverflow.com/a/62017005/9517183

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/UsersAutocomplete.js
```diff
@@ -0,0 +1,153 @@
...
+          userInput
+        );
+        setCompletionUserPublicProfiles(userPublicProfiles);
+        setLoading(false);
+      } catch (err) {
+        setLoading(false);
+        setError(err);
+        console.log('Could not load the users: ', err);
```
AlexandreSi, line 58: nitpicking: console.error

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/UsersAutocomplete.js
```diff
@@ -0,0 +1,153 @@
...
+              text: userPublicProfile.username || '(no username)',
+              value: userPublicProfile.id,
+            };
+          })
+        );
+        setLoading(false);
+      } catch (err) {
+        setLoading(false);
+        setError(err);
+        console.log('Could not load the users: ', err);
```
AlexandreSi, line 100: console.error as well

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/UsersAutocomplete.js
ClementPasteau: good catch, I meant to use it and forgot!

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/ExtensionOptionsEditor.js
ClementPasteau: yup, you're right :+1:

PR #3076, opened 2021-09-22
Zoom on events sheet

### LINKED ISSUE #3041: Zoom in the eventsheet
A quick win for a good first issue can be to add the zoom in/out in the event sheet.

## Description
Actually, there is no way to increase the size of the font in the event sheet unless we're in a browser but the whole page is zoomed.

## Solution suggested

In `<Eventsheet>` component we can add a font style size in **em** unit to increase or decrease the size of the whole event.
With the  `<Eventsheet>` component we have to put on a listener for `Ctrl+Mouse wheel Up / Down`, linked to the font size on the component.

- [ ] Control would be `Ctrl+Mouse wheel Up / Down`
- [ ] This zoom value should be saved in the local preference.
- [ ] Event sheet must be updated in the meanwhile.
- [ ] There is some CSS to fix for the comments and groups.
- [ ] Must prevent the default behavior for the browser.

Here the idea:

https://user-images.githubusercontent.com/1670670/133116656-72303956-9ce7-4159-a85b-39cbd8d47c38.mp4

AlexandreSi:
Resolves #3041 

@Bouh, thank you for the features suggestion, it helped a lot!

This is what I implemented:
 - ~Control would be Ctrl+Mouse wheel Up / Down~
   - Instead I used the already implemented event listener on Cmd + {=|-}
 - ✅ This zoom value should be saved in the local preference.
 - ✅ Event sheet must be updated in the meanwhile.
 - ✅ There is some CSS to fix for the comments and groups.
   - I also changed link and JS (for this one, the code cannot be updated on each zoom in or out, but if you refresh, the font is changed)
 - ✅ Must prevent the default behavior for the browser.
 - ❇️ I added Zoom In and Out in the context menu when right clicking on the events tree


Bouh:
Wow, that is so cool, it works well.
Mouse zoom with the mouse wheel is the main shortcut to have, SIlver-Streak was suggesting ctrl+ / - for accessibility as a second shortcut.
A mouse wheel implementation is living in the runtime I think `event.deltaY` is used for get the delta of the wheel.

AlexandreSi:
There is an implementation in the runtime but it would take time to implement something similar in the UI component in my opinion.
I suggest we add the feature as is, and if more users ask for the zoom with the mouse wheel, we add it. What do you think?

4ian:
> I suggest we add the feature as is, and if more users ask for the zoom with the mouse wheel, we add it. What do you think?

+1 on this, I think it's more than enough to have the shortcuts + menu. I personally don't recall using Cmd/Ctrl+mousewheel to zoom in my browser, and it's hard to judge how fast the zoom should be. Let's start simple and we can add this later if asked again :)

newIDE/app/src/EventsSheet/index.js
```diff
@@ -790,6 +798,19 @@ export default class EventsSheet extends React.Component<Props, State> {
     );
   };
 
+  onZoomEvent = (towards: string): ((domEvent?: KeyboardEvent) => void) => {
```
4ian, line 801: `nit: towards: 'IN' | 'OUT'`

newIDE/app/src/EventsSheet/index.js
```diff
@@ -790,6 +798,19 @@ export default class EventsSheet extends React.Component<Props, State> {
...
+    const factor = towards === 'IN' ? 1 : -1;
+    return (domEvent?: KeyboardEvent) => {
+      if (domEvent) {
```
4ian, line 804: Maybe explain why we're stopping the propagation/prevent default behavior

newIDE/app/src/EventsSheet/index.js
```diff
@@ -1300,6 +1322,18 @@ export default class EventsSheet extends React.Component<Props, State> {
                               ),
                               click: this._openEventsContextAnalyzer,
                             },
+                            { type: 'separator' },
+                            {
+                              label: i18n._(t`Zoom In`),
+                              click: () => this.onZoomEvent('IN')(),
+                              accelerator: 'CmdOrCtrl+=',
+                            },
```
4ian, line 1330: Do you think we should have a max zoom level (open question)?

newIDE/app/src/UI/TextField.js
```diff
@@ -82,7 +82,7 @@ type Props = {|
   margin?: 'none' | 'dense',
   fullWidth?: boolean,
   style?: {|
-    fontSize?: 14 | 18,
+    fontSize?: 14 | 18 | string,
```
4ian, line 85: Should this be a number?

newIDE/app/src/MainFrame/Preferences/PreferencesContext.js
```diff
@@ -264,6 +266,7 @@ export const initialPreferences = {
     useUndefinedVariablesInAutocompletion: false,
     useGDJSDevelopmentWatcher: true,
     eventsSheetUseAssignmentOperators: false,
+    eventsSheetZoomLevel: 14,
```
4ian, line 269: Open question: do you think we should name this `eventsSheetFontSize`?

4ian (approved):
Great stuff! ✨ Added a few nits/questions but implementation seems pretty clean!

newIDE/app/src/UI/Theme/Global/EventsSheet.css
```diff
@@ -7,7 +7,6 @@
     /* Use the fonts provided by the operating system(s) as possible. */
     /* If you update this font list, be sure to do it in all the other places using fonts in the codebase. */
     font-family: var(--gdevelop-font-family) !important;
-    font-size: 14px !important;
     background: var(--event-sheet-event-tree-background-color);
 }
```
Bouh, line 12: Thumbnail images of objects on conditions and actions are static, like the font size, increase the width/height of the img, maybe by a scale factor?

Bouh (approved):

newIDE/app/src/EventsSheet/index.js
AlexandreSi: 👍  forgot flow types reflexes !

newIDE/app/src/EventsSheet/index.js
AlexandreSi: I put a min level so that the user is not list if everything disappeared.
I can't think of a case where a max could be useful but humans find a way. I'll put one

newIDE/app/src/UI/TextField.js
AlexandreSi: Well, it looks a bit awkward like this, not really happy with that.
It seems that TextFields can have only 2 values for their fontsize, which seems ok to me, for consistence's sake.
But in this case, I wanted the TextField to have an adaptable fontsize with `0.8em` for instance. Hence this line

newIDE/app/src/MainFrame/Preferences/PreferencesContext.js
AlexandreSi: I wondered a bit. We could use it to scale icons' sizes as well (not implemented in this PR) so the naming seems ok with that in mind.

AlexandreSi: Just saw that @Bouh add a comment for this purpose. I will add it then

newIDE/app/src/UI/TextField.js
4ian: Ok! You could almost list the valid values? The idea was indeed to enforce consistency by having this kind of discussion "You're adding yet-another fontSize? You sure?" :) Maybe it's not too bad to list the values we're using, which is some kind of "living documentation" of the consistency across the app.

newIDE/app/src/EventsSheet/index.js
```diff
@@ -169,6 +169,7 @@ export default class EventsSheet extends React.Component<Props, State> {
   _eventSearcher: ?EventsSearcher;
   _searchPanel: ?SearchPanel;
   _containerDiv = React.createRef<HTMLDivElement>();
+  _zoomLevel = {min: 1, max: 50};
```
4ian, line 172: uber nit: if these are just constant values I would rather store them outside of the component (to keep the interface of the component as lean as possible and show that they won't ever be changed according to the component or something like this)

AlexandreSi: Good idea

newIDE/app/src/EventsSheet/EventsTree/Renderers/CommentEvent.js
```diff
@@ -29,7 +28,6 @@ const styles = {
     flexWrap: 'wrap',
     padding: 5,
     overflow: 'hidden',
-    minHeight: 35,
```
4ian, line 32: Just thought about it, might be worth checking if a comment without any text is shown properly or not.

AlexandreSi: Nope, you're right, i'll look into it

newIDE/app/src/UI/TextField.js
AlexandreSi: Oh yeah, great idea!

PR #3077, opened 2021-09-22
Remove any reference to friction in particle emitter

### LINKED ISSUE #2313: Particle friction is not working.

AlexandreSi:
Resolves #2313
Forgot to treat this issue in #3058 
The removal of the related action/expression/condition could create bugs in game if they used any of these but i couldn't find them before removing them. They must have been incomplete and removed by `StripUnimplementedInstructionsAndExpressions`


4ian (approved):
> The removal of the related action/expression/condition could create bugs in game if they used any of these but i couldn't find them before removing them. They must have been incomplete and removed by StripUnimplementedInstructionsAndExpressions

Yep, so sounds good to remove them. Thanks for cleaning that up!

PR #3086, opened 2021-09-23
Fix: allow color properties to be used in events.

D8H:
The `Color` property type was not considered as a string and the properties could be used in string expressions.

Steps to reproduce:
* Open the fog of war example
* Run the game, the slilders are blue
* Set the thumb color property of the Slider extension to the color type
* Run the game, the sliders are white instead of blue

Related PR: https://github.com/4ian/GDevelop/pull/2980

### DIFF
--- a/newIDE/app/src/EventsFunctionsExtensionsLoader/MetadataDeclarationHelpers.js
+++ b/newIDE/app/src/EventsFunctionsExtensionsLoader/MetadataDeclarationHelpers.js
@@ -388,11 +388,7 @@ export const declareBehaviorPropertiesInstructionAndExpressions = (
     const propertyLabel =
       property.getLabel() || i18n._(t`${propertyName} property`);
 
-    if (
-      propertyType === 'String' ||
-      propertyType === 'Choice' ||
-      propertyType === 'Color'
-    ) {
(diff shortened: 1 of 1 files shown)


4ian:
This is actually already handled: https://github.com/4ian/GDevelop/blob/d4a3722ec8eafe9117f6509acc9225e3cc15438a/newIDE/app/src/EventsFunctionsExtensionsLoader/MetadataDeclarationHelpers.js#L507

But the action/condition are not following the convention of being a string (the action is to always set a value, there is no operator).

I think the real problem here might be that changing a property type should trigger warnings that this may invalidate actions/conditions using it?

D8H:
> This is actually already handled:
> 
> https://github.com/4ian/GDevelop/blob/d4a3722ec8eafe9117f6509acc9225e3cc15438a/newIDE/app/src/EventsFunctionsExtensionsLoader/MetadataDeclarationHelpers.js#L507
> 
> But the action/condition are not following the convention of being a string (the action is to always set a value, there is no operator).
> 
> I think the real problem here might be that changing a property type should trigger warnings that this may invalidate actions/conditions using it?

I see, so there is an action and condition defined but not an expression.

4ian (approved):
Sounds like the right thing to have indeed!

PR #3087, opened 2021-09-23
Add condition to test if platformer is on a given platform

### LINKED ISSUE #3029: [Platformer] Add a condition to pick the current floor
## Description

It's complicated to detect on which platform the character is.
https://github.com/4ian/GDevelop/issues/1920

## Solution suggested

I think a condition like: "_PARAM0_ is on _PARAM2_" could help building platformer events extensions. For instance, a behavior that makes the character follows a moving platform when he's jumping from it.

AlexandreSi:
Resolves #3029

### DIFF
--- a/Extensions/PlatformBehavior/Extension.cpp
+++ b/Extensions/PlatformBehavior/Extension.cpp
@@ -723,5 +723,5 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
            .AddParameter("behavior", _("Behavior"), "PlatformerObjectBehavior")
            .AddParameter("objectList", _("Platforms"), "", false)
            .AddCodeOnlyParameter("conditionInverted", "")
-           .SetFunctionName("gdjs.platform.isOnPlatform");
+           .SetFunctionName("gdjs.evtTools.platform.isOnPlatform");
 }
--- a/Extensions/PlatformBehavior/Extension.cpp
+++ b/Extensions/PlatformBehavior/Extension.cpp
@@ -711,4 +711,17 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
         .MarkAsAdvanced()
         .SetFunctionName("ChangePlatformType");
   }
+
+  extension.AddCondition("IsObjectOnGivenFloor",
+                         _("Is object on given floor"),
+                         _("Check if an object is on a given floor"),
+                         _("_PARAM0_ is on floor _PARAM2_"),
+                         _("Platforms"),
+                         "CppPlatform/Extensions/platformicon24.png",
+                         "CppPlatform/Extensions/platformicon16.png")
+           .AddParameter("objectList", _("Object"), "", false)
+           .AddParameter("behavior", _("Behavior"), "PlatformerObjectBehavior")
+           .AddParameter("objectList", _("Platforms"), "", false)
+           .AddCodeOnlyParameter("conditionInverted", "")
+           .SetFunctionName("gdjs.platform.isOnPlatform");
 }
--- a/Extensions/PlatformBehavior/Extension.cpp
+++ b/Extensions/PlatformBehavior/Extension.cpp
@@ -41,7 +41,6 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
         std::make_shared<PlatformerObjectBehavior>(),
         std::make_shared<gd::BehaviorsSharedData>());
 
-#if defined(GD_IDE_ONLY)
     aut.AddCondition("IsMoving",
                      _("Is moving"),
                      _("Check if the object is moving (whether it is on the "
@@ -53,8 +52,7 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
         .AddParameter("object", _("Object"))
         .AddParameter("behavior", _("Behavior"), "PlatformerObjectBehavior")
         .MarkAsSimple()
-        .SetFunctionName("IsMoving")
-        .SetIncludeFile("PlatformBehavior/PlatformerObjectRuntimeBehavior.h");
+        .SetFunctionName("IsMoving");
 
     aut.AddCondition("IsOnFloor",
                      _("Is on floor"),
@@ -66,8 +64,7 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
         .AddParameter("object", _("Object"))
         .AddParameter("behavior", _("Behavior"), "PlatformerObjectBehavior")
         .MarkAsSimple()
-        .SetFunctionName("IsOnFloor")
-        .SetIncludeFile("PlatformBehavior/PlatformerObjectRuntimeBehavior.h");
+        .SetFunctionName("IsOnFloor");
 
     aut.AddCondition("IsOnLadder",
                      _("Is on ladder"),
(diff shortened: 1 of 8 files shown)


Extensions/PlatformBehavior/platformtools.ts
```diff
@@ -0,0 +1,18 @@
+namespace gdjs {
+  export namespace platform {
+    export const isOnPlatform = function (
+      objectsLists1,
+      behavior,
+      objectsLists2,
+      inverted
```
4ian, line 7: Let's add type for these. 
objectLists are `Hashtable<Array<gdjs.RuntimeObject>>` 
behavior is a string, can you rename it to behaviorName?

Extensions/PlatformBehavior/platformruntimebehavior.ts
```diff
@@ -229,6 +229,19 @@ namespace gdjs {
     static NORMALPLAFTORM = 0;
     static JUMPTHRU = 1;
     static LADDER = 2;
+
+    static isOnPlatformTest(object1, object2, behaviorName) {
```
4ian, line 233: Let's have types there too (gdjs.RuntimeObject, gdjs.RuntimeObject, string)

Extensions/PlatformBehavior/platformruntimebehavior.ts
```diff
@@ -229,6 +229,19 @@ namespace gdjs {
...
+      // Check if the objects exist and share the behavior
+      if (object1 === null || !object1.hasBehavior(behaviorName)) {
```
4ian, line 235: I think object1 is guaranteed to be not null, same as object2.

4ian: Actually the behavior is also guaranteed by the action calling it, so no need to check.

Extensions/PlatformBehavior/platformruntimebehavior.ts
```diff
@@ -229,6 +229,19 @@ namespace gdjs {
...
+        return false;
+      }
+
+      const behavior1 = object1.getBehavior(behaviorName);
+      return (
+        behavior1.isOnFloor() &&
+        behavior1._onFloor.getFloorPlatform().owner.id === object2.id
```
4ian, line 242: Could we expose this as a method `isOnFloorObject`? Better to avoid PlatformRuntimeBehavior accessing internals of PlatformerObjectRuntimeBehavior.

Extensions/PlatformBehavior/platformruntimebehavior.ts
```diff
@@ -229,6 +229,19 @@ namespace gdjs {
...
     static JUMPTHRU = 1;
     static LADDER = 2;
+
+    static isOnPlatformTest(object1, object2, behaviorName) {
+      // Check if the objects exist and share the behavior
+      if (object1 === null || !object1.hasBehavior(behaviorName)) {
+        return false;
+      }
+
+      const behavior1 = object1.getBehavior(behaviorName);
```
4ian, line 239: You can probably cast this (`as gdjs.PlatformerObjectRuntimeBehavior`)

AlexandreSi:
I don't have typing reflexes still, I'm on it!

Extensions/PlatformBehavior/Extension.cpp
```diff
@@ -758,8 +709,19 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
...
+  extension.AddCondition("IsObjectOnGivenFloor",
+                         _("Is object on given floor"),
+                         _("Test if an object is on a given floor."),
+                         _("_PARAM0_ is on floor _PARAM2_"),
+                         _("Platforms"),
+                         "CppPlatform/Extensions/platformicon24.png",
+                         "CppPlatform/Extensions/platformicon16.png")
+           .AddParameter("objectList", _("Object"), "", false)
+           .AddParameter("behavior", _("Behavior"), "PlatformerObjectBehavior")
+           .AddParameter("objectList", _("Platforms"), "", false)
```
AlexandreSi, line 724: I have a doubt here @4ian, I don't know if Platforms can be misleading, because this condition actually test if the platformers is on any kind of object.

4ian: I think it's fair to call them platforms, you can in theory pass any kind of objects, but in this case the condition will simply be false and it will filter out objects that are not the floor, and so that don't even have the platform behavior.

For a user, it makes sense when using this condition to first choose an object, with the platformer behavior, and then to select the "Platform" on which to test as a floor.

Extensions/PlatformBehavior/platformtools.ts
```diff
@@ -0,0 +1,18 @@
+namespace gdjs {
+  export namespace platform {
```
4ian, line 2: Let's put this inside the `evtTools` namespace (gdjs.evtTools.platform).

The `gdjs` namespace is fairly crowded and so for stuff that are not purely a JS api, but a "glue function" used only by events, it's better to use the `evtTools` namespace (it won't "pollute" the generated GDJS documentation).

4ian (approved):

Extensions/PlatformBehavior/Extension.cpp
D8H: Why not adding a behavior parameter to constraint the "Platforms" objectList?

4ian: That's a good question. It turns out that we don't really need it. We could add it to ensure someone select only the objects that can be platform, but it's not even necessary, so we left it without... do you think it would be safer to have it?
Or just to get a validation on the user side?

tristanbob: If I am understanding this correctly, we have to decide between these user experiences:

1) User selects an object that doesn't have platform behavior, and wonders why the condition didn't work.

or

2) User tries to select an object that doesn't have platform behavior, and wonders why it is not showing up in the autocompletion in the condition.

4ian: I think you're correct, it's basically:
1. no check (no behavior parameter), but condition will never be true (because your object is not even a platform, so no chance at all it is a floor)
2. a behavior parameter, so that it will be flagged in red if you select an object that does not have the Platform behavior.

The 2 might actually be better as giving a hint to the user.

tristanbob: I agree that #2 is the better user experience.

4ian: @AlexandreSi If that's ok I think we should make another PR to add this parameter (despite what I initially thought, it's probably better).

Note that in the future, the editor could be smarter and limit the selection to objects having the behavior (right now you can select any object and then BAM the editor is screaming at you with an error telling you that this object does not have the proper behavior).

Extensions/PlatformBehavior/platformruntimebehavior.ts
```diff
@@ -229,6 +229,17 @@ namespace gdjs {
...
+    static isOnPlatformTest(
+      object1: gdjs.RuntimeObject,
+      object2: gdjs.RuntimeObject,
+      behaviorName: string
+    ): boolean {
+      const behavior1 = object1.getBehavior(
+        behaviorName
+      ) as PlatformerObjectRuntimeBehavior;
+      return behavior1.isOnFloorObject(object2);
+    }
```
D8H, line 242: This method expose a method from PlatformerObjectRuntimeBehavior in a static context. I think it makes more sense to have it in PlatformerObjectRuntimeBehavior.
Also, I personally avoid to use "test" in a method name because it has a "unit test case" connotation and the "Is" prefix already stand for a condition.

PR #3092, opened 2021-09-24
Fix Folder being set for both Examples and Starters

### LINKED ISSUE #3089: Default folder when opening an Example
## Description

When you try to open an "Example", the Default Folder is not set. Which means that the "Open" button doesn't do anything.
However, it is set when you go on the "Starters" tab (and stays if you go back to the "Example" tab)
It can be confusing if you don't understand you need to set a folder (for newcomers)

image
image

## Solution suggested

- Set a default folder for the "Examples" tab, the same way it is set for the "Starters" tab
- Add a small warning (as a toaster) "Select a folder first!" when you try to click "Open" without a folder

## Alternatives considered

We could also trigger the Folder selection when trying to Open an example/starter (and then trigger the opening of the Example once the folder is selected) -> more work

ClementPasteau:
Fixes #3089 

Moved the logic to the `CreateProjectDialog` level rather than the `LocalStarters` only

image


4ian:
Ok I now understand that the problem was the dialog could be opened from the games showcase too. Sounds good then!

4ian (approved):

PR #3101, opened 2021-09-27
Add expression for sprite flipping

AlexandreSi:


AlexandreSi:
@4ian, maybe i'll just merge the removal of the `#if defined(...)` and we'll see later to add those expressions if we decide that booleans can be used!

4ian:
Yep sounds good! Let's only clean the #ifdef for now and see the case for boolean expressions later.

4ian (approved):
Thanks! 
Squash/merge with a commit message containing "Don't show in changelog" to avoid this being pulled in the changelog (because not impacting the user) (if it was a more important refactoring that is still useful for developers/contributors to notice, we would write "Only show in developer changelog" and it would be put in a special category in the changelog).

PR #3102, opened 2021-09-27
Add possibility to start and stop particle emitter

AlexandreSi:
To bring a solution to this
https://forum.gdevelop-app.com/t/stop-particles/24905

### DIFF
--- a/Extensions/ParticleSystem/ExtensionSubDeclaration1.cpp
+++ b/Extensions/ParticleSystem/ExtensionSubDeclaration1.cpp
@@ -267,4 +267,22 @@ void ExtensionSubDeclaration1(gd::ObjectMetadata& obj) {
                    "CppPlatform/Extensions/particleSystemicon16.png")
       .AddParameter("object", _("Object"), "ParticleEmitter")
       .UseStandardRelationalOperatorParameters("number");
+
+  obj.AddAction("StartEmission",
+                _("Start emission"),
+                _("Refull tank (if not infinite) and start emission of the particles."),
+                _("Start emission of _PARAM0_"),
+                _("Common"),
+                "CppPlatform/Extensions/particleSystemicon24.png",
+                "CppPlatform/Extensions/particleSystemicon16.png")
+      .AddParameter("object", _("Object"), "ParticleEmitter");
+
+  obj.AddAction("StopEmission",
+                _("Stop emission"),
+                _("Stop the emission of particles."),
+                _("Stop emission of _PARAM0_"),
+                _("Common"),
+                "CppPlatform/Extensions/particleSystemicon24.png",
+                "CppPlatform/Extensions/particleSystemicon16.png")
+      .AddParameter("object", _("Object"), "ParticleEmitter");
 }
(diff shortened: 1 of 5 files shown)


Bouh:
Think to start a new branch from master, there are expressions IsFlippedX and IsFlippedY on this draft PR.

AlexandreSi:
That's very true, I'll fix that, thanks!

Extensions/ParticleSystem/ExtensionSubDeclaration1.cpp
```diff
@@ -267,4 +267,22 @@ void ExtensionSubDeclaration1(gd::ObjectMetadata& obj) {
                    "CppPlatform/Extensions/particleSystemicon16.png")
       .AddParameter("object", _("Object"), "ParticleEmitter")
       .UseStandardRelationalOperatorParameters("number");
+
+  obj.AddAction("StartEmission",
+                _("Start emission"),
+                _("Refull tank (if not infinite) and start emission of the particles."),
```
4ian, line 273:
```suggestion
                _("Refill tank (if not infinite) and start emission of the particles."),
```

4ian (approved):
🚀

Extensions/ParticleSystem/ExtensionSubDeclaration1.cpp
AlexandreSi: Oups, I forgot that I could accept this suggestion directly. Thanks!

AlexandreSi:
@4ian I added 2 commits:
- One that should bring clarity (why restart emitter when changing capacity?) but may need fixes for games that expect the emitter to start when changing its capacity.
- One for code tidiness
Let me know!

AlexandreSi:
But since the Change Capacity action was not working before my previous PR, the first bullet point is ok

4ian:
> One that should bring clarity (why restart emitter when changing capacity?) 

Isn't this actually a behavior change introduced by https://github.com/4ian/GDevelop/commit/2febbf439f24eb1f5153264c926a044e07109e7c? 
In which case your change here is actually "fixing" this behavior change (not resetting the emission when changing the capacity/flow, which is better).

> One for code tidiness

Maybe we should take the habit of adding the new name next to Deprecated (`SetHidden() // Deprecated - use "SetXxx" instead.`). Otherwise LGTM!

4ian:
> One that should bring clarity (why restart emitter when changing capacity?) 

Isn't this actually a behavior change introduced by https://github.com/4ian/GDevelop/commit/2febbf439f24eb1f5153264c926a044e07109e7c? 
In which case your change here is actually "fixing" this behavior change (not resetting the emission when changing the capacity/flow, which is better).

> One for code tidiness

Maybe we should take the habit of adding the new name next to Deprecated (`SetHidden() // Deprecated - use "SetXxx" instead.`). Otherwise LGTM!

4ian:
> One that should bring clarity (why restart emitter when changing capacity?) 

Isn't this actually a behavior change introduced by https://github.com/4ian/GDevelop/commit/2febbf439f24eb1f5153264c926a044e07109e7c? 
In which case your change here is actually "fixing" this behavior change (not resetting the emission when changing the capacity/flow, which is better).

> One for code tidiness

Maybe we should take the habit of adding the new name next to Deprecated (`SetHidden() // Deprecated - use "SetXxx" instead.`). Otherwise LGTM!

4ian (approved):

Issue #3003, opened 2021-09-03
Value error highlight exceeds text field [Visual Bug]


Elairyx7301:
## Describe the bug
When having a long error that goes over multiple lines, the error highlight exceeds the input field.

## To Reproduce
Steps to reproduce the behavior. Note that this is just one example and only the one shown in the image below. You can reproduce this with basically any input field.

1. Change scene variable value
2. Enter `sedutperspiciatisundeomnisistenatuserrorsitvoluptatemaccusantiumsddolorem` as value
(or use anything else that is long enough to exceed one line and is in the wrong format/data type)
3. Look closely at the end of the first line

![image](https://user-images.githubusercontent.com/83909121/132052579-d4500c55-9df2-436b-a0a4-2158896e6ace.png)

The error does not necessarily have to be one full line, it only needs to be 1) connected (no space character between important characters) and 2) be in multiple lines.

Note that some line/character highlights seem to be inconsistent. Here, the last few characters are not marked, even though the "value" is just one "word". Unfortunately, I haven't been able to find a pattern to when it's inconsistent or not.

## Other details
\* No project example needed.
GDevelop 5.0.0-beta115 based on GDevelop.js 4.0.99-0-release
PC / Desktop GDevelop version, Windows
Regular build, no nightly version

Bouh:
Might be a Material-UI problem?

4ian:
I don't think so, the highlight is home made and probably the styling is not properly "aligned" with the text field :)

Issue #2656, opened 2021-05-17
Object reordering goes wrong


PascalLadalle:
In the objects list, if you drag an object somewhere higher than its current position, it works fine.
If you drag it below its current position, it goes one spot too far.
If you have A, B, C, D and try to move A between C and D, you get B, C, D, A.

Reported [on the forum](https://forum.gdevelop-app.com/t/bugs-gdevelop-interface/32670)
Confirmed on beta110

Oxey405:
Hello ! Could we have screenshots ?

PascalLadalle:
@Oxey405 Sure.
From left to right, initial position, expected result, actual result.
![image](https://user-images.githubusercontent.com/25978830/125301603-6d31a100-e32b-11eb-956f-af48db375a5a.png)

PR #3104, opened 2021-09-29
UI improvements

### LINKED ISSUE #3085: Project name overflowing in project manager
## Describe the bug
The project name at the top of the project manager overflows when the project name is long

## To Reproduce
1. Go to project manager
2. Rename the project's name to something long
3. Reopen the project manager (This might also be a bug, the project manager needs to be reopern for the project name to update in the UI)
4. See the overflow of the text

Project=name overflow-syndrome

## Other details
* Linux mint 20.2
* beta177, desktop

One fix might be to add `...` if the project name is too long
Like done here:
sol

Also, maybe limit the project name to maybe 100 characters
:)

AlexandreSi:
Resolves #3085 #2656 #3003


newIDE/app/src/UI/DrawerTopBar.js
```diff
@@ -43,9 +50,11 @@ export default (props: Props) => {
             <Close />
           </IconButton>
         )}
-        <Typography variant="h6" style={styles.title}>
-          {props.title}
-        </Typography>
+        <Tooltip title={props.title} placement="bottom" enterDelay={500}>
```
4ian, line 53: Let's use 300ms as the delay, as done for the optional tooltip on icon buttons: 
 https://github.com/4ian/GDevelop/blob/2d899d7c52590da0f4ffab392d2f4b9a82e957d7/newIDE/app/src/UI/IconButton.js#L62-L71
 
 If you want to go the extra mile, this constant `tooltipEnterDelay` can be exported in a `UI/Tooltip.js` file to ensure consistency should we tweak this in the future. 

(I'm hesitant to say to abstract the entire material-ui `Tooltip` component in a separate file, because its usage is very much dependent on the context, dependent on material-ui, and because `newIDE/app/src/UI/DrawerTopBar.js` is already in the UI folder, so it has the "right" to use material-ui's Tooltip directly).

newIDE/app/src/UI/DrawerTopBar.js
```diff
@@ -5,6 +5,7 @@ import Toolbar from '@material-ui/core/Toolbar';
 import IconButton from '@material-ui/core/IconButton';
 import Typography from '@material-ui/core/Typography';
 import Close from '@material-ui/icons/Close';
+import { Tooltip } from '@material-ui/core';
```
4ian, line 8: I tend to prefer `import Tooltip from '@material-ui/core/Tooltip'` as it helps tree shaking which is not done automatically with this version. See https://v4.mui.com/guides/minimizing-bundle-size/

This being said, we're using so much of `material-ui/core` that I'm not sure it's worth it (it's definitely worth for `@material-ui/icons`), and should probably be linted if we want to enforce ti.

4ian (approved):
See two nitpickings, otherwise LGTM! :)

newIDE/app/src/UI/DrawerTopBar.js
AlexandreSi: Ok, thanks for this criterium for being in the UI folder or not. For now, I will just define a constant for the delay.

AlexandreSi:
@4ian I added a commit to force update instance list when modifying instances properties (I had troubles reorganizing z orders on my project with a lot of instances).
The last commit is meant to be coherent: unsaved changes were passed down when children in row mode, I don't think they shouldn't be passed if children are in column mode.

newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/index.js
```diff
@@ -70,6 +70,7 @@ const styles = {
     marginTop: 22, //Properly align with the text field
     paddingLeft: 12,
     paddingRight: 12,
+    boxSizing: 'border-box',
```
4ian, line 73: Good catch!!

newIDE/app/src/UI/Tooltip.js
```diff
@@ -0,0 +1 @@
+export const tooltipEnterDelay = 300;
```
4ian, line 1: nit: // @flow ;)

newIDE/app/src/InstancesEditor/InstancePropertiesEditor/index.js
```diff
@@ -25,6 +25,7 @@ type Props = {|
   layout: gdLayout,
   instances: Array<gdInitialInstance>,
   onEditObjectByName: string => void,
+  onInstancesModified?: Instances => void,
```
4ian, line 28: Actually, this is a `Array<gdInitialInstance>` that is passed to onInstancesModified - this is more precise than a `Instances`.

`Instances` is a generic type for PropertiesEditor - as PropertiesEditor can be used for anything else that would have "properties".
In theory, we would have typed PropertiesEditor as a generic component in Flow that would take a type `Instances` as parameter (a `PropertiesEditor<Instances>` if you follow me). 
And Flow would have deduced that this `Instances` type is in this case a `Array<gdInitialInstance>`.

In practice it was too hard, so we just typed this as a "Object" type (i.e: a bit like "any"):
image

Let's avoid leaking this to the parent, and use Array<gdInitialInstance>, much like the `instances` prop that is passed down.

4ian (approved):

newIDE/app/src/InstancesEditor/InstancePropertiesEditor/index.js
AlexandreSi: Ok, thanks for the precision!

4ian:
Great stuff!

PR #3105, opened 2021-09-29
Do not open expression parameters dialog if no parameters

### LINKED ISSUE #864: "There is nothing to configure" popup
## Describe the bug
When you add an expression that has no options/settings/parameters, you still get a popup with a message saying "There is nothing to configure" which you have to close. I'd rather this popup not display if there's nothing to configure...

## To Reproduce
1. Add a new action
2. Search for/click on 'Value of a scene variable'
3. Click the sigma button next to the 'Value' field
4. Search for/click on 'Identifier of the last touch'
5. Note pointless popup

nothing

AlexandreSi:
Resolves #864

### DIFF
--- a/newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/index.js
+++ b/newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/index.js
@@ -222,12 +222,24 @@ export default class ExpressionField extends React.Component<Props, State> {
     });
   };
 
+  _shouldOpenParametersDialog = (
+    expressionInfo: EnumeratedExpressionMetadata
+  ): boolean => {
+    return expressionInfo.parameters.length > 0;
+  };
+
   _handleExpressionChosen = (expressionInfo: EnumeratedExpressionMetadata) => {
-    this.setState({
-      popoverOpen: false,
-      parametersDialogOpen: true,
-      selectedExpressionInfo: expressionInfo,
-    });
+    let newState = { popoverOpen: false };
+    if (this._shouldOpenParametersDialog(expressionInfo)) {
+      newState = {
+        ...newState,
+        parametersDialogOpen: true,
+        selectedExpressionInfo: expressionInfo,
+      };
+    } else {
+      this.insertExpression(expressionInfo, []);
+    }
+    this.setState(newState);
   };
 
   insertExpression = (
--- a/newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/index.js
+++ b/newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/index.js
@@ -225,6 +225,8 @@ export default class ExpressionField extends React.Component<Props, State> {
   _shouldOpenParametersDialog = (
     expressionInfo: EnumeratedExpressionMetadata
   ): boolean => {
+    // If there is no parameter to fill for the selected expression, no need to open the dialog.
+    // Non visible parameters (like code only ones) should have already been filtered out.
     return expressionInfo.parameters.length > 0;
   };
 


AlexandreSi:
@4ian this is your input:

> Yeah, that's the "easy but dirty" way of doing it ;) (dirty because user will see the start of the dark overlay fading in and out - or maybe worse a beginning of the window to appear 😉). Proper solution is "simply" to not show the modal, but for this you have to analyze the expression to see that there is no parameters - knowing that there can be some "hidden" parameters that are asked by the expression to work properly (notably, the most common is a reference to the scene).
> 
> Nothing impossible again, but a bit of refactoring to do :)

It seems to me that my new function `_shouldOpenParametersDialog` should be a bit more elaborate than that, but the ParameterMetadata type does not have a notion of hidden. Are you talking about code only parameters ?

4ian:
> Are you talking about code only parameters ?

Correct! I think they are the only "hidden" parameters. 
Though another case is if you selected an object already, then the first parameter "Object" should not be shown. This is the case in the action/condition already: if you select an object, then an action, you won't have the object name parameter shown again, which is fairly natural:

image
 
 Note that if I was to search for this action and select it instead, the "Object" parameter would be shown: 
 
image

> You can think of the first (and most usual) approach as being "object oriented" (Object first, then the method, then parameters, much like when you write "`Object.MyMethod(Param1, Param2)`" in a programming language). The second approach is also supported to allow searching in everything (This is a bit like the old "C like" approach: `Class_MyMethod(Object, Param1, Param2)`)

This logic is done here:

https://github.com/4ian/GDevelop/blob/2d899d7c52590da0f4ffab392d2f4b9a82e957d7/newIDE/app/src/EventsSheet/InstructionEditor/InstructionParametersEditor.js#L138-L154

In case of expressions, the expression parameter dialog is always showing all parameters I think... so everything I said here is not really applicable to expressions :D 

So I think you're fine with a function filtering code only parameters. Not sure if we should extract these two functions into a pair of `getExpressionVisibleParametersCount` and `getInstructionVisibleParametersCount` (remember, instruction = action or condition) in a file.
Maybe they could live respectively in `EnumerateExpressions.js` and `EnumerateInstructions.js` to consolidate this logic with the other logic used to fetch expressions/instructions. Otherwise, keeping them in the only components using them is fine too I think.

AlexandreSi:
When developing on this feature, I was focusing on the component `ExpressionParametersEditor` and not on the `InstructionParametersEditor` that you mentionned (and that I'm having troubles to know where this component appears).

From what I see:
- We're working on the component `GenericExpressionField` that uses `ExpressionSelector`
- `ExpressionSelector` uses `enumerateAllExpressions` to list all possible expressions
- Those expressions are listed by `enumerateExpressionMetadataMap` that already filters out code only parameters

Do you think we should add another filter to be sure?

https://github.com/4ian/GDevelop/blob/5e3430aea54c3a791f67d71cc4c0c1a36c676273/newIDE/app/src/InstructionOrExpression/EnumerateExpressions.js#L36-L68

4ian:
tl;dr: you're right! No need to filter again.

>  I was focusing on the component ExpressionParametersEditor and not on the InstructionParametersEditor that you mentionned (and that I'm having troubles to know where this component appears).

To be clear, this is `InstructionParametersEditor`:

image

The fields are using GenericExpressionField. They are indeed showing an `ExpressionSelector` when you click on the right button:
image

Finally, when you choose an expression in this list, a new dialog open to allow you to enter parameters (rather than writing them by yourself). This is using `ExpressionParametersEditor`
image

> Those expressions are listed by enumerateExpressionMetadataMap that already filters out code only parameters

This is right.
I was mistaken because the ExpressionParametersEditor (and its corresponding ExpressionParametersEditorDialog) is taking a gdExpressionMetadata - i.e the whole metadata without any filtering done:

image

We can actually see that there is a `getNonCodeOnlyParametersCount` function already present in `ExpressionParametersEditor`:

https://github.com/4ian/GDevelop/blob/5e3430aea54c3a791f67d71cc4c0c1a36c676273/newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/ExpressionParametersEditor.js#L37-L44

But as you said, the EnumeratedExpressionMetadata that you actually handle has the filtering done already, so no need to do it again!

newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/index.js
```diff
@@ -222,12 +222,24 @@ export default class ExpressionField extends React.Component<Props, State> {
     });
   };
 
+  _shouldOpenParametersDialog = (
+    expressionInfo: EnumeratedExpressionMetadata
+  ): boolean => {
+    return expressionInfo.parameters.length > 0;
```
4ian, line 228: Let's add a comment telling that non visible parameters (like "code only" parameters) are already filtered out.

4ian (approved):

D8H:
Hello, Entropy found a regression that seems to be related to this change.

Steps to reproduce:
* Add a scene variable numeric affectation
* Click on the zeta
* Choose: Common > Angle > Angle
* The ExpressionParametersEditorDialog doesn't show up and the expression is like this: `undefinied.Angle()`

Please, see the Discord topic. Entropy did a video: https://discord.com/channels/258623956158906368/614526835807617044/895831416636583946

4ian:
Yep that was a regression, it should be fixed in release 120 : https://github.com/4ian/GDevelop/issues/3145 👍

PR #3110, opened 2021-09-30
Display contributions (examples & extensions) on the user profile

ClementPasteau:
- Moved edit button to be part of the page, rather than the actions
- Created new component for Contributions, fetching extensions & examples based on the userId (so that it can be re-used for a possible public profile in the future)
- Created function to fetch user public profile from userId, but ended up not using it because I can just use the existing profile. (but can be useful later on)

Tested with test data with and without contributions.
Tested username edit still updates correctly the profile.

image

image

### DIFF
--- a/newIDE/app/src/Profile/ContributionsDetails.js
+++ b/newIDE/app/src/Profile/ContributionsDetails.js
@@ -2,9 +2,7 @@
 import { Trans } from '@lingui/macro';
 
 import * as React from 'react';
-import { Accordion } from '../UI/Accordion';
-import { AccordionHeader } from '../UI/Accordion';
-import { AccordionBody } from '../UI/Accordion';
+import { AccordionHeader, AccordionBody, Accordion } from '../UI/Accordion';
 import { IconContainer } from '../UI/IconContainer';
 import { Column, Line } from '../UI/Grid';
 import {
@@ -17,6 +15,7 @@ import {
 } from '../Utils/GDevelopServices/Example';
 import PlaceholderLoader from '../UI/PlaceholderLoader';
 import Text from '../UI/Text';
+import BackgroundText from '../UI/BackgroundText';
 
 type ContributionLineProps = {|
   fullName: string,
@@ -43,11 +42,11 @@ export const ContributionLine = ({
 );
 
 type ExamplesAccordionProps = {|
-  examples: ?Array<ExampleShortHeader>,
+  examples: Array<ExampleShortHeader>,
   exampleError: ?Error,
 |};
 
-const ExamplesAccordion = ({
+export const ExamplesAccordion = ({
   examples,
   exampleError,
 }: ExamplesAccordionProps) => {
@@ -55,14 +54,14 @@ const ExamplesAccordion = ({
     return (
       <Column>
         <Line alignItems="center">
-          <Text size="title">
+          <Text>
             <Trans>Error retrieving the examples</Trans>
           </Text>
         </Line>
       </Column>
     );
 
-  return examples ? (
+  return (
     <Accordion>
       <AccordionHeader>
         <Text displayInlineAsSpan>
@@ -83,17 +82,15 @@ const ExamplesAccordion = ({
         </Column>
       </AccordionBody>
     </Accordion>
-  ) : (
-    <PlaceholderLoader />
   );
 };
 
 type ExtensionsAccordionProps = {|
-  extensions: ?Array<ExtensionShortHeader>,
+  extensions: Array<ExtensionShortHeader>,
   extensionError: ?Error,
 |};
 
-const ExtensionsAccordion = ({
+export const ExtensionsAccordion = ({
   extensions,
   extensionError,
 }: ExtensionsAccordionProps) => {
@@ -101,14 +98,14 @@ const ExtensionsAccordion = ({
     return (
       <Column>
         <Line alignItems="center">
-          <Text size="title">
+          <Text>
             <Trans>Error retrieving the extensions</Trans>
           </Text>
         </Line>
       </Column>
     );
 
-  return extensions ? (
+  return (
     <Accordion>
       <AccordionHeader>
         <Text displayInlineAsSpan>
@@ -127,8 +124,6 @@ const ExtensionsAccordion = ({
         </Column>
       </AccordionBody>
     </Accordion>
-  ) : (
-    <PlaceholderLoader />
   );
 };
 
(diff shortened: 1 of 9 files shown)


newIDE/app/src/Profile/ContributionsDetails.js
```diff
@@ -0,0 +1,210 @@
...
+    <Column expand>
+      <Text noMargin>{fullName}</Text>
+      <Text noMargin size="body2">
+        {shortDescription}
+      </Text>
+    </Column>
+  </Line>
+);
+
+type Props = {
```
4ian, line 45: Exact type? ({||})

newIDE/app/src/Profile/ContributionsDetails.js
```diff
@@ -0,0 +1,210 @@
...
+  );
+
+  React.useEffect(
+    () => {
+      fetchUserExamples();
+    },
+    [fetchUserExamples]
+  );
+
+  const ExtensionsAccordion = () => {
```
4ian, line 106: Could you either move this to a different component outside this one (to avoid redeclaring new components at each render, which will make React diffing useless) or rename these to renderXxx with a React.useCallback?

newIDE/app/src/Profile/SubscriptionDetails.js
```diff
@@ -13,43 +13,51 @@ type Props = {
 };
 
 export default ({ subscription, onChangeSubscription }: Props) =>
-  subscription && subscription.planId ? (
+  subscription ? (
```
4ian, line 16: I think users that used to have a subscription but don't have it anymore may have a subscription object but with an empty planId.

ClementPasteau: I think that's fine though? The logic on the `planId` is done below to display the subscription or not.
This check is mainly a proxy to a `loading` prop, to ensure the call is finished, so we can display the title

4ian: You're right, that's totally fine to use this as a way to know if we're loading or not!

newIDE/app/src/Profile/ContributionsDetails.js
```diff
@@ -0,0 +1,234 @@
...
+
+const ExamplesAccordion = ({
+  examples,
+  exampleError,
+}: ExamplesAccordionProps) => {
+  if (exampleError)
+    return (
+      <Column>
+        <Line alignItems="center">
+          <Text size="title">
```
4ian, line 58: I would prefer avoiding using a title for displaying an error. Let's just have a default sized text?

newIDE/app/src/Profile/ContributionsDetails.js
```diff
@@ -0,0 +1,234 @@
...
+              <Trans>Contributions</Trans>
+            </Text>
+          </Line>
+        </Column>
+      )}
+      <ExtensionsAccordion
+        extensions={extensions}
+        extensionError={extensionError}
+      />
+      <ExamplesAccordion examples={examples} exampleError={exampleError} />
```
4ian, line 230: This will show two placeholder loaders - could we have a single one (so that when it's loaded, everything is shown at once?)

newIDE/app/src/Profile/ContributionsDetails.js
```diff
@@ -0,0 +1,234 @@
...
+    <>
+      {extensions && examples && (
+        <Column>
+          <Line alignItems="center">
+            <Text size="title">
+              <Trans>Contributions</Trans>
+            </Text>
+          </Line>
+        </Column>
+      )}
```
4ian, line 225: I think we can this title always shown (it's part of the structure of the page I would say).

newIDE/app/src/Profile/ContributionsDetails.js
```diff
@@ -0,0 +1,234 @@
...
+      <ExtensionsAccordion
+        extensions={extensions}
+        extensionError={extensionError}
+      />
+      <ExamplesAccordion examples={examples} exampleError={exampleError} />
+      <AssetsAccordion examples={examples} />
+    </>
```
4ian, line 232: Idea: we can maybe add a `BackgroundText` (once the loading is done) telling "Missing some contributions? Check that your username is added to each example and extensions you've made or contributed to."

4ian:
Last stuff in addition to my other comments: I think we should have this as much as possible in the Storybook. The top level component is doing some calls so it's not ideal to test, but the accordions could be put in two stories (each having a case where it's working well and a case where it's erroring) so we can quickly check how it looks if we do changes later.

newIDE/app/src/Profile/ContributionsDetails.js
ClementPasteau: I like that, however what will the user do concretely if they realise they're not in the contributors? Open an issue? Send a message on Discord? 
We should maybe be more specific in the message.
How about:
"Missing some contributions? If you are the author, then create a Pull Request on the corresponding repository after adding your username in the authors (example or extension), otherwise ask the author to add you as a contributor!"

4ian: Ok, this maybe then?

"Missing some contributions? If you are the author, create a Pull Request on the corresponding GitHub repository after adding your username in the authors of the example or the extension - or directly ask the original author to add your username."

newIDE/app/src/Profile/ContributionsDetails.js
```diff
@@ -0,0 +1,241 @@
...
+      {examples && extensions ? (
+        <>
+          <ExtensionsAccordion
+            extensions={extensions}
+            extensionError={extensionError}
+          />
+          <ExamplesAccordion examples={examples} exampleError={exampleError} />
+          <AssetsAccordion examples={examples} />
+          <BackgroundText>
+            <Text>
```
4ian, line 226: No need for a `<Text>`, you can just have `<BackgroundText>`

newIDE/app/src/stories/index.js
```diff
@@ -3964,6 +3978,33 @@ storiesOf('UserPublicProfileChip', module)
...
+      />
+    </>
+  ))
+  .add('no contributions', () => (
+    <>
+      <ExtensionsAccordion extensions={[]} extensionError={null} />
+      <ExamplesAccordion examples={[]} exampleError={null} />
+    </>
+  ))
+  .add('with errors', () => (
```
4ian, line 4001: Love this, having even the error states will make sure we handle these properly in the interface.

newIDE/app/src/Profile/ContributionsDetails.js
```diff
@@ -0,0 +1,241 @@
+// @flow
+import { Trans } from '@lingui/macro';
+
+import * as React from 'react';
+import { Accordion } from '../UI/Accordion';
+import { AccordionHeader } from '../UI/Accordion';
+import { AccordionBody } from '../UI/Accordion';
```
4ian, line 7: Nitpicking: Group in a single `import { ... }`?

4ian (approved):

PR #3112, opened 2021-09-30
Change profile.id calls to firebaseUser.id

ClementPasteau:
This ensures we don't prevent an action because the profile is
not fetched yet

Do not show in changelog


4ian (approved):
Looks good!

PR #3115, opened 2021-10-01
Display error if username contains non-ascii chars

ClementPasteau:
non-ascii chars
image

Empty value or spaces
image


newIDE/app/src/Profile/EditProfileDialog.js
```diff
@@ -31,6 +31,8 @@ export const getUsernameErrorText = (error: ?AuthError) => {
 
   if (error.code === 'auth/username-used')
     return 'This username is already used, please pick another one.';
+  if (error.code === 'auth/malformed-username')
+    return 'You cannot use this username, please only pick common characters.';
```
4ian, line 35: We can be a bit more specific: You cannot use this username. Use a short username with only alphanumeric characters, as well as _ and -.

4ian (approved):

PR #3120, opened 2021-10-01
Add username validation in the create & edit forms

ClementPasteau:
Add frontend validation for username field, both in create & edit form.
Empty value still allowed in created form, in case the user doesn't want to set a username for the moment.

Screenshot 2021-10-01 at 14 04 07
Screenshot 2021-10-01 at 14 04 26

### DIFF
--- a/newIDE/app/src/Profile/UsernameField.js
+++ b/newIDE/app/src/Profile/UsernameField.js
@@ -1,0 +1,40 @@
+// @flow
+import { Trans } from '@lingui/macro';
+import * as React from 'react';
+import TextField from '../UI/TextField';
+
+type Props = {|
+  value: string,
+  onChange: Function,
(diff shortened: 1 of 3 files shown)


newIDE/app/src/Profile/UsernameField.js
```diff
@@ -0,0 +1,40 @@
...
+export const usernameFormatError =
+  'Please pick a short username with only alphanumeric characters as well as _ and -';
+
+export const UsernameField = ({
+  value,
+  onChange,
+  errorText,
+  allowEmpty,
+}: Props) => {
+  const _getUsernameErrorText = () =>
```
4ian, line 27: nit: `getUsernameErrorText`

4ian (approved):

PR #3123, opened 2021-10-01
Give possibility to enter one's bio

AlexandreSi:


ClementPasteau (commented):
lgtm, should we also add the text on the ProfileDetails component to show that this value is actually used somewhere?

AlexandreSi:
Good question. It is not used for now but I feel it would be a bit awkward to see it on this component but I don't know why.

4ian:
That's a good question. I feel we could indeed display it so that we have a start of a "real" profile shown...
The "you are connected as xxxx" is a bit awkward already though.

newIDE/app/src/Profile/EditProfileDialog.js
```diff
@@ -97,6 +99,23 @@ export default class EditDialog extends Component<Props, State> {
...
             errorText={getUsernameErrorText(error)}
           />
+          <TextField
+            value={this.state.form.description}
+            floatingLabelText={<Trans>Bio</Trans>}
+            fullWidth
+            multiline
+            rows={3}
+            rowsMax={5}
+            hintText={t`I use GDevelop to make the videogame of my dreams`}
```
4ian, line 109: I would mark this as a question: "What are you using GDevelop for?" :)

ClementPasteau (approved):

4ian (approved):

PR #3125, opened 2021-10-01
Rework profile details

AlexandreSi:
Change My profile page to show the user's email and bio, introduced in #3123

Result:

image

### DIFF
--- a/newIDE/app/src/Profile/ProfileDetails.js
+++ b/newIDE/app/src/Profile/ProfileDetails.js
@@ -25,7 +25,15 @@ export default ({ profile, onEditProfile }: Props) => {
           <Line alignItems="center">
             <Avatar src={getGravatarUrl(profile.email || '', { size: 40 })} />
             <Spacer />
-            <Text size="title">{profile.username}</Text>
+            <Text
+              size="title"
+              style={{
+                color: profile.username ? 'black' : '#888',
+              }}
+            >
+              {profile.username ||
+                i18n._(t`Edit your profile to pick a username!`)}
+            </Text>
           </Line>
           <Line>
             <TextField
@@ -38,7 +46,12 @@ export default ({ profile, onEditProfile }: Props) => {
           </Line>
           <Line>
             <TextField
-              value={profile.description || i18n._(t`No bio defined`)}
+              value={
+                profile.description ||
+                i18n._(
+                  t`No bio defined. Edit your profile to tell us what you are using GDevelop for!`
+                )
+              }
               readOnly
               fullWidth
               multiline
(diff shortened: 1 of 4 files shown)


Silver-Streak:
SUPER minor comment, and while I know this isn't a huge deal, as an outsider looking in I'd really love to see these PRs with descriptions when submitted, so I can help connect the community on the discord/forums to any WIP upcoming changes to help reduce duplicate issues/fears/etc.

4ian:
Ah interesting remark - I think also it might be due to the PR being a draft but there should be no problem ensuring we add at least a quick description for PRs in the future :) 👍

AlexandreSi:
Hi @Silver-Streak I didn't realize you were using the PRs as a source of info!
I will add a description to my PRs from now on, starting with this one!

AlexandreSi:
@4ian I am not sure what to put instead of "No bio defined", something like "Edit your profile to tell us what you are using GDevelop for!" ?
Maybe I can add the same logic for the username. For now, there will be nothing next to the avatar if the user does not have a username

4ian:
> Maybe I can add the same logic for the username. For now, there will be nothing next to the avatar if the user does not have a username

Great idea!

As for the bio, yes you can go for a "No bio defined. Edit your profile to tell us what you are using GDevelop for!" :)

AlexandreSi:
Final result:
image
image

newIDE/app/src/Profile/ProfileDetails.js
```diff
@@ -17,25 +19,61 @@ type Props = {
...
+    <I18n>
+      {({ i18n }) => (
+        <Column>
+          <Line alignItems="center">
+            <Avatar src={getGravatarUrl(profile.email || '', { size: 40 })} />
+            <Spacer />
+            <Text
+              size="title"
+              style={{
+                color: profile.username ? 'black' : '#888',
```
4ian, line 31: Risky as could not work well with a future theme (even if gray is fairly safe, black is not and won't show well on dark themes).
Instead, keep the default color and use something like `opacity: 0.8` to dim a bit the text.

newIDE/app/src/Profile/ProfileDetails.js
```diff
@@ -17,25 +19,61 @@ type Props = {
...
+                )
+              }
+              readOnly
+              fullWidth
+              multiline
+              floatingLabelText={<Trans>Bio</Trans>}
+              floatingLabelFixed={true}
+              inputStyle={{
+                color: profile.description ? 'black' : '#888',
+              }}
```
4ian, line 62: Here, prefer passing a hint text when the value is empty :) 
This way, no need for inputStyle.

newIDE/app/src/Profile/ProfileDetails.js
AlexandreSi: Oh yes, I forgot about the themes!

newIDE/app/src/Profile/ProfileDetails.js
AlexandreSi: I thought that there was a problem with hint text, but not at all indeed. Will use it

4ian (approved):

PR #3127, opened 2021-10-03
Fix #3097 (dialogue variables even if dialogue is not running

### LINKED ISSUE #3097: Dialogue Tree Extension: DialogueTree::Variable only returns the correct value if dialogue is running
## The bug
When using the Dialogue Tree extension's expression DialogueTree::Variable("someVariableName"), the expression gives a value of 0 unless the condition Dialogue is running would return true. This makes it difficult to use dialogue variables to control game state.

## To Reproduce
1. use Dialogue Tree to run dialogue
2. have that dialogue branch set a Yarn variable
3. end the dialogue (Dialogue is no longer running)
4. attempt to use that variable in an expression
5. See that the expression gives 0

## Other details
* The Set dialogue state variable action does not work this way; dialogue variables can be set at any time during the game. At the very least setting a dialogue variable and getting a dialogue variable should behave consistently.

MechanicalPen:
This PR will fix a design discrepancy where setting a dialogue state variable in an action works whether or not the player is in dialogue but getting a dialogue state variable through an expression or condition does not work if the player is not in dialogue.

### DIFF
--- a/Extensions/DialogueTree/dialoguetools.ts
+++ b/Extensions/DialogueTree/dialoguetools.ts
@@ -739,7 +739,7 @@ namespace gdjs {
    * @param key The name of the variable you want to get the value of
    */
   gdjs.dialogueTree.getVariable = function (key: string) {
-    if (this.dialogueIsRunning && key in this.runner.variables.data) {
+    if (this.runner.variables && key in this.runner.variables.data) {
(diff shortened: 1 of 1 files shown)


4ian (approved):
Looks good! Thanks for fixing this.

PR #3130, opened 2021-10-04
Fix login problems

### LINKED ISSUE #3113: Empty Authors/Usernames bug
### Empty author names
Empty author names can be added by:
1) Type fast and press enter before the search result come
2) Input wrong user names

The other problem is that if there is a empty author name and the project properties is re-opened, all the authors are remove, no matter if the usernames are valid. **The valid usernames are only removed when there are more than two empty usernames**
 
Video provided below:

https://user-images.githubusercontent.com/73597906/135487446-c20436e0-c993-40da-b08e-e0ed6a969a09.mp4

### Empty Usernames

There can be empty usernames/ with spaces like seen in the video below

**If it's an empty username, it won't come up in the author search (Not shown in video)**

https://user-images.githubusercontent.com/73597906/135487435-cb66a7d6-b133-4c80-a55b-3462fb86f007.mp4
## Can't get my username back
I was changing my username to empty, with only spaces, etc.. back and forth and recording the bug with the empty username. Now I can't get my user name back (`Leo Red (Midhil)`), is there any solution? It is definitely a bug

myusername

## To Reproduce
The step to reproduce can be seen in the videos :)

## Other details
* Linux Mint 20.2
* beta118

AlexandreSi:
Resolves #3113

### DIFF
--- a/newIDE/app/src/Utils/GDevelopServices/Authentication.js
+++ b/newIDE/app/src/Utils/GDevelopServices/Authentication.js
@@ -65,10 +65,11 @@ export default class Authentication {
     onAuthStateChanged(this.auth, user => {
       if (user) {
         this.firebaseUser = user;
-        if (this._onUserChangeCallBack) this._onUserChangeCallBack();
       } else {
         this.firebaseUser = null;
       }
+
+      if (this._onUserChangeCallBack) this._onUserChangeCallBack();
     });
   }
 
--- a/newIDE/app/src/Utils/GDevelopServices/Authentication.js
+++ b/newIDE/app/src/Utils/GDevelopServices/Authentication.js
@@ -65,6 +65,7 @@ export default class Authentication {
     onAuthStateChanged(this.auth, (user => {
       if (user) {
         this.firebaseUser = user;
+        if (this._onUserChangeCallBack) this._onUserChangeCallBack();
       } else {
         this.firebaseUser = null;
       }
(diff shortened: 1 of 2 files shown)


AlexandreSi:
There are 2 apparently useless commits.
I made the first commit because when use the Chrome Debugger, I noticed that `this` appeared undefined, and binding them would "define" this in the debugger.
But Flow raised the error "Unnecessary binding" so I removed them and it still works.
I never noticed that in debugger, so good to know

newIDE/app/src/Utils/GDevelopServices/Authentication.js
```diff
@@ -57,22 +57,23 @@ export default class Authentication {
...
-  _onUserChangeCb: ?() => void = null;
+  _onUserChangeCallBack: ?() => void = null;
 
   constructor() {
     const app = initializeApp(GDevelopFirebaseConfig);
     this.auth = getAuth(app);
     onAuthStateChanged(this.auth, user => {
       if (user) {
         this.firebaseUser = user;
+        if (this._onUserChangeCallBack) this._onUserChangeCallBack();
```
4ian, line 68: I reckon we can even put this at the end of the function, to always call the callback when the user has changed.
This way, we handle the user being disconnected for some reason by firebase (the callback is called => _fetchUserProfile is called => the authenticatedUser is cleared).

Actually what we had before:
image

4ian (approved):
> I made the first commit because when use the Chrome Debugger, I noticed that this appeared undefined, and binding them would "define" this in the debugger.

I think this is because of the babel transforming the code, this can be renamed to _this or this2 or something like this (see the pun? 😎)

> But Flow raised the error "Unnecessary binding" so I removed them and it still works.

Yep with an arrow function it's not necessary to bind this.

4ian (approved):
🚀

PR #3132, opened 2021-10-04
Allow user to verify their email

ClementPasteau:
Since the creation of user profiles, email didn't need to be verified.
This check is not to block any feature, but rather to ensure the user has selected their correct email, and maybe in the future we can push information to their email if they're willing. (for example, stats on their games)

Tested:

- Verifying email after signup then opening profile -> button not visible
- Verifying email after signup with profile open, then click on "Send it again" -> button disappears without sending email
- Clicking too many times on "send it again" (Firebase "too-many-requests" error) -> message shown to user to wait
- Verifying email via the popup and waiting for the popup to reload the user -> success message
- Verifying email via the popup but closing the popup before verification is done -> button disappears either by reloading the profile or clicking again on the button

Screenshot 2021-10-04 at 18 07 03
Screenshot 2021-10-04 at 18 09 46
Screenshot 2021-10-04 at 18 09 57
Screenshot 2021-10-05 at 14 36 06
Screenshot 2021-10-05 at 10 05 41

### DIFF
--- a/newIDE/app/src/Profile/AuthenticatedUserProvider.js
+++ b/newIDE/app/src/Profile/AuthenticatedUserProvider.js
@@ -70,7 +70,7 @@ export default class AuthenticatedUserProvider extends React.Component<
   }
 
   _resetAuthenticatedUser() {
-    this.setState({
+    this.setState(({ authenticatedUser }) => ({
       authenticatedUser: {
         ...initialAuthenticatedUser,
         onLogout: this._doLogout,
@@ -83,40 +83,10 @@ export default class AuthenticatedUserProvider extends React.Component<
         getAuthorizationHeader: () =>
           this.props.authentication.getAuthorizationHeader(),
       },
-    });
+    }));
   }
 
-  _reloadFirebaseProfile = () => {
-    const { authentication } = this.props;
-
-    authentication.getFirebaseUser((err, firebaseUser: ?FirebaseUser) => {
-      if (err && err.unauthenticated) {
-        return this.setState({
-          authenticatedUser: {
-            ...this.state.authenticatedUser,
-            authenticated: false,
-            profile: null,
-            usages: null,
-            limits: null,
-            subscription: null,
-          },
-        });
-      } else if (err || !firebaseUser) {
-        console.log('Unable to fetch user profile', err);
-        return;
-      }
-
-      this.setState({
-        authenticatedUser: {
-          ...this.state.authenticatedUser,
-          authenticated: true,
-          firebaseUser,
-        },
-      });
-    });
-  };
-
-  _fetchUserProfile = () => {
+  _reloadFirebaseProfile = (callback?: FirebaseUser => void) => {
     const { authentication } = this.props;
 
     authentication.getFirebaseUser((err, firebaseUser: ?FirebaseUser) => {
@@ -132,7 +102,7 @@ export default class AuthenticatedUserProvider extends React.Component<
           },
         }));
       } else if (err || !firebaseUser) {
-        console.log('Unable to fetch user profile', err);
+        console.error('Unable to fetch user profile', err);
         return;
       }
 
@@ -145,52 +115,59 @@ export default class AuthenticatedUserProvider extends React.Component<
           },
         }),
         () => {
-          if (!firebaseUser) return;
-
-          authentication
-            .getUserProfile(authentication.getAuthorizationHeader)
-            .then(user =>
-              this.setState(({ authenticatedUser }) => ({
-                authenticatedUser: {
-                  ...authenticatedUser,
-                  profile: user,
-                },
-              }))
-            );
-          getUserUsages(
-            authentication.getAuthorizationHeader,
-            firebaseUser.uid
-          ).then(usages =>
-            this.setState(({ authenticatedUser }) => ({
-              authenticatedUser: {
-                ...authenticatedUser,
-                usages,
-              },
-            }))
-          );
-          getUserSubscription(
-            authentication.getAuthorizationHeader,
-            firebaseUser.uid
-          ).then(subscription =>
-            this.setState(({ authenticatedUser }) => ({
-              authenticatedUser: {
-                ...authenticatedUser,
-                subscription,
-              },
-            }))
-          );
-          getUserLimits(
-            authentication.getAuthorizationHeader,
-            firebaseUser.uid
-          ).then(limits =>
-            this.setState(({ authenticatedUser }) => ({
-              authenticatedUser: {
-                ...authenticatedUser,
-                limits,
-              },
-            }))
-          );
+          if (!firebaseUser || !callback) return;
+          callback(firebaseUser);
         }
+      );
+    });
+  };
+
+  _fetchUserProfile = () => {
+    const { authentication } = this.props;
+
+    this._reloadFirebaseProfile((firebaseUser: FirebaseUser) => {
+      authentication
+        .getUserProfile(authentication.getAuthorizationHeader)
+        .then(user =>
+          this.setState(({ authenticatedUser }) => ({
+            authenticatedUser: {
+              ...authenticatedUser,
+              profile: user,
+            },
+          }))
+        );
+      getUserUsages(
+        authentication.getAuthorizationHeader,
+        firebaseUser.uid
+      ).then(usages =>
+        this.setState(({ authenticatedUser }) => ({
+          authenticatedUser: {
+            ...authenticatedUser,
+            usages,
+          },
+        }))
+      );
+      getUserSubscription(
+        authentication.getAuthorizationHeader,
+        firebaseUser.uid
+      ).then(subscription =>
+        this.setState(({ authenticatedUser }) => ({
+          authenticatedUser: {
+            ...authenticatedUser,
+            subscription,
+          },
+        }))
+      );
+      getUserLimits(
+        authentication.getAuthorizationHeader,
+        firebaseUser.uid
+      ).then(limits =>
+        this.setState(({ authenticatedUser }) => ({
+          authenticatedUser: {
+            ...authenticatedUser,
+            limits,
+          },
+        }))
       );
     });
   };
--- a/newIDE/app/src/Profile/AuthenticatedUserProvider.js
+++ b/newIDE/app/src/Profile/AuthenticatedUserProvider.js
@@ -24,6 +24,7 @@ import AuthenticatedUserContext, {
 } from './AuthenticatedUserContext';
 import CreateAccountDialog from './CreateAccountDialog';
 import EditProfileDialog from './EditProfileDialog';
+import EmailVerificationPendingDialog from './EmailVerificationPendingDialog';
 
 type Props = {|
   authentication: Authentication,
@@ -40,6 +41,7 @@ type State = {|
   editInProgress: boolean,
   authError: ?AuthError,
   resetPasswordDialogOpen: boolean,
+  emailVerificationPendingDialogOpen: boolean,
   forgotPasswordInProgress: boolean,
 |};
 
@@ -57,6 +59,7 @@ export default class AuthenticatedUserProvider extends React.Component<
     editInProgress: false,
     authError: null,
     resetPasswordDialogOpen: false,
+    emailVerificationPendingDialogOpen: false,
     forgotPasswordInProgress: false,
   };
 
@@ -75,11 +78,43 @@ export default class AuthenticatedUserProvider extends React.Component<
         onEdit: () => this.openEditProfileDialog(true),
         onCreateAccount: () => this.openCreateAccountDialog(true),
         onRefreshUserProfile: this._fetchUserProfile,
+        onRefreshFirebaseProfile: this._reloadFirebaseProfile,
+        onSendEmailVerification: this._doSendEmailVerification,
         getAuthorizationHeader: () =>
           this.props.authentication.getAuthorizationHeader(),
       },
     });
   }
+
+  _reloadFirebaseProfile = () => {
+    const { authentication } = this.props;
+
+    authentication.getFirebaseUser((err, firebaseUser: ?FirebaseUser) => {
+      if (err && err.unauthenticated) {
+        return this.setState({
+          authenticatedUser: {
+            ...this.state.authenticatedUser,
+            authenticated: false,
+            profile: null,
+            usages: null,
+            limits: null,
+            subscription: null,
+          },
+        });
+      } else if (err || !firebaseUser) {
+        console.log('Unable to fetch user profile', err);
+        return;
+      }
+
+      this.setState({
+        authenticatedUser: {
+          ...this.state.authenticatedUser,
+          authenticated: true,
+          firebaseUser,
+        },
+      });
+    });
+  };
 
   _fetchUserProfile = () => {
     const { authentication } = this.props;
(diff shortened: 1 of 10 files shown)


newIDE/app/src/Profile/ProfileDialog.js
```diff
@@ -88,10 +89,12 @@ export default class ProfileDialog extends Component<Props, State> {
               <Tab label={<Trans>Services Usage</Trans>} value="usage" />
             </Tabs>
             {this.state.currentTab === 'profile' &&
-              (authenticatedUser.authenticated && authenticatedUser.profile ? (
+              (authenticatedUser.authenticated &&
+              authenticatedUser.profile &&
+              authenticatedUser.firebaseUser ? (
```
ClementPasteau, line 94: not needed anymore

newIDE/app/src/Utils/GDevelopServices/Authentication.js
```diff
@@ -140,6 +142,8 @@ export default class Authentication {
   getFirebaseUser = (cb: (any, ?FirebaseUser) => void) => {
     if (!this.isAuthenticated()) return cb({ unauthenticated: true });
 
+    this.auth.currentUser.reload();
```
ClementPasteau, line 145: i'm wondering if the reload should happen in this function (which is called in multiple places) or in another function, like `fetchFirebaseUser`

ClementPasteau: I think we can leave it here. the function `getFirebaseUser` is used in only 1 place, when the full profile info is fetched. I think it makes sense to reload the data from Firebase too at that time to make sure we've got correct information.

newIDE/app/src/Profile/AuthenticatedUserProvider.js
```diff
@@ -75,12 +78,44 @@ export default class AuthenticatedUserProvider extends React.Component<
...
+            limits: null,
+            subscription: null,
+          },
+        });
+      } else if (err || !firebaseUser) {
+        console.log('Unable to fetch user profile', err);
+        return;
+      }
+
+      this.setState({
```
4ian, line 109: Use `({authenticatedUser}) => ({authenticatedUser: {...authenticatedUser, ...}})` form which is a bit verbose but will ensure you don't erase the state with an old one.

newIDE/app/src/Profile/AuthenticatedUserProvider.js
```diff
@@ -75,12 +78,44 @@ export default class AuthenticatedUserProvider extends React.Component<
...
+            ...this.state.authenticatedUser,
+            authenticated: false,
+            profile: null,
+            usages: null,
+            limits: null,
+            subscription: null,
+          },
+        });
+      } else if (err || !firebaseUser) {
+        console.log('Unable to fetch user profile', err);
```
4ian, line 105: Might as well be a console.error

newIDE/app/src/Profile/AuthenticatedUserProvider.js
```diff
@@ -75,12 +78,44 @@ export default class AuthenticatedUserProvider extends React.Component<
...
       },
     });
   }
 
+  _reloadFirebaseProfile = () => {
+    const { authentication } = this.props;
+
+    authentication.getFirebaseUser((err, firebaseUser: ?FirebaseUser) => {
+      if (err && err.unauthenticated) {
+        return this.setState({
```
4ian, line 94: Use `({authenticatedUser}) => ({authenticatedUser: {...authenticatedUser, ...}})` form.

This _reloadFirebaseProfile is exactly the beginning of `_fetchUserProfile`. I reckon we can implement `_fetchUserProfile` using it then (either by passing it an optional callback to be called with the `firebaseUser` or by reworking this to return a promise).

newIDE/app/src/Profile/EmailVerificationPendingDialog.js
```diff
@@ -0,0 +1,90 @@
...
+        ) : (
+          <FlatButton
+            label={<Trans>Cancel and close</Trans>}
+            key="close"
+            primary={false}
+            onClick={onClose}
+          />
+        ),
+      ]}
+      title={undefined}
```
4ian, line 53: You can just simply not write the title prop I think?

newIDE/app/src/Profile/EmailVerificationPendingDialog.js
```diff
@@ -0,0 +1,90 @@
...
+      maxWidth="sm"
+      cannotBeDismissed={true}
+      open
+      noMargin
+    >
+      {!isVerified && (
```
4ian, line 59: Uber nitpicking: use `isVerified ? (...) : (...)`? 
Might emphasize that this dialog has 2 states, no more.

newIDE/app/src/stories/index.js
```diff
@@ -3876,12 +3876,15 @@ storiesOf('ProfileDetails', module)
...
       onEditProfile={action('edit profile')}
     />
   ))
   .add('loading', () => (
-    <ProfileDetails profile={null} onEditProfile={action('edit profile')} />
+    <ProfileDetails
+      authenticatedUser={fakeAuthenticatedButLoadingAuthenticatedUser}
+      onEditProfile={action('edit profile')}
+    />
   ));
```
4ian, line 3889: Can you add a story for `EmailVerificationPendingDialog` with:

- fakeAuthenticatedButLoadingAuthenticatedUser (i.e: "loading" state)
- fakeIndieAuthenticatedUser (i.e: "not email verified user")
- fakeIndieAuthenticatedAndEmailVerifiedUser (a new fixture to add :)) (i.e: "email verified user").

This story will help to quickly ensure the design of the dialog is ok when we refactor/change things later (because this is the kind of dialog we don't see every day, so it's especially important they have a story so that we can visit them easily to ensure they are doing good :))

newIDE/app/src/Utils/UseInterval.js
```diff
@@ -0,0 +1,27 @@
+// @flow
+import { useEffect, useRef } from 'react';
+
+/**
+ * Creates an interval effect for a callback, with a specified delay.
+ */
+export const useInterval = (callback: Function, delay: number | null) => {
```
4ian, line 7: Avoid Function in most cases. Here it can be simply `() => void`

newIDE/app/src/Utils/UseInterval.js
```diff
@@ -0,0 +1,27 @@
...
+  });
+
+  useEffect(
+    () => {
+      function tick() {
+        if (savedCallback.current) savedCallback.current();
+      }
+
+      if (delay !== null) {
+        let id = setInterval(tick, delay);
```
4ian, line 21: nitpicking: can be `const id`

newIDE/app/src/Utils/GDevelopServices/Authentication.js
```diff
@@ -141,8 +144,25 @@ export default class Authentication {
 
   getFirebaseUser = (cb: (any, ?FirebaseUser) => void) => {
     if (!this.isAuthenticated()) return cb({ unauthenticated: true });
+    this.auth.currentUser.reload().then(() => cb(null, this.firebaseUser));
```
4ian, line 147: Interesting, why is this useful/needed?

newIDE/app/src/Utils/GDevelopServices/Authentication.js
```diff
@@ -141,8 +144,25 @@ export default class Authentication {
...
+  };
 
-    cb(null, this.firebaseUser);
+  sendFirebaseEmailVerification = (cb: () => void): Promise<void> => {
+    return this.auth.currentUser.reload().then(() => {
+      if (this.firebaseUser && !this.firebaseUser.emailVerified) {
```
4ian, line 152: ubernitpicking: I tend to early return when you want to dismiss a "not happy path". This saves one indentation and allow to reason like "if this case happens, just bail out".

```suggestion
      if (!this.firebaseUser || this.firebaseUser.emailVerified) return;
```

This reads fairly clearly: "if you don't have a user or it's verified, just stop here".
And you can continue reading the code without remembering "ok there is maybe a else/else if later in the code".

newIDE/app/src/Utils/GDevelopServices/Authentication.js
```diff
@@ -141,8 +144,25 @@ export default class Authentication {
...
+        sendEmailVerification(this.auth.currentUser)
+          .then(() => {
+            cb();
+          })
+          .catch((error: Error) => {
+            showErrorBox({
```
4ian, line 158: Nit: prefer `then(() => { cb(); }, (error: Error) => { ... })`; in most cases, because the catch will swallow errors that are launched by cb (making debugging of an unrelated error in your callback difficult).
With the `then(cb, errorCb)` form, the error callback is only called for the sendEmailVerification related errors - which is good

newIDE/app/src/Profile/SubscriptionPendingDialog.js
```diff
@@ -11,34 +11,13 @@ import CircularProgress from '@material-ui/core/CircularProgress';
 import BackgroundText from '../UI/BackgroundText';
 import VerifiedUser from '@material-ui/icons/VerifiedUser';
 import Text from '../UI/Text';
+import { useInterval } from '../Utils/UseInterval';
```
4ian, line 14: Nice refactoring 👍

newIDE/app/src/Profile/ProfileDetails.js
```diff
@@ -4,67 +4,121 @@ import { Trans, t } from '@lingui/macro';
...
+      authenticatedUser.onSendEmailVerification();
+      setEmailSent(true);
+      setTimeout(() => setEmailSent(false), 3000);
+    },
+    [authenticatedUser]
+  );
+
+  const loadUserProfile = React.useCallback(
+    () => authenticatedUser.onRefreshUserProfile(),
+    [authenticatedUser.onRefreshUserProfile] // eslint-disable-line react-hooks/exhaustive-deps
```
4ian, line 38: Add a comment explaining why this is disabled

newIDE/app/src/Profile/ProfileDetails.js
```diff
@@ -4,67 +4,121 @@ import { Trans, t } from '@lingui/macro';
...
-  return profile ? (
+export default ({ onEditProfile, authenticatedUser }: Props) => {
+  const profile = authenticatedUser.profile;
+  const firebaseUser = authenticatedUser.firebaseUser;
+  const [emailSent, setEmailSent] = React.useState<boolean>(false);
+  const sendEmail = React.useCallback(
+    () => {
+      authenticatedUser.onSendEmailVerification();
+      setEmailSent(true);
+      setTimeout(() => setEmailSent(false), 3000);
```
4ian, line 31: We're "leaking" this timeout, in the sense that if I close the dialog it will fire after 3 seconds. Not a huge problem (react will probably still complain we try to modify the state of a closed component).
Would still maybe be a good idea to use `useIsMounted`, and in the setTimeout callback, check `if (!isMounted.current) return;`

newIDE/app/src/Profile/EmailVerificationPendingDialog.js
```diff
@@ -0,0 +1,90 @@
...
+      )}
+      {isVerified && (
+        <Column>
+          <Line>
+            <Text>
+              <Trans>Your email is now verified!</Trans>
+            </Text>
+            <Spacer />
+            <VerifiedUser />
+          </Line>
```
4ian, line 85: Design speaking: look at the alignment of the icon:
image

We probably want to align it vertically. We also have the loading that was horizontally centered, so we might want to do the same thing:
image

Something like?

```suggestion
          <Line justifyContent="center" alignItems="center">
            <VerifiedUser />
            <Spacer />
            <Text>
              <Trans>Your email is now verified!</Trans>
            </Text>
          </Line>
```

(also note that this kind of design nits are easier to refactor/fix later if we have a storybook story :))

4ian:
> we should probably also add an overall timer on the dialog, to ensure no-one falls asleep in front of it and makes requests every 3seconds - and do the same for the SubcriptionsDialog?

Not sure if it already happened - though we could use either an exponential backoff or limit to ~5 minutes (some people can take a lot of time). I think anyway very few people would keep this open (and it you "navigate" outside of the tab if it's in the browser, the intervals might be stopped by the browser?).
For now no need to make things more complicated though I would say :)

newIDE/app/src/Utils/GDevelopServices/Authentication.js
ClementPasteau: When opening the profile, the function `_fetchUserProfile` calls this function to ensure the profile is updated.
However, it gets the Firebase profile from the state, meaning that if the prop `emailVerified` has changed, it will not be propagated. Thus, reload is needed here.

So we can either:

- just call `onRefreshFirebaseProfile` on profile dialog open
- or call the whole `onRefreshUserProfile` but call a `reload()` on the firebase profile.

I suggest the second and I can add a comment explaining why this is needed

4ian: Fair enough. This method `getFirebaseUser` is only used by `_fetchUserProfile`, so seems ok/low risk!

newIDE/app/src/stories/index.js
ClementPasteau: added 2 stories, turns out the "non-verified user" is the same as the one "loading" :) (and waiting to be verified!)

newIDE/app/src/Profile/AuthenticatedUserProvider.js
```diff
@@ -75,20 +78,22 @@ export default class AuthenticatedUserProvider extends React.Component<
...
-  _fetchUserProfile = () => {
+  _reloadFirebaseProfile = (callback?: FirebaseUser => void) => {
     const { authentication } = this.props;
 
     authentication.getFirebaseUser((err, firebaseUser: ?FirebaseUser) => {
       if (err && err.unauthenticated) {
         return this.setState(({ authenticatedUser }) => ({
           authenticatedUser: {
-            ...authenticatedUser,
+            ...this.state.authenticatedUser,
```
4ian, line 96: Careful here, this could be a regression.

newIDE/app/src/Utils/GDevelopServices/Authentication.js
```diff
@@ -141,8 +144,27 @@ export default class Authentication {
 
   getFirebaseUser = (cb: (any, ?FirebaseUser) => void) => {
     if (!this.isAuthenticated()) return cb({ unauthenticated: true });
+    // In order to fetch the latest firebaseUser properties (like emailVerified)
+    // We have to call the reload method.
```
4ian, line 148: nit:
```suggestion
    // In order to fetch the latest firebaseUser properties (like emailVerified),
    // we have to call the reload method.
```

newIDE/app/src/fixtures/GDevelopServicesTestData.js
```diff
@@ -19,6 +19,13 @@ import {
 } from '../Utils/GDevelopServices/Asset';
 
 export const indieFirebaseUser: FirebaseUser = {
+  uid: 'indie-user',
+  providerId: 'fake-provider.com',
+  email: 'indie-user@example.com',
+  emailVerified: false,
+};
+
+export const indieFirebaseVerifiedUser: FirebaseUser = {
```
4ian, line 28: ubernitpicking (sorry): it may be a question of style, but I would name this... `indieVerifiedFirebaseUser` 😁
Because this thing IS a "FirebaseUser", because it has the type "FirebaseUser". So this means Firebase can't be removed from User, they must stay together. So the adjective "Indie" and "Verified" comes before: verifiedIndieFirebaseUser or indieVerifiedFirebaseUser.

Sorry 😅

4ian (approved):
Lgtm!

newIDE/app/src/Profile/AuthenticatedUserProvider.js
ClementPasteau: interesting, must have come out of a bad conflict resolution. :+1:

newIDE/app/src/fixtures/GDevelopServicesTestData.js
ClementPasteau: Good feedback :)

4ian:
Great to see this email verification finally done!

PR #3137, opened 2021-10-05
Add public profile

AlexandreSi:
Gives the possibility to display another user's profile when clicking on:
- TBD


newIDE/app/src/stories/componentStories/ProfileDetails.stories.js
```diff
@@ -0,0 +1,44 @@
...
+  component: ProfileDetails,
+  decorators: [paperDecorator, muiDecorator],
+  argTypes: {
+    profile: {
+      control: { type: 'radio' },
+      options: ['Complete profile', 'Without username nor bio'],
+      defaultValue: 'Complete profile',
+      mapping: {
+        'Complete profile': indieUserProfile,
+        'Without username nor bio': {
```
4ian, line 23: This is not typed and could break if one day username/description is not part anymore of the type/renamed. Either make a `indieWithUsernameAndDescriptionUserProfile` or add a `({ ... }): Profile` around this so that we're sure the type stays correct in the future :)

newIDE/app/src/Utils/UsersAutocomplete.js
```diff
@@ -83,7 +83,7 @@ export const UsersAutocomplete = (props: Props) => {
         );
         setUsers(
           Object.keys(userPublicProfilesByIds).map(userId => {
-            const userPublicProfile: UserPublicProfile =
+            const userPublicProfile: UserPublicProfileSearch =
```
4ian, line 86: @AlexandreSi I know this PR is not finished but I still took a look after looking at your backend PR - in the backend the `UserPublicProfileByIds` is indeed returning a `UserPublicProfile` (and not a `UserPublicProfileSearch` - and this is ok!).
So I guess here we need to keep a `UserPublicProfile`. 

Is this overfetching? If it was the search part of the autocomplete yes, but because this is the part used to display usernames and later possibly avatars, I don't think so or that it has an importance.

So I think we can keep UserPublicProfile here. At least does not sounds like a huge overfetching.

4ian: (Or adapt the backend to return `UserPublicProfileSearch` for the getUserPublicProfileByIds? That's a possibility too, but would be part I guess of an upcoming PR to clean the separation between UserPublicProfile endpoints and UserPublicProfileSearch :))

AlexandreSi:
Opening #3141 because of conflicts too complicated to solve

PR #3139, opened 2021-10-06
Prevent calling a user update on every authState change.

ClementPasteau:
This fixes a race condition where the fetchUser would be called too early, causing a conflict with the signup function while creating the user in the database.

Do not show in changelog


4ian (approved):

4ian:
@AlexandreSi Take a look and let us know if you have questions, as I think it's an interesting change :)

PR #3140, opened 2021-10-06
Feature/email change

ClementPasteau:
Give the ability to users to change their email address!

Tested:
- changing directly after login, before email verified
- changing after email verified
- verifying email, then changing it
- changing email after being connected for a while (errors -> need to login again for identity verification)
- changing email after reloading the app
- changing email to an existing email (error)

Screenshot 2021-10-06 at 15 36 37
Screenshot 2021-10-06 at 15 36 43
Screenshot 2021-10-06 at 15 59 13

### DIFF
--- a/newIDE/app/src/Profile/ChangeEmailDialog.js
+++ b/newIDE/app/src/Profile/ChangeEmailDialog.js
@@ -13,6 +13,7 @@ import {
 import LeftLoader from '../UI/LeftLoader';
 import { ColumnStackLayout } from '../UI/Layout';
 import TextField from '../UI/TextField';
+import { getEmailErrorText } from './CreateAccountDialog';
 
 type Props = {|
   firebaseUser: FirebaseUser,
@@ -25,19 +26,6 @@ type Props = {|
 type State = {|
   form: ChangeEmailForm,
 |};
-
-export const getEmailErrorText = (error: ?AuthError) => {
-  if (!error) return undefined;
-
-  if (error.code === 'auth/invalid-email') return 'This email is invalid';
-  if (error.code === 'auth/email-already-in-use')
-    return 'This email was already used for another account';
-  if (error.code === 'auth/operation-not-allowed')
-    return 'Service seems to be unavailable, please try again later';
-  if (error.code === 'auth/requires-recent-login')
-    return 'Please log out and log in again to verify your identify, then change your email';
-  return undefined;
-};
 
 export default class ChangeEmailDialog extends Component<Props, State> {
   state = {
--- a/newIDE/app/src/Profile/ProfileDetails.js
+++ b/newIDE/app/src/Profile/ProfileDetails.js
@@ -11,7 +11,7 @@ import RaisedButton from '../UI/RaisedButton';
 import TextField from '../UI/TextField';
 import { I18n } from '@lingui/react';
 import FlatButton from '../UI/FlatButton';
-import { ColumnStackLayout } from '../UI/Layout';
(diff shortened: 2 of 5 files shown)


newIDE/app/src/Profile/AuthenticatedUserProvider.js
```diff
@@ -241,7 +247,7 @@ export default class AuthenticatedUserProvider extends React.Component<
     );
   };
 
-  _doForgotPassword = (form: ForgotPasswordForm) => {
+  _doForgotPassword = (form: LoginForm) => {
```
ClementPasteau, line 250: the `ForgotPasswordForm` didn't make sense as it was using the `LoginForm` data

newIDE/app/src/Profile/ChangeEmailDialog.js
```diff
@@ -0,0 +1,105 @@
...
+  onChangeEmail: (form: ChangeEmailForm) => void,
+  changeEmailInProgress: boolean,
+  error: ?AuthError,
+|};
+
+type State = {|
+  form: ChangeEmailForm,
+|};
+
+export const getEmailErrorText = (error: ?AuthError) => {
```
ClementPasteau, line 29: should I create a `EmailField` similar to the `UsernameField` so that the validation is in one place?

newIDE/app/src/Profile/ProfileDetails.js
```diff
@@ -17,11 +17,12 @@ import { type AuthenticatedUser } from './AuthenticatedUserContext';
 import { useIsMounted } from '../Utils/UseIsMounted';
 
 type Props = {|
-  onEditProfile: Function,
```
4ian, line 20: 👍

newIDE/app/src/Profile/ProfileDetails.js
```diff
@@ -118,6 +119,13 @@ export default ({ onEditProfile, authenticatedUser }: Props) => {
               />
             </Line>
             <Line justifyContent="flex-end">
+              <RaisedButton
+                label={<Trans>Change my email</Trans>}
+                primary
```
4ian, line 124: Let's remove primary to give it a bit less visibility compare to editing the profile

newIDE/app/src/Utils/GDevelopServices/Authentication.js
```diff
@@ -169,6 +171,44 @@ export default class Authentication {
...
+  ) => {
+    return updateEmail(this.firebaseUser, form.email)
+      .then(() => {
+        console.log('Email successfully changed in Firebase');
+        return getAuthorizationHeader();
+      })
+      .then(authorizationHeader => {
+        if (!this.firebaseUser) {
+          console.error('Cannot edit user if not logged in');
+          throw new Error('Cannot edit user if not logged in');
```
4ian, line 186: Nit: period at the end of the sentence and also because this is an error message that is not translated, I would be very precise: "Cannot finish editing the user email address as the user is not logged in anymore."

Because when the time will come for this error to appear, we will be very glad to have something that is very precise about what happened.

newIDE/app/src/Profile/ChangeEmailDialog.js
4ian: We could at least move `getEmailErrorText` to a single function in EmailField - even if we don't have a component at least we're having a single handling of the errors (and sure, some errors don't make sense for the account creation or login, but we don't care - better handle too much than too little).

PR #3146, opened 2021-10-07
Fix opening up the parameters dialog for object or behaviors expressions having no parameters

### LINKED ISSUE #3145: [Urgent] All expressions depending on behavior on an object are undefined
## Describe the bug
All expressions depending on behavior on an object are `undefined` when we use the list, the auto complete works fine.
image

## To Reproduce
Steps to reproduce the behaviour:
1. Add a behavior on an object
2. Use an expression from this behavior somewhere in the events.
3. See `undefined.undefined::CellHeight()`

## Other details
B119
Desktop & web app

ClementPasteau:
Fixes #3145


4ian:
Good to merge after comments addressed, let's name this "Fix opening up the parameters dialog for object or behaviors expressions having no parameters" (to be very precise, because this would happen to object expressions too).

PR #3149, opened 2021-10-07
Prevent effects being re-applied on recycled objects

### LINKED ISSUE #3062: Effects bug
## Describe the bug
if you create an infinite falling objects event, AND if you create another event to remove the objects when its out of screen, OR by just adding the behavior (remove when out of screen), the effect settings (numbers) will start to increase every time the object spawn in endless time, and this applies to all effects.
in my example i used glow effect, but the bug happens in every effect as well.

1. Go to object behavior
2. Click on Effects
3. Pick any effect you want
4. Set the behavior (delete when out of screen)
5. Got events
6. Create an action: create object every frame, X:random in range(0,screen width) Y:0
7. Preview the scene and you should see every spawned object effects getting increased in every frame

Link for the example:
[Effects bug.zip](https://github.com/4ian/GDevelop/files/7189528/Effects.bug.zip)

weirddd

* GDevelop 5.0.0-beta 117 / Desktop
* an example of how this bug may face the user: if the user made a simple bullet shoot game and added "delete when off screen" behavior to the bullets, and then he wanted to add effects for the bullets, the bug will happen.

ClementPasteau:
Fixes #3062 

Turns out Sprites (which are the only objects supporting reinitialization) 
that are recycled are applying the effects multiple times, as the effects are not being removed from the object.

### DIFF
--- a/GDJS/Runtime/runtimeobject.ts
+++ b/GDJS/Runtime/runtimeobject.ts
@@ -806,7 +806,7 @@ namespace gdjs {
     }
 
     /**
-     * Remove all effects
+     * Remove all effects.
      */
     clearEffects(): boolean {
       this._rendererEffects = {};


4ian:
I think it works well like you did. I have just one thing I'm not sure about: should we call removeEffect on each effect, or can we take a shortcut and just empty `this._rendererEffects` object?
This is because effects (i.e: PixiJS filters) don't have a need to have a "destroy" method called, so we could just clear the object, which would be a bit faster.

PR #3158, opened 2021-10-09
Use buffer variable when starting/stopping particle emission

AlexandreSi:
To be able to use Repeat X Seconds to start and stop emission.
image

Otherwise, you start and stop the emission on a single frame

### DIFF
--- a/Extensions/ParticleSystem/particleemitterobject.ts
+++ b/Extensions/ParticleSystem/particleemitterobject.ts
@@ -40,6 +40,7 @@ namespace gdjs {
     flow: number;
     /** Destroy the object when there is no particles? */
     destroyWhenNoParticles: boolean;
+    isStarted: boolean;
   };
 
   export type ParticleEmitterObjectData = ObjectData &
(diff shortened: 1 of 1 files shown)


4ian:
While this work, it can also have another unintended side effect, because a start of emission is not immediately considered in the events:

```
Condition : Variable RestartEmission is true
Action: Start emission of Particles

Condition: Particles does not emit anylonger
Action: Destroy Particles
```

Will destroy particles, even if you just restarted the emission -> not great.

What you encountered is a more complex problem I think, which is the fact that there is no easy way to express a "else" in GDevelop events. 
We could consider adding an event containing "Actions" and "Else Actions", but it's a bit more complex than just doing a "if/else" in programming, because we're dealing with list of objects in the events. See this topic where I try to explain two problems:
https://forum.gdevelop-app.com/t/if-else-event-we-have-a-solution/21168/10?u=4ian

The solution could be to actually create two new different event kinds: a "logical if/else (actions to be run if the conditions are false)" and a "object if/else (actions to be run on the objects not fulfilling the conditions)" - but I think we need to very carefully design something so that it's obvious how this/these thing(s) would work.

PR #3162, opened 2021-10-10
Add load JSON resource to variable action

arthuro555:
Adds an action to load a JSON resource into a variable

Example: [jsontest.zip](https://github.com/4ian/GDevelop/files/7317230/jsontest.zip)

### DIFF
--- a/Extensions/JSON/JsExtension.js
+++ b/Extensions/JSON/JsExtension.js
@@ -1,0 +1,68 @@
+// @flow
+/**
+ * This is a declaration of an extension for GDevelop 5.
+ *
+ * ℹ️ Changes in this file are watched and automatically imported if the editor
+ * is running. You can also manually run `node import-GDJS-Runtime.js` (in newIDE/app/scripts).
+ *
+ * The file must be named "JsExtension.js", otherwise GDevelop won't load it.
+ * ⚠️ If you make a change and the extension is not loaded, open the developer console
+ * and search for any errors.
+ *
+ * More information on https://github.com/4ian/GDevelop/blob/master/newIDE/README-extensions.md
+ */
+
+/*::
+// Import types to allow Flow to do static type checking on this file.
+// Extensions declaration are typed using Flow (like the editor), but the files
+// for the game engine are checked with TypeScript annotations.
+import { type ObjectsRenderingService, type ObjectsEditorService } from '../JsExtensionTypes.flow.js'
+*/
+
+module.exports = {
+  createExtension: function (
+    _ /*: (string) => string */,
+    gd /*: libGDevelop */
+  ) {
+    const extension = new gd.PlatformExtension();
+    extension
+      .setExtensionInformation(
+        'JSON',
+        _('JSON'),
+        _('A set of JSON loading actions'),
+        'Arthur Pacaud (arthuro555)',
+        'MIT'
+      )
+      .setExtensionHelpPath('/all-features/json');
+
+    extension
+      .addAction(
+        'LoadJSONResource',
+        _('Load JSON resource'),
+        _('Loads a JSON resource into a scene variable.'),
+        _('Load JSON resource _PARAM1_ in variable _PARAM2_'),
+        _('JSON'),
+        'JsPlatform/Extensions/filesystem_load_file24.png',
+        'JsPlatform/Extensions/filesystem_load_file16.png'
+      )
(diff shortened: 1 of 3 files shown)


4ian:
Should we rather work on the core to expose variables to the events based extensions? This would make this trivial to implement as an extension outside of the main repository (as this is fairly specific, and we tend to have extensions grouping a number of related capabilities).

arthuro555:
I guess? Though personally i think that such an extension (can only ever be made in JavaScript/another language of we add a new platform and exposes directly feature from the core engine) would belong bundled inside the engine, especially since other events based extensions might want to use JSON resources (for example a translations extension could load a translations catalog), and would require that feature to work with a JSON resource. Since we do not currently have extensions cross dependencies and since this is a primitive action to work with a core engine feature (JSON resources) I thought it wouldn't belong as an events based extension.

arthuro555:
If the problem is the extension only having one action we can also move JSON conversion actions/expression to that extension, i had only not done it because I don't want to deal with a breaking change 😬 
Personally i think that JSON should have it's own category, and be part of the engine in order to work with JSON resources.

arthuro555:
ping

arthuro555:
ping

D8H:
> Should we rather work on the core to expose variables to the events based extensions? This would make this trivial to implement as an extension outside of the main repository (as this is fairly specific, and we tend to have extensions grouping a number of related capabilities).

I have a use case for this. I'm doing scalar field level editor example to let people draw a field and save it (with Filesystem). They could load the scalar field at the beginning of a scene to use it as platforms to do a Worms-like for instance. Using a resource will allow games to work in browsers.

It could work with a user extension too, but I'm not sure this is a specific feature. The instance editor makes level loading from resource files avoidable for most cases, but I guess that it can still be useful to implement a level editor with GDevelop to enforce level rules and quality of life tools. In this case the game developer will probably want to embed levels as resources.

For me it seems less specific than Filesystem as resources can be read in any platform.

Spiral

Bouh:
> Personally I think that JSON should have its own category, and be part of the engine in order to work with JSON resources.

I support you!


The big point here is the use of a JSON resource in the game project.
This is very practical because currently, you have to make a request on a file to get the content.
This requires having the right path to the file.
The handling of filesystem delimiter expressions, plus the expressions to retrieve the right file are a big hindrance to use. (In addition, the expression to return the current folder of the game returns the wrong thing, [see the issue](https://github.com/4ian/GDevelop/issues/3596)).

Managing a JSON file as a resource makes the process much simpler because the power of the file selector compared to a bunch of expressions leading to the wrong place makes me lean in favor of the file selector for importing a JSON file as a resource to the project.

arthuro555:
Ping

krunkster:
There is a lot of value in this implementation for developers right now if committed.
There is also a precedent in GDevelop for TileMaps and Dialog Trees in loading JSON data... why not let developers have this more generic avenue to load other types of JSON data with the benefits of the built in resource management.
Please consider accepting this change as-is with some renewed urgency.

arthuro555:
Ping, this is a demanded feature as made clear by the numerous support comments here, and is a no-brainer to review.

4ian:
Can we estimate how hard it would be to add a "Scene Variable" to the extension parameters?
- Ideally we would have a single "variable" parameter type, that can have a global/scene or object variable. BUT this is another task (https://trello.com/c/FwShg063/487-add-support-for-a-generic-variable-parameter-that-would-allow-to-choose-any-variable-object-variable-scene-global-extension-vari)
- Here, we can for now just add 2 parameters: global or scene variables (let's forget object variables for now, unless it's easy).

- Then we expose it to JS. We don't expose it to events (this would need a single "variable" parameter type, so that we can support "local" variables (or even extension variables)).

We release a new version and this extension can now be an extension made in the editor.
And I'm happy to merge it, because it's less demanding in terms of maintenance rather than having this into the builtin extensions that will be supported forever.

arthuro555:
I still believe this is functionality so essential to the use of JSON resources that it should be included in the base engine, but I moved the functionality over to this extension: https://github.com/GDevelopApp/GDevelop-extensions/pull/446

> Then we expose it to JS. We don't expose it to events

That is something that already works and that I've been doing for quite a while already in private extensions. I modify in the JSON the parameter type to "scenevar", then `eventsFunctionContext.getArgument()` already returns the reference to the variable.

Issue #3006, opened 2021-09-04
Auto complete is too sticky!


Bouh:
## Description

The autocomplete is so useful but it's a awful to use, the autocomplete is too sticky.
You wanna insert new line you need cancel the autocomplete
You need insert a space it still sticky.

Here some use case GDevelop vs VSCode.
I suggest to mimic VSCode, you can see on bottom right my keyboard inputs.

Space don't discard autocomplete on GD:

https://user-images.githubusercontent.com/1670670/132092240-7de2c676-3a5f-4f3a-bf9c-c86feaa14c67.mp4

Text input position is not inside the parenthesis by default + Autocomplete is still present at the end.
Can we add an auto close for parenthesis by default for avoid to make errors?

https://user-images.githubusercontent.com/1670670/132092308-48a783a7-7b88-40b8-8753-4b2c863a3b62.mp4

Autocomplete pop when operator are trigger + Autocomplete still present at the end

https://user-images.githubusercontent.com/1670670/132092459-6ad09a15-0b73-49f6-8e17-6a791287ced3.mp4

However we're autocomplete pop, this is the more anoying things:

https://user-images.githubusercontent.com/1670670/132092594-2d3ec4b0-7c35-49ef-b89e-7d69bb42b4ae.mp4

We can't use enter for return carriage:

https://user-images.githubusercontent.com/1670670/132092624-a114a462-156e-42ca-a891-9b4558ad587a.mp4

Most of time I disable the autocomplete  because of these things. The navigation is bad, despite the powerful autocomplete.

ClementPasteau:
Great videos @Bouh - I have to admit, this is still one of the things that I find annoying when I use the app too 🙈 
That would be great to perfect it for the user experience.

Technically, there are multiple autocomplete fields across the app, but I think they are all based on the `src/UI/SemiControlledAutoComplete` component (which we can test easily with Storybook)

PR #3167, opened 2021-10-11
Improve Expression autocomplete

ClementPasteau:
Trying to address the points raised in #3006 

**Note that some error cases displayed below should not happen really often as it's mainly when the autocomplete is not used and texts are typed completely manually, the common use case is to rely on the autocomplete (Enter/Tab) and is quite intuitive**

Different points addressed:

## Adding both parentheses at the end of the autocomplete option, with the caret in the right position depending on the parameters.

Case where the expression doesn't have parameters -> both parentheses added and caret at the end
image

Case where the expression has parameters -> both parentheses added and caret inside them
image

## Not displaying the exact matches in the autocomplete options (as we assume the user is already sure about it) + not registering the submit "Enter" key, as the option is already selected -> goes to new line

When typing, options are shown
image

Once typed fully, the option disappear, and we can press Enter to go to new line
image

If autocomplete is used without typing fully the text
image

## Not displaying all the options available when using the operator "+"

After typing a "+", no options displayed
image

After typing a letter, options appear
image

## Not displaying options when adding a space at the end of the text entered

Adding a space after an incomplete text
image

Adding a space after an object text
image

## Show Expression description

Description right of the title
image

No description if not an expression
image

### Video showing those cases

https://user-images.githubusercontent.com/4895034/136820403-7e31eeed-97c4-4671-8309-804c27eed69c.mp4

### DIFF
--- a/newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/ExpressionAutocompletionsDisplayer.js
+++ b/newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/ExpressionAutocompletionsDisplayer.js
@@ -360,7 +360,8 @@ export default function ExpressionAutocompletionsDisplayer({
           </Paper>
           {selectedCompletionIndex !== null &&
             expressionAutocompletions[selectedCompletionIndex].kind ===
-              'Expression' && (
+              'Expression' &&
+            !expressionAutocompletions[selectedCompletionIndex].isExact && (
               <Paper variant="outlined" square style={styles.container}>
                 <ScrollView autoHideScrollbar>
                   <Column>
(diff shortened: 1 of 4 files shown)


ClementPasteau:
Thanks for checking @Bouh !

For your first point, this is quite a tricky one...
If we want to avoid that, it means we have to differentiate 2 states in the app:

- When the user is typing -> the caret is in the middle of possible autocomplete options -> suggest them
- When the user has stopped typing and is just navigating in the input -> do not suggest anything -> and maybe have an option like in VSCode to pop the suggestions again with ctrl+space

Quite some work to have something like that, I'm not sure about the technical strategy to get there 🤔 

For your second point, it is quite painful indeed.
It seems to be coming from the `ExpressionParser` which is removing spaces at the end of the searched string to create the ExpressionNode, and thus searching on the trimmed value. I'm investigating further to see if we can avoid that without causing bugs.

arthuro555:
Does this also fix the autocomplete not coming up when modifying a function call?
UP9JeB0tLB

ClementPasteau:
Hey @arthuro555, I've implemented a fix in #3174 that should help with the problem you're having :)
I'll let you test that in the next release!

Issue #1463, opened 2020-02-23
Ghost collision mask


PascalLadalle:
From user 'vinurd' on the [forum](https://forum.gdevelop-app.com/t/the-collision-of-the-platform-object-is-not-removed-although-the-object-has-disappeared/22836):
![image](https://user-images.githubusercontent.com/25978830/75119294-bf641a80-5681-11ea-88bf-ad71dcdbac12.png)
When this specific block is deleted, the hitbox remains, even though the instance is deleted (confirmed in debugger). Instance UID is 114. 
Testing collision Player-rock shows that Player is not in collision with rock when it is standing on this "ghost".
![image](https://user-images.githubusercontent.com/25978830/75119362-9abc7280-5682-11ea-9a81-30a42d75f452.png)

Nothing special in the json for that object:
![image](https://user-images.githubusercontent.com/25978830/75119309-f0dce600-5681-11ea-9c41-9d2859a05561.png)

Project link:
https://dropmefiles.com/ZXHrj

I tried a few things, including removing all Cyrillic letters from the project, no luck.
vinurd said that if he deletes that rock, the bug is moved to another one.

4ian:
Can you explain the exact steps to reproduce the issue? I've not being able to have the issue, getting this Rock destroyed does not produce a ghost for me.

(if possible when posting a project like this, specify the scene name, and what to do, this greatly improve the time for me to inspect these bugs. Thanks!)

Bouh:
I've spend 2 hours on this and i've found nothing.
The eventsheet should be simplify, remove useless events not related to punch, rocks and player.
Same for objects on scene. Keep the minimum.

PascalLadalle:
I didn't need to do anything special to trigger the bug, just moved the player straight to that rock, clicked several times to punch/delete it and jumped on the ghost. 
Strange that the bug doesn't occur with you. 🤔 
Here's a video from vinurd: https://www.youtube.com/watch?time_continue=60&v=nyoAlImRDWE&feature=emb_logo

I agree that the events aren't very clean, but the instance is deleted, so the collision should be impossible...right?

Bouh:
I had the problem, I spent two hours trying to figure out why he was there.
The user should delete anything that is not useful for the bug.
I no longer debug complete projects there is too much possible interference. And it's a waste of time searching through all the events and scenes.

zatsme:
Long time no see 😃

I just had to try this as I remembered from last year, maybe longer, the same issue from before.

[...]

PR #3175, opened 2021-10-19
Use wrapper class to store behavior and its bounding box in RTree

### LINKED ISSUE #2986: Light obstacles persist when the object is deleted
## Describe the bug
@HarsimranVirk 
Light obstacles persist when the object is deleted too fast.

## To Reproduce
[Project example.zip](https://github.com/4ian/GDevelop/files/7070939/TrableShadow.zip)

In the project example, drag&drop fast an object in collision with another one for deleting them, the Light obstacle isn't deleted.
(Might be a failure of the light update function?)
In the first part of the video everything works, in the second (+20sc) the last cube is deleted correctly but the light obstacle is still present.

https://user-images.githubusercontent.com/1670670/131223119-2ca043ac-7852-41e6-ab85-6a09b2c1dba4.mp4

## Other details
* B115 and all builds before 26 May, 2021
* [Original forum bug report](https://forum.gdevelop-app.com/t/problem-with-removing-the-light/32910)

AlexandreSi:
Resolves #2986, and maybe:
- #1463

### DIFF
--- a/Extensions/Lighting/lightobstacleruntimebehavior.ts
+++ b/Extensions/Lighting/lightobstacleruntimebehavior.ts
@@ -30,13 +30,10 @@ namespace gdjs {
      * Add a light obstacle to the list of existing obstacles.
      */
     addObstacle(obstacle: gdjs.LightObstacleRuntimeBehavior) {
-      if (obstacle.currentBehaviorAABBHolder)
-        obstacle.currentBehaviorAABBHolder.updateAABBFromOwner();
-      else
-        obstacle.currentBehaviorAABBHolder = new gdjs.BehaviorAABBHolder(
-          obstacle
-        );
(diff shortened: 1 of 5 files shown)


D8H:
> @D8H we have two tests failing for the platformer when doing this change. It seems that the platformer object is following the floor even when it's moved away (platform going from 0;0 to -100;-100, which is ok) but also when going down too fast.
> 
> This is strange because I fail to see how these changes would how the floor is followed 🤔

The tests are on moving platforms so the rbush is modified.

> Question for you: if you look at the failing tests here: https://gdevelop.semaphoreci.com/jobs/8630c8f8-54f3-4f73-b1cd-d66d22491b10, do you remember if in your PR doing the changes to the platformer, these tests are working/were changed?

I will do a PR to update the test suite. It will give a diff and it could be interesting to run the new test on this change just in case.

PR #3210, opened 2021-10-28
Fix a glitch where we hear a sound being played at a high volume for a split second

### LINKED ISSUE #3054: Sound playback can glitch when played back at lower volume levels
## Describe the bug
This appears to be a resurfacing of https://github.com/4ian/GDevelop/issues/2490

Omar on the forums reported that they were getting weird audio popping when sound files were played rapidly in their project. https://forum.gdevelop-app.com/t/sound-bug-since-version-107-till-117/35041/

I created an example project to try and test this out, but could not reproduce the issue.

Others were able to reproduce the issue with their project, and the impacted user was able to reproduce the issue with my project.

My best guess is that something in the audio code is getting overloaded at some point, because the sound files have to be played _very_ rapidly for about 20-30 seconds before it occurs.

Here's a video of it occurring with their example project. You can see the audio level blip near the end. 
https://streamable.com/hameau

## To Reproduce
1. Load their example project.
2. Run the preview
3. Rapidly press the 1 and 2 keys for 20-30 seconds.

Here is Omar's example project (I cleaned up some bad resource paths and other issues)
[SoundBugExample.zip](https://github.com/4ian/GDevelop/files/7179921/SoundBugExample.zip)

I've also included a copy of my example project.
[Audiotest.zip](https://github.com/4ian/GDevelop/files/7179924/Audiotest.zip)

I have also tested their project by having proper preloading via resources rather than via events.

## Other details
* Windows 10 Pro 64 bit
* B117

ClementPasteau:
This should fix #3054 

Thanks to the project shared in the issue, I was able to reproduce the bug really consistently, and without having to smash the button (almost 1 in 5 key press)
I tried analyzing the timings to see if it was an issue where the volume was set too late compared to when it started to play, but I couldn't find anything consistent.

As discussed, the `play()` method needs to happen before changing anything on the sound, otherwise it is not registered.
So I've moved all the parameters change (volume, loop & pitch) just after this method, instead of waiting for a 'play' event, which can happen 1 frame too late and trigger this weird sound.

This doesn't fix the issue by itself.
In order to avoid hearing a split-second sound at 100% volume, I changed the cached sounds to have a volume at 0 instead of 100. This ensures that if the sound/music starts playing before we have time to change the volume, then we don't hear it.

Here is an example project I've used to test: [Audiotest.zip](https://github.com/4ian/GDevelop/files/7498323/Audiotest.zip)
 

I've also added it to the tests in the code.

### DIFF
--- a/GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
+++ b/GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
@@ -469,9 +469,9 @@ namespace gdjs {
     createHowlerSound(
       soundName: string,
       isMusic: boolean,
-      volume,
-      loop,
-      rate
+      volume: float,
+      loop: boolean,
+      rate: float
     ): HowlerSound {
       const soundFile = this._getFileFromSoundName(soundName);
       const cacheContainer = isMusic ? this._loadedMusics : this._loadedSounds;


ClementPasteau:
I've attempted a new fix.
In all cases (playSound, playSoundOnChannel, playMusic, playMusicOnChannel) -> I call the `.play()` method as soon as possible. Which is as soon as the `HowlerSound()` is created (from an existing `Howl()` or not).
Then I set the parameters volume/pitch/loop.
Then we add the listener on the 'play' event.

I've done various tests:
- play same sound with different volumes on different channels, at different speeds (with key press)
- play it on a channel and also not on a channel

and it seems to have removed the problem. Or maybe because we're not waiting for the 'play' event anymore, it's happening so fast that even if the problem is here, we can't hear it.

Is there something I could be still missing? @arthuro555

arthuro555 (approved):
Looks good to me! Try to also move the `once`s before the `.play()` so that it goes through the `HowlerSound` callbacks array which gets called instantaneously instead of passing through howler events listeners which supposedly rely on the js events queue.

GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
```diff
@@ -560,42 +562,41 @@ namespace gdjs {
...
           sound.pause();
           this._pausedSounds.push(sound);
         }
       });
+      // We need to first play the sound in order to change its parameters like volume.
+      sound.play();
+      sound
+        .setLoop(loop)
+        .setVolume(volume / 100)
+        .setRate(pitch);
```
4ian, line 577: @arthuro555 @ClementPasteau Correct me if I'm wrong, but if the sound is not loaded in memory, then:

- createHowlerSound is called (no problem). It gives you back a gdjs.HowlerSound (based on a Howl that is created).
- We then call `play` on gdjs.HowlerSound. Because it is not `isLoaded` yet, then the `this._id` is not set. Only `this._howl.once('load', () => this.play())` is called.
- We then call `setLoop`/`setVolume`/`setRate`. Problem here: because the sound is not loaded, the `this._id` is null. So the volume is not applied (and same for the pitch/loop).

Conclusion: the first time the sound is loading, we risk running a sound without applying the proper settings?

I could not notice this while testing, but this seems risky - especially for long musics that are long to load if not preloaded (if we skip these, we now have a background music not at the right volume and not looping. 😬)

4ian: I wonder if the solution is to rework the setters (currently they don't do anything if the sound is not loaded => because of this we can have invisible bugs) so that they store in the gdjs.HowlerSound the required volume/pitch/loop state, and apply it directly if they can. And if they can't it's ok, because the required volume/pitch/loop state can be applied in `play()` (i.e: as soon as you get the id => you set the volume/pitch/loop, because the source of truth is now the gdjs.HowlerSound).

ClementPasteau: You're right for your first comment. I haven't noticed any issue either, but it could be because we only have short sounds.
As you suggested, it could be worth creating an audio playground project, with all types and sizes of sounds & musics playing on different channels so we can experiment.
I'll try to do that next so that we feel more confident pushing a fix

arthuro555: Oh right that's the reason i had put it in the once in the first place. Indeed storing the settings and applying them on play sounds like the optimal way, it would also make the JavaScript api less "buggy" (as in these potential issues wouldn't be triggerable by misusing the API either).

ClementPasteau:
Alright new attempt. I've done 2 things:

- stored the properties volume/rate/loop in the HowlerSound object, so that we can either:
  - apply the props directly if the sound is already playing (in the `setVolume` for instance)
  - or use the stored props in the `play()` as soon as the sound is loaded and the ID is generated after the howl.play() method
  - also return those stored values in the getters (and avoid relying on the howl)

This didn't solve the glitch by itself though. I've also:
- Changed the cached sounds so that their volume is at 0 by default, instead of 1.0 (I checked Howl doc) This allow having the opposite glitch effect => the sound may not be heard for a split second, but I couldn't notice it. It's wayyyy less impacting than at 100%!

GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
```diff
@@ -132,37 +154,35 @@ namespace gdjs {
...
-     * @returns The current instance for chaining.
      */
-    setRate(rate: float): this {
+    setRate(rate: float): void {
+      this._rate = rate;
+      // If the sound has already started playing, then change the value directly.
       if (this._id !== null) {
         rate = gdjs.HowlerSoundManager.clampRate(rate);
         this._howl.rate(rate, this._id);
       }
```
4ian, line 169: It's important to keep the return this because this allow to chain calls ("fluent" interface) - this would be a breaking change to change this.

GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
```diff
@@ -560,42 +602,45 @@ namespace gdjs {
...
-        music
-          .setLoop(loop)
-          .setVolume(volume / 100)
-          .setRate(pitch);
         if (this._paused) {
           music.pause();
           this._pausedSounds.push(music);
         }
       });
+      // We need to first play the music in order to change its parameters like volume.
```
4ian, line 641: Are these comments useful? I don't see why we're telling this because we're not changing any parameter?

GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
```diff
@@ -654,6 +698,25 @@ namespace gdjs {
       this._pausedSounds.length = 0;
     }
 
+    preloadAudioFile(
```
4ian, line 701: Is there a reason for moving this to the public interface? We now have a risk users will call and rely on this function - which we probably don't want to?

GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
```diff
@@ -68,6 +86,10 @@ namespace gdjs {
           this._id === null ? '__default' : this._id
         );
         this._id = newID;
+        // Set the howl properties as soon as the sound is played and we have its ID.
+        this._howl.volume(this._volume, newID);
+        this._howl.loop(this._loop, newID);
+        this._howl.rate(this._rate, newID);
```
4ian, line 92: Use gdjs.HowlerSoundManager.clampRate on the rate you're passing to this._howl.rate

GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
```diff
@@ -37,6 +37,21 @@ namespace gdjs {
      */
     private _howl: Howl;
 
+    /**
+     * The volume at which the sound is being played.
+     */
+    private _volume: integer;
+
+    /**
+     * Wether the sound is being played in a loop or not.
```
4ian, line 46: *Whether

GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
ClementPasteau: As I removed the calls to theses functions from the file itself, I though it wasn't useful anymore. I'll add it back

GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
ClementPasteau: nope, they're outdated

GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
ClementPasteau: Probably a misunderstanding from my side. I didn't think making it outside would make it available to the users, and I felt it was cleaner this way. My bad!

4ian: No worries! 
In not performance sensitive code it's better to keep things either inside the function or outside the whole class if it's a pure helper function - in both cases it's not accessible by the outside world (so it's even better than putting an underscore to say "this is private don't call me").

PR #3217, opened 2021-10-30
Updated wording of particle types (ParticleEmitterEditor.js)

tristanbob:
Updated wording of particle types.

### DIFF
--- a/newIDE/app/src/ObjectEditor/Editors/ParticleEmitterEditor.js
+++ b/newIDE/app/src/ObjectEditor/Editors/ParticleEmitterEditor.js
@@ -44,7 +44,7 @@ export default class ParticleEmitterEditor extends React.Component<
         ))}
         <SelectField
           fullWidth
-          floatingLabelText={<Trans>Particles kind</Trans>}
+          floatingLabelText={<Trans>Particle type</Trans>}
           value={particleEmitterObject.getRendererType()}
           onChange={(e, i, value: string) => {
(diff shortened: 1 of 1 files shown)


arthuro555 (approved):
👍

newIDE/app/src/ObjectEditor/Editors/ParticleEmitterEditor.js
```diff
@@ -57,15 +57,15 @@ export default class ParticleEmitterEditor extends React.Component<
         >
           <SelectOption
             value={gd.ParticleEmitterObject.Point}
-            primaryText={t`Point`}
+            primaryText={t`Circle`}
           />
           <SelectOption
             value={gd.ParticleEmitterObject.Line}
-            primaryText={t`Line`}
+            primaryText={t`Rectangle`}
```
Bouh, line 64: Humm no, all settings below this select are for a line, like **length** and **thickness**.
If we rename it rectangle you have to edit the parameters below for width and height, I recommend going back to **line**!

4ian: This is very true, I was a bit too fast. I'm changing this back to line. 
Thanks for your careful review!

4ian: Done in aeecb0e29f5a14f3e199a99268c57bd519cff5ce

PR #3218, opened 2021-10-31
Tween refactor

arthuro555:
A refactor of the tweens to have less duplicate code, more types, and better readability.
Also updates shifty js to 2.16.0.


Silver-Streak:
@arthuro555 a long while ago, we talked about adding tween options to the camera. (Tween camera position and zoom) so people didn't need to keep adding speciap objects for it. Any chance of that getting added to this PR? 

No worries it not, but thought I'd check.

Bouh:
Don't merge it without comparing it with the test project 😉

arthuro555:
My plan was originally to add tweens that do not require the behavior (variables, camera...) and an interpolation expression with eatings, but i found the code to be messy so i decided to do this cleanup first.

> Don't merge it without comparing it with the test project 😉

Already did before submitting the PR, i did not spot anything put of the ordinary

4ian:
Thanks for the PR! So for now this PR should not result in any change/new stuff at all being added right? This is just refactoring (which is good and better to have a separate PR just for this)?

arthuro555:
Exactly :)

Extensions/TweenBehavior/tweenruntimebehavior.ts
```diff
@@ -1,7 +1,50 @@
+/// <reference path="shifty.d.ts" />
+const logger = new gdjs.Logger('Tween behavior');
```
4ian, line 2: If I'm not mistaken, this will create the logger in the global namespace (i.e: global variable), you should put it inside `namespace gdjs {`

Extensions/TweenBehavior/tweenruntimebehavior.ts
```diff
@@ -1,7 +1,50 @@
...
 namespace gdjs {
-  declare var shifty: any;
-  namespace shifty {
-    export type Tweenable = any;
+  export interface RuntimeScene {
```
4ian, line 4: Would this work with other extensions doing the same? Do you have a link to how TypeScript is handling this?

Extensions/TweenBehavior/tweenruntimebehavior.ts
```diff
@@ -58,119 +100,102 @@ namespace gdjs {
...
+        tweenable,
         false,
         startTime,
         totalDuration
       );
-    }
 
-    _getTween(identifier) {
-      return this._tweens[identifier];
+      // Attach it to the scene as it will become active and acttive tweens should be on the scene.
```
4ian, line 134: acttive => active

Extensions/TweenBehavior/tweenruntimebehavior.ts
```diff
@@ -1283,39 +1046,29 @@ namespace gdjs {
...
-    if (!shiftyJsScene) {
-      return;
-    }
+    if (!shiftyJsScene) return;
 
     // Stop and explictly remove all tweenables to be sure to drop
     // all references to the tweenables of the scene.
     shiftyJsScene.stop(false);
-    shiftyJsScene.tweenables.forEach(shiftyJsScene.remove.bind(shiftyJsScene));
+    shiftyJsScene.empty();
```
4ian, line 1055: 👍

Extensions/TweenBehavior/tweenruntimebehavior.ts
```diff
@@ -58,119 +100,102 @@ namespace gdjs {
...
+      tweenable
+        .tween(tweenConfig)
+        .then(() => {
+          if (this._tweens[identifier])
+            this._tweens[identifier].hasFinished = true;
+
+          if (destroyObjectWhenFinished)
+            this.owner.deleteFromScene(this._runtimeScene);
+        })
+        .catch((e) => {
```
4ian, line 149: Shifty was at least in the past throwing errors when you attempted to stop a tween and restart it if it was not finished (despite not being an "error"/"exception" by itself - it's a valid use case).
Is this still the case? If so we should not log anything.

Extensions/TweenBehavior/tweenruntimebehavior.ts
```diff
@@ -58,119 +100,102 @@ namespace gdjs {
...
+      this._tweens[identifier].instance.stop(jumpToDest);
     }
 
-    _tweenHasFinished(identifier) {
+    private _tweenHasFinished(identifier: string): boolean {
       return this._tweens[identifier].hasFinished;
     }
 
-    _removeObjectFromScene(identifier) {
-      this._removeTween(identifier);
```
4ian, line 128: Weird that we were calling _removeTween - not sure what was the reason or if this was useful 🤔

4ian (commented):
Very nice refactoring! ✨
See comments for a few points where we must be careful (I know that shifty handling of promises was a bit special, at least in the past).

Extensions/TweenBehavior/tweenruntimebehavior.ts
arthuro555: Whoops 😬

Extensions/TweenBehavior/tweenruntimebehavior.ts
arthuro555: Yes, it would. It is basically [declarations mergining](https://www.typescriptlang.org/docs/handbook/declaration-merging.html). As you can see on [TypeScript AST viewer](https://ts-ast-viewer.com/#code/MYGwhgzhAEBKDK0DeAoa0AOAnAlgNzABcBTaAfTAC5oA7AVwFsAjYraAXmgAYBuNaYAHsaEQljrBCgrAAoqtRiywBKZNAC+-MDOLVRuGgHNVqdOizFCdLDWhi6xfps0ocNElgBmYYKQTJ+Jmp6ZlYUFzcPb184RFNoMAALeR1qBFV2AD5oPEEcABNwlCA), this gets merged with the class type:
image

4ian: Ok! Resolving this comment then :)

Extensions/TweenBehavior/tweenruntimebehavior.ts
Wend1go: I'm not sure myself after all this time. ^^
I think the tween wasn't deleted automatically when it finished.

arthuro555: Well if the object is removed the tween not being removed shouldn't matter, right?

4ian: I agree this shoud not matter - just asking to see if there was not a hidden issue. I think it's safe to remove this line!

Extensions/TweenBehavior/tweenruntimebehavior.ts
4ian: @arthuro555 We'll need to be sure about this before going ahead with the PR :)

arthuro555: Yes, to need to test this

arthuro555: Neither that: 
image
Nor that (whether the tween was playing before or not):
image

Logged any error in the console, I am assuming that this behavior has since changed in shifty.

4ian: @jeremyckahn hope it's ok to quickly ping you to have your blessing on this: we're making the assumption that tweens won't "throw" in Shifty - is this ok to do?

I see in Shifty code that the promise can be rejected, but only in the case where we actually defined a catch/reject handler:
image

Are there other cases where this can happen?
- If yes, then let's probably not log anything (we don't want to spam logs in a game, decreasing performance) if this can happen as part of the normal operation of a game.
- If not, then let's entirely remove the catch.

4ian:
Alright, I think we're good to go but I pinged the Shifty author to get confirmation about the catch for the promises - I really don't want to make an emergency bugfix release on a week-end because we found too late that some games are having the promises rejected and we're putting tons of logs through the console :)

Extensions/TweenBehavior/tweenruntimebehavior.ts
jeremyckahn: Hi @4ian, thanks for the ping! It's no trouble at all.

The only scenario where I expect a Shifty tween to throw an error is when its `.then` function is called (which will happen implicitly if it is being `await`ed) and it is also `cancel`led:

https://github.com/jeremyckahn/shifty/blob/199774dc8e03b7de1a1971997193e692a5626fe1/src/tweenable.js#L452-L459

https://github.com/jeremyckahn/shifty/blob/199774dc8e03b7de1a1971997193e692a5626fe1/src/tweenable.js#L635-L654

I haven't thoroughly reviewed this PR, but it does seem that there is at least one place where GDevelop is calling a `Tweenable`'s `.then` function: https://github.com/4ian/GDevelop/pull/3218/files#diff-90447101490944a2b12792b193b1f177c27b721e583d175e6eec438a89f232f9R142-R149

FWIW, I believe using Shifty's Promise API is good practice, so that's fine to leave in. However, because of that I think GDevelop should also anticipate the rejection caused by a potential future `cancel` call. It seems to me that the `catch` we're talking about should remain in case a `cancel` call is ever added.

I hope this is helpful. I'm excited to see GDevelop use the latest version of Shifty! Let me know if I can answer any other questions. 🙂

Extensions/TweenBehavior/shifty.d.ts
```diff
@@ -0,0 +1,571 @@
+// Shifty.js 2.16.0 type definitions by arthuro555
```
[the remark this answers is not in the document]
arthuro555: I had actually begun a PR to autogenerate typings from the jsdoc comments 👀 Since it requires changing those comments quite a lot though, and for some reason i cannot get jsdoc to work, I was afraid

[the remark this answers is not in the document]
jeremyckahn: Ah, I understand. There might be a few pitfalls in there, but I'm sure we can get them resolved. I'd be happy to help out where I can, so just let me know what I can do. Even just getting a PR going with what you've started on will help the process along quite a bit! 🙂

Extensions/TweenBehavior/tweenruntimebehavior.ts
4ian: > However, because of that I think GDevelop should also anticipate the rejection caused by a potential future cancel call. It seems to me that the catch we're talking about should remain in case a cancel call is ever added.

Sounds good!

@arthuro555 Can we remove the log in the catch, because it could in the future be called for a cancelled tween, which is not an "error" by itself? Then we're good to go!

4ian (approved):
Let's remove the logging in case of a rejected tween promise, so that if in the future we add a way to cancel a tween we don't spam the console with false "errors". :)

4ian (approved):

4ian:
Seems all good - I hope we broke nothing but hopefully everything is covered and refactored so that it's easier to work on this later :)

PR #3220, opened 2021-10-31
Update esbuild

arthuro555:
Updates esbuild to the latest version and enable code minification (won't impact debugging since we have source maps).
Tested by running the tweens example, nothing seems to have broken (and if we trust esbuild to not be broken, it shouldn't).


arthuro555:
Note that there are breaking changes but most shouldn't affect us as they either affect ES modules, imports, or CSS which we do not use. 

The only notable change is the [change of the programmatic API (which is accounted by this PR)](https://github.com/evanw/esbuild/blob/master/CHANGELOG.md#090).

Note that there is an issue with semaphore: I am using npm@7, and the package-lock.json is therefore in v2. Since it expects a v1 one, it'll regenerate it to a v1 one, modifying the cache key and preventing getting the previously installed node_modules folder.

4ian:
Thanks! 👍
Any chance you can keep using the v1 of the lockfile for a bit? :)

arthuro555:
Sure, I just thought it may be a problem that semaphore breaks if the installation phase modifies the package-lock,jaon

4ian:
Well it seems it actually showed an issue :) The installation phase did modify the package-lock.json and that should not happen (you should always be able to install without changes to this file). 
This is because semaphore is still using npm 6. We can upgrade it later, but it also means that real users, like me, using npm 6 will still have changes made on their local working copy. We should wait for npm 6 to almost disappear before updating this then (i.e: wait a bit before basically forcing users to upgrade) :)

PR #3221, opened 2021-11-01
Fix contributor list

Midhil457:
- Fix wrong links
- Refactor some descriptions
- Add contributors


4ian (approved):

4ian:
Thank you! I've reverted a comment style change because `//` is preferable in most cases, even for multiline comments, because it's easier to re-order line/copy part of the comment/it's more symmetric. We use only `/* */` for JSDoc comments or something inside a line.

Issue #3186, opened 2021-10-22
Endless loop of inclusion when including events into the same scene


Oxey405:
## Describe the bug
Hello ! There is a bug when previewing and compiling that makes an endless loop of events inclusion.
Basicly, you can include events of scene A into scene A so when compliing GDevelop tries to include scene A into scene A wich include scene A into itself etc... (a little bit like inception)

## To Reproduce
1. Create a scene (let's call it "sceneA")
2. click on "Choose and add an event" in the top bar.
3. Choose to include events from sceneA
4. Click on Preview
5. Watch as GDevelop gets into an endless loop of including scene into itself
## Screenshot
![screenshot of the way to do the bug](https://user-images.githubusercontent.com/61246500/138493239-9abf18ce-79a8-452f-b655-d80ca002b675.png)

## Other details
* WIndows 10
* Desktop v. 5.120
* Easy to fix issue

#3228

AlexandreSi:
Closing this issue as you seem to have fixed the issue yourself! Congrats!

PR #3228, opened 2021-11-05
prevented a bug of endless include loop that could have broken the compiler

Oxey405:
# Introduction
Hi ! I have **prevented** a bug that ~~was not in the issues~~ was reported by me #3186 but that I discovered and that could have broken the GDevelop compiler.
# Explaining the bug
You could i**nclude the events of a scene into itself** and creating a **endless loop** of including events wich **lock gdevelop when previewing** the game. Here is an example :
example image
_("Inclure les évenements de [...]" =  "Include events from [...]")_
 # How I prevented it
I **modified the renderer for the event link system** so it will not let users do this bug.
Maybe we should also modify the core of the engine to disallow a selfincluding scene but for now it's a good start.

### DIFF
--- a/newIDE/app/src/EventsSheet/EventsTree/Renderers/LinkEvent/ExternalEventsAutoComplete.js
+++ b/newIDE/app/src/EventsSheet/EventsTree/Renderers/LinkEvent/ExternalEventsAutoComplete.js
@@ -9,8 +9,11 @@ import SemiControlledAutoComplete, {
   type DataSource,
 } from '../../../../UI/SemiControlledAutoComplete';
 
-const getList = (currentSceneName: string, project: ?gdProject): DataSource => {
-  if (!project) {
+const getList = (
+  currentSceneName: ?string,
+  project: ?gdProject
+): DataSource => {
+  if (!project || !currentSceneName) {
     return [];
   }
 
@@ -36,7 +39,7 @@ type Props = {|
   isInline?: boolean,
   onRequestClose?: () => void,
   onApply?: () => void,
-  sceneName: string,
+  sceneName?: string,
 |};
 
 export default class ExternalEventsAutoComplete extends React.Component<
--- a/newIDE/app/src/EventsSheet/EventsTree/Renderers/LinkEvent/ExternalEventsAutoComplete.js
+++ b/newIDE/app/src/EventsSheet/EventsTree/Renderers/LinkEvent/ExternalEventsAutoComplete.js
@@ -27,7 +27,7 @@ const getList = (
     }));
 
   const layouts = enumerateLayouts(project)
-    .filter(layout => layout.getName() !== currentSceneName || externalEvents)
+    .filter(layout => layout.getName() !== currentSceneName)
     .map(layout => ({
       text: layout.getName(),
       value: layout.getName(),
--- a/newIDE/app/src/EventsSheet/EventsTree/Renderers/LinkEvent/ExternalEventsAutoComplete.js
+++ b/newIDE/app/src/EventsSheet/EventsTree/Renderers/LinkEvent/ExternalEventsAutoComplete.js
@@ -8,7 +8,7 @@ import SemiControlledAutoComplete, {
   type DataSource,
 } from '../../../../UI/SemiControlledAutoComplete';
 
-const getList = (project: ?gdProject): DataSource => {
+const getList = (currentSceneName: string, project: ?gdProject): DataSource => {
   if (!project) {
     return [];
   }
@@ -19,10 +19,12 @@ const getList = (project: ?gdProject): DataSource => {
       value: externalEvents.getName(),
     })
   );
-  const layouts = enumerateLayouts(project).map(layout => ({
-    text: layout.getName(),
-    value: layout.getName(),
-  }));
+  const layouts = enumerateLayouts(project)
+    .filter(layout => layout.getName() !== currentSceneName)
+    .map(layout => ({
+      text: layout.getName(),
+      value: layout.getName(),
+    }));
   return [...externalEvents, { type: 'separator' }, ...layouts];
 };
 
@@ -53,6 +55,7 @@ export default class ExternalEventsAutoComplete extends React.Component<
       onApply,
       isInline,
       project,
+      sceneName,
     } = this.props;
 
     return (
@@ -64,7 +67,7 @@ export default class ExternalEventsAutoComplete extends React.Component<
         onChange={onChange}
         onRequestClose={onRequestClose}
         onApply={onApply}
-        dataSource={getList(project)}
+        dataSource={getList(sceneName, project)}
         openOnFocus={!isInline}
         ref={field => (this._field = field)}
       />
--- a/newIDE/app/src/EventsSheet/EventsTree/Renderers/LinkEvent/ExternalEventsAutoComplete.js
+++ b/newIDE/app/src/EventsSheet/EventsTree/Renderers/LinkEvent/ExternalEventsAutoComplete.js
@@ -1,5 +1,4 @@
 // @flow
-import { string } from 'prop-types';
 import * as React from 'react';
 import {
   enumerateLayouts,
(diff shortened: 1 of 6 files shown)


Bouh:
https://github.com/4ian/GDevelop/issues/3186

Oxey405:
> #3186
Oh yes, I forgot

newIDE/app/src/EventsSheet/EventsTree/Renderers/LinkEvent/index.js
```diff
@@ -175,6 +175,7 @@ export default class LinkEvent extends React.Component<EventRendererProps, *> {
               <ExternalEventsAutoComplete
                 project={this.props.project}
                 value={target}
+                sceneName={this.props.scope.layout.getName()}
```
arthuro555, line 178: The layout is not guaranteed to be present (for example in events based functions)

```suggestion
                sceneName={this.props.scope.layout ? this.props.scope.layout.getName() : ""}
```

arthuro555: Remember to add the prop to the flow type as well:

https://github.com/4ian/GDevelop/pull/3228/files#diff-5213df436229c337cb945f865908464ae1d144072d017e97da8f042a327bfc12R31-R38

Oxey405:
Hi arthuro ! I have commited your suggestion !

newIDE/app/src/EventsSheet/EventsTree/Renderers/LinkEvent/ExternalEventsAutoComplete.js
```diff
@@ -8,8 +9,11 @@ import SemiControlledAutoComplete, {
   type DataSource,
 } from '../../../../UI/SemiControlledAutoComplete';
 
-const getList = (project: ?gdProject): DataSource => {
-  if (!project) {
+const getList = (
+  currentSceneName: ?string,
+  project: ?gdProject
+): DataSource => {
+  if (!project || !currentSceneName) {
```
4ian, line 16: I don't think we should return an empty array if the currentSceneName is null/undefined/falsy? 
In events based function, the currentSceneName will be undefined, and still we should autocomplete with the layouts and external events?

newIDE/app/src/EventsSheet/EventsTree/Renderers/LinkEvent/ExternalEventsAutoComplete.js
```diff
@@ -1,4 +1,5 @@
 // @flow
+import { string } from 'prop-types';
```
4ian, line 2: Remove this line it's useless

arthuro555:
> In events based function, the currentSceneName will be undefined, and still we should autocomplete with the layouts and external events?

Iirc the code generation of link events exits out early when there is no layout, so no layout or external events are available if there is no layout and therefore that behavior is i think correct

newIDE/app/src/EventsSheet/EventsTree/Renderers/LinkEvent/ExternalEventsAutoComplete.js
Oxey405: I'm gonna do it

newIDE/app/src/EventsSheet/EventsTree/Renderers/LinkEvent/ExternalEventsAutoComplete.js
Oxey405: returning an empty array is not a problem i think

4ian: Can you give an example showing that this is ok? :)

Oxey405: Hey ! When there is no current layout, you are inside an extension, and there link events do not work anyways ([see here](https://github.com/4ian/GDevelop/blob/36fb4ec9b2cf5529ddf621be8a0a0e84f8572f37/GDJS/GDJS/Extensions/Builtin/CommonInstructionsExtension.cpp#L54)), so no autocompletions should be shown

4ian: Ok thanks!

newIDE/app/src/EventsSheet/EventsTree/Renderers/LinkEvent/ExternalEventsAutoComplete.js
4ian: @Oxey405 you still need to do this :)

Oxey405: Yeah I know I'm late and I have school so you know, my calendar is booked up really quickly, I'm gonna do it tomorrow I promise  ! :)

Oxey405: @4ian done ! I hope you find no more problems with the PR

newIDE/app/src/EventsSheet/EventsTree/Renderers/LinkEvent/ExternalEventsAutoComplete.js
```diff
@@ -8,21 +8,31 @@ import SemiControlledAutoComplete, {
...
-    })
-  );
-  const layouts = enumerateLayouts(project).map(layout => ({
-    text: layout.getName(),
-    value: layout.getName(),
-  }));
+    }));
+
+  const layouts = enumerateLayouts(project)
+    .filter(layout => layout.getName() !== currentSceneName || externalEvents)
```
4ian, line 30: Why this `|| externalEvents`? I don't think this is doing what you think?

newIDE/app/src/EventsSheet/EventsTree/Renderers/LinkEvent/index.js
```diff
@@ -156,7 +156,7 @@ export default class LinkEvent extends React.Component<EventRendererProps, *> {
                 tabIndex={0}
               >
                 {target || (
-                  <Trans>{`<Enter the name of external events>`}</Trans>
+                  <Trans>{`<Enter the name of external events/scene>`}</Trans>
```
4ian, line 159: Let's revert this, we don't want to encourage including scene events.

Oxey405: ok

newIDE/app/src/EventsSheet/EventsTree/Renderers/LinkEvent/ExternalEventsAutoComplete.js
Oxey405: It's to allow including the events in the scene linked to the externals events like externalevents1 is linked to scene1 but it indeed causes a bug that allow the same never ending loop in the compiler. I am removing it.

4ian: externalEvents is an array, so it will always be true, so this filter will filter nothing => so whatever you tried to do, it is probably not what you intended. Be sure to double check every line and ensure JavaScript is properly doing what you think ;)

newIDE/electron-app/app/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
```
4ian, line 1: Even if it's not a big deal, you should not include this file in your commit - always check/add to your commit only the file that are useful for your PR (this one is not) ;)

4ian (approved):
Looks good, thanks!

Oxey405:
Yay !

PR #3234, opened 2021-11-10
Fix clamping of volume in the sound manager

4ian:
Ensure the sound is clamped between 0 and 1, otherwise it's possible that actions changing the volume would set something too high/too low (<0 or >1) which would get ignored by Howler.

### DIFF
--- a/GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
+++ b/GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
@@ -26,7 +26,7 @@ namespace gdjs {
       return 0;
     }
     return volume;
-  }
+  };
 
   /**
    * A thin wrapper around a Howl object with:
--- a/GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
+++ b/GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
@@ -16,11 +16,23 @@ namespace gdjs {
   };
 
   /**
+   * Ensure the volume is between 0 and 1.
+   */
+  const clampVolume = (volume: float): float => {
+    if (volume > 1.0) {
+      return 1.0;
+    }
+    if (volume < 0) {
+      return 0;
+    }
+    return volume;
+  }
+
+  /**
    * A thin wrapper around a Howl object with:
    * * Handling of callbacks when the sound is not yet loaded.
    * * Automatic clamping when calling `setRate` to ensure a valid value is passed to Howler.js.
-   *
-   * See https://github.com/goldfire/howler.js#methods for the full documentation.
+   * * Automatic clamping when calling `setVolume` so that the volume is always between 0 and 1.
    *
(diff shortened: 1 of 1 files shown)


4ian:
This was a problem in a game doing fading of a music, using a timer. Because the music was not loaded, I think the volume was directly trying to be set at something >1 (because the timer was fast). In the past, the volume would have started at 1 directly. But now that musics/sounds are starting at 0, it would have been staying at 0 (because if you pass a volume >1 to Howler, it gets rejected).

In any case, it seems safer to clamp the volume - as we should not be able to go beyond 1 or below 0.

GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
```diff
@@ -64,7 +76,7 @@ namespace gdjs {
 
     constructor(howl: Howl, volume: float, loop: boolean, rate: float) {
       this._howl = howl;
-      this._volume = volume;
+      this._volume = clampVolume(volume);
```
ClementPasteau, line 79: nice. This seems indeed safer to clamp to the volume the same way it is done for the rate.

I'm not sure what's cleanest though.
- storing the clamped value (like you did for the volume)
- storing the original value, but clamp it when playing in the .play() method (like I did for the rate, L104)

In any case, we should do the same for both

4ian: That's a good remark. I've tested on the current version, with a sound starting at volume 0 and an action increasing the volume at each frame, and another action displaying the volume inside a text:

The volume gets "clamped" by Howler, in the sense that it reaches 0.999... and then when you try to increase it a bit more (to 1.0000001 for example), Howler rejects it. So the maximum value stays there.
It would act similarly in the other direction (you would get clamped at 0.01...).

In both cases:
- it means the current GDevelop version "properly" clamp the volume between 0 and 1 for sounds (i.e: between 0 and 100 for events). At least it's not allowing this value to go outside the bounds! So it means we must store the clamped value (to avoid allowing a volume of 200, 300, 400... or -0.5, -1, -100...)

Then for the question of clamping the rate:
- it's currently not clamped (you can have a pitch/rate of 10 and it's a valid value). But for technical reasons, Howler can't go higher than 4 it seems - so we clamp it but we store the original value (because in the future another audio library could maybe play sounds faster).

So I think we are good in terms of behavior?

ClementPasteau: Really good testing 👍 
Indeed, I think we're good in terms of behavior.
I would maybe add a comment in the code to explain why this architecture and avoid someone to refactor it in the future.
Either on the property, the setter or the constructor (or all of them 🙃)

4ian: Great idea, I'll add comments so this is an explicit behavior.

Issue #2281, opened 2021-02-05
Separate action applied multiple times


PascalLadalle:
It seems that the action "move A away from B" applies repel force to A for each instance of B.

https://user-images.githubusercontent.com/25978830/107097876-41574000-680e-11eb-830c-7c0f26337f90.mp4

![image](https://user-images.githubusercontent.com/25978830/107097967-66e44980-680e-11eb-8d97-c2105a3ffd5c.png)
![image](https://user-images.githubusercontent.com/25978830/107097988-706db180-680e-11eb-8abc-9ef51d86ec5e.png)

On a side note, shouldn't the action phrase be consistent with the action label?
![image](https://user-images.githubusercontent.com/25978830/107098065-aad74e80-680e-11eb-9d91-f379cf0d17f3.png)

GD5b104

4ian:
That's indeed a behavior less than ideal. This is because each object around the moving object (here the player) is checked for collision. Then, each of them will gives back a vector to separate the moving object out of the collided object.
This works well when objects are not on the same axis: 
![image](https://user-images.githubusercontent.com/1280130/107101274-b591e380-680e-11eb-8184-aa60af3e772e.png)

But objects that return a vector on the same axis will get them added:
![image](https://user-images.githubusercontent.com/1280130/107101320-d5290c00-680e-11eb-87d8-a0381e3a6ea1.png)

Instead we could project each vector on the axis and only keep the longest:
![image](https://user-images.githubusercontent.com/1280130/107101437-3d77ed80-680f-11eb-989f-35015a2e6eca.png)

But not sure this would work for things that are not axis aligned, for example this:
![image](https://user-images.githubusercontent.com/1280130/107101671-00f8c180-6810-11eb-89f1-158966dc0cb6.png)

Seems like there is no obvious solution, but something to look at more in depth: https://gamedev.stackexchange.com/questions/68486/separating-axis-theorem-with-multiple-polygons

85ayush:
I would love to work on this issue, I am new to open source contribution, can you please guide me?

PR #3236, opened 2021-11-11
Fix the "separate" action when there is several obstacles.

D8H:
## Before the fix
Every vector from the collisionTest are summed. This doesn't work very well when there is more than one obstacle.

SeparateOld

## After the fix

* It finds the longest vector
* project the other ones on its normal 
* apply the longest vector + the longest projected vector if no conflict (between projected vectors)

The downside is that it use 2 static arrays to store the collisionTest results.

SeparateFix

## Example project
The above screenshots are from the "SeparateCases" scene. There is also the Marching Square platform that is a bit more stable.
* https://www.dropbox.com/s/yhc4rl6955247kp/SeparateExample.zip?dl=1

### DIFF
--- a/GDJS/Runtime/runtimeobject.ts
+++ b/GDJS/Runtime/runtimeobject.ts
@@ -28,6 +28,109 @@ namespace gdjs {
   };
 
   /**
+   * Arrays and data structure that are (re)used by
+   * {@link RuntimeObject.separateFromObjects} to avoid any allocation.
+   */
+  const separateFromObjectsStatics: {
+    moveXArray: Array<float>;
+    moveYArray: Array<float>;
+  } = {
+    moveXArray: [],
+    moveYArray: [],
+  };
+
+  /**
+   * Move the object using the results from collisionTest call.
+   *
+   * See {@link RuntimeObject.separateFromObjects}
+   *
+   * @param moveXArray
+   * @param moveYArray
+   * @return true if the object was moved
+   */
+  const moveFollowingSeparatingVectors = (
+    object: RuntimeObject,
+    moveXArray: Array<float>,
+    moveYArray: Array<float>
+  ): boolean => {
+    if (moveXArray.length === 0) {
+      moveXArray.length = 0;
+      moveYArray.length = 0;
+      return false;
+    }
+    if (moveXArray.length === 1) {
+      // Move according to the results returned by the collision algorithm.
+      object.setPosition(
+        object.getX() + moveXArray[0],
+        object.getY() + moveYArray[0]
+      );
+      moveXArray.length = 0;
+      moveYArray.length = 0;
+      return true;
+    }
+
+    // Find the longest vector
+    let squaredDistanceMax = 0;
+    let distanceMaxIndex = 0;
+    for (let index = 0; index < moveXArray.length; index++) {
+      const moveX = moveXArray[index];
+      const moveY = moveYArray[index];
+
+      const squaredDistance = moveX * moveX + moveY * moveY;
+      if (squaredDistance > squaredDistanceMax) {
+        squaredDistanceMax = squaredDistance;
+        distanceMaxIndex = index;
+      }
+    }
+
+    const distanceMax = Math.sqrt(squaredDistanceMax);
+    // unit vector of the longest vector
+    const uX = moveXArray[distanceMaxIndex] / distanceMax;
+    const uY = moveYArray[distanceMaxIndex] / distanceMax;
+
+    // normal vector of the longest vector
+    const vX = -uY;
+    const vY = uX;
+
+    // Project other vectors on the normal
+    let scalarProductMin = 0;
+    let scalarProductMax = 0;
+    for (let index = 0; index < moveXArray.length; index++) {
+      const moveX = moveXArray[index];
+      const moveY = moveYArray[index];
+
+      const scalarProduct = moveX * vX + moveY * vY;
+      scalarProductMin = Math.min(scalarProductMin, scalarProduct);
+      scalarProductMax = Math.max(scalarProductMax, scalarProduct);
+    }
+
+    // Apply the longest vector
+    let deltaX = moveXArray[distanceMaxIndex];
+    let deltaY = moveYArray[distanceMaxIndex];
+
+    // Apply the longest projected vector if they all are in the same direction
+    // Some projections could have rounding errors,
+    // they are considered negligible under a 1 for 1,000,000 ratio.
+    const scalarProductMinIsNegligible =
+      -scalarProductMin < scalarProductMax / 1048576;
+    const scalarProductMaxIsNegligible =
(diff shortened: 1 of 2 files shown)


D8H:
It's probably not the right time to merge something like this. But it can open the discussion.

4ian:
That's totally awesome :) I have to dig more in how it's done but I guess the idea of using a normal vector and projecting the smaller vector onto it is a great one. Did you find articles about this or something that you figured out by yourself?

D8H:
> That's totally awesome :) I have to dig more in how it's done but I guess the idea of using a normal vector and projecting the smaller vector onto it is a great one. Did you find articles about this or something that you figured out by yourself?

I drew some cases and that seems to be a solution. The math sound logical but maybe it's not the best way.

Bouh:
This PR can solve this issue: https://github.com/4ian/GDevelop/issues/2281

GDJS/Runtime/runtimeobject.ts
```diff
@@ -1617,40 +1618,128 @@ namespace gdjs {
...
+      return this._doSeparateFromObjects(moveXArray, moveYArray);
+    }
+
+    /**
+     * Separate the object from others objects, using the results from collisionTest call.
+     * @param moveXArray
+     * @param moveYArray
+     * @return true if the object was moved
+     */
+    private _doSeparateFromObjects(
```
4ian, line 1665: Maybe we could find a slightly different names, like "_moveFollowingSeparatingVectors"?

Also, nitpicking: I see this method is only accessing `this.xxx` public methods. I wonder if we could then extract it outside of the class (passing the object as parameter)?
This just helps clarify that this is a "helper" for the two other functions, and make sure this is not used outside of the class.

GDJS/Runtime/runtimeobject.ts
```diff
@@ -2135,6 +2224,19 @@ namespace gdjs {
      */
     static forcesGarbage: Array<gdjs.Force> = [];
 
+    /**
+     * Arrays and data structure that are (re)used by RuntimeObject.separateFromObjects to
+     * avoid any allocation.
+     * @static
+     */
+    private static separateFromObjectsStatics: {
```
4ian, line 2232: I think we can even extract this... outside of the class! :) 
That would avoid us to expose this member on the class interface

GDJS/Runtime/runtimeobject.ts
```diff
@@ -1617,40 +1618,128 @@ namespace gdjs {
...
+      let deltaY = moveYArray[distanceMaxIndex];
+
+      // Apply the longest projected vector if they all are in the same direction
+      // Some projections could have rounding errors,
+      // they are considered negligible under a 1 for 1,000,000 ratio.
+      if (
+        -scalarProductMin < scalarProductMax / 1048576 !==
+        scalarProductMax < -scalarProductMin / 1048576
+      ) {
+        if (scalarProductMin !== 0) {
```
4ian, line 1731: If scalarProductMin is very very very small (-0.0001) and scalarProductMax is very large (10000), then we will be there (because 
`-scalarProductMin < scalarProductMax / 1048576 !== scalarProductMax < -scalarProductMin / 1048576` => `0.0001 < 10000 / 1048576 !== 10000 < 0.0001 / 1048576` => `0.0001 < 0.0095 !== 10000 < 9.5e-11` => `true !== false` => `true`)

BUT because scalarProductMin is not exactly 0, we will move according to scalarProductMin.

Should we instead look at if scalarProductMin is not negligible against scalarProductMax?

4ian:
Added a few nitpickings and a question. Great job again, looking forward to have this implemented :)

GDJS/Runtime/runtimeobject.ts
```diff
@@ -1617,40 +1618,128 @@ namespace gdjs {
...
+          this.getX() + moveXArray[0],
+          this.getY() + moveYArray[0]
+        );
+        moveXArray.length = 0;
+        moveYArray.length = 0;
+        return true;
+      }
+
+      // Find the longest vector
+      let distanceSqMax = 0;
```
AlexandreSi, line 1686: I don't know if I am the only one who has a problem with abbreviation, but I think it would be clearer with `distanceSq` => `squaredDistance` and `distanceSqMax` => `maxSquaredDistance`

GDJS/tests/tests/runtimeobject.separateFromObjects.js
```diff
@@ -0,0 +1,117 @@
+// @ts-check
+/**
+ * Common tests for gdjs game engine.
+ * See README.md for more information.
+ */
+
+describe('gdjs.RuntimeObject.separateFromObjects', () => {
```
AlexandreSi, line 7: ❤️

AlexandreSi (approved):
It looks great! I just have a small comment on variable names 😅.
And thanks for the images as examples of the result.
Do you maybe have a drawing that sums up the projections that you are doing to get the final deltas?

GDJS/Runtime/runtimeobject.ts
D8H: Yes, I don't like them either. I only use a few that think are common like "Sq" or "Itr". I will make the change but I will keep the max at the end because it's a maximum of the distances and not a distance of a maximum.

GDJS/Runtime/runtimeobject.ts
```diff
@@ -27,6 +27,109 @@ namespace gdjs {
...
+    let deltaY = moveYArray[distanceMaxIndex];
+
+    // Apply the longest projected vector if they all are in the same direction
+    // Some projections could have rounding errors,
+    // they are considered negligible under a 1 for 1,000,000 ratio.
+    const scalarProductMinIsNegligible =
+      -scalarProductMin < scalarProductMax / 1048576;
+    const scalarProductMaxIsNegligible =
+      scalarProductMax < -scalarProductMin / 1048576;
+    if (scalarProductMinIsNegligible !== scalarProductMaxIsNegligible) {
```
4ian, line 118: Way clearer 👍

4ian (approved):
Very nice! We'll probably want to merge this after the next version, so that we don't have a stable version having both the new platformer and this PR at the same time (as this can produces slight changes when dealing with obstacles - even if I'm confident this is better for all games!).

D8H:
> Very nice! We'll probably want to merge this after the next version, so that we don't have a stable version having both the new platformer and this PR at the same time (as this can produces slight changes when dealing with obstacles - even if I'm confident this is better for all games!).

Yes, it will make the support easier.

GDJS/Runtime/runtimeobject.ts
AlexandreSi: I guess abbreviations are always obvious to the one using them (and myself from the past is not the same as myself in the future!)
Ok for the max! Thanks

4ian:
Merging this now that v5.0.122 is out! :)

PR #3241, opened 2021-11-12
Add logging on audio actions on non-existing channels

arthuro555:
Adds some logging and prevent crashes when using a HowlerSound method on a channel where no sound is playing.


GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
```diff
@@ -655,7 +766,7 @@ namespace gdjs {
     }
 
     getSoundOnChannel(channel: integer): HowlerSound {
```
4ian, line 768: What do you think of changing the return type to `HowlerSound | null`, rather than using the null object pattern? This might be clearer for the user - better fail early than later?

arthuro555: Ok

4ian:
Back to this! :) I think we can go ahead, but can you remove all exclamation marks and replace them by a period? 
The "exclamation mark" is to be reserved to exceptional use cases, and for this it's more than probable that games will have these logs. The importance of a log is to be communicated through the level of the log (info, warning, error) and not through the exclamation mark (this can be very tiresome if all logs are yelling at you :)).

4ian (approved):

4ian:
Thank you!

Issue #2775, opened 2021-07-01
Automatically generate collision masks for TileMap objects


4ian:
## Description

Currently gdevelop can parse tiled files and render tilemaps from them.
The user however still needs to manually create collision objects within GDevelop.

## Solution suggested

Tiled lets you create shape objects and collisions on tiles.
The tilemap extension can use this data to create collision shape objects - potentially the user needs to be able to:

* enable/disable this feature - for shape objects and/or tile shapes (more expensive if each tile has a collision - so not sure if we want to allow that)
* only create collisions for shape objects that are in a specific group, maybe even detect the tiled data group a collision belongs to when colliding with it

Note that the engine is not optimized to handle an object with a lot of hitboxes. This means we may have to rework how collision handling is done.

This is linked to https://trello.com/c/3B6otFoa/481-add-ability-to-automatically-generate-collision-shapes-to-the-tilemap-extension and discussed here: https://github.com/4ian/GDevelop/discussions/2366

Bouh:
Note in the checklist, the collision shapes are for the moment convex only in GD5.
The support of concave collisions does not exist.

I tell you because Tiled editor has a[ free line mode for creating collision shapes.](https://youtu.be/72x3jaJrc-0?t=304)

Silver-Streak:
To Bouh's point, I'm not sure how/if you'd be able to implement Tiled collisions unless Concave collisions became possible.

Silver-Streak:
Related to this, LDtk is going to add a collision creation method to its IntGrid layers: https://github.com/deepnight/ldtk/issues/357

As I am personally fond of/biased towards LDtk over Tiled, I'd love to see that method implemented (since that gets GD5 closer to full IntGrid support for LDtk, which has amazing possibilities)

willianholtz:
> Related to this, LDtk is going to add a collision creation method to its IntGrid layers: [deepnight/ldtk#357](https://github.com/deepnight/ldtk/issues/357)
> 
> As I am personally fond of/biased towards LDtk over Tiled, I'd love to see that method implemented (since that gets GD5 closer to full IntGrid support for LDtk, which has amazing possibilities)

My vote is the focus on Tiled, sorry @Silver-Streak  haha. I think it's simpler, direct and cleaner. And to work with tiled it's much better. I used LDtk for a while, actually I still use it, but the real thing is that I prefer Tiled. What is missing in Tiled in my opinion, is the export of maps created (one by one) in .world, which in LDtk already exists.

[...]

PR #3245, opened 2021-11-14
[RuntimeObject] Allow grid based object to optimize collision checks

D8H:
This is built over:
* https://github.com/4ian/GDevelop/pull/3236

It adds a new method to get the hitboxes around a given area: `getHitBoxesAround(left: float, top: float, right: float, bottom: float)`
This method is used in collision checks so grid based objects like tile map or marching squares can be optimized.

The only impact on performances is on `separateFromObjects` when another object has more than 4 hitboxes. In this case, the AABB of `this` is calculated to try to get less hit boxes. Objects with a lot of hitboxes will probably be a tile map where optimizations can take place.

It will help to write a PR for this issue (the next thing I will do):
* https://github.com/4ian/GDevelop/issues/2775

### DIFF
--- a/Extensions/Lighting/lightruntimeobject-pixi-renderer.ts
+++ b/Extensions/Lighting/lightruntimeobject-pixi-renderer.ts
@@ -21,11 +21,10 @@ namespace gdjs {
     _debugMode: boolean = false;
     _debugLight: PIXI.Container | null = null;
     _debugGraphics: PIXI.Graphics | null = null;
+
     /**
-     * The vertices instances are actually shared
-     * with {@link gdjs.RuntimeObject.hitBoxes}.
-     * This is very hacky and hitBoxes should
-     * probably be used directly if needed.
+     * A polygon updated when vertices of the light are computed
+     * to be a polygon bounding the light and its obstacles.
      */
     _lightBoundingPoly: gdjs.Polygon;
 
@@ -59,15 +58,8 @@ namespace gdjs {
       this._indexBuffer = new Uint16Array([0, 1, 2, 0, 2, 3]);
       this.updateMesh();
       this._isPreview = runtimeScene.getGame().isPreview();
+      this._lightBoundingPoly = gdjs.Polygon.createRectangle(0, 0);
 
-      // Shadow copy of the object hit box.
-      // Is this really intended?
-      this._lightBoundingPoly = new gdjs.Polygon();
-      for (let i = 0; i < 4; i++) {
-        this._lightBoundingPoly.vertices.push(
-          runtimeObject.getHitBoxes()[0].vertices[i]
-        );
-      }
       this.updateDebugMode();
 
       // Objects will be added in lighting layer, this is just to maintain consistency.
@@ -399,22 +391,18 @@ namespace gdjs {
       }
 
       // Synchronize light bounding polygon with the hitbox.
-      const lightHitboxPoly = this._object.getHitBoxes()[0];
+      // Note: we suppose the hitbox is always a single rectangle.
+      const objectHitBox = this._object.getHitBoxes()[0];
       for (let i = 0; i < 4; i++) {
-        for (let j = 0; j < 2; j++) {
-          // This is actually completely useless as
-          // this._lightBoundingPoly.vertices[i] === lightHitboxPoly.vertices[i]
-          this._lightBoundingPoly.vertices[i][j] =
-            lightHitboxPoly.vertices[i][j];
-        }
+        this._lightBoundingPoly.vertices[i][0] = objectHitBox.vertices[i][0];
+        this._lightBoundingPoly.vertices[i][1] = objectHitBox.vertices[i][1];
       }
-      const obstaclesCount = lightObstacles.length;
-      // This array could be avoided.
-      const obstacleHitBoxes = new Array<Iterable<gdjs.Polygon>>(
-        obstaclesCount
-      );
-      for (let i = 0; i < obstaclesCount; i++) {
-        obstacleHitBoxes[i] = lightObstacles[
+
+      // Create the list of polygons to compute the light vertices
+      const obstaclePolygons: Array<gdjs.Polygon> = [];
+      obstaclePolygons.push(this._lightBoundingPoly);
+      for (let i = 0; i < lightObstacles.length; i++) {
+        const obstacleHitBoxes = lightObstacles[
           i
         ].behavior.owner.getHitBoxesAround(
           searchAreaLeft,
@@ -422,14 +410,11 @@ namespace gdjs {
           searchAreaRight,
           searchAreaBottom
         );
-      }
-      const obstaclePolygons: Array<gdjs.Polygon> = [];
-      obstaclePolygons.push(this._lightBoundingPoly);
-      for (let i = 0; i < obstaclesCount; i++) {
-        for (const hitbox of obstacleHitBoxes[i]) {
+        for (const hitbox of obstacleHitBoxes) {
           obstaclePolygons.push(hitbox);
         }
       }
+
       let maxX = this._object.x + this._radius;
       let minX = this._object.x - this._radius;
       let maxY = this._object.y + this._radius;
@@ -454,10 +439,6 @@ namespace gdjs {
           }
         }
       }
-      // obstaclePolygons[0] === this._lightBoundingPoly
-      // this._lightBoundingPoly === this._object.hitBoxes
-      // This is actually modifying the object hit boxes.
-      // Is it intended?
       obstaclePolygons[0].vertices[0][0] = minX;
       obstaclePolygons[0].vertices[0][1] = minY;
       obstaclePolygons[0].vertices[1][0] = maxX;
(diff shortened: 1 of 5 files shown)


GDJS/Runtime/runtimeobject.ts
```diff
@@ -1640,40 +1684,134 @@ namespace gdjs {
...
+          for (const otherObject of otherObjects) {
+            if (otherObject.id === this.id) {
+              continue;
+            }
+            const otherHitBoxes = otherObject.getHitBoxesAround(
+              aabb.min[0],
+              aabb.min[1],
+              aabb.max[0],
+              aabb.max[1]
+            );
```
D8H, line 1714: The only impact should be in `SeparateFromObjects` where an AABB is asked on 1 object.
If the object has less than 4 hit boxes we could avoid to calculate its AABB and just call the "non optimized" method.

D8H:
The performance impact should be nearly inexistent now.

4ian (commented):
Looks good I think from a quick check. Great stuff!! 😄

It's important to understand that `getHitboxesAround` is taking the area in the scene space coordinates. So objects like tilemap will have to consider any transformation they have (i.e: they must convert from the scene space coordinates to their local coordinates, to then find the hitboxes in the area) (unless I'm mistaken).

GDJS/Runtime/runtimeobject.ts
```diff
@@ -1253,6 +1256,37 @@ namespace gdjs {
...
+     *
+     * This method can be override by grid based objects to optimize collision checks.
+     *
+     * When overriding this method, the following ones should be overridden too:
+     * * {@link getHitBoxes}
+     * * {@link getAABB}
+     * * {@link updateHitBoxes}
+     * * {@link updateAABB}
+     *
+     * @param left bound of the area
```
4ian, line 1274: Let's precise "left bound of the area, in scene coordinates" and for the 3 others too :)

4ian:
> This method is used in collision checks so grid based objects like tile map or marching squares can be optimized.

This is used in:
- platformer engine
- The separate action: `separateFromObjects`, `separateFromObjectsList` 
- The collision condition (`collisionTest`)
- The raycasting condition (`raycastTest`)
- The point condition/`isCollidingWithPoint`

Anything else you think we might have missed?

D8H:
> > This method is used in collision checks so grid based objects like tile map or marching squares can be optimized.
> 
> This is used in:
> 
>     * platformer engine
> 
>     * The separate action: `separateFromObjects`, `separateFromObjectsList`
> 
>     * The collision condition (`collisionTest`)
> 
>     * The raycasting condition (`raycastTest`)
> 
>     * The point condition/`isCollidingWithPoint`
> 
> 
> Anything else you think we might have missed?

In `runtimescene-pixi-renderer`, the call is to draw hitboxes for debugging which is ok.

```
davy@PC2014-06:/mnt/c/WorkspaceGDevelop/GDevelop/GDJS/Runtime$ grep -rn "getHitBoxes(" .
./pixi-renderers/runtimescene-pixi-renderer.ts:212:        const hitboxes = object.getHitBoxes();
./runtimeobject.ts:1246:    getHitBoxes(): gdjs.Polygon[] {
./runtimeobject.ts:1287:      return this.getHitBoxes();
./runtimeobject.ts:1646:      const hitBoxes = this.getHitBoxes();
./runtimeobject.ts:1654:        let otherHitBoxesArray = otherObject.getHitBoxes();
./runtimeobject.ts:1706:      const hitBoxes = this.getHitBoxes();
./runtimeobject.ts:1718:            let otherHitBoxesArray = otherObject.getHitBoxes();
davy@PC2014-06:/mnt/c/WorkspaceGDevelop/GDevelop/GDJS/Runtime$
```

I didn't change the lighting extension because I don't know it enough, but I could take a look. 
In `platformerobjectruntimebehavior`, the call is on the character which is legit. If a character is a tilemap, it should be small enough.

```
davy@PC2014-06:/mnt/c/WorkspaceGDevelop/GDevelop/Extensions$ grep -rn "getHitBoxes(" .
./Lighting/lightruntimeobject-pixi-renderer.ts:60:          runtimeObject.getHitBoxes()[0].vertices[i]
./Lighting/lightruntimeobject-pixi-renderer.ts:389:      const lightHitboxPoly = this._object.getHitBoxes()[0];
./Lighting/lightruntimeobject-pixi-renderer.ts:399:        obstacleHitBoxes[i] = lightObstacles[i].behavior.owner.getHitBoxes();
./PlatformBehavior/platformerobjectruntimebehavior.ts:2045:      for (const hitBox of behavior.owner.getHitBoxes()) {
davy@PC2014-06:/mnt/c/WorkspaceGDevelop/GDevelop/Extensions$
```

There is no usage in the extension repository.
```
davy@PC2014-06:/mnt/c/WorkspaceGDevelop/GDevelop-extensions/Extensions$ grep -rn "getHitBoxes(" .
davy@PC2014-06:/mnt/c/WorkspaceGDevelop/GDevelop-extensions/Extensions$
```

Also, I wanted to ask for making `getHitBoxes` an `Iterable` to avoid to force grid based implementations to produce a flat array.

GDJS/Runtime/runtimeobject.ts
```diff
@@ -1253,6 +1256,37 @@ namespace gdjs {
       return this.hitBoxes;
     }
 
+    /**
+     * Get the hit boxes for the object around a given area.
```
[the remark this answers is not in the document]
D8H: The following lines explain it:
```
     * The hit boxes don't need to actually overlapping the area,
     * but the ones that do must be returned.
```
This is method can return every hit-boxes when there is no specific optimizations. And optimized implementations can give useless hit-boxes if it's a good trade of.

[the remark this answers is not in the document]
D8H: It could be: "Return at least all the hit boxes overlapping a given area"

D8H:
I made the changes for the lighting extension.
This extension implementation do very odd things and probably need optimizations. I added a few comments.

D8H:
I fix a bug while testing it with a tile map platformer.

4ian:
Agree on having more tests once the tile map implementation is done (we could also have a GridHitBoxesTestRuntimeObject or something like this if we feel like we need a test object for this).

For the lighting thanks for adding comments where things are not optimal - the end of the implementation was maybe a bit too rushed when we wanted to get it done.

4ian:
@D8H Do you want to rebase this on master? :) As the previous PR is now merged and going in the new release of today, I think we're getting ready to get this in, so then the PR on tilemaps can progress :)

4ian:
Thank you for the rebase. I aim to review and merge this by Friday (it's essentially the lights where I need to double check your comments) so you are assured we have this in to do the tilemap stuff :)

Extensions/Lighting/lightruntimeobject-pixi-renderer.ts
```diff
@@ -379,38 +387,54 @@ namespace gdjs {
...
         for (let j = 0; j < 2; j++) {
+          // This is actually completely useless as
+          // this._lightBoundingPoly.vertices[i] === lightHitboxPoly.vertices[i]
           this._lightBoundingPoly.vertices[i][j] =
             lightHitboxPoly.vertices[i][j];
         }
       }
       const obstaclesCount = lightObstacles.length;
-      const obstacleHitBoxes = new Array(obstaclesCount);
+      // This array could be avoided.
```
4ian, line 412: You're correct this intermediate array is not useful.

Extensions/Lighting/lightruntimeobject-pixi-renderer.ts
```diff
@@ -430,6 +454,10 @@ namespace gdjs {
           }
         }
       }
+      // obstaclePolygons[0] === this._lightBoundingPoly
+      // this._lightBoundingPoly === this._object.hitBoxes
+      // This is actually modifying the object hit boxes.
+      // Is it intended?
```
4ian, line 460: I don't think this is intended.

The idea in the algorithm is to always have a rectangle which is at least as large as the light (i.e the rectangle which has a width/height of `2*radius` with as center the center of the light), but if obstacles are outside this rectangle we extend it.

But this should not modify the hitbox. So the renderer here should keep its own `this._lightBoundingPoly` without pointing to the vertices of the object hitbox - which is wrong (but thankfully not creating issues, so far at least)

4ian (commented):
So the issues in the light runtime object renderer boil down to:

- `_lightBoundingPoly` is, wrongly, a shallow copy of the hitbox, sharing the vertices. Instead we should change this so that it owns its own vertices.
  Note that`_lightBoundingPoly` is still useful I think. It's a quadrilateral at least as large as the light radius, and can be extended if obstacles are going beyond it.
- `obstacleHitBoxes` can be removed, because we can directly create `obstaclePolygons`.

I'll make the fixes tomorrow (unless you want to give it a try). Then if we've not forgot any call to `getHitBoxesAround`, I think we'll be good to go.

4ian:
I've hopefully fixed the wrong stuff for the light - not perfect yet I guess but better.

4ian (approved):

4ian:
Looks good to go, merging this now! Thanks for working on this, looking forward to how this can be useful with the tilemap collisions :D

PR #3256, opened 2021-11-17
Add support to display and post user achievements

AlexandreSi:
Design is a WIP, so are the achievements

- My Profile

image

- Other User's Profile

image


newIDE/app/src/EventsSheet/index.js
```diff
@@ -1084,406 +1115,427 @@ export default class EventsSheet extends React.Component<Props, State> {
...
+// EventsSheet is a wrapper so that the component can use multiple
+// context in class methods while correctly exposing the interface.
+const EventsSheet = (props, ref) => {
+  React.useImperativeHandle(ref, () => ({
+    updateToolbar,
+  }));
+
+  const component = React.useRef<?EventsSheetComponentWithoutHandle>(null);
+  const updateToolbar = () => {
+    console.log(component);
```
ClementPasteau, line 1520: log

newIDE/app/src/Profile/AuthenticatedUserProvider.js
```diff
@@ -227,6 +229,7 @@ export default class AuthenticatedUserProvider extends React.Component<
         console.error('Error while loading user limits:', error);
       }
     );
+    this._onBadgesChanged();
```
ClementPasteau, line 232: should this function rather be called `_fetchUserBadges` and passed to the context via the function `onBadgesChanges = this._fetchUserBadges` ?

newIDE/app/src/Utils/GDevelopServices/User.js
```diff
@@ -48,3 +75,68 @@ export const getUserPublicProfile = (
...
+    onBadgesChanged,
+  } = authenticatedUser;
+  if (!badges || !firebaseUser) return null;
+  if (badges.map(badge => badge.achievementId).includes(achievementId)) {
+    return null;
+  }
+  console.log(`posting achievement ${achievementId}`);
+
+  const userId = firebaseUser.uid;
+  return getAuthorizationHeader().then(authorizationHeader =>
```
ClementPasteau, line 96: am I right thinking we have no way to protect the endpoint to avoid people from giving themselves the badges at this stage ?

newIDE/app/src/Utils/GDevelopServices/User.js
```diff
@@ -48,3 +75,68 @@ export const getUserPublicProfile = (
     .get(`${GDevelopUserApi.baseUrl}/user-public-profile/${id}`)
     .then(response => response.data);
 };
+
+const postBadgeIfLocked = (
```
ClementPasteau, line 79: I was confused at first about the signification of "locked", can we find something more meaningful?
Should this be `assignBadgeIfNotClaimed` or just `assignBadge`?

newIDE/app/src/Utils/GDevelopServices/User.js
```diff
@@ -48,3 +75,68 @@ export const getUserPublicProfile = (
...
+): Function => {
+  const { badges } = authenticatedUser;
+  if (!badges) return callback;
+  if (badges.map(badge => badge.achievementId).includes(achievementId)) {
+    return callback;
+  }
+
+  return (...args) => {
+    postBadgeIfLocked(authenticatedUser, achievementId);
+    console.log(args);
```
ClementPasteau, line 139: log

newIDE/app/src/Utils/GDevelopServices/User.js
```diff
@@ -48,3 +75,68 @@ export const getUserPublicProfile = (
...
+        if (err.response.status === 409) {
+          console.warn('Badge already exists');
+        } else {
+          throw err;
+        }
+      })
+  );
+};
+
+export const addPostBadgePreHookIfLocked = (
```
ClementPasteau, line 126: same here, this function is hard to understand from its name (I don't have a better idea yet...)

newIDE/app/src/Utils/GDevelopServices/User.js
```diff
@@ -48,3 +75,68 @@ export const getUserPublicProfile = (
...
+  authenticatedUser: AuthenticatedUser,
+  achievementId: string,
+  callback: Function
+): Function => {
+  const { badges } = authenticatedUser;
+  if (!badges) return callback;
+  if (badges.map(badge => badge.achievementId).includes(achievementId)) {
```
ClementPasteau, line 133: aren't you doing this logic twice? here and in the `postBadgeIfLocked` function

newIDE/app/src/Utils/GDevelopServices/User.js
AlexandreSi: Only trivial badges will be created this way, so even if someone hacks the system, they can only create themselves the "first web export", "first event", etc.

newIDE/app/src/Profile/Achievement/AchievementList.js
```diff
@@ -0,0 +1,90 @@
...
+      }, {});
+
+      setFormattedAchievements(
+        achievements.map(achievement => {
+          if (Object.keys(badgeByAchievementId).includes(achievement.id)) {
+            return {
+              ...achievement,
+              unlockedAt: badgeByAchievementId[achievement.id].unlockedAt,
+            };
+          } else {
```
ClementPasteau, line 55: you don't need the else here

newIDE/app/src/Utils/GDevelopServices/User.js
AlexandreSi: Yes, I can put it in common: but the first is to add the hook or not (and prevent code execution if badge is already owned), the other is to make the call or not, in case the hook is still set.

newIDE/app/src/EventsSheet/index.js
```diff
@@ -235,11 +252,25 @@ export default class EventsSheet extends React.Component<Props, State> {
     fontSize: 14,
   };
 
+  constructor(props: ComponentProps) {
+    super(props);
+    this.addNewEvent = addPostBadgePreHookIfLocked(
+      this.props.authenticatedUser,
+      'trivial_first-event',
```
ClementPasteau, line 259: should those magic strings be in a list of constants somewhere? To avoid typing mistakes.

I'd recommend a new file `src/Utils/GDevelopServices/Achievement.js` or `Badge.js` -> you can add the list here, plus the typings, plus the api calls

newIDE/app/src/Utils/GDevelopServices/User.js
```diff
@@ -17,6 +19,19 @@ export type UserPublicProfileByIds = {|
   [key: string]: UserPublicProfile,
 |};
 
+export type Badge = {|
```
ClementPasteau, line 22: see the other comment, I'd move all that to a `Badge.js` or `Achievement.js`?

newIDE/app/src/Profile/Achievement/AchievementList.js
```diff
@@ -0,0 +1,90 @@
...
+  useEffect(
+    () => {
+      const badgeByAchievementId = badges.reduce((acc, badge) => {
+        acc[badge.achievementId] = badge;
+        return acc;
+      }, {});
+
+      setFormattedAchievements(
+        achievements.map(achievement => {
+          if (Object.keys(badgeByAchievementId).includes(achievement.id)) {
```
4ian, line 50: You can probably just do
```suggestion
          if (!!badgeByAchievementId[achievement.id]) {
```

if the key is not defined or the value undefined, it will be falsy.
This avoids to do a O(n) search in a temporary array for each achievement.

newIDE/app/src/Profile/Achievement/AchievementList.js
```diff
@@ -0,0 +1,90 @@
...
+              style={
+                achievement.unlockedAt
+                  ? styles.unlockedAchievement
+                  : styles.lockedAchievement
+              }
+            >
+              <Text>{achievement.name}</Text>
+              {achievement.unlockedAt && (
+                <Text>
+                  <DateFormat>{achievement.unlockedAt}</DateFormat>
```
4ian, line 79: While it's convenient, we should probably avoid DateFormat in the hope of upgrading one day to js lingui v3:

image

... or we can roll our own DateFormat (that importes I18n then calls i18n._) component, that we put in UI.

newIDE/app/src/Profile/Achievement/AchievementList.js
```diff
@@ -0,0 +1,90 @@
...
+          }
+        })
+      );
+    },
+    [badges, achievements]
+  );
+
+  return (
+    <Column>
+      <div style={styles.achievementsContainer}>
```
4ian, line 66: Maybe give a try at `ScrollView` component?

newIDE/app/src/Profile/Achievement/UserAchievements.js
```diff
@@ -0,0 +1,103 @@
...
+import { ResponsiveLineStackLayout } from '../../UI/Layout';
+import Text from '../../UI/Text';
+import {
+  getAchievements,
+  type Badge,
+  type Achievement,
+} from '../../Utils/GDevelopServices/User';
+
+import AchievementList from './AchievementList';
+// $FlowFixMe[missing-export] TODO: check that this is the correct way to import SVG in the project. If yes, follow https://github.com/facebook/flow/issues/7121.
```
4ian, line 16: We don't import SVG directly but use Material-UI "SvgIcon" component that wrap SVG for us. 
For example, look at newIDE/app/src/UI/CustomSvgIcons/Chrome.js

See usage in the doc:
https://v4.mui.com/components/icons/#svgicon
https://v4.mui.com/api/svg-icon/

newIDE/app/src/Profile/Achievement/UserAchievements.js
```diff
@@ -0,0 +1,103 @@
...
+  return (
+    <Column>
+      <ResponsiveLineStackLayout>
+        <div style={styles.leftContainer}>
+          <div style={styles.summary}>
+            <Trophy height="100px" />
+            {displayError ? (
+              <Text>
+                <Trans>Unexpected error</Trans>
+              </Text>
```
4ian, line 70: Prefer a AlertMessage of kind error

```
<AlertMessage kind="error">
            <Trans>Unable to display your achievements for now.</Trans>
            <Trans>
              Please check your internet connection or try again later.
            </Trans>
          </AlertMessage>
          ```

File: newIDE/app/src/Profile/Achievement/UserAchievements.js [hunk 16]
```diff
@@ -0,0 +1,103 @@
...
+          </div>
+        </div>
+        <div style={styles.rightContainer}>
+          {displayError ? (
+            <Text>
+              <Trans>
+                An error happened when loading achievements, please try again
+                later.
+              </Trans>
+            </Text>
```
4ian, line 90: If there is an error, I would just display a AlertMessage (see other comment) and avoid having these two columns entirely.

newIDE/app/src/Profile/AuthenticatedUserProvider.js
4ian: Agree with this: 
- the function should be named after what it does: this._fetchUserBadges. 
- but it is passed to the context with a name indicating when it should be called (when badges are changed: onBadgesChanged).

These are two different responsibilities. onBadgesChanged is plugged into this._fetchUserBadges.

newIDE/app/src/Profile/AuthenticatedUserProvider.js
```diff
@@ -242,6 +245,23 @@ export default class AuthenticatedUserProvider extends React.Component<
     }));
   };
 
+  _onBadgesChanged = () => {
+    const { firebaseUser } = this.state.authenticatedUser;
+    if (!firebaseUser) return;
+    getUserBadges(firebaseUser.uid).then(
```
4ian, line 251: nit: use async/await for this function?

newIDE/app/src/Utils/GDevelopServices/User.js
```diff
@@ -48,3 +75,68 @@ export const getUserPublicProfile = (
     .get(`${GDevelopUserApi.baseUrl}/user-public-profile/${id}`)
     .then(response => response.data);
 };
+
+const postBadgeIfLocked = (
+  authenticatedUser: AuthenticatedUser,
+  achievementId: string
+): ?Promise<?Badge> => {
```
4ian, line 82: Returning *maybe* a promise (containing *maybe* a badge) seems weird/dangerous.

Instead, always return a promise:
```suggestion
): Promise<?Badge> => {
```

So that the caller can safely `await` you. You can then either:
* mark the function `async`, and benefit from await and from the automatic boxing of `return null` into a `return Promise.resolve(null)`.
* or just replace `return null` by `return Promise.resolve(null)`.

I would go the async route if possible, but the second option work.
The important is to always return a promise, even if it turns out to be immediately set to a result.

newIDE/app/src/Utils/GDevelopServices/User.js
4ian: I think it makes sense, you can both:
- add a comment to tell "Check if the badge is owned now, to avoid executing any extra code if the badge is already owned".
- Put `badges.map(badge => badge.achievementId).includes(achievementId)` in a helper function `hasBadgeWithAchievementId` to make this a tiny bit less cryptic.

newIDE/app/src/EventsSheet/index.js
```diff
@@ -1084,406 +1115,427 @@ export default class EventsSheet extends React.Component<Props, State> {
...
+                          indexInList: eventsList.getEventsCount(),
+                        });
+                      }}
+                      onOpenExternalEvents={onOpenExternalEvents}
+                      onOpenLayout={onOpenLayout}
+                      searchResults={eventsSearchResultEvents}
+                      searchFocusOffset={searchFocusOffset}
+                      onEventMoved={this._onEventMoved}
+                      showObjectThumbnails={
+                        this.props.userPreferences.values
```
4ian, line 1187: uber nitpicking: you could `..., userPreferences } = this.props;` at the beginning of the render to avoid this.props.userPreferences everywhere and reduce a bit the diff here.

newIDE/app/src/EventsSheet/index.js
```diff
@@ -1084,406 +1115,427 @@ export default class EventsSheet extends React.Component<Props, State> {
...
+  const component = React.useRef<?EventsSheetComponentWithoutHandle>(null);
+  const updateToolbar = () => {
+    console.log(component);
+    if (component.current) component.current.updateToolbar();
+  };
+  return (
+    <AuthenticatedUserContext.Consumer>
+      {(authenticatedUser: AuthenticatedUser) => (
+        <PreferencesContext.Consumer>
+          {userPreferences => (
```
4ian, line 1527: nitpicking: Use `useContext` as you're in a functional component - and say good bye to these ugly indentation of render props :D 

It's a good first step toward converting one day the entirety of EventsSheet into a functional component.

newIDE/app/src/Profile/Achievement/AchievementList.js
```diff
@@ -0,0 +1,90 @@
...
+  },
+  unlockedAchievement: {
+    ...achievementBaseline,
+    color: 'green',
+    display: 'flex',
+    justifyContent: 'space-between',
+  },
+  lockedAchievement: {
+    ...achievementBaseline,
+    color: '#CCCCCC',
```
4ian, line 35: For the color:
- For something locked, prefer reducing the opacity so that the theme color is preserved.
- For unlocked, I would use something from the theme. For example, `message.valid`.

```
const gdevelopTheme = React.useContext(ThemeContext);
//...
  style={{...styles.something, color: gdevelopTheme.messages.valid}}
```

newIDE/app/src/Profile/Achievement/AchievementList.js
```diff
@@ -0,0 +1,90 @@
...
+const styles = {
+  achievementsContainer: {
+    maxHeight: 250,
+    overflowY: 'scroll',
+  },
+  unlockedAchievement: {
+    ...achievementBaseline,
+    color: 'green',
+    display: 'flex',
+    justifyContent: 'space-between',
```
4ian, line 31: This could maybe be a `Line` with justifyContent="space-between".
In general I tend to do all the layouting with Line/Column/Spacer/ScrollView as much as possible.

newIDE/app/src/Utils/GDevelopServices/User.js
AlexandreSi: I knew that when writing this, there was something odd 😓

newIDE/app/src/Profile/Achievement/UserAchievements.js
AlexandreSi: Really clear, thanks!

AlexandreSi:
For the translations, is there another part of the code that does not hold every possibility as for this case (they are stored in the could). Should I add a script during extraction that downloads the achievements file and add a key for each achievement title and description?

newIDE/app/src/Utils/GDevelopServices/Badge.js
```diff
@@ -0,0 +1,107 @@
...
+};
+
+/**
+ * Check if user has already claimed the achievement, to avoid executing
+ * any extra code if that's the case.
+ */
+export const addCreateBadgePreHookIfNotClaimed = (
+  authenticatedUser: AuthenticatedUser,
+  achievementId: string,
+  callback: Function
```
4ian, line 89: I'll try to find if we can type this exactly with Flow.
The risk of using Function is that we lose all type information after calling `addCreateBadgePreHookIfNotClaimed`.

newIDE/app/src/Export/ExportLauncher.js
```diff
@@ -70,11 +75,35 @@ export default class ExportLauncher extends Component<Props, State> {
...
   componentWillUnmount() {
     this.buildsWatcher.stop();
   }
 
+  constructor(props: Props) {
+    super(props);
+    this._setupAchievementHook();
+  }
+  componentDidUpdate(prevProps: Props, prevState: State) {
+    this._setupAchievementHook();
```
4ian, line 89: Definitely React hooks are a blessing compared to "hacking" a component functions 😬

newIDE/app/src/Profile/Achievement/resources/trophy.svg
```diff
@@ -0,0 +1,13 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 528.9 511.62">
```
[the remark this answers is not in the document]
Bouh: Yes, you wrote a new UI component somewhere else. So no need to keep the original svg file.
[`newIDE/app/src/UI/CustomSvgIcons/Trophy.js`](https://github.com/4ian/GDevelop/pull/3256/files#diff-a74a3b5bb2de30d73bf592d2aa7ffb0662a9a822c8b02da871a8154064b36b09)

newIDE/app/src/Export/ExportLauncher.js
AlexandreSi: Oh yes, completely!

newIDE/app/src/Profile/Achievement/AchievementList.js
```diff
@@ -0,0 +1,135 @@
...
+        if (b.unlockedAt && a.unlockedAt) {
+          return b.unlockedAt - a.unlockedAt;
+        } else if (a.unlockedAt && !b.unlockedAt) {
+          return -1;
+        } else if (!a.unlockedAt && b.unlockedAt) {
+          return 1;
+        } else {
+          return 0;
+        }
+      });
```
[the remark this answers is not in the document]
AlexandreSi: Good idea

newIDE/app/src/UI/Grid.js
```diff
@@ -14,6 +14,7 @@ export const Line = (props: {|
   justifyContent?: string,
   expand?: boolean,
   overflow?: string,
+  padding?: string | number,
```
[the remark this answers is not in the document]
AlexandreSi: I plead guilty. I'll fix this

newIDE/app/src/Profile/Achievement/resources/trophy.svg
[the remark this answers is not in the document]
AlexandreSi: I was aware that SvgIcon adds its own svg tag. Thanks!

newIDE/app/src/Utils/GDevelopServices/Badge.js
4ian: @AlexandreSi Give a try at making this generic?

`addCreateBadgePreHookIfNotClaimed<T> = (..., callback: T): T => { ... }`

AlexandreSi: I've searched a bit but it seems complicated.
We always return a function with the same signature as callback. 
Writing this, I realize that we don't return anything, i'll fix this (well typing is useful!)

AlexandreSi: image

And found this post https://stackoverflow.com/questions/69005241/cant-return-a-function-from-function-that-returns-a-constrained-generic-type

AlexandreSi: I think flow cannot make sure that the function we return is actually T?

4ian: I'm giving a try at if we can do this without `// $FlowFixMe` (equivalent to // @ts-ignore), but if not we can:
- still use T in the definition of `addCreateBadgePreHookIfNotClaimed`
- Use // $FlowFixMe inside `addCreateBadgePreHookIfNotClaimed`, so that the only risk, not properly typed stuff is inside the function, but the callers are properly typed including the return type (`Function` is equivalent to `any`, so this spills the beans and we'll lose confidence in all the code calling this).

4ian:
Merged ✨ Thanks for the hard work on this and on the underlying system! 👍

ClementPasteau:
🎉  lot of work indeed!

PR #3259, opened 2021-11-18
Feature/allow user receive game stats

ClementPasteau:
This tackles different things:

- Create a checkbox in the profile to allow the user to receive weekly game stats on their registered games
- Improve the Analytic Widget to display in order:
  - A login form if not logged in
  - a Register button if logged in but game isn't registered
  - a warning if the game is registered but not with your account
  - a Get weekly stats button if registered but not receiving stats
  - Analytics widget if registered & receiving game stats
  
All game Analytics can still be accessed via the Games dashboard.

This also ensures that every time an export is done, the game details are updated, meaning:
- its name
- its author

Screenshot 2021-11-18 at 16 41 27
Screenshot 2021-11-18 at 16 42 17
Screenshot 2021-11-18 at 16 47 29
Screenshot 2021-11-18 at 16 42 41
Screenshot 2021-11-18 at 16 41 56
Screenshot 2021-11-18 at 16 41 09


newIDE/app/src/Export/ExportLauncher.js
```diff
@@ -113,94 +141,100 @@ export default class ExportLauncher extends Component<Props, State> {
...
-          this.setState({
-            exportStep: 'upload',
+    try {
+      setStep('register');
+      const profile = authenticatedUser.profile;
+      const getAuthorizationHeader = authenticatedUser.getAuthorizationHeader;
+      const gameId = project.getProjectUuid();
+      const authorName = project.getAuthor() || 'Unspecified author';
+      const gameName = project.getName() || 'Untitled game';
+      if (profile) {
```
4ian, line 155: We could even run this logic in a separate function async that we dont await - in the sense that we launch it and immediately launch the export step (so there is no "register" step).
This could avoid a few hundreds milliseconds that should in theory not be waited for when building a game :)

newIDE/app/src/Export/ExportLauncher.js
```diff
@@ -113,94 +141,100 @@ export default class ExportLauncher extends Component<Props, State> {
...
+        exportState: this.state.exportState,
+      };
+      setStep('export');
+      this.setState({
+        stepCurrentProgress: 0,
+        stepMaxProgress: 0,
+        errored: false,
+        build: null,
+      });
+      const preparedExporter = await exportPipeline.prepareExporter(
```
4ian, line 194: Way clearer, great job!

newIDE/app/src/GameDashboard/GameRegistration.js
```diff
@@ -116,22 +121,102 @@ export const GameRegistration = ({
...
 
+  const _onAcceptGameStatsEmail = React.useCallback(
+    async () => {
+      if (!profile || !project) return;
+
+      setAcceptGameStatsEmailInProgress(true);
+      try {
+        await onAcceptGameStatsEmail();
+      } catch (error) {
+        console.error('Unable to acccept game stats email.', error);
```
4ian, line 132: accept

newIDE/app/src/GameDashboard/GameRegistration.js
```diff
@@ -116,22 +121,102 @@ export const GameRegistration = ({
...
+        showErrorBox({
+          rawError: error,
+          errorId: 'game-stats-email-error',
+          message:
+            'Unable to accept game stats email. ' +
```
4ian, line 137: No space at the end here.

newIDE/app/src/GameDashboard/GameRegistration.js
```diff
@@ -116,22 +121,102 @@ export const GameRegistration = ({
...
+
   React.useEffect(
     () => {
       loadGame();
     },
     [loadGame]
   );
 
-  if (!authenticated) {
+  if (!project || hideIfRegistered) {
```
4ian, line 154: Regression here it seems: If `hideIfRegistered === true`, this component will be never rendered.

newIDE/app/src/Utils/GDevelopServices/Authentication.js
```diff
@@ -285,6 +288,33 @@ export default class Authentication {
...
+  acceptGameStatsEmail = async (
+    getAuthorizationHeader: () => Promise<string>
+  ) => {
+    const { currentUser } = this.auth;
+    if (!currentUser)
+      throw new Error(
+        'Tried to accept game stats email while not authenticated.'
+      );
+
+    return getAuthorizationHeader()
```
4ian, line 300: This code seems duplicated.

newIDE/app/src/Utils/GDevelopServices/User.js
```diff
@@ -48,3 +48,25 @@ export const getUserPublicProfile = (
...
+        {
+          params: { userId: userId },
+          headers: { Authorization: authorizationHeader },
+        }
+      );
+    })
+    .then(response => response.data)
+    .catch(error => {
+      console.error('Error while editing user:', error);
+      throw error.response.data;
```
4ian, line 70: throw error - otherwise you risk an error because response/data might not be defined, and this is not consistent anyway with other functions that don't have a catch and just throw the original error.

newIDE/app/src/Export/ExportLauncher.js
ClementPasteau: Yeah I thought about that and ended up waiting for it because it wasn't too long. 
I'll do it to see if it works well 👍

newIDE/app/src/Utils/GDevelopServices/User.js
ClementPasteau: I think I copied this from somewhere else, I'll make sure this is fixed too.

4ian (approved):
Excellent! Make sure to test a few online packaging and the HTML5 export to check we don't have any obvious regression.

PR #3260, opened 2021-11-18
[Platformer] Fix collision detections of platforms that are rotated at runtime.

D8H:
It fixes:
* https://github.com/4ian/GDevelop/issues/3240

United and checked on the example attached on the issue.


4ian (approved):

4ian:
Thank you! 👍
I'll apply the same logic to the LightObstacleRuntimeBehavior and PathfindingObstacleRuntimeBehavior - anything that relies on the RBush actually.
In the future we could have a RBush handled by the engine directly, but for now better have the same fix everywhere.

PR #3261, opened 2021-11-18
[JSON Manager] Avoid to fetch again an object that is already being fetched.

D8H:
I didn't unit it. It would need to mock `XMLHttpRequest`. As there is no test on `JsonManager`, I didn't write one. I only manually smoke tested that GDevelop can open a project with a tilemap.
* https://github.com/4ian/GDevelop/discussions/3257


4ian:
Should be good 👍

Issue #3244, opened 2021-11-14
Editing Object Variables in Object properties


Midhil457:
## Description
Currently to edit object variables, user has to right click and click `edit variables`. Which is very difficult to find for new users. It would make sense to have it in the object properties. Like shown here:

![variables](https://user-images.githubusercontent.com/73597906/141677049-a23e4da2-6791-4c24-947d-4be1721a3efc.png)
## Solution suggested
I am happy to do it myself, The screenshot shows my attempt at it. But, I have having some issues figuring out how to get the object variables and other things I should make sure of and who the `VariablesList` works. It would be great if someone could help me. Currently when the tab is opened, it results in a crash (obviously)

Bouh:
I agree, it don't have really sense to make it available by the context menu while other options are in the object edition.

4ian:
Agree variables should be there - it's part of what an object is, and it's already in the right click menu so for consistency they should be there.

PR #3263, opened 2021-11-19
Move Object variables to Object Editor Dialog

Midhil457:
Moved Object variables to Object Editor Dialog (#3244)

- [x] Added object variables to object editor
- [x] Update `Edit object variable` action to direct to the new tab
- [x] (Tried) Remove unnessesary code
- [x] Fix VariableList component not taking up all the area

I need help with the last one, the component seems to be moved to the side. I have spent a few days trying to fix it, with no good result (Most likely due to my lack of experience :sweat_smile: ) , it would be great if someone could help find and fix the problem.

EDIT: Fixed it, see next comment
Everything else seems to work perfectly!

gdghalf

:)

### DIFF
--- a/newIDE/app/src/EventsSheet/ParameterFields/GlobalVariableField.js
+++ b/newIDE/app/src/EventsSheet/ParameterFields/GlobalVariableField.js
@@ -64,10 +64,17 @@ export default class GlobalVariableField extends React.Component<
             }}
             emptyExplanationMessage={
               <Trans>
-                Global variables are variables that are persisted across the
-                scenes during the game.
+                Global variables are variables that are shared amongst all the
+                scenes of the game.
               </Trans>
             }
+            emptyExplanationSecondMessage={
+              <Trans>
+                For example, you can have a variable called UnlockedLevelsCount
+                representing the number of levels unlocked by the player.
+              </Trans>
+            }
+            helpPagePath={'/all-features/variables/global-variables'}
             onComputeAllVariableNames={onComputeAllVariableNames}
           />
         )}
--- a/newIDE/app/src/ObjectEditor/ObjectEditorDialog.js
+++ b/newIDE/app/src/ObjectEditor/ObjectEditorDialog.js
@@ -196,29 +196,27 @@ const InnerDialog = (props: InnerDialogProps) => {
         />
       )}
       {currentTab === 'variables' && (
-        <Column expand noMargin>
-          <VariablesList
-            variablesContainer={props.object.getVariables()}
-            emptyExplanationMessage={
-              <Trans>
-                When you add variables to an object, any instance of the object
-                put on the scene or created during the game will have these
-                variables attached to it.
-              </Trans>
-            }
-            emptyExplanationSecondMessage={
-              <Trans>
-                For example, you can have a variable called Life representing
-                the health of the object.
-              </Trans>
-            }
-            helpPagePath={'/all-features/variables/object-variables'}
-            onSizeUpdated={
-              forceUpdate /*Force update to ensure dialog is properly positioned*/
-            }
-            onComputeAllVariableNames={props.onComputeAllVariableNames}
-          />
-        </Column>
+        <VariablesList
+          variablesContainer={props.object.getVariables()}
+          emptyExplanationMessage={
+            <Trans>
+              When you add variables to an object, any instance of the object
+              put on the scene or created during the game will have these
+              variables attached to it.
+            </Trans>
+          }
+          emptyExplanationSecondMessage={
+            <Trans>
+              For example, you can have a variable called Life representing the
+              health of the object.
+            </Trans>
+          }
+          helpPagePath={'/all-features/variables/object-variables'}
+          onSizeUpdated={
+            forceUpdate /*Force update to ensure dialog is properly positioned*/
+          }
+          onComputeAllVariableNames={props.onComputeAllVariableNames}
+        />
       )}
       {currentTab === 'effects' && (
         <EffectsList
--- a/newIDE/app/src/ObjectEditor/ObjectEditorDialog.js
+++ b/newIDE/app/src/ObjectEditor/ObjectEditorDialog.js
@@ -197,6 +197,7 @@ const InnerDialog = (props: InnerDialogProps) => {
         />
       )}
       {currentTab === 'variables' && (
+        <Column expand>
         <VariablesList
         variablesContainer={props.object.getVariables()}
         emptyExplanationMessage={
@@ -223,6 +224,7 @@ const InnerDialog = (props: InnerDialogProps) => {
         )}
         
         />
+        </Column>
       )}
       {currentTab === 'effects' && (
         <EffectsList
--- a/newIDE/app/src/ObjectEditor/ObjectEditorDialog.js
+++ b/newIDE/app/src/ObjectEditor/ObjectEditorDialog.js
@@ -212,6 +212,7 @@ const InnerDialog = (props: InnerDialogProps) => {
                 the health of the object.
               </Trans>
             }
+            helpPagePath={'/all-features/variables/object-variables'}
             onSizeUpdated={
               forceUpdate /*Force update to ensure dialog is properly positioned*/
             }
--- a/newIDE/app/src/ObjectEditor/ObjectEditorDialog.js
+++ b/newIDE/app/src/ObjectEditor/ObjectEditorDialog.js
@@ -23,13 +23,11 @@ import HotReloadPreviewButton, {
   type HotReloadPreviewButtonProps,
 } from '../HotReload/HotReloadPreviewButton';
 import EffectsList from '../EffectsList';
-import VariablesList from '../VariablesList/index'
-import EventsRootVariablesFinder from '../Utils/EventsRootVariablesFinder';
+import VariablesList from '../VariablesList/index';
 
 type Props = {|
   open: boolean,
   object: ?gdObject,
-  variables: ?gdObject.variables,
 
   onApply: () => void,
   onCancel: () => void,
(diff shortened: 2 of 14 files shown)


Midhil457:
I was able to fix the issue by using a Column and adding `expand` and `noMargin` :partying_face:
fixvar

newIDE/app/src/ObjectEditor/ObjectEditorDialog.js
```diff
@@ -188,6 +196,36 @@ const InnerDialog = (props: InnerDialogProps) => {
...
+        onSizeUpdated={
+          forceUpdate /*Force update to ensure dialog is properly positioned*/
+        }
+        onComputeAllVariableNames={() =>
+          EventsRootVariablesFinder.findAllObjectVariables(
+            props.project.getCurrentPlatform(),
+            props.project,
+            props.layout,
+            props.object
+        )}
```
4ian, line 224: props.layout will never exist/be set, so this will never work properly. I think we can keep this call to findAllObjectVariables in SceneEditor/index.js and pass it as a prop:

```suggestion
        onComputeAllVariableNames={props.onComputeAllVariableNames}
```

and in SceneEditor/index.js:

```
<ObjectEditorDialog
  ...
  onComputeAllVariableNames={() => EventsRootVariablesFinder.findAllObjectVariables(...)}
/>
```

This way we avoid polluting ObjectEditorDialog by passing it the layout - which would be only useful for this and so would be meaning passing "too much information" to the component while it should not have access to the layout.

newIDE/app/src/ObjectEditor/ObjectEditorDialog.js
```diff
@@ -23,10 +23,13 @@ import HotReloadPreviewButton, {
   type HotReloadPreviewButtonProps,
 } from '../HotReload/HotReloadPreviewButton';
 import EffectsList from '../EffectsList';
+import VariablesList from '../VariablesList/index'
+import EventsRootVariablesFinder from '../Utils/EventsRootVariablesFinder';
 
 type Props = {|
   open: boolean,
   object: ?gdObject,
+  variables: ?gdObject.variables,
```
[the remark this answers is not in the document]
4ian: This error is also reported by Flow, the type checker, that you can find in the CI runs: https://gdevelop.semaphoreci.com/jobs/c3aa34e5-238e-4c65-b021-87a1ed661b32

newIDE/app/src/ObjectsList/index.js
```diff
@@ -10,7 +10,6 @@ import SortableVirtualizedItemList from '../UI/SortableVirtualizedItemList';
 import Background from '../UI/Background';
 import SearchBar from '../UI/SearchBar';
 import NewObjectDialog from '../AssetStore/NewObjectDialog';
```
4ian, line 12: Changes in the file seem all good 👍

newIDE/app/src/VariablesList/VariablesEditorDialog.js
```diff
@@ -63,14 +62,6 @@ const VariablesEditorDialog = ({
       cannotBeDismissed={true}
       onRequestClose={onCancelChanges}
       secondaryActions={[
```
4ian, line 64: You've removed a feature here, this means we can't open the object variables anymore when we edit the variables of an instance.
Add it back and find a way instead to open the object editor with the variable tab opened :) 

This means you'll have to see where this is used, search for `onEditObjectVariables` in the codebase.
You'll find it in SceneEditor/index.js (ignore the things related to `UseSceneEditorCommands`, it's a totally different component), here precisely:

https://github.com/4ian/GDevelop/blob/2517b4740111a3dc790f27f6c7b4d671b26a0f1d/newIDE/app/src/SceneEditor/index.js#L1315-L1331

If you read this, you see this opens a variable editor for the object. You need to:
1) Replace this by a call to `this.editObject(object, 'variables')` - simple! :) 
2) A bit harder: now you must clean all usage of `this.state.variablesEditedObject` and remove `<VariablesEditorDialog` which is related to `this.state.variablesEditedObject`. Why? Because this dialog is now useless, as we moved the variables edition inside the object editor :)

4ian (commented):
That's a good start! Thanks for taking the time to do this 👍

See my comments, there are a few points notably one to:
- Put back the button to open object variables in the instance variables (careful when doing this, you must call this explicitly in your PR description because otherwise we risk losing features without noticing)
- Adapt the SceneEditor/index.js which was also containing an object variable editor, that we can remove.

Thanks! :)

Midhil457:
Thank you for the review!
I have made the changes according to the review :)

One change I want to make but need the green flag is on replacing the empty message with the `EmptyPlaceHolder` (If that's the name of it) to match the effects and behavior tabs.

Should the `+ Add` button be moved to the bottom? To match the other tab's style. It wouldn't adversly affect the VariableDialog much. I don't exactly know how to do it though (I did try it a few times) :sweat_smile: So I would appreciate some help

4ian:
> One change I want to make but need the green flag is on replacing the empty message with the EmptyPlaceHolder (If that's the name of it) to match the effects and behavior tabs.

I think that would be a good idea! 
Feel free to go ahead and do this. Make sure to also test the result in the case of the scene or global variable editor :) 

> Should the + Add button be moved to the bottom? To match the other tab's style. It wouldn't adversly affect the VariableDialog much. I don't exactly know how to do it though (I did try it a few times) 😅 So I would appreciate some help

We can go back to this later, for now it's still ok this way! :)

newIDE/app/src/SceneEditor/index.js
```diff
@@ -1194,6 +1194,26 @@ export default class SceneEditor extends React.Component<Props, State> {
...
+                variablesEditedInstance.getObjectName()
+              );
+              return variablesEditedObject
+                ? EventsRootVariablesFinder.findAllObjectVariables(
+                    project.getCurrentPlatform(),
+                    project,
+                    layout,
+                    variablesEditedObject
+                  )
+                : [];
```
4ian, line 1215: This seems wrong, you check the object of the edited instance, but we're in the object editor, so we're editing an object.

instead, just pass `this.state.editedObjectWithContext.object` to the last argument of findAllObjectVariables?

newIDE/app/src/SceneEditor/index.js
```diff
@@ -1352,37 +1372,6 @@ export default class SceneEditor extends React.Component<Props, State> {
             }}
           />
         )}
-        {!!this.state.variablesEditedObject && (
```
4ian, line 1355: You also need to clean the unused `variablesEditedObject` in the state, the type and the now useless `editObjectVariables`. 
Otherwise, this will clutter the file and create fake complexity for the future developer working of the file.

blurymind:
This still keeps the instance variables available in the left side properties panel right? My only concern is in not adding a requirement for an extra click to get to it

4ian:
Yes, this just relocates the object variables from their own window into a tab in the object editor.
Instance variables are untouched :)

Midhil457:
> This seems wrong, you check the object of the edited instance, but we're in the object editor, so we're editing an object.

instead, just pass this.state.editedObjectWithContext.object to the last argument of findAllObjectVariables?

It was `this.state.editedObjectWithContext.object` at first, but the test gave me an error at that line, after some testing, that was what solved it, I didn't see that it was calling instance variables

> You also need to clean the unused variablesEditedObject in the state, the type and the now useless editObjectVariables.
Otherwise, this will clutter the file and create fake complexity for the future developer working of the file.

I tried removing as many references that I thought were useless, so I might have missed some even this time


- Changed empty message to EmptyPlaceholderMessage
- Added props for setting Help Page Path
- The instance variable doc doesn't exist. I will create one in the next few weeks, but still have added the link as help path


Changes have been applied to:

- [x] Scene Variable Editor
- [x] Global Variable Editor
- [x] Object Variable Editor
- [x] Instance Variable Editor
- [x] Scene Variable Editor in Scene Variable Field
- [x] Object Variable Editor in Object Variable Field
- [x] Global Variable Editor in Global Variable Field

newIDE/app/src/ObjectEditor/ObjectEditorDialog.js
```diff
@@ -188,6 +196,36 @@ const InnerDialog = (props: InnerDialogProps) => {
           onUpdateBehaviorsSharedData={props.onUpdateBehaviorsSharedData}
         />
       )}
+      {currentTab === 'variables' && (
+        <Column expand noMargin>
```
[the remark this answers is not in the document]
Bouh: `useFullHeight` is missing here?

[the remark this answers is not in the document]
Midhil457: It doesn't seem to help with the positioning issue
Done it like: `<Column expand noMargin useFullHeight>`

newIDE/app/src/EventsSheet/ParameterFields/GlobalVariableField.js
```diff
@@ -64,10 +64,17 @@ export default class GlobalVariableField extends React.Component<
...
               <Trans>
-                Global variables are variables that are persisted across the
-                scenes during the game.
+                Global variables are variables that are shared amongst all the
+                scenes of the game.
               </Trans>
             }
+            emptyExplanationSecondMessage={
+              <Trans>
+                For example, you can have a variable called UnlockedLevelsCount
```
4ian, line 73: Thanks for making this consistent with the project manager 👍

4ian:
I believe I fixed the issues of centering, and also rework the VariablesList component and the variables editor dialog so that it's also centered here and the toolbar stay visible at all time :) 
Also a bunch of Flow typing fixes.

Should be good to go!

Object variables:
image

Object variables (empty):
image

Instance variables in the panel:
image

Instance variables in their editor:
image

Instance variables in their editor (empty):
image

4ian (approved):

PR #3275, opened 2021-11-21
Add conditions and expressions to compare the coordinates of the bounding box of an object

4ian:
This should be a useful addition. My main "concerns" about this are:

- Confusion for users. This is currently "hidden" in a "bounding box" category and there is an explanation about what the bounding box is.
  image
- Risk of over reliance on this, while the position should be preferred. Most of the usage, even in extensions should be and remain using the X;Y position, which is the origin of the object. The bounding box is only for situations where we want to know the area surrounding the (collision masks of an) object.
- 6 more methods on `gdjs.RuntimeObject`. As these are "shortcut" methods, I think it's ok. But it's also cluttering a bit the interface.

Let me know what you think!


D8H:
> 6 more methods on gdjs.RuntimeObject. As these are "shortcut" methods, I think it's ok. But it's also cluttering a bit the interface.

Maybe they can be moved as static helper methods and shown as object condition/expression ?

4ian:
Yep I was thinking of this, but this might be just adding more complexity than anything! (because that's more conditions to be considered as object conditions while not being object conditions, and for expressions it's not even supported).
Not a big deal I guess, the AABB is an important concept for a RuntimeObject anyway.

tristanbob:
I think these conditions are a good addition and very useful for dealing with camera movement.  It's critical to know if any part of the object is on the screen (even when rotated).

It is also the only way to know the REAL center of an object.  Every other point (origin, center) can be modified by the user.

4ian:
Speaking of any part, it's important to understand that the bounding box here is the hit boxes bounding box - it's still possible that an object has a large part outside of these hit boxes. 
There is in the engine the concept of the "visibility bounding box" (encapsulating the image) but 1) it's not really implemented by object and 2) it's more for rendering. 

I reckon everything should be done with the AABB, apart from very specific stuff that would handle the rendering of objects (like choosing to render an object or not to save CPU/GPU when the object "visibility AABB" is "far"). For this, we will continue to require access to JS with getVisibilityAABB. Anyway for now it's the same as the hit boxes AABB.

Issue #3226, opened 2021-11-04
Zooming in or out of a scene doesn't change the scrolling speed


ClementPasteau:
## Describe the bug
Not really a bug and more of a non-optimal UI behavior.

When using a trackpad and scrolling to navigate through a project scene, the speed is in pixels/s, meaning that if we zoom out, the speed of navigating will not increase.
It is really painful if we have a lot of objects on the scene and we want to move around fast.

## To Reproduce
1. Create any project
2. Add elements to the scene
3. Zoom out a few times
4. Try to scroll left and right using a trackpad
5. See that it barely moves

PR #3279, opened 2021-11-23
Improve editor scrolling speed to be adapted based on zoom

ClementPasteau:
Addresses #3226 

- increase default speed
- add a speed factor based on zoom

https://user-images.githubusercontent.com/4895034/142996573-74c858cc-ef15-4a1a-b634-7adc0294b269.mov

### DIFF
--- a/newIDE/app/src/InstancesEditor/index.js
+++ b/newIDE/app/src/InstancesEditor/index.js
@@ -164,12 +164,16 @@ export default class InstancesEditor extends Component<Props> {
     });
 
     this.pixiRenderer.view.onwheel = (event: any) => {
+      const zoomFactor = this.getZoomFactor();
       if (this.keyboardShortcuts.shouldZoom()) {
         this.zoomOnCursorBy(-event.deltaY / 5000);
       } else if (this.keyboardShortcuts.shouldScrollHorizontally()) {
-        this.viewPosition.scrollBy(-event.deltaY / 10, 0);
+        const deltaX = event.deltaY / (5 * zoomFactor);
+        this.viewPosition.scrollBy(-deltaX, 0);
       } else {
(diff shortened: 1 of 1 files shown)


newIDE/app/src/InstancesEditor/index.js
```diff
@@ -167,9 +167,12 @@ export default class InstancesEditor extends Component<Props> {
       if (this.keyboardShortcuts.shouldZoom()) {
         this.zoomOnCursorBy(-event.deltaY / 5000);
       } else if (this.keyboardShortcuts.shouldScrollHorizontally()) {
-        this.viewPosition.scrollBy(-event.deltaY / 10, 0);
+        this.viewPosition.scrollBy(-event.deltaY / 5, 0);
```
4ian, line 170: Is it normal not to use the zoomFactor here?

ClementPasteau: I had not tested this behavior as it was never triggered with my tests.
But now I realise this happens when the user press "alt" and then scroll vertically -> trigger a horizontal scroll.
(I didn't know this was a common scrolling feature 🤔 )
I guess it should apply the same logic as well!

4ian: I guess same logic yeah! 
This was adapted from software doing the same, especially useful for traditional mouses/no trackpad which don't have a horizontal scroll.

Bouh:
So smoothhhh.
Is it also in point, mask, physics, and resources preview?

ClementPasteau:
@Bouh This is not affecting those screens. 
I don't think this is necessary as the objects usually fit the screen well + the scrolling seems better handled there.
By testing it, I couldn't find any frustration, so leaving it as is :)

4ian (approved):

PR #3284, opened 2021-11-24
Decrease build limits correctly after building

ClementPasteau:
Realised that the build limits was never decreasing after a build because we're not fetching the limits again.

We could only do a call to limits, but I thought we might as well get the whole profile (easier technically as well)

https://user-images.githubusercontent.com/4895034/143212101-0fda9594-7867-48e5-a91c-aae6e0506157.mov


4ian:
Excellent! That was often reported.

Issue #785, opened 2018-12-03
Lock position/rotation button not effective


zatsme:
## Describe the bug
In the scene editor you can lock the position and angle of an instance, but it does not work!

![image](https://user-images.githubusercontent.com/17770123/49395047-51a29a80-f72d-11e8-8b00-8406153a6728.png)

Once you check this button you can no longer select the instance, however you can drag select and once you've done that you can use any of the widgets to change the instance.

I believe this is unexpected behaviour and therefore a bug.

Bouh:
I'have selected the instance by list of instance and we have the widgets, and in properties we can aslo change the all values when lock option in checked

4ian:
Lock is meant only as a way to not being able to "select" the instance from the scene editor. So Instance list is fine. Changing properties once it's selected is also fine. 
But being able to select it by clicking and dragging to draw a selection is indeed not the intended behavior and should be fixed if it's the case.

zatsme:
I might be being stupid now, but if you no longer can select or drag select, then it will be a problem to find the instance in the list if you have a lot.

Surely the best behavior would be to be able to select the instance, but to hide the widgets?  That way it would be much easier to unlock the instance when required and at the same time make it impossible to move/rotate/resize in the editor!

🤔

Lizard-13:
Note that it isn't so hard to find locked instances since there's a lock icon in them.
But I have to admit that if there are tons of instances, the instances panel is not the best friend, with the instances ordered in an "arbitrary" way, not Z-order nor object type, I think it's the order in which they were added.

blurymind:
perhaps when an instance gets locked, it should also become deselected?

4ian:
> I might be being stupid now, but if you no longer can select or drag select, then it will be a problem to find the instance in the list if you have a lot of them.

The way it was done in GD4 was that, when you right click on the instance, you had an option to unlock it.

> Surely the best behavior would be to be able to select the instance, but to hide the widgets? That way it would be much easier to unlock the instance when required and at the same time make it impossible to move/rotate/resize in the editor!

The issue is that if for example you select a bunch of instances, including the locked instance, then moving a non locked instance will also move the locked instance (at least that's how it would work right now).

But maybe ideally we should be able to still select them, but any drag/resize/angle should not do anything. Should we also show the properties in the left panel? I think so otherwise you have no way to unlock the instance.

zatsme:
[...]

PR #3286, opened 2021-11-24
Allow selection of locked objects in the editor

ClementPasteau:
Addressing #785 

I think that, like suggested in the issue, when an object is locked, and we still allow its selection, but prevent any action done on it, has a better UX overall.
See video for example.

https://user-images.githubusercontent.com/4895034/143222784-4a40ddef-571e-45e5-ab12-5a558145d2f4.mov

### DIFF
--- a/newIDE/app/src/InstancesEditor/InstancesMover.js
+++ b/newIDE/app/src/InstancesEditor/InstancesMover.js
@@ -84,8 +84,14 @@ export default class InstancesMover {
     this.totalDeltaX += deltaX;
     this.totalDeltaY += deltaY;
 
+    const nonLockedInstances = instances.filter(
+      instance => !instance.isLocked()
+    );
+
     // It will magnet the corner nearest to the grabbing position
-    const initialSelectionAABB = this._getOrCreateSelectionAABB(instances);
+    const initialSelectionAABB = this._getOrCreateSelectionAABB(
+      nonLockedInstances
+    );
     if (!initialSelectionAABB) return;
     const magnetLeft = this._startX < initialSelectionAABB.centerX();
     const magnetTop = this._startY < initialSelectionAABB.centerY();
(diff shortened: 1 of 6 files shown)


4ian:
Interesting! Can you confirm it's also not going to be deleted if I press delete? 
(I might want to select a bunch of stuff on my level to delete them, but not the locked background).

Bouh:
For me, locking an object means not doing any actions on it.
It is rather useful to select a lot of objects and act on them but not on the one that is locked.
I would go even further and remove it from the selection. It would only be accessible from the instance list.

For me blocking = does nothing.

4ian:
> I would go even further and remove it from the selection. It would only be accessible from the instance list.

I can see the point of still allowing selection: so that we can unlock it. But while it's locked, nothing should happen to it:
- it should not be moved
- it should not be rotated
- maybe it should not be copied? (unsure about this)
- maybe it should not be deleted? (unsure about this too)

For now the locking prevents the selection *by clicking* but you can still draw a selection rectangle around it and select/move/delete it.

I think the behavior is an improvement. We could maybe draw the selection around the object using another color when the object is locked?

Bouh:
Indeed, the selection is useful.
I am of the opinion that it should not be deleted or copied.

4ian:
@ClementPasteau Give a try at not deleting the object when locked and when we press delete or click the delete button of the toolbar.
We can deal with the copy behavior separately (I agree it should not be copied, but I also agree this might be surprising because it's selected 🤔.. for now a locked object can be copied, so we can keep it if it's too complex to remove).

ClementPasteau:
I've improved the behavior to change the color of the selected rectangle if locked + remove the resize/rotate buttons + adapt the rectangle to be only on the non-locked items. 
+ disabled the delete action.

Using it, it's quite neat!

https://user-images.githubusercontent.com/4895034/143271292-02bf3120-a7cd-4144-ba0e-178976625f21.mov

4ian (approved):
💯

Issue #941, opened 2019-02-21
Offset tiled sprites become extremely pixelated and drop in frames over time.


SomeWeirdo:
## Describe the bug
Tiled sprites that are offset to appear like they're moving become pixelated only after minutes, this doesn't occur in the preview, but happens on mobile exports and WiFi preview.

Here's a video showcasing the issue https://streamable.com/ol45y I have sped up the video to show the issue, but the actual recording was only 3 minutes long and I've just shortened it to 30 seconds.

This can get to the point where the sprite becomes one single large pixel and runs at about a single frame a second.

I'm using this to create an infinite parallax background that automatically moves without the player needing to move.

## To Reproduce
1. Create a tiled sprite
2. Offset it by any amount and make the condition "always" (or leave blank)
3. Watch as overtime the sprite gets more and more pixelated and more jittery

## Other details
Latest version of GDevelop 5 at the time of writing this, issue doesn't seem to happen in preview mode on macOS High Sierra, 8GB Ram, Intel Iris Plus Graphics 630, but happens on export on Android and WiFi preview for IOS/Android from what I've tested.

Unfortunately this issue is halting the finishing of my game, so I'm really hoping someone can look into this and get it fixed. Thank you.

zatsme:
Can you can post a zip file with a small example with sprites?

4ian:
This might be related to a floating point rounding issue in the graphics cards. To avoid this, you can try to store the offset in a variable, apply a modulo to it (with the width/height of the texture: `mod(Variable(MyOffset), 1000)`) before applying to the offset?

Marichalazo:
The same problem happens to me. 
Try not to stretch the sprite so much and duplicate it. It worked for me.

SomeWeirdo:
@4ian Never done anything like that before, I'm not sure how I would be able to store an offset and then use it later.

Also @zatsme I've attached a file. The cloud in my example looks like this in only 2 minutes or so when running on WiFi Preview or export. Going to need a bit of help for a workaround in the meantime I suppose.

[![IMG-0065.png](https://i.postimg.cc/x1nrf1Wk/IMG-0065.png)](https://postimg.cc/tYSvDpRy)
[Tile_Issue.zip](https://github.com/4ian/GDevelop/files/2896228/Tile_Issue.zip)

zatsme:
[...]

PR #3287, opened 2021-11-24
Fix TileSprite being correctly displayed with an offset

ClementPasteau:
Fixes https://github.com/4ian/GDevelop/issues/941

I've briefly looked at this issue again, since we are now using the official `PIXI.TilingSprite`.
The issue still happens on Android.

It actually is a known issue by Pixi.JS (https://github.com/pixijs/pixijs/issues/7891) because of precision issues.
The fix is indeed to only take the remaining part of the offset based on the texture's size.
Looking good now!


Bouh:
I talked with someone from a `WebGL & WebGPU` discord server, the suggested solution was what you did.
image

Bouh (approved):

4ian:
Thanks!

PR #3298, opened 2021-11-26
Add carousel component

AlexandreSi:
In order to display showcases, tutorials and starters on the start page.

See [Storybook](http://gdevelop-storybook.s3.amazonaws.com/refurbish-homepage/latest/index.html?path=/story/carousel--loading-without-title-skeleton)

As I am not sure about the interface and the data needed to open a starter, the types are very simple and should be completed when integrating it.


Bouh:
Great work it looks and works so great, Netflix can be worried 😁

- Missing the white background under the left arrow button
- Left arrow button doesn't overlap the thumbnail like for the right arrow, buttons are not the same on both sides.
- Missing the button **Browse all**
- Title can be bigger and bolder.

Current PR:
image

Figma mockup:
image

newIDE/app/src/MainFrame/EditorContainers/StartPage/Carousel.css
```diff
@@ -0,0 +1,19 @@
+ul.MuiGridList-root {
```
4ian, line 1: Interesting, is there a risk we conflict with other usage of "MuiGridList" in the future?

newIDE/app/src/MainFrame/EditorContainers/StartPage/Carousel.js
```diff
@@ -0,0 +1,338 @@
...
+  );
+};
+
+const extractVideoIdFromYoutubeLink = (link: string): ?string => {
+  const url = new URL(link);
+  const params = url.searchParams;
+  return params.get('v') || null;
+};
+
+const constructImageLinkFromYoutubeLink = (link: string): string => {
```
4ian, line 319: This should be returned by the API in the future I guess? cc @ClementPasteau

newIDE/app/src/MainFrame/EditorContainers/StartPage/Carousel.js
```diff
@@ -0,0 +1,338 @@
...
+  imageSource: string,
+|};
+
+type SkeletonThumbnail = {|
+  skeleton: boolean,
+  key: string,
+|};
+
+type Props = {|
+  title: string,
```
4ian, line 34: Consider passing a React.Node so that the parents can pass a `title={<Trans>Showcase</Trans>}`

newIDE/app/src/MainFrame/EditorContainers/StartPage/Carousel.js
```diff
@@ -0,0 +1,338 @@
...
+        | YoutubeThumbnail
+        | GameTemplateThumbnail
+        | ShowcaseThumbnail
+        | SkeletonThumbnail
+    ): ?React.Node => {
+      if (!item.link && !item.imageSource && !item.skeleton) return null;
+      if (item.link) {
+        return (
+          <a href={item.link} target="_blank">
+            <img
```
4ian, line 131: Maybe we can add a "alt" and "title" with some content (a string) coming from the item (typically a title)

newIDE/app/src/MainFrame/EditorContainers/StartPage/Carousel.js
```diff
@@ -0,0 +1,338 @@
...
+        <Text size="title">{title}</Text>
+        <GridList
+          cols={itemsToDisplay.length}
+          cellHeight={cellHeight}
+          spacing={cellSpacing}
+          style={styles.gridList}
+          ref={scrollView}
+        >
+          {itemsToDisplay.map(item => (
+            <GridListTile key={computeItemKey(item)}>
```
4ian, line 290: Would be interesting to check if we can focus the tiles using the keyboard, for accessibility i think it would be great to support "tabbing" between them.

AlexandreSi:
Thank you for your feedback @Bouh!
- Missing the white background under the left arrow button and feft arrow button doesn't overlap the thumbnail like for the right arrow, buttons are not the same on both sides.
  - I actually made that "on purpose", because I thought I would do something like Netflix:
  
  image

  But with Material GridList I can't have the thumbnail visible under the arrow. So I kept it like that, it didn't feel odd at the moment, but I can try to change it to have the same behavior as for the right arrow.

- Missing the button Browse all
  - Will do! 
- Title can be bigger and bolder
  - Will do as well

4ian:
Agree on Bouh remarks! I agree I had an odd feeling with the left arrow, not sure what the good solution is (put it on top? hide when not useful a bit like how you did?) but surely something to do here! 👍

newIDE/app/src/MainFrame/EditorContainers/StartPage/Carousel.css
AlexandreSi: Well it would if this style is present in all components, I am not sure about that though: if there is a `MuiGridList` in a component, and this carousel in a dialog for example, will it affect the first?
Anyway, I can add an id/data to the div that wraps the `MuiGridList` and change the CSS to have a `>`

newIDE/app/src/MainFrame/EditorContainers/StartPage/Carousel.js
AlexandreSi: When discussing it, we were not sure, but I have not seen what has been made in the API side. This logic can be used as a default though.

newIDE/app/src/MainFrame/EditorContainers/StartPage/Carousel.js
AlexandreSi: I'll check!

newIDE/app/src/MainFrame/EditorContainers/StartPage/Carousel.js
4ian: Ok we'll see then, a first draft of the API should be done now, probably the API should indicate the image to use (I think there will be some logic so that we always host the images). I'll let you double check! :)

newIDE/app/src/MainFrame/EditorContainers/StartPage/Carousel.css
4ian: I think so, the inclusion of the Carousel.css will *not* be scoped to the component (that's also why we avoid using raw CSS files in the app, they are global and will suffer from traditional CSS flaws: unability to find what is unused/used, risk of impacting other components if you're not careful).

Take a look at if:
1. inline styles would work (passing `style={styles.xxx}` to the component). It's usually good enough for most stuff.
2. if you need to use advanced CSS or modify embedded tags, then you can use Material-UI v4 styling solution. See an example here: https://github.com/4ian/GDevelop/blob/2a386cdcf168437a81b4476eb96d2ec79cb34270/newIDE/app/src/UI/SemiControlledAutoComplete.js#L80-L98
  - This gives you a useStyles hook than you can call and pass the resulting classes to the component. See https://v4.mui.com/api/grid-list/#gridlist-api for this specific component
3. As a last resort, a few global CSS classes can work! :) In this case it's surely safer to use a classname and change the CSS to have a `>`
  
Note that the makeStyles API of Material-UI will change in MUI v5, so between this and a CSS file with a proper classname, I'm not sure what I prefer. I usually just try hard at using only inline styles (to keep the styling inside the component with 0 risks of it leaking outside - the style IS the component) and use the other two solutions as last resorts.

AlexandreSi: I didn't know you could specify pseudo elements and states styles with makeStyles! That's great with it!

newIDE/app/src/MainFrame/EditorContainers/StartPage/Carousel.js
AlexandreSi: I tried to do something, specifying a tab index offset and adding tab index to tiles, it works but it's not perfect. We can perfect it later if the specs are clear.

newIDE/app/src/UI/Carousel.js
```diff
@@ -0,0 +1,329 @@
...
+
+import Text from './Text';
+import { Column, Line, Spacer } from './Grid';
+import GDevelopThemeContext from './Theme/ThemeContext';
+import { useResponsiveWindowWidth } from './Reponsive/ResponsiveWindowMeasurer';
+
+type Thumbnail = {|
+  id: string,
+  title: string,
+  description: string,
```
ClementPasteau, line 19: is the description needed for a Thumbnail?

newIDE/app/src/UI/Carousel.js
```diff
@@ -0,0 +1,329 @@
...
+    },
+    [renderImage, cellWidth, imageHeight]
+  );
+
+  const renderItemTitle = React.useCallback(
+    (item: ThumbnailType | SkeletonThumbnail): ?React.Node => {
+      if (!displayItemTitles) return null;
+      return (
+        <>
+          <Spacer />
```
ClementPasteau, line 176: spacer looks weird here, should it be more where the function `renderItemTitle` is used?

newIDE/app/src/UI/Carousel.js
```diff
@@ -0,0 +1,329 @@
...
+            <div style={{ ...styles.itemTitleContainer, width: cellWidth }}>
+              <Text noMargin style={styles.itemTitle}>
+                {item.title}
+              </Text>
+            </div>
+          ) : (
+            <Skeleton
+              variant="rect"
+              height={titleHeight}
+              width={(cellWidth / 3) * (1 + 2 * Math.random())}
```
ClementPasteau, line 187: I don't understand why there's a `Math.random()` at first reading. Maybe leave a comment?

newIDE/app/src/UI/Carousel.js
```diff
@@ -0,0 +1,329 @@
...
+    scrollViewElement.addEventListener('touchleave', handleScrollEnd);
+  }, []);
+
+  return (
+    <Line noMargin expand>
+      <div
+        style={{ ...styles.arrowContainer, width: arrowWidth }}
+        onClick={onClickArrow('left')}
+      >
+        {!!items && shouldDisplayLeftArrow && <ArrowBackIos />}
```
ClementPasteau, line 279: should all the checks on `items` be more `items && items.length` ? As this is an array, [] will be truthy

newIDE/app/src/UI/Carousel.js
AlexandreSi: You're right, I thought I would need it to add a description for accessibility purposes. Maybe I can add a `title` prop to the item title. What do you think?

newIDE/app/src/UI/Carousel.js
AlexandreSi: Well, if we don't render the title, there is a spacer although it is useless. So I figured I would put it here

newIDE/app/src/UI/Carousel.js
AlexandreSi: So that it feels more "alive". Without it, it would be rectangles of the same length
image

newIDE/app/src/UI/Carousel.js
AlexandreSi: Good catch, I'm on it

newIDE/app/src/UI/Carousel.js
AlexandreSi: Comment added

newIDE/app/src/UI/Carousel.js
ClementPasteau: Descriptions can be long, not sure they would have their place as the title attribute of the item. Maybe I'm wrong?
I'd say remove it for the moment

newIDE/app/src/UI/Carousel.js
```diff
@@ -0,0 +1,356 @@
...
+      flatButtonProps.onClick = onBrowseAllClick;
+    } else {
+      flatButtonProps.onClick = () => {};
+    }
+
+    return (
+      <FlatButton
+        {...flatButtonProps}
+        label={<Trans>Browse all</Trans>}
+        tabIndex={tabIndexOffset + itemsToDisplay.length}
```
4ian, line 221: I really think the tabIndex are not necessary, tabIndex="0" should work on the tiles and are not needed for buttons. Much like z-index, this will create "fights" between components one day or the other.

newIDE/app/src/UI/Carousel.js
```diff
@@ -0,0 +1,356 @@
...
+    if (!onBrowseAllClick && browseAllLink) {
+      flatButtonProps.onClick = () => {
+        Window.openExternalURL(browseAllLink);
+      };
+      flatButtonProps.target = '_blank';
+    } else if (onBrowseAllClick) {
+      flatButtonProps.onClick = onBrowseAllClick;
+    } else {
+      flatButtonProps.onClick = () => {};
+    }
```
4ian, line 215: I would avoid this. Just pass onClick={props.onBrowseAllClick} and let the parent decide what to do.

newIDE/app/src/UI/Carousel.js
```diff
@@ -0,0 +1,356 @@
...
+            />
+          )}
+        </>
+      );
+    },
+    [cellWidth, titleHeight]
+  );
+
+  const renderBrowseAllButton = (): React.Node => {
+    const flatButtonProps: $Shape<FlatButtonProps> = { onClick: () => {} };
```
ClementPasteau, line 205: To avoid using $Shape, I would create the onClick function first, with `let onClick = () => {}` then assign it, then use it in the FlatButton directly ?
`<FlatButton onClick={onClick} ... />`

newIDE/app/src/UI/Text.js
```diff
@@ -48,7 +48,15 @@ export default ({
   displayInlineAsSpan,
 }: Props) => (
   <Typography
-    variant={size === 'title' ? 'h6' : size === 'body2' ? 'body2' : 'body1'}
+    variant={
+      size === 'bold-title'
+        ? 'h5'
+        : size === 'title'
+        ? 'h6'
```
4ian, line 55: title is less large than a bold title.. can be confusing? 🤔

newIDE/app/src/UI/Carousel.js
4ian: Or just a bunch of ifs inside the prop if really it's super complicated ;) 
`onClick={() => { if (...) else (...) }}`

newIDE/app/src/UI/Text.js
4ian: (as discussed, my mistake - resolving this!)

newIDE/app/src/UI/Carousel.js
AlexandreSi: Could be great to have a formation about accessibility to better understand that.

newIDE/app/src/stories/componentStories/Carousel.stories.js
```diff
@@ -0,0 +1,181 @@
...
+      displayItemTitles={false}
+    />
+  );
+};
+
+export const LoadingWithTitleSkeleton = () => (
+  <Carousel title={<Trans>Showcase</Trans>} items={null} />
+);
+export const Showcases = () => {
+  type ShowcaseThumbnail = {|
```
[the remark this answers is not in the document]
AlexandreSi: I feel that a function will be needed to transform a ShowcasedGame to a ShowcaseThumbnail to:
- add a onClick function
- define an id
- select a link?
So I am not sure the Carousel should adapt to this type.

newIDE/app/src/UI/Carousel.js
```diff
@@ -0,0 +1,375 @@
...
+      scrollViewElement.scrollTo({
+        left: roundScroll(scrollViewElement.scrollLeft),
+        behavior: 'smooth',
+      });
+    },
+    [roundScroll]
+  );
+
+  const onFocusItem = event => {
+    event.target.scrollIntoViewIfNeeded();
```
4ian, line 294: This will probably crash on Firefox which seems not to support this: https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoViewIfNeeded

Consider using https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView with `behavior: "smooth" `?

newIDE/app/src/UI/Carousel.js
```diff
@@ -0,0 +1,375 @@
...
+  );
+
+  const roundScroll = React.useCallback(
+    (value: number): number => {
+      return Math.round(value / widthUnit) * widthUnit;
+    },
+    [widthUnit]
+  );
+
+  const onClickArrow = (direction: 'left' | 'right') =>
```
4ian, line 258: On a very small viewport (use the Storybook viewport "small mobile"), the scroll won't scroll because I guess it's smaller than one tile. There is maybe a Math.min to add somewhere to ensure we scroll at least the size of a tile or at least to the next "logical" page:

image

newIDE/app/src/UI/Carousel.js
```diff
@@ -0,0 +1,375 @@
...
+            onClick={
+              onBrowseAllClick ||
+              (browseAllLink
+                ? () => {
+                    Window.openExternalURL(browseAllLink);
+                  }
+                : () => {})
+            }
+            label={<Trans>Browse all</Trans>}
+            tabIndex={0}
```
4ian, line 335: Remove the tabIndex for this button, it's a button so it should by default properly keyboard focusable

newIDE/app/src/UI/FlatButton.js
```diff
@@ -5,7 +5,7 @@ import { Spacer } from './Grid';
 
 // We support a subset of the props supported by Material-UI v0.x FlatButton
 // They should be self descriptive - refer to Material UI docs otherwise.
-type Props = {|
+export type Props = {|
```
4ian, line 8: Probably you can remove this export

newIDE/app/src/UI/FlatButton.js
```diff
@@ -22,6 +22,7 @@ type Props = {|
     flexShrink?: 0,
   |},
   target?: '_blank',
+  tabIndex?: number,
```
4ian, line 25: (as per other comment, remove this as not needed - we never want to use anything other than 0 for tabIndex, and this component is already focusable)

newIDE/app/src/stories/componentStories/Carousel.stories.js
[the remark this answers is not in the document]
4ian: Ok that's fair!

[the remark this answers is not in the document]
4ian: It's the responsibility of the new "home page" to then do the work of :
- fetching the data (using the context)
- rendering carousels, passing them the appropriate props with if needed a quick transformation.

4ian (approved):
Good to go when the comments related to tabIndex/the scroll on a viewport smaller than a tile/focus on Safari/Firefox are dealt with :)

newIDE/app/src/UI/Carousel.js
AlexandreSi: I changed with scrollIntoView as you recommended it but without the behavior, not sure that it's what you want when navigating with tab

PR #3299, opened 2021-11-27
Fix tabs in preference dialog not always on top

### LINKED ISSUE #3293: "GDevelop Preferences" tabs should always remain at top, even after scrolling down
## Description

Currently, when the user scrolls down the GDevelop Preferences screen, they lose access to the tabs at the top.  The user must scroll up to the top to use these tabs.

https://user-images.githubusercontent.com/8879811/143513470-09bd4fa0-9914-4575-b2c9-2538f60f2a38.mp4

## Solution suggested

"GDevelop Preferences" tabs should always remain at the top, even after scrolling down

Midhil457:
- Removed the `Preference` Title as no other similar dialogs (project properties, profile) had a title text. All of them had the Tabs as title
- Fix Tabs not always on top. Fixes #3293 

Pre-comparision

### DIFF
--- a/newIDE/app/src/MainFrame/Preferences/PreferencesDialog.js
+++ b/newIDE/app/src/MainFrame/Preferences/PreferencesDialog.js
@@ -74,12 +74,15 @@ const PreferencesDialog = ({ i18n, onClose }: Props) => {
         <div>
           <Tabs value={currentTab} onChange={setCurrentTab}>
             <Tab label={<Trans>Preferences</Trans>} value="preferences" />
-            <Tab label={<Trans>Hints &amp; explanations</Trans>} value="hints" />
+            <Tab
+              label={<Trans>Hints &amp; explanations</Trans>}
+              value="hints"
+            />
             <Tab label={<Trans>Keyboard Shortcuts</Trans>} value="shortcuts" />
           </Tabs>
-        </div>}
+        </div>
+      }
     >
-      
       {currentTab === 'preferences' && (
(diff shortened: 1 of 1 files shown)


newIDE/app/src/MainFrame/Preferences/PreferencesDialog.js
```diff
@@ -67,15 +67,22 @@ const PreferencesDialog = ({ i18n, onClose }: Props) => {
       onRequestClose={onClose}
       cannotBeDismissed={true}
       open
-      title={<Trans>GDevelop Preferences</Trans>}
+      noTitleMargin
       maxWidth="sm"
       noMargin
+      title={
+        <div>
```
Bouh, line 74: This Div is required ?

newIDE/app/src/MainFrame/Preferences/PreferencesDialog.js
```diff
@@ -67,15 +67,22 @@ const PreferencesDialog = ({ i18n, onClose }: Props) => {
       onRequestClose={onClose}
       cannotBeDismissed={true}
       open
-      title={<Trans>GDevelop Preferences</Trans>}
+      noTitleMargin
```
Bouh, line 70: 👍

Bouh (approved):

4ian:
Thanks, looking better with the tabs staying on top. I've fixed the extra div, looks good to go 👍

PR #3300, opened 2021-11-27
Move scene variables into separate category

Midhil457:
- Moved the scene variables into seperate catergory inside. Moved from `Variables` to `Variables/Scene variables`
- Organize variables catergory


Bouh (approved):

4ian:
Are we sure about this? I guess consistency wins over making scene variables slightly easier to access over global variables?

Bouh:
I agree with the tidying up, people will see much better the distinction between global and scene variables.

4ian:
Let's give this a try then! :)

PR #3301, opened 2021-11-28
Add usage of custom ICE servers to P2P extension

arthuro555:
When using WebRTC in a serious game over the internet, STUN/TURN servers become a necessity, both to have quick connections and to be able to connect at all if the internet provider is very strictly forbidding P2P. While PeerJS uses the Google STUN server by default, it still is missing a TURN server (which is arguably more important).
This PR allows using custom ICE servers (STUN or TURN servers), both for people that do not want to rely on a 3rd party and to be able to have a TURN server.

Tested in the p2p networking example
  -> Not using the action doesn't impact the game -> Not using a custom ICE uses the default one as intended
  -> Putting the google STUN server doesn't impact the game -> The custom ICE is indeed used and working
  -> Putting an invalid STUN URL stops the game's netcode from working -> Removal of the default ICEs works


Extensions/P2P/JsExtension.js
```diff
@@ -154,6 +154,30 @@ module.exports = {
...
 
+    extension
+      .addAction(
+        'UseOwnICEServer',
+        _('Use a custom ICE server'),
+        _(
+          'Disables the default ICE (STUN or TURN) servers list and use one of your own. ' +
+            'Note that it is recommended to add at least 1 self-hosted STUN and TURN server ' +
+            'for games that are not over LAN but over the internet. ' +
+            'This action is cumulable, you can use it multiple times to add multiple servers. ' +
```
4ian, line 165: cumulable is not english I think?
```suggestion
            'This action cane be used multiple times to add multiple servers. ' +
```

4ian (approved):

Extensions/P2P/JsExtension.js
arthuro555:
```suggestion
            'This action can be used multiple times to add multiple servers. ' +
```
You've made a typo too 👀

PR #3306, opened 2021-11-29
Fix example library error text

Midhil457:
- Fix alert mentioning extensions instead of examples
- Use `Update` instead of `Upgrade`for both example and extension dialog  as upgrade sounded like the user had to get a subscription and `update` is a bit of common in the UI


Bouh (approved):

4ian:
Thanks! :)

PR #3310, opened 2021-11-30
Add collapse animation to list component

Midhil457:
Add a collaspe animation to list component. See https://github.com/4ian/GDevelop/discussions/3278 for context

I also wanted to add a rotating arrow icon, but I wasn't able to figure it out.
## Before

https://user-images.githubusercontent.com/73597906/144038078-2ef4d2c9-da94-4989-bdb8-e365d0a55652.mp4

## After
https://user-images.githubusercontent.com/73597906/144037910-16d032db-09d9-4b4d-9e88-b1298ada512c.mp4

### DIFF
--- a/newIDE/app/src/UI/List.js
+++ b/newIDE/app/src/UI/List.js
@@ -264,22 +264,19 @@ export const ListItem = React.forwardRef<ListItemProps, ListItemRefType>(
               )
             ) : null}
             {renderListItemSecondaryAction()}
-          
           </MUIListItem>
           <Collapse in={isItemOpen} timeout="auto" unmountOnExit>
-                <MUIList
-                  component="div"
-                  disablePadding
-                  style={{
-                    paddingLeft: 16,
-                    ...props.nestedListStyle,
-                  }}
-                  dense={useDenseLists}
-                >
-                
-                  {renderNestedItems()}
-                </MUIList>
-
+            <MUIList
+              component="div"
+              disablePadding
+              style={{
+                paddingLeft: 16,
+                ...props.nestedListStyle,
+              }}
+              dense={useDenseLists}
+            >
+              {renderNestedItems()}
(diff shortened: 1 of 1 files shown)


newIDE/app/src/UI/List.js
```diff
@@ -265,7 +265,7 @@ export const ListItem = React.forwardRef<ListItemProps, ListItemRefType>(
             ) : null}
             {renderListItemSecondaryAction()}
           </MUIListItem>
-          {isItemOpen && (
+          <Collapse in={isItemOpen} timeout="auto" unmountOnExit>
```
4ian, line 268: Could you verify if the unmountOnExit effectively works? I don't see it in the version of Material-UI we're using currently.

Also this List is used in other places that in the Project Manager. Can you verify what the result in other places using a `List` with nested items (typically, the action/condition selector maybe?).

Thanks!

Midhil457: I can confirm that the unmountOnExit works, you can see below that in the dev tools the component is unmounted when the list is closed otherwise it would just be hidden

No unmountOnExit:

https://user-images.githubusercontent.com/73597906/144232339-35d72cc9-a5dc-4d79-98ff-efab451f8e2d.mp4

With unmountOnExit:

https://user-images.githubusercontent.com/73597906/144232374-4fa7a1fc-3c99-4405-a542-ba496fc160b0.mp4

As far as I have tested the list works fine in the actions editor too;


https://user-images.githubusercontent.com/73597906/144232453-ce201595-9e1f-4185-a107-7aac057cedd8.mp4

4ian: Thanks for checking! Sounds good then, let's give this a try :)

4ian:
Thank you! Looking good 👍

PR #3313, opened 2021-12-01
[TileMap] Collision mask object

D8H:
This is an implementation for:
* https://github.com/4ian/GDevelop/issues/2775

## Try it
A custom build of GDevelop with the feature
* https://www.dropbox.com/s/dxibwleb4stgxs5/gdevelop-tilemap-collision-beta138-2022-07-12.7z?dl=1

The platformer tile map example with a rotating tile map
* Project: https://www.dropbox.com/s/zyarg6r2rfsip02/TilemapCollisionPlatformer.zip?dl=1
* Build: https://liluo.io/games/70f526f7-c35b-4db9-b63f-7ebc2c3e8989?dev=true

LDTK platformer example converted in Tiled
* Project: https://www.dropbox.com/s/dwmll3wmzdtztmk/TileMapLDTK-Platformer.zip?dl=1
* Build: https://liluo.io/instant-builds/e4d872e6-29f9-4c01-835f-7e6ea571bb28?dev=true

Various tile maps
* Project: https://www.dropbox.com/s/c3m3umd9fve76gm/TileMapExamples.zip?dl=0
* Build: https://liluo.io/instant-builds/37c1b03d-c23d-4c01-bd60-756398f3640d?dev=true

Not a Vania
* Project: https://www.dropbox.com/s/d1y00hezzswc0te/not-a-vania-collision-layer.zip?dl=1
* Build: https://liluo.io/games/c8389601-2960-4197-baa7-f92e3a976c00?dev=true
* PR: https://github.com/GDevelopApp/GDevelop-examples/pull/343

### A tile map model (EditableTileMap)

This model role is to give an efficient way of accessing the tile map content (and maybe modify it in the future).

 * It is used for rending in TileMapRuntimeObjectPixiRenderer
 * and hitboxes handling in TransformedCollisionTileMap.

#### Parsers
 * Tile map files are parsed into this model by TiledTileMapLoader.
 * This allows to support new file format with only a new parser.

### The collision mask object (TransformedCollisionTileMap)

A tile map transformed with an affine transformation to follow the position and angle of the object.

#### Efficient collision

The object has a custom implementation of `getHitboxesAround`. It gives an access to hitboxes for collision handling in:
- O(1) relatively to the map dimension
- O(n) where n is the surface of the other object in tiles which is likely to be very small).

### Tile map sharing (TileMapRuntimeManager and TileMapManager)

To use a tile map with collisions, a user can create 4 objects:
- one for the the rendering
- one for the solid platforms
- one for the jumpthrus
- one for the ladders

To avoid to have 4 copies of the same tile map in memory, this manager puts the tile map in cache and avoid unnecessary parsing.

Every instance with the same files path in properties will share the same EditableTileMap and TileTextureCache.

This might came in handy in the future when editing a tile map from the IDE.

### Library building for the GUI (resolved)
I would like GDJS to be able to use the TypeScript sources directly to allow fast iterations when debugging and build a packaged JavaScript library from them for the GUI. The issue is that GDJS use namespaces, but packaged libraries use modules and they doesn't work well with each other.
Using the packaged library for both would need an extra step. Could `import-GDJS-Runtime.js` automatically build the lib if a file has changed?

### Concave hitboxes
It doesn't check if the hitboxes defined in tiles are convex.
It could use the algo from the NavMesh generator to split them in convex hitboxes when loading the tile set.

### Ellipse hitboxes
It doesn't handle ellipse hitboxes.
It could approximate them with polygons.

### Objects layers
Layer with tiles located by position instead of a grid are ignored for the collision mask. I consider that they will mostly be used for decoration and it doesn't seem important to handle it for the 1st version at least.

### Layer offsets
It can be useful to have a layer shifted from half a tile to place platform more precisely for instance. Tiled allows it, it should be handled in the future.


4ian:
Sorry I'm as usual a bit behind in terms of reviews! 

> The issue is that GDJS use namespaces, but packaged libraries use modules and they doesn't work well with each other.

Yep, all GDJS is "namespaced" (because it was easier when converting to TS).

> Using the packaged library for both would need an extra step. Could import-GDJS-Runtime.js automatically build the lib if a file has changed?

I think we could add a new step in this to also build each lib that would be in the extension, either by declaring them somewhere or just by hardcoding them in the script? 
The IDE should take care of looking at file changes and relaunching the script.

D8H:
> > Using the packaged library for both would need an extra step. Could import-GDJS-Runtime.js automatically build the lib if a file has changed?
> 
> I think we could add a new step in this to also build each lib that would be in the extension, either by declaring them somewhere or just by hardcoding them in the script? The IDE should take care of looking at file changes and relaunching the script.

Each library will have their own package.json?

Having them in a sub folder of the extension would make them easier to found but they will need to be excluded from the extension build. Do you think it's better to have them inside or outside of the extension folder?

4ian:
> Each library will have their own package.json?

Maybe, though if we can avoid running a npm install for each that would be cool. I was thinking of a subfolder or something like this to start.

> Having them in a sub folder of the extension would make them easier to found but they will need to be excluded from the extension build. Do you think it's better to have them inside or outside of the extension folder?

Good question. I was thinking of a subfolder inside the extension folder, and then adapt the build to exclude this folder and build independently each of these.
Maybe I'm overcomplicated stuff though. But if we want something that can be consumed by the IDE and the game engine, that sounds like a solution - and this force the library to be decoupled from the game engine build.

D8H:
I can't find a solution to import the shared library module in the GDJS part. I looked at what was done for Pixi, but I'm not sure how to create the global variable and I would like to avoid to have a list of every exposed classes to maintain.

> Note that we also modified the bundled `pixi.js` file to create a global variable called `GlobalPIXIModule`, containing the `PIXI` object.
> Note that we could use `export as namespace`, but this crash the TypeScript compiler.

Do you know a good solution to use the shared library from the GDS part?

Extensions/TileMap/collision/TransformedTileMap.ts
```diff
@@ -0,0 +1,742 @@
...
+        left: float,
+        top: float,
+        right: float,
+        bottom: float
+      ): Iterable<gdjs.Polygon> {
+        // Return the hitboxes from the tiles that overlap
+        // the AABB of the area in the tile map basis.
+        // Some of these tiles are not event in the given area
+        // but this is a good trade of between the number of
+        // useless returned hitboxes and the time to find them.
```
D8H, line 190: This is what is used for the implementation of `getHitboxesAround`. This is O(1) over the number of tiles in the map.

D8H:
The library works for the GUI part but not with the GDJS part.

I tried to mimic what was done for Pixi:
```JavaScript
var PIXI=function(t){"use strict"

var GlobalPIXIModule = { PIXI: PIXI };
//# sourceMappingURL=pixi-legacy.min.js.map
```

But, the library seems to have extra code to handle the module:
```JavaScript
!(function (e, t) {
  'object' == typeof exports && 'object' == typeof module
    ? (module.exports = t())
    : 'function' == typeof define && define.amd
    ? define([], t)
    : 'object' == typeof exports
    ? (exports.TileMapHelper = t())
    : (e.TileMapHelper = t());
})('undefined' != typeof self ? self : this, function () {
  return (() => {
    'use strict'
```

I hope there is a solution that doesn't involve modifying the packaged library because it might be a bit hard to automatize.

I committed a version with the library project. The extension doesn't compile but this will allow you to check it if you want.
The library can be generated in the extension sub folder `GDevelop\Extensions\TileMap\helper` with this command.
```
GDevelop\SharedLibs\TileMapHelper> npm run tsc
```

Is there anything you would like me to try for solving this issue?

4ian:
> But, the library seems to have extra code to handle the module

Correct, this code is traditional for a "UMD" library, because it works if imported (with Node.js require) and as a global in a browser.
See this pattern explained here: https://www.typescriptlang.org/docs/handbook/declaration-files/templates/global-plugin-d-ts.html

It seems you could configure Webpack to export your library as UMD, so it can be consumed from the IDE (as a Node.js module) or from the extension (as a browser global): https://www.typescriptlang.org/docs/handbook/modules.html#code-generation-for-modules

We might also consider rollup (https://rollupjs.org/guide/en/) as it's faster than Webpack, or just use `esbuild` and inject with code with the `banner` and `footer` option. Give a try with Webpack as "UMD" for now and see if it works better in GDJS (you should have a global variable exposed with your library).

If you struggle for too long let me know and I'll finally take a properly look! :)

D8H:
> It seems you could configure Webpack to export your library as UMD, so it can be consumed from the IDE (as a Node.js module) or from the extension (as a browser global): https://www.typescriptlang.org/docs/handbook/modules.html#code-generation-for-modules

I'm not sure to follow. I think that the library is already a UMD according to the "webpack.config.js".
```
      library: {
        name: "TileMapHelper",
        type: "umd",
      },
```
Do you mean that extra configurations is required for UMD libraries to be used as browser global?
Does "browser global" means stored in a global variable to be used as a namespace?

4ian:
Back to this! I can't find the webpack.config.js you're talking about? Did you commit the files to build the library? I can't seem to find this folder `GDevelop\SharedLibs\TileMapHelper` where to run `npm run build`?

> Do you mean that extra configuration is required for UMD libraries to be used as browser global?

In theory you instruct the bundler (rollup, webpack...) to build your library as UMD (in the config you mentioned) and the name that will be used (TileMapHelper). In the browser, this will be available as a global variable with this name.

> Does "browser global" means stored in a global variable to be used as a namespace?

Yes basically, a global variable.

What is sure is that we'll want to avoid at all cost doing an automation to change the packaged library. We want to have it compiled, probably using esbuild like the rest of the game engine, and then bundled (might be worth looking at rollup instead of webpack, or just pass to esbuild some custom code to add as the ["banner"](https://esbuild.github.io/api/#banner) and "footer") so that it's in a "UMD" library that we can either import or just put in the browser as a file and get a global variable to use.

I can see this is what is already done in TileMapHelper.js, so it seems it's good on this side?
image

SharedLibs/TileMapHelper/webpack.config.js
```diff
@@ -0,0 +1,47 @@
...
+    entry: "./index.ts",
+    output: {
+      library: {
+        name: "TileMapHelper",
+        type: "umd",
+      },
+      filename: "TileMapHelper.js",
+      path: path.resolve(root, "../../Extensions/TileMap/helper/"),
+      globalObject: '(typeof self !== "undefined" ? self : this)',
+    },
```
D8H, line 19: I see it here.

D8H: I can try with esbuild or rollup if you want.

Silver-Streak:
@D8H 

I am working on a point and click example game, and it has made me realize some pretty fundamental issues/enhancements around the hitbox editors for sprites, including no native method for automatic concave polygon splitting.

Before I open up a discussion for a path forward, I wanted to check:

Will the collision mask object for this PR be a whole new user creatable object, or just something autogenerated with tilemaps?

Also, is this going to be adding automatic polygon splitting for concave polygons? I found some js libraries that handle this and was going to include them in the discussion, but if your PR here already covers that I didn't want to repeat work. I couldn't find it in the code as I am way out of practice, but I did see you mention something about it at the start of the pr.

D8H:
I tried Rollup. I can't make it works with PIXI. In the index, when I comment the imports of the classes that use PIXI, it builds the library. The error message give no clue, it's as the ts plugin give the ts files directly to rollup, but the plugin works when there is no dep to PIXI. 
I tried to use tsc + rollup to test without the plugin, but rollup couldn't resolve any import.

I did the PIXI import like this in the library.
```
import * as PIXI from "pixi.js";
```
How should the PIXI dependency be configured to work with Rollup?

D8H:
The library can be build like this (it uses RollUp)
```
cd SharedLibs/TileMapHelper
npm run build
```
But, the produced d.ts file is missing the namespace TileMapHelper. It needs to be added manually which is very not practical. It could be done with a script that remove every occurrence of `declare` and adds one line at the beginning and at the end, but there is probably a better way of doing this.

Extensions/TileMap/helper/TileMapHelper.js
```diff
@@ -0,0 +1,2 @@
...
+//# sourceMappingURL=TileMapHelper.js.map
```
ClementPasteau, line 2: should we add a comment to say where this file is coming from? This could be helpful in the future if we want to update it

Extensions/TileMap/helper/dts/model/TileMapModel.d.ts
```diff
@@ -0,0 +1,330 @@
...
+   * @param tag
+   * @returns The tag that is used to filter tiles.
+   */
+  hasTag(tag: string): boolean;
+  /**
+   * The hitboxes positioning is done by {@link TransformedCollisionTileMap}.
+   * @param tag
+   * @returns The hit boxes for this tile.
+   */
+  getHiBoxes(tag: string): PolygonVertices[] | undefined;
```
ClementPasteau, line 320: typo `HitBoxes`? (is this used?)

Extensions/TileMap/helper/dts/tiled/Tiled.d.ts
```diff
@@ -0,0 +1,339 @@
+import { float, integer } from '../model/CommonTypes';
+/**
+ * Tiled JSON format.
+ */
+export declare type TiledMap = {
+  /** Hex-formatted color (#RRGGBB or #AARRGGBB) (optional) */
+  backgroundcolor?: string;
```
ClementPasteau, line 7: all those properties are lowercase, should we prefer using camelCase? => `backgroundColor`?

Extensions/TileMap/tests/tilemapcollisionmaskruntimeobject.spec.js
```diff
@@ -0,0 +1,229 @@
...
+  let runtimeScene;
+  /**
+   * @type {gdjs.TileMapCollisionMaskRuntimeObject}
+   */
+  let tileMap;
+  beforeEach(async function () {
+    runtimeScene = createScene();
+    tileMap = addTileMapCollisionMask(runtimeScene);
+    // TODO find a clean way to wait for the json to be read.
+    await delay(10);
```
ClementPasteau, line 113: 10 seconds seems like a lot to read a json

SharedLibs/TileMapHelper/karma.conf.js
```diff
@@ -0,0 +1,42 @@
+module.exports = function (config) {
```
ClementPasteau, line 1: do we need a whole different karma conf for the TileMapHelper?

SharedLibs/TileMapHelper/src/render/TileMapManager.ts
```diff
@@ -0,0 +1,137 @@
...
+import { TiledMap } from "../tiled/Tiled";
+import { TiledTileMapLoader } from "../tiled/TiledTileMapLoader";
+import { EditableTileMap } from "../model/TileMapModel";
+import { TileTextureCache } from "./TileTextureCache";
+import { PixiTileMapHelper } from "./TilemapPixiHelper";
+
+import PIXI = GlobalPIXIModule.PIXI;
+
+/**
+ * An holder to share tile maps across the 2 extension objects.
```
ClementPasteau, line 11: nit `an` => `a`

Extensions/TileMap/TileMapRuntimeManager.ts
```diff
@@ -0,0 +1,133 @@
...
+        this._runtimeScene = runtimeScene;
+        this._manager = new TileMapHelper.TileMapManager();
+      }
+
+      /**
+       * @param instanceHolder Where to set the manager instance.
+       * @returns The shared manager.
+       */
+      static getManager(runtimeScene: gdjs.RuntimeScene) {
+        // @ts-ignore
```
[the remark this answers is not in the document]
D8H: It's because this attribute doesn't exist in the runtime scene. The platformer and the pathfinding also use this to share a manager.

Extensions/TileMap/helper/dts/tiled/Tiled.d.ts
D8H: It's actually the definition of the JSON used by Tiled, a tile map editor.

SharedLibs/TileMapHelper/karma.conf.js
D8H: Given that the other Karma run with the bundled lib, I guess it's necessary. This configuration has the advantage of allowing to write tests in TypeScript.

Extensions/TileMap/helper/TileMapHelper.js
D8H: I added a readme:
* https://github.com/D8H/GDevelop/tree/TileMapHitBoxes/Extensions/TileMap/helper

Extensions/TileMap/TileMapRuntimeManager.ts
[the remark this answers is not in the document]
4ian: I think @arthuro555 made something nice in Shift:

```
namespace gdjs {
  export interface RuntimeScene {
    _tweens: Map<string, shifty.Tweenable>;
  }
  ...
```

SharedLibs/TileMapHelper/src/render/TilemapPixiHelper.ts
```diff
@@ -0,0 +1,274 @@
...
+} from "../model/TileMapModel";
+import { TileTextureCache } from "./TileTextureCache";
+
+import PIXI = GlobalPIXIModule.PIXI;
+
+export class PixiTileMapHelper {
+  /**
+   * Parse a Tiled map JSON file,
+   * exported from Tiled (https://www.mapeditor.org/)
+   * into a generic tile map data (`GenericPixiTileMapData`).
```
4ian, line 16: Is this comment still up to date?
(the `GenericPixiTileMapData` part)

SharedLibs/TileMapHelper/src/render/TilemapPixiHelper.ts
```diff
@@ -0,0 +1,274 @@
...
+
+    return textureCache;
+  }
+
+  /**
+   * Re-renders the tilemap whenever its rendering settings have been changed
+   *
+   * @param pixiTileMap
+   * @param tileMap
+   * @param textureCache
```
4ian, line 109: Enter a description for these

SharedLibs/TileMapHelper/src/render/TilemapPixiHelper.ts
```diff
@@ -0,0 +1,274 @@
...
+   * Re-renders the collision mask
+   *
+   * @param pixiGraphics
+   * @param tileMap
+   * @param typeFilter
+   * @param outlineSize
+   * @param outlineColor
+   * @param outlineOpacity
+   * @param fillColor
+   * @param fillOpacity
```
4ian, line 202: Either enter a description or don't list these

SharedLibs/TileMapHelper/src/tiled/Tiled.ts
```diff
@@ -0,0 +1,487 @@
+import { float, integer } from "../model/CommonTypes";
```
4ian, line 1: Could you name this file like "TiledFormat"?

SharedLibs/TileMapHelper/src/tiled/TiledTileMapLoader.ts
```diff
@@ -0,0 +1,162 @@
...
+import {
+  extractTileUidFlippedStates,
+  decodeBase64LayerData,
+} from "./TiledLoaderHelper";
+
+/**
+ * It creates a {@link EditableTileMap} from a Tiled JSON.
+ */
+export class TiledTileMapLoader {
+  static load(pako: any, tiledMap: TiledMap): EditableTileMap | null {
```
4ian, line 17: Any reason why parseAtlas, which also is related to TiledMap, is not in this file?

Extensions/TileMap/collision/TransformedTileMap.ts
```diff
@@ -0,0 +1,746 @@
...
+
+        return {
+          next: () => {
+            let listNext = listItr.next();
+            while (listNext.done) {
+              const layerNext = layerItr.next();
+              if (layerNext.done) {
+                return listNext;
+              }
+              listItr = layerNext.value
```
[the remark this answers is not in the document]
D8H: It's done once per layers. It's just an iteration there is no calculus nor big allocation in `getHitboxes` (we could even make it allocation free if we break the iterable contract by returning a singleton).

Extensions/TileMap/collision/TransformedTileMap.ts
```diff
@@ -0,0 +1,746 @@
...
+        }
+        if (this.layer.isFlippedVertically(this.x, this.y)) {
+          tileTransformation.flipY(height / 2);
+        }
+        if (this.layer.isFlippedDiagonally(this.x, this.y)) {
+          tileTransformation.flipDiagonally();
+        }
+        tileTransformation.preConcatenate(layerTransformation);
+
+        // TODO To handle tile map modification, update lengths.
```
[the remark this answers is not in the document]
D8H: Not soon, at some point, we could try to allow events to modify the map. It would mean a per instance modification and, as there is several object for one tile map, it would need an identifier to tell which instance is the collision mask of a tile map instance.
I'm replacing it with "// The tile map can't change at runtime so the existing arrays can be reused safely."

SharedLibs/TileMapHelper/src/render/TilemapPixiHelper.ts
D8H: Will the documentation generation get the right method signature?

SharedLibs/TileMapHelper/src/tiled/TiledTileMapLoader.ts
D8H: `parseAtla` is is dependent to PIXI and use very few data from the Tiled file. I wanted to keep things separated and wait for a new format to see if we can extract an interface to allow atlas cutting. Because, I guess that having both a parser and a render dependency in the same code is a bad idea.

SharedLibs/TileMapHelper/src/render/TilemapPixiHelper.ts
arthuro555: Yes, that signature information can already be gotten through static analysis, and our current documentation generator relies on typescript to get the function signature.

Extensions/TileMap/tilemapcollisionmaskruntimeobject.ts
```diff
@@ -0,0 +1,490 @@
...
+        // RuntimeObject.hitBoxes point to the same polygons as the hitboxes
+        // from the grid. The hitboxes from the grid are updated according to
+        // the transformation at demand.
+        // This force all the hitboxes to update.
+        // The hitboxes array is built by _updateTileMap().
+      }
+      this.hitBoxesDirty = false;
+      this._renderer.redrawCollisionMask();
+      this.updateAABB();
+    }
```
D8H, line 184: I'm not really proud of this.

Create a `updateAllHitboxes` in the model would need to add a method on each composition level which is not great and it's an operation that should not be seen as the "right way". It's needed only because `RuntimeObject` force to use an `Array`. So I think it's better to keep it outside of the model for clarity.

Looping here with the knowledge of the whole model is clearly not an option.

We could update the `Array` elements of `this.hitboxes`, it's not necessary as the polygon instances don't change, but it can make the code clearer.

Extensions/TileMap/tilemapcollisionmaskruntimeobject.ts
```diff
@@ -0,0 +1,490 @@
...
+        this,
+        runtimeScene
+      );
+      this._updateTileMap();
+
+      // *ALWAYS* call `this.onCreated()` at the very end of your object constructor.
+      this.onCreated();
+    }
+
+    rgbToNumber = function (components: [integer, integer, integer]): integer {
```
4ian, line 79: Could this be a function not part of the class interface?

Extensions/TileMap/tilemapcollisionmaskruntimeobject.ts
```diff
@@ -0,0 +1,490 @@
...
+    updateHitBoxes(): void {
+      this.updateTransformation();
+      // Update the RuntimeObject hitboxes.
+      for (const hitboxes of this._collisionTileMap.getAllHitboxes(
+        this._collisionMaskTag
+      )) {
+        // RuntimeObject.hitBoxes point to the same polygons as the hitboxes
+        // from the grid. The hitboxes from the grid are updated according to
+        // the transformation at demand.
+        // This force all the hitboxes to update.
```
4ian, line 178: Typo: This forces [...]

Extensions/TileMap/collision/TransformedTileMap.ts
```diff
@@ -0,0 +1,747 @@
...
+       */
+      constructor(source: TileMapHelper.EditableTileMap, tag: string) {
+        this._source = source;
+        this.tag = tag;
+        this._layers = new Map<integer, TransformedCollisionTileMapLayer>();
+        for (const sourceLayer of source.getLayers()) {
+          // TODO should there be a list of EditableTileMapLayer to avoid a cast?
+          const tileLayer = sourceLayer as TileMapHelper.EditableTileMapLayer;
+          if (!tileLayer) {
+            // TODO Handle collision mask on object layers?.
```
4ian, line 48: Don't formulate this as a question but rather as an affirmation: "This is skipped because we don't handle collisions on object layers."

Extensions/TileMap/collision/TransformedTileMap.ts
```diff
@@ -0,0 +1,747 @@
...
+     * @see {@link getHitboxesAround} It gives a fast access to hitboxes for collision handling.
+     */
+    export class TransformedCollisionTileMap {
+      /**
+       * The model that describes the tile map.
+       */
+      private _source: TileMapHelper.EditableTileMap;
+      tag: string;
+      private _layers: Map<integer, TransformedCollisionTileMapLayer>;
+      // TODO Tiled allows to offset the layers
```
4ian, line 16: Document this as a limitation/unsupported feature.

Extensions/TileMap/collision/TileMapCollisionMaskRender.ts
```diff
@@ -0,0 +1,87 @@
+namespace gdjs {
+  export namespace TileMap {
+    import PIXI = GlobalPIXIModule.PIXI;
+
+    /**
+     * This render is only useful for debugging purposes.
+     * @see {@link PixiTileMapHelper.updatePixiCollisionMask}, the render used by the GUI.
+     */
+    export class TileMapCollisionMaskRender {
```
4ian, line 9: This is a renderer, not a render, right?

4ian (approved):

Extensions/TileMap/tilemapcollisionmaskruntimeobject.ts
```diff
@@ -76,10 +76,6 @@ namespace gdjs {
       this.onCreated();
     }
 
-    rgbToNumber = function (components: [integer, integer, integer]): integer {
```
4ian: Nice!

Extensions/TileMap/tilemapcollisionmaskruntimeobject.ts
4ian: Could we do a method called "exceptionallyUpdateAllHitboxesEvenIfCostlyToCompute"? A long name while discourage usage and will be easy to find in the codebase should this be tried to be used again.

D8H:
> Could we do a method called "exceptionallyUpdateAllHitboxesEvenIfCostlyToCompute"? A long name while discourage usage and will be easy to find in the codebase should this be tried to be used again.

This is method is overriding the RuntimeObject one it can be renamed.

PR #3321, opened 2021-12-01
Fix tweens automatically deleting the object sometimes affecting newly created objects

### LINKED ISSUE #3316: When object is deleted by tween action, that tween activates on other object instances (Regression in v122, not present in v121)
## Describe the bug

When an object with a Tween is deleted, the Tween becomes active on another instance of that object.

https://user-images.githubusercontent.com/8879811/144277176-4b4f0748-c6ee-40fa-ade0-9e7f3a5dd1a2.mp4

## To Reproduce

0. Install GDevelop v122.  This problem doesn't exist on v121.
1. Create an object and add tween behavior
2. Add a tween action and configure Tween to delete object when complete
3. Object will tween and then delete (properly)
4. However, another instance of that object gets the Tween activated on them (wrongly)

## Other details

Windows 10

4ian:
Fix #3316

### DIFF
--- a/Extensions/TweenBehavior/tweenruntimebehavior.ts
+++ b/Extensions/TweenBehavior/tweenruntimebehavior.ts
@@ -107,6 +107,19 @@ namespace gdjs {
       return true;
     }
 
+    onDestroy() {
+      const shiftyJsScene = this._runtimeScene.shiftyJsScene;
+      if (!shiftyJsScene) return;
+
+      // Stop and delete all tweens of the behavior - otherwise they could:
+      // - continue to point to the behavior, and so to the object (memory leak),
+      // - affect the object in case it's recycled (wrong/hard to debug behavior).
+      for (const identifier in this._tweens) {
+        this._tweens[identifier].instance.stop();
+        shiftyJsScene.remove(this._tweens[identifier].instance);
+      }
+    }
+
     private _addTween(
       identifier: string,
       easingValue: string,


Extensions/TweenBehavior/tweenruntimebehavior.ts
```diff
@@ -107,6 +107,19 @@ namespace gdjs {
       return true;
     }
 
+    onDestroy() {
```
4ian, line 110: @arthuro555 FYI, when deleting an object with a tween behavior, we were not cleaning the tweens running in Shifty on this behavior. It was not easy to spot before the refactor because we were calling removeTween on a tween that was just finishing (so we were not leaking this tween, but potentially leaking *other* tweens going on the object!).

`onDestroy` is guaranteed to be called when the object is removed, and so it's the ideal place to clean all the tweens. So we both fix the issue of the tween wrongly continuing to apply after it's supposed to delete the object, and also fixing the other tweens that could also run on it.

Extensions/TweenBehavior/tweenruntimebehavior.ts
```diff
@@ -107,6 +107,19 @@ namespace gdjs {
...
+      const shiftyJsScene = this._runtimeScene.shiftyJsScene;
+      if (!shiftyJsScene) return;
+
+      // Stop and delete all tweens of the behavior - otherwise they could:
+      // - continue to point to the behavior, and so to the object (memory leak),
+      // - affect the object in case it's recycled (wrong/hard to debug behavior).
+      for (const identifier in this._tweens) {
+        this._tweens[identifier].instance.stop();
```
4ian, line 118: @jeremyckahn Quick question about Shifty (I did not have the time to check that in the source code before going to bed): is calling remove on the Shifty scene not enough to remove properly a tween? I had to call `stop` here (and also at the other place where a tween is removed), otherwise the tween would continue to run despite the call to remove. 

Is this normal/me not reading the docs or a potential bug? :)

jeremyckahn: Hi @4ian, `Tweenable`s are **not** stopped when removed from a `Scene`. They are simply removed from the `Scene`'s data structure: https://github.com/jeremyckahn/shifty/blob/f634dd6f6aac6b25a2fea3d2b2b2184d9ba949ed/src/scene.js#L88-L96

Because the `remove` function returns the removed `Tweenable`, you should be able to do something like `scene.remove(tween).stop()`. Would that meet your needs?

As for whether this is the ideal behavior, it's a tough call. While I can see it being preferable to `stop` a removed tween most of the time, I can also imagine a scenario where that's not desired and the workaround would be a little awkward. The existing API makes it pretty easy to `stop` a removed tween, so it seems safer to leave it as-is.

4ian: Sounds good to leave as is then, I was not sure if this was made on purpose or not! 👍

PR #3324, opened 2021-12-02
Fix debugger actions being always displayed after first use

### LINKED ISSUE #3296: Top-right icons change to debugger icons on non-debugger tab
## Describe the bug

Tested on version 5.0.122

https://user-images.githubusercontent.com/8879811/143534314-d0d6e50b-b653-48ca-bf6b-40c60d8fc114.mp4

## To Reproduce
1. Preview a game
2. Open the Debugger
3. Close preview but leave Debugger tab open
4. Change to a non-Debugger tab
5. Preview a game
6. The top-right icons change to the debugger icons (play/pause/etc)

ClementPasteau:
Fixes #3296 

As the top component doesn't re-render when isActive = false, lifecyclemethods of `Debugger/index.js` will not be triggered and isActive will never go from `true` to `false`

### DIFF
--- a/newIDE/app/src/Debugger/index.js
+++ b/newIDE/app/src/Debugger/index.js
@@ -41,7 +41,6 @@ const isUnavoidableLibraryWarning = ({ group, message }: Log): boolean =>
 type Props = {|
   project: gdProject,
   setToolbar: React.Node => void,
-  isActive: boolean,
   previewDebuggerServer: PreviewDebuggerServer,
 |};
 
@@ -79,8 +78,6 @@ export default class Debugger extends React.Component<Props, State> {
   _debuggerLogs: Map<number, LogsManager> = new Map();
 
   updateToolbar() {
-    if (!this.props.isActive) return;
-
     this.props.setToolbar(
       <Toolbar
         onPlay={() => this._play(this.state.selectedId)}
@@ -102,15 +99,7 @@ export default class Debugger extends React.Component<Props, State> {
   }
 
   componentDidMount() {
-    if (this.props.isActive) {
-      this._registerServerCallbacks();
-    }
-  }
-
-  componentWillReceiveProps(nextProps: Props) {
-    if (nextProps.isActive && !this.props.isActive) {
-      this._registerServerCallbacks();
-    }
+    this._registerServerCallbacks();
   }
 
   componentWillUnmount() {
--- a/newIDE/app/src/MainFrame/EditorContainers/DebuggerEditorContainer.js
+++ b/newIDE/app/src/MainFrame/EditorContainers/DebuggerEditorContainer.js
@@ -24,9 +24,8 @@ export class DebuggerEditorContainer extends React.Component<
   };
 
   shouldComponentUpdate(nextProps: RenderEditorContainerProps) {
-    // Prevent any update to the editor if the editor is not active,
-    // and so not visible to the user.
-    return nextProps.isActive;
+    // We render updates only if the component is active, or becoming active.
+    return this.props.isActive || nextProps.isActive;
   }
 
   getProject(): ?gdProject {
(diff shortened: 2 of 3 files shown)


newIDE/app/src/MainFrame/EditorContainers/DebuggerEditorContainer.js
```diff
@@ -66,6 +66,13 @@ export class DebuggerEditorContainer extends React.Component<
     }
   }
 
+  setDebuggerToolbar = node => {
+    if (!this.props.isActive) {
+      return;
+    }
+    this.props.setToolbar(node);
```
ClementPasteau, line 73: if we use the `setToolbar` function, what is the `updateToolbar` for?

newIDE/app/src/Debugger/index.js
```diff
@@ -107,12 +105,6 @@ export default class Debugger extends React.Component<Props, State> {
     }
   }
 
-  componentWillReceiveProps(nextProps: Props) {
```
4ian, line 110: What about the component did mount? If isActive is false at the mounting of the component, then the server callbacks will never be added.
I guess we should then remove entirely isActive prop from this component if it's not used anymore.

newIDE/app/src/MainFrame/EditorContainers/DebuggerEditorContainer.js
4ian: This function `updateToolbar` is required on the interface of a XxxEditorContainer (i.e: the components being the tabs in `MainFrame`). This imperative method is called when a tab is focused:

See in MainFrame:
image
image

In theory, `updateToolbar` is only called when the tab is focused, so it's safe to ask the children to update the toolbar (See `if (this.editor) this.editor.updateToolbar();` in this file).

Where when the children trigger an update to the toolbar, it will usually not cause any issue because this usually happen after a user interaction... but here the debugger can react to an outside signal. And so it would call setToolbar even if it's not active! 
The MainFrame will not protect against this and will happily replace the toolbar with the last one that called setToolbar.

So I wonder if the real fix is not to be done in MainFrame actually:
* Search for `setToolbar: setEditorToolbar`
* We should replace this by a call to `setEditorToolbar` ONLY if the tab is active. Otherwise we ignore it.
* Other calls to setEditorToolbar looks safe.

4ian:
Just a note to show the usefulness of isActive:

Here I have open the start page, the scene editor, the scene events, the debugger, and external events.
I then close the project manager. Because it's a component very high in the tree, it makes a re-render of a big part of the app. You can see that in the profiler:

image

But you can see that while we re-renderered the Project Manager drawer, the toolbar, the tabs and the start page:
image

The rest (the scene editor, the scene events, the debugger and the external events) are NOT updated:
image

If I switch from the start page to the debugger, you can see that only: the top level stuff, the start page (that I'm quitting) and the Debugger (that I'm switching to) are rendered:

image

Notice the two "colored bars" which are the start page and the debugger :) The scene editor, scene events and external events are untouched!

You can also notice that the Project Manager drawer starts to be rendered, but then stops because we have a `shouldComponentUpdate` down in the tree that stops the rendering if the project manager is closed:

image

That's it! Not linked to your issue, but important to keep these `shouldComponentUpdate` in each tab "container".

newIDE/app/src/MainFrame/EditorContainers/DebuggerEditorContainer.js
```diff
@@ -24,9 +24,8 @@ export class DebuggerEditorContainer extends React.Component<
   };
 
   shouldComponentUpdate(nextProps: RenderEditorContainerProps) {
-    // Prevent any update to the editor if the editor is not active,
-    // and so not visible to the user.
-    return nextProps.isActive;
+    // We render updates only if the component is active, or becoming active.
```
4ian, line 27: or if the component is becoming inactive (which is what solves our problem here ;)). Important to call this out.

```suggestion
    // We stop updates when the component is inactive. 
    // If it's active, was active or becoming active again we let update propagate.
    // Especially important to note that when becoming inactive, a "last" update is allowed.
```

PR #3325, opened 2021-12-02
Fix sound not playing after the first time played

### LINKED ISSUE #3315: "Play sound" action skips if used repeatedly in 5.0.122
## Describe the bug
It appears that the "Play Sound" action is skipping repeated uses in 5.0.122. It will play the sound file correctly the first time, then not play anything the second or third time. This might be related to the default volume that was set to fix the other known sound issues?

Interestingly, "Play Sound on a channel" is not impacted at all, so this exists as a valid workaround. I have confirmed this behavior doesn't occur if reverting to .121 or older.

## To Reproduce
1. Add an event with a trigger condition (press any key + trigger once, timer, etc) and the "Play a sound" action.
2. Preview the game.
3. Trigger the sound. The first time will play, the 2nd and 3rd time will not, then it will start playing normally again.

Another user reported it and provided a test project, I cleaned it up a bit after being able to reproduce it, available here: [Audio Test.zip](https://github.com/4ian/GDevelop/files/7635320/Audio.Test.zip)

## Other details
OS: Windows 10 Pro 
GD5 Version: 5.0.122
This was reported in this thread: https://forum.gdevelop-app.com/t/audio-playback-bug/36625 
Thanks to user Rasterisko for their sample video and test project.

While obviously this should be fixed, I wonder if this is a good instance to mark this event (and the play music events) as legacy, to guide users to use the play sound on a channel events instead. 

No strong feelings either way, just a thought since the sound channel is better for 99.9% of scenarios than non-sound channel, and play sound file is better than play music in most cases as well, this could lead to better utilization by newer users.

ClementPasteau:
Fixes #3315 

Not sure I understand why this fixes it, there are some listeners on those arrays that were probably impacted when this was moved later in the function.
Tested with the audio-playground in the repo and all sounds work great now 👍

### DIFF
--- a/GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
+++ b/GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
@@ -618,6 +618,7 @@ namespace gdjs {
         loop,
         pitch
       );
+      this._storeSoundInArray(this._freeSounds, sound);
       sound.once('play', () => {
         if (this._paused) {
           sound.pause();
@@ -625,7 +626,6 @@ namespace gdjs {
         }
       });
       sound.play();
-      this._storeSoundInArray(this._freeSounds, sound);
     }
 
     playSoundOnChannel(
@@ -644,6 +644,7 @@ namespace gdjs {
         loop,
         pitch
       );
+      this._sounds[channel] = sound;
       sound.once('play', () => {
         if (this._paused) {
           sound.pause();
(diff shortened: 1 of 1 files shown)


4ian:
> Not sure I understand why this fixes it, there are some listeners on those arrays that were probably impacted when this was moved later in the function.

I guess this is related to a race condition between the time the sound starts to "play" (the event "play") and the fact that we store it in the array of all free sounds/musics where we try to reuse the first slot available. The race condition must somehow trigger the same slot to be used, or something like this.

PR #3332, opened 2021-12-04
Improve extensions updating

arthuro555:
- Show a badge in the project manager when extensions are updateable:
Badge in PM
- Add option to update an extension or open its store page (install dialogue):
When update available
When an update is not available
- Show on the install dialogue whether you are updating or reinstalling the extension:
image
image

A potential issue is that the rebuilding of the updates list is currently done every time the MainFrame is rendered. This shouldn't be a huge issue as it isn't too expensive, but it's something to look for. I am unsure how to do it otherwise since afaik there is no real way currently to "watch" the updating of events functions (to update the update state whenever the version of one changes).


4ian (commented):
Interesting stuff :) 
A few things from the perspective of the interface:

- we should probably have the badge shown on each extension? Otherwise you need to click every single extension to find which one to update. Not a great user experience.

A few things that comes to my mind when reading the code:

- I know that `contextType` is convenient, but it's also not scalable. We only have one other place where it's used - and where it should be removed. I would prefer to rework the components needing them to be function component with hooks and use `useContext` (that implies refactoring, but is good for the long term health of the codebase).
- You've moved the ExtensionStoreStateProvider down in the tree inside MainFrame, which is something we want to avoid at all cost to avoid complexifying MainFrame. You can also see that this is a symptom of a more important problem: this ExtensionStoreStateProvider is now doing 2 things:
  - it's responsible for fetching and giving to you (you = the consumer) access to extensions (like before, which is fine).
  - but it's now also responsible for reading the current project extension list and computing which extensions are to be upated.
  
  This second responsibility should not be beared by this provider, because it's specific to the project (especially if we rework the editor one day to open multiple projects). Instead, it's probably a helper hook or component or function that should be used by the ProjectManager?
  
- I see why you made _ProjectManager, but same thing here we would rather need to use a function component rather than dividing the component in two.
- More important, I think that redoing the calculation at every render of MainFrame can't be accepted. MainFrame is re-rendered for things as simple as a toolbar button being toggled between enable and disabled.

For this last point, this means that if we want to keep a silky smooth 60 fps interface, we need to do all re-renders in <16ms. Which is basically 8ms for us and 8ms for React and the browser to do their stuff. If we start doing things like recomputing the list of extensions to update at every single re-render, including even when the project manager is closed, we easily add 0.05ms on a recent, fairly expensive computer. This means 0.3ms on a cheap computer. If we accept to do this for 15 features, we now have added 4.5ms and we're way out of budget.
This is called "death by a thousand cuts" basically :)

That's a lot but I think it's important we get this right!

arthuro555:
> I know that contextType is convenient, but it's also not scalable. We only have one other place where it's used - and where it should be removed. I would prefer to rework the components needing them to be function component with hooks and use useContext (that implies refactoring, but is good for the long term health of the codebase).
> [...]
> I see why you made _ProjectManager, but same thing here we would rather need to use a function component rather than dividing the component in two.

:+1: I was thinking about doing that but was also thinking that it may be risky to refactor such a core component on the IDE. If you think it is better that way though, I can do it.

> More important, I think that redoing the calculation at every render of MainFrame can't be accepted. 
> [...]
> We want to avoid at all cost to avoid complexifying MainFrame

I agree, hence why i stated it as an issue in the PR comment, but I do not see how to get around that. We need to update the update state each time an events functions extension is modified, as otherwise we'll get "zombie notifications" that continue showing up after updating the extension or changing the version number manually to match the latest one. As project elements are not observable (which btw i proposed to change alongside enforcing the creation of an undo stack entry for each project mutation before), we need to either do this or prop drill a callback to all components that may mutate an events functions extension or it's version to rebuild it in MainFrame. I think the current version is cleaner and complexifies MainFrame less than the prop drill alternative. I still believe the true solution would be making project state observable to avoid prop drilling and only rendering components that care about a part of the project state changing, not their common ancestor. 

We could make a custom hook that builds the update state from the project instance and extension store context to decomplexify MainFrame, but then the problem of zombie notifications come back: we need to know when the events functions extensions are mutated or rebuild the state each frame, and since MainFrame is their common ancestor that will happen each time MainFrame is rerendered except it'll be multiplied by each mounted component that has the hook and additionally happen when the component rerenders without MainFrame rerendering which sounds like a worse idea.

> that should be used by the ProjectManager

Note that not only the ProjectManager but also the extension store uses the update state

> This second responsibility should not be beared by this provider

That is a good point, I think update state should probably be moved to it's own context then.

> we should probably have the badge shown on each extension? Otherwise you need to click every single extension to find which one to update. Not a great user experience.

That's what the little dot on the right menu button is for ;) It is hidden by the context menu on most screenshots but you can see it in the first one.

arthuro555:
So... Do you have an idea on how to fix that or should the PR be abandoned as unacceptable because of the too frequent computations?

4ian:
Sorry for the late reply. For now, could we only do that at the opening of the project manager and at the opening of an extension context menu/dialog maybe?

AlexandreSi:
What I see can be done with the latest modifications on the ProjectManager:

In terms of features:

- Have the badge with the number of extensions to update on the title `Functions/Behaviors`
- Have a badge on each updatable extension
- When clicking on one of those extensions, it now opens the ExtensionInstallDialog - so no need to change anything here (or maybe the naming of the button).
  - Maybe we could add more info on this dialog, something like: ~~`1.0.0`~~ => `1.1.0`. I think we should add this information, because the user should see something that explains why there was a badge on this extension.

And how to do it:

- Extract the Functions/Behaviors `ProjectStructureItem` into a new component that will use the `ExtensionStoreContext` and the `EventsFunctionsExtensionsContext`
- Add an effect to this new component so that it checks how many extensions are to be updated once mounted.

PR #3339, opened 2021-12-06
Fix light textures sometimes not behaving properly when close to obstacles

### LINKED ISSUE #3224: Light with Texture doesn't work properly when touching Light obstacle object
## Describe the bug
Light with texture doesn't work properly when coming in contact with Light obstacle object
https://gfycat.com/singlecriminalabalone

image

## To Reproduce
1. Add Light
2. Set it's texture (I set light radius to 50 and created a yellow circle sprite in Piskel)
3. Create an object and add Light Obstacle Behaviour
4. When two come in contact, the light doesn't work like it's supposed to (Shown in the video)

## Other details
* Windows 8.1 x64
* GDevelop Desktop build 121

ClementPasteau:
Fixes #3224 

Fixed by not setting the Fragment Color when outside the texture coordinates.
I'm not an GLSL expert though, can this cause issues somewhere else?

image

### DIFF
--- a/Extensions/Lighting/lightruntimeobject-pixi-renderer.ts
+++ b/Extensions/Lighting/lightruntimeobject-pixi-renderer.ts
@@ -546,9 +546,11 @@ namespace gdjs {
   varying vec2 vPos;
 
   void main() {
-      vec2 topleft = vec2(center.x - radius, center.y - radius);
-      vec2 texCoord = (vPos - topleft)/(2.0 * radius);
-      gl_FragColor = vec4(color, 1.0) * texture2D(uSampler, texCoord);
+    vec2 topleft = vec2(center.x - radius, center.y - radius);
+    vec2 texCoord = (vPos - topleft)/(2.0 * radius);
+    gl_FragColor = (texCoord.x > 0.0 && texCoord.x < 1.0 && texCoord.y > 0.0 && texCoord.y < 1.0)
+      ? vec4(color, 1.0) * texture2D(uSampler, texCoord)
+      : vec4(0.0, 0.0, 0.0, 0.0);
   }`;
   }
 


Bouh (approved):
Lgtm, discard just cancel the pixel rendering it's the more straightforward solution 👍

4ian:
I think that works but I'd like some GLSL specialist to tell us if that's the most optimised way to do it.

Bouh:
I'm not a glsl expert too, I think in that case, we could change the Alpha to 0 for the pixels outside of the UVs (texture coordinates).
Or/and use a Clamp instead of an `if` statement?

ClementPasteau:
We can easily avoid the `discard()` function but I'm not sure we can avoid the `if` ?

tested & works:
```
void main() {
      vec2 topleft = vec2(center.x - radius, center.y - radius);
      vec2 texCoord = (vPos - topleft)/(2.0 * radius);
      if (texCoord.x > 0.0 && texCoord.x < 1.0 && texCoord.y > 0.0 && texCoord.y < 1.0) {
        gl_FragColor = vec4(color, 1.0) * texture2D(uSampler, texCoord);      
      }
    }
```

Bouh:
```
void main() {

      vec4 rgba = vec4(color, 0.0);
      vec2 topleft = vec2(center.x - radius, center.y - radius);
      vec2 texCoord = (vPos - topleft)/(2.0 * radius);
      if (texCoord.x > 0.0 && texCoord.x < 1.0 && texCoord.y > 0.0 && texCoord.y < 1.0) {
             rgba =  vec4(color, 1.0);
      } else {
             rgba =  vec4(color, 0.0);
      }
      gl_FragColor = rgba * texture2D(uSampler, texCoord);
    }
```

ClementPasteau:
The app doesn't crash when the condition is false 🤔 

I've been trying to stress test an example with all those different GLSL snippets of code, by adding lots of lights, whilst monitoring the FPS. 
But I can't seem to stress it enough, the FPS always stays max.
Do you know any high-consuming example in terms of light?

4ian:
Best bet would be to try on a mobile phone, but it's unlikely you can measure differences like this (well, maybe unless you use thousands of lights, with not a lot of obstacles to avoid the game to be CPU bounded).

4ian:
I would be more comfortable always setting the gl_FragColor and just adjust the opacity if we're outside the bounds :)

ClementPasteau:
new suggestion (which works) -> always set the gl_FragColor but instead of using the texture if we're out of bounds, just set the FragColor to a default vec4

4ian (approved):

4ian:
lgtm!

PR #3340, opened 2021-12-06
Renovate home page 🪚

AlexandreSi:
Change start page to:
- Display starters for new users
- Display new tutorials
- Display latest games made with GDevelop
- Display button to open user's profile

Also:
- Move folder selection to another dialog that will host all the future project settings (project name for example).
- Remove Starters tab from the "Create Project" dialog
- Rename Startpage to Homepage
- Add split menu to `Open Project` button with recent projects

Screenshots:
- Desktop:
  image

- Mobile
  image

- Click on showcased game
  image

- Recent files as split menu
  image

- Check Storybook for error handling:
  - On Carousel : https://gdevelop-storybook.s3.amazonaws.com/refurbish-homepage-for-good/latest/index.html?path=/story/carousel--with-error
  - On Home Page : https://gdevelop-storybook.s3.amazonaws.com/refurbish-homepage-for-good/latest/index.html?path=/story/homepage--with-server-side-errors


AlexandreSi:
TODO before this start page is released:
- Change starter thumbnails to be a 16/9 image ? Maybe we should change the API to return a dict `[size: string]: url` instead of a list in `previewImageUrls`

tristanbob:
This is a great improvement!  However, seeing those words so close together without any borders looks messy to me.

Can you find a way to visually separate each of the 3 horizontal rows?  Perhaps a shaded background for each row? 

Also, please feel free to ignore this request as I have no formal skills in UI design.  :)

newIDE/app/src/CommandPalette/CommandsList.js
```diff
@@ -112,7 +112,7 @@ const commandsList: { [CommandName]: CommandMetadata } = {
     area: 'PROJECT',
     displayText: t`Apply changes to the running preview`,
   },
-  OPEN_START_PAGE: { area: 'IDE', displayText: t`Open start page` },
+  OPEN_START_PAGE: { area: 'IDE', displayText: t`Show Home` },
```
ClementPasteau, line 115: change key as well?

newIDE/app/src/GamesShowcase/ShowcasedGameButtons.js
```diff
@@ -0,0 +1,138 @@
...
+      {/* $FlowFixMe */}
+      <StackLayout {...stackLayoutProps}>
+        {firstLinks.map((link, index) => (
+          <LinkButton key={index} link={link} showcasedGame={showcasedGame} />
+        ))}
+      </StackLayout>
+      {otherLinks.length > 0 && (
+        <>
+          {(windowWidth === 'small' || forceColumn) && <Spacer />}
+          {/* $FlowFixMe */}
```
ClementPasteau, line 122: and that

newIDE/app/src/GamesShowcase/ShowcasedGameButtons.js
```diff
@@ -0,0 +1,138 @@
...
+import Twitch from '../UI/CustomSvgIcons/Twitch';
+import ItchIo from '../UI/CustomSvgIcons/ItchIo';
+import Microsoft from '../UI/CustomSvgIcons/Microsoft';
+import Apple from '../UI/CustomSvgIcons/Apple';
+import { useResponsiveWindowWidth } from '../UI/Reponsive/ResponsiveWindowMeasurer';
+
+const getFullUrl = url =>
+  url.startsWith('/') ? 'https://gdevelop-app.com' + url : url;
+
+const getLinkIconAndLabel = (url: string, type: string) => {
```
ClementPasteau, line 30: should we change this to a swtich/case? I find it easier to read knowing it's based only on 1 attribute `type`

4ian:
> Can you find a way to visually separate each of the 3 horizontal rows? Perhaps a shaded background for each row?

I think this is a good remark, we probably can add simply more visual space between each row of thumbnails - that's what is done on YouTube/Twitch for example. This will help to associate each title to the thumbnail row and provide a better separation :)

newIDE/app/src/MainFrame/ElectronMainMenu.js
```diff
@@ -193,7 +193,7 @@ const buildAndSendMenuTemplate = (
         enabled: !!project,
       },
       {
-        label: i18n._(t`Show Start Page`),
+        label: i18n._(t`Show Home`),
         onClickSendEvent: 'main-menu-open-start-page',
```
ClementPasteau, line 197: @4ian I suppose we shouldn't change the event id for tracking purpose?

newIDE/app/src/MainFrame/index.js
```diff
@@ -1417,7 +1424,7 @@ const MainFrame = (props: Props) => {
       setState(state => ({
         ...state,
         editorTabs: openEditorTab(state.editorTabs, {
-          label: i18n._(t`Start Page`),
+          label: i18n._(t`Home`),
           projectItemName: null,
           renderEditorContainer: renderStartPageContainer,
```
ClementPasteau, line 1429: change container name to `renderHomePageContainer`?

newIDE/app/src/MainFrame/index.js
```diff
@@ -2199,10 +2207,29 @@ const MainFrame = (props: Props) => {
...
                   onCreate: () => openCreateDialog(),
+                  onCreateFromExampleShortHeader: onCreateFromExampleShortHeader,
+                  onOpenFromExampleShortHeader: async (
+                    storageProvider,
+                    fileMetadata
+                  ) => {
+                    await getStorageProviderOperations(storageProvider);
+                    const state = await openFromFileMetadata(fileMetadata);
+
+                    if (state) {
```
ClementPasteau, line 2218: you can probably only have one if:
`if (state && state.currentProject)`

newIDE/app/src/stories/componentStories/UserChip.stories.js
```diff
@@ -0,0 +1,27 @@
+// @flow
```
ClementPasteau, line 1: I know this PR is already long, but should we move the existing `UserPublicProfileChip` to this file too?
I think they're quite similar and it'd be good to have them next to each other

newIDE/app/.gitignore
```diff
@@ -16,7 +16,6 @@ build
 
 # misc
 .DS_Store
-.env
```
Bouh, line 19: If we want to leak the keys why not, but I think it's unexpected.

arthuro555:
Personally i would a list of last edited projects too if it is not empty as after the first time, that's what most users will really care about (opening their project fast).

Midhil457:
A few opinions:
- Imo, I think the `Open projects manager` and `Open a project` buttons should be given more importance. The main and the most important use of the Home page is to give quick and easy access to those buttons. Right now, it feels like the buttons are tucked away to give way for the carousel. Maybe seperate them into seperate reactangles. Something like this: (I am just trying to prove a point)
Hm

- I don't see a point in having the game showcase in the Start Page. The other two make sense as they give easy access to the examples and tutorials. I don't think it's the right place to promote the games made with GDevelop
- Maybe the carousel is given a bit too much emphasis and I don't think the main menu is the right place to give this much importance to it. User might get overwhelmed due to the amount of colors and info that will be given to them when they open GDevelop and drive thier attension from the main things of the Start Page (The open buttons and controls)

I am not an UI designer or anything, and I might be biased as the current Start Page is what I have been seeing for a while. I am just trying to convey my opinions

:)

newIDE/app/src/CommandPalette/CommandsList.js
AlexandreSi: I don't know if it's worth changing all the references to the start page for "home" since it could be seen only as a labelling issue. Start page is still easy to me. But if it bugs you, I can do it

newIDE/app/src/GamesShowcase/ShowcasedGameButtons.js
```diff
@@ -0,0 +1,138 @@
...
+    stackLayoutProps = { ...stackLayoutProps, noMargin: true };
+  } else {
+    stackLayoutProps = { ...stackLayoutProps, noColumnMargin: true };
+  }
+
+  const firstLinks = links.slice(0, 3);
+  const otherLinks = links.slice(3);
+  return (
+    <>
+      {/* $FlowFixMe */}
```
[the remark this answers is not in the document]
AlexandreSi: I made this component following our discussion to have an option to force the column layout. However, `ResponsiveLineStackLayout` and `ColumnStackLayout` don't have the same props so Flow raises errors on the use of `stackLayoutProps`. It seemed unsolvable to me at the moment

newIDE/app/src/MainFrame/index.js
AlexandreSi: I copy-pasted without further thinking, good idea!

newIDE/app/src/stories/componentStories/UserChip.stories.js
AlexandreSi: Oh good idea, will do!

newIDE/app/.gitignore
AlexandreSi: @4ian told me that this file is not used anymore, if you have one, I think you can delete it.

newIDE/app/src/GamesShowcase/ShowcasedGameButtons.js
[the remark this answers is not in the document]
AlexandreSi: I will try a bit harder to find a solution

newIDE/app/.gitignore
Bouh: Oh ok sorry

4ian: Yep it's gone for long! :)

newIDE/app/src/GamesShowcase/ShowcasedGameButtons.js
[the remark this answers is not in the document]
4ian: Maybe the only good solution is to make a component for it, which can be a few lines:

`const LineOrColumnStackLayout = ({forceColumn, children}: {|...|}) => forceColumn ? <ColumnStackLayout>{children}</ColumnStackLayout> : <ResponsiveLineStackLayout>{children}</ResponsiveLineStackLayout>`
?

newIDE/app/src/MainFrame/EditorContainers/StartPage/index.js
```diff
@@ -63,190 +74,356 @@ type Props = {|
...
+              )}
+            </>
+          )}
+        </I18n>
+      );
+    }
+  ),
+  // Prevent any update to the editor if the editor is not active,
+  // and so not visible to the user.
+  (prevProps, nextProps) => nextProps.isActive
```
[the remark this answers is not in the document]
4ian: Might be a good idea to start doing this for all editors as we go, to avoid stale props (though for the StartPage the risk seems low)

newIDE/app/src/UI/ClosableTabs.js
```diff
@@ -19,7 +19,7 @@ const styles = {
     width: '100%',
     position: 'relative',
     textAlign: 'initial',
-    height: '100%',
+    minHeight: 0,
```
4ian, line 22: Let's verify in all editors (scene editor, events editor, debugger, extension, external layout, external events) that they still display properly :)

newIDE/app/src/Utils/GDevelopServices/Tutorial.js
```diff
@@ -2,14 +2,14 @@
 import axios from 'axios';
 import { GDevelopAssetApi } from './ApiConfigs';
 
-export type Tutorial = {
+export type Tutorial = {|
```
4ian, line 5: Note that in theory we can't guarantee that the backend won't one day return an additional key.

newIDE/app/src/stories/componentStories/Carousel.stories.js
```diff
@@ -90,76 +84,76 @@ export const Showcases = () => {
     {
       id: 'lil-bub-s-hello-earth',
       title: "Lil BUB's HELLO EARTH",
-      onClick: () => console.log('Clicked on lil-bub-s-hello-earth'),
+      onClick: () => action('Clicked on lil-bub-s-hello-earth'),
```
4ian, line 87: 👍

newIDE/app/src/Utils/GDevelopServices/Tutorial.js
AlexandreSi: Yes of course, but we should tell Flow that we are confident the type will be strictly this. I think it's a different issue that Flow cannot do anything about

4ian:
Thanks both for your feedbacks :) 

I agree this is a big change so might take a bit of time for everyone to adapt! 
I also agree that existing users have a clear use case : "I want to open my project, and do it fast". For this:
- We could add a split button like this, with a menu showing the recent projects:
  image
- But we already have multiple things: the command bar has a open recent project (which I use personally often), there is the Ctrl+O shortcut (desktop app) and the File > Open Recent menu too (desktop app). 

I think this is almost independent from the redesign, in the sense that this new home page is not making easier nor harder to open a project. Sure the button is on the top right, but I don't see it as a big deal (it's still blue, i.e: a primary button).
I don't think someone who made and saved a project will have a hard time finding it, and as I said we can always add a split button with a context menu :) But this is a new feature!

If it's really too painful, we could always think later of adding a carousel with your most recent projects actually :) But I think this is not needed for now, let's see when using this home page if we really need that.

For "having the game showcase in the Start Page", this is indeed new, but I think this is interesting to give ideas and give a bit more emphasis to games made with GDevelop. Note that it's at the bottom of the screen, so the user eye will first scan the top of the page which is about 1) template and 2) tutorials.

For "User might get overwhelmed due to the amount of colors and info", I think we need larger spaces between each carousel row. The idea is for this page *to be* visually "full of colors", in the sense that we want to show 4 things:

1. Look at all the nice templates to get you started (give inspiration).
2. Look at all the tutorials, you'll never be lost (show that tutorials are existing and not far at all).
3. Look at all the nice games, you can do it and have your game listed here too! (give hope, inspiration, give motivation and recognition to creators).
4. And finally note all the "secondary buttons" : you can obviously open a project, or create one from scratch.

The point 4 is for existing users or power users. The points 1, 2 and 3 are for new users... and existing users too! (with this, we give more emphasis to new templates, new tutorials, and new games).

A few final notes, you can note:
- we don't have the logo anymore. It's because you're here to build games, not to look at a big logo.
- we don't have a click anymore to do to see the templates or examples. The existing start page is bad from a UX point of view because it forces you to do a click before even seeing what you can do with GDevelop. 

> I am not an UI designer or anything, and I might be biased as the current Start Page is what I have been seeing for a while. I am just trying to convey my opinions

No problem, appreciate your feedback! I think you and Arthuro have good points about opening recent projects, but I also think this is a use case that can be solved with a split button (without eating the space given to the rest) or if really needed in the future a new carousel (but I would really wait to see if it's useful... as power users we can use the command bar if we want to be quick).
I think if we add some large spaces this screen will be less "compact" and should be a good first screen to see what's going on in the GDevelop world and that could be useful for everyone :)

newIDE/app/src/GamesShowcase/ShowcasedGameButtons.js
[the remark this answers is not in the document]
AlexandreSi: Found a way to do it!

newIDE/app/src/MainFrame/index.js
AlexandreSi: When changing it, I realized that the formatting is prone to confusion: the second if is a oneliner.

newIDE/app/src/MainFrame/ElectronMainMenu.js
4ian: Let's keep the same yes!

newIDE/app/src/MainFrame/index.js
4ian: Good catch, let's keep the exact same logic as now to avoid a edge case or risk any change in the behavior

newIDE/app/src/MainFrame/EditorContainers/StartPage/index.js
```diff
@@ -401,21 +462,25 @@ export const StartPage = React.memo<Props>(
...
-                      outputPath
-                    )
+                    selectedExample
+                      ? onCreateFromExampleShortHeader(
+                          setIsOpening,
+                          onOpenProjectJustCreated(true)
+                        )(i18n, selectedExample, outputPath)
+                      : onCreateBlank(onOpenProjectJustCreated(false))(
+                          i18n,
+                          outputPath
```
4ian, line 482: Could we investigate using the promise rather than a callback?

`onCreateBlank(i18n, outputPath).then(onOpenProjectJustCreated(true))`

or even 
```
await onCreateBlank(i18n, outputPath);
onOpenProjectJustCreated(true);
```

looks more readable to me and I think the callback is not providing more value than a promise that would return an object (rather than passing arguments to the callback)?

newIDE/app/src/MainFrame/EditorContainers/StartPage/index.js
```diff
@@ -82,6 +85,10 @@ type Props = {|
     storageProvider: empty,
     fileMetadata: empty
   ) => Promise<void>,
+
+  // Blank project creation
+  onCreateBlank: OnCreateBlankFunction,
+  onOpenBlank: (storageProvider: empty, fileMetadata: empty) => Promise<void>,
```
4ian, line 91: I fail to see what's the difference between onOpenBlank from onOpenFromExampleShortHeader... the args are the same and I wonder if it should be just one thing? Basically it's telling the MainFrame "please open this file with this storage" (which is often a file path + the LocalFileStorageProvider on the desktop app, or a URL + the UrlStorageProvider on the web-app), so it should work the same 🤔.

newIDE/app/src/ProjectCreation/services/BrowserCreation.js
```diff
@@ -0,0 +1,55 @@
...
+  onOpenCallback(project, null, null);
+};
+
+export const onCreateFromExampleShortHeader = (
+  isOpeningCallback: boolean => void,
+  onOpenCallback: (
+    storageProvider: StorageProvider,
+    fileMetadata: FileMetadata
+  ) => void
+) => async (i18n: I18nType, exampleShortHeader: ExampleShortHeader) => {
```
4ian, line 35: See my other comment, could we:
- go from a callback to a promise with a returned object?
- collapse the high order function into a single function with an object as arg: `{ setIsOpening, i18n, exampleShortHeader }`

Could simplify how we use these functions :)

newIDE/app/src/MainFrame/EditorContainers/StartPage/index.js
4ian: (There might be a difference, but it's unclear to me so far)

Midhil457:
> For "User might get overwhelmed due to the amount of colors and info", I think we need larger spaces between each carousel row. The idea is for this page to be visually "full of colors", in the sense that we want to show 4 things:
Look at all the nice templates to get you started (give inspiration).
Look at all the tutorials, you'll never be lost (show that tutorials are existing and not far at all).
Look at all the nice games, you can do it and have your game listed here too! (give hope, inspiration, give motivation and recognition to creators).
And finally note all the "secondary buttons" : you can obviously open a project, or create one from scratch.

I like the idea. But, won't comparision arise? Like, a user might be discouraged by the games in the showcase too. Also, when opening GDevelop and starting to work on a project, the user would most likely trying to get focus and work on it for a few hours at min. And the color full home screen might be a focus puller. Though I agree it might take a few months for users to adapt, so these issues might not be an issue anymore

AlexandreSi:
@4ian Due to sleeplessness, I could gather your last feedbacks and simplify A LOT the logic around project creation and then their opening.

newIDE/app/src/UI/ClosableTabs.js
AlexandreSi: Looks okay to me!

4ian (approved):
Seems good to go in this state!

4ian:
This being said, we need to fix the thumbnails for the examples/"starters", but I think @Bouh was on it?

Bouh:
> This being said, we need to fix the thumbnails for the examples/"starters", but I think @Bouh was on it?

Yes I've started working on it, there are two thumbnails to finish, I'll do it later this evening.

AlexandreSi:
> This being said, we need to fix the thumbnails for the examples/"starters", but I think @Bouh was on it?

I'll work on adding a description to starters as well

newIDE/app/src/MainFrame/EditorContainers/HomePage/index.js
```diff
@@ -0,0 +1,565 @@
...
+        forceUpdateEditor,
+      }));
+
+      const windowWidth = useResponsiveWindowWidth();
+      const authenticatedUser = React.useContext(AuthenticatedUserContext);
+      const { getRecentProjectFiles } = React.useContext(PreferencesContext);
+      const [
+        recentProjectFiles,
+        setRecentProjectFiles,
+      ] = React.useState<?Array<FileMetadataAndStorageProviderName>>(null);
```
4ian, line 154: @AlexandreSi I think you might not need a state for this, `getRecentProjectFiles` is safe to read at any time and should always return an array, even if empty.

newIDE/app/src/MainFrame/EditorContainers/HomePage/index.js
```diff
@@ -0,0 +1,565 @@
...
+      const [
+        selectedExample,
+        setSelectedExample,
+      ] = React.useState<?ExampleShortHeader>(null);
+
+      const buildRecentProjectFilesMenuTemplate = React.useCallback(
+        (recentProjectFiles: Array<FileMetadataAndStorageProviderName>) => (
+          i18n: I18nType
+        ) =>
+          recentProjectFiles.map(file => ({
```
4ian, line 214: Simply call `getRecentProjectFiles` inside here? So:
- you pay for this call only when the menu is built (when the user click on the split button)
- better, you are sure this menu is up to date if a game was opened/closed in the meantime (since the Home page was loaded).

AlexandreSi: I don't know why I have a tendency to create factories. Will pay more attention to this in the future!
Anyway, it is changed now.

ClementPasteau (approved):
lgtm!

D8H:
A lot of users publish their games on itch.io and they ask a 1.26 ratio image:

> The cover image is used whenever itch.io wants to link to your project from another part of the site. Required (Minimum: 315x250, Recommended: 630x500)

What about using the same ratio in GDevelop?
I think it will make life easier for example contributors and examples will be more likely to have a cover.

AlexandreSi:
Could be a good idea!
Nevertheless, Gamejolt uses ratio around 16/9 and Steam has lots of formats, including 16/9.
I can't find any reason why they chose this particular ratio that doesn't seem to be a standard (around 5/4 nonetheless).

For reference, this is how a 1.26 ratio image would render in the current carousel: 
image

AlexandreSi:
Or we could make sure the image is contained but it would mean adding a background color:
image

blurymind:
Holly crap, was Dead estate and Elect head made in Gdevelop :open_mouth:

4ian:
@blurymind these are just images to compare the size with games on Steam/other platforms :)

PR #3341, opened 2021-12-07
Improve Export flow

ClementPasteau:
As part of improving the export flow, the first step is to improve the toolbar by making the preview & publish actions more central and more visible/intuitive for everyone.

Can be tested on the storybook, but here are some screenshots of the dev environment:
image

image

A second step is to improve the home screen of the publication flow.
This improvement has in mind the idea of:
- Making the "1-click web export in a few seconds" more visible, directly on the first publish screen
- Making this export also more available to everyone, by removing the limit of exports
- Clarify the other export options for beginners by splitting them into "assisted exports" / "manual exports"
  - assisted exports will focus on packaging the game for the user, and will have tutorials on how to upload to itch.io & co / how to publish an app to Steam / how to publish on the play store / how to upload to Facebook Instant games
  - manual exports will be for "advanced users", with a minimal description focusing on exporting the game files for desktop/mobile/html.
  
Some screenshots (not entirely finished):
image
image
image
image


arthuro555:
Does the publish button really need such focus? It is one you rarely use after all, compared to the rest of them.

ClementPasteau:
Thanks for taking a look @arthuro555 . 🙂
As you mentioned, the publish action is less used than the rest, however that's what we're trying to change!
We believe that, if we show to newcomers how easy publishing is (with the hosted web version in particular), by:
- Making the link more accessible
- Improving the export flow to be more intuitive
- Probably also removing the limits of builds on the hosted web version

Then we will have more users exporting their games & interested in sharing what they create to their friends & networks.

In the end, this benefits GDevelop as it makes it visible to more and more people!

Fun story: I discovered last week that there was an export button in the toolbar, I was always doing File>Export...
So I think it deserves to be put forward!

newIDE/app/src/MainFrame/Toolbar/PreviewAndPublishButtons.js
```diff
@@ -0,0 +1,192 @@
...
+                    previewState.overridenPreviewExternalLayoutName,
+              },
+            ]}
+          />
+          <FlatButton
+            onClick={exportProject}
+            disabled={!hasProject}
+            icon={
+              <img
+                alt={t`Publish`}
```
4ian, line 172: This won't work, you must use i18n._ to transform `t` (which gives an object) to a string

newIDE/app/src/MainFrame/Toolbar/index.js
```diff
@@ -1,101 +1,92 @@
...
+export type ToolbarInterface = {|
+  setEditorToolbar: any => void,
 |};
 
-export class MainFrameToolbar extends PureComponent<Props, State> {
-  state = {
-    editorToolbar: null,
-  };
+export default React.forwardRef<MainFrameToolbarProps, ToolbarInterface>(
+  function LayerEffectNameField(props: MainFrameToolbarProps, ref) {
```
4ian, line 29: Name to update ;)

newIDE/app/src/stories/componentStories/MainFrameToolbar.stories.js
```diff
@@ -0,0 +1,157 @@
...
+  <MainFrameToolbar
+    {...defaultProps}
+    isPreviewEnabled
+    hasProject
+    previewState={{
+      isPreviewOverriden: false,
+      overridenPreviewExternalLayoutName: null,
+      overridenPreviewLayoutName: null,
+      previewExternalLayoutName: null,
+      previewLayoutName: 'test',
```
4ian, line 78: Nitpicking: try to give names like "fake layout", "fake external layout".
This can be useful to see if there was a mistake in a component (out of my mind, as an example: if a story is showing "The scene fake external layout is launched" rather than "The scene test is launched")

tristanbob:
I agree with the comments that the "10 seconds away from having your game online" is a HUGE unique feature of GDevelop that should get more attention.  I think many users don't even realize that it is available!

One suggestion for improvement, let's use a more standard icon for "publish":

https://materialui.co/icon/publish

Screenshot 2021-12-07 093632

newIDE/app/src/MainFrame/Toolbar/PreviewAndPublishButtons.js
```diff
@@ -0,0 +1,192 @@
...
+                  />
+                }
+                label={
+                  hasPreviewsRunning ? (
+                    <Trans>Update</Trans>
+                  ) : (
+                    <Trans>Preview</Trans>
+                  )
+                }
+                exceptionalTooltipForToolbar={<Trans>{previewTooltip}</Trans>}
```
4ian, line 107: Don't use `Trans` like this. Ask @AlexandreSi if he remembers why ;) 
By doing this, you generate a translation which is `<0/>`, i.e: a unique variable to translate. Compare this with: `Hello {name}` => this will generate a translation in French of `Bonjour {name}` (and maybe a different word order in another language). But how should `{name}` be translated in French? It should not! That's the same here => you probably want to either pass an object resulting from a `t` macro (i.e: a MessageDescriptor) and you use i18n inside the button OR (better I think here), you pass something that is already considered translated (i.e: a translated string with `i18n._` or a `Trans`, though I'm not sure tooltips work with components, just with strings)

newIDE/app/src/MainFrame/Toolbar/PreviewAndPublishButtons.js
```diff
@@ -0,0 +1,192 @@
...
+            ]}
+          />
+          <ElementWithMenu
+            element={
+              <FlatButton
+                onClick={onHotReloadPreview}
+                disabled={!isPreviewEnabled}
+                icon={
+                  <img
+                    alt={t`Preview`}
```
4ian, line 83: Same as below, this must be a string, not an object.

4ian (commented):
Added a few various comments :)

4ian:
> One suggestion for improvement, let's use a more standard icon for "publish":

Agree! Almost forgot to add this comment! Currently the icon is really outdated. I'll be looking at if we can have a new one designed! I was also thinking of something that would remind of an app icon on an app store (sure you can publish on the web, but I try to convey the idea of "packaging into a "real" app", even if it's on the web)

tristanbob:
I like "cloud upload" even better than "publish":

Screenshot 2021-12-07 095814

https://materialui.co/icon/cloud-upload

Midhil457:
> Fun story: I discovered last week that there was an export button in the toolbar, I was always doing File>Export...
So I think it deserves to be put forward!

Wait, there is? :confused:. Oh

Now, I do agree that it needs more focus and also agree with tristan that the icon is awful at showing it's use

ClementPasteau:
Compact mode is an interesting idea! 
We could indeed imagine having an option in the Preferences>Appearance section for hiding the buttons text.

newIDE/app/src/Export/BrowserExporters/BrowserOnlineWebExport.js
```diff
@@ -61,7 +64,11 @@ export const browserOnlineWebExportPipeline: ExportPipeline<
 
   renderHeader: () => <ExplanationHeader />,
 
-  renderLaunchButtonLabel: () => <Trans>Publish online</Trans>,
+  renderLaunchButtonLabel: () => <Trans>Generate link</Trans>,
+
+  customStepsProgress: (build: ?Build, loading: boolean) => (
```
4ian, line 69: Call this `renderCustomStepsProgress` (a function should be a verb, unless it's a Component, and this verb is often `render` for things returning `React.Node`s)

newIDE/app/src/GameDashboard/GameRegistration.js
```diff
@@ -69,13 +68,13 @@ export const GameRegistration = ({
...
       const { id } = profile;
       setError(null);
-      setUnavailableReason(null);
       try {
         const game = await getGame(
           getAuthorizationHeader,
           id,
           project.getProjectUuid()
         );
+        setUnavailableReason(null);
```
4ian, line 77: Why this change? (curious, I've not thought about it)

newIDE/app/src/UI/Messages/InfoBar.js
```diff
@@ -1,55 +1,39 @@
...
-  }
-}
+  React.useEffect(
+    () => {
+      if (show) {
+        setTimeout(() => {
+          hideInfoBar();
+        }, 3000);
+      }
+    },
```
4ian, line 23: In theory you should clear the timeout in the effect cleanup to avoid calling the callback after the component is already gone ("memory" leak)

newIDE/app/src/UI/Messages/InfoBar.js
```diff
@@ -1,55 +1,39 @@
...
 
 type Props = {|
-  identifier: AlertMessageIdentifier,
   message: React.Node,
   touchScreenMessage?: React.Node,
   show: boolean,
+  hideInfoBar: () => void,
 |};
 
-export default class InfoBar extends React.PureComponent<Props> {
```
AlexandreSi, line 18: One class down 📈

newIDE/app/src/UI/Messages/DismissableInfoBar.js
```diff
@@ -0,0 +1,51 @@
...
+  show: boolean,
+|};
+
+const DismissableInfoBar = ({
+  identifier,
+  show,
+  touchScreenMessage,
+  message,
+}: Props) => {
+  const preferencesContext = React.useContext(PreferencesContext);
```
AlexandreSi, line 24: To rename `preferences` for consistency's sake

newIDE/app/src/UI/Messages/InfoBar.js
```diff
@@ -1,55 +1,39 @@
...
-} from '../../MainFrame/Preferences/PreferencesContext';
-import { ScreenTypeMeasurer } from '../Reponsive/ScreenTypeMeasurer';
+import { useScreenType } from '../Reponsive/ScreenTypeMeasurer';
 
 type Props = {|
-  identifier: AlertMessageIdentifier,
   message: React.Node,
   touchScreenMessage?: React.Node,
   show: boolean,
+  hideInfoBar: () => void,
```
AlexandreSi, line 10: For consistency, you can rename it just `hide`.

newIDE/app/src/UI/Toolbar.js
```diff
@@ -8,8 +8,8 @@ type ToolbarProps = {|
 
 type ToolbarGroupProps = {|
   children?: React.Node,
-  firstChild?: true, // Not used, but could be useful one day for styling
-  lastChild?: true, // Not used, but could be useful one day for styling
+  firstChild?: boolean,
+  lastChild?: boolean,
```
AlexandreSi, line 12: 👌

newIDE/app/src/UI/FlatButton.js
```diff
@@ -22,22 +24,31 @@ type Props = {|
...
  */
 export default class FlatButton extends React.Component<Props, {||}> {
   render() {
-    const { label, primary, icon, keyboardFocused, ...otherProps } = this.props;
+    const {
+      label,
+      primary,
+      icon,
+      keyboardFocused,
+      exceptionalTooltipForToolbar,
```
AlexandreSi, line 40: This `exceptional` props is a bit scary. Can you add documentation to explain why it is here?

newIDE/app/src/Profile/CreateProfile.js
```diff
@@ -12,11 +12,17 @@ type Props = {
   message?: React.Node,
   onLogin: () => void,
   onCreateAccount: () => void,
+  justifyContent?: 'center',
```
AlexandreSi, line 15: It feels weird to have just one possibility. It's the same as having true/false. I don't know if it's a good practice to have it like that.
But i'm not clear on what would be a good solution given Line prop `justifyContent: string`

newIDE/app/src/MainFrame/Toolbar/index.js
```diff
@@ -1,101 +1,92 @@
...
+            canDoNetworkPreview={props.canDoNetworkPreview}
+            isPreviewEnabled={props.isPreviewEnabled}
+            previewState={props.previewState}
+            hasPreviewsRunning={props.hasPreviewsRunning}
+            exportProject={props.exportProject}
+            hasProject={props.hasProject}
           />
         </ToolbarGroup>
-        {this.state.editorToolbar || <ToolbarGroup />}
+        {editorToolbar || <ToolbarGroup />}
```
AlexandreSi, line 88: You add a toolbarGroup to even the display? The `lastChild` is not useful here?

newIDE/app/src/MainFrame/Toolbar/PreviewAndPublishButtons.js
```diff
@@ -0,0 +1,202 @@
...
+  isPreviewEnabled,
+  hasPreviewsRunning,
+  previewState,
+  setPreviewOverride,
+  exportProject,
+  hasProject,
+}: PreviewAndPublishButtonsProps) {
+  return (
+    <ThemeConsumer>
+      {muiTheme => (
```
AlexandreSi, line 44: Same as https://github.com/4ian/GDevelop/pull/3341/files#r765974848

newIDE/app/src/MainFrame/Toolbar/PreviewAndPublishButtons.js
```diff
@@ -0,0 +1,202 @@
...
+                  ) : (
+                    <Trans>
+                      Launch a preview of the scene, right click for more
+                    </Trans>
+                  )
+                }
+              />
+            }
+            openMenuWithSecondaryClick
+            buildMenuTemplate={(i18n: I18nType) => [
```
AlexandreSi, line 121: What about extracting these 2 buildMenuTemplate into callbacks to improve the component's readability ?

newIDE/app/src/Export/LocalExporters/index.js
```diff
@@ -9,108 +7,49 @@ import { localFacebookInstantGamesExportPipeline } from './LocalFacebookInstantG
...
+import { onlineElectronExporter } from '../GenericExporters/OnlineElectronExport';
+import { electronExporter } from '../GenericExporters/ElectronExport';
 
-export const getLocalExporters = (): Array<Exporter> => [
+export const localOnlineWebExporter: Exporter = {
+  ...onlineWebExporter,
+  exportPipeline: localOnlineWebExportPipeline,
+};
+
+export const localAssistedExporters: Array<Exporter> = [
```
AlexandreSi, line 23: Looks cleaner like that! 🧹

newIDE/app/src/Profile/CreateProfile.js
4ian: I've used this in another component in the sense that it's telling you "either don't specify it (probably by default)" or you can specify it but it's so exceptional that I give you just one possibility (center the text).
In the sense: please don't think of abusing this to use "flex-end", that's a no! :D

newIDE/app/src/UI/Messages/DismissableInfoBar.js
4ian: Correct, context does not appear in the variable name (much like we don't call a `useState` variable `somethingState` and `setSomethingState` but just `something/setSomething`).

newIDE/app/src/Export/GenericExporters/OnlineWebExport.js
```diff
@@ -2,13 +2,86 @@
...
+    : getBuildArtifactUrl(build, 's3Key') || '';
+
+  const onOpen = () => {
+    if (buildPending) return;
+    Window.openExternalURL(value);
+  };
+
+  const onCopy = () => {
+    if (buildPending) return;
+    navigator.clipboard.writeText(value);
```
AlexandreSi, line 53: I've never used it but we seem to have a Clipboard class for that

newIDE/app/src/Export/ExportDialog/index.js
```diff
@@ -0,0 +1,198 @@
...
+import { Tab, Tabs } from '../../UI/Tabs';
+import ExportHome from './ExportHome';
+
+const styles = {
+  icon: { width: 40, height: 40 },
+  disabledItem: { opacity: 0.6 },
+  content: { padding: 8 },
+};
+
+export type ExporterSection = 'assisted' | 'manual' | '';
```
AlexandreSi, line 22: Is the value `''` useful? It may look like a hack to not handle a possibly null value in a component state.

newIDE/app/src/Export/BrowserExporters/index.js
```diff
@@ -9,109 +7,50 @@ import { browserHTML5ExportPipeline } from './BrowserHTML5Export';
...
+import { onlineElectronExporter } from '../GenericExporters/OnlineElectronExport';
+import { electronExporter } from '../GenericExporters/ElectronExport';
 
-export const getBrowserExporters = (): Array<Exporter> => [
+export const localOnlineWebExporter: Exporter = {
+  ...onlineWebExporter,
+  exportPipeline: browserOnlineWebExportPipeline,
+};
+
+export const localAssistedExporters: Array<Exporter> = [
```
AlexandreSi, line 23: Oh this is even cleaner than I expected!

AlexandreSi:
A few things here and there, but looks good to me!

newIDE/app/src/Export/GenericExporters/OnlineWebExport.js
4ian: Good observation. Speaking of which, it should probably be adapted to use the navigator.clipboard APIs now that they are implemented everywhere :) 
Note that the existing Clipboard is unfortunately limited to objects. We could rework it to also expose simple read/write texts apis.

4ian: (we can add a TODO)

arthuro555: Btw the clipboard class needs to be updated to use the web clipboard API if it can, right now it uses either the electron API or stores and read from a javascript variable if not in electron, making the contents of the clipboard not pasteable somewhere else on the web editor.

newIDE/app/src/GameDashboard/GameRegistration.js
ClementPasteau: I've noticed that the call to game can happen multiple times sometimes (probably a hidden problem?). 
In this case, I didn't have access to the game, so the unavailableReason was 'unauthorized', meaning the value will switch from 'null' to 'unauthorized' making the component blink from 'something to display' to 'loader' multiple times.

newIDE/app/src/MainFrame/Toolbar/index.js
ClementPasteau: hmm I've just tested without, it's not enough -> in order for the Preview buttons to be centered, it needs 2 Flex elements, one on left & one on right.

4ian (approved):
Nothing really blocking now I think! Great job on this :)

AlexandreSi (approved):

PR #3342, opened 2021-12-07
Improve resilience of resources fetching of the desktop app

4ian:
Do more requests at the same time but retry the failing ones.

Don't show in changelog

### DIFF
--- a/newIDE/app/src/ProjectsStorage/ResourceFetcher/LocalResourceFetcher.spec.js
+++ b/newIDE/app/src/ProjectsStorage/ResourceFetcher/LocalResourceFetcher.spec.js
@@ -103,7 +103,10 @@ describe('LocalResourceFetcher', () => {
       onProgress,
     });
 
-    // Verify that download was done and reported as failed.
+    // Verify that download was done and reported as failed, even after 2 tries.
+    expect(
+      optionalRequire.mockElectron.ipcRenderer.invoke
+    ).toHaveBeenCalledTimes(2);
     expect(
       optionalRequire.mockElectron.ipcRenderer.invoke
     ).toHaveBeenCalledWith(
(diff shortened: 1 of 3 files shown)


newIDE/app/src/ProjectsStorage/ResourceFetcher/LocalResourceFetcher.spec.js
```diff
@@ -115,4 +118,57 @@ describe('LocalResourceFetcher', () => {
...
+      () => false
+    );
+    mockFn(
+      optionalRequire.mockElectron.ipcRenderer.invoke
+    ).mockImplementationOnce(() =>
+      Promise.reject(new Error('Fake download failure'))
+    );
+    mockFn(
+      optionalRequire.mockElectron.ipcRenderer.invoke
+    ).mockImplementationOnce(() => Promise.resolve());
```
ClementPasteau, line 144: that's a super cool test :) 
from the doc, you can apparently chain `mockImplementationOnce()`
so:
```
mockFn(
      optionalRequire.mockElectron.ipcRenderer.invoke
    ).mockImplementationOnce(() =>
      Promise.reject(new Error('Fake download failure'))
    ).mockImplementationOnce(() => Promise.resolve());
```

ClementPasteau (approved):
This looks cool, however I'm not entirely sure why this is needed? What were the problems before?

newIDE/app/src/ProjectsStorage/ResourceFetcher/LocalResourceFetcher.spec.js
4ian: Done 👍

4ian:
> This looks cool, however I'm not entirely sure why this is needed? What were the problems before?

Right sorry I've not even explained.
I've increased the number of parallel file downloads done when opening on the desktop app a project that has external URLs (basically: an example) (the web-app does not download external URLs and just them as is - apart when you export the game, but the desktop app for now prefers to download everything at the beginning to handle only local files).
Problem: sometimes when you do too many parallel downloads, some might fail (that's the reason why I used a promise pool originally: using a Promise.all would hammer your computer/the server asking them to fetch all resources at once). I think 50 is a fairly safe number, and if some resources are failing, you are anyway given the choice to retry. See https://gdevelop-storybook.s3.amazonaws.com/master/latest/index.html?path=/story/resourcefetcher-resourcefetcherdialog--with-errors

BUT this is not a good user experience to fail "too easily". So I added this retry so we at least try to fetch another time a resource that would be failing.

So we have both: fast fetching of resources (up to 50 requests at once), fairly resilient (intermittent request failing? Just retry) and if it's still not working the user will be asked to retry or abandon.

4ian:
Merging because at least one CI pass and the rest is because of a separate install issue.

Issue #3219, opened 2021-10-31
Improvements to Instances List


tristanbob:
## Description

The instances list provides great information but is missing some basic capabilities. 

![image](https://user-images.githubusercontent.com/8879811/139564107-d6dd03fa-9472-4fe7-b5e5-da9fe5680d49.png)

## Solution suggested

- Add the ability to sort by clicking on the names of the columns (clicking twice reverses the sort)
- Add the ability to select an instance and hit the delete key (and confirm delete)
- Add the ability to lock/unlock by clicking on the lock icon
- Add the ability to multi-select (great for mass deletion)
--Not sure what to do with Properties panel... either disable it or make changes apply to all selected items?

4ian:
All good ideas!

> --Not sure what to do with Properties panel... either disable it or make changes apply to all selected items?

That's already how the properties panel is working ;) You can select multiple instance sand changes will apply to all of them.

Bouh:
Adressed in #3343

PR #3343, opened 2021-12-08
Instances List improvements

rs4231199:
for #3219

- [x] Add the ability to sort by clicking on the names of the columns (clicking twice reverses the sort)
- [ ] ~Add the ability to select an instance and hit the delete key (and confirm delete)~
- [x] Add the ability to lock/unlock by clicking on the lock icon
- [x] Add the ability to multi-select (great for mass deletion)

### DIFF
--- a/newIDE/app/src/InstancesEditor/InstancesList/index.js
+++ b/newIDE/app/src/InstancesEditor/InstancesList/index.js
@@ -6,6 +6,7 @@ import {
   Table as RVTable,
   Column as RVColumn,
 } from 'react-virtualized';
+import KeyboardShortcuts from '../../UI/KeyboardShortcuts';
 import ThemeConsumer from '../../UI/Theme/ThemeConsumer';
 import SearchBar, { useShouldAutofocusSearchbar } from '../../UI/SearchBar';
 const gd /*TODO: add flow in this file */ = global.gd;
@@ -17,7 +18,7 @@ type State = {|
 type Props = {|
   instances: gdInitialInstancesContainer,
   selectedInstances: Array<gdInitialInstance>,
-  onSelectInstances: (Array<gdInitialInstance>) => void,
+  onSelectInstances: (Array<gdInitialInstance>, boolean) => void,
 |};
 
 type RenderedRowInfo = {
@@ -48,6 +49,10 @@ export default class InstancesList extends Component<Props, State> {
   instanceRowRenderer: ?typeof gd.InitialInstanceJSFunctor;
   table: ?typeof RVTable;
   _searchBar = React.createRef<SearchBar>();
+  _keyboardShortcuts = new KeyboardShortcuts({
+    isActive: () => false,
+    shortcutCallbacks: {},
+  });
 
   componentDidMount() {
     if (useShouldAutofocusSearchbar() && this._searchBar.current)
@@ -86,7 +91,11 @@ export default class InstancesList extends Component<Props, State> {
 
   _onRowClick = ({ index }: { index: number }) => {
     if (!this.renderedRows[index]) return;
-    this.props.onSelectInstances([this.renderedRows[index].instance]);
+
+    this.props.onSelectInstances(
+      [this.renderedRows[index].instance],
+      this._keyboardShortcuts.shouldMultiSelect()
+    );
   };
 
   _rowGetter = ({ index }: { index: number }) => {
@@ -107,7 +116,7 @@ export default class InstancesList extends Component<Props, State> {
 
   _selectFirstInstance = () => {
     if (this.renderedRows.length) {
-      this.props.onSelectInstances([this.renderedRows[0].instance]);
+      this.props.onSelectInstances([this.renderedRows[0].instance], false);
     }
   };
 
@@ -129,7 +138,11 @@ export default class InstancesList extends Component<Props, State> {
       <ThemeConsumer>
         {muiTheme => (
           <div style={styles.container}>
-            <div style={{ flex: 1 }}>
+            <div
+              style={{ flex: 1 }}
+              onKeyDown={this._keyboardShortcuts.onKeyDown}
+              onKeyUp={this._keyboardShortcuts.onKeyUp}
+            >
               <AutoSizer>
                 {({ height, width }) => (
                   <RVTable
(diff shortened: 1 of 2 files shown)


arthuro555:
GitHub automatically requests a review when opening a non-draft PR because of the code owners file

rs4231199:
> Thanks for the PR! You've asked for the review already, do you want me to check this now or later when adding the other improvements?
@4ian 
I think there is no common code between this functionality and the rest. So, a review now is possible.
Or you can do it later. It doesn't really matter. I hope to push more changes today. :)

rs4231199:
> Add the ability to lock/unlock by clicking on the lock icon

Should I add unlock emoji(:unlock: ) so that users can click on it to lock the instance?

image

Img: This is what it looks like.

4ian:
@rs4231199 Good question, I would say consider:
- Either move to using a (small) `IconButton` with an icon from Material UI like LockOpen and Lock:
  image
  
- Or use a unlock emoji *but* use some opacity so that it's not hard to find which things are locked from those who are not.

But I think ideally a IconButton with `size="small"` would be good. Because an emoji is not supposed to be clickable, there won't be any support for hover, press, keyboard focus, i.e: it would not be accessible :)

newIDE/app/src/SceneEditor/index.js
```diff
@@ -546,13 +546,15 @@ export default class SceneEditor extends React.Component<Props, State> {
 
   _onSelectInstances = (
     instances: Array<gdInitialInstance>,
+    multiSelect: boolean,
     centerView: boolean = true
```
4ian, line 550: Seems that this 3rd argument is unused (always true), we might think of removing it.

4ian (commented):
Looks good so far! :)

newIDE/app/src/SceneEditor/index.js
rs4231199: Yes, It wasn't used anywhere. I removed it.

arthuro555:
> I was thinking of implementing this by maintaining some history that would allow undo and redo operations.

I started a discussion about the undo stack here: https://github.com/4ian/GDevelop/discussions/2739#discussioncomment-1794562

4ian:
@rs4231199 You're right that this is a problem because we're in the case of editors sharing the same "data" and so undo/redo is not consistent.
I would argue that we don't want to address this in this PR. Let's merge this even if we do not handle deletion (or if we do, without undo/redo for now). 

We can think about how to better handle undo/redo across editors separately (the discussion that Arthuro555 linked :))

rs4231199:
I left a console.log statement in the code by mistake. Otherwise it is
ready for a review.

Do we want a ctrl+A(select all) shortcut? Users can first filter instances
using search bar situated below instances list. Then they can select all
filtered instances.
It would be useful if users want to lock/unlock all instances containing a
certain keyword(let's say "button").

newIDE/app/src/InstancesEditor/InstancesList/index.js
```diff
@@ -105,20 +122,86 @@ export default class InstancesList extends Component<Props, State> {
...
+        const compString = (x: string, y: string): number => {
+          if (typeof x === 'string' && typeof y === 'string') {
+            x = x.toLowerCase();
+            y = y.toLowerCase();
+          }
+
+          if (x < y) return direction * 1;
+          if (x > y) return direction * -1;
+          return 0;
+        };
```
4ian, line 169: Move this out of the function and out of the component. This will help reducing the pressure of the JS engine garbage collector by avoiding to re-declare a function.
Also use no abbreviations: `compareStrings`.

newIDE/app/src/InstancesEditor/InstancesList/index.js
```diff
@@ -105,20 +122,86 @@ export default class InstancesList extends Component<Props, State> {
...
+  _renderLockCell = ({ rowData }: { rowData: RenderedRowInfo }) => {
+    return (
+      <IconButton
+        size="small"
+        onClick={() => {
+          rowData.instance.setLocked(!rowData.instance.isLocked());
+        }}
+      >
+        {rowData.instance.isLocked() && <Lock />}
+        {!rowData.instance.isLocked() && <LockOpen />}
```
4ian, line 134: Use a ternary operator :)

newIDE/app/src/InstancesEditor/InstancesList/index.js
```diff
@@ -69,7 +82,7 @@ export default class InstancesList extends Component<Props, State> {
         this.renderedRows.push({
           instance,
           name,
-          locked: instance.isLocked() ? '🔒' : '',
+          locked: instance.isLocked() ? 'true' : 'false',
```
4ian, line 85: Could you store the boolean directly instead?

newIDE/app/src/InstancesEditor/InstancesList/index.js
```diff
@@ -105,20 +122,86 @@ export default class InstancesList extends Component<Props, State> {
...
+          case 'x':
+            return direction * (parseFloat(a.x) - parseFloat(b.x));
+          case 'y':
+            return direction * (parseFloat(a.y) - parseFloat(b.y));
+          case 'angle':
+            return direction * (parseFloat(a.angle) - parseFloat(b.angle));
+          case 'layer':
+            return compString(a.layer, b.layer);
+          case 'locked':
+            return compString(a.locked, b.locked);
```
4ian, line 183: See my other comment, prefer two booleans ;)

4ian (commented):
I've left a few comments. Once they are addressed and the console.log removed, I think it's good to go!

> Do we want a ctrl+A(select all) shortcut? Users can first filter instances
using search bar situated below instances list. Then they can select all
filtered instances.

Do you think it's easy to implement cleanly? If yes we could have it. Otherwise I would keep it for another PR!

rs4231199:
> > Do we want a ctrl+A(select all) shortcut? Users can first filter instances
> > using search bar situated below instances list. Then they can select all
> > filtered instances.
> 
> Do you think it's easy to implement cleanly? If yes we could have it. Otherwise I would keep it for another PR!

Not implementing it, because `ctrl+A` is used to select the text from search bar.

PR should be good to merge now.

4ian (approved):
LGTM! Thanks for working on this :)

PR #3346, opened 2021-12-10
Clean up after new homepage

AlexandreSi:


newIDE/app/src/MainFrame/EditorContainers/HomePage/index.js
```diff
@@ -264,6 +265,7 @@ export const HomePage = React.memo<Props>(
         setPreCreationDialogOpen(false);
         setOutputPath(computeDefaultProjectPath());
         onOpenProjectAfterCreation({ project, storageProvider, fileMetadata });
+        setIsOpening(false);
```
4ian, line 268: You should also set this to false if there is an early return (this could happen if the project can't be created for some reasons, like can't create the path on desktop).

4ian (approved):

newIDE/app/src/stories/componentStories/LocalProjectPreCreationDialog.stories.js
```diff
@@ -0,0 +1,35 @@
+// @flow
```
ClementPasteau, line 1: 😍

Issue #3344, opened 2021-12-09
Collision mask is not being considered exactly as it was made


Rasterisko:
### DESCRIBED THE BUG

It seems that the collision system is not considering exactly how the mask was made, but by completing a rectangle with it.

### TO REPRODUCE

- Create a object with platform character behavior and another object with platform behavior .

- Make a complex colision box like this and try to interact them:

![Colision](https://user-images.githubusercontent.com/95868625/145429265-31026d5f-69ac-4729-8e6d-fa51bb699a76.png)

and it looks like it's happening this:

![Colision edited](https://user-images.githubusercontent.com/95868625/145430322-eeb94581-906c-4585-967a-500a3af29a74.png)

### OTHER DETAILS

- Arthuro555 described in this post briefly what may be happening: https://forum.gdevelop-app.com/t/problems-with-the-collision-mask/36848/14

- GDevelop 5.0.123 on Windows 10

4ian:
Custom hitboxes are indeed not supported by the platformer object behavior - only the rectangle surrounding the red polygons will be used. By the way your polygon looks concave which is also not supported in the game engine (platformer engine or not)?

Only custom hit boxes on the platforms are followed exactly.

AlexandreSi:
> Custom hitboxes are indeed not supported by the platformer object behavior

Maybe I don't fully understand what "support" means, but if you specify a custom hitbox that is a rectangle, it works just fine. Maybe we should add a warning, similar to #3348.

AlexandreSi:
We could even add on the visual editor, for objects with platformer behavior, another rectangle that represents the rectangle surrounding all the masks that will be used when the game is running?

tristanbob:
The physics behavior uses a non-standard collision mask that is configured inside the behavior properties ("Behaviors" tab). 

Perhaps it makes sense to "show" the platformer collision mask on the behavior tab?  (with a note that explains what it shows, and to modify the collision masks using "Properties -> Edit collision masks".

arthuro555:
> Maybe I don't fully understand what "support" means, but if you specify a custom hitbox that is a rectangle

My understanding is that only the AABB is supported, not the hitbox itself. So if you would make a triangular hitbox, which should be valid, it'll keep on using object bounds instead of the actual hitbox:
![image](https://user-images.githubusercontent.com/19349038/145718925-f17105b6-9ca0-4ba3-8758-dfc58036a6c1.png)

AlexandreSi:
[...]

PR #3348, opened 2021-12-10
Add warning when editing the collision mask of a platformer object

### LINKED ISSUE #3312: "Flip object horz." on platformer allows jumping after wall contact
This is a very complex bug and I can't say what could be the origin for it. I feel like it has something to do with the platformer character checking ground while inside of ground.

## Describe the bug

A platformer character is granted another jump when not in contact with a wall/platform anymore if the player is using a custom (smaller) hitbox and is flipped horizontally.

## To Reproduce
1. Create a "player" object
2. Change/Decrease hitbox size

image

3. Give it the **platformer character** behavior
4. Create a "wall" object
5. Give it the **platform** behavior
6. Add following events:
> *Left key is pressed, player is not horizontally flipped > flip hor.: yes*
> *Right key is pressed, player is horizontally flipped > flip hor.: no*

image

7. Start preview
8. Jump at wall (hitbox player **is** in collision with hitbox platform)
9. Flip the character / move away from wall while being airborne (hitbox player **not** in collision with hitbox platform)
10. Jump again (without "touching" ground/floor)
*(Go to 8.)*

## Note

- Smaller/custom hitbox **is mandatory** and won't work without!
- Flipping character **is mandatory** and won't work without!
- Triggering "flip: yes" and "flip: no" only once will not prevent it

## Video

https://user-images.githubusercontent.com/83909121/144070267-b130ab3a-2acb-4f4f-a895-469c736bfd27.mp4

image

## All Events

image

### Game controls

**Move:** Arrow keys *(platformer default)*
**Jump:** Space, Shift *(platformer default)*
**Toggle debug boxes:** Left Control (LControl)
**Restart scene:** R
**Exit game:** Escape (Esc)

### Project files

.json Project file (should download assets upon opening): [game (3).zip](https://github.com/4ian/GDevelop/files/7626658/game.3.zip)
Web example: https://game-previews.gdevelop-app.com/1638283835649-844836/index.html

## Other details
GDevelop 5.0.122 based on GDevelop.js 4.0.99-0-release
PC / Desktop GDevelop version, Windows
Regular build, no nightly version

AlexandreSi:
Closes #3312

https://user-images.githubusercontent.com/32449369/145590989-eaa6fe57-2794-4779-a161-9ee9d3165410.mp4

### DIFF
--- a/newIDE/app/src/ObjectEditor/Editors/SpriteEditor/CollisionMasksEditor/index.js
+++ b/newIDE/app/src/ObjectEditor/Editors/SpriteEditor/CollisionMasksEditor/index.js
@@ -76,9 +76,14 @@ const CollisionMasksEditor = (props: Props) => {
     spriteIndex
   );
 
-  const objectHasPlatformerBehavior = props.object.hasBehaviorNamed(
-    'PlatformerObject'
-  );
+  const objectHasPlatformerBehavior = props.object
+    .getAllBehaviorNames()
+    .toJSArray()
+    .map(behaviorName => props.object.getBehavior(behaviorName))
+    .some(
+      behavior =>
+        behavior.getTypeName() === 'PlatformBehavior::PlatformerObjectBehavior'
+    );
 
   const updateCollisionMasks = React.useCallback(
     () => {
--- a/newIDE/app/src/ObjectEditor/Editors/SpriteEditor/CollisionMasksEditor/index.js
+++ b/newIDE/app/src/ObjectEditor/Editors/SpriteEditor/CollisionMasksEditor/index.js
@@ -24,6 +24,7 @@ import EditorMosaic, {
 } from '../../../../UI/EditorMosaic';
 import { useResponsiveWindowWidth } from '../../../../UI/Reponsive/ResponsiveWindowMeasurer';
 import Background from '../../../../UI/Background';
+import AlertMessage from '../../../../UI/AlertMessage';
 const gd: libGDevelop = global.gd;
 
 const horizontalMosaicNodes: EditorMosaicNode = {
@@ -73,6 +74,10 @@ const CollisionMasksEditor = (props: Props) => {
     animationIndex,
     directionIndex,
     spriteIndex
+  );
+
+  const objectHasPlatformerBehavior = props.object.hasBehaviorNamed(
+    'PlatformerObject'
   );
 
   const updateCollisionMasks = React.useCallback(
@@ -269,6 +274,18 @@ const CollisionMasksEditor = (props: Props) => {
               </Line>
             </React.Fragment>
           )}
+          {!!sprite &&
+            !sprite.isCollisionMaskAutomatic() &&
+            objectHasPlatformerBehavior && (
+              <AlertMessage kind="warning">
+                <Trans>
+                  You currently use a custom collision mask for an object that
+                  has the Platformer behavior. To prevent issues when flipping
+                  object horizontally while grabbing an edge, make sure your
+                  collision mask does not move when flipping it.
+                </Trans>
+              </AlertMessage>
+            )}
           {!sprite && (
             <EmptyMessage>
               <Trans>
--- a/newIDE/app/src/ObjectEditor/Editors/SpriteEditor/CollisionMasksEditor/index.js
+++ b/newIDE/app/src/ObjectEditor/Editors/SpriteEditor/CollisionMasksEditor/index.js
@@ -1,6 +1,8 @@
 // @flow
 import { Trans } from '@lingui/macro';
+import { I18n } from '@lingui/react';
 import React from 'react';
+import every from 'lodash/every';
 import FlatButton from '../../../../UI/FlatButton';
 import EmptyMessage from '../../../../UI/EmptyMessage';
 import { Line, Column } from '../../../../UI/Grid';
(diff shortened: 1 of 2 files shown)


newIDE/app/src/ObjectEditor/Editors/SpriteEditor/CollisionMasksEditor/index.js
```diff
@@ -75,6 +76,10 @@ const CollisionMasksEditor = (props: Props) => {
     spriteIndex
   );
 
+  const objectHasPlatformerBehavior = props.object.hasBehaviorNamed(
```
4ian, line 79: Slightly more complicated, but to be very precise we would need to iterate on behaviors and see if any of them has a type which is `PlatformBehavior::PlatformerObjectBehavior`.

This can be done with `object.getAllBehaviorNames().toJSArray()` and then iterate and for each behavior, get it (`getBehavior`) and on the resulting object uses `getTypeName`.

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/CollisionMasksEditor/index.js
```diff
@@ -269,6 +274,18 @@ const CollisionMasksEditor = (props: Props) => {
               </Line>
             </React.Fragment>
           )}
+          {!!sprite &&
```
4ian, line 277: I think we could move this into the `Hints/index.js` file? In an effort to centralise all this "extra knowledge" in a single place (no strong feeling on this though if this is complicated)

Bouh:
The warning message does not really warn with reason.

`You currently use a custom collision mask for an object that has the Platformer behavior.`
Yes ok, what is the point to explain me me something I do? 
If this action I made has consequences on something else it must be describe.

Because for me the second part explain something else:
` To prevent issues when flipping an object horizontally while grabbing an edge, make sure your collision mask does not move when flipping it.`

This last sentence have no relation with the first sentence, it describe something else.
If the implicit explaination is that the collision mask isn't supported by platformer behavior, it's not describe in these sentences, that is the point that should be explain in the first sentence IMO.

Bouh:
I say that because I read this issue just before https://github.com/4ian/GDevelop/issues/3344

AlexandreSi:
As I commented in #3344, I think that there are 2 problems that could require 2 warnings:

- If you define a custom collision mask that is not a single rectangle, prepare for surprises (and we could add checks on the collision mask to display this warning only if it is not a single rectangle)
- If you define a custom collision, make sure it does not move when flipping it horizontally because when grabbing edges, it could do funny things (but if you didn't activate the edge grabbing, you can ignore that).

AlexandreSi:
> `You currently use a custom collision mask for an object that has the Platformer behavior.` Yes ok, what is the point to explain me me something I do? If this action I made has consequences on something else it must be describe.

I guess I wanted to make sure that the user understands why this warning appears. That way, he can easily find a workaround for what he is doing or refer to that in the forum.

> Because for me the second part explain something else: ` To prevent issues when flipping an object horizontally while grabbing an edge, make sure your collision mask does not move when flipping it.`
> 
> This last sentence have no relation with the first sentence, it describe something else. If the implicit explaination is that the collision mask isn't supported by platformer behavior, it's not describe in these sentences, that is the point that should be explain in the first sentence IMO.

You think it would be better to say: "A platformer with a custom collision mask could experiment issues if it is flipped horizontally grabbing an edge. To avoid that, make sure your collision mask does not move when flipping it."

tristanbob:
People may not know what causes a collision mask to "move".

"If the collision masks are not centered vertically, the object may act erratically when the flip action is performed near a wall."

AlexandreSi:
> People may not know what causes a collision mask to "move".
> 
> "If the collision masks are not centered vertically, the object may act erratically when the flip action is performed near a wall."

That's where I am always confused what is vertically center vs horizontally center 😅.

New proposal:
You currently use a custom collision mask for an object that has the Platformer behavior. A platformer with a custom collision mask could experiment issues if it is flipped horizontally near a wall. To avoid that, make sure your collision mask does not move when flipping it by centering it.

tristanbob:
You are right, I should have said "centered horizontally".

> You currently use a custom collision mask for an object that has the Platformer behavior. A platformer with a custom collision mask could experiment issues if it is flipped horizontally near a wall. To avoid that, make sure your collision mask does not move when flipping it by centering it.

This looks fine, but it has a small typo:   experiment => experience

AlexandreSi:
Thanks for the catch!

Elairyx7301:
> You currently use a custom collision mask for an object that has the Platformer behavior. A platformer with a custom collision mask could experiment issues if it is flipped horizontally near a wall. To avoid that, make sure your collision mask does not move when flipping it by centering it.

Simply centering the hitbox will prevent the bug but can be misleading and users might understand it as center horizontally *and* vertically and afaik that's not all required. Furthermore, I have a feeling that this warning is getting a little too long and I can see users (including me if I didn't know what could happen) being confused and having to read that phrase multiple times to understand what causes the error/"bug".

I see the first sentence as completely redundant because it's not saying anything the following sentences aren't mentioning and also "You currently use a custom collision mask" doesn't cause the bug.

Also, isn't the "Platformer" behavior something different?

I've never heard "experiment" in that context before. Was it meant to be "experience" or is that a correct sentence?

"Move" is unclear when describing the collision mask. In this context, "move" could even be unclear in definition. If I flip a square, doesn't the top right point of the collision mask not move to the left side (instantly)?

If GDevelop could check the hitbox points, I think we need to only show that warning if it's not centered horizontally (if possible).

- - -
Proposal:
"Your Platformer Character object has a custom collision mask that is not centered on the X-axis. This can cause issues when flipping the object horizontally near a wall."

AlexandreSi:
I hear you
In this case, I think we can do something more complex that would benefit #3344 as well. I'll work on it next week.

PR #3350, opened 2021-12-10
Move the copy button for the web export link to the end of the input

4ian:
Also use a component doing the layout of the button on the right.

Don't show in changelog

image
image

### DIFF
--- a/newIDE/app/src/Export/GenericExporters/OnlineWebExport.js
+++ b/newIDE/app/src/Export/GenericExporters/OnlineWebExport.js
@@ -1,5 +1,6 @@
 // @flow
 import { Trans } from '@lingui/macro';
+import { t } from '@lingui/macro';
 import * as React from 'react';
 import Text from '../../UI/Text';
 import { Column, Line } from '../../UI/Grid';
@@ -10,8 +11,10 @@ import {
 } from '../../Utils/GDevelopServices/Build';
 import RaisedButton from '../../UI/RaisedButton';
 import Window from '../../Utils/Window';
-import Paste from '../../UI/CustomSvgIcons/Paste';
+import Copy from '../../UI/CustomSvgIcons/Copy';
 import InfoBar from '../../UI/Messages/InfoBar';
+import IconButton from '../../UI/IconButton';
+import { TextFieldWithButtonLayout } from '../../UI/Layout';
 
 export const ExplanationHeader = () => (
   <Column noMargin alignItems="center" justifyContent="center">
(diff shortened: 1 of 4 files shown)


newIDE/app/src/UI/TextField.js
```diff
@@ -249,6 +253,14 @@ export default class TextField extends React.Component<Props, {||}> {
                 min: props.min,
                 step: props.step,
               },
+              // Input adornment:
+              endAdornment: props.endAdornment ? (
```
ClementPasteau, line 257: interesting I didn't know adornment was a thing 👍

newIDE/app/src/Export/GenericExporters/OnlineWebExport.js
```diff
@@ -56,20 +59,42 @@ export const WebProjectLink = ({ build, loading }: WebProjectLinkProps) => {
...
+        noFloatingLabelText
+        renderTextField={() => (
+          <TextField
+            value={value}
+            readOnly
+            fullWidth
+            endAdornment={
+              <IconButton
+                disabled={!!buildPending}
+                onClick={() => onCopy()}
```
ClementPasteau, line 73: this can just be `onCopy` - it's probably not important any more, but I think with React Classes, this used to cause a re-render because the function was re-created every time?
Anyway, I think it's more clean 😇

ClementPasteau (approved):
You felt that 2 buttons was too much ? :)
I liked the idea that the copy button was the "primary" action -> the next thing you're gonna do, so it catches the eye. we're losing this now I guess

newIDE/app/src/Export/GenericExporters/OnlineWebExport.js
4ian: This won't help re-renders in this particular case (because onCopy is not a React.useCallback) BUT you're totally right that we should strive to have as few functions as possible (especially because with this, we're not far from being able to entirely switch IconButton to a pure component/React.memo and avoid a bunch of useless renders!)

Fixed :)

newIDE/app/src/UI/TextField.js
4ian: I remembered having seen that in the material-ui docs, turns out to be a good fit here! :)

4ian:
So there were two things: 
- purely on the design side of things, raised buttons should not be touching each others (this could have been solved with a Spacer too), and should in theory not grow in height (except if the text is wrapped).
- But more importantly: while I agree that we discussed about the copy being important, I think we missed that it's important because we want people to *share* the game: copying the URL is just a tool for doing that. But I think that whatever we do here (two buttons like before or one like now with an adornment, which is I think still fairly easy to see), we're missing a large thing: we should add buttons to easily share the game (and this could be in this dialog, or even in the game itself). 

Let's talk about this on Monday! I think there is a larger thing about sharing that can be done and that will be 10x more impactful than the copy button. That's why I took the liberty of slightly reducing its visibility (but also at the same time, this is a fairly common UX pattern to have a copy button embedded in an input field, so I think it's fine!).

Issue #2883, opened 2021-08-07
Show or Hide layer does not show or hide objects from that layer for Conditions


Silver-Streak:
## Describe the bug
When you use the "Show/Hide Layer" action, while all of the objects are _visually_ hidden from the game, from a logic perspective they're still not treated as hidden. 

## To Reproduce
1. Create any project
2. Add a layer.
3. Add objects to the new layer.
4. Add an event that hides the layer.
5. Launch the debugger.
6. Check the objects. "Is hidden?" stays as no.

Here's the GUI layer hidden: 
![image](https://user-images.githubusercontent.com/24237149/128593235-3377ebf7-3155-4f74-bc59-af5ee404e782.png)

Here's the left text object on the GUI layer (while the layer is hidden), still showing 'Is hidden?" as false.
![image](https://user-images.githubusercontent.com/24237149/128593244-efe37ec6-288f-4cba-b3f1-292e9cab1cf8.png)

[PlatformerBUG.zip](https://github.com/4ian/GDevelop/files/6948796/PlatformerBUG.zip)

I've tweaked an existing project that was built by someone for the vertical moving platform bug to show this. 
A to hide the layer, D to show it. 
W to hide the left text object, E to show it.
The platform is tinted green when the left text object is visible, red when it's hidden.
The object is treated as visible unless you manually hide it.

You can work around this by setting up some events that checks for the visibility of an object's layer using expressions, then hiding the object if it is still visible, but this seems like a messy solution if you have a lot of objects.

If this behavior is intentional, it can cause some issues for users that want to hide a large quantity of objects, or use object visibility conditions to do object selection.  In this case, I'd recommend adding a parameter to the show/hide layer actions for whether the user wants to also treat the objects as hidden. Default it to "no" to keep current default functionality.

## Other details
OS: Windows 10 pro
GD5 version: B114
Extra context: Found this while trying to help Creativesamurai on the discord deal with layer-based menus.

4ian:
I would say that currently this "works as intended" as hiding/showing an object is different from hiding/showing a layer.
But I understand this can create some confusion. Should be either reworked or made clearer in the description/sentences?

Bouh:
The logic shouldn't be that visibility have to be inherited from the layer?

It make more sense, the object isn't visible, it's hidden due to the layer, if the object condition "is visible" return true while the object isn't visible, its a bit weirdy, don't you think?

Silver-Streak:
[...]

PR #3355, opened 2021-12-13
Change layer visibility checkbox & add a helper text

ClementPasteau:
Addresses #2883 

Also changed the checkbox wording & logic to be more intuitive. (Disabled by default)

Layer:
image

Lighting layer:
image

Tooltip on layer panel:
image

### DIFF
--- a/newIDE/app/src/LayersList/LayerEditorDialog.js
+++ b/newIDE/app/src/LayersList/LayerEditorDialog.js
@@ -139,12 +139,19 @@ const LayerEditorDialog = (props: Props) => {
               </Text>
             )}
             <InlineCheckbox
-              label={<Trans>Layer visible</Trans>}
-              checked={layer.getVisibility()}
+              label={<Trans>Hide the layer</Trans>}
+              checked={!layer.getVisibility()}
               onCheck={(e, checked) => {
-                layer.setVisibility(checked);
+                layer.setVisibility(!checked);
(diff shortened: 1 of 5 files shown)


Bouh:
The interface is less cumbersome with a tooltip notes :)

image

4ian:
> The interface is less cumbersome with a tooltip notes :)

That's an interesting concept but this is not used in any place in the app, I think in this case we should make it so that other part of the app are reworked to use this for consistency sake.

@ClementPasteau can you try replace the Text by a FormHelperText to give it less visual importance?
According to this, it could be put inside the label, using a Line then a FormHelperText (no need for the markdown text here)
image

newIDE/app/src/LayersList/LayerEditorDialog.js
```diff
@@ -139,12 +139,19 @@ const LayerEditorDialog = (props: Props) => {
...
+              label={<Trans>Hide the layer</Trans>}
+              checked={!layer.getVisibility()}
               onCheck={(e, checked) => {
-                layer.setVisibility(checked);
+                layer.setVisibility(!checked);
                 forceUpdate();
               }}
+              tooltipOrHelperText={
+                <Trans>
+                  This setting changes visibility of the entire layer. Objects
```
4ian, line 150: visibility => the visibility

4ian (approved):

PR #3357, opened 2021-12-13
Add possibility to choose name before project creation

AlexandreSi:


newIDE/app/src/ProjectCreation/services/BrowserCreation.js
```diff
@@ -27,16 +29,22 @@ export const onCreateBlank = async ({
   sendNewGameCreated('');
 
   const project = gd.ProjectHelper.createNewGDJSProject();
-  return { project, storageProvider: null, fileMetadata: null };
+  return {
+    project,
+    storageProvider: null,
+    fileMetadata: { projectName, fileIdentifier: '' },
```
AlexandreSi, line 35: @4ian I feel like it is a hack but making fileIdentifier optional calls for a big refactor. I don't know if the project name should be part of FileMetadata

AlexandreSi: Nevermind, I just realized my code is nonsense

newIDE/app/src/MainFrame/EditorContainers/HomePage/index.js
```diff
@@ -260,11 +264,13 @@ export const HomePage = React.memo<Props>(
           const projectMetadata = await onCreateBlank({
             i18n,
             outputPath,
+            projectName: newProjectName,
           });
           if (!projectMetadata) return;
           const { project, storageProvider, fileMetadata } = projectMetadata;
           setPreCreationDialogOpen(false);
           setOutputPath(computeDefaultProjectPath());
+          setNewProjectName(generateName());
```
ClementPasteau, line 273: Is that necessary if you already have the state value intialised with the same function ?

newIDE/app/src/MainFrame/index.js
```diff
@@ -1987,7 +1987,12 @@ const MainFrame = (props: Props) => {
     else if (!!fileMetadata) state = await openFromFileMetadata(fileMetadata);
 
     if (state) {
-      if (state.currentProject) state.currentProject.resetProjectUuid();
+      if (state.currentProject) {
+        const { currentProject } = state;
+        currentProject.resetProjectUuid();
+        if (fileMetadata && fileMetadata.projectName)
+          currentProject.setName(fileMetadata.projectName);
+      }
```
ClementPasteau, line 1995: As discussed, can we save the project into the json here? to avoid having wrong uuid & name in the json!

newIDE/app/src/ProjectCreation/ProjectPreCreationDialog.js
```diff
@@ -0,0 +1,107 @@
...
+  projectName,
+  onChangeProjectName,
+}: Props): React.Node => {
+  const [projectNameError, setProjectNameError] = React.useState<?React.Node>(
+    null
+  );
+
+  let projectNameToDisplay =
+    projectName === null || projectName === undefined
+      ? 'Bonjour les cocos'
```
ClementPasteau, line 38: ? :)

newIDE/app/src/ProjectsStorage/index.js
```diff
@@ -10,6 +10,7 @@ import { type AppArguments } from '../Utils/Window';
 export type FileMetadata = {|
   /** The file id, path or local path according to the provider. */
   fileIdentifier: string,
+  projectName?: string,
```
ClementPasteau, line 13: should this be nullable?

newIDE/app/src/stories/componentStories/ProjectPreCreationDialog.stories.js
```diff
@@ -0,0 +1,45 @@
...
+};
+
+export const Open = () => {
+  const [projectName, setProjectName] = React.useState('Project Name');
+  return (
+    <ProjectPreCreationDialog
+      open
+      outputPath="/path/to/project/file.json"
+      onChangeOutputPath={action('change output path')}
+      onClose={() => action('click on close')()}
```
ClementPasteau, line 23: this can just be `action(xxx)` ? (same below)

ClementPasteau (commented):
small comments, lgtm!

newIDE/app/src/ProjectCreation/ProjectPreCreationDialog.js
AlexandreSi: 😬

newIDE/app/src/MainFrame/EditorContainers/HomePage/index.js
AlexandreSi: It is to make sure that if the user creates a new project, closes it and then creates another one, there is a new project name suggestion instead of the last project name he entered

newIDE/app/src/ProjectsStorage/index.js
AlexandreSi: I have not encountered a case in which it is useful.

newIDE/app/src/MainFrame/EditorContainers/HomePage/index.js
4ian: Related question: is this necessary to then initialize the name in the useState?

newIDE/app/src/MainFrame/index.js
4ian: Might be more complex than just calling a save function, but we should think about it for either this PR or another PR.

newIDE/app/src/MainFrame/EditorContainers/HomePage/index.js
AlexandreSi: The homepage is never unmounted so I would say yes. But there's maybe something I don't see

newIDE/app/src/MainFrame/index.js
AlexandreSi: I tried and then I realized it wouldn't be this easy indeed, I would say another PR

4ian: Correct it's not easy because we need to act differently than a traditional save (like if it fails, we probably ignore the problem) (we also don't want to trigger the window to select a save destination like done in the web-app, so we need a flag to tell "please save silently and if it fails or there is no obvious place where to save, just don't care.")

AlexandreSi: There's also a challenge on synchronicity: the saveProject method returns if currentProject is not defined, but at the moment of this function, it is never defined.

newIDE/app/src/ProjectCreation/ProjectPreCreationDialog.js
```diff
@@ -0,0 +1,107 @@
...
+      : projectName;
+
+  const onValidate = React.useCallback(
+    () => {
+      setProjectNameError(null);
+      if (!projectName) {
+        setProjectNameError(<Trans>Your project needs a name</Trans>);
```
4ian, line 45:
```suggestion
        setProjectNameError(<Trans>Your project needs a name.</Trans>);
```

AlexandreSi: I think I'll use a completely different formulation, this one seems a bit weird.

4ian: Yep thanks!

4ian (approved):

newIDE/app/src/ProjectCreation/services/BrowserCreation.js
```diff
@@ -1,63 +1,49 @@
...
-  storageProvider: StorageProvider,
-  fileMetadata: FileMetadata,
-|}> => {
+}) => {
   try {
     const example = await getExample(exampleShortHeader);
     sendNewGameCreated(example.projectFileUrl);
     return {
       storageProvider: UrlStorageProvider,
+      projectName,
```
4ian, line 45: Much better!

newIDE/app/src/MainFrame/EditorContainers/HomePage/index.js
```diff
@@ -261,20 +261,30 @@ export const HomePage = React.memo<Props>(
...
           if (!projectMetadata) return;
-          const { project, storageProvider, fileMetadata } = projectMetadata;
+          const {
+            project,
+            storageProvider,
+            fileMetadata,
+            projectName,
+          } = projectMetadata;
+
+          // Once project is created, reinitialize default values for variables related to project creation.
```
4ian, line 278: Thanks for the comment. Any chance this could be moved at the moment the project is created, rather than having to remember to "clean" after the creation?
Not sure if it makes sense, but I figure it's more intuitive to have set at the right moment before creating/opening a dialog than having to remember to clean these later.

ClementPasteau (approved):
🎉

PR #3360, opened 2021-12-14
[Platformer] Split tests in 5 files.

D8H:
It should make it easier to navigate in platformer tests.

The 472 tests are here and the diff line count seems to match.


Extensions/PlatformBehavior/tests/PlatformerTestHelper.js
```diff
@@ -0,0 +1,229 @@
+
```
4ian, line 1: Because this file will add global variables, I wonder if we should at least change the name of `makeTestRuntimeScene`? To avoid conflicts/make it clear this is not a common util to be reused in other tests?

4ian:
Thank you!

PR #3367, opened 2021-12-15
Fix 'Not' condition not working with most object conditions

### LINKED ISSUE #3233: 'Not' condition breaks on automatic object selection
## Describe the bug

When using an object condition that uses GDevelops built-in object selection mechanism, it never triggers, when it should. This doesn't affect object conditions that do the object picking "manually", like the "cursor on object" condition.

## Cause

The cause is as far as I see due to two conflicting assumptions (but those assumptions are just my assumption of what's happening, they may be wrong :p).

The not condition sets the return value that needs to be negated to true by default, since no condition = always true, and it assumes that if there are subconditions, they will override it to their return value.

In the other hand, the default automatic object selection assumes that by default the return value is false, and never explicitly sets it to false, it only sets it to true each time an instance meets the condition.

## To Reproduce

1. Make a Not condition with the cursor on object condition
2. Make a not condition with point in object condition
3. See how the latter never triggers when the first one works as expected
4. Look in the sources debugger tab at how the generated code never sets the return value of the later to false, only to true constantly, meaning that the inverted return value will always be false.

## Project reproduction

[autojump.zip](https://github.com/4ian/GDevelop/files/7529542/autojump.zip)

arthuro555:
The 'Not' condition was giving the variable `IsTrue` a default value of `true`, where it is expected by the default object picking code for that variable to default to be false. It likely wasn't noticed as only some conditions break, the ones that do `if (condition) IsTrue = true;`, not the ones that do `IsTrue = condition` (most of them).

Fixes #3233

Tested with the repro project of #3233 to see if the bug was fixed, and with a previously working object condition, no condition, and a free condition to check for regressions.

### DIFF
--- a/GDJS/GDJS/Extensions/Builtin/CommonInstructionsExtension.cpp
+++ b/GDJS/GDJS/Extensions/Builtin/CommonInstructionsExtension.cpp
@@ -278,7 +278,7 @@ CommonInstructionsExtension::CommonInstructionsExtension() {
               outputCode +=
                   codeGenerator.GenerateBooleanFullName(
                       "condition" + gd::String::From(i) + "IsTrue", context) +
-                  ".val = true;\n";
+                  ".val = false;\n";
             }
 
             for (unsigned int cId = 0; cId < conditions.size(); ++cId) {


4ian:
Sorry for the delay! This seems indeed correct. The condition return should be assumed false by default. I probably set this to true because I was thinking "because it's a not we're inverting this", but this is not making sense: the logic to determine the condition result should be the same (i.e: false by default). The "not" is applied only on the result of each condition.

So seems safe and a correct fix 👍

PR #3368, opened 2021-12-15
Improve local IP getting

arthuro555:
When having virtualization software installed, a computer has fake local network interfaces and gives an unusable local IP. This adds a check to filter out interfaces from common virtualization software. Tested with my own PC that has VirtualBox installed.

### DIFF
--- a/newIDE/electron-app/app/Utils/LocalNetworkIpFinder.js
+++ b/newIDE/electron-app/app/Utils/LocalNetworkIpFinder.js
@@ -2,18 +2,12 @@ const os = require('os');
 
 /** @returns {string[]} */
 const getLocalNetworkIps = () => {
-  var interfaces = os.networkInterfaces();
-  var addresses = [];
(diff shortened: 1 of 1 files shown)


newIDE/electron-app/app/Utils/LocalNetworkIpFinder.js
```diff
@@ -2,18 +2,12 @@ const os = require('os');
...
-      var address = interfaces[k][k2];
-      if (address.family === 'IPv4' && !address.internal) {
-        addresses.push(address.address);
-      }
-    }
-  }
-
-  return addresses;
+  return Object.entries(os.networkInterfaces())
+    .flatMap(([name, interface]) =>
```
4ian, line 6: ubernitpicking: this should be `interfaces` as it's an array

4ian (approved):

4ian:
Thanks 👍

PR #3373, opened 2021-12-17
Allow extensions to define property groups.

D8H:
It will help users to understand behavior configuration and find the property they looking for.

### Embedded behavior

PropertyGroupEmbeddedExtention

### Event behavior

PropertyGroupEventExtention

### Event behavior editor with autocompletion

PropertyGroupEventExtentionEditor

### DIFF
--- a/newIDE/app/src/BehaviorsEditor/Editors/BehaviorPropertiesEditor.js
+++ b/newIDE/app/src/BehaviorsEditor/Editors/BehaviorPropertiesEditor.js
@@ -14,25 +14,29 @@ import { type Schema } from '../../PropertiesEditor';
 type Props = BehaviorEditorProps;
 
 export default class BehaviorPropertiesEditor extends React.Component<Props> {
-  render() {
-    const { behavior, behaviorContent, object } = this.props;
+  _getPropertyGroupNames = (): Array<string> => {
+    const { behavior, behaviorContent } = this.props;
     const properties = behavior.getProperties(behaviorContent.getContent());
 
-    const groupNames = [];
+    const groupNames = new Set<string>();
     const propertyNames = properties.keys();
     for (let i = 0; i < propertyNames.size(); i++) {
       const name = propertyNames.at(i);
       const property = properties.get(name);
       const group = property.getGroup() || '';
-      if (!groupNames.includes(group)) {
-        groupNames.push(group);
-      }
+      groupNames.add(group);
     }
-    groupNames.sort((a, b) => a.localeCompare(b));
+    return [...groupNames].sort((a, b) => a.localeCompare(b));
+  };
+
+  render() {
+    const { behavior, behaviorContent, object } = this.props;
+    const properties = behavior.getProperties(behaviorContent.getContent());
+
+    const groupNames = this._getPropertyGroupNames();
 
     let schemaIsEmpty = true;
-    const propertiesSchemas: Schema[] = [];
-    for (const groupName of groupNames) {
+    const propertiesSchemas: Schema[] = groupNames.map(groupName => {
       const propertiesSchema = propertiesMapToSchema(
         properties,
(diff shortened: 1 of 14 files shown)


D8H:
I added groups for behaviors with a lot of property.

PathFindingProperties TopDownProperties PlatformerProperties PlatformProperties

newIDE/app/src/BehaviorsEditor/Editors/BehaviorPropertiesEditor.js
```diff
@@ -7,30 +7,78 @@ import propertiesMapToSchema from '../../PropertiesEditor/PropertiesMapToSchema'
...
 type Props = BehaviorEditorProps;
 
 export default class BehaviorPropertiesEditor extends React.Component<Props> {
+  _getPropertyGroupNames = (): Array<string> => {
+    const { behavior, behaviorContent } = this.props;
+    const properties = behavior.getProperties(behaviorContent.getContent());
+
+    const groupNames = new Set<string>();
+    const propertyNames = properties.keys();
+    for (let i = 0; i < propertyNames.size(); i++) {
```
4ian, line 23: nitpicking: you can use keys().toJSArray().forEach if you prefer to manipulate a JS array

4ian: I think something like:

return [...new Set<string>(properties.keys().toJSArray().map(propertyName => property.get(propertyName).getGroup() || "")).sort(...)

can work (just a matter of slightly more functional style)

newIDE/app/src/BehaviorsEditor/Editors/BehaviorPropertiesEditor.js
```diff
@@ -7,30 +7,78 @@ import propertiesMapToSchema from '../../PropertiesEditor/PropertiesMapToSchema'
...
+              }
+              return [
+                <Text key={groupNames[index] + 'title'} size="title">
+                  {groupNames[index]}
+                </Text>,
+                <PropertiesEditor
+                  key={groupNames[index]}
+                  schema={propertiesSchema}
+                  instances={[behaviorContent]}
+                />,
```
4ian, line 78: Do you think we could put this logic of the title rendering inside the schema and PropertiesEditor?

Probably when converting the properties to a Schema, we can enhance the existing type of field containing other fields:

image

and render the title here:
image


This needs some logic when converting the properties to, well, group them inside the returned Schema :) The advantage is that this is not specific to behaviors and could be used for objects properties, instance properties and other places where properties are shown (the resource editor)

newIDE/app/src/PropertiesEditor/PropertiesMapToSchema.js
```diff
@@ -18,12 +18,21 @@ export default (
...
   ) => void,
-  object: ?gdObject
+  object: ?gdObject,
+  groupFilter: ?string
 ): Schema => {
   const propertyNames = properties.keys();
   const propertyFields = mapFor(0, propertyNames.size(), i => {
     const name = propertyNames.at(i);
     const property = properties.get(name);
+    if (groupFilter || groupFilter === '') {
```
4ian, line 28: If we use the approach of doing the grouping here, by putting fields into "container" fields with titles, we avoid doing N*M iterations, N being the number of properties and M the number of groups. Not a big deal, but when rendering large list of properties, this can start to be visible on low end devices I think.

4ian (commented):
Love this grouping of properties! This is long overdue! 

I was also thinking of one day allowing some properties to be shown on the same row as others (like X, Y positions).
I wonder though how to do this properly (we would maybe need to have a separate hierarchy that is just containing the property names in a structure mimicking how they should be presented on screen)... but groups are a good start in any case! Way clearer.

PR #3376, opened 2021-12-17
Apply rbush to platformer obstacles manager

### LINKED ISSUE #1463: Ghost collision mask

AlexandreSi:
Resolves #1463 
Might not resolve it but it's a good beginning, following fixes for similar bugs on light and pathfinder obstacles.

@D8H Could you have a look on the 2 tests failing? As your eyes are trained on these tests, maybe you can spot something interesting.
Good news, none of your recent tests fail but there are these 2 ugly ducklings.

@4ian the relatively good news is that the 2 failing tests are the same as when we tried for the first time. From what I remember, were not able to reproduce them in the platformer starter with draggable platforms.

### DIFF
--- a/Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
+++ b/Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
@@ -988,8 +988,10 @@ namespace gdjs {
      * Update _potentialCollidingObjects member with platforms near the object.
      */
     private _updatePotentialCollidingObjects(maxMovementLength: float) {
+      const object = this.owner;
+
       this._manager.getAllPlatformsAround(
-        this.owner,
+        object,
         maxMovementLength,
         this._potentialCollidingObjects
       );
@@ -998,7 +1000,7 @@ namespace gdjs {
       // is not considered as colliding with itself, in the case that it also has the
       // platform behavior.
       for (let i = 0; i < this._potentialCollidingObjects.length; ) {
-        if (this._potentialCollidingObjects[i].behavior.owner === this.owner) {
+        if (this._potentialCollidingObjects[i].behavior.owner === object) {
           this._potentialCollidingObjects.splice(i, 1);
         } else {
           i++;
--- a/Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
+++ b/Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
@@ -1562,12 +1562,14 @@ namespace gdjs {
         const deltaMaxY = Math.abs(
           behavior._requestedDeltaX * behavior._slopeClimbingFactor
         );
-        const { highestGround, isCollidingAnyPlatform } =
-          behavior._findHighestFloorAndMoveOnTop(
-            behavior._potentialCollidingObjects,
-            -deltaMaxY,
-            deltaMaxY
-          );
+        const {
+          highestGround,
+          isCollidingAnyPlatform,
+        } = behavior._findHighestFloorAndMoveOnTop(
+          behavior._potentialCollidingObjects,
+          -deltaMaxY,
+          deltaMaxY
+        );
         if (highestGround && highestGround !== this._floorPlatform) {
           behavior._setOnFloor(highestGround);
         }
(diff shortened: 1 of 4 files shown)


D8H:
It follows the platform whereever it goes. I would guess that the AABB of the platform is not yet updated when the moving platform following check is done and that it only checks that the platform is in the surrounding according to RBush.

```
LOG: 'Falling 0 -99.75'
LOG: 'Falling 0 -31.5'
LOG: 'OnFloor 0 -30'
LOG: 'OnFloor 0 -30'
LOG: 'OnFloor 0 -30'
LOG: 'OnFloor 0 -30'
LOG: 'OnFloor 0 -30'
LOG: 'OnFloor 0 -30'
LOG: 'OnFloor 0 -30'
LOG: 'OnFloor 0 -30'
LOG: 'OnFloor 0 -30'
LOG: 'OnFloor -1000 -1020'
HeadlessChrome 96.0.4664 (Windows 10.0.0) gdjs.PlatformerObjectRuntimeBehavior (falling) falls when a platform is moved away FAILED
        Error: expected false to equal true
            at Assertion.be.Assertion.equal (node_modules/expect.js/index.js:216:10)
            at Assertion.<computed> [as be] (node_modules/expect.js/index.js:69:24)
            at context.<anonymous> (C:/WorkspaceGDevelop/GDevelop/Extensions/PlatformBehavior/tests/JumpAndFallingPlatformer.spec.js:92:58)
HeadlessChrome 96.0.4664 (Windows 10.0.0): Executed 1 of 1 (1 FAILED) ERROR (0.008 secs / 0.009 secs)
```

D8H:
We can add check on the actual current AABB of the platform or make the RBush update more predictable with this fix:
* https://github.com/4ian/GDevelop/pull/2602

D8H:
I did a PR, the tests are passing when merged with this branch.
* https://github.com/4ian/GDevelop/pull/3379

4ian:
Thanks @AlexandreSi for digging again on this. 

> I did a PR, the tests are passing when merged with this branch.

Nice! So this is avoiding the issue of the RBush being returning objects that are moved away. Can you explain again a situation where this can happen (I know we discussed it, but it would be useful to have it written again :))

You also said "The check is probably required even if the AABB where updated correctly anyway."
Could you also explain in a few words why this would be the case?

Finally, do you think the extra checks introduced in #3379 will have an impact on performance?

To recap my understanding:
- This PR is adding a level of indirection to make the AABB stored in the Rbush "immutable". This is the right thing to do because 1) the RBush is built around AABB being immutable (the only way to change one should be to remove it and add it again) and 2) we can't guarantee that AABB returned by objects are not mutated, because any time the position of an object is changed, then the AABB is invalidated and the old one is lost. If we don't fix this, we risk having issues of "ghost platforms", because when trying to remove/add again the object AABB when updating the RBush, the old AABB would not be found and not removed.
- The other PR by @D8H is fixing the issue of the RBush being not always exact, returning platforms that may actually be gone.
  - Is there the reverse problem possible? The RBush being not always exact, platforms being "just arrived"/teleported in the search area being not returned by the AABB, and so we're missing a potential obstacle during ~1 frame?

D8H:
> You also said "The check is probably required even if the AABB where updated correctly anyway." Could you also explain in a few words why this would be the case?
> 
> Finally, do you think the extra checks introduced in #3379 will have an impact on performance?

I though that the RBush contract was to give every objects from an area or more, but maybe this is just the bugs fixed by this PR and the one that fix the RBush updates order (https://github.com/4ian/GDevelop/pull/2602).

It will have an impact, but not a very huge one, but it's probably not necessary with (https://github.com/4ian/GDevelop/pull/2602). I can put a TODO to remove it when the issue from (https://github.com/4ian/GDevelop/pull/2602) is fixed.

>       * Is there the reverse problem possible? The RBush being not always exact, platforms being "just arrived"/teleported in the search area being not returned by the AABB, and so we're missing a potential obstacle during ~1 frame? I think this is the case because you said in your PR "It won't make it up for platforms that should have been found by RBush. So, this PR still stand: [...]" (so this is related to my question above, can you explain a situation where this happens?)

There is an explanation in the description of this PR:
* https://github.com/4ian/GDevelop/pull/2602

AlexandreSi:
So if I sum up:
- we could merge this PR along with #3379 to resolve the mentioned bug of having ghost behaviors but it could have a performance issue.
- #2602 is a separate but similar issue that will be resolved on its own. And when it will be resolved, changes from #3379 won't be useful anymore.

4ian:
Let's add  #3379 on top of this and a TODO that the extra check of the #3379 PR can be removed when #2602 is done.

AlexandreSi:
@4ian If it lgty, we can merge this PR!

4ian (approved):
LGTM! Should be safe to go ahead as tests are covering this :)

PR #3386, opened 2021-12-20
Improve homepage

AlexandreSi:
- ~~Start from a template~~ => Start from an example
- ~~Project settings~~ => New project
- Add icons to `Create blank project` and `Browse All` buttons
- Change `Create blank project` color
- Add adornment to generate new random name

image

image


AlexandreSi:
@arthuro555 @Silver-Streak First iteration on your feedbacks!

arthuro555:
Nice! One of the points that made the community the most upset is the removal of the option to create a blank project from the File > New project menu, was that addressed too?

Midhil457:
Don't you think there should be another way to get to the example other than the small `Browse all` button, Maybe like a tab when creating an project like before (which I think gave the example a bit more visibilty). I always think the `browse all` as a secondary way to get to the main way, with another prominent way too (Though it isn't always the case)

Silver-Streak:
> Don't you think there should be another way to get to the examples other than the small `Browse all` button, Maybe like a tab when creating an project like before (which I think gave the example a bit more visibilty). It feels like all the high quality example are hidden behind that single button which I think isn't a good thing. If the user misses that single button, they are missing all the examples. I always think the `browse all` as a secondary way to get to the main way, with another prominent way too (Though it isn't always the case)

Then, as mentioned above, the browse all modal should still have a create new blank project on the upper right.

I don't personally think there is an issue with cleaning up the interface in the way has been done now. I do think having *any* buttons just be plain text/link text is a bad idea. It blends in too much and reminds me a bit of sites attempting to be Youtube  but not understanding the design ethos. 
Youtube has all of its carousel completely segmented from navigation functionality, while the buttons on their pages (navigation and otherwise) are set up in a way that doesn't allow them to be confused as plaintext.

I think the current layout is fine, but all buttons should be full buttons, potentially with relevant icons (blank project with some kind of plus icon, show more/browse all with some kind of list icon, etc)

arthuro555:
🤔 Personally, i think that the "new project" form for blank projects should be moved to being a tab in the modal with the example and tutorial browsers, that way File > Create a project would give you the option to create a blank project directly, while having tabs to easily switch over to an example to use as template by the click of a tab, and allowing to crate a blank project if you do not find an example that satisfies you after all.

AlexandreSi:
The quickest way for me to fix that is adding this button:
image
What do you think about it?

I feel you all, and we're doing our best to keep everyone satisfied, but none of us has been trained on UI/UX aspects, so I hope we will found a product designer soon!

Bouh:
> What do you think about it?

The button looks like it's lost to get here.

Raised button on the homepage and we're good.

AlexandreSi:
> > What do you think about it?
> 
> The button looks like it's lost to get here.
> 
> Raised button on the homepage and we're good.

I think that with the icon and the color, the button should be more visible on the homepage.

I was suggesting a button `Create a blank project` on this `Create a new project` dialog, that is opened when using `File` > `Create a new project` sub menu.

ClementPasteau (approved):

ClementPasteau:
I like it this way :) 
Let's merge this improvement for the moment and see how it feels in day to day use.
We can still improve it later on!

newIDE/app/src/ProjectCreation/CreateProjectDialog.js
```diff
@@ -67,99 +67,189 @@ export type OnCreateFromExampleShortHeaderFunction = ({|
...
+    ],
+    [onClose, openPreCreationDialog]
+  );
 
-  _onChangeTab = (newTab: CreateProjectDialogTabs) => {
-    this.setState({
-      currentTab: newTab,
-    });
+  const secondaryActions = React.useMemo(
+    () => [
```
ClementPasteau, line 126: should we simplify this to 
```
if (currentTab === 'games-showcase') return [<FlatButton ... />]
if (currentTab === 'examples') return [<FlatButton .../>]
```

rather than hiding the logic inside the array

newIDE/app/src/ProjectCreation/CreateProjectDialog.js
```diff
@@ -67,99 +67,189 @@ export type OnCreateFromExampleShortHeaderFunction = ({|
...
+        });
+      } else {
+        projectMetadata = await onCreateBlank({
+          i18n,
+          outputPath,
+          projectName: newProjectName,
+        });
+      }
+
+      if (!projectMetadata) return;
```
ClementPasteau, line 180: I'd add a console.error as well here, this should not happen

newIDE/app/src/ProjectCreation/CreateProjectDialog.js
```diff
@@ -67,99 +67,189 @@ export type OnCreateFromExampleShortHeaderFunction = ({|
...
+              >
+                <Tab label={<Trans>Examples</Trans>} value="examples" />
+                <Tab label={<Trans>Tutorials</Trans>} value="tutorials" />
+                <Tab
+                  label={<Trans>Games showcase</Trans>}
+                  value="games-showcase"
+                />
+              </Tabs>
+              {currentTab === 'examples' && (
+                <Column noMargin expand useFullHeight>
```
ClementPasteau, line 220: why has this tab a column and the others not? Should this be part of the ExampleStore?

newIDE/app/src/ProjectCreation/LocalExamples.js
```diff
@@ -22,20 +22,6 @@ type Props = {|
...
-  showErrorBox({
-    message: i18n._(
-      t`Unable to create the game in the specified folder. Check that you have permissions to write in this folder: ${outputPath} or choose another folder.`
-    ),
-    rawError,
-    errorId: 'local-example-creation-error',
-  });
-};
-
 export default function LocalExamples({
```
ClementPasteau, line 25: If you're not importing `LocalExamples` or `BrowserExamples` anymore, are they still useful?

newIDE/app/src/ProjectCreation/CreateProjectDialog.js
AlexandreSi: Oh yes you're right, I brainless copy-pasted

newIDE/app/src/ProjectCreation/CreateProjectDialog.js
AlexandreSi: Well, ExampleStore and CreateProjectDialog need the ProjectPreCreationDialog, so I figured it would be better having only one instance of it. But the logic and the component is highly similar to what is done in the homepage. It could even be refactored to be in a provider.

newIDE/app/src/ProjectCreation/LocalExamples.js
AlexandreSi: I forgot to remove them, thanks!

newIDE/app/src/ProjectCreation/CreateProjectDialog.js
AlexandreSi: In both function `onCreateBlank` et `onCreateFromExampleShortHeader` they display error boxes if an error is thrown, with the raw error.

PR #3389, opened 2021-12-21
Add possibility to use Cmd+Enter/Ctrl+Enter to submit a dialog

AlexandreSi:
- `canSubmitLastAction` props added to Dialog component
- `canSubmitLastAction` added to most of dialogs that have more than 1 action that is not just `Close`
  - Tested for each of these dialogs
- Problems encountered:
  - Dialogs that use `SemiControlledTextField` were problematic because there's a possible logic to commit on blur. So I added the on blur callback.
  - I couldn't add `canSubmitLastAction` to `ExpressionParametersEditorDialog` because the focused field when we want to submit the dialog is a `GenericExpressionField` that wraps `SemiControlledTextfield`. It seems that there's a race condition that prevents the state to be updated on blur. I couldn't wrap my head around this issue and find a solution.

My recommendation is to merge this PR without fixing the behavior with `GenericExpressionField` (or spending less than 1 hour on it) even if it can be frustrating for the users to not being able to use this feature on this important field/dialog. But on the other dialogs, it's a relief to have it working.

### DIFF
--- a/newIDE/app/src/UI/Dialog/index.js
+++ b/newIDE/app/src/UI/Dialog/index.js
@@ -97,7 +97,6 @@ type DialogContentStyle = {
   flexDirection?: 'row',
 };
 
-
 const findAndClickButton = (ref: {| current: ?HTMLElement |}): void => {
   if (!ref.current || ref.current.childElementCount === 0) return;
 
@@ -187,8 +186,13 @@ export default (props: Props) => {
       return;
     }
     if (shouldSubmit(event)) {
+      event.stopPropagation();
+      const element = document.activeElement;
+      if (element) {
+        element.blur();
+      }
       if (onSubmit === 'lastAction') {
-        findAndClickButton(actionsRef)
+        findAndClickButton(actionsRef);
       } else if (!!onSubmit) {
         onSubmit();
       }
--- a/newIDE/app/src/UI/Dialog/index.js
+++ b/newIDE/app/src/UI/Dialog/index.js
@@ -114,7 +114,6 @@ export default (props: Props) => {
   const preferences = React.useContext(PreferencesContext);
   const backdropClickBehavior = preferences.values.backdropClickBehavior;
   const size = useResponsiveWindowWidth();
-  const actionsRef = React.useRef<?HTMLElement>(null);
 
   const dialogActions = React.useMemo(
     () => (
@@ -122,9 +121,7 @@ export default (props: Props) => {
         {secondaryActions && (
           <div key="secondary-actions">{secondaryActions}</div>
         )}
-        <div key="actions" ref={actionsRef}>
-          {actions}
-        </div>
+        <div key="actions">{actions}</div>
       </React.Fragment>
     ),
     [actions, secondaryActions]
--- a/newIDE/app/src/UI/Dialog/index.js
+++ b/newIDE/app/src/UI/Dialog/index.js
@@ -59,18 +59,11 @@ type Props = {|
   onRequestClose?: () => void,
   /**
    * If specified, will be called when the dialog is dismissed in a way where changes
-   * must be kept.
+   * must be kept or when using the submit keyboard shortcut.
    * This is not applicable to all dialogs. Some dialogs may have no `onApply` and just a
    * single `onRequestClose`.
    */
   onApply?: ?() => void,
-  /**
-   * If specified, allows user to close the dialog with an action based on
-   * `shouldSubmit` key event. `lastAction` option simulates a click on the
-   * last (often the main) action button given (if this last action is a button
-   * with split menu, it simulates click on the main button).
-   */
-  canSubmitLastAction?: boolean,
 
   cannotBeDismissed?: boolean, // Currently unused.
 
@@ -97,26 +90,6 @@ type DialogContentStyle = {
   flexDirection?: 'row',
 };
 
-const findAndClickButton = (ref: {| current: ?HTMLElement |}): void => {
-  if (!ref.current || ref.current.childElementCount === 0) return;
-
-  const actionsElements = ref.current.children;
-  if (!actionsElements) return;
-
-  let target = actionsElements[actionsElements.length - 1];
-  if (target.matches('button')) {
-    // Raised or flat buttons
-    target.click();
-  } else {
-    // Search first button in target. Could be either:
-    // - button wrapped in LeftLoader component
-    // - button in ButtonGroup (e.g. RaisedButtonWithSplitMenu)
-    // - or both
-    target = target.querySelector('button:first-child');
-    if (target) target.click();
-  }
-};
-
 /**
  * A enhanced material-ui Dialog that can have optional secondary actions
  * and no margins if required.
@@ -127,7 +100,6 @@ export default (props: Props) => {
     secondaryActions,
     actions,
     open,
-    canSubmitLastAction,
     onRequestClose,
     maxWidth,
     noMargin,
@@ -192,16 +164,16 @@ export default (props: Props) => {
         return;
       }
 
-      if (canSubmitLastAction && shouldSubmit(event)) {
+      if (onApply && shouldSubmit(event)) {
         event.stopPropagation();
         const element = document.activeElement;
         if (element) {
           element.blur();
         }
-        findAndClickButton(actionsRef);
+        onApply();
       }
     },
-    [onCloseDialog, canSubmitLastAction]
+    [onCloseDialog, onApply]
   );
 
   return (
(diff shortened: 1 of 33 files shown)


newIDE/app/src/UI/Dialog/index.js
```diff
@@ -40,26 +44,33 @@ const styles = {
...
    * single `onRequestClose`.
    */
   onApply?: ?() => void,
+  /**
+   * If specified, allows user to close the dialog with an action based on
+   * `shouldSubmit` key event. `lastAction` option simulates a click on the
+   * last (often the main) action button given (if this last action is a button
+   * with split menu, it simulates click on the main button).
+   */
+  onSubmit?: 'lastAction' | (() => void),
```
ClementPasteau, line 73: We should avoid props that can have different types (here, `string` and `function`)
We should either have `?string` (and if null, decide to not do anything in the logic)
Or we could also decide to pass the function we want triggered when the submit key is called instead of `lastAction` which would allow us in the future to have some custom logic in there, that would be different than when the user clicks on the action?

ClementPasteau: If we change the `onSubmit` to be a function, this would add some duplication when defining a dialog, but at least it would be clear what it does + you wouldn't need the `findAndClickButton` function

newIDE/app/src/UI/Dialog/index.js
```diff
@@ -86,6 +97,26 @@ type DialogContentStyle = {
   flexDirection?: 'row',
 };
 
+const findAndClickButton = (ref: {| current: ?HTMLElement |}): void => {
```
ClementPasteau, line 100: the risk with this function is if we create a new fancy button in the dialog, the querySelector may not work anymore

newIDE/app/src/UI/Dialog/index.js
AlexandreSi: As a user of the component of the Dialog, I don't wanna spend time thinking about this, because I think that 99% of the time, the onSubmit will consist of simulate a click on the last action (proof is that the possibility to pass a function is not used in all the dialogs I reviewed).
I can instead use a props `clickOnLastActionOnSubmit: boolean = false`

newIDE/app/src/UI/Dialog/index.js
AlexandreSi: We don't add buttons everyday and we would only need to change a bit of logic/a css selector so I'm pretty confident it won't cause troubles before a long time

newIDE/app/src/UI/Dialog/index.js
ClementPasteau: I like having to think what all actions should do when I create a dialog as a dev :) but you're right saying this hasn't been used consistently.
Then yes, let's use a boolean then, `submitLastAction: boolean  = false` ? or `canSubmitLastAction` ?

4ian:
> Dialogs that use SemiControlledTextField were problematic because there's a possible logic to commit on blur. So I added the on blur callback.

Indeed, great idea to force a blur on the focused element. Sounds fair.

>  It seems that there's a race condition that prevents the state to be updated on blur

There is possibly indeed more than one "event loop" involved because of a setTimeout or a promise somewhere I guess. 
This is fine for now.


I know I'm arriving a bit late on this, but I still wonder: instead of simulating on the last action, could we... call `onApply`? 
This would go in the direction of giving information to the dialog about "what is cancelling" (onRequestClose) and "what is applying changes" (onApply).

We could even consider that in the future each dialog would automatically register commands in the command palette to "close and cancel changes" or "close and apply changes".
This goes in the direction of standardising dialogs. For example, we could go one step further later and:
- have the dialog automatically add an action button "Ok", "Apply" or "Close" (leaving the choice of the name as a prop, probably). This would call `onApply` or `onRequestClose` if `onApply` is not specified.
- have the dialog automatically add an action button "Cancel" if `onApply` is specified. This button would call `onRequestClose`.

Dialogs would still have a "additionalActions" prop that can be used to add custom action buttons (much like we already have secondaryActions).

I know that this is a bit late but still... there might be something really interesting about going in the direction of generalising the usage of onApply. This gives more knowledge to the dialog about what is inside and the "semantic" of the dialog.

ClementPasteau:
I was of the same opinion, but also realised that some dialogs don't necessarily have an `onApply` as it defined for the moment, and still could use a CMD+enter action.
For instance, a dialog which doesn't have fields in it, when submitting, doesn't need to "apply" the changes made to the fields, and yet could benefit from a `continue`(?) action. 
For example when accessing the extension installation dialog.

4ian:
Yep, but that's why I think we should:
- audit all/most dialogs to have a `onApply` (fairly easy with Flow => make the onApply mandatory and what for Flow to report all errors). This onApply should do the same as the Ok/Next/Install/whatever button.
- we can still have dialogs without a "onApply", in which case Cmd/Ctrl+Enter will trigger the onRequestClose (like Esc is doing)?

AlexandreSi:
I think I'm having troubles figuring what this `onApply` means. I think it's due to the fact that dialogs are very free to do whatever they want.
As you say, in the future, we could have "less free" dialogs that have to respect a few concepts: "What is cancelling?", "What does it mean to apply the dialog?", etc. So that we can automatically map shortcuts to these concepts.
Maybe instead of having a single UI Dialog, we will have a WaitingDialog, an ActionDialog and maybe an InformativeDialog.

For now:
- We have 57 dialogs without `onApply`
- The main obstacle to doing what you're saying (specifying `onApply` for the 24 dialogs that currently have `canSubmitLastAction`) is that this click simulation has to respect the state of the button (`disabled` or not). If we were to apply your suggestion, we would have to duplicate the disabling logic.

I can take time to have this refacto of having 3 different Dialogs and have a `isDisabled` props for the ActionDialog that would apply to the button and the shortcut. But I guess it would need 2 days to do it.

4ian:
> If we were to apply your suggestion, we would have to duplicate the disabling logic.

Correct. That's a good remark. This means for each dialog were we would add onApply, we would:
- connect the action button and the onApply probably to the same or almost the same callback function (preferably **exactly the same**), 
- ...which would do again the check for the disabling (preferably using **the exact same boolean as passed to the button**)

If we improve Dialog in the future, or create job specific dialogs, we could enforce this semantic.

> I can take time to have this refactor of having 3 different Dialogs and have a isActionDisabled props for the ActionDialog that would apply to both the button and the shortcut. 

Do you think we could do without having the 3 different dialogs?
Just thinking about it quickly, I can see some dialogs being not waiting, informative or action (like the objects or variables editor).
To begin, I was thinking of just standardizing the usage of onApply (though it's still optional, like a waiting dialog has no onApply) (and an informative does have a onApply, but just doing the same as onRequestClose).

newIDE/app/src/UI/Dialog/index.js
```diff
@@ -110,14 +114,20 @@ export default (props: Props) => {
...
-    </React.Fragment>
-  ) : (
-    actions
+  const dialogActions = React.useMemo(
+    () => (
+      <React.Fragment>
+        {secondaryActions && (
+          <div key="secondary-actions">{secondaryActions}</div>
+        )}
+        <div key="actions" ref={actionsRef}>
```
4ian, line 125: This ref should be useless now I think

4ian (approved):
🎉

D8H:
I can't download an extension with the nightly (no issue with the stable one). There is a mismatch React version error.
```
Invariant Violation: Minified React error #321; visit https://reactjs.org/docs/error-decoder.html?invariant=321 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
```
In development mode:
```
Uncaught Invariant Violation: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.
    at invariant (http://localhost:3000/static/js/4.chunk.js:33583:19)
    at Object.throwInvalidHookError (http://localhost:3000/static/js/4.chunk.js:46909:7)
    at Object.useCallback (http://localhost:3000/static/js/4.chunk.js:62043:25)
    at ExtensionInstallDialog.render (http://localhost:3000/static/js/1.chunk.js:2255:77)

The above error occurred in the <ExtensionInstallDialog> component:
    in ExtensionInstallDialog (at ExtensionStore/index.js:124)
    in ExtensionStore (at NewBehaviorDialog.js:264)
```
I checked on the previous commit. It was working fine.

ClementPasteau:
Fix created in #3426

PR #3390, opened 2021-12-21
Change exports to send game id

ClementPasteau:
Do not show in changelog.

Keeping the design choice of having 1 unique link per build, to be able to manage different builds/versions for one single game.
In order to link the builds to a game, we have to send the gameId.

### DIFF
--- a/newIDE/app/src/Export/BrowserExporters/BrowserOnlineCordovaExport.js
+++ b/newIDE/app/src/Export/BrowserExporters/BrowserOnlineCordovaExport.js
@@ -153,7 +153,8 @@ export const browserOnlineCordovaExportPipeline: ExportPipeline<
   launchOnlineBuild: (
     exportState: ExportState,
     authenticatedUser: AuthenticatedUser,
-    uploadBucketKey: string
+    uploadBucketKey: string,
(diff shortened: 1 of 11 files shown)


tristanbob:
Can we make an option to make a unique URL?  I have been posting my builds like a dev log and I love the ability to click on my Twitter links and experience what the game felt like and compare it to my current build.

ClementPasteau:
Interesting idea, I didn't know people had a use for the multiple urls!
Technically it is feasible yes, we could keep both behaviours and have some kind of toggle in the interface:
- Deploy to a unique URL (as it's doing now) -> to share experimental versions (or history)
- Deploy to the official game URL (with the ID of the game) -> for the "stable" and latest version

4ian:
What about two "stages"? :) We could offer to deploy on a "live" URL or a test one. Or keep temporary, unique URLs but that are removed after X days (the live one stays forever and can be updated)

ClementPasteau:
Yep you're right, I've updated the type to indicate that it can be undefined.
However, all the functions to create builds will all have one defined from now on.

4ian (approved):

PR #3391, opened 2021-12-21
Fix/export profile calls

ClementPasteau:
Do not show in changelog


newIDE/app/src/Export/ExportDialog/ExportLauncher.js
```diff
@@ -121,7 +121,9 @@ export default class ExportLauncher extends Component<Props, State> {
       builds: [this.state.build],
       onBuildUpdated: (build: Build) => {
```
ClementPasteau, line 122: This function is called every time a build is polled to check its status. When the status turns to 'done' or 'error', then the poll stops (after the last call)

AlexandreSi (approved):

PR #3396, opened 2021-12-22
Add possibility to collapse and expand events to certain levels

AlexandreSi:
Add 2 possibilities in context menu when clicking on event
- Collapse All
- Expand to given level or all levels
demo_collapse_events

Notes:
- I couldn't add a function to collapse/expand only selected events: the way we treat a multiselection involves copying events (via `InsertNewEvent`) so when we call `SetFolded` on them, it doesn't impact the project events.
- We could add shortcuts but I don't know the standard: Ctrl+0 to collapse all and Ctrl+Y to expand to level Y? I don't know what would be a "expand all" shortcut.
- Technical: I didn't use an `ArbitraryEventsWorker`, it could be done quite easily but I don't see what benefits we would take out of it.


Core/GDCore/IDE/Events/EventsListUnfolder.h
```diff
@@ -19,4 +19,29 @@ class GD_CORE_API EventsListUnfolder {
...
+  static void UnfoldToLevel(gd::EventsList& list,
+                            const int8_t maxLevel,
+                            const std::size_t currentLevel = 0) {
+    if (maxLevel >= 0 && currentLevel > maxLevel) return;
+
+    for (size_t i = 0; i < list.size(); ++i) {
+      gd::BaseEvent& event = list[i];
+      event.SetFolded(false);
+      if (event.CanHaveSubEvents() && event.GetSubEvents().size() > 0 &&
+          (maxLevel == -1 || currentLevel <= maxLevel)) {
```
ClementPasteau, line 42: maybe I'd add a comment above the function, explaining that when `maxLevel == -1` it means, unfold until reaching max

newIDE/app/src/EventsSheet/index.js
```diff
@@ -671,6 +679,27 @@ export class EventsSheetComponentWithoutHandle extends React.Component<
...
+    },
+    {
+      label: i18n._(t`Expand all to level`),
+      submenu: [
+        {
+          label: i18n._(t`All`),
+          click: () => this.expandToLevel(-1),
+        },
+        { type: 'separator' },
+        ...[0, 1, 2, 3, 4, 5, 6, 7, 8].map(index => {
```
ClementPasteau, line 694: I'd be interested to hear users' feedback on wether 0-8 is enough or not

AlexandreSi: I remember seeing this 1 to 9 depth. The good thing is it matches the keyboard if we want to add shortcuts.

ClementPasteau (approved):
lgtm :)

4ian (approved):
lgtm!

Bouh:
> We could add shortcuts but I don't know the standard: Ctrl+0 to collapse all and Ctrl+Y to expand to level Y? I don't know what would be a "expand all" shortcut.

CTRL + 0 /1/2/3/4/5 ... is nice. 

Don't worry too much about shortcuts because we'll find a satisfying layout with over time and new shortcuts if we have more to come later.
The shortcuts are editable in the preferences so it's not annoying if it's not so perfect, in any case the expand-all, collapse-all it is absolutely necessary it's so great for productivity! 🤩

Bouh:
For a correct shortcut, I suggest keeping the left hand on the keyboard and the right hand on the mouse.
That seems ridiculous but a shortcut have to be effective with a real benefit.
CTRL+Y starting to be hard for kids with little hands :p 
If we have to drop the mouse to press two keys far away with two hands it is faster to do the same thing with the pointer ^^

It's better if the shortcut has a toggle function on it.

**CTRL+Space** to collapse first level and same for expanding on selection.
And **CTRL+SHIFT+Space** = Expand all/Collapse all on selection.

PR #3397, opened 2021-12-22
Add missing input keys for events

### LINKED ISSUE #3392: NumpadReturn key isn't listed in the auto complete
## Describe the bug
`NumpadReturn` with keycode `3013` isn't visible in the autocomplete for some reason.

image

## Other details
* 5.0.124
* Web app
* Desktop
* W10

ClementPasteau:
Fixes #3392 

I've copied the keys being handled in `inputtools.ts` and reordered them as well to be the same.

image

There are 3 keys suggested in the autocomplete that don't seem handled by the engine: `LBracket`, `RBracket`, `Semicolon`
I'd be tempted to remove them, but also wary I'm missing something.

cc @Bouh


newIDE/app/src/EventsSheet/ParameterFields/KeyField.js
```diff
@@ -95,12 +105,9 @@ const keyNames = [
...
   'Pause',
-  'RControl',
-  'LControl',
-  'RAlt',
-  'LAlt',
-  'RShift',
-  'LShift',
+  'LBracket', // seems not working
+  'RBracket', // seems not working
+  'SemiColon', // seems not working
```
4ian, line 110: Semicolon seems working according to https://keycode.info/ (keycode 186) (notice the lowercase c). Might be something we need to add in the game engine?

4ian (approved):
Approving this to go ahead after checking the semicolon :)

GDJS/Runtime/events-tools/inputtools.ts
```diff
@@ -68,18 +68,18 @@ namespace gdjs {
         RAlt: 2018,
         LSystem: 1091,
         RSystem: 2091,
-        /*"Menu": sf::Keyboard::Menu ,
```
ClementPasteau, line 71: I went back to 2013 to understand why those values had been commented, but looks like they were added commented from the beginning :)
Something I'm missing or can we add them without risk?

4ian: Can be added without risk.

4ian: Can be added without risk.

4ian: Can be added without risk!

4ian: Can be added without risk!

PR #3400, opened 2021-12-23
Fix getUserHomePath action

Bouh:
Fix https://forum.gdevelop-app.com/t/userhomepath-doesnt-working-gives-error/37093


4ian:
Thanks!

PR #3402, opened 2021-12-23
[ShapePainter] Rotation, scaling and fliping

D8H:
It allows users to move a shape painter like a sprite.
Extensions can get the relative coordinates to handle mouse events.
ShapePainterTranformationExpression

Test project: https://www.dropbox.com/s/bf8i3ototgwvjqn/ShapePainterTransformationTest.zip?dl=1
Test video: https://user-images.githubusercontent.com/2611977/147290697-6e4108bc-6ccd-4488-9cf4-849b91c51069.mp4

### DIFF
--- a/Extensions/PrimitiveDrawing/Extension.cpp
+++ b/Extensions/PrimitiveDrawing/Extension.cpp
@@ -669,7 +669,7 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
   obj.AddAction("FlipX",
                 _("Flip the object horizontally"),
                 _("Flip the object horizontally"),
-                _("Flip horizontally _PARAM0_ : _PARAM1_"),
+                _("Flip horizontally _PARAM0_: _PARAM1_"),
                 _("Effects"),
                 "res/actions/flipX24.png",
                 "res/actions/flipX.png")
@@ -680,7 +680,7 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
   obj.AddAction("FlipY",
                 _("Flip the object vertically"),
                 _("Flip the object vertically"),
-                _("Flip vertically _PARAM0_ : _PARAM1_"),
+                _("Flip vertically _PARAM0_: _PARAM1_"),
                 _("Effects"),
                 "res/actions/flipY24.png",
                 "res/actions/flipY.png")
@@ -731,7 +731,7 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
   obj.AddAction("RotationAnchor",
                 _("Center of rotation"),
                 _("Change the center of rotation of an object relatively to the object origin."),
-                _("Change the center of rotation of _PARAM0_ : _PARAM1_; _PARAM2_"),
+                _("Change the center of rotation of _PARAM0_: _PARAM1_; _PARAM2_"),
                 _("Angle"),
                 "res/actions/position24.png",
                 "res/actions/position.png")
@@ -741,7 +741,7 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
       .MarkAsAdvanced();
 
   obj.AddExpression("ToDrawingX",
-                    _("X drawing coordinate of a point form the scene"),
+                    _("X drawing coordinate of a point from the scene"),
                     _("X drawing coordinate of a point from the scene"),
                     _("Position"),
                     "res/actions/position.png")
@@ -750,7 +750,7 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
       .AddParameter("expression", _("Y scene position"));
 
   obj.AddExpression("ToDrawingY",
-                    _("Y drawing coordinate of a point form the scene"),
+                    _("Y drawing coordinate of a point from the scene"),
                     _("Y drawing coordinate of a point from the scene"),
                     _("Position"),
                     "res/actions/position.png")
@@ -759,7 +759,7 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
       .AddParameter("expression", _("Y scene position"));
 
   obj.AddExpression("ToSceneX",
-                    _("X scene coordinate of a point form the drawing"),
+                    _("X scene coordinate of a point from the drawing"),
                     _("X scene coordinate of a point from the drawing"),
                     _("Position"),
                     "res/actions/position.png")
@@ -768,7 +768,7 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
       .AddParameter("expression", _("Y drawing position"));
 
   obj.AddExpression("ToSceneY",
-                    _("Y scene coordinate of a point form the drawing"),
+                    _("Y scene coordinate of a point from the drawing"),
                     _("Y scene coordinate of a point from the drawing"),
                     _("Position"),
                     "res/actions/position.png")
--- a/Extensions/PrimitiveDrawing/Extension.cpp
+++ b/Extensions/PrimitiveDrawing/Extension.cpp
@@ -5,6 +5,7 @@ Copyright (c) 2008-2016 Florian Rival (Florian.Rival@gmail.com)
 This project is released under the MIT License.
 */
 
+#include "GDCore/Extensions/Metadata/MultipleInstructionMetadata.h"
 #include "GDCore/Extensions/PlatformExtension.h"
 #include "GDCore/Tools/Localization.h"
 #include "ShapePainterObject.h"
@@ -643,42 +644,27 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
       .UseStandardOperatorParameters("number")
       .MarkAsAdvanced();
 
-  obj.AddAction("ScaleX",
+  obj.AddExpressionAndConditionAndAction("number",
+                "ScaleX",
                 _("Scale on X axis"),
-                _("Modify the scale of the width of an object."),
+                _("the width's scale of an object"),
                 _("the width's scale"),
                 _("Size"),
-                "res/actions/scale24.png",
-                "res/actions/scale.png")
+                "res/actions/scaleWidth24.png")
       .AddParameter("object", _("Object"), "Drawer")
-      .UseStandardOperatorParameters("number")
+      .UseStandardParameters("number")
       .MarkAsAdvanced();
 
-  obj.AddAction("ScaleY",
+  obj.AddExpressionAndConditionAndAction("number",
+                "ScaleY",
                 _("Scale on Y axis"),
-                _("Modify the scale of the height of an object."),
+                _("the height's scale of an object"),
                 _("the height's scale"),
                 _("Size"),
-                "res/actions/scale24.png",
-                "res/actions/scale.png")
+                "res/actions/scaleHeight24.png")
       .AddParameter("object", _("Object"), "Drawer")
-      .UseStandardOperatorParameters("number")
+      .UseStandardParameters("number")
       .MarkAsAdvanced();
-
-  obj.AddExpression("ScaleX",
-                    _("Scale of the width of an object"),
-                    _("Scale of the width of an object"),
-                    _("Size"),
-                    "res/actions/scaleWidth.png")
-      .AddParameter("object", _("Object"), "Drawer");
-
-  obj.AddExpression("ScaleY",
-                    _("Scale of the height of an object"),
-                    _("Scale of the height of an object"),
-                    _("Size"),
-                    "res/actions/scaleHeight.png")
-      .AddParameter("object", _("Object"), "Drawer");
-
 
   obj.AddAction("FlipX",
                 _("Flip the object horizontally"),
--- a/Extensions/PrimitiveDrawing/Extension.cpp
+++ b/Extensions/PrimitiveDrawing/Extension.cpp
@@ -702,6 +702,23 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
       .AddParameter("yesorno", _("Activate flipping"))
       .MarkAsSimple();
 
+  obj.AddCondition("FlippedX",
+                   _("Horizontally flipped"),
+                   _("Check if the object is horizontally flipped"),
+                   _("_PARAM0_ is horizontally flipped"),
+                   _("Effects"),
+                   "res/actions/flipX24.png",
+                   "res/actions/flipX.png")
+      .AddParameter("object", _("Object"), "Drawer");
+
+  obj.AddCondition("FlippedY",
+                   _("Vertically flipped"),
+                   _("Check if the object is vertically flipped"),
+                   _("_PARAM0_ is vertically flipped"),
+                   _("Effects"),
+                   "res/actions/flipY24.png",
+                   "res/actions/flipY.png")
+      .AddParameter("object", _("Object"), "Drawer");
 
   obj.AddAction("Width",
                 _("Width"),
--- a/Extensions/PrimitiveDrawing/Extension.cpp
+++ b/Extensions/PrimitiveDrawing/Extension.cpp
@@ -632,5 +632,98 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
       .SetFunctionName("AreCoordinatesRelative")
       .SetIncludeFile("PrimitiveDrawing/ShapePainterObject.h");
 
+  obj.AddAction("Scale",
+                _("Scale"),
+                _("Modify the scale of the specified object."),
+                _("the scale"),
+                _("Size"),
+                "res/actions/scale24.png",
+                "res/actions/scale.png")
+      .AddParameter("object", _("Object"), "Drawer")
+      .UseStandardOperatorParameters("number")
+      .MarkAsAdvanced();
+
+  obj.AddAction("ScaleX",
+                _("Scale on X axis"),
+                _("Modify the scale of the width of an object."),
+                _("the width's scale"),
+                _("Size"),
+                "res/actions/scale24.png",
+                "res/actions/scale.png")
+      .AddParameter("object", _("Object"), "Drawer")
+      .UseStandardOperatorParameters("number")
+      .MarkAsAdvanced();
+
+  obj.AddAction("ScaleY",
+                _("Scale on Y axis"),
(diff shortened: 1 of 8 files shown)


Extensions/PrimitiveDrawing/shapepainterruntimeobject-pixi-renderer.ts
```diff
@@ -268,21 +301,165 @@ namespace gdjs {
...
+    }
+
+    updatePositionX(): void {
+      if (this._object._absoluteCoordinates) {
+        this._graphics.pivot.x = 0;
         this._graphics.position.x = 0;
+      } else {
+        // Make the drawing rotate around the anchor.
+        this._graphics.pivot.x = this._object.getRotationAnchorX();
+        // Multiply with the scale to have the scale anchor
```
ClementPasteau, line 321: nit: `Multiply by`

Extensions/PrimitiveDrawing/shapepainterruntimeobject-pixi-renderer.ts
```diff
@@ -4,13 +4,29 @@ namespace gdjs {
...
+      y: 0,
+    };
 
     constructor(
       runtimeObject: gdjs.ShapePainterRuntimeObject,
       runtimeScene: gdjs.RuntimeScene
     ) {
       this._object = runtimeObject;
       this._graphics = new PIXI.Graphics();
+      //this._container.addChild(this._graphics);
```
ClementPasteau, line 29: remove?

Extensions/PrimitiveDrawing/shapepainterruntimeobject-pixi-renderer.ts
```diff
@@ -4,13 +4,29 @@ namespace gdjs {
   class ShapePainterRuntimeObjectPixiRenderer {
     _object: gdjs.ShapePainterRuntimeObject;
     _graphics: PIXI.Graphics;
+    /**
+     * Graphics positions can need updates when shapes are added,
+     * this avoids to do it each time.
+     */
+    _positionXIsUpToDate = false;
```
ClementPasteau, line 11: any reason why we initialise at false? (it feels conter-intuitive to me)

Extensions/PrimitiveDrawing/shapepainterruntimeobject-pixi-renderer.ts
```diff
@@ -268,21 +301,165 @@ namespace gdjs {
...
+          this._graphics.pivot.x * Math.abs(this._graphics.scale.x);
       }
     }
 
-    updateYPosition(): void {
-      if (!this._object._absoluteCoordinates) {
-        this._graphics.position.y = this._object.y;
-      } else {
+    updatePositionY(): void {
+      if (this._object._absoluteCoordinates) {
```
ClementPasteau, line 330: maybe a lack of knowledge from my end, so apologies! 
But, what's the added value of allowing the object to have absolute coordinates?

Also, should this be called `useAbsoluteCoordinates` to make it clear this is a boolean and not actual coordinates?

ClementPasteau (commented):
It looks great! 🎉

Extensions/PrimitiveDrawing/shapepainterruntimeobject-pixi-renderer.ts
```diff
@@ -268,21 +301,165 @@ namespace gdjs {
...
+      }
+      // updatePositionY() uses scale.y
+      this._positionYIsUpToDate = false;
+    }
+
+    getDrawableX(): float {
+      if (this._object._absoluteCoordinates) {
+        return this._graphics.getLocalBounds().x;
+      }
+      let localBound = this._graphics.getLocalBounds().left;
```
4ian, line 384: nitpicking: let's either use x or left, but be consistent to avoid confusion for the reader :) (same for y/top below)

4ian: (unless they mean something different, but I don't think so?)

4ian (commented):
Looks great! I've added a bunch of nitpicking/minor comments.

Do you think we should also allow to define a custom hitbox? We could imagine having a shape painter draw arbitrary stuff but wanting to keep always the same hitbox.
Let me know what you think :)

Extensions/PrimitiveDrawing/shapepainterruntimeobject-pixi-renderer.ts
D8H: This is an existing property that allows user to draw directly in the scene basis without having every ShapePainter instances stacked at (0; 0) in the instance editor.

Yes, this name is less confusing.

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -632,5 +633,148 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+                _("Change the height of an object."),
+                _("the height"),
+                _("Size"),
+                "res/actions/scale24.png",
+                "res/actions/scale.png")
+      .AddParameter("object", _("Object"), "Drawer")
+      .UseStandardOperatorParameters("number")
+      .MarkAsAdvanced();
+
+  obj.AddAction("RotationAnchor",
```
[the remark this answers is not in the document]
D8H: None of the property setters actions start with a "Set".
Is this a legacy thing?
Is a "Set" prefix is preferred over consistency for this extension?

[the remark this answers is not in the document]
4ian: Yep that's a legacy bad decision. While we can name conditions and expressions with the name of the "property" directly, because the "get" is implied (an expression is always a getter, a condition is always a check), the actions are "verbs" (like a function changing something, which should start with a verb). 
I prefer to always use a verb, and in 90% of the cases "Set", to 1) better differentiate actions and conditions and 2) be consistent with `AddExpressionAndConditionAndAction`.

This follows the idea that conditions/actions/expressions are like this code:

```
if (something.Angle() < 30) { // A condition
  something.SetAngle( // An action
    something.Angle() + 10 // An expression
  )  
}
```

We could argue that:
- some "fluent apis", popularised by jQuery, allow to use the same function as a getter (without args) and a setter (with args), but while this is "pretty", it makes static analysis harder. Not a problem for us, but still we'll avoid this.
- if we use "Set", then we should use "Get", but we don't do that for 1) conciseness and 2) conditions are a bit more than a getter, they are a way to compare/check something.

So starting from now I'll try to enforce the "Set" for actions :)

Extensions/PrimitiveDrawing/shapepainterruntimeobject.ts
```diff
@@ -438,23 +450,294 @@ namespace gdjs {
...
+    isFlippedX(): boolean {
+      return this._flippedX;
+    }
+
+    isFlippedY(): boolean {
+      return this._flippedY;
+    }
+
+    /**
+     * Get the scale of the object (or the geometric mean of the X and Y scale in case they are different).
```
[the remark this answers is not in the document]
D8H: The geometric mean gives the same value for the same surface:
Scale 4 is 4*4=16 times the base surface
Scale (1; 16) = 4 is 1*16=16 times the base surface

The square root is more costly but the value can actually be used whereas it's almost impossible to make sense out of a geometric mean.
For instance, it could be used to square a rectangular scaling: object.setScale(getScale()).

Extensions/PrimitiveDrawing/shapepainterruntimeobject.ts
```diff
@@ -438,23 +450,294 @@ namespace gdjs {
...
+
+    transformToSceneX(x: float, y: float) {
+      return this.transformToScene(x, y)[0];
+    }
+
+    transformToSceneY(x: float, y: float) {
+      return this.transformToScene(x, y)[1];
+    }
+
+    updateHitBoxes(): void {
```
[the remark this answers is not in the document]
D8H: It would make it harder to understand the "absolute coordinates" impact. If we wanted to really rely on the base implementation we should probably use a protected method with an extra boolean but this is very specific to the ShapePainter so it will probably pollute the base class.
To over-engineering it, it could be abstract and implemented with an hitbox builder.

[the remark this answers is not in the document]
4ian: Yep, I was thinking about a :

```
if (!absolute coordinates) { 
  super.updateHitBoxes(); // Use getDrawableX, getDrawableY, getWidth, getHeight 
} else { ... }
```

But might not be worth it.

Extensions/PrimitiveDrawing/shapepainterruntimeobject-pixi-renderer.ts
4ian: Yeah the naming is a bit confusing - though it was already there so nothing blocking for this PR.

Extensions/PrimitiveDrawing/shapepainterruntimeobject-pixi-renderer.ts
D8H: Done (it was not marked as outdated)

Extensions/PrimitiveDrawing/Extension.cpp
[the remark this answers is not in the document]
D8H: Ok, `SetWidth` too or this could make it hard if the action move up to Object?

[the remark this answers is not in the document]
4ian: Good question.. let's keep `Width` then for consistency with the base object. And use "Set" for all "new" actions :)

D8H:
> Do you think we should also allow to define a custom hitbox? We could imagine having a shape painter draw arbitrary stuff but wanting to keep always the same hitbox. Let me know what you think :)

We probably should. Do we add an action to set a rectangle hitbox? No polygon handling?

The custom hitbox and custom center could even also be properties but I guess it will probably need to be calculate so they might not be useful.

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -632,5 +633,148 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+                _("Size"),
+                "res/actions/scaleHeight24.png")
+      .AddParameter("object", _("Object"), "Drawer")
+      .UseStandardParameters("number")
+      .MarkAsAdvanced();
+
+  obj.AddAction("FlipX",
+                _("Flip the object horizontally"),
+                _("Flip the object horizontally"),
+                _("Flip horizontally _PARAM0_ : _PARAM1_"),
```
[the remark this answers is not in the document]
D8H: I made the changes in Sprite too, but do we loose the translation?

Extensions/PrimitiveDrawing/shapepainterruntimeobject.ts
```diff
@@ -438,23 +450,294 @@ namespace gdjs {
...
+      if (!this._customCenter) {
+        this._customCenter = [0, 0];
+      }
+      this._customCenter[0] = x;
+      this._customCenter[1] = y;
+    }
+
+    /**
+     * @returns The center X relatively to the drawing origin
+     * (where `getCenterX()` is relative to the top left drawable bound and scaled).
```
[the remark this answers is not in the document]
D8H: Changed in Sprite too.

Extensions/PrimitiveDrawing/shapepainterruntimeobject.ts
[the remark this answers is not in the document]
D8H: If you're ok with this solution, do I harmonize it for the other objects in another PR?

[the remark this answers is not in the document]
4ian: This would be a breaking change for other objects, so might need to think if this is worth it?

4ian:
> We probably should. Do we add an action to set a rectangle hitbox? No polygon handling?

Let's start with an action to set a rectangle hitbox. Because:
- it's what we'll need in most cases.
- we can always add support for polygons later - but I would see this part of a larger effort to add support for custom hitboxes for all objects.

Extensions/PrimitiveDrawing/Extension.cpp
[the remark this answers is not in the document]
4ian: We sadly loose the translation, but Crowdin will suggest the old translation because it will see that the source string is very similar. So it's "not too bad". 
Correctness in english is more important anyway.

Extensions/PrimitiveDrawing/shapepainterruntimeobject.ts
```diff
@@ -438,23 +450,294 @@ namespace gdjs {
...
+    }
+
+    getCenterX(): float {
+      if (!this._customCenter) {
+        return super.getCenterX();
+      }
+      let centerX =
+        this._customCenter[0] +
+        (this.getX() - this.getDrawableX()) / Math.abs(this._scaleX);
+      return centerX * Math.abs(this._scaleX);
```
4ian, line 519: Can we simplify in:

```suggestion
        return this._customCenter[0] * Math.abs(this._scaleX) + this.getX() - this.getDrawableX();
```
?

Extensions/PrimitiveDrawing/Extension.cpp
```diff
@@ -632,5 +633,162 @@ void DeclarePrimitiveDrawingExtension(gd::PlatformExtension& extension) {
...
+  obj.AddExpressionAndConditionAndAction("number",
+                "ScaleX",
+                _("Scale on X axis"),
+                _("the width's scale of an object"),
+                _("the width's scale"),
+                _("Size"),
+                "res/actions/scaleWidth24.png")
+      .AddParameter("object", _("Object"), "Drawer")
+      .UseStandardParameters("number")
+      .MarkAsAdvanced();
```
D8H, line 656: I should probably revert the use of `AddExpressionAndConditionAndAction` to keep the same name as Sprite to make a generalization easier.

4ian: I'm not sure... if we generalize this later, the full name of the actions (PrimitiveDrawing::ScaleX) will be different anyway. We'll probably move these to the base RuntimeObject and "redirect" any existing action to the base RuntimeObject methods.

What will be important will be to easily find all the actions and ensure we've followed the exact same parameters. In this sense, AddExpressionAndConditionAndAction actually helps because it's less code to update in the future :)

Extensions/PrimitiveDrawing/tests/shapepainterruntimeobject.pixiruntimegamewithassets.spec.js
```diff
@@ -0,0 +1,135 @@
...
+    expect(object.getWidth()).to.be(13 * 2); // ...Neither is the size
+    expect(object.getHeight()).to.be(23 * 2);
+    expect(object.getAABB()).to.eql({
+      // The hit boxes/AABB are rotated:
+      min: [4.54415587728429, 14.54415587728429],
+      max: [55.45584412271571, 65.45584412271572],
+    });
+  });
+
+  it.only('properly computes bounds of the object (custom center)', async () => {
```
D8H, line 82: There is an only.

4ian: Thanks for checking after me :)

Extensions/PrimitiveDrawing/shapepainterruntimeobject-pixi-renderer.ts
```diff
@@ -268,20 +305,177 @@ namespace gdjs {
...
+        this.updatePositionY();
+        this._positionYIsUpToDate = true;
+      }
+    }
+
+    updateTransformationIfNeeded() {
+      if (!this._transformationIsUpToDate) {
+        this.updatePositionIfNeeded();
+        this._graphics.updateTransform();
+      }
```
4ian, line 362: @D8H We're missing a `this._transformationIsUpToDate = true` here no?

Extensions/PrimitiveDrawing/shapepainterruntimeobject-pixi-renderer.ts
```diff
@@ -268,20 +305,177 @@ namespace gdjs {
...
+    /**
+     * @returns The drawing origin relatively to the drawable top left corner.
+     */
+    getOriginY() {
+      return -this._graphics.getLocalBounds().top;
+    }
+
+    transformToDrawing(point: FloatPoint): FloatPoint {
+      this.updateTransformationIfNeeded();
+      this._graphics.updateTransform();
```
4ian, line 458: This seems useless, I'm removing this line.

D8H: This is useful. Otherwise, there is a one frame delay. Try to change the angle by 20° and display the hitboxes.

D8H: No, sorry, the call to this._graphics.updateTransform is done by updateTransformationIfNeeded. It can be removed.

4ian: Thanks for checking 👍

Extensions/PrimitiveDrawing/shapepainterruntimeobject-pixi-renderer.ts
4ian: I've added it and fixed an issue where setting center was not updating the transform :)

4ian (approved):
I think this is good now! I hope we've not forgot any case where the transform/position are not updated properly.

I'm not entirely satisfied with the name `getOriginX`, as it's the origin from the drawable X position... but it's just in the renderer so not that terrible. Maybe we could name it something like `getOriginXFromDrawableX` or something like this. Not pretty at all, but at least explains what it really is.

Apart from that, seems all good! Super glad to have this :)

D8H:
> I'm not entirely satisfied with the name `getOriginX`, as it's the origin from the drawable X position... but it's just in the renderer so not that terrible. Maybe we could name it something like `getOriginXFromDrawableX` or something like this. Not pretty at all, but at least explains what it really is.

Maybe a parallel can be done with `this._animationFrame.getPoint('origin')` from sprites: `getInFrameOriginX`, `getFrameRelativeOriginX` or `getRelativeOriginX`?

4ian:
getFrameRelativeOriginX seems a good idea!

PR #3403, opened 2021-12-24
[VisualDebugger] Show objects position instead of origin.

D8H:
This the change for this proposition:
* https://github.com/4ian/GDevelop/discussions/3395

### DIFF
--- a/GDJS/Runtime/pixi-renderers/runtimescene-pixi-renderer.ts
+++ b/GDJS/Runtime/pixi-renderers/runtimescene-pixi-renderer.ts
@@ -230,11 +230,9 @@ namespace gdjs {
         debugDraw.fill.alpha = 0.3;
 
         // Draw Center point
-        const centerPointX = object.getDrawableX() + object.getCenterX();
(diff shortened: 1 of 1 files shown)


Silver-Streak:
I didn't get a chance to respond in the discussion, but one of the main things when the visual debugger was getting added was to show all points. So the origin point *should not* be replaced as users will still need visual representation of its placement, although I agree showing the position point *in addition* is helpful

4ian:
Agree that we can probably:
- have "Position" as being the "unrotated, unscaled, unflipped" X;Y position of the object.
- In case the origin point is not at the exact same position, then we show it too. This can be the case for sprite objects that have a "origin" point, but which is impacted by the rotation/scale/flip.
  - When the origin is exactly at the object position, let's not show it to avoid cluttering the rendering :)

D8H:
Tested on:
* an unrotated Sprite: only the position label and point is visible.
* a rotating Sprite: the origin point is shown and makes its loop.

4ian:
Thanks! Looks much better to always show the proper position :) And the origin/points for sprites (and we can extend to other objects in the future).

PR #3406, opened 2021-12-26
[Platformer] Add a condition to check which controls are used.

D8H:
This adds a condition to check which controls are used.

This is useful for extensions that need to know which keys were pressed but doesn't know the mapping done by the scene events (that could use a game-pad...).

This extension review gives more context:
* https://github.com/GDevelopApp/GDevelop-extensions/pull/325#issuecomment-1001016702

I didn't do a release condition because it would have take an extra 8 bytes and extensions can do it by themselves, but it would have been more convenient.
Do you think we should add it?

Test project:
* https://www.dropbox.com/s/3z5fblixwxgdrmq/PlatformerUseControl.zip?dl=1

### DIFF
--- a/Extensions/PlatformBehavior/Extension.cpp
+++ b/Extensions/PlatformBehavior/Extension.cpp
@@ -501,9 +501,9 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
         .SetFunctionName("SimulateControl");
 
     aut.AddScopedCondition("IsUsingControl",
-                  _("Use control"),
-                  _("A control was applied from a default control or a simulate action."),
-                  _("_PARAM0_ uses the _PARAM2_ key"),
+                  _("Control pressed or simulated"),
+                  _("A control was applied from a default control or a simulated by an action."),
+                  _("_PARAM0_ has the _PARAM2_ key pressed or simulated"),
                   _("Controls"),
                   "res/conditions/keyboard24.png",
                   "res/conditions/keyboard.png")
--- a/Extensions/PlatformBehavior/Extension.cpp
+++ b/Extensions/PlatformBehavior/Extension.cpp
@@ -500,6 +500,20 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
         .MarkAsAdvanced()
         .SetFunctionName("SimulateControl");
 
+    aut.AddScopedCondition("IsUsingControl",
+                  _("Use control"),
+                  _("A control was applied from a default control or a simulate action."),
+                  _("_PARAM0_ uses the _PARAM2_ key"),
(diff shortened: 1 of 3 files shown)


Extensions/PlatformBehavior/Extension.cpp
```diff
@@ -500,6 +500,20 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
         .MarkAsAdvanced()
         .SetFunctionName("SimulateControl");
 
+    aut.AddScopedCondition("IsUsingControl",
+                  _("Use control"),
```
4ian, line 504: Let's name this "Control pressed or simulated"

Extensions/PlatformBehavior/Extension.cpp
```diff
@@ -500,6 +500,20 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
...
+                  _("A control was applied from a default control or a simulate action."),
```
4ian, line 505:
```suggestion
                  _("A control was applied from a default control or a simulated by an action."),
```

Extensions/PlatformBehavior/Extension.cpp
```diff
@@ -500,6 +500,20 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
...
+                  _("_PARAM0_ uses the _PARAM2_ key"),
```
4ian, line 506:
```suggestion
                  _("_PARAM0_ has the _PARAM2_ key pressed or simulated"),
```

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -1007,6 +1027,38 @@ namespace gdjs {
       }
     }
 
+    /**.
+     * @param input The string expression of the control action [Left,Right,Up,Down,Ladder,Jump,Release,Release Ladder].
```
4ian, line 1031:
```suggestion
     * @param input The control to be tested [Left,Right,Up,Down,Ladder,Jump,Release,Release Ladder].
```

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -1007,6 +1027,38 @@ namespace gdjs {
...
+     * @returns true if the key was used from the last `doStepPreEvents` call.
```
4ian, line 1032:
```suggestion
     * @returns true if the key was used since the last `doStepPreEvents` call.
```

4ian (approved):
Good to go after comments are addressed! Thanks :)

4ian (approved):

PR #3408, opened 2021-12-26
[Platformer] Add 2 tests on jump at 30 fps and 120 fps.

D8H:
These tests show why Verlet integration should be used.

The difference is sensible.
* at 30 fps the max height delta is 165 pixels
* at 60 fps the max height delta is 150 pixels
* at 120 fps the max height delta is 142.5 pixels


4ian (approved):
Very clear and useful thanks!

PR #3410, opened 2021-12-27
Display builds in the Game Dashboard

ClementPasteau:
The goal is to start differentiating:
- the builds created to test your game, share easily with the community and keep a history of versions published.
- the live version of the game, the most finished one, which we can put forward to users, gathering them in some kind of game platform

First step: Allow the user to manage their builds!

image
image


newIDE/app/src/GameDashboard/GameDetailsDialog.js
```diff
@@ -225,6 +279,70 @@ export const GameDetailsDialog = ({
             </Line>
           </ColumnStackLayout>
         ) : null}
+        {currentTab === 'builds' ? (
```
ClementPasteau, line 282: I could start extracting this to its own component, but it's quite simple for the moment (may allow build deletion in the future...)

newIDE/app/src/Utils/GDevelopServices/Build.js
```diff
@@ -185,13 +185,15 @@ export const getBuild = (
 
 export const getBuilds = (
   getAuthorizationHeader: () => Promise<string>,
-  userId: string
+  userId: string,
+  gameId?: string
```
ClementPasteau, line 189: should I create 2 functions to show the difference between `getUserBuilds` and `getGameBuild` ?

newIDE/app/src/GameDashboard/GameDetailsDialog.js
```diff
@@ -80,6 +97,10 @@ export const GameDetailsDialog = ({
   );
   const [isGameMetricsLoading, setIsGameMetricsLoading] = React.useState(false);
 
+  const [gameBuilds, setGameBuilds] = React.useState<?(Build[])>(null);
+  const [isGameBuildsLoading, setIsGameBuildsLoading] = React.useState(false);
```
4ian, line 101: Could you consider no error (null) and no builds (null, NOT an empty list) as loading? Would avoid one state.

newIDE/app/src/GameDashboard/GameDetailsDialog.js
4ian: I think that's fine for now, we can extract when needed, or now if you feel like it.

newIDE/app/src/Utils/GDevelopServices/Build.js
4ian: Maybe, could be clearer! Not a big fan of multiplying the number of optional parameters, it's almost always hard to understand. Better either :
- do another function (clearer)
- pass an object as argument for the options (but we don't need this flexibility)

4ian (approved):

PR #3412, opened 2021-12-28
[Top-Down] Make the movement FPS independent.

D8H:

It also does small optimizations
* avoid a sqrt every frame
* avoid an arctan, a cos and a sin when the object stays still.

I didn't do a legacy mode because it only have a small impact on the acceleration. I think it won't be noticeable.


4ian:
Awesome 🤩 I had to expand the deltas to understand how using the previous velocity was helpful.
Super excited to have this fix done.

PR #3417, opened 2021-12-31
[Platformer] Add a frame rate independent movement mode.

D8H:

It also fix the ledge grabbing: when the character is on exactly on right Y, it can now grab.
It add tests to check that the trajectory is the same on different frame rate for jump/fall and walking.

The existing tests have been adapted to this new mode. Only the basic jump test uses the legacy mode. I think it allows to check that the legacy mode works without having to maintain 2 test suites. But, if you want to reduce regression risk during the transition, I can add the old suite in a legacy folder.

I sometimes made tests less constraint. This is to only check what is really matter for the test to help readability and make them easier to maintain.

### DIFF
--- a/Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
+++ b/Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
@@ -1890,7 +1890,9 @@ namespace gdjs {
         behavior._requestedDeltaY -= previousJumpSpeed * timeDelta;
 
         // Fall
-        // This is arbitrary. It used to not be obvious.
+        // The condition is a legacy thing.
+        // There is no actual reason not to fall at 1st frame.
+        // Before a refactoring, it used to not be this obvious.
         if (!this._jumpingFirstDelta) {
           behavior._fall(timeDelta);
         }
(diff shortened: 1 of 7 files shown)


4ian:
So cool! I'll check that tomorrow :)

4ian (commented):
Overall looks great!

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -1842,27 +1870,42 @@ namespace gdjs {
...
       }
+
+      if (this._behavior._useLegacyTrajectory) {
+        behavior._requestedDeltaY -= previousJumpSpeed * timeDelta;
+
+        // Fall
+        // The condition is a legacy thing.
+        // There is no actual reason not to fall at 1st frame.
+        // Before a refactoring, it used to not be this obvious.
+        if (!this._jumpingFirstDelta) {
```
4ian, line 1896: I indeed can't remember why I added this 🤔

4ian (approved):

4ian:
Thank you for working on this! ✨

PR #3421, opened 2021-12-31
[PathFinding] Make the movement frame rate independent.

D8H:

It also does small optimizations
* avoid an arctan, a cos and a sin when the object doesn't change of direction.

I didn't do a legacy mode because it only have a small impact on the acceleration. I think it won't be noticeable.

### DIFF
--- a/Extensions/PathfindingBehavior/tests/commonpathfindingruntimebehavior.spec.js
+++ b/Extensions/PathfindingBehavior/tests/commonpathfindingruntimebehavior.spec.js
@@ -1,10 +1,11 @@
 // @ts-check
-describe('gdjs.PathfindingRuntimeBehavior', function () {
+describe.only('gdjs.PathfindingRuntimeBehavior', function () {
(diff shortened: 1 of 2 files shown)


4ian (approved):

Extensions/PathfindingBehavior/tests/commonpathfindingruntimebehavior.spec.js
```diff
@@ -1,10 +1,11 @@
 // @ts-check
-describe('gdjs.PathfindingRuntimeBehavior', function () {
+describe.only('gdjs.PathfindingRuntimeBehavior', function () {
```
D8H, line 2: I forgot an `only` here, it's still on the master. I'll do a PR.

4ian: Thanks for noticing! We should probably create a linter rule or github action or something like this that warns against `.only`

PR #3426, opened 2022-01-03
Fix Extension Install dialog being functional to use hooks

ClementPasteau:
regression introduced in #3389


4ian (approved):

AlexandreSi (approved):
😅 Thanks for fixing it!

PR #3429, opened 2022-01-04
Add notion of origin to extensions

AlexandreSi:
Do not show in changelog.

Will allow to have a different behavior depending on the origin of the extension (`extensions-store` or else)

### DIFF
--- a/newIDE/app/src/AssetStore/InstallAsset.js
+++ b/newIDE/app/src/AssetStore/InstallAsset.js
@@ -419,6 +419,11 @@ export const addSerializedExtensionsToProject = (
       'unserializeFrom',
       project
     );
+
+    if (serializedExtension.url) {
+      // extensions with url field are extensions that come from the store.
+      newEventsFunctionsExtension.setOrigin('extensions-store', serializedExtension.name)
+    }
   });
 
   return eventsFunctionsExtensionsState.loadProjectEventsFunctionsExtensions(
--- a/newIDE/app/src/AssetStore/InstallAsset.js
+++ b/newIDE/app/src/AssetStore/InstallAsset.js
@@ -400,7 +400,8 @@ export const downloadExtensions = async (
 export const addSerializedExtensionsToProject = (
   eventsFunctionsExtensionsState: EventsFunctionsExtensionsState,
   project: gdProject,
-  serializedExtensions: Array<SerializedExtension>
+  serializedExtensions: Array<SerializedExtension>,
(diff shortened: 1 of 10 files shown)


newIDE/app/src/AssetStore/InstallAsset.js
```diff
@@ -419,6 +419,14 @@ export const addSerializedExtensionsToProject = (
       'unserializeFrom',
       project
     );
+
+    if (serializedExtension.url) {
+      // extensions with url field are extensions that come from the store.
+      newEventsFunctionsExtension.setOrigin(
+        'extensions-store',
```
4ian, line 426: nit: We could even go for 'gdevelop-extensions-store' as we used 'gdevelop-assets-store' I think for resources

AlexandreSi: 👍

AlexandreSi:
I don't know if I should add a function that fills the json read when importing an extension so that the output fits the type `SerializedExtension`

4ian:
> I don't know if I should add a function that fills the json read when importing an extension so that the output fits the type SerializedExtension

I don't know but I think we should not. Apart from validating a few fields, it should be the json of the unserializer to read the extension object (i.e: `EventsFunctionsExtension::UnserializeFrom`, in the C++ code).
I think the problem lies in the SerializedExtension type. I've fixed it in https://github.com/4ian/GDevelop/commit/e2afa946a25df7eab8772b83008368ab92389928

This type is NOT guaranteed to contain the fields of a header. It's probably not, and if it does it's just because the headers are using the JSON of an extension as a basis.

newIDE/app/src/AssetStore/InstallAsset.js
```diff
@@ -419,6 +419,14 @@ export const addSerializedExtensionsToProject = (
       'unserializeFrom',
       project
     );
+
+    if (serializedExtension.url) {
```
4ian, line 423: Are you sure this can even happen? I'm not sure because the SerializedExtension type is a representation of this:

https://resources.gdevelop-app.com/extensions/DoubleClick.json

which does not contain the URL, as this is only in the "header":

https://resources.gdevelop-app.com/extensions/DoubleClick-header.json

The type was sadly misguiding you :(

AlexandreSi: Indeed, I may have confused a few things when testing

4ian (approved):
LGTM in this current state :)

PR #3430, opened 2022-01-04
Discover extensions and behaviors from action or condition list

AlexandreSi:
Add buttons in actions/conditions list to open the NewBehavior dialog or the ExtensionsSearch dialog, depending if an object is selected.

demo_add_extension_on_the_fly

I think that the wording of the buttons can be better : "Discover other actions/conditions in behaviors/extensions" or something like that?

### DIFF
--- a/newIDE/app/src/BehaviorsEditor/NewBehaviorDialog.js
+++ b/newIDE/app/src/BehaviorsEditor/NewBehaviorDialog.js
@@ -31,6 +31,13 @@ import EventsFunctionsExtensionsContext from '../EventsFunctionsExtensionsLoader
 import { installExtension } from '../AssetStore/ExtensionStore/InstallExtension';
 import DismissableInfoBar from '../UI/Messages/DismissableInfoBar';
 import ScrollView, { type ScrollViewInterface } from '../UI/ScrollView';
+import AuthenticatedUserContext from '../Profile/AuthenticatedUserContext';
+import {
+  ACHIEVEMENT_FEATURE_FLAG,
+  addCreateBadgePreHookIfNotClaimed,
+  TRIVIAL_FIRST_BEHAVIOR,
+  TRIVIAL_FIRST_EXTENSION,
+} from '../Utils/GDevelopServices/Badge';
 
 const styles = {
   disabledItem: { opacity: 0.6 },
@@ -89,6 +96,15 @@ export default function NewBehaviorDialog({
   const eventsFunctionsExtensionsState = React.useContext(
     EventsFunctionsExtensionsContext
   );
+  const authenticatedUser = React.useContext(AuthenticatedUserContext);
+
+  const installDisplayedExtension = ACHIEVEMENT_FEATURE_FLAG
+    ? addCreateBadgePreHookIfNotClaimed(
+        authenticatedUser,
+        TRIVIAL_FIRST_EXTENSION,
+        installExtension
+      )
+    : installExtension;
 
   const platform = project.getCurrentPlatform();
   const behaviorMetadata: Array<EnumeratedBehaviorMetadata> = React.useMemo(
@@ -126,7 +142,7 @@ export default function NewBehaviorDialog({
     ({ type }) => !!deprecatedBehaviorsInformation[type]
   );
 
-  const chooseBehavior = (
+  const _chooseBehavior = (
     i18n: I18nType,
     { type, defaultName }: EnumeratedBehaviorMetadata
   ) => {
@@ -136,6 +152,13 @@ export default function NewBehaviorDialog({
 
     return onChoose(type, defaultName);
   };
+  const chooseBehavior = ACHIEVEMENT_FEATURE_FLAG
+    ? addCreateBadgePreHookIfNotClaimed(
+        authenticatedUser,
+        TRIVIAL_FIRST_BEHAVIOR,
+        _chooseBehavior
+      )
+    : _chooseBehavior;
 
   const canBehaviorBeUsed = (behaviorMetadata: EnumeratedBehaviorMetadata) => {
     // An empty object type means the base object, i.e: any object.
@@ -263,12 +286,12 @@ export default function NewBehaviorDialog({
               </React.Fragment>
             )}
             {currentTab === 'search' && (
-              <ExtensionStore // TODO
+              <ExtensionStore
                 project={project}
                 isInstalling={isInstalling}
                 onInstall={async extensionShortHeader => {
                   setIsInstalling(true);
-                  const wasExtensionInstalled = await installExtension(
+                  const wasExtensionInstalled = await installDisplayedExtension(
                     i18n,
                     project,
                     eventsFunctionsExtensionsState,
--- a/newIDE/app/src/BehaviorsEditor/NewBehaviorDialog.js
+++ b/newIDE/app/src/BehaviorsEditor/NewBehaviorDialog.js
@@ -33,7 +33,6 @@ import DismissableInfoBar from '../UI/Messages/DismissableInfoBar';
 import ScrollView, { type ScrollViewInterface } from '../UI/ScrollView';
 import AuthenticatedUserContext from '../Profile/AuthenticatedUserContext';
 import {
-  ACHIEVEMENT_FEATURE_FLAG,
   addCreateBadgePreHookIfNotClaimed,
   TRIVIAL_FIRST_BEHAVIOR,
   TRIVIAL_FIRST_EXTENSION,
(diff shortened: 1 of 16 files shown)


newIDE/app/src/EventsSheet/InstructionEditor/NewInstructionEditorDialog.js
```diff
@@ -146,6 +171,30 @@ export default function NewInstructionEditorDialog({
     }
   };
 
+  const _addBehavior = (type: string, defaultName: string) => {
+    if (!chosenObject) return;
+    setNewBehaviorDialogOpen(false);
```
ClementPasteau, line 176: should the closing of the dialog happen after the behavior is added? In case it fails or if there is a confirmation dialog happening?

newIDE/app/src/EventsSheet/InstructionEditor/NewInstructionEditorDialog.js
```diff
@@ -146,6 +171,30 @@ export default function NewInstructionEditorDialog({
...
+
+    addBehaviorToObject(project, chosenObject, type, defaultName);
+
+    // Re-choose the same object to force recomputation of chosenObjectInstructionsInfoTree
+    chooseObject(chosenObject.getName());
```
ClementPasteau, line 181: is this a hack? I'm not familiar with this panel, just wondering if there's a function for that, like 'forceUpdate'

newIDE/app/src/Utils/Behavior.js
```diff
@@ -0,0 +1,39 @@
+// @flow
+import newNameGenerator from './NewNameGenerator';
+import Window from './Window';
+
+const gd: libGDevelop = global.gd;
+
+export const hasBehaviorWithType = (object: gdObject, type: string) => {
+  const allBehaviorNames = object.getAllBehaviorNames().toJSArray();
+
+  return allBehaviorNames
```
ClementPasteau, line 10: can this be only one filter?

```
return allBehaviorNames.filter(
    behaviorName => object.getBehavior(behaviorName).getTypeName() === type
).length
```

newIDE/app/src/Utils/Behavior.js
```diff
@@ -0,0 +1,39 @@
...
+
+export const addBehaviorToObject = (
+  project: gdProject,
+  object: gdObject,
+  type: string,
+  defaultName: string
+) => {
+  if (hasBehaviorWithType(object, type)) {
+    const answer = Window.showConfirmDialog(
+      "There is already a behavior of this type attached to the object. It's possible to add this behavior again, but it's unusual and may not be always supported properly. Are you sure you want to add this behavior again?"
```
ClementPasteau, line 24: nit: `"may not be always supported"` => `"may not always be supported"`

ClementPasteau (commented):
Look pretty cool!
The only feedback I'd have is to always display the button fixed at the bottom of the dialog, because I feel that people will not think of scrolling to the bottom of the list, especially if they already have loads of conditions/actions available.
Imo, the button isn't big and wouldn't take up much space.

4ian:
I'm not necessarily sure about showing the button at a fixed position, I think we could try with the button at the bottom of the list?

newIDE/app/src/BehaviorsEditor/NewBehaviorDialog.js
```diff
@@ -89,6 +104,15 @@ export default function NewBehaviorDialog({
   const eventsFunctionsExtensionsState = React.useContext(
     EventsFunctionsExtensionsContext
   );
+  const authenticatedUser = React.useContext(AuthenticatedUserContext);
+
+  const installDisplayedExtension = ACHIEVEMENT_FEATURE_FLAG
```
4ian, line 109: I think we can do without the feature flag as it's now added and in production :)

newIDE/app/src/EventsSheet/InstructionEditor/InstructionOrObjectSelector.js
```diff
@@ -104,6 +108,15 @@ export default class InstructionOrObjectSelector extends React.PureComponent<
     this.props.chosenInstructionType
   );
 
+  update = () => {
```
4ian, line 111: Nit: can we call this updateInstructions or reenumerateInstructions?

newIDE/app/src/EventsSheet/InstructionEditor/NewInstructionEditorDialog.js
AlexandreSi: Not really a hack, It's just that the computation of displayed instructions only happens when you choose an object, not when a change in the object behaviors has been detected. And this change happens at the C++ level so there is not much we can do to make this refresh cleaner.

newIDE/app/src/EventsSheet/InstructionEditor/NewInstructionEditorDialog.js
```diff
@@ -146,6 +167,22 @@ export default function NewInstructionEditorDialog({
...
+  };
+
+  const onAddExtension = () => {
```
4ian, line 180: nit: onExtensionAdded or onExtensionInstalled, to show that this is something happening at the end (and not something to be called when the user wants to add an extension)

newIDE/app/src/Utils/Behavior.js
AlexandreSi: True, I'll fix it.

newIDE/app/src/BehaviorsEditor/NewBehaviorDialog.js
AlexandreSi: True, I'm removing them

newIDE/app/src/EventsSheet/InstructionEditor/InstructionOrExpressionSelector/index.js
```diff
@@ -164,6 +170,18 @@ export default class InstructionOrExpressionSelector<
                         initiallyOpenedPath: this.initialInstructionTypePath,
                         selectedItemRef: this._selectedItem,
                       })}
+                      {onClickMore && (
+                        <Line justifyContent="center">
+                          <RaisedButton
+                            primary
+                            icon={<Add />}
+                            onClick={onClickMore}
+                            label={<Trans>Add new behavior</Trans>}
```
4ian, line 179: Would "Add a new behavior to the object" work? Other "Add a new behavior".

newIDE/app/src/EventsSheet/InstructionEditor/NewInstructionEditorDialog.js
4ian: Let's add a comment "This is not done because a change in the object behaviors is not detected by React at this level" or something like this :)

4ian: (One could argue we could pass a key={/*object name + all the names and type of the behaviors in an attempt to create a key reflecting the behaviors of the object*/} - but unsure if this is more clean/efficient or not).

newIDE/app/src/EventsSheet/InstructionEditor/InstructionOrExpressionSelector/index.js
AlexandreSi: It reads better. I will check the display is not cumbersome on mobile

newIDE/app/src/EventsSheet/InstructionEditor/InstructionOrObjectSelector.js
```diff
@@ -356,18 +370,33 @@ export default class InstructionOrObjectSelector extends React.PureComponent<
...
+                              .initialInstructionTypePath,
+                            selectedItemRef: this._selectedItem,
+                          })}
+                          {onClickMore && (
+                            <Line justifyContent="center">
+                              <RaisedButton
+                                primary
+                                icon={<Add />}
+                                onClick={onClickMore}
+                                label={<Trans>Search new extensions</Trans>}
```
4ian, line 394: You suggested  "Discover other actions/conditions in behaviors/extensions", maybe we could have here:

"Search for new actions in extensions" or "Search for new conditions in extensions" (according to isCondition)

or "Find new [...]" if "Search for" is too long?

newIDE/app/src/EventsSheet/InstructionEditor/InstructionOrExpressionSelector/index.js
AlexandreSi: I went for "Add a new behavior to the object" here, and "Add a new extension to the project" for free instructions

AlexandreSi: Oups, didn't see your suggestion below, I will use "Search for new actions in extensions"

ClementPasteau (approved):
lgtm!

PR #3432, opened 2022-01-05
Introduce stable URLs for games

ClementPasteau:
This PR does quite a few things (sorry for the mono-commit!)

In summary:
- improved some components readability + optimisation of network calls
- prevent navigating away from the build dialog at critical moments, resulting in confusing the user when the page is reset (while uploading / while building for web)
- improved the builds list accessible from the Export dialog and from the Games dashboard
  - a web build can be promoted as the latest version for the game
  - a web build can also be unpublished
  - the web build url can be opened directly from the list
  - the list of web builds does not expire any more (only the desktop & mobile builds expire after 7 days)
- change the link of the web build to be:
  - either the stable game page link if the build corresponds to the one linked to the game (especially useful for the first build someone creates, automatically linked to the game)
  - or the build link for subsequent builds, with an option to publish easily to the stable game page
- Added a button to open the stable game page from the game dashboard

Few screenshots:

image
image
image
image
image

### DIFF
--- a/newIDE/app/src/Export/Builds/BuildsDialog.js
+++ b/newIDE/app/src/Export/Builds/BuildsDialog.js
@@ -6,54 +6,26 @@ import Dialog from '../../UI/Dialog';
 import HelpButton from '../../UI/HelpButton';
 import FlatButton from '../../UI/FlatButton';
 import Builds from '.';
-import AuthenticatedUserContext, {
-  type AuthenticatedUser,
-} from '../../Profile/AuthenticatedUserContext';
+import { type AuthenticatedUser } from '../../Profile/AuthenticatedUserContext';
 import useForceUpdate from '../../Utils/UseForceUpdate';
-import { getGame, type Game } from '../../Utils/GDevelopServices/Game';
+import { type Game } from '../../Utils/GDevelopServices/Game';
 
 type Props = {|
   authenticatedUser: AuthenticatedUser,
-  gameId: string,
+  game: Game,
   open: boolean,
   onClose: () => void,
+  onGameUpdated: () => void,
 |};
 
-const BuildsDialog = ({ authenticatedUser, gameId, open, onClose }: Props) => {
+const BuildsDialog = ({
+  authenticatedUser,
+  game,
+  open,
+  onClose,
+  onGameUpdated,
+}: Props) => {
   const forceUpdate = useForceUpdate();
-  const { getAuthorizationHeader, profile } = React.useContext(
-    AuthenticatedUserContext
-  );
-  const [game, setGame] = React.useState<?Game>(null);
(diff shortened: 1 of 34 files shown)


newIDE/app/src/GameDashboard/GameRegistration.js
```diff
@@ -152,9 +152,11 @@ export const GameRegistration = ({
 
   React.useEffect(
     () => {
-      loadGame();
+      if (!game) {
+        loadGame();
+      }
     },
-    [loadGame]
+    [loadGame, game]
```
4ian, line 159: @ClementPasteau Interesting, why was this necessary?

newIDE/app/src/Export/Builds/BuildsDialog.js
```diff
@@ -1,56 +1,85 @@
...
-          authenticatedUser={authenticatedUser}
-          gameId={gameId}
-        />
-      </Dialog>
-    );
-  }
-}
+
+const BuildsDialog = ({ authenticatedUser, gameId, open, onClose }: Props) => {
+  // Force the Dialog repositioning
```
4ian, line 23: Move this comment at the place where forceUpdate is used :)

4ian (commented):
lgtm overall!

newIDE/app/src/GameDashboard/GameRegistration.js
ClementPasteau: I noticed that the game was fetched multiple times.
I assume that `loadGame` is a callback listening to `profile` and during this process, the limits are checked, so the profile is refetched, so the function change, and so the game is loaded again.
It's not mandatory but it saves unnecessary calls

4ian: Perfect then :)

newIDE/app/src/Export/Builds/index.js
```diff
@@ -3,12 +3,15 @@ import React, { Component } from 'react';
...
+import { type Game } from '../../Utils/GDevelopServices/Game';
 import BuildsWatcher from './BuildsWatcher';
 
 type Props = {|
   onBuildsUpdated?: () => void,
   authenticatedUser: AuthenticatedUser,
-  gameId: string,
+  game?: ?Game,
+  onGameUpdated?: Game => void,
+  isGameLoading: boolean,
```
4ian, line 14: Nitpicking: can't you consider game === null as loading?

ClementPasteau: No because you could open the builds dialog before you game is registered, and then game will be null.
Or we can decide to hide the button to see the builds, if the game is not registered & loaded

4ian (approved):

Issue #3359, opened 2021-12-14
Add suggested values list for mouse button pressed/released (text)


Elairyx7301:
## Description

A very small request but I still think it's important and definitely helpful to have a list of possible/accepted (left, right, middle) values to also avoid misspellings.

**Current:**
![image](https://user-images.githubusercontent.com/83909121/146030570-a3aa1129-4ff0-492e-bfae-491de442a319.png)

**Suggested:**
![image](https://user-images.githubusercontent.com/83909121/146030781-3630a8be-fe2f-4cf3-9057-354c3947b0dd.png)

## Solution suggested
Add suggested values list for mouse button pressed/released (text)

PR #3433, opened 2022-01-05
Improve Mouse extension to autocomplete possible values for expression

ClementPasteau:
Adresses #3359 

image

### DIFF
--- a/Core/GDCore/Extensions/Builtin/MouseExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/MouseExtension.cpp
@@ -227,7 +227,9 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsMouseExtension(
           "res/conditions/mouse24.png",
           "res/conditions/mouse.png")
       .AddCodeOnlyParameter("currentScene", "")
-      .AddParameter("string", _("Expression generating the button to check"))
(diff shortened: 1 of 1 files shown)


AlexandreSi:
Looks great! I'm just wondering if these are the only possibilities. I found this:
https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/buttons

But I'm not sure we support those. At least, I'm not sure if the list you added is a suggestion, or if it behaves like an enum and you can't choose another value than these 3

4ian:
We don't support those (we have a "enum like" content) *but* this could be cool to add the auxiliary ("wheel") button and the "browser back/forward" buttons.

[image](https://github.com/4ian/GDevelop/blob/ad3c7e4fad611e17fd827ebd349bd881fa4bca17/GDJS/Runtime/pixi-renderers/runtimegame-pixi-renderer.ts#L453-L475)

Approving this because this is still an improvement :)

Bouh:
Is these values are case sensitive? Please no 👀

ClementPasteau:
Yes they are. Not ideal, but the autocomplete should help avoid mistakes

4ian:
They are/always have been... but I must admit we could reduce the requirements here. There won't ever be a "Left" button and a "left" button being two different things.

PR #3441, opened 2022-01-08
Update step icon to be hollow

Midhil457:
Before:
icons

After:
icon2

I didn't change activate and deactivate icon (The tick and cross is white) as it might cause some contrast issues and not visible at all in some themes. How can the current theme colors retrieved? So that maybe the svg could change it's colors to maybe fit the theme?


4ian (approved):

4ian:
> How can the current theme colors retrieved? So that maybe the svg could change it's colors to maybe fit the theme?

SVG are considered as separate images and I think there is no easy way to "replace" a single color in them (though it would be awesome to allow to create SVG with colors adapted to the theme).
For now the best way is to play around with the CSS filter that are applied by the theme for icons.

4ian:
Don't forget to run `npm run format` ;)

newIDE/app/src/UI/CustomSvgIcons/Behaviors/Step.js
```diff
@@ -10,8 +10,11 @@ export default props => (
           fill="#2B3990"
           points="7.5 4.5 3 0 3 3 0 3 0 6 3 6 3 9"
         />
-        <circle id="Oval" fill="#27AAE1" cx="12" cy="4.5" r="3" />
-        <circle id="Oval" fill="#FFFFFF" cx="12" cy="4.5" r="1.5" />
+        <path
+          fill="#27aae1"
+          d="m 12,1.5 a 3,3 0 0 0 -3,3 3,3 0 0 0 3,3 3,3 0 0 0 3,-3 3,3 0 0 0 -3,-3 z M 12,3 A 1.5,1.5 0 0 1 13.5,4.5 1.5,1.5 0 0 1 12,6 1.5,1.5 0 0 1 10.5,4.5 1.5,1.5 0 0 1 12,3 Z"
+          id="Oval"
+        />
```
Bouh, line 17:
```suggestion
<circle cx="12" cy="4.5" r="2" fill="none" stroke="#27aae1" stroke-width="1.2" />
```

Bouh (approved):

newIDE/app/src/UI/CustomSvgIcons/Behaviors/Step.js
Midhil457: Done!, changed the `stroke-width` so that the circle is the same size

Issue #3354, opened 2021-12-13
UI improvement: Mark functions in extension with an icon


Wend1go:
## Description

When working on a larger external extension it gets hard to find the function you are looking for in the list.
There should be a visual way to distinguish between conditions, actions and expressions.

## Solution suggested

Mockup:
![image](https://user-images.githubusercontent.com/7413059/145811692-351a4b42-15e4-4a89-8a2e-32e0127f5b6e.png)

**Describe the solution**
Adding an icon in front of each function name makes it easy to see if we are dealing with a condition, action or expression.

## Alternatives considered

One could choose a name prefix for functions like "actionDoThis" / "aDoThis" or "condition_IsThis" / "cIsThis" but since the function names are also later visible in the event sheet when they get used this would make the events harder to read.

4ian:
Totally, that's an easy win to have an icon. Also a good opportunity to add a "lock" or something like this to show private functions/methods.

Bouh:
If you need some icons based on the function icon.

Function: ![function32](https://user-images.githubusercontent.com/1670670/145891241-0b262957-b474-4ffe-8033-9c78173b7665.png)


StringExpression: ![functionStringExpression](https://user-images.githubusercontent.com/1670670/145891212-ccfa0c13-0c54-40aa-adca-100ba37a0f0f.png)
Action: ![functionAction](https://user-images.githubusercontent.com/1670670/145891215-e5fa1bbe-0e92-4990-861c-c863cd02f726.png)
Condition: ![functionCondition](https://user-images.githubusercontent.com/1670670/145891216-c86e2d56-62e2-4b8d-9ba8-0cbfd13063c0.png)
Expression: ![functionExpression](https://user-images.githubusercontent.com/1670670/145891217-0ae0010c-6079-4f27-9a8f-5d88c4fffa53.png)

Bouh:
A new couple of icons in SVG.

Action, condition, expression.

![image](https://user-images.githubusercontent.com/1670670/146393688-6bfbcfd5-27a5-4514-942b-642d90e66ac0.png)


[icons_functions.zip](https://github.com/4ian/GDevelop/files/7728238/icons_functions.zip)

Midhil457:
I woulld like to take on this and did manage to make a bit of progress. Though I am having some doubts,
[...]

PR #3447, opened 2022-01-09
[WIP] Add badge option to sortable list

Midhil457:
I need some help with this
Currently the badge is shown with the given badge content. But, it should achieve a bit more things if it needs to be useable
Some of the code is still plceholder/not complete and I want to change it when the way to do is figured out

#### Customizing badge
I want the badge to be customizable. but, I don't know how to achieve this. I tried passing the styles as a prop, but it doesn't seem to work. I doesn't seem to give an error, just doesn't apply the styles. I don't know if a whole component can be passed as a prop and the thumbnail could be made to be it's children.

#### Don't render badge when not needed
I am currently using a if else statement to check if badge content is given. But, I don't think this is the best way to do it

### DIFF
--- a/newIDE/app/src/UI/SortableVirtualizedItemList/ItemRow.js
+++ b/newIDE/app/src/UI/SortableVirtualizedItemList/ItemRow.js
@@ -116,20 +116,19 @@ class ItemRow<Item> extends React.Component<Props<Item>> {
           const leftContent = () => {
             if (getBadge && getThumbnail) {
               return (
-                <Badge badgeContent={25} style={getBadge} >
-                <ListIcon
-                  iconSize={24}
-                  src={getThumbnail()}
-                  cssAnimation={
-                    scaleUpItemIconWhenSelected && selected
-                      ? 'scale-and-jiggle 0.8s forwards'
-                      : ''
-                  }
-                />
-              </Badge>
-              )
-            }
-            else if (!getBadge && getThumbnail) {
+                <Badge badgeContent={25} style={getBadge}>
+                  <ListIcon
+                    iconSize={24}
+                    src={getThumbnail()}
+                    cssAnimation={
+                      scaleUpItemIconWhenSelected && selected
+                        ? 'scale-and-jiggle 0.8s forwards'
+                        : ''
+                    }
+                  />
+                </Badge>
+              );
+            } else if (!getBadge && getThumbnail) {
               return (
                 <ListIcon
                   iconSize={24}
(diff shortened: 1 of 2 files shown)


AlexandreSi:
Hi @Midhil457!
Thanks for suggesting an improvement!

Can you give some context on this feature so that I can best help you with it?

Here are a couple questions that usually help:

- What is the goal you're trying to achieve?
- Where do you feel it will be used?

Midhil457:
I want to add an option to show a badge on the listItem’s icon,

I am trying to implement #3354. I want private function to be shown with a badge with a slashed eye on the extension’s icon. But, the sorted list component doesn’t give access to the icon which makes it hard to add a badge. So, I was trying to add an adoption to have a badge.

It can be used in to notify the user about something about something in the list ex: In the object list, badges could notify the number of effects on an object (if any) like done for layers currently

I am trying to make the badge customisable I suit different purposes (In the object list, number should be shown with background color. But, if a icon has to be shown like i am tying to do with the functions list, background color should be disabled)

:)

AlexandreSi:
Ok I see what you want to do.
To achieve #3354 you would need to do the following:

- Give the props `getItemThumbnail` to this component
 
 https://github.com/4ian/GDevelop/blob/801d58695c8aaa39d5d8f5cca73d91d2832c37fd/newIDE/app/src/EventsFunctionsList/index.js#L317-L337
- This props should be a function with the signature `Item => string` where `string` is the url/uri of the icon.

From what I understand, you want to add 2 other features:

1. the possibility to display a badge along with the icon to give more information to the user about the `Item`, like what is done  for the layers as you say:
  image

2. And you guess that it could be also useful to be able to give an icon instead of just a number, for example:
  image


Regarding the possibility to add a badge (1.), I think it's a good idea. I just think it should be used carefully, we don't want to overwhelm users with information that is not completely useful (a good way to know if it is useful could be to wait for someone to ask for the feature). To implement it, you would need to add a props `getItemBadgeContent` with this signature `Item => number`. The styling (e.g. background color) is handled by the Badge component.

Regarding the possibility to use an icon in the badge (2.), I am not sure that this is something Material Design recommends but the API allows it (as you can see above).
If we want to meet the root need that is "How to give more information about an ItemRow, on top of the already present icon?", you can maybe dig into Material guidelines for list items: https://material.io/components/lists#specs

Midhil457:
> Regarding the possibility to use an icon in the badge (2.), I am not sure that this is something Material Design recommends but the API allows it (as you can see above).

After some experimenting, I was able to successfully add an icon directly to the text content. It looks like this,
privateicon

It might still not be compliant with the material guidline. But, this is a easier way to do it than having a badge.

I am closing this PR as adding a badge is no longer needed. I will include the current implementation with the main PR to add icons to functions

:)

AlexandreSi:
Well I like this layout!
Maybe it's not compliant but it feels more natural.

Will review your PR then!

Issue #3282, opened 2021-11-23
Toggle for automatically starting timers


Elairyx7301:
## Short summary

Add a toggle to timers that prevents them from starting automatically at the beginning of the scene (if those timers are used in the events).

## Description

Even though that it can be really handy that timers are started automatically, many users seem to be confused about it and this behavior can be rather counterintuitive for many but especially new users.

Often, users want to only start a timer when a certain event happens/occurs. This would currently require to pause the timer at beginning of scene, resume for the event, let the timer run and have another condition make the logic work or reset the timer for as long as event conditions are not met.

## Solution suggested

After discussing it with community members, we've come to the conclusion that it would be best to add a toggle switch or boolean to timers allowing the user to avoid having timers started automatically by the engine if they are used in the code to reduce confusion and extra workarounds for timers that only want to be started with a certain condition, event or with a delay.

The decision was a toggle with [default: no] so already existing games and projects aren't affected by this huge change to the core of timers.

## Potential visual mockup (frontend)

![image](https://user-images.githubusercontent.com/83909121/143087692-8c4c2e28-3138-40ad-92f4-d08a9114f2a9.png)

## Alternatives considered

Give a hint to the user that timers are started automatically at beginning of scene.

### Original issue: #3065

Elairyx7301:
Are there any comments about this feature or idea? Is this change likely to happen or not?

D8H:
Hello Elairyx, just to let you know a PR is opened for this.

PR #3450, opened 2022-01-10
New timer conditions with an operator.

D8H:
Implementation for:
* https://github.com/4ian/GDevelop/discussions/3446
* https://github.com/4ian/GDevelop/issues/3282

It still allows every operator.

Example with scene and object timers:
* https://www.dropbox.com/s/zmd89fjpozt7cu2/Permissive.zip?dl=1

### DIFF
--- a/Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
@@ -846,7 +846,7 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
       .AddCodeOnlyParameter("conditionInverted", "")
       .MarkAsSimple();
 
-  // Deprecated and replaced by TimerComparison
+  // Deprecated and replaced by CompareObjectTimer
   obj.AddCondition(
          "ObjectTimer",
          _("Value of an object timer"),
@@ -860,7 +860,7 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
       .AddParameter("expression", _("Time in seconds"))
       .SetHidden();
 
-  obj.AddCondition("ObjectTimerComparison",
+  obj.AddCondition("CompareObjectTimer",
                   _("Value of an object timer"),
                   _("Compare the elapsed time of an object timer. This condition doesn't start the timer."),
                   _("The timer _PARAM1_ of _PARAM0_ _PARAM2_ _PARAM3_ seconds"),
--- a/Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
@@ -869,7 +869,7 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
                   "res/conditions/timer.png")
     .AddParameter("object", _("Object"))
     .AddParameter("string", _("Timer's name"))
-    .AddParameter("relationalOperator", _("Sign of the test"), "number")
+    .AddParameter("relationalOperator", _("Sign of the test"), "time")
     .AddParameter("expression", _("Time in seconds"))
     .SetManipulatedType("number");
 
--- a/Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
@@ -846,6 +846,7 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
       .AddCodeOnlyParameter("conditionInverted", "")
       .MarkAsSimple();
 
+  // Deprecated and replaced by TimerComparison
   obj.AddCondition(
          "ObjectTimer",
          _("Value of an object timer"),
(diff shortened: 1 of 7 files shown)


Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
```diff
@@ -856,7 +857,21 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
          "res/conditions/timer.png")
       .AddParameter("object", _("Object"))
       .AddParameter("string", _("Timer's name"))
-      .AddParameter("expression", _("Time in seconds"));
+      .AddParameter("expression", _("Time in seconds"))
+      .SetHidden();
+
+  obj.AddCondition("ObjectTimerComparison",
```
4ian, line 863: Let's use a verb for the name: `CompareObjectTimer`

Core/GDCore/Extensions/Builtin/TimeExtension.cpp
```diff
@@ -33,7 +33,22 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsTimeExtension(
                     "res/conditions/timer.png")
       .AddCodeOnlyParameter("currentScene", "")
       .AddParameter("expression", _("Time in seconds"))
-      .AddParameter("string", _("Timer's name"));
+      .AddParameter("string", _("Timer's name"))
+      .SetHidden();
+
+  extension
+      .AddCondition("TimerComparison",
```
4ian, line 40: And let's use `CompareTimer` here :)

GDJS/Runtime/events-tools/runtimescenetools.ts
```diff
@@ -106,11 +123,35 @@ namespace gdjs {
         return timeManager.getTimer(timerName).getTime() / 1000;
       };
 
-      export const getTimeFromStartInSeconds = function (runtimeScene) {
+      /**
+       * This is used by the condition to return false when a timer doesn't exist
+       * no mater the relational operator.
```
4ian, line 128:
```suggestion
       * no matter the relational operator.
```

GDJS/Runtime/events-tools/runtimescenetools.ts
```diff
@@ -106,11 +123,35 @@ namespace gdjs {
...
+       */
+      export const getTimerElapsedTimeInSecondsOrNaN = function (
+        runtimeScene: gdjs.RuntimeScene,
+        timerName: string
+      ) {
+        const timeManager = runtimeScene.getTimeManager();
+        if (!timeManager.hasTimer(timerName)) {
+          return Number.NaN;
+        }
+        console.log(timeManager.getTimer(timerName).getTime() / 1000);
```
4ian, line 141: Let's remove the console.log ;)

GDJS/Runtime/runtimeobject.ts
```diff
@@ -1780,6 +1787,22 @@ namespace gdjs {
       return this._timers.get(timerName).getTime() / 1000.0;
     }
 
+    /**
+     * Get a timer elapsed time.
+     *
+     * This is used by the condition to return false when a timer doesn't exist
+     * no mater the relational operator.
```
4ian, line 1794: ditto here

4ian (commented):
I know this is a draft but still added a review ;)

4ian:
Looks great, thank you! 
@charburgx Good to keep in mind for next tutorials => the usage of timer conditions now include the operator :) 

@D8H @ClementPasteau @AlexandreSi  I wonder if for conditions like this, we could in the editor select a default operator (greater than)?

ClementPasteau:
Yeah that could be a nice touch of UX, not having to select the operator, defaulted to the most used.
And we could do that for most conditions/actions too. It's a tiny time saver, but can add up to quite a lot :)

D8H:
Yes, the parameter could be marked as optional with a default value.

What do you think of some sort of radio buttons instead of a combobox?
Classic radio buttons take a lot more space and it's harder to separate visually. I wonder if the symbol could be the radio button itself. It would allow to have a row of small buttons (like a switch but with more than 2 states), but we loose the labels (though the symbols might be enough and they probably there only to fill the combobox).

D8H:
It's not very pretty, but it's not ugly either. It could make it faster to choose an operation while we think of a better look and feel.

operatorButtons

4ian:
As you say, it's not the prettiest thing nor super ugly. I wonder if we could group them?
https://v4.mui.com/components/button-group/

And as order: <, <=, >=, > ?

D8H:
The group button seems to be for having the same look on each button (it seems to override the syle of its children)

I removed the margin and used FlatButton for the unselected ones. I think it has a modern look. Maybe a background color could give more cohesiveness, but I kind of like it like this.

image

tristanbob:
I like the idea of offering these operators, but it doesn't feel intuitive yet.  It is not obvious that there are 3 parameters that can be changed.

How about a drop-down list that showed the operator and the text description?

```
(field) Timer's name 

(pulldown) 
< "Is lesser than"
<= "Is lesser than or equal" 
>= "Is greater than or equal"  <default selection>
> "Is greater than"

(field) Time in seconds 
```

tristanbob:
I guess I should have just said, I like the consistency of doing the same comparison we do for Variables ("=" and "!=" can be removed)

image

D8H:
> I guess I should have just said, I like the consistency of doing the same comparison we do for Variables ("=" and "!=" can be removed)

If we change the operator component, it will change everywhere.

Issue #3453, opened 2022-01-10
Editor crashing following an error due to removing a childNode not belonging to its parent


AlexandreSi:
## Describe the bug

Error message: 

```
Error: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
```

Many users reported this bug:

- #3444 
- #3443

  > estava tranquilo para empezar en linea y me aparecio si quierescompartirlo a amigos y eso y yo le puse si
luego de poner el correo y eso al volver me aparece que no puedo entrar por un error

- #3440 

  > the bug happens a few seconds after I open Gdevelop, and it happens
even when I log in before the bug happens, a message appears and
nothing else works in the app, thanks for the help

- #3423 
- #3422 
- #3419 
- #3418 
- #3416 
- #3415 
- #3409 
- #3381 
- #3370 

## To Reproduce

Couldn't reproduce the bug so far

## Other details

We can notice that the bug appears only **in the webapp** and on no version prior to v124.

Most of the users that reported the bug are working on Linux but I think that Linux users are more likely to have a GitHub account, that could explain the proportion.

## Complete stacktrace

```
Error: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
    at ca (https://editor.gdevelop-app.com/static/js/61.cb175572.chunk.js:2:253606)
    at Pa (https://editor.gdevelop-app.com/static/js/61.cb175572.chunk.js:2:257481)
    at Da (https://editor.gdevelop-app.com/static/js/61.cb175572.chunk.js:2:261044)
    at https://editor.gdevelop-app.com/static/js/61.cb175572.chunk.js:2:274313
    at Object.t.unstable_runWithPriority (https://editor.gdevelop-app.com/static/js/61.cb175572.chunk.js:2:285080)
    at Ou (https://editor.gdevelop-app.com/static/js/61.cb175572.chunk.js:2:274247)
    at Eu (https://editor.gdevelop-app.com/static/js/61.cb175572.chunk.js:2:274023)
    at wu (https://editor.gdevelop-app.com/static/js/61.cb175572.chunk.js:2:273368)
    at yu (https://editor.gdevelop-app.com/static/js/61.cb175572.chunk.js:2:272389)
    at qa (https://editor.gdevelop-app.com/static/js/61.cb175572.chunk.js:2:271258)
    at Object.enqueueSetState (https://editor.gdevelop-app.com/static/js/61.cb175572.chunk.js:2:220650)
[...]

PR #3456, opened 2022-01-10
Improve a few things

AlexandreSi:
- Close Extension details after it was successfully installed
- Add github action to automatically comment on new issues that match #3453

### DIFF
--- a/newIDE/app/src/AssetStore/ExtensionStore/ExtensionsSearchDialog.js
+++ b/newIDE/app/src/AssetStore/ExtensionStore/ExtensionsSearchDialog.js
@@ -101,18 +101,24 @@ export default function ExtensionsSearchDialog({
             isInstalling={isInstalling}
             onInstall={async extensionShortHeader => {
               setIsInstalling(true);
-              onInstallExtension(extensionShortHeader);
-              const wasExtensionInstalled = await installDisplayedExtension(
-                i18n,
-                project,
-                eventsFunctionsExtensionsState,
-                extensionShortHeader
-              );
-              if (wasExtensionInstalled && onExtensionInstalled)
-                onExtensionInstalled(extensionShortHeader);
+              try {
+                onInstallExtension(extensionShortHeader);
+                const wasExtensionInstalled = await installDisplayedExtension(
+                  i18n,
+                  project,
+                  eventsFunctionsExtensionsState,
+                  extensionShortHeader
+                );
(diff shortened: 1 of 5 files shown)


newIDE/app/src/AssetStore/ExtensionStore/ExtensionsSearchDialog.js
```diff
@@ -101,18 +101,25 @@ export default function ExtensionsSearchDialog({
             isInstalling={isInstalling}
             onInstall={async extensionShortHeader => {
```
ClementPasteau, line 102: should we move this function out of the render to have more clarity on what it does?

newIDE/app/src/AssetStore/ExtensionStore/ExtensionsSearchDialog.js
```diff
@@ -101,18 +101,25 @@ export default function ExtensionsSearchDialog({
...
+                  if (onExtensionInstalled)
+                    onExtensionInstalled(extensionShortHeader);
+                  return true;
+                }
 
-              setExtensionWasInstalled(wasExtensionInstalled);
-              setIsInstalling(false);
+                return false;
+              } finally {
+                setIsInstalling(false);
```
ClementPasteau, line 121: this is a bit confusing to me, you're returning a value just before, and have a `finally` too. Does it work?
Should we be clearer and call the function `setIsInstalling` on every branch:
- before returning true
- before returning false
- in a catch for the try

4ian (approved):

newIDE/app/src/AssetStore/ExtensionStore/ExtensionsSearchDialog.js
AlexandreSi: The code in `finally` is always called, if there is an error or not. Doing that, I make sure that isInstalling comes back to false, that's to say, buttons are re-enabled.

I prefer this way of doing as we do not duplicate lines unnecessarily and it's quite clear to determine which part of the code is called regardless of what happens in that `try`

newIDE/app/src/AssetStore/ExtensionStore/ExtensionsSearchDialog.js
AlexandreSi: Good idea, I'll move it in my next PR (the yml is not correctly parsed)

PR #3471, opened 2022-01-11
Update Build Progress to display timed out builds

### LINKED ISSUE #3448: Timed out exports still showing as loading/building
A Discord user reports that his builds from yesterday still appear as loading:
image
I assume they timed out, and GDevelop just doesn't realize it.
Could we do something about that? Giving a real status report of whether the build is uploading, building or failed? Or perhaps a Cancel button?
User was concerned that the concurrent uploads were eating his bandwidth. Perhaps we could put a queue mechanism to avoid multiple concurrent uploads, if there is not one already? I guess multiple uploads make it more likely for all builds to fail, for slow connections.

ClementPasteau:
Fix #3448 

It doesn't solve the fact that some builds stay in "pending" in the GDevelop services.
But at least it doesn't fool the user showing it will finish and it will not. 

Check the storybook for the components design.

### DIFF
--- a/newIDE/app/src/Export/Builds/BuildProgressAndActions.js
+++ b/newIDE/app/src/Export/Builds/BuildProgressAndActions.js
@@ -91,13 +91,16 @@ export default ({
     AuthenticatedUserContext
   );
   const config = buildTypesConfig[build.type];
+  const estimatedTime = config.estimatedTimeInSeconds(build);
   const secondsSinceLastUpdate = Math.abs(
     differenceInSeconds(build.updatedAt, Date.now())
   );
-  const estimatedRemainingTime = Math.max(
-    config ? config.estimatedTimeInSeconds(build) - secondsSinceLastUpdate : 0,
-    0
-  );
+  const estimatedRemainingTime = estimatedTime - secondsSinceLastUpdate;
+  const isStillRunning = estimatedRemainingTime > 0;
+  const hasJustOverrun =
+    estimatedRemainingTime < 0 && estimatedRemainingTime < -estimatedTime;
+  const hasTimedOut =
+    estimatedRemainingTime < 0 && estimatedRemainingTime > 2 * -estimatedTime;
 
   const onDownload = (key: BuildArtifactKeyName) => {
     const url = getBuildArtifactUrl(build, key);
@@ -136,7 +139,7 @@ export default ({
       {({ i18n }) =>
         build.status === 'error' ? (
           <React.Fragment>
-            <Line alignItems="center">
+            <Line alignItems="center" justifyContent="center">
               <Text>
                 <Trans>Something wrong happened :(</Trans>
               </Text>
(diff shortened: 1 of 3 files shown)


newIDE/app/src/stories/index.js
```diff
@@ -2575,38 +2568,6 @@ storiesOf('BuildStepsProgress', module)
     />
   ));
 
-storiesOf('BuildProgressAndActions', module)
```
AlexandreSi, line 2578: So nice to see this file getting skinnier

newIDE/app/src/Export/Builds/BuildProgressAndActions.js
```diff
@@ -156,32 +154,53 @@ export default ({
...
+                      : 0
+                  }
+                  variant={
+                    estimatedRemainingTime > 0 ? 'determinate' : 'indeterminate'
+                  }
+                />
+                <Spacer />
+              </>
+            )}
+            {estimatedRemainingTime > 0 && (
```
AlexandreSi, line 176: I'm having troubles figuring out if one of those 4 conditioned rendering could be rendered at the same time.
Maybe can you name the booleans `const isLongerThanEstimatedButNotTooMuch = estimatedRemainingTime < 0 &&
              estimatedRemainingTime > -estimatedTime` or something simpler 😅
Or have a `if` with `else`s to clearly what rendering can happen together or not.

ClementPasteau: yep, I like that 👍  will do

newIDE/app/src/stories/componentStories/BuildProgressAndActions.stories.js
```diff
@@ -0,0 +1,77 @@
...
+  title: 'Export/BuildProgressAndActions',
+  component: BuildProgressAndActions,
+  decorators: [muiDecorator, paperDecorator],
+};
+
+export const Errored = () => (
+  <BuildProgressAndActions build={erroredCordovaBuild} game={game1} />
+);
+
+export const PendingElectronBuild = () => (
```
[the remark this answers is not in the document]
ClementPasteau: I like the idea, but I've tried quickly and couldn't get something nice working 🤔 
I also need to change the prop publicWebBuildId, but only if it's a type web (didn't search long how to do that)

For the amount we use storybook, I'm leaving it like that :)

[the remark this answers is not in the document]
4ian: Note: in an effort to help future developers or future us to discover all the use cases of a component, it might be worth having a bunch of different states explicitly in the stories. The control can be still useful to avoid an exponential number of cases, but I think the most important cases should be separate stories so that I can quickly browse them and look and say "ok nothing crashed and everything is looking good, so it should be ok".

In a more distant future where we do screenshots of important components, it will also be helpful to have these important use cases as separate stories (rather than getting the tool, like Chromatic, doing screenshots to change the controls).

Overall the idea is to find a balance between "not too much stories" and "not too much controls" :)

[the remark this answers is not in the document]
AlexandreSi: There are 2 conflicting use-cases:

- As a dev, did I break something and you want to see all possible states rapidly only browsing the side menu
- As a designer, I want to make sure that all possibilities are covered so controls are more exhaustive

If we have a product designer, we'll see if they want to use the storybook as such

newIDE/app/src/Export/Builds/BuildProgressAndActions.js
```diff
@@ -91,13 +91,16 @@ export default ({
...
   );
-  const estimatedRemainingTime = Math.max(
-    config ? config.estimatedTimeInSeconds(build) - secondsSinceLastUpdate : 0,
-    0
-  );
+  const estimatedRemainingTime = estimatedTime - secondsSinceLastUpdate;
+  const isStillRunning = estimatedRemainingTime > 0;
+  const hasJustOverrun =
+    estimatedRemainingTime < 0 && estimatedRemainingTime < -estimatedTime;
+  const hasTimedOut =
```
AlexandreSi, line 102: Crystal clear now 💎

AlexandreSi (approved):

PR #3474, opened 2022-01-11
Remove an only in tests

D8H:
I'm very sorry for this, I hope it didn't do too much harm.
It was added by
* https://github.com/4ian/GDevelop/pull/3421


AlexandreSi:
It might be worth installing `eslint-plugin-ban`: https://timdeschryver.dev/blog/dont-commit-focused-tests#eslint-plugin-ban

4ian:
Merging so that tests are back on track as soon as possible :) Would totally be a good idea to have a lint rule for this.

PR #3475, opened 2022-01-11
Open extension details

AlexandreSi:
In the project manager, add possibility to view details of an extension installed from the store.

demo_extension_details_from_project_manager


newIDE/app/src/AssetStore/ExtensionStore/ExtensionStoreContext.js
```diff
@@ -71,7 +73,8 @@ export const ExtensionStoreStateProvider = ({
     () => {
       // Don't attempt to load again resources and filters if they
       // were loaded already.
-      if (extensionShortHeadersByName || isLoading.current) return;
+      if (Object.keys(extensionShortHeadersByName).length || isLoading.current)
```
ClementPasteau, line 76: 👍

newIDE/app/src/ProjectManager/index.js
```diff
@@ -789,472 +841,622 @@ export default class ProjectManager extends React.Component<Props, State> {
...
+                            primaryText={<Trans>Global variables</Trans>}
+                            leftIcon={<VariableTree />}
+                            onClick={this._openProjectVariables}
+                          />,
+                          <ListItem
+                            key="icons"
+                            primaryText={<Trans>Icons</Trans>}
+                            leftIcon={<PhotoLibrary />}
+                            onClick={() =>
+                              this.props.onOpenPlatformSpecificAssets()
```
ClementPasteau, line 909: I think all those onClick can be simplified to not create a function every time: `onClick={this.props.onOpenPlatformSpecificAssets}`

newIDE/app/src/ProjectManager/index.js
```diff
@@ -789,472 +841,622 @@ export default class ProjectManager extends React.Component<Props, State> {
       onReloadEventsFunctionsExtensions,
       onInstallExtension,
     } = this.props;
-    const { renamedItemKind, renamedItemName, searchText } = this.state;
+    const {
+      renamedItemKind,
+      renamedItemName,
+      searchText,
+      openedExtensionShortHeader,
+      openedExtensionName,
```
ClementPasteau, line 849: it's a bit confusing to me as to why we need both the `openedExtensionName` and the `openedExtensionName` in the state.
Aren't they redundant?

newIDE/app/src/ProjectManager/index.js
AlexandreSi: True, I'll change it

newIDE/app/src/ProjectManager/index.js
AlexandreSi: `openedExtensionShortHeader` is the object returned by the API that can be considered as the truth whereas `openedExtensionName` is the name of the extension in the project, name that can be modified by the user.

ClementPasteau: oh right, ok got it.

newIDE/app/src/ProjectManager/ProjectManagerItems.js
```diff
@@ -0,0 +1,304 @@
...
+                borderBottom: `1px solid ${muiTheme.listItem.separatorColor}`,
+                ...style,
+              }}
+              primaryText={label}
+              leftIcon={leftIcon}
+              displayMenuButton
+              buildMenuTemplate={(i18n: I18nType) => [
+                {
+                  label: i18n._(t`Edit`),
+                  click: () => onEdit(),
```
ClementPasteau, line 166: this can be just `onEdit` (and all the others below)

newIDE/app/src/ProjectManager/ProjectManagerItems.js
```diff
@@ -0,0 +1,304 @@
...
+          />
+        ) : null
+      }
+      primaryText={name}
+      editingName={isEditingName}
+      onEdit={() => onEdit(extensionShortHeadersByName)}
+      onDelete={onDelete}
+      addLabel={t`Add a New Extension`}
+      onAdd={onAdd}
+      onRename={newName => {
```
ClementPasteau, line 289: this can just be `onRename={onRename}`?

ClementPasteau (approved):
lgtm!

newIDE/app/src/ProjectManager/index.js
```diff
@@ -714,6 +533,24 @@ export default class ProjectManager extends React.Component<Props, State> {
...
+    eventsFunctionsExtension: gdEventsFunctionsExtension,
+    name: string
+  ) => {
+    const originName = eventsFunctionsExtension.getOriginName();
+    if (originName !== 'gdevelop-extension-store') {
+      this.props.onOpenEventsFunctionsExtension(name);
+      return;
+    }
+    const originIdentifier = eventsFunctionsExtension.getOriginIdentifier();
+    const extensionShortHeader = extensionShortHeadersByName[originIdentifier];
```
4ian, line 547: Is this possible to be undefined? Will this create an issue?

newIDE/app/src/ProjectManager/ProjectManagerItems.js
AlexandreSi: Thanks for spotting it, I had many versions and my eyes were tired.

newIDE/app/src/ProjectManager/ProjectManagerItems.js
```diff
@@ -0,0 +1,304 @@
...
+      return (
+        <ListItem
+          style={{
+            backgroundColor: muiTheme.listItem.groupBackgroundColor,
+            borderBottom: `1px solid ${muiTheme.listItem.separatorColor}`,
+          }}
+          nestedListStyle={
+            indentNestedItems ? undefined : styles.noIndentNestedList
+          }
+          {...otherProps}
```
ClementPasteau, line 61: i'm usually not a big fan of using otherProps like that, it's convenient to reduce the size of a component, but when you're trying to understand how a prop is used (for instance `primaryText`), you can't find it easily with ctrl+f

newIDE/app/src/ProjectManager/ProjectManagerItems.js
AlexandreSi: True. I copy-pasted the existing, I don't know why it was like that at first.

newIDE/app/src/ProjectManager/index.js
4ian: Worth adding a comment here and above/below detailing the logic :
"If the extension is not coming from the store, or is but can't be found, edit it directly:"
"Otherwise open its details first"

newIDE/app/src/ProjectManager/index.js
```diff
@@ -714,6 +533,24 @@ export default class ProjectManager extends React.Component<Props, State> {
     this._onProjectItemModified();
   };
 
+  _onEditEventsFunctionExtension = (
```
4ian, line 536: nitpicking: you could even take the side of using a long name like  `_onEditEventsFunctionExtensionOrSeeDetails` to show the "or" part that means that something will differ according to extensions.

4ian (approved):
LGTM!

newIDE/app/src/ProjectManager/ProjectManagerItems.js
4ian: (Agree - but note that the existing code was already doing that - my fault) (so I give it a pass unless you have time and want to refactor this)

newIDE/app/src/ProjectManager/index.js
AlexandreSi: Good idea

PR #3477, opened 2022-01-11
Update Build storybook check with storybook urls

AlexandreSi:


.github/workflows/build-storybook.yml
```diff
@@ -55,10 +55,32 @@ jobs:
         run: aws s3 sync ./build-storybook/ s3://gdevelop-storybook/$(git rev-parse --abbrev-ref HEAD)/latest/ --delete
         working-directory: newIDE/app
 
+      - name: Compute urls
+        run: |
+          export STORYBOOK_S3_URL_COMMIT=http://gdevelop-storybook.s3-website-us-east-1.amazonaws.com/$(git rev-parse --abbrev-ref HEAD)/commit/$(git rev-parse HEAD)/index.html
+          export STORYBOOK_S3_URL_LATEST=http://gdevelop-storybook.s3-website-us-east-1.amazonaws.com/$(git rev-parse --abbrev-ref HEAD)/latest/index.html
+          echo "::set-env name=STORYBOOK_S3_URL_COMMIT::$STORYBOOK_S3_URL_COMMIT"
```
4ian, line 62: Seems you need to use https://docs.github.com/en/actions/learn-github-actions/workflow-commands-for-github-actions#setting-an-environment-variable

4ian (approved):

AlexandreSi:
Closing it as I'm not succeeding in having this done. Here is a sum up:

- Original goal: display the urls to see the storybook updated by this branch for each pushed commit
- Possible solutions:
  - Change the PR description to add the URLs: Seemed too hard to get the description, append (or change) the content and post it back. I felt like we could erase content in the description.
  - Create a comment: we could end up with a lot of comments on the PR and spam users
  - Update the Check Run object: We cannot easily get the id of the Check Run (that is not the same as the Run or the Status) that is needed to use the action I set. See https://github.com/actions/starter-workflows/issues/292.

PR #3484, opened 2022-01-12
Start behavior descriptions with a verb.

D8H:
It tries to give uniformity to behavior descriptions by starting with a verb and using plural indefinite articles for "objects".


Silver-Streak:
I see a lot of instances where plurals are being used. Normally you'd want to be singular since the behavior is going to only be applied to one object.

e.g. 
"Anchor objects to the window's bounds."
is probably better as
"Anchor an object to the window's bounds."

"Flag objects as being obstacles to light. The light emitted by light objects will be stopped by the object."
is probably better as
"Flag an object as an obstacle on the lighting layer. Any light emitted by light objects will be stopped by the obstacle."

"Flag objects as being obstacles for pathfinding."
is probably better as
"Flag an object as an pathfinding obstacle. Any currently moving pathfinding objects will navigate around the obstacle."

etc.

These are minor tweaks, and even if these aren't adjusted the existing changes you've made still look good to me.

D8H:
> I see a lot of instances where plurals are being used. Normally you'd want to be singular since the behavior is going to only be applied to one object.

I used the plural to express a generality like "Objects/Instances that have this behavior...", but I'm not sure if this is a good idea (I don't know English enough to tell).

Silver-Streak:
That could still work, too. Where is this going to show up? The list of behaviors or after it's added to the object?

4ian:
Merged because this is a good improvement in the way behaviors are presented :)

PR #3495, opened 2022-01-15
[Platformer] No longer reset "can jump" when jump speed reaches 0

D8H:
`canJump` was reset at the `Jumping` to `Falling` transition.
This makes air jump behavior counts a jump that doesn't happen.

### DIFF
--- a/Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
+++ b/Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
@@ -1799,9 +1799,13 @@ namespace gdjs {
     }
 
     enter(from: State) {
-      // canJump is used to make air jumps.
-      // It must not be reset during a jump,
-      // where "jump" means the jump + falling states.
+      // Only forbid jumping when starting to fall from a platform,
+      // not when falling during a jump. This is because the Jumping
+      // state has already set `_canJump` to false and we don't want to reset
+      // it again because it could have been set back to `true` to allow
+      // for an "air jump".
+      // Transition from Falling to Falling state should not happen, 
+      // but don't change anything if this ever happen.
       if (from !== this._behavior._jumping && from !== this) {
         this._behavior._canJump = false;
       }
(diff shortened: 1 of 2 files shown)


Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -1797,8 +1798,13 @@ namespace gdjs {
       this._behavior = behavior;
     }
 
-    enter() {
-      this._behavior._canJump = false;
+    enter(from: State) {
+      // canJump is used to make air jumps.
+      // It must not be reset during a jump,
+      // where "jump" means the jump + falling states.
+      if (from !== this._behavior._jumping && from !== this) {
```
4ian, line 1805: Is this possible for the Falling state (or any other state) to enter from itself?

4ian (approved):

4ian:
Thank you! It seems indeed to be a bug to disallow jumping again after the jump reached its peak - this should only be reset when the jump first start or when starting to fall from a platform (or another state, but not sure if it's possible, I don't have the "state machine transitions" in mind).

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
D8H: It's not possible, but I thought it could avoid regression if that change for some reason.
I'm fine to either remove it or add a comment.

D8H:
> Thank you! It seems indeed to be a bug to disallow jumping again after the jump reached its peak - this should only be reset when the jump first start or when starting to fall from a platform (or another state, but not sure if it's possible, I don't have the "state machine transitions" in mind).

The bug should only happens when the jump ends, when currentJumpSpeed reaches 0.

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
4ian: I've updated the comment to call this out 👍

PR #3501, opened 2022-01-16
Improve functions list

Midhil457:
Implements suggestions made in #3354 and some other improvements too!
- [x] Show icon when function is private
- [x] Show icons for functions
- [x] Disable `Make private` option for lifecycle functions
- [x] Update icons in function chooser
- [x] Update icons in behavior function chooser
- [x] Remove icon in properties editor

iconfunctions

One annoying thing about the way I added the private icon in the list is that it makes the tooltip `[object Object]`. I couldn't figure out how to fix it
tooptip

### Other
The `GDJS/package-lock.json` and `newIDE/app/package-lock.json` changed automatically withotu me doing anything. And it seems `newIDE/app/package-lock.json` is having soem conflits. Should I revert the changes done to it?


Bouh:
As for the Layers list, the interactive icons are on the right, can we do the same thing for the eye icon? 
Nit: Might if a lock icon is more speaking to make a function private or not as you can access to it.
But an eye can be interpreted as visible/usable or not in the condition/action list so not a big deal.

image

4ian:
> As for the Layers list, the interactive icons are on the right, can we do the same thing for the eye icon you just added?

I think this might be complicated, but also not sure if we want to make this option take a place so important in the interface (i.e: have an "eye" or "lock" icon for every single function, which means it's as important as the name, icon or the right click menu). I would rather see it as a "lock" or "eye" icon only on those who are actually private - and let the user right click or edit the function to change it, because this is not an usual operation?

> One annoying thing about the way I added the private icon in the list is that it makes the tooltip [object Object]. I couldn't figure out how to fix it

This is probably because your "hacked" the label to display an icon, so you passed a React node instead of a text which was expected.
For now the virtual list has no way to display more than one icon - we'll have to adapt it once we figure out what we want to do exactly (I was thinking of a "lock" badge on the icon, but could be your solution or Bouh solution).

AlexandreSi:
Thank you for this clear recap of your PR!

On my side, I feel it doesn't make sense to add icons on the right-side of the row, as we have the icon for the context menu. So this left-side icon seems handy to me (and displaying it only when private seems the logical way to me).

Regarding the icon, it seems to me that the lock icon could be ambiguous to me: "is it locked for modification?". In that way, I like the VisibilityOff icon better. As another suggestion, what do you all think of the key icon?
image
Not sure if it's clearer.

Regarding the `[object Object]`, we will need to adapt how the component `ItemRow` defines the div `title` attribute.

Concerning the lock files, it is a bit weird, maybe it's due to us not committing it, maybe it's an issue with your node version.
Either way, when the PR is ready, I can checkout on it and fix it.

Finally, regarding the icons you added in your PR, can you explain why there are SVG icons as well as new png icons of different sizes?

arthuro555:
IIRC private functions are just hidden but not completely unusable (if you modified the JSON yourself you would still be able to use it), and I have used it in the past to deprecate an extension function, so I think the eye might be a better fit since that describes best what privatizing functions does.

GDJS/package-lock.json
```diff
@@ -1,3806 +1,8 @@
 {
   "name": "gdevelop-javascript-platform",
   "version": "0.0.0",
-  "lockfileVersion": 2,
```
arthuro555: I don't think this downgrade of the lockfile, please remove it from the PR (`git reset master ./GDJS/package-lock.json && git restore ./GDJS/package-lock.json`)

newIDE/app/src/EventsFunctionsList/index.js
```diff
@@ -37,7 +37,17 @@ export type EventsFunctionCreationParameters = {|
...
 const getEventsFunctionName = (eventsFunction: gdEventsFunction) =>
-  eventsFunction.getName() + (eventsFunction.isPrivate() ? ' (private)' : '');
+  eventsFunction.isPrivate() ? (
+    <React.Fragment>
+      <VisibilityOffIcon
+        fontSize="small"
+        style={{ marginRight: 5, verticalAlign: 'bottom' }}
+      />
+      {eventsFunction.getName()}
+    </React.Fragment>
```
arthuro555, line 47: Nitpick: you do not need to write out `React.Fragment`, instead you can just use empty brackets:
```suggestion
    <>
      <VisibilityOffIcon
        fontSize="small"
        style={{ marginRight: 5, verticalAlign: 'bottom' }}
      />
      {eventsFunction.getName()}
    </>
```

newIDE/app/src/EventsFunctionsList/index.js
```diff
@@ -37,7 +37,17 @@ export type EventsFunctionCreationParameters = {|
...
+  ) : (
+    <>{eventsFunction.getName()}</>
+  );
```
arthuro555, line 50: No need for a fragment here? A text can always be used as a react child directly.
```suggestion
  ) : eventsFunction.getName();
```

newIDE/app/src/EventsFunctionsList/index.js
```diff
@@ -123,6 +133,54 @@ export default class EventsFunctionsList extends React.Component<Props, State> {
...
+          return 'res/functions/destroy32.png';
+        } else {
+          return 'res/functions/action32.png';
+        }
+      case 1:
+        return 'res/functions/condition32.png';
+      case 2:
+        return 'res/functions/expression32.png';
+      case 3:
+        return 'res/functions/expression32.png';
```
arthuro555, line 181: Don't use the numbers directly, that makes the code unreadable and hard to understand 😱 
```suggestion
      case gd.EventsFunction.Action:
        if (
          eventsFunction.getName() === 'onSceneLoaded' ||
          eventsFunction.getName() === 'onFirstSceneLoaded' ||
          eventsFunction.getName() === 'onCreated'
        ) {
          return 'res/functions/create32.png';
        }
        if (
          eventsFunction.getName() === 'onScenePreEvents' ||
          eventsFunction.getName() === 'onScenePostEvents' ||
          eventsFunction.getName() === 'doStepPreEvents' ||
          eventsFunction.getName() === 'doStepPostEvents'
        ) {
          return 'res/functions/step32.png';
        }
        if (
          eventsFunction.getName() === 'onScenePaused' ||
          eventsFunction.getName() === 'onDeActivate'
        ) {
          return 'res/functions/deactivate32.png';
        }
        if (
          eventsFunction.getName() === 'onSceneResumed' ||
          eventsFunction.getName() === 'onActivate'
        ) {
          return 'res/functions/activate32.png';
        }
        if (
          eventsFunction.getName() === 'onSceneUnloading' ||
          eventsFunction.getName() === 'onDestroy'
        ) {
          return 'res/functions/destroy32.png';
        } else {
          return 'res/functions/action32.png';
        }
      case gd.EventsFunction.Condition:
        return 'res/functions/condition32.png';
      case gd.EventsFunction.Expression:
        return 'res/functions/expression32.png';
      case gd.EventsFunction.StringExpression:
        return 'res/functions/expression32.png';
```

arthuro555 (approved):
Looks overall good to me, apart from a few nitpicks! Personally, I am fine with this design as it is as well, but do discuss it if you think there is a need to.

GDJS/package-lock.json
arthuro555: (Ditto for newIDE/app)

Midhil457:
> Regarding the [object Object], we will need to adapt how the component ItemRow defines the div title attribute
 Can the title attribute be removed? And a custom tooltip could be added (see below)

We could add a tooltip to the icon showing something like `This function won't be shown in in the events editor`. This could help show what privatizing means. And also help with the icon confustion

> Finally, regarding the icons you added in your PR, can you explain why there are SVG icons as well as new png icons of different sizes?

These are new icons made by @Bouh and these could come in handy later.
Also because the list component took the path to the file while it was easier to have a svg for the function chooser

> As for the Layers list, the interactive icons are on the right, can we do the same thing for the eye icon you just added?

I choose to have it on the left also because the icon won't be visible if the function name is long. I didn't want to go through the hacky process to make it always visible.


@arthuro555 Thanks for teh rview, I'll make the changes soon :)

newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionPropertiesEditor.js
```diff
@@ -172,7 +164,6 @@ export default class EventsFunctionPropertiesEditor extends React.Component<
             {renderConfigurationHeader ? renderConfigurationHeader() : null}
             <ResponsiveLineStackLayout alignItems="center" noMargin>
               <Line alignItems="center" noMargin>
-                <img src="res/function32.png" alt="" style={styles.icon} />
```
AlexandreSi, line 175: Good idea to remove this icon as it is not really useful now that the more specialized icon is present in the EventsFunctionsList. 

Nonetheless, I think that a user that select a different type here should see the icon in the EventsFunctionsList updated with the corresponding one.

I think you can use the props `onConfigurationUpdated` to force update the `EventsFunctionsList` used by the `EventsFunctionsExtensionEditor` (you would need to add a ref to the component)

newIDE/app/src/EventsFunctionsList/index.js
AlexandreSi: This will resolve the title issue when the function is public.

newIDE/app/src/EventsFunctionsList/index.js
AlexandreSi: To resolve the title issue in this case, you can change the title props to the div in `ItemRow` to be:

```
title={typeof itemName === 'string' ? itemName : undefined}
```

And you can set your own title here, putting the `eventsFunction.getName()` in a span.

newIDE/app/src/EventsFunctionsList/index.js
```diff
@@ -123,6 +133,54 @@ export default class EventsFunctionsList extends React.Component<Props, State> {
     );
   };
 
+  _getFunctionThumbnail = (eventsFunction: ?gdEventsFunction) => {
```
AlexandreSi, line 136: You can move this function out of the class to lighten re-renders.

newIDE/app/src/EventsFunctionsList/index.js
```diff
@@ -123,6 +133,54 @@ export default class EventsFunctionsList extends React.Component<Props, State> {
...
+    switch (eventsFunction.getFunctionType()) {
+      default:
+        return 'res/functions/function32.png';
+      case 0:
+        if (
```
AlexandreSi, line 141: I wonder if it would be a godd idea to have a nested switch:
```js
switch (eventsFunction.getName()) {
  case 'onSceneLoaded':
  case 'onFirstSceneLoaded':
  case 'onCreated':
    return 'res/functions/create32.png';
  case ...
}
```
You could move up the 3 other cases that are shorter, and have this longer case as the last one for clarity's sake.

AlexandreSi:
A few changes and it should be good to go!

newIDE/app/src/EventsFunctionsList/index.js
arthuro555: This is neither changing state nor calling props callbacks that may in turn trigger a rerender higher up, so this shouldn't trigger any rerender already.

newIDE/app/src/EventsFunctionsList/index.js
arthuro555: Disregard this, Alex's proposition is better

newIDE/app/src/EventsFunctionsList/index.js
AlexandreSi: You would need to change the type of the props itemName to be `string | React.Element` or something like this to be correct

Bouh:
Some SVG are already used in the engine.
There is no need for the png files if the SVG can be used instead.
And the SVG give more flexibility. if you want an SVG of 32px just edit the SVG with width and height attributes.

4ian:
Agree with @Bouh, I think we can officially start using SVG almost everywhere in the app for icons, if it works :)

AlexandreSi:
Well `ItemRow` is used in different places with a `getThumbnail` function that returns a path to a png image.
So this refactor could be done in a different PR, It might be a bit complicated.

AlexandreSi:
@Midhil457 Let me know if I can help

Midhil457:
- [x] Use svg instead of png for function icons
- [x] Remove pngs of icons
- [x] Use nested case instead of if statement 
- [x]  Fix tooltip
- [x] Add tooltip to private icon
- [x] Automatically upudate icons when the function type is changed

> Nonetheless, I think that a user that select a different type here should see the icon in the EventsFunctionsList updated with the corresponding one.
I think you can use the props onConfigurationUpdated to force update the EventsFunctionsList used by the EventsFunctionsExtensionEditor (you would need to add a ref to the component)

I wasn't able to figure how to do this, I am still learning :sweat_smile: I would be great if someone could help

:)

newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionPropertiesEditor.js
4ian: I took care of this 👍

4ian:
I've fixed the renaming of the private functions ("[Object object]") and also fixed the icons updated (see my commits). We should be good! :)

newIDE/app/src/EventsFunctionsList/index.js
```diff
@@ -36,8 +38,23 @@ export type EventsFunctionCreationParameters = {|
   name: ?string,
 |};
 
+const renderEventsFunctionLabel = (eventsFunction: gdEventsFunction) =>
+  eventsFunction.isPrivate() ? (
+    <>
+      <Tooltip title="This function won't be visible in the events editor">
```
Midhil457, line 44: One thing I forgot to mention in the checklist was that I added a tooltip to the icon. The goal is to give more context about the icon. Though I am not really sure about the wording.  is `Events editor` a thing?

4ian: Events editor or events sheet are synonym and right :)

4ian:
Thanks @Midhil457 for you work on this 👍

PR #3512, opened 2022-01-18
Fix focus loss project name

AlexandreSi:
Resolves https://forum.gdevelop-app.com/t/blank-project-name-issue/37558

### DIFF
--- a/newIDE/app/src/AssetStore/ExampleStore/index.js
+++ b/newIDE/app/src/AssetStore/ExampleStore/index.js
@@ -65,7 +65,7 @@ export const ExampleStore = ({ isOpening, onOpen, focusOnMount }: Props) => {
     () => ({
       add: filtersState.addFilter,
       remove: filtersState.removeFilter,
-      chosenTags: Array.from(filtersState.chosenFilters),
+      chosenTags: filtersState.chosenFilters,
     }),
     [filtersState]
   );
--- a/newIDE/app/src/AssetStore/ExampleStore/index.js
+++ b/newIDE/app/src/AssetStore/ExampleStore/index.js
@@ -58,6 +58,15 @@ export const ExampleStore = ({ isOpening, onOpen, focusOnMount }: Props) => {
       searchBarRef.current.focus();
   });
 
+  const tagsHandler = React.useMemo(
+    () => ({
+      add: filtersState.addFilter,
+      remove: filtersState.removeFilter,
+      chosenTags: Array.from(filtersState.chosenFilters),
+    }),
+    [filtersState]
+  );
+
   return (
     <React.Fragment>
       <ResponsiveWindowMeasurer>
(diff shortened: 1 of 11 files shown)


newIDE/app/src/UI/SearchBar.js
```diff
@@ -205,13 +206,20 @@ const SearchBar = React.forwardRef<Props, SearchBarInterface>(
         if (
           shouldAutofocusSearchbar &&
           tagsHandler &&
-          tagsHandler.chosenTags.length === 0
+          tagsHandler.chosenTags.length === 0 &&
+          previousChosenTagsCount.current > 0
         )
           focus();
       },
       [tagsHandler, shouldAutofocusSearchbar]
```
4ian, line 214: This may indicate also an issue where the tagsHandler was not stable between re-renders?

4ian (approved):

newIDE/app/src/UI/SearchBar.js
AlexandreSi: Yes you're right. setting the `tagsHandler` to:
```
const tagsHandler = React.useMemo(() => ({
    add: filtersState.addFilter,
    remove: filtersState.removeFilter,
    chosenTags: Array.from(filtersState.chosenFilters),
  }), [filtersState.chosenFilters])
```
fixes the issue. I think I'll add both as it is safer.

AlexandreSi: After thinking it through, I don't know if it's correct adding it.

AlexandreSi: After thinking it through again, I guess memoizing things is never bad

AlexandreSi: But indeed, tagsHandler seemed to change, but I didn't find why

newIDE/app/src/AssetStore/ExampleStore/index.js
```diff
@@ -39,13 +43,33 @@ export const ExampleStore = ({ isOpening, onOpen }: Props) => {
...
+    [shouldAutofocusSearchbar, focusOnMount]
+  );
+
+  const tagsHandler = React.useMemo(
+    () => ({
+      add: filtersState.addFilter,
+      remove: filtersState.removeFilter,
+      chosenTags: Array.from(filtersState.chosenFilters),
+    }),
+    [filtersState]
```
4ian, line 70: A bit nervous about this useMemo, but I guess if addFilter/removeFilter/chosenFilters change, then filtersState change so it should be correct.

newIDE/app/src/AssetStore/ExtensionStore/index.js
```diff
@@ -61,6 +61,15 @@ export const ExtensionStore = ({
       )
     : null;
 
+  const tagsHandler = React.useMemo(
+    () => ({
+      add: filtersState.addFilter,
+      remove: filtersState.removeFilter,
+      chosenTags: Array.from(filtersState.chosenFilters),
+    }),
+    [filtersState]
```
4ian, line 70: (ditto)

newIDE/app/src/ProjectCreation/services/BrowserCreation.js
```diff
@@ -12,16 +12,14 @@ import {
...
 const gd: libGDevelop = global.gd;
 
-// Signatures of this function and its arguments contain useless arguments
-// because this function has to match the signature of LocalCreation.onCreateBlank
-// that needs these arguments.
 export const onCreateBlank: OnCreateBlankFunction = async ({
   i18n,
-  outputPath,
-  projectName,
+  settings,
```
4ian, line 17: Way cleaner 👍

4ian (approved):
LGTM!

PR #3514, opened 2022-01-19
Fix: Only move unlocked instances when using keyboard in the scene editor

AlexandreSi:
Resolves https://forum.gdevelop-app.com/t/locked-object-moving/37568/6


4ian (approved):
lgtm!

PR #3526, opened 2022-01-21
"Start (or reset) the timer"

tristanbob:
Change wording to make it match the short description of the action.

Are there any other places this needs to be changed?


4ian:
I think these are the only places - we only have two timers kind (objects and scene timers) and these are their sentences that you properly modified. Thanks!

PR #3527, opened 2022-01-21
Set animation to 0 if value is an empty string

### LINKED ISSUE #3518: Animation number error
If the Animation field in the properties panel is empty, it shows "1.50995e+08".

AlexandreSi:
Fixes #3518 

@4ian I'm not familiar with the instances properties editor, but I don't know why in `InstancePropertiesEditor._renderInstancesProperties`, in the instanceSchema, `Animation` is set as a string.

EDIT

Maybe because `string` is the fallback if there is no type set.


4ian:
> Maybe because string is the fallback if there is no type set.

Correct, the property is constructed directly from the value and this gives a string type:
image

We should set the type to "number" to fix this.

4ian:
(I took care of this in another commit ;))

PR #3534, opened 2022-01-21
[Platformer] Add a parameter "try to preserve the current speed" to the "max falling speed" action.

D8H:
This is used by this extension (with JS) to avoid the object to move back up when the wall sliding is activated (max falling speed is reduced):
* https://github.com/GDevelopApp/GDevelop-extensions/pull/325

It also makes explicit when speed, acceleration and deceleration are horizontal.


4ian (approved):
👍✨

PR #3537, opened 2022-01-22
[Platformer] Add actions to allow to dash toward the ground.

D8H:
This adds
* an action to abort a jump
* an action to change the current falling speed when falling and not jumping.

It allows to make a dashing toward the ground like in Dead Cells.

In the abort action, I put a boolean parameter to keep the current falling speed or not. But, I think it might be better to always set the falling speed to 0 when aborting a jump because user probably won't understand where the falling speed comes from.

A merged demo build: https://liluo.io/instant-builds/242e0f29-c0c8-4b9f-9c39-b2ab72cad509?dev=true
* https://github.com/4ian/GDevelop/pull/3539

The demo project: https://www.dropbox.com/s/exajqjrd4j92pzj/platformer-vertical-dash.zip?dl=1

Note: in the events, I should have put the max speed action before the speed action because it's capped. We could not cap it to avoid this kind of issue but in this case the speed expression can be greater than the max speed expression which might not be great either.
VerticalDash

### DIFF
--- a/Extensions/PlatformBehavior/Extension.cpp
+++ b/Extensions/PlatformBehavior/Extension.cpp
@@ -372,6 +372,19 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
         .AddParameter("object", _("Object"))
         .AddParameter("behavior", _("Behavior"), "PlatformerObjectBehavior");
 
+    aut.AddScopedAction(
+           "AbortJump",
+           _("Abort jump"),
+           _("Abort the current jump. This action doesn't have any effect "
+             "when the character is not jumping."),
+           _("Abort the current jump of _PARAM0_"),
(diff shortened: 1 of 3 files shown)


4ian (approved):

4ian:
> We could not cap it to avoid this kind of issue but in this case the speed expression can be greater than the max speed expression which might not be great either.

Yes I think it's a bit risky to allow this. Let's assume that people will modify the max speed before modifying the speed itself.

4ian:
> But, I think it might be better to always set the falling speed to 0 when aborting a jump because user probably won't understand where the falling speed comes from.

Yeah possibly. Let's maybe in another PR change this to always reset the falling speed to 0, and we can add back the parameter in the future if required.

PR #3539, opened 2022-01-23
[Platformer] Add an action to change the current horizontal speed.

D8H:
This adds an action to change the current horizontal speed.

It allows to make a dashing that follows the floor and handles collision without any "separate" action.

A merged demo build: https://liluo.io/instant-builds/242e0f29-c0c8-4b9f-9c39-b2ab72cad509?dev=true
* https://github.com/4ian/GDevelop/pull/3537

The demo project: https://www.dropbox.com/s/bcf1qqv27w2i63t/platformer-horizontal-dash.zip?dl=1

HorizontalDash

### DIFF
--- a/Extensions/PlatformBehavior/Extension.cpp
+++ b/Extensions/PlatformBehavior/Extension.cpp
@@ -619,8 +619,9 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
 
     aut.AddScopedAction("SetCurrentSpeed",
                      _("Current horizontal speed"),
-                     _("Change the current horizontal speed of the object (in pixels per "
-                       "second)."),
+                     _("Change the current horizontal speed of the object "
+                     "(in pixels per second). The object moves to the left "
+                     "with negative values and to the right with positive ones"),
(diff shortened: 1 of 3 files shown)


Extensions/PlatformBehavior/Extension.cpp
```diff
@@ -583,6 +583,19 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
         .MarkAsAdvanced()
         .SetFunctionName("GetCurrentJumpSpeed");
 
+    aut.AddScopedAction("SetCurrentSpeed",
+                     _("Current horizontal speed"),
+                     _("Change the current horizontal speed of the object (in pixels per "
```
4ian, line 588: This horizontal speed is different from other speeds as it has the notion of direction (i.e: it's not absolute). I think we should somehow call this out in the naming/description?

D8H: Yes, we could add "The object moves to the left with negative values and to the right with positive ones" in the description of action, condition and expression.

I'll do a PR to change this and remove the parameter from the PR.

PR #3544, opened 2022-01-23
[Platformer] Always stop the character vertically when the jump is aborted.

D8H:
Also
* Add a small test on abortJump.
* Add description some about the sign of current speeds
* Try to make descriptions more standard (as it would be with an all-in-one definition).


4ian (approved):

4ian:
LGTM! :)

PR #3546, opened 2022-01-24
[Platformer] Allow the character to go down from a jumpthru.

D8H:
It also removes the dependency from PlatformerObjectRuntimeBehavior to BehaviorRBushAABB as it's an internal usage that PlatformRuntimeBehavior should not expose.

Test project: https://www.dropbox.com/s/gbnry9mlix4ht7e/PlatformJumpthruTest.zip?dl=1


4ian (approved):
🚀💯👍

Issue #3536, opened 2022-01-22
Tags, categories and keywords - search feature is inefficient


PascalLadalle:
I feel like I'm having more and more trouble finding what I'm looking for in GDevelop, and I'm supposed to be an experienced user.
The search fields, be they for events or examples, return too much noise.

Here, for instance, I want a mouse action, but can't remember the name, so I write "mouse".
The cursor action I'm looking for appears **below** the unrelated force actions.
I suppose it's because of "**MO**vement **US**ing forc**E**s", but what's the point of that? How comes such a far-fetched match comes before a full word category match?
And I can't use quotes in that field for a strict search, to filter the noise out.
![image](https://user-images.githubusercontent.com/25978830/150649597-a2cb2f9b-1915-47f6-a912-2e1c2c87fe4f.png)

Now, in the Examples section. I'm looking for an example with dragging.
The **first** result that actually contains my keyword is, as you see with the scrollbar, near the end of the results (about 15th position).
![image](https://user-images.githubusercontent.com/25978830/150649782-323d50ad-7fd4-46f2-9008-18bd00f11a1c.png)
I suppose that all the results before that one use the draggable behavior and have been tagged as such, but I think the search result priority should be: title > description > invisible tags.
On a side note, this search didn't return the example I was looking for, because it's called "Move object with mouse joint". 😅 

I hope you can do something about it. 😃

Silver-Streak:
The most recent update definitely did something with the examples search.

It feels like the search logic got hampered somehow. Searching for an example by partial name is now impossible, its like its going Tag > exact name search only, whereas most search engines go Exact Match > Partial Match > Metadata (tags/description/etc)

I know the layout change was intentional, was the search result change intentional? It definitely seems much less accurate than before.

AlexandreSi:
Thank you for taking the time to report this!
I felt like there was something off with the search.

We didn't change the logic for the search, only the UI/UX. I think your suggestion of having Exact Match > Partial Match > Metadata should be the way.
Maybe even: Exact Match in Title > Exact Match in Description > Partial Match in Title > Partial Match in Description > Metadata

4ian:
I think there are two things here:

- First a "placebo effect" on the search itself - because the logic was unchanged. I think you have the impression that it's less effective because the search is not updated anymore on each keystroke (contrary to the old search and the asset store search for example).
  - @AlexandreSi I reckon we should ensure the search is done at each change in the search bar, so that we don't have to press enter.
[...]

PR #3551, opened 2022-01-25
Improve search in Example & Extension Store

AlexandreSi:
First try to solve #3536 

* Added a text highlight to bring insights to the user about why this line is displayed
* Remove tags from string to search into
* Used fuse.js to carry out the search (synchrone search)

demo_search_highlight_bis

This new search is only applied to the Example Store for the moment as it contains only 100ish items. Would need to implement the search in a web worker if we want to apply this search on every store.

You can see that fuse finds matches for strings with the same letter as the searched text (`queries` is matched with `fire` because there is `erie` in `queries`). So it can display results that could be a bit disturbing. But I think that this new search is better than the one we have now.

Fuse settings:

- `threshold: 0.35`: this is applied to the score of a match. The score is a function of the bitmap score as explained in fuse documentation. I found this value testing a few searches as the ones in the gif and this one seemed for ok than others.
- `minMatchCharLength: 2`: It is the minimum number of consecutive letters that fuse can give as a match  (The above gif has been recorded with the value 2):

  I think that the value 3 should be the one to go with (less results displayed, it is clearer).
  Nonetheless, with 3, if you search `fi`, you won't have the results with `fire` because it only matches 2 letters. But you will have `difficulty` or `modified`. But I don't think searches with only 2 letters will be the main case.
  
A few screenshots with different themes:
Screenshot 2022-01-26 at 15 06 43
Screenshot 2022-01-26 at 15 03 39
Screenshot 2022-01-26 at 14 59 31


Bouh:
The color is very flashy, can we test something less contrasted/opaque, this bother me when reading.
Yes 2 return too much results, I think people search at least a word, and if this one is too complex or boring to write they type only the beginning.

newIDE/app/src/UI/Search/UseSearchStructuredItem.js
```diff
@@ -0,0 +1,187 @@
...
+        if (discardSearch) {
+          console.info(
+            'Discarding search results as a new search was launched.'
+          );
+          return;
+        }
+        setSearchResults(
+          filterSearchItems(
+            results
+              // .filter(result => result.score <= 0.3)
```
4ian, line 159: Comment to remove

newIDE/app/src/UI/Search/UseSearchStructuredItem.js
```diff
@@ -0,0 +1,187 @@
...
+              .map(result => ({
+                ...result.item,
+                matches: result.matches,
```
4ian, line 162: I wonder if we should rather return as results:
`?Array<{|item: SearchItem, matches: Array<{| key: string, indices: number[][] |}> |}>`

To avoid "polluting the item with matches".

newIDE/app/src/Utils/GDevelopServices/Example.js
```diff
@@ -14,6 +14,8 @@ export type ExampleShortHeader = {|
   authorIds?: Array<UserPublicProfileSearch>,
   previewImageUrls: Array<string>,
   gdevelopVersion: string,
+  // Used to highlight text when searching for keywords in examples.
+  matches?: Array<{| key: string, indices: number[][] |}>,
```
4ian, line 18: I'm a bit concerned by this, I wonder if we can rather store the search results as `Array<{|item: SearchItem, matches: ...|}>` in the extension store context and use it like this rather than polluting Example.

4ian:
Seems like some weights are not properly setup ;)

image

ClementPasteau (approved):
nice improvement already!

Issue #3554, opened 2022-01-26
Clicking outside the extension installer pop-up window will initiate an update (and overwrite existing version)


tristanbob:
## Describe the bug

Clicking outside the extension installer pop-up window will initiate an update (and overwrite existing version).

## To Reproduce

1) Open Project Manager menu
2) Select an extension 
3) Click outside the pop-up window (such as accidentally missing the "Open in editor" button by a few pixels)
4) The extension is reinstalled from the extension from the repository, overwriting any changes (and losing work)

Two problems:
1) It should require confirmation before performing the update (new installs don't need confirmation)
2) Clicking outside the pop-up window should act like "Cancel" and go back to Project Manager.

## Other details

* Windows 10
* GDevelop 5.0.126

## Video

https://user-images.githubusercontent.com/8879811/151196919-42724fd1-50dc-4d4e-96f5-b883a83453a7.mp4

4ian:
Can you confirm you have the "click outside" behavior set to "Apply changes" in the preferences? 

@ClementPasteau We should remove the onApply from this dialog, it's the second time we're getting the feedback.

ClementPasteau:
I've just tested it and I can confirm this happens because the preferences are set as "Apply changes when clicking outside of Dialog"

I don't think removing `onApply` is a good idea, because it would then remove the ability to use the shortcut to submit the dialog (CMD+Enter)
I would rather suggest removing the ability to click outside the dialog?

(it's one or the other, not sure what's best)

4ian:
> I don't think removing onApply is a good idea, because it would then remove the ability to use the shortcut to submit the dialog (CMD+Enter)

True, but I wonder if this dialog is not special and we should not consider that it could be 'Ctrl+Enter'ed?
Otherwise dismissing the click outside is another solution but not sure what's best yeah 😅

tristanbob:
I can't think of any reason on this particular pop-up where using "control+enter" makes sense.  The "control+enter" keypress only makes sense when your cursor focus is inside a field (i.e. you just typed something) and you want to save it.  In this case, this pop-up has NO fields to edit so that use-case can never happen.  Therefore, we don't need the OnApply logic.  Am I understanding this correctly?

Secondly, can we add a confirmation prompt to the extension upgrade process, specifically when an extension is already installed?

4ian:
I think both things make sense yeah, no "apply" here (i.e: no ctrl+enter) and probably confirm if we know it's an upgrade.

[...]

PR #3556, opened 2022-01-26
Remove the ability to install extension via shortcut or click outside to prevent mistakes

ClementPasteau:
addressing #3554 

- remove ability to submit via shortcut
- add warning when updating an extension

image

### DIFF
--- a/newIDE/app/src/AssetStore/ExtensionStore/ExtensionInstallDialog.js
+++ b/newIDE/app/src/AssetStore/ExtensionStore/ExtensionInstallDialog.js
@@ -22,6 +22,7 @@ import { ColumnStackLayout } from '../../UI/Layout';
 import { IconContainer } from '../../UI/IconContainer';
 import { UserPublicProfileChip } from '../../UI/User/UserPublicProfileChip';
 import RaisedButton from '../../UI/RaisedButton';
+import Window from '../../Utils/Window';
 
 const getTransformedDescription = (extensionHeader: ExtensionHeader) => {
   if (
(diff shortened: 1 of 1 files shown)


newIDE/app/src/AssetStore/ExtensionStore/ExtensionInstallDialog.js
```diff
@@ -86,7 +87,17 @@ const ExtensionInstallDialog = ({
   const canInstallExtension = !isInstalling && isCompatible;
   const onInstallExtension = React.useCallback(
     () => {
-      if (canInstallExtension) onInstall();
+      if (canInstallExtension) {
+        if (alreadyInstalled) {
+          const answer = Window.showConfirmDialog(
+            'This extension is already in your project, this will install the latest version. Do you want to continue?'
```
4ian, line 93: This extension is already in your project, this will install the latest version. You may have to do some adaptations to make sure your game continue to work. Do you want to continue?

4ian (approved):

tristanbob:
This looks great!  I'm not sure the best place to mention this but I have two ideas for improvement:

1) Show installed and proposed version numbers
```
Installed version: 1.2.3
Proposed version: 1.2.3
```

2) Add a "Release notes" field to all extensions that can be access from this dialogue.

ClementPasteau:
Those are really valid points @tristanbob.

Part of our plan is to improve the behaviors & extensions UI overall (better description, maybe an image showing what it helps doing etc...)
I'll add those 2 suggestions to our list when we come to it :)

PR #3557, opened 2022-01-26
Allow event based functions to set a default value for boolean parameters.

D8H:
Manually tested with `yesorno` and `TrueOrFalse`.


4ian:
👍

PR #3558, opened 2022-01-27
Allow to set an empty color property value.

D8H:
It allows extensions to default on a color when the string is empty (the slider extension use this).

In the slider extension, I changed the optional color properties type from string to color and it works.
* https://github.com/GDevelopApp/GDevelop-extensions/pull/326#issuecomment-1022122724

I fixed the color property condition, the operator type was set to "color", but it seems the generator only handles "string" and "number".

Also see
* https://github.com/4ian/GDevelop/pull/3086
* https://github.com/4ian/GDevelop/issues/2211


4ian:
Looks good, thanks! 👍

PR #3561, opened 2022-01-28
Use "Untitled Xxx" instead of "New Xxx" in the project manager for default names

4ian:
Make it clearer these are items to be renamed, rather than buttons.

Don't show in changelog

### DIFF
--- a/newIDE/app/src/MainFrame/index.js
+++ b/newIDE/app/src/MainFrame/index.js
@@ -789,63 +789,6 @@ const MainFrame = (props: Props) => {
     toolbar.current.setEditorToolbar(editorToolbar);
   };
 
-  const addLayout = () => {
-    const { currentProject } = state;
-    if (!currentProject) return;
-
-    const name = newNameGenerator('New scene', name =>
-      currentProject.hasLayoutNamed(name)
-    );
-    const newLayout = currentProject.insertNewLayout(
-      name,
-      currentProject.getLayoutsCount()
-    );
-    newLayout.updateBehaviorsSharedData(currentProject);
-    _onProjectItemModified();
-  };
-
-  const addExternalLayout = () => {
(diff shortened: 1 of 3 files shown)


newIDE/app/src/MainFrame/index.js
```diff
@@ -789,63 +789,6 @@ const MainFrame = (props: Props) => {
     toolbar.current.setEditorToolbar(editorToolbar);
   };
 
-  const addLayout = () => {
```
ClementPasteau, line 792: nice! weird we didn't catch this duplication before 🤔

newIDE/app/src/ProjectManager/index.js
```diff
@@ -308,7 +304,7 @@ export default class ProjectManager extends React.Component<Props, State> {
   _addEventsFunctionsExtension = (index: number) => {
     const { project } = this.props;
 
-    const newName = newNameGenerator('NewExtension', name =>
+    const newName = newNameGenerator('UntitledExtension', name =>
```
ClementPasteau, line 307: `Untitled extension` to be consistent with the other names?

ClementPasteau (approved):
1 comment, but then good!

newIDE/app/src/MainFrame/index.js
4ian: Yeah 😅 Was long for a long time I guess.

newIDE/app/src/ProjectManager/index.js
4ian: Extension names must be valid "identifiers" (in the sense: usable in expressions), so like objects can't contain space.
This could be "fixed" by showing the extension full name (the user friendly name), but that would also hide this "identifier" that is used in expressions. For now I'll keep it this way!

PR #3562, opened 2022-01-29
Display objects in categories when creating a new object

4ian:
image


arthuro555:
Nice! Though I would have grouped sprites, tiled sprites, and 9-patch sprites under an "Image" (or "Sprite") category, and put the rest of the General category under a "Misc." category.

PR #3574, opened 2022-01-31
Improve Scene handling

### LINKED ISSUE #3560: Improve starting scene selection
## Description

Currently, when a game starts, the scene that plays is based on its position in the list of scenes.  This is not obvious to many new users.

## Solution suggested

- [ ] Provide a visual indicator of the scene that is the "starting scene"
 -  Small icon on the starting scene (could be placed left of text, or left of triple-dots)
 -  Hovering on that icon shows a tool-tip "Starting scene"

- [ ] Allow users to specify the starting scene
 - Right-click on scene name allows users to select "Make starting scene"

## Screenshot

Untitled

AlexandreSi:
2 improvements in this PR:
- Fixes #3560 (only the main issue - the drag n drop of project manager items will be handled in another PR)
- When creating a blank project, it automatically creates an Untitled Scene and opens it, so that new users will have a step less to create their first game!

demo_set_first_scene

There's a tooltip on the flag icon that says `This scene will be used as the first scene displayed at game launch.`

---

EDIT:
- Wording has been changed to `Set as start scene` and `This scene will be used as the start scene.`

### DIFF
--- a/newIDE/app/src/ProjectManager/index.js
+++ b/newIDE/app/src/ProjectManager/index.js
@@ -736,14 +736,7 @@ export default class ProjectManager extends React.Component<Props, State> {
                                   t`This scene will be used as the start scene.`
                                 )}
                               >
-                                <Flag
-                                  color="disabled"
-                                  fontSize="small"
-                                  style={{
-                                    verticalAlign: 'bottom',
-                                    marginLeft: 5,
-                                  }}
-                                />
+                                <Flag color="disabled" fontSize="small" />
                               </Tooltip>
                             ) : (
                               undefined
--- a/newIDE/app/src/ProjectManager/index.js
+++ b/newIDE/app/src/ProjectManager/index.js
@@ -1,5 +1,6 @@
 // @flow
 import { Trans } from '@lingui/macro';
+import { I18n } from '@lingui/react';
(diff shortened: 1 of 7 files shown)


4ian (approved):
LGTM! Good stuff to translate the default item labels.

newIDE/app/src/ProjectManager/index.js
```diff
@@ -641,484 +651,555 @@ export default class ProjectManager extends React.Component<Props, State> {
...
+                            name === firstLayoutName ? (
+                              <Tooltip
+                                title={i18n._(
+                                  t`This scene will be used as the start scene.`
+                                )}
+                              >
+                                <Flag
+                                  color="disabled"
+                                  fontSize="small"
+                                  style={{
```
ClementPasteau, line 742: should this style be integrated in the `<Item>` itself? so that we don't have to copy it again if we want to add a new icon later? (probably not, but it feels cleaner)

ClementPasteau (approved):
lgtm! 

One question though, should we always have a starting scene?
Meaning:
- "untitled scene" is defaulted as starting scene
- deleting a starting scene will transfer it to the next one
- deleting all and creating one will default it to starting scene

PR #3577, opened 2022-01-31
Add a button to open events from the scene editor context menu

4ian:
Tested on scene and external layouts.
Should help a tiny bit some people not finding the events tab during their first usage of GDevelop.

### DIFF
--- a/newIDE/app/src/MainFrame/index.js
+++ b/newIDE/app/src/MainFrame/index.js
@@ -2160,6 +2160,11 @@ const MainFrame = (props: Props) => {
                   projectItemName: editorTab.projectItemName,
                   setPreviewedLayout,
                   onOpenExternalEvents: openExternalEvents,
+                  onOpenEvents: (sceneName: string) =>
+                    openLayout(sceneName, {
+                      openEventsEditor: true,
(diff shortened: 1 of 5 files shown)


newIDE/app/src/MainFrame/index.js
```diff
@@ -2160,6 +2160,11 @@ const MainFrame = (props: Props) => {
                   projectItemName: editorTab.projectItemName,
                   setPreviewedLayout,
                   onOpenExternalEvents: openExternalEvents,
+                  onOpenEvents: (sceneName: string) =>
+                    openLayout(sceneName, {
+                      openEventsEditor: true,
+                      openSceneEditor: false,
```
ClementPasteau, line 2166: just to confirm, this is not closing the scene editor right?

ClementPasteau (approved):
lgtm!

newIDE/app/src/MainFrame/index.js
4ian: Correct, this is telling to open the events editor and to not open the scene editor if it's not open already. If it's opened already (which is most probably the case), it stays untouched. In particular, it's not closed :)

PR #3579, opened 2022-01-31
[PathFinding] Remove the dependency from PathfindingRuntimeBehavior to BehaviorRBushAABB

D8H:
Remove the dependency from PathfindingRuntimeBehavior to BehaviorRBushAABB as its an internal usage that PathfindingObstacleRuntimeBehavior should not expose.

### DIFF
--- a/Extensions/PathfindingBehavior/pathfindingobstacleruntimebehavior.ts
+++ b/Extensions/PathfindingBehavior/pathfindingobstacleruntimebehavior.ts
@@ -83,12 +83,12 @@ namespace gdjs {
       searchArea.maxX = x + radius;
       // @ts-ignore
       searchArea.maxY = y + radius;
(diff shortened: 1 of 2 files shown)


Extensions/PathfindingBehavior/pathfindingobstacleruntimebehavior.ts
```diff
@@ -70,7 +70,7 @@ namespace gdjs {
       x: float,
       y: float,
       radius: float,
-      result: gdjs.BehaviorRBushAABB<gdjs.PathfindingObstacleRuntimeBehavior>[]
+      result: gdjs.PathfindingObstacleRuntimeBehavior[]
     ): any {
```
AlexandreSi, line 74: Should we have typed the return of this function to avoid this type mismatch?

AlexandreSi (approved):
LGTM

PR #3580, opened 2022-01-31
[Light] Remove the dependency from LightRuntimeObjectPixiRenderer to BehaviorRBushAABB

D8H:
Remove the dependency from LightRuntimeObjectPixiRenderer to BehaviorRBushAABB as its an internal usage that LightObstaclesManager should not expose.

### DIFF
--- a/Extensions/Lighting/lightobstacleruntimebehavior.ts
+++ b/Extensions/Lighting/lightobstacleruntimebehavior.ts
@@ -54,7 +54,7 @@ namespace gdjs {
       object: gdjs.RuntimeObject,
       radius: number,
       result: gdjs.LightObstacleRuntimeBehavior[]
-    ) {
+    ): void {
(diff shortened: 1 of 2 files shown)


Extensions/Lighting/lightobstacleruntimebehavior.ts
```diff
@@ -53,7 +53,7 @@ namespace gdjs {
     getAllObstaclesAround(
       object: gdjs.RuntimeObject,
       radius: number,
-      result: gdjs.BehaviorRBushAABB<gdjs.LightObstacleRuntimeBehavior>[]
+      result: gdjs.LightObstacleRuntimeBehavior[]
     ) {
```
AlexandreSi, line 57: Same as for the other PR: Adding the return type for this function could have helped

AlexandreSi (approved):

PR #3583, opened 2022-01-31
Sort actions/conditions in categories with icons

4ian:
This rework the extension to have a category (which can be set for extensions made in the editor) and to be able to specify icons to be displayed in the action/condition/expression selector.

This will be useful for:
- first time users, as it gives a less "busy" appearance for this list. 
- all users as the icons will help to quickly find what we need. 
- developers as you have less groups to specify (everything is grouped inside your extension group)

This is just reworking the organisation and not changing the name of the groups/actions/conditions (apart for the group of 3 or 4 of them, because they had to be moved to another extension).

- [ ] We could in theory use the icon url of the extension, but for now the group icons are set in another way.
- [ ] Because expressions are listing the object expressions, unsure if we should support icons here too.
- [x] Fix full group name finishing by /

I went ahead with categories being: general, input, ads, device, audio, network, third-party, advanced.
Most stuff should get in general unless they are advanced concepts or adapted to another category. This is a bit arbitrary but I don't want too many categories, this would defeat the purpose.

Conditions:
image
image

Actions:
image
image

Expressions too:
image
image

--- 

Side note: we could even think of going toward a 3 step dialog, much like objects. Instead of opening groups with a drop down effect, you would click it and it would open a second part of the dialog, like objects. This would help people consider  extensions as some kind of "always existing objects".

But for now I don't want to introduce a large change.


Core/GDCore/Extensions/Builtin/AdvancedExtension.cpp
```diff
@@ -80,9 +71,9 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAdvancedExtension(
                       "a string, an empty string is considered as \"false\". "
                       "If it's a number, 0 is considered as \"false\"."),
                     _("Parameter _PARAM0_ is true"),
-                    _("Functions"),
-                    "res/function24.png",
-                    "res/function16.png")
+                    "",
+                    "res/function32.png",
+                    "res/function32.png")
```
ClementPasteau, line 76: why have you changed some instructions icons to 32, and not others?

Core/GDCore/Extensions/Builtin/AudioExtension.cpp
```diff
@@ -49,7 +52,7 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAudioExtension(
                  _("Stop the sound of a channel"),
                  _("Stop the sound on the specified channel."),
                  _("Stop the sound of channel _PARAM1_"),
-                 _("Audio/Sounds on channels"),
+                 _("Sounds on channels"),
                  "res/actions/son24.png",
```
ClementPasteau, line 56: like here

Core/GDCore/Extensions/Builtin/CameraExtension.cpp
```diff
@@ -31,7 +33,7 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsCameraExtension(
           _("Camera center X position"),
           _("the X position of the center of a camera"),
           _("the X position of camera _PARAM4_ (layer: _PARAM3_)"),
-          _("Layers and cameras"),
+          "",
           "res/conditions/camera24.png")
```
ClementPasteau, line 37: if the group for the extension already has this icon defined, is it needed to define it per instruction?

Core/GDCore/Extensions/Builtin/FileExtension.cpp
```diff
@@ -163,20 +166,21 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsFileExtension(
...
                     _("Test if the specified storage exists."),
                     _("Storage _PARAM0_ exists"),
-                    _("Storage"),
+                    "",
                     "res/conditions/fichier24.png",
                     "res/conditions/fichier.png")
       .AddParameter("string", _("Storage name"))
       .MarkAsAdvanced();
 
+  // TODO: Move this
```
ClementPasteau, line 175: todo

Core/GDCore/Extensions/Builtin/MathematicalToolsExtension.cpp
```diff
@@ -18,13 +18,63 @@ BuiltinExtensionsImplementer::ImplementsMathematicalToolsExtension(
...
+                     "res/actions/position.png")
+      .SetHelpPath("/all-features/expressions")
+      .AddParameter("expression", _("Minimum value"))
+      .AddParameter("expression", _("Maximum value"));
+
+  extension
+      .AddExpression("RandomWithStep",
+                     _("Random value in steps"),
+                     _("Random value in steps"),
+                     _("Random"),
```
ClementPasteau, line 66: do we need a category for those expressions, considering the others don't have one?

Core/GDCore/Extensions/Builtin/SceneExtension.cpp
```diff
@@ -14,67 +14,20 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsSceneExtension(
...
-                     _("Random value in steps"),
-                     _("Random value in steps"),
-                     _("Random"),
-                     "res/actions/position.png")
-      .SetHelpPath("/all-features/expressions")
-      .AddParameter("expression", _("Minimum value"))
-      .AddParameter("expression", _("Maximum value"))
-      .AddParameter("expression", _("Step"));
-
+  // TODO: move this.
```
ClementPasteau, line 25: I think you can remove this now

Core/GDCore/Extensions/Builtin/TimeExtension.cpp
```diff
@@ -14,21 +14,26 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsTimeExtension(
   extension
       .SetExtensionInformation(
           "BuiltinTime",
-          _("Time"),
+          _("Timers and time"),
```
ClementPasteau, line 17: It feels weird to have 2 times "time" in the text (pun intended)

What about `Time tools`? or `Time mechanics` or `Time manipulation`

Core/GDCore/Extensions/Platform.h
```diff
@@ -52,7 +53,6 @@ class GD_CORE_API Platform {
    */
   virtual gd::String GetFullName() const { return "Unnamed platform"; }
 
-#if defined(GD_IDE_ONLY)
```
ClementPasteau, line 55: 🎉

Core/GDCore/Extensions/Platform.h
```diff
@@ -149,28 +160,18 @@ class GD_CORE_API Platform {
...
+  };
 
  private:
   std::vector<std::shared_ptr<PlatformExtension>>
       extensionsLoaded;  ///< Extensions of the platform
   std::map<gd::String, CreateFunPtr>
       creationFunctionTable;  ///< Creation functions for objects
+  std::map<gd::String, InstructionOrExpressionGroupMetadata>
+      instructionOrExpressionGroupMetadata;
+  static InstructionOrExpressionGroupMetadata badInstructionOrExpressionGroupMetadata;
```
ClementPasteau, line 174: I'm not sure I understand why this is called `bad...`?

Core/GDCore/Extensions/PlatformExtension.h
```diff
@@ -5,8 +5,7 @@
  * project is released under the MIT License.
  */
 
-#ifndef GDCORE_PLATFORMEXTENSION_H
-#define GDCORE_PLATFORMEXTENSION_H
+#pragma once
```
ClementPasteau, line 8: what does this do?

Extensions/DialogueTree/JsExtension.js
```diff
@@ -28,12 +28,16 @@ module.exports = {
     extension
       .setExtensionInformation(
         'DialogueTree',
-        _('Dialogue Tree (Experimental)'),
+        _('Dialogue Tree (experimental)'),
```
ClementPasteau, line 31: Is this extension still experimental?

Extensions/ExampleJsExtension/JsExtension.js
```diff
@@ -32,6 +32,9 @@ module.exports = {
       'Florian Rival',
       'MIT'
     );
+    extension
```
ClementPasteau, line 35: can you remind me why we need this DummyExtension?

GDJS/GDJS/Extensions/Builtin/AdvancedExtension.cpp
```diff
@@ -17,9 +17,6 @@ namespace gdjs {
 AdvancedExtension::AdvancedExtension() {
   gd::BuiltinExtensionsImplementer::ImplementsAdvancedExtension(*this);
 
-  GetAllConditions()["Toujours"].SetFunctionName(
```
ClementPasteau, line 20: why remove?

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/ExtensionOptionsEditor.js
```diff
@@ -190,6 +191,37 @@ export const ExtensionOptionsEditor = ({
...
           />
+          <SemiControlledAutoComplete
+            floatingLabelText={<Trans>Category (shown in the editor)</Trans>}
+            fullWidth
+            value={eventsFunctionsExtension.getCategory()}
+            onChange={category => {
+              eventsFunctionsExtension.setCategory(category);
+              forceUpdate();
+            }}
+            dataSource={[
```
ClementPasteau, line 202: did I also see 'network' and 'device' as categories?

newIDE/app/src/InstructionOrExpression/EnumerateInstructions.js
```diff
@@ -63,6 +63,15 @@ const freeInstructionsToRemove = {
   ],
 };
 
+const getExtensionPrefix = (extension: gdPlatformExtension): string => {
```
ClementPasteau, line 66: function duplicated? Should be shared maybe?

ClementPasteau (commented):
sick!

Core/GDCore/Extensions/Builtin/AdvancedExtension.cpp
4ian: Ideally I would have changed them all to a svg file ;) The reason is that the old IDE was working only with PNG. Now we could have single SVG icon everywhere. At least we're not forced to feed the IDE with two version of the PNG, as the browser will resize them when displaying them (the old IDE needed a specific size). So I just changed as I went some icons to use the larger version we have.

Core/GDCore/Extensions/Builtin/AudioExtension.cpp
4ian: Just no time to change everything, that's the reason ;)

Core/GDCore/Extensions/Builtin/CameraExtension.cpp
4ian: Ideally no, we would define only the group icon and the instruction would inherit from it. But this is not how it works today. 
We could at least rework this in the future so that the default icon for all instructions/expressions are the extension icon but I've not done it because it's a lot of work.
But that would be neat, we would remove 2 lines for each instruction, and could still keep specific icons for some instructions.

Core/GDCore/Extensions/Builtin/TimeExtension.cpp
4ian: Agree but maybe let's not rename this in this PR, to avoid users having trouble finding this if they follow tutorials?
Also the repetition is not that much of a problem I would say, especially because it's different in other languages :)

Core/GDCore/Extensions/Platform.h
4ian: This is the null object pattern: https://en.wikipedia.org/wiki/Null_object_pattern
The badXxx objects are objects that are valid objects but empty and returned in case of you trying to access something that is not existing (like here, a group that would not have been declared). 

This is a logical error but we avoid crashing the app by returning a null pointer or something like this (the C++ guarantee that you will get an object, but if you asked for an non existing one, you will get the "null object").

Core/GDCore/Extensions/PlatformExtension.h
4ian: pragma once is the equivalent of ifndef/define/endif

This avoids .h include files to be included more than once. If your .cpp file includes two .h files, and these two .h files include the same one, it must be included only once (otherwise C++ will complain about double declarations of classes/functions).
The answer with recent compilers since a few years is to use #pragma once to ask the compiler to do this deduplication.

Old C++ stuff basically!

Extensions/DialogueTree/JsExtension.js
4ian: Maybe not indeed!

Extensions/ExampleJsExtension/JsExtension.js
4ian: As an example for anyone wanting to try a new extension. A sort of living documentation. Also can be useful to quickly check that an extension can do something or as a sandbox.

GDJS/GDJS/Extensions/Builtin/AdvancedExtension.cpp
4ian: Moved to CommonInstructionsExtension :)

newIDE/app/src/EventsFunctionsExtensionEditor/OptionsEditorDialog/ExtensionOptionsEditor.js
4ian: Great catch!

newIDE/app/src/InstructionOrExpression/EnumerateInstructions.js
4ian: Indeed, I'll import this one from the other file.

Core/GDCore/Extensions/Builtin/FileExtension.cpp
4ian: Done

Core/GDCore/Extensions/Builtin/MathematicalToolsExtension.cpp
4ian: Removing for consistency!

Core/GDCore/Extensions/Builtin/TimeExtension.cpp
4ian: I'll keep this for now to avoid confusing people searching for this group :)

D8H:
There are some extensions that would fit well inside built-in groups, for instance:
* the drag camera extension in "Layer and camera"
* the mouse extensions in "Mouse and touch"
* extended variable extension in "Variables"

The issue is that it will make it difficult to find the actions of a freshly installed extension.

> Side note: we could even think of going toward a 3 step dialog, much like objects. Instead of opening groups with a drop down effect, you would click it and it would open a second part of the dialog, like objects. This would help people consider extensions as some kind of "always existing objects".

The 1st column could plain list the extensions. A selection could act as a filter for the existing 2nd column. This way merging groups across extensions won't be an issue anymore. This will let users browse by extension or by semantic.

D8H:
There is a prefix issue on some function that has been moved.
* 127
```JavaScript
          "conditions": [
            {
              "type": {
                "inverted": false,
                "value": "Egal"
              },
              "parameters": [
                "1",
                "<",
                "0"
              ],
              "subInstructions": []
            }
          ],
```
* Master
```JavaScript
          "conditions": [
            {
              "type": {
                "inverted": false,
                "value": "BuiltinCommonInstructions::Egal"
              },
              "parameters": [
                "0",
                "<",
                "1"
              ],
              "subInstructions": []
            }
          ],
```

4ian:
Oops, thanks for checking after me! I'll check that.

> This would allow to replace one action by another without scrolling too much.
> The issue is that it could make difficult to find the actions of a freshly installed extension.

Yep I'm not sure about this, for now these extensions will live in the own category.. at least it's simple.

> Also, automatically extend sub groups when extending a group (or selecting in the 1st column) could save a lot of clicks (for instance the variable group).

Yep, that's why I'm also thinking of the "2nd column" approach, like done for objects. Because this allows the 2nd column to display stuff sorted in categories that are shown as sub headers, so no extra clicks.

D8H:
The new function might not be detected by the generator because it writes a no name function

```JavaScript
val = (-(12345), "<=", 0);
```

instead of converting it to an operator
```JavaScript
val = (-(1234) <= 0);
```

4ian:
Should be properly fixed this time in 49c5da6505e5dc3617ef6a636bc71d662eca3805 - tested manually the new functions (always, compare numbers, compare strings).

PR #3587, opened 2022-02-01
Fix search highlights that made disappear non exact matches

AlexandreSi:


newIDE/app/src/UI/Search/HighlightedText.js
```diff
@@ -39,6 +39,7 @@ const HighlightedText = ({
   matchesCoordinates: number[][],
   styleToApply: { backgroundColor?: string, color?: string },
 |}): React.Node[] => {
+  if (matchesCoordinates.length === 0) return [text];
```
4ian, line 42: question: This matches everything? or nothing?

ClementPasteau (approved):
neat!

newIDE/app/src/UI/Search/HighlightedText.js
AlexandreSi: It's a security in case the filtering actions after the search make the coordinates empty.

AlexandreSi: Well, I'm checking

AlexandreSi: Yep, that's it, if the coordinates are empty, we return a map on `returnText` that would be empty

PR #3588, opened 2022-02-01
Highlight add new event button

AlexandreSi:
My try at having the button glow to make it more visible, only when there is no event

demo_glowing_button

### DIFF
--- a/newIDE/app/src/AssetStore/ExampleStore/ExampleListItem.js
+++ b/newIDE/app/src/AssetStore/ExampleStore/ExampleListItem.js
@@ -20,7 +20,7 @@ import { openExampleInWebApp } from './ExampleDialog';
 import { UserPublicProfileChip } from '../../UI/User/UserPublicProfileChip';
 import HighlightedText from '../../UI/Search/HighlightedText';
 import GDevelopThemeContext from '../../UI/Theme/ThemeContext';
-import { type SearchMatches } from '../../UI/Search/UseSearchStructuredItem';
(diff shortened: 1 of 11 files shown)


tristanbob:
I like the color glow effect. 

However, why not use the same strategy that you recently used when creating a new project?  Previously users had to click "Add scene", but you skipped that step and created it for them.  Why not create an empty event?

4ian:
I think there is a limit to find between "presenting you with ready made stuff" and "let you discover the concepts by your own". In the case of a scene, it was less important to teach it to a new user. What is important is that a user discovers what's an object, what's a behavior, what's an event. But adding a scene to a project can be discovered later (I click on the project manager and "wow there is already a scene but I can make more of them").

For events, it's a bit different. People have to learn that they can add an event, composed of conditions and actions. If we skip the events creation, they might assume there is a unique single event and that each condition is associated to a unique action (we already had this confusion for some users). So I think it's better if they do themselves the work of clicking on "Add event", so they understand this can be repeated.

AlexandreSi:
As I'm away for a bit, if you want to merge the contrast boost, feel free to remove the last commit!

4ian:
I've merged the contrast fix 👍 Letting this open so we can continue discussing this.

Bouh:
Careful in this pr there is not just a Glow effect, you include changes for research matches.

AlexandreSi:
Oups, I did it again!
Closing it any way, we will soon change it I guess.

PR #3589, opened 2022-02-01
Add errors when zipped file for web is too big

ClementPasteau:
It's been pointed out multiple times that exporting a game sometimes times out.

The goal of this PR is to take a first step towards giving more information to the user about the limitations of the backend services in terms of publishing a game.
This PR only addresses Web Builds and the fact that if the files are too big, then the packaging will not work.
Plus if they are too big, they shouldn't be uploaded to the web, as this will impact the experience, so we can then give them recommendations on how to optimise their assets.

The limit here is set at 250Mb.

Also fixed various design issues in the builds when errored.

Error appearing in the local app:
Screenshot 2022-02-01 at 18 27 30

Error appearing in the web app: (size is just for test)
Screenshot 2022-02-01 at 18 28 25

If the build is forced, the backend rejects the files too, and the build is errored + message in logs:
Screenshot 2022-02-01 at 18 28 04

### DIFF
--- a/newIDE/app/src/Export/BrowserExporters/BrowserOnlineWebExport.js
+++ b/newIDE/app/src/Export/BrowserExporters/BrowserOnlineWebExport.js
@@ -144,6 +144,7 @@ export const browserOnlineWebExportPipeline: ExportPipeline<
       textFiles,
       basePath: '/export/',
       onProgress: context.updateStepProgress,
+      sizeLimit: 250 * 1024 * 1024,
     });
   },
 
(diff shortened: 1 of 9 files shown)


arthuro555:
That byte count doesn't look very user-friendly, could it be shown in mega-bytes instead?

Also, is the limit of the build service really at 250MB? I thought it was behind Cloudflare, and read somewhere Cloudflare blocked uploads of >100mb.

ClementPasteau:
I agree the message should be improved!

And no, there is no limitation from Cloudflare, it's more related to our backend architecture.
(I did read something about Cloudflare too, but technically we're not uploading via cloudflare :))

4ian (approved):
Let's display MB instead of bytes (and divide by 1000*1000) (or display MiB and divide by 1024*1024 according to https://en.wikipedia.org/wiki/Byte#Multiple-byte_units).

We could also report the error as part of the response of the promise (by typing as `{| isTooLarge: true |} | {| outputFilename: string |}` for the local app, and a bit differently for the web app). For now I think it's ok.

PR #3597, opened 2022-02-04
Improve selector search

### LINKED ISSUE #3536: Tags, categories and keywords - search feature is inefficient

AlexandreSi:
Resolves #3536 

Applies the same search behavior as for the example and extension stores.

demo_highlight_search_instruction

### DIFF
--- a/newIDE/app/src/EventsSheet/InstructionEditor/InstructionOrObjectSelector.js
+++ b/newIDE/app/src/EventsSheet/InstructionEditor/InstructionOrObjectSelector.js
@@ -58,6 +58,7 @@ import { type EventsScope } from '../../InstructionOrExpression/EventsScope.flow
 import {
   type SearchResult,
   tuneMatches,
+  sharedFuseConfiguration,
 } from '../../UI/Search/UseSearchStructuredItem';
 
 const gd: libGDevelop = global.gd;
@@ -69,13 +70,6 @@ const styles = {
 };
 
 const DISPLAYED_INSTRUCTIONS_MAX_LENGTH = 20;
-const sharedFuseConfiguration = {
-  minMatchCharLength: 1,
-  threshold: 0.35,
-  includeMatches: true,
-  ignoreLocation: true,
-  ignoreFieldNorm: true,
-};
 
 export type TabName = 'objects' | 'free-instructions';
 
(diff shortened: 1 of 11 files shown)


AlexandreSi:
I used a Fuse client for the tags as well because I think the users will expect to have the same fuzzy behavior between all the displayed results.

newIDE/app/src/EventsSheet/InstructionEditor/InstructionOrObjectSelector.js
```diff
@@ -145,8 +169,77 @@ export default class InstructionOrObjectSelector extends React.PureComponent<
...
+    });
   }
 
+  _search = (searchText: string) => {
+    if (searchText === '') return;
+
+    this.setState({
+      searchResults: {
+        objects: this.objectSearchApi
+          ? this.objectSearchApi.search(`'${searchText}`).map(result => ({
```
4ian, line 208: What is this ' useful for?

newIDE/app/src/UI/Search/UseSearchStructuredItem.js
```diff
@@ -34,16 +43,12 @@ const tuneMatchIndices = (match: SearchMatch, searchText: string) => {
     .filter(Boolean);
 };
 
-const tuneMatches = (
-  result: {| item: any, matches: SearchMatch[] |},
-  searchText: string
-) => {
-  return result.matches.map(match => ({
+export const tuneMatches = <T>(result: SearchResult<T>, searchText: string) =>
```
4ian, line 46: 👍

4ian (approved):

newIDE/app/src/EventsSheet/InstructionEditor/InstructionOrObjectSelector.js
AlexandreSi: I'm not sure but the results felt better : https://fusejs.io/examples.html#extended-search

PR #3601, opened 2022-02-04
Fix help paths of objects

Bouh:
Fix path for Tile map and Bitmap Text.
Video help path lead to an empty page but I notified the content creator team to fill the wiki page.


4ian:
Somehow I missed this! Thanks :)

jack-ii:
The help path for the **Light** object is still wrong.

Bouh:
The Update will be push in the next release. And page for light object is currently empty. URL will be on the objects page of our wiki.

jack-ii:
I thought of linking or copying to this, better than on a non-existent page. But doesn't matter.
https://wiki.gdevelop.io/gdevelop5/all-features/lighting/reference#lights

Bouh:
Someone at GDevelop will be in charge of these changes, it's on the rails 😉

PR #3607, opened 2022-02-09
Start search from currently selected event in Event sheet

AlexandreSi:


AlexandreSi:
@4ian I don't know if `VectorInt` is a great name, as the `gd::String::npos` is converted to `-1` in JS

Core/GDCore/IDE/Events/EventsPositionFinder.h
```diff
@@ -0,0 +1,50 @@
...
+  };
+  virtual ~EventsPositionFinder();
+
+  /**
+   * Return the positions of all searched events
+   */
+  const std::vector<std::size_t> GetPositions() { return positions; }
+
+  /**
+   * Return the positions of all searched events
```
4ian, line 33:
```suggestion
   * Add an event for which the position must be reported in `GetPositions`.
```

Core/GDCore/IDE/Events/EventsPositionFinder.h
```diff
@@ -0,0 +1,50 @@
...
+ public:
+  EventsPositionFinder() {
+    index = 0;
+  };
+  virtual ~EventsPositionFinder();
+
+  /**
+   * Return the positions of all searched events
+   */
+  const std::vector<std::size_t> GetPositions() { return positions; }
```
4ian, line 30: Return a const ref:
```suggestion
  const std::vector<std::size_t>& GetPositions() const { return positions; }
```

This won't change the usage in JS, but this will allow to avoid making a copy of this vector in memory in C++.

Also make the method const because it does not change anythign and this will be useful to someone else who wants to use your class as a const reference.

Core/GDCore/IDE/Events/EventsPositionFinder.h
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * \brief Scans an event list to retrieve the position of a list of searched
+ * events when the events list is flattened.
+ *
+ * \ingroup IDE
+ */
+class GD_CORE_API EventsPositionFinder : public ArbitraryEventsWorker {
+ public:
+  EventsPositionFinder() {
+    index = 0;
```
4ian, line 23: Uber nitpicking

```suggestion
  EventsPositionFinder() : index(0) {
```

The initializer list is recommended because it avoids to initialize to a garbage value then replace this value.

GDevelop.js/__tests__/Core.js
```diff
@@ -2426,6 +2426,68 @@ describe('libGD.js', function () {
...
+        // │   └── evt002 <-- 1
+        // ├── evt01
+        // │   └── evt010 <-- 2
+        // └── evt02
+        //     └── evt020
+        // evt1 <------------ 3
+        // └── evt10 <------- 4
+        const events = new gd.EventsList();
+
+        var evt0 = new gd.StandardEvent();
```
4ian, line 2445: Nit: use `let` instead of `var`

newIDE/app/.flowconfig
```diff
@@ -24,4 +24,4 @@
 
 [options]
 module.ignore_non_literal_requires=true
-sharedmemory.hash_table_pow=21
+sharedmemory.hash_table_pow=22
```
4ian, line 27: More memory for flow?

4ian:
VectorInt sounds fine because there are no unsigned ints in JS.

Core/GDCore/IDE/Events/EventsPositionFinder.h
AlexandreSi: Ok I see, thanks!

4ian (approved):

PR #3611, opened 2022-02-10
Fix compatibility with instructions broken in #3583

4ian:


4ian:
@D8H FYI, this is how I fixed what I broke. I'm merging this as I don't want to keep the master branch broken but let me know if you have feedback about this!

PR #3622, opened 2022-02-13
Use the extension name as file name when saving it.

D8H:
It can save a bit of time when doing changes for an extension review. GDevelop already remember the last folder so it allows to just click on save and commit.


4ian:
Great stuff 👍

PR #3625, opened 2022-02-15
A few UI improvements here and there

### LINKED ISSUE #965: Theme is applied late

### LINKED ISSUE #2213: Arrows move objects and active handles
## Describe the bug
When moving objects with the arrows on the scene, if you drag a blue handle to move the view, arrows then move both the objects and the handle/view.

## To Reproduce
1. Select objects on the scene.
2. Zoom the view on them.
3. Move them by pressing the right arrow a few seconds.
4. Grab the upper handle to recenter the view on the objects.
5. Note position of one object, then press the arrow again. The view moves because the handle is still active. Grab the handle to recenter on the objects and confirm that the position has changed.

GD5 beta 103

### LINKED ISSUE #2813: Buttons not Pinned to bottom of screen in Object Group or Global Variable Lists, display grid colors not saved

### LINKED ISSUE #3519: Identical action appears twice
The identical action is displayed twice when you search for "Create an Object".

AlexandreSi:
Closes #3519
Closes #2813
Closes #2213
Closes #965

### DIFF
--- a/Extensions/Lighting/JsExtension.js
+++ b/Extensions/Lighting/JsExtension.js
@@ -243,7 +243,7 @@ module.exports = {
     objectsEditorService.registerEditorConfiguration(
       'Lighting::LightObject',
       objectsEditorService.getDefaultObjectJsImplementationPropertiesEditor({
-        helpPagePath: '/objects/light',
+        helpPagePath: '/all-features/lighting/reference',
       })
     );
   },
--- a/newIDE/app/public/index.html
+++ b/newIDE/app/public/index.html
@@ -13,7 +13,8 @@
         margin: 0;
         padding: 0;
         height: 100%;
-        background-color: #f0f0f0;
+        background-color: #000;
+        transition: background-color 0.5s linear;
       }
 
       body {
--- a/newIDE/app/src/InstancesEditor/FullSizeInstancesEditorWithScrollbars.js
+++ b/newIDE/app/src/InstancesEditor/FullSizeInstancesEditorWithScrollbars.js
@@ -158,6 +158,7 @@ export default class FullSizeInstancesEditorWithScrollbars extends Component<
                     onChange={this._handleXChange}
                     style={styles.xScrollbar}
                     orientation="horizontal"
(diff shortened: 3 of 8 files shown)


4ian (approved):

Extensions/Lighting/JsExtension.js
```diff
@@ -243,7 +243,7 @@ module.exports = {
     objectsEditorService.registerEditorConfiguration(
       'Lighting::LightObject',
       objectsEditorService.getDefaultObjectJsImplementationPropertiesEditor({
-        helpPagePath: '/objects/light',
+        helpPagePath: '/all-features/lighting/reference',
```
Bouh, line 246: This is a temporary redirection, Wesley and Max will rework the object light page 👍

newIDE/app/public/index.html
```diff
@@ -13,7 +13,8 @@
         margin: 0;
         padding: 0;
         height: 100%;
-        background-color: #f0f0f0;
+        background-color: #000;
```
4ian, line 16: Also check in Electron main.js what color we define for the background, I'm not sure but setting the same might avoid a flash at the opening (can't remember if we wait or not for the page to be at least loaded before showing the window).

AlexandreSi: Good catch!

PR #3652, opened 2022-02-16
Show a warning message and a warning underline in the events sheet when a variable name looks wrong (quotes) or contains spaces

4ian:
See the reason for this here: https://github.com/4ian/GDevelop/discussions/2919#discussioncomment-2185923

image
image
image
image
image
image


D8H:
What if someone try to use an expression, for instance: `Variable(StatVariableName)`?

4ian:
@D8H I can maybe add a check if the variable name is containing parenthesis and warn again against a possible usage of an expression, which won't work here.

PR #3655, opened 2022-02-17
Add ability to select default layer for adding objects

mxsdev:
This allows the user to select the default layer objects are added to. This is implemented through a menu next to tag selection:

image

The layer button `newIDE/app/src/UI/EditorMosaic/LayerButton.js` is copied from `newIDE/app/src/UI/EditorMosaic/TagsButton.js`. Between these and the close button there's a bit of code duplication, though I don't think any kind of abstract "MosaicButton" class is necessary.

### DIFF
--- a/newIDE/app/src/InstancesEditor/InstancesAdder.js
+++ b/newIDE/app/src/InstancesEditor/InstancesAdder.js
@@ -40,15 +40,21 @@ export default class InstancesAdder {
   _temporaryInstances: Array<gdInitialInstance>;
   _instancesEditorSettings: InstancesEditorSettings;
   _zOrderFinder = new gd.HighestZOrderFinder();
+  _defaultLayer: string;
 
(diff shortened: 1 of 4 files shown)


4ian:
This would be a nice feature to have, but I wonder if this is the right place to put it at all?
The objects list is about editing objects, while here we add a button that looks like it could be used to filter objects, while it's actually changing the layer when you drag an object to the scene editor.

Instead, I wonder if we could make the currently "selected" layer into the layers list:
image

We could have a background color or icon set to the currently selected layer, and probably an indication somewhere (near the cursor position on the scene editor?) of the name of this layer. I've seen this done in some software and this sounds a bit more intuitive?

Bouh:
Yes, I think it would be better to select a layer in the layers panel.

AlexandreSi:
This feature has been introduced in #5046

PR #3656, opened 2022-02-17
Select parameter content when editing it form the event sheet

### LINKED ISSUE #1559: Text cursor not at the end of the value in case of num input
# Describe the bug
When a num type input is selected to change, the default cursor goes to the starting of the variable but I think it should be consistent with other type of data for example strings, where it goes to the end by default. 

# How to Reproduce
1. Open any project
2. Try to change a num value
3. See error as given below

String type input
image

Num type input
image

### LINKED ISSUE #2111: Text cursor position in the events sheet
When clicking on a field to edit it, the position of the text cursor is not consistent in all fields:
- Value/Text: left of the text. (with some exceptions, like tween easings)
image

- Name (object, variable, etc): right of the text. 
image

Names (and easings) are available through dropdown menus, so it might be related to that? Either way, I think it should be consistent.
Ideally, I would like the content of fields to be always highlighted when I click on the field. That way, whatever I want to do, I only need one keypress: left to go to the beginning, right to go to the end, backspace to erase (and access the dropdowns), or simply write the new name/text/value.

GD5b101

AlexandreSi:
Closes #1559
Closes #2111

Used @PascalLadalle suggestion: Select all content to quickly go to either the beginning or the end of the input.

demo_select_all_parameter

### DIFF
--- a/Extensions/Lighting/JsExtension.js
+++ b/Extensions/Lighting/JsExtension.js
@@ -243,7 +243,7 @@ module.exports = {
     objectsEditorService.registerEditorConfiguration(
       'Lighting::LightObject',
       objectsEditorService.getDefaultObjectJsImplementationPropertiesEditor({
-        helpPagePath: '/objects/light',
+        helpPagePath: '/all-features/lighting/reference',
       })
     );
   },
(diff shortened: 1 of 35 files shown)


AlexandreSi:
Closing because of bad branch manipulation

PR #3657, opened 2022-02-17
Select parameter content when editing it form the event sheet

### LINKED ISSUE #1559: Text cursor not at the end of the value in case of num input

### LINKED ISSUE #2111: Text cursor position in the events sheet

AlexandreSi:
Closes #1559
Closes #2111

Used @PascalLadalle suggestion: Select all content to quickly go to either the beginning or the end of the input.

demo_select_all_parameter

### DIFF
--- a/newIDE/app/src/EventsSheet/InlineParameterEditor.js
+++ b/newIDE/app/src/EventsSheet/InlineParameterEditor.js
@@ -100,7 +100,8 @@ export default class InlineParameterEditor extends React.Component<
       () => {
         // Give a bit of time for the popover to mount itself
         setTimeout(() => {
-          if (this._field && this._field.focus) this._field.focus(true);
+          if (this._field && this._field.focus)
+            this._field.focus(/*selectAll=*/ true);
         }, 10);
       }
     );
--- a/newIDE/app/src/EventsSheet/InlineParameterEditor.js
+++ b/newIDE/app/src/EventsSheet/InlineParameterEditor.js
@@ -100,7 +100,7 @@ export default class InlineParameterEditor extends React.Component<
       () => {
         // Give a bit of time for the popover to mount itself
         setTimeout(() => {
-          if (this._field && this._field.focus) this._field.focus();
+          if (this._field && this._field.focus) this._field.focus(true);
         }, 10);
       }
     );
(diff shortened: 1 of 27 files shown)


newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/index.js
```diff
@@ -179,9 +179,18 @@ export default class ExpressionField extends React.Component<Props, State> {
     }
   }
 
-  focus = () => {
+  focus = (selectAll: boolean = false) => {
     if (this._field) {
       this._field.focus();
+      if (selectAll) {
+        this._inputElement = this._field ? this._field.getInputNode() : null;
```
4ian, line 186: Is this line necessary?

newIDE/app/src/EventsSheet/InlineParameterEditor.js
```diff
@@ -100,7 +100,7 @@ export default class InlineParameterEditor extends React.Component<
       () => {
         // Give a bit of time for the popover to mount itself
         setTimeout(() => {
-          if (this._field && this._field.focus) this._field.focus();
+          if (this._field && this._field.focus) this._field.focus(true);
```
4ian, line 103: ubernitpicking: 
```suggestion
          if (this._field && this._field.focus) this._field.focus(/*selectAll=*/true);
```

4ian (approved):

newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/index.js
AlexandreSi: You're right, useless!

PR #3661, opened 2022-02-17
Move into a specific group the deprecated property of platformer object behavior

4ian:
Previously, this layout (reported by a French user) was making things look like the last 3 properties were deprecated - while they are not! 

image

Should be clearer now:
image

### DIFF
--- a/Extensions/PlatformBehavior/PlatformerObjectBehavior.cpp
+++ b/Extensions/PlatformBehavior/PlatformerObjectBehavior.cpp
@@ -8,14 +8,14 @@ This project is released under the MIT License.
 #include "PlatformerObjectBehavior.h"
 
 #include <iostream>
-#include <memory>
 #include <map>
+#include <memory>
 
-#include "GDCore/Tools/Localization.h"
 #include "GDCore/CommonTools.h"
 #include "GDCore/Project/Layout.h"
-#include "GDCore/Serialization/SerializerElement.h"
 #include "GDCore/Project/PropertyDescriptor.h"
+#include "GDCore/Serialization/SerializerElement.h"
+#include "GDCore/Tools/Localization.h"
 #include "PlatformBehavior.h"
 
 void PlatformerObjectBehavior::InitializeContent(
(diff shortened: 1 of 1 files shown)


Extensions/PlatformBehavior/PlatformerObjectBehavior.cpp
```diff
@@ -112,9 +122,10 @@ bool PlatformerObjectBehavior::UpdateProperty(
...
   else if (name == _("Can grab platform ledges"))
     behaviorContent.SetAttribute("canGrabPlatforms", (value == "1"));
-  else if (name == _("Automatically grab platform ledges without having to move horizontally"))
+  else if (name == _("Automatically grab platform ledges without having to "
+                     "move horizontally"))
     behaviorContent.SetAttribute("canGrabWithoutMoving", (value == "1"));
-  else if (name == _("Use frame per second dependent trajectories (deprecated)"))
+  else if (name == "useLegacyTrajectory")
     behaviorContent.SetAttribute("useLegacyTrajectory", (value == "1"));
   else if (name == _("Can go down from jumpthru platforms"))
```
D8H, line 130: Why this change?

Extensions/PlatformBehavior/PlatformerObjectBehavior.cpp
```diff
@@ -79,18 +82,25 @@ PlatformerObjectBehavior::GetProperties(
...
       .SetValue(behaviorContent.GetBoolAttribute("canGrabWithoutMoving", false)
                     ? "true"
                     : "false")
       .SetType("Boolean");
-  properties[_("Grab offset on Y axis")].SetGroup(_("Ledge")).SetValue(
-      gd::String::From(behaviorContent.GetDoubleAttribute("yGrabOffset")));
-  properties[_("Grab tolerance on X axis")].SetGroup(_("Ledge")).SetValue(gd::String::From(
-      behaviorContent.GetDoubleAttribute("xGrabTolerance", 10)));
-  properties[_("Use frame per second dependent trajectories (deprecated)")]
-      .SetGroup(_("Jump"))
```
D8H, line 93: I wonder if "frame rate dependent" could be better.

Extensions/PlatformBehavior/PlatformerObjectBehavior.cpp
4ian: Good practice to start using short identifier for property names, and labels for what to show to the user. Apart from clarity, this also ensure the label is only at one place in the code (I forgot to update the other and so the property update was broken without any warning).

PR #3662, opened 2022-02-17
Do not select sprite when using mouse middle click to pan in scene editor

### LINKED ISSUE #2278: Scene Editor - Do not clear selection when mouse wheel is pressed
## Description
When moving objects with the mouse across the scene, sometimes it is necessary to move around the scene; for this, use the mouse with the mouse wheel held down. This resets the selected objects on the scene.

## Solution suggested
Do not clear selection when mouse wheel is pressed.

AlexandreSi:
This PR changes 2 things:

- The first commit propagates right click events to pixi sprites, instead of having the pixi view catching it and handling it. It allows for example the following manipulation:
  - Select Player sprite (left click)
  - Right click on it
  - Edit Player's properties
  - Back to the scene, right click on Enemy sprite
  - Edit Enemy's properties
  => Before that, the context menu when right clicking on the Enemy was referring to the Player. I am not sure a user already reported that but it feels more natural to me.

- The second commit fixes #2278: When you multi select sprites, if you want to use the middle click to pan the scene to select more, if your mid click is on an object, it currently selects it and you lose your previous selection.

### DIFF
--- a/newIDE/app/src/InstancesEditor/InstancesRenderer/LayerRenderer.js
+++ b/newIDE/app/src/InstancesEditor/InstancesRenderer/LayerRenderer.js
@@ -22,6 +22,12 @@ export default class LayerRenderer {
   layer: gdLayer;
   viewPosition: ViewPosition;
   onInstanceClicked: gdInitialInstance => void;
+  onInstanceRightClicked: ({|
+    offsetX: number,
+    offsetY: number,
+    x: number,
+    y: number,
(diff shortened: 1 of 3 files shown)


4ian:
> Before that, the context menu when right clicking on the Enemy was referring to the Player. I am not sure a user already reported that but it feels more natural to me.

Totally, great catch. Verified in Figma and that's what they do.

PR #3668, opened 2022-02-18
Improve copy pasting instances

AlexandreSi:
- When pasting instances on the scene:
  - select the new instances instead of keeping the original instances selected (See https://forum.gdevelop-app.com/t/improve-usability-of-the-scene-editor/21515)
    demo_select_pasted

  - trigger history change
- Fix a bug on redoing with Ctrl+Shift+Z


AlexandreSi:
I kinda like this "pasted where the cursor is" behavior, but I don't know if my feeling is shared

tristanbob:
I agree with Davy.  The center of the copied objects should be at the location of the mouse.

Issue #3667, opened 2022-02-18
Jump Thru Platform Fall Glitch


MrNick666:
I've found a glitch where if you're jumping and your sprite hits a platform on it's edge (at the last point of your jump) you fall through the platform below it as if it's not there. This happens if two separate Jump Thru platforms are butted against each other too.
I can consistently reproduce this jumping in either direction. [Drop down is set to off and grab ledges is set to off].

https://media.giphy.com/media/4FhNZ3PGOPVMV5P6na/giphy.gif

I've uploaded a quick video illustration and this test project.

https://drop.devloprog.org/r/3YsG6XY7rH#WR0c7Awd7HHcCIYv8Q4WqKultMdSgpwbCsAoeSzdK+c=

I'm using the latest release of GDevelop [GDevelop 5.0.127 based on GDevelop.js 4.0.99-0-release] (tested on three mobiles [all Android 11] and two PCs [Windows 10] and get the same results.

D8H:
I couldn't reproduce it.
Do you have the issue with the exact project you upload?

D8H:
I probably don't try the right thing.
Do you understand why I don't go though the floor watching this video (don't mind my 1st 2 attempts)?

https://user-images.githubusercontent.com/2611977/154743046-ff2191fa-a8fe-43ed-8bd4-08ff7874377f.mp4

I you don't see what I do wrong. I would suggest to use ShareX to recode the issue in an mp4 file as the gif has a low frame rate so it's hard do see what happen.

MrNick666:
I see it. It might be easier to make the moving sprite a box (I should have done that) as it will be easier to see the edges. You need to hit the floor at the same time as the edge of the sprite is parallel to the higher platform, if that makes sense.

Hopefully this gif might illustrate it better:

![Alt Text](https://media.giphy.com/media/hlq8wPkjh67iCSnZtn/giphy.gif)

D8H:
Oh ok, thanks for the explanation, I think I understand now.

MrNick666:
I'm not sure if you've been able to replicate it.

I've been seeing the point where it generally happens and, if I turn on grab ledges and enable them, it seems to be the point where the sprite would grab the edge of the platform - if that helps any?

D8H:
Thank you a lot for reporting this. It should be fixed in the next release.

AlexandreSi:
I'm closing it, we'll reopen if the issue is still not fixed!

PR #3672, opened 2022-02-18
[Platformer] Add a test about landing on jumpthru while overlapping another one.

D8H:
Add a test in attempt to reproduce this issue:
* https://github.com/4ian/GDevelop/issues/3667#issuecomment-1044747394


D8H:
Opened a PR for a fix instead:
https://github.com/4ian/GDevelop/pull/3677

PR #3673, opened 2022-02-18
Autocompletion for external layout name fields.

D8H:
It shows every external layouts as the scene asked when creating an external layout is only used for the object list and external layouts can be shared with several scenes.
There is no expression in my knowledge that has an external layout parameter so it's not handled to avoid dead code.

image


4ian (approved):
Looks good to me!

AlexandreSi (approved):

PR #3676, opened 2022-02-20
[Platformer] Fix: Default useLegacyTrajectory to true for old project files.

D8H:
This actually broke the compatibility. On old projects, the new trajectory was used by default (even if the checkbox was checked).

The work around is to uncheck and recheck the checkbox of useLegacyTrajectory to force a defined value.

I though that the c++ part was giving the default value but it's only for the view. Behaviors read from old project files are read to undefined.


4ian:
Ah good catch, I totally missed that too. It's true that it's a bit unexpected because behavior content data is not flowing through the extension in C++/JS if not touched.
This is actually a bit of a pain too when developing or adding new attributes (because default values can't be easily enforced).
We could maybe add something to `gd::Behavior` that would be used to "validate" the data of a behavior, so that the extension can force a value, remove old values, etc...

For now, your fix looks good 👍 The fact that nobody noticed may mean that the change was small enough to be not a problem at all :)

PR #3677, opened 2022-02-21
[Platformer] fix: jumpthrus no longer push the character below when colliding.

D8H:
Test and fix this issue:
* https://github.com/4ian/GDevelop/issues/3667#issuecomment-1044747394

### DIFF
--- a/Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
+++ b/Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
@@ -778,12 +778,16 @@ namespace gdjs {
         this._findPlatformHighestRelativeYUnderObject(platform, context);
         let highestRelativeY = context.getFloorDeltaY();
         if (
-          // When following the floor, ignore jumpthrus that are higher than the character bottom.
-          this._state === this._onFloor &&
-          platform !== this._onFloor.getFloorPlatform() &&
           platform.getPlatformType() ===
             gdjs.PlatformRuntimeBehavior.JUMPTHRU &&
-          highestRelativeY < 0
+          // When following the floor, ignore jumpthrus that are higher than the character bottom.
+          ((this._state === this._onFloor &&
+            platform !== this._onFloor.getFloorPlatform() &&
+            highestRelativeY < 0) ||
+            // A jumpthrus should never constrain a character to go below.
+            // Jumpthrus are considered as obstacles at the 1st frame they are overlapping the character
+            // because it allows it to land on them, but they shouldn't push on its head.
+            context.allowedMinDeltaY !== previousAllowedMinDeltaY)
         ) {
           // Don't follow jumpthrus that are higher than the character bottom.
           // Revert side effect on the search context.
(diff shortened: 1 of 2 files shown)


Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -778,12 +778,16 @@ namespace gdjs {
...
             gdjs.PlatformRuntimeBehavior.JUMPTHRU &&
-          highestRelativeY < 0
+          // When following the floor, ignore jumpthrus that are higher than the character bottom.
+          ((this._state === this._onFloor &&
+            platform !== this._onFloor.getFloorPlatform() &&
+            highestRelativeY < 0) ||
+            // A jumpthrus should never constrain a character to go below.
+            // Jumpthrus are considered as obstacles at the 1st frame they are overlapping the character
+            // because it allows it to land on them, but they shouldn't push on its head.
+            context.allowedMinDeltaY !== previousAllowedMinDeltaY)
```
4ian, line 790: Make sense. Can you tell me what are the reasons that would make allowedMinDeltaY to change? (just for me to fully understand what could cause this)

Extensions/PlatformBehavior/tests/JumpAndFallingPlatformer.spec.js
```diff
@@ -1343,9 +1342,98 @@ describe('gdjs.PlatformerObjectRuntimeBehavior', function () {
...
+        expect(object.getBehavior('auto1').isFallingWithoutJumping()).to.be(
+          true
+        );
+        expect(object.getBehavior('auto1').isMoving()).to.be(true);
+      }
+      // The character is right at the left of "jumpthru".
+      expect(object.getX()).to.be(0);
+      // The character is right above "bottomJumpthru".
+      expect(object.getY()).to.be.below(
+        bottomJumpthru.getY() - object.getHeight()
```
4ian, line 1414: Do you want to check if it's just between `bottomJumpthru.getY() - object.getHeight()` and something like `bottomJumpthru.getY() - object.getHeight() - 1` or `bottomJumpthru.getY() - object.getHeight() - 2`?

4ian (approved):

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
D8H: If a solid platform is on the top of the character, it can land to another platform if it's not squeezed between the 2.

If the obstacle were a solid platform the character would have been stopped before it when moving on X and it would not have been used for floor search. So this scenario can't happen with solid platforms unless it's a moving platform. In this case, the character is pushed in the ground and move back up when the moving platform goes away.

Extensions/PlatformBehavior/tests/JumpAndFallingPlatformer.spec.js
D8H: Yes, It will be safer.

4ian (approved):
LGTM!

PR #3682, opened 2022-02-21
Let users define if their game can be played with a keybord, a gamepad or a mobile

D8H:
It doesn't store the data in the project, but I'm not against it.

image

I tried to keep the same layout as in Liluo, but it makes a big empty space.

image

### DIFF
--- a/newIDE/app/src/GameDashboard/GameDetailsDialog.js
+++ b/newIDE/app/src/GameDashboard/GameDetailsDialog.js
@@ -37,6 +37,9 @@ import { type PublicGame } from '../Utils/GDevelopServices/Game';
 import PlaceholderLoader from '../UI/PlaceholderLoader';
 import PublicGamePropertiesDialog from '../ProjectManager/PublicGamePropertiesDialog';
 import TextField from '../UI/TextField';
+import KeyboardIcon from '@material-ui/icons/Keyboard';
+import SportsEsportsIcon from '@material-ui/icons/SportsEsports';
+import SmartphoneIcon from '@material-ui/icons/Smartphone';
 
 const styles = {
   tableRowStatColumn: {
@@ -146,6 +149,7 @@ export const GameDetailsDialog = ({
     const { id } = profile;
 
     try {
+      const { playWithKeyboard, playWithGamepad, playWithMobile } = publicGame;
       // Set public game to null as it will be refetched automatically by the callback above.
       setPublicGame(null);
       const gameId = project.getProjectUuid();
@@ -153,6 +157,9 @@ export const GameDetailsDialog = ({
         authorName: project.getAuthor() || 'Unspecified publisher',
         gameName: project.getName() || 'Untitle game',
         description: project.getDescription() || '',
+        playWithKeyboard,
+        playWithGamepad,
+        playWithMobile,
       });
       const authorAcls = getAclsFromAuthorIds(project.getAuthorIds());
       await setGameUserAcls(getAuthorizationHeader, id, gameId, authorAcls);
@@ -286,6 +293,11 @@ export const GameDetailsDialog = ({
                     </Trans>
                   </Text>
                 </Line>
+              </Line>
+              <Line expand justifyContent="flex-start" alignItems="center">
+                {publicGame.playWithKeyboard && <KeyboardIcon />}
+                {publicGame.playWithGamepad && <SportsEsportsIcon />}
+                {publicGame.playWithMobile && <SmartphoneIcon />}
               </Line>
               <TextField
                 value={publicGame.gameName}
--- a/newIDE/app/src/ProjectManager/PublicGameProperties.js
+++ b/newIDE/app/src/ProjectManager/PublicGameProperties.js
@@ -4,6 +4,7 @@ import { Trans } from '@lingui/macro';
 import SemiControlledTextField from '../UI/SemiControlledTextField';
 import { UsersAutocomplete } from '../Utils/UsersAutocomplete';
 import { ColumnStackLayout } from '../UI/Layout';
+import Checkbox from '../UI/Checkbox';
 
 type Props = {|
   project: gdProject,
@@ -13,6 +14,12 @@ type Props = {|
   description: ?string,
   setAuthorIds: (string[]) => void,
   authorIds: string[],
+  setPlayableWithKeyboard: boolean => void,
+  playWithKeyboard: boolean,
+  setPlayableWithGamepad: boolean => void,
+  playWithGamepad: boolean,
+  setPlayableWithMobile: boolean => void,
+  playWithMobile: boolean,
 |};
 
 function PublicGameProperties({
@@ -23,6 +30,12 @@ function PublicGameProperties({
   description,
   setAuthorIds,
   authorIds,
+  setPlayableWithKeyboard,
+  playWithKeyboard,
+  setPlayableWithGamepad,
+  playWithGamepad,
+  setPlayableWithMobile,
+  playWithMobile,
 }: Props) {
   return (
     <ColumnStackLayout noMargin>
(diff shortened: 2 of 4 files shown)


newIDE/app/src/ProjectManager/PublicGameProperties.js
```diff
@@ -56,6 +69,29 @@ function PublicGameProperties({
           </Trans>
         }
       />
+      {// This view is used for public game properties as well as project properties.
+      // The following properties are not shown in project properties.
+      setPlayableWithKeyboard &&
+        setPlayableWithGamepad &&
+        setPlayableWithMobile && (
+          <React.Fragment>
+            <Checkbox
```
D8H, line 78: This is not clean, but we will probably display the check boxes in both views at some point.

newIDE/app/src/ProjectManager/PublicGamePropertiesDialog.js
```diff
@@ -33,6 +36,24 @@ function applyPublicPropertiesToProject(
...
+  playWithKeyboard: boolean,
+  playWithGamepad: boolean,
+  playWithMobile: boolean,
+|};
+
+function applyPublicPropertiesToGame(
+  publicGame: PublicGame,
+  newProperties: PublicGameOnlyProperties
+) {
+  publicGame.playWithKeyboard = newProperties.playWithKeyboard;
```
4ian, line 52: It's a bit against the "React way" to mutate things in place, as in the future other components relying on the publicGame could not see the object was changed and don't re-render.
Instead, I would instanciate a new publicGame with the updated attributes:

`const updatedPublicGame = { ...publicGame, playWithKeyboard: ..., ...}`

And I would pass it to the parent inside the `onApply`.
Finally, in the `onApply={...}` I would call `updateGameFromProject` by passing it the updatedPublicGame:
`updateGameFromProjectAndPublicGame(updatedPublicGame)`

(for the project, it's a bit different because objects from libGD.js *are* mutated in place, and this is done almost everywhere).

And then it's the parent responsibility to update its public game, which it will do already because the game is updated.

---

OR we add the properties in the project now (even if not displayed in the project settings dialog).

D8H: I'll go for the 2nd option as it will avoid to revert it in a few days.

4ian: Good decision! :)

Core/GDCore/Project/Project.cpp
```diff
@@ -736,6 +751,18 @@ void Project::SerializeTo(SerializerElement& element) const {
...
+    playableDevicesElement.AddChild("").SetStringValue("keyboard");
+  }
+  if (isPlayableWithGamepad) {
+    playableDevicesElement.AddChild("").SetStringValue("gamepad");
+  }
+  if (isPlayableWithMobile) {
+    playableDevicesElement.AddChild("").SetStringValue("mobile");
+  }
+
   // Compatibility with GD <= 5.0.0-beta101
```
4ian, line 766: You also need to add the 3 booleans to Project::Init (to make sure they are copied when a copy in C++ is done - ask me if the reason for this method to exist is unclear to you and I or Alexandre/Clement can explain)

4ian (approved):

Bouh:
Could be less wordy by just adding a second title?
`Compatible devices` sound better I think.
image

4ian:
Just pointing out that this is *not* the compatible devices, this is what the inputs can be used to play to the game.

Silver-Streak:
I think this is a really good idea, but as mentioned on the discord: Should gamepads be listed if the gamepad extension isn't a native extension at this point?

(alternatively, should the gamepad extension be migrated to a native extension?)

ClementPasteau:
I've seen more and more games supporting gamepads, so in my opinion, it makes sense to make it a thing as big as "keyboard/mouse" and "mobile".
I like the idea of migrating it to a native extension too, it would make it available for all games and we can ensure it is super robust.

newIDE/app/src/ProjectManager/PublicGameProperties.js
```diff
@@ -56,6 +69,29 @@ function PublicGameProperties({
...
+              label={<Trans>Playable with keyboards</Trans>}
+              checked={playWithKeyboard || false}
```
ClementPasteau, line 80: you can transform these with `{!!playWithKeyboard}` (creates a boolean)

newIDE/app/src/ProjectManager/PublicGameProperties.js
ClementPasteau: It's a good remark, I thought it'd be a good thing to only have 1 component because the properties made sense to be in both places.
But we may move towards not sharing them in the future and have different components alltogether so we can control the order the things appear

newIDE/app/src/ProjectManager/PublicGamePropertiesDialog.js
```diff
@@ -64,6 +83,9 @@ const PublicGamePropertiesDialog = ({
         name,
         description: description || '',
         authorIds,
+        playWithKeyboard: playWithKeyboard || false,
```
ClementPasteau, line 86: same here, you can probably just do `!!playWithKeyboard`

Core/GDCore/Project/Project.cpp
D8H: Should I also set a default value in the constructor?
Not every attributes are initialized and I'm not sure how to choose.

newIDE/app/src/GameDashboard/GameDetailsDialog.js
```diff
@@ -287,6 +293,11 @@ export const GameDetailsDialog = ({
                   </Text>
                 </Line>
               </Line>
+              <Line expand justifyContent="flex-start" alignItems="center">
```
ClementPasteau, line 296: (are we sure this doesn't create a weird empty line when none are true?)

newIDE/app/src/ProjectManager/PublicGameProperties.js
```diff
@@ -56,6 +69,29 @@ function PublicGameProperties({
...
         }
       />
+      {// This view is used for public game properties as well as project properties.
+      // The following properties are not shown in project properties.
+      setPlayableWithKeyboard &&
+        setPlayableWithGamepad &&
+        setPlayableWithMobile && (
+          <React.Fragment>
+            <Checkbox
+              label={<Trans>Playable with keyboards</Trans>}
```
ClementPasteau, line 79: `playable with keyboard` without **s** ?

newIDE/app/src/ProjectManager/PublicGameProperties.js
```diff
@@ -56,6 +69,29 @@ function PublicGameProperties({
...
+              checked={!!playWithKeyboard}
+              onCheck={(e, checked) => setPlayableWithKeyboard(checked)}
+            />
+            <Checkbox
+              label={<Trans>Playable with gamepads</Trans>}
```
ClementPasteau, line 84: same here, without s? (yes multiple gamepads can be used, but keeping the generic word "gamepad" should mean that multiple can be used)

newIDE/app/src/GameDashboard/GameDetailsDialog.js
D8H: It adds a bit of space.
image

newIDE/app/src/ProjectManager/PublicGameProperties.js
D8H: Grammatically should it be either "Playable with keyboards" or "Playable with a keyboard"?

ClementPasteau: I feel like we can say "playable with keyboard", but if not I'd got for "playable with a keyboard"

4ian (approved):
Good enough for me, I'll let you iron out the last comments and merge :)

newIDE/app/src/ProjectManager/PublicGameProperties.js
D8H: Maybe "Keyboard playable"?

ClementPasteau: doesn't sound great, let's go for "Playable with a keyboard" + "Playable with a gamepad" :)

ClementPasteau (approved):
last wording fix, then let's merge!

PR #3690, opened 2022-02-24
Let game creators define an orientation on mobile for Liluo.io.

D8H:
It moves the orientation combobox from project property view to public game property view and send it to the game-api.


D8H:
Wrong branch

PR #3691, opened 2022-02-24
Let game creators define an orientation on mobile for Liluo.io.

D8H:
It moves the orientation combobox from project property view to public game property view and send it to the game-api.

image

image

image

### DIFF
--- a/newIDE/app/src/GameDashboard/GameDetailsDialog.js
+++ b/newIDE/app/src/GameDashboard/GameDetailsDialog.js
@@ -324,7 +324,7 @@ export const GameDetailsDialog = ({
                 disabled
                 fullWidth
                 floatingLabelText={
-                  <Trans>Device orientation (for iOS and Android)</Trans>
+                  <Trans>Device orientation (for mobile)</Trans>
                 }
                 value={publicGame.orientation}
               >
--- a/newIDE/app/src/GameDashboard/GameDetailsDialog.js
+++ b/newIDE/app/src/GameDashboard/GameDetailsDialog.js
@@ -159,6 +159,7 @@ export const GameDetailsDialog = ({
         playWithKeyboard: project.isPlayableWithKeyboard(),
         playWithGamepad: project.isPlayableWithGamepad(),
         playWithMobile: project.isPlayableWithMobile(),
+        orientation: project.getOrientation(),
       });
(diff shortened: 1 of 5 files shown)


newIDE/app/src/GameDashboard/GameDetailsDialog.js
```diff
@@ -319,6 +320,21 @@ export const GameDetailsDialog = ({
                 multiline
                 rows={5}
               />
+              <SelectField
+                disabled
+                fullWidth
+                floatingLabelText={
+                  <Trans>Device orientation (for iOS and Android)</Trans>
```
ClementPasteau, line 327: This can be renamed to show this is not only ios/android. Maybe `Device orientation (for mobile)`

newIDE/app/src/ProjectManager/PublicGameProperties.js
```diff
@@ -69,6 +76,18 @@ function PublicGameProperties({
           </Trans>
         }
       />
+      <SelectField
+        fullWidth
+        floatingLabelText={
+          <Trans>Device orientation (for iOS and Android)</Trans>
```
ClementPasteau, line 82: we could rename it here too?

newIDE/app/src/GameDashboard/GameDetailsDialog.js
```diff
@@ -319,6 +320,21 @@ export const GameDetailsDialog = ({
                 multiline
                 rows={5}
               />
+              <SelectField
```
ClementPasteau, line 323: Does the disabled SelectField look OK?
I'm wondering if we should just display a disabled Textfield in this case (with a mapping from the value to a readable text)

ClementPasteau (approved):
A few comms, otherwise lgtm!

newIDE/app/src/GameDashboard/GameDetailsDialog.js
4ian: I would also add "for Liluo.io" maybe too? What do you think?

D8H: The project properties view rework will probably make it clear. As every field is actually for Liluo.io.

4ian: Ok!

PR #3703, opened 2022-02-27
Split the "Center the camera on an object within limits" in 2 actions

D8H:
This replace the "Center the camera on an object within limits" by the "Move back the camera within bounds" action. This new action can be used after the "Center the camera on an object" action to get the same effect.

This allow to constraint camera within bounds after any camera movements:
* Drag camera
* Zoom camera
* Smooth following
* ...

Tested on the new platformer example:
* https://www.dropbox.com/s/3hzpx7zemmnu2bi/Platformer_REVIEW.zip?dl=1

### DIFF
--- a/Core/GDCore/Extensions/Builtin/CameraExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/CameraExtension.cpp
@@ -266,6 +266,32 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsCameraExtension(
       .SetDefaultValue("\"\"")
       .AddParameter("expression", _("Camera number (default : 0)"), "", true)
       .SetDefaultValue("0")
+      .MarkAsAdvanced()
+      .SetHidden();  // Deprecated
+
+  extension
+      .AddAction(
+          "ClampCamera",
+          _("Move back the camera within bounds"),
+          _("Move back the camera inside the specified bounds."),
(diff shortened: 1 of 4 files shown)


tristanbob:
This is one of the camera improvements we need, thank you!  I recommend changing the wording to:

```
Name: "Enforce camera boundaries"
Description: "Enforce camera boundaries by moving the camera back inside specified boundaries"
Sentence: "Enforce camera boundaries (Left: _PARAM1_,  Right: _PARAM3_, Top: _PARAM2_,  Bottom: _PARAM4_, Layer: _PARAM5_)
```
(You can include the Camera number in the sentence if you want, but I believe Bouh and I have been leaving it out)

AlexandreSi:
I'm having troubles finding out what this PR changes. I don't know what to look at in the project you joined 😅 .
I can't find the sentence "Move back the camera within bounds"

D8H:
When game creators write a smooth camera, they will handle the bounds check themselves because they can't use the feature that is tie to the non-smooth camera following action.

So they will do something like this:

image

Whereas with the split suggested by this PR, they could do:

image

4ian:
I think that sounds good to have this new action.

For the old one, could we do a slightly smoother deprecation:
- We for now just add a "hint" message: "Please avoid using this and prefer the new action "Move back the camera within bounds", which is more flexible" on the old action.
- We hide the old action in a next version.

I'm a bit afraid of confusion for people not finding the old action.

4ian (approved):
👍

Issue #3135, opened 2021-10-04
I also need the links for the driver app changed on the . That driver app is no longer receiving jobs. I need potential drivers to call us if they want to drive.  So instead of saying "SUBMIT YOUR APPLICATION TODAY!" , need it to say "CONTACT US TODAY!" with the 4049026442 phone number and support@mule2gopro.com email under


yiniam:
## Describe the bug

Please double check that the bug is not already reported in the issues list.

## To Reproduce

* Please include a link to a game if possible!
* If applicable, add screenshots to help explain your problem.

## Other details
* Include any OS/browser version/smartphone that you're using
* Which version of GDevelop are you using? The desktop app or the web-app?
* Add any other context about the problem here.

PR #3705, opened 2022-02-28
Polish Platformer Properties

Elairyx7301:
- Optimized order of parameters
- ~~Unified all properties from ["xy"] to [_("xy")]~~
(I now see why it was like that in the first place)
- Change "Can grab ledge" default

### DIFF
--- a/Extensions/PlatformBehavior/PlatformerObjectBehavior.cpp
+++ b/Extensions/PlatformBehavior/PlatformerObjectBehavior.cpp
@@ -19,105 +19,117 @@ This project is released under the MIT License.
 #include "PlatformBehavior.h"
 
 void PlatformerObjectBehavior::InitializeContent(
-	gd::SerializerElement& behaviorContent) {
-	
-	behaviorContent.SetAttribute("ignoreDefaultControls", false);
-	
-	behaviorContent.SetAttribute("gravity", 1000);
-	behaviorContent.SetAttribute("jumpSpeed", 600);
-	behaviorContent.SetAttribute("jumpSustainTime", 0.2);
-	behaviorContent.SetAttribute("maxFallingSpeed", 700);
-	behaviorContent.SetAttribute("useLegacyTrajectory", false);
-	
-	behaviorContent.SetAttribute("ladderClimbingSpeed", 150);
-	
-	behaviorContent.SetAttribute("canGrabPlatforms", true);
-	behaviorContent.SetAttribute("canGrabWithoutMoving", true);
-	behaviorContent.SetAttribute("xGrabTolerance", 10);
-	behaviorContent.SetAttribute("yGrabOffset", 0);
-	
-	behaviorContent.SetAttribute("canGoDownFromJumpthru", true);
-	behaviorContent.SetAttribute("acceleration", 1500);
-	behaviorContent.SetAttribute("deceleration", 1500);
-	behaviorContent.SetAttribute("maxSpeed", 250);
-	behaviorContent.SetAttribute("slopeMaxAngle", 60);
+    gd::SerializerElement &behaviorContent) {
+
+  behaviorContent.SetAttribute("ignoreDefaultControls", false);
+
+  behaviorContent.SetAttribute("gravity", 1000);
+  behaviorContent.SetAttribute("jumpSpeed", 600);
+  behaviorContent.SetAttribute("jumpSustainTime", 0.2);
+  behaviorContent.SetAttribute("maxFallingSpeed", 700);
+  behaviorContent.SetAttribute("useLegacyTrajectory", false);
+
+  behaviorContent.SetAttribute("ladderClimbingSpeed", 150);
+
+  behaviorContent.SetAttribute("canGrabPlatforms", true);
+  behaviorContent.SetAttribute("canGrabWithoutMoving", true);
+  behaviorContent.SetAttribute("xGrabTolerance", 10);
+  behaviorContent.SetAttribute("yGrabOffset", 0);
+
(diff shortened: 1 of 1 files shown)


Extensions/PlatformBehavior/PlatformerObjectBehavior.cpp
```diff
@@ -19,99 +19,99 @@ This project is released under the MIT License.
...
+      .SetValue(behaviorContent.GetBoolAttribute("ignoreDefaultControls") ? "false" : "true")
+      .SetType("Boolean");
+
+    properties[_("Gravity")]
+      .SetGroup(_("Jump"))
+      .SetValue(gd::String::From(behaviorContent.GetDoubleAttribute("gravity")));
+    properties[_("Jump speed")]
+      .SetGroup(_("Jump"))
+      .SetValue(gd::String::From(behaviorContent.GetDoubleAttribute("jumpSpeed")));
+    properties[_("jumpSustainTime")]
```
4ian, line 63: Unfortunately that should be the reverse:

Property internal names should not be translated. `SetLabel(_("Translatable name"))` should be used instead.

4ian: (otherwise, you ask translators on Crowdin to translate "jumpSustainTime", and this will be desynchronized with `UpdateProperty` method that was not updated)

Extensions/PlatformBehavior/PlatformerObjectBehavior.cpp
```diff
@@ -19,99 +19,99 @@ This project is released under the MIT License.
...
-  std::map<gd::String, gd::PropertyDescriptor> properties;
-
-  properties[_("Gravity")].SetGroup(_("Jump")).SetValue(
-      gd::String::From(behaviorContent.GetDoubleAttribute("gravity")));
-  properties[_("Jump speed")].SetGroup(_("Jump")).SetValue(
-      gd::String::From(behaviorContent.GetDoubleAttribute("jumpSpeed")));
-  properties["jumpSustainTime"]
+#
+if defined(GD_IDE_ONLY)
+std::map < gd::String, gd::PropertyDescriptor >
```
4ian, line 48: Please run clang-format or the formatter from VS Code on this file because the code style of the file is not followed :) Autoformatting should always be used when you can.

Elairyx7301: @4ian Thank you for the review. I apologize for all the mistakes.
I'm new to working with the code and I just dove into it. I don't think I understood your first comment, sorry. What do you mean by `SetLabel(_("Translatable name")) should be used instead.`? What should I use `SetLabel` instead of?
I will fix these issues soon! :)

Elairyx7301:
I have now made the changes in a new commit. Please let me know if I did anything wrong and I will fix it asap 😄

Extensions/PlatformBehavior/PlatformerObjectBehavior.cpp
```diff
@@ -19,105 +19,117 @@ This project is released under the MIT License.
...
   behaviorContent.SetAttribute("slopeMaxAngle", 60);
-  behaviorContent.SetAttribute("canGrabPlatforms", false);
-  behaviorContent.SetAttribute("canGrabWithoutMoving", true);
-  behaviorContent.SetAttribute("yGrabOffset", 0);
-  behaviorContent.SetAttribute("xGrabTolerance", 10);
-  behaviorContent.SetAttribute("useLegacyTrajectory", false);
-  behaviorContent.SetAttribute("canGoDownFromJumpthru", true);
 }
 
-#if defined(GD_IDE_ONLY)
```
AlexandreSi, line 41: This line should be removed entirely, it has no use today

AlexandreSi: And the `#endif` at the last line as well

Elairyx7301:
Thank you, changes are done.
There is the `properties["useLegacyTrajectory"]` property. The label says "(deprecated)" and the description tells to "let this unchecked". Can this part also be removed since the user is explicitly advised not to use this or is this a advanced/old experimental feature that wants to be left in the code?
-> line 30, line 71-78 in 18b0ddb100f3dfca8c7f86fa370bff0a0730a1a6

AlexandreSi:
Hi @Elairyx7301 !
I realize I don't know why you opened this PR! Could you point out the forum post or the Discord discussion about this?
Or maybe just what you want to improve or fix with this?

AlexandreSi:
Ok I see, that's true it could be better ordered!
As discussed in here: https://github.com/4ian/GDevelop/discussions/3502, parameters are ordered alphabetically, so your changes won't change a thing unfortunately.

There are multiple options, but more complicated than ordering properties:
- Implement that ordering of parameters when defining them
- Add possibility to group parameters in a row (so that you can have Acceleration and Deceleration side by side)
  - It is currently implemented in the generic PropertiesEditor but not in the properties of a behavior
- Build a custom properties editor for the platformer character behavior (as done for the Physics2 behavior with file https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/index.js)

Elairyx7301:
Oh, I see. `Implement that ordering of parameters when defining them` sounds like the best method to quickly change the order of parameters in the first place to me but that's a task I can't take (yet).
Thank you for pointing me to the other issue about it :)
I guess I will close this PR then?

Elairyx7301:
Thank you everyone for the reviews and hints on what I had to improve in my doings helping me understand the background of this engine a bit better. I think the best way out is to make **#3135** happen. With that, ordering parameters won't be a problem now or in the future anymore.

PR #3707, opened 2022-02-28
Clarify separation between object and instance in instance properties panel

AlexandreSi:
demo_instance_properties_panel_separated


4ian (approved):
Is it worth using React.useCallback for the methods?
Apart from that, looks good!

AlexandreSi:
> Is it worth using React.useCallback for the methods?

I'll have a look, lots of methods are using lots of props

PR #3708, opened 2022-02-28
Set the button to add a behavior at the center when there is no behaior attached to an object.

D8H:
image

### DIFF
--- a/newIDE/app/src/BehaviorsEditor/EmptyBehaviorsPlaceholder.js
+++ b/newIDE/app/src/BehaviorsEditor/EmptyBehaviorsPlaceholder.js
@@ -21,13 +21,18 @@ const EmptyBehaviorsPlaceholder = (props: Props) => (
       />
     )}
     buttonJustification="center"
+    outlined={false}
   >
-    <Text size="title">
-      <Trans>Add your first behavior</Trans>
-    </Text>
-    <Text>
-      <Trans>Behaviors add features to objects in a matter of clicks.</Trans>
-    </Text>
+    <Line justifyContent="center" expand>
+      <Text size="title">
+        <Trans>Add your first behavior</Trans>
+      </Text>
+    </Line>
+    <Line justifyContent="center" expand>
+      <Text>
+        <Trans>Behaviors add features to objects in a matter of clicks.</Trans>
+      </Text>
+    </Line>
     <Line justifyContent="center" expand>
       <RaisedButton
         key="add-behavior-line"
--- a/newIDE/app/src/BehaviorsEditor/EmptyBehaviorsPlaceholder.js
+++ b/newIDE/app/src/BehaviorsEditor/EmptyBehaviorsPlaceholder.js
@@ -4,25 +4,40 @@ import * as React from 'react';
 import HelpButton from '../UI/HelpButton';
 import Text from '../UI/Text';
 import { EmptyPlaceholder } from '../UI/EmptyPlaceholder';
+import Add from '@material-ui/icons/Add';
+import { Column, Line } from '../UI/Grid';
+import RaisedButton from '../UI/RaisedButton';
 
-const EmptyBehaviorsPlaceholder = () => (
+type Props = {
+  openNewBehaviorDialog: () => void,
+};
+
+const EmptyBehaviorsPlaceholder = (props: Props) => (
   <EmptyPlaceholder
-    renderButtons={() => <HelpButton helpPagePath="/behaviors" />}
+    renderButtons={() => (
+      <HelpButton
+        label={<Trans>Read the doc</Trans>}
+        helpPagePath="/behaviors"
+      />
+    )}
+    buttonJustification="center"
   >
-    <Text>
-      <Trans>There are no behaviors here.</Trans>
+    <Text size="title">
+      <Trans>Add your first behavior</Trans>
     </Text>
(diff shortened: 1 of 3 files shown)


newIDE/app/src/BehaviorsEditor/EmptyBehaviorsPlaceholder.js
```diff
@@ -4,25 +4,41 @@ import * as React from 'react';
...
-      </Trans>
+    <Text align="center">
+      <Trans>Behaviors add features to objects in a matter of clicks.</Trans>
     </Text>
+    <Line justifyContent="center" expand>
+      <RaisedButton
+        key="add-behavior-line"
+        label={<Trans>Add a behavior</Trans>}
+        primary
+        onClick={() => props.openNewBehaviorDialog()}
```
ClementPasteau, line 37: this can probably be just `onClick={props.openNewBehaviorDialog}`

ClementPasteau (commented):
looks good!
How does it look once you've added the first behavior?

newIDE/app/src/BehaviorsEditor/EmptyBehaviorsPlaceholder.js
D8H: Ok, I made the change.

When a behavior is added, the view is the same as before. The "add" button is on bottom right.

4ian:
I think we discuss alternative sentences for effects and variables (including scene/global variables), I would be happy if we can integrate these too and merge this :)

D8H:
> I think we discuss alternative sentences for effects and variables (including scene/global variables), I would be happy if we can integrate these too and merge this :)

For effects, you suggested this and I think it's fine.
> Effects create visual changes to the object.

For variables, you suggested this
> Variables are number, text or data stored inside each instance of the object.
> 
> We can also use a metaphor:
> "A variable is a memory, like a text or number, stored in each instance of an object" (to give the "each" an importance, 1 object but multiple instances).
> "A scene variable is a memory, like a text or number."
> "A global variable is a memory, like a text or number, available in the whole game."

These sentences are still too long for a centered text like this one.
What about "Variables add custom information to an object/scene/project"?

4ian:
> These sentences are still too long for a centered text like this one.

Sure, then either "Variables add custom information to an [object/scene/project]"
or "Variables allow to remember data for an [object/scene/project]"
or "Variables store data in memory for an [object/scene/project]"

?

D8H:
> > These sentences are still too long for a centered text like this one.
> 
> Sure, then either "Variables add custom information to an [object/scene/project]" or "Variables allow to remember data for an [object/scene/project]" or "Variables store data in memory for an [object/scene/project]"
> 
> ?

I'm fine with the 3. "remember" might be human connoted, "store" game save connoted and "add" maybe too much neutral.
I have a preference for "add".

4ian:
Last try then! 
"Variables keep data in memory for ..."

keep = neutral
but memory = something, well, in memory, but not "on disk"/storage so you understand it's ephemeral?

D8H:
> Last try then! "Variables keep data in memory for ..."
> 
> keep = neutral but memory = something, well, in memory, but not "on disk"/storage so you understand it's ephemeral?

The issue with the term "memory is that it's neutral (e.g. memory card) and force to use technical adjectives to give meaning. I guess, we don't want to use "Volatile memory".

4ian:
> I guess, we don't want to use "volatile memory".

Yep, too complex. By memory I try to refer to the human centered view of "your memory", "I memorise something", "I remember something", which is well, a bit what a variable is doing during the lifetime of the object/scene.
Let's see if someone else can find something better!

D8H:
Maybe "Variables keep additional information on an object/scene/project"?

D8H:
I updated the screenshots in the description. I wonder if we should make the variable scope explicit in the titles.

4ian:
> . I wonder if we should make the variable scope explicit in the titles.

I think we should! 👍

4ian (approved):
LGTM once we found a way to make the "Add your first event" background the same as the events background (can we either get rid of the `Paper`? Or make it transparent? With a preference for getting rid of it).

ClementPasteau:
looks great! 👌

D8H:
I added the same placeholder for animation too, but the description I chose is a bit dub and the object name field make it shifted from other tabs.

The placeholder will look better if it's center vertically in its negative space and not its technical parent, but I'm not sure how to do this properly.

AlexandreSi:
I would say to:
- Add a point add the end of the sentence `Events define the rules of a game.`
- For animations: `Animations are a sequence of images that bring life to a sprite.`

4ian:
> The placeholders will look better if they are centered vertically in their negative space instead of their technical parent, but I'm not sure how to do this properly.

That would look better but I'm indeed not sure how to do this negative margin properly (you can put a ref somewhere and use https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect to measure the size of the input, but well that's starting to be dirty/not performant and you'll risk seeing flickering).
Let's maybe just avoid this because we might in the future remove the object name or move it somewhere else (like in the dialog title above tabs, something like this!)

4ian:
"Animations are a sequence of images that bring life to a sprite." => can even be: "Animations are a sequence of images."?

PR #3710, opened 2022-03-01
Search adjustments for instructions

arthuro555:
- Enable weighting based on object size
  - Otherwise, instructions with small names like "Or" will get obscured by other instructions that are seen as better matches as they match the pattern multiple times, for example: 
image
- Increase the weight of the name of the instruction, as that name is more natural, while the group will be less commonly used and shared with too many instructions to give good results.


AlexandreSi (approved):
There was a moment when this ignoreFieldNorm parameter was useful but it seems to work better now with your suggestion, so thank you!

PR #3712, opened 2022-03-01
Onboarding optimisation

ClementPasteau:
The goal is to avoid triggering a session for Userflow, even when the user decide not to continue with the onboarding flow.

(included a refacto of the Changelog Dialog to be a functional component)

Screenshot 2022-03-01 at 12 13 58


newIDE/app/src/MainFrame/Onboarding/OnboardingDialog.js
```diff
@@ -0,0 +1,109 @@
...
+We highly recommend it!
+`;
+
+const OnboardingDialog = () => {
+  const [open, setOpen] = React.useState(false);
+  const [loading, setLoading] = React.useState(false);
+
+  // Initialise userflow when modal opens.
+  React.useEffect(() => {
+    if (isDev) {
```
4ian, line 37: This will activate UserFlow for 100% of the people seeing this dialog. Instead, let's init userflow only when the tour is really started (`startUserflow` method).

newIDE/app/src/MainFrame/Onboarding/OnboardingDialog.js
```diff
@@ -0,0 +1,109 @@
...
+      }
+    );
+  }, []);
+
+  const startUserflow = React.useCallback(async () => {
+    setLoading(true);
+    const userPreferences = loadPreferencesFromLocalStorage();
+    const appLanguage = userPreferences ? userPreferences.language : undefined;
+    await userflow.identify(getUserUUID(), { language: appLanguage });
+    await userflow.start('b1611206-2fae-41ac-b08c-0f8ad72d8c39');
```
4ian, line 63: Do we need to catch and set loading to false to give the user a chance to retry?

4ian (changes requested):
Let's ensure we don't initialize userflow for all users, only for those who actually start the tour :)

newIDE/app/src/MainFrame/Onboarding/OnboardingDialog.js
ClementPasteau: I wasn't sure which method was considered as a new user for userfow.
But let's be safe and not do anything until we're sure we want to show the flow to the user.

newIDE/app/src/MainFrame/Onboarding/OnboardingDialog.js
ClementPasteau: I'll keep it really simple then, re-enable the button and that's it.
Because we don't have much control over such an error.

ClementPasteau:
updated to initialise only when the user clicks on continue 👍

newIDE/app/src/MainFrame/Onboarding/OnboardingDialog.js
```diff
@@ -0,0 +1,122 @@
...
+(🇬🇧 The tour is only available in English)  
+We highly recommend it!
+`;
+
+const OnboardingDialog = () => {
+  const [open, setOpen] = React.useState(false);
+  const [loading, setLoading] = React.useState(false);
+
+  React.useEffect(() => {
+    userflow.on(
```
4ian, line 37: Let's hope this work even if init was not called? Might be worth checking that or moving that inside startUserflow?

Also, this global callback on userflow will be added every time this dialog is shown.
Sure for now it's only once, but as soon as we add a button "Follow the tour" in the homepage that will show this dialog, we start to add more and more `on('flowvisibilitychange'` and that seems risky.

I would do two things (maybe confirm with Alex):
- only update the loading and open state in `startUserflow`.
- so you keep only `isUserflowRunning = isRunning` inside the callback. If you "leak" the callback by creating too much of them, it's not a huge problem but still dirty and a receipt for future problems if someone adds additional logic. So you can add a global boolean "isUserflowEventListenerSetUp" to false by default, and run this `userflow.on` only when the `isUserflowEventListenerSetUp`is false, immediately setting it to `true` after. 

If you do this after the `init`, then we are:
- feeling safer (we don't call any method before userflow.init... I'm always super suspicious about the capacity of 3rd party libraries to be coded well enough to be robust against this).
- don't create a leak of listeners if we show the dialog multiple times.

4ian (approved):
LGTM after we've made the implementation bulletproof by removing the effect to move the `on` after the `init`.

newIDE/app/src/MainFrame/Onboarding/OnboardingDialog.js
ClementPasteau: Agree we can be more robust.
I tested it locally and the `userflow.on()` was correctly registered, even if the init happened afterwards (the modal closed + the flow started correctly)

But I'll improve that

newIDE/app/src/MainFrame/Onboarding/OnboardingDialog.js
```diff
@@ -0,0 +1,129 @@
...
+      userflow.on(
+        // Undocumented legacy userflow event that is fired
+        // "when a flow either becomes active or removed"
+        // (tip given by a tech member of Userflow - it shouldn't be removed
+        // in the near future given the fact that some of their users still use it).
+        'flowvisibilitychange',
+        isRunning => {
+          isUserflowRunning = isRunning;
+        }
+      );
```
4ian, line 60: Still consider putting this into a "initialiseUserFlow" method that will automatically check with a boolean `isUserflowInitialised` (set to false originally) if the init must be done again, and if not, it can be skipped.

This will make this startUserflow method simpler to read, and later if we start other flows we can just call this method :)

4ian (approved):
Looks good after factoring the initialisation and listener set up in a method so that the method can only do the init one, just to be extra safe.

PR #3714, opened 2022-03-01
Make context menus in scene editor and in object list consistent

AlexandreSi:
- Right-click on instance:

  image

- Right-click on background (even if instance or object is selected):

  image

- Right-click on object:
  - Action `Add instance to the scene` adds an instance at the center of the scene

  image

- Duplicate action on instances adds duplicates on the lower right side, at the distance of 2 shift+arrow moves:
   
  image

### DIFF
--- a/newIDE/app/src/InstancesEditor/index.js
+++ b/newIDE/app/src/InstancesEditor/index.js
@@ -56,7 +56,11 @@ export type InstancesEditorPropsWithoutSizeAndScroll = {|
   onInstancesResized: (instances: Array<gdInitialInstance>) => void,
   onInstancesRotated: (instances: Array<gdInitialInstance>) => void,
   selectedObjectNames: Array<string>,
-  onContextMenu: (x: number, y: number) => void,
+  onContextMenu: (
+    x: number,
+    y: number,
+    ignoreSelectedObjectNamesForContextMenu?: boolean
+  ) => void,
   onCopy: () => void,
   onCut: () => void,
   onPaste: () => void,
@@ -208,6 +212,7 @@ export default class InstancesEditor extends Component<Props> {
           offsetY: event.offsetY,
           x: event.clientX,
           y: event.clientY,
+          ignoreSelectedObjectNamesForContextMenu: true,
         });
 
         return false;
@@ -337,7 +342,7 @@ export default class InstancesEditor extends Component<Props> {
       onDownInstance: this._onDownInstance,
       onOutInstance: this._onOutInstance,
       onInstanceClicked: this._onInstanceClicked,
-      onInstanceRightClicked: this._onRightClicked,
+      onInstanceRightClicked: this._onRightClickedOnInstance,
       onInstanceDoubleClicked: this._onInstanceDoubleClicked,
     });
     this.selectionRectangle = new SelectionRectangle({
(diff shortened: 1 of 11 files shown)


newIDE/app/src/SceneEditor/index.js
```diff
@@ -850,8 +863,127 @@ export default class SceneEditor extends React.Component<Props, State> {
...
-    if (this.contextMenu) this.contextMenu.open(x, y);
+  _onContextMenu = (
+    x: number,
+    y: number,
+    ignoreSelectedObjectNamesForContextMenu?: boolean = false
+  ) => {
+    if (!ignoreSelectedObjectNamesForContextMenu) {
+      if (this.contextMenu) this.contextMenu.open(x, y);
+    } else {
+      this.setState({ ignoreSelectedObjectNamesForContextMenu: true }, () => {
```
4ian, line 874: a bit "dirty", I wonder if we should allow some arbitrary parameters to be passed as third option to `contextMenu.open` that would be then passed to buildMenuTemplate? :)

newIDE/app/src/SceneEditor/index.js
```diff
@@ -876,6 +1008,33 @@ export default class SceneEditor extends React.Component<Props, State> {
...
+
+    if (!this.editor) return;
+
+    const newInstances = serializedSelection
+      .map(serializedInstance => {
+        const instance = new gd.InitialInstance();
+        unserializeFromJSObject(instance, serializedInstance);
+        return instance;
+      })
+      .map(instance => {
```
4ian, line 1024: Do you need to do two maps? If you do a single one, this allows to keep the new in the same function as the `delete`, which is looking a tiny bit safer for the reader.

newIDE/app/src/SceneEditor/index.js
AlexandreSi: Yes it is dirty.

I thought about that as well, but I didn't want to change the signature for this often-used method. But this is a case for which it's worth I think.

newIDE/app/src/InstancesEditor/index.js
```diff
@@ -604,20 +609,40 @@ export default class InstancesEditor extends Component<Props> {
...
   |}) => {
     this.lastContextMenuX = offsetX;
     this.lastContextMenuY = offsetY;
-    if (this.props.onContextMenu) this.props.onContextMenu(x, y);
+    if (this.props.onContextMenu) {
+      const _ignoreSelectedObjectNamesForContextMenu =
+        typeof ignoreSelectedObjectNamesForContextMenu === 'boolean'
+          ? ignoreSelectedObjectNamesForContextMenu
+          : false;
+      this.props.onContextMenu(x, y, _ignoreSelectedObjectNamesForContextMenu);
```
4ian, line 644:
```suggestion
      this.props.onContextMenu(x, y, ignoreSelectedObjectNamesForContextMenu || false);
```
or 
```suggestion
      this.props.onContextMenu(x, y, !!ignoreSelectedObjectNamesForContextMenu);
```

Should work (undefined == falsy)

newIDE/app/src/SceneEditor/index.js
4ian: I believe you can also change the `paste` function to remove the two maps yeah, it seems useless..

4ian (approved):
LGTM after resolving comments! :)

PR #3718, opened 2022-03-03
Fix duplicated extension after rename

### LINKED ISSUE #2587: Renaming an extension leaves the original version (until application restart)
## Describe the bug

After renaming an extension it appears like two extensions are installed (one for each name).  The old name doesn't go away until the application is restarted.

## To Reproduce
1. Create new extension with a default behavior
2. Rename extension
3. Try to add a behavior to an object
4. You will see two copies of the behavior (one for the old name, one for the new name)

https://user-images.githubusercontent.com/8879811/116288018-4b724600-a74e-11eb-9376-bd37f55c0301.mp4

fannieyan:
Fixes #2587

Remove extension by old name after renaming.

### DIFF
--- a/newIDE/README.md
+++ b/newIDE/README.md
@@ -23,8 +23,8 @@ npm start # or yarn start
 
 This will open the app in your web browser.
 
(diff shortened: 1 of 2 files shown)


AlexandreSi (approved):
Simple yet effective!

PR #3720, opened 2022-03-03
Add Liluo Link on homepage

ClementPasteau:
We could also add it in the Games Showcase Dialog, but it's slightly harder, as the Dialog is shared for Examples/Tutorials/Showcase games and already has a button at the bottom to create a blank project, so I feel this would just clutter the space.
Or we need to split the dialogs into 3, to have control over each. (could be a good idea, as the title of the dialog is always "create a new project, even for tutorials & games...)

image
image

### DIFF
--- a/newIDE/app/src/MainFrame/AboutDialog.js
+++ b/newIDE/app/src/MainFrame/AboutDialog.js
@@ -272,7 +272,7 @@ export default class AboutDialog extends PureComponent<Props, State> {
             key="website"
             label={<Trans>GDevelop Website</Trans>}
             primary={false}
-            onClick={() => Window.openExternalURL('http://gdevelop-app.com')}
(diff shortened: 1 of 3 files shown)


newIDE/app/src/MainFrame/AboutDialog.js
```diff
@@ -272,7 +272,7 @@ export default class AboutDialog extends PureComponent<Props, State> {
             key="website"
             label={<Trans>GDevelop Website</Trans>}
             primary={false}
-            onClick={() => Window.openExternalURL('http://gdevelop-app.com')}
+            onClick={() => Window.openExternalURL('https://gdevelop.io')}
```
AlexandreSi, line 275: 👍

AlexandreSi (approved):
LGTM

4ian (approved):

Issue #2630, opened 2021-05-10
List of things to polish


Bouh:
## Description

All points from the list bellow have been spotted during this [live stream.](https://www.twitch.tv/videos/1011065478?t=0h35m34s)

- [ ] Undo/redo isn't active in properties panel

- [ ] When an user click for **Add an object**, even if he click on **Cancel** button a sprite object is created

- [ ] getScaleMode / Nearest / Linear aren't take into account in view editor

- [ ] Resource tab isn't refresh after a new resource has been imported in the project, the tab have to be closed and re open for see new resources

- [ ] All objects added on scene goes to the Basic Layer, could be fine to select a layer for choosing where added the object

- [ ] Select and move multiple events

- [ ] Search for **mirror** a character, actually is **flip**, people can also search for **mirror** term, a solution can be add in parenthis (mirror) in flipping action

- [ ] Reading wrong docs (GD4) for functions of GD5
https://clips.twitch.tv/GloriousSpeedyTapirPogChamp-JWlXMGpsrB23GMG3

- [ ] Same clip, sticky input feild when return carriage is used
https://clips.twitch.tv/GloriousSpeedyTapirPogChamp-JWlXMGpsrB23GMG3

- [ ] Continue on the wrong documentation for create a function
Could be fine to seperate GDevelop 4 and 5 wiki, or make visible in large font that's doc is outdated and for GDevelop 4, on all pages.
https://clips.twitch.tv/ScrumptiousIronicFerretGrammarKing-6VHAwAkDWch-tYFg

- [ ] When adding an object via the asset store there are empty lists because they refer to sounds

- [x] Dialog navigating https://clips.twitch.tv/HorribleEmpathicStapleTwitchRPG-mXFRObUXuJgLlArf
Solved in : 
   https://github.com/4ian/GDevelop/pull/2569
   https://github.com/4ian/GDevelop/commit/1991cc634a856725886674c99da79fa0f461db49

- [x] Smoothing option isn't setup on assets from the assets store
Solved in:  
   https://github.com/4ian/GDevelop/commit/076732abd9cf11c837caed96795335265c3134f8

4ian:
Excellent! Thanks for summarising all of this :) 
I've added a warning to the top of the pages of the GDevelop 4 wiki: http://wiki.compilgames.net/doku.php/ :)

anasiez:
My student group and I are working on the issue from the list these days. 
[...]

PR #3723, opened 2022-03-04
Fix resource list not updated when adding some with tab opened

fannieyan:
Partially deals with #2630

Update resource list after adding a resource with resource tab opened. We do not need to close/open the tab again to see resource changes.

### DIFF
--- a/newIDE/app/src/MainFrame/EditorContainers/ResourcesEditorContainer.js
+++ b/newIDE/app/src/MainFrame/EditorContainers/ResourcesEditorContainer.js
@@ -31,6 +31,15 @@ export class ResourcesEditorContainer extends React.Component<RenderEditorContai
     // No updates to be done.
   }
 
+  componentDidUpdate(prevProps: RenderEditorContainerProps) {
+    if (
+      this.editor &&
+      this.props.isActive &&
+      prevProps.isActive !== this.props.isActive
+    )
+      this.editor.refreshResourcesList();
+  }
+
   render() {
     const { project } = this.props;
     if (!project) return null;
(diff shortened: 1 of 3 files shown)


4ian:
I think this works, though in theory you could imagine a resource was added (which is possible) and another one removed (I don't think you can remove for now a resource outside the resources tab... but let's just imagine) - in which case the heuristic of looking at the list length is not working anymore.

It's ok as it can't happen, but I wonder if we could have forced a refresh of the tab when it's becoming active? 🤔

fannieyan:
> I think this works, though in theory you could imagine a resource was added (which is possible) and another one removed (I don't think you can remove for now a resource outside the resources tab... but let's just imagine) - in which case the heuristic of looking at the list length is not working anymore.

Yes, I thought about it and saw that it was not possible (yet) to delete a resource from elsewhere and did not try to compare all the names

> I wonder if we could have forced a refresh of the tab when it's becoming active? 🤔

That should work (and be simpler !) but a bit costly if there is no change and the user is switching between tabs ?

4ian:
> That should work (and be simpler !) but a bit costly if there is no change and the user is switching between tabs ?

This is true (and indeed simpler, which is always welcome), though in practice this should be fairly short (a few milliseconds) and absorbed from the point of view of the user by the tab switching.
In a sense, because this adds an additional check that enumerate the resources at each render, the cost is already not 0 and so also impact a tiny bit the tab switching (and any subsequent update). So it's not unreasonable to say that instead we would force the refresh, but only when the tab is switched.

Give it a try to see if it works and be can then choose what we go ahead with :)

PR #3724, opened 2022-03-04
Added an action to fade sounds on channels

### LINKED ISSUE #3688: add the missing fade audio/music volume action in Gdevelop
Screenshot_85
## Description

I always wondered whether there was a better way to fade audio/music since updating the volume level with a variable can be choppy, and after a few months I discovered that Arthuro mentioned that it exists in the code but does not show up in Gdevelop
and this can only be accessed by using a javascript event.
Javascript code in mention:
runtimeScene.getGame().getSoundManager().getSoundOnChannel(1).fade(0, 1, 1000);

After waiting many months to see whether it was fixed, it never was nor was it ever reported so I decided to open it up.

## Solution suggested

add an action for the code that exists but is not used. I believe this will be a very useful action for controlling the volume levels smoothly once implemented properly

Oxey405:
Hello !
I added a new action to GDevelop !
This action allows to fade sounds on channels smoothly !
This Closes #3688

### DIFF
--- a/Core/GDCore/Extensions/Builtin/AudioExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/AudioExtension.cpp
@@ -346,16 +346,16 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAudioExtension(
   extension
       .AddAction(
           "FadeVolume",
-          _("fade the volume of a sound played on a channel."),
-          _("Fade the volume of a sound played on a channel within a given time is seconds."),
+          _("Fade the volume of a sound played on a channel."),
+          _("Fade the volume of a sound played on a channel within a given time in seconds."),
           _("Fade the volume of a sound on channel n°_PARAM1_ to a volume of _PARAM2_ within a time of _PARAM3_ seconds"),
           _("Sounds on channels"),
           "res/actions/music24.png",
           "res/actions/music.png")
       .AddCodeOnlyParameter("currentScene", "")
-      .AddParameter("expression", _("ID of the channel"), "", true)
-      .AddParameter("expression", _("Volume to interpolate in %"), "", true)
-      .AddParameter("expression", _("Interpolation time in seconds"), "", true)
+      .AddParameter("expression", _("ID of the channel"))
+      .AddParameter("expression", _("Final volume (0-100)"))
+      .AddParameter("expression", _("Fading time in seconds"))
       .MarkAsAdvanced();
 
   extension
--- a/Core/GDCore/Extensions/Builtin/AudioExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/AudioExtension.cpp
@@ -347,15 +347,15 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAudioExtension(
       .AddAction(
           "FadeVolume",
           _("fade the volume of a sound played on a channel."),
-          _("Fade the volume of a sound played on a channel within a given time is milliseconds."),
-          _("Fade the volume of a sound on channel n°_PARAM1_ to a volume of _PARAM2_ within a time of _PARAM3_ ms"),
+          _("Fade the volume of a sound played on a channel within a given time is seconds."),
+          _("Fade the volume of a sound on channel n°_PARAM1_ to a volume of _PARAM2_ within a time of _PARAM3_ seconds"),
           _("Sounds on channels"),
           "res/actions/music24.png",
           "res/actions/music.png")
       .AddCodeOnlyParameter("currentScene", "")
       .AddParameter("expression", _("ID of the channel"), "", true)
       .AddParameter("expression", _("Volume to interpolate in %"), "", true)
-      .AddParameter("expression", _("Interpolation time in ms"), "", true)
+      .AddParameter("expression", _("Interpolation time in seconds"), "", true)
       .MarkAsAdvanced();
 
   extension
--- a/Core/GDCore/Extensions/Builtin/AudioExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/AudioExtension.cpp
@@ -345,10 +345,10 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAudioExtension(
       .MarkAsComplex();
   extension
       .AddAction(
-          "interpolateAudioTo",
-          _("Interpolate audio sound on to volume"),
-          _("Interpolate an audio sound on a channel to a volume within a given time."),
-          _("Interpolate an audio sound on _PARAM2_ to a specific volume of_PARAM1_ within a given time of _PARAM3_ ms"),
+          "FadeVolume",
+          _("fade volume of a sound."),
+          _("fade the volume an audio sound on a channel to a volume within a given time."),
+          _("Fade the volume of a sound on channel n°_PARAM2_ to a volume of _PARAM1__ within a time of _PARAM3 ms"),
           _("Sounds on channels"),
           "res/actions/music24.png",
           "res/actions/music.png")
--- a/Core/GDCore/Extensions/Builtin/AudioExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/AudioExtension.cpp
@@ -348,7 +348,7 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAudioExtension(
           "FadeVolume",
           _("fade volume of a sound."),
           _("fade the volume an audio sound on a channel to a volume within a given time."),
-          _("Fade the volume of a sound on channel n°_PARAM2_ to a volume of _PARAM1__ within a time of _PARAM3 ms"),
+          _("Fade the volume of a sound on channel n°_PARAM1_ to a volume of _PARAM2__ within a time of _PARAM3_ ms"),
           _("Sounds on channels"),
           "res/actions/music24.png",
           "res/actions/music.png")
--- a/Core/GDCore/Extensions/Builtin/AudioExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/AudioExtension.cpp
@@ -348,7 +348,7 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAudioExtension(
           "FadeVolume",
           _("Fade the volume of a sound played on a channel."),
           _("Fade the volume of a sound played on a channel within a given time in seconds."),
-          _("Fade the volume of a sound on channel n°_PARAM1_ to a volume of _PARAM2_ within a time of _PARAM3_ seconds"),
+          _("Fade the volume of a sound on channel n°_PARAM1_ to a volume of _PARAM2_ within _PARAM3_ seconds"),
           _("Sounds on channels"),
           "res/actions/music24.png",
           "res/actions/music.png")
--- a/Core/GDCore/Extensions/Builtin/AudioExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/AudioExtension.cpp
@@ -346,15 +346,15 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAudioExtension(
   extension
       .AddAction(
           "FadeVolume",
-          _("fade volume of a sound."),
-          _("fade the volume an audio sound on a channel to a volume within a given time."),
-          _("Fade the volume of a sound on channel n°_PARAM1_ to a volume of _PARAM2__ within a time of _PARAM3_ ms"),
+          _("fade the volume of a sound played on a channel."),
+          _("Fade the volume of a sound played on a channel within a given time is milliseconds."),
+          _("Fade the volume of a sound on channel n°_PARAM1_ to a volume of _PARAM2_ within a time of _PARAM3_ ms"),
           _("Sounds on channels"),
           "res/actions/music24.png",
           "res/actions/music.png")
       .AddCodeOnlyParameter("currentScene", "")
       .AddParameter("expression", _("ID of the channel"), "", true)
-      .AddParameter("expression", _("Volume to interpolate to (0 to 100)"), "", true)
+      .AddParameter("expression", _("Volume to interpolate in %"), "", true)
       .AddParameter("expression", _("Interpolation time in ms"), "", true)
       .MarkAsAdvanced();
 
--- a/Core/GDCore/Extensions/Builtin/AudioExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/AudioExtension.cpp
@@ -343,6 +343,20 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAudioExtension(
           "res/actions/music.png")
       .AddCodeOnlyParameter("currentScene", "")
       .MarkAsComplex();
+  extension
+      .AddAction(
+          "interpolateAudioTo",
+          _("Interpolate audio sound on to volume"),
+          _("Interpolate an audio sound on a channel to a volume within a given time."),
+          _("Interpolate an audio sound on _PARAM2_ to a specific volume of_PARAM1_ within a given time of _PARAM3_ ms"),
+          _("Sounds on channels"),
+          "res/actions/music24.png",
+          "res/actions/music.png")
+      .AddCodeOnlyParameter("currentScene", "")
+      .AddParameter("expression", _("ID of the channel"), "", true)
+      .AddParameter("expression", _("Volume to interpolate to (0 to 100)"), "", true)
+      .AddParameter("expression", _("Interpolation time in ms"), "", true)
+      .MarkAsAdvanced();
 
   extension
       .AddCondition("MusicPlaying",
--- a/Core/GDCore/Extensions/Builtin/AudioExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/AudioExtension.cpp
@@ -345,15 +345,29 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAudioExtension(
       .MarkAsComplex();
   extension
       .AddAction(
-          "FadeVolume",
+          "FadeSoundVolume",
           _("Fade the volume of a sound played on a channel."),
-          _("Fade the volume of a sound played on a channel within a given time in seconds."),
-          _("Fade the volume of a sound on channel n°_PARAM1_ to a volume of _PARAM2_ within _PARAM3_ seconds"),
+          _("Fade the volume of a sound played on a channel to the specified volume within the specified duration."),
+          _("Fade the sound on channel _PARAM1_ to volume _PARAM2_ within _PARAM3_ seconds"),
           _("Sounds on channels"),
+          "res/actions/son24.png",
+          "res/actions/son.png")
+      .AddCodeOnlyParameter("currentScene", "")
+      .AddParameter("expression", _("Channel identifier"))
+      .AddParameter("expression", _("Final volume (0-100)"))
+      .AddParameter("expression", _("Fading time in seconds"))
+      .MarkAsAdvanced();
+  extension
+      .AddAction(
+          "FadeMusicVolume",
+          _("Fade the volume of a music played on a channel."),
+          _("Fade the volume of a music played on a channel to the specified volume within the specified duration."),
+          _("Fade a music on channel _PARAM1_ to volume _PARAM2_ within _PARAM3_ seconds"),
+          _("Music on channels"),
           "res/actions/music24.png",
           "res/actions/music.png")
       .AddCodeOnlyParameter("currentScene", "")
-      .AddParameter("expression", _("ID of the channel"))
+      .AddParameter("expression", _("Channel identifier"))
       .AddParameter("expression", _("Final volume (0-100)"))
       .AddParameter("expression", _("Fading time in seconds"))
       .MarkAsAdvanced();
--- a/Core/GDCore/Extensions/Builtin/AudioExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/AudioExtension.cpp
@@ -362,7 +362,7 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAudioExtension(
           "FadeMusicVolume",
           _("Fade the volume of a music played on a channel."),
           _("Fade the volume of a music played on a channel to the specified volume within the specified duration."),
-          _("Fade a music on channel _PARAM1_ to volume _PARAM2_ within _PARAM3_ seconds"),
+          _("Fade the music on channel _PARAM1_ to volume _PARAM2_ within _PARAM3_ seconds"),
           _("Music on channels"),
           "res/actions/music24.png",
           "res/actions/music.png")
--- a/GDJS/Runtime/events-tools/soundtools.ts
+++ b/GDJS/Runtime/events-tools/soundtools.ts
@@ -428,6 +428,23 @@ namespace gdjs {
         runtimeScene
           .getSoundManager()
           .unloadAudio(soundFile, /* isMusic= */ true);
+    
+      export const interpolateAudioTo = (
+        runtimeScene: gdjs.RuntimeScene,
+        channel :integer, 
+        toVolume: float, 
+        timeOfFade: float
+        ) => {
+          
+          const sound = runtimeScene.getSoundManager().getSoundOnChannel(channel);
+          if (sound) runtimeScene.getSoundManager().interpolateAudioTo(channel, toVolume, timeOfFade)
+          else {
+            logger.error(
+              `Cannot fade the volume of a non-existing sound on channel ${channel}.`
+            );
+          }
+          
+      }
     }
   }
 }
--- a/GDJS/Runtime/events-tools/soundtools.ts
+++ b/GDJS/Runtime/events-tools/soundtools.ts
@@ -435,9 +435,15 @@ namespace gdjs {
         toVolume: float,
         timeOfFade: float
       ) => {
+        //Clamping volume
+        if(toVolume > 100) {
+          toVolume = 100;
+        }
+        if(toVolume < 0) {
+          toVolume = 0;
+        }
         const sound = runtimeScene.getSoundManager().getSoundOnChannel(channel);
-        if (sound)
-            sound.fade(sound.getVolume(), toVolume, timeOfFade);
+        if (sound) sound.fade(sound.getVolume(), toVolume, timeOfFade);
         else {
           logger.error(
             `Cannot fade the volume of a non-existing sound on channel ${channel}.`
--- a/GDJS/Runtime/events-tools/soundtools.ts
+++ b/GDJS/Runtime/events-tools/soundtools.ts
@@ -429,7 +429,7 @@ namespace gdjs {
           .getSoundManager()
           .unloadAudio(soundFile, /* isMusic= */ true);
 
-      export const interpolateAudioTo = (
+      export const FadeVolume = (
         runtimeScene: gdjs.RuntimeScene,
         channel: integer,
         toVolume: float,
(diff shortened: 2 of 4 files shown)


GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
```diff
@@ -731,6 +731,20 @@ namespace gdjs {
...
+    ): this {
+      const sound = this.getSoundOnChannel(channel);
+      if (sound) {
+        sound.fade(sound.getVolume(), toVolume, timeOfFade);
+      } else {
+        logger.error('cannot interpolate audio on non-existing channel');
+      }
+      return this;
+    }
+
```
arthuro555, line 747:
```suggestion
```

This code is redundant and inconsistent with the rest of the functions for sound manipulation.

GDJS/Runtime/events-tools/soundtools.ts
```diff
@@ -428,6 +428,24 @@ namespace gdjs {
...
+        runtimeScene: gdjs.RuntimeScene,
+        channel: integer,
+        toVolume: float,
+        timeOfFade: float
+      ) => {
+        const sound = runtimeScene.getSoundManager().getSoundOnChannel(channel);
+        if (sound)
+          runtimeScene
+            .getSoundManager()
+            .interpolateAudioTo(channel, toVolume, timeOfFade);
```
arthuro555, line 442:
```suggestion
          sound.fade(sound.getVolume(), toVolume, timeOfFade);
```

Move the code from howler-sound-manager.js here.

Core/GDCore/Extensions/Builtin/AudioExtension.cpp
```diff
@@ -343,6 +343,20 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAudioExtension(
...
+          "interpolateAudioTo",
+          _("Interpolate audio sound on to volume"),
+          _("Interpolate an audio sound on a channel to a volume within a given time."),
+          _("Interpolate an audio sound on _PARAM2_ to a specific volume of_PARAM1_ within a given time of _PARAM3_ ms"),
+          _("Sounds on channels"),
+          "res/actions/music24.png",
+          "res/actions/music.png")
+      .AddCodeOnlyParameter("currentScene", "")
+      .AddParameter("expression", _("ID of the channel"), "", true)
+      .AddParameter("expression", _("Volume to interpolate to (0 to 100)"), "", true)
```
arthuro555, line 357:
```suggestion
      .AddParameter("expression", _("Volume to fade to in %"), "", true)
```

Core/GDCore/Extensions/Builtin/AudioExtension.cpp
```diff
@@ -343,6 +343,20 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAudioExtension(
...
+      .AddParameter("expression", _("Interpolation time in ms"), "", true)
```
arthuro555, line 358:
```suggestion
      .AddParameter("expression", _("Fade duration in ms"), "", true)
```

arthuro555 (commented):
I am worried about this being a process over multiple frames: what happens if the volume is set manually during the fade? What if another fade is started in the middle of another? Should there be any way to check if a fade is currently in progress, if it just ended, or to cancel it?

Core/GDCore/Extensions/Builtin/AudioExtension.cpp
```diff
@@ -343,6 +343,20 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAudioExtension(
           "res/actions/music.png")
       .AddCodeOnlyParameter("currentScene", "")
       .MarkAsComplex();
+  extension
+      .AddAction(
+          "interpolateAudioTo",
+          _("Interpolate audio sound on to volume"),
+          _("Interpolate an audio sound on a channel to a volume within a given time."),
+          _("Interpolate an audio sound on _PARAM2_ to a specific volume of_PARAM1_ within a given time of _PARAM3_ ms"),
```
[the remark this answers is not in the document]
AlexandreSi: I think that we use seconds instead of milliseconds in other functions, could you use seconds here as well so that we are coherent?

Core/GDCore/Extensions/Builtin/AudioExtension.cpp
AlexandreSi: Same here for seconds

GDJS/Runtime/events-tools/soundtools.ts
```diff
@@ -428,6 +428,24 @@ namespace gdjs {
...
           .unloadAudio(soundFile, /* isMusic= */ true);
+
+      export const interpolateAudioTo = (
+        runtimeScene: gdjs.RuntimeScene,
+        channel: integer,
+        toVolume: float,
+        timeOfFade: float
+      ) => {
+        const sound = runtimeScene.getSoundManager().getSoundOnChannel(channel);
+        if (sound)
```
AlexandreSi, line 439: To be safe, you can add some value checks on the inputs:
- Make sure duration is a positive float (and divide it per 1000 once the parameter is asked in seconds)
- Make sure your volume is positive, but I don't know if there should be a maximum value. Maybe just a warning that over 100% it could alter the quality? I'm not even sure about that.

Oxey405:
> Thanks @Oxey405 for submitting this request and @arthuro555 for the review!
> 
> > what happens if the volume is set manually during the fade? What if another fade is started in the middle of another? Should there be any way to check if a fade is currently in progress, if it just ended, or to cancel it?
> 
> I think that @arthuro555 has a good point here. To go further:
> 
> * Is it possible to cancel a fade? If not, should we prevent setting volume on the same channel?

Oh well actually that is a good question... I don't think stopping a fade is avaible with howler sound manager if GDevelop is using [this](https://github.com/Qriva/howler-sound-controller) library.

For the minus volumes the howler sound manager should already handle that but I will implement the limit anyway as it is just 3 lines of code.

Oxey405:
> I am worried about this being a process over multiple frames: what happens if the volume is set manually during the fade? What if another fade is started in the middle of another? Should there be any way to check if a fade is currently in progress, if it just ended, or to cancel it?

This has been tried by me and what happens is basically this :
It stops the fade and set the volume to what you gave...

Core/GDCore/Extensions/Builtin/AudioExtension.cpp
```diff
@@ -343,6 +343,20 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAudioExtension(
           "res/actions/music.png")
       .AddCodeOnlyParameter("currentScene", "")
       .MarkAsComplex();
+  extension
+      .AddAction(
+          "FadeVolume",
+          _("fade volume of a sound."),
+          _("fade the volume an audio sound on a channel to a volume within a given time."),
```
AlexandreSi, line 350: This sentence reads weird, but I don't speak english natively so hard to suggest something else.

GDJS/Runtime/events-tools/soundtools.ts
```diff
@@ -428,6 +428,28 @@ namespace gdjs {
...
+
+      export const FadeVolume = (
+        runtimeScene: gdjs.RuntimeScene,
+        channel: integer,
+        toVolume: float,
+        timeOfFade: float
+      ) => {
+        //Clamping volume
+        if (toVolume > 100) {
+          toVolume = 100;
```
AlexandreSi, line 440: I recommend that you define a new variable that would be

```js
const targetVolume = Math.max(0, Math.min(100, toVolume))
```

But then again, I'm not sure that we should add a max to the volume.

AlexandreSi: Can you add the same positive check on duration and use seconds instead of milliseconds?

AlexandreSi:
> > I am worried about this being a process over multiple frames: what happens if the volume is set manually during the fade? What if another fade is started in the middle of another? Should there be any way to check if a fade is currently in progress, if it just ended, or to cancel it?
> 
> This has been tried by me and what happens is basically this : It stops the fade and set the volume to what you gave...

This behavior seems ok to me.

Core/GDCore/Extensions/Builtin/AudioExtension.cpp
ZachjuKamashi: I don't speak English natively either, but I am able to at the very least compose better sentences. Here would be my suggestions.
Line 349: "Fade the volume level of an audio track"
Line 350: "Fade the volume level of a sound channel within a given time"

AlexandreSi:
To sum up, I think there 3 points that need to be addressed before merging it:
- Is it possible to set a volume greater than 100? I thought it could be useful but as you said, howler [does its own clamping](https://github.com/goldfire/howler.js/blob/master/src/howler.core.js#L1311-L1312) so as well doing it ourselves => ✅ OK
- Duration should also be checked to be greater or equal to 0 (this time, howler does not do the check) => 👷 To do
- What should be the unit of the fade duration: GDevelop's goal is to simplify video game creation. We don't want to lose users with different values expressed in different units. This is already the case:
  - Tween actions have durations expressed in milliseconds
  - The time extension that procures timers and scene elapsed time return values in seconds

  So far, there is no notion of duration in the audio extension. I think we should use seconds because, well, it's easier to the human mind and some users are not familiar with unit conversion. => 👷 To do
  (Of course, we should have seconds everywhere, or at least make sure the unit is written in the parameters, but it's not the case in the tween actions currently, it could be the objective of another merge request)

GDJS/Runtime/events-tools/soundtools.ts
```diff
@@ -428,6 +428,28 @@ namespace gdjs {
...
+        }
+        if (toVolume < 0) {
+          toVolume = 0;
+        }
```
arthuro555, line 444: Don't do your own clamping, and don't do it here!

-> We already have a clamping function at the top of howler-sound-manager that you can use
-> Functions in events tools are supposed to be just functional wrappers to the object oriented APIs, as events are functional. Any logic relevant to the functionality of the API should be present on the object oriented function, in this case `HowlerSound.fade`.

arthuro555 (commented):
It has come to my attention that the `HowlerSound` class has a member `volume` that doesn't get updated when using the fade action. I do not quite remember if it is really important or not, but better look into that.

AlexandreSi:
> It has come to my attention that the HowlerSound class has a member volume that doesn't get updated when using the fade action

Looking at this: https://github.com/goldfire/howler.js/blob/143ae442386c7b42d91a007d0b1f1695528abe64/src/howler.core.js#L1356

It seems to me that `self.volume` is updated. Or maybe you're talking about something else?

GDJS/Runtime/events-tools/soundtools.ts
AlexandreSi: Oh I didn't realize that, I wrote my comment too quickly. Indeed, @Oxey405 your fade method should be implemented in our class `HowlerSoundManager` and you can use `clampVolume` that is defined in the same file.

And as @arthuro555 mentioned it, we should make sure that the `HowlerSound._volume` corresponds to the `Howl` instance `volume` property.

arthuro555:
> Or maybe you're talking about something else?

I am. What you are linking is the ˋHowlˋ class from howler.js, I am talking about [the ˋHowlerSoundˋ class from howler-sound-manager.ts](https://github.com/4ian/GDevelop/blob/master/GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts#L40)

GDJS/Runtime/events-tools/soundtools.ts
arthuro555: > your fade method should be implemented in our class HowlerSoundManager

I disagree, it is already implemented on `HowlerSound`, and it isn't the responsibility of the manager to call methods on the sounds. What I was saying is that the lamping, which is a necessity for the API to work correctly, should be present on the actual API (`HowlerSound.fade`) and not on the events-tools "bridge function". See the volume action for example, we are still calling the method on the `HowlerSound`, but we are not clamping in events-tools: we are clamping in `HowlerSound.setVolume`.

> we should make sure that the `HowlerSound._volume` corresponds to the Howl instance volume property

So I checked thee code again, and we shouldn't set the `_volume` in the fade method. It is there because the Howl doesn't has to have been preloaded, and it will need some time to load before playing. Until then, we cannot access the `Howl` to set the volume or other properties. Therefore, volume and other properties are stored on the `HowlerSound` instance until the Howl is ready, and we can pass on those properties to the Howl. 
Therefore, we should keep the fade settings on a separate member object, and [when starting to play the sound](https://github.com/4ian/GDevelop/blob/9710360180ab41461f95d1fc6b2c36ca04bdb44e/GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts#L107), if a fade was requested for that sound before it was loaded, then the fade should be executed with the stored parameters (after setting the volume, since that is what defines the beginning of the fade).

To ensure that `getVolume` returns the right volume, we should only return `_volume` while the sound is not loaded (`_id === null`), then get the volume from the howl once it started playing.

AlexandreSi: Yes sorry, `HowlerSoundManager` shouldn't handle this new fade method.
And I didn't realize there could be a different behavior if the sound is playing or not. Do you think that a fade action request on a sound not yet playing should be stored in memory and then used when the sound starts playing?

arthuro555: Yes, as that is what the setVolume action does too. In fact, I am surprised that not all properties are handled like that, and only a select few.

Oxey405:
Hey again ! I am sorry for the 2 days I was off I had to do an internship observation...

I fixed what was asked but I discovered a problem yet another bug ;( !
The volume property is not updated when fading which is not cool...

See by yourself ; download this : [AudioFadeTester.zip](https://github.com/4ian/GDevelop/files/8234418/AudioFadeTester.zip)
project and run it on the custom version from this PR and you'll see...

+ I just noticed I failed my formatting so I may need to refix that...

Anyway hope you have a good day @AlexandreSi and thank you for your time !

GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
```diff
@@ -295,6 +295,8 @@ namespace gdjs {
      * @returns The current instance for chaining.
      */
     fade(from: float, to: float, duration: float): this {
+      from = clampVolume(from);
+      to = clampVolume(to);
       if (this._id !== null) this._howl.fade(from, to, duration, this._id);
```
arthuro555, line 300: This can be simplified:

```suggestion
      if (this._id !== null) this._howl.fade(clampVolume(from), clampVolume(to), duration, this._id);
```

Oxey405:
Hey again ! I think I am done now ! I hope anyone can validate it by the next week !

Thank you to all for your help and have a good day ;)

Core/GDCore/Extensions/Builtin/AudioExtension.cpp
```diff
@@ -343,6 +343,20 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAudioExtension(
           "res/actions/music.png")
       .AddCodeOnlyParameter("currentScene", "")
       .MarkAsComplex();
+  extension
+      .AddAction(
+          "FadeVolume",
+          _("fade the volume of a sound played on a channel."),
```
4ian, line 349: The title should be capitalized and without period at the end.

Core/GDCore/Extensions/Builtin/AudioExtension.cpp
```diff
@@ -343,6 +343,20 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAudioExtension(
...
+          _("Fade the volume of a sound played on a channel within a given time is seconds."),
```
4ian, line 350: This is not correct English. You don't need to precise the unit here, it's already implicit in the parameter.

Core/GDCore/Extensions/Builtin/AudioExtension.cpp
```diff
@@ -343,6 +343,20 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAudioExtension(
...
+          _("Fade the volume of a sound on channel n°_PARAM1_ to a volume of _PARAM2_ within a time of _PARAM3_ seconds"),
```
4ian, line 351: - Do we use n° for other actions?
- `in _PARAM3_ seconds`

Core/GDCore/Extensions/Builtin/AudioExtension.cpp
```diff
@@ -343,6 +343,20 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAudioExtension(
...
+          _("Sounds on channels"),
+          "res/actions/music24.png",
+          "res/actions/music.png")
+      .AddCodeOnlyParameter("currentScene", "")
+      .AddParameter("expression", _("ID of the channel"), "", true)
```
4ian, line 356: Is this how we write the other channels in other actions?

Core/GDCore/Extensions/Builtin/AudioExtension.cpp
```diff
@@ -343,6 +343,20 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAudioExtension(
...
+      .AddParameter("expression", _("Volume to interpolate in %"), "", true)
+      .AddParameter("expression", _("Interpolation time in seconds"), "", true)
```
4ian, line 358: Fading time, in seconds

Core/GDCore/Extensions/Builtin/AudioExtension.cpp
```diff
@@ -343,6 +343,20 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAudioExtension(
...
+          "FadeVolume",
+          _("fade the volume of a sound played on a channel."),
+          _("Fade the volume of a sound played on a channel within a given time is seconds."),
+          _("Fade the volume of a sound on channel n°_PARAM1_ to a volume of _PARAM2_ within a time of _PARAM3_ seconds"),
+          _("Sounds on channels"),
+          "res/actions/music24.png",
+          "res/actions/music.png")
+      .AddCodeOnlyParameter("currentScene", "")
+      .AddParameter("expression", _("ID of the channel"), "", true)
+      .AddParameter("expression", _("Volume to interpolate in %"), "", true)
```
4ian, line 357: `Final volume (0-100)`

Core/GDCore/Extensions/Builtin/AudioExtension.cpp
```diff
@@ -343,6 +343,20 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAudioExtension(
...
+      .AddParameter("expression", _("Interpolation time in seconds"), "", true)
+      .MarkAsAdvanced();
```
4ian, line 359: You marked these parameters as optional, remove all the `"", true)`.

GDJS/Runtime/events-tools/soundtools.ts
```diff
@@ -428,6 +428,21 @@ namespace gdjs {
         runtimeScene
           .getSoundManager()
           .unloadAudio(soundFile, /* isMusic= */ true);
+
+      export const FadeVolume = (
```
4ian, line 432: a function should be written in camelCase

GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
```diff
@@ -223,7 +223,8 @@ namespace gdjs {
      * @returns A float from 0 to 1.
      */
     getVolume(): float {
-      return this._volume;
+      if (this._id === null) return this._volume;
```
4ian, line 226: Are you sure about this? Why do we have this._volume? Why is there a usage of it and a usage of this._howl.volume?

4ian (commented):
There is some progress, but you need to pay much more attention to details.
This will be used by dozens of thousands of users in thousands of games, so everything must be perfect :)

GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
arthuro555: this._volume is the volume given when we want to play a sound initially. That way, if the howl isn't loaded yet, we can keep it stored there until it is loaded, and so that we can set the volume correct volume on the howl when it starts playing. 

This is correct, because this._volume contains the future volume of the howl while it loads, and we can get the true howl volume on it directly once it is loaded. The this._volume will not be in sync since the fade function is on the howl and therefore messes with the internal howl volume directly.

That makes me think, is this kind of "automagical" function something we should be using? It is hard to tell what kind of overhead it has. Maybe an asynchronous action would be a good fit for doing that manually? Though, that would probably make the volume change more stuttering since it would be updating the volume at every game frame, not at a "sound update speed". I cannot tell if that would be a problem or not to be honest.

Oxey405:
Hey ! Uhm sorry for the insistance but this PR is ready for review so... if anyone can take care of that and approve it I'd really appreciate it !

Oxey405 (commented):
working code and compliant to the last reviews

Oxey405 (commented):
This code is good to go I think

Core/GDCore/Extensions/Builtin/AudioExtension.cpp
```diff
@@ -343,6 +343,20 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAudioExtension(
           "res/actions/music.png")
       .AddCodeOnlyParameter("currentScene", "")
       .MarkAsComplex();
+  extension
+      .AddAction(
+          "FadeVolume",
```
4ian, line 348: Call this FadeSoundVolume because this works only on sounds

Core/GDCore/Extensions/Builtin/AudioExtension.cpp
```diff
@@ -343,6 +343,20 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAudioExtension(
...
+          _("Fade the volume of a sound played on a channel."),
+          _("Fade the volume of a sound played on a channel within a given time in seconds."),
```
4ian, line 350:
```suggestion
          _("Fade the volume of a sound played on a channel to the specified volume within the specified duration."),
```

Core/GDCore/Extensions/Builtin/AudioExtension.cpp
```diff
@@ -343,6 +343,20 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAudioExtension(
...
+          _("Fade the volume of a sound on channel n°_PARAM1_ to a volume of _PARAM2_ within _PARAM3_ seconds"),
+          _("Sounds on channels"),
+          "res/actions/music24.png",
+          "res/actions/music.png")
+      .AddCodeOnlyParameter("currentScene", "")
+      .AddParameter("expression", _("ID of the channel"))
```
4ian, line 356: Please use the same name as the other parameters referring to a channel identifier :)

Core/GDCore/Extensions/Builtin/AudioExtension.cpp
```diff
@@ -343,6 +343,20 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAudioExtension(
           "res/actions/music.png")
       .AddCodeOnlyParameter("currentScene", "")
       .MarkAsComplex();
+  extension
+      .AddAction(
+          "FadeVolume",
+          _("Fade the volume of a sound played on a channel."),
+          _("Fade the volume of a sound played on a channel within a given time in seconds."),
+          _("Fade the volume of a sound on channel n°_PARAM1_ to a volume of _PARAM2_ within _PARAM3_ seconds"),
```
4ian, line 351:
```suggestion
          _("Fade the sound on channel _PARAM1_ to volume _PARAM2_ within _PARAM3_ seconds"),
```

Core/GDCore/Extensions/Builtin/AudioExtension.cpp
```diff
@@ -343,6 +343,20 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAudioExtension(
...
+          _("Sounds on channels"),
+          "res/actions/music24.png",
```
4ian, line 353: You're using the wrong icon

GDJS/Runtime/events-tools/soundtools.ts
```diff
@@ -428,6 +428,21 @@ namespace gdjs {
         runtimeScene
           .getSoundManager()
           .unloadAudio(soundFile, /* isMusic= */ true);
+
+      export const fadeVolume = (
```
4ian, line 432: Rename to `fadeSoundVolume`

GDJS/Runtime/events-tools/soundtools.ts
```diff
@@ -428,6 +428,21 @@ namespace gdjs {
...
+        runtimeScene: gdjs.RuntimeScene,
+        channel: integer,
+        toVolume: float,
+        timeOfFade: float /*In seconds and converted to ms*/
```
4ian, line 436:
```suggestion
        fadingTime: float /* in seconds */
```

GDJS/Runtime/howler-sound-manager/howler-sound-manager.ts
4ian: > It is hard to tell what kind of overhead it has

You can check the source code of Howler: https://github.com/goldfire/howler.js/blob/143ae442386c7b42d91a007d0b1f1695528abe64/src/howler.core.js#L1295

If webaudio is used, it's using a native method (which is ideal).
otherwise, it's using setInterval. Not great (this could be handled by the game engine), but ok I guess and not too risky.

4ian (commented):
To merge this, we need:
- Comments addressed
- The same action for musics.
- Because `this._volume` is not the source of truth anymore for the volume, its name is now wrong. To keep a clean code, we must rename it to `this._initialVolume`, to show that this value can't be relied on at all time (happy to get your view here @arthuro555).

Thanks! :)

Core/GDCore/Extensions/Builtin/AudioExtension.cpp
Oxey405: Done

GDJS/Runtime/events-tools/soundtools.ts
```diff
@@ -428,6 +428,35 @@ namespace gdjs {
...
+        }
+      };
+      export const fadeMusicVolume = (
+        runtimeScene: gdjs.RuntimeScene,
+        channel: integer,
+        toVolume: float,
+        timeOfFade: float /* in seconds */
+      ) => {
+        const sound = runtimeScene.getSoundManager().getMusicOnChannel(channel);
+        if (sound) sound.fade(sound.getVolume(), toVolume, timeOfFade * 1000);
```
4ian, line 453:
```suggestion
        const music = runtimeScene.getSoundManager().getMusicOnChannel(channel);
        if (music) sound.fade(sound.getVolume(), toVolume, timeOfFade * 1000);
```

GDJS/Runtime/events-tools/soundtools.ts
```diff
@@ -428,6 +428,35 @@ namespace gdjs {
...
+        else {
+          logger.error(
+            `Cannot fade the volume of a non-existing sound on channel ${channel}.`
```
4ian, line 456:
```suggestion
            `Cannot fade the volume of a non-existing music on channel ${channel}.`
```

Core/GDCore/Extensions/Builtin/AudioExtension.cpp
```diff
@@ -343,6 +343,34 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsAudioExtension(
...
+      .AddParameter("expression", _("Channel identifier"))
+      .AddParameter("expression", _("Final volume (0-100)"))
+      .AddParameter("expression", _("Fading time in seconds"))
+      .MarkAsAdvanced();
+  extension
+      .AddAction(
+          "FadeMusicVolume",
+          _("Fade the volume of a music played on a channel."),
+          _("Fade the volume of a music played on a channel to the specified volume within the specified duration."),
+          _("Fade a music on channel _PARAM1_ to volume _PARAM2_ within _PARAM3_ seconds"),
```
4ian, line 365:
```suggestion
          _("Fade the music on channel _PARAM1_ to volume _PARAM2_ within _PARAM3_ seconds"),
```

Oxey405:
Hello ! This PR is I think ready for review... Anyone please 🥺

PR #3730, opened 2022-03-05
[Platformer] Fix the "Is moving" condition by removing the 1 pixel per frame constraint

D8H:
### Context
* A game creator in the forum is having an issue with the "Is moving" condition.
  https://forum.gdevelop.io/t/problems-with-animation-speed-for-plataform-behavior/38470/24
* The game is very low resolution and uses a 60 pixel per second walking speed.
* At about 60 fps, the character moves around 1 pixel per frame (sometimes more, sometimes less).

### Expected behavior
* The "Is moving" condition should be true when the character actually moves even a "little" because "little" is subjective.

### Actual behavior
* The "Is moving" condition is false when the character walks from less than 1 pixel per frame.
* In their game, this result to frequent restart of the walking animation.

### Solution
* The 1 pixel per frame limit is frame rate dependent. Thus, the compatibility can't be kept by adding a minimum speed parameter.
* The existing condition has been made deprecated.
* The new condition doesn't test any minimal speed. Game creators can add another condition for it if they see fit.

The tests could be updated to use the new method.

### DIFF
--- a/Extensions/PlatformBehavior/Extension.cpp
+++ b/Extensions/PlatformBehavior/Extension.cpp
@@ -43,6 +43,7 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
         std::make_shared<PlatformerObjectBehavior>(),
         std::make_shared<gd::BehaviorsSharedData>());
 
+    // Deprecated, use IsMovingEvenALittle instead
     aut.AddCondition("IsMoving",
                      _("Is moving"),
                      _("Check if the object is moving (whether it is on the "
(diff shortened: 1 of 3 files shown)


JovenOculto:
Thanks for this, and btw I love that "because "little" is subjective" XD so true

Extensions/PlatformBehavior/Extension.cpp
```diff
@@ -43,6 +43,7 @@ void DeclarePlatformBehaviorExtension(gd::PlatformExtension& extension) {
         std::make_shared<PlatformerObjectBehavior>(),
         std::make_shared<gd::BehaviorsSharedData>());
 
+    // Deprecated, use IsMovingEvenALittle instead
```
arthuro555, line 46: IsMovingEvenALittle is a pretty verbose and "weird" name. If boolean expressions are added in the future, conditions would be most likely used as such, so we got to name them as carefully as an expression. As a user, I would be weirded out by such a name 😅

D8H: It should be "IsMoving", but the name is already reserved for compatibility.

What do you think of "IsMovingWithoutLowerSpeedLimit" ?

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -312,6 +314,9 @@ namespace gdjs {
 
       //5) Track the movement
       this._hasReallyMoved =
+        Math.abs(object.getX() - oldX) > 0 ||
```
4ian, line 317: Random question: is there a use case for replacing 0 by an epsilon? Like are there moment where we could have a tiny tiny displacement on the X axis? (like when against a wall?)

D8H: Yes, it will be safer.

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -24,6 +24,11 @@ namespace gdjs {
       isCollidingAnyPlatform: false,
     };
 
+    /**
+     * A very small value compare to 1 pixel, yet very huge compare to rounding errors.
+     */
+    private static readonly epsilon = 2 ** -20;
```
4ian, line 30: Might be worth using 1 / (2 << xxx) for consistency with the other epsilons?

4ian (approved):

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
D8H: They are only in tests. I didn't know this syntax, I will change them next time I add a test.

PR #3732, opened 2022-03-07
Make Liluo.io thumbnail configurable

D8H:
image

### DIFF
--- a/newIDE/app/src/Export/ExportDialog/ExportLauncher.js
+++ b/newIDE/app/src/Export/ExportDialog/ExportLauncher.js
@@ -35,6 +35,7 @@ import {
   addCreateBadgePreHookIfNotClaimed,
   TRIVIAL_FIRST_WEB_EXPORT,
 } from '../../Utils/GDevelopServices/Badge';
+import { getWebBuildThumbnailUrl } from '../../Utils/GDevelopServices/Build';
 
 type State = {|
   exportStep: BuildStep,
@@ -330,6 +331,14 @@ export default class ExportLauncher extends Component<Props, State> {
     }));
   };
 
+  _getWebBuildThumbnailUrl = (buildId: string) => {
+    const project = this.props.project;
+    if (!project) {
+      return undefined;
+    }
+    return getWebBuildThumbnailUrl(project, buildId);
+  };
+
   render() {
     const {
       exportStep,
@@ -410,7 +419,12 @@ export default class ExportLauncher extends Component<Props, State> {
           )}
         {authenticatedUser.authenticated &&
           (exportPipeline.renderCustomStepsProgress ? (
-            exportPipeline.renderCustomStepsProgress(build, errored, exportStep)
+            exportPipeline.renderCustomStepsProgress(
+              build,
+              errored,
+              exportStep,
+              this._getWebBuildThumbnailUrl
+            )
           ) : (
             <Line expand>
               <BuildStepsProgress
@@ -420,6 +434,7 @@ export default class ExportLauncher extends Component<Props, State> {
                 stepMaxProgress={stepMaxProgress}
                 stepCurrentProgress={stepCurrentProgress}
                 errored={errored}
+                getThumbnailURL={this._getWebBuildThumbnailUrl}
               />
             </Line>
           ))}
--- a/newIDE/app/src/Export/ExportDialog/ExportLauncher.js
+++ b/newIDE/app/src/Export/ExportDialog/ExportLauncher.js
@@ -331,11 +331,6 @@ export default class ExportLauncher extends Component<Props, State> {
     }));
   };
 
-  _getWebBuildThumbnailUrl = (buildId: string) => {
-    const project = this.props.project;
-    return project && getWebBuildThumbnailUrl(project, buildId);
-  };
-
   render() {
     const {
       exportStep,
--- a/newIDE/app/src/Export/ExportDialog/ExportLauncher.js
+++ b/newIDE/app/src/Export/ExportDialog/ExportLauncher.js
@@ -333,10 +333,7 @@ export default class ExportLauncher extends Component<Props, State> {
 
   _getWebBuildThumbnailUrl = (buildId: string) => {
     const project = this.props.project;
-    if (!project) {
-      return undefined;
-    }
-    return getWebBuildThumbnailUrl(project, buildId);
+    return project && getWebBuildThumbnailUrl(project, buildId);
   };
 
   render() {
@@ -423,7 +420,9 @@ export default class ExportLauncher extends Component<Props, State> {
               build,
               errored,
               exportStep,
-              this._getWebBuildThumbnailUrl
+              buildId =>
+                this.props.project &&
+                getWebBuildThumbnailUrl(this.props.project, buildId)
             )
           ) : (
             <Line expand>
@@ -434,7 +433,6 @@ export default class ExportLauncher extends Component<Props, State> {
                 stepMaxProgress={stepMaxProgress}
                 stepCurrentProgress={stepCurrentProgress}
                 errored={errored}
-                getThumbnailURL={this._getWebBuildThumbnailUrl}
               />
             </Line>
           ))}
--- a/newIDE/app/src/Export/ExportDialog/index.js
+++ b/newIDE/app/src/Export/ExportDialog/index.js
@@ -249,9 +249,6 @@ const ExportDialog = ({
           authenticatedUser={authenticatedUser}
           game={game}
           onGameUpdated={setGame}
-          getThumbnailURL={(buildId: string) =>
-            project && getWebBuildThumbnailUrl(project, buildId)
-          }
         />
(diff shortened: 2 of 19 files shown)


newIDE/app/src/Utils/GDevelopServices/Build.js
```diff
@@ -61,6 +61,29 @@ export const getBuildArtifactUrl = (
...
+    .split('\\');
+  if (path.length < 1) {
+    return '';
+  }
+  const fileName = path[path.length - 1];
+  if (!fileName) {
+    return '';
+  }
+  return `https:/games.gdevelop-app.com/game-${buildId}/${fileName}`;
+};
```
D8H, line 85: The mapping from resource name to build file path should use the same thing as the exporter to avoid this to break if the export change.

newIDE/app/src/GameDashboard/GameCard.js
```diff
@@ -33,68 +49,105 @@ export const GameCard = ({
...
+                  expand
+                  useFullHeight
+                  justifyContent="center"
+                  alignItems="center"
+                >
+                  <Text align="center">
+                    <Trans>No thumbnail set</Trans>
+                  </Text>
+                </Column>
+              </Paper>
```
D8H, line 89: The text doesn't align at the center vertically.

newIDE/app/src/Export/ExportDialog/ExportLauncher.js
```diff
@@ -330,6 +331,14 @@ export default class ExportLauncher extends Component<Props, State> {
     }));
   };
 
+  _getWebBuildThumbnailUrl = (buildId: string) => {
+    const project = this.props.project;
+    if (!project) {
+      return undefined;
+    }
+    return getWebBuildThumbnailUrl(project, buildId);
+  };
```
arthuro555, line 340: The whole function can be simplified to `(buildId: string) => this.props.project && getWebBuildThumbnailUrl(this.props.project, buildId);`. IMO too verbose code takes too much time to read and is therefore less clean.

newIDE/app/src/Export/ExportDialog/index.js
```diff
@@ -237,6 +238,12 @@ const ExportDialog = ({
           authenticatedUser={authenticatedUser}
           game={game}
           onGameUpdated={setGame}
+          getThumbnailURL={(buildId: string) => {
+            if (!project) {
```
arthuro555, line 242: That looks like a repetition of the function I just osted a comment on, if we follow the DRY code principle, `getWebBuildThumbnailUrl` should probably already check and return undefined if the project argument is undefined.

newIDE/app/src/Utils/GDevelopServices/Build.js
```diff
@@ -61,6 +61,29 @@ export const getBuildArtifactUrl = (
...
+
+export const getWebBuildThumbnailUrl = (
+  project: gdProject,
+  buildId: string
+): string => {
+  // TODO use some helper methods to do this?
+  const path = project
+    .getPlatformSpecificAssets()
+    .get('liluo', `thumbnail`)
+    .split('\\');
```
arthuro555, line 76: That looks like it would break if a non-windows path is used?

newIDE/app/src/Utils/GDevelopServices/Build.js
```diff
@@ -61,6 +61,29 @@ export const getBuildArtifactUrl = (
...
+  if (path.length < 1) {
+    return '';
+  }
+  const fileName = path[path.length - 1];
+  if (!fileName) {
+    return '';
+  }
+  return `https:/games.gdevelop-app.com/game-${buildId}/${fileName}`;
```
arthuro555, line 84: The majority of this function is just getting the file name out of a path. This can simply be done using the `path` module, which is easily (and IIRC already is by default by CRA) polyfilled to work just fine on the browser. This would make the function cleaner to read: to be clean, a function must do what you expect it to do from its name. This function is more of a `basename` than a `getThumbnailURL`, making it take more time to understand than it needs to. 
```suggestion
  import { basename } from "path";
  return `https:/games.gdevelop-app.com/game-${buildId}/${basename(path)}`;
```

newIDE/app/src/stories/componentStories/BuildProgressAndActions.stories.js
```diff
@@ -23,55 +23,73 @@ export default {
 };
 
 export const Errored = () => (
-  <BuildProgressAndActions build={erroredCordovaBuild} game={game1} />
+  <BuildProgressAndActions
+    build={erroredCordovaBuild}
+    game={game1}
+    getThumbnailURL={buildId => undefined}
```
arthuro555, line 29: There should be at least one instance of the component with the thumbnail for it to be tested in the storybook.

arthuro555 (commented):
Why, yes, I have begun reading "Clean Code" by Robert C. Martin, how could you tell? 😅 I may be too nitpicky, sorry about that.

newIDE/app/src/Utils/GDevelopServices/Build.js
D8H: Thanks, that's partially answer my TODO comment.

newIDE/app/src/Utils/GDevelopServices/Build.js
D8H: Yes, this is a temporary hack, there is a TODO right before.

arthuro555: Ah, right 😅 My brain automatically skips those 😓

newIDE/app/src/Export/ExportDialog/index.js
D8H: This is a way of seeing this. But, I personally prefer to have a method that ask what it needs and let the caller do what they need to give the right parameters. It's not to `getWebBuildThumbnailUrl` to decide what should happen if there is no project.

newIDE/app/src/Export/ExportDialog/ExportLauncher.js
D8H: Thanks, I made the change.

4ian: This function `_getWebBuildThumbnailUrl` seems unused?

newIDE/app/src/GameDashboard/GameCard.js
4ian: The expand is probably not working because `Column` is expecting to be inside a flexbox container (a column/line is "just" a div with `display: flex`. And with `flex: 1` when `expand` is set, which requires the parent to be already a `display: flex` itself). But `Paper` is not a flexbox container. Set `display: flex` to Paper to see if it works better

newIDE/app/src/GameDashboard/GameCard.js
```diff
@@ -33,68 +49,105 @@ export const GameCard = ({
...
-                icon={<TimelineIcon />}
-                label={<Trans>Analytics</Trans>}
-                onClick={onOpenAnalytics}
+            ) : (
+              <Paper
+                variant="outlined"
+                style={{
+                  width: '240px',
+                  height: '135px',
+                  whiteSpace: 'normal',
```
4ian, line 76:
```suggestion
                  whiteSpace: 'normal',
                  display: 'flex',
```

(see other comment for explanation)

newIDE/app/src/GameDashboard/GameCard.js
```diff
@@ -33,68 +49,105 @@ export const GameCard = ({
...
+                <Column
+                  expand
+                  useFullHeight
+                  justifyContent="center"
+                  alignItems="center"
+                >
+                  <Text align="center">
+                    <Trans>No thumbnail set</Trans>
+                  </Text>
+                </Column>
```
4ian, line 88: Give a try at `EmptyMessage` (unrelated to centering, just that it's made for this use case of showing that something is missing):

```suggestion
                <EmptyMessage>
                  <Trans>No thumbnail set</Trans>
                </EmptyMessage>
```

4ian (approved):

4ian (approved):

newIDE/app/src/GameDashboard/GameCard.js
```diff
@@ -33,68 +44,98 @@ export const GameCard = ({
...
               />
-              <FlatButton
-                icon={<PlaylistPlayIcon />}
-                label={<Trans>Builds</Trans>}
-                onClick={onOpenBuilds}
+            ) : (
+              <Paper
+                variant="outlined"
+                style={{
+                  width: thumbnailWidth,
```
ClementPasteau, line 68: why not add those styles in the `styles` object at the top? same for image styles

ClementPasteau (approved):
I think you can merge this :)

In the future, we may have to consider validating the format of the picture provided by the user, to avoid non 16/9 formats.
check frontend/backend?

PR #3733, opened 2022-03-07
Make clearer that the publishing is an export

### LINKED ISSUE #3729: Eliminate confusion at game export
- [ ] Rename the button PUBLISH ON STORES to PUBLISH YOUR GAME, like the name of the next window that opens.
Or better EXPORT YOUR GAME

There is no upload to stores. Many who want an exe or apk don't think about the stores.
publish, build, blah.. all they want at this point is Export :)
Today we had the third angry(!) forum threat because someone couldn't find the online build options anymore after the rework.

___

What do you think about exporting the manually HTML5 build optional as .zip/.7z?
It would be much easier to load it on itchio, etc.

4ian:
And a tiny fix too.

This removes the notion of "stores". I think this is not the last time we rework this, but for now this should avoid people wrongly clicking on "Build manually".

image

Fixes #3729

### DIFF
--- a/newIDE/app/src/BehaviorsEditor/index.js
+++ b/newIDE/app/src/BehaviorsEditor/index.js
@@ -128,6 +128,7 @@ const BehaviorsEditor = (props: Props) => {
             }
             actionLabel={<Trans>Add a behavior</Trans>}
             helpPagePath="/behaviors"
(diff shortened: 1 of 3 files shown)


ClementPasteau (approved):
👍

Bouh (approved):

PR #3735, opened 2022-03-07
Add possibility to replace text in text events (Comments and Groups)

### LINKED ISSUE #3385: Find/replace will find matches in comments and event groups, but will not replace the text
## Describe the bug

Find/replace will find matches in comments and event groups, but will not replace the text in those items.

## To Reproduce
1. Create a comment and event group with the words "player"
2. Use the search and replace tool to search for "player"
3. You will see the comments and event group are highlighted and counted in the "results" counter
4. Click "REPLACE" and notice that comments and event groups are not modified

## Video

https://user-images.githubusercontent.com/8879811/146714759-f824880a-bdab-4490-aef9-dd25a7d93cf3.mp4

## Other details
* Windows 10
* GDevelop 5.0.124

fannieyan:
Fixes #3385

Add possibility to replace text in text events like Comments and Groups names

replace-in-event-text

### DIFF
--- a/Core/GDCore/IDE/Events/EventsRefactorer.cpp
+++ b/Core/GDCore/IDE/Events/EventsRefactorer.cpp
@@ -573,7 +573,7 @@ void EventsRefactorer::ReplaceStringInEvents(gd::ObjectsContainer& project,
     }
 
     if (inEventStrings) {
-      bool eventStringModified = ReplaceStringInStringEvent(
+      bool eventStringModified = ReplaceStringInEventSearchableStrings(
           project, layout, events[i], toReplace, newString, matchCase);
     }
 
@@ -684,13 +684,14 @@ bool EventsRefactorer::ReplaceStringInConditions(
   return somethingModified;
 }
 
-bool EventsRefactorer::ReplaceStringInStringEvent(gd::ObjectsContainer& project,
-                                                  gd::ObjectsContainer& layout,
-                                                  gd::BaseEvent& event,
-                                                  gd::String toReplace,
-                                                  gd::String newString,
-                                                  bool matchCase) {
-  vector<gd::String> newStringEvents;
+bool EventsRefactorer::ReplaceStringInEventSearchableStrings(
+    gd::ObjectsContainer& project,
+    gd::ObjectsContainer& layout,
+    gd::BaseEvent& event,
+    gd::String toReplace,
+    gd::String newString,
+    bool matchCase) {
+  vector<gd::String> newEventStrings;
   vector<gd::String> stringEvent = event.GetAllSearchableStrings();
 
   for (std::size_t sNb = 0; sNb < stringEvent.size(); ++sNb) {
@@ -698,10 +699,10 @@ bool EventsRefactorer::ReplaceStringInStringEvent(gd::ObjectsContainer& project,
         matchCase ? stringEvent[sNb].FindAndReplace(toReplace, newString, true)
                   : ReplaceAllOccurencesCaseUnsensitive(
                         stringEvent[sNb], toReplace, newString);
-    newStringEvents.push_back(newStringEvent);
+    newEventStrings.push_back(newStringEvent);
   }
 
-  bool somethingModified = event.ReplaceAllInSearchableString(newStringEvents);
+  bool somethingModified = event.ReplaceAllSearchableStrings(newEventStrings);
 
   return somethingModified;
 }
--- a/Core/GDCore/IDE/Events/EventsRefactorer.cpp
+++ b/Core/GDCore/IDE/Events/EventsRefactorer.cpp
@@ -9,6 +9,8 @@
 #include <memory>
 
 #include "GDCore/CommonTools.h"
+#include "GDCore/Events/Builtin/CommentEvent.h"
+#include "GDCore/Events/Builtin/GroupEvent.h"
 #include "GDCore/Events/Event.h"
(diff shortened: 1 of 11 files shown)


Core/GDCore/IDE/Events/EventsRefactorer.cpp
```diff
@@ -677,6 +686,36 @@ bool EventsRefactorer::ReplaceStringInConditions(
...
+                        stringEvent[sNb], toReplace, newString);
+
+    if (newStringEvent != stringEvent[sNb]) {
+      if (event.GetType() == "BuiltinCommonInstructions::Group") {
+        gd::GroupEvent& groupEvent = dynamic_cast<gd::GroupEvent&>(event);
+        groupEvent.SetName(newStringEvent);
+      } else if (event.GetType() == "BuiltinCommonInstructions::Comment") {
+        gd::CommentEvent& commentEvent = dynamic_cast<gd::CommentEvent&>(event);
+        commentEvent.SetComment(newStringEvent);
+      }
```
4ian, line 711: This seems a good place where we can use "polymorphism" instead of a switch case (https://refactoring.guru/replace-conditional-with-polymorphism).

In other words: 
because you can't really know from the outside of many strings there are and how to replace them, you can create a method doing this whole job inside BaseEvent (by default, it does nothing) and then redefine it for each event.

You can call this method: `ReplaceInAllSearchableStrings` or something similar. It should I guess takes the old string, new string, and whether to match case (and return true if something was modified).

The advantage: in the future, you only need to redefine this method if we create a new event type and it will work automatically.
The disadvantage: you put something related to "refactoring" events inside the event itself... but well that's just a search/replace, it makes sense to have this in an event class :)

newIDE/app/src/AssetStore/InstallAsset.js
```diff
@@ -246,7 +246,8 @@ export const addAssetToProject = async ({
             parameter.defaultValue,
             true,
             true,
-            true
+            true,
+            false
```
4ian, line 250: Good opportunity I think to use "named parameters" to make the code clearer: 
```
/*matchCase=*/ true, 
/*inConditions=*/ true,
...
```

Ideally we would use an option object rather than a set of booleans (it's not a great style of code) but well!

Core/GDCore/IDE/Events/EventsRefactorer.cpp
```diff
@@ -677,6 +684,28 @@ bool EventsRefactorer::ReplaceStringInConditions(
   return somethingModified;
 }
 
+bool EventsRefactorer::ReplaceStringInStringEvent(gd::ObjectsContainer& project,
+                                                  gd::ObjectsContainer& layout,
+                                                  gd::BaseEvent& event,
+                                                  gd::String toReplace,
+                                                  gd::String newString,
+                                                  bool matchCase) {
+  vector<gd::String> newStringEvents;
```
4ian, line 693: Nitpicking:
```suggestion
  vector<gd::String> newEventStrings;
```

There is a single event, but with (potentially) multiple strings.

Core/GDCore/IDE/Events/EventsRefactorer.h
```diff
@@ -208,6 +209,19 @@ class GD_CORE_API EventsRefactorer {
                                      gd::String newString,
                                      bool matchCase);
 
+  /**
+   * Replace all occurrences of a gd::String in string events (comments and
+   * groups)
```
4ian, line 214:
```suggestion
   * Replace all occurrences of a gd::String in strings of events (for example: comments and
   * name of groups).

```

Core/GDCore/IDE/Events/EventsRefactorer.h
```diff
@@ -208,6 +209,19 @@ class GD_CORE_API EventsRefactorer {
...
 
+  /**
+   * Replace all occurrences of a gd::String in string events (comments and
+   * groups)
+   *
+   * \return true if something was modified.
+   */
+  static bool ReplaceStringInStringEvent(gd::ObjectsContainer& project,
+                                         gd::ObjectsContainer& layout,
+                                         gd::BaseEvent& stringEvent,
```
4ian, line 220:
```suggestion
                                         gd::BaseEvent& event,
```

That's just an event? A "string event" does not really exist as a notion (and should not exist I would say).

Core/GDCore/IDE/Events/EventsRefactorer.h
```diff
@@ -208,6 +209,19 @@ class GD_CORE_API EventsRefactorer {
                                      gd::String newString,
                                      bool matchCase);
 
+  /**
+   * Replace all occurrences of a gd::String in string events (comments and
+   * groups)
+   *
+   * \return true if something was modified.
+   */
+  static bool ReplaceStringInStringEvent(gd::ObjectsContainer& project,
```
4ian, line 218:
```suggestion
  static bool ReplaceStringInEventStrings(gd::ObjectsContainer& project,
```

or even `ReplaceStringInEventSearchableStrings` to be very precise, but I'll leave the choice up to you!

4ian:
> I feel like my implementation is a bit weird though, because it depends on the getAllSearchableStrings method (especially for the mapping between the vector index and attributes). What do you think ?

That's unusual, but a careful naming can help show that there is some mirroring between these methods. I think that's fine and a good idea to avoid pulling too much responsibility into the event methods! So great idea.

4ian:
LGTM! 👍

PR #3738, opened 2022-03-08
Use custom collision masks for draggable behavior

### LINKED ISSUE #1298: Draggable behavior ignore the shape and size of custom hitbox and rotation of object

fannieyan:
Fixes #1298

custom-collision-masks-use

Edit : Added toggle in behaviors
image

### DIFF
--- a/Extensions/DraggableBehavior/draggableruntimebehavior.ts
+++ b/Extensions/DraggableBehavior/draggableruntimebehavior.ts
@@ -14,9 +14,11 @@ namespace gdjs {
      * When the owner is being dragged, no other manager can start dragging it.
      */
     _draggedByDraggableManager: DraggableManager | null = null;
+    _checkCollisionMask: boolean;
 
     constructor(runtimeScene, behaviorData, owner) {
       super(runtimeScene, behaviorData, owner);
+      this._checkCollisionMask = behaviorData.checkCollisionMask ? true : false;
     }
 
     updateFromBehaviorData(oldBehaviorData, newBehaviorData): boolean {
@@ -193,10 +195,11 @@ namespace gdjs {
       }
       const position = this.getPosition(runtimeScene, draggableRuntimeBehavior);
       if (
-        !draggableRuntimeBehavior.owner.insideObject(
-          position[0],
-          position[1]
-        ) ||
+        !draggableRuntimeBehavior.owner.insideObject(position[0], position[1])
+      ) {
+        return false;
+      } else if (
+        draggableRuntimeBehavior._checkCollisionMask &&
         !draggableRuntimeBehavior.owner.isCollidingWithPoint(
           position[0],
           position[1]
--- a/Extensions/DraggableBehavior/draggableruntimebehavior.ts
+++ b/Extensions/DraggableBehavior/draggableruntimebehavior.ts
@@ -193,7 +193,10 @@ namespace gdjs {
       }
       const position = this.getPosition(runtimeScene, draggableRuntimeBehavior);
       if (
-        !draggableRuntimeBehavior.owner.insideObject(position[0], position[1])
+        !draggableRuntimeBehavior.owner.isCollidingWithPoint(
+          position[0],
+          position[1]
+        )
       ) {
         return false;
       }
--- a/Extensions/DraggableBehavior/draggableruntimebehavior.ts
+++ b/Extensions/DraggableBehavior/draggableruntimebehavior.ts
@@ -193,6 +193,10 @@ namespace gdjs {
       }
       const position = this.getPosition(runtimeScene, draggableRuntimeBehavior);
       if (
+        !draggableRuntimeBehavior.owner.insideObject(
+          position[0],
+          position[1]
+        ) ||
         !draggableRuntimeBehavior.owner.isCollidingWithPoint(
           position[0],
           position[1]
--- a/Extensions/DraggableBehavior/tests/draggableruntimebehavior.spec.js
+++ b/Extensions/DraggableBehavior/tests/draggableruntimebehavior.spec.js
@@ -105,11 +105,11 @@ describe('gdjs.DraggableRuntimeBehavior', function () {
         object.setPosition(450, 500);
         object2.setPosition(450, 500);
         if (firstInFront) {
-          object.setCustomZOrder(2);
-          object2.setCustomZOrder(1);
+          object.setZOrder(2);
+          object2.setZOrder(1);
         } else {
-          object.setCustomZOrder(1);
-          object2.setCustomZOrder(2);
+          object.setZOrder(1);
+          object2.setZOrder(2);
         }
 
         // Drag'n'drop
@@ -221,11 +221,11 @@ describe('gdjs.DraggableRuntimeBehavior', function () {
         object.setPosition(450, 500);
         object2.setPosition(450, 500);
         if (firstInFront) {
-          object.setCustomZOrder(2);
-          object2.setCustomZOrder(1);
+          object.setZOrder(2);
+          object2.setZOrder(1);
         } else {
-          object.setCustomZOrder(1);
-          object2.setCustomZOrder(2);
+          object.setZOrder(1);
+          object2.setZOrder(2);
         }
 
         // Drag'n'drop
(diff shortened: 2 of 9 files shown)


GDJS/tests/tests/Extensions/testruntimeobject.js
```diff
@@ -40,6 +40,13 @@
     this.hitBoxesDirty = true;
   }
 
+  setCustomZOrder(zorder) {
```
AlexandreSi, line 43: Was it needed to add this method for this feature?

fannieyan: For some reason, setZOrder with a TestRuntimeObject triggered an error (`cannot read _parentID of undefined`). I suspected it was because the method `getRendererObject` of `TestRuntimeObject` does not return `undefined` (which is the default in RuntimeObject) and setting it to return undefined broke a test in another file...

AlexandreSi: You encountered an error when running tests with your modifications?

Extensions/DraggableBehavior/tests/draggableruntimebehavior.spec.js
```diff
@@ -34,20 +34,22 @@ describe('gdjs.DraggableRuntimeBehavior', function () {
...
-  var object2 = new gdjs.RuntimeObject(runtimeScene, {
+  object.setCustomWidthAndHeight(10, 10);
+  var object2 = new gdjs.TestRuntimeObject(runtimeScene, {
     name: 'obj1',
     type: '',
     behaviors: [{ name: 'Behavior1', type: 'DraggableBehavior::Draggable' }],
     variables: [],
     effects: [],
   });
+  object2.setCustomWidthAndHeight(10, 10);
```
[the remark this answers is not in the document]
fannieyan: `isCollidingWithPoint` considers a point outside of the polygon if its position is on the edge of the polygon, so I had to make the object bigger so that the test mouse could pick the object

GDJS/tests/tests/Extensions/testruntimeobject.js
fannieyan: With my modifications, I had to make my object wider, so I had to use TestRuntimeObject, and changing the class of the object used in the test made it break at the setZOrder method

Extensions/DraggableBehavior/draggableruntimebehavior.ts
```diff
@@ -193,7 +193,10 @@ namespace gdjs {
       }
       const position = this.getPosition(runtimeScene, draggableRuntimeBehavior);
       if (
-        !draggableRuntimeBehavior.owner.insideObject(position[0], position[1])
```
4ian, line 196: Nice! Would be good to change the JS documentation of `insideObject` to tell that "`isCollidingWithPoint` is available if you want to check if a point is inside the object collision mask."

GDJS/tests/tests/Extensions/testruntimeobject.js
4ian: > setting it to return undefined broke a test in another file

Any chance you can share this other test/quickly check why it's failing?

I think in theory we should return undefined, but not sure why we returned an empty object (which looks dodgy).

Extensions/DraggableBehavior/draggableruntimebehavior.ts
4ian: Actually, I think it may be worth to do first a AABB check (`insideObject`) and then a precise check (`isCollidingWithPoint`).

Otherwise, we risk to run complex code (`isCollidingWithPoint`) for potential a lot of objects (if noted as `N`, and complexity of the point-polygon collision check as `C`, it results in `N*C`).
With `insideObject`, the checks should be simpler, so the cost is "only" `N*c` (`c` being smaller than `C`).

A really good solution would be to have a way to check objects that are in a certain area (which is possible if storing objects in a spatial datastructure like a R-Tree, giving a search in `O(log N)`)... but for now let's not implement that because it would need to be offered by the game engine itself I think.

fannieyan: Alright !

GDJS/tests/tests/Extensions/testruntimeobject.js
4ian: I wonder if we should rework this test to work with a `TestRuntimeObjectWithFakeRenderer`? and have `TestRuntimeObject` return null as the renderer?

4ian (commented):
Fair enough! Last stuff: this is normally considered as a breaking change because this can change the way a game works.

For example, some games may want to use a non precise check because... because I'm not sure why they would. But that's something to check!

Either we think we need to keep backward compatibility, and we'll add a boolean "do a precision check against the object collision masks: true/false" (false if the attribute is not set) (and true by default for new behaviors).
OR we can consider this is not a big change and we ship without a backward compatability.

tristanbob:
> Either we think we need to keep backward compatibility, and we'll add a boolean "do a precision check against the object collision masks: true/false" (false if the attribute is not set) (and true by default for new behaviors).
> OR we can consider this is not a big change and we ship without a backward compatibility.

I think adding a boolean parameter would be the best way to go.  It offers backward compatibility and flexibility for users.  

For instance, a user might want players to be able to drag a very small object by using the large AABB of the image, but when not being dragged, they want the object to interact with only the visible object that is described by a small collision mask.

fannieyan:
Added toggle in behavior to allow users to chose whether they want to use collision masks or bounding box
image

Extensions/DraggableBehavior/tests/draggableruntimebehavior.spec.js
```diff
@@ -96,6 +104,70 @@ describe('gdjs.DraggableRuntimeBehavior', function () {
       expect(object.getY()).to.be(700);
     });
 
+    it('can drag an object without collision mask check', function () {
+      object.setPosition(450, 500);
+      object.rotateTowardAngle(45, 0, runtimeScene);
```
4ian, line 109: Any reason you're not using `setAngle` instead of `rotateTowardAngle`?

Extensions/DraggableBehavior/tests/draggableruntimebehavior.spec.js
```diff
@@ -96,6 +104,70 @@ describe('gdjs.DraggableRuntimeBehavior', function () {
...
+      expect(object.getY()).to.be(600);
+
+      object.rotateTowardAngle(0, 0, runtimeScene);
+    });
+
+    it('can drag an object with collision mask check', function () {
+      object2.setPosition(450, 500);
+      object2.rotateTowardAngle(45, 0, runtimeScene);
+
+      // Should not drag in bounding box but not in hitbox
```
4ian, line 135: Could you rephrase these comments in the test?
It's a bit hard to understand the double negation "not drag in bounding box but not in hitbox".
For example, explain what is "in" (the dragged point is in the bounding box but the dragged point is also not in the hitbox).

4ian (approved):
👍👍

Extensions/DraggableBehavior/tests/draggableruntimebehavior.spec.js
fannieyan: Wow, I somehow did not see that the method existed 😅

PR #3741, opened 2022-03-09
[Quick win] Add missing "Audio" category for extensions

Bouh:
I was reviewing the Volume Falloff extension and noticed that isn't assigned to the Audio category because this one is missing.

image


4ian:
👍

PR #3744, opened 2022-03-10
Allow game owners to share their ownership with other users.

D8H:
The backend automatically add owners in the author list if they're missing.

image

image

### DIFF
--- a/newIDE/app/src/GameDashboard/GameDetailsDialog.js
+++ b/newIDE/app/src/GameDashboard/GameDetailsDialog.js
@@ -43,7 +43,7 @@ import TextField from '../UI/TextField';
 import KeyboardIcon from '@material-ui/icons/Keyboard';
 import SportsEsportsIcon from '@material-ui/icons/SportsEsports';
 import SmartphoneIcon from '@material-ui/icons/Smartphone';
-import StarIcon from '@material-ui/icons/Star';
+import Crown from '../UI/CustomSvgIcons/Crown';
 import { I18n } from '@lingui/react';
 
 const styles = {
@@ -299,7 +299,7 @@ export const GameDetailsDialog = ({
                                   size="small"
                                   icon={
                                     ownerUsernames.includes(username) ? (
-                                      <StarIcon />
+                                      <Crown />
                                     ) : (
                                       undefined
                                     )
--- a/newIDE/app/src/GameDashboard/GameDetailsDialog.js
+++ b/newIDE/app/src/GameDashboard/GameDetailsDialog.js
@@ -7,7 +7,6 @@ import FlatButton from '../UI/FlatButton';
 import { Line, Spacer } from '../UI/Grid';
 import {
   type Game,
-  type ApiOnlyPublicGameInfo,
   updateGame,
   deleteGame,
   getPublicGame,
@@ -39,6 +38,7 @@ import HelpButton from '../UI/HelpButton';
 import { type PublicGame } from '../Utils/GDevelopServices/Game';
 import PlaceholderLoader from '../UI/PlaceholderLoader';
 import PublicGamePropertiesDialog from '../ProjectManager/PublicGamePropertiesDialog';
+import { type PartialGameChanges } from '../ProjectManager/PublicGamePropertiesDialog';
 import TextField from '../UI/TextField';
 import KeyboardIcon from '@material-ui/icons/Keyboard';
 import SportsEsportsIcon from '@material-ui/icons/SportsEsports';
@@ -150,7 +150,7 @@ export const GameDetailsDialog = ({
   );
 
   const updateGameFromProject = async (
-    apiOnlyPublicGameInfo: ApiOnlyPublicGameInfo
+    apiOnlyPublicGameInfo: PartialGameChanges
   ) => {
     if (!project || !profile) return;
     const { id } = profile;
--- a/newIDE/app/src/GameDashboard/GameDetailsDialog.js
+++ b/newIDE/app/src/GameDashboard/GameDetailsDialog.js
@@ -174,18 +174,21 @@ export const GameDetailsDialog = ({
         orientation: project.getOrientation(),
         // The thumbnailUrl is updated only when a build is made public.
       });
-      const authorAcls = getAclsFromUserIds(project.getAuthorIds().toJSArray());
-      const ownerAcls = getAclsFromUserIds(partialGameChange.ownerIds);
-      await setGameUserAcls(getAuthorizationHeader, id, gameId, {
-        ownership: ownerAcls,
-        author: authorAcls,
-      });
+      try {
+        const authorAcls = getAclsFromUserIds(
+          project.getAuthorIds().toJSArray()
+        );
+        const ownerAcls = getAclsFromUserIds(partialGameChange.ownerIds);
+        await setGameUserAcls(getAuthorizationHeader, id, gameId, {
+          ownership: ownerAcls,
+          author: authorAcls,
+        });
+      } catch (error) {
+        console.error('Unable to update the game owners or authors:', error);
+      }
+      onGameUpdated(updatedGame);
     } catch (error) {
       console.error('Unable to update the game:', error);
-    }
-    // This should always be true in theory.
-    if (updatedGame || oldPublicGame) {
-      onGameUpdated(updatedGame || oldPublicGame);
     }
   };
 
--- a/newIDE/app/src/GameDashboard/GameDetailsDialog.js
+++ b/newIDE/app/src/GameDashboard/GameDetailsDialog.js
@@ -159,6 +159,8 @@ export const GameDetailsDialog = ({
     if (!project || !profile) return;
     const { id } = profile;
 
+    // Set public game to null as it will be refetched automatically by the callback above.
+    setPublicGame(null);
     try {
       const gameId = project.getProjectUuid();
       const updatedGame = await updateGame(getAuthorizationHeader, id, gameId, {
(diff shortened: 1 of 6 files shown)


newIDE/app/src/Utils/GDevelopServices/Game.js
```diff
@@ -25,6 +26,10 @@ export type PublicGame = {
   categories?: string[],
 };
 
+export type ApiOnlyPublicGameInfo = {
```
4ian, line 29: Can you use a strict type `{| ... |}`?

4ian: Also, one thing about this type: the name "ApiOnly" looks like something that would be just sent by the API, or just received by the API. But this is not exposed at any point by the API, it's not an entity that exists (and that's a good thing :)), and it's only used for a component basically.

So could we:
- move this into the component file using it in its props (and import in the parent file)
- find a name that explains what's in without talking about the api itself?

4ian: Basically, this can be understood as a "PublicGameChanges" or "GameChanges" or maybe better: "PartialGameChanges". I.e:
It's a description of the changes to do to a Game, though it's only a partial description (the rest of the changes are "casted" from the Project to the Game).

And we can keep it in a component, because the API has no knowledge of the existence of this (it's just a detail of the IDE, but the API itself expect you to send a `Game` to its update endpoint :) And this is good!)

newIDE/app/src/GameDashboard/GameDetailsDialog.js
```diff
@@ -37,10 +38,12 @@ import HelpButton from '../UI/HelpButton';
 import { type PublicGame } from '../Utils/GDevelopServices/Game';
 import PlaceholderLoader from '../UI/PlaceholderLoader';
 import PublicGamePropertiesDialog from '../ProjectManager/PublicGamePropertiesDialog';
+import { type PartialGameChanges } from '../ProjectManager/PublicGamePropertiesDialog';
```
ClementPasteau, line 41: you can merge this line with the one above

newIDE/app/src/GameDashboard/GameDetailsDialog.js
```diff
@@ -37,10 +38,12 @@ import HelpButton from '../UI/HelpButton';
...
 import TextField from '../UI/TextField';
 import KeyboardIcon from '@material-ui/icons/Keyboard';
 import SportsEsportsIcon from '@material-ui/icons/SportsEsports';
 import SmartphoneIcon from '@material-ui/icons/Smartphone';
+import Crown from '../UI/CustomSvgIcons/Crown';
```
ClementPasteau, line 46: 👑

newIDE/app/src/GameDashboard/GameDetailsDialog.js
```diff
@@ -210,11 +219,10 @@ export const GameDetailsDialog = ({
   );
 
   const authorUsernames =
-    publicGame && publicGame.authors
-      ? publicGame.authors
-          .map(author => (author ? author.username : null))
-          .filter(Boolean)
-      : [];
+    publicGame && publicGame.authors.map(author => author.username);
```
ClementPasteau, line 222: some users don't have a username defined, so I would keep the `.filter(Boolean)` to ensure we don't have a `null` object in the authors (or owners) array

newIDE/app/src/GameDashboard/GameDetailsDialog.js
```diff
@@ -278,7 +286,7 @@ export const GameDetailsDialog = ({
                       alignItems="center"
                       noMargin
                     >
-                      {authorUsernames && (
+                      {authorUsernames && ownerUsernames && (
```
ClementPasteau, line 289: it may be an edge case, but I don't think we need the ownerUsernames here, only the authorUsernames.
the check on ownerUsernames can be done below, only when you want to add the crown icon

newIDE/app/src/ProjectManager/PublicGameProperties.js
```diff
@@ -125,6 +129,22 @@ function PublicGameProperties({
               </Trans>
             }
           />
+          {// This view is used for public game properties as well as project properties.
+          // This property is not shown in project properties.
+          setOwnerIds && (
```
ClementPasteau, line 134: should the owners be before the authors? In order of importance?

newIDE/app/src/Utils/GDevelopServices/Game.js
```diff
@@ -131,12 +132,19 @@ export const getGameUrl = (game: ?Game) => {
   return GDevelopGamesPlatform.getGameUrl(game.id);
 };
 
+export const getAclsFromOwnerIds = (
```
ClementPasteau, line 135: this new function is doing the same as below, any reason to create it rather than keep the one before and specify the "feature"?

ClementPasteau: Actually you don't even need the feature, as you're specifying after the return of the function.
So maybe this could just become `getAclsFromUserIds()` ?

newIDE/app/src/GameDashboard/GameDetailsDialog.js
```diff
@@ -146,7 +149,9 @@ export const GameDetailsDialog = ({
     [loadPublicGame]
   );
 
-  const updateGameFromProject = async () => {
+  const updateGameFromProject = async (
+    apiOnlyPublicGameInfo: PartialGameChanges
```
ClementPasteau, line 153: now that you've changed the type name, maybe also change the variable name? `partialGameChanges`?

newIDE/app/src/GameDashboard/GameDetailsDialog.js
```diff
@@ -609,9 +624,9 @@ export const GameDetailsDialog = ({
               open={isPublicGamePropertiesDialogOpen}
               project={project}
               publicGame={publicGame}
-              onApply={() => {
+              onApply={apiOnlyPublicGameInfo => {
```
ClementPasteau, line 627: same here `partialGameChanges`

newIDE/app/src/ProjectManager/PublicGameProperties.js
D8H: I though it was better to have owners after because they are picked from authors. Something like: "To which authors do you want to grant ownership?"

newIDE/app/src/GameDashboard/GameDetailsDialog.js
```diff
@@ -165,8 +174,18 @@ export const GameDetailsDialog = ({
...
+          project.getAuthorIds().toJSArray()
+        );
+        const ownerAcls = getAclsFromUserIds(partialGameChange.ownerIds);
+        await setGameUserAcls(getAuthorizationHeader, id, gameId, {
+          ownership: ownerAcls,
+          author: authorAcls,
+        });
+      } catch (error) {
+        console.error('Unable to update the game owners or authors:', error);
+      }
```
D8H, line 188: When this catch happen, the game detail dialog will show a loading indicator indefinitely because `publicGame`is null and `onGameUpdated` is never called to refetch it.
This is not a big issue because it will probably only happen if the service can't be reached, but I find it odd that the component force a refetch with knowledge of other component behavior.

ClementPasteau: I think you mean the catch just below?
In this case, can we, in the catch, call again `setPublicGame(oldGame)` (you defined the variable but haven't used it :)
And also show an error box with `showErrorBox()` asking to try again? 

As you mention, it shouldn't happen much, so no need for a fancy UI

ClementPasteau (approved):
All good!

newIDE/app/src/GameDashboard/GameDetailsDialog.js
```diff
@@ -165,11 +172,40 @@ export const GameDetailsDialog = ({
...
+          message:
+            i18n._(t`Unable to update the game owners or authors.`) +
+            ' ' +
+            i18n._(t`Verify your internet connection or try again later.`),
+          rawError: error,
+          errorId: 'game-acls-update-error',
+        });
+      }
+      // Set public game to null to have it been refetched by the callback above.
+      // TODO this probably needs a refactor.
```
4ian, line 196: Can you explain what/why is this refactoring needed, what is the suboptimal behavior that results from this?

newIDE/app/src/Utils/GDevelopServices/Game.js
```diff
@@ -12,7 +12,8 @@ export type PublicGame = {
   authorName: string, // this corresponds to the publisher name
   publicWebBuildId?: ?string,
   description?: string,
-  authors: Array<?UserPublicProfile>,
```
4ian, line 15: @ClementPasteau @D8H Please verify that the backend is never sending null/undefined for authors.
I'm pretty sure I had the issue and I had to fix it by adding this `?` and making the UI robust against this.

4ian (approved):
Be sure to triple check that UserPublicProfile can't be null when returned by the backend. It was not the case in the past and this created crashs in the development version.

newIDE/app/src/GameDashboard/GameDetailsDialog.js
D8H: `publicGame` is set to null knowing that another component will detect this as an initialization state and fetch the Game again.
This component should either ask for game fetching or do it itself and signal the update.

ClementPasteau: actually the `publicGame` being set to null is here to trigger the show of a loader by the top component (because there is no public game, it considers it is loading)
It is the `onGameUpdated` that triggers the refetch. so, not that bad?

D8H: ok

PR #3745, opened 2022-03-10
Allow users to reorder editor tabs

fannieyan:
Allow users to reorder editor tabs by drag-and-dropping them to wanted position.

reorder-editor-tabs-with-drag-and-drop

Ripple effect is still displayed on focus (when using Tab to navigate between tabs) but I had to remove it on touch so that the preview image only takes the tab and prevent this from happening : 
image

### DIFF
--- a/newIDE/app/src/MainFrame/EditorTabs/DraggableEditorTabs.js
+++ b/newIDE/app/src/MainFrame/EditorTabs/DraggableEditorTabs.js
@@ -1,0 +1,158 @@
+// @flow
+import * as React from 'react';
+
+import { makeDragSourceAndDropTarget } from '../../UI/DragAndDrop/DragSourceAndDropTarget';
+import { ScreenTypeMeasurer } from '../../UI/Reponsive/ScreenTypeMeasurer';
+import { ColumnDropIndicator } from './DropIndicator';
+import {
+  type EditorTabsState,
+  type EditorTab,
+  getEditors,
+  getCurrentTabIndex,
+} from './EditorTabsHandler';
+import {
+  ClosableTabs,
+  ClosableTab,
+  type ClosableTabProps,
+} from '../../UI/ClosableTabs';
+
+type DraggableClosableTabsProps = {|
+  hideLabels?: boolean,
+  editorTabs: EditorTabsState,
+  onClickTab: (index: number) => void,
+  onCloseTab: (editor: EditorTab) => void,
+  onCloseOtherTabs: (editor: EditorTab) => void,
+  onCloseAll: () => void,
+  onTabActive: (editor: EditorTab) => void,
+  onDropTab: (fromIndex: number, toIndex: number) => void,
+|};
+
+export function DraggableClosableTabs({
+  hideLabels,
+  editorTabs,
+  onClickTab,
+  onCloseTab,
+  onCloseOtherTabs,
+  onCloseAll,
+  onTabActive,
+  onDropTab,
+}: DraggableClosableTabsProps) {
+  let draggedTabIndex: number | typeof undefined = undefined;
+
+  return (
+    <ClosableTabs hideLabels={hideLabels}>
+      {getEditors(editorTabs).map((editorTab, id) => {
+        const isCurrentTab = getCurrentTabIndex(editorTabs) === id;
+        return (
+          <DraggableClosableTab
+            index={id}
+            label={editorTab.label}
+            key={editorTab.key}
+            id={`tab-${editorTab.key.replace(/\s/g, '-')}`}
+            active={isCurrentTab}
+            onClick={() => onClickTab(id)}
+            onClose={() => onCloseTab(editorTab)}
+            onCloseOthers={() => onCloseOtherTabs(editorTab)}
+            onCloseAll={onCloseAll}
+            onActivated={() => onTabActive(editorTab)}
+            closable={editorTab.closable}
+            onBeginDrag={() => {
+              draggedTabIndex = id;
+              return editorTab;
+            }}
+            onDrop={toIndex => {
+              if (draggedTabIndex !== undefined) {
+                onDropTab(draggedTabIndex, id);
+                draggedTabIndex = undefined;
+              }
+            }}
+          />
+        );
+      })}
+    </ClosableTabs>
+  );
+}
+
+type DraggableClosableTabProps = {|
+  index: number,
+  onBeginDrag: () => EditorTab,
+  onDrop: (toIndex: number) => void,
+  ...ClosableTabProps,
+|};
+
+export function DraggableClosableTab({
+  index,
+  id,
+  active,
+  onClose,
+  onCloseOthers,
+  onCloseAll,
+  label,
+  closable,
+  onClick,
+  onActivated,
+  onBeginDrag,
+  onDrop,
+}: DraggableClosableTabProps) {
+  const DragSourceAndDropTarget = makeDragSourceAndDropTarget<EditorTab>(
+    'draggable-closable-tab'
+  );
+
+  return (
+    <ScreenTypeMeasurer>
+      {screenType => (
+        <DragSourceAndDropTarget
+          beginDrag={onBeginDrag}
+          canDrag={() => {
+            // On touchscreens, we allow DnD on active tab so that user can scroll
+            if (screenType === 'touch' && !active) return false;
+            // We want "Home" tab to stay on the left
+            return index !== 0;
+          }}
+          canDrop={() => true}
+          drop={() => {
+            onDrop(index);
+          }}
+        >
+          {({ connectDragSource, connectDropTarget, isOver, canDrop }) => {
+            // If on a touch screen, setting the whole item to be
+            // draggable would prevent scroll. Set the icon only to be
+            // draggable if the item is not selected. When selected,
+            // set the whole item to be draggable.
+            const canDragOnlyIcon = screenType === 'touch' && !active;
+
+            // Add an extra div because connectDropTarget/connectDragSource can
+            // only be used on native elements
+            const dropTarget = connectDropTarget(
+              <div
+                style={{
+                  display: 'flex',
+                  flexShrink: 0,
+                  // transform: 'translate3d(0,0,0)',
+                }}
+              >
+                <ClosableTab
+                  id={id}
+                  active={active}
+                  onClose={onClose}
+                  onCloseOthers={onCloseOthers}
+                  onCloseAll={onCloseAll}
+                  label={label}
+                  closable={closable}
+                  onClick={onClick}
+                  onActivated={onActivated}
+                  key={id}
+                />
+                {isOver && <ColumnDropIndicator />}
+              </div>
+            );
+
+            if (!dropTarget) return null;
+
+            return canDragOnlyIcon ? dropTarget : connectDragSource(dropTarget);
+          }}
+        </DragSourceAndDropTarget>
+      )}
+    </ScreenTypeMeasurer>
+  );
+}
--- a/newIDE/app/src/MainFrame/EditorTabs/DropIndicator.js
+++ b/newIDE/app/src/MainFrame/EditorTabs/DropIndicator.js
@@ -1,0 +1,30 @@
+// @flow
+import * as React from 'react';
+import ThemeConsumer from '../../UI/Theme/ThemeConsumer';
+
+const styles = {
+  columnDropIndicator: {
+    borderRight: '1px solid #18dcf2',
+    borderLeft: '1px solid #18dcf2',
+    width: 7,
+    marginLeft: '-1px',
+    height: '100%',
+    boxSizing: 'border-box',
+  },
+};
+
+export function ColumnDropIndicator() {
+  return (
+    <ThemeConsumer>
+      {gdevelopTheme => (
+        <div
+          style={{
+            ...styles.columnDropIndicator,
+            backgroundColor: gdevelopTheme.closableTabs.selectedBackgroundColor,
+            borderColor: gdevelopTheme.closableTabs.backgroundColor,
+          }}
+        />
+      )}
+    </ThemeConsumer>
+  );
+}
--- a/newIDE/app/src/MainFrame/EditorTabs/EditorTabsHandler.js
+++ b/newIDE/app/src/MainFrame/EditorTabs/EditorTabsHandler.js
@@ -1,17 +1,17 @@
 // @flow
 import * as React from 'react';
 import findIndex from 'lodash/findIndex';
-import { EventsEditorContainer } from './EditorContainers/EventsEditorContainer';
-import { DebuggerEditorContainer } from './EditorContainers/DebuggerEditorContainer';
-import { EventsFunctionsExtensionEditorContainer } from './EditorContainers/EventsFunctionsExtensionEditorContainer';
-import { ExternalEventsEditorContainer } from './EditorContainers/ExternalEventsEditorContainer';
-import { ExternalLayoutEditorContainer } from './EditorContainers/ExternalLayoutEditorContainer';
-import { ResourcesEditorContainer } from './EditorContainers/ResourcesEditorContainer';
-import { SceneEditorContainer } from './EditorContainers/SceneEditorContainer';
+import { EventsEditorContainer } from '../EditorContainers/EventsEditorContainer';
+import { DebuggerEditorContainer } from '../EditorContainers/DebuggerEditorContainer';
+import { EventsFunctionsExtensionEditorContainer } from '../EditorContainers/EventsFunctionsExtensionEditorContainer';
+import { ExternalEventsEditorContainer } from '../EditorContainers/ExternalEventsEditorContainer';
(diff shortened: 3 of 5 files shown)


AlexandreSi:
Reminder of our discussion: Is it possible to put a tab at the first position (that's to say left to Home)?

It could be worth exploring how we implement `react-dnd` to see if it's possible to see if the pointer is over the left or right half of the first tab to position the separator and determine the index.

newIDE/app/src/UI/ClosableTabs.js
```diff
@@ -81,30 +81,26 @@ type ClosableTabsProps = {|
...
-          const tabItemContainerStyle = {
-            maxWidth: '100%', // Tabs should take all width
-            flexShrink: 0, // Tabs height should never be reduced
-            display: hideLabels ? 'none' : 'flex',
-            flexWrap: 'nowrap', // Single line of tab...
-            overflowX: 'auto', // ...scroll horizontally if needed
-            backgroundColor: muiTheme.closableTabs.containerBackgroundColor,
-          };
+export function ClosableTabs({ hideLabels, children }: ClosableTabsProps) {
+  return (
```
AlexandreSi, line 85: That can be reduced to `export const ClosableTabs = ({ hideLabels, children }: ClosableTabsProps) => (` and remove the return.

newIDE/app/src/MainFrame/EditorTabs/DropIndicator.js
```diff
@@ -0,0 +1,30 @@
...
+    marginLeft: '-1px',
+    height: '100%',
+    boxSizing: 'border-box',
+  },
+};
+
+export function ColumnDropIndicator() {
+  return (
+    <ThemeConsumer>
+      {gdevelopTheme => (
```
AlexandreSi, line 19: See component `ToolbarSeparator` to use the context that allows you to access the theme from the function. This way you won't have to use this `#18dcf2`

newIDE/app/src/MainFrame/EditorTabs/DraggableEditorTabs.js
```diff
@@ -0,0 +1,155 @@
...
+  hideLabels,
+  editorTabs,
+  onClickTab,
+  onCloseTab,
+  onCloseOtherTabs,
+  onCloseAll,
+  onTabActive,
+  onDropTab,
+}: DraggableClosableTabsProps) {
+  let draggedTabIndex: number | typeof undefined = undefined;
```
AlexandreSi, line 40: Prefer using the hook useState:
```js
const [draggedTabIndex, setDraggedTabIndex] = React.useState<?number>(null);
```

newIDE/app/src/MainFrame/EditorTabs/EditorTabsHandler.js
```diff
@@ -289,3 +289,24 @@ export const getEventsFunctionsExtensionEditor = (
...
+  toIndex: number
+): EditorTabsState => {
+  if (toIndex < fromIndex) toIndex += 1;
+  const currentEditorTabs = [...getEditors(editorTabsState)];
+  const movingTab = currentEditorTabs[fromIndex];
+  currentEditorTabs.splice(fromIndex, 1);
+  currentEditorTabs.splice(toIndex, 0, movingTab);
+
+  let currentTabIndex: number = getCurrentTabIndex(editorTabsState);
+  if (fromIndex === currentTabIndex) currentTabIndex = toIndex;
```
AlexandreSi, line 305: Can you give a name to those booleans to improve readability?
This one could be named
```
const theMovingTabIsTheCurrentTab = fromIndex === currentTabIndex
```

And, as a suggestion:

```
const movingTabWentFromLeftOfCurrentTabToRight = fromIndex < currentTabIndex && toIndex >= currentTabIndex
```

newIDE/app/src/MainFrame/EditorTabs/DraggableEditorTabs.js
```diff
@@ -0,0 +1,155 @@
...
+            return index !== 0;
+          }}
+          canDrop={() => true}
+          drop={() => onDrop(index)}
+        >
+          {({ connectDragSource, connectDropTarget, isOver, canDrop }) => {
+            // If on a touch screen, setting the whole item to be
+            // draggable would prevent scroll. Set the icon only to be
+            // draggable if the item is not selected. When selected,
+            // set the whole item to be draggable.
```
AlexandreSi, line 119: Maybe the hook `useLongTouch` can be of use here. I've not used it so far so I'm not sure, but if you want to improve the mobile user experience, it could be handy.

newIDE/app/src/MainFrame/EditorTabs/DraggableEditorTabs.js
fannieyan: Using state triggers a rerender when `onBeginDrag()` is called making us lose the drag, so it is not possible

newIDE/app/src/MainFrame/EditorTabs/DraggableEditorTabs.js
4ian, line 119: Outdated comment ;)

newIDE/app/src/MainFrame/EditorTabs/DropIndicator.js
```diff
@@ -0,0 +1,30 @@
+// @flow
+import * as React from 'react';
+import ThemeConsumer from '../../UI/Theme/ThemeConsumer';
+
+const styles = {
+  columnDropIndicator: {
+    borderRight: '1px solid #18dcf2',
+    borderLeft: '1px solid #18dcf2',
```
4ian, line 8: As the color will be overriden, might be worth putting a non descriptive one (black) or just use two CSS properties to avoid specifiying it entirely.

newIDE/app/src/MainFrame/EditorTabs/EditorTabsHandler.js
```diff
@@ -289,3 +289,24 @@ export const getEventsFunctionsExtensionEditor = (
 
   return null;
 };
+
+export const moveTabToPosition = (
+  editorTabsState: EditorTabsState,
+  fromIndex: number,
+  toIndex: number
+): EditorTabsState => {
+  if (toIndex < fromIndex) toIndex += 1;
```
ClementPasteau, line 298: I'd always avoid modifying parameters passed to a function, and instead create new ones.
Here, you could create a new index called `destinationIndex = toIndex < fromIndex ? toIndex + 1 : toIndex`

newIDE/app/src/MainFrame/index.js
```diff
@@ -1893,6 +1891,15 @@ const MainFrame = (props: Props) => {
     }));
   };
 
+  const _onDropEditorTab = (fromIndex: number, toIndex: number) => {
```
4ian, line 1894: While (wrongly) used in the file, don't prefix a local variable with an underscore. I think this is because this was adapted in the past from the class component which had underscore to hide private functions - but they are not private by design (because inside the component function, so not exposed by default on its interface, that's one advantage of React hooks/functional components)

newIDE/app/src/MainFrame/EditorTabs/EditorTabsHandler.js
```diff
@@ -289,3 +289,24 @@ export const getEventsFunctionsExtensionEditor = (
...
+  const currentEditorTabs = [...getEditors(editorTabsState)];
+  const movingTab = currentEditorTabs[fromIndex];
+  currentEditorTabs.splice(fromIndex, 1);
+  currentEditorTabs.splice(toIndex, 0, movingTab);
+
+  let currentTabIndex: number = getCurrentTabIndex(editorTabsState);
```
ClementPasteau, line 304: Is it required for you to cast this variable to a number? 
If so, then the function `getCurrentTabIndex` could be better typed to indicate it returns a number.

newIDE/app/src/MainFrame/index.js
```diff
@@ -2139,25 +2146,18 @@ const MainFrame = (props: Props) => {
...
+      <DraggableClosableTabs
+        hideLabels={!!props.integratedEditor}
+        editorTabs={state.editorTabs}
+        onClickTab={(id: number) => _onChangeEditorTab(id)}
+        onCloseTab={(editorTab: EditorTab) => _onCloseEditorTab(editorTab)}
+        onCloseOtherTabs={(editorTab: EditorTab) =>
+          _onCloseOtherEditorTabs(editorTab)
+        }
+        onCloseAll={_onCloseAllEditorTabs}
+        onTabActive={(editorTab: EditorTab) => _onEditorTabActive(editorTab)}
```
4ian, line 2158: `onTabActivated`

(sadly, _onEditorTabActive was not a great name, because `on` should be followed from a verb: `_onEditorTabActivated`)

newIDE/app/src/MainFrame/EditorTabs/EditorTabsHandler.js
ClementPasteau: And I would add a comment explaining why this +1 happens. 
Something like `// If the tab is being dragged backwards, then ...`

newIDE/app/src/UI/ClosableTabs.js
```diff
@@ -192,6 +188,9 @@ export function ClosableTab({
                 id={id ? `${id}-button` : undefined}
                 {...longTouchForContextMenuProps}
                 focusRipple
+                // If touch ripple is not disabled, dragged preview will use the size of the ripple
+                // and it will be too big
```
4ian, line 192: Uber nitpicking (good practice says that a comment ends by a period like a sentence):
```suggestion
                // If the touch ripple is not disabled, the dragged preview will use the size of the ripple
                // and it will be too big.
```

newIDE/app/src/MainFrame/EditorTabs/DraggableEditorTabs.js
```diff
@@ -0,0 +1,155 @@
...
+            id={`tab-${editorTab.key.replace(/\s/g, '-')}`}
+            active={isCurrentTab}
+            onClick={() => onClickTab(id)}
+            onClose={() => onCloseTab(editorTab)}
+            onCloseOthers={() => onCloseOtherTabs(editorTab)}
+            onCloseAll={onCloseAll}
+            onActivated={() => onTabActive(editorTab)}
+            closable={editorTab.closable}
+            onBeginDrag={() => {
+              draggedTabIndex = id;
```
ClementPasteau, line 60: as Alex mentions, you can use a useState hook, and here call `setDraggedTabIndex(id)`

newIDE/app/src/MainFrame/EditorTabs/DraggableEditorTabs.js
4ian: If this is just a local variable, then it's fine. Though:

```suggestion
  let draggedTabIndex: number | undefined = undefined;
```

or better:

```suggestion
  let draggedTabIndex: ?number = null;
```

In most cases, avoid "undefined". 

Undefined could means "there is not even a value here", like a parameter "was not even passed". It's possibly a sign of an error.
While null means "This value is defined, but it's defined to be null, surely because it's represents a state where it has no valid value". With `null`, we can clearly see that you meant what you wrote, why seeing "undefined" while debugging is much more worrying and hard to interpret because it might means that you passed something that is not even existing.

null make it clear that it's, well, a null value (and not a value that is not even defined)

newIDE/app/src/MainFrame/EditorTabs/DraggableEditorTabs.js
```diff
@@ -0,0 +1,155 @@
...
+              return editorTab;
+            }}
+            onDrop={toIndex => {
+              if (draggedTabIndex !== undefined) {
+                onDropTab(draggedTabIndex, id);
+                draggedTabIndex = undefined;
```
ClementPasteau, line 66: here too `setDraggedTabIndex(null)`

newIDE/app/src/MainFrame/EditorTabs/DraggableEditorTabs.js
```diff
@@ -0,0 +1,155 @@
...
+            onCloseOthers={() => onCloseOtherTabs(editorTab)}
+            onCloseAll={onCloseAll}
+            onActivated={() => onTabActive(editorTab)}
+            closable={editorTab.closable}
+            onBeginDrag={() => {
+              draggedTabIndex = id;
+              return editorTab;
+            }}
+            onDrop={toIndex => {
+              if (draggedTabIndex !== undefined) {
```
ClementPasteau, line 64: I think that you don't need your index to be undefined, you can initialise it with `null` and set it to `null` just below.
This check could become `if (typeof draggedTabIndex === number)` (to avoid being confused with 0)

newIDE/app/src/MainFrame/EditorTabs/DraggableEditorTabs.js
fannieyan: I agree, using long touch is a good idea. (And now that I tried it, allowing active tab to move on drag makes it become very slippery since you could try to scroll without paying too much attention to where you start scrolling...)

But long touch is already used to display the context menu, so I am not sure how to trigger one but not the other ? :/

newIDE/app/src/MainFrame/EditorTabs/DraggableEditorTabs.js
fannieyan: Using state triggers a rerender when onBeginDrag() is called making us lose the drag, so it is not possible :/

newIDE/app/src/MainFrame/EditorTabs/EditorTabsHandler.js
```diff
@@ -289,3 +289,24 @@ export const getEventsFunctionsExtensionEditor = (
...
+export const moveTabToPosition = (
+  editorTabsState: EditorTabsState,
+  fromIndex: number,
+  toIndex: number
+): EditorTabsState => {
+  if (toIndex < fromIndex) toIndex += 1;
+  const currentEditorTabs = [...getEditors(editorTabsState)];
+  const movingTab = currentEditorTabs[fromIndex];
+  currentEditorTabs.splice(fromIndex, 1);
+  currentEditorTabs.splice(toIndex, 0, movingTab);
```
D8H: I wonder if the `toIndex` increment should be outside of this method because even if there is no bug, the method probably not do what it says it does.

If we do this step by step:
TabDrop

* E from 4 to 2
```
0 1 2 3 4 5 6
A B C D E F G

  currentEditorTabs.splice(4, 1);

0 1 2 3 4 5 6
A B C D F G

  currentEditorTabs.splice(2 + 1, 0, movingTab);

0 1 2 3 4 5 6
A B C E D F G
```
It actually moved E from 4 to 3.

* C from 2 to 4

```
0 1 2 3 4 5 6
A B C D E F G

  currentEditorTabs.splice(2, 1);

0 1 2 3 4 5 6
A B D E F G

  currentEditorTabs.splice(4, 0, movingTab);

0 1 2 3 4 5 6
A B D E C F G
```

In the other case, I see no issue.

newIDE/app/src/MainFrame/EditorTabs/EditorTabsHandler.js
```diff
@@ -289,3 +289,24 @@ export const getEventsFunctionsExtensionEditor = (
...
+
+  let currentTabIndex: number = getCurrentTabIndex(editorTabsState);
+  if (fromIndex === currentTabIndex) currentTabIndex = toIndex;
+  else if (fromIndex < currentTabIndex && toIndex >= currentTabIndex)
+    currentTabIndex -= 1;
+  else if (fromIndex > currentTabIndex && toIndex <= currentTabIndex)
+    currentTabIndex += 1;
+
+  return { editors: currentEditorTabs, currentTab: currentTabIndex };
```
D8H, line 319: I think it's better to avoid to use the same variable for the condition and result.

newIDE/app/src/MainFrame/EditorTabs/DraggableEditorTabs.js
AlexandreSi: As discussed with @fannieyan, using a state variable here triggers a re-render of the component when a tab is picked and dragged and it messes with react-dnd. Maybe there's something wrong in the way the drag n drop was implemented because I can't see a way in which you can have a state variable but no re-render when it changes @4ian  @ClementPasteau

newIDE/app/src/MainFrame/EditorTabs/EditorTabsHandler.js
```diff
@@ -289,3 +289,42 @@ export const getEventsFunctionsExtensionEditor = (
 
   return null;
 };
+
+export const moveTabToTheRightOfHoveredTab = (
```
ClementPasteau, line 293: super clean 👍

ClementPasteau (approved):
nice job! I think it looks good to me! 🎉

AlexandreSi (approved):

D8H:
This feature will be really nice, I like to reorder tabs a certain way.

PR #3748, opened 2022-03-11
Add size actions and conditions for sprite and tiled sprite objects

fannieyan:
Closes #3530.

Height and width conditions for Sprite object
image

Size, Height and Width actions for Sprite object
image

Size action for Tiled Sprite Object
image

### DIFF
--- a/Core/GDCore/Extensions/Builtin/SpriteExtension/SpriteExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/SpriteExtension/SpriteExtension.cpp
@@ -235,7 +235,7 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsSpriteExtension(
       .UseStandardRelationalOperatorParameters("number")
       .MarkAsAdvanced();
 
-  obj.AddAction("ChangeSize",
+  obj.AddAction("SetSize",
                 _("Size"),
                 _("Change the size of an object."),
                 _("Change the size of _PARAM0_: set to _PARAM1_x_PARAM2_"),
--- a/Core/GDCore/Extensions/Builtin/SpriteExtension/SpriteExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/SpriteExtension/SpriteExtension.cpp
@@ -238,8 +238,7 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsSpriteExtension(
   obj.AddAction("ChangeSize",
                 _("Size"),
                 _("Change the size of an object."),
-                _("Change the size of _PARAM0_: _PARAM1_ (width) "
-                  ", _PARAM2_ (height)"),
+                _("Change the size of _PARAM0_: set to _PARAM1_x_PARAM2_"),
                 _("Size"),
                 "res/actions/scale24.png",
                 "res/actions/scale.png")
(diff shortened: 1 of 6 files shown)


ClementPasteau (approved):
👀  lgtm!

Core/GDCore/Extensions/Builtin/SpriteExtension/SpriteExtension.cpp
```diff
@@ -211,6 +223,32 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsSpriteExtension(
...
+
+      .AddParameter("object", _("Object"), "Sprite")
+      .UseStandardRelationalOperatorParameters("number")
+      .MarkAsAdvanced();
+
+  obj.AddAction("ChangeSize",
+                _("Size"),
+                _("Change the size of an object."),
+                _("Change the size of _PARAM0_: _PARAM1_ (width) "
+                  ", _PARAM2_ (height)"),
```
D8H, line 242: Parameter names usually follow this `name1: _PARAM1_, name2: _PARAM2_`, but for dimension a lighter syntax can be used.

The action sentences are not very consistent about this, I don't know which one to suggest:
* "Modify the size of a camera"
  "Change the size of camera _PARAM2_ of _PARAM1_ to _PARAM3_*_PARAM4_"

* "Change the size of the game window"
  "Set game window size to _PARAM1_x_PARAM2_ (also update game resolution: _PARAM3_)"

Maybe the cross character '✕' would be better?

GDJS/Runtime/spriteruntimeobject.ts
```diff
@@ -1180,10 +1191,21 @@ namespace gdjs {
...
+     * @param newWidth The new width of the object, in pixels.
+     * @param newHeight The new height of the object, in pixels.
+     */
+    setSize(newWidth: float, newHeight: float): void {
+      if (this._animationFrameDirty) {
+        this._updateAnimationFrame();
       }
+      this._setWidth(newWidth);
+      this._setHeight(newHeight);
     }
```
D8H, line 1209: I think it's not necessary to define 2 methods only to avoid 2 conditions. The frame will be updated only once either way.

D8H: Never mind what I said, I look it too quickly.

4ian: Agree that if we can avoid _setWidth/_setHeight it's better.

From the point of view of the reader, you don't immediately understand "why are there two methods with the same name, but one with an underscore, i.e: private? Is it public or private?"

I think here we can keep just a single pair: setWidth/setHeight and call them both

Extensions/TiledSpriteObject/Extension.cpp
```diff
@@ -135,6 +135,21 @@ void DeclareTiledSpriteObjectExtension(gd::PlatformExtension& extension) {
       .SetFunctionName("GetHeight")
       .SetIncludeFile("TiledSpriteObject/TiledSpriteObject.h");
 
+  obj.AddAction("Size",
```
4ian, line 138: Action should be named `SetXxx` (that's not enforced in the whole codebase, but I try to use this now for all actions).
Conditions and expressions can be named `Xxx` (without "Get").

Core/GDCore/Extensions/Builtin/SpriteExtension/SpriteExtension.cpp
4ian: I would go for `Change the size of _PARAM0_: set to _PARAM1_x_PARAM2_` because:
- it mimics the traditional "Change [subject] of [object]: [operator] [value]". (you can consider the size as being an attribute, and its value is a dimension, so a tuple, and you can consider the operator is forced to be "=").

GDJS/Runtime/spriteruntimeobject.ts
D8H: Maybe the code is small enough to be inlined?

PR #3751, opened 2022-03-13
Add D as shortcut to toggle disabled event

Bouh:
- [x] D allow to toggle disabled event.
- [x] Added in command palette.
- [x] Added in shortcut settings.

One problem persist, I can make it works when I edit the shortcut with another key, but the shortcut shown in the context menu isn't the key setup in settings.

I choose D because CTRL+D is used by the browser, and SHIFT+D is already used in GDevelop to add a sub-event.
Also CTRL+D is generally uised to duplicate in other software so D is the best choice and intuitive.

### DIFF
--- a/newIDE/app/src/CommandPalette/CommandsList.js
+++ b/newIDE/app/src/CommandPalette/CommandsList.js
@@ -52,6 +52,7 @@ export type CommandName =
   | 'ADD_STANDARD_EVENT'
   | 'ADD_SUBEVENT'
   | 'ADD_COMMENT_EVENT'
+  | 'TOGGLE_EVENT'
   | 'CHOOSE_AND_ADD_EVENT'
   | 'EVENTS_EDITOR_UNDO'
   | 'EVENTS_EDITOR_REDO'
@@ -279,6 +280,7 @@ const commandsList: { [CommandName]: CommandMetadata } = {
     displayText: t`Add a sub-event to the selected event`,
   },
   ADD_COMMENT_EVENT: { area: 'EVENTS', displayText: t`Add a comment` },
+  TOGGLE_EVENT: { area: 'EVENTS', displayText: t`Toggle disabled event` },
   CHOOSE_AND_ADD_EVENT: {
     area: 'EVENTS',
     displayText: t`Choose and add an event...`,
--- a/newIDE/app/src/CommandPalette/CommandsList.js
+++ b/newIDE/app/src/CommandPalette/CommandsList.js
@@ -52,7 +52,7 @@ export type CommandName =
   | 'ADD_STANDARD_EVENT'
   | 'ADD_SUBEVENT'
   | 'ADD_COMMENT_EVENT'
-  | 'TOGGLE_EVENT'
+  | 'TOGGLE_EVENT_DISABLED'
   | 'CHOOSE_AND_ADD_EVENT'
   | 'EVENTS_EDITOR_UNDO'
   | 'EVENTS_EDITOR_REDO'
@@ -280,7 +280,7 @@ const commandsList: { [CommandName]: CommandMetadata } = {
     displayText: t`Add a sub-event to the selected event`,
   },
   ADD_COMMENT_EVENT: { area: 'EVENTS', displayText: t`Add a comment` },
-  TOGGLE_EVENT: { area: 'EVENTS', displayText: t`Toggle disabled event` },
+  TOGGLE_EVENT_DISABLED: { area: 'EVENTS', displayText: t`Toggle disabled event` },
   CHOOSE_AND_ADD_EVENT: {
     area: 'EVENTS',
     displayText: t`Choose and add an event...`,
--- a/newIDE/app/src/CommandPalette/CommandsList.js
+++ b/newIDE/app/src/CommandPalette/CommandsList.js
@@ -280,7 +280,10 @@ const commandsList: { [CommandName]: CommandMetadata } = {
     displayText: t`Add a sub-event to the selected event`,
   },
   ADD_COMMENT_EVENT: { area: 'EVENTS', displayText: t`Add a comment` },
-  TOGGLE_EVENT_DISABLED: { area: 'EVENTS', displayText: t`Toggle disabled event` },
+  TOGGLE_EVENT_DISABLED: {
+    area: 'EVENTS',
+    displayText: t`Toggle disabled event`,
+  },
   CHOOSE_AND_ADD_EVENT: {
     area: 'EVENTS',
     displayText: t`Choose and add an event...`,
--- a/newIDE/app/src/EventsSheet/Toolbar.js
+++ b/newIDE/app/src/EventsSheet/Toolbar.js
@@ -48,11 +48,6 @@ export class Toolbar extends PureComponent<Props> {
             src="res/ribbon_default/commentaireadd32.png"
             tooltip={t`Add a comment`}
           />
-          <ToolbarIcon
-            onClick={this.props.onToggleDisabledEvent}
-            src="res/ribbon_default/subeventadd32.png"
-            tooltip={t`Toggle disabled selected event`}
-          />
           <ElementWithMenu
             element={
               <ToolbarIcon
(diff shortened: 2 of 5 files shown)


AlexandreSi:
Thanks for the suggestion! If Ctrl+D is known as to be the shortcut to duplicate things in other softwares, maybe it's better to have another shortcut?

Elairyx7301:
I love the idea to have a key or key combination with which you can toggle if an event is enabled or disabled.
I had the thought that toggling enabled/disabled may be so universal that could deserve it's own key instead of a key combination.

Now I don't have all key commands and combinations in mind that do something for default GDevelop or browsers but my first thought dragged me to the function keys like F2, F6 etc. (I know F5 starts preview and F11 toggles fullscreen).
- Maybe this should be another key than function keys.

What do you think about just one special key instead of a combination for it? This might make the workflow faster as well 🤷‍♂️

Bouh:
I would like very much that we do not block on a futile point which is the attribution of a key which will come to be revised in the future to have a coherent set of keyboard shortcuts.
I don't understand why D or T key doesn't fit you, it works very well, It doesn't matter which key you want if we can customize it.

**I need a little help as stated in the 1st message for:**
Have in the right click menu, the shortcut configured in the options and not the hardcoded one as it is currently.

Elairyx7301:
Sorry if I transferred my message inefficiently. I don't have anything about D, T or any other key being used for it. I also don't disagree with the default being a key combination, since - as you said - it can be changed to whatever you want.

That's perfectly fine, I just thought about making it the quickest default and wanted to ask for opinion on it :)

newIDE/app/src/EventsSheet/index.js
```diff
@@ -638,6 +639,7 @@ export class EventsSheetComponentWithoutHandle extends React.Component<
       label: i18n._(t`Toggle disabled`),
       click: () => this.toggleDisabled(),
       enabled: this._selectionCanToggleDisabled(),
+      accelerator: 'D',
```
D8H, line 642: I see something like `accelerator: getElectronAccelerator(shortcutMap['OPEN_PROJECT']),` for customisable shortcuts.

Bouh: A hook is used with `shortcutMap` I didn't see how to use it without making a whole refacto of this class into a functional component.
Is there a converter for it? It seems tricky to do manually.

newIDE/app/src/EventsSheet/Toolbar.js
```diff
@@ -47,6 +48,11 @@ export class Toolbar extends PureComponent<Props> {
             src="res/ribbon_default/commentaireadd32.png"
             tooltip={t`Add a comment`}
           />
+          <ToolbarIcon
+            onClick={this.props.onToggleDisabledEvent}
+            src="res/ribbon_default/subeventadd32.png"
+            tooltip={t`Toggle disabled selected event`}
+          />
```
AlexandreSi, line 55: These lines add a new button in the event sheet toolbar, I'm not sure we want to have yet another button in this toolbar, with the same icon as for "Add a sub event"

newIDE/app/src/CommandPalette/CommandsList.js
```diff
@@ -52,6 +52,7 @@ export type CommandName =
   | 'ADD_STANDARD_EVENT'
   | 'ADD_SUBEVENT'
   | 'ADD_COMMENT_EVENT'
+  | 'TOGGLE_EVENT'
```
AlexandreSi, line 55: Can I suggest that you use `TOGGLE_EVENT_DISABLED` to be more clear?

newIDE/app/src/EventsSheet/index.js
AlexandreSi: To make it work, you can do:

```js
_buildEventContextMenu = (i18n: I18nType, userShortcutMap: ShortcutMap) => [
  ...,
  {
      label: i18n._(t`Toggle disabled`),
      click: () => this.toggleDisabled(),
      enabled: this._selectionCanToggleDisabled(),
      accelerator: this.props.preferences.values.userShortcutMap['TOGGLE_EVENT_DISABLED'].replace('Key', ''),
  }
```

newIDE/app/src/EventsSheet/ToolbarCommands.js
```diff
@@ -38,6 +39,10 @@ const ToolbarCommands = (props: Props) => {
     handler: props.onAddCommentEvent,
   });
 
+  useCommand('TOGGLE_EVENT', true, {
+    handler: props.onToggleDisabledEvent,
```
AlexandreSi, line 43: Here, you give the possibility to toggle disabled any events though not every event can be disabled (comment events for example).

The method `toggleDisabled` of the component `EventsSheet` should toggle an event only if it is executable (see method `_selectionCanToggleDisabled`):


```js
  toggleDisabled = () => {
    getSelectedEvents(this.state.selection).forEach(event =>
      if (event.isExecutable()) event.setDisabled(!event.isDisabled())
    );
    this._saveChangesToHistory(() => {
      if (this._eventsTree) this._eventsTree.forceEventsUpdate();
    });
  };
```

But doing that, we can see that we could add an entry to the history even though no event were changed. So we could add:

```js
  toggleDisabled = () => {
    let shouldBeSaved = false;
    getSelectedEvents(this.state.selection).forEach(event =>
      if (event.isExecutable()) {
        event.setDisabled(!event.isDisabled())
        shouldBeSaved = true;
      }
    );
    if (shouldBeSaved) {
      this._saveChangesToHistory(() => {
        if (this._eventsTree) this._eventsTree.forceEventsUpdate();
      });
    }
  };
```

newIDE/app/src/EventsSheet/index.js
AlexandreSi: I just realized that we have a method `getShortcutDisplayName` so, it can be:

```js
    ...
    accelerator: getShortcutDisplayName(this.props.preferences.values.userShortcutMap['TOGGLE_EVENT_DISABLED'])
```

AlexandreSi (changes requested):
I forgot this shortcuts can be changed in the preferences so sorry for the point about the letter.
The PR needs a few changes (details in the comments) and away it goes!

Bouh:
Thanks for the review guys :D
I addressed the remaining points and it works as expected to me.
Ready for the last check and go 🚀

newIDE/app/src/EventsSheet/ToolbarCommands.js
```diff
@@ -38,6 +39,10 @@ const ToolbarCommands = (props: Props) => {
     handler: props.onAddCommentEvent,
   });
 
+  useCommand('TOGGLE_EVENT_DISABLED', true, {
```
ClementPasteau, line 42: here, instead of `true`, we should know when the command can be used or not.
I'd suggest passing down a prop `canToggleEventDisabled` coming from the component above, corresponding to `hasEventSelected(this.state.selection)`

ClementPasteau (commented):
why remove it from the toolbar? I think it's a good idea to have it next to the "remove selected events" icon, but instead of a "remove" icon, it could be a "pause" icon ?

Something like this (but it's not transparent)
Screenshot 2022-03-16 at 18 29 12

Bouh:
> why remove it from the toolbar? I think it's a good idea to have it next to the "remove selected events" icon, but instead of a "remove" icon, it could be a "pause" icon ?
> 
> Something like this (but it's not transparent) Screenshot 2022-03-16 at 18 29 12

It's looks like a pause debugger to me :s 
I think we can do without this button, I don't wanna overload the toolbar with it.

ClementPasteau (commented):
alright! LGTM, but I'll let @AlexandreSi approve

newIDE/app/src/EventsSheet/Toolbar.js
```diff
@@ -15,6 +15,7 @@ type Props = {|
   onAddCommentEvent: () => void,
   allEventsMetadata: Array<EventMetadata>,
   onAddEvent: (eventType: string) => Array<gdBaseEvent>,
+  onToggleDisabledEvent: () => void,
```
ClementPasteau, line 18: I think you forgot `canToggleEventDisabled: boolean` here

AlexandreSi (approved):
Tested and approved!

PR #3754, opened 2022-03-15
Sprites are no longer bouncing of one pixel when the camera is moving

D8H:
With pixel perfect mode activated, sprites are no longer bouncing of one pixel when the camera is moving.

### Broken one (notice how the character seems to move over grasses)
https://liluo.io/instant-builds/5eccedb0-d30f-4b93-a3ac-bf57e2a0c4e5?dev=true

### Fixed one
https://liluo.io/instant-builds/a950d977-35c9-4137-b9f7-325416993a02?dev=true

## Pixel rounding with float camera position
The grid is the pixels on screen and the small arrows are the sprite rounding

FloatCamera1

When the camera moves from less than one pixel, it changes the direction of the character rounding (in orange), but the platform (in green) stays at the same location on screen. Thus, the character appears to move one pixel to the left in the referential of the platform and it will dance from one pixel right and left.

FloatCamera2

### DIFF
--- a/GDJS/Runtime/pixi-renderers/layer-pixi-renderer.ts
+++ b/GDJS/Runtime/pixi-renderers/layer-pixi-renderer.ts
@@ -13,7 +13,7 @@ namespace gdjs {
   export class LayerPixiRenderer {
     _pixiContainer: PIXI.Container;
 
-    _layer: any;
+    _layer: gdjs.Layer;
     _renderTexture: PIXI.RenderTexture | null = null;
     _lightingSprite: PIXI.Sprite | null = null;
     _runtimeSceneRenderer: any;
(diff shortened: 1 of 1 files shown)


4ian:
Awesome! We should document somewhere that for pixel perfect rendering, you should strive to a round factor for the camera. Possibly in the form of hints in the app.

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.ts
```diff
@@ -72,10 +72,26 @@ namespace gdjs {
...
+        // Otherwise the camera position fractional part is added to
+        // the sprite one and it changes in which direction sprites are rounded.
+        // It makes sprites rounding inconsistent with each other
+        // and they seems to move on pixel left and right.
+        this._pixiContainer.position.x = Math.round(
+          this._pixiContainer.position.x
+        );
+        this._pixiContainer.position.y = Math.round(
+          this._pixiContainer.position.y
+        );
```
D8H, line 93: Actually, PIXI uses `floor`. We will have to use floor too. This will avoid flickering when centering the camera on an object in the most common cases: when the frame and the sprite dimension parity is the same.

We could then suggest game creators to use even dimension for screen and sprites (or integer center) when using pixel rounding.

4ian: So the conclusion is that we should use floor here and make a note/comment explaining that this is to use the same rounding as PixiJS (can you explain at which level/for which primitive this rounding is used in PIxiJS)?
Thanks!

PR #3755, opened 2022-03-15
Fix text input wrongly staying visible even if hidden or layer visibility turned off

4ian:


GDJS/Runtime/runtimescene.ts
```diff
@@ -681,11 +681,16 @@ namespace gdjs {
               this._runtimeGame
                 .getEffectsManager()
                 .updatePreRender(object.getRendererEffects(), object);
+
+              // Perform pre-render update only if the object is visible
+              // (including if there is no visibility AABB returned previously).
+              object.updatePreRender(this);
```
4ian, line 687: Most objects should not have their updatePreRender called if they are not visible (either hidden or far from the camera), which make sense.
@D8H to be sure, can you verify this won't impact the changes you made on the shape painter object? I know there were some X/Y position dirty flags or transformation dirty flag and I want to be sure not running updatePreRender is not an issue.

D8H: I did a manual test with a slider going outside of the screen. `Slider.X()` and `Slider.BoundingBoxLeft()` are still refreshed outside of the screen and looking at the code `updatePreRender` only calls `updatePositionIfNeeded` which is also called by `updateHitBoxes`.
ShapePainter position and collision seems to work fine when moving outside of screen.

D8H: Just out of curiosity, did you made the change only for efficiency or to avoid a side effect?

4ian: > ShapePainter position and collision seems to work fine when moving outside of screen.

Excellent!

> Just out of curiosity, did you made the change only for efficiency or to avoid a side effect?

For efficiency :) By which I mean:
- It should not matter if the engine call "too much" updatePreRender (i.e: even if it call it on instances outside of the screen that won't be useful for the rendering, it should not impact the game).
- but when I read the code, I realised that we were calling this on all instances at every frame, which is sub optimal especially considered that we try to do some culling. 

So if ShapePainter was wrongly behaving with this change, the fault would have been on the ShapePainter object - not on this optimisation (because the engine is right not to call `updatePreRender`). 
But because I've been still looking at objects using `updatePreRender` to be sure there was no regression/no bad assumptions made by these objects, I wanted to get your confirmation that it looks alright :) 

Note that in the future we could store the instances in a RTree/spatial data structure and avoiding iterating on all instances entirely at every frame, we could just ask the RTree/spatial data structure for the instances in the camera area (this will avoid iterating on N objects, while there are only M visible on screen, especially useful if N is becoming very large).

PR #3759, opened 2022-03-17
Add built-in leaderboard feature (actions, conditions, administration)

AlexandreSi:
Adds a tab in the game dashboard to manage the future games leaderboards.

- create a leaderboard:
  demo_leaderboard_admin_create

- change leaderboards settings and delete an entry:
  demo_leaderboard_admin_entry_deletion

- delete a leaderboard:
  demo_leaderboard_admin_leaderboard_deletion

To see how api calls errored are handled and displayed, you can check out the [story of the component here](https://gdevelop-storybook.s3.amazonaws.com/give-me-leaderboard/latest/index.html?path=/story/leaderboardadmin--with-errors)

---

### TODO:
- [x] Add pagination for the entries
- [x] Try to fix the error in the console about unique keys in I18n provider
- [x] Fix an odd behavior when switching from a leaderboard configured to display only best entries to another configured to display all entries (race condition between 2 requests, something to change in the interactions between the provider and the admin component)


newIDE/app/src/UI/SelectField.js
```diff
@@ -29,6 +29,7 @@ type Props = {|
     flex?: 1,
     width?: 'auto',
   },
+  inputStyle?: {| fontSize: '0.875em' |},
```
4ian, line 32: Why was this change needed?

newIDE/app/src/UI/EmptyPlaceholder.js
```diff
@@ -43,7 +45,8 @@ export const EmptyPlaceholder = (props: Props) => (
             label={props.actionLabel}
```
4ian, line 45: Consider if you could be using `LeftLoader` to display a loader next to a button (that's what we do, but maybe a loader inside the button is better)

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
```diff
@@ -0,0 +1,835 @@
+// @flow
+import React from 'react';
```
4ian, line 2: Prefer `import * as React from 'react'` to benefit from `React.Node` types

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
```diff
@@ -0,0 +1,835 @@
...
+    | 'entriesFetching'
+    | 'entryDeletion'
+    | 'leaderboardsFetching'
+    | 'leaderboardNameUpdate'
+    | 'leaderboardSortUpdate'
+    | 'leaderboardPlayerUnicityDisplayChoiceUpdate'
+    | 'leaderboardCreation'
+    | 'leaderboardReset'
+    | 'leaderboardDeletion',
+  message: React$Node,
```
4ian, line 79: everywhere: `React$Node` => `React.Node`

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
```diff
@@ -0,0 +1,835 @@
...
+
+  const _listLeaderboards = React.useCallback(
+    () => {
+      const fetchAndHandleError = async () => {
+        disableActions(true);
+        setApiError(null);
+        try {
+          await listLeaderboards();
+        } catch (err) {
+          console.error(err);
```
4ian, line 194: Prefer to always give some context, useful to understand errors when reading the console: 
```suggestion
          console.error(`Error while listing leaderboards:`, err);
```

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
```diff
@@ -0,0 +1,835 @@
...
+    [disableActions, listLeaderboards]
+  );
+
+  const _fetchLeaderboardEntries = async () => {
+    disableActions(true);
+    setApiError(null);
+    try {
+      await fetchLeaderboardEntries();
+    } catch (err) {
+      console.error(err);
```
4ian, line 219: (same for every logged error)

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
```diff
@@ -0,0 +1,835 @@
...
+    setApiError(null);
+    try {
+      await resetLeaderboard();
+    } catch (err) {
+      console.error(err);
+      setApiError({
+        action: 'leaderboardReset',
+        message: (
+          <Trans>
+            An error ocurred when resetting the leaderboard, please close the
```
4ian, line 277:
```suggestion
            An error occurred when resetting the leaderboard, please close the
```

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
```diff
@@ -0,0 +1,835 @@
...
+    | 'leaderboardSortUpdate'
+    | 'leaderboardPlayerUnicityDisplayChoiceUpdate'
+    | 'leaderboardCreation'
+    | 'leaderboardReset'
+    | 'leaderboardDeletion',
+  message: React$Node,
+  itemId?: string,
+|};
+
+const WrappedError = ({ children }: { children: React$Node }) => (
```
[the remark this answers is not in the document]
4ian: (could have been named CenteredError)?

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
```diff
@@ -0,0 +1,835 @@
...
+            onKeyPress={event => {
+              if (event.key === 'Enter' && !isRequestPending) {
+                _updateLeaderboard(i18n, { name: newName });
+              }
+            }}
+            disabled={isRequestPending}
+          />
+          {!isRequestPending && (
+            <IconButton
+              style={{ padding: 0, marginLeft: 5 }}
```
4ian, line 475: See if you can add a `<Spacer />` instead of marginLeft: 5.

`Spacer` helps to keep the margin consistent in the whole app.

newIDE/app/src/UI/Dialog/index.js
```diff
@@ -24,6 +24,10 @@ const styles = {
     display: 'flex',
     flexDirection: 'row',
   },
+  flexColumnBody: {
+    display: 'flex',
+    flexDirection: 'column',
+  },
```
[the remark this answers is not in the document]
AlexandreSi: It was not working, I'll check why

newIDE/app/src/UI/SelectField.js
AlexandreSi: There is no way in the `SelectField` component to change the input font size, as done in the `TextField` component [here](https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/UI/TextField.js#L99-L103). And `0.875em` corresponds to the `body2` MUI variant

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
AlexandreSi: Oh yes thanks!

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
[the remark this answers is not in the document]
AlexandreSi: Good suggestion!

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
AlexandreSi: Will do!

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
AlexandreSi: Yes true, I'll change that

newIDE/app/src/UI/Dialog/index.js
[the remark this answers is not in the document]
AlexandreSi: Only the component `NewInstructionEditorDialog` uses the `flexRowBody` props and it seems to work fine when not set

newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardEntriesTable.js
```diff
@@ -0,0 +1,155 @@
...
+import PlaceholderLoader from '../../UI/PlaceholderLoader';
+import Text from '../../UI/Text';
+import { textEllipsisStyle } from '../../UI/TextEllipsis';
+import { type LeaderboardDisplayData } from '../../Utils/GDevelopServices/Play';
+
+type Props = {|
+  entries: ?Array<LeaderboardDisplayData>,
+  onDeleteEntry: (entryId: string) => Promise<void>,
+  disableActions: boolean,
+  erroredEntry?: {| entryId: string, message: React$Node |},
```
4ian, line 29: React$Node => React.Node

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
```diff
@@ -0,0 +1,820 @@
...
+import Label from '@material-ui/icons/Label';
+import Fingerprint from '@material-ui/icons/Fingerprint';
+import Update from '@material-ui/icons/Update';
+import Today from '@material-ui/icons/Today';
+import Sort from '@material-ui/icons/Sort';
+import PeopleAlt from '@material-ui/icons/PeopleAlt';
+import SwapVertical from '@material-ui/icons/SwapVert';
+import Refresh from '@material-ui/icons/Refresh';
+import Delete from '@material-ui/icons/Delete';
+import {
```
4ian, line 23: I think that for code splitting, it's recommended to import using `@material-ui/core/Xxx`. See https://v4.mui.com/guides/minimizing-bundle-size/

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
```diff
@@ -0,0 +1,820 @@
...
+      );
+      return;
+    }
+    disableActions(true);
+    setApiError(null);
+    try {
+      await updateLeaderboard(payload);
+      if (payload.name) setIsEditingName(false);
+    } catch (err) {
+      console.error(`An error occurred when updating leaderboard: ${err}`);
```
4ian, line 157: For errors in the Browser console, prefer to pass the error as a separate object: console.error("An error [...]", err);

so that it's not stringified and keep containing all the useful info.

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
```diff
@@ -0,0 +1,820 @@
...
+
+  const _listLeaderboards = React.useCallback(
+    () => {
+      const fetchAndHandleError = async () => {
+        disableActions(true);
+        setApiError(null);
+        try {
+          await listLeaderboards();
+        } catch (err) {
+          console.error(`An error occurred when fetching leaderboards: ${err}`);
```
4ian, line 194: (ditto and everywher)

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
```diff
@@ -0,0 +1,820 @@
...
+          actionLabel={<Trans>Create a leaderboard</Trans>}
+          onAdd={() => {
+            _createLeaderboard();
+          }}
+          isLoading={isRequestPending}
+        />
+      </Line>
+    );
+
+  const leaderboardDescription = (
```
4ian, line 429: nit: this is a function, so it should begin with a verb (render, build, create...)

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
```diff
@@ -0,0 +1,820 @@
...
+          <TextField
+            ref={newNameTextFieldRef}
+            margin="none"
+            style={{ width: 125, fontSize: 14 }}
+            maxLength={50}
+            value={newName}
+            errorText={newNameError}
+            onChange={(e, text) => setNewName(text)}
+            onKeyPress={event => {
+              if (event.key === 'Enter' && !isRequestPending) {
```
4ian, line 447: Replace all event.key === xxx with `shouldValidate` or similar

newIDE/app/src/Leaderboard/LeaderboardProvider.js
```diff
@@ -0,0 +1,406 @@
...
+    await doDeleteLeaderboardEntry(
+      authenticatedUser,
+      gameId,
+      currentLeaderboardId,
+      entryId
+    );
+    fetchEntries({ uri: pageIndex > 0 ? mapPageIndexToUri[pageIndex] : null });
+  };
+
+  // --- Navigation ---
```
4ian, line 324: Nitpicking: if you feel like you need to write a comment to separate things, it means that your code is becoming too long (though I agree it's hard to reduce).
What about making here a custom hook "useLeaderboardEntriesNavigation" that returns navigateToNextPage (and previous and first page)?

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
```diff
@@ -0,0 +1,820 @@
...
+                      <MUITextField
+                        {...params}
+                        margin="dense"
+                        label={<Trans>Leaderboard name</Trans>}
+                        variant="filled"
+                      />
+                    )}
+                  />
+                  <IconButton
+                    onClick={_createLeaderboard}
```
4ian, line 687: I'm not a big fan of naming functions with an underscore - they are not private or anything, they are just local variables. Can you name these "onCreateLeaderboard" to show that it's what is called to react to a semantic event from the user?

4ian (approved):

newIDE/app/src/Utils/GDevelopServices/Play.js
```diff
@@ -0,0 +1,239 @@
...
+  gameId: string,
+  leaderboardId: string,
+  options: {| pageSize: number, onlyBestEntry: boolean, forceUri: ?string |}
+): Promise<{|
+  entries: LeaderboardEntry[] | LeaderboardExtremePlayerScore[],
+  nextPageUri: ?string,
+|}> => {
+  const uri =
+    options.forceUri || `/game/${gameId}/leaderboard/${leaderboardId}/entries`;
+  // $FlowFixMe
```
ClementPasteau, line 107: what's that for?

newIDE/app/src/Leaderboard/LeaderboardProvider.js
```diff
@@ -0,0 +1,321 @@
...
+          pageSize,
+          onlyBestEntry: displayOnlyBestEntry,
+        }
+      );
+      if (!fetchedEntries) return;
+
+      let entriesToDisplay: LeaderboardDisplayData[] = [];
+      if (displayOnlyBestEntry) {
+        entriesToDisplay = fetchedEntries.map(entry =>
+          // $FlowIgnore
```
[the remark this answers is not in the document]
AlexandreSi: Nope, conflict in my mind with `// @ts-ignore`

newIDE/app/src/Leaderboard/LeaderboardProvider.js
```diff
@@ -0,0 +1,406 @@
...
+      });
+    },
+    [gameId, authenticatedUser]
+  );
+
+  const fetchEntries = React.useCallback(
+    async (options?: {| uri?: ?string |}) => {
+      if (!currentLeaderboardId) return;
+
+      let uriToUse = null;
```
ClementPasteau, line 221: you don't need a let and can probably do `const utiToUse = options && options.uri ? options.uri || null` or something like that

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
AlexandreSi: True, it feels better with the `on` instead of the underscores.

newIDE/app/src/GameDashboard/GameDetailsDialog.js
```diff
@@ -208,15 +217,29 @@ export const GameDetailsDialog = ({
...
     } catch (error) {
       console.error('Unable to delete the game:', error);
+      if (
+        error.response &&
+        error.response.data &&
+        error.response.data.code === 'game-deletion/leaderboards-exist'
+      ) {
+        setGameUnregisterErrorText(
+          i18n._(
+            t`You cannot unregister a game that has active leaderboards. To delete them, go in tab Leaderboards, and delete them one by one.`
```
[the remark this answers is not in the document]
AlexandreSi: Thanks!

newIDE/app/src/GameDashboard/GameDetailsDialog.js
```diff
@@ -268,11 +291,16 @@ export const GameDetailsDialog = ({
...
+          fullHeight={currentTab === 'leaderboards'}
+          onRequestClose={() => {
+            if (!isLoading) onClose();
+          }}
           maxWidth="md"
           actions={[
             <FlatButton
-              label={<Trans>Close</Trans>}
+              label={isLoading ? null : <Trans>Close</Trans>}
+              icon={isLoading ? <CircularProgress size={20} /> : null}
```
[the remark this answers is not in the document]
AlexandreSi: Yeah you're right, the more I used it and saw this loader, the more it felt weird.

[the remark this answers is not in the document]
4ian: It's true that we can just disable it. The loading is shown by another component in the dialog, near where the action really happened (i.e: the click)

newIDE/app/src/UI/PlaceholderError.js
```diff
@@ -4,27 +4,34 @@ import * as React from 'react';
 import { Column, Line } from './Grid';
 import RaisedButton from './RaisedButton';
 import EmptyMessage from './EmptyMessage';
+import GDevelopThemeContext from './Theme/ThemeContext';
 
 type Props = {|
   children: React.Node,
   onRetry?: () => void | Promise<void>,
+  kind?: 'error' | 'valid' | 'warning',
```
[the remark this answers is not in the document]
AlexandreSi: I did not think it through, I just copied the possible values in the theme. Maybe it could be useful to display an error, and when it is resolved, it is then displayed in green saying "Error is fixed".

Maybe it's far-fetched and I can remove this option

newIDE/app/src/UI/PlaceholderError.js
```diff
@@ -4,27 +4,34 @@ import * as React from 'react';
...
+    <Column expand alignItems="center">
+      <EmptyMessage style={kind ? { color: theme.message[kind] } : undefined}>
+        {children}
+      </EmptyMessage>
+      <Line>
+        {onRetry && (
+          <RaisedButton
+            primary
+            label={<Trans>Retry</Trans>}
+            onClick={() => {
```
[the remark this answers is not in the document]
AlexandreSi: Yep, will do, thanks!

newIDE/app/src/Utils/GDevelopServices/Play.js
AlexandreSi: This is because of the fixed-in-0.159-version-of-flow bug: https://github.com/facebook/flow/issues/5294.

The fact that I use the response below in the `extractNextPageUriFromLinkHeader` method make it fail

newIDE/app/src/Leaderboard/LeaderboardProvider.js
```diff
@@ -0,0 +1,406 @@
...
+        ...state,
+        displayOnlyBestEntry: shouldDisplayOnlyBestEntries(
+          newLeaderboardsByIds[leaderboardsIds[0]]
+        ),
+        leaderboardsByIds: newLeaderboardsByIds,
+        currentLeaderboard: newLeaderboardsByIds[leaderboardsIds[0]],
+        currentLeaderboardId: leaderboardsIds[0],
+      };
+    default: {
+      throw new Error(`Unhandled type: ${action.type}`);
```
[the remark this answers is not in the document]
AlexandreSi: True, it's a remain from a copy-paste, maybe for the dev experience, to make sure we see that the type is not handled.
But with typing, we already have a few warnings.

newIDE/app/src/Leaderboard/LeaderboardProvider.js
```diff
@@ -0,0 +1,406 @@
...
+      fetchEntries();
+    },
+    [currentLeaderboardId, displayOnlyBestEntry, fetchEntries]
+  );
+
+  return (
+    <LeaderboardContext.Provider
+      value={{
+        leaderboards: !!leaderboardsByIds
+          ? // $FlowFixMe
```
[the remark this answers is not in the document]
AlexandreSi: Seems like `Object.values(object)` returns `Array<mixed>`  instead of `Array<U>` where `object: {[key: string]: U}`

newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardEntriesTable.js
```diff
@@ -0,0 +1,155 @@
...
+              <Text size="body2">
+                <Trans>No entries</Trans>
+              </Text>
+            </Column>
+          ) : (
+            <Line noMargin justifyContent="flex-end">
+              <IconButton
+                tooltip={t`Go to first page`}
+                disabled={!navigation.goToFirstPage}
+                onClick={navigation.goToFirstPage || (() => {})}
```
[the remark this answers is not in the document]
AlexandreSi: Yes, it feels better

Bouh:
On a pure functional review:

- Hit `Refresh` or `Retry` ?
image

- Two different select list in same component 😬 
 image

- What is for this id?
image

AlexandreSi:
> Hit Refresh or Retry ?

I guess I will remove the advise, seeing the Retry button should be straight forward enough

> Two different select list in same component 😬

I am not sure about what you're implying: 2 select is forbidden ? or because they are open at the same time?

> What is for this id?
> Ellipsis on text is active even if the id is completely visible

That's my bad, the id should be an uuid, say 5 strings separated by hyphens. I'll change it and you'll see.

Regarding the tooltip: I put tooltips everywhere, even if the text is not partially hidden. There's no easy way to know if CSS has made the text overflow in order to know if we should add the tooltip or not.

ClementPasteau (approved):
lgtm after the final small feedbacks!

Should we merge this PR only when the actions/conditions/expressions for leaderboards are available? Otherwise it's pretty useless :)

newIDE/app/src/GameDashboard/GameDetailsDialog.js
```diff
@@ -208,15 +217,29 @@ export const GameDetailsDialog = ({
...
       await deleteGame(getAuthorizationHeader, id, game.id);
       onGameDeleted();
     } catch (error) {
       console.error('Unable to delete the game:', error);
+      if (
+        error.response &&
+        error.response.data &&
+        error.response.data.code === 'game-deletion/leaderboards-exist'
+      ) {
+        setGameUnregisterErrorText(
```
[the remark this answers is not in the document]
AlexandreSi: This is an expected error, so no need to console log something. If a user tells us that the error message is displayed, we can only tell them to do what the message says.

4ian:
Agree on merging this PR once we have the associated extension ready to merge too :) 

(Alas, GitHub does not allow to "stack PRs", and rebasing can be a bit hard)

newIDE/app/src/UI/SelectField.js
AlexandreSi: I'll change for 14 so that it's consistent with `TextField`

Bouh:
> > Two different select list in same component 😬
> 
> I am not sure about what you're implying: 2 select is forbidden ? or because they are open at the same time?

There are visually different for no reason.

AlexandreSi:
> > > Two different select list in same component 😬
> > 
> > 
> > I am not sure about what you're implying: 2 select is forbidden ? or because they are open at the same time?
> 
> There are visually different for no reason.

True! I didn't think of that.
The first is a bit different as it is an autocomplete (useful when you have 25 leaderboards and don't want to scroll).
The second one only has 3 possibilities so it felt a bit overkill to use the same component.

But I just noticed I'm maybe not using the best component, I'll have a look.

AlexandreSi:
Thanks for spotting it! Here's what we have now.

image

4ian:
I'm challenging the choice to revert to a Material-UI Select. It's a choice we already made in the UI to prioritise the native controls of the platform whenever we can. By choosing the Select, we are maybe more "consistent" in this dialog, but less consistent with the rest of the interface.

So I think we identified the wrong problem: it was not the select with 3 choices that was wrong (that's a select, we use what the native platform is providing, we can't do better than this - especially on mobile). It's the autocomplete that is "a bit weird" (because it's used as a "combobox": a select where you can type).

So either we choose to:
- accept the "weird" autocomplete acting as a select
- or go back to a native select for it (but we lose the ability to search)

but let's not use a material-ui select while we already have a SelectField in the `UI` folder.

AlexandreSi:
TODO: 
- Clear leaderboard context when switching projects
- Add help links in dialog buttons

4ian (approved):

PR #3766, opened 2022-03-21
Allow users to save scores in leaderboards

fannieyan:
- [x] Create action to save scores in given leaderboard
- [x] Create condition to check if last save failed
- [x] Create expression to get last save's status code
- [x] Create new field with leaderboard names (instead of ID)
- [x] Add button to new field that opens leaderboard admin to create a new one

Storybook is available to test Leaderboard field => look for LeaderboardIdField

Action
action
Action UI
Inline action

EDIT : updated button placement :  
image
image


Extensions/Leaderboards/leaderboardstools.ts
```diff
@@ -0,0 +1,75 @@
...
+  export namespace evtTools {
+    export namespace leaderboards {
+      let _scoreLastSentAt: number | null = null;
+      let _lastScore: number;
+      let _lastPlayerName: string;
+      let _lastErrorCode: number;
+
+      export const setPlayerScore = function (
+        leaderboardId: string,
+        score: gdjs.Variable,
```
4ian, line 12: (The score and player name should be passed as `string`, though double check the parameters defined in the extension declaration)

fannieyan: Alright !
The score too ? Even though it should be sent as a number to our API ?

4ian: Ah you're right, a `float` is fine for the score (`float` is a type alias for `number` in TypeScript). So in the parameters of the extension, this will be an "expression" (which should have been just named "number", but that's how it's called).

newIDE/app/src/EventsSheet/ParameterFields/LeaderboardIdField.js
```diff
@@ -0,0 +1,144 @@
...
+        )
+      }
+      renderButton={style => (
+        <>
+          <RaisedButton
+            icon={<OpenInNew />}
+            disabled={!leaderboards}
+            primary
+            style={style}
+            onClick={() => setIsTextInput(!isTextInput)}
```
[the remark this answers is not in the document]
AlexandreSi: As discussed together, this comment is not relevant

newIDE/app/src/EventsSheet/ParameterFields/LeaderboardIdField.js
```diff
@@ -0,0 +1,147 @@
...
+import { type Leaderboard } from '../../Utils/GDevelopServices/Play';
+import LeaderboardContext from '../../Leaderboard/LeaderboardContext';
+import OpenInNew from '@material-ui/icons/OpenInNew';
+import { t } from '@lingui/macro';
+import Toggle from '../../UI/Toggle';
+import { LargeSpacer } from '../../UI/Grid';
+import LeaderboardDialog from '../../SceneEditor/LeaderboardDialog';
+import GenericExpressionField from './GenericExpressionField';
+
+const getLeaderboardNameByValue = (
```
AlexandreSi, line 19: It's not clear what is a value here: is it a leaderboard id? name? quotedId?

newIDE/app/src/EventsSheet/ParameterFields/LeaderboardIdField.js
```diff
@@ -0,0 +1,147 @@
...
+  leaderboards: ?Array<Leaderboard>,
+  value: string
+): string => {
+  if (!leaderboards) return value;
```
AlexandreSi, line 23: I don't know if the whole uuid is necessary, maybe juste the first hyphen as it is done in the admin? I am not sure about that though

newIDE/app/src/EventsSheet/ParameterFields/LeaderboardIdField.js
```diff
@@ -0,0 +1,147 @@
...
+  const quotedIdLeaderboards = leaderboards.map(leaderboard => ({
+    id: `"${leaderboard.id}"`,
+    name: leaderboard.name,
+  }));
+  const leaderboard = quotedIdLeaderboards.find(
+    leaderboard => leaderboard.id === value
```
AlexandreSi, line 29: To avoid having to define a new variable `quotedIdLeaderboards`, your test here could be:

```js
leaderboards.find( leaderboard => `"${leaderboard.id}"` === value )
`
``

File: newIDE/app/src/EventsSheet/ParameterFields/LeaderboardIdField.js [hunk 6]
```diff
@@ -0,0 +1,147 @@
...
+  );
+  return leaderboard ? leaderboard.name : value;
+};
+
+export function LeaderboardIdField(props: ParameterFieldProps) {
+  const { leaderboards } = React.useContext(LeaderboardContext);
+  const [isAdminOpen, setIsAdminOpen] = React.useState(false);
+
+  const isCurrentValueInLeaderboardList =
```
AlexandreSi, line 38: 💜

newIDE/app/src/EventsSheet/ParameterFields/LeaderboardIdField.js
AlexandreSi: I see you did that just below, maybe there's something I'm missing

Extensions/Leaderboards/JsExtension.js
```diff
@@ -0,0 +1,108 @@
...
+        '',
+        'JsPlatform/Extensions/leaderboard.svg',
+        'JsPlatform/Extensions/leaderboard.svg'
+      )
+      .addParameter('leaderboardId', 'Leaderboard', '', false)
+      .addParameter('expression', 'Score to register for the player', '', false)
+      .addParameter('string', 'Name to register for the player', '', false)
+      .addParameter(
+        'scenevar',
+        _('Variable where to store the saved score'),
```
4ian, line 58:
```suggestion
        _('Variable where to store the saved score (optional)'),
```

Extensions/Leaderboards/JsExtension.js
```diff
@@ -0,0 +1,108 @@
...
+      )
+      .getCodeExtraInformation()
+      .setIncludeFile('Extensions/Leaderboards/leaderboardstools.js')
+      .setFunctionName('gdjs.evtTools.leaderboards.setPlayerScore');
+
+    extension
+      .addCondition(
+        'LastSentEntrySaveFailed',
+        _('Last entry failed to save'),
+        _('Check if the last sent entry failed to save'),
```
4ian, line 78:
```suggestion
        _('Check if the last sent entry failed to save in the leaderboard.'),
```

Extensions/Leaderboards/JsExtension.js
```diff
@@ -0,0 +1,108 @@
...
+        _('Last entry failed to be saved in leaderboard'),
```
4ian, line 79:
```suggestion
        _('Last entry failed to be saved in the leaderboard'),
```

Extensions/Leaderboards/JsExtension.js
```diff
@@ -0,0 +1,108 @@
...
+        _(''),
+        'JsPlatform/Extensions/leaderboard.svg',
+        'JsPlatform/Extensions/leaderboard.svg'
+      )
+      .getCodeExtraInformation()
+      .setIncludeFile('Extensions/Leaderboards/leaderboardstools.js')
+      .setFunctionName('gdjs.evtTools.leaderboards.lastEntrySaveFailed');
```
4ian, line 86: For JS functions, always start with a verb (for condition names, this is not forced, so "LastEntrySaveFailed" is fine)

```suggestion
      .setFunctionName('gdjs.evtTools.leaderboards.hasLastEntrySaveFailed');
```

Extensions/Leaderboards/leaderboardstools.ts
```diff
@@ -0,0 +1,70 @@
...
+          (!!_scoreLastSentAt && Date.now() - _scoreLastSentAt < 500)
+        ) {
+          errorVar.setString('Wait before sending a new score.');
+        } else {
+          const baseUrl = 'https://api.gdevelop-app.com/play';
+          fetch(
+            `${baseUrl}/game/${gdjs.projectData.properties.projectUuid}/leaderboard/${leaderboardId}/entry`,
+            {
+              body: JSON.stringify({
+                playerName: playerName,
```
4ian, line 31: I think we could send playerUuid by exposing `_makePlayerUuid` from runtimeGame (that you can get from the runtime scene)

Extensions/Leaderboards/JsExtension.js
```diff
@@ -0,0 +1,108 @@
...
+        _('Save player score'),
+        _("Save the player's score to a giver leaderboard."),
+        _(
+          'Send to leaderboard _PARAM0_ the score _PARAM1_ with player name: _PARAM2_.'
+        ),
+        '',
+        'JsPlatform/Extensions/leaderboard.svg',
+        'JsPlatform/Extensions/leaderboard.svg'
+      )
+      .addParameter('leaderboardId', 'Leaderboard', '', false)
```
4ian, line 53: Almost sure you'll need to pass the current scene too (search for "addCodeOnlyParameter" and "currentScene")

newIDE/app/src/EventsSheet/ParameterRenderingService.js
```diff
@@ -109,6 +113,7 @@ const inlineRenderers: { [string]: ParameterInlineRenderer } = {
   trueorfalse: renderInlineTrueFalse,
   operator: renderInlineOperator,
   relationalOperator: renderInlineRelationalOperator,
+  leaderboardId: InlineLeaderboardIdField,
```
AlexandreSi, line 116: For consistency's sake, I would name it `renderInlineLeaderboardIdField`

newIDE/app/src/EventsSheet/index.js
```diff
@@ -124,6 +127,7 @@ type ComponentProps = {|
   ...Props,
   authenticatedUser: AuthenticatedUser,
   preferences: Preferences,
+  leaderboards: LeaderboardState,
```
AlexandreSi, line 130: `leaderboards` doesn't seem to be the best name but hard to find another one: `leaderboardManager`? `leaderboardHandler`?

newIDE/app/src/MainFrame/EditorContainers/EventsEditorContainer.js
```diff
@@ -62,25 +63,27 @@ export class EventsEditorContainer extends React.Component<RenderEditorContainer
...
-        project={project}
-        scope={{
-          layout,
-        }}
-        globalObjectsContainer={project}
-        objectsContainer={layout}
-        events={layout.getEvents()}
-        onOpenExternalEvents={this.props.onOpenExternalEvents}
-      />
+      <LeaderboardProvider gameId={project.getProjectUuid()}>
```
AlexandreSi, line 66: As discussed yesterday together, is it here the best place to put the provider? If you use the action in an external events editor, will it work?

newIDE/app/src/EventsSheet/ParameterFields/LeaderboardIdField.js
```diff
@@ -0,0 +1,147 @@
...
+              onClick={() => setIsAdminOpen(true)}
+            />
+            <LargeSpacer />
+            <Toggle
+              labelPosition="right"
+              toggled={isTextInput}
+              onToggle={() => setIsTextInput(!isTextInput)}
+              style={{ marginTop: 8 }}
+              disabled={!leaderboards}
+            />
```
4ian, line 115: Can we try a RaisedButtonWithSplitMenu?

The default button is the OpenInNew icon that opens the admin. And the arrow menu would show a single menu item "Switch to an expression editor" or "Switch to leaderboard selector".

newIDE/app/src/MainFrame/EditorContainers/ExternalEventsEditorContainer.js
```diff
@@ -180,4 +181,11 @@ export class ExternalEventsEditorContainer extends React.Component<
 
 export const renderExternalEventsEditorContainer = (
   props: RenderEditorContainerPropsWithRef
-) => <ExternalEventsEditorContainer {...props} />;
+) =>
+  props.project ? (
```
4ian, line 185: Is not having a project even possible?

newIDE/app/src/SceneEditor/LeaderboardDialog.js
```diff
@@ -0,0 +1,37 @@
...
+        <FlatButton
+          label={<Trans>Close</Trans>}
+          onClick={onClose}
+          key={'Cancel'}
+        />,
+      ]}
+      open={open}
+      cannotBeDismissed={true}
+      onRequestClose={onClose}
+      title="Leaderboards"
```
4ian, line 28:
```suggestion
      title={<Trans>Leaderboards</Trans>}
```

newIDE/app/src/stories/componentStories/ParameterFields/LeaderboardIdField.stories.js
```diff
@@ -0,0 +1,155 @@
...
+        setDisplayOnlyBestEntry: action('setDisplayOnlyBestEntry'),
+        createLeaderboard: () => {
+          throw new Error('createLeaderboard');
+        },
+        listLeaderboards: action('listLeaderboards'),
+        selectLeaderboard: leaderboardId => {
+          setCurrentLeaderboard(leaderboardsByIds[leaderboardId]);
+        },
+        updateLeaderboard: () => {
+          throw new Error('updateLeaderboard');
```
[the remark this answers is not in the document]
AlexandreSi: The `action` method is here to make sure that the right callback is plugged to the right button when clicking on it, you can see in the panel below a line that says something like `Click on 'callbackName'`

newIDE/app/src/SceneEditor/LeaderboardDialog.js
```diff
@@ -0,0 +1,37 @@
+//@flow
```
4ian, line 1: Should this file really be inside SceneEditor folder? It could be outside, maybe even inside a `src/Leaderboards` folder

newIDE/app/src/stories/componentStories/ParameterFields/LeaderboardIdField.stories.js
```diff
@@ -0,0 +1,155 @@
...
+
+import LeaderboardContext from '../../../Leaderboard/LeaderboardContext';
+import {
+  type Leaderboard,
+  type LeaderboardSortOption,
+} from '../../../Utils/GDevelopServices/Play';
+
+const gd: libGDevelop = global.gd;
+
+export default {
```
4ian, line 22: 🤩 stories 🤩

newIDE/app/src/EventsSheet/ParameterFields/LeaderboardIdField.js
```diff
@@ -0,0 +1,147 @@
...
+        renderButton={style => (
+          <>
+            <RaisedButton
+              icon={<OpenInNew />}
+              primary
+              style={style}
+              onClick={() => setIsAdminOpen(true)}
+            />
+            <LargeSpacer />
+            <Toggle
```
AlexandreSi, line 109: Is it possible to have something that explains what this toggle does? A tooltip or a title? or an icon `Info`?

newIDE/app/src/MainFrame/EditorContainers/EventsEditorContainer.js
4ian, line 66: Do we really need the provider to be here, couldn't there be a single LeaderboardProvider at the top of the app? (Providers.js)

newIDE/app/src/EventsSheet/ParameterFields/LeaderboardIdField.js
AlexandreSi: And it would explain why it is disabled when leaderboards could not have been fetched

4ian (commented):
Lots of great stuff! 
My main concern is about the fact that we're "polluting" EventsSheet with the leaderboard provider and the call to `listLeaderboards`. Could we have:

- The provider as high as possible in the component tree.
- The field (or the inline field) to use the context and call listLeaderboards?

newIDE/app/src/EventsSheet/ParameterFields/LeaderboardIdField.js
4ian: (the idea is to hide this "advanced" thing, and to avoid a Toggle without label which is a bit hard to understand)

newIDE/app/src/EventsSheet/index.js
AlexandreSi: This prop should be optional I think, when used in an External function, there is no notion of project/gameId. There might be a bug when opening an extension and editing its events sheets

newIDE/app/src/EventsSheet/ParameterFields/LeaderboardIdField.js
fannieyan: No you are right, since we return only the name when found, we do not need to create quotedIdLeaderboards

newIDE/app/src/MainFrame/EditorContainers/EventsEditorContainer.js
AlexandreSi: I don't know if it would conflict with the provider that provides the game dashboard. We wouldn't want the user to select a leaderboard for a different game

4ian: I think the "bottom most" provider is used when you `useContext`. I agree we don't want to mix leaderboards between games. And that if we put the provider too deep in the tree (like in the field), well you have to reload leaderboards every time a field is rendered.

newIDE/app/src/EventsSheet/ParameterFields/LeaderboardIdField.js
```diff
@@ -0,0 +1,147 @@
...
+  }));
+  const leaderboard = quotedIdLeaderboards.find(
+    leaderboard => leaderboard.id === value
+  );
+  return leaderboard ? leaderboard.name : value;
+};
+
+export function LeaderboardIdField(props: ParameterFieldProps) {
+  const { leaderboards } = React.useContext(LeaderboardContext);
+  const [isAdminOpen, setIsAdminOpen] = React.useState(false);
```
4ian, line 36: Could we have a React effect that reloads the leaderboard if we have none loaded? Or even if they are loaded, it's ok to update them silently:

newIDE/app/src/MainFrame/EditorContainers/EventsEditorContainer.js
fannieyan: Isn't it a bit strange to place the Provider on top of the App even though you may start the app with no project (it needs a gameId and it seems a bit complex to retrieve the gameId because of the difference browser/local app) ?
I suggest we make it wrap the editorsTabs in MainFrame instead and keep the fetch call in EventsSheet ?

newIDE/app/src/MainFrame/EditorContainers/ExternalEventsEditorContainer.js
fannieyan: I think in practice it would not be possible, but `RenderEditorContainerPropsWithRef` is a common type for all editor tab contents that do not necessarily need a project to be opened (like extensions)

newIDE/app/src/EventsSheet/ParameterFields/LeaderboardIdField.js
fannieyan: This function will be used in inline display of parameter, I am not sure about changing the display of uuid specifically when there could be anything (like an expression instead)

newIDE/app/src/MainFrame/EditorContainers/ExternalEventsEditorContainer.js
4ian: In this case, do a: 

```
    if (!project) {
      //TODO: Error component
      return <div>No project opened!</div>;
    }
```

This "TODO" is there since the beginning of the app... but well it's never shown, it's just to give a hint to a future developer that would rework the main frame and forget to pass the `project`.

newIDE/app/src/EventsSheet/ParameterFields/LeaderboardIdField.js
4ian: I think it's fine to return the whole value. We've not being able to identify a leaderboard id => we display everything.

Extensions/Leaderboards/JsExtension.js
fannieyan: Is it to get the playerUuid or did I miss something ?

Bouh:
image
I suggest a Gear icon in replacement of the external window, the button allows to *setup* the leaderboard. External sound like to me "open something out of the engine".

Extensions/Leaderboards/JsExtension.js
4ian: Yes it's to get the `runtimeScene` passed to your function (from which you can use `getGame` (https://docs.gdevelop.io/GDJS%20Runtime%20Documentation/classes/gdjs.RuntimeScene.html#getGame) to get the gdjs.RuntimeGame

fannieyan:
> I suggest a Gear icon in replacement of the external window, the button allows to _setup_ the leaderboard. External sound like to me "open something out of the engine".

@Bouh I wanted to match with the scene variable field in which the button opens the Variable editor. Though the Leaderboard admin does more than variable editor indeed. What do you think ? ^^

fannieyan:
New look for leaderboard input !

image

newIDE/app/src/UI/RaisedButtonWithSplitMenu.js
```diff
@@ -15,6 +15,7 @@ type Props = {|
   icon?: React.Node,
   onClick: ?() => void,
   buildMenuTemplate: (i18n: I18nType) => Array<MenuItemTemplate>,
+  style?: Object,
```
4ian, line 18: In general it's good to type as precisely as possible the style object, so that you don't encourage future developers to pass arbitrary styles. In this case:

```
{|
    marginTop?: number,
    marginBottom?: number,
    marginLeft?: number,
    marginRight?: number,
    margin?: number,
    flexShrink?: 0,
  |}
```

This also helps to refactor in the future UI components, because just looking at the type you know "ok, these are all the styles that can be passed, nothing more, so it's manageable I don't need to handle everything in theory"

newIDE/app/src/EventsSheet/index.js
```diff
@@ -267,6 +271,18 @@ export class EventsSheetComponentWithoutHandle extends React.Component<
 
   componentDidMount() {
     this.setState({ allEventsMetadata: enumerateEventsMetadata() });
+    if (this.props.leaderboardsManager) {
+      this.props.leaderboardsManager
```
4ian, line 275: Is there any chance we could do this inside the inline renderers? 
Or at least, have the inline renderers do that once (with a useEffect) only if the leaderboards are not loaded?

I'm not sure if it would work, I'm just trying to remove as much logic as possible from this poor EventsSheet

4ian (approved):
Ok for me to go ahead! With these:
- See if we can avoid the logic in EventsSheet or not.
- Let's quickly do a new PR with additional stuff passed to the leaderboard entry endpoint so that we can improve the backend logic later without breaking changes.

newIDE/app/src/EventsSheet/index.js
fannieyan: I tried fetching in inlineRenderer, with a fetch only if leaderboards is null, but it still called the API once for every occurence of the field in the EventsSheet the first time it renders (My guess is that the first time, all the useEffect are called at the same time and leaderboards are still null at the time). 😕

fannieyan: This will be treated in the next PR (here : https://github.com/4ian/GDevelop/pull/3779) that will move the listLeaderboards the API call from EventsSheet to the inlineField/ParameterField and ensure that only one call is made at once by adding a variable in the Provider component.
(More specific style typing too)

4ian: Sounds good 👍

newIDE/app/src/UI/RaisedButtonWithSplitMenu.js
4ian: (resolving as this will be done in the next PR)

PR #3767, opened 2022-03-21
Add key expressions to "Simulate control" action for TopDown

Elairyx7301:
Clang format does not seem to be present in the file.
Should that be changed as well?


4ian (approved):
Looks good as is!

PR #3769, opened 2022-03-22
Allow to specify a slug to be used in liluo.io games URL

D8H:
The "user name" drop-down list contains the name of the user that is currently used for the slug and the current user name. Thus, its a choice between not changing the slug or putting its own name.

A game slug change is ignore if the user slug doesn't match the current user.

I didn't add the slug in the game detail view. I guess that the button to open the game in the list and be redirected to the slug will be used by users to check the slug.

image

image


newIDE/app/src/GameDashboard/GameDetailsDialog.js
```diff
@@ -184,27 +192,36 @@ export const GameDetailsDialog = ({
...
+        console.error(
+          'Unable to update the game owners or authors:',
+          error.response || error.message
+        );
         showErrorBox({
           message:
             i18n._(t`Unable to update the game owners or authors.`) +
             ' ' +
-            i18n._(t`Verify your internet connection or try again later.`),
+            ((error.response && error.response.data) ||
```
4ian, line 203: Usually we avoid surfacing the response data directly to the user - in case we want more info we would ask the backend to return an error with a body like `{ code: 'area/what-happened', }`, so that the UI can react accordingly (or display a generic message if the code is not present/not expected).

newIDE/app/src/GameDashboard/GameDetailsDialog.js
```diff
@@ -160,19 +160,27 @@ export const GameDetailsDialog = ({
     const { id } = profile;
 
     // Set public game to null as it will be refetched automatically by the callback above.
+    const notUpdatedGame = publicGame;
```
4ian, line 163: nit: Even more precisely, this is the "notUpdatedPublicGame"

newIDE/app/src/ProjectManager/PublicGameProperties.js
```diff
@@ -13,7 +13,10 @@ import {
   allGameCategories,
   getCategoryName,
 } from '../Utils/GDevelopServices/Game';
+import AuthenticatedUserContext from '../Profile/AuthenticatedUserContext';
+import { Line, Spacer } from '../UI/Grid';
 import { I18n } from '@lingui/react';
+import { string } from 'prop-types';
```
4ian, line 19: Remove this line - I'm adding a warning in eslint to avoid this in the future.

4ian (commented):
I'll let someone else approve, I've added a few comments, but looks good overall! :)

arthuro555:
> Thus, its a choice between not changing the slug or putting its own name.

The UI should reflect that, then. Instead of a dropdown, it should be just showing the text, and, if you are logged into a different account than the current slug and an owner, a button to change the slug to your own username.

4ian:
For the UI, we'll keep a drop-down because in the future you should be able to specify another owner (if we added a "double opt in": a new owner will have to accept ownership of a game).
The UI is starting to be a bit bloated, we might want to insert section titles, but not a mandatory thing to get this merged.

newIDE/app/src/GameDashboard/PublicGameProperties.js
```diff
@@ -162,6 +194,35 @@ function PublicGameProperties({
             multiline
             rows={5}
           />
+          {setUserSlug && setGameSlug && (
+            <Line>
```
4ian, line 198: Double check if it looks good on a small (mobile) screen :)

newIDE/app/src/GameDashboard/PublicGamePropertiesDialog.js
```diff
@@ -11,12 +11,15 @@ import {
...
+import AuthenticatedUserContext from '../Profile/AuthenticatedUserContext';
 
 /**
  * Changes that are not stored in the Project.
  */
 export type PartialGameChange = {|
-  ownerIds?: Array<string>,
+  ownerIds: Array<string> | null,
+  userSlug: string | null,
+  gameSlug: string | null,
```
4ian, line 22: Any reason not to do that: 

```suggestion
  ownerIds?: Array<string>,
  userSlug?: string,
  gameSlug?: string,
```

A property with a `?` after its name indicated that you're not forced to specify it. This means it will be `undefined`, and the semantic is "I've not specified it so don't even attempt to change/modify it".
While a `null` has a slighlty different meaning. This can be understood as "I specified this property, so I probably want to set it to null, i.e: remove it."

newIDE/app/src/Export/GenericExporters/OnlineWebExport/OnlineGamePropertiesDialog.js
```diff
@@ -78,7 +78,12 @@ export const OnlineGamePropertiesDialog = ({
         orientation: orientation || 'default',
       })
     ) {
-      await onApply({ discoverable });
+      await onApply({
```
4ian, line 81: See my other comment, I would rather not specify things that are not changing rather than having this feeling when reading the code that it's going to set them to null.

4ian (approved):

newIDE/app/src/GameDashboard/PublicGamePropertiesDialog.js
D8H: As it's used at 2 places, I was worried that optional attributes could lead to forbid to update the UI for the browser version.
But, you're right, null has a meaning, I will use optional attributes.

4ian: Totally, you're right that the big default of allowing to not specify an attribute is that.. well you can forget to specify it :) 
What I usually do is that I set the property as mandatory (I remove the `?`), I check everywhere in the code where Flow is complaining, and if it's ok I put back the `?`. Not ideal but better than nothing!

newIDE/app/src/GameDashboard/PublicGameProperties.js
D8H: Should the drop-down list be disable if there is only one choice?

4ian: For the game list: we should use a `ResponsiveLineStackLayout` so that on mobile it goes on one column (the thumbnail then the rest)

PR #3772, opened 2022-03-23
Improve UI again

### LINKED ISSUE #3761: Copy and paste variable objects
I believe we must have a regression here! Now to copy object variables, we must do more steps than before

https://user-images.githubusercontent.com/23488207/159143367-3583705e-3a69-4abb-bcaf-aa41942fbe21.mp4

AlexandreSi:
Fixes #3736 
Fixes #3761 

image

### DIFF
--- a/newIDE/app/src/VariablesList/index.js
+++ b/newIDE/app/src/VariablesList/index.js
@@ -429,8 +429,8 @@ export default class VariablesList extends React.Component<Props, State> {
 
     return (
       <Column noMargin expand useFullHeight>
-        {allVariables.length ? (
-          <React.Fragment>
+        <React.Fragment>
+          {allVariables.length ? (
             <ScrollView autoHideScrollbar>
               <SortableVariablesListBody
                 variablesContainer={this.props.variablesContainer}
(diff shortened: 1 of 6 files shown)


newIDE/app/src/VariablesList/index.js
```diff
@@ -429,8 +429,8 @@ export default class VariablesList extends React.Component<Props, State> {
 
     return (
       <Column noMargin expand useFullHeight>
-        {allVariables.length ? (
-          <React.Fragment>
+        <React.Fragment>
```
ClementPasteau, line 432: I'm not sure you need the fragment any more?

4ian:
image

That's a lot of (well, 2) primary buttons. Can we exceptionally hide the bottom right "Add" when there are no variables (i.e: when the empty placeholder is shown)?

newIDE/app/src/stories/FixedHeightFlexContainer.js
```diff
@@ -1,3 +1,4 @@
+// @flow
```
4ian, line 1: Nice

4ian (approved):

PR #3773, opened 2022-03-24
Add expression to format player name to send it to leaderboard

AlexandreSi:


Extensions/Leaderboards/tests/leaderboardstools.spec.js
```diff
@@ -0,0 +1,55 @@
+describe('Leaderboards', () => {
```
4ian, line 1: Add `// @ts-check`

Extensions/Leaderboards/tests/leaderboardstools.spec.js
```diff
@@ -0,0 +1,55 @@
...
+      expect(gdjs.evtTools.leaderboards.formatPlayerName(null))
+        .to.be.a('string')
+        .and.match(/^Player\d{5}/);
+      expect(gdjs.evtTools.leaderboards.formatPlayerName(5))
+        .to.be.a('string')
+        .and.match(/^Player\d{5}/);
+      expect(gdjs.evtTools.leaderboards.formatPlayerName(undefined))
+        .to.be.a('string')
+        .and.match(/^Player\d{5}/);
+      expect(gdjs.evtTools.leaderboards.formatPlayerName(() => {}))
```
4ian, line 39: Typing + the extension parameter types should prevent anything but a string to reach the function.

4ian (approved):

Extensions/Leaderboards/tests/leaderboardstools.spec.js
AlexandreSi: If a user uses an expression, or a JS event, it could store in a variable all sorts of things, couldn't it? Or are there securities that I'm not aware of?

4ian: A GDevelop expression are "type safe" and will always guarantee that you pass a proper string (otherwise the expression is not valid and won't even be executed, it will return an empty string or 0 to the action/condition using it).

But you're right, a JS event could call this with an improper type, so it's ok to protect against it.

PR #3777, opened 2022-03-26
Specify if you are updating or reinstalling in the extension store page

arthuro555:
Adapted from prior work on #3332

Make the update dialogue aware of if you are updating or reinstalling an extension update, and change the button text accordingly.

### DIFF
--- a/newIDE/app/src/AssetStore/ExtensionStore/ExtensionInstallDialog.js
+++ b/newIDE/app/src/AssetStore/ExtensionStore/ExtensionInstallDialog.js
@@ -24,7 +24,6 @@ import { UserPublicProfileChip } from '../../UI/User/UserPublicProfileChip';
 import RaisedButton from '../../UI/RaisedButton';
 import Window from '../../Utils/Window';
 import { useExtensionUpdate } from './UseExtensionUpdates';
-import { enumerateEventsFunctionsExtensions } from '../../ProjectManager/EnumerateProjectItems';
 
 const getTransformedDescription = (extensionHeader: ExtensionHeader) => {
   if (
--- a/newIDE/app/src/AssetStore/ExtensionStore/ExtensionInstallDialog.js
+++ b/newIDE/app/src/AssetStore/ExtensionStore/ExtensionInstallDialog.js
@@ -23,6 +23,8 @@ import { IconContainer } from '../../UI/IconContainer';
 import { UserPublicProfileChip } from '../../UI/User/UserPublicProfileChip';
 import RaisedButton from '../../UI/RaisedButton';
 import Window from '../../Utils/Window';
+import { useExtensionUpdates } from './UseExtensionUpdates';
+import { enumerateEventsFunctionsExtensions } from '../../ProjectManager/EnumerateProjectItems';
 
 const getTransformedDescription = (extensionHeader: ExtensionHeader) => {
   if (
@@ -46,6 +48,7 @@ type Props = {|
   onInstall: () => Promise<void>,
   onEdit?: () => void,
   alreadyInstalled: boolean,
+  project: gdProject,
 |};
 
 const ExtensionInstallDialog = ({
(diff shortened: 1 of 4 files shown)


newIDE/app/src/AssetStore/ExtensionStore/ExtensionInstallDialog.js
```diff
@@ -46,6 +48,7 @@ type Props = {|
   onInstall: () => Promise<void>,
   onEdit?: () => void,
   alreadyInstalled: boolean,
```
4ian, line 50: If you pass the project, the dialog can now infer if the extension is alreadyInstalled. So we should probably remove this prop to avoid any risk of inconsistency?

newIDE/app/src/AssetStore/ExtensionStore/ExtensionInstallDialog.js
```diff
@@ -55,7 +58,13 @@ const ExtensionInstallDialog = ({
   onInstall,
   onEdit,
   alreadyInstalled,
+  project,
 }: Props) => {
+  const installedExtensions = new Set(
+    enumerateEventsFunctionsExtensions(project).map(ext => ext.getName())
+  );
```
4ian, line 65: Could you useMemo to avoid recomputing these updates/installed extensions at each render?
Could we also return the installedExtensions from the hook so we avoid to enumerate twice the extensions?

newIDE/app/src/AssetStore/ExtensionStore/ExtensionInstallDialog.js
```diff
@@ -117,8 +126,12 @@ const ExtensionInstallDialog = ({
             label={
               !isCompatible ? (
                 <Trans>Not compatible</Trans>
-              ) : alreadyInstalled ? (
```
4ian, line 120: Any mention of alreadyInstalled in the dialog should be replaced by a boolean infered from what is returned by the hook I guess.

newIDE/app/src/AssetStore/ExtensionStore/UseExtensionUpdates.js
```diff
@@ -0,0 +1,55 @@
...
+
+type UpdateMetadata = {|
+  type: 'patch' | 'minor' | 'major',
+  currentVersion: string,
+  newestVersion: string,
+|};
+type ExtensionUpdates = Map<string, UpdateMetadata>;
+
+export const useExtensionUpdates = (project: gdProject): ExtensionUpdates => {
+  const { extensionShortHeadersByName } = useContext(ExtensionStoreContext);
```
4ian, line 15: Would be a good exercise to check how easy it is to test this kind of hook :)

4ian (approved):

newIDE/app/src/AssetStore/ExtensionStore/ExtensionInstallDialog.js
arthuro555: I hadn't done it because the project is not something we can use as a dependency for react hooks or that will trigger a rerender on an update of the list of extensions, so the memo could keep a wrong outdated memoized list of installed extensions. While it is very unlikely for something else to happen while in this menu, and not a huge issue to have that button mislabeled, it is dangerous to make these kinds of assumptions that "this will never happen here" imo. Especially if one day we add remote collaboration and a lot of stuff could be done with the project while someone does something else on a completely different menu.

4ian: There are a few things that would allow to avoid a future problem of outdated memoized list:
- Any change could generate a new JS object (in the sense, not the same reference) so that any useMemo/useContext is updated.
- Maybe we'll have a solution like `const project = useCollaborativeProject()` so that any dialog will progressively be made collaborative.
- In any case, the issue of "this project is potentially outdated" is limited to this dialog - you just need to close the dialog and re-open it to avoid the issue.

Not a big deal here anyway, but some dialogs are lagging so I'm now more carful about this.
Note that when we add collaborative updates of a `project`, useMemo(..., [project]) should *still* work. It's not because you have a project that can be updated at any time that you don't want to optimise your renderings (in which case you probably want to use what I said before).

newIDE/app/src/AssetStore/ExtensionStore/UseExtensionUpdates.js
```diff
@@ -0,0 +1,94 @@
...
+            project.getEventsFunctionsExtension(extension.name).getVersion(),
+            extension.version
+          )
+        : null,
+    [project, extension]
+  );
+};
+
+// TODO: Use this hook to display a visual hint when an installed extension can be updated
+export const useAllExtensionUpdates = (
```
4ian, line 64: This is dead and untested code, so I'm removing it (you can always use these commits to find it again later :))

PR #3535, opened 2022-01-21
Add asynchronous events & Wait action

arthuro555:
Closes #2975

Aims at adding asynchronous event support to GDevelop.

TODO (polish):

- [x] Write tests for code generation
- [x] Write tests for the async task manager
- [x] ~~Add an instruction that uses PromiseTask to ensure it works~~ Unit tests show that it works
- [x] Add documentation on making an asynchronous action
- [x] Decide about: the design of LongLivedObjectsList (is it the best data structure to use/can we do better?)
- [ ] Decide about: could avoid creating and splitting events into AsyncEvent?
- [x] Fix the test with object groups
- [x] Avoid replacing a parameter when assigning `asyncObjectsLists`
- [x] Add test case about TaskGroup code generation (object async action)
- [x] Decide about: how things should work when an async action is used on an object
- [x] Check the code generation works with the hot reloading (stable unique identifiers, and the hot reloader should probably replace the callbacks of the async tasks that are waiting :))
   - ~~Will be done in another PR since it is a bit complex~~ Nevermind, it was extremely easy after all 😅 
- [x] Revert unwanted changes done by the autoformatter
- [ ] FINAL REVIEW 🎉

DONE:

- [x] Add wait action
- [x] Add a way to define instruction as asynchronous
- [x] Add basic code generation
- [x] Make the generated code functional
  - [x] Backup/restore object lists
  - [x] Ensure preprocessing works on all events
  - [x] Test all edge cases
- [x] Ensure support of events functions
  - [x] Find out why it randomly turns asynv on its own
- [x] Pass runtimeScene to the callback without leaking runtimeScene
- [x] Write wiki page
- [x] Add visual indicator to asynchronous actions

This PR uses a kind-of workaround to make this work: While preprocessing the events, iterate over actions, and if an asynchronous one is found, move them all with the subevents to an "asynchronous" subevent. This is because otherwise, it is difficult to generate the asynchronous callback function. The part that would do the generation of that function would be `gd::EventsCodeGenerator::GenerateActionsListCode`. That function doesn't have access to subevents though, it is the responsibility of an event's code generator to mix actions, conditions, and subevents code generation as it likes. Splitting that code generation between both is hard and error prone. By moving affected actions and subevents together to an internal subevent, we can do a custom mix of actions and subevents, allowing us to have one function responsible for all code generation of those subevents and actions again, allowing us to have a less error-prone and concise code generation.

Of course, this isn't without issues: 
[...]

PR #3778, opened 2022-03-26
[P2P] Add a way to check if a connection was successful

arthuro555:
Adds a condition to see if the last connection attempt succeeded and an async action to await that results (based on #3535).
This is useful to display an error message when a connection failed.

This branch is without the asynchronous code, to keep the changed files readable and reviewable. Use the branch `p2p-async` from my fork, a version with the changes from the async PR, to actually try the feature.

### DIFF
--- a/Extensions/P2P/B_p2ptools.ts
+++ b/Extensions/P2P/B_p2ptools.ts
@@ -123,6 +123,11 @@ namespace gdjs {
       let lastError = '';
 
       /**
+       * Set to true if the last connection attempt succeeded.
+       */
+      let lastConnectionSucessful = false;
+
+      /**
        * List of IDs of peers that just disconnected.
        */
       const disconnectedPeers: string[] = [];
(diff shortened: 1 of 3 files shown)


4ian:
Alright, should we make a "SetOptionallyAsync" and slightly rework the code generation so that it checks if an action is Async or optionnally async and a "await" is set to true on the instruction? 

That would avoid duplications of such actions :)

arthuro555:
Updated the OR to use the new optional async feature!

Extensions/P2P/B_p2ptools.ts
```diff
@@ -230,10 +238,60 @@ namespace gdjs {
         if (peer === null) return;
         const connection = peer.connect(id);
         connection.on('open', () => {
+          // See line 158 for the failure case.
```
4ian, line 241: Referring to a line is fragile (this will break in the future with the person breaking it not even knowing they broke a comment by adding a new line or doing a refactoring far from this). Can you refer to something else, like a function name or anything else?

arthuro555:
Yes, that would be a good idea. Alternatively, I've been thinking, maybe we could try to do the same thing as for the await tween action: use a separate action for awaiting. That would be good for the separation of concerns I think, and make the awaiting even more explicit for users. Of course, that comes at the risk of obscuring the possibility of awaiting... 🤔

arthuro555:
Bump! I don't remember what my last comment was about, but looking at the code it looks ready to be merged.

arthuro555:
bump

PR #3781, opened 2022-03-28
Improve UX with game not registered or with 0 leaderboard

AlexandreSi:


newIDE/app/src/UI/SelectField.js
```diff
@@ -34,6 +34,7 @@ type Props = {|
 
   floatingLabelText?: React.Node,
   helperMarkdownText?: ?string,
+  helperText?: React.Node,
```
ClementPasteau, line 37: do we have to add this parameter? or can we just re-use the `helperMarkdownText` (even without markdown)

newIDE/app/src/EventsSheet/ParameterFields/LeaderboardIdField.js
```diff
@@ -122,8 +135,8 @@ export function LeaderboardIdField(props: ParameterFieldProps) {
               buildMenuTemplate={i18n => [
                 {
                   label: isTextInput
-                    ? i18n._(t`Switch to expression`)
-                    : i18n._(t`Switch to select`),
+                    ? i18n._(t`Select the leaderboard from a list`)
+                    : i18n._(t`Enter the leaderboard id as an expression`),
```
ClementPasteau, line 140: I'm not sure I understand what this means.
If this is for the user to enter the id themselves, should this be more `type the leaderboard ID manually`?

newIDE/app/src/UI/SelectField.js
AlexandreSi: Markdown text does not accept anything else than a string as a source so I'm not sure it could be done safely.

ClementPasteau: I think the `SemiControlledTextField` uses a similar prop and has it used sometimes with:
`helperMarkdownText={i18n._(t`As a percent of the game width.`)}`

It would require having i18n imported where the field is used...

I'm not happy with both approaches 😅

AlexandreSi: Yes, I thought of adding this props because it would be the first parameter field to include i18n so not a big fan.

AlexandreSi: switching to new branch https://github.com/4ian/GDevelop/pull/3782 due to conflict hell

PR #3786, opened 2022-03-29
Add action to display leaderboard in iframe

AlexandreSi:


Extensions/Leaderboards/JsExtension.js
```diff
@@ -111,6 +111,37 @@ module.exports = {
...
+      .addCodeOnlyParameter('currentScene', '')
+      .addParameter('leaderboardId', 'Leaderboard', '', false)
+      .getCodeExtraInformation()
+      .setIncludeFile('Extensions/Leaderboards/leaderboardstools.js')
+      .setFunctionName('gdjs.evtTools.leaderboards.displayLeaderboard');
+
+    extension
+      .addAction(
+        'CloseLeaderboardView',
+        _('Close leaderboard view'),
```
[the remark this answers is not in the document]
AlexandreSi: Well, they are a bit different : if you open X leaderboards, you don't open X views: the first leaderboard opens the view, the followings just change the href. So the close leaderboard view seems ok to me.
BUT for sure, it may be not clear for our users. Maybe I can add "If a leaderboard is already open, it will replace it in the view"? in the action `DisplayLeaderboard`?

Extensions/Leaderboards/JsExtension.js
```diff
@@ -111,6 +111,37 @@ module.exports = {
       .setIncludeFile('Extensions/Leaderboards/leaderboardstools.js')
       .setFunctionName('gdjs.evtTools.leaderboards.formatPlayerName');
 
+    extension
+      .addAction(
+        'DisplayLeaderboard',
+        _('Display leaderboard'),
+        _('Display leaderboard on top of the game.'),
+        _('Display leaderboard _PARAM1_.'),
```
4ian, line 119: No period at the end of an events sheet sentence

Extensions/Leaderboards/leaderboardstools.ts
```diff
@@ -92,6 +96,119 @@ namespace gdjs {
...
+        if (_leaderboardViewIframe) {
+          _leaderboardViewIframe.src = targetUrl;
+        } else {
+          const domElementContainer = runtimeScene
+            .getGame()
+            .getRenderer()
+            .getDomElementContainer();
+          if (!domElementContainer) {
+            logger.error(
+              "Div covering game couldn't be found, leaderboard cannot be displayed."
```
4ian, line 150:
```suggestion
              "The div element covering the game couldn't be found, the leaderboard cannot be displayed."
```

Extensions/Leaderboards/leaderboardstools.ts
```diff
@@ -92,6 +96,119 @@ namespace gdjs {
...
+          iframe.style.height = '100%';
+          iframe.style.left = '0px';
+          iframe.style.width = '100%';
+          iframe.style.border = 'none';
+          _leaderboardViewIframe = iframe;
+          if (typeof window !== 'undefined') {
+            _leaderboardViewClosingCallback = (event: MessageEvent) => {
+              receiveMessage(runtimeScene, event);
+            };
+            (window as any).addEventListener(
```
4ian, line 170: Is `window as Window` working?

Extensions/Leaderboards/JsExtension.js
```diff
@@ -111,6 +111,37 @@ module.exports = {
       .setIncludeFile('Extensions/Leaderboards/leaderboardstools.js')
       .setFunctionName('gdjs.evtTools.leaderboards.formatPlayerName');
 
+    extension
+      .addAction(
+        'DisplayLeaderboard',
+        _('Display leaderboard'),
+        _('Display leaderboard on top of the game.'),
```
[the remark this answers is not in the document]
4ian: Agree with the other comment saying that we can also say "If a leaderboard was already displayed on top of the game, the new leaderboard will replace it"

Extensions/Leaderboards/JsExtension.js
```diff
@@ -111,6 +111,37 @@ module.exports = {
...
+      .addParameter('leaderboardId', 'Leaderboard', '', false)
+      .getCodeExtraInformation()
+      .setIncludeFile('Extensions/Leaderboards/leaderboardstools.js')
+      .setFunctionName('gdjs.evtTools.leaderboards.displayLeaderboard');
+
+    extension
+      .addAction(
+        'CloseLeaderboardView',
+        _('Close leaderboard view'),
+        _('Close the view of the leaderboard.'),
```
4ian, line 134:
```suggestion
        _('Close the leaderboard currently displayed on top of the game.'),
```

To be a bit more precise?

Extensions/Leaderboards/leaderboardstools.ts
AlexandreSi: No, TS doesn't seem to be able to cast it

Extensions/Leaderboards/JsExtension.js
```diff
@@ -111,6 +111,37 @@ module.exports = {
...
+        '',
+        'JsPlatform/Extensions/leaderboard.svg',
+        'JsPlatform/Extensions/leaderboard.svg'
+      )
+      .addCodeOnlyParameter('currentScene', '')
+      .addParameter('leaderboardId', 'Leaderboard', '', false)
+      .getCodeExtraInformation()
+      .setIncludeFile('Extensions/Leaderboards/leaderboardstools.js')
+      .setFunctionName('gdjs.evtTools.leaderboards.displayLeaderboard');
+
```
4ian, line 129: Let's add a condition "LeaderboardViewErrored" ("The leaderboard view had an error when opening") which returns true when we detect the leaderboard view could not be open, so that the game can react to this :)

Extensions/Leaderboards/leaderboardstools.ts
```diff
@@ -92,6 +96,119 @@ namespace gdjs {
...
+      ) {
+        const gameId = gdjs.projectData.properties.projectUuid;
+        const targetUrl = `https://liluo.io/games/${gameId}/leaderboard/${leaderboardId}`;
+        if (!(await checkLeaderboardAvailability(targetUrl))) {
+          logger.error('Leaderboard data could not be fetched, doing nothing');
+          return;
+        }
+
+        if (_leaderboardViewIframe) {
+          _leaderboardViewIframe.src = targetUrl;
```
4ian, line 142: For the loading error handling:
- Set back the `_leaderboardViewIframeErrored` to false here.
- In theory the event listener for "error" and "load" will then set it to true (or to false again).

Extensions/Leaderboards/leaderboardstools.ts
```diff
@@ -92,6 +96,119 @@ namespace gdjs {
...
+        } else {
+          const domElementContainer = runtimeScene
```
4ian, line 144: For the loading error handling:
- Set back the `_leaderboardViewIframeErrored` to false here.

Extensions/Leaderboards/leaderboardstools.ts
```diff
@@ -92,6 +96,119 @@ namespace gdjs {
...
+            .getGame()
+            .getRenderer()
+            .getDomElementContainer();
+          if (!domElementContainer) {
+            logger.error(
+              "The div element covering the game couldn't be found, the leaderboard cannot be displayed."
+            );
+            return;
```
4ian, line 152: For the loading error handling:
- Set back the `_leaderboardViewIframeErrored` to true here.

Extensions/Leaderboards/leaderboardstools.ts
```diff
@@ -92,6 +96,119 @@ namespace gdjs {
...
+          }
+          const iframe = document.createElement('iframe');
+
+          iframe.src = targetUrl;
```
4ian, line 156: For the loading error handling:
- Before setting `src`, add some `addEventListener` to listen to `load` and `error`. 
- In case of error, log it and set  the `_leaderboardViewIframeErrored` to true, then close the leaderboard view (the iframe could not load, we don't want to soft lock the game on a blank screen).
- In the "load" event: Set it to false, for robustness/safety sake.

Extensions/Leaderboards/leaderboardstools.ts
```diff
@@ -92,6 +96,119 @@ namespace gdjs {
...
+            );
+          }
+          domElementContainer.appendChild(iframe);
+        }
+      };
+
+      export const closeLeaderboardView = function (
+        runtimeScene: gdjs.RuntimeScene
+      ) {
+        if (!_leaderboardViewIframe) {
```
4ian, line 183: ~~For the loading error handling:~~
- ~~Reset `_leaderboardViewIframeErrored` to false at the beginning of this method~~

Actually don't do it, because it's important that we can close the leaderboard in case of an error, and still be able to check from the game conditions that an error happened.

The error flag should be reset to false only when we open a new leaderboard view.

Extensions/Leaderboards/JsExtension.js
4ian: (see other comments about "loading error handling")

Extensions/Leaderboards/JsExtension.js
```diff
@@ -111,6 +111,37 @@ module.exports = {
...
+    extension
```
4ian, line 130: I wonder if we also should have a condiiton "LeaderViewOpened" which means the iframe was created and is displayed on top of the game (even if it is still loading). This way we have:

- An action to open the leaderboard view.
- We have a condition to know if it's opened (useful to pause your game or do something in the background).
- We have a condition to check if it errored (so you can at least display "oops, try again later") and could not be displayed.
- We have an action to close the leaderboard (though I guess in the majority of cases, it's the player that will close it).

4ian (commented):
Good to go once we've bulletproofed how the leaderboard view react in case of error.
This can be:
- A loading error.
- "Worse", an error like a 404 which displays the wrong page. I'm not sure about this one. Maybe we could try to detect in the iframe the presence of the close button or some DOM element with a well known id. If it's not found 2 seconds after the loading of the iframe is done => we close the leaderboard view considering it errored.

This last case is more tricky (we have to ensure that as a parent we can read the content of the iframe... which is not guaranteed because browser could put restrictions => need more investigation).

Let's at least make the loading robust in case of loading error (which you can test by entering a wrong url I think).

Extensions/Leaderboards/leaderboardstools.ts
```diff
@@ -120,61 +146,161 @@ namespace gdjs {
...
+                  true
+                );
+              }
+              if (displayLoader) {
+                domElementContainer.appendChild(_loaderContainer);
+              }
+              domElementContainer.appendChild(_leaderboardViewIframe);
+            }
+          })
+          .catch(() => {
```
4ian, line 285: These .catch after a then will catch errors launched by `checkLeaderboardAvailability`, but also by the entire body of the `then`. If it's not meant to do that, consider using then 2nd argument instead: `.then(() => {...}, (error) => {...})`.
(Catching everything seems like a good idea, but could also hide real errors like syntax errors, unhandled exception and generally it's better to let them just go rather than hiding them behind an error message that is unrelated and confusing)

Extensions/Leaderboards/leaderboardstools.ts
```diff
@@ -120,61 +146,161 @@ namespace gdjs {
...
+              }
+
+              const iframe = document.createElement('iframe');
+
+              resetErrorTimeout(runtimeScene);
+              iframe.src = targetUrl;
+              iframe.id = 'leaderboard-view';
+              iframe.style.position = 'absolute';
+              if (displayLoader) {
+                // To trigger iframe loading and be able to listen to its events, use `opacity: 0` instead of `visibility: hidden` or `display: none`
```
4ian, line 259: Very useful comment, thanks!

Extensions/Leaderboards/leaderboardstools.ts
```diff
@@ -7,10 +7,36 @@ namespace gdjs {
...
+      _loaderContainer.style.width = '100%';
+      _loaderContainer.style.justifyContent = 'center';
+      _loaderContainer.style.alignItems = 'center';
+      const _loader = document.createElement('img');
+      _loader.setAttribute('width', '50px');
+      _loader.setAttribute(
+        'src',
+        'data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIGZpbGw9Im5vbmUiIHZpZXdCb3g9IjAgMCAyNCAyNCI+CjxjaXJjbGUgb3BhY2l0eT0nMC4yNScgY3g9IjEyIiBjeT0iMTIiIHI9IjEwIiBzdHJva2U9IiNGRkZGRkYiIHN0cm9rZS13aWR0aD0iNCI+PC9jaXJjbGU+CjxwYXRoIG9wYWNpdHk9JzAuNzUnIGZpbGw9IiNGRkZGRkYiIGQ9Ik00IDEyYTggOCAwIDAxOC04VjBDNS4zNzMgMCAwIDUuMzczIDAgMTJoNHptMiA1LjI5MUE3Ljk2MiA3Ljk2MiAwIDAxNCAxMkgwYzAgMy4wNDIgMS4xMzUgNS44MjQgMyA3LjkzOGwzLTIuNjQ3eiI+PC9wYXRoPgo8L3N2Zz4='
+      );
+      _loader.animate(
```
4ian, line 31: animate supports started with `iOS 13.1`... I think it's good but we could almost put a try/catch around it to allow this to work (without animation) without crashing the game on iOS 12/13 (even if it's <2% of iOS usage).

4ian: Also, will this consume CPU even if not rendered? 🤔

Extensions/Leaderboards/leaderboardstools.ts
```diff
@@ -7,10 +7,36 @@ namespace gdjs {
       let _lastPlayerName: string;
       let _lastErrorCode: number;
       let _leaderboardViewIframe: HTMLIFrameElement | null = null;
+      let _leaderboardViewIframeErrored: boolean = false;
+      let _leaderboardViewIframeLoading: boolean = false;
+      let _leaderboardViewIframeLoaded: boolean = false;
+      let _errorTimeoutId: NodeJS.Timeout | null = null;
```
4ian, line 13: Interesting, why the choice of `NodeJS.Timeout`?

Extensions/Leaderboards/leaderboardstools.ts
```diff
@@ -120,61 +146,161 @@ namespace gdjs {
...
+      export const displayLeaderboard = function (
+        runtimeScene: gdjs.RuntimeScene,
+        leaderboardId: string,
+        displayLoader: boolean
+      ) {
+        _leaderboardViewIframeErrored = false;
+        _leaderboardViewIframeLoaded = false;
+        _leaderboardViewIframeLoading = true;
+        const gameId = gdjs.projectData.properties.projectUuid;
+        const targetUrl = `http://localhost:3001/games/${gameId}/leaderboard/${leaderboardId}?inGameEmbedded=true&dev=true`;
```
4ian, line 224: Add a TODO when you leave development urls ;) Easier to spot before merging

Extensions/Leaderboards/leaderboardstools.ts
```diff
@@ -120,61 +146,161 @@ namespace gdjs {
...
-              true
+        }, 5000);
+      };
+
+      export const displayLeaderboard = function (
+        runtimeScene: gdjs.RuntimeScene,
+        leaderboardId: string,
+        displayLoader: boolean
+      ) {
+        _leaderboardViewIframeErrored = false;
```
4ian, line 220: Last (?) edge case: if I call displayLeaderboard multiple times quickly with different leaderboard ids.
Like, `displayLeaderboard(.., 1)`, `displayLeaderboard(.., 2)` and finally `displayLeaderboard(.., 3)`.

There is small, but non zero, chance that 1 will take time and the response (of `checkLeaderboardAvailability`) will arrive after the response for 2 and 3.

So you will end up displaying the leaderboard 1 instead of the last one that was requested - which will be surprising and surely considered as a bug/strange behavior.

There is no simple way to avoid this - ideally we would have something to "cancel" promises that are "in flight" and stop execution of anything left.
As this is not easily doable in JS, you can still store the `_requestedLeaderboardId`. You set it directly at the beginning of the call - so we are assured this is up to date with the latest call.
When you received the response of checkLeaderboardAvailability, you check `if (_requestedLeaderboardId !== leaderboardId)`. If that's true, you are getting a response for something outdated. You simply bail out (and don't error - you just don't care about the response anymore).

Extensions/Leaderboards/JsExtension.js
```diff
@@ -86,6 +91,48 @@ module.exports = {
       .setIncludeFile('Extensions/Leaderboards/leaderboardstools.js')
       .setFunctionName('gdjs.evtTools.leaderboards.hasLastEntrySaveFailed');
 
+    extension
+      .addCondition(
```
4ian, line 95: Nitpicking: we took the convention to start all conditions and actions by a capital (PascalCase)

Extensions/Leaderboards/JsExtension.js
```diff
@@ -86,6 +91,48 @@ module.exports = {
...
+        'isLeaderboardViewErrored',
+        _('Leaderboard display is errored'),
```
4ian, line 97:
```suggestion
        _('Leaderboard display has errored'),
```

Extensions/Leaderboards/JsExtension.js
```diff
@@ -86,6 +91,48 @@ module.exports = {
...
+        _('Check if the display of the leaderboard errored.'),
+        _('Leaderboard display is errored'),
```
4ian, line 99:
```suggestion
        _('Leaderboard display has errored'),
```

Extensions/Leaderboards/JsExtension.js
```diff
@@ -86,6 +91,48 @@ module.exports = {
...
+      .getCodeExtraInformation()
+      .setIncludeFile('Extensions/Leaderboards/leaderboardstools.js')
+      .setFunctionName('gdjs.evtTools.leaderboards.isLeaderboardViewErrored');
+
+    extension
+      .addCondition(
+        'isLeaderboardViewLoaded',
+        _('Leaderboard display is loaded'),
+        _('Check if the display of the leaderboard has finished loading.'),
+        _('Leaderboard display is loaded'),
```
4ian, line 113: I would even call this "Leaderboard display has loaded and is displayed on screen"

4ian (approved):
We're close! I think this will be ready to go after this round of comments!

Extensions/Leaderboards/leaderboardstools.ts
AlexandreSi: > animate supports started with iOS 13.1... I think it's good but we could almost put a try/catch around it to allow this to work (without animation) without crashing the game on iOS 12/13 (even if it's <2% of iOS usage).

Good idea, I'll add it

> Also, will this consume CPU even if not rendered? 🤔

Good question, I'll check

Extensions/Leaderboards/leaderboardstools.ts
AlexandreSi: That's the type returned by `setTimeout` 😅

Extensions/Leaderboards/leaderboardstools.ts
AlexandreSi: 😓

Extensions/Leaderboards/leaderboardstools.ts
AlexandreSi: I couldn't find something on the internet explaining how Animate would behave on a node that is not in the DOM.
I made a test on a dedicated HTML file. Here are the performance analysis by Chrome DevTools:
- First case: Loader is inserted in the div at start, then animated at 1.5s.

image

- Second case: Loader is inserted in the div at 3s, bu still animated at 1.5s

image

The animation seems to not require CPU when the element is not displayed: CPU line is hardly visible but:
- in the first case, we can see a little "peak" at 1.5s 
- in the second case, there seems to be a small CPU Usage when the element appears at 3s but none when the animation starts at 1.5s

AlexandreSi: Another point is that I found a few articles claiming that infinite CSS animations can consume a lot of CPU. But the Animate API doesn't seem to inject CSS on the element.

4ian: > The animation seems to not require CPU when the element is not displayed

Ok! I'm a bit paranoid on having any DOM work that would still the browser CPU time from the game and that would make the browser not 100% dedicated on running the game frames.

PR #3787, opened 2022-03-29
Add a condition to check if the mouse is inside the window.

D8H:
When the mouse is outside of the canvas GDevelop doesn't update the mouse position and mouse button released (it seems a very hard thing to do). This can cause 2 issues:

1. game creators can't write events to check if the mouse stopped near the canvas edge or left it.
2. game creator can't write events to check if the mouse is released outside of the canvas (PIXI has a mouseupoutside event, maybe it can be possible to have something like it).

## EdgeScrollCamera  extension (1)
* https://github.com/GDevelopApp/GDevelop-extensions/pull/384

Example from Tristan extension with a fix on scrolling up:
* https://liluo.io/games/6fd0cf36-b14b-4180-b7a2-dc5d7613d04a?dev=true
  Try to move the mouse quickly to leave from the bottom. It will often not scroll. Whereas leaving from the top will always scroll.

## QIX example (2)
There is a bug in this game: when leaving the canvas while drawing and releasing the button outside, the game will continue to draw even if the mouse button is not pressed.
* https://d8h.itch.io/quick-impulse-xpansion
With the new condition the drawing could be ended when the mouse leave the canvas.

### DIFF
--- a/Core/GDCore/Extensions/Builtin/MouseExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/MouseExtension.cpp
@@ -191,6 +191,17 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsMouseExtension(
   extension.AddDuplicatedExpression("SourisY", "MouseY").SetHidden();
 
   extension
+      .AddCondition("IsMouseInsideCanvas",
+                    _("Mouse is inside the window"),
+                    _("Check if the mouse is inside the window."),
+                    _("The mouse is inside the window"),
+                    "",
+                    "res/conditions/mouse24.png",
+                    "res/conditions/mouse.png")
+      .AddCodeOnlyParameter("currentScene", "")
+      .MarkAsAdvanced();
+
+  extension
       .AddCondition("MouseButtonPressed",
                     _("Mouse button pressed or touch held"),
                     _("Check if the specified mouse button is pressed or "
--- a/Core/GDCore/Extensions/Builtin/MouseExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/MouseExtension.cpp
@@ -192,9 +192,9 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsMouseExtension(
 
   extension
       .AddCondition("IsMouseInsideCanvas",
-                    _("Mouse is inside the window"),
-                    _("Check if the mouse is inside the window."),
-                    _("The mouse is inside the window"),
+                    _("Mouse cursor is inside the window"),
+                    _("Check if the mouse cursor is inside the window."),
+                    _("The mouse cursor is inside the window"),
                     "",
                     "res/conditions/mouse24.png",
                     "res/conditions/mouse.png")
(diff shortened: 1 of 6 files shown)


Core/GDCore/Extensions/Builtin/MouseExtension.cpp
```diff
@@ -190,6 +190,17 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsMouseExtension(
   extension.AddDuplicatedCondition("SourisY", "MouseY").SetHidden();
   extension.AddDuplicatedExpression("SourisY", "MouseY").SetHidden();
 
+  extension
+      .AddCondition("IsMouseInsideCanvas",
+                    _("Mouse is inside the window"),
+                    _("Check if the mouse is inside the window."),
```
4ian, line 196:
```suggestion
                    _("Check if the mouse cursor is inside the window."),
```

Core/GDCore/Extensions/Builtin/MouseExtension.cpp
```diff
@@ -190,6 +190,17 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsMouseExtension(
   extension.AddDuplicatedCondition("SourisY", "MouseY").SetHidden();
   extension.AddDuplicatedExpression("SourisY", "MouseY").SetHidden();
 
+  extension
+      .AddCondition("IsMouseInsideCanvas",
+                    _("Mouse is inside the window"),
```
4ian, line 195:
```suggestion
                    _("Mouse cursor is inside the window"),
```

Core/GDCore/Extensions/Builtin/MouseExtension.cpp
```diff
@@ -190,6 +190,17 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsMouseExtension(
...
+                    _("Check if the mouse is inside the window."),
+                    _("The mouse is inside the window"),
```
4ian, line 197:
```suggestion
                    _("The mouse cursor is inside the window"),
```

4ian (commented):
Looking good overall - can you check with a phone/touchscreen how is this reacting? This should not trigger a mouseleave, but I want to be sure.

D8H:
I updated the example. With the new fix, the condition works on touch device too.

D8H:
> On my phone with [the example](https://liluo.io/games/6fd0cf36-b14b-4180-b7a2-dc5d7613d04a?dev=true), I load the webpage, touch to nothing, and the ship move top the upper-left corner without I had to touch.

The default mouse position is (0, 0). The extension can't do anything about it.

> (Optionnal: And on PC when I switch to another windows or app as the cursor is detected outside then the latest direction is used to the ship. The problem is the game still running and the ship still moving even if I don't play the game because I'm focused on another app/tabs/window.)

Yes, I hesitated to do the same kind of condition for focus. It could be interesting to be used to show a pause menu.

4ian (approved):
Looks good with the proper handling for touches 👍

PR #3788, opened 2022-03-30
Rework the Game publishing on Liluo + allow defining if the game is discoverable

ClementPasteau:
Multiple things:

- Rework the Liluo publishing workflow to display a preview of what the game will look like, based on the project properties
- Rework the build creation, so that it doesn't save game information. Only when the build is published to the game.
- Add the thumbnail to the Game Edit Form in the game dashboard to reuse it in the publication flow.
- Add the discoverable attribute to the Game edit forms
- Fix UX loading interactions a bit everywhere + warning & error messages
- Reduce Checkbox's padding to fix the overlapping when multiple checkboxes are displayed
- Allow saving the project automatically when publishing

Screen Recording 2022-03-30 at 17 04 37


newIDE/app/src/Export/ExportDialog/ExportLauncher.js
```diff
@@ -157,30 +155,22 @@ export default class ExportLauncher extends Component<Props, State> {
     }
   };
 
-  registerAndUpdateGame = async () => {
+  registerGameIfNot = async () => {
```
4ian, line 158: Good renaming 👍

newIDE/app/src/Export/GenericExporters/OnlineWebExport/OnlineGameLink.js
```diff
@@ -0,0 +1,384 @@
...
+              message={<Trans>Copied to clipboard!</Trans>}
+              visible={showCopiedInfoBar}
+              hide={() => setShowCopiedInfoBar(false)}
+            />
+          </Dialog>
+          {game && build && (
+            <OnlineGamePropertiesDialog
+              project={project}
+              buildId={build.id}
+              open={isOnlineGamePropertiesDialogOpen}
```
4ian, line 363: What do you think of:

```suggestion
          {game && build && isOnlineGamePropertiesDialogOpen && (
            <OnlineGamePropertiesDialog
              project={project}
              buildId={build.id}
```

newIDE/app/src/GameDashboard/PublicGameProperties.js
```diff
@@ -57,55 +66,91 @@ function PublicGameProperties({
...
-            value={name}
-            onChange={setName}
-            autoFocus
-          />
-          {setCategories && (
-            <SemiControlledMultiAutoComplete
-              hintText={t`Select a genre`}
-              floatingLabelText={<Trans>Genres</Trans>}
-              helperText={
+          <Line noMargin>
```
4ian, line 80: Take a look at what happens on a small screen. You might want to use a ResponsiveLineStackLayout to go to a 1 column stack on small screens automatically (the message "To update your thumbnail" might need to be moved to a small component and have its place changed according to the window size though..).

newIDE/app/src/GameDashboard/PublicGamePropertiesDialog.js
```diff
@@ -16,7 +16,8 @@ import { type PublicGame } from '../Utils/GDevelopServices/Game';
  * Changes that are not stored in the Project.
```
4ian, line 16: Good change of folder 👍

4ian (approved):

newIDE/app/src/Export/GenericExporters/OnlineWebExport/OnlineGameLink.js
```diff
@@ -0,0 +1,384 @@
...
+              </Text>
+              <LinearProgress />
+            </>
+          )}
+          <Dialog
+            title={<Trans>Share your game</Trans>}
+            actions={dialogActions}
+            open={isShareDialogOpen}
+            onRequestClose={() => setIsShareDialogOpen(false)}
+          >
```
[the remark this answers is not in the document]
ClementPasteau: Yeah I'm not sure why it was done like that, I copied it without thinking. Will change 👍

newIDE/app/src/UI/Checkbox.js
```diff
@@ -2,6 +2,16 @@
 import * as React from 'react';
 import FormControlLabel from '@material-ui/core/FormControlLabel';
 import Checkbox from '@material-ui/core/Checkbox';
+import { makeStyles } from '@material-ui/core/styles';
+
+// Reduce checkbox size to avoid overlapping with other checkboxes.
+const useStyles = makeStyles({
+  root: {
+    marginLeft: 9,
```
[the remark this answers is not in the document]
ClementPasteau: Material UI was applying a padding of 9px and a margin of -9px, I'm not sure why... Maybe to have a large ripple effect around the checkbox.
Because we were reducing the size of the line, the checkbox were overlapping with top & bottom checkboxes

I'll double check everywhere

newIDE/app/src/GameDashboard/PublicGameProperties.js
ClementPasteau: As I think mobile is not the first use, I'll leave the message where it is for the moment. If this changes, this will be a nice problem to fix :)

newIDE/app/src/BrowserApp.js
```diff
@@ -71,6 +71,7 @@ export const create = (authentication: Authentication) => {
               renderExportDialog={props => (
                 <ExportDialog
                   project={props.project}
+                  saveProject={props.saveProject}
```
4ian, line 74: I would call this `onSaveProject` to show that it's a callback. In the sense that it's something to call to show your intention, as the child component, to do an action, but this is just something that you ask, and you delegate the doing to another function (which is probably called saveProject in a parent somewhere)

newIDE/app/src/Export/GenericExporters/OnlineWebExport/OnlineGamePropertiesDialog.js
```diff
@@ -0,0 +1,143 @@
...
+  const [orientation, setOrientation] = React.useState<string>(
+    project.getOrientation()
+  );
+  const [discoverable, setDiscoverable] = React.useState<boolean>(
+    !!game.discoverable
+  );
+  const thumbnailUrl = getWebBuildThumbnailUrl(project, buildId);
+
+  const saveProjectAndPublish = async () => {
+    await saveProject();
```
4ian, line 63: Do you think you should try/catch around this line?

newIDE/app/src/MainFrame/index.js
```diff
@@ -2266,6 +2266,7 @@ const MainFrame = (props: Props) => {
             openSubscriptionDialog(true);
           },
           project: state.currentProject,
+          saveProject,
```
4ian, line 2269: (see here the similar prop drilling we've done for "onChangeSubscription", that was called "onChangeSubscription" and not "changeSubscription". This is because a prop is called in *reaction* to something. It almost reads as english: "on change subscription requested, please close the export dialog and open the subscription dialog").

newIDE/app/src/UI/SemiControlledMultiAutoComplete.js
```diff
@@ -53,6 +53,7 @@ export default function SemiControlledMultiAutoComplete(props: Props) {
             ) ||
             (props.optionsLimit && props.value.length >= props.optionsLimit)
           }
+          getOptionSelected={(option, value) => option.value === value.value}
```
4ian, line 56: Interesting, why this?

ClementPasteau: this is the warning we've seen in the console if you remember? 
a MUI autocomplete tries to know if an option is selected or not.
by default, it compares the option itself (which is not equal because it's an object)

4ian: Great!

4ian (approved):
Looks good after the comments are addressed 👍
Great improvements.

newIDE/app/src/Export/GenericExporters/OnlineWebExport/OnlineGamePropertiesDialog.js
ClementPasteau: the existing function already has a try catch / showErrorBox behavior built into it. So I think it's not necessary

PR #3789, opened 2022-03-30
Fix the camera rounding to work when following objects with even dimensions

D8H:
Fix the camera rounding to work when following objects with even dimensions instead of odd ones.

I actually used `ceil` because otherwise, when the character jumps, it seems to move from one pixel on screen.

Direct following
* https://liluo.io/instant-builds/23ae9daf-9240-4ac8-9456-7fd6ce36cb4f?dev=true

Linear following
* https://liluo.io/instant-builds/cb33a8eb-7840-40ae-8622-1722ae730e2c?dev=true

There will be 2 alerts next to each other, but I think it's not an issue.

image

### DIFF
--- a/GDJS/Runtime/pixi-renderers/layer-pixi-renderer.ts
+++ b/GDJS/Runtime/pixi-renderers/layer-pixi-renderer.ts
@@ -85,10 +85,10 @@ namespace gdjs {
         // the sprite one and it changes in which direction sprites are rounded.
         // It makes sprites rounding inconsistent with each other
         // and they seems to move on pixel left and right.
-        this._pixiContainer.position.x = Math.round(
+        this._pixiContainer.position.x = Math.floor(
           this._pixiContainer.position.x
         );
-        this._pixiContainer.position.y = Math.round(
+        this._pixiContainer.position.y = Math.floor(
           this._pixiContainer.position.y
         );
       }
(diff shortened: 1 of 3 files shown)


GDJS/Runtime/pixi-renderers/layer-pixi-renderer.ts
```diff
@@ -85,10 +85,10 @@ namespace gdjs {
         // the sprite one and it changes in which direction sprites are rounded.
         // It makes sprites rounding inconsistent with each other
         // and they seems to move on pixel left and right.
-        this._pixiContainer.position.x = Math.round(
+        this._pixiContainer.position.x = Math.ceil(
```
4ian, line 88: Maybe add a link to this PR in the comment so that we can find the context back and why ceil was chosen.

4ian (approved):
Thanks! 👍

GDJS/Runtime/pixi-renderers/layer-pixi-renderer.ts
```diff
@@ -81,14 +81,36 @@ namespace gdjs {
...
+        // onScreenPosition = onScenePosition - cameraPosition
+        // onScreenPosition = 980.75 - 200.75
+        // onScreenPosition = 780
+        //
+        // Doing the calculus with rounding:
+        // onScreenPosition = floor(onScenePosition + ceil(-cameraPosition))
+        // onScreenPosition = floor(980.75 + ceil(-200.75))
+        // onScreenPosition = floor(980.75 - 200)
+        // onScreenPosition = floor(780.75)
+        // onScreenPosition = 780
```
D8H, line 109: I added this explanation.

4ian: Excellent! Love this example, that's very clear now (notably the fact that there is a minus in the positioning, so ceil instead of floor).

4ian:
Thanks again for investigating this!

PR #3792, opened 2022-03-31
Use an icon for the Home tab

4ian:
Can help memorising that this is the home page and not the project manager (even if a change of layout will be needed at some point to truly fix the navigation) + take less space + visually more appealing.

image
image
image

Also fix an unused style and fix the `height` of `closableTabs` not respected previously (the tab height was depending on the text, not on this setting).


newIDE/app/src/UI/ClosableTabs.js
```diff
@@ -192,12 +189,15 @@ export function ClosableTab({
...
+                    ...styles.tabLabelAndIcon,
+                    height: muiTheme.closableTabs.height,
                     color: textColor,
                     fontFamily: muiTheme.closableTabs.fontFamily,
                   }}
                 >
-                  {label}
+                  {icon}
+                  {icon && label ? <Spacer /> : null}
+                  <span style={styles.tabLabel}>{label}</span>
```
ClementPasteau, line 200: this doesn't create a weird space with the span if there's no label?

newIDE/app/src/stories/componentStories/ClosableTabs.stories.js
```diff
@@ -0,0 +1,304 @@
+// @flow
```
ClementPasteau, line 1: 🎉

newIDE/app/src/UI/ClosableTabs.js
```diff
@@ -192,12 +189,15 @@ export function ClosableTab({
...
                   style={{
-                    ...styles.tabLabel,
+                    ...styles.tabLabelAndIcon,
+                    height: muiTheme.closableTabs.height,
                     color: textColor,
                     fontFamily: muiTheme.closableTabs.fontFamily,
                   }}
                 >
-                  {label}
+                  {icon}
```
ClementPasteau, line 198: for accessibility purpose, should a label be used as an alt/title for the icon/tab?

ClementPasteau (approved):
lgtm 👍

newIDE/app/src/UI/ClosableTabs.js
4ian: Great remark, I've added a title to the Home icon, which will give a tooltip and an equivalent text for screen readers ("Provides a human-readable title for the element that contains it. The title element may be rendered by a graphical user agent as a tooltip. It may be rendered as speech by a speech synthesizer.")

newIDE/app/src/UI/ClosableTabs.js
4ian: Good spot. It's ok because the span size is 0x0 if no label is provided, so visually it's ok. But in an effort to reduce useless DOM element in the document, I've added a check 👍

PR #3794, opened 2022-04-01
Fine tune leaderboards

AlexandreSi:


newIDE/app/src/Utils/GDevelopServices/Play.js
```diff
@@ -105,7 +121,7 @@ export const listLeaderboardEntries = async (
   nextPageUri: ?string,
 |}> => {
   const uri =
-    options.forceUri || `/game/${gameId}/leaderboard/${leaderboardId}/entries`;
+    options.forceUri || `/game/${gameId}/leaderboard/${leaderboardId}/entry`;
```
4ian, line 124: 👍

Extensions/Leaderboards/JsExtension.js
```diff
@@ -75,29 +75,57 @@ module.exports = {
       )
       .getCodeExtraInformation()
       .setIncludeFile('Extensions/Leaderboards/leaderboardstools.js')
-      .setFunctionName('gdjs.evtTools.leaderboards.setPlayerScore');
+      .setFunctionName('gdjs.evtTools.leaderboards.savePlayerScore');
+
+    extension
+      .addCondition(
```
4ian, line 81: I just had an idea: should we ask, for these 3 conditions related to the score sent to a leaderboard (errored, succeeded, is saving), for the leaderboard id? That's not making things way more complicated for the creator, and this can help the creators to check what happened if they have multiple leaderboards.
I had this idea while reading the code where you mention that a refactoring is needed to store the last player name/last score/last saved at/potentially other stuff indexed by leaderboard id.

I'm not entirely sure but I think this will naturally come as you rework the checks to be per leaderboard rather than global.

Note that the leaderboard display view is single, so it's fine to have the conditions related to this not taking any parameter.

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
```diff
@@ -377,6 +377,8 @@ export const LeaderboardAdmin = ({ onLoading, project }: Props) => {
 
   React.useEffect(() => {
     if (currentLeaderboard) onFetchLeaderboardEntries();
+    // This has to be executed on component mount to refresh entries on each admin opening
+    // eslint-disable-next-line
   }, []);
```
4ian, line 382: Thank you for fixing my stuff 😇

AlexandreSi: 🤗

Extensions/Leaderboards/leaderboardstools.ts
```diff
@@ -45,74 +116,196 @@ namespace gdjs {
...
+        hasEnded,
+      }: {
+        hasEnded: boolean;
+      }): ScoreSavingData | null {
+        const dateFieldToUse = hasEnded
+          ? 'lastScoreSavingSucceededAt'
+          : 'lastScoreSavingStartedAt';
+        const scoreSavingDataArray = Object.values(
+          _savingDataByLeaderboard
+        ).filter((scoreSavingData) => !!scoreSavingData[dateFieldToUse]);
```
4ian, line 129: This is a bit bypassing the type system, I would rather do something like:


```suggestion
        const scoreSavingDataArray = Object.values(
          _savingDataByLeaderboard
        ).filter((scoreSavingData) => !!(hasEnded
          ? scoreSavingData.lastScoreSavingSucceededAt
          : scoreSavingData.lastScoreSavingStartedAt));
```

(sure you evaluate the condition multiple times, but it's not less efficient than accessing a property by a string and at least TS will tell you when you rename a property that you must do it here)

Extensions/Leaderboards/leaderboardstools.ts
```diff
@@ -45,74 +116,196 @@ namespace gdjs {
...
+        if (scoreSavingDataArray.length === 0) return null;
+
+        let lastScoreSavingData = scoreSavingDataArray[0];
+        scoreSavingDataArray.forEach((scoreSavingData) => {
+          const currentItemDate = scoreSavingData[dateFieldToUse];
+          const lastItemDate = lastScoreSavingData[dateFieldToUse];
```
4ian, line 135: Same here. If you really don't like repeat the condition (which I can understand, this is making the code harder to read), use a closure (see other comment)

Extensions/Leaderboards/leaderboardstools.ts
4ian: Or better for readibilty, use a function: `filter((scoreSavingData) => !!getDateField(scoreSavingData))`

with `getDateField = (scoreSavingData: ScoreSavingData) => hasEnded ? scoreSavingData.xxx : scoreSavingData.yyy)`.

You pay a function call which is has some overhead (callstack, jump, cache trashing), but we don't care about this (not performance sensitive code)

Extensions/Leaderboards/leaderboardstools.ts
```diff
@@ -45,74 +116,196 @@ namespace gdjs {
...
-                return response.statusText;
-              } else {
-                _lastScore = score;
-                _lastPlayerName = playerName;
-                _lastErrorCode = response.status;
-                return response.text();
+
+            scoreSavingData.closeSaving();
+
+            response.text().then(
```
4ian, line 231: Nit: 
```suggestion
            return response.text().then(
```

So that someone who would in the future await/`then` on the initial promise is not surprised about this not being finished before their new code is called.

Extensions/Leaderboards/leaderboardstools.ts
```diff
@@ -45,74 +116,196 @@ namespace gdjs {
...
       };
 
-      export const hasLastEntrySaveFailed = function () {
-        return _lastErrorCode && _lastErrorCode >= 400;
+      export const isSaving = function (leaderboardId?: string): boolean {
+        return getLeaderboardOrLastLeaderboardState(
+          'isScoreSaving',
+          leaderboardId
+        );
+      };
```
4ian, line 277: Prefer a typesafe equivalent:
```suggestion
      
      export const isSaving = function (leaderboardId?: string): boolean {
        const savingState = getLeaderboardOrLastLeaderboardSavingState(leaderboardId);
        return savingState ? savingState.isScoreSaving : false
      };
```

Extensions/Leaderboards/leaderboardstools.ts
```diff
@@ -45,74 +116,196 @@ namespace gdjs {
...
+          }
+          return _savingDataByLeaderboard[leaderboardId][state];
         }
+        const lastScoreSavingData = getLastScoreSavingData({
+          hasEnded: state === 'hasScoreBeenSaved',
+        });
+        if (!lastScoreSavingData) {
+          return false;
+        }
+        return lastScoreSavingData[state];
```
4ian, line 269: Nits:
- Prefer to return the ScoreSavingData or null
- Choose a name (ScoreSavingData or ScoreSavingState) and use this rather than "LeaderboardState" to not create multiple concepts in the head of the reader.

```suggestion
      const getLeaderboardOrLastLeaderboardSavingData = function (
        leaderboardId?: string
      ): boolean {
        if (leaderboardId) {
          return _savingDataByLeaderboard[leaderboardId] || null;
        }
        const lastScoreSavingData = getLastScoreSavingData({
          hasEnded: state === 'hasScoreBeenSaved',
        });
        return lastScoreSavingData || null;
```

4ian (approved):
LGTM with typesafe functions + normalised naming (ScoreSavingData or ScoreSavingState or LeaderboardState => pick one) 👍

AlexandreSi:
Thanks for the feedback on the type safety, it is way clearer.

PR #3798, opened 2022-04-03
Add string selector to "Change input type" for text input object

Elairyx7301:
What's the difference between `.useStandardParameters()` and `.addParameter()`?
Is that change ok like this?


Elairyx7301:
Also, is it normal that all my commits are shown above?
> > image

Can I only select the relevant commits or is it just how it is? 🤔

4ian:
> Can I only select the relevant commits or is it just how it is? 🤔

Not a big deal here, but you should always go to the master branch, pull the latest changes, then create a new branch and start your changes here (and then open a PR thanks to this branch).

> What's the difference between `.useStandardParameters()` and `.addParameter()`?
> Is that change ok like this?

Unfortunately that won't work like this, because useStandardParameters is not handling the second parameter.
`useStandardParameters` is a special function that internally calls multiple times `addParameter` to add, well, some conventional parameters according to the type being passed.

We could add support for this second parameter, but this needs to be done carefully I think in the C++ codebase.

PR #3800, opened 2022-04-03
[TopDownMovement] Add input condition and velocity action.

D8H:
I'm working on an extension to make the Platformer Character and the Top-Down Movement behaviors pixel perfect. To do so, I needed some new functions:

Action
* Speed on the X axis
* Speed on the Y axis

Condition
* Control pressed or simulated

Expression
* Stick angle

## Examples
Top-down movement on a grid (it uses a cell as a big pixel):
* https://liluo.io/games/6e9ff72a-3e34-412a-9e62-1658548e1459?dev=true
  It also allows to do random PNJ movement without relying on path finding.

Pixel perfect platformer (notice that the character always end up on a pixel when moving left or right)
* https://liluo.io/instant-builds/d214365c-0b65-4d11-b65b-b533ba1d2259

### DIFF
--- a/Extensions/TopDownMovementBehavior/Extension.cpp
+++ b/Extensions/TopDownMovementBehavior/Extension.cpp
@@ -145,6 +145,28 @@ void DeclareTopDownMovementBehaviorExtension(gd::PlatformExtension& extension) {
       .SetIncludeFile(
           "TopDownMovementBehavior/TopDownMovementRuntimeBehavior.h");
 
+    aut.AddScopedCondition("IsUsingControl",
+                  _("Control pressed or simulated"),
+                  _("A control was applied from a default control or a simulated by an action."),
+                  _("_PARAM0_ has the _PARAM2_ key pressed or simulated"),
+                  _("Controls"),
+                  "res/conditions/keyboard24.png",
+                  "res/conditions/keyboard.png")
+        .AddParameter("object", _("Object"))
+        .AddParameter("behavior", _("Behavior"), "TopDownMovementBehavior")
+        .AddParameter("stringWithSelector",
+                    _("Key"),
+                    "[\"Left\", \"Right\", \"Up\", \"Down\", \"Stick\"]")
+        .MarkAsAdvanced();
+
+  aut.AddExpression("StickAngle",
+                    _("Stick angle"),
+                    _("Return the angle of the simulated stick input (in degrees)"),
+                    _("Controls"),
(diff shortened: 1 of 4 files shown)


Extensions/TopDownMovementBehavior/Extension.cpp
```diff
@@ -145,6 +145,28 @@ void DeclareTopDownMovementBehaviorExtension(gd::PlatformExtension& extension) {
       .SetIncludeFile(
           "TopDownMovementBehavior/TopDownMovementRuntimeBehavior.h");
 
+    aut.AddScopedCondition("IsUsingControl",
+                  _("Control pressed or simulated"),
+                  _("A control was applied from a default control or a simulated by an action."),
```
4ian, line 150:
```suggestion
                  _("A control was applied from a default control or simulated by an action."),
```

Extensions/TopDownMovementBehavior/topdownmovementruntimebehavior.ts
```diff
@@ -35,10 +35,11 @@ namespace gdjs {
...
     private _upKey: boolean = false;
     private _downKey: boolean = false;
-    private _leftKeyPressedDuration: integer = -1;
-    private _rightKeyPressedDuration: integer = -1;
-    private _upKeyPressedDuration: integer = -1;
-    private _downKeyPressedDuration: integer = -1;
+    private _leftKeyPressedDuration: integer = 0;
+    private _rightKeyPressedDuration: integer = 0;
+    private _upKeyPressedDuration: integer = 0;
+    private _downKeyPressedDuration: integer = 0;
```
4ian, line 41:
```suggestion
    private _leftKeyPressedDuration: float = 0;
    private _rightKeyPressedDuration: float = 0;
    private _upKeyPressedDuration: float = 0;
    private _downKeyPressedDuration: float = 0;
```

4ian (approved):

4ian:
The examples look awesome 🔥

Extensions/TopDownMovementBehavior/topdownmovementruntimebehavior.ts
D8H: `getElapsedTime` gives milliseconds.

4ian: Well I think we can still have sub milliseconds precision?

PR #3803, opened 2022-04-05
Allow to define a default value for a string with options property.

D8H:
image

### DIFF
--- a/newIDE/app/src/EventsBasedBehaviorEditor/EventsBasedBehaviorPropertiesEditor.js
+++ b/newIDE/app/src/EventsBasedBehaviorEditor/EventsBasedBehaviorPropertiesEditor.js
@@ -124,10 +124,14 @@ export default class EventsBasedBehaviorPropertiesEditor extends React.Component
 
   _setChoiceExtraInfo = (property: gdNamedPropertyDescriptor) => {
     return (newExtraInfo: Array<string>) => {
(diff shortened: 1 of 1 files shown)


newIDE/app/src/EventsBasedBehaviorEditor/EventsBasedBehaviorPropertiesEditor.js
```diff
@@ -124,10 +124,14 @@ export default class EventsBasedBehaviorPropertiesEditor extends React.Component
...
   _setChoiceExtraInfo = (property: gdNamedPropertyDescriptor) => {
     return (newExtraInfo: Array<string>) => {
+      const defaultValueIndex = getExtraInfoArray(property).indexOf(
+        property.getValue()
+      );
       const vectorString = new gd.VectorString();
       newExtraInfo.forEach(item => vectorString.push_back(item));
       property.setExtraInfo(vectorString);
       vectorString.delete();
+      property.setValue(newExtraInfo[defaultValueIndex]);
```
4ian, line 134: Any chance this is defaultValueIndex can be -1? In which case, prefer to do
```suggestion
      property.setValue(newExtraInfo[defaultValueIndex] || "");
```

to avoid passing undefined.
For example, if newExtraInfo is empty.

4ian (approved):
Looks good but careful about potential edge cases

Issue #3801, opened 2022-04-04
Upgrade to Electron 18


4ian:
It's a long time (1.5 year) since we last updated Electron. We're lagging behind several major versions now and it may be the reason for some performance issue on some OS.

We should:
- Upgrade to the latest Electron version
- Adapt code using `remote` to use `@electron/remote`...
  -  ...before migrating (in another PR, less urgent) to exposing just the necessary apis through the Electron preload.js script
- Adapt code using ipc (we have to expose ipc through preload.js too).
- Adapt code of the third party editor bridges
- Also upgrade the games runtime I think (can be done in a separate PR).
- Ensure "it works". Check on Windows/macOS/Linux, ensure auto update works.
  - Check if it works better with Flatpak (https://github.com/flathub/flathub/pull/2964)

I'm ready to open a large bounty to do that if someone wants to take this task :) 
Bonus point if you do the removal of `remote`/`@electron/remote` entirely in favor of an API exposed in preload.js - but I'm willing not to do that first to ensure we can release asap a version with a new Electron version without breakages.

4ian:
@nilaymaj You did the last update, would you be interested in taking this? Happy to open a large bounty if you're interested, let me know :)

nilaymaj:
Sure, let me try this - it's much larger than the earlier upgrade (10 major versions!). I'll start with going through and summarising the Electron changelog, may take a couple of days.

4ian:
We're probably only interesting in the breaking changes ;) https://www.electronjs.org/docs/latest/breaking-changes

From a quick look, these are the changes we need to take care about:
- "Removed: remote module" => see my original message (we should use `@electron/remote` and in the long term move toward a "preload.js" approach, but this should be done carefully one feature by one).
- "Removed: app.allowRendererProcessReuse" => remove the related code and check there are no issues on Windows when loading/saving files (we had some issues with `fs` on Windows on Electron 8 with this flag).
- "Removed: BrowserWindow extension APIs" => should switch to using `session` object methods.
- "Deprecated: WebContents new-window event" => to replace with `setWindowOpenHandler`
- `contextIsolation` should be set to false (and nodeIntegration kept to `true`). In the long term this should be removed (in favor of the "preload.js" approach).
- "API Changed: shell.openItem is now shell.openPath" => quick update for us to do too.

PR #3806, opened 2022-04-06
Upgrade to Electron 18

nilaymaj:
This PR is to upgrade the Electron version used for the desktop app to v18. (#3801)

The approach I'm following is to bump the Electron version one major version at a time, encountering and resolving breaking changes and regressions as they arise. Progress till now:

- **v8.2.1 -> v9.4.4**: 
  - **Deprecation:** `BrowserWindow` extension APIs deprecated in favour of `session` API.
  - **Regression:** accessing local files using the `file` protocol breaks (see Electron [issue](https://github.com/electron/electron/issues/23757). [Fix](https://github.com/electron/electron/pull/28489) is backported only to v11). I've checked and it's indeed resolved in v11, so I'm ignoring this for now.
  - **API change:** synchronous `shell.openItem` changed to asynchonous `shell.openPath`.

- **v9.4.4 -> v10.4.7**:
  - **Deprecation:** `remote` module deprecated, now available only on setting the flag `enableRemoteModule`. For this version bump we use the flag instead of performing the full `remote` refactor.
  - **Regression:** External editors get clipped to height 150px (probably due to a Chrome rendering update). Looking into it now.

- **v10.4.7 -> v11.5.0**: No relevant breaking changes, no regressions found yet.

- **v11.5.0 -> v12.2.3**:
  - **Default changed:** `contextIsolation` default value is `true` now. To use Node.js APIs (like `require`) in the renderer process, we keep `contextIsolation` false and `nodeIntegration` true.
  - **Deprecation:** `remote` module is deprecated. Need to switch all `remote` module usage to `@electron/remote`.

- **v12.2.3 -> v13.6.9**:
  - **Deprecation:** `new-window` event is replaced by separate method `setWindowOpenHandler`, where our callback's return value specifies whether action is allowed or deny.
  - **Removal:** Extension API on `BrowserWindow` is removed. We've already refactored this in the v9 upgrade above.

- **v13.6.9 -> v14.2.9**:
  - **Usage change:** In addition to module initialization call, the `@electron/remote` module must now be explicitly enabled for each `BrowserWindow` separately.
  - **Removal:** the deprecated `remote` module is now removed. One usage missed earlier was found and fixed. Searches for `'remote'` (as in `require('remote')`) and `.remote` (as in `electron.remote`) have 0 results now. Hopefully that's all.
  - **Removal:** the `allowRendererProcessReuse` flag is now removed. We've already removed this in the v11 upgrade.

- **v14.2.9 -> v15.5.1**:
  - **Default changed:** the default value of `nativeWindowOpen` is now true. Doesn't seem to affect, previews and external editors are working fine.

- **v15.5.1 -> v16.2.1**: No relevant breaking changes, no regressions found yet.

- **v16.2.1 -> v17.3.1**: No relevant breaking changes, no regressions found yet.

- **v17.3.1 -> v18.0.1**: No relevant breaking changes, no regressions found yet.

A side note: not necessary but after reaching v18, if possible we may think about enabling the `webSecurity` flag. Will require some changes like adding a custom file protocol for accessing local files (and then modifiying all file urls to use this custom protocol)... not sure if it's worth it.

Minor fixes and improvements to do:
- Add error handling to all `shell.openPath` calls (perhaps by adding a tiny wrapper function). A `console.error` should do for reporting the error?

### DIFF
--- a/newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionOpener.js
+++ b/newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionOpener.js
@@ -2,7 +2,8 @@
 import optionalRequire from '../../Utils/OptionalRequire.js';
 const fs = optionalRequire('fs');
 const electron = optionalRequire('electron');
-const dialog = electron ? electron.remote.dialog : null;
+const remote = optionalRequire('@electron/remote');
+const dialog = electron ? remote.dialog : null;
 
 const readJSONFile = (filepath: string): Promise<Object> => {
   if (!fs) return Promise.reject('Filesystem is not supported.');
@@ -24,7 +25,7 @@ const readJSONFile = (filepath: string): Promise<Object> => {
 export default class LocalEventsFunctionsExtensionOpener {
   static chooseEventsFunctionExtensionFile = (): Promise<?string> => {
     if (!dialog) return Promise.reject('Not supported');
-    const browserWindow = electron.remote.getCurrentWindow();
+    const browserWindow = remote.getCurrentWindow();
 
     return dialog
       .showOpenDialog(browserWindow, {
--- a/newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionOpener.js
+++ b/newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionOpener.js
@@ -1,9 +1,8 @@
 // @flow
 import optionalRequire from '../../Utils/OptionalRequire.js';
 const fs = optionalRequire('fs');
-const electron = optionalRequire('electron');
 const remote = optionalRequire('@electron/remote');
-const dialog = electron ? remote.dialog : null;
+const dialog = remote ? remote.dialog : null;
 
 const readJSONFile = (filepath: string): Promise<Object> => {
   if (!fs) return Promise.reject('Filesystem is not supported.');
--- a/newIDE/electron-app/app/ModalWindow.js
+++ b/newIDE/electron-app/app/ModalWindow.js
@@ -38,7 +38,6 @@ const loadModalWindow = ({
     center: true,
     webPreferences: {
       webSecurity: false,
-      enableRemoteModule: true,
       // Allow Node.js API access in renderer process
       nodeIntegration: true,
       contextIsolation: false,
--- a/newIDE/electron-app/app/ModalWindow.js
+++ b/newIDE/electron-app/app/ModalWindow.js
@@ -39,6 +39,7 @@ const loadModalWindow = ({
     webPreferences: {
       webSecurity: false,
       nodeIntegration: true,
+      enableRemoteModule: true,
     },
   };
 
--- a/newIDE/electron-app/app/ModalWindow.js
+++ b/newIDE/electron-app/app/ModalWindow.js
@@ -48,6 +48,9 @@ const loadModalWindow = ({
   modalWindow = new BrowserWindow(windowOptions);
   modalWindow.setMenu(null);
 
+  // Enable `@electron/remote` module for renderer process
+  require('@electron/remote/main').enable(modalWindow.webContents);
+
   ipcMain.removeAllListeners(readyChannelName);
   ipcMain.on(readyChannelName, event => {
     onReady(modalWindow);
--- a/newIDE/electron-app/app/ModalWindow.js
+++ b/newIDE/electron-app/app/ModalWindow.js
@@ -38,8 +38,10 @@ const loadModalWindow = ({
     center: true,
     webPreferences: {
       webSecurity: false,
-      nodeIntegration: true,
       enableRemoteModule: true,
+      // Allow Node.js API access in renderer process
+      nodeIntegration: true,
+      contextIsolation: false,
     },
   };
 
--- a/newIDE/electron-app/app/ModalWindow.js
+++ b/newIDE/electron-app/app/ModalWindow.js
@@ -72,11 +72,11 @@ const loadModalWindow = ({
     }
   });
 
-  //Prevent opening any website or url inside Electron.
-  modalWindow.webContents.on('new-window', (e, url) => {
-    console.info('Opening in browser (because of new-window): ', url);
-    e.preventDefault();
-    electron.shell.openExternal(url);
+  // Prevent opening any website or url inside Electron
+  modalWindow.webContents.setWindowOpenHandler(details => {
+    console.info('Opening in browser (because of new window): ', details.url);
+    electron.shell.openExternal(details.url);
+    return { action: 'deny' };
   });
 
   modalWindow.on('closed', event => {
--- a/newIDE/electron-app/app/main.js
+++ b/newIDE/electron-app/app/main.js
@@ -77,8 +77,10 @@ app.on('ready', function() {
     y: args.y,
     webPreferences: {
       webSecurity: false, // Allow to access to local files,
-      nodeIntegration: true,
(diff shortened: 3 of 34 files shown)


newIDE/electron-app/app/main.js
```diff
@@ -46,17 +49,18 @@ const args = parseArgs(process.argv.slice(isDev ? 2 : 1), {
...
 protocol.registerSchemesAsPrivileged([{ scheme: 'gdide' }]);
 
+// TODO: Linked issue is resolved in v10.1.4+. Set to true (or remove) and check if stuff works
 // Should be set to true, which will be the default value in future Electron
 // versions, but then causes an issue on Windows where the `fs` module stops
 // working in the renderer process.
 // See https://github.com/electron/electron/issues/22119
 // For now, disable this as we rely heavily on `fs` in the renderer process.
-app.allowRendererProcessReuse = false;
+app.allowRendererProcessReuse = true;
```
nilaymaj, line 58: I think I committed this by mistake - this flag will be removed in v14 so I should've handled it later. Also, I haven't checked if it works on Windows on v10.4.7 yet (TODO)

arthuro555:
👏 This is a needed change. I attempted this not long ago but tried to go overkill mode and replace all usages of the remote modules with electron IPC but that quickly got into a big refactoring mess that I ended up abandoning. I hope you will succeed where I failed 😎

newIDE/electron-app/app/main.js
4ian: Feel free to go ahead with the upgrade and consider this is not needed anymore, as we want to land on a version that is at least Electron 14 I think :) 

So you can test on Windows once you have upgraded to a "sufficiently recent" version.
In other words, I don't want you to lose time trying to understand a bug that would have been fixed by a newer Electron version ;)

nilaymaj: Checked on Electron v11 on Windows, `fs` module is working (can open local project). Resolving this comment now.

4ian: I believe it was working but failing after you reload one or twice the development version by pressing Ctrl+R. But I think they resolved this at some point according to the issue on GitHub :)

4ian:
Great progress! I believe changing the remote module will be the most time consuming/painful stuff, but this is great already 👍

> A side note: not necessary but after reaching v18, if possible we may think about enabling the webSecurity flag. Will require some changes like adding a custom file protocol for accessing local files (and then modifiying all file urls to use this custom protocol)... not sure if it's worth it.

Not for now, but we'll definitely move toward this (much like we'll move toward removing the `remote`/`@electron/remote` module entirely, but this will be for later)

4ian:
(we'll also move toward removing nodeIntegration, but same thing that's a work in progress to do later which implies to refactor all usage of node packages in the renderer. Doable, but later and progressively! :))

nilaymaj:
Okay, I've done a search for `electron.remote` and refactored all such uses to the `@electron/remote` module. I think there are some usages left that did not appear in the search, because the deprecation warning is still printed in the console. These cases will be easier to find in v14 when the `remote` module is removed entirely.

A quick question: I saw [this comment](https://github.com/4ian/GDevelop/blob/master/newIDE/app/src/Utils/Window.js#L281-L282) in the source code. Do you think it's possible for this rare case to happen with the `@electron/remote` import too? While the linked use case may not be critical, other cases use `remote` for more important purposes - which is why the current changes are to fail hard in case `electron` is defined but `remote` isn't. Is that approach okay, or should this case be explicitly handled?

nilaymaj:
Reached Electron v18. Major versions after v14 didn't really seem to affect GDevelop in any way, so progress went rapid fast (hopefully not too fast). There's some things still to consider:
- The new `shell.openPath` API returns a promise specifying if the call was successful. What should happen if the call does fail? All the use cases of this API are on not-so-critical user actions (open project folder, open export folder) - and for invalid paths, at least on my Linux desktop I get a native dialog saying the path does not exist.
- See the question in my previous comment.

Now, towards full testing:
- [ ] Check that there are no UI regressions in the editor.
- [ ] Test native/file operations (project read-writes, game resources, exports, etc) for each platform (both in dev and production versions)
- [ ] Test scene editor and preview of multiple games containing varying elements (videos, html text inputs, etc).
- [ ] Exported games are still on the older Electron version so those should be okay, but maybe still test them just to be sure?
- [ ] Test that the debugger's functionalities are working fine.

4ian:
Awesome work! Thanks a lot for your super clear upgrade log. I'm busy this weekend but I'll try to take a look asap!
Great job!

newIDE/electron-app/app/ModalWindow.js
```diff
@@ -38,13 +38,19 @@ const loadModalWindow = ({
     center: true,
     webPreferences: {
       webSecurity: false,
+      enableRemoteModule: true,
```
4ian, line 41: Not sure if this is still need with the latest Electron version(s)?

newIDE/electron-app/app/main.js
```diff
@@ -84,7 +80,10 @@ app.on('ready', function() {
     y: args.y,
     webPreferences: {
       webSecurity: false, // Allow to access to local files,
+      enableRemoteModule: true,
```
4ian, line 83: (same question)

4ian (commented):
Great stuff! 
We can update games in a second time once we've confirmed this is working on Linux/macOS/Windows.
(and did a test to check that the auto update is working well - we'll do that once we've tested there are no obvious bugs)

newIDE/electron-app/app/ModalWindow.js
nilaymaj: Right, this isn't needed anymore. Removed now.

newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionOpener.js
```diff
@@ -2,7 +2,8 @@
 import optionalRequire from '../../Utils/OptionalRequire.js';
 const fs = optionalRequire('fs');
 const electron = optionalRequire('electron');
-const dialog = electron ? electron.remote.dialog : null;
+const remote = optionalRequire('@electron/remote');
+const dialog = electron ? remote.dialog : null;
```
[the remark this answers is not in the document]
nilaymaj: Yeah, this is related to my second query above. Using `remote` as check for these cases does make sense though, I've fixed this everywhere.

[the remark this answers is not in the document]
4ian: Thanks!

4ian:
I got a quick try at this, seems to work fine on Windows 11!
There is an issue where devtools can't be opened if you run the app from an elevated command prompt. There are workarounds, though they seem not perfect: https://github.com/microsoft/accessibility-insights-web/pull/2625/files

Apart from that, seems like the preview, export, debugger, file opening/saving works. Debugging with source maps now seems to work properly :) 

About your previous questions:

>  Is that approach okay, or should this case be explicitly handled?

I think it's fine. The try/catch was maybe due to an Electron bug in the past. Let's keep it but not add it for other usages.

> The new shell.openPath API returns a promise specifying if the call was successful. What should happen if the call does fail

Nothing critical, so I would say to just catch the error and log it at least. If necessary, we'll handle it later better, according to what the caller does with this.

Still have to test automatic export and probably merge this into an `experimental-build/xxx` branch so we can test production versions!

willianholtz:
Hey @4ian, did you ever test on a Linux distro, like Arch Linux itself?

4ian:
I usually give a quick try on Ubuntu, but feel free to double check on Arch Linux/other distros once we have an experimental version built :)

PR #3810, opened 2022-04-06
Send digest with leaderboard entry creation call

AlexandreSi:


AlexandreSi:
TODO: Clear leaderboard context when switching projects

Extensions/Leaderboards/leaderboardstools.ts
```diff
@@ -1,7 +1,15 @@
 namespace gdjs {
+  declare var jsSHA: any;
+
   const logger = new gdjs.Logger('Leaderboards');
   export namespace evtTools {
     export namespace leaderboards {
+      const computeDigest = (payload: string): string => {
+        const shaObj = new jsSHA('SHA-256', 'TEXT', { encoding: 'UTF8' });
```
ClementPasteau, line 8: out of curiosity, how did you pick this library?

ClementPasteau (commented):
lgtm, I'll let @4ian approve this one :)

GDJS/Runtime/libs/sha256.js
```diff
@@ -0,0 +1,740 @@
+/**
```
arthuro555, line 1: Maybe a minified version could be used since this file is untransformed?

Extensions/Leaderboards/leaderboardstools.ts
```diff
@@ -1,7 +1,15 @@
 namespace gdjs {
+  declare var jsSHA: any;
```
arthuro555, line 2: It doesn't seem that hard to type. At least even if not describing the whole library then its usage here:
```suggestion
  declare const jsSHA: (algorithm: "SHA-256", type: "TEXT", options: {encoding: "UTF8"}) => { getHash: (format: "B64") => string, update: (payload: string) => void, };
```

arthuro555 (approved):

Extensions/Leaderboards/leaderboardstools.ts
4ian: Library implementing just a single algorithm and lightweight. Something not focused on performance or completness (we don't want multiple megabytes of dependencies) but something that works entirely in JS (or wasm).

Extensions/Leaderboards/leaderboardstools.ts
4ian: Actually I think the library has some d.ts that we can include: https://github.com/Caligatio/jsSHA/blob/master/dist/sha256.d.ts

GDJS/GDJS/IDE/ExporterHelper.cpp
```diff
@@ -534,6 +534,7 @@ void ExporterHelper::AddLibsInclude(bool pixiRenderers,
   InsertUnique(includesFiles, "logger.js");
   InsertUnique(includesFiles, "gd.js");
   InsertUnique(includesFiles, "libs/rbush.js");
+  InsertUnique(includesFiles, "libs/sha256.js");
```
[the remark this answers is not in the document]
AlexandreSi: Good question. If it is only used (and meant to be used) by the leaderboard extension, I could only put it in the extension folder. What do you think?

Extensions/Leaderboards/leaderboardstools.ts
AlexandreSi: I'll see how to include it

Extensions/Leaderboards/JsExtension.js
```diff
@@ -28,13 +28,13 @@ module.exports = {
...
-        _('Leaderboards'),
+        _('Leaderboards (experimental)'),
         _('Allow your game to send scores to your leaderboards.'),
         'Florian Rival',
         'Open source (MIT License)'
       )
-      .setCategory('Leaderboards');
-    extension
+      .setExtensionHelpPath('/all-features/leaderboards')
+      .setCategory('Leaderboards')
```
arthuro555, line 37: A category is meant to be generic and applicable to multiple extensions. Maybe something like "GDevelop services", "Online services" or "Liluo services" would be better?

PR #3814, opened 2022-04-07
Add warnings when updating an extension with breaking changes

arthuro555:
The warning messages are now clearer, translatable, and only shown if the update type is unknown or breaking (major).

### DIFF
--- a/newIDE/app/src/AssetStore/ExtensionStore/ExtensionInstallDialog.js
+++ b/newIDE/app/src/AssetStore/ExtensionStore/ExtensionInstallDialog.js
@@ -1,5 +1,5 @@
 // @flow
-import { Trans } from '@lingui/macro';
+import { Trans, t } from '@lingui/macro';
 import React from 'react';
 import Dialog from '../../UI/Dialog';
 import FlatButton from '../../UI/FlatButton';
@@ -47,6 +47,7 @@ type Props = {|
   onInstall: () => Promise<void>,
   onEdit?: () => void,
   project: gdProject,
+  i18n: I18nType,
 |};
 
 const ExtensionInstallDialog = ({
@@ -56,6 +57,7 @@ const ExtensionInstallDialog = ({
   onInstall,
   onEdit,
   project,
+  i18n,
 }: Props) => {
   const alreadyInstalled = project.hasEventsFunctionsExtensionNamed(
     extensionShortHeader.name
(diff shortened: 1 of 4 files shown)


newIDE/app/src/AssetStore/ExtensionStore/ExtensionInstallDialog.js
```diff
@@ -94,18 +97,29 @@ const ExtensionInstallDialog = ({
   const onInstallExtension = React.useCallback(
     () => {
       if (canInstallExtension) {
-        if (alreadyInstalled) {
-          const answer = Window.showConfirmDialog(
-            'This extension is already in your project, this will install the latest version. You may have to do some adaptations to make sure your game still works. Do you want to continue?'
-          );
```
D8H, line 100: It might be important to keep this message. It was added to avoid extension creators to override the extension update they are working on by mistake.
Maybe, it could detect an upgrade with the same or lower version and change the color of the update button or something like this.

arthuro555:
Ping

D8H:
I tested manually. I don't see any issue.

* Update
Update

* Major update
MajorUpdate
 
* Unknown
Unknown

D8H:
What do you think about handling the case when a downgrade is done?
It will happen when someone is working on an extension update or is trying one. Changing the button to "Downgrade" could help to avoid mistakes.

Also, maybe the installed version could be shown next to the new version, with something like this:
```JavaScript
              {extensionUpdate ?
              (<Trans>Version{' ' + extensionShortHeader.version + ' '} (installed:{' ' + extensionUpdate.currentVersion})</Trans>) :
              (<Trans>Version{' ' + extensionShortHeader.version}</Trans>)}
```

newIDE/app/src/AssetStore/ExtensionStore/ExtensionInstallDialog.js
```diff
@@ -95,16 +98,24 @@ const ExtensionInstallDialog = ({
       if (canInstallExtension) {
         if (alreadyInstalled) {
           const answer = Window.showConfirmDialog(
-            'This extension is already in your project, this will install the latest version. You may have to do some adaptations to make sure your game still works. Do you want to continue?'
+            (extensionUpdate && extensionUpdate.type) === 'major'
+              ? i18n._(
+                  t`This extension update contains a breaking change. You will have to do some adaptations to make sure your game still works. We advise to back up your game before proceeding. Do you want to continue?`
```
D8H, line 103: > You will have to do some adaptations to make sure your game still works.
It's a nitpick but for me "make sure" mean "verify" and doing changes is more than verifying.
I guess it should be either:
`Make sure your game still works and do required adaptations.`
`Adaptations may be required for your game to keep working.`

newIDE/app/src/AssetStore/ExtensionStore/ExtensionInstallDialog.js
```diff
@@ -95,16 +98,24 @@ const ExtensionInstallDialog = ({
...
+                )
+              : (extensionUpdate && extensionUpdate.type) === 'unknown'
+              ? i18n._(
+                  t`The latest version will be installed, but it isn't indicated whether this update includes breaking changes. You may have to do some adaptations to make sure your game still works. We advise to back up your game before proceeding. Do you want to continue?`
+                )
+              : i18n._(
+                  t`The latest version of the extension is about to be installed. While the extension notifies that there should be no breaking changes, be aware that any modifications you might have made to the extension since installing it will be discarded.  Do you want to continue?`
```
D8H, line 110: > While the extension notifies that there should be no breaking changes, be aware that any modifications you might have made to the extension since installing it will be discarded.

It's hard to see what is the logic behind this.
Does it warns against possible behavior changes that reviews could have missed?
Does it imply that the breaking changes could come from the loss of user-made changes?

What seems important here is that user-made changes might be lost. The message seems too long and will probably not be read.

```suggestion
                  t`Be aware that any modifications you might have made to the extension since installing it will be discarded. Do you want to continue?`
```

newIDE/app/src/AssetStore/ExtensionStore/index.js
```diff
@@ -165,23 +166,28 @@ export const ExtensionStore = ({
...
-            const wasInstalled = await onInstall(selectedExtensionShortHeader);
-            // An errorBox is already displayed by `installExtension`.
-            if (wasInstalled) setSelectedExtensionShortHeader(null);
-          }}
-          onClose={() => setSelectedExtensionShortHeader(null)}
-        />
-      )}
+          </ColumnStackLayout>
+        )}
+      </ResponsiveWindowMeasurer>
```
arthuro555, line 171:
```suggestion
          )}
        />
      </ColumnStackLayout>
```

PR #3817, opened 2022-04-08
Fix: update the game details for Liluo.io when a game is published for the 1st time.

D8H:


newIDE/app/src/Export/GenericExporters/OnlineWebExport/OnlineGameLink.js
```diff
@@ -103,14 +126,42 @@ const OnlineGameLink = ({
       try {
         setIsGameLoading(true);
         const game = await getGame(getAuthorizationHeader, id, gameId);
-        setGame(game);
+
+        let updatedGame = null;
+        try {
+          const isPublishedForTheFirstTime =
+            build && game.publicWebBuildId === build.id;
```
ClementPasteau, line 133: I'm not sure this will work 🤔 

this is the same condition as the variable above `isBuildPublished` which ensures that the current build is the build published.
I'm afraid that with this condition, the update to the game will always be called when you push your build live (and then will be a duplicate of the existing `onGameUpdate` below?

D8H: When a game is already published, this condition can't be true at the openning of the dialog because it's only after the game details are verified by the user that `publicWebBuildId` is updated with the `build.id`.

I manually tested it and it works, I don't say it's clean though.

ClementPasteau: right, I see. I agree with you that it's not the cleanest 🤔  but can't think of a good solution for the moment...

D8H:
This issue is already fixed by another PR.

PR #3818, opened 2022-04-08
Fix: the checkboxes playable with keyboards, gamepads and touch devices where randomly checked when publishing a game for the first time.

D8H:
I guess that, in C++, a non affected boolean variable value is what even there was on the memory.


4ian:
Yep, uninitialised memory! This could have been far worse 😬

PR #3819, opened 2022-04-08
Add a button "Go to public profile" on Liluo.io in profile details

fannieyan:
image
image

### DIFF
--- a/newIDE/app/src/Profile/ProfileDetails.js
+++ b/newIDE/app/src/Profile/ProfileDetails.js
@@ -16,6 +16,7 @@ import RaisedButton from '../UI/RaisedButton';
 import UserAchievements from './Achievement/UserAchievements';
 import { type Badge } from '../Utils/GDevelopServices/Badge';
 import Window from '../Utils/Window';
+import { getUserPublicProfileUrl } from '../Utils/GDevelopServices/User';
 
 type DisplayedProfile = {
   id: string,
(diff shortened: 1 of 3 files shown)


newIDE/app/src/Profile/ProfileDetails.js
```diff
@@ -40,25 +43,49 @@ const ProfileDetails = ({
...
+                    ? i18n._(t`Edit your profile to pick a username!`)
+                    : i18n._(t`No username`))}
+              </Text>
+            </Line>
+            {profile.id && (
+              <RaisedButton
+                label={i18n._(t`Access public profile`)}
+                onClick={() =>
+                  Window.openExternalURL(
+                    profile.username
```
AlexandreSi, line 76: Could this be a function in GDevelopServices/User.js?

newIDE/app/src/Profile/ProfileDetails.js
```diff
@@ -40,25 +44,43 @@ const ProfileDetails = ({
   onEditProfile,
   badges,
 }: Props) => {
+  const isDev = process.env.NODE_ENV === 'development';
   return profile ? (
```
4ian, line 48: You should have something for this in the Window.js file :)

fannieyan: Actually we don't need it here anymore since I moved this check in the Utils function's file. Should I still use Window.isDev() in the utils file ? (If it is the case, I saw the node_env check in ApiConfig, should it be changed too ?)

ClementPasteau: I believe we can change the node_env usage in the ApiConfig too. 
Change it and ensure api calls are going to the right endpoints locally?

newIDE/app/src/Utils/GDevelopServices/User.js
```diff
@@ -56,3 +57,8 @@ export const getUserPublicProfile = (
     .get(`${GDevelopUserApi.baseUrl}/user-public-profile/${id}`)
     .then(response => response.data);
 };
+
+export const getUserPublicProfileUrl = (userId: string, username: ?string) =>
+  username
+    ? `https://liluo.io/${username}${isDev ? '?dev=true' : ''}`
```
ClementPasteau, line 63: to follow the same logic as the other calls, maybe move this logic to ApiConfigs.js inside `GDevelopGamesPlatform`?

ClementPasteau (approved):
lgtm 👍

PR #3820, opened 2022-04-08
New touch event functions to allow to write extensions.

D8H:
## Example
* Project: https://www.dropbox.com/s/t50wfgkz1kgni0n/TouchSmallestExample.zip?dl=1
* Build: https://liluo.io/games/e8f40f0a-6e5f-4eca-97a4-e7732621325b?dev=true

NewTouchEvents
OldTouchEvents

### DIFF
--- a/Core/GDCore/Extensions/Builtin/MouseExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/MouseExtension.cpp
@@ -334,7 +334,7 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsMouseExtension(
       .AddCondition(
           "HasAnyTouchStarted",
           _("A new touch has started"),
-          _("Check if a touch has started. The touch identifiers can be "
+          _("Check if a touch has just started on this frame. The touch identifiers can be "
             "accessed using StartedTouchId() and StartedTouchCount()."),
           _("A new touch has started"),
           _("Multitouch"),
@@ -346,7 +346,7 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsMouseExtension(
       .AddExpression(
           "StartedTouchCount",
           _("Started touch count"),
-          _("The number of touches that have started. The touch identifiers can be "
+          _("The number of touches that have just started on this frame. The touch identifiers can be "
             "accessed using StartedTouchId()."),
           _("Multitouch"),
           "res/conditions/touch.png")
@@ -356,7 +356,7 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsMouseExtension(
       .AddExpression(
           "StartedTouchId",
           _("Started touch identifier"),
-          _("The identifier of the touch that has started. The touch number of touches can be "
+          _("The identifier of the touch that has just started on this frame. The touch number of touches can be "
             "accessed using StartedTouchCount()."),
           _("Multitouch"),
           "res/conditions/touch.png")
--- a/Core/GDCore/Extensions/Builtin/MouseExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/MouseExtension.cpp
@@ -311,7 +311,8 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsMouseExtension(
           _("Multitouch"),
           "res/conditions/touch24.png",
           "res/conditions/touch.png")
-      .AddCodeOnlyParameter("currentScene", "");
+      .AddCodeOnlyParameter("currentScene", "")
+      .SetHidden();
 
   extension
       .AddCondition(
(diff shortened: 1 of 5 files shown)


GDJS/Runtime/inputmanager.ts
```diff
@@ -344,12 +363,22 @@ namespace gdjs {
...
 
+    /**
+     * @deprecated
+     */
     popEndedTouch(): integer | undefined {
-      return this._endedTouches.shift();
+      const touchId = this._endedTouches[this._lastEndedTouchIndex];
+      this._lastEndedTouchIndex++;
+      return touchId;
     }
```
arthuro555, line 382: It concerns me that now, once an ID has been pushed into `_endedTouches` or `_startedTouches`, it will never get removed, so spam touching the screen will fill up memory slowly but steadily. I'd suggest clearing both arrays at the end of the frame since we usually only need to process the start or end of a touch once anyways and to keep the memory footprint low.

GDJS/Runtime/inputmanager.ts
```diff
@@ -309,16 +326,18 @@ namespace gdjs {
...
 
     onTouchStart(identifier: integer, x: float, y: float): void {
-      this._startedTouches.push(identifier);
-      this._touches.put(identifier, { x: x, y: y, justEnded: false });
+      // Add 1 to the identifier to avoid identifiers taking
+      // the GDevelop default variable value which is 0.
+      this._startedTouches.push(identifier + 1);
+      this._touches.put(identifier + 1, { x: x, y: y, justEnded: false });
```
arthuro555, line 332:
```suggestion
      const internalID = this.getInternalIdentifier(identifier);
      this._startedTouches.push(internalID);
      this._touches.put(internalID, { x: x, y: y, justEnded: false });
```

2/5

GDJS/tests/tests/inputmanager.js
```diff
@@ -3,7 +3,29 @@
...
+    objects: [],
+    instances: [],
+  });
+  runtimeScene.getGame.getElapsedTime = function () {
+    return 1000 / 60;
+  };
+  const inputManager = runtimeScene
+  .getGame()
+  .getInputManager();
+  const inputTool = gdjs.evtTools.input;
```
arthuro555, line 28: nit: `inputTools`?

GDJS/Runtime/inputmanager.ts
D8H: It seems it was already the case, because nothing was forcing users to pop every events.

```JavaScript
    onFrameEnded(): void {
      //Only clear the ended touches at the end of the frame.
      for (const id in this._touches.items) {
        if (this._touches.items.hasOwnProperty(id)) {
          const touch = this._touches.items[id];
          if (touch.justEnded) {
            this._touches.remove(id);
          }
        }
      }
      this._startedTouches.length = 0;
      this._endedTouches.length = 0;
      this._releasedKeys.clear();
      this._releasedMouseButtons.length = 0;
      this._mouseWheelDelta = 0;
    }
```

arthuro555: Ah, I missed that, all good then 👍

GDJS/Runtime/inputmanager.ts
```diff
@@ -309,16 +326,18 @@ namespace gdjs {
     }
```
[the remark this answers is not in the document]
D8H: I hesitated to do it. I though is was too verbose and could actually make the code less readable as all the `+ 1` all are in the same location and it is hard to find a function name that is understandable without documentation, but I'm ok to make the change. It would be an "external" identifier as it's the code used by events.

GDJS/tests/tests/inputmanager.js
```diff
@@ -3,7 +3,29 @@
...
+  });
+  const runtimeScene = new gdjs.RuntimeScene(runtimeGame);
+  runtimeScene.loadFromScene({
+    layers: [{ name: '', visibility: true, effects: [] }],
+    variables: [],
+    behaviorsSharedData: [],
+    objects: [],
+    instances: [],
+  });
+  runtimeScene.getGame.getElapsedTime = function () {
```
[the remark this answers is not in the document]
D8H: It was copy-pasted from a movement behavior. It's, indeed, not necessary here.

GDJS/Runtime/inputmanager.ts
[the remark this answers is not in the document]
arthuro555: > A local variable would probably be clearer.

Unless the variable name is very verbose, understanding it would require the comment to be above each local variable, which is not awesome (especially for the one liner in `onTouchMove`). I think a function is cleaner because it allows the explanation comment to only be used once, and when editing the other functions you do not need to know why an identifier needs 1 added to it. Ensuring an ID is not 0 is another concern than adding, modifying and deleting a touch, and should therefore be separated in my opinion

> It would be an "external" identifier as it's the identifier used by events.

Fair, I'd suggest naming it `publicID` then.

> I though is was too verbose and could actually make the code less readable

In that case we should probably ask for a 3rd opinion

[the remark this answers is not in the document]
4ian: > It would be an "external" identifier as it's the identifier used by events.

That's fair! It's indeed a public/external identifier.

> > I though is was too verbose and could actually make the code less readable
> In that case we should probably ask for a 3rd opinion

My decision would be to go with the more verbose alternative suggested by Arthuro555: a function to "do the +1".
Sure it's not a lot, but:
- it's not performance sensitive code (we don't care about an extra function call).
- it's more verbose, but it also force you to think "am I right to translate this raw identifier into a public identifier?" and "because it's verbose, I'll try to use it only once" 
  - We could almost think of two separate types (but no need to do this, not sure we can express this in TypeScript anyway).
- Finally, it's a good place to put a comment.

GDJS/Runtime/inputmanager.ts
4ian: (resolving)

Core/GDCore/Extensions/Builtin/MouseExtension.cpp
```diff
@@ -326,8 +327,54 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsMouseExtension(
...
+          "res/conditions/touch.png")
+      .AddCodeOnlyParameter("currentScene", "")
+      .AddParameter("expression", _("Touch index"));
+
+  extension
+      .AddCondition(
+          "HasTouchEnded",
+          _("A touch has ended"),
+          _("Check if a touch has ended."),
+          _("The touch has ended with identifier: _PARAM1_"),
```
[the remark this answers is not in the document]
4ian: Agree!

Core/GDCore/Extensions/Builtin/MouseExtension.cpp
```diff
@@ -326,8 +327,54 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsMouseExtension(
...
+          _("Multitouch"),
+          "res/conditions/touch24.png",
+          "res/conditions/touch.png")
       .AddCodeOnlyParameter("currentScene", "");
 
+  extension
+      .AddExpression(
+          "StartedTouchCount",
+          _("Started touch count"),
+          _("The number of touches that have started. The touch identifiers can be "
```
4ian, line 349:
```suggestion
          _("The number of touches that have just started on this frame. The touch identifiers can be "
```

GDJS/Runtime/inputmanager.ts
```diff
@@ -295,6 +304,14 @@ namespace gdjs {
       return this._touches.get(identifier).y;
     }
 
+    /**
+     * @param identifier the touch identifier
+     * @returns true if the touch has just ended.
+     */
+    hasTouchEnded(identifier: integer): boolean {
```
4ian, line 311: nit/idea: we could rename all `identifier` parameters to `rawIdentifier` (coming from the browser API) or `publicIdentifier` (the one stored by GDevelop)

4ian (approved):
Love this!

4ian:
@D8H Please prepare a note for the changelog and let's update at least the examples and list what need updates in the wiki.

Issue #3666, opened 2022-02-18
Camera will not reposition if in the same event as Resolution Change and/or Zoom


Silver-Streak:
## Describe the bug
Working with a user on the discord around a potential scaling method to allow for cleaner low-resolution rotation/camera/etc, I discovered that you cannot reposition a camera in the same event that the zoom and resolution are changed (I'm unable to narrow down if it's the combination of the two, or just zoom, or what).

I've confirmed this happen during mid scene, or if it's at the beginning of the scene. It's almost as if the camera will not move correctly unless it is done the frame *after* the zoom/resolution changes.

## To Reproduce
1. Create a new project
2. Set the resolution to non-screen native (160x90, 320x180, etc)
3. Add something to the scene.
4. Add an event with "at the beginning of the scene" condition.
5. Add an action to change the resolution by a factor (Let's say a 160x90 game, by 12x factor, to 1920x1080)
6. Add an action to change the zoom on the layer to 12x
7. Add actions to change the camera center back to the default starting x/y (for 160x90, that'd be 80,45)
8. Start the scene. The camera will be at 960x540 and remain there.

If you add actions to restart the scene, since the zoom/resolution change already occurred, the position will update correctly.
OR if you add an event with "At the beginning of the scene" inverted and with Trigger once, with the actions for the camera move, the position will update correctly.

I've also tested by just having resolution/zoom updates associated with a key, but if the position actions are in the same event, they do not occur unless you press the key twice.

You can see an example here: 
Edit: Helps if I attach a zip with content in it, re-uloaded example project.
[ScalerTest.zip](https://github.com/4ian/GDevelop/files/8095330/ScalerTest.zip)

I've got scenes set up for both scenarios mentioned above.

## Other details
* Windows 10 Pro 64 Bit
* GDevelop 5.0.127

PR #3821, opened 2022-04-09
Fix the cameras position to keep them centered on the object they follow when the game resolution is changed

D8H:
* https://github.com/4ian/GDevelop/issues/3666

# Changes
- Also fix the camera position for pixel-art games when a zoom and a game resolution change is used to have big pixels

# Examples
* Project: https://www.dropbox.com/s/frf2sjp0718hm9e/ResolutionCameraTest.zip?dl=1

## Example 1 - UI or scrolling?
The scene is 800x600 fill with orange.

image

### Camera event disabled
* The camera is at the default position. It's probably an UI layer.
* The camera upper left corner stays at the same position.

image

### Camera event enabled
* The camera is not at the default position. It's probably a scrolling layer. When a camera following extension is used it will be the case.
* The camera center stays at the same position.

image

## Example 2 - Pixel UI

The camera center doesn't need to be moved by the engine.

image

image

image

### DIFF
--- a/GDJS/Runtime/layer.ts
+++ b/GDJS/Runtime/layer.ts
@@ -43,17 +43,22 @@ namespace gdjs {
       oldGameResolutionOriginX: float,
       oldGameResolutionOriginY: float
     ): void {
-      // Adapt position of the camera center as:
-      // * Most cameras following a player/object on the scene will be updating this
-      // in events anyway.
+      // Adapt position of the camera center only if the camera has never moved as:
+      // * When the camera follows a player/object, it will rarely be at the default position.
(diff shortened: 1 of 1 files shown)


4ian (commented):
I'm not against this, but how do you think we could document this to make it clear? Somewhere on the wiki?

D8H:
> I'm not against this, but how do you think we could document this to make it clear? Somewhere on the wiki?

We could, but I see it as a bug fix. I can't see any case where the camera has been moved and should anchor to the top left. Well, it's already isn't in this case actually, so why try?

4ian (approved):

PR #3823, opened 2022-04-12
Show syntax errors in the event sheet.

D8H:
## Problem
Parsing an expression can be time consuming, it should not be done every time the event sheet is scrolled and new expressions appear on screen. A cache of AST must be used.
The AST was dependent from the context where the expression was used (like the object and function declarations). The cache would be invalidated if there is a change in the context. This is complicated thus bug prone.

## Technical solution
The type checking was extracted from `ExpressionParser2` to `ExpressionValidator`. This way `ExpressionParser2` build an AST that is independent from context. A given string will always give the exact same AST.
The type checking takes a lot less time than an AST parsing (about 1 to 10). It can be done every time.

## Old comment
It could be interesting to check object existence in expressions one day.

image

The behavior renamer and the code generator still use the parser directly because `ParameterMetadataTools::IterateOverParameters` make the change a bit hard. It might be interesting to do it if it can speed up the preview.


Core/GDCore/Events/Expression.h
```diff
@@ -24,32 +29,54 @@ class GD_CORE_API Expression {
...
+  /**
+   * @brief Get the expression node.
+   *
+   * @param type the expression type
+   * @param parser the parser to use if this expression never has been parsed
+   * before.
+   * @return std::unique_ptr<gd::ExpressionNode>
+   */
+  std::unique_ptr<gd::ExpressionNode> GetRootNode(
+      const gd::String& type, gd::ExpressionParser2& parser) const;
```
D8H, line 68: The type should probably be passed by the constructor, but it seems like a huge refactor.

newIDE/app/src/EventsSheet/EventsTree/Instruction.js
```diff
@@ -146,14 +166,44 @@ const Instruction = (props: Props) => {
...
-          const parameterType = parameterMetadata.getType();
+          const parameterSubType = parameterMetadata.getType();
+          const parameterType = gd.ParameterMetadata.isExpression(
+            'number',
+            parameterSubType
+          )
+            ? 'number'
+            : gd.ParameterMetadata.isExpression('string', parameterSubType)
+            ? 'string'
+            : parameterSubType;
```
D8H, line 177: The parser should probably understand the subtypes of "string" and "number" when they are passed in `ParseExpression()` as it already does these check in some internal steps.

newIDE/app/src/EventsSheet/EventsTree/Instruction.js
```diff
@@ -146,14 +166,50 @@ const Instruction = (props: Props) => {
...
+              globalObjectsContainer,
+              objectsContainer
+            );
+            const expressionNode = instruction
+              .getParameter(parameterIndex)
+              .getRootNode(parameterType, parser)
+              .get();
+            const expressionValidator = new gd.ExpressionValidator();
+            expressionNode.visit(expressionValidator);
+            expressionIsValid = expressionValidator.getErrors().size() === 0;
```
D8H, line 195: Maybe this should be cached too?

D8H:
How can an external event sheet be attached to another scene? Is it only possible when the scene was removed?

newIDE/app/src/EventsSheet/EventsTree/Instruction.js
```diff
@@ -146,14 +166,50 @@ const Instruction = (props: Props) => {
           }
 
           const parameterMetadata = metadata.getParameter(parameterIndex);
-          const parameterType = parameterMetadata.getType();
+          const parameterSubType = parameterMetadata.getType();
```
4ian, line 169: We should probably at least get this logic into a `gd.ParameterMetadata.isParsableAsExpression` or something like this, or `gd.ParameteraMetadata.getParentType` or something similar.

newIDE/app/src/EventsSheet/EventsTree/Instruction.js
4ian: Probably yes!

newIDE/app/src/EventsSheet/EventsTree/Instruction.js
4ian: I'm not sure what the cost is exactly, so hard to say. Ideally we want to avoid caching, but it also means that for now we're running validation at every render, which can be time consuming.

Add some performance.now() to see what is the time to run this validation on average? And on a very large expression (1000 characters?)

Core/GDCore/Events/Expression.cpp
```diff
@@ -0,0 +1,41 @@
...
+Expression& Expression::operator=(const Expression& expression) {
+  plainString = expression.plainString;
+  node = nullptr;
+  return *this;
+};
+
+Expression::~Expression(){};
+
+std::unique_ptr<ExpressionNode> Expression::GetRootNode(
+    const gd::String& type, gd::ExpressionParser2& parser) const {
```
4ian, line 34: Is there a risk that the parser being passed changes compared to the "memoized" node? We should at least log this, to see when and why it happens, and what are the risks.

4ian: (and we could if necessary memoize according to the parser address in memory or the content of the parser, like the objects container, but this can be risky too)

4ian (commented):
We would need to continue:
- Timing of the validation during the render.
- When/why the parser could change when calling getRootNode. Let's at least log this and try this.
  - In particular with external events (even if links should copy them).
  - In particular if you change the scene associated to the external events (at this moment, they don't change in memory, but the objects containers are not the same anymore!).

D8H:
It remains 7 red tests on 4 independent issues. I have a good idea of how to fix them.

D8H:
All tests should be green now. I'll do a pass on the todo and add tests if needed.

D8H:
There are still some todo to be solved and tests to be added, but I guess it's ready enough for a review, though it can wait till Monday.

Core/GDCore/IDE/Events/ExpressionValidator.h
```diff
@@ -56,52 +71,238 @@ class GD_CORE_API ExpressionValidator : public ExpressionParser2NodeWorker {
...
       errors.push_back(node.diagnostic.get());
     }
   }
 
+  void RaiseError(const gd::String &type, 
+      const gd::String &message, const ExpressionParserLocation &location) {
+    auto diagnostic = std::move(gd::make_unique<ExpressionParserError>(
+        type, message, location));
+    errors.push_back(diagnostic.get());
+    supplementalErrors.push_back(std::move(diagnostic));
```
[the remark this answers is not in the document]
D8H: It's to keep the errors alive, but it probably doesn't work as leaving the helper method could destroy the validator.

Core/GDCore/IDE/Events/ExpressionTypeFinder.h
```diff
@@ -0,0 +1,189 @@
...
+    auto leftSideType = gd::ExpressionLeftSideTypeFinder::GetType(
+        platform, 
+        globalObjectsContainer,
+        objectsContainer,
+        node);
+    if (leftSideType == "number" || leftSideType == "string") {
+      type = leftSideType;
+    }
+    else {
+      type = "number|string";
```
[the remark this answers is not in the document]
D8H: The `type` returned by this worker is a mix of:
- an expected type looking up like a parameter declaration
- an actual type looking down, but only looking at the most left branch (using ExpressionLeftSideTypeFinder)

This logic was built with the constraint of following a parser that can't know the right side. Now that it is extracted, it could be enhanced but we probably want to wait a stable version before doing it.

About this case specifically, it must be understood as follow "In a general manner, a bracket variable assessor expression is either a number or a string. According to the actual used expression, which one is it?". The type is then used to give more precise errors or autocompletions focusing on either number or string.

Core/GDCore/Extensions/Metadata/MetadataProvider.cpp
```diff
@@ -390,6 +392,88 @@ MetadataProvider::GetBehaviorAnyExpressionMetadata(const gd::Platform& platform,
...
+              functionCall.objectName, functionCall.behaviorName);
+      const gd::ExpressionMetadata &metadata = MetadataProvider::GetFunctionCallMetadata(
+          platform, globalObjectsContainer, objectsContainer, functionCall);
+
+      if (IsBadExpressionMetadata(metadata)) {
+        return nullptr;
+      }
+
+      // TODO use a badMetadata instead of a nullptr?
+      const gd::ParameterMetadata* parameterMetadata = nullptr;
```
D8H, line 459: Should it be done?

Core/tests/ExpressionParser2.cpp
```diff
@@ -194,64 +226,83 @@ TEST_CASE("ExpressionParser2", "[common][events]") {
...
       REQUIRE(validator.GetErrors()[0]->GetMessage() ==
+              "You must add the operator + between texts or expressions. For "
+              "example: \"Your name: \" + VariableString(PlayerName).");
+      REQUIRE(validator.GetErrors()[2]->GetMessage() ==
+              "You must enter a text (between quotes) or a valid expression call.");
+
+      REQUIRE(validator.GetErrors()[1]->GetMessage() ==
               "The expression has extra character at the end that should be "
               "removed (or completed if your expression is not finished).");
     }
```
D8H, line 267: These 2 extra errors happen because the validation is done on the tree without checking that the empty node at the right of the "," operator is actually an unfinished expression.
These errors could be pertinent but 3 errors might be a bit too much. Should this case be handled to remove these 2 errors?

Core/tests/ExpressionParser2.cpp
```diff
@@ -573,22 +673,36 @@ TEST_CASE("ExpressionParser2", "[common][events]") {
...
+      REQUIRE(validator.GetErrors().size() == 3);
+      REQUIRE(validator.GetErrors()[1]->GetMessage() ==
               "No operator found. Did you forget to enter an operator (like +, "
               "-, * or /) between numbers or expressions?");
-      REQUIRE(validator.GetErrors()[0]->GetStartPosition() == 4);
+      REQUIRE(validator.GetErrors()[1]->GetStartPosition() == 4);
+      // TODO Should error be removed?
+      REQUIRE(validator.GetErrors()[2]->GetMessage() ==
+              "You must enter a number.");
+    }
```
D8H, line 689: Here additional errors happen because "%" is not a reserved character and it's read as an identifier. This PR adds a second error that is not really pertinent, but I don't see an easy solution to remove it. Should it be dug?

The test that follows use a reserved character and only the pertinent message is shown. I guess, that operators commonly used in other language can't be added to the list of reserved character because of side effects.

Core/GDCore/Extensions/Metadata/MetadataProvider.cpp
4ian: If not necessary for this PR, no.

Core/tests/ExpressionParser2.cpp
4ian: >  I guess, that operators commonly used in other language can't be added to the list of reserved character because of side effects.

Maybe we could handle "well known" characters like % to ensure we deliver a proper error message (like, "the modulo operator is not supported, using `mod` instead"). Not a big priority though, it's rather a "nice to have".

D8H:
I did a quick stat of the time to validate expressions. The plateau is the huge expressions. There is a lot more of them than in the previous graph to better see the difference between tiny and huge expressions.
There are 5 values near 30 ms and 32 values near 20 ms, I guess it's GC that might not be related to the expression validation.

image

Core/GDCore/IDE/Events/ExpressionValidator.h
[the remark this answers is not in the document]
D8H: Using static constants for error messages will probably make the code hard to read, but I don't see any other solution.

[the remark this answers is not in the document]
4ian: Could you store a copy of the diagnostics?

In any case I would really really avoid this pattern of storing raw pointers in a vector and the unique_ptr in a supplementalErrors.
This makes the intention not clear (one could drop the supplementalError, because they think it's "just extra stuff I don't care about"), invalidating all pointers stored in the validator.
We probably need another approach.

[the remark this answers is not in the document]
4ian: Could you store a copy of the diagnostics?

In any case I would really really avoid this pattern of storing raw pointers in a vector and the unique_ptr in a supplementalErrors.
This makes the intention not clear (one could drop the supplementalError, because they think it's "just extra stuff I don't care about"), invalidating all pointers stored in the validator.
We probably need another approach.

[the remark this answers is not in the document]
D8H: > Could you store a copy of the diagnostics?

Do you mean to copy errors from the AST to return a vector of unique pointers?

> one could drop the supplementalError, because they think it's "just extra stuff I don't care about"

By "one", do you mean maintainers of this class? because `supplementalErrors`is a protected member (and should probably be private).

4ian:
As discussed:
- `gd.ParameterMetadata.getExpressionValueType` returns a reference to a temporary, which result in a deleted/garbage string being returned. We could consider using en enum here (for the return value, maybe not yet for the parameter).
- Because `ParseExpression` is called in `GenerateExpressionCode`, we actually always reparse everything, even if we switch gd::Expression to share (with a shared_ptr) the `gd::ExpressionNode` when it's copied (which should be safe because expressions and nodes are immutable).
  - This implies refactoring `IterateOverParametersWithIndex` to pass a gd::Expression, not the plain string equivalent.

D8H:
The new way to do the type checking is heavier on metadata accesses. I did a small optimization to lesser its impact.

On a big game export, it saves 7% (from 2.90 seconds to 2.88 seconds). It also save about half the time when doing the type checking (to show error in the events sheet for instance).

## Full export

### Before
Export-Before

### After
Export-After

## Type checking

### Before
Validator-Before

### After
Validator-After

## Type finding

### Before
TypeChecking-Before

### After
TypeChecking-After

D8H:
Without parsing, it saves 10% (from 2.88 seconds to 2.60 seconds. It might worth the risk.

Export-NoParsing

GDJS/GDJS/IDE/ExporterHelper.cpp
```diff
@@ -134,6 +114,28 @@ bool ExporterHelper::ExportProjectForPixiPreview(
...
+  ExportResources(fs, exportedProject, options.exportPath);
+
+  previousTime = LogTimeSpent("Resource export", previousTime);
+
+  // Compatibility with GD <= 5.0-beta56
+  // Stay compatible with text objects declaring their font as just a filename
+  // without a font resource - by manually adding these resources.
+  AddDeprecatedFontFilesToFontResources(
+      fs, exportedProject.GetResourcesManager(), options.exportPath);
+  // end of compatibility code
```
D8H, line 137: Do you think it's ok to move this after the code generation?
I guess the generated code is using resource names so changing the resource pathes should probably not impact the code generation.

Core/GDCore/Project/Layout.cpp
```diff
@@ -505,18 +505,16 @@ gd::String GD_CORE_API GetTypeOfBehavior(const gd::ObjectsContainer& project,
                                          gd::String name,
                                          bool searchInGroups) {
   for (std::size_t i = 0; i < layout.GetObjectsCount(); ++i) {
-    vector<gd::String> behaviors = layout.GetObject(i).GetAllBehaviorNames();
-    for (std::size_t j = 0; j < behaviors.size(); ++j) {
-      if (layout.GetObject(i).GetBehavior(behaviors[j]).GetName() == name)
-        return layout.GetObject(i).GetBehavior(behaviors[j]).GetTypeName();
+    const auto &object = layout.GetObject(i);
+    if (object.HasBehaviorNamed(name)) {
+      return object.GetBehavior(name).GetTypeName();
```
D8H, line 510: `GetAllBehaviorNames()` allocate a vector. As the new type checking access the metadata more often, it made the code generation slower. I replaced it with `HasBehaviorNamed()` but it does the same `find` call as `GetBehavior`. I preferred to leave things unoptimized for code clarity (it would only have saved 1ms on a 2.8 second long export).

GDJS/GDJS/IDE/ExporterHelper.cpp
4ian: Maybe... though the code generation could in theory raise an error because somehow you're using an action/condition with a "fontResource" and this resource does not exist.
I think the risk is small, but exists... What's the reason behind you wanting to move this?

Side note:
- At some point I experimented about exporting the resources in an asynchronous matter: we issue commands to copy files, which are launched on the Node.js side, and at the end of the export (or whenever the content of a file is needed, but this should not be required I think) we wait on all copies to be done.
  The idea is to parallelise the I/O (copying exports) and the events generation (CPU work, then I/O at the end).
  
  (This is a different problem though, so if we do this it should be done in another PR :) Don't make your work here more complex with this optimisation for now)

D8H: It avoid to parse the expressions when exporting a second time which save around 10% of the exporting time. If it's risky, it can wait a different release.

Core/GDCore/Project/Layout.cpp
4ian: My god this was so un-optimised 😅
HasBehaviorNamed + GetBehavior is fine :) We at least avoid the unnecessary copy.

I know there are other stuff like the extension/instruction metadata that are doing a lot of useless loops too. We could take a look at converting these to hashtable/unordered_map or something that helps getting constant time accesses :)

D8H: Yes, I did some other similar optimizations. I don't know if I've changed them all but at least the ones I found with the monitoring tool.

4ian (approved):
Looks good to go - and let's hope to have all the optimisations (avoid reparsing everything at each preview/export) in another PR later 😄

PR #3829, opened 2022-04-14
Polish particle actions

Elairyx7301:
# Changes
- Replaced `parameter x` with corresponding info like `start color`
- Added dots to end of descriptions
- Other polish

# Questions
I tried my best to figure them out but I have no idea what the actions, condition and/or expressions for the following do so I had to leave it like it is. I have found "*RendererParam*" in the web app expression list but I still don't know what it's doing.
 
- **EmitterAngleA, EmitterAngleB:**
https://github.com/4ian/GDevelop/blob/master/Extensions/ParticleSystem/ExtensionSubDeclaration3.cpp#L228

- **RendererParam1, RendererParam2:**
https://github.com/4ian/GDevelop/blob/master/Extensions/ParticleSystem/ExtensionSubDeclaration3.cpp#L180

- **ParticleAngle1, ParticleAngle2:**
https://github.com/4ian/GDevelop/blob/master/Extensions/ParticleSystem/ExtensionSubDeclaration3.cpp#L342
https://github.com/4ian/GDevelop/blob/master/Extensions/ParticleSystem/ExtensionSubDeclaration2.cpp#L210

**Note:**
"*ParticleAngle2*" seems to be the "*Spray cone angle*". I have tried close to everything but I can't notice a difference when changing "*ParticleAngle1*". What does it do?

## Reference image
image

 
 
---
Only relevant pushes are:
2840d420095b94026765ceaff8b569f6f5da0db8 [remove spaces inside parentheses](https://github.com/4ian/GDevelop/commit/2840d420095b94026765ceaff8b569f6f5da0db8)
89849dde07c50e26f2c15993d1aa94d2a3e998f4 [clarify/unify names and descriptions](https://github.com/4ian/GDevelop/commit/89849dde07c50e26f2c15993d1aa94d2a3e998f4)[](https://github.com/Elairyx7301)
9b02f6bf87914b13028b151711829bb99cdf2d12 [clarify/unify expressions](https://github.com/4ian/GDevelop/commit/9b02f6bf87914b13028b151711829bb99cdf2d12)


4ian:
Thanks! I did an additional commit to remove a file that was change previously (don't be afraid of doing additional commits, we'll "squash" them into one at the end).

Extensions/ParticleSystem/ExtensionSubDeclaration2.cpp
```diff
@@ -16,30 +16,30 @@ This project is released under the MIT License.
...
       .AddParameter("object", _("Object"), "ParticleEmitter")
-      .AddParameter("color", _("Final color"));
+      .AddParameter("color", _("End color"));
 
   obj.AddAction(
          "ParticleRed1",
-         _("Red color, parameter 1"),
-         _("Modify parameter 1 of the red color."),
-         _("the parameter 1 of red color"),
+         _("Red start color"),
```
4ian, line 40: Any chance you could change all references to colors to be "Start color red component"?

Because red/green/blue is a "component" of the overall color :)

4ian (approved):
I'm fine with most of these changes. If you can update everywhere the colors to be "[End/start] color [red/green/blue] component", that would be awesome.

Elairyx7301:
Sure thing! Will do that tomorrow :)

Elairyx7301:
@4ian Requested changes are done.
Do you know about the "*EmitterAngle*", "*RendererParam*" and "*ParticleAngle*" act./cond./expr. and what they do? I would change them as well but I can't find what they do or what parameter of the "object settings" they change/check.

4ian:
I think EmitterAngle is the same as the object angle. This would need a double check in the code, but this can be removed if that's the case.
RendererParam I'm not sure (was it the texture?)
ParticleAngle I think is something that was used in the past but is not anymore (would need a check in the code too).

I think this is good enough to merge once tests are passing :)

4ian:
Merged! Thanks for taking the time to do this 🎉 And let's see if we can improve the rest by digging a bit into the code.

tristanbob:
> I think EmitterAngle is the same as the object angle. This would need a double check in the code, but this can be removed if that's the case. 

I agree.  To set the direction that the particle emitter will point, we can simply use the object.angle() conditions/actions/expressions that users are already familiar with.  This also matches what happens when the user rotates the particle emitter object in the scene editor.  So I agree that it is not needed and can be removed.

PR #3830, opened 2022-04-14
Add image resource field

AlexandreSi:
Use it for particle emitter "Change image" action

### DIFF
--- a/Extensions/ParticleSystem/ExtensionSubDeclaration3.cpp
+++ b/Extensions/ParticleSystem/ExtensionSubDeclaration3.cpp
@@ -136,12 +136,13 @@ void ExtensionSubDeclaration3(gd::ObjectMetadata& obj) {
 
   obj.AddAction("Texture",
                 _("Image"),
-                _("Change the image of particles, valid name is the resource name from the resource tab."),
+                _("Change the image of particles ( if displayed )."),
                 _("Change the image of particles of _PARAM0_ to _PARAM1_"),
                 _("Common"),
                 "CppPlatform/Extensions/particleSystemicon24.png",
                 "CppPlatform/Extensions/particleSystemicon16.png")
       .AddParameter("object", _("Object"), "ParticleEmitter")
+      .SetParameterLongDescription(_("Valid name is the resource name used for an asset."))
       .AddParameter("string", _("New image"));
 
   obj.AddCondition(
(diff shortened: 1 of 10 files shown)


Extensions/ParticleSystem/ExtensionSubDeclaration3.cpp
```diff
@@ -135,14 +135,25 @@ void ExtensionSubDeclaration3(gd::ObjectMetadata& obj) {
...
+                _("Change the image of particles (if displayed)."),
+                _("Change the image of particles of _PARAM0_ to _PARAM1_"),
+                _("Advanced"),
+                "CppPlatform/Extensions/particleSystemicon24.png",
+                "CppPlatform/Extensions/particleSystemicon16.png")
+      .AddParameter("object", _("Object"), "ParticleEmitter")
+      .AddParameter("string", _("Image to use"))
+      .SetParameterLongDescription("Indicate the name of the resource");
+
+  obj.AddAction("TextureAmongResources",
```
D8H, line 148: Actions must now begin with a verb.

If one day a field can do autocompletion as well as expression, one action will need to be deprecated, is it ok?

newIDE/app/src/EventsSheet/ParameterFields/ImageResourceField.js
```diff
@@ -0,0 +1,58 @@
...
+
+import * as React from 'react';
+import ResourceSelector from '../../ResourcesList/ResourceSelector';
+import ResourcesLoader from '../../ResourcesLoader';
+import {
+  type ParameterFieldProps,
+  type ParameterFieldInterface,
+} from './ParameterFieldCommons';
+
+const ImageResourceField = React.forwardRef<
```
4ian, line 12: Any chance you can add it to a story? :)

Extensions/ParticleSystem/ExtensionSubDeclaration3.cpp
4ian, line 148: SetTextureFromResource

(I sadly in the past not used verb at the beginning of actions. But you should have a verb now for actions, Set being the default one)

4ian (approved):

newIDE/app/src/EventsSheet/ParameterFields/ImageResourceField.js
AlexandreSi: Yes, will do!

PR #3832, opened 2022-04-15
Hide advanced parameters and properties from beginners.

D8H:
I made a quick way for advance users to get ride of the beginner mode, but it's probably not be a good practice to replace a button when it's clicked.

InstructionBasic

InstructionAdvance

BehaviorBasic

BehaviorAdvance

### DIFF
--- a/newIDE/app/src/PropertiesEditor/index.js
+++ b/newIDE/app/src/PropertiesEditor/index.js
@@ -26,12 +26,14 @@ import {
   ColumnStackLayout,
 } from '../UI/Layout';
 import RaisedButton from '../UI/RaisedButton';
+import FlatButton from '../UI/FlatButton';
 import UnsavedChangesContext, {
   type UnsavedChanges,
 } from '../MainFrame/UnsavedChangesContext';
 import { Line, Spacer } from '../UI/Grid';
 import Text from '../UI/Text';
 import useForceUpdate from '../Utils/UseForceUpdate';
+import PreferencesContext from '../MainFrame/Preferences/PreferencesContext';
 
 // An "instance" here is the objects for which properties are shown
 export type Instance = Object; // This could be improved using generics.
@@ -45,6 +47,7 @@ export type ValueFieldCommonProperties = {|
   getExtraDescription?: Instance => string,
   disabled?: boolean,
   onEditButtonClick?: Instance => void,
+  usageComplexity?: number,
 |};
 
 // "Primitive" value fields are "simple" fields.
@@ -140,6 +143,7 @@ type Props = {|
   resourceSources?: ?Array<ResourceSource>,
   onChooseResource?: ?ChooseResourceFunction,
   resourceExternalEditors?: ?Array<ResourceExternalEditor>,
+  shouldShowAdvancedParameter?: boolean,
 |};
 
 const styles = {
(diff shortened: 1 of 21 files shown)


newIDE/app/src/PropertiesEditor/index.js
```diff
@@ -210,6 +214,22 @@ const getFieldLabel = (instances: Instances, field: ValueField): any => {
...
+      (field.usageComplexity && field.usageComplexity > 5) ||
+      (field.children && containsAdvancedFields(field.children))
+  );
+};
+
+const containsBasicFields = (schema: Schema): boolean => {
+  return schema.some(
+    field =>
+      !(field.usageComplexity && field.usageComplexity > 5) ||
+      (field.children && containsBasicFields(field.children))
```
D8H, line 229: Flow says that there is no `field.children` in `Shema`. I don't understand why.

4ian:
I order if we should use an Accordion (https://github.com/4ian/GDevelop/blob/a217565919a7a9e990db02c82a6af9c635ea689e/newIDE/app/src/UI/Accordion.js) for this? We use it already for things like behavior and this could make sense to put advanced parameters inside. This would make them out of order, but generally advanced parameters are at the end anyway :)

And this is more adapted to something that toggles the visibility of its children!

D8H:
> I order if we should use an Accordion (https://github.com/4ian/GDevelop/blob/a217565919a7a9e990db02c82a6af9c635ea689e/newIDE/app/src/UI/Accordion.js) for this? We use it already for things like behavior and this could make sense to put advanced parameters inside. This would make them out of order, but generally advanced parameters are at the end anyway :)
> 
> And this is more adapted to something that toggles the visibility of its children!

It will probably work well for parameters but I'm not sure how to do it for properties as there are categories:
- add one accordion per category might be to heavy and a bit annoying
- add one at the end will remove the context that might be needed to understand the properties.
- putting the categories a 2nd time inside the accordion is probably not something to do either.

It should probably work the same for both properties and parameters.

AlexandreSi:
This is interesting to mark complexity on a scale, for sure a good decision for the future!
I really feel like that the Show/Hide advanced parameters should be [behavior/instruction/object-type]-wise. A user that learns how to use advanced properties for platformer behavior doesn't learn how to use advanced properties for the physics behavior.

D8H:
> This is interesting to mark complexity on a scale, for sure a good decision for the future! I really feel like that the Show/Hide advanced parameters should be [behavior/instruction/object-type]-wise. A user that learns how to use advanced properties for platformer behavior doesn't learn how to use advanced properties for the physics behavior.

Personally, I think it's a matter of mental state not knowledge:
- I've never managed to make any working video game with the tools I used before. Ho, no, there is so much going on their I won't make it, I don't understand => the UI must hide frightening stuffs
- Ho, but I can do a platformer that easily, it's not that hard to make a game. I can do anything! => the UI should not annoy the user with toggle show/hide, they can take it.

AlexandreSi:
On my side, for each example I make, I use collisions, and each time I add the condition, I wonder if I should put Yes or No for the edge condition although it has never been useful to me => it confuses me and slows each time I set this condition.
But I know how to use each parameter of the Platformer behavior and I prefer seeing them all.
But I don't know if my experience is applicable to everyone

4ian:
Agree that this could ideally be per action/condition/behavior :)
You don't want the entire app to become complex at once, but progressively grow with you :) 

You even probably want to hide back stuff after checking some advanced stuff!

Not sure exactly about the UI should look like though - it's true that an accordion inside an accordion is not pretty.
Maybe something different.

tristanbob:
@D8H Thanks for this idea, I think it is a great one!

What is the UI for the extension creator?  For behavior properties, I was imagining another toggle similar to "Visible in editor".   There is probably enough space in the Property name field to put the "Advanced" toggle to the left of the visibility toggle.

The same "Advanced" toggle could be added to function parameters (in the same location, to the right of the parameter name).

On a side-note, both of those toggles should have basic instructions added on mouse-hover.

For the user experience, I like the way you have done it with "Show advanced parameters".  However, I think you should replace "Always show advanced parameters" with "Hide advanced parameters".  Then make this setting sticky, so it always remembers the last state it was left in.

Bouh:
An accordion is too much.
I like that the advanced settings buttons appear among the basic settings.

I agree with what [Tristan says](https://github.com/4ian/GDevelop/pull/3832#issuecomment-1103846933).

PR #3834, opened 2022-04-17
Add await tween action

arthuro555:
Adds an async action for awaiting the end of a tween. Depends on the following PR: #3535

### DIFF
--- a/Extensions/TweenBehavior/tweenruntimebehavior.ts
+++ b/Extensions/TweenBehavior/tweenruntimebehavior.ts
@@ -80,6 +80,18 @@ namespace gdjs {
     ];
   }
 
+  class TweenAwaiterTask extends gdjs.AsyncTask {
+    tween: TweenRuntimeBehavior.TweenInstance;
+    constructor(tween: TweenRuntimeBehavior.TweenInstance) {
+      super();
+      this.tween = tween;
+    }
+
+    update(runtimeScene: RuntimeScene) {
+      return this.tween.hasFinished;
+    }
+  }
+
   export class TweenRuntimeBehavior extends gdjs.RuntimeBehavior {
     private _tweens: Record<string, TweenRuntimeBehavior.TweenInstance> = {};
     private _runtimeScene: gdjs.RuntimeScene;
@@ -916,11 +928,19 @@ namespace gdjs {
       if (!this._tweenExists(identifier)) return;
 
       this._tweens[identifier].instance.stop();
+      // Since once removed it cannot play again, consider it has finished playing (for {this.awaitTween})
+      this._tweens[identifier].hasFinished = true;
       if (this._runtimeScene.shiftyJsScene)
         this._runtimeScene.shiftyJsScene.remove(
           this._tweens[identifier].instance
         );
       delete this._tweens[identifier];
+    }
+
+    awaitTween(identifier: string) {
+      return this._tweenExists(identifier)
+        ? new TweenAwaiterTask(this._tweens[identifier])
+        : new gdjs.ResolveTask();
     }
 
     /**
--- a/Extensions/TweenBehavior/JsExtension.js
+++ b/Extensions/TweenBehavior/JsExtension.js
@@ -37,45 +37,45 @@ module.exports = {
       )
       .setExtensionHelpPath('/behaviors/tween');
 
-      const easingChoices = JSON.stringify([
-        'linear',
-        'easeInQuad',
-        'easeOutQuad',
-        'easeInOutQuad',
-        'easeInCubic',
-        'easeOutCubic',
-        'easeInOutCubic',
-        'easeInQuart',
-        'easeOutQuart',
-        'easeInOutQuart',
-        'easeInQuint',
-        'easeOutQuint',
-        'easeInOutQuint',
-        'easeInSine',
-        'easeOutSine',
-        'easeInOutSine',
-        'easeInExpo',
-        'easeOutExpo',
-        'easeInOutExpo',
-        'easeInCirc',
-        'easeOutCirc',
-        'easeInOutCirc',
-        'easeOutBounce',
-        'easeInBack',
-        'easeOutBack',
-        'easeInOutBack',
-        'elastic',
-        'swingFromTo',
-        'swingFrom',
-        'swingTo',
-        'bounce',
-        'bouncePast',
-        'easeFromTo',
-        'easeFrom',
-        'easeTo',
-      ]);
-  
-      extension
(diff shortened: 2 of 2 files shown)


4ian:
Nice this should work well once we have the async action merged :)

4ian (commented):
For this PR, we should probably ensure the wait task is somehow dismissed if the object having the tween behavior is destroyed?

4ian:
> > For this PR, we should probably ensure the wait task is somehow dismissed if the object having the tween behavior is destroyed?
> 
> 🤔 If the tween behavior gets disabled or if the owner object gets deleted, the task should be deleted for sure, but I am less sure about whether the callback should be run or not. Technically could be seen as the end of a tween and I wonder if it couldn't break logic to have a callback just vanish and never be run.

Agree, I think it's weird to have a callback never run (if you don't want it to run, just add a condition to check if the object count is not 0 or something like this). In our model, any action that is "awaited" will always finish one day and the rest of the actions/sub events will for sure run "one day".

Extensions/TweenBehavior/tweenruntimebehavior.ts
```diff
@@ -104,16 +116,10 @@ namespace gdjs {
...
-      if (!shiftyJsScene) return;
-
       // Stop and delete all tweens of the behavior - otherwise they could:
       // - continue to point to the behavior, and so to the object (memory leak),
       // - affect the object in case it's recycled (wrong/hard to debug behavior).
-      for (const identifier in this._tweens) {
-        this._tweens[identifier].instance.stop();
-        shiftyJsScene.remove(this._tweens[identifier].instance);
-      }
+      for (const identifier in this._tweens) this.removeTween(identifier);
```
4ian, line 122: Is there a risk of memory leak? How long is the object/behavior still leaving in memory after it's removed from the scene?

arthuro555: At the beginning of the next frame, since removeTween calls hasFinished on the tween, the AsyncTasksManager should run the callback and delete the task, which has the reference to the tween which likely captured the RuntimeObject with its step function. 

The only risk of memory leaks because of that is if the object is somehow deleted without `onDestroy` being called, but that will most likely never happen.

4ian: I have bad gut feeling about this - didn't we had an issue about calling removeTween at some wrong point and we had to fix it but it was hard to debug?

I wonder if tasks awaiting should somehow be notified immediately that the object is destroyed and so they are considered as resolved (at the next tick sure, but so that they can drop the reference to the tween immediately).

What I mean is that the object instances are "hold" by the runtime scene, who is the owner of them, and objects own their behavior. When objects are destroyed, they notify their behaviors that they are being destroyed (onDestroy).
While the runtime scene can maybe decide to keep or not objects/behaviors in a cache to recycle them, it probably won't and so they should be removed from memory. In a language like C++, we would probably `delete` the objects.

But there, because the tasks have a reference to the tween, they force the behavior to stay in memory and be GC'ed later. It's a bit going in the wrong direction. It means a task could do stuff with the tween, while it's deleted. In C++, tasks could keep a weak pointer. In JS we can't, it's not a big deal, but we should still strive to avoid keeping references to things that are logically destroyed.
Otherwise, this liberty that we're taking is going to bite us one day when a task will be updated to check something else, that will call something else, that will end up modify a behavior that was supposed to be destroyed and still has somehow a side effect.

I'm not sure what the good solution is exactly, but tasks awaiting on a tween should probably be notified, so they are "aborted" or "force resolved" in a way.

arthuro555: > In C++, tasks could keep a weak pointer.

We can, actually, but Opera doesn't support it yet: https://caniuse.com/mdn-javascript_builtins_weakref

arthuro555: Is this satisfactory? 0322e0ecd9ed9c3b70c0eb34e137c9486ab0c96c
Or have I misunderstood what you wanted?

4ian: Yes I think that was the overall idea 👍 Be 100% sure that the lifecycle of objects/tween behavior/tasks is precise and consistently handled.
I think I saw a leak of the tasks, I will add a comment.

Extensions/TweenBehavior/tweenruntimebehavior.ts
```diff
@@ -76,6 +76,27 @@ namespace gdjs {
...
+      super(objectInstance);
+      this.tween = tween;
+    }
+
+    shouldResolve() {
+      return this.tween.hasFinished;
+    }
+
+    onObjectDeleted(): void {
+      //@ts-ignore - We are destoying the task to avoid leaking memory.
```
4ian, line 95:
```suggestion
      //@ts-ignore - We destroy the task to avoid leaking memory/allow tween memory to be reclaimed as soon as possible.
```

GDJS/Runtime/AsyncTasksManager.ts
```diff
@@ -58,13 +58,65 @@ namespace gdjs {
...
+   * get rid of the object reference and resolve to avoid memory leaks.
+   */
+  export abstract class ObjectBoundTask extends AsyncTask {
+    public readonly objectInstance: RuntimeObject;
+    private hasBeenDeleted = false;
+
+    constructor(objectInstance: RuntimeObject) {
+      super();
+      this.objectInstance = objectInstance;
+      objectInstance.registerDestroyCallback(() => {
```
4ian, line 81: We have to call `unregisterDestroyCallback` at some place, otherwise we risk creating a lot of tasks that will resolve and still be in memory because retained by the callback (which is retaining `this`) (and we'll lose precious CPU cycles too).

GDJS/Runtime/AsyncTasksManager.ts
```diff
@@ -58,13 +58,65 @@ namespace gdjs {
...
+        this.hasBeenDeleted = true;
+        this.onObjectDeleted();
+
+        // @ts-ignore At this point, the subclass code will not be called again,
```
4ian, line 85: Great comment and great idea 👍 Will reduce the risks of having leaks or memory not immediately claimed by the JS engine.

4ian (commented):
(Sorry I thought I already wrote some comments earlier today but seems I forgot to submit them somehow)

arthuro555:
(Bump, this should be ready for a review again)

PR #3840, opened 2022-04-20
Send events when opening and finalizing events extraction as function

AlexandreSi:
So that we know how often this feature is used, and if we modify/market it, we can track how good/bad it works

### DIFF
--- a/newIDE/app/src/EventsSheet/index.js
+++ b/newIDE/app/src/EventsSheet/index.js
@@ -120,6 +120,7 @@ type Props = {|
     extensionName: string,
     eventsFunction: gdEventsFunction
   ) => void,
+  onBeginCreateEventsFunction: () => void,
   unsavedChanges?: ?UnsavedChanges,
 |};
 
@@ -1316,6 +1317,7 @@ export class EventsSheetComponentWithoutHandle extends React.Component<
       onChooseResource,
       resourceExternalEditors,
       onCreateEventsFunction,
+      onBeginCreateEventsFunction,
     } = this.props;
     if (!project) return null;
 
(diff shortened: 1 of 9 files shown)


newIDE/app/src/EventsSheet/index.js
```diff
@@ -1502,6 +1504,7 @@ export class EventsSheetComponentWithoutHandle extends React.Component<
                         serializedEventsToExtract: null,
                       })
                     }
+                    onOpen={onBeginCreateEventsFunction}
```
4ian, line 1507: I would avoid `onOpen` (the parent is the one responsible for the child, so they should know when they open it) and add the call to onBeginCreateEventsFunction in `extractEventsToFunction`

4ian (approved):
LGTM without the onOpen

D8H:
I wonder if data about cancelled forms could be even more infighting than than filled forms.

AlexandreSi:
> I wonder if data about cancelled forms could be even more in-lighting than than filled forms.
> 
> * Do people see the form and get scared away or do they try to fill something and get stuck?

That's a piece of information that we should be able to deduct from this: Out of 53 openings, only 3 went through => 50 cancellations. Or do you feel we will miss something with this logic?

newIDE/app/src/EventsSheet/index.js
AlexandreSi: True, I will change that

D8H:
> > I wonder if data about cancelled forms could be even more in-lighting than than filled forms.
> > 
> > * Do people see the form and get scared away or do they try to fill something and get stuck?
> 
> That's a piece of information that we should be able to deduct from this: Out of 53 openings, only 3 went through => 50 cancellations. Or do you feel we will miss something with this logic?

What is filled when the dialog is cancelled could help understand what was not understood, but maybe this is too much data to send or not really necessary. I'm not sure.

AlexandreSi:
> What is filled when the dialog is cancelled could help understand what was not understood, but maybe this is too much data to send or not really necessary. I'm not sure.

I think that's too much data for the service we use (hard to visualize).

Identify what is blocking users on this dialog could be the goal of a UX study.

PR #3841, opened 2022-04-20
Add possibility to duplicate object group

AlexandreSi:
image

image

Also:
- Use React sortable virtualized list to render object groups list
- Match group context menu with object one


ClementPasteau (commented):
lgtm 🎉 
I'll let someone else approve, just because I'm not familiar with this part

newIDE/app/src/ObjectGroupsList/index.js
```diff
@@ -231,18 +154,51 @@ export default class GroupsListContainer extends React.Component<Props, State> {
...
+    const newGroup = container.insertNew(
+      newName,
+      container.getPosition(group.getName()) + 1
+    );
+
+    unserializeFromJSObject(
+      newGroup,
+      serializeToJSObject(group),
+      'unserializeFrom'
     );
```
D8H, line 188: It's probably the way it's done in other places in the code, but I find it strange to rely on serialization to copy the attributes of the group. Shouldn't containers have a `duplicate` method instead?

newIDE/app/src/ObjectGroupsList/index.js
```diff
@@ -340,45 +332,59 @@ export default class GroupsListContainer extends React.Component<Props, State> {
...
-      key: 'add-groups-row',
-      object: null,
-    });
 
     // Force List component to be mounted again if globalObjectGroups or objectGroups
     // has been changed. Avoid accessing to invalid objects that could
     // crash the app.
     const listKey = objectGroups.ptr + ';' + globalObjectGroups.ptr;
 
+    const renamedGroupWithContext = fullList.find(
```
4ian, line 345: Maybe useless to run the find if the group with context is null?

4ian (approved):

newIDE/app/src/ObjectGroupsList/index.js
AlexandreSi: I guess so but it kinda grows the scope of this PR 😅 
It's an improvement we could do later maybe? Wait for another time we need a duplicate action?

AlexandreSi:
Added in the last commits:
- Allow drop of local items on first displayed global item to put local items at the end of the local list (or right after last displayed local item if list is filtered)

PR #3848, opened 2022-04-21
Add analytics for extensions and examples.

D8H:
Should it be smoke tested?

### DIFF
--- a/newIDE/app/src/AssetStore/ExampleStore/index.js
+++ b/newIDE/app/src/AssetStore/ExampleStore/index.js
@@ -12,6 +12,10 @@ import { ExampleListItem } from './ExampleListItem';
 import { ResponsiveWindowMeasurer } from '../../UI/Reponsive/ResponsiveWindowMeasurer';
 import { ExampleDialog } from './ExampleDialog';
 import { type SearchMatch } from '../../UI/Search/UseSearchStructuredItem';
+import {
+  sendExampleDetailsOpened,
+  sendExampleChosenAsProject,
+} from '../../Utils/Analytics/EventSender';
 
 const styles = {
   searchBar: {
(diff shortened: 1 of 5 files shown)


newIDE/app/src/Utils/Analytics/EventSender.js
```diff
@@ -252,6 +252,42 @@ export const sendAssetAddedToProject = ({
...
+}: {|
+  id: string,
+  name: string,
+|}) => {
+  if (isDev || !client) return;
+
+  client.recordEvent('example-details-opened', { id, name });
+};
+
+export const sendExampleChosenAsProject = ({
```
4ian, line 279: I think this one is not needed because there is `sendNewGameCreated` already giving this info?

4ian (approved):

newIDE/app/src/Utils/Analytics/EventSender.js
```diff
@@ -146,6 +146,18 @@ export const sendExportLaunched = (exportKind: string) => {
   });
 };
 
+export const sendExampleDetailsOpened = ({
+  id,
+  name,
+}: {|
```
D8H, line 152: `sendNewGameCreated` send the project url. It won't be easy to follow.
The url is not known until the full example information are get. Do you think the id could be used instead?

4ian: Fair enough, then in this case I would recommend to send more info: `sendNewGameCreated({ templateName: example.projectFileUrl, exampleId: example.id, exampleSlug: example.slug })`

So we still have the old information ("templateName"... not super adapted but well that's what we have already) and the id and slug to do some stats (id because it's unique, slug because it's more readable and also unique!).

newIDE/app/src/AssetStore/ExampleStore/index.js
```diff
@@ -116,6 +120,10 @@ export const ExampleStore = ({ isOpening, onOpen, focusOnMount }: Props) => {
                     exampleShortHeader={exampleShortHeader}
                     matches={getExampleMatches(exampleShortHeader)}
                     onChoose={() => {
+                      sendExampleDetailsOpened({
+                        id: exampleShortHeader.id,
+                        name: exampleShortHeader.name,
```
4ian, line 125: Also send the slug (unique as the id, but easier to read).

4ian:
LGTM after adding the slug/id :)

PR #3849, opened 2022-04-22
Save project on creation

ClementPasteau:
Tested on Desktop and web, for an empty project and an example

Screen Recording 2022-04-22 at 13 47 21


4ian (approved):
Looks good, if you tested on web and desktop, blank or from an example, I guess it's fine.

PR #3850, opened 2022-04-22
Make loader screen cover input fields

fannieyan:
Do not show in changelog

Fixes input still showing over the loader div
image

### DIFF
--- a/Extensions/Leaderboards/leaderboardstools.ts
+++ b/Extensions/Leaderboards/leaderboardstools.ts
@@ -562,8 +562,9 @@ namespace gdjs {
 
               resetLeaderboardDisplayErrorTimeout(runtimeScene);
 
-              _leaderboardViewIframe =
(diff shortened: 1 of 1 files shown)


Extensions/Leaderboards/leaderboardstools.ts
```diff
@@ -105,6 +105,8 @@ namespace gdjs {
       _loaderContainer.style.width = '100%';
       _loaderContainer.style.justifyContent = 'center';
       _loaderContainer.style.alignItems = 'center';
+      _loaderContainer.style.position = 'absolute';
```
AlexandreSi, line 108: Is the `absolute` needed?

fannieyan: Z-index must be used with "positioned box". I am not sure if there are other positions that are more "neutral" than absolute and that is not "static" ?

Issue #2975, opened 2021-08-26
[$200+ Bounty] Asynchronous events


arthuro555:
## Description

Asynchronous events are events that do not stop the execution flow of the events sheet. Instead, they will continue to run in the background while the game continues running the game loop. Once the action finishes its task, the object context at the time of its execution is restored and actions and subevents after that action are run. 

They are used in the event sheets like normal actions but have a clear visual clue and links to documentation to explain their difference from normal ones.

You can find a complete discussion about the feature [here](https://forum.gdevelop-app.com/t/asynchronous-instructions-in-gdevelop/26197/3?u=arthuro555).

### Tasks required to claim the bounty
- [ ] Add asynchronous actions to the core and JavaScript platform (`.SetAsynchronous()` that can be called on actions/conditions when declaring them, code generation, etc)
- [ ] Add a way to back up and restore (without deleted objects) the object context. 
  - Investigate if other stuff should be included in this context. Objects are a central part of the events metaphor in GDevelop, so they are for sure included.
- [ ] Add a "wait" condition/action that resolves after the time passed as a parameter.
- [ ] Verify how asynchronous actions behave when used in custom behaviors or events based functions.
  - Notably, a deleted object should probably result in any of the "wait" related to it (i.e: launched in a behavior method) to be discarded.

### Additional tasks
- [ ] Convert existing asynchronous actions to that system (Network, Firebase, Filesystem...)
- [ ] Add asynchronous conditions
- [ ] Add support for creating events based asynchronous actions.
- [ ] Add an action that accepts asynchronous actions as subinstructions that runs them all at once and continues once they are all finished (equivalent to JavaScript `Promise.all`)
- [ ] Add action that runs all subinstructions asynchronous actions and continues once any of the finishes (equivalents to `Promise.any`)
- [ ] Check if it's possible to add asynchronous expressions?

## Alternatives considered

You can find a discussion about alternatives on [the forum](https://forum.gdevelop-app.com/t/asynchronous-instructions-in-gdevelop/26197/3?u=arthuro555).

arthuro555:
Note that ultimately this could allow for much powerful features such as first class functions (by making a condition that accepts a variable, sets the variable to a uid, maps that uid to a resolver function, and then having an action that takes that variable, gets the resolver from the uid and calls it to effectively call the function.

4ian:
Thanks for opening this! 👍
I've reordered a bit the list of tasks to:

[...]

PR #3852, opened 2022-04-23
Add support for asynchronous actions & Wait action

4ian:
Based on #3535
Closes https://github.com/4ian/GDevelop/issues/2975

The code generation context were slightly adapted to avoid considering the objects coming from outside of an async event as "not declared": 
  - If they are declared by a parent, they should be kept considered as "declared" in the async callbacks and sub events. What I mean is that having an async event (async action) is not changing the logic of picking of objects. 

Instead, we check when getting the lists from the parent (`declareObjectListFromParent`) if it's an async list or not (`ShouldUseAsyncObjectsList`). For this, we:
- We get the nearest async parent (which is the context itself if it's async, or the nearest parent otherwise).
- Either the object was used in or after the async parent. So it's not coming from an async list.
- Or it was not, so we ask the async parent if it was already declared in its parent(s). If yes: it's an async list. If not: it's declared as usual in the async parent, so it's not an async list.

Note that the naive implementation of this was more costly because all async contexts, when saving objects in async object lists, were saving all objects needed in their children, even if not used before one (or multiple) other children async callbacks.
This is solved by 1) not saving objects in the new `LongLivedObjectsList` if it's already coming from a parent `LongLivedObjectsList` (so it's still important that `LongLivedObjectsList` has a connection to its parent).
 
Other changes:
- `InheritsAsAsyncCallbackFrom` is just `InheritsFrom` with the async depth counter increased.


4ian:
Remaining to do:

- [ ] Fix async event still listed in the list of event types
- [ ] Mark it as experimental
- ???
- [ ] Merge!

arthuro555:
>  Fix async event still listed in the list of event types

🤔 It didn't do that for me before... Maybe I forgot to rename the filter after moving the event from the common instructions extension to its own async extension?

> Remove "Time" category for the Wait action

🤔 What category should it go to if not time?

---

I'd gladly do those changes but I am afraid that this is not my branch anymore, so I can't commit... 😓

4ian:
> I'd gladly do those changes but I am afraid that this is not my branch anymore, so I can't commit... 😓

It's a bit cumbersome but you can actually open a PR to be merged in... another PR (this branch) ;)

I've done the fix.
The extensions are now acting as a category, so there was no need to have this extra "Time" category. Also a confusion in the event type.

4ian:
Aaaand it's merged! 🎉🎉🎉

I anticipate this to be super super useful and used a lot in future games.
Thanks a lot @arthuro555 for your work on this, this will open a new way of building game events :) We can now have other actions like the tween added - do you want to set up the PR for it? :)

PR #3857, opened 2022-04-26
Add UI to customize leaderboard appearance

AlexandreSi:
For now, only the formatting of the score can be configured

Component story: http://gdevelop-storybook.s3-website-us-east-1.amazonaws.com/customize-leaderboard/latest/index.html?path=/story/leaderboardappearancedialog--default


newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardAppearanceDialog.js
```diff
@@ -0,0 +1,330 @@
...
+            <Trans>Score formatting</Trans>
+          </Text>
+          <Column>
+            <Line>
+              <SelectField
+                fullWidth
+                value={scoreType}
+                floatingLabelText={<Trans>Score display</Trans>}
+                onChange={(e, i, newValue) =>
+                  // $FlowIgnore
```
ClementPasteau, line 182: should we instead do a `if (newValue !== 'custom' && newValue !== 'time') setScoreType('time')}` ?

I'm not sure this is that useful tbh, as it's checked in the backend too.

newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardAppearanceDialog.js
```diff
@@ -0,0 +1,330 @@
...
+                    </Line>
+                    <Line expand>
+                      <TextField
+                        fullWidth
+                        floatingLabelText={<Trans>Suffix</Trans>}
+                        maxLength={10}
+                        value={suffix}
+                        hintText={t`Ex: gold`}
+                        onChange={(e, newValue) => {
+                          setSuffix(newValue);
```
ClementPasteau, line 227: From your storybook example, it looks like the user has to enter the space in the suffix.
Should the space be always included by default meaning they will not forget it? (I can't think of a case where a user would not want a space between their score & suffix?)

newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardAppearanceDialog.js
```diff
@@ -0,0 +1,330 @@
...
+      ? leaderboardCustomizationSettings.scoreFormatting.suffix
+      : ''
+  );
+  const [precision, setPrecision] = React.useState<number>(
+    leaderboardCustomizationSettings &&
+      leaderboardCustomizationSettings.scoreFormatting.type === 'custom'
+      ? leaderboardCustomizationSettings.scoreFormatting.precision
+      : 0
+  );
+  const [units, setUnits] = React.useState<string>(
```
ClementPasteau, line 103: maybe call that `timeUnits/setTimeUnits` for readability

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
```diff
@@ -169,6 +171,10 @@ export const LeaderboardAdmin = ({ onLoading, project }: Props) => {
           ? 'leaderboardNameUpdate'
           : payload.sort
           ? 'leaderboardSortUpdate'
+          : payload.visibility
```
ClementPasteau, line 174: maybe this could become a small function `getApiError(payload)` with a switch case?
returning :
`{ action: xxx, message: yyy}`

newIDE/app/src/Leaderboard/LeaderboardScoreFormatter.js
```diff
@@ -0,0 +1,79 @@
...
+
+export const formatCustomScore = (
+  score: number,
+  options: LeaderboardScoreFormattingCustom
+): string => {
+  const roundedScore =
+    Math.round(score * 10 ** options.precision) / 10 ** options.precision;
+  return `${options.prefix}${roundedScore.toFixed(
+    Math.max(0, options.precision)
+  )}${options.suffix}`;
```
ClementPasteau, line 70: see other comment.
I'm wondering if prefix/suffix should automatically add a space before/after the score

newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardAppearanceDialog.js
4ian: We might want to display something like `15$`, so requiring the space is maybe good - though a bit weird I agree.

ClementPasteau: yup, as discussed, even if a bit weird, it's better 👍

newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardAppearanceDialog.js
AlexandreSi: Yes, I think that it would make the code heavy for something that is safe already

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
```diff
@@ -794,90 +835,117 @@ export const LeaderboardAdmin = ({ onLoading, project }: Props) => {
...
+              leaderboardCustomizationSettings={
+                currentLeaderboard
+                  ? currentLeaderboard.customizationSettings
+                  : undefined
+              }
+              onClose={() => setIsEditingAppearance(false)}
+              onSave={async (
+                customizationSettings: LeaderboardCustomizationSettings
+              ) => {
+                try {
```
ClementPasteau, line 938: should we catch and show an error box if it errors?

ClementPasteau (approved):

Bouh:
## Suggestion

image

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
AlexandreSi: That's what `onUpdateLeaderboard` does

Bouh:
##  Suggestion

image

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
ClementPasteau: so you don't need the try here as it will not throw?

AlexandreSi: I'm using this for the `finally` so that, whatever happens (unexpected error), the dialog is closed and you can see the admin dialog where api errors are displayed

PR #3858, opened 2022-04-27
Add expressions to specify decimal place where to round, floor and ceil

AlexandreSi:
`roundTo(15.583, 2) => 15.58`
`ceilTo(15.583, 2) => 15.59`
`floorTo(15.583, 1) => 15.5`

Second argument is optional and will be considered as 0 if not present

### DIFF
--- a/GDJS/Runtime/events-tools/commontools.ts
+++ b/GDJS/Runtime/events-tools/commontools.ts
@@ -306,6 +306,51 @@ namespace gdjs {
       ): number {
         return distance * Math.sin(gdjs.toRad(angle));
       };
+
+      /**
+       * Rounds a number to the Nth decimal place
+       * @param {float} value
+       * @param {number} decimalPlace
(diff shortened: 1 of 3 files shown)


GDJS/Runtime/events-tools/commontools.ts
```diff
@@ -306,6 +306,51 @@ namespace gdjs {
...
+       * @param {number} decimalPlace
+       * @returns the rounded value
+       */
+      export const roundTo = function (
+        value: float,
+        decimalPlace: number
+      ): number {
+        if (!decimalPlace || !Number.isInteger(decimalPlace))
+          return Math.round(value);
+        return Math.round(value * 10 ** decimalPlace) / 10 ** decimalPlace;
```
4ian, line 322: (paraniod mode: double check this ** notation support in browser)

4ian (approved):
LGTM

PR #3860, opened 2022-04-27
Remove variable parameters from SavePlayerScore action.

D8H:
image

### DIFF
--- a/Extensions/Leaderboards/leaderboardstools.ts
+++ b/Extensions/Leaderboards/leaderboardstools.ts
@@ -159,9 +159,7 @@ namespace gdjs {
         runtimeScene: gdjs.RuntimeScene,
         leaderboardId: string,
         score: float,
-        playerName: string,
-        responseVar: gdjs.Variable,
-        errorVar: gdjs.Variable
+        playerName: string
       ) {
         let scoreSavingState: ScoreSavingState;
         if (_scoreSavingStateByLeaderboard[leaderboardId]) {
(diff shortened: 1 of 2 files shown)


Extensions/Leaderboards/leaderboardstools.ts
```diff
@@ -233,30 +227,25 @@ namespace gdjs {
                 response.statusText
               );
               scoreSavingState.setError(errorCode);
-              errorVar.setString(errorCode);
               return;
             }
 
             scoreSavingState.closeSaving();
 
             return response.text().then(
```
AlexandreSi, line 235: This can be removed I think

D8H:
I modified the example not to use these parameters and check the "submit" button after a submission and another run.
* https://github.com/GDevelopApp/GDevelop-examples/pull/256/commits/7dedfb7312d2e57f1792ce16881dfc87e02b4772

Extensions/Leaderboards/leaderboardstools.ts
D8H: There is condition to check a failure. An expression can be added later to give more details about the error, but this is not mandatory.

AlexandreSi: That's ok with me

AlexandreSi (approved):

Extensions/Leaderboards/leaderboardstools.ts
4ian: Yes I think it's good to keep 👍

4ian (approved):

PR #3862, opened 2022-04-27
Update supported filetypes for resources

arthuro555:
[Untested]

Adds webm, jpeg and webp as supported formats in the IDE. Those formats are all [already supported by PIXI](https://github.com/pixijs/pixijs/blob/8c666471b8b49fb39b2d6f3f64a283153d337fd7/packages/loaders/src/LoaderResource.ts#L1347), widely used, fast and modern (except JPEG)

### DIFF
--- a/newIDE/app/src/ResourcesList/ResourceSource.js
+++ b/newIDE/app/src/ResourcesList/ResourceSource.js
@@ -26,7 +26,7 @@ export const allResourceKindsAndMetadata = [
   {
     kind: 'image',
     displayName: t`Image`,
-    fileExtensions: ['png', 'jpg'],
+    fileExtensions: ['png', 'jpg', 'jpeg', 'webp'],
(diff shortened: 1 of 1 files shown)


4ian:
Can we get a table of support across browsers? We might want to warn users for formats not supported widely.

Silver-Streak:
I was just digging into this to respond to a forum post. Seems that most if not all modern browsers support it, and electron 18 (if not much older) would as well. 

https://developers.google.com/speed/webp/faq

tristanbob:
> Can we get a table of support across browsers? We might want to warn users for formats not supported widely.

I love this website for these type of questions:

https://caniuse.com/webp
91.38% (supported) + 1.83% (partially supported)	= 93.21%

https://caniuse.com/webm
77% (supported) + 17.3% (partially supported) = 94.30%

Bouh:
> [Untested]
> 

Did someone test it?

PR #3864, opened 2022-04-28
More leaderboard customization

AlexandreSi:
Add parameter in leaderboard appearance dialog to control number of entries to display

Current layout for number of entries to display:

image

New button to choose the default leaderboard to display on Liluo

demo_default_leaderboard

### DIFF
--- a/newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardAppearanceDialog.js
+++ b/newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardAppearanceDialog.js
@@ -209,6 +209,9 @@ function LeaderboardAppearanceDialog({
             />,
           ]}
         >
+          <Text size="title">
+            <Trans>Table settings</Trans>
+          </Text>
           <Line>
             <TextField
               fullWidth
--- a/newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardAppearanceDialog.js
+++ b/newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardAppearanceDialog.js
@@ -171,6 +171,7 @@ function LeaderboardAppearanceDialog({
     }
     setIsLoading(true);
     const customizationSettings = {
+      defaultDisplayedEntriesNumber,
       scoreTitle,
       scoreFormatting:
         scoreType === 'custom'
@@ -215,6 +216,9 @@ function LeaderboardAppearanceDialog({
               key={'save'}
             />,
           ]}
+          onApply={() => {
+            onSaveSettings(i18n);
+          }}
         >
           <Text size="title">
             <Trans>Table settings</Trans>
--- a/newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardAppearanceDialog.js
+++ b/newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardAppearanceDialog.js
@@ -33,6 +33,9 @@ const unitToAbbreviation = {
   millisecond: 'ms',
 };
 
+const isWholeNumber = (value: any): boolean =>
+  !isNaN(value) && Number.isInteger(value);
+
 const getIdentifierFromUnits = (units: {|
   smallestUnit: LeaderboardScoreFormattingTimeUnit,
   biggestUnit: LeaderboardScoreFormattingTimeUnit,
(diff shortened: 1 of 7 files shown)


Bouh:
I'll just chime on the missing help button to the documentation. Otherwise LGTM 👍

AlexandreSi:
I'm working on the wiki page for the appearance to add a Help button in the dialog

newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardAppearanceDialog.js
```diff
@@ -134,6 +150,25 @@ function LeaderboardAppearanceDialog({
       setScoreTitleError(i18n._(t`Title cannot be empty.`));
       return;
     }
+    if (
+      isNaN(defaultDisplayedEntriesNumber) ||
+      !Number.isInteger(defaultDisplayedEntriesNumber)
```
ClementPasteau, line 155: maybe create a tiny helper `isValidEntry()` that does those 2 checks (and re-use below)

newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardAppearanceDialog.js
```diff
@@ -175,6 +216,36 @@ function LeaderboardAppearanceDialog({
...
+            <Trans>Table settings</Trans>
+          </Text>
+          <Line>
+            <TextField
+              fullWidth
+              type="number"
+              floatingLabelText={<Trans>Number of entries to display</Trans>}
+              value={
+                isNaN(defaultDisplayedEntriesNumber)
+                  ? ''
```
ClementPasteau, line 229: if it's a `type="number"` this should probably be a number too no?
(This check may not be needed as you're doing it above too)

newIDE/app/src/Leaderboard/LeaderboardProvider.js
```diff
@@ -122,13 +125,30 @@ const reducer = (state: ReducerState, action: ReducerAction): ReducerState => {
         displayOnlyBestEntry: action.payload,
       };
     case 'UPDATE_OR_CREATE_LEADERBOARD':
+      let leaderboardsByIdsWithUpdatedPrimaryFlags;
```
ClementPasteau, line 128: you can probably initiate this variable to `const leaderboardsByIdsWithUpdatedPrimaryFlags = {}` so that you just assign value/key afterwards instead of reassigning the object

newIDE/app/src/Leaderboard/LeaderboardProvider.js
```diff
@@ -122,13 +125,30 @@ const reducer = (state: ReducerState, action: ReducerAction): ReducerState => {
...
+        Object.entries(state.leaderboardsByIds).forEach(
+          ([leaderboardId, leaderboard]) => {
+            leaderboardsByIdsWithUpdatedPrimaryFlags[leaderboardId] = {
+              ...leaderboard,
+              // $FlowFixMe: known error where Flow returns mixed for object value https://github.com/facebook/flow/issues/2221
+              primary: action.payload.primary ? undefined : leaderboard.primary,
+            };
+          }
+        );
+        leaderboardsByIdsWithUpdatedPrimaryFlags[action.payload.id] =
```
ClementPasteau, line 144: you're doing the same thing twice (in `if` and `else`) this could be outside of if/else? (and reverse your if logic)

newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardAppearanceDialog.js
AlexandreSi: It's necessary to allow users to delete the whole number and type a new, otherwise, when deleting the last digit, it sets the value to 0

newIDE/app/src/Leaderboard/LeaderboardProvider.js
AlexandreSi: Indeed, that's possible! Thanks
I was paranoid that Flow wouldn't accept it, I feel like TS wouldn't have accepted it (you set keys in an untyped object)

newIDE/app/src/Utils/GDevelopServices/Play.js
```diff
@@ -121,14 +124,14 @@ export const listGameLeaderboards = async (
 
   const { uid: userId } = firebaseUser;
   const authorizationHeader = await getAuthorizationHeader();
-  return axios
-    .get(`${GDevelopPlayApi.baseUrl}/game/${gameId}/leaderboard`, {
+  const response = await axios.get(
+    `${GDevelopPlayApi.baseUrl}/game/${gameId}/leaderboard?deleted=false`,
```
ClementPasteau, line 128: nipicking: maybe this function could be called `listActiveGameLeaderboards()` to be explicit that you will not receive the deleted ones if you call it

ClementPasteau (approved):
lgtm!

newIDE/app/src/Utils/GDevelopServices/Play.js
AlexandreSi: Good idea!

PR #3865, opened 2022-04-28
Remove the dependency to SFML.

D8H:


Core/GDCore/String.cpp
```diff
@@ -112,7 +86,7 @@ String::const_iterator String::end() const
 String String::FromLocale( const std::string &localizedString )
 {
 #if defined(WINDOWS)
-    return FromSfString(sf::String(localizedString)); //Don't need to use the current locale, on Windows, std::locale is always the C locale
+    return FromUTF8(localizedString); //Don't need to use the current locale, on Windows, std::locale is always the C locale
```
D8H, line 89: I'm not sure about this.

Core/GDCore/String.cpp
```diff
@@ -164,7 +133,7 @@ String String::FromWide( const std::wstring &wstr )
 std::string String::ToLocale() const
 {
 #if defined(WINDOWS)
-    return ToSfString().toAnsiString();
+    return m_string;
```
D8H, line 136: I'm not sure about this either.

Core/GDCore/Vector2.h
```diff
@@ -0,0 +1,360 @@
+// This code comes from the SFML.
```
D8H, line 1: Should it be formatted?

D8H:
On Windows, the UTF8 seams give issues:
```
C:\WorkspaceGDevelop\GDevelop\build-tests\Core>GDCore_tests.exe
terminate called after throwing an instance of 'std::out_of_range'
  what():  [gd::String::substr] starting pos greater than size
```

Core/CMakeLists.txt
```diff
@@ -88,5 +89,5 @@ if(BUILD_TESTS)
 	add_executable(GDCore_tests ${test_source_files})
 	set_target_properties(GDCore_tests PROPERTIES BUILD_WITH_INSTALL_RPATH FALSE) #Allow finding dependencies directly from build path on Mac OS X.
 	target_link_libraries(GDCore_tests GDCore)
-	target_link_libraries(GDCore_tests ${sfml_LIBRARIES})
+	target_link_libraries(GDCore_tests ${CMAKE_DL_LIBS})
```
D8H, line 92: I had to add this. It was probably done implicitly by the SFML dependency.

Core/GDCore/Vector2.h
4ian: Not mandatory for C++ files (for now).

```suggestion
// This is adapted from SFML (https://github.com/SFML/SFML).
```

Core/tests/Utf8Tests.cpp
```diff
@@ -19,10 +18,8 @@ TEST_CASE("Utf8 String", "[common][utf8]") {
   SECTION("ctor & conversions") {
     gd::String str = u8"UTF8 a été testé !";
 
-    sf::String sfStr = str;
     std::u32string u32str = str.ToUTF32();
 
-    REQUIRE(str == gd::String::FromSfString(sfStr));
     REQUIRE(str == gd::String::FromUTF32(u32str));
   }
```
4ian, line 25: As said in another comment, let's add some tests (check size, substr, assignement from a string to another, append a string to another) with emojis and CJK characters.

D8H:
I tried to add the ansi conversion from SFML, but it doesn't make any difference in the tests and tests.
* ExpressionParser2NaugtyStrings would be interesting to run but it fails at opening the file.
* I thought that the failure on memory consumption was due to the extra string conversion but it seems to act the same without it.

```
C:\WorkspaceGDevelop\GDevelop\build-tests\Core>xcopy /y "C:\WorkspaceGDevelop\GDevelop\Binaries\Output\Release_Windows\GDCore.dll" .
C:\WorkspaceGDevelop\GDevelop\Binaries\Output\Release_Windows\GDCore.dll
1 fichier(s) copié(s)

C:\WorkspaceGDevelop\GDevelop\build-tests\Core>GDCore_tests.exe -e
Start checking naughty strings...

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
GDCore_tests.exe is a Catch v1.0 b52 host application.
Run with -? for options

-------------------------------------------------------------------------------
ExpressionParser2 - Naughty strings
  Check that no naughty string crash the parser
-------------------------------------------------------------------------------
../Core/tests/ExpressionParser2NaugtyStrings.cpp:16
...............................................................................

../Core/tests/ExpressionParser2NaugtyStrings.cpp:44: FAILED:
explicitly with message:
  Can't open ../Core/tests/ExpressionParser2NaugtyStrings.cpp-blns.txt

-------------------------------------------------------------------------------
EventsList
  Memory consumption
-------------------------------------------------------------------------------
../Core/tests/Common.cpp:29
...............................................................................

../Core/tests/Common.cpp:78: FAILED:
  REQUIRE( 1500 >= endMemory - startMemory )
with expansion:
  1500 >= 2692
with message:
  Memory used: 2692KB

===============================================================================
test cases:   40 |   38 passed | 2 failed
assertions: 3180 | 3178 passed | 2 failed
```

4ian:
> ExpressionParser2NaugtyStrings would be interesting to run but it fails at opening the file.

Strange, can you try with a hardcoded path? Try with backslash and forward slashes.

> I thought that the failure on memory consumption was due to the extra string conversion but it seems to act the same without it.

I think this test is not relevant on Windows because the memory layouting must be different. ~2-3M of memory consumption is fine. 
This test was rather to protect some kind of memory leak where memory consumption would just to easily 50 or 100M.

> I tried to add the ansi conversion from SFML, but it doesn't make any difference in the tests result.

Thanks for testing! 
If we can somehow run ExpressionParser2NaugtyStrings and the memory test, then everything will run right?
So if the ANSI conversion is not making any change.... let's revert this commit and simplify the code so we just don't do conversions, much like on Linux/macOS :) 

These are tests only, so it only matters that they run. The GDevelop.js version is compiled by Emscripten, so it does not depend if its compiled on Windows, mac or Linux - so we should not introduce any risk of regression.

D8H:
> If we can somehow run ExpressionParser2NaugtyStrings and the memory test, then everything will run right?

I forgot to mention that I disabled tests that check exceptions with `-e` because it seems to make the suite crash on Windows.

4ian:
Ah nice! What if you try to add a try/catch around them? Otherwise we can remove them or add `#if !defined(WINDOWS)` (or similar) around the REQUIRE_THROWS.

D8H:
About the naughty strings test, the script current folder wasn't right.

I reverted the changes on Ansi conversion. Maybe modern Windows use a standard library that handle UTF8. This script allows to run the tests.
```
xcopy /y "..\Binaries\Output\Release_Windows\GDCore.dll" Core\
"Core\GDCore_tests.exe" -e
PAUSE
```

4ian (approved):
🎉
This should be safe if all tests are passing and again we should not have really touched codepaths used by Emscripten for UTF8.
Let's merge this :)

PR #3866, opened 2022-04-29
Encode thumbnail URL before sending it to the GDevelop services

ClementPasteau:
This fixes the case where you would use a thumbnail which has non-conventional characters in the name.
Then the URL saved in the GDevelop services would be different than the actual URL of the thumbnaill.

### DIFF
--- a/newIDE/app/src/Utils/GDevelopServices/Build.js
+++ b/newIDE/app/src/Utils/GDevelopServices/Build.js
@@ -80,7 +80,11 @@ export const getWebBuildThumbnailUrl = (
   }
   // The exporter put asset files directly in the build folder.
   // It's not factorized with the exporter because it's a temporary solution.
-  return `https://games.gdevelop-app.com/game-${buildId}/${fileName}`;
+  // TODO: Upload the thumbnail separately, so that this URL is not defined by the frontend.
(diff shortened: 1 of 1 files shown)


newIDE/app/src/Utils/GDevelopServices/Build.js
```diff
@@ -80,7 +80,10 @@ export const getWebBuildThumbnailUrl = (
   }
   // The exporter put asset files directly in the build folder.
   // It's not factorized with the exporter because it's a temporary solution.
-  return `https://games.gdevelop-app.com/game-${buildId}/${fileName}`;
+  const uri = `https://games.gdevelop-app.com/game-${buildId}/${fileName}`;
+  // The backend services encode the file URLs when uploaded, so we need to do the same before saving the value.
```
AlexandreSi, line 84: It's a case where you expect the same behaviour between frontend and backend. I wonder if it is worth having unit tests for that, so that when changing it in the future, you really ask yourself if this has some impact

AlexandreSi (approved):

newIDE/app/src/Utils/GDevelopServices/Build.js
AlexandreSi: Ok, maybe just add a TODO then

PR #3870, opened 2022-05-03
Add tutorials in app

fannieyan:
Added dismissable tutorials : 
- In Events Sheet
imageimage
- In Variables Editors
image
- In Object selection
image
- In Sprite Animation
image
- In Publish Game
image
- In extension store
image

Added Tutorial buttons: 
- In empty states
image
- In Externals
image

### DIFF
--- a/newIDE/app/src/EventsSheet/EventsTree/index.js
+++ b/newIDE/app/src/EventsSheet/EventsTree/index.js
@@ -439,6 +439,7 @@ export default class ThemableEventsTree extends Component<EventsTreeProps, *> {
                 description={<Trans>Events define the rules of a game.</Trans>}
                 actionLabel={<Trans>Add an event</Trans>}
                 helpPagePath="/events"
+                tutorialId="intro-event-system"
                 onAdd={() =>
                   this.props.onAddNewEvent(
                     'BuiltinCommonInstructions::Standard',
--- a/newIDE/app/src/EventsSheet/EventsTree/index.js
+++ b/newIDE/app/src/EventsSheet/EventsTree/index.js
@@ -32,6 +32,11 @@ import ThemeConsumer from '../../UI/Theme/ThemeConsumer';
 import BottomButtons from './BottomButtons';
 import { EmptyPlaceholder } from '../../UI/EmptyPlaceholder';
 import { CorsAwareImage } from '../../UI/CorsAwareImage';
+import { Line } from '../../UI/Grid';
+import { type Preferences } from '../../MainFrame/Preferences/PreferencesContext';
+import { type Tutorial } from '../../Utils/GDevelopServices/Tutorial';
+import TutorialMessage from '../../Hints/TutorialMessage';
+import getTutorial from '../../Hints/getTutorial';
 const gd: libGDevelop = global.gd;
 
 const getThumbnail = ObjectsRenderingService.getThumbnail.bind(
@@ -227,6 +232,9 @@ type EventsTreeProps = {|
   windowWidth: WidthType,
   eventsSheetHeight: number,
   fontSize?: number,
+
+  preferences: Preferences,
+  tutorials: ?Array<Tutorial>,
 |};
 
 // A node displayed by the SortableTree. Almost always represents an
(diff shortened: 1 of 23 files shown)


newIDE/app/src/EventsSheet/EventsTree/index.js
```diff
@@ -267,6 +275,27 @@ export default class ThemableEventsTree extends Component<EventsTreeProps, *> {
...
+    const {
+      values: { hiddenTutorialHints: previousHiddenTutorialHints },
+    } = prevProps.preferences;
+    if (
+      hiddenTutorialHints['intro-event-system'] !==
+      previousHiddenTutorialHints['intro-event-system']
+    ) {
+      this.setState({
+        ...this.state,
+        treeData: this.state.treeData.slice(
```
4ian, line 291: Is there a way to do this filtering but without relying on the fact that this tutorial line is at the end (which might change in the future, and would create an invisible bug)?

newIDE/app/src/EventsSheet/index.js
```diff
@@ -1355,207 +1356,219 @@ export class EventsSheetComponentWithoutHandle extends React.Component<
...
-                    }
-                    onCloseSearchPanel={() => {
-                      this._closeSearchPanel();
-                    }}
-                    onGoToNextSearchResult={() =>
-                      this._ensureEventUnfolded(goToNextSearchResult)
-                    }
-                    searchFocusOffset={searchFocusOffset}
-                  />
+              <TutorialContext.Consumer>
```
4ian, line 1359: Can you instead `useContext`  in the `EventsSheet` at the bottom of the file? (and pass it to `EventsSheetComponentWithoutHandle`)

newIDE/app/src/Export/ExportDialog/index.js
```diff
@@ -174,6 +175,11 @@ const ExportDialog = ({
       ]}
       secondaryActions={[
         <HelpButton key="help" helpPagePath={exporter.helpPage} />,
+        <TutorialButton
```
4ian, line 178: Can we only show this if the exporter name is `local-html5` or `browser-html5`?

newIDE/app/src/Hints/getTutorial.js
```diff
@@ -0,0 +1,28 @@
+// @flow
+import { type Tutorial } from '../Utils/GDevelopServices/Tutorial';
+import { type Preferences } from '../MainFrame/Preferences/PreferencesContext';
+
+/**
+ * Function that returns a tutorial if it can be found, otherwise returns null.
```
4ian, line 6: Uber nitpicking:
Good practice is to avoid saying "Function that returns [...]" and just say "Returns [...]". This is because you should not need to repeat the nature/type of the commented thing, because it should be obvious.

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/SpacedDismissableTutorialMessage.js
```diff
@@ -0,0 +1,22 @@
+// @flow
+import * as React from 'react';
+import { Column, Spacer } from '../../../UI/Grid';
+import useDismissableTutorialMessage from '../../../Hints/useDismissableTutorialMessage';
+
+/**
+ * TODO : Use context directly in SpriteEditor
```
4ian, line 7: Sad but ok there.

Ubernitpicking, no space before colon:
```suggestion
 * TODO: Use context directly in SpriteEditor
```

newIDE/app/src/UI/TutorialButton/index.js
```diff
@@ -0,0 +1,43 @@
...
+ * The button that can be used in any dialog to open a Youtube tutorial
+ */
+const TutorialButton = (props: PropsType) => {
+  const { tutorials } = React.useContext(TutorialContext);
+  if (!tutorials || !props.tutorialId) return props.renderIfNotFound || null; // Loading or errored, do not display the tutorial.
+  const tutorial: ?Tutorial = tutorials.find(
+    tutorial => tutorial.id === props.tutorialId
+  );
+  if (!tutorial) {
+    console.warn(`Tutorial not found`);
```
4ian, line 26: Either remove the log or:

```suggestion
    console.warn(`Tutorial with id "${tutorialId}" not found.`);
```

newIDE/app/src/EventsSheet/EventsTree/index.js
4ian: Yes, it's not ideal but can be a bit more bullet proof by filtering (though you also need to add something to differentiate the tutorial from the rest)

fannieyan: I added a key `eventstree-tutorial-node` to the tutorial node (I felt it was more reliable than using "indexInList")! :)

4ian: Sounds good!

PR #3871, opened 2022-05-04
Fix not setting authors correctly when publishing to Liluo

ClementPasteau:
When publishing a build, only the slug was set, not the authors (via the ACLs)

### DIFF
--- a/newIDE/app/src/Export/GenericExporters/OnlineWebExport/OnlineGameLink.js
+++ b/newIDE/app/src/Export/GenericExporters/OnlineWebExport/OnlineGameLink.js
@@ -42,6 +42,8 @@ import {
   getGameSlugs,
   type Game,
   type GameSlug,
+  getAclsFromUserIds,
+  setGameUserAcls,
 } from '../../../Utils/GDevelopServices/Game';
 import AuthenticatedUserContext from '../../../Profile/AuthenticatedUserContext';
 import AlertMessage from '../../../UI/AlertMessage';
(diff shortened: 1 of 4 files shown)


newIDE/app/src/Export/GenericExporters/OnlineWebExport/OnlineGameLink.js
```diff
@@ -201,34 +278,15 @@ const OnlineGameLink = ({
...
-            });
-            return false;
-          }
+        // Then set the authors.
+        const authorSuccess = await tryUpdateAuthors(i18n);
+        if (!authorSuccess) {
+          return false;
+        }
+        // Then set the slug if it has been filled.
+        const slugSuccess = await tryUpdateSlug(partialGameChange, i18n);
```
D8H, line 287: When the author update doesn't succeed, the game is updated but not the slug. Is it ok?

4ian: Good catch, we should probably run both in parallel :)

newIDE/app/src/Export/GenericExporters/OnlineWebExport/OnlineGameLink.js
```diff
@@ -125,6 +127,80 @@ const OnlineGameLink = ({
...
+            'Unable to update the game slug:',
+            error.response || error.message
+          );
+          showErrorBox({
+            message:
+              i18n._(
+                t`Unable to update the game slug. A slug must be 6 to 30 characters long and only contains letters, digits or dashes.`
+              ) +
+              ' ' +
+              i18n._(t`Verify your internet connection or try again later.`),
```
AlexandreSi, line 191: It's the same as above, can you see an elegant way to deduplicate it? I'm not sure there's one, but maybe you can see something

newIDE/app/src/Export/GenericExporters/OnlineWebExport/OnlineGameLink.js
ClementPasteau: yep! will do

newIDE/app/src/Export/GenericExporters/OnlineWebExport/OnlineGameLink.js
ClementPasteau: Hmm not sure there's one that doesn't complexify it more 😞

newIDE/app/src/Export/GenericExporters/OnlineWebExport/OnlineGamePropertiesDialog.js
```diff
@@ -92,6 +87,11 @@ export const OnlineGamePropertiesDialog = ({
         orientation: orientation || 'default',
       })
     ) {
+      // If the project has been modified, then save it.
```
ClementPasteau, line 90: This fixes the bug where the project was saved before the modifications were applied. So it was useless........

newIDE/app/src/Export/GenericExporters/OnlineWebExport/OnlineGamePropertiesDialog.js
```diff
@@ -153,6 +153,7 @@ export const OnlineGamePropertiesDialog = ({
         setDiscoverable={setDiscoverable}
         displayThumbnail
         thumbnailUrl={thumbnailUrl}
+        disabled={isLoading}
```
ClementPasteau, line 156: Disabled the whole form while the game is submitting, for a better UX (otherwise only the button disables)

4ian (approved):

PR #3872, opened 2022-05-05
Improve points and collision masks editors

AlexandreSi:
## Points editor

- Allow user to erase a coordinate whole value
- Highlight points on the preview
  - for the selected point in the list
  - for the hovered point in the list
- Change Z index of selected and hovered points in the preview
  - to prevent Origin point to cover newly added point for instance
- Take all available space in row

demo_points_editor

## Collision masks editor

- Allow user to erase a coordinate whole value
- Add possibility to reorder collision masks points

demo_collision_mask_editor

## Other

- Remove react warnings

## Yet to do

- Fix the behavior that transforms `1.5` in `1.500004572158` that prevents user to fine-tune a value

### DIFF
--- a/newIDE/app/src/ObjectEditor/Editors/SpriteEditor/CollisionMasksEditor/PolygonsList.js
+++ b/newIDE/app/src/ObjectEditor/Editors/SpriteEditor/CollisionMasksEditor/PolygonsList.js
@@ -1,5 +1,9 @@
 // @flow
 import * as React from 'react';
+import { Trans, t } from '@lingui/macro';
+import DeleteIcon from '@material-ui/icons/Delete';
+import WarningIcon from '@material-ui/icons/Warning';
+import AddIcon from '@material-ui/icons/Add';
 import {
   Table,
   TableBody,
@@ -8,7 +12,6 @@ import {
   TableRow,
   TableRowColumn,
 } from '../../../../UI/Table';
-import { SortableContainer, SortableElement } from 'react-sortable-hoc';
 import { mapVector } from '../../../../Utils/MapFor';
 import styles from './styles';
 import VerticeRow from './VerticeRow';
@@ -19,19 +22,13 @@ import {
 } from '../../../../UI/Accordion';
 import Text from '../../../../UI/Text';
 import IconButton from '../../../../UI/IconButton';
-import DeleteIcon from '@material-ui/icons/Delete';
-import WarningIcon from '@material-ui/icons/Warning';
-import AddIcon from '@material-ui/icons/Add';
 import FlatButton from '../../../../UI/FlatButton';
-import { Trans, t } from '@lingui/macro';
 import { Column, Line, Spacer } from '../../../../UI/Grid';
 import RaisedButtonWithSplitMenu from '../../../../UI/RaisedButtonWithSplitMenu';
 import AlertMessage from '../../../../UI/AlertMessage';
 import GDevelopThemeContext from '../../../../UI/Theme/ThemeContext';
 import ScrollView from '../../../../UI/ScrollView';
 const gd = global.gd;
-
-const SortableVerticeRow = SortableElement(VerticeRow);
 
 type VerticesTableProps = {|
   vertices: gdVectorVector2f,
@@ -45,29 +42,32 @@ type VerticesTableProps = {|
 |};
 
 const VerticesTable = (props: VerticesTableProps) => {
-  const updateVerticeX = (vertice, newValue) => {
+  const draggedVerticeIndex = React.useRef<?number>(null);
+
+  const updateVerticeX = (vertice: gdVector2f, newValue: number) => {
     // Ensure vertice stays inside the sprite bounding box.
     vertice.set_x(Math.min(props.spriteWidth, Math.max(newValue, 0)));
     props.onUpdated();
   };
 
-  const updateVerticeY = (vertice, newValue) => {
+  const updateVerticeY = (vertice: gdVector2f, newValue: number) => {
     // Ensure vertice stays inside the sprite bounding box.
     vertice.set_y(Math.min(props.spriteHeight, Math.max(newValue, 0)));
     props.onUpdated();
   };
 
+  const dropVertice = (oldIndex: number, newIndex: number) => {
+    if (oldIndex === newIndex) return;
+    gd.moveVector2fInVector(
+      props.vertices,
+      oldIndex,
+      newIndex > oldIndex ? newIndex - 1 : newIndex
+    );
+    props.onUpdated();
+  };
+
   return (
     <Column expand>
-      {props.hasWarning && (
-        <AlertMessage kind="warning">
-          <Trans>
-            The polygon is not convex. Ensure it is, otherwise the collision
-            mask won't work.
-          </Trans>
-        </AlertMessage>
-      )}
-      <Spacer />
       <Table>
         <TableHeader>
           <TableRow>
@@ -82,17 +82,25 @@ const VerticesTable = (props: VerticesTableProps) => {
           </TableRow>
         </TableHeader>
         <TableBody>
-          {mapVector(props.vertices, (vertice, j) => (
-            <SortableVerticeRow
-              index={j}
-              key={j}
-              disabled
+          {mapVector(props.vertices, (vertice, verticeIndex) => (
+            <VerticeRow
+              key={vertice.ptr}
+              parentVerticeId={props.vertices.ptr.toString()}
+              setDragged={() => {
+                draggedVerticeIndex.current = verticeIndex;
+              }}
+              drop={() => {
+                const { current } = draggedVerticeIndex;
+                if (!current && current !== 0) return;
+                dropVertice(current, verticeIndex);
+                draggedVerticeIndex.current = null;
+              }}
               verticeX={vertice.get_x()}
               verticeY={vertice.get_y()}
               onChangeVerticeX={newValue => updateVerticeX(vertice, newValue)}
               onChangeVerticeY={newValue => updateVerticeY(vertice, newValue)}
               onRemove={() => {
-                gd.removeFromVectorVector2f(props.vertices, j);
+                gd.removeFromVectorVector2f(props.vertices, verticeIndex);
                 props.onUpdated();
               }}
               canRemove={props.vertices.size() > 3}
@@ -100,6 +108,15 @@ const VerticesTable = (props: VerticesTableProps) => {
           ))}
         </TableBody>
       </Table>
+      <Spacer />
+      {props.hasWarning && (
+        <AlertMessage kind="warning">
+          <Trans>
+            The polygon is not convex. Ensure it is, otherwise the collision
+            mask won't work.
+          </Trans>
+        </AlertMessage>
+      )}
       <Line justifyContent="center">
         <FlatButton
           icon={<AddIcon size="small" />}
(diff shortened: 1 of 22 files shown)


Bouh:
> Highlight points on the preview
> * for the selected point in the list
> * for the hovered point in the list

- Do you think you can add these highlights for the collision mask too? All points of the collision mask are Red.
- And as you can see you scrolled on the preview because it's a tiny sprite, you think the sprite can fit better to the preview to avoid us zoom-in?

Bouh:
> > Highlight points on the preview
> > 
> > * for the selected point in the list
> > * for the hovered point in the list
> 
> * Do you think you can add these highlights for the collision mask too? All points of the collision mask are Red.
> * And as you can see you scrolled on the preview because it's a tiny sprite, you think the sprite can fit better to the preview to avoid us zoom-in?


> Yet to do
> Fix the behavior that transforms 1.5 in 1.500004572158 that prevents user to fine-tune a value

So cool because, there is a new post related to it today: https://forum.gdevelop.io/t/feature-collision-integers-as-an-option/39313

AlexandreSi:
> * Do you think you can add these highlights for the collision mask too? All points of the collision mask are Red.

I could, with a different color

> * And as you can see you scrolled on the preview because it's a tiny sprite, you think the sprite can fit better to the preview to avoid us zoom-in?

I'll see what I can do, not sure that it will be easy to compute a zoom factor based on the sprite size

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/CollisionMasksEditor/PolygonsList.js
```diff
@@ -19,20 +22,14 @@ import {
...
 import FlatButton from '../../../../UI/FlatButton';
-import { Trans, t } from '@lingui/macro';
 import { Column, Line, Spacer } from '../../../../UI/Grid';
 import RaisedButtonWithSplitMenu from '../../../../UI/RaisedButtonWithSplitMenu';
 import AlertMessage from '../../../../UI/AlertMessage';
 import GDevelopThemeContext from '../../../../UI/Theme/ThemeContext';
 import ScrollView from '../../../../UI/ScrollView';
 const gd = global.gd;
 
-const SortableVerticeRow = SortableElement(VerticeRow);
```
4ian, line 34: Good riddance! :)

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/CollisionMasksEditor/VerticeRow.js
```diff
@@ -1,66 +1,134 @@
...
+          canDrag={() => true}
+          canDrop={() => true}
+          drop={() => {
+            props.drop();
+          }}
+        >
+          {({ connectDragSource, connectDropTarget, isOver, canDrop }) =>
+            connectDropTarget(
+              <tr
+                className="MuiTableRow-root"
```
4ian, line 70: Can you explain the reason for this? It's seems a bit risky because there is no guarantee MuiTableRow-root is 1) injected in the CSS already (if we've not used a table in another place, we won't have it) and 2) won't change at the next update.

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/CollisionMasksEditor/VerticeRow.js
```diff
@@ -1,66 +1,134 @@
...
+                      setVerticeXInputValue(value);
+                      const valueAsNumber = parseFloat(value);
+                      if (!isNaN(valueAsNumber)) {
+                        props.onChangeVerticeX(valueAsNumber);
+                      }
+                    }}
+                  />
+                </TableRowColumn>
+                <TableRowColumn style={styles.coordinateColumn}>
+                  <TextField
```
4ian, line 100: Maybe a SemiControlledTextField would have worked? Give it a look, it can maybe avoid you to have these 2 states.

In theory, it is made to solve the issue you had: it's is "semi controlled" in the sense that the value is coming from the value prop BUT when the user change the value, the user can enter anything.

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/PointsEditor/PointRow.js
```diff
@@ -12,90 +12,146 @@ import Text from '../../../../UI/Text';
...
+                inputStyle={
+                  props.selected
+                    ? { color: muiTheme.listItem.selectedTextColor }
+                    : undefined
+                }
+                value={pointXInputValue}
+                type="number"
+                id="point-x"
+                onChange={(e, value) => {
+                  setPointXInputValue(value);
```
4ian, line 100: Same, consider a SemiControlledTextField to allow anything to be input and have the onChange called at each keystroke, but the value entered won't be replaced until the component is blurred.

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/PointsEditor/PointsPreview.js
```diff
@@ -64,7 +101,14 @@ const PointsPreview = (props: Props) => {
 
   const onEndDragPoint = () => {
     const draggingWasDone = !!state.draggedPoint;
-    if (draggingWasDone) props.onPointsUpdated();
+    if (draggingWasDone) {
+      props.onPointsUpdated();
+      if (state.draggedPointKind && state.draggedPoint) {
```
4ian, line 106: Can you add a comment explaining why this is done?

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/PointsEditor/styles.js
```diff
@@ -1,15 +1,11 @@
-//TODO: Factor with styles.js from LayersList.
+// @flow
```
4ian, line 1: 🙏

4ian (approved):
LGTM if you try SemiControlledTextField to avoid storing extra states.

tristanbob:
> > * And as you can see you scrolled on the preview because it's a tiny sprite, you think the sprite can fit better to the preview to avoid us zoom-in?
> 
> I'll see what I can do, not sure that it will be easy to compute a zoom factor based on the sprite size

The zoom frustrates me as well.  Selecting a smarter initial zoom (so that image takes up most of visible area) is the preferred fix.  

A less preferred option might to make a slider that controls zoom.  Thoughts?

AlexandreSi:
I managed to set the zoom "correctly". Do you think the zoom range is still needed? It feels to me because for a small sprite that makes the zoom maxxed, one could think the zoom is broken when trying to zoom even more (same for large sprites)

tristanbob:
> I managed to set the zoom "correctly". Do you think the zoom range is still needed? It feels to me because for a small sprite that makes the zoom maxxed, one could think the zoom is broken when trying to zoom even more (same for large sprites)

That is great! No need to add a slider, the existing controls are great for fine tuning.

AlexandreSi:
I added the slider so that the user can understand they maxxed out the zoom.

image

Bouh:
The new interface can be [tested here.](http://gdevelop-storybook.s3-website-us-east-1.amazonaws.com/improve-points-editor/latest/index.html?path=/story/spriteeditor-and-related-editors--spriteeditor)

First I have to said this link to Storybook is awesome!
Here my last wishes:
- [ ] Have a selectable line in collision mask, like in point editor.
- [ ] Image preview isn't centered and don't fit well, maybe it is just in Storybook.

The new slider is nice it will help users on touch devices to understand we can zoom.

AlexandreSi:
I couldn't make so that the selected point have priority when at the same place as another point. It requires more work (z index is not working in SVGs) and it does not make the tool unusable.

newIDE/app/src/UI/Slider.js
```diff
@@ -0,0 +1,24 @@
+// @flow
```
4ian, line 1: Thank you for this abstraction!

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/PointsEditor/PointRow.js
```diff
@@ -6,96 +6,144 @@ import IconButton from '../../../../UI/IconButton';
...
+    <ThemeConsumer>
+      {muiTheme => (
+        <TableRow
+          style={{
+            backgroundColor: props.selected
+              ? muiTheme.listItem.selectedBackgroundColor
+              : muiTheme.list.itemsBackgroundColor,
+          }}
+          onClick={() => props.onClick(props.pointName)}
+          onMouseEnter={() => props.onMouseEnter(props.pointName)}
```
4ian, line 52: I wonder if we should prefer onPointerEnter to handle other things than a mouse

Bouh:
We will gain in convenience with this is improvements thank you 💯

tristanbob:
> First I have to say this link to Storybook is awesome! Here my last wishes:

Wow, yes the storybook is very useful for testing!

> * [x]  Have a selectable line in collision mask, like in point editor.

I think he is saying to use the box outline to highlight points on the collision mask when the mouse hovers over the table on the right.  I agree that this would be much preferred to the current "turn point yellow" 

> The new slider is nice it will help users on touch devices to understand we can zoom.

I agree, this slider does work great.  :)  Thanks @AlexandreSi!

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/PointsEditor/PointRow.js
```diff
@@ -6,96 +6,144 @@ import IconButton from '../../../../UI/IconButton';
...
+                }
+                value={roundTo(pointY, POINT_COORDINATE_PRECISION).toString()}
+                type="number"
+                id="point-y"
+                onChange={value => {
+                  const valueAsNumber = parseFloat(value);
+                  if (!isNaN(valueAsNumber))
+                    props.onChangePointY(valueAsNumber);
+                }}
+                onBlur={event => {
```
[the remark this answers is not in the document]
AlexandreSi: Without the onBlur, a user can leave the input empty, having then a quantic state: will it be 0? will it be something else? Well, it will be the last valid value, if you were erasing `5,643`, you will have 5 as a value. Adding this onBlur, I set empty values to 0 so that it's clear to the user what it will be

Silver-Streak:
These changes look amazing. I have a big list of things I wrote up when I was first working on my adventure game template, and this covers most of them! Fantastic work.

The only things immediately missing are:
1. Concave hull slicing into multiple convex shapes. (Complicated and likely needs a separate story. I'll write up a request... eventually 😆 )
2. The ability to double click on a vertice to add a new point right at the double click. This is pretty common when I was testing other engines for adventure game template ideas.

No worries if these aren't feasible here, but wanted to get them out of my brain now wrather than sitting in my notes for more months.

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/PointsEditor/PointRow.js
AlexandreSi: Yes, good idea! I didn't think of that

AlexandreSi:
> I think he is saying to use the box outline to highlight points on the collision mask when the mouse hovers over the table on the right. I agree that this would be much preferred to the current "turn point yellow"

@tristanbob Thanks for the suggestion, I was not happy with this yellow. Instead, circles are outlined. Much better I think.

> No worries if these aren't feasible here, but wanted to get them out of my brain now wrather than sitting in my notes for more months.

@Silver-Streak  Those are 2 whole new features I think. I'll get these features live and will have your suggestions turned into a discussion so that we don't lose them!

AlexandreSi:
When I think that I "only" wanted to allow users to erase a coordinate value instead of having the default 0!

oupsi

PR #3874, opened 2022-05-05
Fix history not working well for Inline Event change

ClementPasteau:
Screen Recording 2022-05-05 at 17 40 27

### DIFF
--- a/newIDE/app/src/EventsSheet/index.js
+++ b/newIDE/app/src/EventsSheet/index.js
@@ -154,7 +154,6 @@ type State = {|
   inlineEditing: boolean,
   inlineEditingAnchorEl: ?HTMLElement,
   inlineInstructionEditorAnchorEl: ?HTMLElement,
-  inlineEditingChangesMade: boolean,
   inlineEditingPreviousValue: ?string,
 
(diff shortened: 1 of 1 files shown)


newIDE/app/src/EventsSheet/index.js
```diff
@@ -805,24 +810,32 @@ export class EventsSheetComponentWithoutHandle extends React.Component<
...
   };
 
   closeParameterEditor = (shouldCancel: boolean) => {
-    if (shouldCancel) {
-      const { instruction, parameterIndex } = this.state.editedParameter;
-      if (!instruction || !this.state.inlineEditingPreviousValue) return;
-
+    const { instruction, parameterIndex } = this.state.editedParameter;
+    if (!instruction) {
+      return;
```
4ian, line 820: Exceptionally, can we avoid a early return an do a:

```
if (instruction) {
  if (shouldCancel && ...)
  if (!shouldCancel && ...)
}

// and then the rest:
const { inlineEditingAnchorEl } = this.state;
this.setState(...)
```

I'm a bit paranoid but I want to be sure that if, for some reason, instruction is null, we still close the editor.

4ian (approved):
Great stuff!

PR #3885, opened 2022-05-09
Fix web-app recent files wrongly containing URLs to examples

4ian:
Tested on the web-app: examples are not added to the recent files list (and so not re-opened, unless you saved them on Google Drive).


ClementPasteau (approved):
LGTM!

PR #3889, opened 2022-05-11
Allow to choose installation directory in the Windows installer

4ian:
Following discussion ([here, in French](https://twitter.com/Florianrival/status/1524319855339622400)), and more generally feedback from confused users about where GDevelop is installed, this should:
- Show the path before installation - and allow to change it.
- Allow to install machine wide.

### DIFF
--- a/newIDE/electron-app/package.json
+++ b/newIDE/electron-app/package.json
@@ -44,6 +44,10 @@
       "executableName": "GDevelop",
       "publish": null
     },
+    "nsis": {
+      "oneClick": false,
+      "allowToChangeInstallationDirectory": true
+    },
     "appx": {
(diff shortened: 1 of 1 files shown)


4ian:
Testing build with the updated installer that asks if you want a machine wide or per-user installation, and allow you to choose the installation folder (Program Files will be used by default for machine wide): https://gdevelop-releases.s3.amazonaws.com/experimental-build/nsis-assisted-installer/latest/GDevelop%205%20Setup%205.0.133-beta2.exe

Side note, according to: https://nsis.sourceforge.io/Docs/Chapter4.html#silent, these switches can be used for a silent installation in a given directory:
` GDevelop 5 Setup[...].exe /S /D=C:\Program Files\GDevelop 5`

PR #3895, opened 2022-05-12
Disable autocomplete for Text Input objects and add 5px padding left

Bouh:
Close https://github.com/GDevelopApp/GDevelop-games-platform/issues/88

+ Add 5 px padding on left because this depend of the browser, and the text is too much stuck on left for now.

### DIFF
--- a/Extensions/TextInput/textinputruntimeobject-pixi-renderer.ts
+++ b/Extensions/TextInput/textinputruntimeobject-pixi-renderer.ts
@@ -47,6 +47,8 @@ namespace gdjs {
       const isTextArea = this._object.getInputType() === 'text area';
       this._input = document.createElement(isTextArea ? 'textarea' : 'input');
       this._input.style.border = '1px solid black';
+      this._input.autocomplete = 'off';
+      this._input.style.paddingLeft = '5px';
       this._input.style.borderRadius = '0px';
       this._input.style.backgroundColor = 'white';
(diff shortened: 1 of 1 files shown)


4ian:
I'm worried about paddingLeft, this seems a quick but dirty solution. It should be at least visible in the scene editor, possibly it mus be configurable, and there is an investigation to do to see how it interact with a large border.

Extensions/TextInput/textinputruntimeobject-pixi-renderer.ts
```diff
@@ -47,6 +47,8 @@ namespace gdjs {
       const isTextArea = this._object.getInputType() === 'text area';
       this._input = document.createElement(isTextArea ? 'textarea' : 'input');
       this._input.style.border = '1px solid black';
+      this._input.autocomplete = 'off';
+      this._input.style.paddingLeft = '5px';
```
Bouh, line 51:
```suggestion
```

Bouh (commented):
Right, so I remove the padding left line.

ClementPasteau:
I've just rebased and tested the whole thing to make sure it was fine. It looks alright to me.
The only thing is that we're disabling autocomplete for all types of inputs, I'm wondering if we should keep it in some cases?

I can't think of a case where we'd like it to be enabled.

image

ClementPasteau (approved):

PR #3896, opened 2022-05-12
Rework asset store

ClementPasteau:
As part of revamping the Asset Store, the first step is to redo the homepage of the asset store, and suggest to creators packs of assets they can import and use easily.

In this PR:

- Moved all stores stories to their own files with a better structure to test easily
- Fixed a bug where the ExtensionStore would show a loader
- Retrieve the Asset Packs from the backend services
- Show the list on the Asset Store homepage, with a folded filters panel
- Redirect to the previous Asset listing on pack selection, filtered on this pack + open the panel

In the future PRs we will probably:

- Rework the Asset Pack Screen to be independent from the listing, with a better view of the assets it contains + previews + allow adding all of them at once
- Rework the filters to be actual filters (size, color, animations, etc...)
- Rework the Asset screen to be more exhaustive about what the asset contains

Few screenshots in different sizes and themes:

image
image
image
image

### DIFF
--- a/newIDE/app/src/AssetStore/AssetsHome.js
+++ b/newIDE/app/src/AssetStore/AssetsHome.js
@@ -3,7 +3,13 @@ import * as React from 'react';
 import { CorsAwareImage } from '../UI/CorsAwareImage';
 import Text from '../UI/Text';
 import { type AssetPacks } from '../Utils/GDevelopServices/Asset';
-import { GridListTile, GridList, Paper } from '@material-ui/core';
+import {
+  GridListTile,
+  GridList,
+  Paper,
+  makeStyles,
+  createStyles,
+} from '@material-ui/core';
 import { shouldValidate } from '../UI/KeyboardShortcuts/InteractionKeys';
 import { Line, Column } from '../UI/Grid';
 import ScrollView from '../UI/ScrollView';
@@ -32,6 +38,18 @@ const styles = {
   },
 };
 
+const useStylesForGridListItem = makeStyles(theme =>
+  createStyles({
+    root: {
+      '&:focus': {
+        border: `2px solid ${theme.palette.primary.main}`,
+        outline: 'none',
+      },
+      '&:focus-visible': { outline: 'unset' },
+    },
+  })
+);
+
 type Props = {|
   assetPacks: AssetPacks,
   onPackSelection: string => void,
@@ -41,6 +59,7 @@ export const AssetsHome = ({
   assetPacks: { starterPacks },
   onPackSelection,
 }: Props) => {
+  const classesForGridListItem = useStylesForGridListItem();
   return (
     <ThemeConsumer>
       {muiTheme => (
@@ -53,6 +72,7 @@ export const AssetsHome = ({
           >
             {starterPacks.map((pack, index) => (
               <GridListTile
+                classes={classesForGridListItem}
                 key={pack.tag}
                 tabIndex={0}
                 onKeyPress={(
--- a/newIDE/app/src/AssetStore/AssetsHome.js
+++ b/newIDE/app/src/AssetStore/AssetsHome.js
@@ -16,6 +16,7 @@ import { Line, Column } from '../UI/Grid';
 import ScrollView from '../UI/ScrollView';
 import ThemeConsumer from '../UI/Theme/ThemeConsumer';
 import { useResponsiveWindowWidth } from '../UI/Reponsive/ResponsiveWindowMeasurer';
+import ThemeContext from '../UI/Theme/ThemeContext';
 
 const columns = 3;
 const columnsForSmallWindow = 2;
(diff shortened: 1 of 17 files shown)


4ian:
Great start! 
Comments in no particular order:
- The focus ring/selection ring is a bit cut for the first line. We might want to do the styling by ourselves. Not a deal breaker
- No feedback when pressing/clicking, let's add something.
- The discover is a Text, but in this design it seems it's used like a title, but it does not look like one. We should at least go for a size="something-that-is-supposed-to-be-a-title", as we can't use the default text styling for something that is not a text. Might be something to check with the design.
- As the filters are not making really a sense for these packs, should we hide them from this first page?

ClementPasteau:
> The focus ring/selection ring is a bit cut for the first line. We might want to do the styling by ourselves. Not a deal breaker

Fixed!
> No feedback when pressing/clicking, let's add something.

Fixed!

> The discover is a Text, but in this design it seems it's used like a title, but it does not look like one. We should at least go for a size="something-that-is-supposed-to-be-a-title", as we can't use the default text styling for something that is not a text. Might be something to check with the design.

Fixed!

> As the filters are not making really a sense for these packs, should we hide them from this first page?

For the moment, because those filters are still "Categories", I thought keeping them until the proper filters are in place, and we can then get rid of it.

AlexandreSi:
Looks great!

I will have a quick look on how to better display the cards on mobile, the title and the number of assets that are truncated don't look that good.

newIDE/app/src/AssetStore/AssetsHome.js
```diff
@@ -0,0 +1,122 @@
...
+|};
+
+export const AssetsHome = ({
+  assetPacks: { starterPacks },
+  onPackSelection,
+}: Props) => {
+  const classesForGridListItem = useStylesForGridListItem();
+  return (
+    <ThemeConsumer>
+      {muiTheme => (
```
AlexandreSi, line 65: Better use `useContext(ThemeContext)`

newIDE/app/src/AssetStore/AssetsHome.js
```diff
@@ -0,0 +1,122 @@
...
+                    <Line justifyContent="space-between" noMargin>
+                      <Text style={styles.packTitle} size="body2">
+                        {pack.name}
+                      </Text>
+                      <Text
+                        style={styles.packTitle}
+                        color="primary"
+                        size="body2"
+                      >
+                        {pack.assetsCount} Assets
```
AlexandreSi, line 110: To translate

newIDE/app/src/AssetStore/index.js
```diff
@@ -71,13 +81,33 @@ export const AssetStore = ({
...
             onRequestSearch={() => {}}
             style={styles.searchBar}
             ref={searchBar}
             id="asset-store-search-bar"
           />
+          {!isOnHomePage && <Spacer />}
+          <Line justifyContent="left" noMargin={isOnHomePage}>
+            {isOnHomePage ? (
+              <Column>
+                <Text size="title">Discover</Text>
```
AlexandreSi, line 96: To translate

newIDE/app/src/AssetStore/AssetsHome.js
```diff
@@ -107,7 +112,7 @@ export const AssetsHome = ({
                         color="primary"
                         size="body2"
                       >
-                        {pack.assetsCount} Assets
+                        {pack.assetsCount} <Trans>Assets</Trans>
```
AlexandreSi: You can write `<Trans>{pack.assetsCount} Assets</Trans>`. It can give more context to translators

AlexandreSi (approved):
One last change and it seems ok to me!

newIDE/app/src/AssetStore/index.js
```diff
@@ -87,41 +120,77 @@ export const AssetStore = ({
...
+                    <IconButton onClick={() => setIsFiltersOpen(false)}>
+                      <DoubleChevronArrow />
+                    </IconButton>
+                  </Line>
+                  <>
+                    <CategoryChooser
+                      allItemsLabel={<Trans>All assets</Trans>}
+                      allFilters={filters}
+                      filtersState={filtersState}
+                      setIsOnHomePage={setIsOnHomePage}
```
4ian, line 150: setIsOnHomePage seems not a proper name for this prop. Take the point of view of CategoryChooser: 
it does not even know what is a home page. A home page of what?
It just know about its own filters/categories that you give to him. But has no idea when you would want to change if you're on a home page or not - that's not his responsibility! :) 

Instead, what you, as the parent (AssetStore) really want to know is: 
tell me when a tag is clicked and I will react accordingly (here, by... setting we're not on the home page anymore, which is not the business of the child)

Something like:
```suggestion
                      onChoiceChanged={() => setIsOnHomePage(false)}
```

newIDE/app/src/UI/Search/CategoryChooser.js
```diff
@@ -68,12 +70,15 @@ const MemoizedTagsTree = React.memo<MemoizedTagsTreeProps>(function TagsTree({
           : ''
       }
       defaultExpanded={[]}
-      onNodeSelect={() => {}}
+      onNodeSelect={() => setIsOnHomePage && setIsOnHomePage(false)}
```
4ian, line 73: A way to see something is wrong: setIsOnHomePage is called always with false.
This means the prop is not fully used and so is not representing what it's really doing.
It seems here it's just calling it when a label is chosen?

4ian:
Quick note to make clear a behavior: @D8H, you updated the behavior of the filters/search so that they go automatically to the existing result search page. I think it make sense. Just flagging this for @ClementPasteau.
This is just a first rework so as long as it make sense it's good.

newIDE/app/src/UI/Search/CategoryChooser.js
D8H: Yes, I though about it too and kept it like this because it will be removed in a few day.
Do you think it must be done?

4ian: Well I'm always worried of "this is just temporary and I will remove it for sure in two days" 😇
But if you think that's a detail, then no big deal.

D8H: No, you're right it might take some time before the old filters can be removed. I made the change.

PR #3909, opened 2022-05-15
Allow event extensions to define a condition and an action with an operator based on an expression

D8H:
This is built over:
- https://github.com/4ian/GDevelop/pull/4437
- https://github.com/4ian/GDevelop/pull/4436

It allows:
- to write conditions and actions with operators in event-based extensions
- to use refined type for automatic operands in built-in or event-based extensions

I added a test to check that expression and condition of an ExpressionAndCondition function are renamed.
I tried to add a test where an expression is inside its own condition, but this can't be done with the current suite because it seems that the event function metadata are not initialized by the test.

(The story is not really useful because it needs Core to be rebuilt)
http://gdevelop-storybook.s3-website-us-east-1.amazonaws.com/ExpressionAndCondition/latest/index.html?path=/story/eventsfunctionsextensioneditor-eventsfunctionconfigurationeditor--default-behavior-function

image

image

image

I chose to forbid to add parameters on the action because I don't see a usecase for it. It can be added another time if necessary.
image

## Example
I manually tested some rename cases:
* Project: https://www.dropbox.com/s/rs5aa0w5d4ts4s2/ExpressionAndConditionTest.zip?dl=1
* Build: https://liluo.io/games/c6b2bba7-f208-413e-9ae3-ad01851cb627?dev=true

image

image


4ian:
Nice! This is corresponding to this card: https://trello.com/c/zwo0T1E7/479-add-in-events-function-a-condition-expression-type-that-would-allow-to-generate-both-an-expression-and-a-condition-to-check-it

> there is probably some work to do an event refactorers.

I guess this would work by launching the expression refactorer first, then the condition?

Core/GDCore/IDE/WholeProjectRefactorer.cpp
```diff
@@ -393,8 +384,18 @@ void WholeProjectRefactorer::RenameBehaviorEventsFunction(
   const gd::EventsFunction& eventsFunction =
       eventsFunctions.GetEventsFunction(oldFunctionName);
 
-  if (eventsFunction.GetFunctionType() == gd::EventsFunction::Action ||
-      eventsFunction.GetFunctionType() == gd::EventsFunction::Condition) {
+  // The order is important
```
D8H, line 387: I'm not exactly sure why. So I didn't give details, but with the other way around, expressions inside their own condition weren't renamed.

newIDE/app/src/EventsFunctionsExtensionsLoader/MetadataDeclarationHelpers.js
```diff
@@ -590,18 +649,23 @@ export const declareEventsFunctionParameters = (
...
+    );
+  } else if (functionType === gd.EventsFunction.StringExpressionAndCondition) {
+    ((instructionOrExpression: any): gdMultipleInstructionMetadata).useStandardParameters(
+      'string'
+    );
+  }
+
   // By convention, latest parameter is always the eventsFunctionContext of the calling function
   // (if any).
   instructionOrExpression.addCodeOnlyParameter('eventsFunctionContext', '');
```
D8H, line 671: The operator and operand parameters are put at the end to avoid to mess withe the user defined parameter indexes that are used in sentense like "_PARAM3_".
The issue with this is that when a user add a new parameter, it breaks the compatibility. This should not be a huge issue as most ExpressionAndCondition won't have any extra parameters.

Core/GDCore/IDE/WholeProjectRefactorer.cpp
4ian: Ah I just saw this comment. Ok then!

D8H:
I only cover the feature for the function renamer.  Do you think other refactors could be important to cover too?

4ian:
Nice thanks for the other tests :) 
Are we now covering:
- Extension renaming
- Function renaming
- Parameter move

for the new ExpressionAndCondition ? Maybe we should have this both for an extension and for a behavior? Not sure about this but because we have a "The order is important" and we don't really know why, tests might save us of a future regression here :)

D8H:
It seems that moving parameters on expression never worked because the scene or object and behavior parameters where not taking int account.
Should I extract the fix in an other PR?

D8H:
The `ExpressionAndCondition` now has the same cover as `Expression` and `Condition`. 

Free function
- Extension renaming
- Function renaming
- Parameter move

Behavior function
- Extension renaming
- Function renaming
- Parameter move
- Behavior renaming

I also updated the project example to check it manually.

D8H:
There is still one issue that remains: adding new parameters make the operator and operand parameters to shift as they are at the end. We should do a refactorer for this.

Adding new parameters to an `ExpressionAndCondition` will still break extensions compatibility when they are updated in the repository. We could try to add the 2 parameters at the beginning, but it seems hard to do. What is your view on this?

4ian:
I'm not sure but intuitively it would be better to have these operator and operands parameter at a fixed place, at the beginning just after the "context" parameters, to avoid them being moved (so that you can add a new parameter, and the extension will continue to work when upgraded as long as you handle this parameter as being "optional").
Otherwise, it will be a pain (and even if we write a refactorer, it's... not a great design if we had the operator/operand at the end?).

I.e:

- First: Current Scene OR Object, then Behavior
- Then: Operator
- Then: Operand
- Then: the rest (because we can't know which of these parameters are an "identifier" and which are just additional parameters
- Then: the automatically insert eventsFunctionContext

At least that's what I would say intuitively?

> For instance, the timer condition put the identifier first but the camera condition put the layer after which is not horrible either.

Can you give the examples you're mentioning?

D8H:
I have to check that it works correctly with `Choices`, `Color` or other types.

Core/GDCore/Project/EventsFunction.h
```diff
@@ -154,7 +226,20 @@ class GD_CORE_API EventsFunction {
   gd::EventsList& GetEvents() { return events; };
 
   /**
-   * \brief Return the parameters of the function.
+   * \brief Return the parameters of the function that are used in the events.
+   *
+   * \note During code/extension generation, new parameters are added
```
4ian, line 231: Thanks for adding this note

Core/GDCore/Project/EventsFunctionsContainer.h
```diff
@@ -139,6 +150,9 @@ class GD_CORE_API EventsFunctionsContainer
   void Init(const gd::EventsFunctionsContainer& other) {
     return SerializableWithNameList<gd::EventsFunction>::Init(other);
   };
+
+private:
+FunctionSource source;
```
4ian, line 155: Indent this line

Core/GDCore/Project/EventsFunctionsContainer.h
```diff
@@ -25,7 +25,18 @@ namespace gd {
...
- public:
+public:
+  enum FunctionSource {
+      Extension,
+      Object,
+      Behavior};
+
+  EventsFunctionsContainer(FunctionSource source_) : source(source_) {}
+
+  FunctionSource GetSource() const {
```
4ian, line 36: Let's add a comment why this is needed.

Core/GDCore/Project/EventsFunctionsExtension.h
```diff
@@ -20,7 +20,9 @@ class Project;
 }  // namespace gd
 
 namespace gd {
-
+// TODO Remove the EventsFunctionsContainer extension and make it an attribute.
```
4ian, line 23: I think you mean the inheritance?
```suggestion
// TODO Remove the EventsFunctionsContainer inheritance and make it an attribute.
```

GDJS/GDJS/Events/CodeGeneration/EventsCodeGenerator.cpp
```diff
@@ -116,14 +116,15 @@ gd::String EventsCodeGenerator::GenerateLayoutCode(
 
 gd::String EventsCodeGenerator::GenerateEventsFunctionCode(
     gd::Project& project,
+    const gd::EventsFunctionsContainer functionContainer,
```
4ian, line 119: Nit: 
```suggestion
    const gd::EventsFunctionsContainer functionsContainer,
```

GDJS/GDJS/Events/CodeGeneration/EventsCodeGenerator.h
```diff
@@ -65,6 +67,7 @@ class EventsCodeGenerator : public gd::EventsCodeGenerator {
    */
   static gd::String GenerateEventsFunctionCode(
       gd::Project& project,
+      const gd::EventsFunctionsContainer functionContainer,
```
4ian, line 70: Pass it by reference?
```suggestion
      const gd::EventsFunctionsContainer& functionsContainer,
```

Otherwise we'll pay for a copy every time and this can quickly adds up in terms of performance.

GDevelop.js/__tests__/GDJS.js
```diff
@@ -479,6 +479,7 @@ describe('libGD.js - GDJS related tests', function () {
         new gd.EventsFunctionsExtensionCodeGenerator(project);
       const code =
         eventsFunctionsExtensionCodeGenerator.generateFreeEventsFunctionCompleteCode(
+          new gd.EventsFunctionsExtension(),
```
4ian, line 482: This will leak, not a huge deal but prefer to declare and then delete it.

newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
```diff
@@ -292,29 +293,48 @@ export default class EventsFunctionParametersEditor extends React.Component<
...
+    const typeShownFirstIndex = firstParameterIndex;
+    const labelShownFirstIndex =
+      firstParameterIndex +
+      (eventsFunction.getFunctionType() === gd.EventsFunction.ActionWithOperator
+        ? 1
+        : 0);
+    const isParameterTypeShown = index => {
       // The first two parameters of a behavior method should not be changed at all,
       // so we even hide their description and type to avoid cluttering the interface.
       // Same thing for an object which has mandatory Object parameter.
```
4ian, line 324: Move this comment before the declaration of "typeShownFirstIndex" (because that's where the logic is now)

newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionParametersEditor.js
```diff
@@ -292,29 +293,48 @@ export default class EventsFunctionParametersEditor extends React.Component<
...
-        (!eventsBasedBehavior && !eventsBasedObject) ||
-        (!!eventsBasedBehavior && index >= 2) ||
-        (!!eventsBasedObject && index >= 1)
-      );
+      return index >= typeShownFirstIndex;
+    };
+    const isParameterDescriptionShown = index => {
+      // The first two parameters of a behavior method should not be changed at all,
+      // so we even hide their description and type to avoid cluttering the interface.
+      // Same thing for an object which has mandatory Object parameter.
```
4ian, line 330: Same, move to the line just before `const labelShownFirstIndex`

newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionPropertiesEditor.js
```diff
@@ -213,13 +241,28 @@ export default class EventsFunctionPropertiesEditor extends React.Component<
...
+                      ? gd.EventsFunction.Expression
+                      : type === gd.EventsFunction.StringExpressionAndCondition
+                      ? gd.EventsFunction.ExpressionAndCondition
+                      : type
+                  }
                   floatingLabelText={<Trans>Function type</Trans>}
                   fullWidth
                   disabled={!!freezeEventsFunctionType}
-                  onChange={(e, i, value: string) => {
+                  onChange={(e, i, valueSting: string) => {
```
4ian, line 254:
```suggestion
                  onChange={(e, i, valueString: string) => {
```

newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionPropertiesEditor.js
```diff
@@ -213,13 +241,28 @@ export default class EventsFunctionPropertiesEditor extends React.Component<
             <ResponsiveLineStackLayout alignItems="center" noMargin>
               <Line alignItems="center" noMargin>
                 <SelectField
-                  value={type}
+                  value={
+                    type === gd.EventsFunction.StringExpression
```
4ian, line 245: Add a comment explaining why you're doing this.

newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionPropertiesEditor.js
```diff
@@ -237,69 +280,223 @@ export default class EventsFunctionPropertiesEditor extends React.Component<
...
+                type === gd.EventsFunction.StringExpressionAndCondition ? (
+                <SemiControlledTextField
+                  commitOnBlur
+                  floatingLabelText={
+                    eventsBasedBehavior &&
+                    (type === gd.EventsFunction.ExpressionAndCondition ||
+                      type ===
+                        gd.EventsFunction.StringExpressionAndCondition) ? (
+                      <Trans>
+                        Sentence in Events Sheet (automatically suffixed by "of
```
4ian, line 493: That could even be:
```suggestion
                        Name in Events Sheet (automatically suffixed by "of
```

4ian: (because it's a name, not really a sentence anymore)

newIDE/app/src/EventsFunctionsExtensionEditor/index.js
```diff
@@ -189,9 +190,10 @@ export default class EventsFunctionsExtensionEditor extends React.Component<
...
         this._objectsContainer
       );
-    } else {
+    } else if (eventsFunctionsExtension) {
       gd.EventsFunctionTools.freeEventsFunctionToObjectsContainer(
         project,
+        eventsFunctionsExtension,
         eventsFunction,
         this._globalObjectsContainer,
         this._objectsContainer
```
4ian, line 199: Add a `else throw new Error("No extension, behavior or object was specified when loading a function")`

newIDE/app/src/EventsFunctionsExtensionEditor/index.js
```diff
@@ -1029,6 +1040,11 @@ export default class EventsFunctionsExtensionEditor extends React.Component<
                     eventsFunction={selectedEventsFunction}
                     eventsBasedBehavior={selectedEventsBasedBehavior}
                     eventsBasedObject={selectedEventsBasedObject}
+                    eventsFunctionsContainer={
+                      (selectedEventsBasedEntity &&
```
4ian, line 1044: That's where a common interface would make the code cleaner, right?

newIDE/app/src/EventsFunctionsExtensionsLoader/MetadataDeclarationHelpers.js
```diff
@@ -301,6 +302,12 @@ export const isExtensionLifecycleEventsFunction = (functionName: string) => {
   );
 };
 
+export const removeTailingDot = (description: string): string => {
```
4ian, line 305:
```suggestion
export const removeTrailingDot = (description: string): string => {
```

I think

newIDE/app/src/EventsFunctionsExtensionsLoader/MetadataDeclarationHelpers.js
```diff
@@ -350,6 +422,34 @@ export const declareInstructionOrExpressionMetadata = (
   }
 };
 
+const shiftSentenceParamIndexes = (
```
4ian, line 425: Let's make a quick unit test for this ;)

newIDE/app/src/EventsFunctionsExtensionsLoader/MetadataDeclarationHelpers.js
```diff
@@ -910,37 +1140,69 @@ export const declareObjectPropertiesInstructionAndExpressions = (
...
+  const parameters = (functionType === gd.EventsFunction.ActionWithOperator &&
+  getterFunction
+    ? getterFunction
+    : eventsFunction
+  ).getParameters();
+
+  mapVector(parameters, addParameter, 0, userDefinedFirstParameterIndex);
+
+  if (functionType === gd.EventsFunction.ExpressionAndCondition) {
+    ((instructionOrExpression: any): gdMultipleInstructionMetadata).useStandardParameters(
```
4ian, line 1188: Do you think we should have been naming this `useStandardOperatorParameters`?

newIDE/app/src/EventsFunctionsExtensionsLoader/MetadataDeclarationHelpers.js
```diff
@@ -910,37 +1140,69 @@ export const declareObjectPropertiesInstructionAndExpressions = (
...
+    ((instructionOrExpression: any): gdInstructionMetadata).useStandardOperatorParameters(
+      getterFunction &&
+        getterFunction.getFunctionType() ===
+          gd.EventsFunction.StringExpressionAndCondition
+        ? 'string'
+        : 'number'
+    );
+  }
+
+  mapVector(parameters, addParameter, userDefinedFirstParameterIndex);
```
4ian, line 1205: I'm a bit unease at this additional parameter which is "not standard". Could we instead remove it and add, in the  `addParameter`, a check on the index?

4ian: mapVector, like map in JS, pass the index as the second parameter to the callback.

newIDE/app/src/EventsFunctionsExtensionEditor/EventsFunctionConfigurationEditor/EventsFunctionPropertiesEditor.js
D8H: Yes, there is probably a better phrasing but I want to keep "sentence" because it's like a technical term for GDevelop and I think it's more important that users to understand where it's used rather than what should be filled.

newIDE/app/src/EventsFunctionsExtensionEditor/index.js
D8H: Indeed

newIDE/app/src/EventsFunctionsExtensionsLoader/MetadataDeclarationHelpers.js
D8H: Probably, I can create a card for it.

newIDE/app/src/EventsFunctionsExtensionsLoader/MetadataDeclarationHelpers.js
D8H: I could create a local helper method that would filter the indexes between `mapVector` and `addParameter`.

Core/GDCore/Extensions/Metadata/InstructionMetadata.cpp
```diff
@@ -118,6 +120,8 @@ InstructionMetadata& InstructionMetadata::UseStandardOperatorParameters(
     }
   } else {
     AddParameter("operator", _("Modification's sign"), type);
+    // The type "string" is not forced because it's declined in several subtype
+    // (see ParameterMetadata).
     AddParameter(type == "number" ? "expression" : type, _("Value"));
```
D8H, line 125: I'm thinking to use this to handle types like `color`. I'm not sure this is intended though.
`ParameterMetadata::IsExpression` could be used here to make it more explicit and robust. Unless, it should only be "string" but it seems fine to do the change.

Core/GDCore/Extensions/Metadata/ValueTypeMetadata.h
```diff
@@ -0,0 +1,227 @@
...
+    }
+    return false;
+  }
+
+  /**
+   * \brief Return the expression type from the parameter type.
+   * Declinations of "number" and "string" types (like "forceMultiplier" or
+   * "sceneName") are replaced by "number" and "string".
+   */
+  static const gd::String &GetExpressionValueType(const gd::String &parameterType);
```
D8H, line 199: This and...

D8H:
This is how the function properties look for an expression. It's not ideal but it's not worst than having the expression type buried in the the function type where users are ask about it at the creation and could suppose there is no more choices.

I don't want to spend time to find a better design when it will be redesigned from scratch soon. This reuse what already exists and is still an improvement.

image

Core/GDCore/Project/EventsFunction.h
```diff
@@ -115,20 +117,81 @@ class GD_CORE_API EventsFunction {
...
+   */
+  EventsFunction& SetExpressionType(const gd::ValueTypeMetadata& type) {
+    expressionType = type;
+    return *this;
+  }
+
+  /**
+   * \brief Get the type of the expression
+   */
+  const gd::ValueTypeMetadata& GetExpressionType() const { return expressionType; }
```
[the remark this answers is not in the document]
4ian: Maybe GetPrimitiveValueType?

Core/GDCore/Project/EventsFunctionsContainer.h
```diff
@@ -25,7 +25,24 @@ namespace gd {
...
+  EventsFunctionsContainer(FunctionSource source_) : source(source_) {}
+
+  /**
+   * \brief Get the source of the function container.
+   * 
+   * \note For instance, it can be useful to handle specific parameters for
+   * behaviors.
+   */
+  FunctionSource GetSource() const {
+    return source;
```
4ian, line 43: Nit: Could this be a "Owner" rather than a Source?

4ian (approved):
I think I reviewed everything! Looks good generally speaking 👍

4ian:
This will be huge for extension creators

Core/GDCore/Extensions/Metadata/ParameterMetadata.h
```diff
@@ -156,21 +172,20 @@ class GD_CORE_API ParameterMetadata {
...
    * \see gd::ParameterMetadata::GetType
+   * \deprecated Use gd::ValueTypeMetadata instead.
    */
   static bool IsObject(const gd::String &parameterType) {
-    return parameterType == "object" || parameterType == "objectPtr" ||
-           parameterType == "objectList" ||
-           parameterType == "objectListOrEmptyIfJustDeclared" ||
-           parameterType == "objectListOrEmptyWithoutPicking";
+    return gd::ValueTypeMetadata::TypeIsObject(parameterType);
   }
```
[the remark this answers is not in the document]
D8H: Doesn't the "Is" prefix only work for object methods? It feels strange in a static context because there is no object as subject.

Core/GDCore/Project/EventsFunction.h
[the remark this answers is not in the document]
D8H: Sounds good

Core/GDCore/Extensions/Metadata/ParameterMetadata.h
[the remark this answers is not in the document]
4ian: I think this convention can work for any function, even static ones, because it's still properly english ("is XXX something?" -> yes it is/no it's not -> it's definitely a boolean that you return) (while "XXX is something" -> is it an affirmation or a question?). 

At least that's what the folks choose for React (`shouldComponentUpdate` and not `componentShouldUpdate`).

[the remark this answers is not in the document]
D8H: ok, I'll do the change

4ian (approved):

PR #3910, opened 2022-05-16
Allow user to limit score values

fannieyan:
set extreme score

### DIFF
--- a/newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardAppearanceDialog.js
+++ b/newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardAppearanceDialog.js
@@ -197,6 +197,7 @@ function LeaderboardAppearanceDialog({
           }}
           secondaryActions={[
             <HelpButton
+              key="help"
               helpPagePath="/interface/games-dashboard/leaderboard-administration"
               anchor="change_the_appearance_of_a_leaderboard"
             />,
--- a/newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardSortOptionsDialog.js
+++ b/newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardSortOptionsDialog.js
@@ -20,9 +20,6 @@ import Checkbox from '../../UI/Checkbox';
 import { FormHelperText } from '@material-ui/core';
 import { MarkdownText } from '../../UI/MarkdownText';
 
-const isWholeNumber = (value: any): boolean =>
-  value === null || Number.isInteger(value);
-
 type SortOptions = {|
   sort: LeaderboardSortOption,
   extremeAllowedScore: ?number,
@@ -30,7 +27,7 @@ type SortOptions = {|
 
 type Props = {
   open: boolean,
-  sort?: LeaderboardSortOption,
+  sort: LeaderboardSortOption,
   extremeAllowedScore?: number,
   onSave: SortOptions => Promise<void>,
   onClose: () => void,
@@ -82,10 +79,10 @@ function LeaderboardSortOptionsDialog({
         return;
       }
     }
-    if (!isWholeNumber(extremeAllowedScoreValue)) {
+    if (isNaN(extremeAllowedScoreValue)) {
       setExtremeAllowedScoreError(
         i18n._(
-          t`The extreme score must be a whole number between ${extremeAllowedScoreMin} and ${extremeAllowedScoreMax}.`
+          t`Limit cannot be empty, uncheck or fill a value between ${extremeAllowedScoreMin} and ${extremeAllowedScoreMax}.`
         )
       );
       return;
@@ -168,9 +165,9 @@ function LeaderboardSortOptionsDialog({
                   <FormHelperText style={{ display: 'inline' }}>
                     <MarkdownText
                       source={i18n._(
-                        `Any submitted score that is ${
-                          sortOrder === 'ASC' ? 'lower' : 'higher'
-                        } than the set value will not be saved in the leaderboard.`
+                        sortOrder === 'ASC'
+                          ? `Any submitted score that is lower than the set value will not be saved in the leaderboard.`
+                          : `Any submitted score that is higher than the set value will not be saved in the leaderboard.`
                       )}
                     />
                   </FormHelperText>
@@ -192,9 +189,11 @@ function LeaderboardSortOptionsDialog({
                   fullWidth
                   type="number"
                   floatingLabelText={
-                    <Trans>
-                      {sortOrder === 'ASC' ? 'Minimum' : 'Maximum'} score
-                    </Trans>
+                    sortOrder === 'ASC' ? (
+                      <Trans>Minimum score</Trans>
+                    ) : (
+                      <Trans>Maximum score</Trans>
+                    )
                   }
                   value={extremeAllowedScoreValue}
                   errorText={extremeAllowedScoreError}
@@ -205,7 +204,7 @@ function LeaderboardSortOptionsDialog({
                       setExtremeAllowedScoreError(null);
                     }
 
-                    setExtremeAllowedScoreValue(parseInt(newValue));
+                    setExtremeAllowedScoreValue(parseFloat(newValue));
                   }}
                 />
               </Line>
--- a/newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardSortOptionsDialog.js
+++ b/newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardSortOptionsDialog.js
@@ -1,0 +1,220 @@
+// @flow
+import React from 'react';
+import { t, Trans } from '@lingui/macro';
+import { I18n } from '@lingui/react';
+import { type I18n as I18nType } from '@lingui/core';
+
+import Dialog from '../../UI/Dialog';
+import FlatButton from '../../UI/FlatButton';
+import { ColumnStackLayout } from '../../UI/Layout';
+import RaisedButton from '../../UI/RaisedButton';
+import SelectField from '../../UI/SelectField';
+import SelectOption from '../../UI/SelectOption';
+import Text from '../../UI/Text';
+import TextField from '../../UI/TextField';
+
+import { type LeaderboardSortOption } from '../../Utils/GDevelopServices/Play';
+import { Column, LargeSpacer, Line } from '../../UI/Grid';
+import HelpButton from '../../UI/HelpButton';
+import Checkbox from '../../UI/Checkbox';
+import { FormHelperText } from '@material-ui/core';
+import { MarkdownText } from '../../UI/MarkdownText';
+
+const isWholeNumber = (value: any): boolean =>
+  value === null || Number.isInteger(value);
+
+type SortOptions = {|
+  sort: LeaderboardSortOption,
+  extremeAllowedScore: ?number,
+|};
+
+type Props = {
+  open: boolean,
+  sort?: LeaderboardSortOption,
+  extremeAllowedScore?: number,
+  onSave: SortOptions => Promise<void>,
+  onClose: () => void,
+};
+
+const extremeAllowedScoreMax = Number.MAX_SAFE_INTEGER;
+const extremeAllowedScoreMin = Number.MIN_SAFE_INTEGER;
+
+function LeaderboardSortOptionsDialog({
+  open,
+  onClose,
+  onSave,
+  sort,
+  extremeAllowedScore,
+}: Props) {
+  const [isLoading, setIsLoading] = React.useState<boolean>(false);
+
+  const [
+    extremeAllowedScoreError,
+    setExtremeAllowedScoreError,
+  ] = React.useState<?string>(null);
+  const [
+    displayExtremeAllowedScoreInput,
+    setDisplayExtremeAllowedScoreInput,
+  ] = React.useState<boolean>(extremeAllowedScore !== undefined);
+  const [
+    extremeAllowedScoreValue,
+    setExtremeAllowedScoreValue,
+  ] = React.useState<number>(0);
+
+  const [sortOrder, setSortOrder] = React.useState<LeaderboardSortOption>(
+    sort || 'ASC'
+  );
+
+  const onSaveSettings = async (i18n: I18nType) => {
+    if (displayExtremeAllowedScoreInput) {
+      if (extremeAllowedScoreValue > extremeAllowedScoreMax) {
+        setExtremeAllowedScoreError(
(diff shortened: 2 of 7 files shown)


newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardAppearanceDialog.js
```diff
@@ -197,6 +197,7 @@ function LeaderboardAppearanceDialog({
           }}
           secondaryActions={[
             <HelpButton
+              key="help"
```
AlexandreSi, line 200: 💜

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
```diff
@@ -977,6 +1008,30 @@ export const LeaderboardAdmin = ({ onLoading, project }: Props) => {
...
+              |}) => {
+                try {
+                  await onUpdateLeaderboard(i18n, {
+                    ...sortOptions,
+                  });
+                } finally {
+                  setIsEditingSortOptions(false);
+                }
+              }}
+              sort={currentLeaderboard ? currentLeaderboard.sort : undefined}
```
AlexandreSi, line 1027: You can put this check with `{isEditingSortOptions ?` because it is non-sense otherwise

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
```diff
@@ -977,6 +1008,30 @@ export const LeaderboardAdmin = ({ onLoading, project }: Props) => {
...
+              onClose={() => setIsEditingSortOptions(false)}
+              onSave={async (sortOptions: {|
+                sort: LeaderboardSortOption,
+                extremeAllowedScore: ?number,
+              |}) => {
+                try {
+                  await onUpdateLeaderboard(i18n, {
+                    ...sortOptions,
+                  });
+                } finally {
```
AlexandreSi, line 1023: 👍

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
```diff
@@ -483,6 +489,25 @@ export const LeaderboardAdmin = ({ onLoading, project }: Props) => {
       </Line>
     );
 
+  const getExtremeAllowedScoreText = (currentLeaderboard: Leaderboard) => {
```
AlexandreSi, line 492: I think you can either:
- Have this function outside the component because it does not use anything that belongs to the component (neither the props nor the hooks)
- Have this function take no input so that it uses current leaderboard from the component, and it will itself do the checks of current leaderboard definition that you're currently doing below

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
```diff
@@ -483,6 +489,25 @@ export const LeaderboardAdmin = ({ onLoading, project }: Props) => {
...
+    const maxOrMin = currentLeaderboard.sort === 'ASC' ? 'min' : 'max';
```
AlexandreSi, line 493: We should translate those `min` and `max`

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
```diff
@@ -483,6 +489,25 @@ export const LeaderboardAdmin = ({ onLoading, project }: Props) => {
...
+    if (
+      currentLeaderboard.extremeAllowedScore === null ||
+      currentLeaderboard.extremeAllowedScore === undefined
+    ) {
+      return '';
+    }
+    let customizedScoreText = currentLeaderboard.extremeAllowedScore;
+    if (currentLeaderboard.customizationSettings) {
+      customizedScoreText = formatScore(
+        // $FlowIgnore
```
AlexandreSi, line 503: Can you explain in the comment why this is safe to ignore?

newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardSortOptionsDialog.js
```diff
@@ -0,0 +1,220 @@
...
+import TextField from '../../UI/TextField';
+
+import { type LeaderboardSortOption } from '../../Utils/GDevelopServices/Play';
+import { Column, LargeSpacer, Line } from '../../UI/Grid';
+import HelpButton from '../../UI/HelpButton';
+import Checkbox from '../../UI/Checkbox';
+import { FormHelperText } from '@material-ui/core';
+import { MarkdownText } from '../../UI/MarkdownText';
+
+const isWholeNumber = (value: any): boolean =>
```
AlexandreSi, line 23: Can you reuse the function for leaderboardAppearanceDialog if they have the same name, or do they have a different logic?

newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardSortOptionsDialog.js
```diff
@@ -0,0 +1,220 @@
...
+  value === null || Number.isInteger(value);
+
+type SortOptions = {|
+  sort: LeaderboardSortOption,
+  extremeAllowedScore: ?number,
+|};
+
+type Props = {
+  open: boolean,
+  sort?: LeaderboardSortOption,
```
AlexandreSi, line 33: I think this is mandatory

newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardSortOptionsDialog.js
```diff
@@ -0,0 +1,220 @@
...
+      if (extremeAllowedScoreValue < extremeAllowedScoreMin) {
+        setExtremeAllowedScoreError(
+          i18n._(
+            t`Extreme score must be equal or higher than ${extremeAllowedScoreMin}.`
+          )
+        );
+        return;
+      }
+    }
+    if (!isWholeNumber(extremeAllowedScoreValue)) {
```
AlexandreSi, line 85: I wonder if we should have this constraint of a whole number: one could want to allow a max score of 2.7

newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardSortOptionsDialog.js
```diff
@@ -0,0 +1,220 @@
...
+      setExtremeAllowedScoreError(
+        i18n._(
+          t`The extreme score must be a whole number between ${extremeAllowedScoreMin} and ${extremeAllowedScoreMax}.`
+        )
+      );
+      return;
+    }
+    setIsLoading(true);
```
AlexandreSi, line 93: It is never set to false, should it be somewhere? I am not sure since the parent closes the dialog if the api call fails

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
fannieyan: Actually, we don't need to ignore 😅

newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardSortOptionsDialog.js
```diff
@@ -0,0 +1,220 @@
...
+            <Checkbox
+              label={
+                <>
+                  <Line noMargin>
+                    <Trans>Limit scores</Trans>
+                  </Line>
+                  <FormHelperText style={{ display: 'inline' }}>
+                    <MarkdownText
+                      source={i18n._(
+                        `Any submitted score that is ${
```
AlexandreSi, line 171: Translation will be funky: `Un score soumis qui est lower que 5 ne sera pas sauvegardé`.
Prefer having the choice between 2 whole translations

newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardSortOptionsDialog.js
```diff
@@ -0,0 +1,220 @@
...
+          </ColumnStackLayout>
+          {displayExtremeAllowedScoreInput && (
+            <Column noMargin>
+              <Line>
+                <LargeSpacer />
+                <TextField
+                  fullWidth
+                  type="number"
+                  floatingLabelText={
+                    <Trans>
```
AlexandreSi, line 195: In this case, you have nothing to translate. You should use:
```js
{sortOrder === 'ASC' ? <Trans>Minimum score</Trans> : <Trans>Maximum score</Trans>}
```

newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardSortOptionsDialog.js
fannieyan: Indeed, and we can actually only check if isNaN function returns true (extremeAllowedScore is reset if we uncheck the checkbox so the input should never be null, always a number)

newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardSortOptionsDialog.js
AlexandreSi: Otherwise, translators will receive the string `{0}` to translate I think

fannieyan: Good to know!

newIDE/app/src/stories/componentStories/LeaderboardSortOptionsDialog.stories.js
```diff
@@ -0,0 +1,23 @@
+// @flow
```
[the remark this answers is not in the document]
4ian: Stories 😍

newIDE/app/src/stories/componentStories/LeaderboardSortOptionsDialog.stories.js
```diff
@@ -0,0 +1,23 @@
...
+export default {
+  title: 'LeaderboardSortOptionsDialog ',
+  component: LeaderboardSortOptionsDialog,
+  decorators: [paperDecorator, muiDecorator],
+};
+
+export const Default = () => (
+  <LeaderboardSortOptionsDialog
+    open
+    onClose={() => action('onClose')()}
```
4ian, line 18:
```suggestion
    onClose={action('onClose')}
```

is not working?

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
```diff
@@ -150,12 +152,38 @@ const getApiError = (payload: LeaderboardUpdatePayload): ApiError => ({
...
+    return '';
+  }
+  let customizedScoreText = currentLeaderboard.extremeAllowedScore;
+  if (currentLeaderboard.customizationSettings) {
+    customizedScoreText = formatScore(
+      currentLeaderboard.extremeAllowedScore,
+      currentLeaderboard.customizationSettings.scoreFormatting
+    );
+  }
+  return ` (${maxOrMin}: ${customizedScoreText})`;
```
4ian, line 174: Some languages may not use this construction "`(max: 150)`" but instead write thing like `(150: max)` or even not use colon/parenthesis.

So it's better to put a `Trans` or `i18n._(t...)` on a larger sentence, like the 4 possibilities:

Higher is better
Higher is better (max : X)
Lower is better
Higher is better (min : X)

newIDE/app/src/GameDashboard/LeaderboardAdmin/LeaderboardSortOptionsDialog.js
```diff
@@ -0,0 +1,219 @@
...
+          <Text size="title">
+            <Trans>Sort order</Trans>
+          </Text>
+          <ColumnStackLayout>
+            <SelectField
+              fullWidth
+              value={sortOrder}
+              floatingLabelText={<Trans>Scores sort order</Trans>}
+              onChange={(e, i, newValue) => {
+                // $FlowIgnore
```
4ian, line 144: Add a quick explanation to ease the work for the future reader: // $FlowFixMe - newValue is blabla

Also FlowFixMe is the only officially supported name.

newIDE/app/src/stories/componentStories/LeaderboardSortOptionsDialog.stories.js
AlexandreSi: Sometimes it is necessary to do that to prevent JS to enter an infinite loop

4ian: True, I wonder if it's not a max depth to set to Storybook to avoid it trying to log recursive stuff (weird that they don't handle this properly).

Would be good to spend 5 minutes investigating, otherwise good to go!

fannieyan: Not sure why, but it makes the story freeze/crash when clicking on Close button

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
```diff
@@ -977,6 +991,34 @@ export const LeaderboardAdmin = ({ onLoading, project }: Props) => {
...
+                try {
+                  await onUpdateLeaderboard(i18n, {
+                    ...sortOptions,
+                  });
+                } finally {
+                  setIsEditingSortOptions(false);
+                }
+              }}
+              sort={
+                isEditingSortOptions && currentLeaderboard
```
AlexandreSi, line 1011: Sorry I was not clear 😅 
I meant that the dialog should not be open if there is no current leaderboard. So you could have:

```js
{isEditingSortOptions && currentLeaderboard ? (
            <LeaderboardSortOptionsDialog ... />) : null}
```
This way, you can avoid the checks on `currentLeaderboard` definition in the props, and giving the default `ASC` value that is not "true"

fannieyan: oh yes alright !

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
```diff
@@ -633,16 +651,12 @@ export const LeaderboardAdmin = ({ onLoading, project }: Props) => {
...
-            });
-          }}
-          tooltip={t`Change sort direction`}
+          onClick={() => setIsEditingSortOptions(true)}
+          tooltip={t`Edit`}
           edge="end"
           disabled={isRequestPending || isEditingName}
         >
-          <SwapVertical />
+          <Brush />
```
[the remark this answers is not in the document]
AlexandreSi: I think your point is valid as well, let's keep the Brush icon

PR #3912, opened 2022-05-16
Fix border on div that represents the frame of an image

### LINKED ISSUE #3838: Slight offset in the hitbox editor
There is a slight offset of the displayed sprite and the white rectangle, in the collision masks editor.
If you look at the left and top sides, the sprite shows beyond the rectangle, while in the right and lower sides, there is a clear separation between the sprite and the rectangle.
image
GD5v131

In the points editor, only the right and lower sides seem to be misaligned.
image

AlexandreSi:
Fixes #3838

Before

image

After

image


4ian (approved):
🤩

PR #3913, opened 2022-05-16
Fix scale of volume on fade action

### LINKED ISSUE #3907: Wrong range value for audio/sound fading event.
The range for fade the volume of a sound played on a channel is set to 0-100 according to description but actual test results show it's 0 - 1

AlexandreSi:
Fixes #3907


4ian:
> It could generate "pseudo"-regressions but this action was created not that long ago so I feel like it's okay to release this with a warning: "If you used this action, please change your scale from 0-1 to 0-100"

Yes, totally.

PR #3915, opened 2022-05-16
[Asset Store] Add animation, viewpoint, dimensions, object type and license filters

D8H:
[Test in live in Storybook](http://gdevelop-storybook.s3-website-us-east-1.amazonaws.com/asset-store/tag-filter/latest/index.html?path=/story/assetstore-newobjectdialog--default)

I'm undecided about class names. I would like to avoid confusion between classes from the view and the ones from the model.


newIDE/app/src/AssetStore/AssetStoreFilterPanel.js
```diff
@@ -0,0 +1,309 @@
...
+              value={range}
+              min={min}
+              max={max}
+              step={step}
+              scale={scale}
+              marks={true}
+              valueLabelDisplay="auto"
+              onChange={(event, newValue) => setRange(newValue)}
+            />
+          </AccordionBody>
```
D8H, line 171: I'm not sure what is the bet way to add some margin on both sides of the slider.

Bouh:
- [ ] Replace categories by **Filter** ? image

newIDE/app/src/AssetStore/AssetStoreFilterPanel.js
```diff
@@ -0,0 +1,359 @@
...
+        <Accordion key={filterKey} defaultExpanded>
+          <AccordionHeader>
+            <Text displayInlineAsSpan>{title}</Text>
+          </AccordionHeader>
+          <AccordionBody>
+            <ColumnStackLayout>
+              {choices.map(tag => (
+                <InlineCheckbox
+                  key={tag.value}
+                  label={tag.label ? i18n._(tag.label) : tag.value}
```
ClementPasteau, line 58: can tags not have a label?

newIDE/app/src/AssetStore/AssetStoreFilterPanel.js
ClementPasteau: I'd suggest adding custom style to the slider, the same way we've done for other material ui components.
(search in the code `makeStyles()`)
I would add some padding left / right to the `root` https://v4.mui.com/api/slider/#css)

newIDE/app/src/AssetStore/AssetStoreFilterPanel.js
```diff
@@ -0,0 +1,359 @@
...
+        <Accordion key={filterKey} defaultExpanded>
+          <AccordionHeader>
+            <Text displayInlineAsSpan>{title}</Text>
+          </AccordionHeader>
+          <AccordionBody>
+            <HexColorField
+              disableAlpha
+              fullWidth
+              color={color}
+              onChange={color => {
```
ClementPasteau, line 206: this can probably just be `onChange=setColor`

newIDE/app/src/AssetStore/AssetStoreSearchFilter.js
```diff
@@ -0,0 +1,168 @@
...
+  }
+}
+
+/**
+ * Modulo operation (the remainder after dividing one number by another)
+ * @param x Dividend value.
+ * @param y Divisor value.
+ * @returns Return the remainder for the values.
+ */
+const mod = function(x: number, y: number): number {
```
ClementPasteau, line 121: Why can't you use the remainder operation `%` ?

newIDE/app/src/AssetStore/HexColorField.js
```diff
@@ -0,0 +1,115 @@
...
+
+type State = {|
+  color: string,
+|};
+
+const hexToNullableRGBColor = (color: string): RGBColor | null => {
+  return /^#{0,1}[0-9a-fA-F]{6}$/.test(color) ? hexToRGBColor(color) : null;
+};
+
+export class HexColorField extends React.Component<Props, State> {
```
ClementPasteau, line 47: Why not use a functional component? (especially for something we just created)

newIDE/app/src/AssetStore/HexColorField.js
```diff
@@ -0,0 +1,115 @@
...
+          floatingLabelFixed
+          helperMarkdownText={this.props.helperMarkdownText}
+          type="text"
+          hintText={t`Code like #ff8844`}
+          value={this.state.color}
+          onChange={event => this._handleChange(event.target.value)}
+          onBlur={this._handleBlur}
+          ref={textField => (this._textField = textField)}
+          onKeyPress={ev => {
+            console.log(`Pressed keyCode ${ev.key}`);
```
ClementPasteau, line 95: log here

newIDE/app/src/UI/EmptyPlaceholder.js
```diff
@@ -52,7 +53,15 @@ export const EmptyPlaceholder = (props: Props) => (
             primary
             onClick={props.onAdd}
             disabled={!!props.isLoading}
-            icon={props.isLoading ? <CircularProgress size={24} /> : <Add />}
+            icon={
```
ClementPasteau, line 56: Can we also refactor the name of the action to be `onAction` rather than `onAdd`?
Then the action is either "add" by default, or something else if another icon is provided

newIDE/app/src/UI/Search/UseSearchItem.js
```diff
@@ -4,18 +4,39 @@ import { type ChosenCategory } from './FiltersChooser';
...
   searchItems: ?Array<SearchItem>,
   chosenCategory: ?ChosenCategory,
-  chosenFilters: Set<string>
+  chosenFilters: Set<string>,
+  searchFilters?: Array<SearchFilter<SearchItem>>
 ): ?Array<SearchItem> => {
   if (!searchItems) return null;
 
   const startTime = performance.now();
+  // TODO do only one call to filter for efficiency.
```
ClementPasteau, line 39: is this to avoid multiple calls when selecting multiple filters quickly?

newIDE/app/src/AssetStore/AssetStoreContext.js
```diff
@@ -77,6 +115,73 @@ export const AssetStoreStateProvider = ({
...
+  // for the memo.
+  const currentFilters = [
+    animatedFilter,
+    viewpointFilter,
+    dimensionFilter,
+    objectTypeFilter,
+    colorFilter,
+    licenseFilter,
+  ];
+  const [searchFilters, setSearchFilters] = React.useState<
```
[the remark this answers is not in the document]
D8H: The setter is called in AssetStoreFilterPanel when the user change a criteria value.

newIDE/app/src/AssetStore/AssetStoreSearchFilter.js
D8H: I also prefer the formula with the double `%` but it's not a lot more simple and this method was used in the engine so I kept the same implementation.

newIDE/app/src/AssetStore/HexColorField.js
D8H: It's a copy of an existing component and I wasn't sure it would be kept.

newIDE/app/src/AssetStore/AssetStoreContext.js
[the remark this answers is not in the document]
4ian: Calling a setState function (`setSearchFilters`) seems weird though.

Ensuring that `currentFilters` stays stable across render if the items inside have not changed is a great thing :) 
But the way to do it should be to do a `useMemo` (which looks a bit weird to be honest):

```
const currentFilters = React.useMemo(() => [
     animatedFilter,
    viewpointFilter,
    dimensionFilter,
    objectTypeFilter,
    colorFilter,
    licenseFilter,
    ], [
     animatedFilter,
    viewpointFilter,
    dimensionFilter,
    objectTypeFilter,
    colorFilter,
    licenseFilter,
    ])
```

which will properly return always the same array (so you get the stability you need) if dependencies (the second array) are always the same.

newIDE/app/src/AssetStore/AssetStoreContext.js
```diff
@@ -77,6 +115,73 @@ export const AssetStoreStateProvider = ({
...
+  ] = React.useState<ColorAssetStoreSearchFilter>(
+    new ColorAssetStoreSearchFilter()
+  );
+  const [
+    licenseFilter,
+    setLicenseFilter,
+  ] = React.useState<LicenseAssetStoreSearchFilter>(
+    new LicenseAssetStoreSearchFilter()
+  );
+  const assetFiltersState = {
```
4ian, line 154: This object will be a new one at each render, do you need to `useMemo` on it? 
Otherwise it's passed to `const assetStoreState = React.useMemo(` and will make it change at each render even if the filters did not change?

newIDE/app/src/AssetStore/AssetStoreContext.js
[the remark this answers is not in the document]
4ian: EDIT: I may have not grasped entirely the problem, but this state does not looks like a state and so we should avoid it, I think useMemo is better.

newIDE/app/src/AssetStore/HexColorField.js
ClementPasteau: I feel like we should keep it, like you mentioned, the usage is different, so we may end up modifying it in the future without affecting the other (and vice versa)

newIDE/app/src/UI/EmptyPlaceholder.js
ClementPasteau: weird, maybe just because VSCode is lost, it's only used 2 or 3 times across the app though, so the refactor could be done manually

newIDE/app/src/AssetStore/AssetStoreFilterPanel.js
```diff
@@ -0,0 +1,355 @@
...
+}: MultipleChoiceFilterProps) => {
+  return (
+    <I18n>
+      {({ i18n }) => (
+        <Accordion key={filterKey} defaultExpanded>
+          <AccordionHeader>
+            <Text displayInlineAsSpan>{title}</Text>
+          </AccordionHeader>
+          <AccordionBody>
+            <ColumnStackLayout>
```
4ian, line 52: Can we remove margins for this (`noMargin`), or is the AccordionBody not adding margins and so we need it?

newIDE/app/src/UI/Search/UseSearchItem.js
ClementPasteau: oh right, I misread the line for "filter for efficiency" instead of "to .filter() for efficiency" :)

newIDE/app/src/AssetStore/AssetStoreFilterPanel.js
4ian: I would recommend to go for a 
`<Column><Line noMargin><Slider ...></Line></Column>`

4ian: A slider by default, like any component, takes all the width available, which is good. It's the parent job to add layouting with margins (i.e: Line, Column)

ClementPasteau (commented):
Great work overall 🌮 
lgtm as it's quite maintainable and easy to read.
A few things to cleanup, maybe in the next PR
I'll let @4ian review it too

newIDE/app/src/AssetStore/AssetStoreSearchFilter.js
```diff
@@ -0,0 +1,168 @@
...
+  isSatisfiedBy(searchItem: AssetShortHeader): boolean {
+    if (!this.color) return true;
+    const targetHsl = rgbToHsl(this.color.r, this.color.g, this.color.b);
+    const dominantRgb = hexNumberToRGBColor(searchItem.dominantColors[0]);
+    const dominantHsl = rgbToHsl(dominantRgb.r, dominantRgb.g, dominantRgb.b);
+    const targetSaturation = targetHsl[1];
+    const dominantSaturation = dominantHsl[1];
+    let score = 0;
+    if (targetSaturation === 0) {
+      // Hue is not relevent.
```
4ian, line 142:
```suggestion
      // Hue is not relevant.
```

newIDE/app/src/AssetStore/HexColorField.js
4ian: Do you think a prop like `preferHexFormat={true}` would work to reuse ColorField?

newIDE/app/src/UI/Search/UseSearchItem.js
```diff
@@ -4,18 +4,39 @@ import { type ChosenCategory } from './FiltersChooser';
 import shuffle from 'lodash/shuffle';
 import SearchApi from 'js-worker-search';
 
+export interface SearchFilter<SearchItem> {
+  isSatisfiedBy(searchItem: SearchItem): boolean;
+}
+
+export class TagSearchFilter<SearchItem: { tags: Array<string> }>
```
4ian, line 11: Is this class really needed? It seems to only have one subclass?

D8H: It's to keep the logic about filtering on tags in this file. It might help to refactor when the old filters will be dropped or merged with the new filters.

newIDE/app/src/AssetStore/AssetStoreFilterPanel.js
D8H: It looks like AccordionBody not adding margins. Without it, the content is not aligned with the title.

newIDE/app/src/AssetStore/HexColorField.js
D8H: There are already a lot of color conversions, it might make it hard to understand.

newIDE/app/src/AssetStore/AssetStoreSearchFilter.js
D8H: `-14 % 8 = -6` and `-14 mod % 8 = 2`, but I guess Clement meant a formula like `(x % y + x) % y` to avoid a call to floor.
It's true that it can't be negative but I think a mod make it more explicit.
uh, I copied pasted a documentation that is a bit confusing.

newIDE/app/src/AssetStore/AssetStoreContext.js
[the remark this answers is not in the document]
D8H: I made a change. It seems cleaner but I'm not sure if there is something else to change.

Initially, I did it by analogy to `useFilters()`, so it probably have the same issue. It create a new instance each call.

newIDE/app/src/AssetStore/AssetStoreFilterPanel.js
4ian: Ok 👍

newIDE/app/src/AssetStore/HexColorField.js
```diff
@@ -0,0 +1,116 @@
...
+  floatingLabelText,
+  helperMarkdownText,
+  onChange,
+  color,
+}: Props) => {
+  const [colorString, setColorString] = React.useState<string>(
+    color ? rgbColorToHex(color.r, color.g, color.b) : ''
+  );
+
+  // It conserves the imputed text if the color didn't changed.
```
4ian, line 70:
```suggestion
  // It keeps the inputed text if the color has not changed.
```

4ian (approved):
LGTM after a few additional comments :)

PR #3920, opened 2022-05-18
Reorder effects

### LINKED ISSUE #3749: Allow re-ordering of effects (object/layers)
## Description

I recently learned that the way effects are ordered will change how the effects are applied.  

For example, "glow + outline" looks different than "outline + glow".

If users want to adjust the order of effects today, they need to delete and recreate the effects. 

## Solution suggested

Allow users to change the order that effects (both objects and layers) are listed.  

Preferred method: Drag and drop the effects to reorder

Alternative method: Add "Move Up" and "Move Down" in a right-click menu

AlexandreSi:
Fixes #3749

Add checks on names as well

demo_effects_list

### DIFF
--- a/newIDE/app/src/EffectsList/index.js
+++ b/newIDE/app/src/EffectsList/index.js
@@ -1,9 +1,11 @@
 // @flow
-import { Trans } from '@lingui/macro';
-import { t } from '@lingui/macro';
+import * as React from 'react';
+import { t, Trans } from '@lingui/macro';
 import { I18n } from '@lingui/react';
 import { type I18n as I18nType } from '@lingui/core';
-import * as React from 'react';
+import MoreVert from '@material-ui/icons/MoreVert';
+import Add from '@material-ui/icons/Add';
+
 import { Column, Line, Spacer } from '../UI/Grid';
 import SelectField from '../UI/SelectField';
 import SelectOption from '../UI/SelectOption';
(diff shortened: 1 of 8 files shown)


newIDE/app/src/EffectsList/index.js
```diff
@@ -144,118 +204,197 @@ export default function EffectsList(props: Props) {
...
-                                  <MarkdownText
-                                    source={effectMetadata.description}
+                          {({
+                            connectDragSource,
+                            connectDropTarget,
+                            isOver,
+                            canDrop,
+                          }) =>
+                            connectDropTarget(
+                              <div
```
ClementPasteau, line 239: should those divs be <Columns> and <Lines> ?

newIDE/app/src/EffectsList/index.js
```diff
@@ -144,118 +204,197 @@ export default function EffectsList(props: Props) {
...
+                                key={effect.ptr}
+                                style={{
```
ClementPasteau, line 241: seeing inline style makes me wonder if we're doing something wrong

ClementPasteau (approved):
lgtm overall! quick question in comments

newIDE/app/src/EffectsList/index.js
AlexandreSi: You cannot give a React component to the connect functions of the drag n drop, they require pure HTML component :/

4ian:
LGTM!

newIDE/app/src/EffectsList/index.js
ClementPasteau: 👍

Issue #3927, opened 2022-05-19
New variable editor displays type select options white on white


Bouh:
## Describe the bug
For a selected row, the select list of type is hard to read.
![image](https://user-images.githubusercontent.com/1670670/169288435-dc5b1239-0ce5-4f25-90e3-9d4b8af354da.png)

Only in White theme, Dark and Nord theme are ok.

## Other details
* 5.0.133, W10

AlexandreSi:
I'm curious to see how it looks like with dark themes.
It looks like the text color set for selected row is also given to children on Windows.

Bouh:
By themes in order:
![image](https://user-images.githubusercontent.com/1670670/169296542-33320465-8982-4ca4-9093-a0409a803be5.png)

AlexandreSi:
Thanks!

PR #3928, opened 2022-05-19
New variables list improvements

AlexandreSi:
To fix:
- [x] https://github.com/4ian/GDevelop/issues/3927

@Silver-Streak ideas:
- [x] I would recommend docking the copy/paste/delete/undo/search bar at the top, so users don't have to scroll back up if they have a large list of variables.
- [ ] There probably needs to be a copy/paste variable option on the right-click menu. If there is one, I can't seem to activate it (only copy/paste text in a variable name/value seems to work)
- [x] When searching, if the top level of a structure/array is found, the children aren't shown. Not sure if that is intentional or not (or if it should be). Likewise, results that are structures or array cannot be expanded, which feels like it reduces the effectiveness/usefulness of searching somewhat, since I'd have to unsearch and then manually find the structure/array again.
- [ ] Probably needs a right-click "Collapse/Expand this level" option for child variables to collapse/expand all children of the current variable at the level selected.
- [x] When adding a new variable via the "Add variable" button, the window doesn't jump/scroll to it. So the user may not know a variable was created at the bottom of the list and may hit the button a few times. I'd probably recommend jumping to any new variable as the focus, since it was created via a conscious action by the dev clicking the button. 
- [ ] CTRL+A, select all rows.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -18,6 +18,20 @@
     {
       "type": "node",
       "request": "launch",
+      "name": "newIDE/app Jest tests (current file)",
+      "program": "${workspaceFolder}/newIDE/app/node_modules/.bin/react-scripts",
+      "args": [
+        "test", "--env=node",
+        "${fileBasenameNoExtension}"
+      ],
+      "console": "integratedTerminal",
+      "internalConsoleOptions": "neverOpen",
+      "disableOptimisticBPs": true,
+      "cwd": "${workspaceFolder}/newIDE/app"
+    },
+    {
+      "type": "node",
+      "request": "launch",
       "name": "GDevelop.js Jest tests (current file)",
       "program": "${workspaceFolder}/GDevelop.js/node_modules/.bin/jest",
       "args": [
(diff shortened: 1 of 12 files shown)


AlexandreSi:
I won't add the right click behavior as it would require to have a 3-dots additional button on each line, that would make the lines heavier.
With the fixed position toolbar, it may prove itself less necessary. We'll see if we have more requests on this in the future

AlexandreSi:
@Bouh I pushed a new version.
Storybook should be at this url http://gdevelop-storybook.s3-website-us-east-1.amazonaws.com/fine-tune-new-variables-list/latest/index.html

Can you try the select?

Bouh:
To be sure, I started the application locally, it works fine now!

image

Bouh:
If possible to add:
- [ ] CTRL+A, select all rows.
I add this to your first post.

AlexandreSi:
I was having trouble adding shortcuts to the component, and I just realized that the tree view has implemented accessibility guidelines, so it captures key events. So much that I cannot detect the A key type at my level. And Material UI does not allow to remove the accessibility shortcuts.

So Ctrl+A will be for another time (maybe TreeView is not that useful at the end of the day)

Issue #1595, opened 2020-03-28
Events are sometimes hard to drag


PascalLadalle:
## Describe the bug
When dragging a line of the events sheet, there is sometimes a position conflict that makes it impossible to know where the line will be dropped and it triggers a lag proportional to the duration of the position conflict. It seems this happens mostly when moving a line to a space **higher** in the list.

[GDevelop 5platformer.json 2020-03-28 21-24-49.mp4.zip](https://github.com/4ian/GDevelop/files/4397853/GDevelop.5platformer.json.2020-03-28.21-24-49.mp4.zip)

## To Reproduce
1. Open the platformer example
2. Grab an event somewhere down the list
3. Move it up one slot

## Other details
Win10 64 bits, GD5 b89

4ian:
Yep dragging events was never a good experience. I'd like at some point to rewrite the rendering of events and the drag'n'dropping in particular to have an approach similar to actions/conditions, with a line drawn at the position of the drop (and red if you can't drop).

Wend1go:
Keyboard shortcuts would be handy too, like:
"CTRL" + "Arrow Up" = Move current event up by one position   
"CTRL" + "Arrow Down" = Move current event down by one position 
"CTRL" + "Arrow Right" = Move current event one level down in the subevent tree
"CTRL" + "Arrow Left" = Move current event one level up in the subevent tree

gautamv95:
@4ian should we rely on some other 3rd party library for drag and drop behavior?
I looked at some of them, and I think they can provide better user experience than the current one, like keyboard shortcuts, better event handling, smooth animations, etc.

4ian:
I don't think there is another 3rd party library doing that apart from react-sortable-tree.
Feel free to suggest one if you find - but I'm thinking of getting rid of react-sortable-tree and using just react-virtualized or react-window to do the virtualization.

gautamv95:
I'll make a detailed list and let you know :)

PR #3936, opened 2022-05-20
Rework events Drag&Drop in events sheet

### LINKED ISSUE #1903: GDevelop freezes when changes the events order.
## About the bug:
The editor freezes when change the events order on the action section from the events sheet on a scene. But the navigation bar (the menu where the options to open, help, etc.) come out does work (only freezes the editor section).

## Steps to Reproduce the bug:
1. Go to any scene.
2. Open the events sheet.
3. Put some actions on the "actions" section.
4. Try to change the actions order.

## Other details:
* OS: Windows 10 2004 19041.388
* GDevelop 5.0.0 Beta-98

fannieyan:
Fixes #1903 
(at least partially)

- Change drop indicator to a slim line between events
- Drop event above/below/as subevent depending on mouse position and event type
- Temporally unfold events when user hovers over them for at least 1 sec while dragging

### DIFF
--- a/newIDE/app/scripts/theme-templates/theme.json
+++ b/newIDE/app/scripts/theme-templates/theme.json
@@ -309,6 +309,14 @@
         }
       }
     },
+    "event-drop-indicator": {
+      "border-color": {
+        "value": "#3c4698"
+      },
+      "outline-color": {
+        "value": "#fff"
+      }
+    },
     "link": {
       "container": {
         "background-color": {
--- a/newIDE/app/src/EventsSheet/EventsTree/ClassNames.js
+++ b/newIDE/app/src/EventsSheet/EventsTree/ClassNames.js
@@ -24,6 +24,8 @@ export const eventsTreeWithSearchResults = 'with-search-results';
 
 export const dropIndicator = 'drop-indicator';
 export const cantDropIndicator = 'cant-drop-indicator';
+export const eventDropIndicator = 'event-drop-indicator';
+export const handle = 'moveHandle';
 
 export const linkContainer = 'link-container';
 
--- a/newIDE/app/src/EventsSheet/EventsTree/DropContainer.js
+++ b/newIDE/app/src/EventsSheet/EventsTree/DropContainer.js
@@ -1,0 +1,178 @@
+// @flow
+import * as React from 'react';
+import { getIndentWidth, type SortableTreeNode } from '.';
+import { eventDropIndicator } from './ClassNames';
+import {
+  isDescendant,
+  moveNodeAbove,
+  moveNodeBelow,
+  moveNodeAsSubevent,
+} from './helpers';
+import { type WidthType } from '../../UI/Reponsive/ResponsiveWindowMeasurer';
+import './style.css';
+
+type ContainerPosition = 'top' | 'bottom-left' | 'bottom-right' | 'bottom';
+type IndicatorPosition = 'top' | 'bottom-right' | 'bottom';
+
+const styles = {
+  dropArea: { zIndex: 1, position: 'absolute' },
+  dropIndicator: {
+    position: 'absolute',
+    zIndex: 2,
+  },
+};
+
+const getTargetPositionStyle = (position: ContainerPosition) => {
+  switch (position) {
+    case 'bottom-left':
+      return { left: '0px', bottom: '0px', top: '50%', right: '95%' };
+    case 'bottom-right':
+      return { left: '5%', right: '0px', bottom: '0px', top: '50%' };
+    case 'top':
+      return { left: '0px', right: '0px', top: '0px', bottom: '50%' };
+    case 'bottom':
+      return { left: '0px', right: '0px', top: '50%', bottom: '0px' };
+    default:
+      return {};
+  }
+};
+
+const getIndicatorPositionStyle = (
+  position: IndicatorPosition,
+  indentWidth: number
+) => {
+  switch (position) {
+    case 'bottom':
+      return { left: '0px', right: '0px', bottom: '-3px' };
+    case 'bottom-right':
+      return { left: indentWidth, right: '0px', bottom: '-3px' };
+    case 'top':
+      return { left: '0px', right: '0px', top: '-3px' };
+    default:
+      return {};
+  }
+};
+
+function DropTargetContainer({
+  containerPosition,
+  indicatorPosition,
+  DnDComponent,
+  canDrop,
+  onDrop,
+  indentWidth,
+}: {|
+  containerPosition: ContainerPosition,
+  indicatorPosition: IndicatorPosition,
+  DnDComponent: any,
+  canDrop: () => boolean,
+  onDrop: () => void,
+  indentWidth: number,
+|}) {
+  return (
+    <DnDComponent canDrop={canDrop} drop={onDrop}>
+      {({ isOver, connectDropTarget, canDrop }) => {
+        return connectDropTarget(
+          <div>
+            {/* Drop area */}
+            <div
+              style={{
+                ...styles.dropArea,
+                ...getTargetPositionStyle(containerPosition),
+              }}
+            />
+            {/* Drop indicator */}
+            <div
+              style={{
+                ...styles.dropIndicator,
+                ...getIndicatorPositionStyle(indicatorPosition, indentWidth),
+                visibility: canDrop && isOver ? 'visible' : 'hidden',
+              }}
+              className={eventDropIndicator}
+            />
+          </div>
+        );
+      }}
+    </DnDComponent>
+  );
+}
+
+export default function DropContainer({
+  node,
+  draggedNode,
+  DnDComponent,
+  onDrop,
+  activateTargets,
+  windowWidth,
+}: {|
+  node: SortableTreeNode,
+  draggedNode: ?SortableTreeNode,
+  DnDComponent: any,
+  onDrop: (
+    moveFunction: ({
+      targetNode: SortableTreeNode,
+      node: SortableTreeNode,
+    }) => void,
+    node: SortableTreeNode
+  ) => void,
+  activateTargets: boolean,
+  windowWidth: WidthType,
+|}) {
+  // We want to allow dropping below if the event has no children OR if the only
+  // child of the event is the dragged one.
+  const canDropBelow =
+    node.event &&
+    (node.children.length === 0 ||
+      (node.children.length === 1 && node.children[0] === draggedNode));
+  const canHaveSubevents = node.event && node.event.canHaveSubEvents();
+  const commonProps = {
+    DnDComponent: DnDComponent,
+    canDrop: () => true,
+    indentWidth: getIndentWidth(windowWidth),
+  };
+  return (
+    <div
+      style={{
+        visibility: activateTargets ? 'visible' : 'hidden',
+      }}
+    >
+      <DropTargetContainer
+        containerPosition="top"
+        indicatorPosition="top"
+        onDrop={() => onDrop(moveNodeAbove, node)}
+        {...commonProps}
+      />
+      {canHaveSubevents && canDropBelow && (
+        <>
+          <DropTargetContainer
+            containerPosition="bottom-right"
+            indicatorPosition="bottom-right"
+            onDrop={() => onDrop(moveNodeAsSubevent, node)}
+            {...commonProps}
+          />
+          <DropTargetContainer
+            containerPosition="bottom-left"
+            indicatorPosition="bottom"
+            onDrop={() => onDrop(moveNodeBelow, node)}
+            {...commonProps}
+          />
+        </>
+      )}
+      {!canHaveSubevents && canDropBelow && (
+        <DropTargetContainer
+          containerPosition="bottom"
+          indicatorPosition="bottom"
+          onDrop={() => onDrop(moveNodeBelow, node)}
+          {...commonProps}
+        />
+      )}
+      {canHaveSubevents && !canDropBelow && (
+        <DropTargetContainer
+          containerPosition="bottom"
+          indicatorPosition="bottom-right"
+          onDrop={() => onDrop(moveNodeAsSubevent, node)}
+          {...commonProps}
+        />
+      )}
+    </div>
+  );
+}
--- a/newIDE/app/src/EventsSheet/EventsTree/DropContainer.js
+++ b/newIDE/app/src/EventsSheet/EventsTree/DropContainer.js
@@ -2,12 +2,7 @@
 import * as React from 'react';
 import { getIndentWidth, type SortableTreeNode } from '.';
 import { eventDropIndicator } from './ClassNames';
-import {
-  isDescendant,
-  moveNodeAbove,
-  moveNodeBelow,
-  moveNodeAsSubevent,
-} from './helpers';
+import { moveNodeAbove, moveNodeBelow, moveNodeAsSubevent } from './helpers';
 import { type WidthType } from '../../UI/Reponsive/ResponsiveWindowMeasurer';
 import './style.css';
 
--- a/newIDE/app/src/EventsSheet/EventsTree/helpers.js
+++ b/newIDE/app/src/EventsSheet/EventsTree/helpers.js
@@ -1,0 +1,76 @@
+import { type SortableTreeNode } from '.';
+
+export type MoveFunctionArguments = {
+  targetNode: SortableTreeNode,
+  node: SortableTreeNode,
+};
+
+export const moveEventToEventsList = ({
+  targetEventsList,
+  movingEvent,
+  initialEventsList,
+  toIndex,
+}: {
+  targetEventsList: gdEventsList,
+  movingEvent: ?gdBaseEvent,
+  initialEventsList: gdEventsList,
+  toIndex: number,
+}) => {
+  if (!movingEvent) return;
+  initialEventsList.moveEventToAnotherEventsList(
+    movingEvent,
+    targetEventsList,
+    toIndex
+  );
+};
+
+export const moveNodeAsSubevent = ({
+  targetNode,
+  node,
+}: MoveFunctionArguments) => {
+  if (!targetNode.event) return;
+  moveEventToEventsList({
+    targetEventsList: targetNode.event.getSubEvents(),
+    movingEvent: node.event,
+    initialEventsList: node.eventsList,
+    toIndex: 0,
+  });
+};
+
+export const moveNodeBelow = ({ targetNode, node }: MoveFunctionArguments) => {
+  if (!targetNode.event) return;
+  const toIndex =
+    node.indexInList < targetNode.indexInList && node.depth === targetNode.depth
+      ? targetNode.indexInList
+      : targetNode.indexInList + 1;
+  moveEventToEventsList({
+    targetEventsList: targetNode.eventsList,
+    movingEvent: node.event,
+    initialEventsList: node.eventsList,
+    toIndex,
+  });
+};
+
+export const moveNodeAbove = ({ targetNode, node }: MoveFunctionArguments) => {
+  const toIndex =
+    node.indexInList < targetNode.indexInList && node.depth === targetNode.depth
+      ? targetNode.indexInList - 1
+      : targetNode.indexInList;
+  moveEventToEventsList({
+    targetEventsList: targetNode.eventsList,
+    movingEvent: node.event,
+    initialEventsList: node.eventsList,
+    toIndex,
+  });
+};
+
+// Optimises react-sortable-tree isDescendant that processes every children with a .some method.
+export const isDescendant = (
+  olderNode: SortableTreeNode,
+  childrenNode: SortableTreeNode
+) => {
+  const parentPath = olderNode.nodePath;
+  const childrenPath = childrenNode.nodePath;
+  if (childrenNode.depth <= olderNode.depth) return false;
+  return parentPath[olderNode.depth] === childrenPath[olderNode.depth];
+};
--- a/newIDE/app/src/EventsSheet/EventsTree/index.js
+++ b/newIDE/app/src/EventsSheet/EventsTree/index.js
@@ -306,12 +306,12 @@ export default class ThemableEventsTree extends Component<
     eventsSheetEventsDnDType
   );
   DropTarget = makeDropTarget<SortableTreeNode>(eventsSheetEventsDnDType);
-  temporarlyUnfoldedNodes: Array<SortableTreeNode>;
+  temporallyUnfoldedNodes: Array<SortableTreeNode>;
   _hoverTimerId: ?number;
 
   constructor(props: EventsTreeProps) {
     super(props);
-    this.temporarlyUnfoldedNodes = [];
+    this.temporallyUnfoldedNodes = [];
     this.eventsHeightsCache = new EventHeightsCache(this);
     this.state = {
       ...this._eventsToTreeData(props.events),
@@ -568,20 +568,20 @@ export default class ThemableEventsTree extends Component<
     );
   };
 
-  _temporarlyUnfoldNode = (isOverLazy: boolean, node: SortableTreeNode) => {
+  _temporallyUnfoldNode = (isOverLazy: boolean, node: SortableTreeNode) => {
     if (!node.event) return;
 
     const { event: nodeEvent } = node;
 
-    const isNodeTemporarlyUnfolded = !this.temporarlyUnfoldedNodes.some(
+    const isNodeTemporallyUnfolded = !this.temporallyUnfoldedNodes.some(
       foldedNode => foldedNode.event && nodeEvent.ptr === foldedNode.event.ptr
     );
     if (isOverLazy) {
       if (!this._hoverTimerId && !node.expanded) {
         this._hoverTimerId = window.setTimeout(() => {
-          if (node.event && isNodeTemporarlyUnfolded) {
+          if (node.event && isNodeTemporallyUnfolded) {
             nodeEvent.setFolded(false);
-            this.temporarlyUnfoldedNodes.push(node);
+            this.temporallyUnfoldedNodes.push(node);
             this.forceEventsUpdate();
           }
         }, 1000);
@@ -593,11 +593,11 @@ export default class ThemableEventsTree extends Component<
   };
 
   _restoreFoldedNodes = () => {
-    this.temporarlyUnfoldedNodes.forEach(
+    this.temporallyUnfoldedNodes.forEach(
       node => node.event && node.event.setFolded(true)
     );
 
-    this.temporarlyUnfoldedNodes = [];
+    this.temporallyUnfoldedNodes = [];
     this.forceEventsUpdate();
   };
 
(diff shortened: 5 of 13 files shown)


newIDE/app/src/EventsSheet/EventsTree/ClassNames.js
```diff
@@ -24,6 +24,8 @@ export const eventsTreeWithSearchResults = 'with-search-results';
 
 export const dropIndicator = 'drop-indicator';
 export const cantDropIndicator = 'cant-drop-indicator';
+export const eventDropIndicator = 'event-drop-indicator';
```
4ian, line 27: Should we actually reuse the existing dropIndicator?

newIDE/app/src/EventsSheet/EventsTree/index.js
```diff
@@ -665,6 +749,7 @@ export default class ThemableEventsTree extends Component<EventsTreeProps, *> {
             ref: list => (this._list = list),
             onScroll: this.props.onScroll,
           }}
+          slideRegionSize={0}
```
4ian, line 752: I think this also reduce the region size on desktop, let's keep it to a good value and investigate what we can do for mobile/touchscreens :) Maybe there is a way to avoid being "stuck".

4ian (commented):
Great progress, this will be way less buggy than what we have now :)

newIDE/app/src/UI/Theme/Global/EventsSheet.css
```diff
@@ -36,202 +36,210 @@
...
+  margin-bottom: -1px;
+  width: 100%;
+  box-sizing: border-box;
+}
+
+/**
+* Drag'n'drop indicator (events)
+*/
+.gd-events-sheet .event-drop-indicator {
+  border: 3px solid var(--event-sheet-event-drop-indicator-border-color);
```
Bouh, line 221: Please make this border property fully customizable in the theme 🙏

newIDE/app/src/UI/Theme/Global/EventsSheet.css
```diff
@@ -36,202 +36,210 @@
...
+  outline: var(--event-sheet-event-drop-indicator-outline-color) solid 1px;
```
Bouh, line 222: Same for the outline, please make it fully customizable in the theme 🙏

Bouh (commented):
- [ ] Can we keep a kind of consistency in the colors of the indicator line?
image
- [ ] I ask, did you think it is possible to drag an a sub-event on the left for making it at the same level of the parent?

1) Drag and move on left
image

2) Release the touch/mouse
image

Bouh:
- [ ] It is possible to hide the handle under the cursor while the DND is active? 
image

newIDE/app/src/UI/Theme/Global/EventsSheet.css
4ian: For consistency we should have the same drop indicator (color and size) everywhere (like in the objects list, group list, etc...)

fannieyan:
@Bouh 
> * [ ]  It is possible to hide the handle under the cursor while the DND is active?
Working with preview is a bit complicated, it may not be worth to try



 > [ ] I ask, did you think it is possible to drag an a sub-event on the left for making it at the same level of the parent?
Actually, you can, but you must use the parent. I added an "option" to just drag left/right for this feature

https://user-images.githubusercontent.com/81410437/169853819-811870a3-0658-4a8a-9b8c-8aec3fe8d52f.mov

newIDE/app/src/EventsSheet/EventsTree/ClassNames.js
fannieyan: Finally I used the same color that is used for closable tabs and object lists. (I added dropIndicator in the theme)
Regarding the size, it may not be lisible enough to keep the same everywhere (the one in object list is quite thin).

It may be good to create a DropIndicator in the UI folder in the future (currently, we have 3-4 drop indicators, that are all implemented in a different way 😅 )

newIDE/app/src/EventsSheet/EventsTree/index.js
fannieyan: Yes, I saw that after committing 😅 
I tried the method of drag&scroll using drop targets (see WIP commit) but it looks like it may not be a problem of rst like I thought as the problem persists. (at least it can scroll up)

newIDE/app/scripts/theme-templates/theme.json
```diff
@@ -309,6 +309,14 @@
         }
       }
     },
+    "event-drop-indicator": {
```
AlexandreSi, line 312: Should this be the same as the fields you added in the different themes? `drop-indicator`, `cannot-drop`, `can-drop`, etc.

newIDE/app/src/EventsSheet/EventsTree/ClassNames.js
```diff
@@ -24,6 +24,8 @@ export const eventsTreeWithSearchResults = 'with-search-results';
...
+export const handle = 'moveHandle';
```
AlexandreSi, line 28: Can I suggest you use kebab-case for this class name for consistency's sake?

newIDE/app/src/EventsSheet/EventsTree/helpers.js
```diff
@@ -0,0 +1,92 @@
...
+    movingEvent: node.event,
+    initialEventsList: node.eventsList,
+    toIndex: 0,
+  });
+};
+
+export const moveNodeBelow = ({ targetNode, node }: MoveFunctionArguments) => {
+  if (!targetNode.event) return;
+  const toIndex =
+    node.indexInList < targetNode.indexInList && node.depth === targetNode.depth
```
AlexandreSi, line 43: Why is there a check on the depth?

newIDE/app/src/EventsSheet/EventsTree/helpers.js
```diff
@@ -0,0 +1,92 @@
...
+    targetEventsList: targetNode.eventsList,
+    movingEvent: node.event,
+    initialEventsList: node.eventsList,
+    toIndex,
+  });
+};
+
+export const moveNodeAbove = ({ targetNode, node }: MoveFunctionArguments) => {
+  const toIndex =
+    node.indexInList < targetNode.indexInList && node.depth === targetNode.depth
```
AlexandreSi, line 56: Same comment as above abouth depth check

newIDE/app/src/EventsSheet/EventsTree/helpers.js
```diff
@@ -0,0 +1,92 @@
...
+      : targetNode.indexInList;
+  moveEventToEventsList({
+    targetEventsList: targetNode.eventsList,
+    movingEvent: node.event,
+    initialEventsList: node.eventsList,
+    toIndex,
+  });
+};
+
+// Optimises react-sortable-tree isDescendant that processes every children with a .some method.
```
AlexandreSi, line 67: Is the comment up to date? I don't see a `.some` in the code

newIDE/app/src/EventsSheet/EventsTree/helpers.js
```diff
@@ -0,0 +1,92 @@
...
+export const isDescendant = (
+  olderNode: SortableTreeNode,
+  childrenNode: SortableTreeNode
+) => {
+  const parentPath = olderNode.nodePath;
+  const childrenPath = childrenNode.nodePath;
+  if (childrenNode.depth <= olderNode.depth) return false;
```
AlexandreSi, line 74: You can do this check as a first thing and then check the paths

newIDE/app/src/EventsSheet/EventsTree/helpers.js
```diff
@@ -0,0 +1,92 @@
...
+  return parentPath.every(
```
AlexandreSi, line 75: Could it be worth checking that parentPath is somewhere in the childPath? Seems moe efficient to me but I'm not sure

newIDE/app/src/EventsSheet/EventsTree/DropContainer.js
```diff
@@ -0,0 +1,295 @@
...
+          }}
+          onDrop={() => onDrop(moveNodeAsSubevent, node)}
+          {...commonProps}
+        />
+      )}
+    </div>
+  );
+}
+
+export function Autoscroll({
```
AlexandreSi, line 237: `AutoScroll`?

newIDE/app/src/EventsSheet/EventsTree/DropContainer.js
```diff
@@ -0,0 +1,295 @@
...
+  height?: number,
+|};
+
+function DropTargetContainer({
+  DnDComponent,
+  canDrop,
+  onDrop,
+  style,
+}: {|
+  DnDComponent: any,
```
AlexandreSi, line 73: Is there a  slightly more precise type you can use? `React.Node` for example?

newIDE/app/src/EventsSheet/EventsTree/DropContainer.js
```diff
@@ -0,0 +1,295 @@
...
+    border: '4px solid black',
+    outline: '1px solid white',
+  },
+};
+
+const getTargetPositionStyles = (
+  indentWidth: number,
+  draggedNodeHeight: number,
+  isDraggedNodeChild: boolean
+) => ({
```
AlexandreSi, line 28: Is it worth mentionning it returns `{[position: string]; DropTargetContainerStyle}` ?

newIDE/app/src/EventsSheet/EventsTree/DropContainer.js
```diff
@@ -0,0 +1,295 @@
...
+  return (
+    <DnDComponent canDrop={canDrop} drop={onDrop}>
+      {({ isOver, connectDropTarget, canDrop }) => {
+        return connectDropTarget(
+          <div>
+            {/* Drop area */}
+            <div
+              style={{
+                ...styles.dropArea,
+                ...style.dropArea,
```
AlexandreSi, line 91: It's a bit confusing to see theses 2 drop area styles. Can you maybe rename `styles` with `baseStyles` or `sharedStyles`?

newIDE/app/src/EventsSheet/EventsTree/DropContainer.js
```diff
@@ -0,0 +1,295 @@
...
+    node.children.length === 1 &&
+    !!draggedNode &&
+    node.children[0].key === draggedNode.key;
+  const isDraggedNodeSameDepthAndBelow =
+    !!draggedNode && isSameDepthAndBelow(node, draggedNode);
+  // We want to allow dropping below if the event has no children OR if the only
+  // child of the event is the dragged one.
+  const canDropBelow =
+    node.event && (node.children.length === 0 || isDraggedNodeNodesOnlyChild);
+  const canHaveSubevents = node.event && node.event.canHaveSubEvents();
```
AlexandreSi, line 145: ~~canHaveSubevents~~ canHaveSubEvents

newIDE/app/src/EventsSheet/EventsTree/DropContainer.js
```diff
@@ -0,0 +1,295 @@
...
+}: {|
+  direction: 'top' | 'bottom',
+  DnDComponent: any,
+  activateTargets: boolean,
+  onHover: () => void,
+|}) {
+  const delayActivationTimer = React.useRef<?TimeoutID>(null);
+  const [show, setShow] = React.useState(false);
+
+  // This drop target overlaps with sibling drag source and cancels drag immediatly.
```
AlexandreSi, line 251: immediately

newIDE/app/scripts/theme-templates/theme.json
fannieyan: Indeed, forgot that one !

newIDE/app/src/EventsSheet/EventsTree/DropContainer.js
```diff
@@ -0,0 +1,295 @@
...
+  node,
+  draggedNode,
+  DnDComponent,
+  onDrop,
+  activateTargets,
+  windowWidth,
+  draggedNodeHeight,
+}: {|
+  node: SortableTreeNode,
+  draggedNode: ?SortableTreeNode,
```
AlexandreSi, line 122: I wonder if this component should exist if there is no draggedNode?

newIDE/app/src/EventsSheet/EventsTree/DropContainer.js
```diff
@@ -0,0 +1,295 @@
...
+      targetNode: SortableTreeNode,
+      node: SortableTreeNode,
+    }) => void,
+    node: SortableTreeNode
+  ) => void,
+  activateTargets: boolean,
+  windowWidth: WidthType,
+  draggedNodeHeight: number,
+|}) {
+  const isDraggedNodeNodesOnlyChild =
```
AlexandreSi, line 135: Is there a typo? `NodeNodes` is a bit confusing

newIDE/app/src/EventsSheet/EventsTree/helpers.js
fannieyan: I actually browsed react-sortable-tree's utils function and was talking about their "isDescendant" method.
https://github.com/frontend-collective/react-sortable-tree/blob/52f0bcf51a80297ab63d5d1cfb0b32f4e3f0e702/src/utils/tree-data-utils.js#L1032

newIDE/app/src/MainFrame/EditorTabs/DropIndicator.js
```diff
@@ -19,8 +19,8 @@ export function ColumnDropIndicator() {
     <div
       style={{
         ...styles.columnDropIndicator,
-        backgroundColor: gdevelopTheme.closableTabs.selectedBackgroundColor,
-        borderColor: gdevelopTheme.closableTabs.backgroundColor,
+        backgroundColor: gdevelopTheme.dropIndicator.canDrop,
+        borderColor: gdevelopTheme.dropIndicator.border,
```
AlexandreSi, line 23: 👌

newIDE/app/src/EventsSheet/EventsTree/index.js
```diff
@@ -134,46 +158,51 @@ class EventContainer extends Component<EventsContainerProps, {||}> {
...
-              disabled /* Use disabled (not event.disabled) as it is true if a parent event is disabled*/
-            }
-            renderObjectThumbnail={this.props.renderObjectThumbnail}
-            screenType={this.props.screenType}
-            windowWidth={this.props.windowWidth}
-            eventsSheetHeight={this.props.eventsSheetHeight}
-          />
+          <div style={styles.eventComponentContainer}>
+            {this.props.connectDragSource(
+              <div style={styles.handle} className={handle} />
```
AlexandreSi, line 165: Styles is defined by both the style props and the className props. The classname was used because it was the only way to customize the rst handle but I think you can use only the style props now.

newIDE/app/src/EventsSheet/EventsTree/index.js
```diff
@@ -239,35 +268,59 @@ type EventsTreeProps = {|
...
   static defaultProps = {
     selection: getInitialSelection(),
   };
   _list: ?any;
   eventsHeightsCache: EventHeightsCache;
+  DragSourceAndDropTarget = makeDragSourceAndDropTarget<SortableTreeNode>(
+    eventsSheetEventsDnDType
+  );
+  DropTarget = makeDropTarget<SortableTreeNode>(eventsSheetEventsDnDType);
+  temporallyUnfoldedNodes: Array<SortableTreeNode>;
```
AlexandreSi, line 312: Should it be `temporaryUnfoldedNodes`?

newIDE/app/src/EventsSheet/EventsTree/helpers.js
AlexandreSi: Is this because the tree is flattened?

newIDE/app/src/EventsSheet/EventsTree/index.js
fannieyan: I cannot believe Google translate failed me like this :(
image

newIDE/app/src/EventsSheet/EventsTree/index.js
```diff
@@ -239,35 +268,59 @@ type EventsTreeProps = {|
...
+  _hoverTimerId: ?number;
```
AlexandreSi, line 313: I think you can clear the timeout if this is not null on component unmount

newIDE/app/src/EventsSheet/EventsTree/index.js
```diff
@@ -557,54 +575,140 @@ export default class ThemableEventsTree extends Component<EventsTreeProps, *> {
...
 
+  _temporallyUnfoldNode = (isOverLazy: boolean, node: SortableTreeNode) => {
+    if (!node.event) return;
+
+    const isNodeTemporallyUnfolded = this.temporallyUnfoldedNodes.some(
+      foldedNode => node.key === foldedNode.key
+    );
+    if (isOverLazy) {
+      if (!this._hoverTimerId && !node.expanded) {
+        if (node.event && !isNodeTemporallyUnfolded) {
```
AlexandreSi, line 586: the check on event presence shoud not be necessary here since you're doing it first thing in the method.

Can destructuring with 
```js
const {event} = node;
if (!event) return;
```
help you?

newIDE/app/src/EventsSheet/EventsTree/index.js
```diff
@@ -557,54 +575,140 @@ export default class ThemableEventsTree extends Component<EventsTreeProps, *> {
...
+                })}
+                DnDComponent={DropTarget}
+                onDrop={this._onDrop}
+                activateTargets={!isDragged && !!this.state.draggedNode}
+                windowWidth={this.props.windowWidth}
+              />
+            </div>
+          );
+
+          if (!dropTarget) return null;
```
AlexandreSi, line 708: Why would it be null?

AlexandreSi:
A great improvement!
I inundated you with comments but that's a complex part of the events sheet

newIDE/app/src/EventsSheet/EventsTree/helpers.js
fannieyan: IndexInList is the index of the event in the `EventsList`.
If the nodes are of same depth, we have to take into account the shift in index that comes from moving the node if the targeted node is after the dragged one. If the nodes are of different depth, there is no shift in index.
And depth check makes indexInList comparison relevant

newIDE/app/src/EventsSheet/EventsTree/helpers.js
fannieyan: I think that is what I do ? I check that the parent path is at the start of child path.
Do you think about a method or function in particular ?

newIDE/app/src/EventsSheet/EventsTree/helpers.js
AlexandreSi: Alright, I think I got the logic here.

newIDE/app/src/EventsSheet/EventsTree/helpers.js
AlexandreSi: I don't know if it's worth adding this comment. You're saying that you could have used RST's isDescendant method but you preferred a more optimized version of your own. But I don't know what to do with this info: you're not overriding a RST default method or using it as an unexpected props for a component.

newIDE/app/src/EventsSheet/EventsTree/helpers.js
4ian: Ubernit:

> And depth check makes indexInList comparison relevant

Prefer to start your condition with node.depth === targetNode.depth, which conveys the idea to the reader "**in the case of the same depth**, then look at this, otherwise do as usual". Otherwise the reader can be confused: "why are we testing this whereas this is not relevant... ah ok we check if it's relevant just after".

newIDE/app/src/EventsSheet/EventsTree/helpers.js
AlexandreSi: I think that when I wrote that comment, I was not completely understanding the method oupsi.
Now it seems quite efficient to me!

newIDE/app/src/EventsSheet/EventsTree/index.js
AlexandreSi: Ahah and I failed you with my suggestion!

newIDE/app/src/EventsSheet/EventsTree/DropContainer.js
4ian, line 145: nit:
```suggestion
  const canHaveSubevents = !!node.event && node.event.canHaveSubEvents();
```

to force this variable to be a boolean

newIDE/app/src/EventsSheet/EventsTree/DropContainer.js
```diff
@@ -0,0 +1,300 @@
...
+  direction: 'top' | 'bottom',
+  DnDComponent: DropTargetComponent<SortableTreeNode>,
+  activateTargets: boolean,
+  onHover: () => void,
+|}) {
+  const delayActivationTimer = React.useRef<?TimeoutID>(null);
+  const [show, setShow] = React.useState(false);
+
+  // This drop target overlaps with sibling drag source and cancels drag immediately.
+  // See: https://github.com/react-dnd/react-dnd/issues/766#issuecomment-388943403
```
4ian, line 260: Very nice to leave these comments to github issues, will be useful should this break in the future.

newIDE/app/src/EventsSheet/EventsTree/helpers.js
```diff
@@ -0,0 +1,89 @@
...
+  moveEventToEventsList({
+    targetEventsList: targetNode.eventsList,
+    movingEvent: node.event,
+    initialEventsList: node.eventsList,
+    toIndex,
+  });
+};
+
+export const moveNodeAbove = ({ targetNode, node }: MoveFunctionArguments) => {
+  const toIndex =
```
4ian, line 55: You're not doing `if (!targetNode.event) return;`, is this normal?
And if no, should we also do `if (!targetNode.event || !node.event) return;` to protect against a node that would not have event (should never happen, but who knows)

newIDE/app/src/EventsSheet/EventsTree/helpers.js
```diff
@@ -0,0 +1,89 @@
...
+  olderNode: SortableTreeNode,
+  childNode: SortableTreeNode
+) => {
+  if (childNode.depth <= olderNode.depth) return false;
+  const parentPath = olderNode.nodePath;
+  const childPath = childNode.nodePath;
+  return parentPath.every((pathValue, index) => pathValue === childPath[index]);
+};
+
+export const isSameDepthAndBelow = (
```
4ian, line 77: Should this be named

```suggestion
export const isSameDepthAndJustBelow = (
```

in the sense that the belowNode must be just before the aboveNode?

newIDE/app/src/EventsSheet/EventsTree/index.js
fannieyan: It looks like there is some logic linked with the classname (the color and size of the handle will change when there is a search for example).
I will move the styles to the move-handle css instead!

newIDE/app/src/EventsSheet/EventsTree/index.js
```diff
@@ -557,54 +577,142 @@ export default class ThemableEventsTree extends Component<EventsTreeProps, *> {
...
+      <DragSourceAndDropTarget
+        beginDrag={() => {
+          this.setState({ draggedNode: node });
+          return node;
+        }}
+        canDrag={() => this._canDrag(node)}
+        canDrop={() => this._canDrop(node)}
+        // Drop operations are handled by DropContainers
+        drop={() => {
+          return;
```
4ian, line 638: Ubernitpicking:
```suggestion
        drop={() => {
          // Drop operations are handled by DropContainers.
          return;
```

newIDE/app/src/EventsSheet/EventsTree/index.js
```diff
@@ -295,6 +348,10 @@ export default class ThemableEventsTree extends Component<EventsTreeProps, *> {
     }
   }
 
+  componentWillUnmount() {
+    this._hoverTimerI && clearTimeout(this._hoverTimerId);
```
4ian, line 352: Typo here: _hoverTimerI

newIDE/app/src/EventsSheet/EventsTree/index.js
```diff
@@ -239,35 +268,59 @@ type EventsTreeProps = {|
...
+export type SortableTreeNode = {
   eventsList: gdEventsList,
   event: ?gdBaseEvent,
   depth: number,
   disabled: boolean,
   indexInList: number,
+  nodePath: Array<number>,
+  children: Array<any>,
+  expanded: boolean,
+  key: number,
```
4ian, line 280: Can you comment what is this key?

newIDE/app/src/EventsSheet/EventsTree/index.js
```diff
@@ -239,35 +268,59 @@ type EventsTreeProps = {|
...
 // event, except for the buttons at the bottom of the sheet.
-type SortableTreeNode = {
+export type SortableTreeNode = {
   eventsList: gdEventsList,
   event: ?gdBaseEvent,
   depth: number,
   disabled: boolean,
   indexInList: number,
+  nodePath: Array<number>,
+  children: Array<any>,
```
4ian, line 278: Can you comment why we need the children stored here?

4ian (commented):
Great progress! 
Let's try to make this as clean as possible to ensure this is understandable/maintainable. In addition to comments, can you check:

- If you can add a boolean that would allow to display "debug version" of the drop targets, that have a outline/border?

AlexandreSi:
I think we can close #1595 after this is merged as well

newIDE/app/src/EventsSheet/EventsTree/DropContainer.js
```diff
@@ -0,0 +1,323 @@
...
+      if (activateTargets) {
+        delayActivationTimer.current = setTimeout(() => {
+          setShow(true);
+        }, 100);
+      } else {
+        setShow(false);
+        clearTimeout(delayActivationTimer.current);
+        delayActivationTimer.current = null;
+      }
+    },
```
4ian, line 292: Should you return a cleanup function from this effect? Is there a risk that unmounting the component would not clear the timeout? (it's 100ms so the risk is minimal, just give it a try to see if using the cleanup function of the effect works)

4ian (approved):
LGTM after these last comments!

newIDE/app/src/EventsSheet/EventsTree/style.css
```diff
@@ -9,19 +9,20 @@
 /**
  * Draggable handle on the left of an event
  */
-.gd-events-sheet .rst__moveHandle {
+.gd-events-sheet .move-handle {
     width: 10px;
-    border: none;
-    box-shadow: none;
-    border-radius: 0;
+    margin: 0.5px;
```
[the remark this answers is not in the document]
fannieyan: If the question is "why I added it", it is because react-sortable-tree had a default border in their .movehandle class (that we do not use anymore). If it is "why this specific value", it is the one that felt the closest to what we have right now :)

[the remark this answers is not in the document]
4ian: Ok thanks 👍

PR #3937, opened 2022-05-20
Fix expression parameter builder

### LINKED ISSUE #3932: Expression Builder - object variables not in drop-down list
## Describe the bug

When using the expression builder, object variables do not show up on the drop-down list unless you open the object variable box first. 

## To Reproduce

use expression builder

go to object -> variables

select object - then on drop down, nothing shows up underneath.

Click on object variable dialog on right, and press close.

Drop down menu now populates.

## Other details
Windows 11 - 64bit
GDevelop 5.0.133 (was also happening on 5.0.132, not certain on earlier versions)

AlexandreSi:
Fixes #3932

### DIFF
--- a/newIDE/app/src/MainFrame/Preferences/PreferencesProvider.js
+++ b/newIDE/app/src/MainFrame/Preferences/PreferencesProvider.js
@@ -73,11 +73,8 @@ export const getInitialPreferences = () => {
 };
 
 const getPreferences = () => {
-  const initialPreferences = getInitialPreferences();
-  const localPreferences = loadPreferencesFromLocalStorage();
-  const preferences = localPreferences
-    ? { ...initialPreferences, ...localPreferences } // Merge initial preferences with local preferences in case a new value has been added to the preferences
-    : initialPreferences;
+  const preferences =
+    loadPreferencesFromLocalStorage() || getInitialPreferences();
   setLanguageInDOM(preferences.language);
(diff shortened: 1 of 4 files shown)


newIDE/app/src/MainFrame/Preferences/PreferencesProvider.js
```diff
@@ -73,8 +73,11 @@ export const getInitialPreferences = () => {
 };
 
 const getPreferences = () => {
-  const preferences =
-    loadPreferencesFromLocalStorage() || getInitialPreferences();
+  const initialPreferences = getInitialPreferences();
+  const localPreferences = loadPreferencesFromLocalStorage();
+  const preferences = localPreferences
+    ? { ...initialPreferences, ...localPreferences } // Merge initial preferences with local preferences in case a new value has been added to the preferences
```
4ian, line 79: Is this really useful? 

It should be done already by loadPreferencesFromLocalStorage:
image

AlexandreSi: I'll dig into it, because it seemed to me that my local preferences missed a value

AlexandreSi: Indeed, that is not needed!

4ian (approved):

PR #3940, opened 2022-05-21
Speed up IDE JS transpilation by avoiding to process locale files

4ian:
These files don't need to be processed by Babel and they are huge. Patch create-react-app to avoid running Babel on them.
- Gitpod: Going from 90s of transpilation to 30s.
- 2017 Macbook Pro (3,5 GHz Intel Core i7): Going from ~3min20 to ~1min10.

TODO: 
- Check built app
- Check Storybook


4ian:
Closing as replaced by #3941

PR #3949, opened 2022-05-24
Set interactive children to false in runtime scene

arthuro555:
[Untested]

Sets `interactiveChildren` to false on the root container. That tells PIXI not to run a recursive hit test function on the containers children, and may improve perfromance.

I am assuming that since it is recursive, putting it to false on the root container is enough, but if I am wrong the layers renderers may also need to have that property set.

See:
https://pixijs.download/dev/docs/PIXI.Container.html#interactiveChildren
https://github.com/pixijs/pixijs/discussions/8358#discussioncomment-2804783


4ian:
Interesting, this may allow to avoid an extra Pixi "Ticker" that checks for what's under the cursor - though I'm not sure if this is currently running in the games.

arthuro555:
To optimize further, we could go to the next level and make a build of PIXI without the interactivity and accessibility plugins at all, since we are not using them.

PR #3950, opened 2022-05-24
Prevent runtime from trying to parse js map files to avoid Chrome to log w…

### LINKED ISSUE #3947: Remove warnings from debugger sourcemaps, this freak out people, but it's benign
## Describe the bug

Can we remove warnings from debugger sourcemaps and logs in general because it gives a feeling that there is a bug to people, but it is totally benign.

**This has been visible for years, experimented user know this is  benign but new people see a "bugged game engine" while this is not source of bugs.**

A little non-exhaustive history of people who **freak out** when they see warnings:

https://forum.gdevelop.io/t/solved-console-error-while-debugging-my-game/39766
https://forum.gdevelop.io/t/solved-errors-in-developer-tools-console/30970
https://forum.gdevelop.io/t/my-game-doesnt-work/23580/5
https://forum.gdevelop.io/t/game-runs-inconcistantly/34246/4
https://forum.gdevelop.io/t/scene-freezes-game/39231/3
https://forum.gdevelop.io/t/game-crashes-when-touching-platformer-character/32134/11
https://discord.com/channels/258623956158906368/258659467887902730/978257735407783986 

## To Reproduce
1. Open the dev tools or debugger in game.

GD5

AlexandreSi:
…arnings

Fixes #3947 

I added leading `// ` to source mapping file url to prevent Chrome to try to load it instead of removing the whole line that, maybe, could prove useful in the future.


4ian (approved):
Would it be worth adding `// Disabled to avoid warnings in Chrome (see https://github.com/4ian/GDevelop/pull/3950) //` instead of just // to self document this workaround?

This could help the future person debugging "why are sourcemaps not working for these"? "Oh they were disabled" "Oh ok they did not even work because they were not included and created warnings."

arthuro555:
> Oh ok they did not even work because they were not included

Most of the files here do have included source maps though.

arthuro555:
This PR can be reverted as now sourcemaps are correctly exported with #4145

4ian:
Can't revert automatically, I guess we need a few find-replace in the source!

PR #3951, opened 2022-05-24
Fix: Allow debugger server to run when not connected to Internet

### LINKED ISSUE #3709: Using the debugger offline
A forum user reports that the debugger doesn't work if we're offline when we open GDevelop, even if we connect afterwards.
On the other hand, if the debugger has been used online, it still works after disconnect.
The debugger should thus work offline, even if GDevelop is started offline.
I confirm the issue on my side, no preview is detected.
Win10, GD5.0.127

[source](https://forum.gdevelop.io/t/debugger-stopped-working/38374/6)

AlexandreSi:
Fixes #3709

### DIFF
--- a/newIDE/electron-app/app/Utils/LocalNetworkIpFinder.js
+++ b/newIDE/electron-app/app/Utils/LocalNetworkIpFinder.js
@@ -10,7 +10,10 @@ const getLocalNetworkIps = () => {
     .map(({ address }) => address);
 };
 
-/** @returns {?string} */
+/**
+ * Returns null if not connected to internet with an IPv4 connection.
+ * @returns {?string}
+ */
 const findLocalIp = () => {
   const ipAddresses = getLocalNetworkIps();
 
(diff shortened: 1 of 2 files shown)


newIDE/electron-app/app/Utils/LocalNetworkIpFinder.js
```diff
@@ -10,7 +10,10 @@ const getLocalNetworkIps = () => {
     .map(({ address }) => address);
 };
 
-/** @returns {?string} */
+/**
+ * Returns null if not connected to internet with an IPv4 connection.
+ * @returns {?string}
+ */
 const findLocalIp = () => {
```
ClementPasteau, line 17: should we change this function to return 127.0.0.1 if ipAddresses === null ?

AlexandreSi: I'm not sure if returning a local (in the scope of the machine) IP address is really expected from this function

ClementPasteau: Then the name of the function is really confusing 😅 
Either way, lgtm 👍

ClementPasteau (approved):

PR #3958, opened 2022-05-25
Add context in search placeholder

fannieyan:
Instead of just having "Search" as placeholder in search bars, we now have the scope of the search.

image
image

### DIFF
--- a/newIDE/app/src/ProjectManager/index.js
+++ b/newIDE/app/src/ProjectManager/index.js
@@ -1069,6 +1069,7 @@ export default class ProjectManager extends React.Component<Props, State> {
               value={searchText}
               onRequestSearch={this._onRequestSearch}
               onChange={this._onSearchChange}
+              placeholder={t`Search manager`}
             />
             {this.state.projectVariablesEditorOpen && (
               <VariablesEditorDialog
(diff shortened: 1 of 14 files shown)


newIDE/app/src/ProjectManager/index.js
```diff
@@ -1069,6 +1069,7 @@ export default class ProjectManager extends React.Component<Props, State> {
               value={searchText}
               onRequestSearch={this._onRequestSearch}
               onChange={this._onSearchChange}
+              placeholder={t`Search manager`}
```
4ian, line 1072: Search in project

AlexandreSi, line 1072: I don't know if this is very clear. If we were really accurate, it would be "Search in layout, external events, external layouts and functions/behaviors names". Might not be worth adding a placeholder here

4ian (approved):

newIDE/app/src/ProjectManager/index.js
4ian: +1, or "Search in project"

AlexandreSi (approved):
LGTM! Just a doubt for the project manager search bar placeholder though

newIDE/app/src/ProjectManager/index.js
fannieyan: Actually, I think a placeholder is necessary for this one, it was one of the few that I did not know what it searched for at first glance!
Florian suggested "Search in project" instead.

AlexandreSi: Let's go with `Search in project` then! We'll see if users are really confused

PR #3964, opened 2022-05-25
Clear authors and publishers when opening project after creation

fannieyan:
In particular, remove authors and publishers when creating a new project from an example.


ClementPasteau (approved):
lgtm!

AlexandreSi (approved):

PR #3965, opened 2022-05-27
Search sort to make the asset color filter not excluding.

D8H:
For colors, the top results are more pertinent, there is no sensible difference in response time and no asset are excluded.

### DIFF
--- a/newIDE/app/src/UI/Search/UseSearchItem.js
+++ b/newIDE/app/src/UI/Search/UseSearchItem.js
@@ -17,24 +17,42 @@ export class TagSearchFilter<SearchItem: { tags: Array<string> }>
   }
 
   getPertinence(searchItem: SearchItem): number {
-    return (
-      this.tags.size === 0 || searchItem.tags.some(tag => this.tags.has(tag))
-    );
+    return this.tags.size === 0 ||
+      searchItem.tags.some(tag => this.tags.has(tag))
+      ? 1
+      : 0;
   }
 }
 
-const partialQuickSort = <Element: any>(
+/**
+ * Approximately sort the elements from biggest to smallest.
+ *
+ * It does a quick sort but only on the left side. It means that elements with
+ * a poor pertinence won't be sorted as well as the one with a good pertinence.
+ *
+ * This allows a O(n) complexity in most cases, but still a O(n²) in worst
+ * cases scenario (when all the values are the same).
+ */
+export const partialQuickSort = <Element: any>(
   searchItems: Array<Element>,
   getValue: (a: Element) => number,
-  pertinenceMax: number
+  valueMin: number,
+  valueMax: number
 ): void => {
+  if (valueMin >= valueMax) {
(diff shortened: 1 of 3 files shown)


newIDE/app/src/UI/Search/UseSearchItem.js
```diff
@@ -16,13 +16,83 @@ export class TagSearchFilter<SearchItem: { tags: Array<string> }>
...
+/**
+ * Approximately sort the elements from biggest to smallest.
+ *
+ * It does a quick sort but only on the left side. It means that elements with
+ * a poor pertinence won't be sorted as well as the one with a good pertinence.
+ *
+ * This allows a O(n) complexity in most cases, but still a O(n²) in worst
+ * cases scenario (when all the values are the same).
+ */
+export const partialQuickSort = <Element: any>(
```
ClementPasteau, line 36: should this function not mutate the array passed and instead return a new sorted array? 
I'm usually quite wary of functions that don't return anything but have modified one of their parameters magically.

newIDE/app/src/UI/Search/UseSearchItem.spec.js
```diff
@@ -15,3 +15,55 @@ describe('UseSearchItem', () => {
...
+      expectedElements.length = 8;
+      expectedElements.fill(pertinence);
+      expect(elements).toStrictEqual(expectedElements);
+    });
+  });
+  test('Sort search results', () => {
+    const elements = [0.2, 0.1, 0.6, 0.4, 1, 0.3, 0.8, 0.5, 0.7, 0.9];
+    partialQuickSort(elements, a => a, 0.1, 1);
+    expect(elements).toStrictEqual(
+      //
```
ClementPasteau, line 47: missing comment here and below too

ClementPasteau:
lgtm overall 👍

newIDE/app/src/UI/Search/UseSearchItem.spec.js
D8H: It's format comments, but the formatter put them on a new line.

newIDE/app/src/UI/Search/UseSearchItem.js
D8H: It's an operation on an array. I don't find it confusing. `push` is also an array operation and it doesn't return a new array.

ClementPasteau: true! but push is called on the object itself as a method, so it warns a bit more that this will affect the object 😇 
I think it's fine either way 👍

ClementPasteau (approved):
good to merge imo

newIDE/app/src/UI/Search/UseSearchItem.spec.js
4ian: `// prettier-ignore` ;)

4ian (approved):
👍

PR #3968, opened 2022-05-29
Fix the move parameter refactor for expressions.

D8H:
The scene or object and behavior parameters were not taking into account so the wrong parameter was moved.

### DIFF
--- a/Core/GDCore/IDE/Events/ExpressionsParameterMover.cpp
+++ b/Core/GDCore/IDE/Events/ExpressionsParameterMover.cpp
@@ -96,7 +96,7 @@ class GD_CORE_API ExpressionParameterMover
       if (behaviorType.empty() && !objectType.empty() &&
           !node.objectName.empty()) {
         // Move parameter of an object function
-        // This refactor only applies on events functions
+        // This refactor only applies on events object functions
         // and events object functions doesn't exist yet.
         // This is a dead code.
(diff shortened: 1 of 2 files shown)


Core/GDCore/IDE/Events/ExpressionsParameterMover.cpp
```diff
@@ -79,37 +79,43 @@ class GD_CORE_API ExpressionParameterMover
...
                             std::move(movedParameterNode));
         };
 
     if (node.functionName == functionName) {
       if (behaviorType.empty() && !objectType.empty() &&
           !node.objectName.empty()) {
         // Move parameter of an object function
+        // This refactor only applies on events functions
+        // and events object functions doesn't exist yet.
+        // This is a dead code.
```
ClementPasteau, line 101: I'm no expert in this part of the code, but if this is dead code, should we remove the lines below and adapt the comment saying this can't happen?

ClementPasteau (commented):
lgtm overall as this is well tested :)

but I'll let @4ian give the final approval

Core/GDCore/IDE/Events/ExpressionsParameterMover.cpp
D8H: A word is missing, it should be "This refactor only applies on events [object] functions..."

4ian (approved):

PR #3970, opened 2022-05-30
Update the mouse buttons status when the cursor comes back inside the game view.

D8H:
Tested on Tristan's game:
https://liluo.io/games/fe638ef5-7609-4072-b0ba-5f7adfaab6ae?dev=true


4ian (approved):
I think this solution is good, we updated as soon as we can and can't do tracking anyway of buttons if the mouse is outside the game view.

PR #3976, opened 2022-06-01
Fix variable editor unmount

### LINKED ISSUE #3967: Object variable does not save as it should
Previous versions of GD, when inserting an object variable and modifying it, whenever we left the object, this variable automatically recorded. Now that doesn't happen anymore, when modifying the value of a variable, and clicking on another object, the variable is reset. Only if we click on something in the panel is recorded.

https://user-images.githubusercontent.com/23488207/170849914-c672120d-6852-4f87-ad94-a804c4bced10.mp4

Notice in the video, that for the variable to be recorded, I have to click above, otherwise the edit is not saved.

AlexandreSi:
Fixes #3967

### DIFF
--- a/newIDE/app/src/VariablesList/VariablesList.js
+++ b/newIDE/app/src/VariablesList/VariablesList.js
@@ -99,6 +99,7 @@ type Props = {|
   emptyPlaceholderTitle?: React.Node,
   emptyPlaceholderDescription?: React.Node,
   helpPagePath?: ?string,
+  commitChangesOnBlur: boolean,
   size?: 'small',
 |};
 
@@ -1078,7 +1079,10 @@ const VariablesList = ({ onComputeAllVariableNames, ...props }: Props) => {
                               disabled={
                                 isCollection || (isInherited && !isTopLevel)
                               }
-                              onChange={() => {}}
+                              onChange={newValue => {
+                                onChangeValue(nodeId, newValue);
+                                forceUpdate();
+                              }}
                               value={
                                 isCollection
                                   ? i18n._(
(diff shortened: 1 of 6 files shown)


ClementPasteau (approved):
lgtm

newIDE/app/src/VariablesList/VariablesList.js
```diff
@@ -94,11 +94,16 @@ export type HistoryHandler = {|
...
   variablesContainer: gdVariablesContainer,
   inheritedVariablesContainer?: gdVariablesContainer,
+  /** Callback executed at mount to compute suggestions. */
   onComputeAllVariableNames?: () => Array<string>,
+  /** To specify if history should be handled by parent. */
   historyHandler?: HistoryHandler,
   emptyPlaceholderTitle?: React.Node,
   emptyPlaceholderDescription?: React.Node,
   helpPagePath?: ?string,
+  /** If set to false, it will commit changes to variables on each input change. It can be expensive, but useful when VariablesList can be unmounted at any time. */
```
4ian, line 104: I was thinking we could even change the SemiControlledTextField to throttle the commits: commit X milliseconds after the last change.
So you get "real time updates" but at the same time not paying for a rendering at each keystroke (only ~500ms after the last change).

Anyway, thanks for the comment, it will work for now!

4ian (approved):

newIDE/app/src/VariablesList/VariablesList.js
AlexandreSi: Oh yes, that could do the trick if there are performance issues!

PR #3978, opened 2022-06-01
Keep assets from a pack in order

D8H:


ClementPasteau (approved):
lgtm 👍

newIDE/app/src/UI/Search/UseSearchItem.js
```diff
@@ -208,6 +208,14 @@ export const useSearchItem = <SearchItem: { tags: Array<string> }>(
     },
     [searchItemsById]
   );
+  const sortedSearchItems: ?Array<SearchItem> = React.useMemo(
+    () => {
+      if (!searchItemsById) return null;
+
+      return Object.keys(searchItemsById).map(id => searchItemsById[id]);
```
ClementPasteau, line 215: I believe you can just return Object.values(searchItemsById) here (and above too)

D8H: Actually, it makes Flow issues, so I reverted the change.

PR #3982, opened 2022-06-02
The asset store remembers the opened page.

D8H:
A more complex state will be needed to handle navigation (to come back to the previous asset details, pack or tag. This PR allows to start on a better base.


ClementPasteau (approved):
lgtm 👍

PR #3983, opened 2022-06-02
Adapt leaderboard admin to new reset API

AlexandreSi:
Leaderboard reset API changed to be asynchronous to fix an issue where leaderboards with too many entries would make the backend timeout.

This PR:
- Make the reset trigger an async process
- Silently refreshes leaderboards every 5s when a leaderboard is resetting

Also:
- When one opens the leaderboard admin from the button in a leaderboard id field, the admin is opened with the selected leaderboard displayed (instead of the default leaderboard)

### DIFF
--- a/newIDE/app/src/EventsSheet/ParameterFields/LeaderboardIdField.js
+++ b/newIDE/app/src/EventsSheet/ParameterFields/LeaderboardIdField.js
@@ -192,6 +192,11 @@ export default React.forwardRef<ParameterFieldProps, ParameterFieldInterface>(
                 onClose={() => setIsAdminOpen(false)}
                 open={isAdminOpen}
                 project={props.project}
+                leaderboardId={
+                  isCurrentValueInLeaderboardList
+                    ? props.value.replace(/"/g, '')
+                    : undefined
+                }
               />
             )}
           </>
--- a/newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
+++ b/newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
@@ -68,7 +68,12 @@ import LeaderboardSortOptionsDialog from './LeaderboardSortOptionsDialog';
 import { type LeaderboardSortOption } from '../../Utils/GDevelopServices/Play';
 import { formatScore } from '../../Leaderboard/LeaderboardScoreFormatter';
 
-type Props = {| onLoading: boolean => void, project?: gdProject |};
+type Props = {|
+  onLoading: boolean => void,
+  project?: gdProject,
+  leaderboardIdToSelectAtOpening?: string,
+|};
+
 type ContainerProps = {| ...Props, gameId: string |};
 
 type ApiError = {|
@@ -172,7 +177,11 @@ const getSortOrderText = (currentLeaderboard: Leaderboard) => {
   return <Trans>Higher is better</Trans>;
 };
 
-export const LeaderboardAdmin = ({ onLoading, project }: Props) => {
+export const LeaderboardAdmin = ({
+  onLoading,
+  project,
+  leaderboardIdToSelectAtOpening,
+}: Props) => {
   const isOnline = useOnlineStatus();
   const windowWidth = useResponsiveWindowWidth();
   const [isEditingAppearance, setIsEditingAppearance] = React.useState<boolean>(
@@ -434,6 +443,14 @@ export const LeaderboardAdmin = ({ onLoading, project }: Props) => {
     // This has to be executed on component mount to refresh entries on each admin opening
     // eslint-disable-next-line
   }, []);
+
+  React.useEffect(
+    () => {
+      if (!!leaderboardIdToSelectAtOpening)
+        selectLeaderboard(leaderboardIdToSelectAtOpening);
+    },
+    [leaderboardIdToSelectAtOpening, selectLeaderboard]
+  );
 
   const onCopy = React.useCallback(
     () => {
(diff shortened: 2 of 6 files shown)


newIDE/app/src/Leaderboard/LeaderboardProvider.js
```diff
@@ -400,6 +403,19 @@ const LeaderboardProvider = ({ gameId, children }: Props) => {
     [currentLeaderboardId, displayOnlyBestEntry, fetchEntries, gameId]
   );
 
+  useInterval(
+    () => {
+      listLeaderboards(true);
+    },
+    !leaderboardsByIds ||
+      Object.values(leaderboardsByIds).every(
+        // $FlowFixMe
```
fannieyan, line 412: Why do we need to/why can we ignore flow here ?

newIDE/app/src/Leaderboard/LeaderboardProvider.js
```diff
@@ -202,11 +205,11 @@ const LeaderboardProvider = ({ gameId, children }: Props) => {
   });
 
   const listLeaderboards = React.useCallback(
-    async () => {
+    async (silent?: boolean) => {
```
fannieyan, line 208: What does "silent" do ? I did not understand why we need to set an interval and fetch leaderboards when we are resetting.

fannieyan (commented):
I have some questions to understand better :)

newIDE/app/src/EventsSheet/ParameterFields/LeaderboardIdField.js
```diff
@@ -192,6 +192,11 @@ export default React.forwardRef<ParameterFieldProps, ParameterFieldInterface>(
                 onClose={() => setIsAdminOpen(false)}
                 open={isAdminOpen}
                 project={props.project}
+                leaderboardId={
+                  isCurrentValueInLeaderboardList
+                    ? props.value.replace(/"/g, '')
```
ClementPasteau, line 197: why do we need to do this?

newIDE/app/src/Leaderboard/LeaderboardProvider.js
```diff
@@ -72,7 +73,9 @@ const reducer = (state: ReducerState, action: ReducerAction): ReducerState => {
       );
       const newCurrentLeaderboard = shouldDefineCurrentLeaderboardIfNoneSelected
         ? primaryLeaderboard || leaderboards[0]
-        : state.currentLeaderboard;
+        : (state.currentLeaderboard &&
+            leaderboardsByIds[state.currentLeaderboard.id]) ||
+          state.currentLeaderboard;
```
ClementPasteau, line 78: hmm, I'm not sure how this logic will behave and why we need it?

newIDE/app/src/GameDashboard/LeaderboardAdmin/index.js
```diff
@@ -605,7 +628,17 @@ export const LeaderboardAdmin = ({ onLoading, project }: Props) => {
...
-      text: (
+      text: currentLeaderboard.resetLaunchedAt ? (
+        <Text size="body2">
+          <Trans>
+            Reset requested the{' '}
+            {i18n.date(currentLeaderboard.resetLaunchedAt, {
+              dateStyle: 'short',
+              timeStyle: 'short',
+            })}
+          </Trans>
```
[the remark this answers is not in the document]
AlexandreSi: Good idea!

newIDE/app/src/Leaderboard/LeaderboardProvider.js
AlexandreSi: I'll try to make this logic clearer

newIDE/app/src/Leaderboard/LeaderboardProvider.js
AlexandreSi: It is ill-named. Without it, when the leaderboards are fetched, when we are waiting for the response, an error is displayed on the UI. So, with it, it's more a `background fetch`

fannieyan:
I just thought, do we display nothing while the leaderboard is resetting ? In my understanding, the entries display will be "frozen" while the leaderboard is resetting, and there will be an additional "reset launched the..." on the reset "tab" ?
I wonder if we could give some more visual cues so that user will understand at first glance that the leaderboard is still resetting ?

AlexandreSi:
There is no freeze of the entries: someone sending a score during a reset could see their score deleted.
I think that a reset is an important action that users keep in mind so I don't think they need another visual indicator.

newIDE/app/src/EventsSheet/ParameterFields/LeaderboardIdField.js
ClementPasteau: 🥐 👍

newIDE/app/src/Leaderboard/LeaderboardProvider.js
```diff
@@ -400,6 +407,40 @@ const LeaderboardProvider = ({ gameId, children }: Props) => {
...
+      previousCurrentLeaderboardResetLaunchedAt.current = currentLeaderboard
+        ? currentLeaderboard.resetLaunchedAt
+        : null;
+    },
+    [currentLeaderboard, currentLeaderboardId, fetchEntries]
+  );
+
+  useInterval(
+    () => {
+      listLeaderboards(true);
```
ClementPasteau, line 433: last nitpick, maybe make this prop a param of an object rather than a random boolean => `listLeaderboards({shouldNotClearBeforeFetching: true})`

ClementPasteau (approved):
lgtm after final nitpick!

PR #3986, opened 2022-06-02
Allow to go back to previous asset pages.

D8H:
This allows to go back to an asset page after exploring tags.
- The filters and textual search are not part of a page state.
- When an asset pack is opened the textual search is ignored and filters are reset. The textual search is applied again when:
  - its validated
  - the back button is used to go back to the initial search result page.
- The history is reset if a textual search happens to avoid inconsistencies as previous textual researches are not save in the history.

### DIFF
--- a/newIDE/app/src/AssetStore/AssetStoreNavigator.js
+++ b/newIDE/app/src/AssetStore/AssetStoreNavigator.js
@@ -1,0 +1,151 @@
+// @flow
+import * as React from 'react';
+import { type FiltersState } from '../UI/Search/FiltersChooser';
+import {
+  type AssetShortHeader,
+  type AssetPack,
+} from '../Utils/GDevelopServices/Asset';
+
+export type AssetStorePageState = {|
+  isOnHomePage: boolean,
+  openedAssetPack: ?AssetPack,
+  openedAssetShortHeader: ?AssetShortHeader,
+  filtersState: FiltersState,
+  ignoreTextualSearch: boolean,
+|};
+
+export type NavigationState = {|
+  previousPages: Array<AssetStorePageState>,
+  getCurrentPage: () => AssetStorePageState,
+  backToPreviousPage: () => void,
+  openHome: () => void,
+  clearHistory: () => void,
+  openSearchIfNeeded: () => void,
+  activateTextualSearch: () => void,
+  openTagPage: string => void,
+  openPackPage: AssetPack => void,
+  openDetailPage: AssetShortHeader => void,
+|};
+
+const noFilter: FiltersState = {
+  chosenCategory: null,
+  chosenFilters: new Set(),
+  addFilter: () => {},
+  removeFilter: () => {},
+  setChosenCategory: () => {},
+};
+
+export const assetStoreHomePageState: AssetStorePageState = {
+  isOnHomePage: true,
+  openedAssetShortHeader: null,
+  openedAssetPack: null,
+  filtersState: noFilter,
+  ignoreTextualSearch: false,
+};
+
+const searchPageState: AssetStorePageState = {
+  isOnHomePage: false,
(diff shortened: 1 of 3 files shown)


newIDE/app/src/AssetStore/AssetStoreNavigator.js
```diff
@@ -0,0 +1,151 @@
...
+    previousPages,
+    getCurrentPage: () => previousPages[previousPages.length - 1],
+    backToPreviousPage: () => {
+      if (previousPages.length > 1) {
+        previousPages.pop();
+        setHistory({ previousPages });
+      }
+    },
+    openHome: () => {
+      previousPages.length = 0;
```
ClementPasteau, line 63: I guess it works, but I've never reset an array by defining its length like that before! (I would have done a `previousPages = [assetStoreHomePageState]`

newIDE/app/src/AssetStore/AssetStoreNavigator.js
```diff
@@ -0,0 +1,151 @@
...
+    },
+    openSearchIfNeeded: () => {
+      const currentPage = previousPages[previousPages.length - 1];
+      if (currentPage.isOnHomePage || currentPage.openedAssetShortHeader) {
+        previousPages.push({
+          isOnHomePage: false,
+          openedAssetShortHeader: null,
+          openedAssetPack: null,
+          filtersState: {
+            chosenCategory: null,
```
ClementPasteau, line 82: can we factor this default state into a variable? something like `emptyFiltersState = { ... }` (because it's used multiple times)

ClementPasteau (approved):
I think it works really well 👍 
see comments to cleanup the code of the different pages so we can reuse most of what we can :)

newIDE/app/src/AssetStore/AssetStoreNavigator.js
ClementPasteau: I don't think it really matters 👍 
I was just surprised this was actually possible

newIDE/app/src/AssetStore/index.js
```diff
@@ -182,77 +179,34 @@ export const AssetStore = ({
...
     if (!assetPacks) return;
 
     sendAssetPackOpened(tag);
 
-    const assetPack = assetPacks.starterPacks.find(
-      assetPack => assetPack.tag === tag
-    );
-    if (assetPack && assetPack.externalWebLink) {
-      Window.openExternalURL(assetPack.externalWebLink);
-      return;
```
D8H, line 208: I removed this, is it okay?
Was it to open the website of the author?

ClementPasteau: not ok! @4ian 

It just got added to test External Paid Packs :)
You need to keep it

D8H: How should it works? I made a change to stays on the discovery page when the user click on it because I guess it only redirect to a website. The asset can't be downloaded or browsed from the app store in this version, right?

newIDE/app/src/AssetStore/index.js
```diff
@@ -182,77 +178,39 @@ export const AssetStore = ({
...
-  // When the back button is pressed, if we were on the asset details page,
-  // we just go back to the search,
-  // if we were on the search, we reset everything to go back to the homepage.
-  const goBack = () => {
-    if (openedAssetShortHeader) {
-      // Going back from Asset page to search.
-      setOpenedAssetShortHeader(null);
+    const assetPack =
+      assetPacks && assetPacks.starterPacks.find(pack => pack.tag === tag);
+    if (assetPack) {
```
ClementPasteau, line 207: sick this help us know if we're on a pack page or not! 🎉

ClementPasteau (approved):
LGTM!

PR #3987, opened 2022-06-03
Add support for displaying external (optionally, paid) asset packs in the Asset Store

4ian:
Also improve the Asset store stories by mocking (some) requests

http://gdevelop-storybook.s3-website-us-east-1.amazonaws.com/feature/external-asset-packs/latest/index.html?path=/story/assetstore-assetstore--default

image

### DIFF
--- a/newIDE/app/src/Utils/GDevelopServices/Asset.js
+++ b/newIDE/app/src/Utils/GDevelopServices/Asset.js
@@ -74,6 +74,8 @@ export type AssetPack = {|
   tag: string,
   thumbnailUrl: string,
   assetsCount: number,
+  externalWebLink?: ?string,
+  userFriendlyPrice?: ?string,
 |};
 
 export type AssetPacks = {|
--- a/newIDE/app/src/stories/componentStories/AssetStore/AssetStore/AssetStore.stories.js
+++ b/newIDE/app/src/stories/componentStories/AssetStore/AssetStore/AssetStore.stories.js
@@ -1,5 +1,6 @@
 // @flow
 import * as React from 'react';
+import withMock from 'storybook-addon-mock';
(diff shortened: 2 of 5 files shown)


newIDE/app/src/Utils/GDevelopServices/Asset.js
```diff
@@ -74,6 +74,8 @@ export type AssetPack = {|
   tag: string,
   thumbnailUrl: string,
   assetsCount: number,
+  externalWebLink?: ?string,
+  userFriendlyPrice?: ?string,
```
ClementPasteau, line 78: is this a good idea to handle the price ourselves? This forces us to monitor packs and update if they do.
Should this be more generic like "Platformer - Paid pack"?

newIDE/app/src/stories/componentStories/AssetStore/AssetStore/AssetStore.stories.js
```diff
@@ -29,3 +53,23 @@ export const Default = () => (
     />
   </AssetStoreStateProvider>
 );
+Default.decorators = [withMock];
+Default.parameters = apiDataFakePacks;
```
ClementPasteau, line 57: 🎉

ClementPasteau (approved):
lgtm overall, just 1 comment in terms of maintainability

newIDE/app/src/Utils/GDevelopServices/Asset.js
4ian: For now I would say it's important because this can makes the difference between clicking on the asset pack to check it or not. As we'll start with paid asset packs made by community members, we can rely on them, by asking paid pack authors to tell us if they change the price :)

PR #3988, opened 2022-06-03
Save template slug in project

fannieyan:
To have some stats on templates used in games

### DIFF
--- a/Core/GDCore/Project/Project.cpp
+++ b/Core/GDCore/Project/Project.cpp
@@ -51,7 +51,7 @@ Project::Project()
     : name(_("Project")),
       version("1.0.0"),
       packageName("com.example.gamename"),
-      templateSlug(NULL),
+      templateSlug(""),
       orientation("landscape"),
       folderProject(false),
       windowWidth(800),
--- a/Core/GDCore/Project/Project.cpp
+++ b/Core/GDCore/Project/Project.cpp
@@ -51,7 +51,7 @@ Project::Project()
     : name(_("Project")),
       version("1.0.0"),
       packageName("com.example.gamename"),
-      templateSlug(""),
+      templateSlug(NULL),
       orientation("landscape"),
       folderProject(false),
       windowWidth(800),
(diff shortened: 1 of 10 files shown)


Core/GDCore/Project/Project.h
```diff
@@ -983,6 +995,7 @@ class GD_CORE_API Project : public ObjectsContainer {
   bool isPlayableWithGamepad;  ///< The project is playable with a gamepad.
   bool isPlayableWithMobile;   ///< The project is playable on a mobile.
   gd::String packageName;  ///< Game package name
+  gd::String templateSlug;
```
AlexandreSi, line 998: Can I suggest a comment that explains what it is?

Core/GDCore/Project/Project.cpp
```diff
@@ -51,6 +51,7 @@ Project::Project()
     : name(_("Project")),
       version("1.0.0"),
       packageName("com.example.gamename"),
+      templateSlug(""),
```
ClementPasteau, line 54: should this be NULL instead of "" ?

Core/GDCore/Project/Project.h
```diff
@@ -983,6 +995,8 @@ class GD_CORE_API Project : public ObjectsContainer {
   bool isPlayableWithGamepad;  ///< The project is playable with a gamepad.
   bool isPlayableWithMobile;   ///< The project is playable on a mobile.
   gd::String packageName;  ///< Game package name
+  gd::String templateSlug; ///< The slug of the template from which is game is
```
ClementPasteau, line 998: typo from which `the` game

Core/GDCore/Project/Project.cpp
4ian: Nope, in C++ a gd::String should always be a string, if you initialise it from a nullptr/NULL (i.e: 0) I guess very bad things can happen ;)

ClementPasteau: Thanks :)

fannieyan: Oops

4ian: Some useful reading on why this would cause an issue: https://stackoverflow.com/questions/49058133/why-doesnt-stdstring-take-a-null-pointer

AlexandreSi (approved):

Issue #3989, opened 2022-06-03
Multiple things 


Bouh:
This is a bug report and a list of thing to polish before the next update.

- [x] In Dark mode, Solarized and new GDevelop default Dark, when I move an event the revamped indicator have a bad contrast (@ClementPasteau fixed)
![image](https://user-images.githubusercontent.com/1670670/171935910-d8e99195-a668-47ab-8adb-449906887f80.png)
![image](https://user-images.githubusercontent.com/1670670/171936242-df69a4ca-f994-4545-94d2-1a5b4ce3cdb6.png)

- [x] With the new theme, panel have no more colors.
![image](https://user-images.githubusercontent.com/1670670/171936724-2decd137-f441-4e9f-a557-3abc03da7a40.png)

- [x] The bold font on the buttons bothers me, I feel like I'm squinting, it's nice, but I feel like it makes me more tired to read.
![image](https://user-images.githubusercontent.com/1670670/171937070-9ac72b40-476b-4b2a-86f1-7f86ce310f98.png)
![image](https://user-images.githubusercontent.com/1670670/171937082-943ec31f-5c14-462a-b865-1a561931be01.png)
![image](https://user-images.githubusercontent.com/1670670/171937095-2464e5a7-71c4-486e-9316-34b4e7d7f844.png)
![image](https://user-images.githubusercontent.com/1670670/171937124-c343f5f3-9d14-47e0-8735-d6c29eeb4550.png)

- [x]  Cancel is unuseable (@ClementPasteau fixed - cannot be canceled so removed button)
![image](https://user-images.githubusercontent.com/1670670/171937436-dcedb6b1-d8fc-4373-8752-778803ea3aab.png)

- [x] The cursor isn't consistent (@ClementPasteau fixed)
![image](https://user-images.githubusercontent.com/1670670/171938387-fe62f399-ddf8-4749-b708-b44b6a32121e.png)

- [x] Click on a tag, and you go to a list of all items with the tag, but this tag isn't in the search bar, or in filters. (@D8H fixed)
![tag](https://user-images.githubusercontent.com/1670670/171938809-41fe3c56-5f73-41e9-bef9-79a79aa958cd.gif)

- [x] I don't remember adding **Laser Shooter** and **Laser** in this **GDevelop Platformer** pack and after a check there is no sprite with this name in the folder. Maybe there is a problem in the asset store. (@D8H fixed)
[...]

PR #3993, opened 2022-06-06
Few fixes Asset store and Events drag and drop

ClementPasteau:
Few fixes listed in #3989 

- Inconsistency of cursor in asset store
- cancel button shown even if not cancelable
- Reset pack disabled for all packs after adding one
- improve style of drop indicator
-

do not show in changelog

### DIFF
--- a/newIDE/app/src/AssetStore/AssetDetails.js
+++ b/newIDE/app/src/AssetStore/AssetDetails.js
@@ -3,7 +3,7 @@ import { t, Trans } from '@lingui/macro';
 import * as React from 'react';
 import { Column, Line, Spacer } from '../UI/Grid';
 import Text from '../UI/Text';
-import Chip from '@material-ui/core/Chip';
+import Chip from '../UI/Chip';
 import RaisedButton from '../UI/RaisedButton';
 import {
   type AssetShortHeader,
--- a/newIDE/app/src/AssetStore/index.js
+++ b/newIDE/app/src/AssetStore/index.js
@@ -235,8 +235,9 @@ export const AssetStore = ({
                   tooltip={t`Back to discover`}
                   onClick={() => {
                     navigationState.openHome();
-                    setIsFiltersPanelOpen(false);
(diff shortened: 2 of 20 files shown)


newIDE/app/src/AssetStore/index.js
```diff
@@ -235,8 +235,9 @@ export const AssetStore = ({
                   tooltip={t`Back to discover`}
                   onClick={() => {
                     navigationState.openHome();
-                    setIsFiltersPanelOpen(false);
                     clearAllFilters(assetFiltersState);
+                    setIsFiltersPanelOpen(false);
+                    setIsAssetPackAdded(false);
                   }}
```
D8H, line 241: After coming back to the pack, it will show as not already added.
Should this boolean be evaluated from what is in he object list?
Do we keep this for a next version?

ClementPasteau: I was thinking of keeping for a next version.
Yes, the button will be enabled again, but the dialog opening up does the logic of comparing the assets and shows a warning "You already have this pack installed, do you want to install it again?"

D8H: ok

D8H (approved):

ClementPasteau:
I tried with the default `style={}` but the Material UI component overrides it 😞

newIDE/app/src/AssetStore/AssetDetails.js
```diff
@@ -3,7 +3,7 @@ import { t, Trans } from '@lingui/macro';
 import * as React from 'react';
 import { Column, Line, Spacer } from '../UI/Grid';
 import Text from '../UI/Text';
-import Chip from '@material-ui/core/Chip';
+import Chip from '../UI/Chip';
```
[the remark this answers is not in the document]
ClementPasteau: yeah that'd be great, we have other components like that too.

4ian:
Ok, thanks for confirming that inline styles would not work there! 👍

PR #3994, opened 2022-06-06
Set search bars to be rounded in the modern theme

4ian:
* Search bar have rounding and margins in most dialogs. Margins are better because they are following the asset store design.
* In lists, they stay "integrated" (no rounding, no margins)

image
image
image
image

Also fix an icon that was too big.

Don't show in changelog


AlexandreSi (approved):
LGTM! Looks nice like that

ClementPasteau (approved):
cool!

PR #3996, opened 2022-06-06
Add Camera zoom condition

Elairyx7301:
I'm *(still)* working with the most basic tools when doing PRs etc. and I hope I will be able to change that soon but until then I'll have to bother you with a bit chaos. Also with that, I can't test the changes right away, sorry about all that.
I have promised to work on this for quite some time so I wanted to give it a go.

@\/AlexandreSi has already told me that the following files need changes:
[Core/GDCore/Extensions/Builtin/CameraExtension.cpp](https://github.com/4ian/GDevelop/blob/master/Core/GDCore/Extensions/Builtin/CameraExtension.cpp)
[GDJS/GDJS/Extensions/Builtin/CameraExtension.cpp](https://github.com/4ian/GDevelop/blob/master/GDJS/GDJS/Extensions/Builtin/CameraExtension.cpp)
[GDJS/Runtime/events-tools/cameratools.ts](https://github.com/4ian/GDevelop/blob/master/GDJS/Runtime/events-tools/cameratools.ts)
but I couldn't find what to change in **.../events-tools/cameratools.ts**, since all I could find was getters and setters and I see them not connected if the element is an action or condition or expression and [`export const getCameraZoom`](https://github.com/4ian/GDevelop/blob/master/GDJS/Runtime/events-tools/cameratools.ts#L132) is already available in the file.

What am I missing?


Elairyx7301:
I hate everything about this **:-x**

Elairyx7301:
See only those commits
[7033138](https://github.com/4ian/GDevelop/pull/3996/commits/703313885ab6c441e3d5f7f29a22c8e29919a281)
[ab3e10b](https://github.com/4ian/GDevelop/pull/3996/commits/ab3e10b4e2e7a3b127669cefbe73fa2cbba93dba)

4ian:
You're on the good track :)  There is indeed nothing to change in cameratool.js, as the condition you're adding is using the same JavaScript function as the expression.

Were you able to test that your new condition is working?

Elairyx7301:
I'm sorry, I was not able to test it. I'm not using any tools for PRs other than the GitHub web version and I'm close to moving out so I will set up a proper workflow etc. on the new computer hopefully very soon.

PR #3999, opened 2022-06-07
Add camera borders expressions and conditions

### LINKED ISSUE #3990: Add new expressions: CameraLeft(), CameraRight(), CameraTop(), CameraBottom()
## Description

The new `Object.BoundingBoxTop()` family of expressions has been very useful and saved me a lot of time.

The borders of cameras would provide a similar benefit.  These would allow users to replace this:

`CameraX("Layer", Camera) - CameraWidth("Layer", Camera)/2`

with this:

`CameraLeft("Layer", Camera)
`
## Solution suggested

Add new expressions for each side of the camera:

- [ ] `CameraTop("Layer", Camera)`
- [ ] `CameraBottom("Layer", Camera)`
- [ ] `CameraLeft("Layer", Camera)`
- [ ] `CameraRight("Layer", Camera)`

fannieyan:
Closes #3990

- Added expressions `CameraCenterX` and  `CameraCenterY` and hid `CameraX` and `CameraY`
- Add `CameraLeft`, `CameraRight`, `CameraBottom`, `CameraTop` expressions and conditions

### DIFF
--- a/GDJS/GDJS/Extensions/Builtin/CameraExtension.cpp
+++ b/GDJS/GDJS/Extensions/Builtin/CameraExtension.cpp
@@ -26,10 +26,10 @@ CameraExtension::CameraExtension() {
   GetAllActions()["SetCameraY"]  // Deprecated
       .SetFunctionName("gdjs.evtTools.camera.setCameraY")
       .SetGetter("gdjs.evtTools.camera.getCameraY");
-  GetAllActions()["CameraCenterX"]
+  GetAllActions()["SetCameraCenterX"]
       .SetFunctionName("gdjs.evtTools.camera.setCameraX")
       .SetGetter("gdjs.evtTools.camera.getCameraX");
-  GetAllActions()["CameraCenterY"]
+  GetAllActions()["SetCameraCenterY"]
       .SetFunctionName("gdjs.evtTools.camera.setCameraY")
       .SetGetter("gdjs.evtTools.camera.getCameraY");
 
--- a/GDJS/GDJS/Extensions/Builtin/CameraExtension.cpp
+++ b/GDJS/GDJS/Extensions/Builtin/CameraExtension.cpp
@@ -45,14 +45,14 @@ CameraExtension::CameraExtension() {
       "gdjs.evtTools.camera.getCameraWidth");
   GetAllConditions()["CameraHeight"].SetFunctionName(
       "gdjs.evtTools.camera.getCameraHeight");
-  GetAllConditions()["CameraLeft"].SetFunctionName(
-      "gdjs.evtTools.camera.getCameraLeft");
-  GetAllConditions()["CameraRight"].SetFunctionName(
-      "gdjs.evtTools.camera.getCameraRight");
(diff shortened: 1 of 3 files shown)


GDJS/Runtime/events-tools/cameratools.ts
```diff
@@ -74,6 +74,62 @@ namespace gdjs {
...
+      export const getCameraBottom = function (
+        runtimeScene: gdjs.RuntimeScene,
+        layer: string,
+        cameraId: integer
+      ): number {
+        if (!runtimeScene.hasLayer(layer)) {
+          return 0;
+        }
+        return (
+          getCameraY(runtimeScene, layer, cameraId) -
```
AlexandreSi, line 128: Should be a plus here I think

fannieyan:
> I wonder if we should have a more specific naming: CameraLeftBorder or something else? I think there are already a lot of conditions/actions regarding camera, camera position limits and else.

I think it wouldn't hurt to be more specific on the expression name, yes!

tristanbob:
> * Rename expressions `CameraX` to `CameraCenterX` and `CameraY` to `CameraCenterY`

Don't we need to keep these for backward compatibility?  I think 4ian suggested hiding the old versions and adding new expressions.

fannieyan:
@tristanbob 

> Don't we need to keep these for backward compatibility? I think 4ian suggested hiding the old versions and adding new expressions.

Yes my bad on the wording, I actually added those and hid the old version ;)

> If we want to make them more descriptive, can I suggest swapping the word order to this: CameraBorderLeft?
This matches the BoundingBoxLeft style.

Makes sense! Will do :)

GDJS/GDJS/Extensions/Builtin/CameraExtension.cpp
```diff
@@ -20,20 +20,39 @@ CameraExtension::CameraExtension() {
...
       .SetFunctionName("gdjs.evtTools.camera.setCameraX")
       .SetGetter("gdjs.evtTools.camera.getCameraX");
-  GetAllActions()["SetCameraY"]
+  GetAllActions()["SetCameraY"]  // Deprecated
       .SetFunctionName("gdjs.evtTools.camera.setCameraY")
       .SetGetter("gdjs.evtTools.camera.getCameraY");
+  GetAllActions()["CameraCenterX"]
+      .SetFunctionName("gdjs.evtTools.camera.setCameraX")
+      .SetGetter("gdjs.evtTools.camera.getCameraX");
+  GetAllActions()["CameraCenterY"]
```
4ian, line 32: This is not obvious (I got bitten already by this), but this should be:

```suggestion
  GetAllActions()["SetCameraCenterX"]
      .SetFunctionName("gdjs.evtTools.camera.setCameraX")
      .SetGetter("gdjs.evtTools.camera.getCameraX");
  GetAllActions()["SetCameraCenterY"]
```

as the convention, enforced by AddExpressionAndConditiondAndAction is that an action starts with a verb: `Set`

4ian (approved):
Make sure to double check the new actions are working, I think they might not be.

PR #4000, opened 2022-06-07
[Physics] Add Mass and Inertia expressions.

D8H:
Theses expression are useful to set a speed with impulses because the speed must be multiplied with these values to do so.
It also fixes a bad copy/paste in the unit of impulses actions.

I saw it requested by users some month ago.

I tested the 2 expression on Tristan's example.


AlexandreSi:
LGTM but I never worked on the Physics Behavior so hard for me to approve :/

tristanbob:
Does this mean if a condition checks the mass every frame, those instances will never sleep?

Can you record if it was sleeping so you can put it back to sleep? Or will that not have the desired effect of saving CPU

D8H:
> Does this mean if a condition checks the mass every frame, those instances will never sleep?
> 
> Can you record if it was sleeping so you can put it back to sleep? Or will that not have the desired effect of saving CPU

This expression is only useful when an impulse needs to be applied and the impulse also wake it up.

PR #4001, opened 2022-06-07
Add hidden "advanced" section to enter data urls as resource

### LINKED ISSUE #3984: Base64 strings as sprites
## Description

It would be nice if we could add sprites not only from file system or from URL, but from Base64 strings as well

AlexandreSi:
Closes #3984

demo_url_chooser

### DIFF
--- a/newIDE/app/src/ResourcesList/BrowserResourceSources.js
+++ b/newIDE/app/src/ResourcesList/BrowserResourceSources.js
@@ -9,18 +9,37 @@ import {
 } from './ResourceSource';
 import { ResourceStore } from '../AssetStore/ResourceStore';
 import path from 'path';
-import { Line } from '../UI/Grid';
+import { Column, Line } from '../UI/Grid';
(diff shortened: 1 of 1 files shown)


4ian:
This is a new feature that we engage ourselves to support forever, so an additional multiplier to the complexity of the resource management. I'm ok with this if we display a warning and maybe even refuse anything larger than X kb. 
Otherwise, Murphy's law/big number law, we will end up with people stuffing 1MB+ resources in their game and complaining "why did GDevelop/my browser/the backend crashed?"

AlexandreSi:
I thought it was supported but not clear how to use.

In that case, maybe it's better not merging that and just answering to the issue that this is a hack that works but that might break in the future (but hopefully it won't be needed in the future).

4ian:
Yes, I've commented on the thread! Let's hold this then, unless we get massive support for this or that we decide it's reasonable to support this officially.

PR #4005, opened 2022-06-08
howler-sound-manager: Preload sounds sequentially

dos1:
Trying to preload all sounds at once can overwhelm the browser
in case of projects with many big sound files. Load the files
sequentially, the same way images are being preloaded.


4ian (approved):
Sounds fair. We could do something more complex like loading a max of 2, 3, ... N files at once, but going one by one is for now surely better than loading everything at once. Thanks!

dos1:
> We could do something more complex like loading a max of 2, 3, ... N files at once

With this code, it should be as easy as calling `loadNextFile` multiple times.

PR #4008, opened 2022-06-08
Rework Asset Pack add to a list of filtered assets

ClementPasteau:


newIDE/app/src/AssetStore/index.js
```diff
@@ -308,17 +305,13 @@ export const AssetStore = ({
...
                               primary
-                              label={
-                                isAssetPackAdded ? (
-                                  <Trans>Asset pack added</Trans>
-                                ) : (
-                                  <Trans>Add pack to my scene</Trans>
-                                )
-                              }
+                              label={<Trans>Add all assets to my scene</Trans>}
                               onClick={() =>
```
D8H, line 309: The text doesn't change because we don't know wether all assets are already added or not?

D8H (approved):

newIDE/app/src/AssetStore/index.js
ClementPasteau: Yes, we don't know at this stage, we do the logic in the dialog that opens up

PR #4011, opened 2022-06-09
Fix music files staying muted after loading

dos1:
This is caused by a bug in recent versions of Howler.js. Since
there's no fix in Howler yet, this change reverts the problematic
commit (https://github.com/goldfire/howler.js/commit/0323af9b843351cacc47893aed8e65cd8ba3b0cb)
on top of v2.2.3 release.

See https://github.com/goldfire/howler.js/issues/1603 for more details.


4ian:
Could you describe what's the issue visible in GDevelop? Music files would simply not be heard at all because they would stay at volume 0?

dos1:
Yup, that's what I've seen - music files often played at 0 volume regardless of what was being set in PlayMusic event. You could see the volume event being stored in Howl's `_queue` array, but it was never actually applied unless you called `_loadQueue` manually. It's essentially a race condition, so it sometimes played correctly, but usually did not.

PR #4012, opened 2022-06-09
Fix camera actions/conditions/expressions broken since introducing new expressions

4ian:
(see https://github.com/4ian/GDevelop/pull/3999)

Removed some logs so that it's easier to spot errors reported.


ClementPasteau (approved):
lgtm

PR #4013, opened 2022-06-09
Add preference toggle to suggest all expressions regardless of type

fannieyan:
- Add toggle in Preferences to suggest all expressions regardless of type
image

Now:
(Note that AnimationName returns a string, and Animation returns a number)
image

### DIFF
--- a/Core/GDCore/IDE/Events/ExpressionCompletionFinder.h
+++ b/Core/GDCore/IDE/Events/ExpressionCompletionFinder.h
@@ -309,12 +309,8 @@ class GD_CORE_API ExpressionCompletionFinder
 
     gd::ExpressionNode* maybeParentNodeAtLocation = finder.GetParentNode();
     gd::ExpressionCompletionFinder autocompletionProvider(
-        platform,
-        globalObjectsContainer,
-        objectsContainer,
-        rootType,
-        searchedPosition,
-        maybeParentNodeAtLocation);
+        platform, globalObjectsContainer, objectsContainer, rootType,
+        searchedPosition, maybeParentNodeAtLocation);
     nodeAtLocation->Visit(autocompletionProvider);
     return autocompletionProvider.GetCompletionDescriptions();
   }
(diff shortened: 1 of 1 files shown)


newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/FormatExpressionCall.js
```diff
@@ -117,3 +117,5 @@ export const formatExpressionCall = (
     return `${functionName}(${functionArgs})`;
   }
 };
+
+// export const wrapExpressionCallWithTypeChange = ()
```
ClementPasteau, line 121: ?

newIDE/app/src/EventsSheet/ParameterFields/FunctionParameterNameField.js
```diff
@@ -9,7 +11,7 @@ export default class FunctionParameterNameField extends Component<
   ParameterFieldProps,
   void
 > {
-  _field: ?GenericExpressionField;
+  _field: ?ExpressionFieldInterface;
```
ClementPasteau, line 14: 😍

newIDE/app/src/MainFrame/Preferences/PreferencesContext.js
```diff
@@ -287,6 +289,7 @@ export const initialPreferences = {
     isAlwaysOnTopInPreview: false,
     backdropClickBehavior: 'nothing',
     eventsSheetCancelInlineParameter: 'apply',
+    eventsSheetUseAllExpressionTypes: false,
```
ClementPasteau, line 292: I'm not sure I agree with the decision to make this prop false by default - meaning it's hardly an improvement from before.
For newcomers, it's really confusing/blocking the first time you use expressions.
Ex: I want to display the speed of an element in a text.
- if set to false, I don't see the `object.GetVelocityX()` autocompletion and I don't understand why, so I don't use it. (and maybe later I realise I can actually use it, but I have to know the expression by heart to type it)
- if set to true, I see the `object.getVelocityX()` autocompletion and I use it, then I get an error telling me I have to use toString() -> it teaches me that they are different types of variables and I have to understand the difference between number and string.

I definitely prefer the second option

fannieyan:
After discussion, I will remove the preference toggle and go with the strategy of wrapping the suggestion with a ToString() when necessary

fannieyan:
Closed and redirected to #4033

PR #4016, opened 2022-06-10
Fix asset thumbnail having a wrong size when navigating the Asset store

ClementPasteau:
This fixes a few things:

- when going back from one asset to the other, we reset the asset to allow setting the zoom again + seeing a loader
- when the screen is really small, the search results were not displayed at all and couldn't be accessed
- when an asset loads, sometimes a scroll is seen and the image jumps

https://user-images.githubusercontent.com/4895034/173111248-f5001da4-82c5-4981-a474-7da6c75f7744.mov

### DIFF
--- a/newIDE/app/src/AssetStore/AssetDetails.js
+++ b/newIDE/app/src/AssetStore/AssetDetails.js
@@ -129,6 +129,8 @@ export const AssetDetails = ({
     () => {
       (async () => {
         try {
+          // Reinitialise asset to trigger a loader and recalculate all parameters. (for instance zoom)
+          setAsset(null);
           const loadedAsset = await getAsset(assetShortHeader);
           setAsset(loadedAsset);
           if (loadedAsset.objectType === 'sprite') {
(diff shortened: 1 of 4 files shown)


AlexandreSi:
LGTM but I'll let someone else approve

newIDE/app/src/AssetStore/AssetDetails.js
```diff
@@ -129,6 +130,8 @@ export const AssetDetails = ({
     () => {
       (async () => {
         try {
+          // Reinitialise asset to trigger a loader and recalculate all parameters. (for instance zoom)
+          setAsset(null);
           const loadedAsset = await getAsset(assetShortHeader);
           setAsset(loadedAsset);
```
D8H, line 136: Why wasn't the new asset size taken into account without it?

ClementPasteau: The ImagePreview component initialize a state value for its zoom based on the props, but does not update this state value when the prop changes.
We can either:
- Never clear the asset (just replace it) meaning that we don't see a loader (the network call happens in the background) when a user clicks on an asset or when they click "back"
- Always clear the asset, triggering a loader that can be really short or long depending on network connection.

I feel like solution 1 could be confusing if a user has a slow connection, they wouldn't have the feedback that their asset is currently loading

4ian (approved):

PR #4018, opened 2022-06-10
Detect all physics collisions

### LINKED ISSUE #4009: Physics behavior not tracking collisions properly.
When collisions happen between frames the physics behavior doesn't count them as collisions, and therefore developers who expect a sound effect or value to change upon collision will be sorely upset.

## To Reproduce
Just play the game at 20 min fps and 20 max fps and compare it to 60/60. There will be missing counts.

[Plinko.zip](https://github.com/4ian/GDevelop/files/8863544/Plinko.zip)

AlexandreSi:
Fixes #4009


Extensions/Physics2Behavior/JsExtension.js
```diff
@@ -3734,7 +3735,7 @@ module.exports = {
         _('Collision'),
         _('Test if two objects collide.'),
         _('_PARAM0_ is colliding with _PARAM2_'),
-        '',
+        _('Collisions'),
         'res/physics32.png',
```
D8H, line 3739: The advantage to leave no category is that it is on top of the list. It makes sens as it might be the most used condition.
The group "Collision" (without an "s") could be used to merge it with the built-in conditions.

Extensions/Physics2Behavior/JsExtension.js
```diff
@@ -3746,6 +3747,42 @@ module.exports = {
       .setIncludeFile('Extensions/Physics2Behavior/physics2tools.js')
       .setFunctionName('gdjs.physics2.objectsCollide');
 
+    extension
+      .addCondition(
+        'CollisionStarted',
+        _('Collision start'),
```
D8H, line 3753: Why not "Collision started"?

Extensions/Physics2Behavior/JsExtension.js
```diff
@@ -3746,6 +3747,42 @@ module.exports = {
...
+      .addCodeOnlyParameter('conditionInverted', '')
+      .getCodeExtraInformation()
+      .setIncludeFile('Extensions/Physics2Behavior/physics2tools.js')
+      .setFunctionName('gdjs.physics2.objectsStartedColliding');
+
+    extension
+      .addCondition(
+        'CollisionStopped',
+        _('Collision stop'),
+        _('Test if two objects stopped colliding.'),
```
D8H, line 3772: "Test" is used in some conditions but should we replace them by "Check"?

Extensions/Physics2Behavior/physics2runtimebehavior.ts
```diff
@@ -48,13 +48,25 @@ namespace gdjs {
...
-        behaviorB.currentContacts.push(behaviorA);
+        // There might be contacts that end during the frame and
+        // start again right away. It is considered a glitch
+        // and should not be detected.
+        let i = behaviorA.contactsEndedBetweenFrames.indexOf(behaviorB);
+        if (i !== -1) {
+          behaviorA.contactsEndedBetweenFrames.splice(i, 1);
+        } else {
+          behaviorA.contactsStartedBetweenFrames.push(behaviorB);
+        }
```
D8H, line 63: It could be a `onContactBegin` method in the behavior.

Extensions/Physics2Behavior/physics2runtimebehavior.ts
```diff
@@ -281,7 +288,13 @@ namespace gdjs {
...
       this.layers = behaviorData.layers;
       this.masks = behaviorData.masks;
-      this.currentContacts = this.currentContacts || [];
+      this.contactsStartedBetweenFrames =
+        behaviorData.contactsStartedBetweenFrames || [];
+      this.contactsStartedBetweenFrames.length = 0;
+      this.contactsEndedBetweenFrames =
+        behaviorData.contactsEndedBetweenFrames || [];
+      this.contactsEndedBetweenFrames.length = 0;
+      this.currentContacts = behaviorData.currentContacts || [];
```
D8H, line 297: `behaviorData` is for properties.

Extensions/Physics2Behavior/physics2runtimebehavior.ts
```diff
@@ -741,11 +754,27 @@ namespace gdjs {
...
     }
 
     doStepPostEvents(runtimeScene) {
       this._updateBodyFromObject();
 
+      // Reset contacts that happened in-between frames
+      this.contactsStartedBetweenFrames.splice(
+        0,
+        this.contactsStartedBetweenFrames.length
+      );
```
D8H, line 772: Setting the length to `0` is easier to read (maybe faster but I'm not sure)

Extensions/Physics2Behavior/physics2runtimebehavior.ts
```diff
@@ -3813,6 +3842,70 @@ namespace gdjs {
...
+        // true anyway.
+        for (
+          let i = 0, len = behavior1.contactsStartedBetweenFrames.length;
+          i < len;
+          ++i
+        ) {
+          if (behavior1.contactsStartedBetweenFrames[i].owner === object2) {
+            return true;
+          }
+        }
```
D8H, line 3856: `Array.some` could be used.

Extensions/Physics2Behavior/JsExtension.js
```diff
@@ -3746,6 +3747,42 @@ module.exports = {
       .setIncludeFile('Extensions/Physics2Behavior/physics2tools.js')
       .setFunctionName('gdjs.physics2.objectsCollide');
 
+    extension
+      .addCondition(
+        'CollisionStarted',
+        _('Collision start'),
+        _('Test if two objects started colliding.'),
```
4ian, line 3754: We generally now prefer "Check" or "Compare" over "Test".

```suggestion
        _('Check if two objects started colliding.'),
```

Extensions/Physics2Behavior/physics2runtimebehavior.ts
4ian: Yes no need to initialise from anything, just set it to `[]`.

Extensions/Physics2Behavior/physics2runtimebehavior.ts
```diff
@@ -3813,6 +3842,70 @@ namespace gdjs {
...
+      }
+
+      // No contact found
+      return false;
+    }
+
+    // Collision
+    static stoppedCollisionTest(
```
4ian, line 3890: ditto

Extensions/Physics2Behavior/physics2tools.ts
```diff
@@ -15,6 +15,36 @@ namespace gdjs {
       );
     };
 
+    export const objectsStartedColliding = function (
```
4ian, line 18: haveObjectsStartedColliding?
and (haeObjectsStoppedColliding)?

Extensions/Physics2Behavior/JsExtension.js
```diff
@@ -3746,6 +3747,42 @@ module.exports = {
...
+      .addParameter('objectList', _('Object'), '', false)
+      .addCodeOnlyParameter('conditionInverted', '')
+      .getCodeExtraInformation()
+      .setIncludeFile('Extensions/Physics2Behavior/physics2tools.js')
+      .setFunctionName('gdjs.physics2.objectsStartedColliding');
+
+    extension
+      .addCondition(
+        'CollisionStopped',
+        _('Collision stop'),
```
4ian, line 3771: `Collision stopped`?

Extensions/Physics2Behavior/physics2runtimebehavior.ts
4ian: Yep, use `length = 0`, this is handled well by JS engines.

Extensions/Physics2Behavior/physics2runtimebehavior.ts
```diff
@@ -241,6 +246,8 @@ namespace gdjs {
     layers: any;
     masks: any;
     shapeScale: number = 1;
+    contactsStartedBetweenFrames: Array<any>;
+    contactsEndedBetweenFrames: Array<any>;
```
4ian, line 250: Because there is nothing really "between frames" (something is always computed as part of the rendering of a frame), I would rather name these:

```suggestion
    contactsStartedThisFrame: Array<any>;
    contactsEndedThisFrame: Array<any>;
```

or 

```suggestion
    contactsJustStarted: Array<any>;
    contactsJustEnded: Array<any>;
```

Indeed, the simulation of the physics engine is done just before the events (in the do pre events). So the contacts are really detected as starting/ending when simulating *this* frame?

Wdyt @D8H @AlexandreSi ?

Extensions/Physics2Behavior/tests/utils.spec.js
```diff
@@ -0,0 +1,108 @@
...
+      items,
+      startedItems,
+      endedItems
+    );
+    for (let i = 0; i < items.length; i++) {
+      expect(items[i]).to.be(expectedResolvedItems[i]);
+    }
+  });
+
+  it('It returns current items with item started twice', () => {
```
4ian, line 79: ubernit:
```suggestion
  it('returns current items with item started twice', () => {
```

Extensions/Physics2Behavior/tests/utils.spec.js
```diff
@@ -0,0 +1,108 @@
...
+    const items = ['A', 'B', 'C'];
+    const startedItems = ['Z', 'Z'];
+    const endedItems = ['Z'];
+    const expectedResolvedItems = [...items, 'Z'];
+    gdjs.physics2.computeCurrentContactsFromStartedAndEndedContacts(
+      items,
+      startedItems,
+      endedItems
+    );
+    for (let i = 0; i < items.length; i++) {
```
4ian, line 89: Try (I'm not sure): `expect(items).to.eql(expectedResolvedItems)`

Extensions/Physics2Behavior/physics2runtimebehavior.ts
```diff
@@ -3813,6 +3842,70 @@ namespace gdjs {
             return true;
           }
         }
+        // If a contact has started between frames and ended right away, it
```
4ian, line 3845:
```suggestion
        // If a contact has started at this frame and ended right away, it
```

Extensions/Physics2Behavior/physics2runtimebehavior.ts
AlexandreSi: Great idea, I'll do this

Extensions/Physics2Behavior/utils.ts
```diff
@@ -0,0 +1,25 @@
+namespace gdjs {
+  export namespace physics2 {
+    export const computeCurrentContactsFromStartedAndEndedContacts = <T>(
+      current: Array<T>,
+      started: Array<T>,
+      ended: Array<T>
+    ): void => {
```
D8H, line 7: It was extracted to be able to test it?

Extensions/Physics2Behavior/physics2runtimebehavior.ts
D8H: The arrays can be cleared at the beginning of `doStepPreEvents`. It will allow to write tests more easily. It won't change how it behaves as these arrays are not used during the rending.

4ian (approved):
LGTM

Extensions/Physics2Behavior/tests/physics2runtimebehavior.spec.js
```diff
@@ -0,0 +1,128 @@
+// TODO: find a way to add assertions at events time to enable this test.
+describe.skip('Physics2RuntimeBehavior', () => {
```
D8H, line 2: This needs to be addressed.
Arrays clearing can be done at the beginning of the preEvent function.

AlexandreSi: The conditions use the functions `hasCollisionStartedBetween` and `hasCollisionStoppedBetween`, and themselves use the arrays `contactsStartedThisFrame` and `contactsStoppedThisFrame` so I can't erase them in the pre events, or is there something that I'm missing?

D8H: It can be in this order:
- preEvent
  - clear the arrays
  - step Box2d (this fill the arrays)
- events (conditions use the arrays)
loop...

Extensions/Physics2Behavior/tests/utils.spec.js
```diff
@@ -0,0 +1,94 @@
+// @ts-check
+
+describe('computeCurrentContactsFromStartedAndEndedContacts', () => {
+  it('returns same current items is nothing happened', () => {
```
fannieyan, line 4:
```suggestion
  it('returns same current items if nothing happened', () => {
```

fannieyan (commented):
Very interesting read!

4ian (approved):
lgtm

Extensions/Physics2Behavior/physics2runtimebehavior.ts
```diff
@@ -3797,27 +3811,89 @@ namespace gdjs {
...
+      // There might be contacts that end during the frame and
+      // start again right away. It is considered a glitch
+      // and should not be detected.
+      let i = this.contactsEndedThisFrame.indexOf(otherBehavior);
+      if (i !== -1) {
+        this.contactsEndedThisFrame.splice(i, 1);
+      } else {
+        this.contactsStartedThisFrame.push(otherBehavior);
+      }
+    }
```
D8H, line 3824: I wonder if it could be interesting to test it as it needed a comment, but I don't see how to write a good test. It will be hard to cover this case with actual physics. I guess, it could be called directly by the test.

D8H (approved):

Extensions/Physics2Behavior/utils.ts
```diff
@@ -0,0 +1,25 @@
...
+        }
+      });
+      endedDuplicate.forEach((endedItem) => {
+        const index = current.indexOf(endedItem);
+        if (index !== -1) {
+          current.splice(index, 1);
+        }
+      });
+    };
+  }
```
D8H, line 24: I think that this method could be much simpler.

End events are eaten by a following start event.
```
[end start]
start [end start]
start [end start] end
[end start] end
```
This means that a `start` is always before a `end`. It could be 2 simple steps:
- add started contacts
- remove ended contacts

Extensions/Physics2Behavior/tests/utils.spec.js
```diff
@@ -0,0 +1,94 @@
...
+      endedItems
+    );
+    expect(items).to.eql(expectedResolvedItems);
+  });
+
+  it('It returns current items without item ended twice', () => {
+    const items = ['A', 'B', 'C'];
+    const startedItems = ['C'];
+    const endedItems = ['C', 'C'];
+    const expectedResolvedItems = ['A', 'B'];
```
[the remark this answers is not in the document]
AlexandreSi: I'm not sure I understood all you meant, but I pushed another version, let me know if it's okay for you

Extensions/Physics2Behavior/utils.ts
```diff
@@ -0,0 +1,25 @@
+namespace gdjs {
+  export namespace physics2 {
+    export const computeCurrentContactsFromStartedAndEndedContacts = <T>(
+      current: Array<T>,
+      started: Array<T>,
+      ended: Array<T>
+    ): void => {
+      const endedDuplicate = [...ended];
```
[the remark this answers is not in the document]
AlexandreSi: I didn't know about it, thanks!

Extensions/Physics2Behavior/utils.ts
AlexandreSi: I'll dig into it

Extensions/Physics2Behavior/tests/utils.spec.js
```diff
@@ -0,0 +1,104 @@
...
+    );
+    expect(contacts).to.eql(expectedResolvedContacts);
+  });
+
+  it('returns current contacts without a contact that ended and also jittered', () => {
+    // Should handle cases where contact C was here and, during the frame:
+    // - contact C ends
+    // - contact C starts
+    // - contact C ends
+    // Contact C should not appear in the current contacts
```
D8H, line 92: Same.

Extensions/Physics2Behavior/tests/physics2runtimebehavior.spec.js
```diff
@@ -0,0 +1,269 @@
...
+  });
+
+  describe('onContactBegin', () => {
+    let runtimeGame;
+    let runtimeScene;
+    beforeEach(() => {
+      [runtimeGame, runtimeScene] = createGameWithSceneWithPhysics2SharedData();
+    });
+
+    it('should add behavior to list of started contacts', () => {
```
D8H, line 225: a description like:
"can detect a contact that start during a physic sub steps"
?

Maybe this test could do a "end" after the "beginning" to show that it's keep in this order?
Something like "detect a contact that is maintained only during some physics intermediate sub-steps" ?

Extensions/Physics2Behavior/tests/utils.spec.js
```diff
@@ -0,0 +1,104 @@
...
+      startedContacts,
+      endedContacts
+    );
+    expect(contacts).to.eql(expectedResolvedContacts);
+  });
+
+  it('returns current contacts with ended contacts removed', () => {
+    const contacts = ['A', 'B', 'C', 'Z'];
+    const startedContacts = [];
+    const endedContacts = ['Z', 'B', 'R'];
```
D8H, line 33: The 'R' can't be removed if it doesn't exist. Every case that can't happens should probably be in a specific describe to make it easier for the reader to understand how the behavior works.

AlexandreSi: I'm not keen on listing things that can't happen somewhere because some things will be forgotten and it can be misleading in the future.

PR #4021, opened 2022-06-11
Add WebManifest generation to builds

arthuro555:
A first baby step towards making the exported games working PWAs 💪 Generates a WebManifest from project data.  

The WebManifest is only made for builds, not previews, since:
1. A preview shouldn't be installable
2. The manifest isn't working there anyways, as icons links are not allowed to be file URLs

As a follow-up, we could add support for the following fields that require being added to the project properties first:
- Background color
- Theme color
- Short name
- Screenshots
- Language

Those properties could also be useful for liluo 😉

### DIFF
--- a/GDJS/GDJS/IDE/ExporterHelper.cpp
+++ b/GDJS/GDJS/IDE/ExporterHelper.cpp
@@ -904,6 +904,7 @@ const gd::String ExporterHelper::GenerateWebManifest(
   "start_url": "./index.html",
   "display": "standalone",
   "background_color": "black",
+  "categories": ["games", "entertainment"],
   "icons": {ICONS}
 })webmanifest")
       .FindAndReplace("{NAME}", project.GetName())
--- a/GDJS/GDJS/IDE/ExporterHelper.cpp
+++ b/GDJS/GDJS/IDE/ExporterHelper.cpp
@@ -858,28 +858,28 @@ const gd::String ExporterHelper::GenerateWebManifest(
 
   {
     std::map<int, gd::String> resourcesForSizes;
-    const auto FileNameForIcon = [&project](const gd::String &platform,
-                                            const int size) {
-      return project.GetPlatformSpecificAssets().Has(
-                 platform, "icon-" + gd::String::From(size))
+    const auto getFileNameForIcon = [&project](const gd::String &platform,
+                                               const int size) {
+      const gd::String iconName = "icon-" + gd::String::From(size);
+      return project.GetPlatformSpecificAssets().Has(platform, iconName)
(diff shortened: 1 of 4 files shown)


GDJS/GDJS/IDE/ExporterHelper.cpp
```diff
@@ -844,4 +845,74 @@ void ExporterHelper::AddDeprecatedFontFilesToFontResources(
...
+const std::array<int, 6> ANDROID_ICONS_SIZES = {36, 48, 72, 96, 144, 192};
+
+const gd::String ExporterHelper::GenerateWebManifest(
+    const gd::Project &project) {
+  const gd::String &orientation = project.GetOrientation();
+  gd::String icons = "[";
+
+  {
+    std::map<int, gd::String> resourcesForSizes;
+    const auto FileNameForIcon = [&project](const gd::String &platform,
```
4ian, line 861: Uber nitpicking:

this is a local variable, so camelCase + this a function, so always start with a verb

```suggestion
    const auto getFileNameForIcon = [&project](const gd::String &platform,
```

GDJS/GDJS/IDE/ExporterHelper.cpp
```diff
@@ -844,4 +845,74 @@ void ExporterHelper::AddDeprecatedFontFilesToFontResources(
...
+                                            const int size) {
+      return project.GetPlatformSpecificAssets().Has(
```
4ian, line 863: Nit: can you `const gd::String iconName = "icon-" + gd::String::From(size)`?

GDJS/GDJS/IDE/ExporterHelper.cpp
```diff
@@ -844,4 +845,74 @@ void ExporterHelper::AddDeprecatedFontFilesToFontResources(
...
+          gd::String(R"({
+        "src": "{FILE}",
+        "sizes": "{SIZE}x{SIZE}" 
+      },)")
+              .FindAndReplace("{SIZE}", gd::String::From(sizeAndFile.first))
+              .FindAndReplace("{FILE}", sizeAndFile.second);
+    }
+  }
+
+  icons = icons.RightTrim(",") + "]";
```
4ian, line 896: Nice, very clean!

4ian (approved):
Nice! So this will for now make the game looks great when added on the home screen on iOS/Android right? This is a good first step!

arthuro555:
I made the last commit from my phone, it is untested. There is a low chance of this actually breaking but I recommend still doing so before merging.

4ian (approved):

4ian:
Thanks!

PR #4023, opened 2022-06-12
Optionally async actions

arthuro555:
This allows to make optionally async action.

- [X] Modify the C++ code to make code generation support it
- [X] Remove setAsync as it is redundant with the specification of async functions
- [x] Allow to toggle the awaiting in the IDE
- [x] Make an optionally async action to test it
- [x] Write tests for optionally async free action
- [x] Write tests for optionally async object action
- [x] Add an events sheet context menu option to toggle the awaiting of an optionally async function


GDJS/Runtime/events-tools/networktools.ts
```diff
@@ -66,6 +70,43 @@ namespace gdjs {
...
+      export const sendAwaitableAsyncRequest = (
+        url: string,
+        body: string,
+        method: string,
+        contentType: string,
+        responseVar: gdjs.Variable,
+        errorVar: gdjs.Variable
+      ) => {
+        return new gdjs.PromiseTask(
+          fetch(url, {
```
arthuro555, line 82: I used fetch here since it returns a promise by default so it is more convenient to make an asynchronous action with. While I tried to keep the behavior the same as with the XmlHttpRequest, there might be some differences in internal behavior still. Is it fine to keep it like this or should I change it to use XmlHttpRequest to be 100% certain that the behavior is the same as the synchronous version?

Core/GDCore/Events/CodeGeneration/EventsCodeGenerator.cpp
```diff
@@ -468,6 +468,14 @@ gd::String EventsCodeGenerator::GenerateActionCode(
         action, *this, context);
   }
 
+  // Get the correct function name depending on whether it should be async or
+  // not
+  const gd::String& functionCallName =
+      instrInfos.IsAsync() &&
+              (!instrInfos.IsOptionallyAsync() || action.IsAwaited())
```
arthuro555, line 475: This expression is used at multiple places in the code. It is not very dry, but it is annoying since it has to use both the Instruction and InstructionMetadata, which are decoupled. Would it be possible to make the gd::Instruction fetch and cache the corresponding Instruction metadata when setting the instruction type? That way, not only could this expression be implemented as a method on gd::Instruction (e.g. something like `bool gd::Instruction::RunsAsAsync()`), but from the IDE and other places it would be faster and more convenient to have the instruction metadata accessible directly instead of having to gp get it from the metadata provider every time.

GDJS/Runtime/events-tools/networktools.ts
4ian: I think fetch is fine! Its support is now more than enough even across dated browsers, and I don't think we risk much in terms of behavior.

Core/GDCore/Events/CodeGeneration/EventsCodeGenerator.cpp
4ian: I think it's better to keep it like this:
- it's not super long
- it's super clear about what it does
- there is no hidden complexity, it's just boolean checks.

But this could be abstracted into a `ShouldRunAsAsync(instrMetadata, instruction)` static function inside a class, I'm not against it.

I would avoid at all cost caching the instruction metadata, because caching implies that:
- you should invalidate the cache when the instruction metadata is updated (which can happen for extensions).
- there is a risk that if the cache is not invalidated of having wrongly generated instructions and more generally terrible bugs.
- if the performance of something is tied to the fetching of metadata, then it's the metadata provider that should be replaced to give access in constant time to an instruction metadata (this could speed up export).

Overall, keep things "simple and stupid": 
- either keep this everywhere
- or create a quick ShouldRunAsAsync helper (which would be a good idea I guess!)
- but no caching that will have implications way beyond this (cache invalidation, yet-another-abstraction, risks of crashs...), all of this to avoid 3 checks or a function that are already super simple and clear :)

4ian:
I'm merging master in this branch as it was well done and only missing tests :)

4ian:
Alright, just one more test or two to test the generation of an optionally async object action and we should be good!

4ian (approved):

PR #4027, opened 2022-06-13
Remove out of label click on checkboxes

### LINKED ISSUE #4024: Clicking empty space > check box
pop
Clicking in the empty spaces will always check the boxes near the clicks in the behavior options.

## To Reproduce
1. Open any behavior that have check box options in it (in my GIF its Particles object).
2. Click on the empty space beside the box.
3. The box will get check and uncheck while you are clicking in the empty place.

## Other details
* GDevelop offline version (Desktop).
* Latest GDevelop official version.
* Even if this is not a bug (built on purpose), it still feels wrong to check the box while you are clicking on empty places.

AlexandreSi:
Fixes #4024

Changes from display flex to block for the from group so that the label does not take the whole available space.

### DIFF
--- a/newIDE/app/src/UI/Checkbox.js
+++ b/newIDE/app/src/UI/Checkbox.js
@@ -1,8 +1,9 @@
 // @flow
 import * as React from 'react';
-import FormControlLabel from '@material-ui/core/FormControlLabel';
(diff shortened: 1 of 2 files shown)


ClementPasteau (approved):
lgtm 👍

PR #4033, opened 2022-06-13
Suggest all expression types in fields

fannieyan:
- Suggest all expression regardless of type in fields
- Wrap expressions returning number by a ToString conversion when using autocompletion or instruction selectors

Test it here :
http://gdevelop-storybook.s3-website-us-east-1.amazonaws.com/suggest-all-expression-types-in-string-fields/latest/index.html?path=/story/parameterfields--stringfield


newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/FormatExpressionCall.js
```diff
@@ -85,9 +85,11 @@ export const getVisibleParameterTypes = (
 
 export const formatExpressionCall = (
   expressionInfo: EnumeratedInstructionOrExpressionMetadata,
-  parameterValues: ParameterValues
+  parameterValues: ParameterValues,
+  shouldConvertToString: boolean
```
ClementPasteau, line 89: never a big fan of those loose booleans.
Can we instead have an options object, that has a `shouldConvertTostring` key? It's easier to read when we call the function to see `{ shouldConvertToString: true }` rather than just `true`

newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/index.js
```diff
@@ -268,12 +268,48 @@ export default class ExpressionField extends React.Component<Props, State> {
...
+      globalObjectsContainer,
+      objectsContainer,
+      expressionType,
+      value,
+    } = this.props;
     const cursorPosition = this._inputElement.selectionStart;
+    const parser = new gd.ExpressionParser2();
 
-    const functionCall = formatExpressionCall(expressionInfo, parameterValues);
+    // We add a character so that getCompletionDescriptionsFor will always return the type of the node
```
ClementPasteau, line 280: I'm not familiar with this part of the code but this seems like a hack 🤔

newIDE/app/src/ExpressionAutocompletion/index.js
```diff
@@ -513,32 +521,118 @@ type InsertedAutocompletion = {|
...
 
+const findCompletedExpressionStartPosition = (expression: string) => {
+  let match;
+  let indexes = [];
+  // We consider that expressions are composed of letters, digits, dot or colons
+  const expressionSeparatorRegex = new RegExp(/[^\w\d.:]/g);
+  while ((match = expressionSeparatorRegex.exec(expression))) {
+    indexes.push(match.index);
+  }
+  return Math.max.apply(null, indexes);
```
ClementPasteau, line 540: you probably can just `Math.max(...indexes)`

newIDE/app/src/ExpressionAutocompletion/index.js
```diff
@@ -513,32 +521,118 @@ type InsertedAutocompletion = {|
...
   addNamespaceSeparator?: ?boolean,
+  shouldConvertToString?: ?boolean,
 |};
 
 type ExpressionAndCaretLocation = {|
   expression: string,
   caretLocation: number,
 |};
 
+const findCompletedExpressionStartPosition = (expression: string) => {
```
ClementPasteau, line 532: it took me a while to understand what we're doing here.
Maybe add a comment with an example of what this function does?

newIDE/app/src/InstructionOrExpression/EnumerateExpressions.js
```diff
@@ -10,15 +10,6 @@ const gd: libGDevelop = global.gd;
 
 const GROUP_DELIMITER = '/';
 
-// Support enumerate expressions that can be string, number or both ("number|string").
```
D8H, line 13: Are we sure we don't need it any more?
I think that it may be better to not suggest functions returning a string when the node is a number.

newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/index.js
4ian: Indeed, @fannieyan can you explain why you need the type at a precise position?
it seems to be to check if some completions are strings, but why is this needed?

fannieyan: This part is used when we choose an expression from the selector that displayed when clicking on the expression button (on the right of the field). It is then inserted where the caret was. There was previously no notion of type (except the type of the field) to determine what could be found in this selector.
It is the reason why I wanted to retrieve what type should the expression be here. I indeed hacked it so that the autocompletion finder is forced to return a type to me so that I would know what kind I needed/if I needed to convert the chosen expression. If a character is not added, if we insert an expression between parenthesis, it would not return autocompletion suggestions (it is as if we simulate that the user wants to type something).

newIDE/app/src/InstructionOrExpression/EnumerateExpressions.js
fannieyan: Indeed, my bad 😅

D8H: But, it's open to discussion, I guess.

fannieyan: It makes sense not to open suggestions for numbers as all numbers can be converted to string while you have to make sure your string is a number before converting it.
Let's keep the filtering for number expressions for now and, if needs arise, we could remove this filtering later?

ClementPasteau: I agree with you on this point, number -> string is the thing users are most likely to do so let's suggest those options, whereas string -> number is quite rare.

My coding background still urges me to see all the functions, regardless of the type, and then I decide what I use...
But our users may not have the same view.

Let's keep the filtering and we'll get more feedback about this :)

newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/index.js
4ian: Can you try to use `ExpressionNodeLocationFinder::GetNodeAtPosition` to get the node at the cursor position, then `ExpressionTypeFinder::GetType` to get the type of this node?

I think this would be less hacky than adding a character and then looking at the completions (there could be none for a good reason).

You'll need to expose these two static methods from the C++, let me know if you struggle to do that (search for `STATIC_` in Bindings.idl)

newIDE/app/src/InstructionOrExpression/EnumerateExpressions.js
4ian: Agree!

4ian (commented):
LGTM once we get the type of the node using the methods to locate a node and then get the type of the node

newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/index.js
```diff
@@ -268,12 +268,46 @@ export default class ExpressionField extends React.Component<Props, State> {
...
+      gd.JsPlatform.get(),
+      globalObjectsContainer,
+      objectsContainer,
+      expressionType,
+      expressionNode,
+      cursorPosition
+    );
+
+    let shouldConvertToString = false;
+    if (expressionInfo.metadata.getReturnType() === 'number') {
```
ClementPasteau, line 299: maybe add a comment on this whole part explaining the logic happening:
"If the field expects a string and one of the completions returns a number, then ... so that we can add the ToString()..."

newIDE/app/src/ExpressionAutocompletion/index.js
```diff
@@ -513,32 +521,124 @@ type InsertedAutocompletion = {|
...
+  }
+  return Math.max(...indexes);
+};
+
+const getNewExpressionWithToStringConversion = ({
+  expression,
+  wordStartPosition,
+  wordEndPosition,
+  insertedWord,
+}: {
```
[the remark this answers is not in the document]
fannieyan: I see what you mean, although I don't really know how to do it 😅 Do you have a lead on what to do?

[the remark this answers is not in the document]
D8H: My guess would have been to give the responsability to find the bounds to `ExpressionCompletionFinder.getCompletionDescriptionsFor` but it gives information to enumerate the choices and the place where to put these choices so this is probably not the right place.
Maybe new class could be used. Something like `FunctionCallFinder`. The bounds can be found on that node and used to insert `ToString(` at the right place maybe.
@4ian Do you think it's overthinking it?

newIDE/app/src/ExpressionAutocompletion/index.js
```diff
@@ -513,32 +521,126 @@ type InsertedAutocompletion = {|
...
   caretLocation: number,
 |};
 
+/**
+ * Returns the position where the last expression call starts,
+ * so that we know where to insert the ToString call.
+ * For example:
+ * '"HelloWorld" + Object.Behavior::' should return 15.
+ */
+const findCompletedExpressionStartPosition = (expression: string) => {
```
4ian, line 538: Let's talk about nodes here:

```suggestion
const findLastNodeStartPosition = (expression: string) => {
```

newIDE/app/src/ExpressionAutocompletion/index.js
```diff
@@ -513,32 +521,126 @@ type InsertedAutocompletion = {|
...
+  // NodeParenthesesBoundsFinder worker to specify the proper bounds for each node.
+  const completedExpressionStartPosition: number = findCompletedExpressionStartPosition(
+    expressionStart
+  );
+
+  const newExpressionStart = expression.substring(
+    0,
+    completedExpressionStartPosition + 1
+  );
+  const completedExpressionStart = expression.substring(
```
4ian, line 571:
```suggestion
  const completedNodeStart = expression.substring(
```

4ian (approved):

PR #4041, opened 2022-06-16
Text to string for all text related to a string for a variable

### LINKED ISSUE #846: Inconsistent condition/action names
## Describe the bug
The conditions/actions for setting the text of a scene/global variable are named differently depending on if you're adding a condition or an action. This makes it hard to find them when searching.

## To Reproduce
1. Click 'Add condition' in the Events editor
2. Expand the Variables group on the left
3. Note the names of the conditions for setting the text of a scene and global variable ('**Text** of a [scene/global] variable')
4. Cancel, and click 'Add action'
5. Expand the Variables group on the left
6. Note the names of the conditions for setting the text of a scene and global variable ('**String** of a [scene/global] variable') (note also that clicking on one of these shows the title 'Modify the **text** of a [scene/global] variable')

I think naming these consistently would be better.

text_vs_string

Bouh:
Because now in the revamped variable editor we use string.
Close #846

### DIFF
--- a/Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
@@ -407,9 +407,9 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
       .UseStandardOperatorParameters("number");
 
   obj.AddAction("ModVarObjetTxt",
-                _("Text of an object variable"),
-                _("Change the text of an object variable."),
-                _("the text of variable _PARAM1_"),
+                _("String of an object variable"),
+                _("Change the string of an object variable."),
+                _("the string of variable _PARAM1_"),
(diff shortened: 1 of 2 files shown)


Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp
```diff
@@ -407,9 +407,9 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsBaseObjectExtension(
       .UseStandardOperatorParameters("number");
 
   obj.AddAction("ModVarObjetTxt",
-                _("Text of an object variable"),
-                _("Change the text of an object variable."),
-                _("the text of variable _PARAM1_"),
+                _("String of an object variable"),
```
ClementPasteau, line 410: in the related issue, I think there was a suggestion to change *all* actions & conditions descriptions to "String (Text)"

PR #4045, opened 2022-06-17
Give me comments

fannieyan:
Capture d’écran 2022-06-17 à 16 16 18

Capture d’écran 2022-06-17 à 16 16 28


AlexandreSi:
I'm curious how the layout evolves if the toggle labels are not the same length:

Here see simulation on the storybook

image

http://gdevelop-storybook.s3-website-us-east-1.amazonaws.com/give-me-comments/latest/index.html?path=/story/gamedashboard-gamecard--default-game-card

newIDE/app/src/UI/Card/CardContent.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+import * as React from 'react';
+import { CardContent as MUICardContent, makeStyles } from '@material-ui/core';
```
AlexandreSi, line 3: To change for

```
import MUICardContent from '@material-ui/core/CardContent'
import { makeStyles } from '@material-ui/core/styles'
```
I think it changes compilation time and maybe bundle size

newIDE/app/src/UI/Card/CardContent.js
```diff
@@ -0,0 +1,21 @@
...
+
+const useStyles = makeStyles({
+  root: {
+    padding: 24,
+    paddingTop: 8,
```
AlexandreSi, line 8: I wonder if we should use `marginsSize` from `Grid` with multipliers

newIDE/app/src/UI/Card/CardHeader.js
```diff
@@ -0,0 +1,28 @@
+// @flow
+import * as React from 'react';
+import { CardHeader as MUICardHeader, makeStyles } from '@material-ui/core';
```
AlexandreSi, line 3: Same remark as for `CardContent`

newIDE/app/src/stories/componentStories/GameDashboard/GameCard.stories.js
```diff
@@ -0,0 +1,42 @@
+// @flow
```
AlexandreSi, line 1: 💜

newIDE/app/src/stories/componentStories/GameDashboard/GamesList.stories.js
```diff
@@ -0,0 +1,77 @@
+// @flow
```
AlexandreSi, line 1: 💜

newIDE/app/src/stories/componentStories/GameDashboard/GamesList.stories.js
```diff
@@ -0,0 +1,77 @@
...
+import { GamesList } from '../../../GameDashboard/GamesList';
+
+export default {
+  title: 'GameDashboard/GameCard',
+  component: GameCard,
+  decorators: [muiDecorator, paperDecorator],
+};
+
+export const WithoutAProjectOpened = () => {
+  const mock = new MockAdapter(axios);
```
AlexandreSi, line 27: Great!

newIDE/app/src/stories/componentStories/GameDashboard/GamesList.stories.js
AlexandreSi: The filename is wrong though (I think)

newIDE/app/src/GameDashboard/ShareDialog.js
```diff
@@ -0,0 +1,133 @@
+// @flow
+import { Trans } from '@lingui/macro';
+import { CircularProgress } from '@material-ui/core';
```
AlexandreSi, line 3:
```
import CircularProgress from '@material-ui/core/CircularProgress';
```

newIDE/app/src/GameDashboard/ShareDialog.js
```diff
@@ -0,0 +1,133 @@
...
+    [getAuthorizationHeader, profile, game]
+  );
+
+  React.useEffect(
+    () => {
+      fetchGameSlug();
+    },
+    [fetchGameSlug]
+  );
+  console.log(gameSlug);
```
AlexandreSi, line 55: To remove

newIDE/app/src/GameDashboard/GameCard.js
```diff
@@ -1,92 +1,263 @@
...
+                    <Column expand alignItems="flex-start" noMargin>
+                      <Toggle
+                        labelPosition="left"
+                        onToggle={onToggleDiscoverable}
+                        toggled={getCurrentState(
+                          game.discoverable,
+                          isDiscoverable
+                        )}
+                        label={<Trans>Make discoverable on Liluo</Trans>}
+                        style={{ marginLeft: 0 }}
```
4ian, line 224: We might want to do the reverse:
- Have no margin by default
- Have any code using already `Toggle` to use `Column` instead around the toggle..

newIDE/app/src/GameDashboard/GameCard.js
```diff
@@ -1,92 +1,263 @@
...
+                </Line>
+                <ElementWithMenu
+                  element={
+                    <IconButton size="small" style={{ padding: 0 }}>
+                      <MoreVert />
+                    </IconButton>
+                  }
+                  buildMenuTemplate={(i18n: I18nType) => [
+                    {
+                      label: i18n._(t`Open details`),
```
AlexandreSi, line 146: Wdyt about:
- Open details
- See builds
- Open analytics
- Manage leaderboards
?

newIDE/app/src/GameDashboard/GameCard.js
```diff
@@ -1,92 +1,263 @@
...
+import Toggle from '../UI/Toggle';
+import MoreVert from '@material-ui/icons/MoreVert';
+import CardContent from '../UI/Card/CardContent';
+import AuthenticatedUserContext from '../Profile/AuthenticatedUserContext';
+import CardHeader from '../UI/Card/CardHeader';
+import ElementWithMenu from '../UI/Menu/ElementWithMenu';
+import { type I18n as I18nType } from '@lingui/core';
+import { type GamesDetailsTab } from './GameDetailsDialog';
+import { useDebounce } from '../Utils/UseDebounce';
+import ShareDialog from './ShareDialog';
```
AlexandreSi, line 29: As a nitpicking:
I like to have imports organized so that they look like:
```
Usual imports (react, i18n)
Material icons
UI components
Other
```
Separated by new lines. It can help sometimes understand the context of the component (Is is really dependant on Material ? Is it self-sufficient or does it use a lot of sub components? etc.

But this is not mandatory

newIDE/app/src/GameDashboard/GameCard.js
```diff
@@ -1,92 +1,263 @@
...
+  const [
+    acceptsGameComments,
+    setAcceptsGameComments,
+  ] = React.useState<?boolean>(null);
+
+  const { getAuthorizationHeader, profile } = React.useContext(
+    AuthenticatedUserContext
+  );
+
+  const updateGameProperties = useDebounce(async parameters => {
```
AlexandreSi, line 69: Can you type parameters?

newIDE/app/src/GameDashboard/GameCard.js
```diff
@@ -1,92 +1,263 @@
...
-                    primary
-                  />
-                </ResponsiveLineStackLayout>
-              </CardActions>
-            </Column>
-          </ResponsiveLineStackLayout>
-        </Card>
+
+                  <Text size="body2" noMargin displayInlineAsSpan>
+                    <Trans>Created on {i18n.date(game.createdAt * 1000)}</Trans>{' '}
```
AlexandreSi, line 135: I'm not sure `{' '}` is still useful here

AlexandreSi (commented):
Great PR, excited to see those new cards!

newIDE/app/src/GameDashboard/GameCard.js
fannieyan: I like this, it looks more organised and clean! :D

newIDE/app/src/UI/Toggle.js
```diff
@@ -17,10 +17,11 @@ type Props = {|
   toggled: boolean,
   onToggle: (e: {||}, toggled: boolean) => void,
   disabled?: boolean,
-  labelPosition: 'right',
+  labelPosition: 'right' | 'left',
 
   style?: {|
     marginTop?: number,
+    marginLeft?: number,
```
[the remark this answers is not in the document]
fannieyan: Actually, I used this props to set marginLeft to 0

newIDE/app/src/GameDashboard/GameCard.js
```diff
@@ -1,92 +1,275 @@
...
+                        </IconButton>
+                      </ResponsiveLineStackLayout>
+                    </Column>
+                  </ResponsiveLineStackLayout>
+                  <ResponsiveLineStackLayout
+                    noMargin
+                    justifyContent="flex-start"
+                  >
+                    <Column expand alignItems="flex-start" noMargin>
+                      <Toggle
```
ClementPasteau, line 231: Should those toggles open a small dialog warning the user what they're about to do so they can confirm?
As a newcomer or an experienced user, I'd feel more reassured by a 2 step process when making my game discoverable/open for feedback

newIDE/app/src/GameDashboard/GameCard.js
```diff
@@ -1,92 +1,275 @@
...
+                        label={<Trans>Make discoverable on Liluo</Trans>}
+                      />
+                      <Toggle
+                        labelPosition="left"
+                        onToggle={onToggleAcceptsGameComments}
+                        toggled={getCurrentState(
+                          game.acceptsGameComments,
+                          acceptsGameComments
+                        )}
+                        label={<Trans>Open for feedback on Liluo</Trans>}
```
ClementPasteau: I feel that we should improve this wording, it's too similar to the one for builds.
Should it be `Show feedback banner on Liluo.io game page`?
And for builds: `Ask for feedback on all build pages`

ClementPasteau: By the way, let's ensure we always put `Liluo.io` everywhere, to be consistent

newIDE/app/src/UI/Toggle.js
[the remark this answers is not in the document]
4ian: I think you can set `marginLeft?: 0` in the typing to show that the only allowed value is 0 (useful to force future developers to avoid taking advantage of this - too much liberty is NOT something we want here).

Or as said in another comment, the component should apply no margin by itself - it's the parent that should in theory do the layouing with Column/ColumnStackLayout/LineStackLayout (or, in case they don't work, Spacer, but that's a band aid).

newIDE/app/src/GameDashboard/ShareDialog.js
```diff
@@ -0,0 +1,148 @@
...
+      try {
+        const gameSlugs = await getGameSlugs(
+          getAuthorizationHeader,
+          profile.id,
+          game.id
+        );
+        setGameSlug(gameSlugs[0]);
+      } catch (error) {
+        console.log(error);
+        setShowAlertMessage(true);
```
ClementPasteau, line 47: Why use the AlertMessage instead of the usual `showErrorBox()` ?

fannieyan: Mmmh, good question. I wonder if the error box wouldn't be too "alarming" since we can still provide a working url (but with game id) ?

newIDE/app/src/GameDashboard/GameCard.js
fannieyan: Dealing with it in #4063! :)

newIDE/app/src/GameDashboard/Feedbacks/FeedbackCard.js
```diff
@@ -0,0 +1,170 @@
...
+  const ratings = getRatings(comment.ratings);
+  const theme = React.useContext(GDevelopThemeContext);
+
+  const [processed, setProcessed] = React.useState<boolean>(
+    !!comment.processedAt
+  );
+  const [
+    pendingProcessedState,
+    setPendingProcessedState,
+  ] = React.useState<?boolean>(null);
```
ClementPasteau, line 76: Can we do everything with only 1 value in the state?

(not sure, with the fact that a user can click twice, and only the second call would be registered, and if fails, then we need to revert to the previous "saved" value)

newIDE/app/src/GameDashboard/Feedbacks/GameFeedback.js
```diff
@@ -0,0 +1,134 @@
...
+        {!authenticatedUser.authenticated && (
+          <EmptyMessage>
+            <Trans>You need to login first to see your game feedbacks.</Trans>
+          </EmptyMessage>
+        )}
+        {authenticatedUser.authenticated &&
+          !displayedFeedbacks &&
+          !isErrored && <PlaceholderLoader />}
+        {authenticatedUser.authenticated && isErrored && (
+          <PlaceholderError onRetry={() => {}}>
```
ClementPasteau, line 84: missing onRetry here, it should be `loadFeedbacks` ?

newIDE/app/src/GameDashboard/GameCard.js
```diff
@@ -1,92 +1,331 @@
...
 
-export const GameCard = ({ game, isCurrentGame, onOpenGameManager }: Props) => {
+type TogglableProperties =
+  | 'discoverable'
+  | 'acceptsBuildComments'
+  | 'acceptsGameComments';
+
+const confirmationMessage = {
+  discoverable: {
+    true: <Trans>Are you sure you want to make this game discoverable?</Trans>,
```
ClementPasteau, line 51: let's take the opportunity to add a bit more info to these messages, to educate users how this work.
This could become:
`You're about to make this game discoverable on categories pages on Liluo.io. Continue?`

newIDE/app/src/GameDashboard/GameCard.js
```diff
@@ -1,92 +1,331 @@
...
+    false: (
+      <Trans>
+        Are you sure you don't want to make this game discoverable anymore?
```
ClementPasteau, line 54: => `You're about to hide this game from categories pages on Liluo.io. Continue?`

newIDE/app/src/GameDashboard/GameCard.js
```diff
@@ -1,92 +1,331 @@
...
+  onUpdateGame,
+}: Props) => {
   const openGameUrl = () => {
     const url = getGameUrl(game);
     if (!url) return;
     Window.openExternalURL(url);
   };
+  const [showShareDialog, setShowShareDialog] = React.useState(false);
+  const [
+    showConfirmationDialog,
```
ClementPasteau, line 97: I'd rename this prop, I got confused and thought it was a boolean deciding if the dialog is opened or not.
Maybe `shownConfirmationDialog`/`setShownConfirmationDialog` (switch focus from "show" to "dialog")

Or maybe better: `confirmationDialogName` ?

newIDE/app/src/UI/Card/CardContent.js
```diff
@@ -0,0 +1,23 @@
+// @flow
+import * as React from 'react';
+import MUICardContent from '@material-ui/core/CardContent';
+import { makeStyles } from '@material-ui/core/styles';
+import { marginsSize } from '../Grid';
```
ClementPasteau, line 5: are those margins really linked?
Having the same make sense, but I'm wondering if we should define it again here, rather than import a margin used from another component 🤔

newIDE/app/src/Utils/GDevelopServices/Play.js
```diff
@@ -294,3 +294,83 @@ export const deleteLeaderboardEntry = async (
...
+  id: string,
+  type: CommentType,
+  gameId: string,
+  buildId?: string,
+  text: string,
+  ratings?: GameRatings,
+  playerId?: string,
+  playerName?: string,
+  contact?: string,
+  createdAt: number,
```
ClementPasteau, line 319: I think there is also an updatedAt prop. not used, but could be useful in the future

newIDE/app/src/GameDashboard/ShareDialog.js
```diff
@@ -0,0 +1,148 @@
...
+    () => {
+      fetchGameSlug();
+    },
+    [fetchGameSlug]
+  );
+
+  const gameUrl = getGameUrl(game, gameSlug);
+  if (!gameUrl) return null;
+
+  const onCopyLinkToClipboard = () => {
```
ClementPasteau, line 65: you seem to have this logic copied twice (here and in OnlineGameLink + onOpen + onShare.
Should they be integrated inside the <ShareLink> component?
The `ShareLink` could just take a `url` and create the textfield with the buttons & the actions associated

newIDE/app/src/GameDashboard/Feedbacks/FeedbackCard.js
fannieyan: I don't think we can; with only one state, we can't know what was the first state while the call is debounced.
As you said, if I click two times, it will do "true -> false -> true" the call will be sent at the second true and I don't think we can easily get back to the first "true" if the call fails

fannieyan: I managed to abstract this logic in a useOptimisticState so that it will be easier to read.

newIDE/app/src/UI/Card/CardContent.js
4ian: Alternative is to use Column/Line and/or a bunch of `<Spacer />`?

fannieyan: It was suggested to keep consistency between spacings.
Actually I just found the getMuiOverrides in CreateTheme file and it should make more sense to delete those "custom" component and directly override those styles

newIDE/app/src/GameDashboard/ShareDialog.js
fannieyan: There is a bit of logic involved in `OnlineGameLink` on the display of ShareButtons (it relies on whether the game is published or not), so I could only factorise a `ShareButton` (and rename `ShareButtons` as `SocialShareButtons`)

newIDE/app/src/UI/Theme/CreateTheme.js
```diff
@@ -203,6 +203,18 @@ export function getMuiOverrides(
         padding: 8,
       },
     },
+    MuiCardHeader: {
```
ClementPasteau, line 206: I didn't know we had this in the app. Nice

newIDE/app/src/GameDashboard/Feedbacks/FeedbackCard.js
```diff
@@ -58,61 +58,35 @@ const getRatings = (ratings: ?GameRatings) => {
...
-    setPendingProcessedState,
-  ] = React.useState<?boolean>(null);
+  const processComment = async (processed: boolean, i18n: I18nType) => {
+    if (!profile) return;
+    try {
+      await updateComment(getAuthorizationHeader, profile.id, {
+        gameId: comment.gameId,
+        commentId: comment.id,
+        processed,
+      });
```
ClementPasteau, line 73: as discussed, you probably need to:
1. retrieve the comment returned from the call `const updatedComment = await udpateComment(...)`
2. call a `onCommentUpdated(updatedComment)` defined in the top component, which replaces the comment in question (findable by ID) by this one.

This would fix the sort applied on the correct updated list.

ClementPasteau (approved):
LGTM 🚀  🚀

PR #4056, opened 2022-06-21
Fix collision once again

### LINKED ISSUE #4049: Physics collision not detected
Following a forum user report, it would seem that the recent fix for the Physics engine (to better detect collision happening between frames) triggered a collision detection issue.

Video with the same project on v135 and v137.
https://drive.google.com/drive/folders/1H3m2fk79061GW8El8j5uwbmKI4FPaH0N

Screenshot of the related event:
image

image

[source](https://forum.gdevelop.io/t/strange-behaviour-after-latest-update-collision-stuff/40368)

AlexandreSi:
As a follow up of #4018 
Closes #4049
And related forum posts

### DIFF
--- a/Extensions/Physics2Behavior/physics2runtimebehavior.ts
+++ b/Extensions/Physics2Behavior/physics2runtimebehavior.ts
@@ -719,10 +719,6 @@ namespace gdjs {
         this.createBody();
       }
 
-      // Reset contacts that happened this frame
-      this.contactsStartedThisFrame.length = 0;
-      this.contactsEndedThisFrame.length = 0;
-
       // Step the world if not done this frame yet
       if (!this._sharedData.stepped) {
         this._sharedData.step(
@@ -759,6 +755,10 @@ namespace gdjs {
 
     doStepPostEvents(runtimeScene) {
       this._updateBodyFromObject();
+
+      // Reset contacts that happened this frame
+      this.contactsStartedThisFrame.length = 0;
+      this.contactsEndedThisFrame.length = 0;
 
       // Reset world step to update next frame
       this._sharedData.stepped = false;
--- a/Extensions/Physics2Behavior/tests/physics2runtimebehavior.spec.js
+++ b/Extensions/Physics2Behavior/tests/physics2runtimebehavior.spec.js
@@ -131,8 +131,8 @@ describe('Physics2RuntimeBehavior', () => {
 
       let hasBounced = false;
       let stepIndex = 0;
-      while (stepIndex < 10 && !hasBounced) {
-        runtimeScene.renderAndStep(1000 / fps);
+
+      runtimeScene.setEventsFunction(() => {
         if (movingObjectBehavior.getLinearVelocityY() > 0) {
           // If the moving object has a positive velocity, it hasn't bounced
           // on the static object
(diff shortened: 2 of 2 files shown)


Extensions/Physics2Behavior/tests/physics2runtimebehavior.spec.js
```diff
@@ -178,8 +181,12 @@ describe('Physics2RuntimeBehavior', () => {
 
       let hasBounced = false;
       let stepIndex = 0;
-      while (stepIndex < 10 && !hasBounced) {
-        runtimeScene.renderAndStep(1000 / fps);
+      // Seems like the events function is called even if hasBounce is true.
```
4ian, line 184: Not sure I understand this comment, of course the events function is supposed to be called, that's normal no?

Extensions/Physics2Behavior/tests/physics2runtimebehavior.spec.js
```diff
@@ -190,28 +197,39 @@ describe('Physics2RuntimeBehavior', () => {
...
+          subStepIndex += 1;
+          if (subStepIndex === 1) {
+            // At first frame, collision should have only started
+            expect(movingObject.getY() < staticObject.getY()).to.be(true);
+            assertCollision(movingObject, staticObject, {
+              started: true,
+              collision: true,
+              stopped: false,
+            });
+          } else if (subStepIndex === 3) {
```
4ian, line 209: Does it make sense to test subStepIndex === 2?

Extensions/Physics2Behavior/tests/physics2runtimebehavior.spec.js
```diff
@@ -190,28 +197,39 @@ describe('Physics2RuntimeBehavior', () => {
...
-
-      // At next frame, end of collision should be detected
-      assertCollision(movingObject, staticObject, {
-        started: false,
-        collision: false,
-        stopped: true,
-      });
+      if (!hasCollisionEndedBeenTested) {
+        runtimeScene.renderAndStep(1000 / fps);
+      }
```
4ian, line 232: You should check for the subStepIndex or anything else that will check your test did run through the end (if I comment `renderAndStep`, the test will work).
The `if` also sounds suspicious (in which case can it be false/true? Is it a flaky test?)

Extensions/Physics2Behavior/physics2runtimebehavior.ts
```diff
@@ -719,10 +719,6 @@ namespace gdjs {
         this.createBody();
       }
 
-      // Reset contacts that happened this frame
-      this.contactsStartedThisFrame.length = 0;
-      this.contactsEndedThisFrame.length = 0;
-
       // Step the world if not done this frame yet
       if (!this._sharedData.stepped) {
         this._sharedData.step(
```
D8H, line 724: I think it should moved inside `step(deltaTime)`.

D8H: I was not detected by tests probably because the checked object was the 1st to run. The test could create an object before and after the ones that are used to cover the case.

AlexandreSi: I guess so, but I'd rather postpone this for another PR in order to fix the current bug

D8H: It's not like it would take a lot of time.
Without this change the condition is still broken for extensions.

4ian (approved):

Extensions/Physics2Behavior/physics2runtimebehavior.ts
4ian: I would rather do ths larger change in another PR, because here we want to fix asap the version that is in the wild, and we can't take the risk of introducing another "subtle" behavior change that will get us deeper into the rabbit hole of bugs.

PR #4057, opened 2022-06-21
Bump newIDE version

AlexandreSi:


ClementPasteau (approved):
OK once the physics fix is merged!

PR #4063, opened 2022-06-24
Add Feedback tab

fannieyan:
Continuation of #4045 (split for an easier review)

- Add a tab to review the feedback sent
- Allow user to mark feedback as processed

Note : 
- The build on which the feedback is sent is not yet displayed (it has not much sense until we can customize them)
- We cannot yet filter feedbacks by build (for the same reasons)

image

Test the new tab here:
http://gdevelop-storybook.s3-website-us-east-1.amazonaws.com/add-feedback-tab/latest/index.html?path=/story/gamedashboard-feedback-gamefeedback--default-game-feedback

### DIFF
--- a/newIDE/app/src/GameDashboard/Feedbacks/FeedbackCard.js
+++ b/newIDE/app/src/GameDashboard/Feedbacks/FeedbackCard.js
@@ -7,19 +7,21 @@ import { I18n } from '@lingui/react';
 import { Card } from '@material-ui/core';
 import CheckCircleOutlineIcon from '@material-ui/icons/CheckCircleOutline';
 import CheckCircleIcon from '@material-ui/icons//CheckCircle';
+
+import { ResponsiveLineStackLayout } from '../../UI/Layout';
 import CardContent from '../../UI/Card/CardContent';
 import Text from '../../UI/Text';
+import { Column, LargeSpacer, Line, Spacer } from '../../UI/Grid';
+import IconButton from '../../UI/IconButton';
+import GDevelopThemeContext from '../../UI/Theme/ThemeContext';
+
+import Rating from './Rating';
 
 import {
   updateComment,
   type Comment,
   type GameRatings,
 } from '../../Utils/GDevelopServices/Play';
-import { ResponsiveLineStackLayout } from '../../UI/Layout';
-import Rating from './Rating';
-import { Column, LargeSpacer, Line, Spacer } from '../../UI/Grid';
-import IconButton from '../../UI/IconButton';
-import GDevelopThemeContext from '../../UI/Theme/ThemeContext';
 import { useDebounce } from '../../Utils/UseDebounce';
 import { type AuthenticatedUser } from '../../Profile/AuthenticatedUserContext';
 
--- a/newIDE/app/src/GameDashboard/Feedbacks/FeedbackCard.js
+++ b/newIDE/app/src/GameDashboard/Feedbacks/FeedbackCard.js
@@ -3,6 +3,7 @@
 import * as React from 'react';
 import { t, Trans } from '@lingui/macro';
 import { I18n } from '@lingui/react';
+import { type I18n as I18nType } from '@lingui/core';
 
 import { Card } from '@material-ui/core';
 import CheckCircleOutlineIcon from '@material-ui/icons/CheckCircleOutline';
@@ -24,14 +25,14 @@ import {
 } from '../../Utils/GDevelopServices/Play';
 import { useDebounce } from '../../Utils/UseDebounce';
 import { type AuthenticatedUser } from '../../Profile/AuthenticatedUserContext';
+import { showErrorBox } from '../../UI/Messages/MessageBox';
 
 const styles = { textComment: { whiteSpace: 'pre-wrap' } };
 
-type Props = {
+type Props = {|
   comment: Comment,
   authenticatedUser: AuthenticatedUser,
-  onUpdateComment: () => ?Promise<void>,
-};
+|};
 
 const getRatings = (ratings: ?GameRatings) => {
   if (!ratings) return null;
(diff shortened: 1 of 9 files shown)


newIDE/app/src/GameDashboard/Feedbacks/FeedbackCard.js
```diff
@@ -0,0 +1,142 @@
...
+  comment: Comment,
+  authenticatedUser: AuthenticatedUser,
+  onUpdateComment: () => ?Promise<void>,
+};
+
+const getRatings = (ratings: ?GameRatings) => {
+  if (!ratings) return null;
+  if (ratings.version === 1) {
+    return [
+      { label: 'Sound', value: ratings.sound },
```
4ian, line 38: Ideally these would be translatable.

newIDE/app/src/GameDashboard/Feedbacks/FeedbackCard.js
```diff
@@ -0,0 +1,142 @@
...
+                  <Text style={styles.textComment}>{comment.text}</Text>
+                </Column>
+              </CardContent>
+            </Column>
+            <IconButton
+              tooltip={isProcessed ? t`Unresolve` : t`Resolve`}
+              onClick={onSetProcessed}
+            >
+              {isProcessed ? (
+                <CheckCircleIcon htmlColor={theme.message.valid} />
```
4ian, line 130: Excellent usage of the theme :)

newIDE/app/src/GameDashboard/Feedbacks/Rating.js
```diff
@@ -0,0 +1,35 @@
...
+});
+
+type Props = {
+  value: number,
+  label: string,
+};
+
+/* Display a rating between 1 and 10. */
+const Rating = ({ value, label }: Props) => {
+  const classes = useStyles();
```
4ian, line 22: Could this be an inline style?

newIDE/app/src/GameDashboard/Feedbacks/FeedbackCard.js
```diff
@@ -0,0 +1,156 @@
...
+      try {
+        await updateComment(getAuthorizationHeader, profile.id, {
+          gameId: comment.gameId,
+          commentId: comment.id,
+          processed: processed,
+        });
+      } catch (error) {
+        console.warn(`Unable to update comment: ${error}`);
+      } finally {
+        await onUpdateComment();
```
ClementPasteau, line 85: should this be in the try rather than the finally?

newIDE/app/src/GameDashboard/Feedbacks/FeedbackCard.js
```diff
@@ -0,0 +1,156 @@
...
+  authenticatedUser,
+  onUpdateComment,
+}: Props) => {
+  const { getAuthorizationHeader, profile } = authenticatedUser;
+  const ratings = getRatings(comment.ratings);
+  const theme = React.useContext(GDevelopThemeContext);
+
+  const [processed, setProcessed] = React.useState<?boolean>(null);
+
+  const isProcessed = processed === null ? !!comment.processedAt : processed;
```
ClementPasteau, line 71: I feel like you can simplify this by:
- initialising the boolean `processed` with !!comment.processedAt
- and you can then always use `processed` instead of `isProcessed`

newIDE/app/src/GameDashboard/Feedbacks/FeedbackCard.js
```diff
@@ -0,0 +1,156 @@
...
+    async ({ processed }: { processed: boolean }) => {
+      if (!profile) return;
+      try {
+        await updateComment(getAuthorizationHeader, profile.id, {
+          gameId: comment.gameId,
+          commentId: comment.id,
+          processed: processed,
+        });
+      } catch (error) {
+        console.warn(`Unable to update comment: ${error}`);
```
ClementPasteau, line 83: this should probably be a console.error()
+ should we display a `showErrorBox()` ? so that the user know something went wrong

newIDE/app/src/GameDashboard/Feedbacks/GameFeedback.js
```diff
@@ -0,0 +1,135 @@
...
+              </EmptyMessage>
+            )}
+            {displayedFeedbacks.length !== 0 && (
+              <ColumnStackLayout expand>
+                {displayedFeedbacks.map((comment: Comment) => (
+                  <FeedbackCard
+                    key={comment.id}
+                    comment={comment}
+                    authenticatedUser={authenticatedUser}
+                    onUpdateComment={loadFeedbacks}
```
ClementPasteau, line 123: why do you need to load all feedbacks every time a comment is updated? This feels overkill / spamming the network.
Can we instead:
- always rely on the state of the component, which is initialised from the values in the first call to loadFeedbacks()
- when a check is done -> update the value in the state optimistically, whilst the network call happens
- if the network calls finished, fine, we do nothing
- if it fails, then the catch() reverts the value in the state with an error message

?

newIDE/app/src/GameDashboard/GameCard.js
```diff
@@ -39,14 +40,66 @@ type Props = {|
...
+  | 'discoverable'
+  | 'acceptsBuildComments'
+  | 'acceptsGameComments';
 
-const getCurrentState = (gameProperty: ?boolean, pendingState: ?boolean) => {
-  return !!(pendingState === null ? gameProperty : pendingState);
+const getConfirmationMessage = (newProperty: {
+  [key: TogglableProperties]: boolean,
+}) => {
+  console.log(newProperty);
```
ClementPasteau, line 51: log

newIDE/app/src/Utils/GDevelopServices/Play.js
```diff
@@ -294,3 +294,87 @@ export const deleteLeaderboardEntry = async (
...
+  gameId: string, // We are always able to link a comment to a game, even if made on a build.
+  buildId?: string, // If defined, the comment is made on a specific build.
+  text: string,
+  ratings?: GameRatings,
+  playerId?: string, // Useful in the future, to link a comment to a user.
+  playerName?: string, // For non-authenticated comments.
+  contact?: string, // In order to be able to contact the user.
+  createdAt: number,
+  updatedAt: number,
+  deletedAt?: number, // For soft delete.
```
4ian, line 323: Do we really need this in the editor?

newIDE/app/src/GameDashboard/Feedbacks/FeedbackCard.js
4ian: Maybe/maybe not, but this should be called `onCommentUpdated()`.

newIDE/app/src/GameDashboard/Feedbacks/GameFeedback.js
4ian: Also I think the way it's done is already optimistic, from something I suggested, but Clement is surely right his approach is simpler/avoids reloading everything everytime (while I surely suggested the "overkill" solution which is to do the change in a state and then reload everything and then clear the state. While here a local change is good enough, no need to reload everything)

newIDE/app/src/GameDashboard/Feedbacks/FeedbackCard.js
```diff
@@ -0,0 +1,170 @@
...
+      if (!profile) return;
+      try {
+        await updateComment(getAuthorizationHeader, profile.id, {
+          gameId: comment.gameId,
+          commentId: comment.id,
+          processed: newProcessedValue,
+        });
+        setProcessed(newProcessedValue);
+      } catch (error) {
+        console.error(`Unable to update comment: ${error}`);
```
4ian, line 92: Prefer:
```suggestion
        console.error(`Unable to update comment:`, error);
```

So that the error can be browsed in the debugger (and not stringified, which is harder to read/browse)

newIDE/app/src/GameDashboard/Feedbacks/FeedbackCard.js
```diff
@@ -0,0 +1,170 @@
...
+const getCurrentState = (pendingState: ?boolean, state: boolean) => {
+  return pendingState === null ? state : pendingState;
+};
+
+const FeedbackCard = ({ comment, authenticatedUser }: Props) => {
+  const { getAuthorizationHeader, profile } = authenticatedUser;
+  const ratings = getRatings(comment.ratings);
+  const theme = React.useContext(GDevelopThemeContext);
+
+  const [processed, setProcessed] = React.useState<boolean>(
```
4ian, line 70: Not mandatory for this PR, but think of if a hook "useOptimisticState" could be done to abstract this. You would pass it the initial state (source of truth), the function to do the change, and it would return the state to display (i.e: the pending or the source of truth) and the function to call to change it (that would change the pending state, and would launch the debounced version of the function to do the change).

4ian (approved):

Issue #4015, opened 2022-06-10
[$20.00 bounty] In a filepath a hashtag broke the preview 


Bouh:
## Describe the bug
In a filepath a hashtag broke the preview of the file in the scene editor, and in the resource tab, but the file is loaded in the preview.

## To Reproduce
Correct: `E:\Réalisations\GDevelop_branding\Press\itch.io\Game Jam\Weekend Jam - 1\Parts\hero.png`
Bugged: `E:\Réalisations\GDevelop_branding\Press\itch.io\Game Jam\Weekend Jam #1\Parts\hero.png`

![image](https://user-images.githubusercontent.com/1670670/173022056-07676dbc-29f2-4ad1-a588-4183ffd2070a.png)

## Other details
GD 5.0.136
Windows 10

jposeyjr:
This is due to how <img> tag treats # and in general, they shouldn't be used in file names. The best solution is likely to add https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent to all file paths to ensure all issues are caught. I don't have time currently to read how to contribute to this and do the changes but this should help anyone looking very quickly.

rishabhguptareal:
i can resolve this can you assign this to me

nifanpinc:
I've submitted a fix for this issue in #8393.

The root cause is that `#` in file paths gets interpreted as a URL fragment identifier by the browser. The fix encodes `#` as `%23` in the resolved local file path within `ResourcesLoader/index.js`, ensuring the full path is correctly loaded.

adamsardo:
I opened a fresh fix in #8581. It handles the `#` fragment problem in both places the older editor-only attempt left open: Electron editor local file URLs via `pathToFileURL`, and GDJS runtime/exported resource URLs by encoding `#` before cache/query parameters are appended.

tianheng2017:
I opened PR #8607 to fix this by encoding local resource file URLs with pathToFileURL and decoding them back with fileURLToPath for local file actions.

PR #4065, opened 2022-06-24
Feat: Fix img hash tag preview issue in corsawareimage.js issue #4015

jposeyjr:
A simple fix for https://github.com/4ian/GDevelop/issues/4015 by encoding the url and removing special chars as recommended by the MDN page. [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) I also attached screen shots showing it now working and the break point showing full path with the # file name. 
Screen Shot 2022-06-24 at 5 39 32 PM
Screen Shot 2022-06-24 at 5 39 54 PM
Screen Shot 2022-06-24 at 5 40 50 PM


4ian:
Hi! Thanks for the PR.
Can you confirm images with a `#` will also work when previewing and exporting a game, both on the desktop app and the web-app?
Also, is there a reason for not using the encodeURIComponent function - could there be other "special" characters like `#` creating issues?

jposeyjr:
Yes it worked on all forms of previewing and exporting from what I could tell. If you have a list of all the different ones let me know and I can go through it for sure. I did start with encodeURIComponent at the start but it didn't fix the issue. I then added the regex to it as well and again it didn't fix it. So I limited the scope to simple the # and it was working, I didn't dig into why the encoded function wasn't working as I don't have a deep understanding of the code base.

jposeyjr:
@4ian I fixed the preview, that was something I thought was working but it wasn't that was my bad :). The export "worked" in that it exported them but there is an error in PIXI.JS when trying to view the game. It seems to encode the space but not the # so that is the same issue. Would that be an issue on PIXI.js repo ?? Or would you want the version we have modified? Sorry not my project so don't want to assume the workflow with vendor scripts. 

Screen Shot 2022-06-26 at 12 09 45 PM
Exported project running index.html on live server extension.
Screen Shot 2022-06-26 at 12 27 21 PM

jposeyjr:
Any updates on this?

jposeyjr:
So part of the problem is that img tags and the way urls work. That was what I fixed but hashtags shouldn’t be in urls at all for files or most things. As they are used for internal anchors in the url schematic. 

When I stepped through the breakpoints it seems pixi doesn’t encode urls either. Honestly all urls should be encoded regardless of solution just from best practices stand point and security.  

It doesn’t help how many different things touch and use the url with no good linking between them. Trying to find the best place to encode it was tough too, but that I leave to my inexperience with the code base.

PR #4069, opened 2022-06-27
Don't display objects already in a group as addable

arthuro555:
Feature request: https://forum.gdevelop.io/t/remove-already-grouped-objects-from-list/40515?u=arthuro555

Filters already selected objects out of the autocomplete for adding new objects to a group.
This is good since:
- They are useless anyways (not letting one remove or re-add them again).
- We have a clearer list of what objects are left to add in the group, giving less room for overseeing an object that needs to be added.

### DIFF
--- a/newIDE/app/src/ObjectGroupEditor/index.js
+++ b/newIDE/app/src/ObjectGroupEditor/index.js
@@ -29,6 +29,7 @@ const ObjectGroupEditor = ({
   onSizeUpdated,
 }: Props) => {
   const [newObjectName, setNewObjectName] = React.useState<string>('');
+  const objectsInGroup = group.getAllObjectsNames().toJSArray();
 
   const removeObject = (objectName: string) => {
     group.removeObject(objectName);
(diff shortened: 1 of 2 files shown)


newIDE/app/src/ObjectGroupEditor/index.js
```diff
@@ -98,6 +96,7 @@ const ObjectGroupEditor = ({
             globalObjectsContainer={globalObjectsContainer}
             objectsContainer={objectsContainer}
             value={newObjectName}
+            excludedObjects={objectsInGroup}
```
4ian, line 99: I think in theory this could be called, to be super explicit:

```suggestion
            excludedObjectOrGroupNames={objectsInGroup}
```

4ian (approved):

4ian:
Thank you! Double check flow warning you that `values` might not exist, you may want to avoid destructuring or slap a FlowFixMe on that.

PR #4071, opened 2022-06-28
Add GDevelop own cloud storage provider

AlexandreSi:
Add a storage provider managed by GDevelop to facilitate project saving on browser app.

TODO:

- [x] Handle what happens when user goes in Project Manager > Save As.
- [x] Add list of cloud projects in future homepage
- [x] Add possibility to delete a cloud project (in homepage)
- [ ] ~~Add possibility to duplicate a cloud project (in homepage)~~ this feature will be added in a future work
- [x] Make sure failure during saving are well-handled
  - The already implemented logic that raises errors with alerts when saving fails is working. 
- [x] Make sure the * near the project name is relevant (might appear after the project was saved following a name change)
- [x] Add logic to re-authenticate when cookie is expired
- [x] Make so that only project from the local file storage provider can be reopened at startup (on electron)
- [x] Update `lastModifiedDate` of a cloud project that has just been saved so that the list of projects in the build section is up to date
  - \>\> Should be refetch the list of cloud projects on each save? Might be costly <<
  - Or should we manually change the lastModifiedDate in `authenticatedUser.cloudProjects`? (if so, let's check that there is no timezone issue)
- [x] Almost the same as above when saving/duplicating (with save as) a cloud project
- [ ] ~~Add an indicator that shows if the project is saved or not~~ Should be ok now with the update on the homepage of the projects that should show that the new "last edited at" date of the current project
- [x] Single file or Folder project select in ProjectPropertiesDialog only makes sense with local file storage provider. Should we add a helper text that mentions this?
- [x] When a user is logging out:
  - [x] We have to make sure that the project resource cookie is erased
  - [ ] ~~If a cloud project is opened, should we close it?~~ for the moment, let's say no

---

New/updated components:

- [Loader modal](https://gdevelop-storybook.s3.amazonaws.com/cloud-save/latest/index.html?path=/story/ui-building-blocks-loadermodal--with-progress)
- [Dialog that allows to chose storage provider](https://gdevelop-storybook.s3.amazonaws.com/cloud-save/latest/index.html?path=/story/storage-providers-writers--user-not-authenticated)
- [Dialog that allows to choose a name when duplicating/saving as a project](https://gdevelop-storybook.s3.amazonaws.com/cloud-save/latest/index.html?path=/story/storage-providers-cloudstorageprovider-cloudsaveasdialog--default)
- [Dangerous dialog that needs the user to enter a string in an input to confirm something](https://gdevelop-storybook.s3.amazonaws.com/cloud-save/latest/index.html?path=/story/ui-building-blocks-confirmdialog--default)


newIDE/app/src/MainFrame/index.js
```diff
@@ -1930,6 +1986,19 @@ const MainFrame = (props: Props) => {
...
+  const onChangeProjectName = async (newName: string): Promise<void> => {
+    if (!currentProject || !currentFileMetadata) return;
+    const storageProviderOperations = getStorageProviderOperations();
+    if (storageProviderOperations.onChangeProjectProperty) {
+      await storageProviderOperations.onChangeProjectProperty(
+        currentProject,
+        currentFileMetadata,
+        { name: newName }
+      );
+      updateWindowTitle();
```
4ian, line 1998: This is a bit "imperative", any chance this could be done automatically in an effect by checking the storage provider and the `currentProject.getName()` directly? :)

AlexandreSi: I'm not sure what to watch in the effect dependencies. `getStorageProvider` does not change, neither does `currentProject`. If the `currentProject` name is modified, the MainFrame component does not see a change in its props.

newIDE/app/src/ProjectsStorage/CloudStorageProvider/CloudProjectWriter.js
```diff
@@ -0,0 +1,175 @@
...
+const createCloudProjectAndInitCommit = async (
+  project: gdProject,
+  authenticatedUser: AuthenticatedUser,
+  name: string
+): Promise<?string> => {
+  try {
+    const cloudProject = await createCloudProject(authenticatedUser, {
+      name,
+    });
+    if (!cloudProject) return;
```
4ian, line 107: Open question: should you throw instead of returning?

newIDE/app/src/ProjectsStorage/ProjectStorageProviders.js
```diff
@@ -81,86 +79,108 @@ const computeInitialFileMetadataToOpen = (
...
-          getStorageProviderOperations: this._getStorageProviderOperations,
-          initialFileMetadataToOpen,
-          getStorageProvider: this._getStorageProvider,
-        })}
-        {renderDialog && renderDialog()}
-      </React.Fragment>
-    );
-  }
-}
+  React.useEffect(
```
4ian, line 159: Can you comment about this effect utility?

newIDE/app/src/ProjectsStorage/StorageProviderListItem.js
```diff
@@ -0,0 +1,90 @@
...
+type Props = {|
+  storageProvider: StorageProvider,
+  onChooseProvider: (storageProvider: StorageProvider) => void,
+|};
+
+const useListItemStyles = makeStyles(theme => {
+  return {
+    root: {
+      border: `1px solid ${theme.palette.divider}`,
+      borderRadius: '5px',
```
4ian, line 27: This is not a UI component 🧐
I'll give this a pass but let's think about it should this appear in another dialog in the future.

newIDE/app/src/UI/Confirm/ConfirmDialog.js
```diff
@@ -0,0 +1,49 @@
...
+|};
+
+function ConfirmDialog(props: Props) {
+  return (
+    <I18n>
+      {({ i18n }) => (
+        <Dialog
+          open={props.open}
+          actions={[
+            <FlatButton
```
4ian, line 28: We should focus this button by default.

newIDE/app/src/UI/Confirm/ConfirmProvider.js
```diff
@@ -0,0 +1,94 @@
...
+      value={{
+        showConfirmDialog: openConfirmDialog,
+        showConfirmDeleteDialog: openConfirmDeleteDialog,
+      }}
+    >
+      {children}
+      {confirmDialogConfig && (
+        <ConfirmDialog
+          open={confirmDialogOpen}
+          onConfirm={() => onConfirm({ deleteDialog: false })}
```
4ian, line 73: nit: Not sure these extra onConfirm/onDismiss help a lot the code clarity?

newIDE/app/src/UI/Confirm/useConfirmDialog.js
```diff
@@ -0,0 +1,31 @@
...
+
+const useConfirmDialog = () => {
+  const { showConfirmDialog, showConfirmDeleteDialog } = React.useContext(
+    ConfirmContext
+  );
+
+  const getConfirmation = (
+    options: ShowConfirmDialogOptions
+  ): Promise<boolean> =>
+    new Promise((resolve, reject) => {
```
4ian, line 17: Don't write reject parameter if you don't intend to use it, to show to the reader that "this will ALWAYS resolve at some point"

newIDE/app/src/Utils/GDevelopServices/Project.js
```diff
@@ -0,0 +1,262 @@
...
+  currentVersion?: string,
+  deletedAt?: string,
+|};
+
+export type CloudProjectWithUserAccessInfo = {|
+  ...CloudProject,
+  lastModifiedAt: string,
+|};
+
+const reAuthenticateAndRetryIfFailed = async <T>(
```
4ian, line 28: nit: maybe be a bit more precise on what "authenticate" means here, it's actually the project credentials.
```suggestion
const refetchCredentialsForProjectAndRetryIfFailed = async <T>(
```

newIDE/app/src/UI/Confirm/ConfirmProvider.js
AlexandreSi: True, I'll clean it

newIDE/app/src/MainFrame/EditorContainers/HomePage/BuildSection.js
```diff
@@ -42,110 +112,286 @@ const BuildSection = ({
...
+      confirmText: projectName,
+    });
+    if (!deleteAnswer) return;
+
+    try {
+      setPendingProject(fileMetadata.fileIdentifier);
+      await deleteCloudProject(authenticatedUser, fileMetadata.fileIdentifier);
+      authenticatedUser.onCloudProjectsChanged();
+    } catch (error) {
+      console.error(
```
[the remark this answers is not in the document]
AlexandreSi: Yes, good idea!

newIDE/app/src/ProjectsStorage/CloudStorageProvider/CloudProjectWriter.js
AlexandreSi: For clarity's sake, it's better so I changed it. But in the end, it won't appear as an error to the user since it will return `wasSaved: false`

newIDE/app/src/ProjectsStorage/StorageProviderListItem.js
AlexandreSi: I thought about creating a UI component but it felt very specialized for this use.
We could create a UI component if we want to use it elsewhere. At that time, it will be easier to know what needs to be put in common

4ian: Ok!

newIDE/app/src/MainFrame/index.js
```diff
@@ -2002,7 +2091,10 @@ const MainFrame = (props: Props) => {
 
     let state: ?State;
     if (project) state = await loadFromProject(project, fileMetadata);
-    else if (!!fileMetadata) state = await openFromFileMetadata(fileMetadata);
+    else if (!!fileMetadata) {
+      if (storageProvider) getStorageProviderOperations({ storageProvider });
```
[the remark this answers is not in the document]
AlexandreSi: It is not needed, it's a remain from my first strategy

newIDE/app/src/ProjectsStorage/CloudStorageProvider/CloudProjectWriter.js
```diff
@@ -0,0 +1,175 @@
...
+  properties: { name: string }
+): Promise<boolean> => {
+  if (!authenticatedUser.authenticated) return false;
+  try {
+    await updateCloudProject(
+      authenticatedUser,
+      fileMetadata.fileIdentifier,
+      properties
+    );
+    const newVersion = await zipProjectAndCommitVersion({
```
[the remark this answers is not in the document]
AlexandreSi: It only saves if the name is changed, and only if it's a cloud stored project so I'm not sure how to proceed with this. I agree with you that it should mention it will save though.

AlexandreSi:
Maybe a thing to do:
When duplicating a cloud project with the save as option, it takes quite some time (4 or 5 api calls).
We display the snack message "Saving..." but it disappears after 3s, and then there is around 3 seconds of emptiness, and then the message "Project properly saved". It doesn't feel right to me but it could be me being nitpicky.

4ian:
> We display the snack message "Saving..." but it disappears after 3s, and then there is around 3 seconds of emptiness

Any chance we could "simply" increase the timeout, or even not dismiss this snackbar until we get the confirmation of the saving? Not a big deal though.

AlexandreSi:
> > We display the snack message "Saving..." but it disappears after 3s, and then there is around 3 seconds of emptiness
> 
> Any chance we could "simply" increase the timeout, or even not dismiss this snackbar until we get the confirmation of the saving? Not a big deal though.

Oh yes, I didn't realize the saving messages are 2 out of the 3 uses of the snack bar. The third one is to display a problem with the auto save. So I'll increase the timeout to 5s

newIDE/app/src/MainFrame/EditorContainers/HomePage/BuildSection.js
```diff
@@ -34,6 +50,61 @@ type Props = {|
   onCreateProject: () => void,
 |};
 
+const getRelativeOrAbsoluteDisplayDate = (
```
4ian, line 53: Could this be in an `Utils` file? I fear these big files.

newIDE/app/src/MainFrame/EditorContainers/HomePage/BuildSection.js
```diff
@@ -34,6 +50,61 @@ type Props = {|
...
+    now.getFullYear() === date.getFullYear() &&
+    now.getMonth() === date.getMonth() &&
+    sevenDaysAgoAtFirstHour.getTime() <= date.getTime()
+  ) {
+    return i18n._(t`This week`);
+  }
+  return i18n.date(date);
+};
+
+const PrettyBreakablePath = ({ path }: { path: string }) => {
```
4ian, line 94: {| |}

newIDE/app/src/Utils/GDevelopServices/Project.js
```diff
@@ -0,0 +1,271 @@
...
+  authenticatedUser: AuthenticatedUser,
+  cloudProjectId: string,
+  zippedProject: any
+): Promise<?string> => {
+  const { getAuthorizationHeader, firebaseUser } = authenticatedUser;
+  if (!firebaseUser) return;
+
+  const { uid: userId } = firebaseUser;
+  const authorizationHeader = await getAuthorizationHeader();
+  // get presigned url to upload new version (that contains new version id)
```
4ian, line 114: Ubernit: capitals + periods in comments.
```suggestion
  // Get a presigned url to upload a new version (the URL will contain the new version id).
```

newIDE/app/src/Profile/AuthenticatedUserProvider.js
```diff
@@ -186,10 +195,36 @@ export default class AuthenticatedUserProvider extends React.Component<
...
-    const firebaseUser = await this._reloadFirebaseProfile();
-    if (!firebaseUser) return;
+    let firebaseUser;
+    try {
+      firebaseUser = await this._reloadFirebaseProfile();
+      if (!firebaseUser) {
+        this.setState(({ authenticatedUser }) => ({
+          authenticatedUser: {
+            ...authenticatedUser,
+            loginState: 'done',
```
4ian, line 214: Maybe a dumb question, but will `loginState: 'done'` be set if there is no error?

AlexandreSi: I'm not sure I see which case you're referring to

newIDE/app/src/ProjectsStorage/CloudStorageProvider/CloudProjectOpener.js
```diff
@@ -0,0 +1,56 @@
...
+import { type MessageDescriptor } from '../../Utils/i18n/MessageDescriptor.flow';
+import { type AuthenticatedUser } from '../../Profile/AuthenticatedUserContext';
+import { type FileMetadata } from '..';
+
+const unzipProject = async (zippedProject: Blob) => {
+  const zipJs: ZipJs = await initializeZipJs();
+
+  return new Promise((resolve, reject) => {
+    try {
+      zipJs.createReader(new zipJs.BlobReader(zippedProject), zipReader => {
```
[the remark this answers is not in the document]
AlexandreSi: I ended up using the `onerror` callback in createWriter/Reader.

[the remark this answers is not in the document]
AlexandreSi: It should cover errors that occurs deeper

newIDE/app/src/ProjectsStorage/CloudStorageProvider/index.js
```diff
@@ -0,0 +1,45 @@
...
+    onOpen: generateOnOpen(authenticatedUser),
+    onSaveProject: generateOnSaveProject(authenticatedUser),
+    onSaveProjectAs: generateOnSaveProjectAs(
+      authenticatedUser,
+      setDialog,
+      closeDialog
+    ),
+    onChangeProjectProperty: generateOnChangeProjectProperty(authenticatedUser),
+    getOpenErrorMessage: (error: Error): MessageDescriptor => {
+      return t`Check that the file exists, that this file is a proper game created with GDevelop and that you have the authorizations to open it.`;
```
ClementPasteau, line 42: I'd make `authorization` singular

newIDE/app/src/ProjectsStorage/StorageProviderListItem.js
```diff
@@ -0,0 +1,90 @@
...
+              {shouldDisplayAuthenticationButtons && (
+                <Line noMargin>
+                  <FlatButton
+                    label={'Create an Account'}
+                    onClick={() => authenticatedUser.onCreateAccount()}
+                  />
+                  <Spacer />
+                  <RaisedButton
+                    primary
+                    label={'Log-in to GDevelop'}
```
ClementPasteau, line 77: missing trans?

newIDE/app/src/UI/Confirm/ConfirmDeleteDialog.js
```diff
@@ -0,0 +1,76 @@
...
+  onDismiss: () => void,
+|};
+
+function ConfirmDeleteDialog(props: Props) {
+  const [textInput, setTextInput] = React.useState<string>('');
+  const canConfirm = textInput === props.confirmText;
+
+  const onConfirm = () => {
+    if (!canConfirm) return;
+    props.onConfirm();
```
ClementPasteau, line 29: can this error for some reason? should we try catch it to not continue clearing the text input?

ClementPasteau: I read the Provider below, and it looks like the onConfirm does: `setConfirmDeleteDialogOpen(false);` 
So it should be fine and should not error.
However, you're trying to empty the text input after hiding the dialog, so your console is probably unhappy about that?
(I'm not sure you need the line below to clear the input?)

newIDE/app/src/Utils/GDevelopServices/ApiConfigs.js
```diff
@@ -41,44 +35,62 @@ export const GDevelopFirebaseConfig = {
   messagingSenderId: '44882707384',
 };
 
+export const GDevelopBuildApi = {
+  baseUrl: isDev
+    ? 'https://api-dev.gdevelop.io/build'
```
ClementPasteau, line 40: youhouuuuuu 🚀

newIDE/app/src/Utils/GDevelopServices/Project.js
```diff
@@ -0,0 +1,271 @@
...
+  currentVersion?: string,
+  deletedAt?: string,
+|};
+
+export type CloudProjectWithUserAccessInfo = {|
+  ...CloudProject,
+  lastModifiedAt: string,
+|};
+
+const refetchCredentialsForProjectAndRetryIfFailed = async <T>(
```
ClementPasteau, line 28: maybe replace `Failed` with `Unauthorized` as this is what really happens?

newIDE/app/src/Utils/GDevelopServices/Project.js
```diff
@@ -0,0 +1,271 @@
...
+    if (error.response && error.response.status === 403) {
+      await getCredentialsForCloudProject(authenticatedUser, cloudProjectId);
+      const response = await apiCall();
+      return response;
+    }
+    throw error;
+  }
+};
+
+const getVersionIdFromPath = (path: string): string => {
```
ClementPasteau, line 46: you call that versionId, and yet you're retrieving the filename. Is it the same thing?

newIDE/app/src/Utils/GDevelopServices/Project.js
```diff
@@ -0,0 +1,271 @@
...
+
+  const { uid: userId } = firebaseUser;
+  const authorizationHeader = await getAuthorizationHeader();
+  const response = await axios.get(
+    `${GDevelopProjectApi.baseUrl}/project/${cloudProjectId}/action/authorize`,
+    {
+      headers: { Authorization: authorizationHeader },
+      params: { userId },
+      validateStatus: status => true,
+      withCredentials: true, // Necessary to save cookie returned by the server.
```
ClementPasteau, line 72: do you need to specify withCredentials for all those calls? You can't use the `projectResourcesClient` you've created?

newIDE/app/src/Utils/GDevelopServices/Project.js
```diff
@@ -0,0 +1,271 @@
...
+  await refetchCredentialsForProjectAndRetryIfFailed(
+    authenticatedUser,
+    cloudProjectId,
+    () =>
+      projectResourcesClient.post(
+        `${GDevelopProjectResourcesStorage.baseUrl}${presignedUrl}`,
+        zippedProject,
+        {
+          headers: { 'content-type': 'application/zip' },
+          withCredentials: true,
```
ClementPasteau, line 130: is withCredentials needed when you use projectResourcesClient?

newIDE/app/src/Utils/GDevelopServices/Project.js
```diff
@@ -0,0 +1,271 @@
...
+};
+
+export const getPresignedUrlForVersionUpload = async (
+  authenticatedUser: AuthenticatedUser,
+  cloudProjectId: string
+): Promise<string> => {
+  const { getAuthorizationHeader, firebaseUser } = authenticatedUser;
+  if (!firebaseUser)
+    throw new Error(
+      'Presigned url is could not be created, user not authenticated.'
```
ClementPasteau, line 236: `is` can be removed

ClementPasteau: by the way, this is the only place where you throw an error, you could probably just return;

newIDE/app/src/fixtures/GDevelopServicesTestData/index.js
```diff
@@ -41,6 +42,21 @@ export const indieVerifiedFirebaseUser: FirebaseUser = {
   emailVerified: true,
 };
 
+export const cloudProjectsForIndieUser: Array<CloudProjectWithUserAccessInfo> = [
+  {
+    id: 'af7a8282-746d-4d3a-8cb8-bb8cd9372143',
+    name: 'Worms 2D',
```
ClementPasteau, line 48: 🪱

newIDE/app/src/fixtures/GDevelopServicesTestData/index.js
```diff
@@ -41,6 +42,21 @@ export const indieVerifiedFirebaseUser: FirebaseUser = {
...
+    createdAt: '2022-02-05T00:36:53.972Z',
+    lastModifiedAt: '2022-02-07T00:36:53.972Z',
+  },
+  {
+    id: 'fb4d878a-1935-4916-b681-f9235475d35c',
+    name: 'Crash Bandicoot',
```
ClementPasteau, line 54: 🎁

newIDE/app/src/stories/componentStories/ConfirmDialog.stories.js
```diff
@@ -0,0 +1,54 @@
+/* eslint-disable import/no-anonymous-default-export */
```
ClementPasteau, line 1: nice, this is annoyingly impacting all our stories files.
Maybe we could exclude this rule for all the stories folder: https://stackoverflow.com/a/41316801

newIDE/app/src/UI/Dialog/index.js
```diff
@@ -106,29 +122,30 @@ export const DialogPrimaryButton = RaisedButton;
...
-    fullHeight,
-    noTitleMargin,
-    id,
-    cannotBeDismissed,
-  } = props;
-
+const Dialog = ({
+  onApply,
+  secondaryActions,
+  isDangerous,
```
ClementPasteau, line 128: You use this prop only for the DeleteDialog if i'm not mistaken?
the naming is a bit confusing, are we planning to use it for further dialogs?
If so, should it be something like "isNewDialogDesign"?

newIDE/app/src/MainFrame/index.js
```diff
@@ -449,6 +422,63 @@ const MainFrame = (props: Props) => {
     []
```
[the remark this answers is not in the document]
AlexandreSi: It feels like to me it's self explanatory: if it's an empty list, it's equivalent to `componentDidMount`.
Is it okay for you?

newIDE/app/src/MainFrame/index.js
```diff
@@ -449,6 +422,63 @@ const MainFrame = (props: Props) => {
...
+          );
+          if (proceed)
+            openFromFileMetadataWithStorageProvider(
+              fileMetadataAndStorageProviderName
+            );
+        } else {
+          // Open the intro dialog if not opening any project.
+          if (introDialog) openIntroDialog(true);
+        }
+        isInitialProjectLoadingDone.current = true;
```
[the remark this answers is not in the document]
AlexandreSi: Good question! I think it's because since everything happens at start and through effects, I wanted to make sure I have a variable that does not trigger re-renders

newIDE/app/src/MainFrame/index.js
```diff
@@ -2330,14 +2444,18 @@ const MainFrame = (props: Props) => {
...
+        message={fileMetadataOpeningMessage}
+      />
       <HelpFinder
         open={helpFinderDialogOpen}
         onClose={() => openHelpFinderDialog(false)}
       />
       <Snackbar
         open={state.snackMessageOpen}
-        autoHideDuration={3000}
+        autoHideDuration={5000}
```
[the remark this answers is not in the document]
AlexandreSi: Maybe it's because I'm on dev, lambda's are on cold start, but yeah it feels long. I could also add the loader modale that displays messages.

newIDE/app/src/ProjectsStorage/StorageProviderListItem.js
AlexandreSi: Yep, thank you!

newIDE/app/src/Utils/GDevelopServices/ApiConfigs.js
AlexandreSi: ⏩

newIDE/app/src/Utils/GDevelopServices/Project.js
AlexandreSi: Well the version is the filename of the zip file to it kinda makes sense

newIDE/app/src/Utils/GDevelopServices/Project.js
AlexandreSi: Good idea, I did the change

newIDE/app/src/Utils/GDevelopServices/Project.js
AlexandreSi: I cleaned all the uses of axios is the file! Thanks for giving me the motivation to do it 😅

newIDE/app/src/stories/componentStories/ConfirmDialog.stories.js
AlexandreSi: Done, thanks!

newIDE/app/src/UI/Dialog/index.js
AlexandreSi: We could use it for other dialogs yes.
There is already a notion of danger in Material ui to set the usual red color so the naming seemed ok.

newIDE/app/src/UI/Confirm/ConfirmDeleteDialog.js
AlexandreSi: The dialog is used in the provider, so at the top level of the app. As a consequence, the component is never unmounted. So at the next dialog appearance, the former text in the input appears.

newIDE/app/src/Profile/AuthenticatedUserContext.js
```diff
@@ -13,7 +14,9 @@ export type AuthenticatedUser = {|
   authenticated: boolean,
   firebaseUser: ?FirebaseUser,
   profile: ?Profile,
+  loginState: 'justOpened' | 'loading' | 'done',
```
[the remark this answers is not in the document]
AlexandreSi: I ended up using `null | 'loggingIn' | 'done'`

newIDE/app/src/ProjectsStorage/CloudStorageProvider/CloudProjectWriter.js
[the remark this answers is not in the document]
AlexandreSi: Good idea, I added it

newIDE/app/src/UI/Confirm/ConfirmDeleteDialog.js
AlexandreSi: The dialog is used in the provider, so at the top level of the app. As a consequence, the component is never unmounted. So at the next dialog appearance, the former text in the input appears.

newIDE/app/src/MainFrame/index.js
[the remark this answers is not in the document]
ClementPasteau: hmm, I'm still confused about the eslint above.
The goal of the message is for the next person to not remove the disable-eslint, because there's a message explaining why we're doing it only once (and why we shouldn't do it for every prop change)

newIDE/app/src/MainFrame/index.js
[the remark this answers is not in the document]
ClementPasteau: alright 👍

newIDE/app/src/MainFrame/index.js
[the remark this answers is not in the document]
ClementPasteau: I think 3s should be enough on prod yeah. I'm not sure what you mean by loader modale?

newIDE/app/src/UI/Confirm/ConfirmDeleteDialog.js
ClementPasteau: oh right 👍

newIDE/app/src/Utils/GDevelopServices/Project.js
ClementPasteau: ok, I got confused to see different things named the same at first read. If you know it it's fine though.
Maybe it's worth a tiny comment for future contributors? // the filename is in the format `xxxid.zip`?

newIDE/app/src/Utils/GDevelopServices/Project.js
ClementPasteau: 🚀 🎸

newIDE/app/src/stories/componentStories/ConfirmDialog.stories.js
ClementPasteau: no, thank you!

newIDE/app/src/UI/Dialog/index.js
ClementPasteau: alright 👍

newIDE/app/src/ProjectsStorage/CloudStorageProvider/CloudProjectWriter.js
```diff
@@ -0,0 +1,182 @@
...
+      project,
+      cloudProjectId: fileMetadata.fileIdentifier,
+    });
+    if (!newVersion) {
+      throw new Error("Couldn't save project following property update.");
+    }
+
+    return true;
+  } catch (error) {
+    // TODO: Determine if a feedback should be given to user so that they can try again if necessary.
```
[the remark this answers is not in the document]
AlexandreSi: It's a bit difficult here.
They were changing the project properties, they got a warning saying that the project would be saved.
At this point, the properties of the project have been changed but:
- Either the name of the cloud project (in our api) has not been changed (because the error comes from `updateCloudProject`)
- Either the name of the cloud project has been saved but a new version has not been uploaded.
In both cases, when they return to the project properties dialog, they see things just as they think they would be, with no discrepancies.

I was thinking of having a snack that appears in the lower right side where you can hit retry so that the api calls can be sent again. If it succeeds, great, but if not, I'm not sure what to do.

Since this component will be added later, I thought that this TODO could wait, having faith in this api calls.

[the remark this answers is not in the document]
ClementPasteau: alright 👍 
It could indeed be a small snackbar saying "there was an error while saving your project, hit ctrl+s to try again"

newIDE/app/src/MainFrame/index.js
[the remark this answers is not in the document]
AlexandreSi: Capture d’écran 2022-07-28 à 11 07 48

I'm taking about this

newIDE/app/src/Utils/GDevelopServices/Project.js
AlexandreSi: Good idea!

newIDE/app/src/MainFrame/index.js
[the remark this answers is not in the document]
ClementPasteau: I didn't see your conversation above about the time it takes to duplicate.
Ideally we'd close this snackbar manually instead of having a timeout, but probably fine to go with 5 seconds for the moment yes 👍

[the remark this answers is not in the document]
AlexandreSi: I changed it back to 3s

ClementPasteau (approved):
LGTM! ☁️ 🌩️ ⛈️ 🌧️ 🌨️ 🌥️ 🌦️ 🌤️ 🌞

4ian (approved):
- There is a double confirm dialog to delete a project, I think just the final confirm dialog (where you have to type the name) is enough.
-

newIDE/app/src/Profile/AuthenticatedUserProvider.js
4ian: I was mentioning the "happy path", where `firebaseUser = await this._reloadFirebaseProfile();` works and does not return a null object. But this is handled at the end of the function already :)

PR #4073, opened 2022-06-28
Avoid bad expression warnings in the console when using force actions

4ian:
See the issue here: https://forum.gdevelop.io/t/fix-these-code-problems/40436/4?u=willianholtz


AlexandreSi (approved):
Neat

PR #4074, opened 2022-06-29
Bust ghost Physics2 behaviors

### LINKED ISSUE #3525: DestroyOutside behavior leaves "ghost" physics colliders (not visible, but take up space)
## Describe the bug

DestroyOutside behavior leaves "ghost" physics colliders (not visible, but take up space).

## To Reproduce

1. Add "DestroyOutside" behavior to an object with Physics behavior
2. Make the object leave the screen and then come back.
3. The object is no longer be visible but it still takes up space in the physics engine

## Play now

https://liluo.io/games/bf2f4d2a-083d-4cc6-affa-e25374765511

## Other details

* Windows 10
* GDevelop desktop, version 5.0.125 and 5.0.126

## Project files

[CatSpawnerWithGhostCatsBug.zip](https://github.com/4ian/GDevelop/files/7909800/CatSpawnerWithGhostCatsBug.zip)

## Video

https://user-images.githubusercontent.com/8879811/150453640-9a6327d1-0a62-42a0-b69d-3cd6cdb6581d.mp4

### LINKED ISSUE #4060: Physics collision is not being detected in 5.0.138
When updating my game [Ball Challenge 2](https://www.crazygames.com/game/ball-challenge-2)  I noticed something strange with the physics.

I updated to version 5.0.138 but the physics collision bug still persists.
In version 5.0.136 everything worked perfectly, as in the video below. (note the sound and bounce)
https://user-images.githubusercontent.com/89740251/175190696-68f25ae1-d646-4d03-b9e6-76b2d3d4817b.mp4

When the ball touches the ground it makes a sound and has the effect of bouncing.

In the video below, in the version 5.0.138 note that the debug text doesn't check for non-collision with the ground, so the effects that were supposed to occur when the ball bounced don't work correctly.
https://user-images.githubusercontent.com/89740251/175190701-1c691c42-2d0f-453f-b437-edea0acafd82.mp4

Here is the debug event that checks for collision.
bug2

Things I've already tried:
- Remove and put the physics behavior;
- Remove the trigger once;
- Change the ball's collision;

AlexandreSi:
Closes #3525

Fixed when trying to fix #4060. A fix for this issue will come soon.

### DIFF
--- a/Extensions/Physics2Behavior/physics2runtimebehavior.ts
+++ b/Extensions/Physics2Behavior/physics2runtimebehavior.ts
@@ -379,6 +379,9 @@ namespace gdjs {
     }
 
     onDeActivate() {
+      this.contactsEndedThisFrame.length = 0;
+      this.contactsStartedThisFrame.length = 0;
+      this.currentContacts.length = 0;
       if (this._body !== null) {
         // When a body is deleted, Box2D removes automatically its joints, leaving an invalid pointer in our joints list
         this._sharedData.clearBodyJoints(this._body);
--- a/Extensions/Physics2Behavior/physics2runtimebehavior.ts
+++ b/Extensions/Physics2Behavior/physics2runtimebehavior.ts
@@ -393,6 +393,10 @@ namespace gdjs {
       }
     }
 
+    onActivate() {
+      this._updateBodyFromObject();
+    }
+
     onDestroy() {
       this.onDeActivate();
     }
--- a/Extensions/Physics2Behavior/physics2runtimebehavior.ts
+++ b/Extensions/Physics2Behavior/physics2runtimebehavior.ts
@@ -645,8 +645,7 @@ namespace gdjs {
     recreateShape() {
       // If there is no body, set a new one
       if (this._body === null) {
-        this.createBody();
-        return;
+        if (!this.createBody()) return;
       }
 
       // Destroy the old shape
@@ -677,7 +676,8 @@ namespace gdjs {
       return this._body;
     }
 
-    createBody() {
+    createBody(): boolean {
+      if (!this.activated()) return false;
       // Generate the body definition
       const bodyDef = new Box2D.b2BodyDef();
 
@@ -715,12 +715,13 @@ namespace gdjs {
       // Update cached size
       this._objectOldWidth = this.owner.getWidth();
       this._objectOldHeight = this.owner.getHeight();
+      return true;
     }
 
     doStepPreEvents(runtimeScene) {
       // Create a body if there is not one
       if (this._body === null) {
-        this.createBody();
+        if (!this.createBody()) return;
       }
 
       // Step the world if not done this frame yet
@@ -775,7 +776,7 @@ namespace gdjs {
     _updateBodyFromObject() {
       // If there is no body, set a new one
       if (this._body === null) {
-        this.createBody();
+        if (!this.createBody()) return;
       }
 
       // The object size has changed, recreate the shape.
@@ -872,8 +873,7 @@ namespace gdjs {
 
       // If there is no body, set a new one
       if (this._body === null) {
-        this.createBody();
-        return;
+        if (!this.createBody()) return;
       }
 
       // Update body type
@@ -896,8 +896,7 @@ namespace gdjs {
 
       // If there is no body, set a new one
       if (this._body === null) {
-        this.createBody();
-        return;
+        if (!this.createBody()) return;
       }
 
       // Update body type
@@ -920,8 +919,7 @@ namespace gdjs {
 
       // If there is no body, set a new one
       if (this._body === null) {
-        this.createBody();
-        return;
+        if (!this.createBody()) return;
       }
 
       // Update body type
@@ -944,8 +942,7 @@ namespace gdjs {
 
       // If there is no body, set a new one
       if (this._body === null) {
-        this.createBody();
-        return;
+        if (!this.createBody()) return;
       }
 
       // Update body bullet flag
@@ -959,8 +956,7 @@ namespace gdjs {
     setFixedRotation(enable): void {
       this.fixedRotation = enable;
       if (this._body === null) {
-        this.createBody();
-        return;
+        if (!this.createBody()) return;
       }
       this._body.SetFixedRotation(this.fixedRotation);
     }
@@ -972,8 +968,7 @@ namespace gdjs {
     setSleepingAllowed(enable): void {
       this.canSleep = enable;
       if (this._body === null) {
-        this.createBody();
-        return;
+        if (!this.createBody()) return;
       }
       this._body.SetSleepingAllowed(this.canSleep);
     }
@@ -981,7 +976,7 @@ namespace gdjs {
     isSleeping(): boolean {
       // If there is no body, set a new one
       if (this._body === null) {
-        this.createBody();
+        if (!this.createBody()) return true;
       }
 
       // Get the body sleeping state
@@ -1008,8 +1003,7 @@ namespace gdjs {
 
       // If there is no body, set a new one
       if (this._body === null) {
-        this.createBody();
-        return;
+        if (!this.createBody()) return;
       }
 
       // Update the body density
@@ -1037,8 +1031,7 @@ namespace gdjs {
 
       // If there is no body, set a new one
       if (this._body === null) {
-        this.createBody();
-        return;
+        if (!this.createBody()) return;
       }
 
       // Update the body friction
@@ -1072,8 +1065,7 @@ namespace gdjs {
 
       // If there is no body, set a new one
       if (this._body === null) {
-        this.createBody();
-        return;
+        if (!this.createBody()) return;
       }
 
       // Update the body restitution
@@ -1102,8 +1094,7 @@ namespace gdjs {
 
       // If there is no body, set a new one
       if (this._body === null) {
-        this.createBody();
-        return;
+        if (!this.createBody()) return;
       }
 
       // Update the body linear damping
@@ -1125,8 +1116,7 @@ namespace gdjs {
 
       // If there is no body, set a new one
       if (this._body === null) {
-        this.createBody();
-        return;
+        if (!this.createBody()) return;
       }
 
       // Update the body angular damping
@@ -1148,8 +1138,7 @@ namespace gdjs {
 
       // If there is no body, set a new one
       if (this._body === null) {
-        this.createBody();
-        return;
+        if (!this.createBody()) return;
       }
 
       // Update the body gravity scale
@@ -1185,8 +1174,7 @@ namespace gdjs {
 
       // If there is no body, set a new one
       if (this._body === null) {
-        this.createBody();
-        return;
+        if (!this.createBody()) return;
       }
 
       // Update the body layers
@@ -1224,8 +1212,7 @@ namespace gdjs {
 
       // If there is no body, set a new one
       if (this._body === null) {
-        this.createBody();
-        return;
+        if (!this.createBody()) return;
       }
 
       // Update the body masks
@@ -1237,8 +1224,7 @@ namespace gdjs {
     getLinearVelocityX(): float {
       // If there is no body, set a new one
       if (this._body === null) {
-        this.createBody();
-        return 0;
+        if (!this.createBody()) return 0;
       }
 
       // Get the linear velocity on X
@@ -1248,7 +1234,7 @@ namespace gdjs {
     setLinearVelocityX(linearVelocityX): void {
       // If there is no body, set a new one
       if (this._body === null) {
-        this.createBody();
+        if (!this.createBody()) return;
       }
 
       // Set the linear velocity on X
@@ -1263,8 +1249,7 @@ namespace gdjs {
     getLinearVelocityY(): float {
       // If there is no body, set a new one
       if (this._body === null) {
-        this.createBody();
-        return 0;
+        if (!this.createBody()) return 0;
       }
 
       // Get the linear velocity on Y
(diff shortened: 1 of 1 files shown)


Extensions/Physics2Behavior/physics2runtimebehavior.ts
```diff
@@ -377,6 +379,9 @@ namespace gdjs {
     }
 
     onDeActivate() {
+      this.contactsEndedThisFrame.length = 0;
```
4ian, line 382: Any risk doing this if we deactivate and then reactive the behavior during the same frame?

AlexandreSi: I guess that the object holding the behavior will forget about its contacts, but not the other objects it was in contact with. So I would say it's not too risky

4ian: Related question: should there be a `if (this._body !== null)` in onContactEnd?

Like, I'm deactivating a physics behavior of an object, so the body is deleted, so the contact with another objects stops.... so `EndContact` is called and the other object is added back to `contactsEndedThisFrame`, despite you just clearing this array before (and it will stay there until it's reactivated and doStepPreEvents is called).

Extensions/Physics2Behavior/physics2runtimebehavior.ts
```diff
@@ -393,7 +398,12 @@ namespace gdjs {
...
     }
 
+    onActivate() {
+      this._updateBodyFromObject();
+    }
+
     onDestroy() {
+      this.destroyedDuringFrameLogic = true;
       this.onDeActivate();
     }
```
D8H, line 408: Should `activate(enable: boolean): void` be called instead?
It will change the value returned by `activated(): boolean` and make `destroyedDuringFrameLogic` useless. It will ensure that `onDeActivate()` won't be called twice.

D8H:
Is there something preventing to write a test case?

Extensions/Physics2Behavior/physics2runtimebehavior.ts
4ian: I would avoid to call `activate(false)` because:
- this is an operation that could involves more stuff later. This is implemented in the runtime behavior, and here we call onDeActivate because "it's the same code", but not because we want to deactivate the behavior. In other words, the code paths could differ in the future, the fact that they are the same is just a coincidence.
- onDestroy should already be called only once anyway?

D8H: Ok, deleting a deactivated instance seem to be a good test case.

AlexandreSi:
> Is there anything preventing from writing a test case?

Nope! I was curious to see if something would shock you with those changes. I'll add the necessary tests

Extensions/Physics2Behavior/physics2runtimebehavior.ts
AlexandreSi: I was wondering about that at one point. I'll with the tests if this can be checked.

Extensions/Physics2Behavior/physics2runtimebehavior.ts
```diff
@@ -673,7 +685,8 @@ namespace gdjs {
       return this._body;
     }
 
-    createBody() {
+    createBody(): boolean {
+      if (!this.activated() || this.destroyedDuringFrameLogic) return false;
```
fannieyan, line 689: Could you explain to me what is the purpose of `this.destroyedDuringFrameLogic` please? :)
Do you have a case in mind when createBody would be called while `this.destroyedDuringFrameLogic` is `true`?

fannieyan (commented):
I just have a question on implementation, but apart from that it looks ok to me (apart the fact that the condition "collision has ended" will always return true after the behaviour is deactivated, but I will be working on it ;)).

Extensions/Physics2Behavior/physics2runtimebehavior.ts
AlexandreSi: When going through events, you can have, in order:
- A. If XXX, apply impulse to Object
- B. if YYY, destroy Object
- C. if ZZZ, set linear damping of Object

If all three conditions are true, when going through B, you destroy the object, and so you do for the box2d body.
When going through C, if you don't know that you destroyed the body, you will recreate it, leaving a ghost body (behavior without a runtime object linked to it).

And keep in mind that instead of having a C event, this can happen during post events.

D8H: I think that no action can be called on a destroyed instance because destroyed instances can't be picked.

fannieyan:
We made up a project with a set of actions to test out collisions:
- An A ball bouncing between cross balls with a lot of collisions that could go undetected. When a collision is detected and the count is incremented by one.
- A coin on a platform that makes the background flash green when they are in collision. The coin will be resized and behavior activated/deactivated

/ ! \ Beware of rapidly flashing colors / ! \
<details><summary>Gifs</summary>

| Bouncing ball | (De)Activating behavior | Resizing |
| -------------- | ----------------------- | --------- |
|  |  |  |

</details>

Extensions/Physics2Behavior/physics2runtimebehavior.ts
AlexandreSi: Oh I didn't think of that.

Extensions/Physics2Behavior/tests/physics2runtimebehavior.spec.js
```diff
@@ -103,7 +103,185 @@ function createObject(runtimeScene, behaviorProperties) {
   return object;
 }
 
-describe('Physics2RuntimeBehavior', () => {
```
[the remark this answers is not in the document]
fannieyan: The same error 😅

Extensions/Physics2Behavior/physics2runtimebehavior.ts
```diff
@@ -244,6 +244,7 @@ namespace gdjs {
     contactsStartedThisFrame: Array<Physics2RuntimeBehavior>;
```
4ian, line 244: can you comment for each of these arrays its responsibility (what does it hold, when it's updated)?

Extensions/Physics2Behavior/tests/physics2runtimebehavior.spec.js
```diff
@@ -104,6 +104,184 @@ function createObject(runtimeScene, behaviorProperties) {
...
+        // It should be false because the condition does not have sense anymore
+        // since the behavior is deactivated.
+        stopped: false,
+      });
+
+      runtimeScene.renderAndStep(1000 / fps);
+      // Objects should have 0 contacts in memory.
+      expect(object1Behavior.currentContacts.length).to.be(0);
+      expect(object1Behavior.contactsEndedThisFrame.length).to.be(0);
+      expect(object1Behavior.contactsStartedThisFrame.length).to.be(0);
```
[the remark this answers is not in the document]
fannieyan: Indeed!

Extensions/Physics2Behavior/tests/physics2runtimebehavior.spec.js
```diff
@@ -104,6 +104,184 @@ function createObject(runtimeScene, behaviorProperties) {
...
+      behavior.getMassCenterX();
+
+      // Object's body should still not exist
+      expect(behavior.getBody()).to.be(null);
+
+      // Reactivate behavior
+      object.activateBehavior('Physics2', true);
+      expect(behavior.getBody()).not.to.be(null);
+
+      // Behavior should have recorded the above setter called.
```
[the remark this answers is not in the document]
fannieyan: I think the comment meant "Behavior should have recorded what was called with its setters while it was de-activated"

I will reword it so

AlexandreSi:
path to `utils.js` should removed from:
- the Physics2 JsExtension file so that it is not included at compilation time
- the karma configuration so that karma does not import it when running tests

Extensions/Physics2Behavior/tests/physics2runtimebehavior.spec.js
```diff
@@ -232,6 +514,257 @@ describe('Physics2RuntimeBehavior', () => {
...
+        collision: true,
+        stopped: false,
+      });
+
+      // Destroy (postEvent operation).
+      runtimeScene.setEventsFunction(() => {
+        movingObject.deleteFromScene(runtimeScene);
+
+        // Collision should be reset on destroyed object and
+        // added to contactsStoppedThisFrame array of the other object.
```
AlexandreSi, line 743: contactsEndedThisFrame

Extensions/Physics2Behavior/tests/physics2runtimebehavior.spec.js
```diff
@@ -318,4 +857,104 @@ describe('Physics2RuntimeBehavior', () => {
...
+          );
+          behaviorTest.setExpectedCollisions(expectedCollisionTest);
+        }
+      });
+      while (stepIndex < 10 && !hasBounced) {
+        runtimeScene.renderAndStep(1000 / fps);
+        stepIndex++;
+      }
+
+      // Should be cleared at next step.
```
AlexandreSi, line 948: Same question as above

Extensions/Physics2Behavior/tests/physics2runtimebehavior.spec.js
```diff
@@ -318,4 +857,104 @@ describe('Physics2RuntimeBehavior', () => {
...
     });
   });
+
+  describe('Behavior interaction with other objects behaviors and extensions.', () => {
+    let runtimeGame;
+    let runtimeScene;
+    beforeEach(() => {
+      [runtimeGame, runtimeScene] = createGameWithSceneWithPhysics2SharedData();
+    });
+    it('BehaviorTest should have access to current started and ended contacts in postEvent.', () => {
```
D8H, line 874: Having the asserts during the postEvent of another behavior or after the rendering is likely the same unless you want to test a behavior that run before the Physics one.

But even in this case, as there is no more side effect in the postEvent of Physics, a less strong tests might be ok for the sake of tests readability. I guess a comment to warn to avoid putting side effect in the postEvent should be enough.

Note that `renderAndStep` actually steps and renders.

fannieyan: Mmmh, do you think this test is still relevant then? Since we already assert quite a lot after renderAndStep in other tests?

Extensions/Physics2Behavior/physics2runtimebehavior.ts
```diff
@@ -241,13 +281,29 @@ namespace gdjs {
     layers: any;
     masks: any;
     shapeScale: number = 1;
+    // Array containing the beginning of contacts reported by onContactBegin. Each contact
+    // should be unique to avoid recording glitches where the object loses and regain
+    // contact between two frames. The array is updated each time the method
+    // onContactBegin is called by the listener, which is only called when stepping
+    // the world i.e. in the first preEvent called by a physics behavior. This array is
+    // cleared just before stepping the world.
```
D8H, line 289: This should be a documentation comment. Even if it's a private member, it allow to read the comment by hovering the attribute in the code.

Extensions/Physics2Behavior/tests/physics2runtimebehavior.spec.js
```diff
@@ -232,6 +521,257 @@ describe('Physics2RuntimeBehavior', () => {
...
+        });
+      });
+      runtimeScene.renderAndStep(1000 / fps);
+      assertCollision(movingObject, staticObject, {
+        started: true,
+        collision: true,
+        stopped: false,
+      });
+
+      // Resize (postEvent operation).
```
D8H, line 678: No longer true. Does the test need some adaptations?

D8H:
Some suggestions to make the tests easier to read.

> it('should bounce during the frame (at low fps), collision should be detected, as well as start and stop of the collision'


```
it('should detect a collision even if the contact stated and ended during the same frame"
// Use a low fps to reproduce more easily.
```

> it('should begin to bounce during the frame (at high fps), collision should be detected, as well as start of the collision, but not the end'


```
it('should detect a collision when the contact doesn't end the same frame it started"
// Use a high fps to reproduce more easily.
```


> it('should not detect a contact while already in contact as a new one (the contact jittered).'

```
it('should not detect a new contact while already in contact with that same object (the contact jittered).'
```


In the following test, `onContactEnd` and `onContactBegin` are actually called at the same time. Testing intermediate state has a very low value and make the test unclear. Only the last asset is interesting because it's actually what the events will get.
The comment is not right, it's not during the events that the signals happen.

```JavaScript
      runtimeScene.setEventsFunction(() => {
        // Manually call onContactEnd and onContactBegin methods to simulate
        // a loss of contact followed by a contact beginning during the event.
        movingObject
          .getBehavior('Physics2')
          .onContactEnd(staticObject.getBehavior('Physics2'));

        assertCollision(movingObject, staticObject, {
          started: false,
          collision: false,
          stopped: true,
        });

        movingObject
          .getBehavior('Physics2')
          .onContactBegin(staticObject.getBehavior('Physics2'));

        assertCollision(movingObject, staticObject, {
          started: false,
          collision: true,
          stopped: false,
        });
      });
```

Extensions/Physics2Behavior/tests/physics2runtimebehavior.spec.js
```diff
@@ -1,4 +1,27 @@
 describe('Physics2RuntimeBehavior', () => {
+  class FakeAutoRemoveBehavior extends gdjs.RuntimeBehavior {
+    shouldDeleteInPreEvent = false;
+
+    constructor(runtimeScene, behaviorData, owner) {
+      super(runtimeScene, behaviorData, owner);
+    }
+
+    shouldAutoRemoveInPreEvent() {
```
4ian, line 9: nit:
`setShouldAutoRemoveInPreEvent`

if you start a method name by "shouldXXX" or "hasXXX" or "isXXX", this gives the impression this is returning a boolean. i.e: a getter rather than a setter.

4ian (approved):

4ian:
Let's merge this and make one last update on the issues that were mentioning this :)

PR #4083, opened 2022-07-06
Select copied element on behaviors and functions lists

D8H:
It could be nice to have the function inserted after the copied one but this is another story. I wanted to keep this PR very tiny.

image

### User story
* I want to add a new action to change a property
* I duplicate an existing action to have less work to do
* In most UI I know, the copied element of a list is selected. So, I edit the selected function which is not the one I should edit.
* I rename the new action. Usually, at this point I realize my mistake and I have start from scratch.
* But sometimes, I don't and this happens:
  * https://github.com/GDevelopApp/GDevelop-extensions/issues/484


ClementPasteau:
lgtm 👍 
the same behavior is applied when duplicating objects in the editor, so it makes total sense.
Also, should there be a "duplicate" option like for objects?

image
image

ClementPasteau (approved):

PR #4085, opened 2022-07-06
Add a duplicate action in the functions list contextual menu.

D8H:
I only did it for functions because behaviors are rarely duplicated.

### DIFF
--- a/newIDE/app/src/EventsFunctionsList/index.js
+++ b/newIDE/app/src/EventsFunctionsList/index.js
@@ -271,6 +271,27 @@ export default class EventsFunctionsList extends React.Component<Props, State> {
     this._editName(newEventsFunction);
   };
 
+  _duplicateEventsFunction = (
+    eventsFunction: gdEventsFunction,
+    newFunctionIndex: number
+  ) => {
+    const { eventsFunctionsContainer } = this.props;
+    const newName = newNameGenerator(eventsFunction.getName(), name =>
+      eventsFunctionsContainer.hasEventsFunctionNamed(name)
+    );
+    let newEventsFunction = eventsFunction.clone();
+    newEventsFunction.setName(newName);
+    newEventsFunction = eventsFunctionsContainer.insertEventsFunction(
+      newEventsFunction,
+      newFunctionIndex
+    );
+    this.props.onEventsFunctionAdded(newEventsFunction);
+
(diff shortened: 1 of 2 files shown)


newIDE/app/src/EventsFunctionsList/index.js
```diff
@@ -271,6 +271,27 @@ export default class EventsFunctionsList extends React.Component<Props, State> {
...
+    const { eventsFunctionsContainer } = this.props;
+    const newName = newNameGenerator(eventsFunction.getName(), name =>
+      eventsFunctionsContainer.hasEventsFunctionNamed(name)
+    );
+    let newEventsFunction = eventsFunction.clone();
+    newEventsFunction.setName(newName);
+    newEventsFunction = eventsFunctionsContainer.insertEventsFunction(
+      newEventsFunction,
+      newFunctionIndex
+    );
```
D8H, line 287: It's a bit strange that `insertEventsFunction`does a copy because here a copy is needed to change the name and another one to insert it.

ClementPasteau: it is, but if it works like that for the moment, no real reason to change I guess...

ClementPasteau (approved):
lgtm!

newIDE/app/src/EventsFunctionsList/index.js
4ian: Careful there is a memory leak here

4ian: (I'll let you find/understand it, because it's important to be able to recognise these ;) - you are right to find this code "weird" as it is. Look also at how InsertEventsFunction is working because you might not need the clone at all).

4ian (approved):
That should be the proper solution! ;) 

(or use `clone` but ensure to `delete` it afterwards, because otherwise you have the cloned stuff that will stay in memory forever, as it's a copy that is stored in the project - but this is more complicated than the good solution you have now)

newIDE/app/src/EventsFunctionsList/index.js
4ian: >  it allows to have 2 functions with the same name in the same list

Exact, this is a shortcoming on the design of the API of the list holding the events function.
If the API was perfect, it would probably force you to pass a new name.

PR #4099, opened 2022-07-13
Replace 'Musics' with 'Music'.

BWPanda:


tristanbob:
I can confirm that "musics" is not a word and can be replaced with "music".

4ian:
Alright then, thank you! 👍

PR #4105, opened 2022-07-15
Fix the box of tilemap collision masks in the editor.

D8H:


ClementPasteau (approved):
I haven't tried it, but lgtm!

PR #4106, opened 2022-07-15
[Platformer] Fix: platformer characters can now jump when moving against a wall that is part of the same object as the floor.

D8H:
This could happen when using a tilemap collision mask object and not sustaining the jump.

It's covered by tests but can also be reproduced on this example:
* https://github.com/GDevelopApp/GDevelop-examples/pull/343

### DIFF
--- a/Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
+++ b/Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
@@ -1715,6 +1715,17 @@ namespace gdjs {
         behavior._overlappedJumpThru.push(this._floorPlatform!);
         behavior._setFalling();
       }
+
+      // It was originally in checkTransitionBeforeY.
+      // The character is ignoring the floor when moving on X to be able to
+      // follow up a slope when moving Y (it enter inside it).
+      // When the current floor and the wall the character is facing is part of
+      // the same instance, the wall is also ignored when moving on X, but the
+      // wall is to high to follow and it is seen as colliding an obstacle
+      // from behind.
+      // Moving against a wall before jumping in this configuration was making
+      // jumps being aborted.
+      behavior._checkTransitionJumping();
     }
 
     beforeMovingX() {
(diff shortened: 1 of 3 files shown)


D8H:
This fix has a side effect: with a very low jump speed while climbing a slope, the character can no longer jump.

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -1723,14 +1723,15 @@ namespace gdjs {
       behavior._requestedDeltaX +=
         this._floorPlatform!.owner.getX() - this._floorLastX;
       // See `beforeUpdatingObstacles` for the logic for the Y axis.
+
+      //Jumping
+      behavior._checkTransitionJumping();
```
4ian, line 1728: Can you explain in plain English why this change enabled to fix the issue? :)

4ian (approved):

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
ClementPasteau: +1 I'm not sure I understand why this needs to happen in the beforeMovingX, and not in the checkTransitionBeforeY anymore?
Also, should it happen in the beforeMovingY() too?

D8H: Sorry I meant to move it in `checkTransitionBeforeX` not `beforeMovingX`. I added a comment.

D8H: > Also, should it happen in the beforeMovingY() too?

The check is about the jump key. It won't change during `doStepPreEvents`.

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -1715,6 +1715,17 @@ namespace gdjs {
...
+
+      // It was originally in checkTransitionBeforeY.
+      // The character is ignoring the floor when moving on X to be able to
+      // follow up a slope when moving Y (it enter inside it).
+      // When the current floor and the wall the character is facing is part of
+      // the same instance, the wall is also ignored when moving on X, but the
+      // wall is to high to follow and it is seen as colliding an obstacle
+      // from behind.
+      // Moving against a wall before jumping in this configuration was making
+      // jumps being aborted.
```
D8H, line 1727: I added this comment.

ClementPasteau: nice! just one typo: "to high" => "too high"

ClementPasteau (approved):

PR #4121, opened 2022-07-20
Show typing errors of object parameters directly on the event sheet.

D8H:
The error was already shown in the instruction editor but not directly on the sheet.
image

It took less time to implement than what I've wasted because of an oversight.


4ian (approved):

ClementPasteau:
🚀

D8H:
I did a bug with this. Groups are always underlined.

4ian:
I opened a fix here: https://github.com/4ian/GDevelop/pull/4134 ;)

PR #4126, opened 2022-07-21
Reduce max size of get started tiles

ClementPasteau:
Before:
image

After:
image

### DIFF
--- a/newIDE/app/src/MainFrame/EditorContainers/HomePage/GetStartedSection.js
+++ b/newIDE/app/src/MainFrame/EditorContainers/HomePage/GetStartedSection.js
@@ -21,7 +21,8 @@ const electron = optionalRequire('electron');
 const styles = {
   grid: {
     textAlign: 'center',
-    maxWidth: LARGE_WIDGET_SIZE * 4, // Avoid tiles taking too much space on large screens.
+    // Avoid tiles taking too much space on large screens.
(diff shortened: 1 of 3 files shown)


newIDE/app/src/MainFrame/EditorContainers/HomePage/GetStartedSection.js
```diff
@@ -21,7 +21,8 @@ const electron = optionalRequire('electron');
 const styles = {
   grid: {
     textAlign: 'center',
-    maxWidth: LARGE_WIDGET_SIZE * 4, // Avoid tiles taking too much space on large screens.
+    // Avoid tiles taking too much space on large screens.
+    maxWidth: LARGE_WIDGET_SIZE * 3 + 5 * 6, // 3 widgets + 5 padding per side
```
D8H, line 25: Was the original intent to show 4 columns?
Should it use `getColumnsFromWidth` to always keep the same margin? Maybe something else already handle it?

D8H (approved):

newIDE/app/src/MainFrame/EditorContainers/HomePage/GetStartedSection.js
ClementPasteau: I think the original was to show 3, but because of padding around them, I ended up putting the value at 4 times the width. Didn't really think about edge cases of screen size.
I'll see if I can have it done automatically yes 👀

Issue #4119, opened 2022-07-20
'Expanded/collapsed' arrows on the event sheet are different colours


BWPanda:
## Describe the bug
The arrows used to show expanded or collapsed groups/sub-events are different colours:
- Expanded (down arrow) is white/light
- Collapsed (right arrow) is black/dark

This means that whatever theme you're using, **one of the arrows is barely visible**.

## To Reproduce
1. Go to an event sheet
2. Create some events and sub-events so the arrows appear
3. Click the arrows to expand/collapse the events
4. Note the arrow colour

Dark theme:
![image](https://user-images.githubusercontent.com/2385329/179870009-739974df-4f7f-4b19-92ac-80cd932b1bbc.png)

Light theme:
![image](https://user-images.githubusercontent.com/2385329/179870093-16f0b02a-bc81-4de4-8e44-af3132223783.png)

## Other details
GDevelop v5.0.138 running on Manjaro Linux (KDE Plasma v5.24.6)

BWPanda:
I tried making a PR: https://github.com/4ian/GDevelop/pull/4129 Not sure if I've done it right though (I'm a web developer, not an app developer). Hope it helps though.

PR #4129, opened 2022-07-22
Issue #4119: Fix colours of expanded/collapsed arrows on event sheet.

BWPanda:
Fixes https://github.com/4ian/GDevelop/issues/4119


ClementPasteau (approved):
Thanks @BWPanda !

If we add another light theme in the future, we may want to revisit the logic as it will not inherit the dark arrows.
But for the moment, it's fine 👍

PR #4130, opened 2022-07-22
Fix layer names in drop-down lists that were translated even though they are defined by users.

D8H:
image

image

### DIFF
--- a/newIDE/app/src/LayersList/EnumerateLayers.js
+++ b/newIDE/app/src/LayersList/EnumerateLayers.js
@@ -1,10 +1,10 @@
 import { mapFor } from '../Utils/MapFor';
 
-const enumerateLayers = layersContainer =>
+const enumerateLayers = (layersContainer, baseLayerLabel: string) =>
   mapFor(0, layersContainer.getLayersCount(), i => {
     return {
       value: layersContainer.getLayerAt(i).getName(),
-      label: layersContainer.getLayerAt(i).getName() || 'Base layer',
+      label: layersContainer.getLayerAt(i).getName() || baseLayerLabel,
     };
   });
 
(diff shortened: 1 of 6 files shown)


newIDE/app/src/UI/SelectOption.js
```diff
@@ -10,6 +10,7 @@ type Props = {|
   value: string | number | boolean,
   primaryText: MessageDescriptor,
   disabled?: boolean,
+  areUserDefinedValues?: boolean,
```
ClementPasteau, line 13: This works.
We've also done something different in `MarkdownText.js`
2 different props:
- source & translatableSource

Check it out, maybe we should copy something similar with `translatablePrimaryText` and `primaryText` ?

newIDE/app/src/LayersList/LayerRemoveDialog.js
```diff
@@ -48,13 +58,21 @@ export default class VariablesEditorDialog extends Component {
...
       .filter(({ value }) => {
         return value !== this.props.layerRemoved;
       })
       .map(({ value, label }) => (
-        <SelectOption key={value} value={value} primaryText={label} />
+        <SelectOption
+          key={value}
+          value={value}
+          primaryText={label}
+          areUserDefinedValues
```
ClementPasteau, line 74: should this be true only if label === 'Base layer'?
and then the SelectOption would translate the base layer text, instead of doing it yourself above, and you don't need to change the logic of the `enumerateLayers`

D8H: It would not work for the property editor unless each choice has the boolean.

newIDE/app/src/LayersList/EnumerateLayers.js
```diff
@@ -5,6 +5,7 @@ const enumerateLayers = layersContainer =>
     return {
       value: layersContainer.getLayerAt(i).getName(),
       label: layersContainer.getLayerAt(i).getName() || 'Base layer',
+      labelIsUserDefined: layersContainer.getLayerAt(i).getName(),
```
4ian, line 8: If this is a boolean, you should cast it with !!

4ian (approved):
I like this new approach better :)

PR #4133, opened 2022-07-22
Fix: tilemaps object boxes is now the same as the tilemaps collision mask object boxes.

D8H:
Tilemaps are no longer cropped automatically, because void spaces could be intended. The right size can be set with Tiled if needed.

It seems that it was already fixed by this PR, but it wasn't merged:
* https://github.com/4ian/GDevelop/pull/2296

## Solution
It relies on the tile map dimensions instead of the render dimensions to avoid bad positioning when there is no tile to draw around the map borders.

## Copping issue
* The crop was probably intended to happen only at right and bottom.

TileMapVoidSpaceBug

## Fix : no cropping
* There is no longer any cropping
TileMapVoidSpaceFix

* It follows what the user chose in Tiled (maybe there is a reason, no judgment)
TileMapVoidSpaceTile

* The collision mask were already using a no cropping strategy because void spaces are very expected
TileMapCollisionMaskBox

* It allow to overlap the tilemap and its collision masks more easily
TileMapVoidSpaceFixOverview

## Manual tests

* GDJS
https://user-images.githubusercontent.com/2611977/180506767-744f0127-ce5d-4263-bdbc-becbe25e2875.mp4

* newIDE
https://user-images.githubusercontent.com/2611977/180506817-b0127396-ace7-430b-aa8a-3ab1d36d7998.mp4

### DIFF
--- a/Extensions/TileMap/JsExtension.js
+++ b/Extensions/TileMap/JsExtension.js
@@ -804,19 +804,16 @@ module.exports = {
         const localPosition = new PIXI.Point();
         this._pixiObject.worldTransform.applyInverse(position, localPosition);
 
-        // Check if the point is inside the object bounds
-        const originalWidth = this._pixiObject.width / this._pixiObject.scale.x;
-        const originalHeight =
-          this._pixiObject.height / this._pixiObject.scale.y;
-
         return (
           localPosition.x >= 0 &&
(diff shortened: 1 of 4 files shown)


4ian (approved):
Lgtm from a quick look!

PR #4138, opened 2022-07-25
Fix package URLs crashing the app on web

ClementPasteau:
I'm not sure when this got introduced, probably when we refactored the export dialog to be different tabs.

### DIFF
--- a/newIDE/app/src/Utils/BlobDownloadUrlHolder.js
+++ b/newIDE/app/src/Utils/BlobDownloadUrlHolder.js
@@ -8,27 +8,31 @@ type Props = {
 
 export const BlobDownloadUrlHolder = ({ blob, children }: Props) => {
   const [blobDownloadUrl, setBlobDownloadUrl] = React.useState('');
+  const [stateBlob, setStateBlob] = React.useState(null);
   React.useEffect(
     () => {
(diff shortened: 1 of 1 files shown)


newIDE/app/src/Utils/BlobDownloadUrlHolder.js
```diff
@@ -8,27 +8,31 @@ type Props = {
 
 export const BlobDownloadUrlHolder = ({ blob, children }: Props) => {
   const [blobDownloadUrl, setBlobDownloadUrl] = React.useState('');
+  const [stateBlob, setStateBlob] = React.useState(null);
```
4ian, line 11:
```suggestion
  const [stateBlob, setStateBlob] = React.useState<?Blob>(null);
```

4ian: Let's maybe rename this to `currentBlob`/`setCurrentBlob`?
To show that it's holding the "current value"?

4ian (approved):

4ian:
(introduced in https://github.com/4ian/GDevelop/commit/8202e6b38e0d5f1172b8162875dde970cdc53986 for the record)

PR #4140, opened 2022-07-25
Handle tile maps with invalid resources

D8H:
* A placeholder is used when a resource is invalid
TileMapMaskError TileMapError

* The border allows to:
  * put mask exactly on top of the tile map more easily
  * see empty masks
  * understand rotated instances box
TileMapCollisionMaskPositionning
TileMapCollisionMaskBorder


newIDE/app/src/ObjectsList/index.js
```diff
@@ -65,6 +65,7 @@ const objectTypeToDefaultName = {
   'BitmapText::BitmapTextObject': 'NewBitmapText',
   'TextEntryObject::TextEntry': 'NewTextEntry',
   'TileMap::TileMap': 'NewTileMap',
+  'TileMap::CollisionMask': 'NewTileMapMask',
```
ClementPasteau, line 68: noob question: What is this for?

ClementPasteau (commented):
lgtm overall

newIDE/app/src/ObjectsList/index.js
D8H: It's when you create a new object, it gives a default name to it.

PR #4145, opened 2022-07-26
Round of maintenance on P2P

arthuro555:
- Update PeerJS
   - Fixes for Safari
   - Better typings
   - New default TURN server
 - Add a JavaScript API to get access to the raw connection. 
   - This allows to send custom messages from JavaScript. This allows to use a binary protocol from JavaScript.
 - Use `Map` instead of JavaScript objects in places where it makes sense
   - This is more performant and less error prone
 - Make a few types stricter/explicit
 - In `sendDataToAll`, send the data directly instead of using `sendDataTo`
   - This avoid making a redundant `get` for every connection on the map containing the connections.

Additionally, not really P2P but I happened to do it while looking into issues with this PR:

 - Fix non-autogenerated sourcemaps not being exported properly
   - It was not listed as files to copy in the GDJS build script
 - Remove old references to cocos from tsconfig excludes


Extensions/P2P/B_p2ptools.ts
```diff
@@ -471,10 +481,17 @@ namespace gdjs {
        */
       export const getConnectedPeer = (): string => connectedPeers[0] || '';
 
+      /**
+       * A JavaScript-only function to get the raw P2P DataConnection.
+       * This can be useful for example when you want to use a binary protocol
+       * instead if GDevelop variables for high-performance networking.
```
4ian, line 487:
```suggestion
       * instead of GDevelop variables for high-performance networking.
```

4ian (approved):
Nice!

PR #4147, opened 2022-07-26
Define types on the Physics2 extension

D8H:


Extensions/Physics2Behavior/physics2runtimebehavior.ts
```diff
@@ -458,6 +483,7 @@ namespace gdjs {
         }
 
         // Delete the body
+        this._body.gdjsAssociatedBehavior = null;
         this._sharedData.world.DestroyBody(this._body);
         this._body = null;
```
D8H, line 488: This is the only actual change.

D8H: I reverted it because it makes a test fail but it's probably the test that should be fixed.

Extensions/Physics2Behavior/physics2runtimebehavior.ts
```diff
@@ -3857,9 +4093,9 @@ namespace gdjs {
       joint.GetBodyB().SetAwake(true);
     }
 
-    getMotorJointCorrectionFactor(jointId) {
+    getMotorJointCorrectionFactor(jointId: integer | string): float {
       // Get the joint
-      const joint = this._sharedData.getJoint(jointId);
+      const joint = this._sharedData.getJoint(jointId) as Box2D.b2MotorJoint;
```
4ian, line 4098: In theory we're not sure yet it's a MotorJoin, but I guess this is not harmful

Extensions/Physics2Behavior/physics2runtimebehavior.ts
```diff
@@ -1360,217 +1409,283 @@ namespace gdjs {
...
       );
-      this._body.ApplyLinearImpulse(
+      body.ApplyLinearImpulse(
         this.b2Vec2(length * Math.cos(angle), length * Math.sin(angle)),
         this.b2Vec2Sec(
           positionX * this._sharedData.invScaleX,
           positionY * this._sharedData.invScaleY
-        )
+        ),
+        // TODO Should let Box2d awake the object itself.
```
4ian, line 1598: Not sure what you mean by this exactly?

Extensions/Physics2Behavior/physics2runtimebehavior.ts
```diff
@@ -1,31 +1,42 @@
...
 
-    // Contact listener to keep track of current collisions
-    contactListener: any;
+    /** Contact listener to keep track of current collisions */
+    contactListener: Box2D.JSContactListener;
     _nextJointId: number = 1;
 
-    // Start with 1 so the user is safe from default variables value (0)
-    joints: any = {};
+    // TODO Define the type: { [key: string]: Box2D.b2Joint }
```
4ian, line 28: Let's do this TODO or just remove it.

4ian (approved):
Not ideal that we have a lot of `body!` but this is the way the extension was written in JS, better to have typings even with a lot of `!` and `as` than nothing!

Extensions/Physics2Behavior/physics2runtimebehavior.ts
D8H: It's because the boolean can be set to true to wake up the object if necessary where we're doing it right before.

Extensions/Physics2Behavior/physics2runtimebehavior.ts
D8H: It's done. I'll remove it.

4ian (approved):

PR #4148, opened 2022-07-26
Fix Safari/iOS unable to load some images in the app

4ian:
This needs testing to check that the web-app works properly for these cases.

Thumbnails from examples work properly:
- [ ] on Safari iOS
- [ ] on Safari macOS
- [ ] on Chrome/Firefox as before

Assets are displayed on the asset store:
- [ ] on Safari iOS
- [ ] on Safari macOS
- [ ] on Chrome/Firefox as before

Assets are displayed in the scene editor:
- [ ] on Safari iOS
- [ ] on Safari macOS
- [ ] on Chrome/Firefox as before

Previews run properly:
- [ ] on Safari iOS
- [ ] on Safari macOS
- [ ] on Chrome/Firefox as before

Export run properly:
- [ ] on Safari iOS
- [ ] on Safari macOS
- [ ] on Chrome/Firefox as before

### DIFF
--- a/newIDE/app/src/UI/CorsAwareImage.js
+++ b/newIDE/app/src/UI/CorsAwareImage.js
@@ -38,7 +38,6 @@ const addSearchParameterToUrl = (
 export const CorsAwareImage = (props: Props) => (
   <img // eslint-disable-line jsx-a11y/alt-text
     {...props}
-    crossOrigin="anonymous"
     src={
       // To avoid strange/hard to understand CORS issues, we add a dummy parameter.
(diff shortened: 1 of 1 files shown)


4ian:
Testing version is being deployed on https://editor-beta.gdevelop.io.

ClementPasteau:
@4ian with the work you're currently doing on resources, is this PR still relevant?

4ian:
I think we can merge this now as it should not make things worse/could actually improve stuff on Safari. And this change will be necessary for supporting resources uploaded on GDevelop Cloud, so let's have it now!

PR #4151, opened 2022-07-28
Rework play section to display Liluo games

ClementPasteau:
https://user-images.githubusercontent.com/4895034/181479886-85056cb3-ca86-46b1-afe3-b496856c2bf3.mov


ClementPasteau:
One last issue I noticed, when the window is resized, the height is not handled correctly.
I've tried posting the new height, from Liluo, when the window resizes, but for some reason:
- it posts the correct new height when I run it on Liluo
- But when embedded in an iframe, it always posts the same height, not the updated one

I don't think it's a big deal, because the only "bug" it creates is that, after resizing the page, there may be some empty space at the end of the page (iframe height does not decrease)

4ian (approved):

AlexandreSi (approved):

PR #4152, opened 2022-07-28
Fix Asset cards text color to be visible on light theme

ClementPasteau:
Screenshot 2022-07-28 at 12 27 20
Screenshot 2022-07-28 at 12 27 32

### DIFF
--- a/newIDE/app/src/UI/Text.js
+++ b/newIDE/app/src/UI/Text.js
@@ -11,7 +11,7 @@ type textSize =
   | 'body'
   | 'body2';
 
-type textColor = 'error' | 'primary' | 'secondary';
+type textColor = 'error' | 'primary' | 'secondary' | 'inherit';
 
 type Props = {|
   /** The text to display. */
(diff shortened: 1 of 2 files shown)


newIDE/app/src/UI/Text.js
```diff
@@ -11,7 +11,7 @@ type textSize =
   | 'body'
   | 'body2';
 
-type textColor = 'error' | 'primary' | 'secondary';
+type textColor = 'error' | 'primary' | 'secondary' | 'inherit';
```
ClementPasteau, line 14: This is a new case where we want the color to be the same on all themes.
Our Text component does not allow this for the moment.
I think setting the color on the container, and using inherit is an OK practice, rather than allowing the `<Text/>` to have a color that can be overriden

4ian: I think in some exceptional use cases where it's the container that defines the color, it's ok to say "this text has no color, it's inheriting it from the parent".

4ian (approved):

Issue #4144, opened 2022-07-25
Physics 2 Behavior full crash in 5.0.139 when switching scenes with lots of objects with physics 2 behavior


krunkster:
## Describe the bug
There appears to be a bug somewhere in the Physics 2 Behavior that results in a full simulation crash with unending loop of box2d uncaught abort errors being thrown in the console log.
[phys2crash.txt](https://github.com/4ian/GDevelop/files/9184163/phys2crash.txt)
The two errors in that log repeat non-stop.

## To Reproduce
[issue-4144-min.zip](https://github.com/4ian/GDevelop/files/9184854/issue-4144-min.zip)
This issue appears to be a combination of:
- Having a lot of physics objects loaded from an external layout
- Loading and unloading a scene with lots of physics objects
If you load my example above and click the reload button a few times it will all come down.

Notes:
If you look at my external layout you will see a big boxed in section of objects.
That was the original size I tried to repro with.... it was not enough.
I added a lot more objects outside of that box in order to reproduce.
The "BlipScene" is not necessary to reproduce the issue, but it makes it happen faster (you can change it to just reload the main scene instead)
Wild guess: destroying an existing scene with lots of physics objects on it makes box2d angry... maybe reset the state of box2d if scene is changing and paused scenes are being destroyed?

## Other details
* Windows
* 5.0.139 Desktop

tristanbob:
Do you happen to adjust the size of the physics objects?  When I was tweening object scale it caused occasional crashes in Box2d.  For me, the solution was to disable physics Behavior while tweening.

krunkster:
It does seem to happen near a position tween of a physics object, but I believe the physics behavior is disabled while I'm tweening already... I'll double check though.

4ian:
If tweening is involved, maybe we reach something like a size of 0 and get a bug because of this.

krunkster:
OK, managed to isolate the problem pretty well (If I do say so myself)
I updated the main description with repro zip.

krunkster:
Ugh... I just had another box2d crash not related to loading scenes....  I'll capture a separate ticket later if I can repro again... it was after playing my game for like 20 minutes.  I honestly feel like the Physics behavior hasn't been stable since 5.0.130 :(

krunkster:
Is there an un-baked version of the version of box2d.js that you are using? I'm trying to debug and it's kinda tough.
```
  gdjs.registerRuntimeSceneUnloadedCallback(function (runtimeScene) {
[...]

PR #4153, opened 2022-07-28
Replace the Box2D library with a WebAssembly one.

D8H:
Fixes https://github.com/4ian/GDevelop/issues/4144

* It fixes a crash of the old Box2D library when a lot of objects with the Physics2 behavior were created.
* The same version of Box2d (2.3.1) is still used so there won't be any behavior change.


GDevelop.js/types/gdbehaviormetadata.js
```diff
@@ -25,6 +25,7 @@ declare class gdBehaviorMetadata {
   getAllStrExpressions(): gdMapStringExpressionMetadata;
   setIncludeFile(includeFile: string): gdBehaviorMetadata;
   addIncludeFile(includeFile: string): gdBehaviorMetadata;
+  addRequiredFile(resourceFile: string): gdBehaviorMetadata;
```
D8H, line 28: I only added a "add" method because I think that there is no need to rewrite the list.
Why does the "set" exist?

GDJS/GDJS/IDE/ExporterHelper.cpp
```diff
@@ -808,6 +811,46 @@ void ExporterHelper::ExportObjectAndBehaviorsIncludes(
...
+      std::cout << "include as resource: " << requiredFile
+                << std::endl;
+      InsertUnique(requiredFiles, requiredFile);
+    }
+  };
+
+  auto addObjectRequiredFiles = [&](const gd::Object &object) {
+    // Ensure needed files are included for the object type and its behaviors.
+
+    // TODO Do the same for objects once we have a case to test it.
```
D8H, line 828: This is to avoid to add dead code

Core/GDCore/Extensions/Metadata/BehaviorMetadata.cpp
```diff
@@ -402,6 +402,16 @@ BehaviorMetadata& BehaviorMetadata::AddIncludeFile(
   return *this;
 }
 
+BehaviorMetadata& BehaviorMetadata::AddRequiredFile(
+    const gd::String& requiredFile) {
+#if defined(GD_IDE_ONLY)
```
AlexandreSi, line 407: I think you can remove those lines

Extensions/Physics2Behavior/tests/physics2runtimebehavior.spec.js
```diff
@@ -1,4 +1,16 @@
 describe('Physics2RuntimeBehavior', () => {
+  const delay = (ms) => new Promise((res) => setTimeout(res, ms));
+
+  beforeEach(async function () {
+    // TODO find a clean way to wait for the Box2D library to load.
+    for (let index = 0; index < 200 && !window.Box2D; index++) {
+      await delay(5);
+    }
+    if (!window.Box2D) {
+      throw new Error('Timeout reading the tile map JSON file.');
```
AlexandreSi, line 10: Is the error message really adapted to the error?

GDJS/GDJS/IDE/ExporterHelper.cpp
```diff
@@ -808,6 +811,46 @@ void ExporterHelper::ExportObjectAndBehaviorsIncludes(
   }
 }
 
+void ExporterHelper::ExportObjectAndBehaviorsRequiredFiles(
+    const gd::Project &project, std::vector<gd::String> &requiredFiles) {
+  auto addRequiredFiles = [&](const std::vector<gd::String> &newRequiredFiles) {
+    for (const auto &requiredFile : newRequiredFiles) {
+      
+      std::cout << "include as resource: " << requiredFile
```
AlexandreSi, line 819: Is this `cout` necessary?

GDJS/GDJS/IDE/ExporterHelper.cpp
AlexandreSi: I don't think I understand the comment. Since the name of the method is `addObject...` I would tend to believe that we are doing it for the objects

GDJS/Runtime/gd.ts
```diff
@@ -558,8 +558,28 @@ namespace gdjs {
...
   };
+
+  const asynchronouslyLibraryPromises: Array<Promise<any>> = [];
+
+  /**
+   * Register a promise that a library will be loaded.
+   *
+   * This method must be called by any library that loads asynchronously.
+   */
+  export const registerAsynchronouslyLibraryPromise = (
```
AlexandreSi, line 569: Should this be named `registerAsynchronouslyLoadedLibraryPromise` or `registerPromiseForLibraryLoading`. I had trouble reading the english of the variable at first

GDJS/GDJS/IDE/ExporterHelper.cpp
4ian: Let's go for:

`TODO: Handle required files declared by objects. For now, no objects has a need for additional required files, so the object metadata do not even have `requiredFiles`.`

GDJS/GDJS/IDE/ExporterHelper.cpp
```diff
@@ -808,6 +811,45 @@ void ExporterHelper::ExportObjectAndBehaviorsIncludes(
...
+        for (std::size_t i = 0; i < objectsContainer.GetObjectsCount(); ++i) {
+          addObjectRequiredFiles(objectsContainer.GetObject(i));
+        }
+      };
+
+  addObjectsRequiredFiles(project);
+  for (std::size_t i = 0; i < project.GetLayoutsCount(); ++i) {
+    const gd::Layout &layout = project.GetLayout(i);
+    addObjectsRequiredFiles(layout);
+  }
```
4ian, line 850: Not "super happy" to have this "logic" traversing objects duplicated, I wonder if it's worth making it into a function/template, or not?

In the future, we might have objects in other places (like... inside objects? :))

4ian: Might be too much forward thinking and not useful though.

GDJS/Runtime/gd.ts
```diff
@@ -558,8 +558,30 @@ namespace gdjs {
       return diff / Math.min(absA + absB, Number.MAX_VALUE) < epsilon;
     }
   };
+
+  const asynchronouslyLoadingLibraryPromises: Array<Promise<any>> = [];
+
+  /**
+   * Register a promise that a library will be loaded.
```
4ian, line 565:
```suggestion
   * Register a promise which will be resolved when a third party library has finished loading (and is required to load before launching the game).
```

GDJS/Runtime/gd.ts
```diff
@@ -558,8 +558,30 @@ namespace gdjs {
...
+   * This method must be called by any library that loads asynchronously.
+   */
+  export const registerAsynchronouslyLoadingLibraryPromise = (
+    promise: Promise<any>
+  ): void => {
+    asynchronouslyLoadingLibraryPromises.push(promise);
+  };
+
+  /**
+   * @returns a promise that all libraries will be loaded.
```
4ian, line 576:
```suggestion
   * @returns a promise resolved when all all third party libraries, which need to be loaded before the game startup, are loaded. If a library fails loading, this will be rejected.
```

GDJS/tests/karma.conf.js
```diff
@@ -23,106 +23,120 @@ module.exports = function (config) {
         timeout: 10000, // Give a bit more time for CIs (the default 2s can be too low sometimes, as a real browser is involved).
       },
     },
+    basePath: '../..',
+    proxies: {
+      '/base/tests-utils/':
+      '/base/GDJS/tests/tests-utils/',
+      '/base/newIDE/app/resources/GDJS/Runtime/Extensions/Physics2Behavior/Extensions/Physics2Behavior/Box2D_v2.3.1_min.wasm.wasm':
```
4ian, line 30: You might want to comment why this line is needed.

GDJS/tests/karma.conf.js
```diff
@@ -23,106 +23,120 @@ module.exports = function (config) {
         timeout: 10000, // Give a bit more time for CIs (the default 2s can be too low sometimes, as a real browser is involved).
       },
     },
+    basePath: '../..',
```
4ian, line 26: Very nice!

GDevelop.js/types/gdbehaviormetadata.js
4ian: I think we should just have been using addIncludeFile, wrong past decision.

GDJS/GDJS/IDE/ExporterHelper.cpp
```diff
@@ -808,6 +811,45 @@ void ExporterHelper::ExportObjectAndBehaviorsIncludes(
...
+      InsertUnique(requiredFiles, requiredFile);
+    }
+  };
+
+  auto addObjectRequiredFiles = [&](const gd::Object &object) {
+    // Ensure needed files are included for the object type and its behaviors.
+
+    // TODO Do the same for objects once we have a case to test it.
+    // No object use WebAssembly library so the requiredFiles attribute
+    // is not implemented yet.
```
D8H, line 827: Should I keep it or change it?

4ian (approved):
🤩🤩 Excellent! 🤩🤩

I wonder if we'll see behavior changes because of the wasm doing maybe not exactly the same arithmetic operations as the asm.js version. 

But anyway we can't stay with the other version crashing when using too much memory :)

GDJS/GDJS/IDE/ExporterHelper.cpp
4ian: Replace by this comment:

```
TODO: Handle required files declared by objects. For now, no objects has a need for additional required files, so the object metadata do not even have `requiredFiles`.
```

GDJS/GDJS/IDE/ExporterHelper.cpp
D8H: I like the Iterator pattern 😃

D8H: But, yes I think it can wait.

4ian: Totally.

PR #4166, opened 2022-08-01
Add support for paying subscriptions with PayPal

4ian:
Took the opportunity to fix the design of the subscription dialog:
image

The payment is made in a separate web browser. As now, but it gives you the choice to pay with Stripe (credit card/Google Pay/Apple Pay) or PayPal:
image

Subscriptions are handled the same, except that with PayPal you can stop them from your PayPal account. You can't upgrade to another subscription with a pro-rated amount though (while Stripe handles this properly) - GDevelop will let you know that your existing subscription will be cancelled and the new one started instead.


newIDE/app/src/Profile/SubscriptionDialog.js
```diff
@@ -33,319 +29,333 @@ import {
...
-
-export default class SubscriptionDialog extends React.Component<Props, State> {
-  state = { isLoading: false, subscriptionPendingDialogOpen: false };
-
-  componentDidMount() {
-    if (this.props.open) {
-      sendSubscriptionDialogShown();
-    }
-  }
+export default function SubscriptionDialog({ open, onClose }: Props) {
```
ClementPasteau, line 69: FC 🙌

newIDE/app/src/Profile/SubscriptionDialog.js
```diff
@@ -33,319 +29,333 @@ import {
...
-      }).then(
-        () =>
-          this.handleUpdatedSubscriptionSuccess(i18n, authenticatedUser, plan),
-        (err: Error) => this.handleUpdatedSubscriptionFailure(i18n, err)
-      );
-    } else {
-      this.setState({
-        subscriptionPendingDialogOpen: true,
-      });
+    // Subscribing from an account with a subscription.
```
ClementPasteau, line 95: you mean with no subscription?

newIDE/app/src/Profile/SubscriptionDialog.js
```diff
@@ -33,319 +29,333 @@ import {
...
+      // Cancelling the existing subscription.
+      const answer = await showConfirmation(cancelConfirmationTexts);
+      if (!answer) return;
+
+      setIsChangingSubscription(true);
+      try {
+        await changeUserSubscription(getAuthorizationHeader, profile.id, {
+          planId: null,
+        });
+        await authenticatedUser.onRefreshUserProfile();
```
ClementPasteau, line 118: I'm wondering if the onRefreshUserProfile() should be embedded in the `changeUserSubscription()` function, to avoid having to call it every time.
Not a big deal though

newIDE/app/src/Profile/SubscriptionDialog.js
```diff
@@ -33,319 +29,333 @@ import {
...
+        setSubscriptionPendingDialogOpen(true);
+        Window.openExternalURL(
+          getRedirectToCheckoutUrl(planId, profile.id, profile.email)
+        );
+      }
+    }
   };
 
-  _renderPrice(plan: PlanDetails): React.Node {
+  const renderPrice = React.useCallback((plan: PlanDetails) => {
```
ClementPasteau, line 199: hmm, a callback with no parameters, this can probably just be a function outside the React.FC

newIDE/app/src/Profile/SubscriptionDialog.js
```diff
@@ -33,319 +29,333 @@ import {
...
+          <ColumnStackLayout noMargin expand>
+            <Line noMargin alignItems="center">
+              <img src="res/diamond.svg" style={styles.diamondIcon} alt="" />
+              <Text>
+                <Trans>
+                  Get a subscription to gain more one-click exports, remove the
+                  GDevelop splashscreen and messages asking you to get a
+                  subscription. With a subscription, you're also supporting the
+                  development of GDevelop, which is an open-source software
+                  maintained by volunteers in their free time.
```
ClementPasteau, line 237: is this message still valid? I'd maybe remove the last part after "open-source software"

newIDE/app/src/Profile/SubscriptionDialog.js
```diff
@@ -33,319 +29,333 @@ import {
...
+                </Trans>
+              </Text>
+            </Line>
+            {getSubscriptionPlans().map(plan => {
+              const isCurrentPlan =
+                !!authenticatedUser.subscription &&
+                authenticatedUser.subscription.planId === plan.planId;
+              return (
+                <Card key={plan.planId || ''} isHighlighted={isCurrentPlan}>
```
ClementPasteau, line 246: nit: if the planId can be null, then maybe use an index for the map key, rather than this, with the risk of having 2 empty strings

newIDE/app/src/Profile/SubscriptionDialog.js
4ian: Done

newIDE/app/src/Profile/SubscriptionDialog.js
4ian: Indeed, update

newIDE/app/src/Profile/SubscriptionDialog.js
4ian: Fixed

newIDE/app/src/Profile/SubscriptionDialog.js
4ian: It's an API call, so a bit hard to do it inside. I'll leave it as is for now!

newIDE/app/src/Profile/SubscriptionDialog.js
4ian: The only empty planId that can be null is the "no subscription". If somehow two plans are null, it means twice the "no subscription" plan... which should never happen (much like you should not have two subscriptions with the same id!)

PR #4168, opened 2022-08-02
Export a basic service worker with games

arthuro555:
With this, this makes all HTML5 builds hosted on https a working PWA 😎 

This adds a basic service worker that prefetches and caches all code and asset files. When the project is reexported, the generated code files and assets cache are invalidated. GDJS Runtime's code cache only expire when GDevelop's version changes.

Issues:
- [ ] GDevelop version checking is using VersionWrapper but this doesn't contains the version of GDJS, only of the project file format
- [ ] index.html fails with forbidden when trying to cache it only (on liluo)

I tested it on preview, a local server, liluo, and netlify. Only liluo had an issue and only with index.html for a rreason I couldn't determine.


Core/GDCore/IDE/Project/ProjectResourcesCopier.h
```diff
@@ -39,14 +39,16 @@ class GD_CORE_API ProjectResourcesCopier {
    * of the resources will be preserved when copying. Otherwise, everything will
    * be send in the destinationDirectory.
    *
-   * \return true if no error happened
+   * \return The list of the exported names of all resources, useful for
+   * prefetching and caching.
    */
-  static bool CopyAllResourcesTo(gd::Project& project,
```
arthuro555, line 44: I checked and calls to this function was using the boolean return value (which always returned true either way)

GDJS/GDJS/IDE/ExporterHelper.cpp
```diff
@@ -852,6 +853,168 @@ const std::array<int, 20> IOS_ICONS_SIZES = {
...
+    name: 'gdjs-generated-code-vGENERATED_CODE_CACHE_VERSION',
+    files: /*GENERATED_CODE_FILES*/,
+  },
+  {
+    name: 'gdjs-static-code-vSTATIC_CODE_CACHE_VERSION',
+    files: /*STATIC_CODE_FILES*/,
+  },
+  {
+    name: 'gdjs-resources-vRESOURCES_CACHE_VERSION',
+    files: /*RESOURCE_FILES*/,
```
arthuro555, line 905: It would be cool if sometime in the future we can do it on a file-pre-file basis and only invalidate the cache of those that have actually changed, but let's keep it simple for now.

PR #4170, opened 2022-08-02
Display staging assets

ClementPasteau:
For testing purpose, on dev only.
This allows testing assets before releasing them.

https://user-images.githubusercontent.com/4895034/182406679-8287f58a-2d75-4362-810b-7230d2287da8.mov

### DIFF
--- a/newIDE/app/src/AssetStore/AssetDetails.js
+++ b/newIDE/app/src/AssetStore/AssetDetails.js
@@ -100,7 +100,14 @@ export const AssetDetails = ({
   onOpenDetails,
 }: Props) => {
   const gdevelopTheme = React.useContext(ThemeContext);
-  const { authors, licenses } = React.useContext(AssetStoreContext);
+  const {
+    authors,
+    licenses,
+    environment,
+    error: filterError,
+    fetchAssetsAndFilters,
+    useSearchItem,
+  } = React.useContext(AssetStoreContext);
   const [asset, setAsset] = React.useState<?Asset>(null);
   const [
     selectedAnimationName,
@@ -113,7 +120,9 @@ export const AssetDetails = ({
         try {
           // Reinitialise asset to trigger a loader and recalculate all parameters. (for instance zoom)
           setAsset(null);
-          const loadedAsset = await getAsset(assetShortHeader);
+          const loadedAsset = await getAsset(assetShortHeader, {
+            environment,
+          });
           setAsset(loadedAsset);
           if (loadedAsset.objectType === 'sprite') {
             // Only sprites have animations and we select the first one.
@@ -127,7 +136,7 @@ export const AssetDetails = ({
         }
       })();
     },
-    [assetShortHeader]
+    [assetShortHeader, environment]
   );
 
   const isImageResourceSmooth = React.useMemo(
@@ -176,11 +185,6 @@ export const AssetDetails = ({
     () => [new SimilarAssetStoreSearchFilter(assetShortHeader)],
     [assetShortHeader]
   );
-  const {
-    error: filterError,
-    fetchAssetsAndFilters,
-    useSearchItem,
-  } = React.useContext(AssetStoreContext);
   const searchResults = useSearchItem('', null, null, similarAssetFilters);
   const truncatedSearchResults = searchResults && searchResults.slice(0, 60);
 
@@ -269,8 +273,7 @@ export const AssetDetails = ({
                       fixedWidth={FIXED_WIDTH}
                     />
                   )}
-                {(asset.objectType === 'tiled' ||
-                  asset.objectType === '9patch') && (
+                {asset.objectType !== 'sprite' && (
                   <div style={styles.previewBackground}>
                     <CorsAwareImage
                       style={styles.previewImage}
--- a/newIDE/app/src/AssetStore/AssetStoreContext.js
+++ b/newIDE/app/src/AssetStore/AssetStoreContext.js
@@ -243,7 +243,7 @@ export const AssetStoreStateProvider = ({
     [isLoading, environment]
   );
 
-  // Preload the assets and filters when the
+  // Preload the assets and filters when the app loads.
   React.useEffect(
     () => {
       // Don't attempt to load again assets and filters if they
--- a/newIDE/app/src/AssetStore/AssetStoreContext.js
+++ b/newIDE/app/src/AssetStore/AssetStoreContext.js
@@ -6,6 +6,7 @@ import {
   type AssetPacks,
   type Author,
   type License,
+  type Environment,
   listAllAssets,
   listAllAuthors,
   listAllLicenses,
@@ -49,6 +50,8 @@ type AssetStoreState = {|
   assetPacks: ?AssetPacks,
   authors: ?Array<Author>,
   licenses: ?Array<License>,
+  environment: Environment,
+  setEnvironment: Environment => void,
   searchResults: ?Array<AssetShortHeader>,
   fetchAssetsAndFilters: () => void,
   error: ?Error,
@@ -70,6 +73,8 @@ export const AssetStoreContext = React.createContext<AssetStoreState>({
   assetPacks: null,
   authors: null,
   licenses: null,
+  environment: 'live',
+  setEnvironment: () => {},
   searchResults: null,
   fetchAssetsAndFilters: () => {},
   error: null,
@@ -130,6 +135,7 @@ export const AssetStoreStateProvider = ({
   const [assetPacks, setAssetPacks] = React.useState<?AssetPacks>(null);
   const [authors, setAuthors] = React.useState<?Array<Author>>(null);
   const [licenses, setLicenses] = React.useState<?Array<License>>(null);
+  const [environment, setEnvironment] = React.useState<Environment>('live');
   const [error, setError] = React.useState<?Error>(null);
   const isLoading = React.useRef<boolean>(false);
 
@@ -195,9 +201,7 @@ export const AssetStoreStateProvider = ({
 
   const fetchAssetsAndFilters = React.useCallback(
     () => {
-      // Don't attempt to load again assets and filters if they
-      // were loaded already.
-      if (assetShortHeadersById || isLoading.current) return;
+      if (isLoading.current) return;
 
       (async () => {
         setError(null);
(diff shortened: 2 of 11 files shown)


newIDE/app/src/AssetStore/AssetStoreContext.js
```diff
@@ -50,7 +50,7 @@ type AssetStoreState = {|
   authors: ?Array<Author>,
   licenses: ?Array<License>,
   searchResults: ?Array<AssetShortHeader>,
-  fetchAssetsAndFilters: () => void,
+  fetchAssetsAndFilters: (showStagingAssets: boolean) => void,
```
4ian, line 53: Would that make sense to pass this as an object? I'm always worried of one boolean, hard to read.

newIDE/app/src/Utils/GDevelopServices/Asset.js
```diff
@@ -143,8 +159,16 @@ export const listAllAssets = (): Promise<AllAssets> => {
 };
 
 export const getAsset = (
-  assetShortHeader: AssetShortHeader
+  assetShortHeader: AssetShortHeader,
+  { isStagingAsset }: { isStagingAsset: boolean }
```
4ian, line 163: {||}

newIDE/app/src/UI/Search/UseSearchItem.js
```diff
@@ -237,8 +237,8 @@ export const useSearchItem = <SearchItem: { tags: Array<string> }>(
         const allIds = Object.keys(searchItemsById);
 
         allIds.forEach(id => {
-          const assetShortHeader = searchItemsById[id];
-          newSearchApi.indexDocument(id, getItemDescription(assetShortHeader));
+          const searchItem = searchItemsById[id];
```
4ian, line 240: Nice!

newIDE/app/src/Utils/GDevelopServices/Asset.js
```diff
@@ -143,8 +159,16 @@ export const listAllAssets = (): Promise<AllAssets> => {
...
 ): Promise<Asset> => {
+  if (isStagingAsset) {
+    return axios
+      .get(
+        `${stagingBucket}/assets-database/assets/${assetShortHeader.id}.json`
```
4ian, line 168: I'm not sure to understand why this is not just a ternary on the url?

newIDE/app/src/AssetStore/AssetStoreContext.js
4ian: Better, could we keep this option in the context?
It's like searchText, it's an option of the search:
- what do you search (`searchText`)
- with which filters (`assetFiltersState`)
- where do you search (`environment: "staging" | null`)

newIDE/app/src/AssetStore/AssetDetails.js
```diff
@@ -404,7 +407,7 @@ export const AssetDetails = ({
             <Line expand noMargin justifyContent="center">
               <BoxSearchResults
                 baseSize={128}
-                onRetry={fetchAssetsAndFilters}
+                onRetry={() => fetchAssetsAndFilters(showStagingAssets)}
```
4ian, line 410: See my other comment, but I have a bad feeling if we start to have two places where the search options are (the context and here).
It means you have to ensure you don't call the method with the wrong boolean.

newIDE/app/src/AssetStore/AssetStoreContext.js
```diff
@@ -236,7 +236,7 @@ export const AssetStoreStateProvider = ({
         isLoading.current = false;
       })();
     },
-    [assetShortHeadersById, isLoading]
+    [isLoading]
```
4ian, line 239: This was not necessary?

newIDE/app/src/AssetStore/AssetDetails.js
```diff
@@ -269,8 +273,7 @@ export const AssetDetails = ({
                       fixedWidth={FIXED_WIDTH}
                     />
                   )}
-                {(asset.objectType === 'tiled' ||
-                  asset.objectType === '9patch') && (
+                {asset.objectType !== 'sprite' && (
```
D8H, line 276: What will happen to users of the 140 if we introduce a new kind of object in the future?

newIDE/app/src/AssetStore/AssetStoreContext.js
```diff
@@ -247,7 +247,7 @@ export const AssetStoreStateProvider = ({
 
       const timeoutId = setTimeout(() => {
         console.info('Pre-fetching assets from asset store...');
-        fetchAssetsAndFilters();
+        fetchAssetsAndFilters(false);
```
4ian, line 250: Here you have an example of the "bad implicitness": you pass false, so it means that other components must consider that by default the environment was the production one (which is fair, but in the future you could want to load staging assets first, and this should need only a change in a single place.)

newIDE/app/src/AssetStore/AssetStoreContext.js
ClementPasteau: yup, I put that in the NavigationState, but it makes sense to rather exist in the context

newIDE/app/src/AssetStore/AssetDetails.js
ClementPasteau: Good question.
I believe that an asset always have an array of previewImageUrls defined.
And the CorsAwareImage handles the case where we pass null.
So if we introduce a new kind of object, it should display the preview image correctly (until we push an update where we display that differently)

I think it's fair to say:
- all assets have previews (shown on Asset Search list), and that's the default display, unless specified otherwise

newIDE/app/src/AssetStore/AssetStoreContext.js
ClementPasteau: Instead of:
- calling this function every time the NewObjectDialog opens (with an effect)
- handling in the context if there are already elements, to decide if the call happens

I think it's cleaner to:
- call this function by the context on first load
- call manually when needed (retry / switch environment)

I didn't notice any issue, but now that I think about it, there may be a slight delay if you open the asset store and the context hasn't started fetching the assets yet (quite unique case)
I'll do more testing

newIDE/app/src/Utils/GDevelopServices/Asset.js
ClementPasteau: This is where it's a bit hacky.
Our current logic asks GDevelop backend for this URL, then makes the call.
But because the backend always returns the prod URL (no matter the environment) I went for calling the json url directly.
I'm not sure what's best:
- modifying the backend to return the "staging" bucket urls when in dev. And then we need to modify the IDE to call prod URLs by default, unless specified otherwise
- keep it like this
- change the logic to avoid making 2 calls whatever the environment, and always fetch the jsons directly

arthuro555: An alternative option would be to add to the asset API a get parameter to specify the environment you want to get the bucket URL for. That way, no need to do a hack where you call the prod API to get the live buckets, since the development API could return both live and staging assets.

ClementPasteau: I like this approach @arthuro555 👍 
I'll go for this I think

4ian: Yep I think it's fair and allow for more flexibility for testing/development.

newIDE/app/src/AssetStore/AssetStoreContext.js
```diff
@@ -236,9 +240,10 @@ export const AssetStoreStateProvider = ({
         isLoading.current = false;
       })();
     },
-    [assetShortHeadersById, isLoading]
```
4ian, line 239: Can you remind me why assetShortHeadersById is not useful?

newIDE/app/src/AssetStore/AssetStoreContext.js
```diff
@@ -236,9 +240,10 @@ export const AssetStoreStateProvider = ({
...
+    [isLoading, environment]
   );
 
+  // Preload the assets and filters when the
```
4ian, line 246: The suspense is huge!

newIDE/app/src/AssetStore/NewObjectDialog.js
```diff
@@ -198,9 +204,21 @@ export default function NewObjectDialog({
...
       openedAssetShortHeader,
+      environment,
     ]
   );
 
+  // Load assets and filters when the dialog is opened.
+  React.useEffect(
+    () => {
+      // This check is to ensure we call fetchAssetsAndFilters every time environment changes.
+      if (environment) {
```
4ian, line 215: environment can't be null right?

newIDE/app/src/AssetStore/NewObjectDialog.js
```diff
@@ -34,6 +34,7 @@ import { installAsset } from './InstallAsset';
 import EventsFunctionsExtensionsContext from '../EventsFunctionsExtensionsLoader/EventsFunctionsExtensionsContext';
 import { useResourceFetcher } from '../ProjectsStorage/ResourceFetcher';
 import { showErrorBox } from '../UI/Messages/MessageBox';
+const isDev = process.env.NODE_ENV === 'development';
```
4ian, line 37: Use `Window.isDev()`

newIDE/app/src/AssetStore/AssetStoreSearchFilter.js
```diff
@@ -94,6 +94,10 @@ export class DimensionAssetStoreSearchFilter
   }
 
   getPertinence(searchItem: AssetShortHeader): number {
+    // Assets with no defined dimensions are always pertinent. (ex: particle emitter)
+    if (!searchItem.width && !searchItem.height) {
```
4ian, line 98: Should that be:

```suggestion
    if (!searchItem.width || !searchItem.height) {
```

for safety?

4ian (approved):
I think this approach using environment is much better. Cleaner, easier to switch, could be extended to another environment in the future if needed. This also avoids the strong distinction between live and staging (or "prod" and "dev"), meaning that we can stay as close as the prod environment as possible in development.

newIDE/app/src/AssetStore/NewObjectDialog.js
ClementPasteau: No, it's a bit of hack, but couldn't find another way to trigger a refetch when the value change 🤔 
You have to do it inside an effect, because if you do it just after the value is set in the state, it's not there yet (and no callback available)
I can improve the comment to indicate why it's a hack

newIDE/app/src/AssetStore/AssetStoreContext.js
ClementPasteau: There used to be a check to ensure the assets are loaded only once for the whole session.
But now that we change the environment, we want the callback to run again so we don't care if there were already results or not.

PR #4178, opened 2022-08-04
Fix memory leak at the end of a cloud project save

AlexandreSi:


newIDE/app/src/MainFrame/index.js
```diff
@@ -351,6 +364,11 @@ const MainFrame = (props: Props) => {
     setFileMetadataOpeningMessage,
   ] = React.useState<?MessageDescriptor>(null);
 
+  // This reference is useful to get the current opened project.
+  // It might be different from the current project an effect or a callback
+  // manipulates when a promise resolves for instance.
+  const currentProjectRef = React.useRef<?gdProject>(null);
```
4ian, line 370:
```suggestion
  /* This reference is useful to get the current opened project, even in the callback of a hook/promise - without risking to read "stale" data .
   * This can be different from the `currentProject` (coming from the state) that an effect or a callback
   * manipulates when a promise resolves for instance.
   * See `isCurrentProjectStale`.
   */
  const currentProjectRef = React.useRef<?gdProject>(null);
```

newIDE/app/src/MainFrame/index.js
```diff
@@ -180,6 +180,19 @@ const findStorageProviderFor = (
   return storageProvider;
 };
 
+/**
+ * Compares a React reference to the current project (truth source)
+ * and a project stored in a variable.
+ * It is useful to detect if the project stored in a variable is still
+ * currently opened.
+ */
+const isCurrentProjectUnchanged = (
```
4ian, line 189:
```suggestion
 * Compares a React reference to the current project (truth source)
 * and a project stored in a variable (coming probably from a React state).
 * It is useful to detect if the project stored in a variable is still
 * valid (still currently opened). If it's not, it means the variable is "stale".
 */
const isCurrentProjectStale = (
```

newIDE/app/src/MainFrame/index.js
```diff
@@ -1821,17 +1852,23 @@ const MainFrame = (props: Props) => {
...
-
-                setState(state => ({
-                  ...state,
-                  currentFileMetadata: fileMetadata,
-                }));
+                if (
+                  isCurrentProjectUnchanged(currentProjectRef, currentProject)
+                ) {
+                  // We do not want to change the current file metadata if the
+                  // project has changed at the moment the promise resolved
```
4ian, line 1866:
```suggestion
                  // project has changed before the promise resolved.
```

newIDE/app/src/MainFrame/index.js
```diff
@@ -1928,17 +1972,22 @@ const MainFrame = (props: Props) => {
...
-            fileMetadata: fileMetadata.name
-              ? fileMetadata
-              : { ...fileMetadata, name: currentProject.getName() },
-            storageProviderName: getStorageProvider().internalName,
+            fileMetadata: enrichedFileMetadata,
+            storageProviderName: storageProviderInternalName,
           });
+          if (isCurrentProjectUnchanged(currentProjectRef, currentProject)) {
+            // We do not want to change the current file metadata if the
+            // project has changed at the moment the promise resolved
```
4ian, line 1984:
```suggestion
            // project has changed before the promise resolved.
```

4ian (approved):

newIDE/app/src/MainFrame/index.js
4ian: (vocabulary used by: https://reactjs.org/docs/hooks-faq.html#why-am-i-seeing-stale-props-or-state-inside-my-function)

PR #4181, opened 2022-08-05
Fix collision mask of rotated tiles that were sometimes misplaced.

D8H:
The 8 cases are now unit tested.

* Tiled
image

* Instance editor
image

* Runtime
image


SharedLibs/TileMapHelper/src/render/TileMapPixiHelper.ts
```diff
@@ -266,17 +266,19 @@ export class PixiTileMapHelper {
               for (let index = 0; index < vertices.length; index++) {
                 let vertexX = vertices[index][0];
                 let vertexY = vertices[index][1];
+                // It's important to do the diagonal flipping first,
```
4ian, line 269: Nice comment!

4ian (approved):

PR #4183, opened 2022-08-05
Add more cloud projects for GDevelop online subscribers

4ian:
I've kept the existing design of the AlertMessage (this could be updated later to follow the new design everywhere in the app, but I stayed simple here and avoided too many changes)

image
image

### DIFF
--- a/newIDE/app/src/MainFrame/EditorContainers/HomePage/BuildSection/MaxProjectCountAlertMessage.js
+++ b/newIDE/app/src/MainFrame/EditorContainers/HomePage/BuildSection/MaxProjectCountAlertMessage.js
@@ -1,0 +1,45 @@
+// @flow
+import * as React from 'react';
+import { Trans } from '@lingui/macro';
+
+import Text from '../../../../UI/Text';
+import RaisedButton from '../../../../UI/RaisedButton';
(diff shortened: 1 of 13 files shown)


D8H (approved):
LGTM

newIDE/app/src/MainFrame/EditorContainers/HomePage/BuildSection/MaxProjectCountAlertMessage.js
```diff
@@ -0,0 +1,45 @@
...
+        <Text size="block-title">
+          <Trans>
+            You've reached your maximum storage of{' '}
+            {limits.capabilities.cloudProjects.maximumCount}
+            cloud-based projects
+          </Trans>
+        </Text>
+        <Text>
+          <Trans>
+            Update to GDevelop Premium to get more storage, one click
```
AlexandreSi, line 38: Should it be upgrade?

AlexandreSi: And we should maybe change the message depending if the user has no subscription (check out our premium plans), an indie (consider upgrading to pro to get more storage) or a pro subscription (damn, too many projects, delete some).

PR #4184, opened 2022-08-05
RuntimeScenePixiRenderer: Wait with showing the cursor until the frame is rendered

dos1:
This minimizes cursor flicker in games that switch between system cursor and
in-game cursor. If the browser updates cursor status asynchronously and the
rendering takes longer than screen refresh, we could easily end up with two
cursors displayed at once for a short moment.

While the same could be done when hiding the cursor as well, it may be not
a bad idea to do the hiding early anyway, as system cursor can be updated
faster than the browser is able to dispatch events, so hiding it early
minimizes the cases where the cursor jumps back to a previous position
after switching to in-game cursor.

### DIFF
--- a/GDJS/Runtime/pixi-renderers/runtimescene-pixi-renderer.ts
+++ b/GDJS/Runtime/pixi-renderers/runtimescene-pixi-renderer.ts
@@ -11,6 +11,7 @@ namespace gdjs {
     _debugDraw: PIXI.Graphics | null = null;
     _debugDrawContainer: PIXI.Container | null = null;
     _profilerText: PIXI.Text | null = null;
+    _showCursorAtNextRender: boolean = false;
     _debugDrawRenderedObjectsPoints: Record<
       number,
       {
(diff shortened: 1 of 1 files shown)


GDJS/Runtime/pixi-renderers/runtimescene-pixi-renderer.ts
```diff
@@ -11,6 +11,7 @@ namespace gdjs {
     _debugDraw: PIXI.Graphics | null = null;
     _debugDrawContainer: PIXI.Container | null = null;
     _profilerText: PIXI.Text | null = null;
+    _showCursor: boolean = false;
```
4ian, line 14: Should we call this

```suggestion
    _showCursorAtNextRender: boolean = false;
```

?

GDJS/Runtime/pixi-renderers/runtimescene-pixi-renderer.ts
```diff
@@ -59,6 +60,10 @@ namespace gdjs {
       // render the PIXI container of the scene
       this._pixiRenderer.backgroundColor = this._runtimeScene.getBackgroundColor();
       this._pixiRenderer.render(this._pixiContainer);
+      if (this._showCursor) {
```
4ian, line 63: Nitpicking: let's add a line break and then a comment explaining why this operation is deferred there :)

4ian (approved):
Sounds good, added two minor comments!

dos1:
Updated according to comments

4ian (approved):

4ian:
Thank you 👏

PR #4186, opened 2022-08-06
Fix raycast condition not working properly on objects with multiple hitboxes

### LINKED ISSUE #3563: Inverted Raycasting condition ignores additional Collision Masks on the same object
## Describe the bug
If a sprite object has multiple collision masks in the same object (to deal with shapes that would otherwise be convex), the inverted raycasting condition will only find points within the _newest created_ collision mask, instead of all of the collision masks of the object.

## Long Version/Why this matters
I am working on a point and click adventure game template/example for the engine. 
Currently, there is not an easy (or really, any) way to replicate proper point and click adventure game movement in the engine.
In classic and modern adventure games, there is a defined "valid movement area" and if you click anywhere within it, the character will pathfind to that point. This is doable in the engine. 

However, in those same games, if you click _outside_ the valid movement area, the character will pathfind to the closest moveable area that _is_ valid. Right now there is not a way to do this in the engine.

The closest way I and anyone on the discord have found to accomplish this _should_ be by defining the movable area as the collision mask of the background, then set up an inverted raycast to find the last point that is still valid for the background sprite object. This almost works, except in most cases players are going to need to set up numerous collision masks within the same object to accommodate non-convex areas that are valid for movement. As mentioned above, the raycast will only find valid points within one collision mask of an object, and only the newest one.

## To Reproduce
1. Create a sprite object. (We'll call this "Destination")
2. Add a custom collision mask of any shape.
3. Add an additional custom collision mask of any shape (overlapping or not).
4. Add another sprite object of any kind (we'll call this "Start").
5. Place both on the scene.
6. Add a text object. 
7. Add it to scene.
8. Add an event with an "Left mouse button is down" condition and inverted raycast that tests from the "Start" object against the MouseX/Y position, against the "Destination" object. Add an inverted "Point is inside object" using the MouseX/Y points and the "Destination" object.
9. Have the action update the text object to show the X/Y variables from the raycast.
10. Preview the scene and click outside the Destination object. No matter where you click, it will only show X/Y locations within the last created collision mask. It will treat the other collision masks as if they were part of another object.

* Here's an online export of an example set up really quickly: [Online Example](https://liluo.io/games/ead6eddb-6600-47f9-b08a-d62ab2fa9ea3)
* Here are the collision masks: 
image
* Here are the events: 
image
* Here's a copy of the project: 
[RayCastTest.zip](https://github.com/4ian/GDevelop/files/7964065/RayCastTest.zip)

## Other details
* Windows 10 Pro 64-bit
* Tested on both desktop and web demo versions of 5.0.126
* This may be related to https://github.com/4ian/GDevelop/issues/3498

4ian:
Fix #3563

TODO:
- Add a test case with an object having two hitboxes.

### DIFF
--- a/GDJS/Runtime/runtimeobject.ts
+++ b/GDJS/Runtime/runtimeobject.ts
@@ -48,9 +48,9 @@ namespace gdjs {
    * {@link RuntimeObject.raycastTest} to avoid any allocation.
    */
   const raycastTestStatics: {
-    result: RaycastTestResult
+    result: RaycastTestResult;
   } = {
-    result: gdjs.Polygon.makeNewRaycastTestResult()
+    result: gdjs.Polygon.makeNewRaycastTestResult(),
   };
 
   /**
@@ -2332,24 +2332,32 @@ namespace gdjs {
       }
 
       // Do a real check if necessary.
-      let testSqDist = closest ? raySqBoundingRadius : 0;
-      const hitBoxes = this.getHitBoxesAround(x, y, endX, endY);
-      for (const hitBox of hitBoxes) {
-        const res = gdjs.Polygon.raycastTest(hitBox, x, y, endX, endY);
-        if (res.collision) {
-          if (closest && res.closeSqDist < testSqDist) {
-            testSqDist = res.closeSqDist;
+      if (closest) {
+        let sqDistMin = raySqBoundingRadius;
+        const hitBoxes = this.getHitBoxesAround(x, y, endX, endY);
+        for (const hitBox of hitBoxes) {
+          const res = gdjs.Polygon.raycastTest(hitBox, x, y, endX, endY);
+          if (res.collision && res.closeSqDist < sqDistMin) {
+            sqDistMin = res.closeSqDist;
             gdjs.Polygon.copyRaycastTestResult(res, result);
-          } else if (
-            !closest &&
-            res.farSqDist > testSqDist &&
+          }
+        }
+      } else {
+        let sqDistMax = 0;
+        const hitBoxes = this.getHitBoxesAround(x, y, endX, endY);
+        for (const hitBox of hitBoxes) {
+          const res = gdjs.Polygon.raycastTest(hitBox, x, y, endX, endY);
+          if (
+            res.collision &&
+            res.farSqDist > sqDistMax &&
             res.farSqDist <= raySqBoundingRadius
           ) {
-            testSqDist = res.farSqDist;
+            sqDistMax = res.farSqDist;
             gdjs.Polygon.copyRaycastTestResult(res, result);
           }
         }
       }
+
       return result;
     }
 
--- a/GDJS/Runtime/runtimeobject.ts
+++ b/GDJS/Runtime/runtimeobject.ts
@@ -2343,7 +2343,7 @@ namespace gdjs {
           }
         }
       } else {
-        let sqDistMax = 0;
+        let sqDistMax = -Number.MAX_VALUE;
         const hitBoxes = this.getHitBoxesAround(x, y, endX, endY);
         for (const hitBox of hitBoxes) {
           const res = gdjs.Polygon.raycastTest(hitBox, x, y, endX, endY);
--- a/GDJS/Runtime/runtimeobject.ts
+++ b/GDJS/Runtime/runtimeobject.ts
@@ -2333,7 +2333,7 @@ namespace gdjs {
 
       // Do a real check if necessary.
       if (closest) {
-        let sqDistMin = raySqBoundingRadius;
+        let sqDistMin = Number.MAX_VALUE;
         const hitBoxes = this.getHitBoxesAround(x, y, endX, endY);
         for (const hitBox of hitBoxes) {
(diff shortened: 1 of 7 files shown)


GDJS/Runtime/runtimeobject.ts
```diff
@@ -2330,16 +2339,14 @@ namespace gdjs {
...
-            }
+            gdjs.Polygon.copyRaycastTestResult(res, result);
+          } else if (
+            !closest &&
+            res.farSqDist > testSqDist &&
+            res.farSqDist <= raySqBoundingRadius
+          ) {
+            testSqDist = res.farSqDist;
+            gdjs.Polygon.copyRaycastTestResult(res, result);
           }
```
D8H, line 2350: The fix is to avoid to override previous results during the loop, right?

I have only some suggestion about readability.

Rename `closest` to `shouldSearchClosest`?
`testSqDist` could be split in `sqDistMin` and `sqDistMax` for clarity

A if-else could better show that only one of the 2 blocks can be executed in one loop. I would even prefer the loop to be inside the if-else.
```JavaScript
if (closest) {
  if ()
}
else {
  if ()
}
```

GDJS/tests/tests/runtimeobject.js
```diff
@@ -228,26 +228,193 @@
...
-    expect(object1.raycastTest(500, 35, 200+9, 35, true).collision).to.be(true);
-    expect(object1.raycastTest(500, 35, 200+10, 35, true).collision).to.be(true);
-    expect(object1.raycastTest(500, 35, 200+10.1, 35, true).collision).to.be(false);
-    expect(object1.raycastTest(500, 35, 200+10.5, 35, true).collision).to.be(false);
-    expect(object1.raycastTest(500, 35, 200+11, 35, true).collision).to.be(false);
+    expect(object1.raycastTest(500, 35, 200 + 9, 35, true).collision).to.be(
+      true
+    );
+    expect(object1.raycastTest(500, 35, 200 + 10, 35, true).collision).to.be(
+      false
```
4ian, line 257: @D8H I'm not sure why this test expected result changed to false. Any idea? (and same for another line below)

GDJS/Runtime/runtimeobject.ts
4ian: > The fix is to avoid to override previous results during the loop, right?

Yes!

> I have only some suggestions about readability.

Fair enough, I see what you're saying! I'll give it a try.

GDJS/Runtime/runtimeobject.ts
```diff
@@ -2323,26 +2332,32 @@ namespace gdjs {
...
+          }
+        }
+      } else {
+        let sqDistMax = 0;
+        const hitBoxes = this.getHitBoxesAround(x, y, endX, endY);
+        for (const hitBox of hitBoxes) {
+          const res = gdjs.Polygon.raycastTest(hitBox, x, y, endX, endY);
+          if (
+            res.collision &&
+            res.farSqDist > sqDistMax &&
```
D8H, line 2352: sqDistMax should probably be set `-Number.MAX_VALUE` to allow a point at a distance = 0 to be detected. I guess `raycastTest` already fulfill the role to ensure values under 0 won't be returned.

The same thing could be done for `sqDistMax`.

4ian: > The same thing could be done for sqDistMax

You mean sqDistMin?

D8H: yes 😄

4ian: I did the change but I'm not sure there is anything to do for sqDistMax, it's already initialised at the maximum radius we are allowing the ray to go.

GDJS/Runtime/runtimeobject.ts
```diff
@@ -2323,26 +2332,32 @@ namespace gdjs {
...
-            ) {
-              testSqDist = res.farSqDist;
-              result = res;
-            }
+      if (closest) {
+        let sqDistMin = raySqBoundingRadius;
+        const hitBoxes = this.getHitBoxesAround(x, y, endX, endY);
+        for (const hitBox of hitBoxes) {
+          const res = gdjs.Polygon.raycastTest(hitBox, x, y, endX, endY);
+          if (res.collision && res.closeSqDist < sqDistMin) {
```
D8H, line 2340: This condition will not allow `res.closeSqDist === raySqBoundingRadius`.
`let sqDistMin = Number.MAX_VALUE;` will also allow values that would be a bit more than raySqBoundingRadius because of eventual rounding error which make the code stronger because it's probably better to find something a bit too far that not finding something that is 1e-16 too far.

GDJS/Runtime/runtimeobject.ts
D8H: Do you mean sqDistMin?

GDJS/Runtime/runtimeobject.ts
4ian: Good catch, this was an issue. I used `Number.MAX_VALUE `.

PR #4189, opened 2022-08-07
Add possibility to seal an instance (impossible to select it)

AlexandreSi:
To resolve https://forum.gdevelop.io/t/new-lock-ui-option/41304

- Add a new boolean to the instance `sealed`
  - When an instance is sealed, pixi will ignore the pixi instance when mouse is over.
  - When selecting multiple instances with Click => Pan => Release click (rectangle selection), the sealed instances won't be selected.
- Add a new checkbox to control this boolean, enabled only if the instance is already locked
  
  image

- Make it possible to leave the sealed state from the instance list:

  image

### DIFF
--- a/Core/GDCore/Project/InitialInstance.cpp
+++ b/Core/GDCore/Project/InitialInstance.cpp
@@ -28,6 +28,7 @@ InitialInstance::InitialInstance()
       width(0),
       height(0),
       locked(false),
+      sealed(false),
       persistentUuid(UUID::MakeUuid4()) {}
 
 void InitialInstance::UnserializeFrom(const SerializerElement& element) {
--- a/Core/GDCore/Project/InitialInstance.cpp
+++ b/Core/GDCore/Project/InitialInstance.cpp
@@ -11,9 +11,7 @@
 #include "GDCore/Project/Project.h"
 #include "GDCore/Serialization/SerializerElement.h"
 #include "GDCore/Tools/UUID/UUID.h"
-#if defined(GD_IDE_ONLY)
 #include "GDCore/Project/PropertyDescriptor.h"
-#endif
 
 namespace gd {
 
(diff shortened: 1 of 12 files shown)


Core/GDCore/Project/InitialInstance.cpp
```diff
@@ -83,7 +83,8 @@ void InitialInstance::SerializeTo(SerializerElement& element) const {
   element.SetAttribute("customSize", HasCustomSize());
   element.SetAttribute("width", GetCustomWidth());
   element.SetAttribute("height", GetCustomHeight());
-  element.SetAttribute("locked", IsLocked());
+  if (IsLocked()) element.SetAttribute("locked", IsLocked());
```
4ian, line 86: Very nice!

Core/GDCore/Project/InitialInstance.h
```diff
@@ -127,19 +127,31 @@ class GD_CORE_API InitialInstance {
...
   void SetLocked(bool enable = true) { locked = enable; }
 
+  /**
+   * \brief Return true if the instance cannot be selected by clicking on it
+   * in the IDE (only applies if instance is also locked).
+   */
+  bool IsSealed() const { return sealed; };
+
+  /**
+   * \brief (Un)lock the initial instance.
```
4ian, line 148: `(Un)seal`?

newIDE/app/src/InstancesEditor/InstancesSelection.js
```diff
@@ -27,34 +27,52 @@ export default class InstancesSelection {
     this.selection.length = 0;
   }
 
-  selectInstance(
+  selectInstance({
+    instance,
+    multiSelect,
+    layersVisibility = null,
+    ignoreSeal = false,
+  }: {
```
4ian, line 35: You probably should {| |} that.

4ian (approved):

Core/GDCore/Project/InitialInstance.h
AlexandreSi: Good catch, ty!

PR #4194, opened 2022-08-09
Fix project save

ClementPasteau:
Do not show in changelog.

The save was not triggered when the project was defined, because the fileMetadata was not initialised anymore.
Also took the opportunity to fix an error appearing when GDJS reloads with recursive-copy, because of latest Node versions (https://github.com/timkendrick/recursive-copy/issues/24)

### DIFF
--- a/newIDE/app/package.json
+++ b/newIDE/app/package.json
@@ -22,7 +22,7 @@
     "prettier": "1.15.3",
     "react-error-overlay": "^6.0.9",
     "react-scripts": "4.0.3",
-    "recursive-copy": "^2.0.10",
+    "recursive-copy": "^2.0.14",
     "recursive-readdir": "^2.2.2",
     "shelljs": "0.8.4",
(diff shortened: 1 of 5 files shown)


newIDE/app/package.json
```diff
@@ -22,7 +22,7 @@
     "prettier": "1.15.3",
     "react-error-overlay": "^6.0.9",
     "react-scripts": "4.0.3",
-    "recursive-copy": "^2.0.10",
```
AlexandreSi, line 25: 🧹

AlexandreSi (approved):
Thanks!

4ian (approved):
Make sure to check no errors when creating a new project on the web-app, on the desktop app (that's what you've fixed) and when opening an example from an url on the web-app :)

ClementPasteau:
tested the 3 of them, all look good 👍

4ian (approved):

PR #4195, opened 2022-08-09
Fix the export of game using Box2d WASM library

D8H:
Fix the export, but the preview no longer work.


4ian:
I wonder what's making the preview to behave differently 🤔

4ian:
A network preview works: 
image

So the issue is because of the preview done locally, which does not use a web server. We run directly the index.html of the game as a local file, thanks to Electron access to the file system:
image

Sadly, it seems it changes the way the file is resolved.

Extensions/Physics2Behavior/Box2D_v2.3.1_min.wasm.js
```diff
@@ -18,7 +18,7 @@ function Fa(b){for(;0<b.length;){var c=b.shift();if("function"==typeof c)c(a);el
 function na(b){if(a.onAbort)a.onAbort(b);b+="";pa(b);qa(b);ua=!0;throw new WebAssembly.RuntimeError("abort("+b+"). Build with -s ASSERTIONS=1 for more info.");}function Pa(b){var c=Qa;return String.prototype.startsWith?c.startsWith(b):0===c.indexOf(b)}function Ra(){return Pa("data:application/octet-stream;base64,")}
 
 // The wasm file path is defined here.
-var Qa="Extensions/Physics2Behavior/Box2D_v2.3.1_min.wasm.wasm";
+var Qa="Box2D_v2.3.1_min.wasm.wasm";
 
 if(!Ra()){var Sa=Qa;Qa=a.locateFile?a.locateFile(Sa,ia):ia+Sa}
```
4ian, line 23: @D8H You can use `locateFile` to avoid having to modify the JS file: https://emscripten.org/docs/porting/files/packaging_files.html#changing-the-data-file-location

4ian: We actually use this here (for another purpose): https://github.com/4ian/GDevelop/blob/21b525ab989bde34e4f5c2861bd15d962e20a9d6/newIDE/app/src/index.js#L91-L95

4ian:
@D8H This is what we can do:
- Pass locateFile when you initialise the library.
- If location.protocol === 'file', locateFile should return `Extensions/Physics2Behavior/Box2D_v2.3.1_min.wasm.wasm`. 
- Otherwise, locateFile should return `Box2D_v2.3.1_min.wasm.wasm`. 

Check that:
- it works in preview and when the game is exported on desktop (i.e: the game runs in Electron)
- it works in network preview, Liluo.io and HTML5 exported on a webserver (i.e: the game runs on any webserver)
- it works when exported to an Android apk (i.e: the game runs on Cordova, which should run the game on a webserver, so this should be equivalent to the previous case... but let's be sure).

Extensions/Physics2Behavior/Box2D_v2.3.1_min.wasm.js
```diff
@@ -7,20 +7,22 @@ function(Box2D) {
...
 
 
-var a;a||(a=typeof Box2D !== 'undefined' ? Box2D : {});var aa;a.ready=new Promise(function(b){aa=b});var ba={},ca;for(ca in a)a.hasOwnProperty(ca)&&(ba[ca]=a[ca]);var da=!1,ea=!1,fa=!1,ha=!1;da="object"===typeof window;ea="function"===typeof importScripts;fa="object"===typeof process&&"object"===typeof process.versions&&"string"===typeof process.versions.node;ha=!da&&!fa&&!ea;var ia="",ja,ka,la,ma;
+var a;a||(a=typeof Box2D !== 'undefined' ? Box2D : {
+  locateFile: function(path, prefix) {
+    return location.protocol === 'file' ?
+        // This is needed to run in preview.
+        prefix + "Extensions/Physics2Behavior/" + path :
+        prefix + path;
+  },
```
D8H, line 16: Do you think it's okay to put it here?

(I only tested a preview and a Liluo build)

4ian: Careful the protocol contains the `:` too.

Extensions/Physics2Behavior/Box2D_v2.3.1_min.wasm.js
```diff
@@ -7,20 +7,22 @@ function(Box2D) {
   Box2D = Box2D || {};
 
 
-var a;a||(a=typeof Box2D !== 'undefined' ? Box2D : {});var aa;a.ready=new Promise(function(b){aa=b});var ba={},ca;for(ca in a)a.hasOwnProperty(ca)&&(ba[ca]=a[ca]);var da=!1,ea=!1,fa=!1,ha=!1;da="object"===typeof window;ea="function"===typeof importScripts;fa="object"===typeof process&&"object"===typeof process.versions&&"string"===typeof process.versions.node;ha=!da&&!fa&&!ea;var ia="",ja,ka,la,ma;
+var a;a||(a=typeof Box2D !== 'undefined' ? Box2D : {
+  locateFile: function(path, prefix) {
+    return location.protocol === 'file' ?
+        // This is needed to run in preview.
```
4ian, line 13:
```suggestion
        // This is needed to run on Electron.
```

Extensions/Physics2Behavior/Box2D_v2.3.1_min.wasm.js
4ian: As for the question, it's a bit dirty and it would be better to have a global initializeBox2D that we can call like `initializeBox2D({ locateFile: ... }).then(Box2D => /*use it/store it somewhere*/)`. But it's ok if it's simpler.

D8H: What do you mean by "global"?
I think each extension should have their own library. For instance, there is another version of Box2D in the deprecated Physics extension.

I guess, we could move the implementation of `locateFile` at the end of the lib file where the promise is registered, but the inside of the library still needs to be modified a bit.

4ian: I meant the bit at the end that register the library :) We currently have this:
```
gdjs.registerAsynchronouslyLoadingLibraryPromise(Box2D(window.Box2D).then(box2d => {
  window.Box2D = box2d;
}));
```

I *think* this could be changed to:
```
gdjs.registerAsynchronouslyLoadingLibraryPromise(initializeBox2D({ locateFile: ... }).then(box2d => {
  window.Box2D = box2d;
}));
```

If you basically rename the IIFE that you put around the library:
image
into `initializeBox2D`:
image

This should work because what you pass to this IIFE (that I renamed `initializeBox2D` is an object that is used to pass options to the library, including locateFile):
image


So in short, the process should be:
- Take the original Box2D .wasm.js file
- Put it all in a IIFE like done already, but name this IIFE `initializeBox2D` (that's unfortunately a global variable).
- Add the promise registration at the end: 
  ```
  gdjs.registerAsynchronouslyLoadingLibraryPromise(initializeBox2D({locateFile: ... }).then(box2d => {
    window.Box2D = box2d;
  }));
  ```
- That's it! 
- If you are worried about `initializeBox2D` being left in the global namespace, you can consider wrapping the whole thing into another IIFE (`(function() { ... })()`) to avoid leaking it.

Extensions/Physics2Behavior/Box2D_v2.3.1_min.wasm.js
D8H: I tested a windows build and there is not the same issue as for preview. Does it use Electron too?

4ian: Alright, weird 🤔 A windows build is definitely Electron based, you can look at the installed files! 
Would you be able to find what "location.protocol" is returning? I want to be 100% sure of what this is.

D8H: I tried again a Windows build withe the new version and it's working. But, I don't have access to the console so I don't know which protocol is used.
I've probably not tested is correctly the 1st time.

Extensions/Physics2Behavior/Box2D_v2.3.1_min.wasm.js
```diff
@@ -1,10 +1,10 @@
-
-var Box2D = (function() {
+(function() {
+var initializeBox2D = (function() {
   var _scriptDir = typeof document !== 'undefined' && document.currentScript ? document.currentScript.src : undefined;
   if (typeof __filename !== 'undefined') _scriptDir = _scriptDir || __filename;
   return (
-function(Box2D) {
-  Box2D = Box2D || {};
+function(Box2D, locateFile) {
```
4ian, line 6: There must be some confusion for the first argument (it's ill named, but that's sadly how it's done by the library). It's nothing more than an object composed of options that are passed to the js library. This parameter will be undefined because you pass window.Box2D, which is undefined, and so only the second argument is useful.

In other words, this parameter should have been better named "options". The library is built as this so it's confusing, but in your mind just consider it's called "options".
So you do NOT need the second argument. Remove it entirely:

- Keep only the first argument and pass `{locateFile: ...}` to it. 
- This is how it's done here:
 
 https://github.com/4ian/GDevelop/blob/3e71c3a63fda336ca016e5e71a02bad84e3a9aa2/newIDE/app/src/index.js#L88-L96

Extensions/Physics2Behavior/Box2D_v2.3.1_min.wasm.js
```diff
@@ -627,6 +622,12 @@ Zv();a.b2Manifold.e_faceA=$v();a.b2Manifold.e_faceB=aw();a.b2_staticBody=bw();a.
 );
 })();
 
-gdjs.registerAsynchronouslyLoadingLibraryPromise(Box2D(window.Box2D).then(box2d => {
+gdjs.registerAsynchronouslyLoadingLibraryPromise(initializeBox2D(window.Box2D, function(path, prefix) {
```
4ian, line 625:
```suggestion
gdjs.registerAsynchronouslyLoadingLibraryPromise(initializeBox2D({ locateFile: (path, prefix) => {
  return ...
}}).then(Box2D => window.Box2D = Box2D)
```

4ian:
LGTM, at least in theory that's all I think is needed! Should work in preview, desktop export (electron), cordova, liluo.io and classic HTML5!

D8H:
Tested on:
- Preview
- Liluo.io
- Windows
- Android

D8H:
The test failure seems unrelated.
```
e 104.0.5112 (Linux 0.0.0): Executed 30 of 890 SUCCESS (0 secs / 3.433 secs) 00:11
e 104.0.5112 (Linux 0.0.0) Firebase extension end-to-end tests Cloud Firestore Base read/write operations should delete and detect presence of a field from Firestore FAILED 00:11
	Error: timeout of 5000ms exceeded 00:11
HeadlessChrome 104.0.5112 (Linux 0.0.0): Executed 31 of 890 (1 FAILED) (0 secs / 8.434 secs) 00:12
e 104.0.5112 (Linux 0.0.0): Executed 32 of 890 (1 FAILED) (0 secs / 8.688 secs)
...
e 104.0.5112 (Linux 0.0.0): Executed 879 of 890 (1 FAILED) (skipped 9) (0 secs / 35.427 secs)
```

4ian:
Thank you! I've relaunched the CI to be sure.

PR #4197, opened 2022-08-11
Upgrade Admob so that it works with the latest Android SDK

4ian:
The build service is being upgraded to use the latest Android SDK (API level 32) as required by the Play Store. This needs to also update the Cordova Admob plugin to work around a crash due to an old google ads service using a dependency crashing when targeting this API level.


ClementPasteau (approved):
👍

Issue #4187, opened 2022-08-06
"Play sound on a channel" looks duplicated because "Unpause sound" function has same name


tristanbob:
## Describe the bug

Users see two actions with very similar names:

"Play a sound on a channel"
"Play the sound of a channel" 

One action is for starting a sound (includes parameters about the sound file)
One action is for playing a sound that was paused/stopped (no parameters other than channel)

I think we should rename the second one to "Play (unpause) a sound on a channel that has been paused/stopped".
(However, I'm not sure if this action is meant for pause and stopped sounds)

![image](https://user-images.githubusercontent.com/8879811/183258732-1f5ae467-2d49-4abe-97b1-2127360dedda.png)

4ian:
Right, we really need a rework of these namings 😅
What about "Resume" or "Continue"? Would that be understandable?

tristanbob:
I like the "Resume" wording. Either of these options sound good to me:

"Resume playing a sound on a channel"
"Resume playing a sound on a channel that was paused/stopped"

BTW, this same change needs to be done to Music actions as well.

Silver-Streak:
Alternative verbiage:

"Resume playing a paused sound on a channel"

PR #4199, opened 2022-08-11
Fix the action name

Bouh:
Fix https://github.com/4ian/GDevelop/issues/4187


4ian:
I also updated the music because consistency is very important :)

PR #4206, opened 2022-08-17
Fix the extension search bar that was only using the first 30 tags alphabetically.

D8H:
Fixes a bug reported by pampogokiraly on Discord.


4ian (approved):
Did you find any performance implications?

D8H:
> Did you find any performance implications?

It was very smooth on my computer.

PR #4207, opened 2022-08-17
Limit the number of autocompletion elements displayed for extensions search tags.

D8H:


newIDE/app/src/UI/SearchBar.js
```diff
@@ -243,6 +245,8 @@ const SearchBar = React.forwardRef<Props, SearchBarInterface>(
       }
     };
 
+    const filterTags = createFilterOptions();
```
4ian, line 248: Can you use `{ limit: 200 }`?
If yes, would be good to create this outside of the component so there is one less thing to instantiate at every render.

newIDE/app/src/UI/SearchBar.js
```diff
@@ -299,6 +303,9 @@ const SearchBar = React.forwardRef<Props, SearchBarInterface>(
                     <Autocomplete
                       id={id}
                       options={tags}
+                      filterOptions={(options, state) =>
```
4ian, line 306: I think using the limit option of createFilterOptions will avoid yet-another-function.

4ian (approved):

PR #4208, opened 2022-08-17
Add an events based object editor prototype

D8H:
### Critical
- Only plain properties can be set on objects, it makes impossible to configures the children sprites.
  - This needs some think through and will be in another PR.
- BehaviorsSharedData are not updated when a behavior is added to a child-object.
  - I guess it's something that could be added after the prototype.
- Events-based object deserialization won't work if an object use another Events-based object because the EventBasedObject may not be deserialized yet.
  - We can address this issue in the next PR.

### Nice to have
- The variable editor of child-objects won't do autocompletion with undeclared variables.

### Useless?
- Allow extension designer to use place holder in the extension?
  - The resource provider is not pass through the extension editor.
  - It could be confusing.
- Should event-based object have private behaviors?
  - The behavior can be attached to a child as a workaround.
- There is no tag filtering on the child-object list.
  - There won't be many children, the feature is a bit useless in this context.

## Technical solution

There are some codes for objects that are very similar to the one for behaviors (for instance `declareObjectInstructionOrExpressionMetadata` and `declareBehaviorInstructionOrExpressionMetadata`) but I think they should not be factorized because it's likely they will keep diverging.


newIDE/app/src/EventsFunctionsExtensionEditor/index.js
```diff
@@ -1021,9 +1411,15 @@ export default class EventsFunctionsExtensionEditor extends React.Component<
...
-                      ) || initialMosaicEditorNodes
+                      isDev !==
+                      mosaicContainsNode(
+                        initialMosaicEditorNodes,
+                        'objects-list'
+                      )
+                        ? getDefaultEditorMosaicNode(
+                            'events-functions-extension-editor'
+                          ) || initialMosaicEditorNodes
+                        : initialMosaicEditorNodes
```
D8H, line 1422: The mosaic is reset to let the `'objects-list'` appear only in dev mode.

Core/GDCore/Project/AbstractEventsBasedEntity.h
```diff
@@ -0,0 +1,152 @@
...
+   */
+  virtual void UnserializeFrom(gd::Project& project,
+                       const gd::SerializerElement& element);
+  ///@}
+
+ private:
+  gd::String name;
+  gd::String fullName;
+  gd::String description;
+  gd::String objectType;
```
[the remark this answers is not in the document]
D8H: No, I forgot to remove it.

Core/GDCore/Project/EventsBasedObject.h
```diff
@@ -0,0 +1,72 @@
...
+  EventsBasedObject(const gd::EventsBasedObject &_eventBasedObject);
+
+  /**
+   * \brief Return a pointer to a new EventsBasedObject constructed from
+   * this one.
+   */
+  EventsBasedObject* Clone() const { return new EventsBasedObject(*this); };
+
+  EventsBasedObject& SetDescription(const gd::String& description_) override {
+    AbstractEventsBasedEntity::SetDescription(description_);
```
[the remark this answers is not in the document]
D8H: This is to refine the returned type to allow call chaining.

Core/GDCore/IDE/EventsFunctionTools.cpp
```diff
@@ -72,4 +73,41 @@ void EventsFunctionTools::BehaviorEventsFunctionToObjectsContainer(
...
+    gd::ObjectsContainer& outputGlobalObjectsContainer,
+    gd::ObjectsContainer& outputObjectsContainer) {
+  // The context is build the same way as free function...
+  FreeEventsFunctionToObjectsContainer(project,
+                                       eventsFunction,
+                                       outputGlobalObjectsContainer,
+                                       outputObjectsContainer);
+
+  // TODO EBO these 2 checks are not necessary for the following code to work.
+  // Should they be removed?
```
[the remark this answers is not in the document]
D8H: Ok, I'll remove the todo.

Core/GDCore/Project/EventsBasedObject.h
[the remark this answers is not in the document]
4ian: I see. We could make the public base class to allow chaining functions calls (but then you lose the type of the child - i think there is a way to workaround this, but not something important at this stage.)

Core/tests/WholeProjectRefactorer.cpp
```diff
@@ -31,134 +31,536 @@
...
+    gd::Instruction instruction;
+    instruction.SetType(
+        "MyEventsExtension::MyEventsBasedObject::" +
+        gd::EventsBasedObject::GetPropertyConditionName("MyProperty"));
+    event.GetConditions().Insert(instruction);
+    eventList.InsertEvent(event);
+  }
+
+  // Create an event in the layout using "MyProperty" expression
+  // TestEvent: ObjectPropertyExpression
```
4ian, line 312: Any chance you could somehow detect the position of the inserted event, and if the `position != ObjectPropertyExpression`, then throw (because the set up would be done incorrectly, so the tests will fail for sure).

GDevelop.js/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ endif()
 
 # Compilation flags (https://emscripten.org/docs/tools_reference/emcc.html):
 SET( CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -O2") # Optimizations during compilation
-# SET( CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} --profiling") # Profiling support
+SET( CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} --profiling") # Profiling support
```
4ian, line 17: To remove before merging :)

newIDE/app/src/AssetStore/InstallAsset.spec.js
```diff
@@ -6,7 +6,6 @@ import {
   filterMissingBehaviors,
   downloadExtensions,
   installAsset,
-  getRequiredExtensionsForEventsFromAsset,
```
4ian, line 9: I'm curious what problem this capability was giving you?

D8H: The asset store wants an EventList to add events on the object.
I could have done a new dialog to create an object without the asset store tab, but I wasn't sure if I should as the asset store could be interesting to have placeholders while designing an extension. I was more convinced that adding events directly at the end of the scene events was not something to do.

newIDE/app/src/MainFrame/Preferences/PreferencesContext.js
```diff
@@ -21,6 +21,8 @@ export type AlertMessageIdentifier =
   | 'function-extractor-explanation'
   | 'events-based-behavior-explanation'
   | 'empty-events-based-behavior-explanation'
+  | 'events-based-object-explanation'
+  | 'empty-events-based-object-explanation'
```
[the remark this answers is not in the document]
4ian: I think like this it works well.

newIDE/app/src/ObjectEditor/ObjectsEditorService.js
```diff
@@ -17,6 +17,9 @@ const ObjectsEditorService = {
       console.warn(
         `Object with type ${objectType} has no editor configuration registered. Please use registerEditorConfiguration to register your editor.`
       );
+      return this.getDefaultObjectJsImplementationPropertiesEditor({
```
4ian, line 20: Isn't there any other check to do?

4ian (approved):
A few more comments.
Two questions to finish:

- Is this mergeable without regressions?
- Are each of your points (especially the critical ones) in your PR description present as a "TODO EBO" in the code? If not it would be great to add them, or make sure to move them to the next PR so that we don't lose the context!

Thanks :)

newIDE/app/src/ObjectEditor/ObjectsEditorService.js
D8H: Yes, I added a check on the existence of the EBO and only warn when it's not found.

D8H:
I tested the last version with a game with 5 event-based behaviors. I could modify the extensions and run the game without any issue.

Core/GDCore/Project/Project.cpp
```diff
@@ -81,22 +82,62 @@ void Project::ResetProjectUuid() { projectUuid = UUID::MakeUuid4(); }
...
+    const gd::String& platformName) const {
+    if (Project::HasEventsBasedObject(type)) {
+      auto &eventsBasedObject = Project::GetEventsBasedObject(type);
+      auto customObject = gd::make_unique<CustomObject>(eventsBasedObject, type);
+      customObject->SetName(name);
+      return customObject;
+    }
+    else {
+      for (std::size_t i = 0; i < platforms.size(); ++i) {
+      if (!platformName.empty() && platforms[i]->GetName() != platformName)
```
4ian, line 94: Nit: Indentation

newIDE/app/src/EventsBasedObjectEditor/EventBasedObjectChildrenEditor.js
```diff
@@ -0,0 +1,278 @@
...
+
+  _onRenameObject = (
+    objectWithContext: ObjectWithContext,
+    newName: string,
+    done: boolean => void,
+    i18n: I18nType
+  ) => {
+    // TODO EBO Use a constant instead a hard coded value "Object".
+    if (newName === 'Object') {
+      showWarningBox(i18n._(t`"Object" is a reserved name for the parent.`), {
```
4ian, line 112: is a reserved name, used for the parent object in the events (actions, conditions, expressions...). Please choose another name.

(for error message, state the problem and then what to do/how to fix the problem)

newIDE/app/src/EventsBasedObjectEditor/index.js
```diff
@@ -0,0 +1,142 @@
...
+                .getEventsFunctions()
+                .getEventsFunctionsCount() === 0 && (
+                <DismissableAlertMessage
+                  identifier="empty-events-based-object-explanation"
+                  kind="info"
+                >
+                  <Trans>
+                    Once you're done, close this dialog and start adding some
+                    functions to the object. Then, test the object by adding it
+                    to an object in a scene.
```
4ian, line 116: Message to update

4ian, line 116: Message to update

Core/GDCore/Project/Project.cpp
```diff
@@ -642,6 +683,29 @@ void Project::UnserializeFrom(const SerializerElement& element) {
   if (currentPlatform == NULL && !platforms.empty())
     currentPlatform = platforms.back();
 
+  // TODO EBO All events based behaviors and events based objects should be
+  // added to the project before the content of events based objects is
+  // unserialized because when unserializing child-objects InsertNewObject and
```
4ian, line 688: Might be a good reason for storing separately the data of an object (something like "gd::BehaviorContent") from the code manipulating it (something like "gd::Behavior") (noting that the naming is not perfect)

4ian, line 688: Might be a good reason for storing separately the data of an object (something like "gd::BehaviorContent") from the code manipulating it (something like "gd::Behavior") (noting that the naming is not perfect)

D8H: I could probably work but Behavior can use the same logic as CustomObject and it gives better encapsulation.
I guess that BehaviorContent was used because the implementation was JS side.

I think that a solution would be to unserialize only empty EventsBasedObjet in a 1st pass and then their content. I have a PR ready for it when this one is merged.

D8H:
I tested the last version with Spherez, Absorbus and Platformer example.
They are playable on preview.
On Spherez, I:
- added a property and a function in a behavior
- used undo on an event sheet
- saved the project and reopen it

4ian (approved):

Issue #4039, opened 2022-06-16
Large Tilemaps cut off Horizontally


Silver-Streak:
## Describe the bug
When a tilemap is past a certain size, the map image will cut off horizontally both in the IDE and in previews. The map dimensions still render correctly in the IDE. This was reported by Discord member (78), and I was able to reproduce it.

## To Reproduce
1. Create a tilemap in Tiled (or in LDTK and export to tiled)
2. Ensure the map has at least one axis around ~8000 pixels large and another axis around ~5000 pixels large.
3. Export as JSON
4. Import as Tilemap object, add to scene.
5. The map will graphically be cut off (horizontally). This will be horizontal regardless of whether the map is larger vertically or horizontally. I'm assuming this is likely just due to how the json data orders data, but it still feels odd.

Attached is a project with both a horizontal and vertical map experiencing this issue.
[TileMapCutoff.zip](https://github.com/4ian/GDevelop/files/8915432/TileMapCutoff.zip)

Some examples:
Veritcal map in Tiled:
![image](https://user-images.githubusercontent.com/24237149/173985486-5a7a284b-4b71-4520-a9a0-5f23597650a6.png)

Vertical map in Editor (and preview):
![image](https://user-images.githubusercontent.com/24237149/173985523-14babc4a-0dbd-48cb-bd96-a000b6828a9e.png)

Horizontal Map in Tiled:
![image](https://user-images.githubusercontent.com/24237149/173985601-473ae313-0480-4391-a72a-3a344893791b.png)

Horizontal map in Editor (and preview): 
![image](https://user-images.githubusercontent.com/24237149/173985628-fb79664a-34c8-4268-9339-8eee5ea1c7c5.png)

Workaround:
You can split any map into ~4000x4000 smaller maps and create multiple tilemap objects, and it'll work fine. Based on how tilemaps work, this really shouldn't be required (Or should, I guess, be automated by the engine to split a larger map into smaller consumable maps).

## Other details
* Windows 10 Pro 64 bit
* Tested with both 5.0.135 and 136

willianholtz:
normal for me..
![Screenshot_20220616_210628](https://user-images.githubusercontent.com/23488207/174197304-68b5bed0-db1f-4a61-936f-0ae3986386ad.png)

Linux: Arch

Silver-Streak:
> Unfortunately I couldn't open your project, it says it's not a valid GD5 file.
> 
> I'd like to be able to test on Linux too, so I did my test and it works.

[...]

PR #4212, opened 2022-08-18
Upgrade the Pixi tile map extension to 3.2.2

D8H:
Cherry picked from:
- https://github.com/4ian/GDevelop/pull/2828

Fixes https://github.com/4ian/GDevelop/issues/4039

image

### DIFF
--- a/Extensions/TileMap/JsExtension.js
+++ b/Extensions/TileMap/JsExtension.js
@@ -956,6 +956,7 @@ module.exports = {
         pixiRenderer
       );
 
+      Tilemap.settings.use32bitIndex = true;
       pixiRenderer.plugins.tilemap =
         pixiRenderer.plugins.tilemap || new Tilemap.TileRenderer();
       this.tileMapPixiObject = new Tilemap.CompositeTilemap();
--- a/Extensions/TileMap/JsExtension.js
+++ b/Extensions/TileMap/JsExtension.js
@@ -956,6 +956,7 @@ module.exports = {
         pixiRenderer
       );
 
+      // This setting allows tile maps with more than 16K tiles.
       Tilemap.settings.use32bitIndex = true;
       pixiRenderer.plugins.tilemap =
         pixiRenderer.plugins.tilemap || new Tilemap.TileRenderer();
--- a/Extensions/TileMap/JsExtension.js
+++ b/Extensions/TileMap/JsExtension.js
@@ -942,7 +942,8 @@ module.exports = {
       instance,
       associatedObject,
       pixiContainer,
-      pixiResourcesLoader
+      pixiResourcesLoader,
+      pixiRenderer
     ) {
       RenderedInstance.call(
         this,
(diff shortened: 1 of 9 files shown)


Extensions/TileMap/tilemapruntimeobject-pixi-renderer.ts
```diff
@@ -26,8 +26,20 @@ namespace gdjs {
       this._object = runtimeObject;
       this._runtimeScene = runtimeScene;
 
+      const pixiRenderer = runtimeScene
+        .getGame()
+        .getRenderer()
+        .getPIXIRenderer();
+
+      PIXI.tilemap.settings.use32bitIndex = true;
```
D8H, line 34: This is what fix the tiles limit. It takes a bit more memory but it should be fine.

ClementPasteau: maybe adding a comment above this line, and the one in `JSExtension` to say why it's important to keep. 
Something like "This is required to allow a high number of tiles"

Extensions/TileMap/JsExtension.js
```diff
@@ -951,10 +952,14 @@ module.exports = {
         instance,
         associatedObject,
         pixiContainer,
-        pixiResourcesLoader
+        pixiResourcesLoader,
+        pixiRenderer
       );
 
-      this.tileMapPixiObject = new Tilemap.CompositeRectTileLayer(0);
+      Tilemap.settings.use32bitIndex = true;
```
ClementPasteau, line 959: I'm probably missing something, but why is this line present twice? in the extension and in the pixi-object-renderer?

ClementPasteau (commented):
lgtm, but I'll let @4ian approve :)

Extensions/TileMap/JsExtension.js
D8H: This is because the renderer is used in the IDE and at runtime.

ClementPasteau: Oh yeah, there're different instances. 👍 thanks

Extensions/TileMap/tilemapruntimeobject-pixi-renderer.ts
4ian: Yep, let's add comments on both lines :)

4ian (approved):
Ok with additional comment to explain why these lines should not be removed :)

Extensions/TileMap/pixi-tilemap/dist/pixi-tilemap.umd.js
```diff
@@ -1,116 +1,384 @@
...
+         * Since {@link TextileResource} is a dual-column format, this should be even for packing
+         * efficiency. The optimal value is usually 4.
+         */
+        TEXTILE_UNITS: 1,
+
+        /** The scaling mode of the combined texture tiling. */
+        TEXTILE_SCALE_MODE: constants.SCALE_MODES.LINEAR,
+
+        /** This will enable 32-bit index buffers. It's useful when you have more than 16K tiles. */
         use32bitIndex: false,
```
D8H, line 99: The attribute already have a documentation.
Should we still add a comment in the code?

ClementPasteau: I'd better be safe than sorry, this helps documenting that we had bugs because we didn't set it to true before

PR #4216, opened 2022-08-22
Improve onboarding banner

ClementPasteau:
2 things:

- disabled auto onboarding message
- replaced card with banner

Goal: increase the visibility & completion of the onboarding for newcomers

Capture d’écran 2022-08-22 à 12 08 07


4ian (approved):

LuniMoon:
One mini change for one card:
Replace the copy/imagery of the first card on the second row:
Instead of "Follow a tutorial", replace with text "GDevelop tutorials" (cause that's where the card is taking them), and use the image from "How to get Started" so the first-comer has more idea of where to go next (which should be Wesley's playlist):
Screenshot 2022-08-22 at 12 33 01

4ian:
> Should the banner and the right card be aligned on the right side?

Good catch. Would be good to find a way to keep these aligned.

> Replace the copy/imagery of the first card on the second row:

Good idea too! Will make the transition between screens easier to understand.

ClementPasteau:
> And should we add the available languages for the tour?

We could, but they're displayed as soon as the user clicks on it, so I'm not sure it will add value.
We can try that later :)

PR #4217, opened 2022-08-22
Merge BehaviorContent into Behavior and add CustomBehavior.

D8H:


Core/GDCore/Project/BehaviorConfigurationContainer.cpp
```diff
@@ -0,0 +1,25 @@
...
+#include "GDCore/Project/PropertyDescriptor.h"
+
+namespace gd {
+
+BehaviorConfigurationContainer::~BehaviorConfigurationContainer(){};
+
+
+std::map<gd::String, gd::PropertyDescriptor> BehaviorConfigurationContainer::GetProperties() const {
+    return GetProperties(content);
+  };
```
D8H, line 17: Should I move this logic in BehaviorJsImplementation like what is done in ObjectJsImplementation?

Core/GDCore/Project/Project.cpp
```diff
@@ -671,16 +703,15 @@ void Project::UnserializeFrom(const SerializerElement& element) {
...
-  // instances from the Project.
   eventsFunctionsExtensions.clear();
   const SerializerElement& eventsFunctionsExtensionsElement =
       element.GetChild("eventsFunctionsExtensions");
   eventsFunctionsExtensionsElement.ConsiderAsArrayOf(
       "eventsFunctionsExtension");
+  // First, only unserialize behaviors and objects names.
+  // As event based objects can contains CustomObject and Custom Object,
+  // this allows them to reference EventBasedBehavior and EventBasedObject
+  // respectively.
```
D8H, line 714: With these 2 passes, we won't have any issues to unserialize extensions that depends on each other.
It could be done only for EventBasedObject but I find it clearer to do it for behavior and free functions too.

Core/GDCore/Project/CustomBehavior.h
```diff
@@ -0,0 +1,40 @@
...
+  virtual std::map<gd::String, gd::PropertyDescriptor>
+  GetProperties(const gd::SerializerElement &behaviorContent) const override;
+  virtual bool UpdateProperty(gd::SerializerElement &behaviorContent,
+                              const gd::String &name,
+                              const gd::String &value) override;
+  virtual void
+  InitializeContent(gd::SerializerElement &behaviorContent) override;
+
+private:
+  const EventsBasedBehavior &eventsBasedBehavior;
```
4ian, line 38: This means we have to be careful not to have any CustomBehavior living in memory after their associated EventsBasedBehavior was deleted/removed.
Do you think of any situation where this could happen?

Core/GDCore/Project/CustomBehavior.h
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * \brief A gd::Behavior that stores its content in JSON and forward the
+ * properties related functions to Javascript with Emscripten.
+ */
+class CustomBehavior : public gd::Behavior {
+public:
+  CustomBehavior(const EventsBasedBehavior &eventsBasedBehavior_,
+                 const gd::String &fullType)
+      : // Name is not important as this object is just a "blueprint"
+        // that is copied (see calls to AddObject).
```
4ian, line 19: Update this comment (AddObject => ?)

Core/GDCore/Project/CustomBehavior.h
```diff
@@ -0,0 +1,40 @@
...
+  GetProperties(const gd::SerializerElement &behaviorContent) const override;
+  virtual bool UpdateProperty(gd::SerializerElement &behaviorContent,
+                              const gd::String &name,
+                              const gd::String &value) override;
+  virtual void
+  InitializeContent(gd::SerializerElement &behaviorContent) override;
+
+private:
+  const EventsBasedBehavior &eventsBasedBehavior;
+  gd::SerializerElement objectContent;
```
4ian, line 39: objectContent?

Core/GDCore/Project/CustomBehavior.cpp
```diff
@@ -0,0 +1,102 @@
...
+bool CustomBehavior::UpdateProperty(gd::SerializerElement &behaviorContent,
+                                    const gd::String &propertyName,
+                                    const gd::String &newValue) {
+  const auto &properties = eventsBasedBehavior.GetPropertyDescriptors();
+  if (!properties.Has(propertyName)) {
+    return false;
+  }
+  const auto &property = properties.Get(propertyName);
+
+  auto &element = objectContent.AddChild(propertyName);
```
4ian, line 72: Should this be objectContent or behaviorContent?

Core/GDCore/Project/BehaviorConfigurationContainer.h
```diff
@@ -0,0 +1,155 @@
...
+class SerializerElement;
+class Project;
+class Layout;
+}  // namespace gd
+
+namespace gd {
+
+/**
+ * \brief Base class for containers of behavior configuration.
+ * They can be attached to objects (Behavior) or layouts (BehaviorsSharedData).
```
4ian, line 25: Thank you for this explanation, very useful when reading the code and trying to understand how this is used.

Core/GDCore/Project/CustomBehavior.h
4ian: Should this be removed? This should be handled by the parent class right?

Core/GDCore/Extensions/Metadata/BehaviorMetadata.cpp
```diff
@@ -42,6 +43,27 @@ BehaviorMetadata::BehaviorMetadata(
...
+        fullname_,
+        // Default name is the name
+        name_,
+        description_,
+        group_,
+        icon24x24_,
+        // Class name is the name, actually unused
+        name_,
+        // It is only used to get the name for GetName.
+        gd::make_unique<gd::Behavior>(),
```
D8H, line 64: It's not ideal as `Get()` will still returns something. We could maybe use a boolean member to return null. This way the class will have a clean interface.

Core/GDCore/Project/CustomBehavior.h
D8H: It will be an issue when a behavior is removed. On the released version, removing an extension and adding it back preserve the behaviors on objects.
We could use a Project and a fullType instead of an EventsBasedBehavior. It will work the same way as before but hidden inside the CustomBehavior.

Core/GDCore/Project/BehaviorConfigurationContainer.cpp
4ian: I'm not sure. Can you explain what is "this logic"?

Core/GDCore/Extensions/Metadata/ObjectMetadata.h
```diff
@@ -307,7 +317,9 @@ class GD_CORE_API ObjectMetadata {
   std::shared_ptr<gd::Object>
       blueprintObject;  ///< The "blueprint" object to be copied when a new
                         ///< object is asked. Can be null in case a creation
-                        ///< function is passed.
+                        ///< function is passed or for events based behaviors
+                        ///< as CustomBehavior are created using
```
4ian, line 321: Wrong comment?

Core/GDCore/IDE/WholeProjectRefactorer.cpp
```diff
@@ -1058,14 +1048,15 @@ bool WholeProjectRefactorer::FixInvalidRequiredBehaviorProperties(
     auto& object = problem.GetSourceObject();
     auto suggestedBehaviorNames =
         GetBehaviorsWithType(object, problem.GetExpectedBehaviorTypeName());
-    auto& behaviorContent = problem.GetSourceBehaviorContent();
+    auto& behavior = problem.GetSourceBehaviorContent();
+    
+    // TODO BC Be confident on the model and remove the metadata check?
```
4ian, line 1053: Are the metadata now useless? If so I guess we can remove the checks.

Core/GDCore/Project/CustomBehavior.h
4ian: I see! Might be safer. 
If we don't change anything (let the reference as you did now) and if we remove an extension, what would happen? Invalid reference => crash? These CustomBehavior are never recreated right, after being created once in objects?

Core/GDCore/Project/BehaviorConfigurationContainer.cpp
D8H: ObjectJsImplementation use something like this where it is done with overloading in C++ for behaviors and methods with the content parameter are exposed.
```JS
var newProperties = self['getProperties'](objectContent);
```
An overloading seems easier to understand, but maybe there is some constraints with JS forbidding it.

Core/GDCore/Project/CustomBehavior.h
D8H: Yes, it doesn't do crash right away but it freezes some interactions with the UI and make stacks all over the place.

Core/GDCore/Project/CustomObject.h
```diff
@@ -45,6 +43,6 @@ class CustomObject : public gd::Object {
   void DoUnserializeFrom(Project& project, const SerializerElement& element) override;
 
   private:
-    const EventsBasedObject& eventsBasedObject;
+    const Project& project;
```
4ian, line 46: Add a comment explaining why this is useful to keep the project here.

Core/GDCore/Project/CustomBehavior.h
```diff
@@ -0,0 +1,38 @@
...
+  virtual std::map<gd::String, gd::PropertyDescriptor>
+  GetProperties(const gd::SerializerElement &behaviorContent) const override;
+  virtual bool UpdateProperty(gd::SerializerElement &behaviorContent,
+                              const gd::String &name,
+                              const gd::String &value) override;
+  virtual void
+  InitializeContent(gd::SerializerElement &behaviorContent) override;
+
+private:
+  const Project &project;
```
4ian, line 37: Same as for CustomObject, add a comment explaining why it's useful to keep the project here.

Core/GDCore/Extensions/Metadata/BehaviorMetadata.cpp
4ian: Having a nullptr (which is what a Unique pointer to nothing is) is fine I guess for now?

Core/GDCore/Extensions/Metadata/BehaviorMetadata.h
```diff
@@ -246,12 +260,18 @@ class GD_CORE_API BehaviorMetadata {
 
   /**
    * \brief Return the associated gd::Behavior, handling behavior contents.
+   * 
+   * \note Returns a dub Behavior for events based behaviors as CustomBehavior
```
4ian, line 264: typo: dub?

Core/GDCore/Extensions/Metadata/ObjectMetadata.h
```diff
@@ -307,7 +317,8 @@ class GD_CORE_API ObjectMetadata {
   std::shared_ptr<gd::Object>
       blueprintObject;  ///< The "blueprint" object to be copied when a new
                         ///< object is asked. Can be null in case a creation
-                        ///< function is passed.
+                        ///< function is passed or for events based objects
+                        ///< as CustomObject are using EventBasedObject.
```
4ian, line 321:
```suggestion
                        ///< function is passed or for events based objects
                        ///<(CustomObject are using EventBasedObject, they don't need blueprints).

```

Core/GDCore/Project/CustomObject.cpp
```diff
@@ -22,8 +22,12 @@ std::unique_ptr<gd::Object> CustomObject::Clone() const {
 // This will be used by the GUI to display custom editors (for sprites for
 // instance)
 std::map<gd::String, gd::PropertyDescriptor> CustomObject::GetProperties() const {
-    const auto &properties = eventsBasedObject.GetPropertyDescriptors();
     auto objectProperties = std::map<gd::String, gd::PropertyDescriptor>();
+    if (!project.HasEventsBasedObject(GetType())) {
```
4ian, line 26: Nice! Certainly less performant, but unlikely to be a bottleneck and much safer in case we somehow end up in a situation where customobject continue to exist after their extensions are gone.

Core/GDCore/Project/CustomObject.h
4ian: (it's obvious now, but maybe not for the reader of the header file)

Core/GDCore/Project/EventsFunctionsExtension.cpp
```diff
@@ -86,6 +86,12 @@ void EventsFunctionsExtension::SerializeTo(SerializerElement& element) const {
 }
 
 void EventsFunctionsExtension::UnserializeFrom(
+    gd::Project& project, const SerializerElement& element) {
```
4ian, line 89: Nice, I think a UnserializeFrom should indeed do the entire job - always.
If you want to partially unserialize, then you use the other functions.

Core/GDCore/Project/EventsFunctionsExtension.h
```diff
@@ -216,10 +216,25 @@ class GD_CORE_API EventsFunctionsExtension : public EventsFunctionsContainer {
...
+  /**
+   * \brief Load the extension without free functions, behaviors and objects
+   * implementation.
+   */
+  void UnserializeExtensionDeclarationFrom(
+      gd::Project& project,
+      const gd::SerializerElement& element);
+
+  /**
+   * \brief Load free functions, behaviors and objects implementation.
```
4ian, line 233:
```suggestion
   * \brief Load free functions, behaviors and objects implementation (in opposition to load just their "declaration" by reading their name).
```

Core/GDCore/Project/EventsFunctionsExtension.cpp
```diff
@@ -138,11 +144,40 @@ void EventsFunctionsExtension::UnserializeFrom(
...
+
+void EventsFunctionsExtension::UnserializeExtensionImplementationFrom(
+    gd::Project& project,
+    const SerializerElement& element) { 
   UnserializeEventsFunctionsFrom(project, element.GetChild("eventsFunctions"));
-  eventsBasedBehaviors.UnserializeElementsFrom(
-      "eventsBasedBehavior", project, element.GetChild("eventsBasedBehaviors"));
-  eventsBasedObjects.UnserializeElementsFrom(
-      "eventsBasedObject", project, element.GetChild("eventsBasedObjects"));
+  auto &behaviorElement = element.GetChild("eventsBasedBehaviors");
```
4ian, line 171: There is a risk the lists were manipulated between the calls to UnserializeExtensionDeclarationFrom and UnserializeExtensionImplementationFrom.
I think this needs at least a check that the lists length are the same, and a check that every name is still the same.
Otherwise, this should probably error (bail out before unserializing if these verifications are not ok).

Core/GDCore/Project/EventsFunctionsExtension.cpp
```diff
@@ -138,11 +144,40 @@ void EventsFunctionsExtension::UnserializeFrom(
     dependencies.push_back(
         UnserializeDependencyFrom(dependenciesElement.GetChild(i)));
 
+  // Only unserialize behaviors and objects names.
+  // As event based objects can contains CustomObject and Custom Object,
```
4ian, line 148:
```suggestion
  // As event based objects can contains objects using CustomBehavior and/or CustomObject,
```

Core/GDCore/Extensions/Metadata/BehaviorMetadata.cpp
D8H: No, I forgot that it doesn't call the Behavior default constructor.

D8H:
I launched 3 games with the last version without any issue. Saving and loading a game project seems to work too.

Core/GDCore/Project/Object.cpp
```diff
@@ -58,46 +60,53 @@ bool Object::RenameBehavior(const gd::String& name, const gd::String& newName) {
...
+    const gd::BehaviorMetadata& behaviorMetadata =
+        gd::MetadataProvider::GetBehaviorMetadata(project.GetCurrentPlatform(),
+                                                  type);
+    if (gd::MetadataProvider::IsBadBehaviorMetadata(behaviorMetadata)) {
+      gd::LogWarning("Tried to create a behavior with an unknown type: " + type
+                     + " on object " + GetName() + "!");
+    // It's probably an Event based that was removed.
+    // Create a custom behavior to preserve the properties values.
+    return initializeAndAdd(
+        gd::make_unique<CustomBehavior>(name, project, type));
```
D8H, line 104: Do you think it's fine?

Core/tests/ProjectSerialization.cpp
```diff
@@ -0,0 +1,207 @@
...
+  REQUIRE(object.HasBehaviorNamed("MyEventsBasedBehavior"));
+
+  auto &behavior = object.GetBehavior("MyEventsBasedBehavior");
+  REQUIRE(behavior.GetTypeName() == "MyEventsExtension::MyEventsBasedBehavior");
+  REQUIRE(behavior.GetProperties().size() == 1);
+  REQUIRE(behavior.GetProperties().at("MyProperty").GetValue() == "481516");
+};
+} // namespace
+
+TEST_CASE("ProjectSerialization", "[common]") {
```
4ian, line 123: Very nice tests!

4ian (approved):

Core/GDCore/IDE/WholeProjectRefactorer.cpp
```diff
@@ -1058,21 +1037,15 @@ bool WholeProjectRefactorer::FixInvalidRequiredBehaviorProperties(
...
 
-    auto& behavior = behaviorMetadata.Get();
     if (suggestedBehaviorNames.empty()) {
       // No matching behavior on the object.
       // Add required behaviors on the object.
 
       auto& expectedBehaviorMetadata = MetadataProvider::GetBehaviorMetadata(
           project.GetCurrentPlatform(), problem.GetExpectedBehaviorTypeName());
-      if (MetadataProvider::IsBadBehaviorMetadata(behaviorMetadata)) {
+      if (MetadataProvider::IsBadBehaviorMetadata(expectedBehaviorMetadata)) {
```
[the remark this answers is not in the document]
D8H: Yes, I guess that if:
- a user add a required behavior "B" property to a behavior "A"
- remove the behavior "B"
- add the behavior "A" to an object
It could have crashed the editor.

It's an interesting case to add I'll add a todo for this.

Core/GDCore/Project/EventsFunctionsExtension.cpp
```diff
@@ -138,11 +144,48 @@ void EventsFunctionsExtension::UnserializeFrom(
     dependencies.push_back(
         UnserializeDependencyFrom(dependenciesElement.GetChild(i)));
 
+  // Only unserialize behaviors and objects names.
+  // As event based objects can contains objects using CustomBehavior and/or
```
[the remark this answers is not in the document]
D8H: When an object is created, the type is used to check if a custom object must be created or not. The event-based object needs to be declared in the project for this check.
Actually, it could work if we create a custom object by default when the type is unknown but it would be impossible to log a warning to differentiate it from the nominal case.

Core/GDCore/Project/Object.cpp
4ian: I think this is fine yes!

Core/GDCore/Project/Layout.cpp
```diff
@@ -387,8 +401,9 @@ void Layout::Init(const Layout& other) {
 
   behaviorsSharedData.clear();
   for (const auto& it : other.behaviorsSharedData) {
+    auto *sharedData = it.second.get();
     behaviorsSharedData[it.first] =
-        std::unique_ptr<gd::BehaviorContent>(it.second->Clone());
+        gd::make_unique<gd::BehaviorsSharedData>(*sharedData);
```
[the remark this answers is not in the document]
D8H: Indeed, I'll have to fix this.
I guess, the tests don't cover much shared data cases. An other todo to add 😅

Issue #4223, opened 2022-08-25
String and number object variables both have same name


BWPanda:
## Describe the bug
When using the Expression Editor to add an object variable to your expression, only one option shows up:

![image](https://user-images.githubusercontent.com/2385329/186548882-c1d0fbd8-76f3-4aa3-904d-2b27cab93663.png)

Whereas I'd expect there to be two options (one for strings and one for numbers (values)), just like for scene and global variables:

![image](https://user-images.githubusercontent.com/2385329/186549000-59788fc3-354c-49f6-839b-ba337ec52ad7.png)

This is confirmed by the fact that two separate object variables are listed when you do a search, it's just that both have the same name:

![image](https://user-images.githubusercontent.com/2385329/186549134-3333eaf8-19b2-4fa1-bd85-968f57da3b18.png)

The problem appears to be here, where both object variable expressions have the same name, so one must override the other:

https://github.com/4ian/GDevelop/blob/master/Core/GDCore/Extensions/Builtin/BaseObjectExtension.cpp#L1100-L1122
```
  obj.AddExpression("Variable",
                    _("Object variable"),
                    _("Value of an object variable"),
                    _("Variables"),
                    "res/actions/var.png")
      .AddParameter("object", _("Object"))
      .AddParameter("objectvar", _("Variable"));

...

  obj.AddStrExpression("VariableString",
                       _("Object variable"),
                       _("Text of an object variable"),
                       _("Variables"),
                       "res/actions/var.png")
      .AddParameter("object", _("Object"))
      .AddParameter("objectvar", _("Variable"));
```

See https://forum.gdevelop.io/t/object-variables-dont-have-value-string-versions/41744/4 for where this was discovered.

BWPanda:
Here's an untested PR: https://github.com/4ian/GDevelop/pull/4224

tristanbob:
@BWPanda thanks for finding this bug and for the fix.  The fix looks good to me, but I will wait for another dev to review and approve.

PR #4224, opened 2022-08-25
Issue #4223: Rename object variables expressions to differentiate them.

BWPanda:
Fixes https://github.com/4ian/GDevelop/issues/4223


tristanbob:
@BWPanda thanks for finding this bug and for the fix. The fix looks good to me, but I will wait for another dev to review and approve.

ClementPasteau (approved):
lgtm! thanks

PR #4225, opened 2022-08-25
Add support for uploading resources for cloud project

4ian:
This allows to upload resources:
- Resources are uploaded to the backend which gives an URL where to upload them. We hash the content to ensure integrity and other nice features in the backend.
- The option to use a resource from an URL was hidden by default behind an "advanced" toggle.
- Reworks the "save as" of the "StorageProviders" to call "onMoveResources", which is plugged by MainFrame to `ResourcesMover`. This helper does the job of looking at the old "StorageProvider", the new "StorageProvider", and call a function that does the move of the resources between these storage providers.
  - This means that this `ResourcesMover` has the knowledge of every storage provider (which is fine!). It works a bit as a matrix between storage providers. 
  - We have a similar `ResourcesFetcher` which allows storage providers to fetch some resources, like resolving urls (when loading a project from a URL) or download files locally (when saving/opening a project from a URL, like an example, into a local file, on the desktop app)
  - ~~This means it could also replace the old "ResourcesFetcher" which was used "only" to import URLs into local files (i.e: this is an operation that should be handled by `ResourcesMover` using the 
 `UrlStorageProvider`=>`LocalFileStorageProvider` function).~~
  - ~~It's very similar to ResourcesFetcher in the sense that it allows to report errors and retry if something bad happened. A good reason to try to remove ResourcesFetcher and use this ResourcesMover only.~~
- There were a few changes to allow PIXI and resources to be loaded with "credentials" (i.e: cookies) being sent. Both for the preview, the images shown in the editor and the export (and the save as).

At the end, for the end user, this looks like this:
- A choice to upload one or more ressources: image
- Resources already uploaded are de-duplicated.
- If you "Save As" to another cloud project, the resources are uploaded to the newly created project, so it just works.
- "Download a copy" will give you a zip file with the project json and resources, to be opened in the desktop app (and it should "just work").

Stories:
- [ResourcesMoverDialog](https://gdevelop-storybook.s3.amazonaws.com/feature/upload-cloud-project-resources/latest/index.html?path=/story/resourcemoverdialog--default)
- [FileToCloudProjectResourceUploader](https://gdevelop-storybook.s3.amazonaws.com/feature/upload-cloud-project-resources/latest/index.html?path=/story/resourceslist-filetocloudprojectresourceuploader--default)

Longer term (but maybe not that far):
- [In another PR] You can use Cloud projects in the desktop app .
  - This includes save from/to a local project.
- [In another PR] You can use the asset store for single files in the desktop app.

Still to do:
- [x] Partially replace ResourcesFetcher by ResourcesMover.
- [x] Rework of LocalCreation/BrowserCreation so that a new project made from an example is loaded by UrlStorageProvider, and then "saved as" immediately with your chosen storage provider. 
- [x] Remove/replace ResourcesFetcher.
- ~~Check to see if cloud projects would work "as is" on the desktop app.~~
  - No because of adaptations need for the export.
- ~~Check if it means we can add the "resource stores" to the ResourceSources on desktop.~~
  - Needs some adaptation to fetch the resources, some props to rework.
- [x] Verify if the PIXI upgrade is actually not needed.
- [x] Fix remaining TODOs
  - [x] Some tests?
  - [x] Organise assets and avoid conflicts in filenames in DownloadFile.
  - [x] Revert changes in newIDE/web-app/scripts/deploy-GDJS-Runtime.js and newIDE/app/src/GameEngineFinder/BrowserS3GDJSFinder.js
- [x] Update DownloadFileStorageProvider to download resources as local files (with well organised assets by types). Use the `useGenericRetryableProcessWithProgress`. 
- [x] Ensure the backend provides proper limitation to prevent abuses for user accounts/IPs.
- [x] Ensure Google Drive still works.

Manual test cases:
- Create blank project (browser and desktop app). Add a new resource. Save.
- Create project from example (browser and desktop app). [Save] Add a new resource. Save.
- Save as to a new project (browser and desktop app). Check resources are copied.
- Download File on the webapp, verify resources are there. Verify project still works.


Bouh:
This PR update PIXI from v6.1.2 to v6.5.1.
I think this is an important thing to mention.

4ian:
Yes, we might actually not even keep the upgrade or do it in a separate PR. I thought it was mandatory to get some resources to load properly, but I think I was mistaken.

newIDE/app/src/Utils/GDevelopServices/Project.spec.js
```diff
@@ -0,0 +1,102 @@
...
+      );
+      expect(
+        extractFilenameFromProjectResourceUrl(
+          'https://project-resources.gdevelop.io/whatever/some/other-folder/1/2/_/6ef87bc678921eb4bfa2d04e5dc6a16b75f7b239f3163d0c5efe64d4cc501711-Pea-Happy.png?some=parameter'
+        )
+      ).toBe(
+        '6ef87bc678921eb4bfa2d04e5dc6a16b75f7b239f3163d0c5efe64d4cc501711-Pea-Happy.png?some=parameter'
+      );
+    });
+    it('extracts filename from a classic URL', () => {
```
AlexandreSi, line 75: I don't feel like the test name really explains what is tested in it.

newIDE/app/src/Utils/GDevelopServices/Project.spec.js
```diff
@@ -0,0 +1,102 @@
...
+          'https://project-resources.gdevelop.io/a9fe5bce-de39-4147-a669-93fc5cd69632/resources/some/other-folder/1/2/_/6ef87bc678921eb4bfa2d04e5dc6a16b75f7b239f3163d0c5efe64d4cc501711-Pea-Happy?some=parameter'
+        )
+      ).toBe('Pea-Happy');
+      expect(
+        extractFilenameFromProjectResourceUrl(
+          'https://project-resources.gdevelop.io/a9fe5bce-de39-4147-a669-93fc5cd69632/resources/some/other-folder/1/2/_/6ef87bc678921eb4bfa2d04e5dc6a16b75f7b239f3163d0c5efe64d4cc501711-Pea-Happy.png?some=parameter'
+        )
+      ).toBe('Pea-Happy.png');
+    });
+    it('extracts filename from a classic URL', () => {
```
AlexandreSi, line 53: Same as below. Plus I don't know why the expected value contains the query parameters and the hash

newIDE/app/src/UI/CorsAwareImage.js
```diff
@@ -38,7 +38,6 @@ const addSearchParameterToUrl = (
 export const CorsAwareImage = (props: Props) => (
   <img // eslint-disable-line jsx-a11y/alt-text
     {...props}
-    crossOrigin="anonymous"
```
AlexandreSi, line 41: I seem to remember you explained to me why this had to be changed but I'm not sure I remember why.

newIDE/app/src/Utils/BrowserArchiver.js
```diff
@@ -57,80 +44,45 @@ function eachCallback<T>(
...
-        );
+  const downloadedBlobs: Array<
+    ItemResult<UrlFileDescriptor>
+  > = await downloadUrlsToBlobs({
+    urlContainers: urlFiles.filter(({ url }) => url.indexOf('.h') === -1), // Should be useless now, still keep it by safety.
+    onProgress,
+  });
 
-        return fetch(urlWithParameters)
-          .then(
```
AlexandreSi, line 97: 🔥 `then` 🤗

newIDE/app/src/MainFrame/EditorContainers/HomePage/index.js
```diff
@@ -72,6 +72,8 @@ export const HomePage = React.memo<Props>(
         onOpenRecentFile,
         onCreateFromExampleShortHeader,
         onCreateBlank,
+
+        // Unused?
```
AlexandreSi, line 76: Might have changed with homepage changes by @ClementPasteau

newIDE/app/src/MainFrame/index.js
```diff
@@ -741,15 +752,14 @@ const MainFrame = (props: Props) => {
...
-
-          return loadFromProject(newProject, fileMetadata);
-        },
-        time => console.info(`Unserialization took ${time} ms`)
-      );
+      const startTime = Date.now();
+      const newProject = gd.ProjectHelper.createNewGDJSProject();
+      newProject.unserializeFrom(serializedProject);
+      const endTime = Date.now();
+      const duration = endTime - startTime;
```
AlexandreSi, line 759:
```suggestion
      const duration =  Date.now() - startTime;
```

newIDE/app/src/MainFrame/index.js
```diff
@@ -1818,14 +1828,20 @@ const MainFrame = (props: Props) => {
...
       saveUiSettings(state.editorTabs);
 
-      const storageProviderOperations = getStorageProviderOperations(
+      // Remember the old storage provider, as we may need to use it to get access
+      // to resources.
+      const oldStorageProvider = getStorageProvider();
+      const oldStorageProviderOperations = getStorageProviderOperations();
+
+      // Get the methods to save the project using the *new* storage provider.
+      const newStorageProvider = getStorageProvider();
```
AlexandreSi, line 1837: I think you need to define `newStorageProvider` after calling the new storage provider operations since it's the second instruction that changes the `currentStorageProvider` in `ProjectStorageProviders`

newIDE/app/src/ObjectsRendering/PixiResourcesLoader.js
```diff
@@ -146,18 +167,18 @@ export default class PixiResourcesLoader {
...
-      }),
-      {
-        scaleMode: PIXI.SCALE_MODES.LINEAR,
-        resourceOptions: {
-          autoPlay: false,
-        },
-      }
-    );
+    const url = ResourcesLoader.getResourceFullUrl(project, resourceName, {
+      disableCacheBurst: true, // Disable cache bursting for video because it prevent the video to be recognized as such (for a local file)
```
AlexandreSi, line 171:
```suggestion
      disableCacheBurst: true, // Disable cache bursting for video because it prevents the video to be recognized as such (for a local file).
```

newIDE/app/src/ProjectsStorage/ResourceMover/BrowserResourceMover.js
```diff
@@ -0,0 +1,234 @@
...
+          if (isURL(resourceFile)) {
+            if (checkIfIsGDevelopCloudBucketUrl(resourceFile)) {
+              if (
+                extractProjectUuidFromProjetResourceUrl(resourceFile) ===
+                newCloudProjectId
+              ) {
+                // Somehow the resource is *already* stored in the new project - surely because
+                // the project resources were partially moved (like when you click "Retry" after some failures
+                // when saving a project as a new cloud project).
+                // Just ignore this resource which is already moved then.
```
AlexandreSi, line 82: 👌 neat comment

newIDE/app/src/ProjectsStorage/ResourceMover/BrowserResourceMover.js
```diff
@@ -0,0 +1,234 @@
...
+
+      return {
+        resource: item.resource,
+        file: new File([blob], item.filename, { type: blob.type }),
+      };
+    })
+    .filter(Boolean);
+
+  // Upload the files just downloaded, for the new project.
+  await getCredentialsForCloudProject(authenticatedUser, newCloudProjectId);
```
AlexandreSi, line 147: Can `onEnsureCanAccessResources` be used here?

newIDE/app/src/ProjectsStorage/ResourceMover/LocalResourceMover.js
```diff
@@ -0,0 +1,69 @@
...
+    );
+    return {
+      erroredResources: [],
+    };
+  },
+  [`${UrlStorageProvider.internalName}=>${
+    LocalFileStorageProvider.internalName
+  }`]: (options: MoveAllProjectResourcesOptions) => {
+    // TODO: Currently unused.
+    // Download resources. Check if this ResourceFetcher can be moved here.
```
AlexandreSi, line 42: Could be great indeed

newIDE/app/src/ProjectsStorage/ResourceMover/index.js
```diff
@@ -0,0 +1,276 @@
...
+  genericError,
+}: ResourceMoverDialogProps) => {
+  const hasErrors =
+    (result && result.erroredResources.length > 0) || !!genericError;
+
+  return (
+    <Dialog
+      actions={[
+        onRetry ? (
+          <DialogPrimaryButton
```
AlexandreSi, line 82: I think the retry button should be on the right side.
One might be tempted to ignore without reading the dialog:
image

newIDE/app/src/ProjectsStorage/GoogleDriveStorageProvider/index.js
```diff
@@ -404,48 +401,51 @@ export default ({
...
-                      })
-                    )
+                  const newFileId = await createNewJsonFile(
+                    selectedFileOrFolder.id,
+                    newFileName
                   );
+                  const newFileMetadata = {
+                    fileIdentifier: newFileId,
+                  };
+                  await options.onMoveResources({ newFileMetadata });
```
AlexandreSi, line 418: I didn't realize moving project from GoogleDrive to Cloud would be available! So nice

newIDE/app/src/ResourcesList/FileToCloudProjectResourceUploader.js
```diff
@@ -0,0 +1,206 @@
...
+          error: 'too-large',
+        };
+      }
+      return null;
+    })
+    .filter(Boolean);
+
+  const canUploadWithThisStorageProvider =
+    storageProvider.internalName === 'Cloud';
+  const isConnected = !!authenticatedUser.profile;
```
AlexandreSi, line 123: `authenticatedUser.authenticated` doesn't work for this?

newIDE/app/src/ResourcesList/FileToCloudProjectResourceUploader.js
```diff
@@ -0,0 +1,206 @@
...
+            />
+          </Column>
+        </Line>
+      </Paper>
+      {invalidFiles.map(erroredFile => {
+        if (erroredFile.error === 'too-large')
+          return (
+            <AlertMessage kind="error">
+              <Trans>
+                The file {erroredFile.filename} is too large. Use files that are
```
AlexandreSi, line 175: Should we specify the max size? Could be a bit frustrating

newIDE/app/src/ResourcesList/FileToCloudProjectResourceUploader.js
```diff
@@ -0,0 +1,206 @@
...
+          </AlertMessage>
+        );
+      })}
+      <LineStackLayout alignItems="center" justifyContent="flex-end" expand>
+        {isUploading ? (
+          <LinearProgress expand value={uploadProgress} variant="determinate" />
+        ) : null}
+        <RaisedButton
+          onClick={onUpload}
+          disabled={!canUploadFiles}
```
AlexandreSi, line 193: Should this be disabled when uploading as well?

newIDE/app/src/ResourcesList/FileToCloudProjectResourceUploader.js
```diff
@@ -0,0 +1,206 @@
...
+      <Paper variant="outlined">
+        <Line expand>
+          <Column expand>
+            <input
+              accept={resourceKindToInputAcceptedFiles[options.resourceKind]}
+              style={styles.fileInput}
+              multiple={options.multiSelection}
+              type="file"
+              ref={inputRef}
+              disabled={!canChooseFiles}
```
AlexandreSi, line 158: Same here, should it be disabled when uplaoding?

newIDE/app/src/MainFrame/index.js
4ian: Good catch, thanks!

newIDE/app/src/ProjectsStorage/ResourceMover/index.js
4ian: Fixed :) And factored for usage in other places.

newIDE/app/src/ResourcesList/FileToCloudProjectResourceUploader.js
4ian: It does work, I just forgot about this! Thanks.

newIDE/app/src/ResourcesList/FileToCloudProjectResourceUploader.js
4ian: It is ;)
`const canChooseFiles = !isUploading && isConnected && canUploadWithThisStorageProvider`

newIDE/app/src/ResourcesList/FileToCloudProjectResourceUploader.js
4ian: It is :)

newIDE/app/src/ProjectsStorage/GoogleDriveStorageProvider/index.js
4ian: This is a actually "easy" in this direction because no custom resources were allowed, just public URLs, so there is nothing to move 😅
The other direction, similarly, will only work if you have public URLs. If you use a resource uploaded to GDevelop cloud, you will get an error for this/these resource(s) (that you can ignore).

In practice, we'll just strongly recommend to use GDevelop cloud. Download a copy will allow you to download the game including the resources you uploaded, so good to do a backup. (to import it back, we either need a specific storage provider to open from a zip file, or we need to ask to go through the desktop app).

newIDE/app/src/ResourcesList/FileToCloudProjectResourceUploader.js
```diff
@@ -0,0 +1,206 @@
...
+    canChooseFiles &&
+    hasSelectedFiles &&
+    invalidFiles.length === 0;
+
+  return (
+    <ColumnStackLayout noMargin>
+      {!isConnected ? (
+        <AlertMessage kind="warning">
+          <Trans>
+            Your need to first create your account, or login, to upload your own
```
Bouh, line 137: Typo 1

newIDE/app/src/ResourcesList/FileToCloudProjectResourceUploader.js
```diff
@@ -0,0 +1,206 @@
...
+            resources.
+          </Trans>
+        </AlertMessage>
+      ) : !canUploadWithThisStorageProvider ? (
+        <AlertMessage kind="warning">
+          <Trans>
+            Your need to first save your game on GDevelop Cloud to upload your
```
Bouh, line 144: Typo 2

newIDE/app/src/ResourcesList/FileToCloudProjectResourceUploader.js
4ian: Good idea, done!

newIDE/app/src/UI/CorsAwareImage.js
4ian: Anonymous was not sending the cookies when fetching images, which we need now (when it's an image hosted on a gdevelop.io bucket, the cookie will be sent)

4ian:
All checks should pass, I've addressed comments and all use cases should work. Notably:
- Download as a file in the web-app now gives an archive with resources.
- You can save from a cloud project to a new cloud project, and resources will be copied.
- You can upload your own resources anywhere a resource is asked, or just open the resources editor and start replacing existing resources.

This should open the room to more good stuff in the future, like having cloud projects on the desktop app, with the ability to save from/to a local file to/from a cloud project.

PR #4227, opened 2022-08-25
Add a property editor for custom objects

D8H:
Custom objects configurations are composed of configurations according to the events-based object children.

In order to allow object configuration composition, the `Object` class is no longer derivated. It has a `ObjectConfiguration` that is an extracted interface for object to handle their configurations.

The entry point of this PR is `CustomObjectPropertiesEditor.js`.

### Manual test
* Check that every object types can be added to the scene (the last 3 are video and tile maps without resources).

image


Core/GDCore/Extensions/Builtin/SpriteExtension/SpriteObject.cpp
```diff
@@ -25,8 +25,7 @@ namespace gd {
 
 Animation SpriteObject::badAnimation;
 
-SpriteObject::SpriteObject(gd::String name_)
-    : Object(name_), updateIfNotVisible(false) {}
+SpriteObject::SpriteObject() : updateIfNotVisible(false) {}
```
D8H, line 28: I kept the classes names, but they should probably be renamed. Maybe in another PR?

newIDE/app/src/ObjectEditor/ObjectsEditorService.js
```diff
@@ -1,3 +1,4 @@
+// @flow
```
D8H, line 1: I added some typing here to avoid mistakes.
Is there other files related to object editors where flow should be added?

newIDE/app/src/ObjectEditor/ObjectsEditorService.js
```diff
@@ -16,16 +19,26 @@ const ObjectsEditorService = {
...
-      console.warn(
-        `Object with type ${objectType} has no editor configuration registered. Please use registerEditorConfiguration to register your editor.`
+    if (project.hasEventsBasedObject(objectType)) {
+      const objectMetadata = gd.MetadataProvider.getObjectMetadata(
+        gd.JsPlatform.get(),
+        objectType
       );
+      return this.getCustomObjectPropertiesEditor({
+        helpPagePath: objectMetadata.getHelpPath(),
+      });
```
D8H, line 29: Do you think that the events-based object generation should register instead.

Core/GDCore/Extensions/Builtin/SpriteExtension/SpriteObject.cpp
4ian: Probably in another PR, after we've proven this PR does not introduce any regression :)

newIDE/app/src/ObjectEditor/ObjectsEditorService.js
4ian: That's a good question. What would be the pros/cons of each?

D8H: It makes sense for extensions from Core and GDJS as them will stay the same for the process life-time.
But, as this list is static, we probably don't want to pollute it with extensions from projects.

4ian: Yeah, probably we don't want to pollute it. The list exists because extensions from Core and GDJS are fixed. But if we can be dynamic, no reason to make things constrained/harder to use with a list.

newIDE/app/src/ObjectEditor/ObjectsEditorService.js
4ian: Nothing in mind, that's good already!

4ian (commented):
That's pretty huge but I've seen nothing suspicious. Will probably need another pass just to make sure it makes sense after re-reading it again.

Core/GDCore/Project/Object.h
```diff
@@ -70,6 +72,13 @@ class GD_CORE_API Object {
     return gd::make_unique<gd::Object>(*this);
   }
 
+  /**
+   * \brief Return the object configuration.
+   */
+  gd::ObjectConfiguration& GetConfiguration();
+
+  gd::ObjectConfiguration& GetConfiguration() const;
```
4ian, line 80: This should return a `const gd::ObjectConfiguration&`:
```suggestion
  const gd::ObjectConfiguration& GetConfiguration() const;

```

Core/GDCore/Project/ObjectConfiguration.h
```diff
@@ -0,0 +1,189 @@
...
+class ArbitraryResourceWorker;
+class InitialInstance;
+class SerializerElement;
+class EffectsContainer;
+}  // namespace gd
+
+namespace gd {
+
+/**
+ * \brief Base class used to represent an object configuration
```
4ian, line 31:
```suggestion
 * \brief Base class used to represent an object configuration.
 * For example, this can be the animations in a sprite, the text, its font, its color in a Text object, etc...
```

Core/GDCore/Project/ObjectConfiguration.h
```diff
@@ -0,0 +1,189 @@
...
+   */
+  ObjectConfiguration();
+
+  /**
+   * Destructor.
+   */
+  virtual ~ObjectConfiguration();
+
+  /**
+   * Must return a pointer to a copy of the object. A such method is needed to
```
4ian, line 48:
```suggestion
   * Must return a pointer to a copy of the object configuration. This method is needed to
```

Core/GDCore/Project/ObjectConfiguration.h
```diff
@@ -0,0 +1,189 @@
...
+  /**
+   * \brief Derived objects can redefine this method to save custom attributes.
+   */
+  virtual void DoSerializeTo(SerializerElement& element) const {};
+};
+
+}  // namespace gd
+
+/**
+ * Objects are usually managed thanks to (smart) pointers.
```
4ian, line 185:
```suggestion
 * Object configurations are usually managed thanks to (smart) pointers.
```

But if this is used nowhere in the codebase, delete this.

Core/GDCore/Project/ObjectConfiguration.h
```diff
@@ -0,0 +1,189 @@
...
+  /**
+   * \brief Unserialize the object.
+   * \see DoUnserializeFrom
+   */
+  void UnserializeFrom(gd::Project& project, const SerializerElement& element);
+  ///@}
+
+ protected:
+  gd::String type;  ///< Which type is the object. ( To test if we can do
+                    ///< something reserved to some objects with it )
```
4ian, line 168:
```suggestion
  gd::String type;  ///< Which type of object is represented by this configuration.
```

Core/GDCore/Project/ObjectConfiguration.h
```diff
@@ -0,0 +1,189 @@
...
+///@{
+  /**
+   * \brief Serialize the object.
+   * \see DoSerializeTo
+   */
+  void SerializeTo(SerializerElement& element) const;
+
+  /**
+   * \brief Unserialize the object.
+   * \see DoUnserializeFrom
```
4ian, line 161:
```suggestion
 * Members functions related to serialization of the object configuration.
 */
///@{
  /**
   * \brief Serialize the object configuration.
   * \see DoSerializeTo
   */
  void SerializeTo(SerializerElement& element) const;

  /**
   * \brief Unserialize the object configuration.
   * \see DoUnserializeFrom
```

Core/GDCore/Project/Project.h
```diff
@@ -464,6 +465,13 @@ class GD_CORE_API Project : public ObjectsContainer {
   std::unique_ptr<gd::Object> CreateObject(const gd::String& type,
                                            const gd::String& name) const;
 
+  /**
+   * Create an object configuration of the given type with the specified name.
```
4ian, line 469:
```suggestion
   * Create an object configuration of the given type.

```

Core/GDCore/Project/Project.h
```diff
@@ -464,6 +465,13 @@ class GD_CORE_API Project : public ObjectsContainer {
   std::unique_ptr<gd::Object> CreateObject(const gd::String& type,
                                            const gd::String& name) const;
 
+  /**
+   * Create an object configuration of the given type with the specified name.
+   * 
+   * \param type The type of the object
```
4ian, line 471:
```suggestion
   * \param type The type of the object configuration.

```

4ian (approved):

Core/GDCore/Project/Project.h
D8H: It's actually the type of the object of the configuration.

D8H:
I tested it on the Platformer, Spherez and Absorbus examples.
I saved, reopened and previewed them with this branch and opened and previewed the saved project on the current release.

PR #4228, opened 2022-08-25
Change points image for div with background image to fix dragging issue

AlexandreSi:
To fix https://forum.gdevelop.io/t/points-deselect-is-not-working/41679
Also improved a bit padding to display the textfields:

**before**
Mobile
image
Desktop
image

**after**
Mobile
image
Desktop
image


4ian (approved):
Let's add a quick comment "Use a div so that when dragged, focus is moved here (and any text field is blurred)"
Also double check it works with touch emulation, just to be sure we're not breaking this on touch screens :)

AlexandreSi:
> Let's add a quick comment "Use a div so that when dragged, focus is moved here (and any text field is blurred)" Also double check it works with touch emulation, just to be sure we're not breaking this on touch screens :)

I added a different comment because this bug was due to a bad fix for the ghost image with Safari and Firefox

4ian (approved):

4ian:
Nice work on the padding!

PR #4238, opened 2022-08-28
Remove useless "Change" prefix in actions full name.

D8H:
I think it important to avoid starting actions full name by "Change" because:
- it doesn't add any information
- make it harder to read and especially to scan because the name is to long and the uppercase letter is not relevant
- It also open the dor to names that looks like sentences which is really bad for scanning.

I've found my self in situations where my review advices were countered by examples from built-in actions which make me lose time arguing. So, I think we should set the example.

Related extensions PR:
- https://github.com/GDevelopApp/GDevelop-extensions/pull/546
- https://github.com/GDevelopApp/GDevelop-extensions/pull/547


4ian (approved):
I agree these were bad namings. Thanks for fixing!

PR #4244, opened 2022-08-29
Add button in changelog dialog to read on Github to benefit from browser translator

AlexandreSi:
To answer:
image

image

I added a button to the ChangelogDialog because it was straightforward.
It could be added to the AboutDialog but we have to decide where to put it, it's less evident.

### DIFF
--- a/newIDE/app/src/MainFrame/Changelog/ChangelogDialog.js
+++ b/newIDE/app/src/MainFrame/Changelog/ChangelogDialog.js
@@ -7,8 +7,8 @@ import FlatButton from '../../UI/FlatButton';
 import Changelog from '.';
 import Text from '../../UI/Text';
 import useForceUpdate from '../../Utils/UseForceUpdate';
-import { Line } from '../../UI/Grid';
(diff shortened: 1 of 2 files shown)


newIDE/app/src/MainFrame/Changelog/ChangelogDialog.js
```diff
@@ -35,12 +40,19 @@ const ChangelogDialog = ({ open, onClose }: Props) => {
...
       open={open}
       onRequestClose={onClose}
+      flexColumnBody
     >
-      <Text>
-        <Trans>
-          GDevelop was upgraded to a new version! Check out the changes.
-        </Trans>
-      </Text>
+      <Line noMargin justifyContent="space-between">
```
ClementPasteau, line 45: Should this be a ResonsiveLineStackLayout so that the button goes below on small screens?

AlexandreSi: Good spot!

PR #4251, opened 2022-08-30
Add an instance render for custom objects

D8H:
As there is no instance editor for event-based object yet, it stacks one instance of each child object and centered them. It will give some what good result for an object button.

The entry point is `RenderedCustomObjectInstance`. The other real change is in `ObjectsRenderingService.createNewInstanceRenderer`. Other changes are only type definition.

### DIFF
--- a/newIDE/app/src/ObjectsRendering/ObjectsRenderingService.js
+++ b/newIDE/app/src/ObjectsRendering/ObjectsRenderingService.js
@@ -44,8 +44,7 @@ const ObjectsRenderingService = {
     'ParticleSystem::ParticleEmitter': RenderedParticleEmitterInstance,
   },
   getThumbnail: function(project: gdProject, object: gdObject) {
-    var objectType = object.getType();
-    console.log(objectType);
+    const objectType = object.getType();
     if (this.renderers.hasOwnProperty(objectType))
       return this.renderers[objectType].getThumbnail(
         project,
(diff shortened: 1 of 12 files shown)


newIDE/app/src/ObjectsRendering/ObjectsRenderingService.js
```diff
@@ -74,9 +75,21 @@ const ObjectsRenderingService = {
         associatedObjectConfiguration,
         pixiContainer,
         PixiResourcesLoader,
+        // This parameter is only used by RenderedTileMapInstance.
         pixiRenderer
```
D8H, line 79: I'm not sure if we should generalize it.

newIDE/app/src/ObjectsRendering/ObjectsRenderingService.js
```diff
@@ -74,9 +75,21 @@ const ObjectsRenderingService = {
         associatedObjectConfiguration,
         pixiContainer,
         PixiResourcesLoader,
+        // This parameter is only used by RenderedTileMapInstance.
```
4ian, line 78: Can you comment why it's used for?

newIDE/app/src/ObjectsRendering/Renderers/RenderedCustomObjectInstance.js
```diff
@@ -0,0 +1,251 @@
...
+    return [];
+  }
+
+  serializeTo(element: gdSerializerElement) {}
+
+  unserializeFrom(element: gdSerializerElement) {}
+}
+
+/**
+ * Renderer for gd.SpriteObject
```
4ian, line 125: Comment to update

newIDE/app/src/ObjectsRendering/ObjectsRenderingService.js
D8H: I've checked and the tile map extension doesn't actually need it.
- https://github.com/4ian/GDevelop/pull/4260

4ian (approved):
lgtm!

D8H:
Everything seems fine.
image

Issue #4245, opened 2022-08-30
Piskel falsely claims file is saved when using keyboard shortcut


lostfictions:
## Describe the bug

When adding a sprite and editing it with Piskel, pressing `Ctrl+S` yields a notification that the sprite was successfully saved. However, on closing the Piskel modal, all changes are lost.

## To Reproduce

1. Create a new blank project
2. Right click the scene backdrop and pick "Insert new..."
3. Pick "Sprite" from the menu. The Properties pane for the new sprite is shown. Click the "Add an animation" button. Optionally, select an existing image.
4. Click "Edit with Piskel" to begin drawing the animation. Add multiple frames if you'd like.
5. Hit Ctrl-S. A "Successfully saved!" toast is shown in the bottom-right corner of the Piskel editor.
   
   ![image](https://user-images.githubusercontent.com/567041/187351731-074d852a-c5f4-4a0b-b73c-ec0b5f61c276.png)
6. Close the Piskel window. No confirmation about closing with unsaved changes is shown.
7. If you started from scratch, the animation shown in the Properties plane is blank. If you started from an existing sprite, it will be shown in its original state before you made edits in Piskel. In both cases, the animation was not actually saved.
 
   ![image](https://user-images.githubusercontent.com/567041/187351911-36a0fdff-f989-478a-b34d-b3996d73211a.png)

---

To actually save the sprite, it looks like you have to open the "..." menu in the top right corner and select "Save". This immediately dismisses the editor. The workflow here is extremely confusing -- I'm really not sure why `Ctrl+S` claims to work but doesn't actually do anything.

![image](https://user-images.githubusercontent.com/567041/187352878-391b6ab2-5e82-4fb0-88fd-69851b530db0.png)

![image](https://user-images.githubusercontent.com/567041/187352921-a85ace7f-c657-4899-bded-d9ad1cf168f5.png)

## Other details

* OS: Linux, using GDevelop Flatpak 5.0.140
* Project file type is set to "Multiple files", since I saw this was recommended for collaboration/source control. I'm unsure whether this makes a difference.

AlexandreSi:
Hi, thanks for reporting this!
I'm currently following your steps and I wonder what action you actually do to "Close the Piskel window".
On my end, to close it, I either have to click save or cancel so I cannot leave Piskel with unsaved changes.

Either way, this Ctrl+S false shortcut should be removed.

AlexandreSi:
[...]

PR #4252, opened 2022-08-31
Attempt to remove modal window frame to hide close button on linux

AlexandreSi:
Workaround for #4245.
Not sure it will work (not using Linux) so we'll wait for next release.

### DIFF
--- a/newIDE/electron-app/app/ModalWindow.js
+++ b/newIDE/electron-app/app/ModalWindow.js
@@ -36,6 +36,10 @@ const loadModalWindow = ({
     backgroundColor,
     modal: true,
     center: true,
+    // Hide top bar to hide close button (needed on Linux) so that
+    // one cannot close the modal window without using Save or Cancel
+    // buttons (and lose work on Piskel for instance) (Workaround for #4245).
+    frame: false,
     webPreferences: {
       webSecurity: false,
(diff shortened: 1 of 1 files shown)


newIDE/electron-app/app/ModalWindow.js
```diff
@@ -36,6 +36,10 @@ const loadModalWindow = ({
     backgroundColor,
     modal: true,
     center: true,
+    // Hide top bar to hide close button (needed on Linux) so that
+    // one cannot close the modal window without using Save or Cancel
+    // buttons (and lose work on Piskel for instance) (Workaround for #4245).
```
ClementPasteau, line 41: The ModalWindow is also used for yarn and jfxr.
Are we sure this removal of the top bar doesn't affect them?

AlexandreSi: It will affect them and for the same logic: we don't want to exit those modal windows without using the `Save` or `Cancel` buttons

ClementPasteau: If we're happy it doesn't prevent closing the window on all 3 tools, then ok 👍

4ian (approved):
Would be good to check on Windows to see if this prevent moving the window around or maximizing it though. Or maybe just apply this on Linux?

PR #4253, opened 2022-08-31
Remove theme filter on checkered background for SVG display in resource store

AlexandreSi:
Theme filter not applied on SVG resource store

image

But still applied for Image resource store

image

### DIFF
--- a/newIDE/app/src/AssetStore/ResourceStore/ResourceCard.js
+++ b/newIDE/app/src/AssetStore/ResourceStore/ResourceCard.js
@@ -59,6 +59,7 @@ type ImageCardProps = {|
   size: number,
   resource: Resource,
   onChoose: () => void,
+  removeThemeFilterAppliedOnBackground?: boolean,
   imageStyle?: {|
     width: number,
     height: number,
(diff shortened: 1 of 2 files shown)


newIDE/app/src/AssetStore/ResourceStore/ResourceCard.js
```diff
@@ -148,6 +158,7 @@ export const ResourceCard = ({ resource, onChoose, size }: Props) => {
           onChoose={onChoose}
           size={size}
           imageStyle={styles.previewIcon}
+          removeThemeFilterAppliedOnBackground // All svg are black and hard to see on background darken by the theme filter so we prefer using the raw checkered background that is quite light #nofilter
```
ClementPasteau, line 161: I'm wondering if this is a bit too quick as a fix, and if we should rather fix the original problem which is: svgs should have a color that is the opposite of the theme, so that they're correctly visible in the store but also everywhere they're used.
(this change does not fix the fact that they're not visible elsewhere in dark theme)

AlexandreSi: I think it would be quite a consequent job to do and could have impact on the performance.
I felt like having this as a fix to make sure they would be visible in the meantime was a good idea (only a few changes for a painpoint that is not a current priority).

newIDE/app/src/ResourcesList/CheckeredBackground.js
```diff
@@ -3,7 +3,7 @@ import * as React from 'react';
 import GDevelopThemeContext from '../UI/Theme/ThemeContext';
 
 type Props = {|
-  style?: Object,
+  style?: {| filter?: 'unset' |},
```
4ian, line 6: Very nice!

4ian (approved):
Lgtm then!

PR #4268, opened 2022-09-05
Expose a gdObject constructor for the asset script.

D8H:
It breaks the encapsulation of Core but it's easier than requiring the asset script to setup a Platform.


Core/GDCore/Project/Object.cpp
```diff
@@ -20,11 +20,20 @@ namespace gd {
...
     : name(name_), configuration(std::move(configuration_)) {
       SetType(type_);
     }
 
+Object::Object(const gd::String& name_,
+               const gd::String& type_,
+               gd::ObjectConfiguration& configuration_)
+    : name(name_), configuration(&configuration_) {
+      SetType(type_);
+    }
```
D8H, line 35: Would it be safer to make a clone of the configuration?
It would force the asset script to destroy the given configuration thought.

Core/GDCore/Project/Object.cpp
```diff
@@ -20,11 +20,20 @@ namespace gd {
...
+Object::Object(const gd::String& name_,
+               const gd::String& type_,
+               std::unique_ptr<gd::ObjectConfiguration> configuration_)
     : name(name_), configuration(std::move(configuration_)) {
       SetType(type_);
     }
 
+Object::Object(const gd::String& name_,
+               const gd::String& type_,
+               gd::ObjectConfiguration& configuration_)
```
4ian, line 32: I think it's super risky to create a configuration from a point to a reference, because it means the unique_ptr will take ownership of the lifetime of the reference. But there is a risk someone would use this by passing a ObjectConfiguration created on the stack:

```
{
  gd::ObjectConfiguration configuration; // Unlikely, but as you can see, easy to write
  gd::Object object("name", "sometype", configuration);
}
// Boom, configuration was released because out of scope and the unique_ptr is pointing at something deleted.
```

Instead, take a pointer (`gd::ObjectConfiguration*`) to show that this is a raw pointer, so the reader must check who is taking ownership of it (answer being in the documentation generally).

Core/GDCore/Project/Object.h
```diff
@@ -40,7 +40,17 @@ class GD_CORE_API Object {
...
+         const gd::String& type,
+         std::unique_ptr<gd::ObjectConfiguration> configuration);
+
+  /**
+   * Create a new object with the name passed as argument.
+   * \param name Object's name
+   */
+  Object(const gd::String& name,
+         const gd::String& type,
+         gd::ObjectConfiguration& configuration);
```
4ian, line 53: Pass a parameter and tell that "the pointer ownership is taken by the object" in the parameter descriptions

Core/GDCore/Project/Object.h
```diff
@@ -40,7 +40,17 @@ class GD_CORE_API Object {
    * Create a new object with the name passed as argument.
    * \param name Object's name
    */
-  Object(const gd::String& name, const gd::String& type, std::unique_ptr<gd::ObjectConfiguration> configuration);
+  Object(const gd::String& name,
```
4ian, line 43: param descriptions are missing/out of date

4ian (approved):
I know this is only for usage in JS (where pointer or references don't matter, because every object created in the webassembly library by JS is never freed automatically) but you still must use a pointer to avoid usage in C++ that would crash because of a double free (because of double ownership).

If you really wanted to keep a reference, then you would have to clone the object configuration (so there reference ownership is kept by the parent).

PR #4273, opened 2022-09-08
Add announcements feed in community tab

4ian:
* As well as a way to display a urgent message on the homepage (that you can dismiss)

image
image


AlexandreSi:
image

I feel like the line should take all the available space

newIDE/app/src/MainFrame/EditorContainers/HomePage/index.js
```diff
@@ -167,49 +170,57 @@ export const HomePage = React.memo<Props>(
...
-                      }
-                      onTabChange={setActiveTab}
-                      onOpenHelpFinder={onOpenHelpFinder}
-                    />
-                  )}
-                  {activeTab === 'play' && <PlaySection />}
-                  {activeTab === 'community' && <CommunitySection />}
+                  <Column
+                    noMargin
+                    expand //TODO: check fullHeight and co on Safari.
```
ClementPasteau, line 175: todo?

newIDE/app/src/AnnouncementsFeed/AnnouncementsFeed.js
```diff
@@ -0,0 +1,100 @@
...
+  addMargins,
+}: AnnouncementsFeedProps) => {
+  const { announcements, error, fetchAnnouncements } = React.useContext(
+    AnnouncementsFeedContext
+  );
+  const { values, showAnnouncement } = React.useContext(PreferencesContext);
+
+  if (error) {
+    return (
+      <PlaceholderError onRetry={fetchAnnouncements}>
```
AlexandreSi, line 35: Should it silently fail instead? It doesn't seem that important to take space in the IDE and show that there's an issue

4ian:
> On the second screen I would have moved the links of our networks to the bottom of the page.

Changing that.

I check the alignments (might be the screens that are wrong) + the vertical line.

newIDE/app/src/AnnouncementsFeed/AnnouncementsFeed.js
```diff
@@ -0,0 +1,100 @@
+// @flow
```
AlexandreSi, line 1: should the file be renamed index since it has the same name as the folder?

newIDE/app/src/MainFrame/Preferences/PreferencesProvider.js
```diff
@@ -447,6 +449,33 @@ export default class PreferencesProvider extends React.Component<Props, State> {
     );
   }
 
+  _showAnnouncement(identifier: string, show: boolean) {
```
ClementPasteau, line 452: looking at the other props in the preferences, it's probably too late now, but I find the choice of name a bit confusing - especially if the only action available is hiding one.

ClementPasteau (approved):
code-wise lgtm!
design-wise, maybe aligning the alert with the rest of the page like Bouh suggests would be nice 👍

newIDE/app/src/AnnouncementsFeed/AnnouncementsFeed.js
4ian: It does :) In the HomePage, if there are no announcements, it does not even try to show the AnnouncementsFeed at the top (and no loading too), so it's effectively "best effort".
Error and loading *are* displayed though in the community page.

newIDE/app/src/AnnouncementsFeed/AnnouncementsFeed.js
4ian: Yes!

newIDE/app/src/MainFrame/EditorContainers/HomePage/index.js
4ian: Checked 👍

newIDE/app/src/MainFrame/Preferences/PreferencesProvider.js
4ian: I agree but I'll favor the consistency with the rest for now!

newIDE/app/src/MainFrame/EditorContainers/HomePage/index.js
4ian: Checked, works well on Safari and Firefox, removing the todo

newIDE/app/src/AnnouncementsFeed/AnnouncementsFeed.js
4ian: Yes, done!

PR #4277, opened 2022-09-09
Create GDevelop Authentication extension

ClementPasteau:
And allow submitting a leaderboard entry with a connected player


Extensions/GDevelopAuthentication/JsExtension.js
```diff
@@ -0,0 +1,144 @@
+// @flow
```
AlexandreSi, line 1: Should the file be renamed `GDevelopPlayerAuthentication`? Feels clearer

Extensions/GDevelopAuthentication/JsExtension.js
```diff
@@ -0,0 +1,144 @@
...
+
+module.exports = {
+  createExtension: function (
+    _ /*: (string) => string */,
+    gd /*: libGDevelop */
+  ) {
+    const extension = new gd.PlatformExtension();
+    extension
+      .setExtensionInformation(
+        'GDevelop Authentication',
```
AlexandreSi, line 30: `Player Authentication` here as well?

Extensions/GDevelopAuthentication/JsExtension.js
```diff
@@ -0,0 +1,144 @@
...
+      .addDependency()
+      .setName('InAppBrowser Cordova plugin')
+      .setDependencyType('cordova')
+      .setExportName('cordova-plugin-inappbrowser');
+
+    extension
+      .addAction(
+        'DisplayAuthenticationBanner',
+        _('Display authentication banner'),
+        _('Display the authentication banner for the player.'),
```
AlexandreSi, line 54: `for the player` is not very clear to me: for the player to make sure they're connected with the right account?

Extensions/GDevelopAuthentication/JsExtension.js
```diff
@@ -0,0 +1,144 @@
...
+      )
+      .setFunctionName(
+        'gdjs.gdevelopAuthentication.displayAuthenticationBanner'
+      );
+
+      extension
+      .addAction(
+        'DisplayAuthenticationWindow',
+        _('Display authentication window'),
+        _('Display the authentication window for the player.'),
```
AlexandreSi, line 74: Same here: in that case, for the player doesn't seem useful.

Extensions/Leaderboards/leaderboardstools.ts
```diff
@@ -155,6 +187,79 @@ namespace gdjs {
...
+        runtimeScene,
+      }: {
+        leaderboardId: string;
+        playerName?: string;
+        playerId?: string;
+        score: number;
+        scoreSavingState: ScoreSavingState;
+        runtimeScene: gdjs.RuntimeScene;
+      }) {
+        const baseUrl = 'https://api.gdevelop-app.com/play';
```
AlexandreSi, line 205: This can be changed for `api.gdevelop.io/play`

Extensions/Leaderboards/leaderboardstools.ts
```diff
@@ -193,61 +300,64 @@ namespace gdjs {
...
-            },
+      };
+
+      export const saveConnectedPlayerScore = function (
+        runtimeScene: gdjs.RuntimeScene,
+        leaderboardId: string,
+        score: float
+      ) {
+        let scoreSavingState: ScoreSavingState;
+        const playerId = 'toto';
```
AlexandreSi, line 320: To change I guess!

Extensions/GDevelopAuthentication/JsExtension.js
```diff
@@ -0,0 +1,144 @@
...
+    extension
+      .setExtensionInformation(
+        'GDevelop Authentication',
+        _('Player Authentication (experimental)'),
+        _('Allow your game to authenticate players.'),
+        'Florian Rival',
+        'Open source (MIT License)'
+      )
+      .setExtensionHelpPath('/all-features/authentication')
+      .setCategory('Authentication');
```
D8H, line 37: Should it be in the same category as leaderboards something like "Player services" or something?

Extensions/Leaderboards/leaderboardstools.ts
```diff
@@ -193,61 +300,64 @@ namespace gdjs {
...
+        scoreSavingState.startSaving({ playerId, score });
+
+        saveScore({
+          leaderboardId,
+          playerId,
+          score,
+          scoreSavingState,
+          runtimeScene,
+        });
       };
```
AlexandreSi, line 361: Really clean changes for the leaderboards!

Extensions/GDevelopAuthentication/gdevelopauthenticationtools.ts
```diff
@@ -0,0 +1,865 @@
...
+     */
+    const fetchAuthenticatedUserFromLocalStorage = (): boolean => {
+      const _gameId = gdjs.projectData.properties.projectUuid;
+
+      const authenticatedUserStorageItem = window.localStorage.getItem(
+        `${_gameId}_authenticatedUser`
+      );
+      if (!authenticatedUserStorageItem) {
+        _checkedLocalStorage = true;
+        return _userId === null;
```
AlexandreSi, line 92: I don't understand in a glance how this is useful, could need a comment

Extensions/GDevelopAuthentication/gdevelopauthenticationtools.ts
```diff
@@ -0,0 +1,865 @@
...
+
+      const _gameId = gdjs.projectData.properties.projectUuid;
+      (window as any).localStorage.removeItem(`${_gameId}_authenticatedUser`);
+      _authenticationBanner = computeNotAuthenticatedBanner(runtimeScene);
+    };
+
+    /**
+     * Retrieves the user information from the local storage, and store
+     * them in the extension variables.
+     * Returns if the fetched user is different than the current one used
```
AlexandreSi, line 81: true or false?

Extensions/GDevelopAuthentication/JsExtension.js
```diff
@@ -0,0 +1,144 @@
...
+
+    extension
+      .addDependency()
+      .setName('InAppBrowser Cordova plugin')
+      .setDependencyType('cordova')
+      .setExportName('cordova-plugin-inappbrowser');
+
+    extension
+      .addAction(
+        'DisplayAuthenticationBanner',
```
D8H, line 52: Where is it displayed?
Should it be an object?
- to let game creator place it where they want
- make it shake with a behavior if they want to catch player eyes
- put more than one instance if the game is played by 2 players

Extensions/GDevelopAuthentication/gdevelopauthenticationtools.ts
```diff
@@ -0,0 +1,865 @@
...
+     * in the variables. (useful when the user changes)
+     */
+    const fetchAuthenticatedUserFromLocalStorage = (): boolean => {
```
AlexandreSi, line 84: Should it be renamed `readAuthenticatedUserInLocalStorage`: I was expecting an API call to be made

Extensions/GDevelopAuthentication/gdevelopauthenticationtools.ts
```diff
@@ -0,0 +1,865 @@
...
+        fetchAuthenticatedUserFromLocalStorage();
+      }
+      return _username || '';
+    };
+
+    /**
+     * Remove the user information from the local storage.
+     */
+    export const logout = (runtimeScene: RuntimeScene) => {
+      if (!_checkedLocalStorage) {
```
AlexandreSi, line 67: I don't understand why this check is done

Extensions/GDevelopAuthentication/gdevelopauthenticationtools.ts
```diff
@@ -0,0 +1,865 @@
...
+        fetchAuthenticatedUserFromLocalStorage();
+      }
+      _username = null;
+      _userToken = null;
```
AlexandreSi, line 71: should userId also be set to `null`?

Extensions/GDevelopAuthentication/gdevelopauthenticationtools.ts
```diff
@@ -0,0 +1,865 @@
...
+            `${_gameId}_authenticatedUser`,
+            JSON.stringify({
+              username: _username,
+              userId: _userId,
+              userToken: _userToken,
+            })
+          );
+          _username = username;
+          _userId = event.data.body.userId;
+          _userToken = event.data.body.token;
```
AlexandreSi, line 157: I feel like those 3 lines should be put above the local storage item setting instruction.

Extensions/GDevelopAuthentication/gdevelopauthenticationtools.ts
```diff
@@ -0,0 +1,865 @@
...
+        runtimeScene
+      );
+      domElementContainer.appendChild(authenticationContainer);
+
+      // Based on which platform the game is running, we open the authentication window
+      // Inside the container, or in a new tab.
+      const electron = runtimeScene.getGame().getRenderer().getElectron();
+      if (electron) {
+        // If we're on Electron, use an iframe.
+        logger.info("We're on Electron");
```
AlexandreSi, line 748: To be removed I guess

Extensions/GDevelopAuthentication/JsExtension.js
arthuro555: I don't think internal names are supposed to have spaces in them, aren't they supposed to be PascalCase?

Extensions/GDevelopAuthentication/JsExtension.js
4ian: Ideally yes but in practice we'll give a "default", "works for most games" solution and in the future we can provide an EBO for this ;)

4ian (commented):
Looks clean! We'll need at some point a super lightweight abstraction over the DOM to avoid super imperative code, to reduce risks of bugs and complexity of handling transitions between use cases (i.e: have the UI be a function of the state). Not urgent though as this is still manageable.

Extensions/GDevelopAuthentication/gdevelopauthenticationtools.ts
```diff
@@ -0,0 +1,865 @@
...
+      const authenticationContainer = computeAuthenticationContainer(
+        runtimeScene
+      );
+      domElementContainer.appendChild(authenticationContainer);
+
+      // Based on which platform the game is running, we open the authentication window
+      // Inside the container, or in a new tab.
+      const electron = runtimeScene.getGame().getRenderer().getElectron();
+      if (electron) {
+        // If we're on Electron, use an iframe.
```
arthuro555, line 747: Why use an iframe on electron? Users are likely to already be logged into their liluo account and link it from their default web browser. It is also a bad security practice as it teaches the users to enter their credentials in untrusted contexts instead of exclusively on the liluo site, and an iframe could most likely be read using a JS event to steal user's passwords :/ 

If it's because you cannot do window messaging when opening in another browser, what multiple other applications I've seen do is to register a custom protocol on the machine and redirect using it from the authentication page to give the token back to the electron application.

4ian: This is indeed the problem (getting users to write credentials in untrusted contexts). And the issue with the browser is indeed getting a message back to the app.
Do you have an idea of how to do it with a browser?
I'm not sure how easy though it is to register a custom protocol.
Another way would be to do polling/long lived connection with the backend using a random token, but also unsure how robust/scalable it is

4ian: Instead of polling, we could also use web socket to notify the game from the backend (by opening a web socket connection with a random token, open Liluo login in the browser with this token and send to the backend the JWT when the user has logged in. The backend can then notify the web socket connection.

We just have to prepare a message to tell the user to go back to the game.

arthuro555: > Another way would be to do polling/long lived connection with the backend using a random token, but also unsure how robust/scalable it is

The AWS API gateway supports WebSockets, so you could do it kind of like this:
1. Connect over WebSocket to an OAuth API, and send an authentication request message
2. That triggers a lambda function via API gateway, send back the WebSocket connection ID
3. Open the Liluo oauth page with the WebSocket connection ID as get parameter
4. Once logged in (or the log in is cancelled), call a lambda function, send back the authentication result to the WebSocket that initiated the log in using the ID passed in as get parameter

Since it'd work on top of AWS for the unique ID generation & creation of long lived connections, and on top of lambdas, it should be pretty scalable I think. 

> I'm not sure how easy though it is to register a custom protocol.

It is pretty easy, but I think that a long lived connection to the backend should be more scalable, less error prone (what if two GDevelop games register the same protocol?) and reduce code duplication between Cordova and Electron

Extensions/GDevelopAuthentication/JsExtension.js
ClementPasteau: Yep, good idea 👍

Extensions/GDevelopAuthentication/gdevelopauthenticationtools.ts
ClementPasteau: The logic is convoluted here, I'll change this.

Extensions/GDevelopAuthentication/gdevelopauthenticationtools.ts
ClementPasteau: yep!

Extensions/GDevelopAuthentication/gdevelopauthenticationtools.ts
ClementPasteau: 👍

Extensions/GDevelopAuthentication/gdevelopauthenticationtools.ts
```diff
@@ -0,0 +1,865 @@
...
+        .getRenderer()
+        .getDomElementContainer();
+      if (!domElementContainer) {
+        handleAuthenticationError(
+          runtimeScene,
+          "The div element covering the game couldn't be found, the authentication banner cannot be displayed."
+        );
+        return;
+      }
+      if (_userToken && _username !== null) {
```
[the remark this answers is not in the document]
ClementPasteau: yep, I don't need the username to be passed down the function, as it's a namespace variable, I'm removing the check

Extensions/GDevelopAuthentication/gdevelopauthenticationtools.ts
ClementPasteau: Those are indeed good ideas 👍 
Completely agree with the fact that it's not good security practices.

As this extension is still experimental, it will not be widely used.
Also, the main usage will be on web to start with, as we gather feedback on usability.
So I'm suggesting to release is as is, and work on an updated version as soon as it is released.

Extensions/PlayerAuthentication/playerauthenticationtools.ts
```diff
@@ -0,0 +1,833 @@
...
+    // Authentication information.
+    let _username: string | null = null;
+    let _userId: string | null = null;
+    let _userToken: string | null = null;
+    let _justLoggedIn = false;
+
+    let _error: any = null;
+    let _checkedLocalStorage: boolean = false;
+
+    // Authentication display
```
D8H, line 15: Do you think it would be easy to spit it in 2 files: one for the logic and one for the presentation?

Extensions/PlayerAuthentication/playerauthenticationtools.ts
```diff
@@ -0,0 +1,833 @@
...
+     * Returns true if the user just logged in.
+     * Useful to update username or trigger messages in the game.
+     */
+    export const hasLoggedIn = () => {
+      if (_justLoggedIn) {
+        _justLoggedIn = false;
+        return true;
+      }
+      return false;
+    };
```
D8H, line 51: A getter should probably always return the same value when it's called twice in a row.

Extensions/PlayerAuthentication/playerauthenticationtools.ts
```diff
@@ -0,0 +1,833 @@
...
+          window.localStorage.setItem(
+            `${_gameId}_authenticatedUser`,
+            JSON.stringify({
+              username: _username,
+              userId: _userId,
+              userToken: _userToken,
+            })
+          );
+
+          _justLoggedIn = true;
```
D8H, line 177: Can this happen in the middle of the events execution?

Extensions/PlayerAuthentication/playerauthenticationtools.ts
ClementPasteau: That is a good idea, I'll see if I can split that

Extensions/PlayerAuthentication/playerauthenticationtools.ts
ClementPasteau: yes, the logic is not great here, I'm updating this

Extensions/PlayerAuthentication/playerauthenticationtools.ts
D8H: The callback comes from `window.addEventListener('message'`.
Does the main loop happens in another callback so they can't overstep each other or something like this?

ClementPasteau: Unsure about the exact details, but I think so.
@4ian can you confirm?

Extensions/PlayerAuthentication/playerauthenticationcomponents.ts
```diff
@@ -0,0 +1,374 @@
...
+
+      _textContainer.appendChild(loggedText);
+      _textContainer.appendChild(changeAccountText);
+      divContainer.appendChild(_textContainer);
+
+      return divContainer;
+    };
+
+    /**
+     * Creates a DOM element reprensenting a banner for the user to know
```
4ian, line 215:
```suggestion
     * Creates a DOM element representing a banner for the user to know
```

4ian (approved):

Extensions/PlayerAuthentication/playerauthenticationcomponents.ts
```diff
@@ -0,0 +1,374 @@
+namespace gdjs {
+  const logger = new gdjs.Logger('Player Authentication');
+  export namespace playerAuthenticationComponents {
+    /**
+     * Create the Iframe that will contain the authentication form. (electron)
```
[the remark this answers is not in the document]
ClementPasteau: useless function indeed now!

Extensions/Leaderboards/JsExtension.js
```diff
@@ -65,6 +65,32 @@ module.exports = {
       .addIncludeFile('Extensions/Leaderboards/leaderboardstools.js')
       .setFunctionName('gdjs.evtTools.leaderboards.savePlayerScore');
 
+    extension
+      .addAction(
+        'SaveConnectedPlayerScore',
+        _('Save connected player score'),
+        _("Save the connected player's score to the given leaderboard."),
+        _(
+          'Send to leaderboard _PARAM1_ the score _PARAM2_ for connected player'
```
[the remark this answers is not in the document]
ClementPasteau: Unsure about this idea. Let's wait and see the usage before doing any change

Extensions/GDevelopAuthentication/gdevelopauthenticationtools.ts
ClementPasteau: update: we've implemented a websocket connection when the application runs on Electron :)

Extensions/PlayerAuthentication/playerauthenticationtools.ts
4ian: JS is single threaded (unless using web workers), so it means callbacks are executed one after the other, and a function who runs, like the rendering of a frame, can't have anything else "happening in between".
In particular, the frame is rendered every 16.66ms (more or less), thanks to requestAnimationFrame. During this time, no callback can run. So it means the callbacks are necessarily launched between frames:

requestAnimationFrame 1 ------> requestAnimationFrame 2 ----> callback A --> requestAnimationFrame 3 ---> callback B, callback C ----> requestAnimationFrame 4.

The story would be different if this was a webworker modifying shared data. But this is not the case.
So the callbacks of an event are always before/after a frame.

PR #4279, opened 2022-09-11
Fix the events-based objects editor that were never showing up in dev mode.

D8H:


newIDE/app/src/UI/EditorMosaic/index.js
```diff
@@ -39,10 +39,13 @@ export const mosaicContainsNode = (
...
     (mosaic === node ||
-      (!!mosaic.first &&
-        mosaicContainsNode(((mosaic.first: ?any): ?EditorMosaicNode), node) &&
-        !!mosaic.second &&
-        mosaicContainsNode(((mosaic.second: ?any): ?EditorMosaicNode), node)))
+      ((!!mosaic.first &&
+        mosaicContainsNode(((mosaic.first: ?any): ?EditorMosaicNode), node)) ||
+        (!!mosaic.second &&
+          mosaicContainsNode(
+            ((mosaic.second: ?any): ?EditorMosaicNode),
```
4ian, line 46: // $FlowFixMe might have be maybe simpler than the double cast?

newIDE/app/src/EventsFunctionsExtensionEditor/index.js
```diff
@@ -1410,9 +1410,11 @@ export default class EventsFunctionsExtensionEditor extends React.Component<
                       )
                     }
                     initialNodes={
-                      isDev !==
+                      isDev ===
```
4ian, line 1413: I'm not sure to follow why isDev is compared to mosaicContainsNode? Any chance you can rework this ternary?

4ian (approved):
Approving to unblock but the logic is hard to parse/read, see if you can rework it?

newIDE/app/src/EventsFunctionsExtensionEditor/index.js
D8H: In development mode, the object list must be shown: the mosaic must be reset to default if the save settings don't contain it.
This will be removed as soon as we release the events-based objects editor.

4ian (approved):

PR #4281, opened 2022-09-12
Fix a crash when renaming a child-object

D8H:
It adds tests for object renaming in events (layout and event-based object function)

### DIFF
--- a/Core/GDCore/IDE/WholeProjectRefactorer.cpp
+++ b/Core/GDCore/IDE/WholeProjectRefactorer.cpp
@@ -1628,19 +1628,18 @@ void WholeProjectRefactorer::ObjectOrGroupRemovedInEventsFunction(
 
 void WholeProjectRefactorer::ObjectOrGroupRenamedInEventsBasedObject(
     gd::Project& project,
-    gd::EventsBasedObject& eventsBasedObject,
     gd::ObjectsContainer& globalObjectsContainer,
-    gd::ObjectsContainer& objectsContainer,
+    gd::EventsBasedObject& eventsBasedObject,
     const gd::String& oldName,
     const gd::String& newName,
     bool isObjectGroup) {
   for (auto &functionUniquePtr : eventsBasedObject.GetEventsFunctions().GetInternalVector()) {
-    auto function = functionUniquePtr.get();
+    auto *function = functionUniquePtr.get();
     WholeProjectRefactorer::ObjectOrGroupRenamedInEventsFunction(
         project,
         *function,
         globalObjectsContainer,
-        objectsContainer,
+        eventsBasedObject,
         oldName,
         newName,
         isObjectGroup);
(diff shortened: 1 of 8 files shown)


Core/GDCore/IDE/WholeProjectRefactorer.cpp
```diff
@@ -1628,19 +1628,18 @@ void WholeProjectRefactorer::ObjectOrGroupRemovedInEventsFunction(
 
 void WholeProjectRefactorer::ObjectOrGroupRenamedInEventsBasedObject(
     gd::Project& project,
-    gd::EventsBasedObject& eventsBasedObject,
     gd::ObjectsContainer& globalObjectsContainer,
```
4ian, line 1631: Open question: is the global objects container even making sense in an events based object?

Core/GDCore/IDE/WholeProjectRefactorer.cpp
```diff
@@ -1628,19 +1628,18 @@ void WholeProjectRefactorer::ObjectOrGroupRemovedInEventsFunction(
...
-    gd::ObjectsContainer& objectsContainer,
+    gd::EventsBasedObject& eventsBasedObject,
     const gd::String& oldName,
     const gd::String& newName,
     bool isObjectGroup) {
   for (auto &functionUniquePtr : eventsBasedObject.GetEventsFunctions().GetInternalVector()) {
-    auto function = functionUniquePtr.get();
+    auto *function = functionUniquePtr.get();
```
4ian, line 1637: 😬

Core/tests/WholeProjectRefactorer.cpp
```diff
@@ -106,14 +108,26 @@ const std::vector<const gd::EventsList *> GetEventsLists(gd::Project &project) {
   return eventLists;
 }
 
+const std::vector<const gd::ObjectsContainer *> GetObjectContainers(gd::Project &project) {
```
4ian, line 111: Should the name be more specific?

```suggestion
const std::vector<const gd::ObjectsContainer *> GetObjectContainersForMyOtherEventsBasedObject(gd::Project &project) {
```

4ian (approved):

Core/GDCore/IDE/WholeProjectRefactorer.cpp
D8H: The `*` was implicit, so it has no impact.

Core/GDCore/IDE/WholeProjectRefactorer.cpp
D8H: No, the GUI hold an empty one to fill the parameter (like behaviors do).

Core/tests/WholeProjectRefactorer.cpp
D8H: No, I should remove this method as I don't use it.
I though I would factorize tests for layout and ebo but it's not actually possible.

PR #4286, opened 2022-09-13
Experimental custom objects for runtime

D8H:


D8H:
Merged with:
- https://github.com/4ian/GDevelop/pull/4294

PR #4294, opened 2022-09-14
Custom objects for runtime

D8H:
This is the last version in development. An older version is built here:
- https://github.com/4ian/GDevelop/pull/4286


Core/GDCore/IDE/ProjectStripper.cpp
```diff
@@ -22,7 +24,28 @@ void GD_CORE_API ProjectStripper::StripProjectForExport(gd::Project& project) {
...
+  // Keep the EventsBasedObject object list because it's useful for the Runtime
+  // to create the child-object.
+  for (unsigned int extensionIndex = 0;
+       extensionIndex < project.GetEventsFunctionsExtensionsCount();
+       ++extensionIndex) {
+    auto &extension = project.GetEventsFunctionsExtension(extensionIndex);
+    auto &eventsBasedObjects = extension.GetEventsBasedObjects();
+    if (eventsBasedObjects.size() == 0) {
+      project.RemoveEventsFunctionsExtension(extension.GetName());
+      extensionIndex--;
```
4ian, line 36: This is risky. If it happens for the first extension you would overflow (well, "underflow") the unsigned int from 0 to -1 which would become the highest unsigned int possible, which would then stop the loop.

Instead, remove the ++extensionIndex and do it only when not removing the extension.

D8H: There should be no issue because the next thing to execute is the test and the increment of the for.

4ian (commented):
Made a first pass! I skipped the tests

Core/GDCore/IDE/ProjectStripper.cpp
4ian: > because the next thing to execute is the increment of the for.

Fair enough

> A while with an index is more risky to me because a misplaced closing bracket can lead to an infinite loop.

Sure, but I was advocating for a `for(unsigned int i = 0;i<xxxx;) { if(remove) {} else {++i;} }`.
At least I've been bitten in the past by unsigned int going from 0 to "max unsigned int" leading to infinite loop or skipping elements.

GDJS/Runtime/events-tools/runtimescenetools.ts
```diff
@@ -211,34 +243,42 @@ namespace gdjs {
...
-        runtimeScene.requestChange(gdjs.SceneChangeRequest.POP_SCENE);
+      export const popScene = function (runtimeScene: gdjs.RuntimeScene) {
+        runtimeScene
+          .getScene()
+          .requestChange(gdjs.SceneChangeRequest.POP_SCENE);
       };
 
-      export const stopGame = function (runtimeScene) {
-        runtimeScene.requestChange(gdjs.SceneChangeRequest.STOP_GAME);
+      export const stopGame = function (runtimeScene: gdjs.RuntimeScene) {
```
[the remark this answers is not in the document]
D8H: You said that when the function calls `getScene`, we should leave a scene parameter for clarity.

GDJS/Runtime/layer.ts
```diff
@@ -287,18 +317,53 @@ namespace gdjs {
       const sinValue = Math.sin(angleInRadians);
       x = cosValue * x - sinValue * y;
       y = sinValue * tmp + cosValue * y;
-      return [x + this.getCameraX(cameraId), y + this.getCameraY(cameraId)];
+      // TODO EBO use an AffineTransformation to avoid chained calls.
```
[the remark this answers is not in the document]
D8H: It was not at the right place.

GDJS/tests/benchmarks/layer.js
```diff
@@ -10,7 +10,8 @@ describe('gdjs.Layer', function() {
   var runtimeScene = new gdjs.RuntimeScene(runtimeGame);
 
   it('benchmark convertCoords and convertInverseCoords', function() {
-    this.timeout(20000);
+    // TODO EBO Understand why the timeout need to be 35s instead of 20s to pass on CI.
+    this.timeout(35000);
```
D8H, line 14: This is not great but I guess it because of the call depth that comes with generalization.

GDJS/Runtime/layer.ts
```diff
@@ -268,15 +263,51 @@ namespace gdjs {
...
+      let position = result || [0, 0];
+      // TODO EBO use an AffineTransformation to avoid chained calls.
+      position = this._runtimeScene.convertCoords(x, y, position);
+      return this.applyLayerInverseTransformation(
+        position[0],
+        position[1],
+        cameraId,
+        position
+      );
+    }
```
D8H, line 289: It's probably because the implementation do 2 calls instead of the same thing but inlined.

GDJS/Runtime/CustomRuntimeObject.ts
```diff
@@ -0,0 +1,587 @@
...
+    _updateUntransformedHitBoxes() {
+      this._isUntransformedHitBoxesDirty = false;
+
+      const oldUnscaledCenterX =
+        (this._unrotatedAABB.max[0] + this._unrotatedAABB.min[0]) / 2;
+      const oldUnscaledCenterY =
+        (this._unrotatedAABB.max[1] + this._unrotatedAABB.min[1]) / 2;
+
+      this._untransformedHitBoxes.length = 0;
+      if (this._instanceContainer.getAdhocListOfAllInstances().length === 0) {
```
4ian, line 189: We should probably avoid calling getAdhocListOfAllInstances() twice, as it's not a trivial operation.

4ian: Ok the result is actually memoized so no need!

GDJS/Runtime/RuntimeInstanceContainer.ts
```diff
@@ -0,0 +1,686 @@
...
+
+    /**
+     * @return the height of the game resolution for a {@link gdjs.RuntimeScene} or the
+     * internal size for a {@link gdjs.CustomRuntimeObject}.
+     */
+    abstract getViewportHeight(): float;
+
+    /**
+     * @return the center X of the game resolution for a {@link gdjs.RuntimeScene} or the
+     * internal size for a {@link gdjs.CustomRuntimeObject}.
```
4ian, line 117: internal size?

GDJS/Runtime/layer.ts
```diff
@@ -287,18 +320,54 @@ namespace gdjs {
...
+    convertInverseCoords(
+      x: float,
+      y: float,
+      cameraId: integer = 0,
+      result: FloatPoint
+    ): FloatPoint {
+      let position = result || [0, 0];
+      // TODO EBO use an AffineTransformation to avoid chained calls.
+      this.applyLayerTransformation(x, y, cameraId, position);
+      return this._runtimeScene.convertInverseCoords(
```
4ian, line 347: Can you explain why these two transformation and usage of this._runtimeScene here?

GDJS/Runtime/runtimescene.ts
```diff
@@ -35,89 +28,74 @@ namespace gdjs {
...
+        : 0;
+      this._cachedGameResolutionHeight = this._runtimeGame
+        ? this._runtimeGame.getGameResolutionHeight()
+        : 0;
       for (const name in this._layers.items) {
         if (this._layers.items.hasOwnProperty(name)) {
           /** @type gdjs.Layer */
           const theLayer: gdjs.Layer = this._layers.items[name];
-          theLayer.onGameResolutionResized();
+          theLayer.onGameResolutionResized(
```
4ian, line 95: Nice, any reason for moving oldGameRsolutionOriginX/Y here? Just make more sense?

newIDE/app/src/EventsFunctionsExtensionEditor/ObjectMethodSelectorDialog.js
```diff
@@ -156,14 +157,13 @@ export default function ObjectMethodSelectorDialog({
         />
         <MethodListItem
           icon={<Step style={styles.icon} />}
-          name={'doStepPreEvents'}
-          disabled={eventsFunctions.hasEventsFunctionNamed('doStepPreEvents')}
+          name={'doStepPostEvents'}
```
4ian, line 160: What's the reason to avoid doStepPreEvents (sorry maybe we discussed that already)?

GDJS/Runtime/CustomRuntimeObject.ts
D8H: It is now, it only rebuilds it if necessary. But you're right, I'll do the change.

GDJS/Runtime/RuntimeInstanceContainer.ts
D8H: The AABB of all children give the default dimension of the object. It's not a great name. I couldn't find something better.

GDJS/Runtime/layer.ts
D8H: `this._runtimeScene` is an `InstanceContainer` but I kept the old name for compatibility with community extensions (the embedded extension were using it so I was afraid to break something)

I think there is an issue there, I used the same method for visual transformations and hitboxes transformations. I'll redo this and write tests.

GDJS/Runtime/runtimescene.ts
D8H: It allows to generalize it to custom objects that have an "internal" dimension change.

newIDE/app/src/EventsFunctionsExtensionEditor/ObjectMethodSelectorDialog.js
D8H: It's easier to add one that remove one so I prefer to keep it hidden until we find a use-case for it.
Object events are mainly there to update the rendering according to scene or parent event changes so it might not be necessary to run events before the scene events.

GDJS/Runtime/layer.ts
D8H: Actually, it works fine. I added some tests.

GDJS/Runtime/SceneLayer.ts
```diff
@@ -0,0 +1,71 @@
...
+ * Copyright 2013-2016 Florian Rival (Florian.Rival@gmail.com). All rights reserved.
+ * This project is released under the MIT License.
+ */
+namespace gdjs {
+  /**
+   * Represents a layer of a container, used to display objects.
+   *
+   * Viewports and multiple cameras are not supported.
+   */
+  export class SceneLayer extends gdjs.Layer {
```
4ian, line 12: Should this be a `RuntimeSceneLayer`? The naming of gdjs.Layer was a mistake to not have the "Runtime".

GDJS/Runtime/SceneLayer.ts
```diff
@@ -0,0 +1,71 @@
+/*
+ * GDevelop JS Platform
+ * Copyright 2013-2016 Florian Rival (Florian.Rival@gmail.com). All rights reserved.
+ * This project is released under the MIT License.
+ */
+namespace gdjs {
+  /**
+   * Represents a layer of a container, used to display objects.
```
4ian, line 8: Update this comment, to explain that it's a scene layer and why it's different from a gdjs.Layer

newIDE/app/src/EventsFunctionsExtensionsLoader/MetadataDeclarationHelpers.js
```diff
@@ -82,15 +82,181 @@ export const declareBehaviorMetadata = (
...
     eventsBasedObject.getFullName() || eventsBasedObject.getName(),
     eventsBasedObject.getDescription(),
     getExtensionIconUrl(extension)
   );
+
+  // TODO EBO Use full type to identify object to avoid collision.
+  // Objects are identified by their name alone.
+  const objectType = eventsBasedObject.getName();
+
+  objectMetadata
```
4ian, line 100: Nice! I wonder if this will be needed to be moved to a separate file at some point if this grow. For now it seems ok!

GDJS/Runtime/RuntimeInstanceContainer.ts
```diff
@@ -0,0 +1,696 @@
...
```
4ian, line 117: Let's write down your explanation:

```suggestion
     * internal size for a {@link gdjs.CustomRuntimeObject} (AABB of all children for the default dimension of the object).
```

4ian (approved):

GDJS/tests/tests/CustomRuntimeObject.js
```diff
@@ -0,0 +1,722 @@
...
+        [64, 64],
+      ]);
+      expect(rootCustomObject.getHitBoxes()[3].vertices).to.eql([
+        [128, 128],
+        [64, 128],
+        [128, 64],
+      ]);
+    });
+
+    it('keeps hit-boxes up to date when its children move and push the bottom-right corner', async () => {
```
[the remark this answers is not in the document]
D8H: When a child is moved, every parents up to the root is invalidated. It would only cover 1 or 2 lines but it's still important to do. I'm adding a todo.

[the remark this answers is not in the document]
D8H: Actually, it seems covered by `describe('with custom objects as children'`.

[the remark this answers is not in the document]
4ian: Perfect then!

GDJS/Runtime/CustomRuntimeObject.ts
```diff
@@ -0,0 +1,587 @@
...
+      return this._flippedX;
+    }
+
+    isFlippedY(): boolean {
+      return this._flippedY;
+    }
+  }
+
+  // Others initialization and internal state management :
+  CustomRuntimeObject.supportsReinitialization = true;
```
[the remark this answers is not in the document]
D8H: Is it really useful? Most objects seems to set it to false.

GDJS/GDJS/Events/CodeGeneration/EventsCodeGenerator.cpp
```diff
@@ -191,9 +196,83 @@ gd::String EventsCodeGenerator::GenerateBehaviorEventsFunctionCode(
...
+  EventsCodeGenerator codeGenerator(globalObjectsAndGroups, objectsAndGroups);
+  codeGenerator.SetCodeNamespace(codeNamespace);
+  codeGenerator.SetGenerateCodeForRuntime(compilationForRuntime);
+
+  // Generate the code setting up the context of the function.
+  gd::String fullPreludeCode =
+      preludeCode + "\n" + "var that = this;\n" +
+      // runtimeScene is supposed to be always accessible, read
+      // it from the object
+      "var runtimeScene = this._instanceContainer;\n" +
```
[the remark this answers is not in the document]
D8H: Is it what is used by scene variable instructions for instance?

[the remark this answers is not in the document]
D8H: I copied it from the behavior generation code.
This is what is pass to every free functions?

GDJS/Runtime/CustomRuntimeObject.ts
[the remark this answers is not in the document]
4ian: In theory it could be a nice time/performance saver if you for example shoot a lot of bullets and destroy them, it would avoid creating/destroying objects and just "reset" a bunch of properties. But I don't have real benchmarks to prove this unfortunately.

GDJS/GDJS/Events/CodeGeneration/EventsCodeGenerator.cpp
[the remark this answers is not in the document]
4ian: In theory I think the code generator should use a more generic "instanceContainer" name and do the "getScene()" if necessary. We should inspect the code generation outputs to see what's the real usage of runtimeScene and see if it could be avoided. For now I guess it's ok-ish.

newIDE/app/src/EventsFunctionsExtensionEditor/ObjectMethodSelectorDialog.js
4ian: Fine with this 👍

GDJS/tests/benchmarks/layer.js
4ian: Resolving as fixed.

GDJS/Runtime/CustomRuntimeObject.ts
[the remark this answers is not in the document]
D8H: ok, it's probably something that can be done in a second version to be safe?

D8H:
Layer benchmark tests are still not stable.

image

GDJS/Runtime/CustomRuntimeObject.ts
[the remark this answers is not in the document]
4ian: I guess so yes. You can disable this for now

PR #4300, opened 2022-09-15
Fix events-based behavior default name

D8H:
It fixes a regression from:
- https://github.com/4ian/GDevelop/pull/4217

The parameter "name" in `PlatformExtension::AddBehavior` and `BehaviorMetadata::BehaviorMetadata` don't have the same meaning. I renamed it to avoid future mistakes.

```Cpp
gd::BehaviorMetadata& PlatformExtension::AddEventsBasedBehavior(
    const gd::String& name,
    const gd::String& fullname,
    const gd::String& description,
    const gd::String& group,
    const gd::String& icon24x24) {
  gd::String nameWithNamespace = GetNameSpace() + name;
  behaviorsInfo[nameWithNamespace] = BehaviorMetadata(GetNameSpace(),
                                                      nameWithNamespace,
```

### DIFF
--- a/Core/GDCore/Extensions/Metadata/BehaviorMetadata.cpp
+++ b/Core/GDCore/Extensions/Metadata/BehaviorMetadata.cpp
@@ -21,7 +21,7 @@ namespace gd {
 
 BehaviorMetadata::BehaviorMetadata(
     const gd::String& extensionNamespace_,
-    const gd::String& name_,
+    const gd::String& nameWithNamespace,
     const gd::String& fullname_,
     const gd::String& defaultName_,
     const gd::String& description_,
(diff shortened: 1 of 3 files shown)


Core/GDCore/Extensions/Metadata/BehaviorMetadata.cpp
```diff
@@ -45,32 +45,32 @@ BehaviorMetadata::BehaviorMetadata(
...
-        name_,
+        nameWithNamespace,
         fullname_,
-        // Default name is the name
-        name_,
+        defaultName_,
         description_,
         group_,
         icon24x24_,
         // Class name is the name, actually unused
```
4ian, line 70: Is this comment still valid?

4ian: Just answering my question, yes it is :)

4ian (approved):

PR #4307, opened 2022-09-16
Add context menu item to center view on the scene editor

AlexandreSi:
Add new methods to:
- ✅ Return to initial position (view matches the game resolution)
- ✅ Fit zoom to selected instances (center of the rectangle that contains them all
- ✅ Fit zoom to the whole scene

  image

- ✅ Select all instances of an object on the scene

  image

  demo_select_instances

Technical side quests done:
- Use Rectangle class everywhere the former rectangle object (`{x, y, width, height}`) was defined
- Remove deprecated keyboard shortcuts class
- Add missing types a bit everywhere it was needed (mainly for instance measurer)

Notes:
- There was a condition that prevented to detect mouse mid button clicks in the deprecated keyboard shortcuts class for Mac-like machines (that prevented to move the scene keeping the mouse mid button pressed) => I didn't keep it


newIDE/app/src/InstancesEditor/ViewPosition.js
```diff
@@ -103,9 +104,17 @@ export default class ViewPosition {
...
+   * Moves view to the rectangle center and returns the ideal zoom
+   * factor to fit to the rectangle.
+   */
+  fitToRectangle(rectangle: Rectangle) {
+    this.viewX = rectangle.centerX();
+    this.viewY = rectangle.centerY();
+    const idealZoomOnX = this._width / rectangle.width();
+    const idealZoomOnY = this._height / rectangle.height();
+
+    return Math.min(idealZoomOnX, idealZoomOnY) - 0.05; //Add margin so that the object don't feel cut
```
D8H, line 117: Zoom are ratios they can't be subtracted. Maybe `* 0.95` ?

newIDE/app/src/InstancesEditor/index.js
```diff
@@ -804,16 +811,69 @@ export default class InstancesEditor extends Component<Props> {
     this.viewPosition.scrollTo(x, y);
   }
 
-  centerView() {
+  zoomToFitContent() {
+    const instanceMeasurer = this.instancesRenderer.getInstanceMeasurer();
+    let rectangle: ?Rectangle;
```
D8H, line 816: Maybe `contentRectangle` or `contentAABB`?

newIDE/app/src/InstancesEditor/index.js
```diff
@@ -804,16 +811,69 @@ export default class InstancesEditor extends Component<Props> {
...
+  centerViewOnLastInstance(instances: Array<gdInitialInstance>) {
+    if (instances.length === 0) return;
 
-    this.viewPosition.scrollToInstance(instances[instances.length - 1]);
+    const instanceMeasurer = this.instancesRenderer.getInstanceMeasurer();
+    let lastInstanceRectangle = instanceMeasurer.getInstanceAABB(
+      instances[instances.length - 1],
+      new Rectangle()
+    );
+    this.viewPosition.fitToRectangle(lastInstanceRectangle);
```
D8H, line 876: Does it fit to the whole selection or just one instance of the selection?

AlexandreSi: I kept the current behaviour: the camera centers on the last selected instance. It's kinda logic since you're selecting them from the instance list, it shows on the canvas what is the instance that you just selected

newIDE/app/src/InstancesEditor/ViewPosition.js
AlexandreSi: Good idea

newIDE/app/src/InstancesEditor/index.js
AlexandreSi: I'll change it!

newIDE/app/src/InstancesEditor/index.js
D8H: I'm not sure it's a good UX. The view will jump from one instance to another and other instances from the selection will go out of the view. User won't have a clear view of what is selected.

newIDE/app/src/InstancesEditor/index.js
```diff
@@ -804,16 +811,72 @@ export default class InstancesEditor extends Component<Props> {
...
+      initialInstances.iterateOverInstances(getInstanceRectangle);
+
+      if (contentAABB) {
+        const idealZoom = this.viewPosition.fitToRectangle(contentAABB);
+        this.setZoomFactor(idealZoom);
+      }
+    }
+  }
+
+  zoomToInitialPosition() {
```
ClementPasteau, line 845: Is this the correct initial position?
If the user change the game start position, does it also change the origin of the canvas? (0,0) (maybe I'm just imagining they can)

newIDE/app/src/InstancesEditor/index.js
```diff
@@ -804,16 +811,72 @@ export default class InstancesEditor extends Component<Props> {
...
     const y = this.props.project.getGameResolutionHeight() / 2;
     this.viewPosition.scrollTo(x, y);
+    this.setZoomFactor(1);
   }
 
-  centerViewOn(instances: Array<gdInitialInstance>) {
-    if (!instances.length) return;
+  zoomToFitSelection(instances: Array<gdInitialInstance>) {
+    if (instances.length === 0) return;
+    const [firstInstance, ...otherInstances] = instances;
```
ClementPasteau, line 854: Is there a risk this throws an error if there is only 1 instance? (trying to decompose `otherInstances`)

newIDE/app/src/SceneEditor/Toolbar.js
```diff
@@ -26,156 +27,137 @@ type Props = {|
...
-          canUndo={this.props.canUndo}
-          redo={this.props.redo}
-          canRedo={this.props.canRedo}
-          deleteSelection={this.props.deleteSelection}
-          toggleWindowMask={this.props.toggleWindowMask}
-          toggleGrid={this.props.toggleGrid}
-          setupGrid={this.props.openSetupGrid}
-          canDeleteSelection={
-            this.props.instancesSelection.getSelectedInstances().length !== 0
+const Toolbar = (props: Props) => {
```
ClementPasteau, line 35: 🎉

newIDE/app/src/SceneEditor/index.js
```diff
@@ -882,8 +880,48 @@ export default class SceneEditor extends React.Component<Props, State> {
...
+        accelerator: 'CmdOrCtrl+numadd',
+      },
+      {
+        label: i18n._(t`Zoom out`),
+        click: this.zoomOut,
+        accelerator: 'CmdOrCtrl+numsub',
+      },
+      {
+        label: i18n._(t`Zoom to fit selection`),
+        click: () => this.zoomToFitSelection(),
```
ClementPasteau, line 910: nit, this can be just `this.zoomToFitSelection` (same below)

newIDE/app/src/UI/KeyboardShortcuts/index.js
```diff
@@ -68,24 +86,81 @@ export default class KeyboardShortcuts {
...
+  shouldMoveView() {
+    return this._spacePressed || this._mouseMidButtonPressed;
+  }
+
+  shouldZoom() {
+    if (isMacLike()) {
+      return this._isControlOrCmdPressed();
+    } else {
+      if (
+        !this._isControlOrCmdPressed() &&
```
ClementPasteau, line 114: nit: this can just be `return !this._isControlOrCmdPressed() && !this._altPressed && !this._shiftPressed`

newIDE/app/src/InstancesEditor/index.js
LuniMoon: I would say:

- Camera zooms in to selected object on the scene
- It does change the zoom (it zooms in to the selected object on the scene).

As for instances repeated across the scene: I see the issue.
Indeed at first I would say "zoom to the instance selected on the object list", but considering that there could be a "duplicated" instance on the scene, I would go to what design software does:
It doesn't allow you to "Zoom to instance with is selected from the object list", but rather selecting an object from the list and "Highlight instances of this object on the project" on the same view, without zooming-in or out.
The copy that they use on the "right-click menu" is "Highlight component instances" or for GDevelop: "Highlight object instances".

newIDE/app/src/UI/KeyboardShortcuts/index.js
```diff
@@ -68,24 +86,81 @@ export default class KeyboardShortcuts {
...
   _isControlOrCmdPressed = () => {
     // On macOS, meta key (Apple/Command key) acts as Control key on Windows/Linux.
     return this._metaPressed || this._ctrlPressed;
   };
 
+  onMouseDown = (evt: MouseEvent) => {
+    if (evt.button === MID_MOUSE_BUTTON) {
+      this._mouseMidButtonPressed = true;
+    } else {
+      this._mouseMidButtonPressed = false;
```
ClementPasteau, line 147: hmm, is it normal to go back to false if the button pressed is not the mid?
(it means pressing mid, then left, would return mid to false)

newIDE/app/src/Utils/TransformRect.js
```diff
@@ -1,10 +1,11 @@
-export default function transformRect(transformPoint, { x, y, width, height }) {
-  const startPos = transformPoint(x, y);
-  const endPos = transformPoint(x + width, y + height);
-  return {
-    x: startPos[0],
-    y: startPos[1],
-    width: endPos[0] - startPos[0],
-    height: endPos[1] - startPos[1],
-  };
+// @flow
```
ClementPasteau, line 1: 🪨

ClementPasteau (commented):
lgtm!

newIDE/app/src/InstancesEditor/index.js
D8H: I don't think it's possible to change the origin of the scene.

newIDE/app/src/InstancesEditor/ViewPosition.js
```diff
@@ -103,9 +104,17 @@ export default class ViewPosition {
     this.viewY = y;
   }
 
-  scrollToInstance(instance: gdInitialInstance) {
```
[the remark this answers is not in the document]
AlexandreSi: I changed the only use of this method to use the method that centers on the instance

newIDE/app/src/InstancesEditor/index.js
```diff
@@ -176,18 +192,22 @@ export default class InstancesEditor extends Component<Props> {
...
-      this.keyboardShortcuts.blur();
-    });
-    this.pixiRenderer.view.addEventListener('mouseover', e => {
-      this.keyboardShortcuts.focus();
-    });
-    this.pixiRenderer.view.addEventListener('mouseout', e => {
-      this.keyboardShortcuts.blur();
-    });
+    this.pixiRenderer.view.addEventListener(
+      'keydown',
```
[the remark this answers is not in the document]
AlexandreSi: What I removed was a hack to ensure that the shortcuts were active only when the pixi canvas was focused or hovered.
My change is a bit different, it adds event listeners on the canvas element so that the browser handles the "is the canvas active?" part

newIDE/app/src/InstancesEditor/InstancesRenderer/LayerRenderer.js
```diff
@@ -234,10 +230,7 @@ export default class LayerRenderer {
       top = Math.min(top, rotatedRectangle[i][1]);
       bottom = Math.max(bottom, rotatedRectangle[i][1]);
     }
-    bounds.left = left;
-    bounds.right = right;
-    bounds.top = top;
-    bounds.bottom = bottom;
+    bounds.set(left, top, right, bottom);
```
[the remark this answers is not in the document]
AlexandreSi: You're right, I'll change it, I encountered this mistake during the refactor

newIDE/app/src/InstancesEditor/index.js
```diff
@@ -804,16 +811,72 @@ export default class InstancesEditor extends Component<Props> {
...
+  zoomToFitContent() {
+    const instanceMeasurer = this.instancesRenderer.getInstanceMeasurer();
+    let contentAABB: ?Rectangle;
+    const getInstanceRectangle = new gd.InitialInstanceJSFunctor();
+    // $FlowFixMe - invoke is not writable
+    getInstanceRectangle.invoke = instancePtr => {
+      // $FlowFixMe - wrapPointer is not exposed
+      const instance = gd.wrapPointer(instancePtr, gd.InitialInstance);
+      if (!contentAABB) {
+        contentAABB = instanceMeasurer.getInstanceAABB(
```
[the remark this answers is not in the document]
AlexandreSi: Not really, because if you create it outside, it will have 0 as default values for the 4 coordinates. So if you have a single instance that fits in [10, 10, 10, 10], it will end up with [0, 0, 10, 10]

newIDE/app/src/InstancesEditor/index.js
AlexandreSi: No, `otherInstances` will be `[]` in the case there is only one instance.

newIDE/app/src/UI/KeyboardShortcuts/index.js
AlexandreSi: I kept the original behaviour that was in the DeprecatedKeyboardShortcut.
I think it makes sense since you usually press a single button at a time.

ClementPasteau (approved):
lgtm!

ClementPasteau (approved):
👍

newIDE/app/src/Utils/Layout.js
```diff
@@ -0,0 +1,34 @@
...
+      instancePtr,
+      gd.InitialInstance
+    );
+    if (instance.getObjectName() === objectName) {
+      objectInstances.push(instance);
+    }
+  };
+  // $FlowFixMe - JSFunctor is incompatible with Functor
+  instances.iterateOverInstances(instanceGetter);
+
```
4ian, line 25: We have a memory leak here: `instanceGetter` should be deleted.

newIDE/app/src/InstancesEditor/index.js
```diff
@@ -804,16 +811,72 @@ export default class InstancesEditor extends Component<Props> {
...
+        );
+      } else {
+        contentAABB.union(
+          instanceMeasurer.getInstanceAABB(instance, new Rectangle())
+        );
+      }
+    };
+    // $FlowFixMe - JSFunctor is incompatible with Functor
+    initialInstances.iterateOverInstances(getInstanceRectangle);
+
```
4ian, line 838: Memory leak: `getInstanceRectangle` should be deleted.

Give a check in the whole codebase for `new gd.InitialInstanceJSFunctor` and check if they are deleted (the answer is no, it's my fault, and this is very bad because in the long run we take too much memory!).

4ian (approved):
LGTM after checking the (existing) memory leak issues (unless I'm mistaken)

4ian (approved):

PR #4310, opened 2022-09-20
Fix shared data initialization and unserializing that wasn't creating properties on the scene for Physics.

D8H:
It was an object slicing, the Physics behavior implementation was never called.


youssef3453:
I was directed by someone form the g develop a forum here after I explained my issue with the physics 2 engine. However, I don't understand what I should do now. I'm not that experienced with GitHub however, I'm familiar with its basic operations (pull and push) please let me know what  I should do to fix my issue. I have G develop PC version on windows

PR #4328, opened 2022-09-27
WIP: Add asset packs that can be purchased in the Asset store home

AlexandreSi:
New component: ImagesDisplay that can be seen in PrivateAssetPackDialog story


newIDE/app/src/AssetStore/AssetStoreContext.js
```diff
@@ -27,8 +31,10 @@ import {
   assetStoreHomePageState,
 } from './AssetStoreNavigator';
 import { type ChosenCategory } from '../UI/Search/FiltersChooser';
+import _ from 'lodash';
```
ClementPasteau, line 34: I think it's not recommended to import the whole library like this, because it can increase the bundle size.
let's do `import shuffle from 'lodash/shuffle';`

(found this article https://www.blazemeter.com/blog/import-lodash-libraries)

newIDE/app/src/AssetStore/AssetStoreContext.js
```diff
@@ -27,8 +31,10 @@ import {
...
 
 const defaultSearchText = '';
+const ACTIVATE_ASSET_PACK_MARKETPLACE = false;
```
ClementPasteau, line 37: feature-flag I presume? maybe add a TODO: remove when marketplace ready

newIDE/app/src/AssetStore/AssetStoreContext.js
```diff
@@ -259,6 +287,25 @@ export const AssetStoreStateProvider = ({
     [fetchAssetsAndFilters, assetShortHeadersById, isLoading]
   );
 
+  // Randomize asset packs when number of asset packs and private asset packs change
```
ClementPasteau, line 290: instead of randomizing with an effect, should you do it just below the `setPrivateAssetPacks(privateAssetPacks);` (line 259) ?
You should have both arrays ready and you can set the value here.

Also, is there a need to have multiple arrays (assetPacks, privateAssetPacks, ordering) or can we just have 1 array in the state `assetPacks` that contain the random ordering of all assetPacks (public & private) ?

newIDE/app/src/AssetStore/AssetsHome.js
```diff
@@ -59,17 +60,108 @@ const useStylesForGridListItem = makeStyles(theme =>
...
+  assetPackRandomOrdering,
   onPackSelection,
+  onPrivateAssetPackSelection,
 }: Props) => {
   const classesForGridListItem = useStylesForGridListItem();
   const windowWidth = useResponsiveWindowWidth();
   const gdevelopTheme = React.useContext(ThemeContext);
 
+  const starterPacksTiles = starterPacks.map((pack, index) => (
+    <GridListTile
```
ClementPasteau, line 81: This feels like you can extract this into its own component `AssetPackTile`

newIDE/app/src/AssetStore/AssetsHome.js
```diff
@@ -59,17 +60,108 @@ const useStylesForGridListItem = makeStyles(theme =>
...
+              {pack.userFriendlyPrice ? ' - ' + pack.userFriendlyPrice : null}
+            </Text>
+          </Line>
+        </Column>
+      </Paper>
+    </GridListTile>
+  ));
+
+  const privateAssetPacksTiles = privateAssetPacks.map(pack => (
+    <GridListTile
```
ClementPasteau, line 121: this feels very similar to the starterAssetPackTile, see other comment, can we create a component that handles both cases (public/private)?

newIDE/app/src/AssetStore/AssetsHome.js
```diff
@@ -59,17 +60,108 @@ const useStylesForGridListItem = makeStyles(theme =>
...
+            <Text style={styles.packTitle} color="primary" size="body2">
+              <Trans>{pack.description}</Trans>
+            </Text>
+          </Line>
+        </Column>
+      </Paper>
+    </GridListTile>
+  ));
+
+  const allTiles = starterPacksTiles
```
ClementPasteau, line 159: see comment in provider. I'm wondering if we could have just 1 array of assetPacks already randomized, (with a prop to know if it's public or private) and you just have to loop through it, rather than doing the logic here

newIDE/app/src/AssetStore/PrivateAssetPackDialog.js
```diff
@@ -0,0 +1,244 @@
...
+        } catch (error) {
+          if (error.response && error.response.status === 404) {
+            setErrorText(
+              <Trans>
+                Asset pack not found - An error occurred, please try again
+                later.
+              </Trans>
+            );
+          } else {
+            setErrorText(<Trans>Unknown error</Trans>);
```
ClementPasteau, line 94: maybe improve that message for the user? something like `An error occurred, please try again later`

newIDE/app/src/AssetStore/AssetStoreContext.js
4ian, line 34: Nit: see if you can just import the specific package you need: lodash/xxx

newIDE/app/src/AssetStore/PrivateAssetPackDialog.js
```diff
@@ -0,0 +1,244 @@
...
+              </>
+            ) : assetPackDetails && sellerPublicProfile ? (
+              <>
+                <Column noMargin>
+                  <Text size="title">{name}</Text>
+                  <Text size="body2">
+                    <Trans>by</Trans>{' '}
+                    <Link
+                      onClick={() => setOpenSellerPublicProfileDialog(true)}
+                      href=""
```
ClementPasteau, line 142: Looks like Link has never been used without an href. 
Are we sure it works fine on web?

newIDE/app/src/UI/ImagesDisplay.js
```diff
@@ -0,0 +1,255 @@
...
+                    <CardMedia>
+                      <CorsAwareImage
+                        src={url}
+                        style={{
+                          ...styles.mobileImageCarouselItem,
+                          height: mobileImageWidth / (16 / 9),
+                        }}
+                        alt={altTextTemplate.replace(
+                          /{imageIndex}/g,
+                          String(selectedImageIndex + 1)
```
ClementPasteau, line 179: I've never used the String() constructor. Maybe use toString() instead?

newIDE/app/src/AssetStore/PrivateAssetPackDialog.js
```diff
@@ -0,0 +1,244 @@
...
+                          </Column>
+                          <Column noMargin expand>
+                            <Text size="sub-title">
+                              <Trans>Licensing</Trans>
+                            </Text>
+                            <Line noMargin alignItems="center">
+                              <Mark fontSize="small" />
+                              <Spacer />
+                              <Text displayInlineAsSpan noMargin>
+                                <Trans>Personal projects</Trans>
```
ClementPasteau, line 199: I guess those elements should be dependent on the Licensing of the Asset Pack?
Or are all private asset packs under the same license?

newIDE/app/src/UI/ImagesDisplay.js
```diff
@@ -0,0 +1,255 @@
...
+              </CardMedia>
+            </Grid>
+          ))}
+        </Grid>
+      </Line>
+    </Column>
+  );
+};
+
+export default ImagesDisplay;
```
ClementPasteau, line 255: I'm not sure about the naming. It feels really... generic. But I don't have a better idea 😅 

`ResponsiveImagesRow` ?

4ian: Could even be ScrollingImagesRow, to show that it's a "carousel of images".

4ian: Actually, it's a gallery of images. ResponsiveImagesGallery?

newIDE/app/src/UI/ImagesDisplay.js
```diff
@@ -0,0 +1,255 @@
...
+        <Line justifyContent="center">
+          <Text noMargin size="body2">
+            {currentlyViewedImageIndex + 1}/{imagesUrls.length}
+          </Text>
+        </Line>
+      </div>
+    );
+  }
+  return (
+    <Column noMargin>
```
4ian, line 198: nit: rather than adding a line around the Grid, you can use a ColumnStackLayout which inserts spacer between child (and no spacer around them, so the component does not introduce unnecessary paddings that make things hard to align).

4ian:
LGTM overall, I let you check with Clement for merging this when ready.

newIDE/app/src/AssetStore/AssetStoreContext.js
AlexandreSi: Thanks!

newIDE/app/src/AssetStore/AssetsHome.js
AlexandreSi: For now they are very alike but this will change when we will display prices. So I'd rather not refactor this into a single component

newIDE/app/src/AssetStore/PrivateAssetPackDialog.js
AlexandreSi: I didn't notice any issue with using that. I will change it for "#" so that React does not translate this as the following html:
```
<a ... href>...</a>
```

newIDE/app/src/UI/ImagesDisplay.js
AlexandreSi: Good idea! I'll use that

newIDE/app/src/UI/ImagesDisplay.js
AlexandreSi: Oh good idea! Thanks

newIDE/app/src/AssetStore/PrivateAssetPackDialog.js
AlexandreSi: For the moment, we decided it will be the same licensing for every asset pack

ClementPasteau (approved):
NIIICE!

ClementPasteau:
bring the 🥐

PR #4329, opened 2022-09-27
Allow event-based objects to define a default name for created objects

D8H:
For instance, for the PanelSpriteButton event-based object, the default name can be "Button".
When users create a new object, it will be called "NewButton" instead of "NewPanelSpriteButton" or "NewObject".

### DIFF
--- a/newIDE/app/src/EventsBasedObjectEditor/index.js
+++ b/newIDE/app/src/EventsBasedObjectEditor/index.js
@@ -121,17 +121,17 @@ export default class EventsBasedObjectEditor extends React.Component<
                       eventsBasedObject.getName()
                     }
                     onChange={text => {
-                      if (!gd.Project.validateName(text)) {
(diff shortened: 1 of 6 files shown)


newIDE/app/src/EventsBasedObjectEditor/index.js
```diff
@@ -105,6 +105,21 @@ export default class EventsBasedObjectEditor extends React.Component<
...
+                commitOnBlur
+                floatingLabelText={
+                  <Trans>Default name for created objects</Trans>
+                }
+                value={
+                  eventsBasedObject.getDefaultName() ||
+                  eventsBasedObject.getName()
+                }
+                onChange={text => {
+                  eventsBasedObject.setDefaultName(text);
```
4ian, line 118: Maybe we should validate it's proper identifier? With the ill named "gd.Project.validateName(newName)"

4ian (approved):

PR #4330, opened 2022-09-27
Add community extensions subsection and warning

D8H:
Also fix a "read more" link loop.

The extension I took is probably well made because Arthuro is the author. But, for other extensions, as we don't necessarily do a review of the description, I think it's important to add a warning to keep a good impression on readers.

image

image

image


D8H:
Merged with:
- https://github.com/4ian/GDevelop/pull/4332

PR #4331, opened 2022-09-27
Add new categories for extensions.

D8H:


D8H:
> Do we need to requalify the category of some builtin extensions?

Yes, I'm merging this PR to be sure the categories can be selected by extension creators in the upcoming update and I'll take a look.

PR #4332, opened 2022-09-28
Add subsections for extensions categories in the Wiki

D8H:
Built over:
- https://github.com/4ian/GDevelop/pull/4330

It handles any category we will add.
Categories are sorted alphabetically.
General is the default category.

Extensions are h4, they no longer appear on the index. This could be an issue.

image

### DIFF
--- a/newIDE/app/scripts/extract-extensions-document.js
+++ b/newIDE/app/scripts/extract-extensions-document.js
@@ -86,7 +86,7 @@ const getAllExtensionAndExtensionShortHeaders = async () => {
 };
 
 
-const getExtensionSectionAndCreatePage = async (extension, extensionShortHeader) => {
+const createExtensionReferencePage = async (extension, extensionShortHeader, isCommunity) => {
   const folderName = getExtensionFolderName(extension.name);
   const referencePageUrl = `${gdevelopWikiUrlRoot}/extensions/${folderName}/reference`;
   const helpPageUrl = getHelpLink(extension.helpPath) || referencePageUrl;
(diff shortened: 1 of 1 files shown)


newIDE/app/scripts/extract-extensions-document.js
```diff
@@ -14,6 +14,7 @@ const {
   improperlyFormattedHelpPaths,
 } = require('./lib/WikiHelpLink');
 const { convertMarkdownToDokuWikiMarkdown } = require('./lib/DokuwikiHelpers');
+const { table } = require('console');
```
4ian, line 17: Bad auto import I guess

newIDE/app/scripts/extract-extensions-document.js
```diff
@@ -85,6 +86,110 @@ const getAllExtensionAndExtensionShortHeaders = async () => {
...
+      ? ` ([[${referencePageUrl}|reference]])`
+      : '') +
+    '\n\n');
+};
+
+const getAllExtensionsSections = (extensionsAndExtensionShortHeaders) => {
+  let extensionSectionsContent = "";
+  const extensionsByCategory = sortKeys(group(extensionsAndExtensionShortHeaders, getExtensionCategory));
+  for (const category in extensionsByCategory) {
+    if (Object.hasOwnProperty.call(extensionsByCategory, category)) {
```
4ian, line 181: This sounds convoluted, is this necessary?

newIDE/app/scripts/extract-extensions-document.js
```diff
@@ -85,6 +85,108 @@ const getAllExtensionAndExtensionShortHeaders = async () => {
   return extensions;
 };
 
+const group = (array, getKey) => {
```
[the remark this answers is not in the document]
4ian: (following lodash's groupBy naming)

newIDE/app/scripts/extract-extensions-document.js
```diff
@@ -85,6 +85,108 @@ const getAllExtensionAndExtensionShortHeaders = async () => {
...
+
+const sortKeys = (table) => {
+  const sortedTable = {};
+  for (const key of Object.keys(table).sort()) {
+    sortedTable[key] = table[key];
+  }
+  return sortedTable;
+}
+
+const getExtensionCategory = pair => pair.extension.category || 'General';
```
4ian, line 110: If used only once, this could be inlined I think as the naming is too generic

newIDE/app/scripts/extract-extensions-document.js
```diff
@@ -85,6 +85,108 @@ const getAllExtensionAndExtensionShortHeaders = async () => {
...
+    'reference.txt'
+  );
+  await fs.mkdir(path.dirname(extensionReferenceFilePath), {
+    recursive: true,
+  });
+  await fs.writeFile(extensionReferenceFilePath, referencePageContent);
+  console.info(`ℹ️ File generated: ${extensionReferenceFilePath}`);
+};
+
+const getExtensionSection = (extension, extensionShortHeader) => {
```
4ian, line 155: rather than getXxx, name generateXxx all the functions generating a text.

newIDE/app/scripts/extract-extensions-document.js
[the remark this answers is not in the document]
D8H: I would have choose this too but found this:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/group

PR #4333, opened 2022-09-28
Check child variable is defined when casting variable array to js object

### LINKED ISSUE #4327: ToJSON will cause error when array has missing keys
Calling ToJSON (GlobalVarToJSON, etc.) can cause a error, which will result in missing frame logic which can cause all sorts of bugs. An array really shouldn't have missing keys, but in Gdevelop it's really easy to cause this type of situation because keys are created dynamically just from checking them.

This error may seem contrived, but it happened to me on accident and it was quite difficult to debug and pinpoint the cause.

## To Reproduce
1. Create a Scene variable "blob" of type Struct
2. Add a child to "blob" called "list" of type Array
3. Add two boolean children to "list"
4. Call ToJSON on "blob" and it will work as expected
5. Now add a conditional to check if "blob.list[5]" is true... this will create a gap in the array.
6. Call ToJSON on "blob" and it will cause an error

Demo Reproduction:
[issue-4327.zip](https://github.com/4ian/GDevelop/files/9661363/issue-4327.zip)

What the "blob" looks like after checking blob.list[5] in condition:
global-vars-after

## Other details
* Windows Desktop Version 5.0.145

AlexandreSi:
Fixes #4327

### DIFF
--- a/GDJS/Runtime/variable.ts
+++ b/GDJS/Runtime/variable.ts
@@ -199,9 +199,7 @@ namespace gdjs {
             // A variable can have empty items in its children array if one inserts
             // a variable at an index greater than highest index. All the array elements
             // in the gap will be empty elements.
-            item === undefined
-              ? arr.push(undefined)
-              : arr.push(item.toJSObject());
+            arr.push(item === undefined ? undefined : item.toJSObject());
           }
           return arr;
       }
--- a/GDJS/Runtime/variable.ts
+++ b/GDJS/Runtime/variable.ts
@@ -199,8 +199,9 @@ namespace gdjs {
             // A variable can have empty items in its children array if one inserts
             // a variable at an index greater than highest index. All the array elements
             // in the gap will be empty elements.
-            if (item === undefined) arr.push(undefined);
-            else arr.push(item.toJSObject());
+            item === undefined
+              ? arr.push(undefined)
+              : arr.push(item.toJSObject());
           }
           return arr;
       }
--- a/GDJS/Runtime/variable.ts
+++ b/GDJS/Runtime/variable.ts
@@ -194,7 +194,14 @@ namespace gdjs {
           return obj;
         case 'array':
           const arr: any[] = [];
-          for (const item of this._childrenArray) arr.push(item.toJSObject());
+          for (const item of this._childrenArray) {
+            // item can be undefined if there are empty elements in the array.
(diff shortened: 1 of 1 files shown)


GDJS/Runtime/variable.ts
```diff
@@ -194,7 +194,14 @@ namespace gdjs {
           return obj;
         case 'array':
           const arr: any[] = [];
-          for (const item of this._childrenArray) arr.push(item.toJSObject());
+          for (const item of this._childrenArray) {
+            // item can be undefined if there are empty elements in the array.
+            // A variable can have empty items in its children array if one inserts
+            // a variable at an index greater than highest index. All the array elements
+            // in the gap will be empty elements.
+            if (item === undefined) arr.push(undefined);
```
ClementPasteau, line 202: can `null` be also a value of the array?

AlexandreSi: It shouldn't be. It's targeted against this use-case:
```
// variable is an array variable with 2 children at indices 0 and 1
variable.getChildAt(5)
// variable is now [XXX, YYY, empty, empty, empty, ZZZ]
```

GDJS/Runtime/variable.ts
```diff
@@ -194,7 +194,14 @@ namespace gdjs {
...
+            else arr.push(item.toJSObject());
```
D8H, line 203: a `?` maybe?

D8H (approved):

GDJS/Runtime/variable.ts
```diff
@@ -194,7 +194,13 @@ namespace gdjs {
           return obj;
         case 'array':
           const arr: any[] = [];
-          for (const item of this._childrenArray) arr.push(item.toJSObject());
+          for (const item of this._childrenArray) {
+            // item can be undefined if there are empty elements in the array.
+            // A variable can have empty items in its children array if one inserts
+            // a variable at an index greater than highest index. All the array elements
+            // in the gap will be empty elements.
+            arr.push(item === undefined ? undefined : item.toJSObject());
```
arthuro555, line 202: Could have been simplified to `arr.push(item && item.toJSObject())`

arthuro555 (commented):
The typing of the internal array should probably be modified to add undefined as a possible child type, to make sure we handle that case everywhere and catch such errors.

PR #4334, opened 2022-09-28
Add invert condition shortcut

AlexandreSi:
`I` is already taken by the instances list opening on the scene, so I set `J`.

I thought that they wouldn't interact since it's not the same "scope", it could deserve a change but that's not highly needed.


4ian (approved):

newIDE/app/src/EventsSheet/index.js
```diff
@@ -544,6 +551,11 @@ export class EventsSheetComponentWithoutHandle extends React.Component<
       label: i18n._(t`Invert Condition`),
       click: () => this._invertSelectedConditions(),
       visible: hasSelectedAtLeastOneCondition(this.state.selection),
+      accelerator: getShortcutDisplayName(
+        this.props.preferences.values.userShortcutMap[
+          'TOGGLE_CONDITION_INVERTED'
+        ] || 'KeyJ'
+      ),
```
D8H, line 558: Was `N` as "negate" already taken?

AlexandreSi: No, but we never use this term in the IDE so I'm not sure it's the best. N is traditionally associated to New

PR #4335, opened 2022-09-28
Fix custom object renderer for buttons

D8H:
It's not a big extraction, but it's very safe and it will allow to see how buttons resizing works in the IDE with a nightly as I can't manage to build macOS version on the experimental branch.
The bonus is that the button won't appear broken in the IDE of previous GDevelop version when we launch it.

### DIFF
--- a/newIDE/app/src/ObjectsRendering/Renderers/RenderedCustomObjectInstance.js
+++ b/newIDE/app/src/ObjectsRendering/Renderers/RenderedCustomObjectInstance.js
@@ -3,6 +3,7 @@ import RenderedInstance from './RenderedInstance';
 import PixiResourcesLoader from '../../ObjectsRendering/PixiResourcesLoader';
 import ResourcesLoader from '../../ResourcesLoader';
 import ObjectsRenderingService from '../ObjectsRenderingService';
+import RenderedTextInstance from './RenderedTextInstance';
 import { mapReverseFor } from '../../Utils/MapFor';
 import * as PIXI from 'pixi.js-legacy';
 
@@ -13,10 +14,16 @@ const gd: libGDevelop = global.gd;
 class ChildInstance {
   x: number;
   y: number;
+  _hasCustomSize: boolean;
+  _customWidth: number;
+  _customHeight: number;
 
   constructor() {
     this.x = 0;
     this.y = 0;
+    this._customWidth = 0;
+    this._customHeight = 0;
+    this._hasCustomSize = false;
   }
 
   getX() {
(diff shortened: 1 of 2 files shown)


newIDE/app/src/ObjectsRendering/Renderers/RenderedCustomObjectInstance.js
```diff
@@ -161,21 +176,25 @@ export default class RenderedCustomObjectInstance extends RenderedInstance {
...
+          const renderer = ObjectsRenderingService.createNewInstanceRenderer(
             project,
             layout,
             // $FlowFixMe Use real object instances.
             childInstance,
             childObjectConfiguration,
             this._pixiObject
           );
+          if (renderer instanceof RenderedTextInstance) {
+            renderer._pixiObject.style.align = 'center';
```
4ian, line 195: Could this be done as a property of the text object?

newIDE/app/src/ObjectsRendering/Renderers/RenderedCustomObjectInstance.js
```diff
@@ -229,32 +248,43 @@ export default class RenderedCustomObjectInstance extends RenderedInstance {
...
+      : this.getDefaultHeight();
+
     for (
       let index = 0;
       index < this.childrenRenderedInstances.length;
       index++
     ) {
       const renderedInstance = this.childrenRenderedInstances[index];
       const childInstance = this.childrenInstances[index];
-      childInstance.x = (defaultWidth - renderedInstance.getDefaultWidth()) / 2;
```
4ian, line 239: For now we can do like this, but in the future we could probably add a setting to choose for each child object instances if they are stretched or centered when rendered in the scene editor.
Could help for some basic cases like buttons or a few other simple objects.

4ian (approved):

newIDE/app/src/ObjectsRendering/Renderers/RenderedCustomObjectInstance.js
D8H: This should probably be the default layout when there is no `initialInstances` defined.

4ian: Ok!

newIDE/app/src/ObjectsRendering/Renderers/RenderedCustomObjectInstance.js
4ian: Let's add a "TODO: this should be a property added to the text object. When this property is added, remove these special cases".

Or if it's really not too complex => add it to the text + TextObject editor.

PR #4336, opened 2022-09-28
Ensure required behavior properties can't be hidden.

D8H:
Required behavior properties must not be hidden because extensions users need to understand why additional behaviors are added automatically and it doesn't work anyway because no values can be set on hidden properties even by the IDE itself.

It confused some extension creators that had an exception at preview because of this.


tristanbob:
Thank you @D8H !

4ian (approved):
Lgtm! If there are other similar "never private" properties in the future it should be in a function/array so that we can consolidate the list in a single place.

PR #4341, opened 2022-09-29
Add an extension category filter

D8H:
The button has a bigger height than the search bar.
Is there a way to change this?

http://gdevelop-storybook.s3-website-us-east-1.amazonaws.com/extension-category-filter/latest/index.html?path=/story/assetstore-extensionstore-extensionsearchdialog--default

### DIFF
--- a/Extensions/TileMap/tests/tilemapcollisionmaskruntimeobject.spec.js
+++ b/Extensions/TileMap/tests/tilemapcollisionmaskruntimeobject.spec.js
@@ -121,10 +121,10 @@ describe('gdjs.TileMapCollisionMaskRuntimeObject', function () {
       // TODO find a clean way to wait for the json to be read.
       for (
         let index = 0;
-        index < 400 && tileMap._collisionTileMap.getDimensionX() === 0;
+        index < 200 && tileMap._collisionTileMap.getDimensionX() === 0;
         index++
       ) {
-        await delay(5);
+        await delay(10);
       }
       if (tileMap._collisionTileMap.getDimensionX() === 0) {
         throw new Error('Timeout reading the tile map JSON file.');
--- a/Extensions/TileMap/tests/tilemapcollisionmaskruntimeobject.spec.js
+++ b/Extensions/TileMap/tests/tilemapcollisionmaskruntimeobject.spec.js
@@ -121,7 +121,7 @@ describe('gdjs.TileMapCollisionMaskRuntimeObject', function () {
       // TODO find a clean way to wait for the json to be read.
       for (
         let index = 0;
-        index < 200 && tileMap._collisionTileMap.getDimensionX() === 0;
+        index < 400 && tileMap._collisionTileMap.getDimensionX() === 0;
         index++
       ) {
         await delay(5);
--- a/newIDE/app/src/AssetStore/ExtensionStore/ExtensionStoreContext.js
+++ b/newIDE/app/src/AssetStore/ExtensionStore/ExtensionStoreContext.js
@@ -28,6 +28,9 @@ type ExtensionStoreState = {|
   error: ?Error,
   searchText: string,
   setSearchText: string => void,
+  allCategories: string[],
+  chosenCategory: string,
+  setChosenCategory: string => void,
   extensionShortHeadersByName: { [name: string]: ExtensionShortHeader },
   filtersState: FiltersState,
 |};
@@ -39,6 +42,9 @@ export const ExtensionStoreContext = React.createContext<ExtensionStoreState>({
   error: null,
   searchText: '',
   setSearchText: () => {},
+  allCategories: [],
+  chosenCategory: '',
+  setChosenCategory: () => {},
   extensionShortHeadersByName: {},
   filtersState: {
     chosenFilters: new Set(),
@@ -67,6 +73,7 @@ export const ExtensionStoreStateProvider = ({
   const preferences = React.useContext(PreferencesContext);
   const { showCommunityExtensions } = preferences.values;
   const [filters, setFilters] = React.useState<?Filters>(null);
+  const [allCategories, setAllCategories] = React.useState<Array<string>>([]);
   const [firstExtensionIds, setFirstExtensionIds] = React.useState<
     Array<string>
   >([]);
@@ -76,6 +83,7 @@ export const ExtensionStoreStateProvider = ({
   const [searchText, setSearchText] = React.useState(
     defaultSearchText || emptySearchText
   );
+  const [chosenCategory, setChosenCategory] = React.useState('');
   const filtersState = useFilters();
 
   const fetchExtensionsAndFilters = React.useCallback(
@@ -91,9 +99,18 @@ export const ExtensionStoreStateProvider = ({
 
         try {
           const extensionRegistry: ExtensionsRegistry = await getExtensionsRegistry();
-          const { extensionShortHeaders, allTags } = extensionRegistry;
+          const {
+            extensionShortHeaders,
+            allTags,
+            allCategories,
+          } = extensionRegistry;
 
           const sortedTags = allTags
+            .slice()
+            .sort((tag1, tag2) =>
+              tag1.toLowerCase().localeCompare(tag2.toLowerCase())
+            );
+          const sortedCategories = allCategories
             .slice()
             .sort((tag1, tag2) =>
               tag1.toLowerCase().localeCompare(tag2.toLowerCase())
@@ -115,6 +132,7 @@ export const ExtensionStoreStateProvider = ({
             defaultTags: sortedTags,
             tagsTree: [],
           });
+          setAllCategories(sortedCategories);
           setFirstExtensionIds(
             extensionRegistry.views
               ? extensionRegistry.views.default.firstExtensionIds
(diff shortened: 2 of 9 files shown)


Extensions/TileMap/tests/tilemapcollisionmaskruntimeobject.spec.js
```diff
@@ -124,7 +124,7 @@ describe('gdjs.TileMapCollisionMaskRuntimeObject', function () {
         index < 200 && tileMap._collisionTileMap.getDimensionX() === 0;
         index++
       ) {
-        await delay(5);
+        await delay(10);
```
AlexandreSi, line 127: The delay was not long enough?

newIDE/app/src/AssetStore/ExtensionStore/ExtensionStoreContext.js
```diff
@@ -39,6 +42,9 @@ export const ExtensionStoreContext = React.createContext<ExtensionStoreState>({
   error: null,
   searchText: '',
   setSearchText: () => {},
+  allCategories: [],
+  chosenCategory: '',
```
AlexandreSi, line 46: Should it be null so that it's clear nothing has been chosen?

newIDE/app/src/AssetStore/ExtensionStore/ExtensionStoreContext.js
```diff
@@ -39,6 +42,9 @@ export const ExtensionStoreContext = React.createContext<ExtensionStoreState>({
   error: null,
   searchText: '',
   setSearchText: () => {},
+  allCategories: [],
```
AlexandreSi, line 45: Should it be null so that it's clear categories were not fetched yet

4ian:
I think we might need *exceptionally* a new UI component: "SearchBarSelectField".
It's prefixed by "SearchBar" because it must be used next to a search bar and would look the same. Here is a quick mockup by fiddling around with the HTML:

image

Maybe you can ask @LuniMoon for a better idea though!

newIDE/app/src/AssetStore/ExtensionStore/index.js
```diff
@@ -94,14 +101,30 @@ export const ExtensionStore = ({
...
+                    setChosenCategory(value);
+                  }}
+                  disableUnderline
+                >
+                  <SelectOption value="" primaryText={t`All categories`} />
+                  {allCategories.map(category => (
+                    <SelectOption value={category} primaryText={category} />
+                  ))}
+                </SelectField>
+                <Column expand>
```
AlexandreSi, line 117: When the screen is small, the 2 fields are not aligned since this column adds horizontal margins.

newIDE/app/src/AssetStore/ExtensionStore/index.js
```diff
@@ -94,14 +101,30 @@ export const ExtensionStore = ({
...
+                <SelectField
+                  value={chosenCategory}
+                  onChange={(e, i, value: string) => {
+                    setChosenCategory(value);
+                  }}
+                  disableUnderline
+                >
+                  <SelectOption value="" primaryText={t`All categories`} />
+                  {allCategories.map(category => (
+                    <SelectOption value={category} primaryText={category} />
```
AlexandreSi, line 114: You should have a warning in the console stating that each `SelectOption` should have a unique key

AlexandreSi: Setting `key={category}` should do the job

newIDE/app/src/UI/Search/UseSearchStructuredItem.js
```diff
@@ -86,6 +89,12 @@ export const filterSearchResults = <
 
   const startTime = performance.now();
   const filteredSearchResults = searchResults
+    .filter(
+      ({ item }) =>
+        !itemCategory ||
+        item.category === itemCategory ||
+        (itemCategory === 'General' && !item.category)
```
AlexandreSi, line 96: This is a bit hard to read because I can't tell the difference between `item.category` and `itemCategory`. I feel like the name of `itemCategory` should be a bit more specific: `chosenItemCategory`? `requestedItemCategory`?

newIDE/app/src/fixtures/GDevelopServicesTestData/FakeExtensionsRegistry.js
```diff
@@ -1,6 +1,11 @@
+// @flow
```
AlexandreSi, line 1: 💜

newIDE/app/src/fixtures/GDevelopServicesTestData/FakeExtensionsRegistry.js
```diff
@@ -967,12 +1151,17 @@ export const fakeExtensionsRegistry: ExtensionsRegistry = {
       headerUrl:
         'https://resources.gdevelop-app.com/extensions/CurrentGameVersion-header.json',
       tags: 'version',
+      category: 'Advance',
```
AlexandreSi, line 1154: Great mock data for the stories!

AlexandreSi (commented):
I added a few comments on readability since I feel like I would be afraid to bring changes to extensions registry or filtering logic if I were to.

Extensions/TileMap/tests/tilemapcollisionmaskruntimeobject.spec.js
D8H: Yes, it's to fix the tests for the CI.

newIDE/app/src/AssetStore/ExtensionStore/ExtensionStoreContext.js
AlexandreSi: Alright. I don't know if it needs a comment then here

newIDE/app/src/AssetStore/ExtensionStore/ExtensionStoreContext.js
D8H: It would force to add a check when building the drop-down list.
If we don't display the drop-down list while the registry is retried, won't that makes the UI jumpy?

LuniMoon:
I've just created a trello card on UI Clarity to standardize search bar design and interaction, actually.
I think the new component is good enough. It could be reused if needed somewhere else.

I did these for reference if needed: 
Screenshot 2022-09-30 at 12 29 55
_(App Design system & Resources > 2. Components > Testing > Forms & Inputs)_

newIDE/app/src/UI/Search/UseSearchStructuredItem.js
```diff
@@ -86,6 +89,12 @@ export const filterSearchResults = <
 
   const startTime = performance.now();
   const filteredSearchResults = searchResults
+    .filter(
+      ({ item }) =>
+        !chosenItemCategory ||
+        item.category === chosenItemCategory ||
+        (chosenItemCategory === 'General' && !item.category)
+    )
```
D8H, line 97: It's probably a bad idea to use `'General'` here as it's common to other searches like examples and resources.

newIDE/app/src/AssetStore/ExtensionStore/ExtensionStoreContext.js
D8H: I added one.

newIDE/app/src/fixtures/GDevelopServicesTestData/FakeExtensionsRegistry.js
4ian: Nice!

newIDE/app/src/fixtures/GDevelopServicesTestData/FakeExtensionsRegistry.js
4ian: Is the typo made on purpose?

newIDE/app/src/UI/SearchBarSelectField.js
```diff
@@ -0,0 +1,173 @@
...
+ * A select field based on Material-UI select field.
+ * To be used with `SelectOption`.
+ */
+const SearchBarSelectField = React.forwardRef<
+  Props,
+  SearchBarSelectFieldInterface
+>((props, ref) => {
+  const inputRef = React.useRef<?HTMLInputElement>(null);
+
+  const focus = () => {
```
4ian, line 83: I guess this could use React.useCallback

newIDE/app/src/AssetStore/ExtensionStore/index.js
```diff
@@ -94,14 +101,33 @@ export const ExtensionStore = ({
...
             <ColumnStackLayout>
-              <SearchBar
-                value={searchText}
-                onChange={setSearchText}
-                onRequestSearch={() => {}}
-                tagsHandler={tagsHandler}
-                tags={filters && filters.allTags}
-                placeholder={t`Search extensions`}
-              />
+              <ResponsiveLineStackLayout noMargin>
```
4ian, line 104: Everyday I thank ResponsiveLineStackLayout for being there for us.

newIDE/app/src/AssetStore/ExtensionStore/ExtensionStoreContext.js
4ian: We could still have it null and display a SelectField that has a single, disabled option called "Loading..." (I agree with you that we don't want a jumpy UI).

BUT as this is maybe more work for not a lot (and we can do this later), I'm ok to avoid this (other components are taking care of showing a proper loader).

4ian (approved):

D8H:
Tested with a category and a tag

image

PR #4346, opened 2022-09-30
Add expression and condition to get highest z order of a layer

### LINKED ISSUE #4298: Fetching the highest active Z-order of objects in the scene.
**Description**
We need to have a function to fetch the highest Z-order value in the scene.

Many times in our games we've been needing to get know the highest Z-order value of current objects in the scene to determine some effects accurate and correct placement in the Z-order list.

- We are forced either to keep track with a variables every time when creating an object in the events to keep track of the highest Z-order value.
- Or to scan through every object to determine the highest object Z-order value (really heavy to do every single frame).
- Or create an dummy object and getting it's value and store it in a variable to use on the next event every time we need to know the highest value.

**Solution**
Since creating new objects in the events, the engine already knows the highest Z-order value and makes the new object to have the highest Z-order value when creating a new object.

It would be immensely helpful to be able to call the highest Z-order value with one function to use when needed.

AlexandreSi:
Resolves #4298 
Add expression and condition to get a layer highest z order.
Defaults to 0 if no object on the layer.

https://user-images.githubusercontent.com/32449369/193551117-82808488-fe63-44a7-b5ec-7ee4d8aa60d5.mov


GDJS/Runtime/runtimescene.ts
```diff
@@ -732,6 +737,8 @@ namespace gdjs {
      * Tool function filling _allInstancesList member with all the living object instances.
      */
     _constructListOfAllInstances() {
+      this._layerHighestZOrders.clear();
```
ClementPasteau, line 740: is this required?

ClementPasteau (commented):
lgtm! 
Should we add some simple unit tests on this in the runtimescene?

GDJS/Runtime/runtimescene.ts
AlexandreSi: Yes, if a layer has a highest z order (let's say 25) but at the next frame, no instance is on the layer, this value won't change whereas it should not be defined.

AlexandreSi:
I added an integration test if you want to have a look

4ian:
@D8H Could you review this as this will impact your work on Custom objects + see if you think it's properly implemented?

GDJS/Runtime/runtimescene.ts
```diff
@@ -62,6 +66,7 @@ namespace gdjs {
     constructor(runtimeGame: gdjs.RuntimeGame) {
       this._instances = new Hashtable();
       this._instancesCache = new Hashtable();
+      this._layerHighestZOrders = new Hashtable();
```
D8H, line 69: Is it indexed on layer names?
Why not add the attribute in the layer class?

GDJS/Runtime/events-tools/cameratools.ts
```diff
@@ -450,6 +450,16 @@ namespace gdjs {
         return runtimeScene.getLayer(layer).getDefaultZOrder();
       };
 
+      export const getLayerHighestZOrder = function (
```
D8H, line 453: I understand why it was added here, but it's not related to camera.
Should we add a `layertools.ts` with only this function for now?

GDJS/Runtime/runtimescene.ts
```diff
@@ -14,13 +14,17 @@ namespace gdjs {
     _eventsFunction: null | ((runtimeScene: RuntimeScene) => void) = null;
     _instances: Hashtable<RuntimeObject[]>;
 
-    //Contains the instances living on the scene
+    // Contains the instances living on the scene
     _instancesCache: Hashtable<RuntimeObject[]>;
```
D8H, line 18: The comments are not at the right place. I reorganized them in the custom PR. Avoid to change them to reduce conflicts.

GDJS/Runtime/runtimescene.ts
```diff
@@ -1158,7 +1187,7 @@ namespace gdjs {
      * in exceptional use cases where you need to loop through all objects,
      * and it won't be performant.
      *
-     * @returns The list of all runtime objects on the scnee
+     * @returns The list of all runtime objects on the scene
```
D8H, line 1190: This typo won't exist soon.

GDJS/Runtime/runtimescene.ts
```diff
@@ -1067,6 +1082,20 @@ namespace gdjs {
...
+     * @returns The highest z order across the layer's object instances
+     */
+    getLayerHighestZOrder(layerName: string): number {
+      if (!this._layerHighestZOrders.containsKey(layerName)) return 0;
+      return this._layerHighestZOrders.get(layerName);
+    }
+
+    _setLayerHighestZOrder(layerName: string, zOrder: number) {
+      this._layerHighestZOrders.put(layerName, zOrder);
+    }
```
D8H, line 1097: Move it to Layer?

GDJS/Runtime/runtimescene.ts
AlexandreSi: Alright, I'll remove it to avoid conflicts

GDJS/Runtime/runtimescene.ts
AlexandreSi: Alright, I'll remove them to avoid conflicts

GDJS/Runtime/events-tools/cameratools.ts
AlexandreSi: You're right, I'll move it

D8H: Sorry, actually, there is already a lot of function unrelated to camera so it can stay here.
The file should probably be renamed one day.

AlexandreSi:
@D8H I move the logic to the layer class. Let me know if it's okay with you

GDJS/Runtime/events-tools/layertools.ts
```diff
@@ -0,0 +1,20 @@
+/*
+ * GDevelop JS Platform
+ * Copyright 2013-2016 Florian Rival (Florian.Rival@gmail.com). All rights reserved.
+ * This project is released under the MIT License.
+ */
+namespace gdjs {
+  export namespace evtTools {
+    export namespace layer {
```
D8H, line 8: Sorry about this, actually, there is already a lot of things in cameratools.ts that are not related to camera so maybe it should just be there.

Core/GDCore/Extensions/Builtin/CameraExtension.cpp
```diff
@@ -589,6 +589,19 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsCameraExtension(
       .SetDefaultValue("\"\"")
       .AddParameter("expression", _("New default Z order"));
 
+  extension
+      .AddExpressionAndCondition("number",
+                                 "LayerHighestZOrder",
+                                 _("Highest Z order of a layer"),
+                                 _("the highest Z order of a layer"),
+                                 _("the highest Z order of layer _PARAM1_"),
```
D8H, line 598: "highest Z order of objects in the layer" ?

GDJS/Runtime/runtimeobject.ts
```diff
@@ -714,10 +714,14 @@ namespace gdjs {
       }
       this.zOrder = z;
       const rendererObject = this.getRendererObject();
+      const theLayer = this._runtimeScene.getLayer(this.layer);
       if (rendererObject) {
-        const theLayer = this._runtimeScene.getLayer(this.layer);
         theLayer.getRenderer().changeRendererObjectZOrder(rendererObject, z);
       }
+      const layerHighestZOrder = theLayer.getHighestZOrder();
```
D8H, line 721: This variable can be inlined.

GDJS/Runtime/runtimescene.ts
```diff
@@ -732,6 +732,10 @@ namespace gdjs {
      * Tool function filling _allInstancesList member with all the living object instances.
      */
     _constructListOfAllInstances() {
+      const layers = gdjs.staticArray(gdjs.RuntimeScene);
+      this._layers.values(layers);
+      layers.forEach((layer) => layer._setHighestZOrder(0));
```
D8H, line 737: Why not iterate on the object attributes directly with a `for in`?

GDJS/Runtime/runtimescene.ts
```diff
@@ -744,6 +748,14 @@ namespace gdjs {
             } else {
               this._allInstancesList.push(list[j]);
             }
+            const layerName = list[j].getLayer();
+            const zOrder = list[j].getZOrder();
```
D8H, line 752: Maybe `list[j]` should be set in a local variable because it's used 3 times.

GDJS/Runtime/runtimescene.ts
```diff
@@ -744,6 +748,14 @@ namespace gdjs {
...
+            if (
+              !this.hasLayer(layerName) ||
+              this.getLayer(layerName).getHighestZOrder() < zOrder
```
D8H, line 755: We can be confident that `getLayer` gives the right layer.

GDJS/tests/tests/runtimescene.js
```diff
@@ -83,100 +83,155 @@ describe('gdjs.RuntimeScene integration tests', function () {
...
+      expect(runtimeScene.hasLayer('')).to.be(true);
+      expect(runtimeScene.hasLayer('MyLayer')).to.be(true);
+
+      const object1 = runtimeScene.createObject('MyObject');
+      const object2 = runtimeScene.createObject('MyObject');
+      const object3 = runtimeScene.createObject('MyObject');
+      if (!object1 || !object2 || !object3) {
+        throw new Error('object should have been created');
+      }
+      expect(object2.getZOrder()).to.be(0);
```
D8H, line 201: It this useful for the test?

GDJS/tests/tests/runtimescene.js
```diff
@@ -83,100 +83,155 @@ describe('gdjs.RuntimeScene integration tests', function () {
...
+      object2.setLayer('MyLayer');
+
+      // Layers highest Z orders should stay at 0 if objects did not change Z their order
```
D8H, line 204: Typo "their Z"

GDJS/Runtime/runtimescene.ts
AlexandreSi: Good idea!

GDJS/Runtime/runtimescene.ts
AlexandreSi: You're right, one check fewer!

GDJS/tests/tests/runtimescene.js
```diff
@@ -83,100 +83,155 @@ describe('gdjs.RuntimeScene integration tests', function () {
...
+
+      // Check highest Z order has been correctly updated for layers and they are stable
+      runtimeScene._updateObjectsPreRender();
+      expect(runtimeScene.getLayer('MyLayer').getHighestZOrder()).to.be(8);
+      expect(runtimeScene.getLayer('').getHighestZOrder()).to.be(0);
+      runtimeScene._updateObjectsPreRender();
+      expect(runtimeScene.getLayer('MyLayer').getHighestZOrder()).to.be(8);
+      expect(runtimeScene.getLayer('').getHighestZOrder()).to.be(0);
+
+      // Change Z orders for default layer
```
D8H, line 219: Layers don't have Z order.
"Change default layer objects Z orders"?

GDJS/tests/tests/runtimescene.js
```diff
@@ -83,100 +83,155 @@ describe('gdjs.RuntimeScene integration tests', function () {
...
+      object3.setZOrder(25);
+
+      runtimeScene._updateObjectsPreRender();
+      expect(runtimeScene.getLayer('MyLayer').getHighestZOrder()).to.be(8);
+      expect(runtimeScene.getLayer('').getHighestZOrder()).to.be(25);
+
+      // Check highest z orders come back to 0 after objects deletion
+      runtimeScene.markObjectForDeletion(object1);
+      runtimeScene.markObjectForDeletion(object2);
+      runtimeScene.markObjectForDeletion(object3);
```
D8H, line 230: It could be interesting to only remove the object with Z order = 25 first.

What about moving one instance in front of the top one?

AlexandreSi:
I updated the tests with your suggestions. Let me know if you feel like another test case is needed!

Core/GDCore/Extensions/Builtin/CameraExtension.cpp
```diff
@@ -590,6 +590,19 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsCameraExtension(
       .SetDefaultValue("\"\"")
       .AddParameter("expression", _("New default Z order"));
 
+  extension
+      .AddExpressionAndCondition("number",
+                                 "LayerHighestZOrder",
+                                 _("Highest Z order of a layer"),
+                                 _("the highest Z order of a layer"),
+                                 _("the highest Z order of objects in the layer _PARAM1_"),
```
D8H, line 598:
```suggestion
                                 _("the highest Z order of objects in the layer"),
                                 _("the highest Z order of objects in the layer _PARAM1_"),
```

Core/GDCore/Extensions/Builtin/CameraExtension.cpp
```diff
@@ -590,6 +590,19 @@ void GD_CORE_API BuiltinExtensionsImplementer::ImplementsCameraExtension(
       .SetDefaultValue("\"\"")
       .AddParameter("expression", _("New default Z order"));
 
+  extension
+      .AddExpressionAndCondition("number",
+                                 "LayerHighestZOrder",
+                                 _("Highest Z order of a layer"),
```
D8H, line 596: The full name must be as concise as possible because it's used in a list.
```suggestion
                                 _("Layer highest Z order"),
```

GDJS/Runtime/runtimescene.ts
```diff
@@ -732,17 +732,29 @@ namespace gdjs {
      * Tool function filling _allInstancesList member with all the living object instances.
      */
     _constructListOfAllInstances() {
+      for (const name in this._layers.items) {
+        if (this._layers.containsKey(name)) {
```
D8H, line 736: This check is not necessary because the object of `_layers` is local to RuntimeScene.
We should remove the other checks, but I'll do it in my PR.

Core/GDCore/Extensions/Builtin/CameraExtension.cpp
AlexandreSi: I prefer keeping `a layer` in this sentence, because of the phrasing here:
image
But I'll add `objects in`

AlexandreSi:
It should be ok now! 😅 
I always feel like I'm not thorough enough when touching the runtime

D8H (approved):

D8H:
One test fail because of a change I made this morning, but it's ok.

PR #4349, opened 2022-10-03
Add a test on sprite hit-boxes after a camera displacement.

D8H:
The test is dumb for the current version but it will be less obvious with custom objects.

### DIFF
--- a/GDJS/tests/tests/spriteruntimeobject.pixiruntimegamewithassets.js
+++ b/GDJS/tests/tests/spriteruntimeobject.pixiruntimegamewithassets.js
@@ -5,7 +5,7 @@
  * * GDJS/tests/games/rotate-flip-around-center-point/ship-rotate-flip.json
  * * GDJS/tests/games/rotated-objects-hitboxes/game.json
  */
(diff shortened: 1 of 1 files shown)


GDJS/tests/tests/spriteruntimeobject.pixiruntimegamewithassets.js
```diff
@@ -388,4 +388,53 @@ describe('gdjs.SpriteRuntimeObject (using a PixiJS RuntimeGame with assets)', fu
       expect(object.getPointY("Origin")).to.be(0);
     });
   });
+
+  it('properly computes hitboxes and point positions after the layer camera has moved', function() {
+    return gdjs.getPixiRuntimeGameWithAssets().then(runtimeGame => {
```
4ian, line 393: nit: We can probably start using async/await in runtime tests:

```suggestion
  it('properly computes hitboxes and point positions after the layer camera has moved', async () => {
    const runtimeGame = await gdjs.getPixiRuntimeGameWithAssets();
```

4ian (approved):

PR #4350, opened 2022-10-03
Add price to assets home private asset packs thumbnails and display them in their dialog

AlexandreSi:
Create a new component to display price => [story](https://gdevelop-storybook.s3.amazonaws.com/add-private-asset-packs-prices/latest/index.html?path=/story/ui-building-blocks-pricetag--default)

image

image

### DIFF
--- a/newIDE/app/src/AssetStore/index.js
+++ b/newIDE/app/src/AssetStore/index.js
@@ -303,7 +303,7 @@ export const AssetStore = ({ project }: Props) => {
                   />
                 )}
                 {isOnHomePage && error && (
-                  <PlaceholderError>
+                  <PlaceholderError onRetry={fetchAssetsAndFilters}>
                     <AlertMessage kind="error">
                       <Trans>
                         An error occurred when fetching the asset store content.
--- a/newIDE/app/src/AssetStore/index.js
+++ b/newIDE/app/src/AssetStore/index.js
@@ -34,6 +34,8 @@ import { AssetDetails } from './AssetDetails';
 import PlaceholderLoader from '../UI/PlaceholderLoader';
 import Home from '@material-ui/icons/Home';
 import PrivateAssetPackDialog from './PrivateAssetPackDialog';
+import PlaceholderError from '../UI/PlaceholderError';
+import AlertMessage from '../UI/AlertMessage';
 
 type Props = {|
   project: gdProject,
@@ -265,9 +267,9 @@ export const AssetStore = ({ project }: Props) => {
                     )}
                   </Background>
                 )}
-                {isOnHomePage && !(assetPacks && privateAssetPacks) && (
-                  <PlaceholderLoader />
-                )}
+                {isOnHomePage &&
+                  !(assetPacks && privateAssetPacks) &&
+                  !error && <PlaceholderLoader />}
                 {isOnHomePage &&
                   assetPacks &&
                   privateAssetPacks &&
@@ -299,6 +301,16 @@ export const AssetStore = ({ project }: Props) => {
                       />
                     }
                   />
+                )}
+                {isOnHomePage && error && (
+                  <PlaceholderError>
+                    <AlertMessage kind="error">
+                      <Trans>
+                        An error occurred when fetching the asset store content.
+                        Please try again later.
+                      </Trans>
+                    </AlertMessage>
+                  </PlaceholderError>
                 )}
                 {openedAssetShortHeader && (
                   <AssetDetails
--- a/newIDE/app/src/UI/BackgroundText.js
+++ b/newIDE/app/src/UI/BackgroundText.js
@@ -16,6 +16,7 @@ const BackgroundText = (props: Props) => (
       <Typography
         variant="body2"
         align="center"
+        component="div"
         style={{
           opacity: 0.6,
           textShadow: `1px 1px 0px ${muiTheme.emptyMessage.shadowColor}`,
--- a/newIDE/app/src/UI/PriceTag.js
+++ b/newIDE/app/src/UI/PriceTag.js
@@ -11,13 +11,18 @@ type Props = {|
   /**
    * To be used when the component is over an element for which
    * we don't control the background (e.g. an image).
-   * This component should logically be invariant to selected theme
-   * when this option is activated
    */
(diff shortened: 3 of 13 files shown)


newIDE/app/src/UI/PriceTag.js
```diff
@@ -0,0 +1,64 @@
...
+import Text from './Text';
+import { makeStyles } from '@material-ui/core';
+
+type Props = {|
+  value: number,
+  /**
+   * To be used when the component is over an element for which
+   * we don't control the background (e.g. an image).
+   * This component should logically be invariant to selected theme
+   * when this option is activated
```
4ian, line 15: Not sure I totally understood the comment

newIDE/app/src/UI/BackgroundText.js
```diff
@@ -16,6 +16,7 @@ const BackgroundText = (props: Props) => (
       <Typography
         variant="body2"
         align="center"
+        component="div"
```
4ian, line 19: Nice!

D8H:
How does the description looks on a small screen?

newIDE/app/src/UI/PriceTag.js
```diff
@@ -0,0 +1,64 @@
...
+      {({ i18n }) => (
+        <div className={classes.container}>
+          <Text noMargin size="sub-title" color="inherit">
+            {i18n
+              .number(value / 100, {
+                minimumFractionDigits: 2,
+                maximumFractionDigits: 2,
+              })
+              .replace(/\D00$/, '')}
+            EUR
```
4ian, line 56: I think the order is usually EUR 100. We could also use the symbol € => €10?

4ian (approved):

newIDE/app/src/UI/PriceTag.js
AlexandreSi: This is for @LuniMoon ;)

newIDE/app/src/UI/PriceTag.js
AlexandreSi: I'll clarify

AlexandreSi:
If you want to make your own tests => https://gdevelop-storybook.s3.amazonaws.com/add-private-asset-packs-prices/latest/index.html?path=/story/assetstore-assetstore-privateassetpackdialog--default

Otherwise, here it is:
image

newIDE/app/src/AssetStore/index.js
```diff
@@ -300,6 +302,16 @@ export const AssetStore = ({ project }: Props) => {
                     }
                   />
                 )}
+                {isOnHomePage && error && (
+                  <PlaceholderError>
```
ClementPasteau, line 306: should we pass an `onRetry` to the component with the function `fetchAssetsAndFilters`

newIDE/app/src/UI/PriceTag.js
```diff
@@ -0,0 +1,69 @@
...
+      borderRadius: 4,
+      padding: '2px 4px',
+      backdropFilter: props =>
+        props.withOverlay && theme.palette.type === 'light'
+          ? 'brightness(40%)'
+          : undefined,
+      backgroundColor: props =>
+        props.withOverlay && theme.palette.type === 'light'
+          ? undefined
+          : theme.palette.background.alternate.startsWith('#')
```
ClementPasteau, line 37: hmmm, that feels like a hack 😅 
Should we try to have all colors in the same format in the themes (start with '#')?

newIDE/app/src/UI/PriceTag.js
```diff
@@ -0,0 +1,69 @@
...
+          ? theme.palette.background.alternate + 'BB' // manually add opacity to the background hex color
+          : theme.palette.background.alternate,
+      color: props =>
+        props.withOverlay && theme.palette.type === 'light'
+          ? '#FAFAFA'
```
ClementPasteau, line 42: I don't mind having a color here, but wondering if we should avoid that and put everything in the theme.json files?

ClementPasteau (approved):
few comments/questions, but lgtm!

newIDE/app/src/UI/PriceTag.js
ClementPasteau: Agree that the opposite order is what I've been used to, but to check with Luni

newIDE/app/src/AssetStore/index.js
AlexandreSi: Good idea! I'll add it

newIDE/app/src/UI/PriceTag.js
LuniMoon: Sorry! that price got lost in translation hehe.

I've just bench a little here and there.

**Amazon**:
Screenshot 2022-10-04 at 11 00 37

**Google Search** results:
Screenshot 2022-10-04 at 11 03 31

**United Nude** (shoes):
Screenshot 2022-10-04 at 11 02 40

**Quicksilver** (clothes):
Screenshot 2022-10-04 at 11 03 11

**Vans** (clothes):
Screenshot 2022-10-04 at 11 04 13

**KLM** (flights)
Screenshot 2022-10-04 at 11 05 01

**I love NY** official store:
Screenshot 2022-10-04 at 11 10 49


So, let's go with the majority, and **keep the "€" at the beginning of the price**.
As for the decimals, we'll use 2 (if needed), and I personally have a preference for a dot separating the cents from the units (I'm from the american school), but I dunno if that's good practice in Europe or the rest of the world (?).
I do see sites using commas and dots interchangeably...

LuniMoon: Note on the side: In my context, **commas are used for big numbers like "2,599" and dots for cents "2.59"**.

But I do know that depending on the language, numbers are written differently (each language has certains rules for using commas and dots, and sometimes they vary from each other... I do have some books as reference if you wanna read about it lol).

AlexandreSi: > I personally have a preference for a dot separating the cents from the units

This should be handled by the native internationalization library.
In French, this is really weird for non-english speaking people to see a price like `1,255.45`. I guess other countries have the same kind of issues.

PR #4352, opened 2022-10-04
Add parameter to file reading actions to remove CR characters from files

### LINKED ISSUE #4246: New Lines/NewLine() Expression are treated differently if stored in variables loaded from external text files.
## Describe the bug
Bluzima on the forums [here](https://forum.gdevelop.io/t/issue-searching-for-newline-character-with-strfind/41814) is trying to load data from a text file into their game, then test if there is a NewLine in it to do an action.
They are using the expression StrFind(VariableString(Loadedtextvarhere),"*" + NewLine()), which should test fine if the text file has any instance of
"whateverprecedingcharactersyouwanthere *
(newlinehere)
whateverelseyouwanthere"

However, if loaded from a text file and stored into memory, while it will display correctly if you load that string variable into a text object, it *will not* evaluate correctly. The same is true if you try to use the TextObject.String() expression to test against, instead of a variable.

What *does* work is if you manually populate a string variable with the text I pasted above. It will evaluate fine.

## To Reproduce
1. Create a text file with whatever characters you want, followed by a new line (return/enter key).
2. Save the text file somewhere you can easily build the Expression-based path to in GDevelop.
3. In a new GDevelop project, load that text file into a variable using the "Load text from _ into scene variable _"
4. Set up a new event that uses the "Compare two values" condition, and tests StrFind(VariableString(YourVariable),"yourcharacter" + NewLine()) is greater than -1.
5.  Set up an action to do anything.
The action will not occur.

There are a few permutations to this, but things Bluzima and I have now tested:
1. String Find against a string variable loaded from a text file (above scenario) - Doesn't work
2. Modify the string of a text object with the variable from 1 above, then String Find against the TextObject.String() expression. - Doesn't work.
3. Create a string variable (via the variable editor or events) with the text copy+pasted from the text file, then String Find against that scene variable - *DOES WORK*
4. Modify the string of a text object with the variable from 3 above, then String Find against the TextObject.String() expression. -  *DOES WORK*.

Everything in red in the screenshot is done with the variable that was created by reading from the text file. Everything in blue is done with the variable that was manually created. 
image

Here's the example project. You will need to extract it to your desktop for the file path to work.
[NewLineTests.zip](https://github.com/4ian/GDevelop/files/9451486/NewLineTests.zip)

## Other details 
* Include any OS/browser version/smartphone that you're using - Win 10 64 
* Which version of GDevelop are you using? The desktop app or the web-app? - 5.0.140
* Add any other context about the problem here. - Forum thread here: https://forum.gdevelop.io/t/issue-searching-for-newline-character-with-strfind/41814/15

AlexandreSi:
Fixes #4246

### DIFF
--- a/Extensions/FileSystem/JsExtension.js
+++ b/Extensions/FileSystem/JsExtension.js
@@ -203,6 +203,15 @@ module.exports = {
         '',
         true
       )
+      .addParameter(
+        'yesorno',
+        _(
+          '(Advanced) Should clean file from superfluous character (\\r) present in files created on Windows (recommended).'
+        ),
+        '',
+        true
+      )
+      .setDefaultValue('yes')
       .getCodeExtraInformation()
       .setIncludeFile('Extensions/FileSystem/filesystemtools.js')
       .setFunctionName('gdjs.fileSystem.loadStringFromFileAsync');
@@ -229,6 +238,15 @@ module.exports = {
         '',
         true
       )
+      .addParameter(
+        'yesorno',
+        _(
+          '(Advanced) Should clean file from superfluous character (\\r) present in files created on Windows (recommended).'
+        ),
+        '',
+        true
+      )
+      .setDefaultValue('yes')
       .getCodeExtraInformation()
       .setIncludeFile('Extensions/FileSystem/filesystemtools.js')
       .setFunctionName('gdjs.fileSystem.loadStringFromFile');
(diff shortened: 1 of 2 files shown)


Extensions/FileSystem/JsExtension.js
```diff
@@ -203,6 +203,15 @@ module.exports = {
         '',
         true
       )
+      .addParameter(
+        'yesorno',
+        _(
+          '(Advanced) Should clean file from superfluous character (\\r) present in files created on Windows (recommended).'
```
4ian, line 209: See if using `setLongDescription` could make something a bit more readable:

```
.addParameter(..., _('Normalize the file content (recommended)')...)
.setLongDescription('This replaces Windows new lines characters ("CRLF") by a single new line character.')
```

4ian (approved):

Extensions/FileSystem/JsExtension.js
AlexandreSi: Good idea!

AlexandreSi:
Just a remark, I think this change could create regressions if people had conditions like:
`StrFind(text, "Hello" + "\r" + NewLine() + "there") >= 0`
 where they should have normally used:
`StrFind(text, "Hello" + NewLine() + "there") >= 0`
Do you think it's okay to do this?

PR #4354, opened 2022-10-04
Allow assets to declare extension dependencies without behaviors

D8H:


newIDE/app/src/AssetStore/InstallAsset.spec.js
```diff
@@ -730,5 +731,43 @@ describe('InstallAsset', () => {
           .toJSArray()
       ).toEqual(['MyBehavior']);
     });
+
+    // TODO EBO Add a test for a custom object that contains another custom objet.
+    // There is 2 cases:
```
4ian, line 736:
```suggestion
    // There are 2 cases:
```

newIDE/app/src/AssetStore/InstallAsset.spec.js
```diff
@@ -730,5 +731,43 @@ describe('InstallAsset', () => {
...
+    // - an event-based object from the same extension (this should already work).
+    // - an event-based object from another extension (this won't work because
+    //   it needs extension dependencies).
+
+    it('install an asset, with an event-based object that is already installed', async () => {
```
4ian, line 741: Should we have a test for "install an asset, with an event-based object that is not already installed" and check that the extension is installed?

Or did I miss it?

4ian (approved):

newIDE/app/src/AssetStore/InstallAsset.spec.js
D8H: The tests are unit-testing the function to get extension to install. Probably to avoid having to mock the call to extensions registry.

newIDE/app/src/Utils/GDevelopServices/Asset.js
```diff
@@ -25,12 +25,8 @@ export type AssetCustomization =
...
     |}
   | {|
-      required: boolean,
-      parameters: Array<SerializedParameterMetadata>,
-      extensions: Array<{|
-        extensionName: string,
-        extensionVersion: string,
-      |}>,
+      extensionName: string,
+      extensionVersion: string,
```
D8H, line 29: I wonder if `AssetCustomization` is still a good name.
Should behavior properties be in the serialized object? In this case, it would just be a list of required extensions.
Behaviors on assets is a dead code. Should it be removed?

4ian: I guess the idea with assets customisation was to offer OPTIONAL behaviors that would be interesting for the object... But this seems outdated/useless. 
So we could rename this (just be sure not to crash if the existing assets don't have the new property key).

4ian: I guess this can be.. requiredExtensions?

4ian (approved):

newIDE/app/src/Utils/GDevelopServices/Asset.js
D8H: ok, I renamed it. I'll have to report it to the asset repository too.

PR #4355, opened 2022-10-04
Allow using private assets in the asset store

ClementPasteau:
TODO:
- [x] Ensure we reset the storage provider when creating a blank project

## Desktop

https://user-images.githubusercontent.com/4895034/193864160-0171ba4a-0a59-411e-8730-382d548ad968.mov

## Asset Pack

https://user-images.githubusercontent.com/4895034/194103947-95240e86-e711-40a2-87cb-d751a59bd1dc.mov

## Web

https://user-images.githubusercontent.com/4895034/193864184-58cffff5-4189-43f6-a424-a37586ed3c02.mov

## Web 2

https://user-images.githubusercontent.com/4895034/193864197-3caa6889-cdcc-4944-a82d-371420515316.mov

## Web 3

image


newIDE/app/src/AssetStore/PrivateAssets/PrivateAssetsAuthorizationProvider.js
```diff
@@ -0,0 +1,151 @@
...
+    const userId = profile.id;
+
+    // If a request is already in progress, don't do anything.
+    if (isLoading.current) return;
+
+    try {
+      isLoading.current = true;
+      await fetchAuthorizationToken(userId);
+    } catch (error) {
+      console.error(error);
```
4ian, line 57: Don't forget the context. An error log = the context + the error.

newIDE/app/src/AssetStore/AssetPackDialog.js
```diff
@@ -66,16 +71,31 @@ export const AssetPackDialog = ({
...
+              : await installPublicAsset({
+                  assetShortHeader,
+                  eventsFunctionsExtensionsState,
+                  project,
+                  objectsContainer,
+                  environment,
+                });
+
+            if (!installOutput) {
+              throw new Error('Unable to install private Asset.');
```
4ian, line 92: The error message mentions a private asset but this can be an error linked to a public asset.

(nit: also, careful about not randomly capitalizing words in sentences.)

newIDE/app/src/AssetStore/AssetStoreContext.js
```diff
@@ -148,6 +149,7 @@ export const AssetStoreStateProvider = ({
   const [assetShortHeadersById, setAssetShortHeadersById] = React.useState<?{
     [string]: AssetShortHeader,
   }>(null);
+  const { ownedAssetShortHeaders } = React.useContext(AuthenticatedUserContext);
```
4ian, line 152: Should this be called ownedPrivateAssetShortHeaders?
(open question, I'm not sure it's better)

newIDE/app/src/AssetStore/AssetStoreContext.js
```diff
@@ -269,10 +276,23 @@ export const AssetStoreStateProvider = ({
         isLoading.current = false;
       })();
     },
-    [isLoading, environment]
+    [isLoading, environment, ownedAssetShortHeaders]
   );
 
-  // Preload the assets and filters when the app loads.
+  // We're listening to the owned assets changing to update the list of assets.
+  // This can happen when the user logs in or logs out.
```
4ian, line 283: Useful comments, thanks!

newIDE/app/src/MainFrame/index.js
```diff
@@ -2180,6 +2183,28 @@ const MainFrame = (props: Props) => {
     return true;
   };
 
+  const canInstallPrivateAsset = async (): Promise<boolean> => {
```
4ian, line 2186: Any chance this can be useCallback'ed to avoid this changing this at every render?

newIDE/app/src/MainFrame/index.js
```diff
@@ -2180,6 +2183,28 @@ const MainFrame = (props: Props) => {
...
+      if (!currentFileMetadata) {
+        return false;
+      }
+      const cloudProjectId = currentFileMetadata.fileIdentifier;
+
+      const cloudProject = await getCloudProject(
+        authenticatedUser,
+        cloudProjectId
+      );
+      return !!cloudProject;
```
4ian, line 2203: You should not need to do this. If the storageProvider is Cloud, the file was either loaded from the Cloud or saved as a cloud project. No need to do a `getCloudProject` request every time we need to know if we can install a private asset.

4ian: In other words, the storageProvider.internalName is the name of the storage provider for the SAVED project. This should be a guarantee. 
- On the desktop app, it's a LocalFile (or soon, a Cloud) storage provider. The file is already saved locally, or will be created in the Cloud.
- On the web app, it's either a UrlStorageProvider (in which case the file is "saved" in a url, so almost nothing is supported) or Cloud (in which case it's saved online).

newIDE/app/src/MainFrame/Providers.js
```diff
@@ -34,6 +34,7 @@ import { GamesShowcaseStateProvider } from '../GamesShowcase/GamesShowcaseContex
 import { TutorialStateProvider } from '../Tutorial/TutorialContext';
 import ConfirmProvider from '../UI/Confirm/ConfirmProvider';
 import { AnnouncementsFeedStateProvider } from '../AnnouncementsFeed/AnnouncementsFeedContext';
+import PrivateAssetsProvider from '../AssetStore/PrivateAssets/PrivateAssetsAuthorizationProvider';
```
4ian, line 37: The name of the default import should be the name of the file. Change either one or the other according to what makes sense.

newIDE/app/src/MainFrame/EditorContainers/BaseEditor.js
```diff
@@ -91,6 +91,7 @@ export type RenderEditorContainerProps = {|
     newName: string,
     cb: (boolean) => void
   ) => void,
+  canInstallPrivateAsset: () => Promise<boolean>,
```
4ian, line 94: See the other comment, this can be a simple `() => boolean`.

newIDE/app/src/AssetStore/PrivateAssets/PrivateAssetsAuthorizationProvider.js
```diff
@@ -0,0 +1,151 @@
...
+    try {
+      const asset = await getPrivateAsset(assetShortHeader, token, {
+        environment,
+      });
+      return asset;
+    } catch (error) {
+      if (error.response && error.response.status === 404) {
+        // If the token is expired, fetch a new one and try again.
+        token = await fetchAuthorizationToken(userId);
+        const asset = getPrivateAsset(assetShortHeader, token, { environment });
```
4ian, line 82: await

newIDE/app/src/AssetStore/PrivateAssets/PrivateAssetsAuthorizationProvider.js
```diff
@@ -0,0 +1,151 @@
...
+    if (!asset) {
+      throw new Error(
+        'Unable to install the asset because it could not be fetched.'
+      );
+    }
+
+    const token =
+      authorizationToken || (await fetchAuthorizationToken(profile.id));
+
+    const assetWithAuthorizedResourceUrls = {
```
4ian, line 115: This smells weird:
- this object is not typed
- the triple spread indicates something is too complex.
- only a single object asset is handled

A function is a good chance to name what you're doing here. What is the operation being done? Why is this necessary? The name of the function should tell that.

newIDE/app/src/AssetStore/PrivateAssets/PrivateAssetsAuthorizationProvider.js
```diff
@@ -0,0 +1,151 @@
...
+      objectsContainer,
+      environment,
+    });
+  };
+
+  return (
+    <PrivateAssetsAuthorizationContext.Provider
+      value={{
+        authorizationToken,
+        updateAuthorizationToken,
```
4ian, line 141: Isn't this a detail? Why is this necessary to communicate the authroization token/update method to children?

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/AnimationPreview.js
```diff
@@ -37,9 +39,11 @@ type Props = {|
   initialZoom?: number,
   fixedHeight?: number,
   fixedWidth?: number,
+  isPrivate?: boolean,
```
4ian, line 42: This is named too generically. What is private: the object, the sprite, the image, the project?

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/AnimationPreview.js
```diff
@@ -129,13 +149,19 @@ const AnimationPreview = ({
...
 
         currentFrameIndexRef.current = newFrameIndex;
         currentFrameElapsedTimeRef.current = newFrameElapsedTime;
         // Ensure we trigger an update if the animation changes,
         // as the refs will not do it.
         if (currentFrameIndex !== newFrameIndex) {
+          imagesLoadedArray.current[currentFrameIndexRef.current] = false;
+          // When the array of loaders changes, wait a bit to display the loader to avoid flickering.
+          loaderTimeout.current = setTimeout(() => {
+            console.warn('image took too long to load, displaying loader');
```
4ian, line 162: nit: Capital + period in sentences.

newIDE/app/src/Profile/AuthenticatedUserContext.js
```diff
@@ -17,6 +18,8 @@ export type AuthenticatedUser = {|
   loginState: null | 'loggingIn' | 'done',
   badges: ?Array<Badge>,
   cloudProjects: ?Array<CloudProjectWithUserAccessInfo>,
+  ownedAssetPackIds: ?Array<string>,
+  ownedAssetShortHeaders: ?Array<AssetShortHeader>,
```
4ian, line 22: Should this be:
```suggestion
  ownedPrivateAssetPackIds: ?Array<string>,
  ownedPrivateAssetShortHeaders: ?Array<AssetShortHeader>,
```

newIDE/app/src/Profile/AuthenticatedUserProvider.js
```diff
@@ -138,6 +140,8 @@ export default class AuthenticatedUserProvider extends React.Component<
         onAcceptGameStatsEmail: this._doAcceptGameStatsEmail,
         getAuthorizationHeader: () =>
           this.props.authentication.getAuthorizationHeader(),
+        ownedAssetPackIds: [], // Not using null in order to trigger an update of the assets context.
+        ownedAssetShortHeaders: [], // Not using null in order to trigger an update of the assets context.
```
AlexandreSi, line 144: I don't understand those comments, they seem to imply something like `lists trigger updates` but I don't see how

newIDE/app/src/Profile/AuthenticatedUserProvider.js
```diff
@@ -277,6 +281,36 @@ export default class AuthenticatedUserProvider extends React.Component<
...
+    }).then(
+      assetShortHeaders =>
+        this.setState(({ authenticatedUser }) => ({
+          authenticatedUser: {
+            ...authenticatedUser,
+            ownedAssetShortHeaders: assetShortHeaders,
+          },
+        })),
+      error => {
+        console.error('Error while loading user assets:', error);
```
4ian, line 311:
```suggestion
        console.error('Error while loading user owned private assets:', error);
```

newIDE/app/src/ProjectsStorage/CloudStorageProvider/CloudResourceMover.js
```diff
@@ -30,94 +38,83 @@ const isBlobURL = (filename: string) => {
...
 
-  const newCloudProjectId = newFileMetadata.fileIdentifier;
+/**
+ * Find the resources stored on GDevelop Cloud that must be downloaded and
+ * uploaded into the new project.
+ */
+const getResourcesToFetchAndUpload = ({
+  project,
+  selectCondition,
+  ignoreCondition,
```
4ian, line 55: Not a big fan of these two things, as I can understand what this is selecting or ignoring.
The word "Condition" here gives no clue (sure it's a function, but the typing should tell me this already).

newIDE/app/src/ProjectsStorage/ResourceFetcher/BrowserResourceFetcher.js
```diff
@@ -6,6 +6,7 @@ import {
   type FetchAllProjectResourcesFunction,
 } from './index';
 import CloudStorageProvider from '../CloudStorageProvider';
+import { moveUrlResourcesToCloudFilesIfPrivate } from '../CloudStorageProvider/CloudResourceMover';
```
4ian, line 9: You can see there is "something weird going on" because you're in a "ResourceFetcher" file and you import something from a "ResourceMover".

Moving !== Fetching.
The first one is to move existing resources from a project to another.
While fetching is to import resources stored somewhere (but not in a project) inside a project.

newIDE/app/src/ProjectsStorage/CloudStorageProvider/CloudResourceMover.js
```diff
@@ -161,15 +162,113 @@ export const moveAllCloudProjectResourcesToCloudProject = async ({
...
+        resourceName,
+        error,
+      });
+    },
+  });
+
+  return result;
+};
+
+export const moveUrlResourcesToCloudFilesIfPrivate = async ({
```
4ian, line 229: See my other comment - what you're doing here is not a "move" of resources between 2 different projects, but a "fetch" of resources inside a single project.
So this should live inside a XxxResourceFetcher file.

newIDE/app/src/ProjectsStorage/CloudStorageProvider/CloudResourceMover.js
```diff
@@ -161,15 +162,113 @@ export const moveAllCloudProjectResourcesToCloudProject = async ({
...
+  onProgress: (number, number) => void,
+|}) => {
+  const result = {
+    erroredResources: [],
+  };
+
+  const cloudProjectId = fileMetadata.fileIdentifier;
+
+  // Get all resources to download.
+  const resourcesToFetchAndUpload = getResourcesToFetchAndUpload({
```
4ian, line 247: I wonder if it would be better NOT to factor the code? Repeating ourselves might be better because it's two different use cases?

newIDE/app/src/ProjectsStorage/CloudStorageProvider/CloudResourceMover.js
```diff
@@ -161,15 +162,113 @@ export const moveAllCloudProjectResourcesToCloudProject = async ({
...
+      result.erroredResources.push({
+        resourceName,
+        error: new Error(errorMessage),
+      });
+    },
+  });
+
+  await uploadResourcesToCloud({
+    resourcesToFetchAndUpload,
+    // // Remove the token used to download the file when re-uploading it.
```
4ian, line 260: Double //

newIDE/app/src/ProjectsStorage/CloudStorageProvider/CloudResourceMover.js
```diff
@@ -161,15 +162,113 @@ export const moveAllCloudProjectResourcesToCloudProject = async ({
...
+    transformFilename: filename => filename.split('?')[0],
```
4ian, line 261: This is a very important thing, that is specific to the private assets. It definitely deserves a named function to explain what it does + a test.

newIDE/app/src/ProjectsStorage/LocalFileStorageProvider/LocalFileResourceMover.js
```diff
@@ -52,8 +52,12 @@ export const moveUrlResourcesToLocalFiles = async ({
       const resource = resourcesManager.getResource(resourceName);
 
       const url = resource.getFile();
-      const extension = path.extname(url);
-      const filenameWithoutExtension = path.basename(url, extension);
+      const urlWithoutQueryParams = url.split('?')[0];
```
4ian, line 55: Same comment as before, definitely needs a named function + test

newIDE/app/src/Utils/GDevelopServices/Shop.js
```diff
@@ -31,3 +31,56 @@ export const listListedPrivateAssetPacks = async (): Promise<
...
+    {},
+    {
+      headers: { Authorization: authorizationHeader },
+      params: { userId },
+    }
+  );
+  return response.data;
+};
+
+export const createAuthorizedUrl = (url: string, token: string): string => {
```
4ian, line 84: I think we're promising too much with this function name. It sounds like a universal way to make an URL authorized. Sure the implementation is simple, but it works only for a specific convention, which works only in the realm of our Shop api.
In other words, name this:
```suggestion
export const createProductAuthorizedUrl = (url: string, token: string): string => {
```

newIDE/app/src/AssetStore/InstallAsset.spec.js
```diff
@@ -5,9 +5,9 @@ import {
   getRequiredBehaviorsFromAsset,
   filterMissingBehaviors,
   downloadExtensions,
-  installAsset,
   filterMissingExtensions,
   sanitizeObjectName,
+  installPublicAsset,
```
AlexandreSi, line 10: Would it be worth adding a test for installPrivateAsset ?

newIDE/app/src/Profile/AuthenticatedUserProvider.js
```diff
@@ -277,6 +281,36 @@ export default class AuthenticatedUserProvider extends React.Component<
...
+      }
+    );
+    listReceivedAssetShortHeaders(authentication.getAuthorizationHeader, {
+      userId: firebaseUser.uid,
+    }).then(
+      assetShortHeaders =>
+        this.setState(({ authenticatedUser }) => ({
+          authenticatedUser: {
+            ...authenticatedUser,
+            ownedAssetShortHeaders: assetShortHeaders,
```
4ian, line 307: The method is called `received`, but the resulting assetShortHeaders are stored into something "owned".
So we must chose to be consistent:
- is this owned?
- or is this received?

newIDE/app/src/Profile/AuthenticatedUserContext.js
4ian: or 
```suggestion
  receivedPrivateAssetPackIds: ?Array<string>,
  receivedPrivateAssetShortHeaders: ?Array<AssetShortHeader>,
```

newIDE/app/src/Profile/AuthenticatedUserProvider.js
4ian: or
```suggestion
        console.error('Error while loading received private assets:', error);
```

newIDE/app/src/Utils/GDevelopServices/Asset.js
```diff
@@ -177,25 +179,38 @@ export const listAllAssets = ({
...
+};
+
+export const getPrivateAsset = async (
+  assetShortHeader: AssetShortHeader,
+  authorizationToken: string,
+  { environment }: {| environment: Environment |}
+): Promise<Asset> => {
+  const assetPackId = assetShortHeader.assetPackId;
+  if (!assetPackId) {
+    throw new Error('Asset is not part of a private asset pack.');
```
4ian, line 206: Reading the code this just tells me "the asset is not part of an asset pack that has an ID".
The fact that it means that it's private is just a coincidence, true for now but that could be false in the future, which is:
"It turns out that only private asset packs have ids".

So we should rename "assetPackId" to "privateAssetPackId" or, maybe preferably, just change the error message to:
```suggestion
    throw new Error('Asset is not part of an asset pack that can be accessed with an id.');
```

newIDE/app/src/AssetStore/PrivateAssets/AuthorizedAssetImage.js
```diff
@@ -0,0 +1,105 @@
...
+      if (!authorizationToken) {
+        updateAuthorizationToken();
+      } else {
+        setAuthorizedUrl(createAuthorizedUrl(url, authorizationToken));
+      }
+    },
+    [authorizationToken, updateAuthorizationToken, url, isImageLoaded]
+  );
+
+  // React.useEffect(
```
AlexandreSi, line 47: Should this effect be kept?

newIDE/app/src/Utils/GDevelopServices/Asset.js
```diff
@@ -278,3 +293,30 @@ export const isPixelArt = (
     return tag.toLowerCase() === 'pixel art';
   });
 };
+
+export const isPrivateAsset = (
+  assetOrAssetShortHeader: AssetShortHeader | Asset
+): boolean => {
+  return assetOrAssetShortHeader.previewImageUrls[0].startsWith(
+    'https://private-assets'
```
4ian, line 301: nit: maybe better be consistent with isPrivateAssetUrl and against the two base urls?

newIDE/app/src/ResourcesList/ResourcePreview/ImagePreview.js
```diff
@@ -77,6 +79,9 @@ type Props = {|
   onSize?: (number, number) => void,
   hideCheckeredBackground?: boolean,
   hideControls?: boolean,
+  isPrivate?: boolean,
```
4ian, line 82: Same comment about private being a very overloaded term, let's qualify this prop as much as we can.

newIDE/app/src/AssetStore/PrivateAssets/AuthorizedAssetImage.js
```diff
@@ -0,0 +1,105 @@
...
+    }
+    // If the image is not authorized, fetch a new authorization token.
+    console.info('Error while fetching image, fetching a new token...');
+    updateAuthorizationToken();
+    setTries(tries + 1);
+  };
+
+  const onImageLoaded = (e: any) => {
+    if (!hideLoader) {
+      setIsImageLoaded(true);
```
AlexandreSi, line 75: Should you set it to true even if the loader should be hidden?

AlexandreSi: Since it impacts the authorization token update logic

newIDE/app/src/AssetStore/PrivateAssets/AuthorizedAssetImage.js
```diff
@@ -0,0 +1,105 @@
...
+  onError,
+  onLoad,
+  hideLoader,
+}: Props) => {
+  const [authorizedUrl, setAuthorizedUrl] = React.useState(null);
+  const [isImageLoaded, setIsImageLoaded] = React.useState(false);
+  const { authorizationToken, updateAuthorizationToken } = React.useContext(
+    PrivateAssetsAuthorizationContext
+  );
+  const [tries, setTries] = React.useState(0);
```
AlexandreSi, line 29: I feel like this could be a ref to avoid trigger rendering

newIDE/app/src/AssetStore/AssetDetails.js
```diff
@@ -254,9 +270,11 @@ export const AssetDetails = ({
             {asset ? (
               <>
                 {asset.objectType === 'sprite' &&
-                  animationResources &&
+                animationResources &&
+                typeof selectedAnimationName === 'string' && // Animation name can be empty string.
```
AlexandreSi, line 274: 💜

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/AnimationPreview.js
```diff
@@ -164,17 +190,46 @@ const AnimationPreview = ({
...
+            // When the array of loaders changes, decide if we display the loader or not.
+            // If all images are loaded, then hide loader for instant display.
+            const hasFinishedLoadingAllResources = !imagesLoadedArray.current.some(
+              hasImageLoaded => !hasImageLoaded
+            );
+            if (hasFinishedLoadingAllResources) {
+              setIsStillLoadingResources(false);
+            }
+            // Image has loaded, so cancel the timeout if it was set.
+            if (loaderTimeout.current) clearTimeout(loaderTimeout.current);
```
AlexandreSi, line 215: Should `loaderTimeout.current` be set back to `null`?

newIDE/app/src/AssetStore/PrivateAssets/PrivateAssetsAuthorizationProvider.js
AlexandreSi, line 115: I feel like this should go into a dedicated helper. And maybe something similar already exists

newIDE/app/src/MainFrame/EditorContainers/BaseEditor.js
AlexandreSi, line 94: Should it go to `extraEditorProps`?

newIDE/app/src/MainFrame/index.js
```diff
@@ -757,6 +758,7 @@ const MainFrame = (props: Props) => {
           fileMetadata,
           storageProvider,
           storageProviderOperations,
+          authenticatedUser,
```
AlexandreSi, line 761: Why add authenticated user to dependencies of this function?

newIDE/app/src/MainFrame/index.js
```diff
@@ -770,6 +772,7 @@ const MainFrame = (props: Props) => {
       getStorageProvider,
       getStorageProviderOperations,
       ensureResourcesAreFetched,
+      authenticatedUser,
```
AlexandreSi, line 775: Ditto: Why add authenticated user to dependencies of this function?

AlexandreSi (commented):
😋 Can't wait to install these yummy private assets!

Z24RLECO (approved):

newIDE/app/src/Profile/AuthenticatedUserProvider.js
ClementPasteau: They mean that there is an effect listening to this value, and if we go back to null it will not be triggered again, so I'm setting it to an empty array.
But maybe that's not the best strategy

newIDE/app/src/MainFrame/index.js
ClementPasteau: Actually, there is one case where it's useful.
When you:
- arrive on the webapp (not storage provider)
- create a project
- save it in the cloud (storage provider = cloud)
- create a new project (storage provider still = cloud)
-> we're then in the case where the new project is not saved yet, and yet the default storage provider is cloud.

If what you're saying is true, then we should be clearing the storage provider when opening up a new project, and it doesn't seem to be the case

newIDE/app/src/AssetStore/PrivateAssets/PrivateAssetsAuthorizationProvider.js
ClementPasteau: This is to allow the `AuthorizedImage` to trigger an update when the `` tag errors.
This is the only way we know that the token expired, the provider itself doesn't know about it

newIDE/app/src/Profile/AuthenticatedUserContext.js
ClementPasteau: yup, I'm renaming everything to be clearer

newIDE/app/src/ProjectsStorage/CloudStorageProvider/CloudResourceMover.js
ClementPasteau: That's what I did at first but really thought I was duplicating everything.
This isn't manipulated much, so we can cope with duplication here

newIDE/app/src/MainFrame/index.js
4ian: > If what you're saying is true, then we should be clearing the storage provider when opening up a new project, and it doesn't seem to be the case

Yes. We should reset it to null or UrlStorageProvider, let's see that on another PR and keep a note here in this PR description :)

ClementPasteau: I've made it here, as it was quite easy to implement. I'll let you review the strategy I picked

newIDE/app/src/ProjectsStorage/CloudStorageProvider/CloudResourceMover.js
ClementPasteau: It did feel like a move, but of resources, rather than of a project.
The code is gonna be incredibly similar

newIDE/app/src/AssetStore/PrivateAssets/AuthorizedAssetImage.js
ClementPasteau: yep, I overengineered the logic here

newIDE/app/src/AssetStore/PrivateAssets/AuthorizedAssetImage.js
ClementPasteau: yep, fixing this

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/AnimationPreview.js
ClementPasteau: entirely right

newIDE/app/src/MainFrame/EditorContainers/BaseEditor.js
ClementPasteau: I'm not sure about the purpose of extraEditorProps.
It looks like it's mainly for optional props, but this one is actually mandatory

newIDE/app/src/MainFrame/index.js
ClementPasteau: In the case of being on web, adding a private asset will only be allowed if you've saved your project on the cloud.
This function will then upload all the resources to the cloud, and the authenticatedUser is required to generate pre-signed URLs with authorization.
(see CloudResourceMover, soon moved to CloudResourceFetcher)

newIDE/app/src/AnnouncementsFeed/index.js
```diff
@@ -87,6 +87,7 @@ export const AnnouncementsFeed = ({
                         }
                       : null
                   }
+                  key={announcement.id}
```
AlexandreSi, line 90: 💜

newIDE/app/src/MainFrame/EditorContainers/BaseEditor.js
AlexandreSi: I understand it differently than the comment: for me it's the space where we put the props that are only used by a single editor in order to avoid putting something only used by the external layout editor in the "base" props

newIDE/app/src/MainFrame/index.js
AlexandreSi: Oh right, my bad, I mixed lines up

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/AnimationPreview.js
```diff
@@ -164,17 +192,49 @@ const AnimationPreview = ({
...
+          resourceName={resourceName}
+          imageResourceSource={getImageResourceSource(resourceName)}
+          isImageResourceSmooth={isImageResourceSmooth(resourceName)}
+          initialZoom={initialZoom}
+          project={project}
+          hideCheckeredBackground={hideCheckeredBackground}
+          hideControls={hideControls}
+          fixedHeight={fixedHeight}
+          fixedWidth={fixedWidth}
+          onImageLoaded={() => {
```
AlexandreSi, line 206: nitpick: You can put this props in a function to improve readability

newIDE/app/src/Utils/GDevelopServices/Shop.js
```diff
@@ -31,3 +31,36 @@ export const listListedPrivateAssetPacks = async (): Promise<
...
+    }
+  );
+  return response.data;
+};
+
+export const createProductAuthorizedUrl = (
+  url: string,
+  token: string
+): string => {
+  return url + '?token=' + encodeURIComponent(token);
```
AlexandreSi, line 59: Should we put a check somewhere that there is not already a `?` in the URL? It shouldn't happen but I'd feel more secure if there was such a thing

newIDE/app/src/AssetStore/NewObjectDialog.js
```diff
@@ -30,11 +30,14 @@ import ScrollView from '../UI/ScrollView';
 import useDismissableTutorialMessage from '../Hints/useDismissableTutorialMessage';
 import RaisedButton from '../UI/RaisedButton';
 import { AssetStoreContext } from './AssetStoreContext';
-import { AssetPackDialog } from './AssetPackDialog';
-import { installAsset } from './InstallAsset';
+import AssetPackDialog from './AssetPackInstallDialog';
```
AlexandreSi, line 33: nitpick: Should you name the imported component AssetPackInstallDialog

AlexandreSi (approved):
LGTM!

PR #4358, opened 2022-10-05
Add array serialisation of multiline strings

arthuro555:
Adds the serialization of multiline strings as arrays, to improve the read-friendliness of GDevelop JSON files.
See discussion on https://github.com/4ian/GDevelop/discussions/4160

### DIFF
--- a/Core/GDCore/Serialization/Serializer.cpp
+++ b/Core/GDCore/Serialization/Serializer.cpp
@@ -111,8 +111,12 @@ gd::String Serializer::ToEscapedXMLString(const gd::String& str) {
 }
 
 namespace {
+
+const char* MultilineStringMagicKey = "__MULTILINESTRING__";
+
 void RapidJsonValueToElement(const Value& value,
                              gd::SerializerElement& element) {
+  
   if (value.IsBool()) {
     element.SetBoolValue(value.GetBool());
   } else if (value.IsNumber()) {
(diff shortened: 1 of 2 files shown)


4ian:
Very interesting! I wonder if instead of a magic string that could be used anywhere, we should ask for this capability at the moment we write or read. Something like:

SetStringOrMultilineStringAsArray
and
GetStringOrMultilineStringFromArray

Then we would just need to go through the code to add places where we support those? What do you think?

arthuro555:
Personally, I like the magic string as it also tells the reader of the JSON that it is not to be read as an array but as text. I also cannot think of any case where Multiline strings are present where the array representation of a strong does not make sense. Buuuuut I see how automagically converting strings between arrays based on a magic string is kind of a code smell...

4ian:
I think the "magic value" approach might work/be safe, but I still have a gut feeling of a bad code smell. Notably:
- this makes writing the JSON manually harder "what's the magic string again I forgot?"
- and harder to read
- ...and what if someone store the exact same magic value as the first element of an array? Sure it should not happen, but Murphy's law says that it will somehow happen one day ;)

In a way, the serialiser is defining a "schema" when we're using it. If we had a tool that could somehow explore all code paths of the SerializeTo/UnserializeFrom methods, it could generate a schema according to the Get/SetXxx calls.

What I suggest here is that when writing/reading some strings, we explicitly tell "this field is a string | Array<string>". I move the issue to the caller of the serialiser instead of the JSON serialisation (i.e: rather than extending JSON with a magic value, we just handle both strings and arrays).

I would say that it would be worth giving a try? It might be just two methods to implement and then go through the code and use them when useful. Note that please don't feel force to then adapt the whole codebase. Adapting extensions to use this string or array of string methods will be more than enough to start!

arthuro555:
nevermind

PR #4361, opened 2022-10-06
Fix points editor

AlexandreSi:


newIDE/app/src/ObjectEditor/Editors/SpriteEditor/PointsEditor/PointRow.js
```diff
@@ -32,114 +31,110 @@ type Props = {|
...
+              ? { color: muiTheme.listItem.selectedTextColor }
+              : undefined
+          }
+          value={props.pointName}
+          id={props.pointName}
+          fullWidth
+          errorText={props.nameError ? 'This name is already taken' : undefined}
+          disabled={!props.onBlur}
+          onBlur={props.onBlur}
+          onChange={value => {}}
```
4ian, line 61: This can be replaced by:

```suggestion
          disabled={!props.onChangePointName}
          commitOnBlur
          onChange={props.onChangePointName}
```

if you replace `onBlur?: (ev: any) => void,` by `onChangePointName: (newName: string) => void` and adapt the caller (there is only one place where to change this).

4ian (approved):

PR #4362, opened 2022-10-06
Reinitialize object default hitboxes when reinitializing it

### LINKED ISSUE #4339: Ghost objects (Error in object's creation from external layouts)
Things going wrong in the creation of external layouts:
Creating objects from an external layout can bring to a strange issue where "ghosts" of the current objects from the external layout are created with a random size, and angle, and not only that, they are invisible, but they can return data of visibility and opacity conditions perfectly, these work like the other objects, but the collisions are most of the time broken (and even convex sometimes)

Why this needs to fix? simple:
Restart systems, let's say you lost a life and you want to restart the level, you gotta delete the current objects and load them back (since you don't want to re-create and place the enemies and bosses you've defeated event by event), you'll just have to delete all objects in the scene and load them back from an external layout. But what if you get a random invisible ghost block in front of your ladder that doesn't let you go further? well...

I was able to reproduce it only in my project, so i had to make a copy of my project and clean it up for the bug test project

Scene without "Ghost objects"
image

Scene with "Ghost objects"
image

The bug test project:
[LayoutBug.zip](https://github.com/4ian/GDevelop/files/9669847/LayoutBug.zip)

Some other details
* Windows 10
* GDevelop 5.0.145

AlexandreSi:
Fixes #4339

### DIFF
--- a/GDJS/Runtime/runtimeobject.ts
+++ b/GDJS/Runtime/runtimeobject.ts
@@ -297,6 +297,8 @@ namespace gdjs {
       this.persistentUuid = null;
       this.pick = false;
       this.hitBoxesDirty = true;
+      this._defaultHitBoxes.length = 0;
+      this._defaultHitBoxes.push(gdjs.Polygon.createRectangle(0, 0));
       this.aabb.min[0] = 0;
       this.aabb.min[1] = 0;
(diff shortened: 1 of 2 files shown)


ClementPasteau (approved):
🚀

PR #4363, opened 2022-10-06
Add a button to export custom objects

D8H:
* There is the custom object `MyButton` in this list, the "export" action is only in the contextual menu of this object.
http://gdevelop-storybook.s3-website-us-east-1.amazonaws.com/custom-object-export/latest/index.html?path=/story/layouteditor-objectslist--default

* It shows this dialog
http://gdevelop-storybook.s3-website-us-east-1.amazonaws.com/custom-object-export/latest/index.html?path=/story/layouteditor-objectexporterdialog--default

The export button is not working because it's not supported from a browser. The "export" action from the contextual menu should not show either in practice but it's forced by the story by setting `onExportObject` (see: https://github.com/4ian/GDevelop/pull/4363#discussion_r994540816).

### DIFF
--- a/newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionWriter.js
+++ b/newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionWriter.js
@@ -88,12 +88,11 @@ export default class LocalEventsFunctionsExtensionWriter {
     customObject: gdObject,
     filepath: string
   ): Promise<void> => {
-    const exportedObject = customObject.clone();
+    const exportedObject = customObject.clone().get();
     exportedObject.setTags('');
     exportedObject.getVariables().clear();
     exportedObject.getEffects().clear();
     exportedObject
-      .get()
       .getAllBehaviorNames()
       .toJSArray()
       .forEach(name => exportedObject.removeBehavior(name));
--- a/newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionWriter.js
+++ b/newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionWriter.js
@@ -93,6 +93,7 @@ export default class LocalEventsFunctionsExtensionWriter {
     exportedObject.getVariables().clear();
     exportedObject.getEffects().clear();
     exportedObject
+      .get()
       .getAllBehaviorNames()
       .toJSArray()
       .forEach(name => exportedObject.removeBehavior(name));
--- a/newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionWriter.js
+++ b/newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionWriter.js
@@ -99,7 +99,6 @@ export default class LocalEventsFunctionsExtensionWriter {
       .toJSArray()
       .forEach(name => exportedObject.removeBehavior(name));
     const serializedObject = serializeToJSObject(exportedObject);
-    exportedObject.delete();
     return writeJSONFile(serializedObject, filepath).catch(err => {
       console.error('Unable to write the object:', err);
       throw err;
--- a/newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionWriter.js
+++ b/newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionWriter.js
@@ -88,6 +88,8 @@ export default class LocalEventsFunctionsExtensionWriter {
     customObject: gdObject,
     filepath: string
   ): Promise<void> => {
+    // TODO Fix the memory crash
+    // I think the clone method is bugged, it doesn't keep the Configuration.
     const exportedObject = customObject.clone().get();
     exportedObject.setTags('');
     exportedObject.getVariables().clear();
@@ -96,7 +98,7 @@ export default class LocalEventsFunctionsExtensionWriter {
       .getAllBehaviorNames()
       .toJSArray()
       .forEach(name => exportedObject.removeBehavior(name));
-    const serializedObject = serializeToJSObject(customObject);
+    const serializedObject = serializeToJSObject(exportedObject);
     exportedObject.delete();
     return writeJSONFile(serializedObject, filepath).catch(err => {
       console.error('Unable to write the object:', err);
--- a/newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionWriter.js
+++ b/newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionWriter.js
@@ -88,8 +88,6 @@ export default class LocalEventsFunctionsExtensionWriter {
     customObject: gdObject,
     filepath: string
   ): Promise<void> => {
-    // TODO Fix the memory crash
-    // I think the clone method is bugged, it doesn't keep the Configuration.
     const exportedObject = customObject.clone().get();
     exportedObject.setTags('');
     exportedObject.getVariables().clear();
--- a/newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionWriter.js
+++ b/newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionWriter.js
@@ -44,7 +44,7 @@ export default class LocalEventsFunctionsExtensionWriter {
             extensions: ['json'],
           },
         ],
-        defaultPath: extensionName ? extensionName : 'Extension.json',
+        defaultPath: extensionName || 'Extension.json',
       })
       .then(({ filePath }) => {
         if (!filePath) return null;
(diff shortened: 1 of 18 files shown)


newIDE/app/src/EventsFunctionsExtensionsLoader/Storage/LocalEventsFunctionsExtensionWriter.js
```diff
@@ -62,4 +62,45 @@ export default class LocalEventsFunctionsExtensionWriter {
       throw err;
     });
   };
+
+  static chooseCustomObjectFile = (objectName?: string): Promise<?string> => {
```
D8H, line 66: Is it ok to add this here?
It could be renamed `LocalProjectPartExporter`, but there is probably a better name.

4ian:
Very nice. We'll want to move the button so that it's a bit more hidden (it's not something you use so frequently, especially for new users who would feel confused by this). But the idea is there 👍

D8H:
> Very nice. We'll want to move the button so that it's a bit more hidden (it's not something you use so frequently, especially for new users who would feel confused by this). But the idea is there 👍

What do you think about moving the action in the object list contextual menu? (as the button, it will only appear for custom objects)

newIDE/app/src/SceneEditor/index.js
```diff
@@ -1387,6 +1412,9 @@ export default class SceneEditor extends React.Component<Props, State> {
                 selectedObjectNames={this.state.selectedObjectNames}
                 canInstallPrivateAsset={this.props.canInstallPrivateAsset}
                 onEditObject={this.props.onEditObject || this.editObject}
+                onExportObject={
+                  eventsFunctionsExtensionWriter ? this.exportObject : null
+                }
```
D8H, line 1417: This disable the export action from the web-app.

newIDE/app/src/ObjectEditor/ObjectExporterDialog.js
```diff
@@ -0,0 +1,103 @@
...
+|};
+
+const ObjectExporterDialog = (props: Props) => {
+  const eventsFunctionsExtensionsState = React.useContext(
+    EventsFunctionsExtensionsContext
+  );
+
+  return (
+    <Dialog
+      secondaryActions={[
```
AlexandreSi, line 54: Given the size of the content, I think you could use a dialog of size `"sm"`

newIDE/app/src/ObjectEditor/ObjectExporterDialog.js
```diff
@@ -0,0 +1,103 @@
...
+      <Column expand>
+        <Line>
+          <Text>
+            <Trans>
+              You can export the object to a file to submit it for the asset
+              store.
+            </Trans>
+          </Text>
+        </Line>
+        <LineStackLayout>
```
AlexandreSi, line 84: You should use the ResponsiveLineStackLayout for a better mobile experience

newIDE/app/src/ObjectEditor/ObjectExporterDialog.js
```diff
@@ -0,0 +1,103 @@
...
+} from '../EventsFunctionsExtensionsLoader/EventsFunctionsExtensionsContext';
+import Window from '../Utils/Window';
+
+const exportCustomObject = (
+  eventsFunctionsExtensionsState: EventsFunctionsExtensionsState,
+  customObject: gdObject
+) => {
+  const eventsFunctionsExtensionWriter = eventsFunctionsExtensionsState.getEventsFunctionsExtensionWriter();
+  if (!eventsFunctionsExtensionWriter)
+    return Promise.reject(new Error('Not supported'));
```
AlexandreSi, line 23: I think this error should contain the context and the true error: `Custom object export error: couldn't find the extension writer`.

newIDE/app/src/ObjectEditor/ObjectExporterDialog.js
```diff
@@ -0,0 +1,103 @@
...
+      open
+      onRequestClose={() => {
+        props.onClose();
+      }}
+    >
+      <Column expand>
+        <Line>
+          <Text>
+            <Trans>
+              You can export the object to a file to submit it for the asset
```
4ian, line 79:
```suggestion
              You can export the object to a file to submit it to the asset
```

newIDE/app/src/ObjectEditor/ObjectExporterDialog.js
```diff
@@ -0,0 +1,103 @@
...
+
+const exportCustomObject = (
+  eventsFunctionsExtensionsState: EventsFunctionsExtensionsState,
+  customObject: gdObject
+) => {
+  const eventsFunctionsExtensionWriter = eventsFunctionsExtensionsState.getEventsFunctionsExtensionWriter();
+  if (!eventsFunctionsExtensionWriter)
+    return Promise.reject(new Error('Not supported'));
+
+  return eventsFunctionsExtensionWriter
```
4ian, line 25: Prefer use async/await

newIDE/app/src/ObjectEditor/ObjectExporterDialog.js
4ian, line 23: Use an async function and do:
```suggestion
    throw new Error('Not supported');
```

newIDE/app/src/ObjectEditor/ObjectExporterDialog.js
```diff
@@ -0,0 +1,103 @@
...
+    .chooseCustomObjectFile(customObject.getName())
+    .then(pathOrUrl => {
+      if (!pathOrUrl) return;
```
AlexandreSi, line 28: Should we raise an error?

newIDE/app/src/SceneEditor/index.js
```diff
@@ -385,6 +389,22 @@ export default class SceneEditor extends React.Component<Props, State> {
     }
   };
 
+  exportObject = (editedObject: ?gdObject) => {
+    if (editedObject) {
+      this.setState({
+        exportedObjectWithContext: {
+          object: editedObject,
```
4ian, line 396: Ubernit:
```suggestion
  exportObject = (object: ?gdObject) => {
    if (object) {
      this.setState({
        exportedObjectWithContext: {
          object,
```

newIDE/app/src/stories/componentStories/LayoutEditor/ObjectExporterDialog.stories.js
```diff
@@ -0,0 +1,43 @@
...
+  component: ObjectExporterDialog,
+  decorators: [muiDecorator, paperDecorator],
+};
+
+const fakeEventsFunctionsExtensionsContext = {
+  loadProjectEventsFunctionsExtensions: async project => {},
+  unloadProjectEventsFunctionsExtensions: project => {},
+  unloadProjectEventsFunctionsExtension: (project, extensionName) => {},
+  reloadProjectEventsFunctionsExtensions: async project => {},
+  getEventsFunctionsExtensionWriter: () => LocalEventsFunctionsExtensionWriter,
```
AlexandreSi, line 27: I wonder if this function should use the action method so that when you click on the button in the story, you can see the action has been called.

newIDE/app/src/ObjectEditor/ObjectExporterDialog.js
4ian: This can be valid if the user cancel the save dialog I think.

4ian (approved):

newIDE/app/src/SceneEditor/index.js
```diff
@@ -385,6 +389,22 @@ export default class SceneEditor extends React.Component<Props, State> {
     }
   };
 
+  exportObject = (editedObject: ?gdObject) => {
```
AlexandreSi, line 392: The naming is a bit misleading: should you name it `openObjectExportDialog`?

newIDE/app/src/SceneEditor/index.js
```diff
@@ -385,6 +389,22 @@ export default class SceneEditor extends React.Component<Props, State> {
...
+    if (editedObject) {
+      this.setState({
+        exportedObjectWithContext: {
+          object: editedObject,
+          // It's not important for the export.
```
AlexandreSi, line 397: Why use the type `ObjectWithContext` if the context is not important? Am I missing something?

AlexandreSi (commented):
A few changes here and there but looks good to me!
The copy of the dialog feels a bit empty, maybe we could have it read by an native english speaker person

newIDE/app/src/SceneEditor/index.js
```diff
@@ -1403,7 +1422,10 @@ export default class SceneEditor extends React.Component<Props, State> {
                   })
                 }
                 getAllObjectTags={this._getAllObjectTags}
-                ref={objectsList => (this._objectsList = objectsList)}
+                ref={
+                  // $FlowFixMe Make this component functional.
+                  objectsList => (this._objectsList = objectsList)
+                }
```
D8H, line 1428: Should we address this?

4ian: Is it better with `ref={this._objectsList}`?

newIDE/app/src/stories/componentStories/LayoutEditor/ObjectExporterDialog.stories.js
D8H: I'm not sure because it's not really a user action, but rather internal to the implementation.

4ian (approved):
Double check that things like moving an object to be a global object, or other things like modifying a global object name in a scene and then switching to another scene, still works. 
Once we're confident everything works, good to go!

D8H:
I did the test, it seems to be ok.

Issue #4366, opened 2022-10-06
Tour not going to "Asset Store" when creating a new object


isaacbmc:
In the tour, when prompted to _add an object_, clicking the `Add a new object` button opens the "New object from scratch" tab instead of the "Asset Store". 

![image](https://user-images.githubusercontent.com/6667151/194429102-c9ed22a4-587b-4e4a-94bc-12053bf94a4a.png)

The tour notifies you that you arent on the correct page:
![image](https://user-images.githubusercontent.com/6667151/194429530-0be1aa8b-d1b5-498d-9b5b-62fe841f1d89.png)

I believe this is occurring if you add a new object before going through the tour, click the `Add a new object button`, navigate to the "New object from scratch" tab, and then go back to the home page to start the tour.

Bouh:
Hi @isaacbmc 
I'm fixing this, thank you for the report. 
It will be fixed for the next release.

ClementPasteau:
closing this issue as this is now fixed

PR #4369, opened 2022-10-07
Add id for the tour

Bouh:
Add id to adds a step in the userflow to fix this issue:
https://github.com/4ian/GDevelop/issues/4366

Tested and working in local with EN, BR, PT, and FR, 

image


D8H (approved):

D8H:
The fail test is not related (tile map loading for a test)

PR #4371, opened 2022-10-07
Notify web users when new version should be installed

ClementPasteau:
Hard to test in real conditions, I've mainly tested in the dev console, having editor.gdevelop.io opened.
Tested on Chrome, Firefox & Safari, the reloading/cache clearing works well.

### DIFF
--- a/newIDE/app/src/MainFrame/UpdaterTools.js
+++ b/newIDE/app/src/MainFrame/UpdaterTools.js
@@ -1,8 +1,9 @@
 // @flow
 // See ElectronEventsBridge, AboutDialog and electron-app/main.js for handling the updates.
 
-import { Trans } from '@lingui/macro';
+import { t, Trans } from '@lingui/macro';
 import React from 'react';
+import useAlertDialog from '../UI/Alert/useAlertDialog';
 
 export type ElectronUpdateStatus = {
   message: string,
@@ -68,6 +69,7 @@ export const useServiceWorkerUpdateStatus = () => {
     serviceWorkerUpdateStatus,
     setServiceWorkerUpdateStatus,
   ] = React.useState<ServiceWorkerUpdateStatus>('unknown');
+
   React.useEffect(() => {
     (async () => {
       if (navigator.serviceWorker) {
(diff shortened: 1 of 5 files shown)


newIDE/app/src/MainFrame/index.js
```diff
@@ -658,6 +661,29 @@ const MainFrame = (props: Props) => {
...
 
+  const serviceWorkerUpdateStatus = useServiceWorkerUpdateStatus();
+
+  useTimeout(
+    React.useCallback(
+      async () => {
+        if (serviceWorkerUpdateStatus === 'update-ready') {
+          const answer = await showConfirmation({
+            title: t`New update available!`,
+            message: t`A new version of GDevelop is available. Would you like to update now?`,
```
AlexandreSi, line 672: We should mention that the page will be reloaded.

newIDE/app/src/MainFrame/index.js
```diff
@@ -658,6 +661,29 @@ const MainFrame = (props: Props) => {
...
+          });
+          if (answer) {
+            await updateServiceWorkerToLatestVersion();
```
AlexandreSi, line 675: I wonder if we should add a try catch here so that if anything fails, since the user is expecting the page to be reloaded, they can see a (native?) alert that informs them that the update has failed and it should be available the next time they close and reopen GDevelop (if that's even true?)

newIDE/app/src/MainFrame/UpdaterTools.js
```diff
@@ -95,6 +95,14 @@ export const useServiceWorkerUpdateStatus = () => {
   return serviceWorkerUpdateStatus;
 };
 
+export const updateServiceWorkerToLatestVersion = async () => {
+  const { serviceWorker } = navigator;
+  if (!serviceWorker) return;
+  const registration = await serviceWorker.getRegistration();
+  if (!registration) return;
+  await registration.unregister();
```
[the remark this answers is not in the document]
ClementPasteau: Yeah the wording of the function is wrong, all the logic should probably be in this function.
1. Unregistering ensures the next reload will load the latest service-worker
2. cache clearing ensure the browser is not pointing at a wrong bundle name (conflict between what the service worker is expecting and what the browser has put in cache)

newIDE/app/src/MainFrame/UpdaterTools.js
```diff
@@ -95,6 +95,21 @@ export const useServiceWorkerUpdateStatus = () => {
...
+    throw new Error('Service worker not available');
+  }
+  const registration = await serviceWorker.getRegistration();
+  if (!registration) {
+    throw new Error('Service worker registration not available');
+  }
+  await registration.unregister();
+  const cacheKeys = await caches.keys();
+  await Promise.all(cacheKeys.map(key => caches.delete(key)));
+  window.location.reload();
```
4ian, line 110: Because of this, this function is more than just a "updateServiceWorkerToLatestVersion", it's a "clearServiceWorkerAndForceReload"

newIDE/app/src/MainFrame/index.js
```diff
@@ -660,6 +663,40 @@ const MainFrame = (props: Props) => {
 
   useDiscordRichPresence(currentProject);
 
+  const serviceWorkerUpdateStatus = useServiceWorkerUpdateStatus();
+
+  useTimeout(
+    React.useCallback(
+      async () => {
+        if (serviceWorkerUpdateStatus === 'update-ready') {
```
4ian, line 671: I would almost put this in a dedicated function in UpdateTools to avoid bloating too much MainFrame?

newIDE/app/src/MainFrame/index.js
```diff
@@ -660,6 +663,40 @@ const MainFrame = (props: Props) => {
...
+          const answer = await showConfirmation({
+            title: t`New update available!`,
+            message: t`A new version of GDevelop is available. Would you like to update now? The page will be reloaded.`,
```
4ian, line 674: The page => The app

newIDE/app/src/MainFrame/index.js
4ian: Something like useServiceWorkerCheckAndAskToUpdate

newIDE/app/src/MainFrame/UpdaterTools.js
```diff
@@ -95,6 +95,21 @@ export const useServiceWorkerUpdateStatus = () => {
...
+  const { serviceWorker } = navigator;
+  if (!serviceWorker) {
+    throw new Error('Service worker not available');
+  }
+  const registration = await serviceWorker.getRegistration();
+  if (!registration) {
+    throw new Error('Service worker registration not available');
+  }
+  await registration.unregister();
+  const cacheKeys = await caches.keys();
```
4ian, line 108: TIL about "caches"! Funny to have such a common word as a global variable.

4ian (approved):

Issue #4340, opened 2022-09-29
Replace current work area "vector scrolls" by UI standard scrolls.


LuniMoon:
## Description
When working on the scene area, I accidentally click on the "vector style" scroll by mistake and i find myself lost somewhere outside my canvas.

_See handles for the scroll:_
![image](https://user-images.githubusercontent.com/103995399/192978933-1aec36c4-1a32-45bd-9c9b-4940fa4e2365.png)

The issue might be due to the fact that the handles are too light, or that they don't look like UI standard scrollbars.

## Solution suggested
Replace the component by a "traditional" 8px scroll handle with a 4px spacing around it to optimize click (ideally, if the user clicks on the 4px margin, the scroll bar should be able to react).
![image](https://user-images.githubusercontent.com/103995399/192979504-538e749d-c578-4ff8-8f2c-3383e8591c21.png)

The scroll bars should be placed: at the bottom, and on the right of the workspace (currently on the left) to align to UI standards.

Silver-Streak:
I think this is a good idea, although I don't normally like UI changes to applications that reduce the size of the main work area. If something like this was implemented, would it be possible for the full scrolls to "collapde/expand" when you hover near them in some way?

LuniMoon:
Good observation!
This is what I've observed on Design softwares:

Scroll bars always have transparency
<img width="148" alt="Screenshot 2022-09-30 at 12 00 30" src="https://user-images.githubusercontent.com/103995399/193246110-eccf545f-3fdd-4445-af46-98c52857c908.png">

Depending on the app, there are 2 interactions:
Hidden by default, and:
1. When the keyboard "space" is pressed, and the "grab hand" appears, the scroll bars (vertical and horizontal) appear and stay visible while the user navigates the workspace (letting them know that these bars exist), and disappear when the user releases the space bar.
2. When the user scrolls vertically with their mouse, the scrolling bars appear, and stay visible for 10 seconds. They disappear after.

Always visible by default:
vertical and horizontal always visible.

Silver-Streak:
These all sound viable. I wonder if it'd be possible to have them as a toggleable setting for which mode people prefer?

LuniMoon:
I would choose one of the 2 ways of interaction.
Having personalisation freedom is also maintenance (for Design but also Engineering).

I'm gonna go for "hidden by default" since it's better for mobile supports (which "grab" interaction can be "drag with one finger" like Google Maps).

Bouh:
[...]

PR #4377, opened 2022-10-11
Attempt to improve scrollbars

ClementPasteau:
Trying to address #4340 

I went for a full custom solution, which requires intricate logic about event listeners but it looks like everything is covered, from what I tested.

I'm unsure about the performance of this solution, as it triggers a lot of updates.
However, I've tested with 6x throttling of the CPU and it still works mainly well (few slowdowns when dragging too quickly)
We're mainly using refs to control the updates + every update is done via a throttle to be close to 60 fps

Storybook link to test: http://gdevelop-storybook.s3-website-us-east-1.amazonaws.com/improve-scrollbars/latest/index.html?path=/story/editor-fullsizeinstanceseditorwithscrollbars--default

image

### DIFF
--- a/newIDE/app/src/InstancesEditor/FullSizeInstancesEditorWithScrollbars.js
+++ b/newIDE/app/src/InstancesEditor/FullSizeInstancesEditorWithScrollbars.js
@@ -1,30 +1,76 @@
 // @flow
-import React, { Component } from 'react';
-import Slider from '@material-ui/core/Slider';
+import React from 'react';
 import ViewPosition from './ViewPosition';
 import throttle from 'lodash/throttle';
 import InstancesEditor, {
   type InstancesEditorPropsWithoutSizeAndScroll,
 } from './index';
-import { ScreenTypeMeasurer } from '../UI/Reponsive/ScreenTypeMeasurer';
+import { useScreenType } from '../UI/Reponsive/ScreenTypeMeasurer';
 import { FullSizeMeasurer } from '../UI/FullSizeMeasurer';
+import useForceUpdate from '../Utils/UseForceUpdate';
+import { useDebounce } from '../Utils/UseDebounce';
+
+const SCROLLBAR_DETECTION_WIDTH = 50;
+const SCROLLBAR_TRACK_WIDTH = 16;
+const SCROLLBAR_THUMB_WIDTH = 8;
+const SCROLLBAR_SIZE = 200;
+const SCROLLBAR_MARGIN = (SCROLLBAR_TRACK_WIDTH - SCROLLBAR_THUMB_WIDTH) / 2;
+
+const THROTTLE_TIME = 1000 / 60; // 60 FPS
 
 const styles = {
   container: {
     overflow: 'hidden',
   },
-  xScrollbar: {
+  xScrollbarDetectionZone: {
     position: 'absolute',
-    top: -13,
     left: 0,
     right: 0,
+    bottom: 0,
+    height: SCROLLBAR_DETECTION_WIDTH,
   },
-  yScrollbar: {
+  xScrollbarTrack: {
+    position: 'relative',
+    display: 'inline-block',
+    width: '100%',
+    marginTop: SCROLLBAR_DETECTION_WIDTH - SCROLLBAR_TRACK_WIDTH,
+    marginRight: SCROLLBAR_TRACK_WIDTH - SCROLLBAR_MARGIN, // leave some margin for the vertical scrollbar
+    height: SCROLLBAR_TRACK_WIDTH,
+  },
+  xThumb: {
+    position: 'relative',
+    width: SCROLLBAR_SIZE,
+    marginTop: SCROLLBAR_MARGIN,
+    height: SCROLLBAR_THUMB_WIDTH,
+    backgroundColor: '#1D1D26',
+    outline: '2px solid #FAFAFA',
+    opacity: 0.3,
+    borderRadius: 4,
+  },
+  yScrollbarDetectionZone: {
     position: 'absolute',
     top: 0,
     bottom: 0,
-    left: -13,
-    transform: 'rotate(180deg)',
+    right: 0,
+    width: SCROLLBAR_DETECTION_WIDTH,
+  },
+  yScrollbarTrack: {
+    position: 'relative',
+    display: 'inline-block',
+    height: '100%',
+    marginLeft: SCROLLBAR_DETECTION_WIDTH - SCROLLBAR_TRACK_WIDTH,
+    marginBottom: SCROLLBAR_TRACK_WIDTH - SCROLLBAR_MARGIN, // leave some margin for the vertical scrollbar
+    width: SCROLLBAR_TRACK_WIDTH,
+  },
+  yThumb: {
+    position: 'relative',
+    height: SCROLLBAR_SIZE,
+    marginLeft: SCROLLBAR_MARGIN,
+    width: SCROLLBAR_THUMB_WIDTH,
+    backgroundColor: '#1D1D26',
+    outline: '1px solid #FAFAFA',
+    opacity: 0.3,
+    borderRadius: 4,
   },
 };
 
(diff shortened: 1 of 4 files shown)


newIDE/app/src/InstancesEditor/ViewPosition.js
```diff
@@ -6,56 +6,56 @@ import { type InstancesEditorSettings } from './InstancesEditorSettings';
...
+  initialWidth: number,
+  initialHeight: number,
   instancesEditorSettings: InstancesEditorSettings,
 |};
 
 export default class ViewPosition {
   viewX: number = 0;
   viewY: number = 0;
-  _width: number;
-  _height: number;
```
AlexandreSi, line 18: I think that the underscore was here to signify that this property is "private" and that one should use the getWidth and getHeight methods. Maybe we should keep them although those methods do nothing but returning these attributes at the moment.

newIDE/app/src/InstancesEditor/StatusBar.js
```diff
@@ -41,22 +41,41 @@ export default class StatusBar {
   }
 
   render() {
-    const padding = 5;
-    const borderRadius = 4;
+    const textPadding = 5;
+    const statusBarPadding = 20;
+    const borderRadius = 6;
+    const textXPosition = Math.round(
```
AlexandreSi, line 47: Doing this, you put the text at the **right** bottom corner although it was at the left bottom corner until now. Was it a requested feature?

AlexandreSi: And one could think that it's badly positioned since it goes too much into the scene. Is it possible to shift the text when the scrollbars appear?

newIDE/app/src/InstancesEditor/ViewPosition.js
ClementPasteau: I was confused with the fact that viewX & viewY don't have this. 
I can revert that for sure, I thought it was more readable

newIDE/app/src/InstancesEditor/StatusBar.js
ClementPasteau: Not really, with the scrollbar added, it was now in the way. 
And as we moved it, Luni & I felt it was more natural to be at the intersection of both scrollbars.
That can be changed if we feel this is not wise

newIDE/app/src/InstancesEditor/FullSizeInstancesEditorWithScrollbars.js
```diff
@@ -33,154 +76,412 @@ type Props = {|
   wrappedEditorRef: ?(?InstancesEditor) => void,
 |};
 
-type State = {|
```
AlexandreSi, line 36: 💜

newIDE/app/src/InstancesEditor/FullSizeInstancesEditorWithScrollbars.js
```diff
@@ -33,154 +76,412 @@ type Props = {|
...
-  _handleYChange = (e: any, value: number) => {
-    const { yMin, yMax } = this.state;
-    // Substract the value from 100 as the slider is 180deg rotated. Not perfect though
-    // as it breaks when using the keyboard arrow keys (when focusing the slider).
-    const yValue = yMin + ((yMax - yMin) * (100 - value)) / 100;
-    this.setState(
-      {
-        yValue,
+      timeoutHidingScrollbarsId.current = setTimeout(() => {
+        console.log('clearing timeout function');
```
AlexandreSi, line 115: To remove

newIDE/app/src/InstancesEditor/FullSizeInstancesEditorWithScrollbars.js
```diff
@@ -33,154 +76,412 @@ type Props = {|
...
+  const xScrollbarThumb = React.useRef<?HTMLDivElement>(null);
+  const yScrollbarDetectionZone = React.useRef<?HTMLDivElement>(null);
+  const yScrollbarTrack = React.useRef<?HTMLDivElement>(null);
+  const yScrollbarThumb = React.useRef<?HTMLDivElement>(null);
+
+  const showScrollbars = React.useRef(false);
+  const timeoutHidingScrollbarsId = React.useRef<?TimeoutID>(null);
+  const isDragging = React.useRef(false);
+
+  const xValue = React.useRef(0);
```
AlexandreSi, line 96: xScrollValue and yScrollValue ?

newIDE/app/src/InstancesEditor/FullSizeInstancesEditorWithScrollbars.js
```diff
@@ -33,154 +76,412 @@ type Props = {|
...
+  const maxYScrollbarTopPosition = `calc(100% - ${SCROLLBAR_SIZE}px - ${SCROLLBAR_TRACK_WIDTH}px)`;
+  const expectedYScrollbarTopPosition = `calc(${((yValue.current -
+    yMin.current) /
+    (yMax.current - yMin.current)) *
+    100 +
+    '%'} - ${SCROLLBAR_SIZE}px / 2)`;
+  const yScrollbarTopPosition = `max(min(${expectedYScrollbarTopPosition}, ${maxYScrollbarTopPosition}), ${minYScrollbarTopPosition})`;
+
+  return (
+    <ScreenTypeMeasurer>
```
AlexandreSi, line 413: You can use the hook useScreenType

newIDE/app/src/InstancesEditor/FullSizeInstancesEditorWithScrollbars.js
ClementPasteau: well technicaly it's not related to scroll, it defines where the viewPosition is positioned

newIDE/app/src/InstancesEditor/StatusBar.js
ClementPasteau: I've pushed it as far as possible. On my screen it looks fine but lmk if you think we should do something else.
I think that making it move depending on scrollbar visibility will be too confusing as it may happen very often?

newIDE/app/src/InstancesEditor/ViewPosition.js
ClementPasteau: reverted

newIDE/app/src/InstancesEditor/FullSizeInstancesEditorWithScrollbars.js
```diff
@@ -33,154 +79,394 @@ type Props = {|
...
+        'mousedown',
+        mouseDownXThumbHandler
+      );
+      yScrollbarThumbElement.addEventListener(
+        'mousedown',
+        mouseDownYThumbHandler
+      );
+    },
+    [mouseDownXThumbHandler, mouseDownYThumbHandler]
+  );
```
AlexandreSi, line 285: Shouldn't there be an empty list of dependencies here?

AlexandreSi (approved):
LGTM!
Have you tried it on mobile with the stories?

newIDE/app/src/InstancesEditor/FullSizeInstancesEditorWithScrollbars.js
```diff
@@ -33,154 +79,394 @@ type Props = {|
...
+
+      xScrollbarThumbElement.addEventListener(
+        'mousedown',
+        mouseDownXThumbHandler
+      );
+      yScrollbarThumbElement.addEventListener(
+        'mousedown',
+        mouseDownYThumbHandler
+      );
+    },
```
4ian, line 283: Should you unregister the event listener? Or don't you do it because it will be destroyed when the element is destroyed?

4ian (commented):
Do you think the mouseup/down/etc event should be written as pointerup/pointerdown/etc to handle potential "pointers" (pencils on tablets?) instead of just the mouse?

Or don't we care because on these (tablets, mobile), we pinch to zoom/move anyway?

newIDE/app/src/InstancesEditor/FullSizeInstancesEditorWithScrollbars.js
```diff
@@ -33,154 +79,394 @@ type Props = {|
...
+      {({ width, height }) => (
+        <div style={styles.container}>
+          <InstancesEditor
+            onViewPositionChanged={
+              screenType !== 'touch' ? handleViewPositionChange : noop
+            }
+            ref={(editor: ?InstancesEditor) => {
+              wrappedEditorRef && wrappedEditorRef(editor);
+              if (editor) {
+                editorRef.current = editor;
```
4ian, line 412: Isn't this risking to keep an outdated editorRef.current, because of the if (editor) that will prevent editorRef.current to be reset to null? Seem like an edge case an unlikely to happen, but I prefer to understand if there is a real reason to do that - better to have a null error later rather than faking the fact that everything is alright by preventing null to be set.

ClementPasteau:
> Do you think the mouseup/down/etc event should be written as pointerup/pointerdown/etc to handle potential "pointers" (pencils on tablets?) instead of just the mouse?
> 
> Or don't we care because on these (tablets, mobile), we pinch to zoom/move anyway?

I'd say we don't display them on mobile (that was the case with previous scrollbars too)

newIDE/app/src/InstancesEditor/FullSizeInstancesEditorWithScrollbars.js
ClementPasteau: I haven't changed this part of the code.
I don't think the result differs tbh, as editorRef.current is null to start with, setting it or not doesn't change the result?
I'm not sure why it was done like this. I'll update

newIDE/app/src/InstancesEditor/FullSizeInstancesEditorWithScrollbars.js
ClementPasteau: Tested. It is removed from the DOM only when the editor tab is closed, and the event listeners are removed because destroyed. 👍

ClementPasteau: tested when changing from one editor to the other, no conflict.

PR #4386, opened 2022-10-12
Add autocompletion for timers, tweens and other extensions.

D8H:
It allows extensions to declare a parameter as an identifier. Expressions filled in parameters of the same identifier will be use for autocompletion (in a similar way as undeclared variables).

image

image


Core/GDCore/IDE/Events/EventsIdentifiersFinder.cpp
```diff
@@ -0,0 +1,254 @@
...
+std::set<gd::String> EventsIdentifiersFinder::FindAllIdentifierExpressions(
+    const gd::Platform& platform,
+    gd::Project& project,
+    gd::Layout& layout,
+    const gd::String& identifierType,
+    const gd::String& contextObjectName) {
+  std::set<gd::String> results;
+
+  const bool isObjectIdentifier = identifierType.find("object") == 0;
+  // The object from the context is only relevent for object identifiers.
```
4ian, line 205:
```suggestion
  // The object from the context is only relevant for object identifiers.
```

Core/tests/EventsVariablesFinder.cpp
```diff
@@ -0,0 +1,360 @@
...
+} // namespace
+
+TEST_CASE("EventsVariablesFinder (FindAllGlobalVariables)", "[common]") {
+  SECTION("Can find global variables in scenes") {
+    gd::Project project;
+    gd::Platform platform;
+    DeclareVariableExtension(project, platform);
+
+    auto &layout = project.InsertNewLayout("Layout1", 0);
+    layout.GetEvents().InsertEvent(UseGlobalVariable("MyGlobalVariable"));
```
4ian, line 124: Nice, this construction is very clear

newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/ExpressionAutocompletionsDisplayer.js
```diff
@@ -313,6 +313,7 @@ export default function ExpressionAutocompletionsDisplayer({
                     : undefined;
 
                   return expressionAutocompletion.kind === 'Text' ||
+                    expressionAutocompletion.kind === 'FullExpression' ||
```
4ian, line 316: Can you explain why this is useful exactly?

4ian (commented):
I've not checked everything in details but overall lgtm!

newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/ExpressionAutocompletionsDisplayer.js
D8H: The autocompletion is not a text but an expression. It will suggest things like `"MyTimer" + VariableString(Index)`. It must not add quote around the autocompletion. I had to define a new kind of completion because of this.

Core/GDCore/IDE/Events/EventsVariablesFinder.h
```diff
@@ -55,8 +54,8 @@ class EventsVariablesFinder {
    */
   static std::set<gd::String> FindAllLayoutVariables(
       const gd::Platform& platform,
-      const gd::Project& project,
-      const gd::Layout& layout);
+      gd::Project& project,
```
4ian, line 57: Why no const anymore?

4ian (approved):
Looks good once we're sure we can't do something to keep the const.

Core/GDCore/IDE/Events/EventsVariablesFinder.h
D8H: Ho, yes, I wanted to ask you. Do you think we should make a duplicated implementation of `ArbitraryEventsWorkerWithContext` for read-only ones?

4ian: That would make sense. We could have a `ReadOnlyArbitraryEventsWorker` and a `ReadOnlyArbitraryEventsWorkerWithContext`. 
Of course, it means doubling the implementation - which is not super great (and at the same time not too much code...).

You could either do that or choose to implement ReadOnlyArbitraryEventsWorker by reusing `ArbitraryEventsWorker` and using `const_cast` to force passing a const argument to a non const argument.

For example, `void Launch(const gd::EventsList& events) { Launch(const_cast<...>(events); };` to have the const version of Launch to launch the Non Const version (because we can guarantee it won't do any non const calls) and reimplement all the VisitXxx as non const version calling the new const version....

At the end it's probably simpler/safer to duplicate the code :)

PR #4387, opened 2022-10-12
Fix leaderboards going crazy

AlexandreSi:


Extensions/Leaderboards/leaderboardstools.ts
```diff
@@ -279,30 +280,42 @@ namespace gdjs {
         let scoreSavingState: ScoreSavingState;
         if (_scoreSavingStateByLeaderboard[leaderboardId]) {
           scoreSavingState = _scoreSavingStateByLeaderboard[leaderboardId];
+          let shouldStartSaving = true;
           if (
+            shouldStartSaving &&
```
ClementPasteau, line 285: this is probably not necessary, as you just created the value to true

ClementPasteau (approved):
nice!

Extensions/Leaderboards/leaderboardstools.ts
AlexandreSi: In case someone starts moving those 3 ifs around, they won't get unexpected errors

PR #4388, opened 2022-10-12
Move some stories to LayoutEditor and ObjectEditor

D8H:
It also adds some object editors that were missing.

This is extracted from
- https://github.com/4ian/GDevelop/pull/4363

It will be useful when we rework the UI.

http://gdevelop-storybook.s3-website-us-east-1.amazonaws.com/stories-object-layout/latest/index.html?path=/story/objecteditor-objecteditordialog--custom-object


newIDE/app/src/stories/componentStories/LayoutEditor/LayersList.stories.js
```diff
@@ -0,0 +1,65 @@
+// @flow
+
+import * as React from 'react';
+import { action } from '@storybook/addon-actions';
+
+// Keep first as it creates the `global.gd` object:
```
4ian, line 6: You're sure the order is important?

4ian (approved):
👍

newIDE/app/src/stories/componentStories/LayoutEditor/LayersList.stories.js
D8H: I don't know, I don't understand this comment. I only copy-paste it with the import that was in the main story.

PR #4399, opened 2022-10-16
Fix grammar; change 'let' to 'leave'.

BWPanda:
"it's recommended to let this unchecked" isn't proper grammar. Should be "it's recommended to **leave** this unchecked" instead.


tristanbob:
I agree that this wording is an improvement.  Thanks for your contribution, @BWPanda!

4ian (approved):

4ian:
Thanks a lot!

PR #4411, opened 2022-10-19
Improve new scrollbars on scene editor canvas

AlexandreSi:
- Make it possible to click and scroll through the "scrollbar detection zones"
- Make mosaic horizontal resize handle easier to grab
- Use same style as for other scrollbars around the editor (+ hover and active states)

Tested on Chrome and Safari

Storybook to test: http://gdevelop-storybook.s3-website-us-east-1.amazonaws.com/fix-scroll/latest/index.html?path=/story/editor-fullsizeinstanceseditorwithscrollbars--default

### DIFF
--- a/newIDE/app/src/InstancesEditor/FullSizeInstancesEditorWithScrollbars.js
+++ b/newIDE/app/src/InstancesEditor/FullSizeInstancesEditorWithScrollbars.js
@@ -84,34 +84,45 @@ const FullSizeInstancesEditorWithScrollbars = (props: Props) => {
     hideScrollbarsAfterDelay();
   }, 500);
 
+  const showScrollbarsThrottled = throttle(
+    () => {
+      if (!showScrollbars.current) {
+        showScrollbars.current = true;
+        forceUpdate();
+      }
+      if (timeoutHidingScrollbarsId.current) {
+        clearTimeout(timeoutHidingScrollbarsId.current);
+        timeoutHidingScrollbarsId.current = null;
+      }
+    },
+    1000,
+    { leading: true, trailing: false }
+  );
+
   const onMouseMoveOverInstanceEditor = React.useCallback(
     (event: MouseEvent) => {
       if (!editorRef.current) {
         return;
       }
-      let shouldDisplayScrollBars = false;
 
-      shouldDisplayScrollBars = !canvasRectangle.containsPoint(
+      const shouldDisplayScrollBars = !canvasRectangle.containsPoint(
         event.clientX,
         event.clientY
       );
 
       if (shouldDisplayScrollBars) {
-        if (!showScrollbars.current) {
-          showScrollbars.current = true;
-          forceUpdate();
-        }
-        if (timeoutHidingScrollbarsId.current) {
-          clearTimeout(timeoutHidingScrollbarsId.current);
-          timeoutHidingScrollbarsId.current = null;
-        }
+        showScrollbarsThrottled();
       } else {
         if (!isDragging.current) {
-          hideScrollbarsAfterDelay();
+          hideScrollbarsAfterDelayDebounced();
         }
       }
     },
-    [canvasRectangle, forceUpdate, hideScrollbarsAfterDelay]
+    [
+      canvasRectangle,
+      hideScrollbarsAfterDelayDebounced,
+      showScrollbarsThrottled,
+    ]
   );
 
   // When the mouse is moving after dragging the thumb:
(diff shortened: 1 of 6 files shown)


newIDE/app/src/InstancesEditor/FullSizeInstancesEditorWithScrollbars.js
```diff
@@ -107,6 +55,18 @@ const FullSizeInstancesEditorWithScrollbars = (props: Props) => {
...
   const forceUpdate = useForceUpdate();
 
+  const canvasRectangle = React.useMemo(() => new Rectangle(), []);
+
+  if (editorRef.current) {
+    const elementBoundingRect = editorRef.current.getBoundingClientRect();
+    canvasRectangle.set({
+      left: elementBoundingRect.left,
+      top: elementBoundingRect.top,
+      right: elementBoundingRect.right - SCROLLBAR_DETECTION_WIDTH,
```
ClementPasteau, line 65: if one comes back from outside the canvas onto the scrollbars, do they appear?

newIDE/app/src/InstancesEditor/FullSizeInstancesEditorWithScrollbars.js
```diff
@@ -124,14 +84,51 @@ const FullSizeInstancesEditorWithScrollbars = (props: Props) => {
...
+      if (!editorRef.current) {
+        return;
+      }
+
+      const shouldDisplayScrollBars = !canvasRectangle.containsPoint(
+        event.clientX,
+        event.clientY
+      );
+
+      if (shouldDisplayScrollBars) {
```
ClementPasteau, line 119: should we add a check here on `showScrollbars.current` so that if it's already true, you don't trigger this function?

newIDE/app/src/InstancesEditor/FullSizeInstancesEditorWithScrollbars.js
```diff
@@ -124,14 +84,51 @@ const FullSizeInstancesEditorWithScrollbars = (props: Props) => {
...
+        showScrollbarsThrottled();
+      } else {
+        if (!isDragging.current) {
+          hideScrollbarsAfterDelayThrottled();
```
ClementPasteau, line 123: same here, if `showScrollbars.current` is already false, we probably don't need to call it? It means they're just moving around

newIDE/app/src/InstancesEditor/index.js
```diff
@@ -848,13 +861,19 @@ export default class InstancesEditor extends Component<Props> {
       const idealZoom = this.viewPosition.fitToRectangle(contentAABB);
       this.setZoomFactor(idealZoom);
     }
+    if (this.props.onViewPositionChanged) {
+      this.props.onViewPositionChanged(this.viewPosition);
+    }
```
ClementPasteau, line 866: maybe this logic you've done 3 times could be in the `setZoomFactor()` function

newIDE/app/src/UI/EditorMosaic/style.css
```diff
@@ -23,7 +23,7 @@
 
 /* Make the horizontal split to have the same thickness as the vertical */
 .mosaic-gd-theme .mosaic-split.-column {
-  margin-top: -2px;
+  margin-top: -3px;
```
ClementPasteau, line 26: hmm, this feels like a random change. Are we sure it doesn't affect anything else?

ClementPasteau:
It looks great 👍 

Just noticed 1 tiny bug:

- if you go over to grab the thumb
- but your mouse leaves the canvas for a millisecond, before coming back to the thumb
- the scrollbars will disappear no matter what you do
- but the next time they appear they stay (if you don't leave the canvas again)

Probably need to cancel any hide timeouts when coming back to the canvas/scrollbars.


https://user-images.githubusercontent.com/4895034/196631320-5a6661f4-3c96-4b61-8f0a-3ae825c72256.mov

newIDE/app/src/InstancesEditor/index.js
4ian: (we need to check in this case the other call sites of setZoomFactor to be sure, if there are any, that it won't impact them)

newIDE/app/src/UI/EditorMosaic/style.css
AlexandreSi: It's not random, it displays enough for it to be the same size as for the vertical handles

newIDE/app/src/InstancesEditor/FullSizeInstancesEditorWithScrollbars.js
AlexandreSi: Yes, because the event listener is one the whole canvas

newIDE/app/src/InstancesEditor/FullSizeInstancesEditorWithScrollbars.js
AlexandreSi: Good idea!

newIDE/app/src/InstancesEditor/index.js
AlexandreSi: Well the need to call onViewPositionChanged is not related to zoom factor, but the view position, so I'll look into calling it in position change functions

ClementPasteau (approved):
LGTM now!

PR #4415, opened 2022-10-20
Prevent pixi from rendering if any dialog is opened

AlexandreSi:


ClementPasteau (approved):
lgtm 👍

newIDE/app/src/InstancesEditor/index.js
```diff
@@ -39,6 +39,12 @@ const styles = {
   dropCursor: { cursor: 'copy' },
 };
 
+const isNoDialogOpened = (): boolean => {
```
4ian, line 42: Do you think this could be an util? Somehow I have the feeling that having this function in its own file would help us, when we upgrade Material-UI one day, to remember/find "oh there is this query selector that is checking a specific class, let's check the name of the class is still up to date"

4ian (approved):
Thousands of CPUs and batteries are thanking you for this!

newIDE/app/src/InstancesEditor/index.js
AlexandreSi: It could a file `UI/MaterialUISpecificUtil.js` ?

4ian: Perfect

PR #4420, opened 2022-10-21
Update instance panel with new layer

AlexandreSi:
This PR make it so that instance properties panel is force updated when a layer is created or renamed.

### DIFF
--- a/newIDE/app/src/SceneEditor/index.js
+++ b/newIDE/app/src/SceneEditor/index.js
@@ -688,6 +688,11 @@ export default class SceneEditor extends React.Component<Props, State> {
   ) => {
     this.props.initialInstances.moveInstancesToLayer(oldName, newName);
     done(true);
+    this.forceUpdatePropertiesEditor();
+  };
+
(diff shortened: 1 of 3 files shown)


newIDE/app/src/SceneEditor/index.js
```diff
@@ -674,6 +674,11 @@ export default class SceneEditor extends React.Component<Props, State> {
   ) => {
     this.props.initialInstances.moveInstancesToLayer(oldName, newName);
     done(true);
+    this.forceUpdatePropertiesEditor();
```
ClementPasteau, line 677: nice! should this be before the done?

4ian (approved):
Lgtm!

D8H (approved):

ClementPasteau (approved):
nice!

newIDE/app/src/SceneEditor/index.js
AlexandreSi: Nope, this callback is what actually renames the layer after all instances has been updated with the new layer name

PR #4428, opened 2022-10-24
Change current file metadata name after renaming a project

AlexandreSi:
Before this PR:
- User renames project
- User saves project
- Still former name in recent project files

After this PR, the name displayed in the recent project files is updated after the save has finished


ClementPasteau (approved):
👍

PR #4432, opened 2022-10-24
Revert service worker update

ClementPasteau:
Doesn't work as expected.

We should look into https://developer.chrome.com/docs/workbox/handling-service-worker-updates/ to:

- add an event listener to the existing sw (maybe using workbox-window package)
- prompt for the same update dialog, when the status goes to "waiting"
- reload page when status goes to "controlling"

### DIFF
--- a/newIDE/app/src/MainFrame/UpdaterTools.js
+++ b/newIDE/app/src/MainFrame/UpdaterTools.js
@@ -1,9 +1,8 @@
 // @flow
 // See ElectronEventsBridge, AboutDialog and electron-app/main.js for handling the updates.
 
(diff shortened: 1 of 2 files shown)


4ian:
> We should look into https://developer.chrome.com/docs/workbox/handling-service-worker-updates/

Note that we can also pass a `config` object to `register` here: https://github.com/4ian/GDevelop/blob/5b1e169557263689315c32bcb622c1e0881f9d76/newIDE/app/src/serviceWorker.js#L25

So maybe this could be useful? Rather than adding another package to listen to things we could already know with this.

4ian (approved):

PR #4433, opened 2022-10-24
Remove starting value in tween variable actions

### LINKED ISSUE #4270: Variable tweens are not setting the variable to starting value at creation.
When creating a variable tween, the starting value is not set as the variable value when creating the tween.
The variable is refreshed only after the next pass.

And this example quits instantly.
image

The only way to work around is to set the Variable as 255 before the Tween starts OR start the tween AFTER the variable checking 
image

Windows 10 with GDev version 5.0.140

AlexandreSi:
Fixes #4270

### DIFF
--- a/Extensions/TweenBehavior/JsExtension.js
+++ b/Extensions/TweenBehavior/JsExtension.js
@@ -637,7 +637,7 @@ module.exports = {
         'AddObjectScaleTween',
         _('Tween object scale'),
         _(
-          'Tweens an object scale from its current scale to a new one (Note: the scale can never be less than 0).'
+          'Tweens an object scale from its current scale to a new one (note: the scale can never be less than 0).'
         ),
         _(
           'Tween the scale of _PARAM0_ to X-scale: _PARAM3_, Y-scale: _PARAM4_ (from center: _PARAM8_) with easing _PARAM5_ over _PARAM6_ms as _PARAM2_'
@@ -671,7 +671,7 @@ module.exports = {
         'AddObjectScaleXTween',
         _('Tween object X-scale'),
         _(
-          'Tweens an object X-scale from its current value to a new one (Note: the scale can never be less than 0).'
+          'Tweens an object X-scale from its current value to a new one (note: the scale can never be less than 0).'
         ),
         _(
           'Tween the X-scale of _PARAM0_ to _PARAM3_ (from center: _PARAM7_) with easing _PARAM4_ over _PARAM5_ms as _PARAM2_'
@@ -704,7 +704,7 @@ module.exports = {
         'AddObjectScaleYTween',
         _('Tween object Y-scale'),
         _(
-          'Tweens an object Y-scale from its current value to a new one (Note: the scale can never be less than 0).'
+          'Tweens an object Y-scale from its current value to a new one (note: the scale can never be less than 0).'
         ),
         _(
           'Tween the Y-scale of _PARAM0_ to _PARAM3_ (from center: _PARAM7_) with easing _PARAM4_ over _PARAM5_ms as _PARAM2_'
@@ -737,7 +737,7 @@ module.exports = {
         'AddTextObjectCharacterSizeTween',
         _('Tween text size'),
         _(
-          'Tweens the text object character size from tis current value to a new one (Note: the size can never be less than 1).'
+          'Tweens the text object character size from its current value to a new one (note: the size can never be less than 1).'
         ),
         _(
           'Tween the character size of _PARAM0_ to _PARAM3_ with easing _PARAM4_ over _PARAM5_ms as _PARAM2_'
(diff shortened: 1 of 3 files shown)


Extensions/TweenBehavior/JsExtension.js
```diff
@@ -118,19 +118,46 @@ module.exports = {
...
       .addIncludeFile('Extensions/TweenBehavior/shifty_setup.js')
       .addIncludeFile('Extensions/TweenBehavior/tweentools.js')
       .setFunctionName('gdjs.evtTools.tween.tweenVariableNumber');
 
+    extension
+      .addAction(
+        'TweenSceneVariableNumber2',
+        _('Tween a number in a scene variable'),
+        _(
+          "Tweens a scene variable's numeric value from current value to another."
```
4ian, line 133:
```suggestion
          "Tweens a scene variable's numeric value from its current value to another."
```
?

4ian (approved):

Extensions/TweenBehavior/JsExtension.js
```diff
@@ -674,9 +735,9 @@ module.exports = {
     behavior
       .addAction(
         'AddTextObjectCharacterSizeTween',
-        _('Add text size tween'),
+        _('Tween text size'),
         _(
-          'Add a tween animation for the text object character size (Note: the size can never be less than 1).'
+          'Tweens the text object character size from tis current value to a new one (Note: the size can never be less than 1).'
```
4ian, line 740:
```suggestion
          'Tweens the text object character size from its current value to a new one (note: the size can never be less than 1).'
```

Extensions/TweenBehavior/JsExtension.js
```diff
@@ -705,9 +766,9 @@ module.exports = {
     behavior
       .addAction(
         'AddObjectOpacityTween',
-        _('Add object opacity tween'),
+        _('Tween object opacity'),
         _(
-          'Add a tween animation for the object opacity (Value between 0 and 255).'
+          'Tweens the object opacity from its current value to a new one (Note: the value shall stay between 0 and 255).'
```
4ian, line 771:
```suggestion
          'Tweens the object opacity from its current value to a new one (note: the value shall stay between 0 and 255).'
```

4ian (approved):

Extensions/TweenBehavior/JsExtension.js
```diff
@@ -179,7 +207,7 @@ module.exports = {
         _('Tween the camera rotation'),
         _('Tweens the camera rotation from the current angle to a new one.'),
         _(
-          'Tween the rotation of camera on layer _PARAM3_ to _PARAM2_ for _PARAM4_ms with easing _PARAM5_ as _PARAM1_'
+          'Tween the rotation of camera on layer _PARAM3_ to _PARAM2_ over _PARAM4_ms with easing _PARAM5_ as _PARAM1_'
```
ClementPasteau, line 210: Just to confirm, does using `_PARAM4_ms` display "ms" in the editor?

AlexandreSi: Yes!

Extensions/TweenBehavior/tweentools.ts
```diff
@@ -158,6 +161,26 @@ namespace gdjs {
         getShiftyScene(runtimeScene).add(tween);
       };
 
+      export const tweenVariableNumber2 = (
```
ClementPasteau, line 164: maybe add a small comment above this function explaining why we use this one now and not the previously deprecated one?

Extensions/TweenBehavior/tweenruntimebehavior.ts
```diff
@@ -239,6 +240,39 @@ namespace gdjs {
...
+    /**
+     * Add an object variable tween.
+     * @param identifier Unique id to identify the tween
+     * @param variable The object variable to store the tweened value
+     * @param toValue End value
+     * @param easingValue Type of easing
+     * @param durationValue Duration in milliseconds
+     * @param destroyObjectWhenFinished Destroy this object when the tween ends
+     */
+    addVariableTween2(
```
ClementPasteau, line 252: same here

ClementPasteau (approved):

Extensions/TweenBehavior/tweenruntimebehavior.ts
4ian: 👍

Extensions/TweenBehavior/tweentools.ts
4ian: 👍

AlexandreSi: I added a detailed comment on the deprecated function

Extensions/TweenBehavior/tweenruntimebehavior.ts
AlexandreSi: I added a detailed comment on the deprecated function

PR #4434, opened 2022-10-25
Fix events sheet not wrapping on small screens

AlexandreSi:


ClementPasteau (approved):
lgtm, can you share a screenshot of the change?

AlexandreSi:
Before:
image


After:
image

tristanbob:
This makes it hard to tell what is a condition vs action.  What could we do visually to help with this?  I see it is a darker color, but how about a slight indentation also?  

I wonder what sub-events will look like.

AlexandreSi:
There's already the left padding for the actions, you can see they're a bit off. I don't know if it's enough.
This display was the default one before a regression was introduced, so I just fixed it.
But the mobile display of the events sheet can be a whole discussion to have

PR #4437, opened 2022-10-25
Refactor to make some ParameterMetadata attributes private

D8H:
Refactor to make some ParameterMetadata attributes private (type, supplementaryInformation and optional).

It will allow to extract a class for type metadata and keep ParameterMetadata interface the same to avoid to refactor in a PR that add features (like an expression type or unit definition)


4ian (approved):
Great stuff 👌

Issue #4351, opened 2022-10-03
Game thumbnail not working


Bouh:
## Describe the bug
The game thumbnail setup in the game didn't work at 100%.

In rare use cases, the thumbnail isn't working on Liluo.

On this game: [https://liluo.io/jianforest/forestkeepers](https://liluo.io/jianforest/forestkeepers)
The thumbnail is set with:
![image](https://user-images.githubusercontent.com/1670670/193664493-c4106571-f683-4ed5-9fc7-19ec6927f1b4.png)

The game loads the image correctly:
```
https://games.gdevelop-app.com/game-810a6244-e7e8-4b42-9f5a-22014deab37e/The%20Forest%20Keepers%20(thumb)%20by%20Jian%20Forest.png
```

The resource in the game data is:
![image](https://user-images.githubusercontent.com/1670670/193665026-bbbf7c76-2354-4eaa-b56b-51cb27b0dfd2.png)

But there is nothing on Liluo:
![image](https://user-images.githubusercontent.com/1670670/193664548-307fd141-06c1-4c33-9177-fd42d45d08a0.png)

I presume the issue is with the special characters or the spaces.
There is also a double backslash in the URL, which might be the cause too!

## Other details
GDevelop 5

Yash-Punia:
Even I think that the main issue are the two backslash appended by windows file management system, did you try using a single forwardslash? I can't reproduce the same problem due to lack of info. I tried using both local web and electron too, and both seem to work fine for me

AlexandreSi:
Hi!
I think it might be due to something missing in the editor: the liluo thumbnail is actually not updated when changing it from the "Project icons" dialog. It should be changed the next time the game is published on Liluo though.

Have you tried publish a new version on Liluo?

Anyway, I'll add a message to explain this until we find a proper solution

Oxey405:
So is this issue resolved then ?

AlexandreSi:
The issue is not really resolved. I just added a warning so that one can be "unblocked".

There is no technical bug, but there is a User Experience bug: we say that if you change the image, it will be updated on Liluo. But **it doesn't** so the behaviour is buggy here.

This is due to the image source we use on Liluo that is - at the moment - the one present in the game build. So this image present on Liluo cannot be updated on the fly.

This issue cannot be solved by someone outside of the core team.

PR #4438, opened 2022-10-25
Add warning message when updating liluo.io thumbnail from the project icons dialog

AlexandreSi:
After a change is detected in the resource selector, this is displayed:
image

I'm not sure the copy is the best though.

This should prevent some misunderstanding related to #4351


ClementPasteau (approved):
copy looks fine to me...

PR #4440, opened 2022-10-25
Make popovers more visible + a few UI improvements

AlexandreSi:
The idea is to make the popovers and poppers more discernable from the underlying background color.

So far:

### Before

image

### After

image

#### 2 panels

image

#### 1 panel

image

#### 2 panels

image

#### 1 panel

image

### Before

image

### After

image

# Also

- Use same DragNDrop library for layers list as for objects and objects groups.
- Improve user experience when renaming layers
- Add object icon in objects group object list:

image

### DIFF
--- a/newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/ExpressionAutocompletionsDisplayer.js
+++ b/newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/ExpressionAutocompletionsDisplayer.js
@@ -14,7 +14,7 @@ import { type ParameterRenderingServiceType } from '../ParameterFieldCommons';
 import { type EnumeratedInstructionOrExpressionMetadata } from '../../../InstructionOrExpression/EnumeratedInstructionOrExpressionMetadata';
 import { Column, Line, Spacer } from '../../../UI/Grid';
 import ObjectsRenderingService from '../../../ObjectsRendering/ObjectsRenderingService';
-import { useTheme } from '@material-ui/styles';
+import GDevelopThemeContext from '../../../UI/Theme/ThemeContext';
 
 const defaultTextStyle = {
   // Break words if they are too long to fit on a single line.
@@ -274,8 +274,7 @@ export default function ExpressionAutocompletionsDisplayer({
   onScroll,
   parameterRenderingService,
 }: Props) {
-  const muiTheme = useTheme();
-
+  const gdevelopTheme = React.useContext(GDevelopThemeContext);
   const scrollView = React.useRef((null: ?ScrollViewInterface));
   const selectedAutocompletionElement = React.useRef(
     (null: ?React$Component<any, any>)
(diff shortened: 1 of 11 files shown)


newIDE/app/src/LayersList/index.js
```diff
@@ -20,96 +19,149 @@ import Background from '../UI/Background';
...
+  });
 
-const SortableLayersListBody = SortableContainer(LayersListBody);
+  return (
+    <Column noMargin expand>
+      {containerLayersList}
+      <DropTarget
+        canDrop={() => true}
+        drop={() => {
+          onDropLayer(-1);
```
4ian, line 140: Would null work?

newIDE/app/src/LayersList/index.js
```diff
@@ -20,96 +19,149 @@ import Background from '../UI/Background';
...
+        draggedLayerIndex,
+        targetIndex < draggedLayerIndex ? targetIndex + 1 : targetIndex
       );
-    });
+      onLayerModified();
+    }
+    draggedLayerIndexRef.current = null;
+  };
+
+  const DropTarget = React.useMemo(() => makeDropTarget('layers-list'), []);
```
4ian, line 76: I think this would work even outside of the component

newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/ExpressionAutocompletionsDisplayer.js
```diff
@@ -273,6 +274,8 @@ export default function ExpressionAutocompletionsDisplayer({
   onScroll,
   parameterRenderingService,
 }: Props) {
+  const muiTheme = useTheme();
```
4ian, line 277: Some would say you could use the gdevelopTheme?

4ian (approved):

newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/ExpressionAutocompletionsDisplayer.js
AlexandreSi: Well, the alternate background color is used for `gdevelopTheme.list.itemsBackgroundColor` and `gdevelop.searchBar.backgroundColor`.
So I could define a new item in the GDevelop theme that would be something like `gdevelopTheme.palette.background.alternateColor`.

Moreover this `useTheme` was used in only one place (`SelectOption`). So maybe I can also make sure this single use uses the gdevelop theme instead.

newIDE/app/src/LayersList/index.js
AlexandreSi: Not sure, I'll check

newIDE/app/src/LayersList/index.js
AlexandreSi: True, I'll change it

PR #4444, opened 2022-10-27
Fix animation ended condition

AlexandreSi:
It was reported by @HelperWesley that the "Animation Finished" condition fires when an animation enters its last frame, not when the last frame has been displayed long enough for the animation to be considered finished.

This PR fixes this:
- Hide current condition to prevent regression
- Create identical condition that behaves correctly

## Example

With this event sheet:

image

Here is what happens:

https://user-images.githubusercontent.com/32449369/198299256-8cc6be6e-2d15-43c6-8212-c1864758b145.mov


ClementPasteau (approved):
nice!

4ian (approved):
Lgtm!

PR #4448, opened 2022-10-27
Modernize scene properties to look the same as behavior properties

D8H:
The story is not really useful because Physics 2 can't be used but we will be able to show shared data of event-based behaviors in the future.
http://gdevelop-storybook.s3-website-us-east-1.amazonaws.com/shared-property-editor/latest/index.html?path=/story/layouteditor-scenepropertiesdialog--default

It uses the PropertiesEditor to give the same level of features as the behavior property editor.

image

### DIFF
--- a/newIDE/app/src/SceneEditor/BehaviorSharedPropertiesEditor.js
+++ b/newIDE/app/src/SceneEditor/BehaviorSharedPropertiesEditor.js
@@ -1,0 +1,53 @@
+// @flow
+import { Trans } from '@lingui/macro';
+
+import * as React from 'react';
+import PropertiesEditor from '../PropertiesEditor';
+import propertiesMapToSchema from '../PropertiesEditor/PropertiesMapToSchema';
+import EmptyMessage from '../UI/EmptyMessage';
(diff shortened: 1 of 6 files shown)


newIDE/app/src/SceneEditor/BehaviorSharedPropertiesEditor.js
```diff
@@ -0,0 +1,53 @@
...
+
+    return (
+      <Column expand>
+        {propertiesSchema.length ? (
+          <PropertiesEditor
+            schema={propertiesSchema}
+            instances={[behaviorSharedData]}
+          />
+        ) : (
+          <EmptyMessage>
```
4ian, line 43: This will show that for all behaviors having shared data (so almost none) right?

4ian (approved):

newIDE/app/src/SceneEditor/BehaviorSharedPropertiesEditor.js
D8H: Yes, this could probably be removed but it's safer to keep it.

PR #4453, opened 2022-10-28
Add back parts of a revert for the TextImput fix.

D8H:
Sorry about the bug. I shouldn't have tried to do this refactor during a big PR. 

Following:
- https://github.com/4ian/GDevelop/pull/4450
- https://github.com/4ian/GDevelop/pull/4294

Manually tested on an example.

image

image

### DIFF
--- a/GDJS/Runtime/pixi-renderers/runtimegame-pixi-renderer.ts
+++ b/GDJS/Runtime/pixi-renderers/runtimegame-pixi-renderer.ts
@@ -409,20 +409,22 @@ namespace gdjs {
      * Convert a point from the canvas coordinates to the dom element container coordinates.
      *
      * @param canvasCoords The point in the canvas coordinates.
+     * @param result The point to return.
      * @returns The point in the dom element container coordinates.
      */
     convertCanvasToDomElementContainerCoords(
-      canvasCoords: FloatPoint
+      canvasCoords: FloatPoint,
+      result: FloatPoint
     ): FloatPoint {
-      const pageCoords: FloatPoint = [0, 0];
-      const gameResolutionWidth = this._game.getGameResolutionWidth();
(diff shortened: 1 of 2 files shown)


GDJS/Runtime/pixi-renderers/runtimegame-pixi-renderer.ts
```diff
@@ -409,20 +409,22 @@ namespace gdjs {
      * Convert a point from the canvas coordinates to the dom element container coordinates.
      *
      * @param canvasCoords The point in the canvas coordinates.
+     * @param result The point to return.
      * @returns The point in the dom element container coordinates.
      */
     convertCanvasToDomElementContainerCoords(
-      canvasCoords: FloatPoint
+      canvasCoords: FloatPoint,
+      result: FloatPoint
```
ClementPasteau, line 417: I don't understand the logic, is this more performant passing the same variable twice?
If so, should it not return the object and just modify it?

D8H: This is a micro optimization to avoid allocating a FloatPoint. It's possible that the JS engine is already optimized to make this useless but it's hard to understand why GC happens so avoiding as many allocations as possible is a safe bet in my opinion.

About the double parameters, the caller might want to keep the point to convert. This is why the 2nd parameter is needed: to store the result and keep the 1st parameter untouched when needed.

4ian: In a way, you can consider these "result" parameters (that you can find in other functions too) as the memory where the function must operate.
Microoptimization, but avoiding unnecessary optimisation seems safe and important if it's not impacting the readability too much.

4ian (approved):

PR #4457, opened 2022-10-28
Handle new format for public asset resources

ClementPasteau:
The main goal is to allow easy modification of existing public assets, without impacting projects that may be using them.

The strategy picked is to add a unique id to the files, corresponding the sha of the file (thus, when it changes, the file changes without affecting the existing one)

This PR is to handle those new file formats.

### DIFF
--- a/newIDE/app/src/ProjectsStorage/LocalFileStorageProvider/LocalFileResourceMover.js
+++ b/newIDE/app/src/ProjectsStorage/LocalFileStorageProvider/LocalFileResourceMover.js
@@ -5,25 +5,18 @@ import { retryIfFailed } from '../../Utils/RetryIfFailed';
 import newNameGenerator from '../../Utils/NewNameGenerator';
 import { type FileMetadata } from '../index';
 import {
-  extractFilenameAndExtensionFromProductAuthorizedUrl,
+  extractFilenameWithExtensionFromProductAuthorizedUrl,
   isProductAuthorizedResourceUrl,
 } from '../../Utils/GDevelopServices/Shop';
 import {
-  extractFilenameAndExtensionFromPublicAssetResourceUrl,
+  extractFilenameWithExtensionFromPublicAssetResourceUrl,
   isPublicAssetResourceUrl,
 } from '../../Utils/GDevelopServices/Asset';
+import { isFetchableUrl } from '../../ResourcesList/ResourceUtils';
 const electron = optionalRequire('electron');
 const ipcRenderer = electron ? electron.ipcRenderer : null;
 const fs = optionalRequire('fs-extra');
 const path = optionalRequire('path');
-
-const isFetchableUrl = (filename: string) => {
-  return (
-    filename.startsWith('http://') ||
-    filename.startsWith('https://') ||
-    filename.startsWith('ftp://')
-  );
-};
 
 type Options = {|
   project: gdProject,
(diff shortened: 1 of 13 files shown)


newIDE/app/src/Utils/GDevelopServices/Asset.js
```diff
@@ -328,3 +332,30 @@ export const listReceivedAssetPacks = async (
...
+const escapeStringForRegExp = string =>
+  string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'); // $& means the whole matched string
+const resourceFilenameRegex = new RegExp(
+  `(${escapeStringForRegExp(
+    GDevelopPublicAssetResourcesStorageBaseUrl
+  )}|${escapeStringForRegExp(
+    GDevelopPublicAssetResourcesStorageStagingBaseUrl
+  )})\\/public-resources\\/(.*)\\/([a-z0-9]{64})_(.*)(\\..*)`
+);
+export const extractFilenameAndExtensionFromPublicAssetResourceUrl = (
```
4ian, line 348: Something that I wanted to do in my still not merged PR about cloud projects on desktop:
don't return "filename and extension", just return the filename with the extension (`myfile.png`).

The caller can extract the extension if they wish (but it's not always the case, at one point the code is concatenating them just after getting them!)

ClementPasteau: true, I'll change that.

newIDE/app/src/ProjectsStorage/ResourceFetcher/index.js
```diff
@@ -39,14 +57,47 @@ type UseResourceFetcherOutput = {|
...
-    options: FetchAllProjectResourcesOptionsWithoutProgress
+    options: CleanAndFetchAllProjectResourcesOptionsWithoutProgress
   ) => Promise<void>,
   /**
    * Render, if needed, the dialog that will show the progress of resources fetching.
    */
   renderResourceFetcherDialog: () => React.Node,
 |};
 
+export const cleanUpResourceNames = ({
```
ClementPasteau, line 68: I've tried creating this function to call it in both fetchers.
The goal is to clean up the names of the resources (and remove the possible unwanted token & sha) before asking the fetcher to fetch it.
Wdyt of the strategy? @4ian

newIDE/app/src/ProjectsStorage/LocalFileStorageProvider/LocalFileResourceMover.js
```diff
@@ -53,25 +53,45 @@ export const moveUrlResourcesToLocalFiles = async ({
       const resource = resourcesManager.getResource(resourceName);
 
       const url = resource.getFile();
-      const {
-        extension,
-        filenameWithoutExtension,
-      } = extractFilenameAndExtensionFromProductAuthorizedUrl(url);
+      let filenameWithExtension;
+      if (isProductAuthorizedResourceUrl(url)) {
```
4ian, line 57: Very clear

newIDE/app/src/ProjectsStorage/ResourceFetcher/index.js
```diff
@@ -39,14 +57,47 @@ type UseResourceFetcherOutput = {|
...
+        resourceFile
+      );
+    } else if (isPublicAssetResourceUrl(resourceFile)) {
+      // Resource is a public asset.
+      cleanedUpFilename = extractFilenameWithExtensionFromPublicAssetResourceUrl(
+        resourceFile
+      );
+    } else {
+      // Resource is a generic url.
+      cleanedUpFilename = path.basename(resourceFile);
```
4ian, line 95: Could we just have this being factored into a function, and reused whenever need in the various resource fetchers?

4ian (approved):

PR #4458, opened 2022-10-28
Fix importing path correctly on web

ClementPasteau:


newIDE/app/src/Utils/GDevelopServices/Shop.js
```diff
@@ -1,8 +1,7 @@
 // @flow
 import axios from 'axios';
+import path from 'path';
```
4ian, line 3: Yes this is indeed correct because we're dealing with paths from URLs

4ian (approved):

PR #4464, opened 2022-10-30
Allow event-based behaviors to declare scene properties

D8H:
It will allow:
- behaviors that need a shared structure (the NavMesh pathfinding for instance)
- a workaround for local variables
  Behaviors use private properties for indexes or intermediate calculus. It pollutes the property list and takes more memory than needed.

Event-based object won't let users configure default values but it's not an issue when they are private properties and real shared properties are rare and none really apply to EBO cases.

http://gdevelop-storybook.s3-website-us-east-1.amazonaws.com/shared-property/latest/index.html?path=/story/eventsbasedbehavioreditor-eventsbasedbehavioreditordialog--default


ClementPasteau:
@4ian could you take a quick look at this when you have time?

Core/GDCore/IDE/WholeProjectRefactorer.h
```diff
@@ -207,6 +207,21 @@ class GD_CORE_API WholeProjectRefactorer {
       const gd::String& oldPropertyName,
       const gd::String& newPropertyName);
 
+  /**
+   * \brief Refactor the project **before** a property of a behavior is
```
4ian, line 211: a property => a shared property

Core/GDCore/Project/CustomBehaviorsSharedData.h
```diff
@@ -0,0 +1,51 @@
...
+#include "GDCore/Project/Project.h"
+#include "GDCore/Project/PropertyDescriptor.h"
+#include "GDCore/Serialization/Serializer.h"
+#include "GDCore/Serialization/SerializerElement.h"
+
+using namespace gd;
+
+namespace gd {
+/**
+ * \brief A gd::Behavior that stores its content in JSON and forward the
```
4ian, line 20: Update this comment :)

GDJS/GDJS/Events/CodeGeneration/BehaviorCodeGenerator.h
```diff
@@ -57,32 +57,69 @@ class BehaviorCodeGenerator {
     return "_set" + propertyName;
   }
 
+  /**
+   * \brief Generate the name of the method to get the value of the shared property
+   * of a behavior.
+   */
+  static gd::String GetBehaviorSharedPropertyGetterName(
+      const gd::String& propertyName) {
+    return "_getShared" + propertyName;
```
4ian, line 66: We should probably ensure properties are not starting with "Shared"? To avoid conflicts in method names?

newIDE/app/src/EventsFunctionsExtensionsLoader/MetadataDeclarationHelpers.js
```diff
@@ -914,6 +937,43 @@ export const declareBehaviorPropertiesInstructionAndExpressions = (
...
+        gd.BehaviorCodeGenerator.getBehaviorSharedPropertySetterName(
+          propertyName
+        )
+      )
+      .setGetter(
+        gd.BehaviorCodeGenerator.getBehaviorSharedPropertyGetterName(
+          propertyName
+        )
+      )
+      // All property actions/conditions/expressions are private, meaning
```
4ian, line 973: Nit: All shared property

4ian (approved):
Overall LGTM!

GDJS/GDJS/Events/CodeGeneration/BehaviorCodeGenerator.h
D8H: I moved shared properties getter and setter in the shared data class to avoid name collisions.

GDJS/GDJS/Events/CodeGeneration/BehaviorCodeGenerator.cpp
```diff
@@ -263,4 +370,23 @@ gd::String BehaviorCodeGenerator::GenerateDoStepPreEventsPreludeCode() {
...
+
+gd::String BehaviorCodeGenerator::GetBehaviorSharedPropertyGetterInternalName(
+    const gd::String& propertyName) {
+  return "_get" + propertyName;
+}
+
+gd::String BehaviorCodeGenerator::GetBehaviorSharedPropertySetterInternalName(
+    const gd::String& propertyName) {
+  return "_set" + propertyName;
+}
```
D8H, line 391: This is the change I made since the last review.

GDJS/GDJS/Events/CodeGeneration/BehaviorCodeGenerator.cpp
```diff
@@ -119,21 +154,51 @@ gd::String BehaviorCodeGenerator::GenerateRuntimeBehaviorTemplateCode(
...
+CODE_NAMESPACE.RUNTIME_BEHAVIOR_CLASSNAMESharedData = class RUNTIME_BEHAVIOR_CLASSNAMESharedData {
+  constructor(sharedData) {
+    INITIALIZE_SHARED_PROPERTIES_CODE
+  }
+  
+  // Shared properties:
+  SHARED_PROPERTIES_CODE
+}
+
+CODE_NAMESPACE.getSharedData = function(instanceContainer, behaviorName) {
```
4ian, line 176: I wonder if this should be `getRUNTIME_BEHAVIOR_CLASSNAMESharedData`?

It works now because the CODE_NAMESPACE is different for each behavior, but in theory it could be the same one day.

4ian (approved):

GDJS/GDJS/Events/CodeGeneration/BehaviorCodeGenerator.cpp
```diff
@@ -148,6 +189,30 @@ CODE_NAMESPACE.RUNTIME_BEHAVIOR_CLASSNAME = class RUNTIME_BEHAVIOR_CLASSNAME ext
...
+CODE_NAMESPACE.RUNTIME_BEHAVIOR_CLASSNAME.getSharedData = function(instanceContainer, behaviorName) {
+  if (!instanceContainer.RUNTIME_BEHAVIOR_CLASSNAMESharedData) {
+    const initialData = instanceContainer.getInitialSharedDataForBehavior(
+      behaviorName
+    );
+    instanceContainer.RUNTIME_BEHAVIOR_CLASSNAMESharedData = new CODE_NAMESPACE.RUNTIME_BEHAVIOR_CLASSNAME.SharedData(
+      initialData
+    );
+  }
+  return instanceContainer.RUNTIME_BEHAVIOR_CLASSNAMESharedData;
```
4ian, line 213: I wonder if this would be useful:

```suggestion
  if (!instanceContainer._RUNTIME_BEHAVIOR_CLASSNAMESharedData) {
    const initialData = instanceContainer.getInitialSharedDataForBehavior(
      behaviorName
    );
    instanceContainer._RUNTIME_BEHAVIOR_CLASSNAMESharedData = new CODE_NAMESPACE.RUNTIME_BEHAVIOR_CLASSNAME.SharedData(
      initialData
    );
  }
  return instanceContainer._RUNTIME_BEHAVIOR_CLASSNAMESharedData;
```

to make it clear this is not to be used directly (instead, `CODE_NAMESPACE.RUNTIME_BEHAVIOR_CLASSNAME.getSharedData` should be used).

4ian (approved):

GDJS/GDJS/Events/CodeGeneration/BehaviorCodeGenerator.cpp
D8H: Oh, yes, I'm making it `_EXTENSION_NAME_RUNTIME_RUNTIME_BEHAVIOR_CLASSNAMESharedData` to avoid name collision.

PR #4468, opened 2022-10-30
Serialize JS code events and extension descriptions as arrays of strings to make collaboration/reviews easier

4ian:

### DIFF
--- a/Core/GDCore/Project/EventsFunctionsExtension.cpp
+++ b/Core/GDCore/Project/EventsFunctionsExtension.cpp
@@ -55,8 +55,6 @@ void EventsFunctionsExtension::SerializeTo(SerializerElement& element) const {
   element.SetAttribute("extensionNamespace", extensionNamespace);
   element.SetAttribute("shortDescription", shortDescription);
   element.AddChild("description").SetMultilineStringValue(description);
-  element.SetAttribute("description", description);
-
   element.SetAttribute("name", name);
   element.SetAttribute("fullName", fullName);
   element.SetAttribute("category", category);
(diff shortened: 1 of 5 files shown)


ClementPasteau (approved):
Looks pretty cool to me.
I assume there's no impact on existing projects?

4ian:
Good question! Existing project will be opened as usual, and when saved the JS code events `inlineCode`, and the `description` of extensions, will be saved as arrays. 
If you open the game in a previous version, this will be read as an empty string so games won't be backward compatible.

4ian:
We actually need to change the GDevelop-extensions deployment step so that any description and any JS code event is serialised to a string, so we keep compatibility with the existing versions of GDevelop. We can then remove this compatibility code once we have enough people using the last version.

Otherwise, extensions won't have descriptions and JS code events won't work

Core/GDCore/Project/EventsFunctionsExtension.cpp
```diff
@@ -54,7 +54,11 @@ void EventsFunctionsExtension::SerializeTo(SerializerElement& element) const {
   element.SetAttribute("version", version);
   element.SetAttribute("extensionNamespace", extensionNamespace);
   element.SetAttribute("shortDescription", shortDescription);
+
+  // TODO Use the multi line serialization in a release that follows the 150.
+  // element.AddChild("description").SetMultilineStringValue(description);
   element.SetAttribute("description", description);
```
D8H, line 60: I made this change.
I'm not use if this is what you had in mind.

PR #4469, opened 2022-10-31
Introduce no-unused-vars ts rule for GDJS & Extensions

ClementPasteau:


Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -1143,6 +1143,14 @@ namespace gdjs {
       return this._gravity;
     }
 
+    /**
+     * Get the slope maximun angle of the Platformer Object.
+     * @returns the slope maximun angle.
+     */
+    getSlopeMaxAngle(): float {
```
AlexandreSi, line 1150: Defining this maybe means that there are also missing expression, condition and actions on this parameter

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -1143,6 +1143,14 @@ namespace gdjs {
       return this._gravity;
     }
 
+    /**
+     * Get the slope maximun angle of the Platformer Object.
+     * @returns the slope maximun angle.
```
4ian, line 1148:
```suggestion
     * Get maximum angle of a slope for the Platformer Object to run on it as a floor.
     * @returns the slope maximum angle, in degrees.
```

4ian (approved):
Love this

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
ClementPasteau: Yeah I agree, but I'm not familiar with the platformer behavior so unsure this is wise exposing this value.
@D8H should we have an expression/condition for this value? (or should it not exist as it is not used?)

D8H: I don't think it would be useful as it's a setting that must be chosen once and for all.

AlexandreSi: One could want to change it though? If your platformer is a skier that is equipped with sealskin, it could go up slopes that are 30° but without them, only 5° (That's just en example)

D8H: This property is what tells apart a slope from a wall. Set to less than 50°, it will look broken.

ClementPasteau:
Alright, I'm merging it like this. 
If we feel there is the need of a new condition/expression/action for the slope angle, this can be done in a new PR

PR #4470, opened 2022-10-31
Fix Animation Preview height & being stuck when duplicating an image

### LINKED ISSUE #4461: Regressions in the animation previews
## Describe the bug

- The preview is not taking the whole height of the window.
- The default zoom seems also broken (maybe linked to the previous bug)
- The loading should probably not be shown at all here.
- The looping seems also broken 

image

## To Reproduce

- Open the preview animation window.
- For the looping issue:
  image

## Other details

* Version 5.1.149, desktop

ClementPasteau:
Fixes #4461

### DIFF
--- a/newIDE/app/src/ObjectEditor/Editors/SpriteEditor/AnimationPreview.js
+++ b/newIDE/app/src/ObjectEditor/Editors/SpriteEditor/AnimationPreview.js
@@ -16,9 +16,6 @@ import useForceUpdate from '../../../Utils/UseForceUpdate';
 import PlaceholderLoader from '../../../UI/PlaceholderLoader';
 
 const styles = {
-  imageContainer: {
-    position: 'relative',
-  },
   loaderContainer: {
     position: 'absolute',
     left: 'calc(50% - 30px)',
--- a/newIDE/app/src/ObjectEditor/Editors/SpriteEditor/AnimationPreview.js
+++ b/newIDE/app/src/ObjectEditor/Editors/SpriteEditor/AnimationPreview.js
@@ -2,7 +2,7 @@
 import { Trans } from '@lingui/macro';
 
 import React from 'react';
-import { Column } from '../../../UI/Grid';
+import { Column, Line } from '../../../UI/Grid';
 import { LineStackLayout } from '../../../UI/Layout';
 import ImagePreview from '../../../ResourcesList/ResourcePreview/ImagePreview';
 import Replay from '@material-ui/icons/Replay';
@@ -48,6 +48,7 @@ type Props = {|
   fixedHeight?: number,
   fixedWidth?: number,
   isAssetPrivate?: boolean,
+  hideAnimationLoader?: boolean,
 |};
 
 const AnimationPreview = ({
@@ -65,6 +66,7 @@ const AnimationPreview = ({
   fixedHeight,
   fixedWidth,
   isAssetPrivate,
+  hideAnimationLoader,
 }: Props) => {
   const forceUdpate = useForceUpdate();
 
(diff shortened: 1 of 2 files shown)


newIDE/app/src/ObjectEditor/Editors/SpriteEditor/AnimationPreview.js
```diff
@@ -161,17 +164,28 @@ const AnimationPreview = ({
...
         if (currentFrameIndex !== newFrameIndex) {
-          imagesLoadedArray.current[currentFrameIndexRef.current] = false;
-          // When the array of loaders changes, wait a bit to display the loader to avoid flickering.
-          loaderTimeout.current = setTimeout(() => {
-            console.warn(
-              'The image took too long to load, displaying a loader.'
-            );
-            setIsStillLoadingResources(true);
-          }, 500);
+          if (newResourceName === currentResourceNameRef.current) {
```
AlexandreSi, line 171: Shouldn't it be the path rather than the name of the file?

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/AnimationPreview.js
```diff
@@ -221,7 +235,7 @@ const AnimationPreview = ({
 
   return (
     <Column expand noOverflowParent noMargin>
-      <div style={styles.imageContainer}>
```
AlexandreSi, line 224: You can remove the style I think

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/DirectionTools.js
```diff
@@ -185,6 +185,7 @@ export default class DirectionTools extends Component<Props, State> {
                 this.setState({ timeBetweenFrames: text })
               }
               isLooping={direction.isLooping()}
+              hideAnimationLoader // No need to show a loader in the Direction Tools.
```
AlexandreSi, line 188: It is true when on Desktop because the resources are local, but should we show the loader for cloud projects?

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/AnimationPreview.js
ClementPasteau: Tried that, but:
- there is a one to one relation between name & path (and a function is called to get the path from the name) so the result should be the same
- but when a file is cached locally, it adds a `?cache={date}` in the url, so the urls actually differ as the animation continues, but the browser is smart enough to used the cached image and not trigger the onLoad.

tldr; I tried but it felt more complex to make it work

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/DirectionTools.js
ClementPasteau: We could, but as this tool is not the "external facing" animations, like in the Asset Store, and more of a "internal tool" to work on your animation, I'm not sure it adds value.

newIDE/app/src/ObjectEditor/Editors/SpriteEditor/AnimationPreview.js
AlexandreSi: Are you sure there is a one to one relation between names and paths for resources? Because today you can have multiple objects using the same resource.

ClementPasteau: I think this is different here, this is in the scope of 1 object & 1 animation only.
The animation knows where to find the resource from the name of this resource.
It uses the function `resourcesLoader.getResourceFullUrl(project, name, {})` in the `DirectionTools`

AlexandreSi (approved):
LGTM

PR #4478, opened 2022-11-02
Do not prevent inline parameter editor to be opened

### LINKED ISSUE #4473: UI locks up the field editing after dragging an actionline when the field is open.
When editing the actions entry field in the editor and then drag the action while the field is opened fo editing it will lock out the field editing until you restart GDevelop.

This has been a problem since forever, seemingly random and getting a long while to accidentally happen.

A short video on how to reproduce.
https://user-images.githubusercontent.com/90930609/199047070-94f304d8-2293-4bdf-8d28-423649f3b51c.mp4

Same issue as this one in here.
https://github.com/4ian/GDevelop/issues/4044

AlexandreSi:
Resolves #4473

### DIFF
--- a/newIDE/app/src/EventsSheet/index.js
+++ b/newIDE/app/src/EventsSheet/index.js
@@ -891,9 +891,6 @@ export class EventsSheetComponentWithoutHandle extends React.Component<
     locatingEvent: gdBaseEvent,
     parameterContext: ParameterContext
   ) => {
-    // Prevent state from changing when clicking on a parameter with inline parameter
-    // editor (it will close the editor).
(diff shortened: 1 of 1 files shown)


ClementPasteau (approved):
👍

D8H (approved):
It will allow to open another inline editor with only one click instead of 2.

AlexandreSi:
> It will allow to open another inline editor with only one click instead of 2.

Well unfortunately no. Around the inline parameter editor, there's a click away listener that listens for any click outside the editor and we close the editor. So a click on any parameter will first close the editor, and the second click re-opens it.

PR #4479, opened 2022-11-02
Allow private event-based behaviors.

D8H:
This will reduce EBO maintenance cost. For instance, every button object kinds will use the same FSM as a behavior internally and when we improve the FSM, we'll only have to copy-paste the new behavior version.

I wanted to add some tests but it would need to declare an event-based behavior and generate its metadata. It seems a lot to cover a few lines. I'm not sure if it's worth the effort and I guess the person who implemented private functions though the same thing.

Manual tests can be done with this project:
- https://www.dropbox.com/s/amqaqeizwgsap0c/PrivateBehavior.zip?dl=0

### DIFF
--- a/Core/GDCore/Extensions/Metadata/BehaviorMetadata.h
+++ b/Core/GDCore/Extensions/Metadata/BehaviorMetadata.h
@@ -236,7 +236,6 @@ class GD_CORE_API BehaviorMetadata {
   }
 
   const gd::String& GetName() const;
-#if defined(GD_IDE_ONLY)
   const gd::String& GetFullName() const { return fullname; }
   const gd::String& GetDefaultName() const { return defaultName; }
   const gd::String& GetDescription() const { return description; }
@@ -257,7 +256,21 @@ class GD_CORE_API BehaviorMetadata {
    * \note An empty string means the base object, so any object.
    */
   const gd::String& GetObjectType() const { return objectType; }
-#endif
+
+  /**
+   * Check if the behavior is private - it can't be used outside of its
+   * extension.
+   */
+  bool IsPrivate() const { return isPrivate; }
+
+  /**
+   * Set that the behavior is private - it can't be used outside of its
+   * extension.
+   */
+  BehaviorMetadata &SetPrivate() {
+    isPrivate = true;
+    return *this;
+  }
 
   /**
    * \brief Return the associated gd::Behavior, handling behavior contents.
@@ -315,6 +328,7 @@ class GD_CORE_API BehaviorMetadata {
   gd::String group;
   gd::String iconFilename;
   gd::String objectType;
+  bool isPrivate = false;
 
   // TODO: Nitpicking: convert these to std::unique_ptr to clarify ownership.
   std::shared_ptr<gd::Behavior> instance;
--- a/newIDE/app/src/BehaviorsEditor/EnumerateBehaviorsMetadata.js
+++ b/newIDE/app/src/BehaviorsEditor/EnumerateBehaviorsMetadata.js
@@ -31,23 +31,22 @@ export const enumerateBehaviorsMetadata = (
           behaviorType,
           behaviorMetadata: extension.getBehaviorMetadata(behaviorType),
         }))
-        .map(({ behaviorType, behaviorMetadata }) =>
-          !behaviorMetadata.isPrivate() ||
-          (eventsFunctionsExtension &&
-            extension.getName() === eventsFunctionsExtension.getName())
-            ? {
-                extension,
-                behaviorMetadata,
-                type: behaviorType,
-                defaultName: behaviorMetadata.getDefaultName(),
-                fullName: behaviorMetadata.getFullName(),
-                description: behaviorMetadata.getDescription(),
-                iconFilename: behaviorMetadata.getIconFilename(),
-                objectType: behaviorMetadata.getObjectType(),
-              }
-            : null
+        .filter(
+          ({ behaviorMetadata }) =>
+            !behaviorMetadata.isPrivate() ||
+            (eventsFunctionsExtension &&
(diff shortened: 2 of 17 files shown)


Core/GDCore/Extensions/Metadata/BehaviorMetadata.h
```diff
@@ -257,7 +256,21 @@ class GD_CORE_API BehaviorMetadata {
...
+   * Check if the behavior is private - it can't be used outside of its
+   * extension.
+   */
+  bool IsPrivate() const { return isPrivate; }
+
+  /**
+   * Set that the behavior is private - it can't be used outside of its
+   * extension.
+   */
+  BehaviorMetadata &SetPrivate() {
```
ClementPasteau, line 270: Should this function take a boolean as parameter? So that one can revert the isPrivate status

newIDE/app/src/BehaviorsEditor/EnumerateBehaviorsMetadata.js
```diff
@@ -30,16 +31,23 @@ export const enumerateBehaviorsMetadata = (
...
-          defaultName: behaviorMetadata.getDefaultName(),
-          fullName: behaviorMetadata.getFullName(),
-          description: behaviorMetadata.getDescription(),
-          iconFilename: behaviorMetadata.getIconFilename(),
-          objectType: behaviorMetadata.getObjectType(),
-        }));
+        .map(({ behaviorType, behaviorMetadata }) =>
+          !behaviorMetadata.isPrivate() ||
+          (eventsFunctionsExtension &&
+            extension.getName() === eventsFunctionsExtension.getName())
```
ClementPasteau, line 37: maybe you could extract this logic outside of the map, as a properly named variable, it will be easier to read.
Something like : 
`const isCurrentlyEditedEventsFunctionExtension = eventsFunctionsExtension && extension.getName() === eventsFunctionsExtension.getName()`

newIDE/app/src/EventsBasedBehaviorsList/index.js
```diff
@@ -22,9 +22,30 @@ import {
...
+import VisibilityOffIcon from '@material-ui/icons/VisibilityOff';
 
 const EVENTS_BASED_BEHAVIOR_CLIPBOARD_KIND = 'Events Based Behavior';
 
+const renderEventsBehaviorLabel = (
+  eventsBasedBehavior: gdEventsBasedBehavior
+) =>
+  eventsBasedBehavior.isPrivate() ? (
+    <>
+      <Tooltip title="This function won't be visible in the events editor">
```
ClementPasteau, line 35: add a `<Trans>` to this title

newIDE/app/src/EventsBasedBehaviorsList/index.js
```diff
@@ -22,9 +22,30 @@ import {
...
+        <VisibilityOffIcon
+          fontSize="small"
+          style={{ marginRight: 5, verticalAlign: 'bottom' }}
```
ClementPasteau, line 38: let's extract the style at the top to be consistent with other files

newIDE/app/src/EventsBasedBehaviorsList/index.js
```diff
@@ -22,9 +22,30 @@ import {
...
+        />
+      </Tooltip>
+      <span title={eventsBasedBehavior.getName()}>
+        {eventsBasedBehavior.getName()}
+      </span>
+    </>
+  ) : (
+    eventsBasedBehavior.getName()
```
ClementPasteau, line 46: should this be a variable instead of calling the function 3 times? (not sure it improves performance...)

newIDE/app/src/InstructionOrExpression/EnumeratedInstructionOrExpressionMetadata.js
```diff
@@ -57,38 +57,35 @@ export const filterEnumeratedInstructionOrExpressionMetadataByScope = <
...
+        eventsBasedBehavior &&
+        eventsFunctionsExtension &&
+        getBehaviorFullType(
+          eventsFunctionsExtension.getName(),
+          eventsBasedBehavior.getName()
+        ) === behaviorMetadata.getName()) ||
+      // Show private non-behavior functions when editing the extension
+      (!behaviorMetadata &&
+        eventsFunctionsExtension &&
+        eventsFunctionsExtension.getName() === extension.getName())
```
ClementPasteau, line 88: this logic seems to be done multiple times. Maybe create a variable for it as well so you can reuse it?

ClementPasteau: It would improve readability

Core/GDCore/Extensions/Metadata/BehaviorMetadata.h
4ian: In C++, this can even be a default parameter (default parameters don't work when exposed in JS, but nice to have in C++).

newIDE/app/src/BehaviorsEditor/EnumerateBehaviorsMetadata.js
4ian: You can use also a filter before the map, so you avoid to mix concerns in a map (slightly easier to read: "I see there is a filter, and then we map to transform the data").

4ian (approved):

Core/GDCore/Extensions/Metadata/BehaviorMetadata.h
D8H: Actually, these class are used declaratively, it won't be mutated much. It's public by default and it can be declared private but it will never need to change after this. The next generation will use new instances.

PR #4481, opened 2022-11-03
Add autocompletion for choices and color property private instructions

D8H:
It uses addExpressionAndConditionAndAction to declare properties instructions. It will be easier to add new property types.

Manually tested for every property type for behavior and object with this project
- https://www.dropbox.com/s/rs5aa0w5d4ts4s2/ExpressionAndConditionTest.zip?dl=1

image


ClementPasteau (commented):
lgtm, I'll let someone else review just in case

4ian (approved):
lgtm!

PR #4483, opened 2022-11-03
Fix number expression fields

D8H:
This is a pretty big regression that I did in this PR:
- https://github.com/4ian/GDevelop/pull/3909

It declares ParameterMetadata with "number" instead of "expression" for ExpressionAndCondition. It's not a mistake but it wasn't handled by the UI.
I think we should replace all the "expression" by "number" at some point. This way it will be consistent with the expression type.


4ian (approved):
We should indeed convert everything to "number" at some point. For now this will work.

PR #4489, opened 2022-11-04
Fix: parameter default value changes were not saved.

D8H:
This is a regression introduced with:
- https://github.com/4ian/GDevelop/pull/3909

It also avoid to serialize some optional attributes when they are empty to make extensions with a lot of properties a bit lighter.

### DIFF
--- a/Core/GDCore/Project/EventsFunction.cpp
+++ b/Core/GDCore/Project/EventsFunction.cpp
@@ -59,11 +59,19 @@ const std::vector<gd::ParameterMetadata>& EventsFunction::GetParametersForEvents
 void EventsFunction::SerializeTo(SerializerElement& element) const {
   element.SetAttribute("name", name);
   element.SetAttribute("fullName", fullName);
-  element.SetAttribute("description", description);
+  if (!description.empty()) {
+    element.SetAttribute("description", description);
+  }
   element.SetAttribute("sentence", sentence);
-  element.SetAttribute("group", group);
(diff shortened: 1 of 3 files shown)


Core/GDCore/Project/EventsFunction.cpp
```diff
@@ -59,11 +59,19 @@ const std::vector<gd::ParameterMetadata>& EventsFunction::GetParametersForEvents
...
-  element.SetAttribute("group", group);
-  element.SetAttribute("getterName", getterName);
-  element.SetBoolAttribute("private", isPrivate);
+  if (!group.empty()) {
+    element.SetAttribute("group", group);
+  }
+  if (!getterName.empty()) {
+    element.SetAttribute("getterName", getterName);
+  }
+  if (isPrivate) {
```
ClementPasteau, line 72: is there a risk the setBoolAttribute is not triggered if we're serializing an object with isPrivate = false?

AlexandreSi (approved):
LGTM!

Core/GDCore/Project/EventsFunction.cpp
D8H: Yes, this is the point.
`GetBoolAttribute` return `false` by default and `isPrivate` is `false` most of the time.

ClementPasteau: Nice!

PR #4492, opened 2022-11-04
Add possibility to use markdown for asset descriptions

AlexandreSi:
In order to credit creators in asset and asset pack descriptions:

image

image

### DIFF
--- a/newIDE/app/src/AssetStore/PrivateAssets/PrivateAssetPackInformationDialog.js
+++ b/newIDE/app/src/AssetStore/PrivateAssets/PrivateAssetPackInformationDialog.js
@@ -13,11 +13,7 @@ import PriceTag, { formatPrice } from '../../UI/PriceTag';
 import FlatButton from '../../UI/FlatButton';
 import AlertMessage from '../../UI/AlertMessage';
 import PlaceholderLoader from '../../UI/PlaceholderLoader';
-import {
-  ColumnStackLayout,
-  ResponsiveLineStackLayout,
-  LineStackLayout,
-} from '../../UI/Layout';
+import { ResponsiveLineStackLayout, LineStackLayout } from '../../UI/Layout';
 import { Column, Line } from '../../UI/Grid';
 import {
   getUserPublicProfile,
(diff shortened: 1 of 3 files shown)


newIDE/app/src/AssetStore/PrivateAssets/PrivateAssetPackInformationDialog.js
```diff
@@ -218,7 +215,9 @@ const PrivateAssetPackDialog = ({
                           <PriceTag value={prices[0].value} />
                           {getBuyButton(i18n)}
                         </Line>
-                        <Text noMargin>{assetPack.longDescription}</Text>
+                        <Text size="body2" displayInlineAsSpan>
+                          <MarkdownText source={assetPack.longDescription} />
```
ClementPasteau, line 219: isn't using markdowntext inside a text component creating warnings in the DOM? (because of p inside h4 or something like that)

I believe this is wider issue that we should fix at some point.
I've seen multiple warnings in the app when markdown text is used.

newIDE/app/src/UI/MarkdownText.js
```diff
@@ -61,6 +61,7 @@ export const MarkdownText = (props: Props) => {
             'standalone-text-container': props.isStandaloneText,
           })}
           components={markdownCustomComponents}
+          linkTarget="_blank"
```
ClementPasteau, line 64: interesting, the `makeMarkdownCustomComponents` should already apply a target blank if this is an `a`.
Is this not working?

newIDE/app/src/AssetStore/PrivateAssets/PrivateAssetPackInformationDialog.js
AlexandreSi: Yes indeed. That's why there's the displayInlineAsSpan, it does not raise errors

newIDE/app/src/UI/MarkdownText.js
AlexandreSi: Discussed together: I didn't have installed the version you specified

ClementPasteau (approved):
🚀

PR #4496, opened 2022-11-04
Fix Posthog alias

ClementPasteau:
Do not show in changelog


4ian (approved):
I *think* this is correct, let's check it works though for real both for an existing user logging and a new user creating their account.

PR #4503, opened 2022-11-04
Add a setting to try the event-based object editor

D8H:


newIDE/app/src/MainFrame/Preferences/PreferencesDialog.js
```diff
@@ -329,6 +330,16 @@ const PreferencesDialog = ({ i18n, onClose }: Props) => {
               </Trans>
             }
           />
+          <Toggle
+            onToggle={(e, check) => setShowEventBasedObjectsEditor(check)}
+            toggled={values.showEventBasedObjectsEditor}
+            labelPosition="right"
+            label={
+              <Trans>
+                Show event-based objects in the extension editor (experimental)
```
4ian, line 339: Nit: I wonder if we should go for "custom objects" rather than "event-based objects" in the UI.

4ian (approved):
👍👍

Issue #4055, opened 2022-06-21
[$30.00 bounty] Allow shape painter objects to use anti-aliasing


tristanbob:
## Description

Drawing circles (and other curved shapes) with shape painters results in jagged lines.  This is true even with the default setting of anti-aliasing enabled for GDevelop projects.

## Solution suggested

Pixi does support anti-aliasing and the circles look much better with it turned on.

Pixi with anti-aliasing = FALSE
https://codepen.io/tristanbob/pen/VwQOrLP

Pixi with anti-aliasing = TRUE
https://codepen.io/tristanbob/pen/GRQaOJM

- [ ] GDevelop should offer the option to enable antialiasing on shape painter objects
 (and enabled by default, unless there are significant downsides)

jposeyjr:
// Disable anti-aliasing(default) to avoid rendering issue (1px width line of extra pixels) when rendering pixel perfect tiled sprites.
@4ian Would you want a warning to pop up for this or is simply changing the props enough?

Midhil457:
I would like to dabble into this. But first I have some doubt
After browsing thorugh the code a bit, I found that the setting in the project properties changes the rendering setting for the textures and the renderer's antialiesing option is always set to false (My understanding might be completely wrong). Is there a reason for this?  

After testing I did find that that the shape painter is being antialiesed, but not very well.

My idea is to completely use the renderer for antialiesing. I am worried if it might break some settings.

:)

jposeyjr:
Yeah but they want it enabled and currently no way to do that. Adding it to the settings is just tedious due to how the client is formed. You have to find a good place to pass it down.

Midhil457:
> Yeah but they want it enabled and currently no way to do that. Adding it to the settings is just tedious due to how the client is formed. You have to find a good place to pass it down.

I didn't understand. Could you clarify?

Midhil457:
Hello? Is it ok if I go forward with my plan?

Oxey405:
Hey ! I worked a bit on that issue and I implemented something that looks I'd say prommessing !
Sending a PR today :)

D8H:
Antialiasing can be heavy. I guess, the `cacheAsBitmap` mode will be interesting to allow to render high quality still image and keep good efficiency.
https://pixijs.download/dev/docs/PIXI.Graphics.html#cacheAsBitmap

Oxey405:
@D8H I'vhe worked on it can you maybe look at it later ?

[...]

PR #4505, opened 2022-11-05
added antialiasing to shape painters

Oxey405:
Hello there !
So I worked on a potential fix to the issue #4055 which asked if anyone could add antialiasing options to shape painters.

For that I used PIXIJS filters to handle all of the actual antialiasing and added 2 new events
- Set Antialiasing on
- Set antialiasing quality (2, 4 or 8)
At the time of me writing this, I'm aware that some things are still missing. I will work on them throughout the day or the week until it's perfect but I think this PR can already be reviewed :)


Oxey405:
Broken PR ignore please

PR #4515, opened 2022-11-07
[PathFinding] Add a property to smooth the path.

D8H:
A build of one of the example:
https://liluo.io/games/47f7973b-f3bc-4779-80be-d4baed26fef6?dev=true

The project:
https://www.dropbox.com/s/tv5c0egzltxud69/PathfindingSmoothingTest.zip?dl=1

image


Extensions/PathfindingBehavior/pathfindingruntimebehavior.ts
```diff
@@ -487,6 +513,156 @@ namespace gdjs {
...
+     *
+     * @param px The x-value of point (px, py).
+     * @param py The y-value of point (px, py)
+     * @param ax The x-value of the line segment's vertex A.
+     * @param ay The y-value of the line segment's vertex A.
+     * @param bx The x-value of the line segment's vertex B.
+     * @param by The y-value of the line segment's vertex B.
+     * @return The distance squared from the point (px, py) to line segment AB.
+     */
+    private static getPointSegmentDistanceSq(
```
D8H, line 617: This should probably be in a tool class, maybe even `simplifyPath`.

Extensions/PathfindingBehavior/PathfindingBehavior.cpp
```diff
@@ -57,6 +59,11 @@ std::map<gd::String, gd::PropertyDescriptor> PathfindingBehavior::GetProperties(
       gd::String::From(behaviorContent.GetDoubleAttribute("gridOffsetY", 0)));
   properties[_("Extra border size")].SetGroup(_("Collision")).SetValue(
       gd::String::From(behaviorContent.GetDoubleAttribute("extraBorder")));
+  properties[_("Smoothing max cell gap")]
+      .SetValue(gd::String::From(
+          behaviorContent.GetDoubleAttribute("smoothingMaxCellGap")))
+      .SetGroup(_("Path smoothing"))
+      .SetDescription(_("It's recommended to leave a max gap of 1 cell."));
```
ClementPasteau, line 66: Is it worth adding a bit more info? Like what should the use expect by increasing/decreasing this value

Extensions/PathfindingBehavior/tests/legacypathfindingruntimebehavior.spec.js
```diff
@@ -1,6 +1,8 @@
 // @ts-check
 describe('gdjs.PathfindingRuntimeBehavior', function () {
   // limit tests cases on the legacy collision methods.
+  // Not that the legacy collision mode is still the only mode that exists
```
ClementPasteau, line 4: typo "Note"

Extensions/PathfindingBehavior/tests/legacypathfindingruntimebehavior.spec.js
```diff
@@ -1,6 +1,8 @@
...
+  // because the new one were never merged.
```
ClementPasteau, line 5: not related to this PR, but why has it never been merged?

ClementPasteau (approved):
few comments but lgtm.
Really nice improvemen!

Extensions/PathfindingBehavior/tests/legacypathfindingruntimebehavior.spec.js
D8H: The obstacle detection is a big issue of the behavior because it only handles default hiboxes without rotation.
The new method was doing true collision checks instead of boxes check. There was a risk that it would be too slow and we would need yet another collision method. So we decided to wait for a better solution.
The better solution could be a NavMesh, but it has the issue to take a long time to build (maybe a worker could solve this) and I haven't implemented incremental updates so it's not really usable.

Extensions/PathfindingBehavior/PathTools.ts
```diff
@@ -0,0 +1,163 @@
...
+     * @param maxGap The maximum distance the edge of the contour may deviate
+     * from the source geometry.
+     * @param simplifiedVertices The simplified path.
+     * @param workingVertices It avoids allocations.
+     */
+    export const simplifyPath = (
+      sourceVertices: FloatPoint[],
+      maxGap: float,
+      simplifiedVertices: FloatPoint[] = [],
+      workingVertices: FloatPoint[] = []
```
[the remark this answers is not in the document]
D8H: This does allocations, it must be avoided in the runtime code.

4ian (approved):
Overall LGTM!

Extensions/PathfindingBehavior/PathfindingBehavior.cpp
D8H: I added a note about what `0` does. Anything other than `0` or `1` is a bit useless. Values less than 1 will only smooth some parts of the path and value over 1 may remove important point. I didn't use a boolean because it might be useful in very rare cases.

Extensions/PathfindingBehavior/tests/legacypathfindingruntimebehavior.spec.js
ClementPasteau: ok thanks 👍

Extensions/PathfindingBehavior/PathTools.ts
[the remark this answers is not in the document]
ClementPasteau: ok 👍

Extensions/PathfindingBehavior/PathfindingBehavior.cpp
ClementPasteau: OK nice!

PR #4516, opened 2022-11-07
[Platformer] The character no longer has a 1-frame delay when following platforms

D8H:
Fixes: https://forum.gdevelop.io/t/stop-player-sliding-on-moving-platform/42693

### DIFF
--- a/Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
+++ b/Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
@@ -262,7 +262,6 @@ namespace gdjs {
       }
 
       //0.2) Track changes in object size
-      this._state.beforeUpdatingObstacles(timeDelta);
       this._onFloor._oldHeight = object.getHeight();
 
       //0.3) Update list of platforms around/related to the object
@@ -329,7 +328,10 @@ namespace gdjs {
       this._lastDeltaY = object.getY() - oldY;
     }
 
-    doStepPostEvents(instanceContainer: gdjs.RuntimeInstanceContainer) {}
+    doStepPostEvents(instanceContainer: gdjs.RuntimeInstanceContainer) {
+      const timeDelta = this.owner.getElapsedTime() / 1000;
+      this._state.doStepPostEvents(timeDelta);
+    }
 
     private _updateSpeed(timeDelta: float): float {
       const previousSpeed = this._currentSpeed;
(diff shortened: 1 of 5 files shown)


Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -1610,7 +1612,7 @@ namespace gdjs {
      * Called before the obstacle search.
```
AlexandreSi, line 1612: I feel like this comment should be changed, I'm having troubles figuring out what it means.

Extensions/PlatformBehavior/tests/FlatFloorPlatformer.spec.js
```diff
@@ -694,9 +694,6 @@ describe('gdjs.PlatformerObjectRuntimeBehavior', function () {
       expect(object.getBehavior('auto1').isFalling()).to.be(false);
       expect(object.getBehavior('auto1').isOnFloor()).to.be(true);
       expect(object.getY()).to.be.within(140.6297999, 140.6298001);
-      // TODO Remove the 1-frame delay
```
AlexandreSi, line 697: 💜

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
```diff
@@ -1665,9 +1667,10 @@ namespace gdjs {
       this._floorLastY = this._floorPlatform!.owner.getY();
     }
 
-    beforeUpdatingObstacles(timeDelta: float) {
-      const object = this._behavior.owner;
-      //Stick the object to the floor if its height has changed.
+    doStepPostEvents(timeDelta: float) {
```
4ian, line 1670: Doing this post event is a big problem because it means your events have no chance to position anything on top of the platformer object - precisely because it runs after the events.

Any chance the fix could be done somehow else?

4ian (changes requested):
Marking as requesting changes because the one frame delay for the events is potentially worse - as it means anything you put on top of the object (let's say, a sword or a helmet on a knight in a platformer game) will be always some/several pixels behind the platformer.

In general, behaviors must always do all their changes before events.

Extensions/PlatformBehavior/platformerobjectruntimebehavior.ts
D8H: I think it had to be done after the events
- events move the platform
- the character is moved accordingly
- the platform and the character are rendered without any gap

But maybe we should keep both because the platform could be moved by another behavior and the character should not fell or be stuck in this case.

D8H:
> Marking as requesting changes because the one frame delay for the events is potentially worse - as it means anything you put on top of the object (let's say, a sword or a helmet on a knight in a platformer game) will be always some/several pixels behind the platformer.
> 
> In general, behaviors must always do all their changes before events.

Should we do an action "Follow current platform"?

AlexandreSi:
How should this action be used? Without condition?
Or maybe with the condition `IsOnFloor`?

PR #4517, opened 2022-11-07
Choose Extension is now a built-in

UlisesFreitas:
Choose extension which is right now a community extension I managed to make that  extension as a built in please check if this is worthed to be integrated.


AlexandreSi:
Hi @UlisesFreitas,
Thanks for submitting this PR!

Could you explain why this extension should be part of the built-in extensions?
It looks like it's useful for quite a specific use-case.

4ian:
Hey @UlisesFreitas - thanks for taking the time to open the PR :) 
For the "core", built-in extensions of GDevelop, we're actually trying to keep them as small as possible in terms of number and to have a maximum of them done as extensions on the GDevelop-extensions repository.
The reason is because this means less maintenance of code, and because what can be done as a non built in extension should stay non built in: it's faster to update, does not rely on additional JS code like the JsExtension.js file that must be maintained when we do changes to extensions, can be updated without doing a new release of GDevelop...

This being said, you could say "sure but this extension is so important that it should be visible by default!". And I agree it's an important one, but I think it's more a UX challenge: how do we make these important extensions easy to find in the UI ? Should we show them somewhere, or show them in a way we can see it's "available but you need to activate it?". For example, should we list it in the actions/conditions/expressions and when it's used, the extension is automatically fetched and added to the project?

I think this could be interesting to do for reviewed extension - to have them "look like they are available by default". Happy to get your thoughts!

UlisesFreitas:
Hi thank you both for taking time to review or at least check this PR, that beign said there is no need really for this to be included for the "core", built-in extensions, I tottaly get the idea of keeping the core as minimal and make sense. So don't worry and just leep developing this great engine.
Regards.

4ian:
No problem and thanks again for what your contributions and what you do with GDevelop 👍

PR #4518, opened 2022-11-07
[Platformer] Make tests with events side effect more realistic.

D8H:
I prefer the test changes to be merged first. It will allow to easily check they worked before the fix and see what we changed in the tests with the fix if we need to come back to this one day.

### DIFF
--- a/Extensions/PlatformBehavior/tests/FlatFloorPlatformer.spec.js
+++ b/Extensions/PlatformBehavior/tests/FlatFloorPlatformer.spec.js
@@ -404,7 +404,9 @@ describe('gdjs.PlatformerObjectRuntimeBehavior', function () {
         expect(object.getY()).to.be(-30); // -30 = -10 (platform y) + -20 (object height)
 
         // Make the platform under the character feet smaller.
-        object.setCustomWidthAndHeight(object.getWidth(), 9);
+        runtimeScene.renderAndStepWithEventsFunction(1000 / 60, () => {
+          object.setCustomWidthAndHeight(object.getWidth(), 9);
+        });
         runtimeScene.renderAndStep(1000 / 60);
         expect(object.getBehavior('auto1').isFalling()).to.be(false);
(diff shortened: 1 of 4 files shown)


Extensions/PlatformBehavior/tests/FlatFloorPlatformer.spec.js
```diff
@@ -404,7 +404,9 @@ describe('gdjs.PlatformerObjectRuntimeBehavior', function () {
         expect(object.getY()).to.be(-30); // -30 = -10 (platform y) + -20 (object height)
 
         // Make the platform under the character feet smaller.
-        object.setCustomWidthAndHeight(object.getWidth(), 9);
+        runtimeScene.renderAndStepWithEventsFunction(1000 / 60, () => {
+          object.setCustomWidthAndHeight(object.getWidth(), 9);
+        });
         runtimeScene.renderAndStep(1000 / 60);
         expect(object.getBehavior('auto1').isFalling()).to.be(false);
```
D8H, line 411: I added a step because the change of state will only happens after the 2nd step as the platform is now moving after `doStepPreEvents` as it should.
I didn't do it on the other tests because they are looping so event if the 1st iteration can't really change of state, it's not a big deal because other step follows.

GDJS/tests/tests/Extensions/TestRuntimeScene.js
```diff
@@ -0,0 +1,22 @@
...
+  constructor(runtimeGame) {
+    super(runtimeGame);
+  }
+
+  /**
+   * @param {float} elapsedTime 
+   * @param {() => void} eventsFunction 
+   */
+  renderAndStepWithEventsFunction(elapsedTime, eventsFunction) {
+    this._eventsFunction = eventsFunction.bind(this, this);
```
4ian, line 18: Prefer a anonymous function, slightly more readable that the old bind (and equally performant, as both create a new function)

4ian (approved):
This testing approach looks good to me

Issue #4523, opened 2022-11-08
Blurry pixels for 9 patch object


HelperWesley:
Blurry 9patch objects

## To Reproduce
1. Create a 9 patch object.

[My project56.zip](https://github.com/4ian/GDevelop/files/9962480/My.project56.zip)

@ClementPasteau

HelperWesley:
Whatever the issue is, it seems limited to 9 patch objects. The object to the left is a tiled sprite, the floor the character is standing on is a 9 patch.

![image](https://user-images.githubusercontent.com/72082749/200609275-efb456ac-3e31-46c3-8bdc-4bfa0bae402e.png)

Oxey405:
I worked on the renderer this week, maybe should I take a look ? :)

D8H:
The isssue is that `cacheAsBitmap` uses `PIXI.SCALE_MODES.LINEAR` for the sprite in cache.(`this._spritesContainer._cacheData.sprite._texture.baseTexture.scaleMode`).
We will probably have to deactivate the optimization when the image smoothing is deactivated on the image used by the 9-patch.

PR #4526, opened 2022-11-09
Panel sprite objects now respect the resource smoothing option

D8H:
Fixes https://github.com/4ian/GDevelop/issues/4523
- https://github.com/4ian/GDevelop/issues/4523

### Solution
It deactivates the `cacheAsBitmap` optimization when the image smoothing is deactivated on the image used by the 9-patch.

### Issue
The isssue is that cacheAsBitmap uses the scale mode PIXI.SCALE_MODES.LINEAR for the sprite in cache (this._spritesContainer._cacheData.sprite._texture.baseTexture.scaleMode).
https://github.com/pixijs/pixijs/blob/main/packages/mixin-cache-as-bitmap/src/index.ts

Ideally, we would need a `cacheAsBitmapScaleMode` property in Pixi.

### Efficiency
There will be 9 times more draw calls, but it's not really an issue as it takes very little CPU time. What really impacts efficiency is actually the tilling. It's very heavy on the GPU. It seems that Pixi uses the WebGL `repeat` texture wrap. So, I guess the only way to improve efficiency was `cacheAsBitmap` (can `cacheAsBitmap` creates giant texture which could be an issue too?).

I don't think it's a big deal as TiledSpriteObject have the exact same efficiency issue and no one complained about it. The editor also render without `cacheAsBitmap` and it's fine as long as there less than 100 big areas to tile.

### DIFF
--- a/Extensions/PanelSpriteObject/panelspriteruntimeobject-pixi-renderer.ts
+++ b/Extensions/PanelSpriteObject/panelspriteruntimeobject-pixi-renderer.ts
@@ -33,27 +33,24 @@ namespace gdjs {
       const StretchedSprite = !tiled ? PIXI.Sprite : PIXI.TilingSprite;
       this._spritesContainer = new PIXI.Container();
       this._wrapperContainer = new PIXI.Container();
-      // @ts-ignore
-      this._centerSprite = new StretchedSprite(new PIXI.Texture(texture));
+      this._centerSprite = new StretchedSprite(
+        new PIXI.Texture(texture.baseTexture)
+      );
       this._borderSprites = [
-        // @ts-ignore
-        new StretchedSprite(new PIXI.Texture(texture)),
-        //Right
+        // Right
+        new StretchedSprite(new PIXI.Texture(texture.baseTexture)),
+        // Top-Right
         new PIXI.Sprite(texture),
-        //Top-Right
-        // @ts-ignore
-        new StretchedSprite(new PIXI.Texture(texture)),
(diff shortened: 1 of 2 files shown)


Extensions/PanelSpriteObject/panelspriteruntimeobject-pixi-renderer.ts
```diff
@@ -77,11 +74,19 @@ namespace gdjs {
...
     ensureUpToDate() {
       if (this._spritesContainer.visible && this._wasRendered) {
-        // Cache the rendered sprites as a bitmap to speed up rendering when
-        // lots of panel sprites are on the scene.
-        // Sadly, because of this, we need a wrapper container to workaround
-        // a PixiJS issue with alpha (see updateOpacity).
-        this._spritesContainer.cacheAsBitmap = true;
+        // PIXI uses PIXI.SCALE_MODES.LINEAR for the cached image:
+        // this._spritesContainer._cacheData.sprite._texture.baseTexture.scaleMode
+        // There seems to be no way to configure this so the optimization is disabled.
```
ClementPasteau, line 79: Should we create an issue on their repository and link the issue in the comment here? 
This would help with documentation when we come back to it one day

ClementPasteau (approved):
LGTM ! 
Let's do a few tests on existing examples/templates that use PanelSprites just to ensure we haven't missed anything.
But looks good.
Nice debugging!

4ian (approved):

PR #4528, opened 2022-11-09
Save the scroll position of the asset store

D8H:
http://gdevelop-storybook.s3-website-us-east-1.amazonaws.com/asset-store-scroll/latest/index.html?path=/story/assetstore-assetstore--default

### DIFF
--- a/newIDE/app/src/AssetStore/AssetStoreNavigator.js
+++ b/newIDE/app/src/AssetStore/AssetStoreNavigator.js
@@ -13,6 +13,7 @@ export type AssetStorePageState = {|
   openedAssetShortHeader: ?AssetShortHeader,
   filtersState: FiltersState,
   ignoreTextualSearch: boolean,
+  scrollPosition?: ?number,
 |};
 
 export type NavigationState = {|
--- a/newIDE/app/src/AssetStore/index.js
+++ b/newIDE/app/src/AssetStore/index.js
@@ -73,7 +73,9 @@ export const AssetStore = ({ project }: Props) => {
   const searchBar = React.useRef<?SearchBarInterface>(null);
 
   const shouldAutofocusSearchbar = useShouldAutofocusSearchbar();
-  const [isFiltersPanelOpen, setIsFiltersPanelOpen] = React.useState(false);
+  const [isFiltersPanelOpen, setIsFiltersPanelOpen] = React.useState(
+    !isOnHomePage && !openedAssetShortHeader
+  );
   const [
     selectedPrivateAssetPackListingData,
     setSelectedPrivateAssetPackListingData,
--- a/newIDE/app/src/AssetStore/index.js
+++ b/newIDE/app/src/AssetStore/index.js
@@ -72,12 +72,6 @@ export const AssetStore = ({ project }: Props) => {
   } = navigationState.getCurrentPage();
   const searchBar = React.useRef<?SearchBarInterface>(null);
 
-  const assetsHome = React.useRef<?AssetsHomeInterface>(null);
-  const boxSearchResults = React.useRef<?BoxSearchResultsInterface>(null);
-  const assetDetails = React.useRef<?AssetDetailsInterface>(null);
-  const scrollView =
-    assetsHome.current || boxSearchResults.current || assetDetails.current;
-
   const shouldAutofocusSearchbar = useShouldAutofocusSearchbar();
   const [isFiltersPanelOpen, setIsFiltersPanelOpen] = React.useState(false);
   const [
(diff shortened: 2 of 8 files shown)


newIDE/app/src/UI/Search/BoxSearchResults.js
```diff
@@ -24,86 +24,126 @@ const styles = {
...
+        if (!gridElement) return 0;
+
+        // TODO Find a clean way to get the scroll position.
+        // Using the internal state of a component is hacky.
+        // Grid probably doesn't expose the scroll position
+        // because it became irrelevant when the dimensions change.
+        // Though, it's easier to use it and the chance that the Grid is
+        // resized is low.
+        return gridElement.state.scrollTop;
+      },
```
D8H, line 62: This is not great.

newIDE/app/src/UI/Search/BoxSearchResults.js
```diff
@@ -24,86 +24,126 @@ const styles = {
...
+export type BoxSearchResultsInterface = {|
+  getScrollPosition: () => number,
+  scrollToPosition: (y: number) => void,
+|};
+
+export const BoxSearchResults = React.forwardRef<
+  // $FlowFixMe The generic type can't pass through.
+  Props<SearchItem>,
+  BoxSearchResultsInterface
+>(
```
D8H, line 36: I didn't find an elegant solution.

newIDE/app/src/UI/Search/BoxSearchResults.js
AlexandreSi: Did you try to declare a ref that would be `scrollTop` that would be updated on grid scroll?
`<Grid onScroll={(event) => {scrollTop.current = event.currentTarget.scrollTop}} ...>`

newIDE/app/src/AssetStore/AssetStoreNavigator.js
```diff
@@ -13,6 +13,7 @@ export type AssetStorePageState = {|
   openedAssetShortHeader: ?AssetShortHeader,
   filtersState: FiltersState,
   ignoreTextualSearch: boolean,
+  scrollPosition?: ?number,
```
AlexandreSi, line 16: Is this still needed?

newIDE/app/src/UI/ScrollView.js
```diff
@@ -47,6 +49,28 @@ export default React.forwardRef<Props, ScrollViewInterface>(
...
+      /**
+       * Scroll the view to the target position.
+       */
+      scrollToPosition: (y: number) => {
+        const scrollViewElement = scrollView.current;
+        if (!scrollViewElement) return;
+
+        const scrollViewYPosition = scrollViewElement.getBoundingClientRect()
+          .top;
+        scrollViewElement.scrollTop += y - scrollViewYPosition;
```
AlexandreSi, line 61: You can maybe put this in common in an util that the `scrollTo` method would use as well.

newIDE/app/src/AssetStore/index.js
```diff
@@ -80,6 +90,37 @@ export const AssetStore = ({ project }: Props) => {
...
+  React.useLayoutEffect(
+    () => {
+      const scrollPosition = navigationState.getCurrentPage().scrollPosition;
+      if (scrollPosition != null && scrollView && !hasScrolled.current) {
+        scrollView.scrollToPosition(scrollPosition);
+      }
+      hasAppliedSavedScrollPosition.current = true;
+    },
+    [isOnHomePage, navigationState, openedAssetShortHeader, scrollView]
+  );
```
D8H, line 112: I did a new version that uses call-backs to get the current scroll position.
The scroll position is applied back in a `useLayoutEffect` to have the right ref to the scrolled views.

At the 1st render, it has no effect on the scroll. It only scroll when the view is rendered again.
It works for the home page and the asset list because the view is always refreshed for some reason, but the asset detail doesn't refresh so the scroll is not applied.

newIDE/app/src/AssetStore/index.js
```diff
@@ -80,6 +90,37 @@ export const AssetStore = ({ project }: Props) => {
   ] = React.useState<?PrivateAssetPackListingData>(null);
   const { onPurchaseSuccessful } = React.useContext(AuthenticatedUserContext);
 
+  // The saved scroll position must not be reset by a scroll event until it
+  // has been applied.
+  const hasAppliedSavedScrollPosition = React.useRef<boolean>(false);
+  // The saved scroll position must not be applied after users have scrolled.
+  const hasScrolled = React.useRef<boolean>(false);
```
D8H, line 97: I bet there is a simpler solution, but couldn't find one.

newIDE/app/src/AssetStore/AssetStoreNavigator.js
D8H: Yes, it's what stores the stack of scroll positions.

newIDE/app/src/AssetStore/index.js
```diff
@@ -80,6 +90,37 @@ export const AssetStore = ({ project }: Props) => {
...
+  const needScrollUpdate = React.useCallback(() => {
```
AlexandreSi, line 98: I feel like this variable name is a check to know if something needs a scroll update although it defines there's a need. What do you think about `setScrollUpdateNeeded` instead?

newIDE/app/src/UI/Search/BoxSearchResults.js
AlexandreSi: Looks like it's a known issue

newIDE/app/src/AssetStore/index.js
```diff
@@ -68,8 +71,11 @@ export const AssetStore = ({ project }: Props) => {
     filtersState,
   } = navigationState.getCurrentPage();
   const searchBar = React.useRef<?SearchBarInterface>(null);
+
   const shouldAutofocusSearchbar = useShouldAutofocusSearchbar();
-  const [isFiltersPanelOpen, setIsFiltersPanelOpen] = React.useState(false);
+  const [isFiltersPanelOpen, setIsFiltersPanelOpen] = React.useState(
+    !isOnHomePage && !openedAssetShortHeader
```
AlexandreSi, line 77: This means the panel is opened when going back from details of an asset to the result list?

newIDE/app/src/AssetStore/index.js
```diff
@@ -80,6 +86,65 @@ export const AssetStore = ({ project }: Props) => {
...
+  );
+
+  const handleScroll = React.useCallback(
+    y => {
+      if (
+        navigationState.getCurrentPage().scrollPosition == null ||
+        hasAppliedSavedScrollPosition.current
+      ) {
+        navigationState.getCurrentPage().scrollPosition = y;
+        // The scroll position is right, we can continue to save it.
```
AlexandreSi, line 141: What do you mean the scroll position is right?

newIDE/app/src/AssetStore/index.js
D8H: I think this state is initialized only when the dialog is opened back.

AlexandreSi (approved):
LGTM!

4ian (commented):
I'm arriving a bit late, but onScroll can be costly on mobile devices/low end devices because you run JS at every single small scrolling amount. I think these event can be set to be "passive", but not sure if it's easy to do with React.
Instead, any chance we query and save the scroll Y position *only* when we're navigating out of the component?

It's a bit of a change but I think it's worth trying because these onScroll triggered 60 frame per second are notouriously creating issues or high CPU usage. And nobody likes a bad web-app that is lagging when you just scroll.

D8H:
In the first version, I used a getter in the ref interface. I'll try with this.

4ian (approved):

PR #4543, opened 2022-11-14
Hide warning if backdrop behavior is set to cancel

ClementPasteau:


newIDE/app/src/Utils/SerializableObjectCancelableEditor.js
```diff
@@ -64,13 +67,19 @@ export const useSerializableObjectCancelableEditor = ({
       if (!serializedElement) return;
 
       let continueCanceling = false;
+      const hasCancelBackdropPreference = backdropClickBehavior === 'cancel';
```
AlexandreSi, line 70: should this `'cancel'` be stored in a variable in the preferences? To avoid regressions if one day we decided to change it?

newIDE/app/src/Utils/SerializableObjectCancelableEditor.js
```diff
@@ -64,13 +67,19 @@ export const useSerializableObjectCancelableEditor = ({
...
 
+      // We show a warning if:
+      // - the user has not set the backdrop click behavior to "cancel", as we assume they know what they are doing
+      // - the user has made a significant number of changes
+      // - the user has been editing for a significant amount of time
```
AlexandreSi, line 75:
```suggestion
      // and, one of those conditions is met:
      // - the user has made a significant number of changes
      // - the user has been editing for a significant amount of time
```

AlexandreSi (approved):
LGTM

newIDE/app/src/Utils/SerializableObjectCancelableEditor.js
ClementPasteau: hmm, the variable is typed properly, so I assume we're fine in terms of catching a regression.
I also think it's clearer to use a string like this, rather than importing the constants everywhere we need them

AlexandreSi: Oh yeah, I didn't check for types

PR #4544, opened 2022-11-14
Include tags in the search for extensions and examples

D8H:
Here, the last result is only from a tag.
I guess, visually, it would be too heavy to display tags for each extension.

image

### DIFF
--- a/newIDE/app/src/UI/Search/UseSearchStructuredItem.js
+++ b/newIDE/app/src/UI/Search/UseSearchStructuredItem.js
@@ -205,6 +205,7 @@ export const useSearchStructuredItem = <
             { name: 'name', weight: 2 },
             { name: 'fullName', weight: 5 },
             { name: 'shortDescription', weight: 1 },
+            { name: 'tags', weight: 4 },
           ],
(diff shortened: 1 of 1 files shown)


newIDE/app/src/UI/Search/UseSearchStructuredItem.js
```diff
@@ -205,6 +205,7 @@ export const useSearchStructuredItem = <
             { name: 'name', weight: 2 },
             { name: 'fullName', weight: 5 },
             { name: 'shortDescription', weight: 1 },
+            { name: 'tags', weight: 4 },
```
D8H, line 208: What do you think of the weight?

4ian: tags are arrays, can you check Fuse.js is working properly with these?

D8H: It handles arrays according to the documentation:
https://fusejs.io/api/options.html#keys

ClementPasteau (approved):
lgtm 👍

PR #4553, opened 2022-11-16
Improve Searchbar

ClementPasteau:
Unify Searchbar across the app.
image

Can be played on storybook: https://gdevelop-storybook.s3.amazonaws.com/improve-search-bars/latest/index.html?path=/story/ui-building-blocks-searchbar--all-options


newIDE/app/src/EventsSheet/InstructionEditor/InstructionOrExpressionSelector/index.js
```diff
@@ -160,13 +160,7 @@ export default class InstructionOrExpressionSelector<
           helpPagePath={helpPagePath}
           ref={searchBar => (this._searchBar = searchBar)}
         />
-        <ScrollView
-          autoHideScrollbar
-          ref={
-            // $FlowFixMe - improper typing of ScrollView?
```
AlexandreSi, line 166: 👌

newIDE/app/src/UI/Theme/NordTheme/theme.json
```diff
@@ -49,6 +49,44 @@
...
     },
+    "search": {
+      "default": {
+        "background-color": {
+          "value": "#494952",
+          "comment": "Palette/Grey/70"
+        },
+        "text-color": {
+          "value": "#C5C5C9",
+          "comment": "Palette/Grey/30"
```
AlexandreSi, line 60: should those comments be removed for non official themes?

newIDE/app/src/UI/Theme/DefaultDarkTheme/theme.json
```diff
@@ -156,6 +156,44 @@
         }
       }
     },
+    "search": {
```
AlexandreSi, line 159: Could you add those new fields in `newIDE/app/scripts/theme-templates/theme.json` ?

AlexandreSi: Speaking of which, should this field be `searchBar` to be more accurate?

newIDE/app/src/UI/TagChips.js
```diff
@@ -1,7 +1,6 @@
 // @flow
 import React from 'react';
 import Chip from '../UI/Chip';
-import randomColor from 'randomcolor';
```
AlexandreSi, line 4: I guess you can remove the package as well

AlexandreSi: 👋

newIDE/app/src/UI/SearchBar.js
```diff
@@ -5,17 +5,10 @@ import * as React from 'react';
...
 import IconButton from './IconButton';
 import TextField from './TextField';
-import {
-  Collapse,
-  Typography,
-  TextField as MuiTextField,
-} from '@material-ui/core';
-import Close from '@material-ui/icons/Close';
-import Search from '@material-ui/icons/Search';
+import { Collapse, TextField as MuiTextField, Paper } from '@material-ui/core';
```
AlexandreSi, line 8: Separate this import in 3 separate imports

newIDE/app/src/UI/Theme/NordTheme/theme.json
ClementPasteau: we could, and yet there're using the official theme colors, so I thought it's not a bad idea to keep them (in case someone in the future want to change them)

AlexandreSi:
Using the stories, here are some questions:
- Should we focus the textfield after the cross button has been clicked?
- The focus ripple for the inline buttons is really hard to see on a Retina screen
- With the search bar with tags, if you set 4 tags:
  - You can focus tags using tab navigation and remove them using the Del key. I would expect as a user that when a tag is removed, the tag that takes its place is focused.
  - When you remove the last tag, the component crashes.

newIDE/app/src/UI/TextField.js
```diff
@@ -265,6 +264,7 @@ export default class TextField extends React.Component<Props, {||}> {
                 max: props.max,
                 min: props.min,
                 step: props.step,
+                style: props.inputStyle,
```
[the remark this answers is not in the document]
ClementPasteau: no, I checked everywhere and it's fine 👍 
I was surprised this was like that indeed

newIDE/app/src/UI/TagChips.js
ClementPasteau: 👋  we'll see each other soon I'm sure

ClementPasteau:
Thanks!
- textfield focused after click on cross OR last tag removed
- when removing a tag, the one taking the place is focused
- focus improved slightly

AlexandreSi (approved):
LGTM

PR #4557, opened 2022-11-17
Improve Instance decimals chaos

ClementPasteau:
Updated logic:

- If a "resize", "drag" or "rotate" is done in the instance editor, then we round the "size", "position" and "angle" to the nearest integer.
   - The assumption is that a user using the instance editor will not be looking for sub pixel precision
   - And if they do, they can still edit the instance properties with as many digits as they want

- The tooltip now displays up to 2 digits on the coordinates of the instances

- The number fields do not update the instance with "0" when a user is still typing (ex: "19," does not update the field)

- The instance number fields (width/height/x/y/angle) can now correctly accept precise float numbers, avoiding excessive and inaccurate digits

Before:

https://user-images.githubusercontent.com/4895034/207826496-089000a1-fe33-4b1d-8cb0-a9caa5b81cd4.mov

https://user-images.githubusercontent.com/4895034/207826511-0a02cd5d-f4df-4631-b18a-b58c4d965581.mov

After:

https://user-images.githubusercontent.com/4895034/207826567-6c2b237b-579c-40ad-ba48-d7936ce3b8d7.mov

https://user-images.githubusercontent.com/4895034/207826574-d421886f-b507-45da-9fea-0b314f244c2c.mov

### DIFF
--- a/newIDE/app/src/InstancesEditor/HighlightedInstance.js
+++ b/newIDE/app/src/InstancesEditor/HighlightedInstance.js
@@ -81,13 +81,14 @@ export default class InstancesSelection {
       highlightedInstance.getObjectName() +
       '\n' +
       'X: ' +
-      parseInt(highlightedInstance.getX()) +
+      Math.round(highlightedInstance.getX() * 100) / 100 + // An instance position can have a lot of decimals, so round to 2 decimals.
       '  Y: ' +
-      parseInt(highlightedInstance.getY()) +
+      Math.round(highlightedInstance.getY() * 100) / 100 + // An instance position can have a lot of decimals, so round to 2 decimals.
       '\n' +
       'Layer: ' +
-      highlightedInstance.getLayer() +
-      '  Z: ' +
+      (highlightedInstance.getLayer() || 'Base layer') +
+      '\n' +
+      'Z: ' +
       highlightedInstance.getZOrder() +
       '\n';
     this.tooltipText.text = tooltipInfo;
--- a/Core/GDCore/Project/InitialInstance.h
+++ b/Core/GDCore/Project/InitialInstance.h
@@ -7,13 +7,14 @@
 #ifndef GDCORE_INITIALINSTANCE_H
 #define GDCORE_INITIALINSTANCE_H
 #include <map>
+
 #include "GDCore/Project/VariablesContainer.h"
 #include "GDCore/String.h"
 namespace gd {
 class PropertyDescriptor;
 class Project;
 class Layout;
-}
+}  // namespace gd
 
 namespace gd {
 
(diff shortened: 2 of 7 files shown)


newIDE/app/src/Utils/Mathematics.js
```diff
@@ -2,3 +2,6 @@
 
 export const roundTo = (value: number, precision: number): number =>
   Math.round(value * Math.pow(10, precision)) / Math.pow(10, precision);
+
+export const truncTo = (value: number, precision: number): number =>
+  Math.trunc(value * Math.pow(10, precision)) / Math.pow(10, precision);
```
D8H, line 7: It probably won't work because `number` can't take any values.
https://www.exploringbinary.com/why-0-point-1-does-not-exist-in-floating-point/

AlexandreSi: I think there's something we don't understand be cause this works:
image
I guess the same would work for trunc

Bouh: Why not used [toFixed()](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed) ? It seem easier for display.

ClementPasteau: Problem with toFixed is that it returns a string, it's not made for operations on a number.
So it'll probably mess up the field.
I'll think about it

AlexandreSi:
Just curious: does it work if you type in `19,2` instead of `19.2`?

ClementPasteau:
> Just curious: does it work if you type in `19,2` instead of `19.2`?

Yes, the number field knows it's a number.
And the value is parsed before saved in the instance, so it will be saved as 19.2

newIDE/app/src/InstancesEditor/HighlightedInstance.js
```diff
@@ -81,13 +81,14 @@ export default class InstancesSelection {
       highlightedInstance.getObjectName() +
       '\n' +
       'X: ' +
-      parseInt(highlightedInstance.getX()) +
+      Math.round(highlightedInstance.getX() * 100) / 100 + // An instance position can have a lot of decimals, so round to 2 decimals.
```
4ian, line 84: Nit: this could be a roundUpTo(2, highlightedInstance.getX()) method, as this could allow in the future to easily adapt the rounding to 0, 1 or 2 decimal places.

4ian (approved):
Overall lgtm

newIDE/app/src/InstancesEditor/HighlightedInstance.js
ClementPasteau: I'm under the impression that this tooltip will disappear, or at least the content will change, as it doesn't seem to be used for these information.
So for the moment, not touching this

PR #4559, opened 2022-11-18
Physics2 behavior no longer step before the first frame

D8H:
* It allows events to access to the initial object positions.

### DIFF
--- a/Extensions/Physics2Behavior/physics2runtimebehavior.ts
+++ b/Extensions/Physics2Behavior/physics2runtimebehavior.ts
@@ -825,7 +825,7 @@ namespace gdjs {
     }
 
     doStepPreEvents(instanceContainer: gdjs.RuntimeInstanceContainer) {
-      // Step the world if not done this frame yet
+      // Step the world if not done this frame yet.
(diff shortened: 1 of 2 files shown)


Extensions/Physics2Behavior/physics2runtimebehavior.ts
```diff
@@ -826,7 +826,11 @@ namespace gdjs {
 
     doStepPreEvents(instanceContainer: gdjs.RuntimeInstanceContainer) {
       // Step the world if not done this frame yet
```
4ian, line 828:
```suggestion
      // Step the world if not done this frame yet.
```

Extensions/Physics2Behavior/tests/physics2runtimebehavior.spec.js
```diff
@@ -1,4 +1,4 @@
-describe('Physics2RuntimeBehavior', () => {
+describe.only('Physics2RuntimeBehavior', () => {
```
4ian, line 1: .only ;)

4ian (approved):

PR #4560, opened 2022-11-18
Add a condition to check if at least one touch is pressed.

D8H:
This is useful for extensions because they can't disable mouse emulation and could have the same event twice.
It avoids each behavior to remember if a touch is pressed when this touch may not concern their owner (when another button is pressed for instance).


D8H:
This is no longer needed with the new touch events system.

PR #4565, opened 2022-11-22
Add a command to generate an action and an expression for a property

D8H:
image

image

image


Core/GDCore/IDE/PropertyFunctionGenerator.cpp
```diff
@@ -0,0 +1,245 @@
...
+gd::String
+PropertyFunctionGenerator::CapitalizeFirstLetter(const gd::String &string) {
+  if (string.empty()) {
+    return string;
+  }
+  return string.substr(0, 1).UpperCase() + string.substr(1);
+}
+
+gd::String
+PropertyFunctionGenerator::UnCapitalizeFirstLetter(const gd::String &string) {
```
AlexandreSi, line 238: Should those 2 functions be in a separate String helpers file? Or in the String class?

Core/tests/PropertyFunctionGenerator.cpp
```diff
@@ -0,0 +1,493 @@
...
+      eventsExtension.GetEventsBasedBehaviors().InsertNew(
+          "MyEventsBasedBehavior", 0);
+  eventsBasedBehavior.SetFullName("My events based behavior");
+  eventsBasedBehavior.SetDescription("An events based behavior for test");
+  eventsBasedBehavior.SetObjectType("");
+  return eventsBasedBehavior;
+};
+} // namespace
+
+TEST_CASE("PropertyFunctionGenerator", "[common]") {
```
AlexandreSi, line 34: 💜

AlexandreSi:
In you screenshots, is it normal that you have `Health health configuration` (double health) in the group name?

newIDE/app/src/EventsBasedBehaviorEditor/EventsBasedBehaviorPropertiesEditor.js
```diff
@@ -225,6 +227,21 @@ export default class EventsBasedBehaviorPropertiesEditor extends React.Component
                           click: () => this._moveProperty(i, i + 1),
                           enabled: i + 1 < properties.getCount(),
                         },
+                        {
+                          label: i18n._(t`Generate functions`),
```
AlexandreSi, line 231: Should it be more specific? Generate functions to get and set value for this property?

newIDE/app/src/EventsBasedBehaviorEditor/EventsBasedBehaviorPropertiesEditor.js
```diff
@@ -225,6 +227,21 @@ export default class EventsBasedBehaviorPropertiesEditor extends React.Component
...
+                          click: () =>
```
AlexandreSi, line 232: What happens if you click twice on this option? There are identical functions that are created?

AlexandreSi (commented):
LGTM but not being used to work on that part, it might be good to have another approval

newIDE/app/src/EventsBasedBehaviorEditor/EventsBasedBehaviorPropertiesEditor.js
D8H: You can't as it gets disabled and there is also a check in the method to avoid overwriting existing functions.

D8H:
> In you screenshots, is it normal that you have `Health health configuration` (double health) in the group name?

The 1st one is the extension full name and the 2nd one is the property group. In this case, users will have to clean up a bit.

newIDE/app/src/EventsBasedBehaviorEditor/EventsBasedBehaviorPropertiesEditor.js
AlexandreSi: Your second suggestion reads fine I think. And not too long in my opinion

D8H: Ok, I'll go for "Generate expression and action".

newIDE/app/src/EventsBasedBehaviorEditor/EventsBasedBehaviorPropertiesEditor.js
D8H: No, it's disabled if one of them exist. I hesitated to make something more intelligent (as the existing function could be of the wrong type and break the generated one that depends on it) but it seems too complicated and hard to understand for users so I kept it simple.

Core/GDCore/Extensions/Metadata/ValueTypeMetadata.h
```diff
@@ -192,6 +192,11 @@ class GD_CORE_API ValueTypeMetadata {
   static const gd::String numberType;
   static const gd::String stringType;
 
+  /**
+   * \brief Return the ValueType name from the property type.
```
AlexandreSi, line 196: Could you add context about where those different types are used? Or a link to where the ValueType is defined/used?

ClementPasteau (approved):
lgtm, but maybe a quick review from @4ian ?

Core/tests/PropertyFunctionGenerator.cpp
4ian: Big thank you for these tests :)

Core/GDCore/IDE/PropertyFunctionGenerator.cpp
```diff
@@ -0,0 +1,245 @@
...
+#include "GDCore/Project/Project.h"
+#include "GDCore/Project/PropertyDescriptor.h"
+#include "GDCore/String.h"
+
+namespace gd {
+
+void PropertyFunctionGenerator::GenerateGetterAndSetter(
+    gd::Project &project, gd::EventsFunctionsExtension &extension,
+    gd::EventsBasedBehavior &eventsBasedBehavior,
+    const gd::NamedPropertyDescriptor &property, bool isSceneProperties) {
```
4ian, line 22: isSceneProperties or `isSharedProperties` ?

4ian (approved):

PR #4566, opened 2022-11-22
Improve login profile

ClementPasteau:
image
image
image
image


newIDE/app/src/UI/User/UserChip.js
```diff
@@ -27,44 +31,56 @@ const styles = {
...
 
-const UserChip = ({ profile, onClick, displayNotificationBadge }: Props) => {
+const UserChip = ({ onOpenProfile }: Props) => {
+  const authenticatedUser = React.useContext(AuthenticatedUserContext);
+  const { profile, onLogin, onCreateAccount, loginState } = authenticatedUser;
+  const displayNotificationBadge = hasPendingNotifications(authenticatedUser);
   const classes = useStyles();
-  return (
+  return loginState === 'loggingIn' ? (
+    <CircularProgress size={20} />
```
AlexandreSi, line 43: Should it be in a `Line` so that it keeps an height similar to a button?

newIDE/app/src/Profile/AdditionalUserInfoDialog.js
```diff
@@ -0,0 +1,287 @@
...
+  {
+    value: 'code-experience',
+    label: t`I've mainly created games with code (Godot, Unreal Engine, Unity, ...)`,
+  },
+  {
+    value: 'nocode-experience',
+    label: t`I have created games with no-code engines (GameMaker, Construct, Buildbox, ...)`,
+  },
+  {
+    value: 'full',
```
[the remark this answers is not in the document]
AlexandreSi: It's complicated afterwards when carrying out data analysis to remember what `full` means

newIDE/app/src/Profile/AdditionalUserInfoDialog.js
```diff
@@ -0,0 +1,287 @@
...
+              disabled={editInProgress}
+            >
+              {hearFromOptions.map(hearFromOption => (
+                <SelectOption
+                  key={hearFromOption.value}
+                  value={hearFromOption.value}
+                  primaryText={hearFromOption.label}
+                />
+              ))}
+            </SelectField>
```
AlexandreSi, line 279: Should those 4 select fields be within a map? To avoid the setting of the not-changing props in 4 different places

AlexandreSi: Or those not-changing props could be an object that is spread in each SelectField

newIDE/app/src/Profile/AuthenticatedUserProvider.js
```diff
@@ -35,6 +36,9 @@ import {
   listReceivedAssetShortHeaders,
   listReceivedAssetPacks,
 } from '../Utils/GDevelopServices/Asset';
+import AdditionalUserInfoDialog from './AdditionalUserInfoDialog';
+import { Trans } from '@lingui/macro';
+import { Snackbar } from '@material-ui/core';
```
AlexandreSi, line 41: `import Snackbar from '@material-ui/core/Snackbar';`

newIDE/app/src/UI/User/UserChip.js
ClementPasteau: The Line wasn't giving the expected height for some reason, so I just changed the size to match the same height 👍

newIDE/app/src/Profile/AdditionalUserInfoDialog.js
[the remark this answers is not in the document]
ClementPasteau: agreed, I went a bit fast on this

newIDE/app/src/Profile/AuthenticatedUserProvider.js
```diff
@@ -672,6 +727,20 @@ export default class AuthenticatedUserProvider extends React.Component<
     });
   };
 
+  openAdditionalUserInfoDialog = (open: boolean = true) => {
```
AlexandreSi, line 730: Is the default value useful?

newIDE/app/src/Profile/AuthenticatedUserProvider.js
```diff
@@ -748,6 +814,20 @@ export default class AuthenticatedUserProvider extends React.Component<
                 error={this.state.authError}
               />
             )}
+            {this.state.additionalUserInfoDialogOpen && (
+              <AdditionalUserInfoDialog
+                onClose={() => {
+                  this.openAdditionalUserInfoDialog(false);
+                  this.showUserSnackbar({
+                    message: <Trans>👋 Welcome to GDevelop!</Trans>,
```
AlexandreSi, line 822: Should it try to show the user username as for after the additional info dialog is closed?

newIDE/app/src/Profile/AdditionalUserInfoDialog.js
ClementPasteau: yeah that would improve the readability 👍

newIDE/app/src/Profile/CreateAccountDialog.js
```diff
@@ -102,19 +114,23 @@ const CreateAccountDialog = ({
...
+      });
+    } catch (error) {
+      console.error(error);
+    }
   };
 
   return (
     <Dialog
-      title={<Trans>Create a new GDevelop account</Trans>}
+      title={null}
```
AlexandreSi, line 133: Since you added a comment for each dialog that does not have a title, should we add one: `This dialog has a custom design to be more welcoming, the title is set in the content`

newIDE/app/src/Profile/AdditionalUserInfoDialog.js
ClementPasteau: actually, because each selectfield depends on a state value, I'm not sure it makes it simpler.
I'll leave it like this for the moment

newIDE/app/src/Profile/AuthenticatedUserProvider.js
ClementPasteau: I copied other functions in this file to be consistent.
It allows calling this function without passing a param

newIDE/app/src/Profile/CreateAccountDialog.js
```diff
@@ -149,54 +159,86 @@ const CreateAccountDialog = ({
...
+              label={<Trans>I want to receive the GDevelop Newsletter</Trans>}
+              checked={getNewsletterEmail}
+              onCheck={(e, value) => {
+                setGetNewsletterEmail(value);
+              }}
+              disabled={createAccountInProgress}
+            />
+          </ColumnStackLayout>
+        </div>
         <BackgroundText>
```
AlexandreSi, line 237: I wonder if this should be a Text with color secondary and a link rather than a background text with markdown

newIDE/app/src/Profile/LoginDialog.js
```diff
@@ -76,91 +73,95 @@ const LoginDialog = ({
...
-        disabled={actionInProgress}
+        disabled={loginInProgress}
       />
     </LeftLoader>,
   ];
 
   return (
     <Dialog
-      title={<Trans>Login to your GDevelop account</Trans>}
+      title={null}
```
AlexandreSi, line 83: Same here for the title

newIDE/app/src/Profile/ForgotPasswordDialog.js
```diff
@@ -0,0 +1,87 @@
...
+const ForgotPasswordDialog = ({ onClose, onForgotPassword }: Props) => {
+  const [email, setEmail] = React.useState('');
+  const [resetDone, setResetDone] = React.useState(false);
+  const [resetInProgress, setResetInProgress] = React.useState(false);
+
+  const doResetPassword = async () => {
+    if (resetInProgress || !email) return;
+    setResetInProgress(true);
+
+    await onForgotPassword({
```
AlexandreSi, line 27: Should you catch an error here?

newIDE/app/src/Profile/UserWelcomeDialog.js
```diff
@@ -0,0 +1,39 @@
...
+import { Column } from '../UI/Grid';
+import Text from '../UI/Text';
+import { useTimeout } from '../Utils/UseTimeout';
+
+type Props = {|
+  onClose: () => void,
+  message: React.Node,
+|};
+
+const UserWelcomeDialog = ({ onClose, message }: Props) => {
```
AlexandreSi, line 16: To remove

newIDE/app/src/UI/Link.js
```diff
@@ -1,23 +1,51 @@
...
+        cursor: !disabled ? 'pointer' : 'default',
+      },
+      '&:focus': {
+        color: !disabled ? theme.link.color.hover : undefined,
+      },
+    },
+  })();
+
+const Link = (props: Props) => {
+  const GDevelopTheme = React.useContext(GDevelopThemeContext);
```
AlexandreSi, line 31: Prefer lowercase for the variable

newIDE/app/src/UI/Link.js
```diff
@@ -1,23 +1,51 @@
...
+  const linkStyles = useLinkStyles(GDevelopTheme, !!props.disabled);
   const onClick = (event: MouseEvent) => {
-    event.preventDefault();
-    props.onClick();
+    if (!props.disabled) {
```
AlexandreSi, line 34: The click should have the default prevented in any case

newIDE/app/src/Profile/ForgotPasswordDialog.js
ClementPasteau: The error is already caught by the function itself, and not shown to the user;
The idea is to prevent giving any hint to a potential attacker if the email is in our DB

newIDE/app/src/Profile/ForgotPasswordDialog.js
```diff
@@ -0,0 +1,87 @@
...
+          onClick={onClose}
+          disabled={resetInProgress}
+        />,
+        !resetDone ? (
+          <LeftLoader isLoading={resetInProgress} key="reset">
+            <DialogPrimaryButton
+              label={<Trans>Reset password</Trans>}
+              primary
+              onClick={doResetPassword}
+              disabled={resetInProgress}
```
AlexandreSi, line 52: To disable if no email or if email does not match a regex? Not sure it's useful though

AlexandreSi (commented):
LGTM, I'll do another review for approval

newIDE/app/src/Profile/ForgotPasswordDialog.js
ClementPasteau: Adding a really simple regex

newIDE/app/src/Profile/CreateAccountDialog.js
ClementPasteau: As this is more of a "keep in mind these are the T&C's, you can click if you want" I like the fact that it's a greyer text, not appearing as a main action.
I think it's fine

AlexandreSi (approved):

PR #4573, opened 2022-11-23
Add analytics data when opening an asset pack or an asset

4ian:


newIDE/app/src/Utils/Analytics/EventSender.js
```diff
@@ -352,24 +366,27 @@ export const sendSubscriptionDialogShown = () => {
...
 };
 
-export const sendAssetAddedToProject = ({
-  id,
-  name,
-}: {|
+export const sendAssetAddedToProject = (options: {|
   id: string,
   name: string,
+  kind: 'public' | 'private' | 'unknown',
```
ClementPasteau, line 383: is there value in adding this prop?
An asset will probably always have the same kind as its pack

newIDE/app/src/Utils/Analytics/EventSender.js
```diff
@@ -289,16 +289,30 @@ export const sendOnboardingManuallyOpened = () => {
   recordEvent('onboarding_manually_opened');
 };
 
-export const sendAssetPackOpened = (assetPackTag: string) => {
-  recordEvent('asset_pack_opened', {
-    assetPackTag,
-  });
+export const sendAssetPackOpened = (options: {|
+  assetPackName: string,
+  assetPackTag?: string,
```
ClementPasteau, line 294: to be consistent with other functions, change type to `assetPackTag: string | null` ? (same for id)

newIDE/app/src/AssetStore/index.js
```diff
@@ -54,6 +57,27 @@ export type AssetStoreInterface = {|
   onClose: () => void,
 |};
 
+const identifyAssetPackKind = ({
```
ClementPasteau, line 60: I know PublicAssetPack should be the same as PublicAssetPack, but as it's not, should we say `isPrivate === !!assetPack.content` ?
(this props is only present on private packs)

ClementPasteau: This feels convoluted to search in the arrays.
And in the future, if both types merge to have the same props, we could add a `type: private | public` and use it?

ClementPasteau (approved):
lgtm, few feedbacks/questions

newIDE/app/src/AssetStore/index.js
4ian: I wonder if it's too fragile, and the day we update the backend to send a content for public packs, we now have wrong analytics. So we broke analytics by adding a field on the backend?

newIDE/app/src/Utils/Analytics/EventSender.js
4ian: Probably yes - but this is more for on the posthog/analytics side, to be able to explore multiple events and do a breakdown (i.e: a "group by" in SQL, a "breakdown" in Posthog) using the same field for multiple events.

4ian: This being said, I was assuming other events where having this "kind" for asset, which is not the case. I'm removing it!

newIDE/app/src/Utils/Analytics/EventSender.js
4ian: Done

newIDE/app/src/AssetStore/index.js
4ian: I'll go ahead with this and this is embedded in a function, so we an always simplify this later.

PR #4576, opened 2022-11-23
Allow extensions to define units of measurement for their properties

D8H:
- It use the scientific notation because it's the easiest to generate and it looks clean.
- It only works for behavior properties. It will be hard to do for objects like the ParticleEmitter because there is no property metadata.
- It doesn't allow event-based extensions to define units. I want to wait that everyone is ok with what is done for the built-in extensions because event-based extensions need a serialization and it's hard to go back.

image


Extensions/PlatformBehavior/PlatformerObjectBehavior.cpp
```diff
@@ -44,38 +45,70 @@ PlatformerObjectBehavior::GetProperties(
...
   std::map<gd::String, gd::PropertyDescriptor> properties;
 
-  properties[_("Gravity")].SetGroup(_("Jump")).SetValue(
-      gd::String::From(behaviorContent.GetDoubleAttribute("gravity")));
-  properties[_("Jump speed")].SetGroup(_("Jump")).SetValue(
-      gd::String::From(behaviorContent.GetDoubleAttribute("jumpSpeed")));
+  properties[_("Gravity")]
+      .SetGroup(_("Jump"))
+      .SetType("Number")
+      .SetMeasurementUnit(gd::MeasurementUnit::GetPixelAcceleration())
```
ClementPasteau, line 51: this isn't Newton because it's measured in pixels right?

Extensions/TopDownMovementBehavior/TopDownMovementBehavior.cpp
```diff
@@ -45,25 +46,46 @@ TopDownMovementBehavior::GetProperties(
...
+      .SetMeasurementUnit(gd::MeasurementUnit::GetPixelAcceleration())
+      .SetValue(
+          gd::String::From(behaviorContent.GetDoubleAttribute("deceleration")));
+  properties[_("Max. speed")]
+      .SetGroup(_("Movement"))
+      .SetType("Number")
+      .SetMeasurementUnit(gd::MeasurementUnit::GetPixelSpeed())
+      .SetValue(
+          gd::String::From(behaviorContent.GetDoubleAttribute("maxSpeed")));
+  properties[_("Rotate speed")]
```
ClementPasteau, line 72: same here, should it be "Rotation speed"?

newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/index.js
```diff
@@ -33,20 +40,40 @@ const NumericProperty = (props: {|
...
+      endAdornment={<UnitAdornment property={property} />}
     />
   );
 };
 
+const UnitAdornment = (props: {| property: gdPropertyDescriptor |}) => {
+  const { property } = props;
+  const measurementUnit = property.getMeasurementUnit();
+  return (
+    <Tooltip
```
ClementPasteau, line 62: on the screenshot, it felt to me that the title has a bit too much top margin from the top of the tooltip.
Is it something than can be improved?

newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/index.js
```diff
@@ -33,20 +40,40 @@ const NumericProperty = (props: {|
...
+      title={
+        <MarkdownText
+          source={getMeasurementUnitDescription(measurementUnit)}
```
ClementPasteau, line 65: can we translate this value? or is it already translated?

newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/index.js
```diff
@@ -215,6 +242,13 @@ const Physics2Editor = (props: Props) => {
               forceUpdate();
             }}
             type="number"
+            endAdornment={
+              <UnitAdornment
+                property={properties.get(
+                  shape === 'Polygon' ? 'PolygonOriginX' : 'shapeDimensionA'
```
ClementPasteau, line 248: Something weird in this file.
There is a check at the top of the textfield to display only if `shape !== 'Polygon'` 
And yet we do the check over and over again 🤔

Core/GDCore/Project/MeasurementUnit.cpp
```diff
@@ -0,0 +1,26 @@
...
+#include "MeasurementUnit.h"
+#include "GDCore/Serialization/SerializerElement.h"
+#include "GDCore/String.h"
+#include <vector>
+
+namespace gd {
+
+MeasurementUnit::~MeasurementUnit() {}
+
+gd::MeasurementUnit MeasurementUnit::unknown = CreateUnknown();
```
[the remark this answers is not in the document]
D8H: Properties are declared using something like a builder. The unit may not be explicitly declared. When this happens we shouldn't make assumption on the unit. For instance, in the future, when we do a dimensional analyzer, an unknown unit will allow to know that we must not warn the user about dimensional issues because we don't know.

Extensions/PathfindingBehavior/PathfindingBehavior.cpp
```diff
@@ -29,36 +29,69 @@ void PathfindingBehavior::InitializeContent(
 
 #if defined(GD_IDE_ONLY)
 std::map<gd::String, gd::PropertyDescriptor> PathfindingBehavior::GetProperties(
-    const gd::SerializerElement& behaviorContent) const {
+    const gd::SerializerElement &behaviorContent) const {
```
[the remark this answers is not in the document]
D8H: It's because I formatted the method.

Extensions/DestroyOutsideBehavior/DestroyOutsideBehavior.cpp
```diff
@@ -25,6 +25,7 @@ DestroyOutsideBehavior::GetProperties(
       .SetValue(gd::String::From(
           behaviorContent.GetDoubleAttribute("extraBorder", 0)))
       .SetType("Number")
+      .SetMeasurementUnit(gd::MeasurementUnit::GetPixel())
       .SetLabel(_("Margin before deleting the object, in pixels"));
```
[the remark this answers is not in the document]
D8H: Do you think we should keep it? I'm fine with either ways.
Do you fear that it could be understood as a temporal margin?

Extensions/PlatformBehavior/PlatformerObjectBehavior.cpp
D8H: and also there is no mass handling in the platformer behavior

newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/index.js
D8H: We could add left and right margin I guess.

newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/index.js
D8H: It's already translated Core side.

newIDE/app/src/PropertiesEditor/PropertiesMapToSchema.js
```diff
@@ -242,4 +249,50 @@ const propertiesMapToSchema = (
...
+      const signedUnit = (power < 0 ? 'per ' : '') + baseUnit.getLabel();
+      const separator = power < 0 ? ', ' : ' ';
+      const firstSeparator = i === 0 ? '' : wasNegativePower ? separator : ' ';
+      const poweredUnit =
+        firstSeparator +
+        signedUnit +
+        (absPower > 1 ? (separator + signedUnit).repeat(absPower - 1) : '');
+      wasNegativePower = power < 0;
+      return poweredUnit;
+    }).join('') +
```
D8H, line 293: I wonder if it's a bad idea to generate the worded unit because it will be hard to translate ("per" is a bit alone).
Should we add a property in MeasurementUnit for it instead?

newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/index.js
D8H, line 65: Maybe we should pass a react component instead of a markdown string to allow more precise layout?

Core/GDCore/Project/MeasurementUnit.cpp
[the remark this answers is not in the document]
ClementPasteau: ok 👍 
Or should it be named, `Unset` ? rather than unknown?

Extensions/DestroyOutsideBehavior/DestroyOutsideBehavior.cpp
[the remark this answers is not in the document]
ClementPasteau: let's keep it, it doesn't hurt imo!

newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/index.js
ClementPasteau: nice!

newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/index.js
D8H: Indeed there is a list to chose the origin for polygons.

Core/GDCore/Project/MeasurementUnit.cpp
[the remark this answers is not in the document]
D8H: I renamed it "undefined".

newIDE/app/src/PropertiesEditor/PropertiesMapToSchema.js
D8H: I did this.

newIDE/app/src/BehaviorsEditor/Editors/Physics2Editor/index.js
D8H: I did this too.

newIDE/app/src/PropertiesEditor/MeasurementUnitDocumentation.js
```diff
@@ -0,0 +1,33 @@
...
+import { Column } from '../UI/Grid';
+import Text from '../UI/Text';
+
+// Strings are used instead of gdMeasurementUnit
+// because the C++ instance would be dead when the component is rendered.
+type Props = {|
+  label: string,
+  description: string,
+  elementsWithWords: string,
+|};
```
D8H, line 13: Is it ok?

newIDE/app/src/PropertiesEditor/MeasurementUnitDocumentation.js
```diff
@@ -0,0 +1,33 @@
...
+  elementsWithWords,
+}: Props) {
+  return (
+    <Column noMargin>
+      <Text size="sub-title">{label}</Text>
+      {description && <Text size="body">{description}</Text>}
+      {elementsWithWords && (
+        <Text size="body">
+          <i>
+            <Trans>{`in ${elementsWithWords}`}</Trans>
```
D8H, line 27: The name is not great. I think it won't help translators much, but for what there is is to translate it may be ok.

Core/GDCore/Project/MeasurementUnit.cpp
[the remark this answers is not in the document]
ClementPasteau: nice!

newIDE/app/src/PropertiesEditor/MeasurementUnitDocumentation.js
ClementPasteau: I'm not sure what "dead" means

newIDE/app/src/PropertiesEditor/MeasurementUnitDocumentation.js
ClementPasteau: I'm not sure the variable name appears in crowding, they will see `in {0}`

ClementPasteau (approved):
lgtm 👍 
Just unsure about the typing comment's meaning

newIDE/app/src/PropertiesEditor/MeasurementUnitDocumentation.js
D8H: The `gdMeasurementUnit` instances are only living during their iteration on the properties when building the property editor view. As the rendering of `MeasurementUnitDocumentation` is done outside of this loop, the memory referenced by the `gdMeasurementUnit` instance is reused and it display wrong values.

ClementPasteau (approved):
maybe improve the comment on typing a bit, then good to go!

Core/GDCore/Project/MeasurementUnitElement.h
```diff
@@ -0,0 +1,46 @@
...
+#include "GDCore/String.h"
+
+namespace gd {
+class SerializerElement;
+}
+
+namespace gd {
+
+/**
+ * \brief An couple of an atomic unit of measurement and its power.
```
4ian, line 20: A couple

Core/GDCore/Project/PropertyDescriptor.h
```diff
@@ -103,12 +105,21 @@ class GD_CORE_API PropertyDescriptor {
     extraInformation.push_back(info);
     return *this;
   }
+  
+  /**
+   * \brief Change the group where this property is displayed to the user, if any.
```
4ian, line 110: Comment to update

4ian (approved):
Looks good generally but the translations won't work when you switch to another language because things are constructed statically. 
Would be worth having a "gd::MeasurementUnit::ApplyTranslation" to changes all the transaltable labels.
Note that it is a good thing to avoid recreating at every call a new MeasurementUnit! Otherwise this creates a lot of temporary just to destroy them after and this is needlessly costly. 
But the labels should be updated when the language is changed in the editor.

D8H:
I did the change, it seems to work.

image

There is a bit of duplicated code but I looked for static blocks in C++ and it seems complicated and risky.

4ian:
> There is a bit of duplicated code but I looked for static blocks in C++ and it seems complicated and risky.

Yes. Terribly risky (there is even a page about it: https://en.cppreference.com/w/cpp/language/siof). Let's avoid doing anything risky in the statics and just keep a good old function applying translations.

PR #4579, opened 2022-11-24
Improve Feedback panel with average rating and option to mark all as read

ClementPasteau:
- Added options inside a "options" icon
- New option to mark all as read
- New section with overall average ratings
- New numbers on the bars, for easier readability

https://user-images.githubusercontent.com/4895034/203826000-5eeb080e-c4fb-4164-8fd1-4e94ba5eca93.mov

### DIFF
--- a/newIDE/app/src/GameDashboard/Feedbacks/GameFeedback.js
+++ b/newIDE/app/src/GameDashboard/Feedbacks/GameFeedback.js
@@ -3,7 +3,7 @@ import * as React from 'react';
 import { t, Trans } from '@lingui/macro';
 import { type I18n as I18nType } from '@lingui/core';
 
-import { Column, Line, Spacer } from '../../UI/Grid';
+import { Column, Line } from '../../UI/Grid';
 import EmptyMessage from '../../UI/EmptyMessage';
 import PlaceholderLoader from '../../UI/PlaceholderLoader';
 import Text from '../../UI/Text';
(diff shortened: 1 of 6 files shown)


newIDE/app/src/GameDashboard/Feedbacks/GameFeedback.js
```diff
@@ -222,119 +245,216 @@ const GameFeedback = ({ i18n, authenticatedUser, game }: Props) => {
...
+        displayedFeedbacksArray
+          .filter(comment => !comment.processedAt)
+          .map(comment =>
+            updateComment(getAuthorizationHeader, profile.id, {
+              gameId: comment.gameId,
+              commentId: comment.id,
+              processed: true,
+            })
+          )
+      );
```
D8H, line 268: Does it send a request for each message?

D8H:
I guess that the action from the menu could be buttons as there is enough space but I don't think that blocking.

D8H (approved):

newIDE/app/src/GameDashboard/Feedbacks/GameFeedback.js
ClementPasteau: yes it does, I think it was the most pragmatic.
We could modify the backend services to only accept a gameId and loop through them if we realise this improves performance.
But as it's not really a critical part of the app, let's go with this and improve that later

PR #4581, opened 2022-11-24
Added lerp_angle() which takes the shortest path between two angles

tristanbob:
- This is my first commit to the core, so I'm unsure of its quality
- I copied another commit to understand what needed changed
- I have not tested this function (let me know of a good way to do this)

### DIFF
--- a/Core/GDCore/Extensions/Builtin/MathematicalToolsExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/MathematicalToolsExtension.cpp
@@ -492,6 +492,16 @@ BuiltinExtensionsImplementer::ImplementsMathematicalToolsExtension(
                      "",
                      "res/mathfunction.png")
       .SetHelpPath("/all-features/expressions");
+
+  extension
+      .AddExpression("lerp_angle",
+                     _("Lerp (Linear interpolation) between two angles"),
+                     _("Linearly interpolates between two angles (in degrees) by taking the shortest direction."),
+                     "",
+                     "res/mathfunction.png")
+      .AddParameter("expression", _("Starting angle, in degrees"))
+      .AddParameter("expression", _("Destination angle, in degrees"))
+      .AddParameter("expression", _("Interpolation value between 0 and 1."));
 }
 
 }  // namespace gd
--- a/Core/GDCore/Extensions/Builtin/MathematicalToolsExtension.cpp
+++ b/Core/GDCore/Extensions/Builtin/MathematicalToolsExtension.cpp
@@ -494,7 +494,7 @@ BuiltinExtensionsImplementer::ImplementsMathematicalToolsExtension(
       .SetHelpPath("/all-features/expressions");
 
   extension
-      .AddExpression("lerp_angle",
+      .AddExpression("lerpAngle",
                      _("Lerp (Linear interpolation) between two angles"),
                      _("Linearly interpolates between two angles (in degrees) by taking the shortest direction."),
                      "",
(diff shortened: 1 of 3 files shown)


Core/GDCore/Extensions/Builtin/MathematicalToolsExtension.cpp
```diff
@@ -492,6 +492,16 @@ BuiltinExtensionsImplementer::ImplementsMathematicalToolsExtension(
                      "",
                      "res/mathfunction.png")
       .SetHelpPath("/all-features/expressions");
+
+  extension
+      .AddExpression("lerp_angle",
```
AlexandreSi, line 497: Can you use camelCase for the expression and function names? There's currently a lack of consistency across all expressions, but I think that camelCase is the most used.

GDJS/GDJS/Extensions/Builtin/MathematicalToolsExtension.cpp
```diff
@@ -71,6 +71,7 @@ MathematicalToolsExtension::MathematicalToolsExtension() {
   GetAllExpressions()["XFromAngleAndDistance"].SetFunctionName("gdjs.evtTools.common.getXFromAngleAndDistance");
   GetAllExpressions()["YFromAngleAndDistance"].SetFunctionName("gdjs.evtTools.common.getYFromAngleAndDistance");
   GetAllExpressions()["Pi"].SetFunctionName("gdjs.evtTools.common.pi");
+  GetAllExpressions()["lerp_angle"].SetFunctionName("gdjs.evtTools.common.lerp_angle");
```
AlexandreSi, line 74: camelCase to use here as well

GDJS/Runtime/events-tools/commontools.ts
```diff
@@ -371,3 +371,18 @@ namespace gdjs {
...
+       * @param angle1 Starting angle, in degrees.
+       * @param angle2 Destination angle, in degrees.
+       * @param x The interpolation value between 0 and 1.
+       * @returns Return the interpolated angle, in degrees.
+       */
+       export const lerp_angle = function (
+        angle1: number,
+        angle2: number,
+        x: float
+      ): number {
```
AlexandreSi, line 386: In typescript (javascript), `number`, `float` and `integer` are the same.
It can help a human understand some things though. Here I would use `float` as return type since `x` is supposed to be a float.

GDJS/Runtime/events-tools/commontools.ts
```diff
@@ -371,3 +371,18 @@ namespace gdjs {
...
 }
+
+      /**
+       * Linearly interpolates between two angles (in degrees) by taking the shortest direction.
+       * @param angle1 Starting angle, in degrees.
+       * @param angle2 Destination angle, in degrees.
+       * @param x The interpolation value between 0 and 1.
+       * @returns Return the interpolated angle, in degrees.
+       */
+       export const lerp_angle = function (
```
AlexandreSi, line 382: This function should be in the namespace `gdjs.evtTools.common`, just below `pi`

AlexandreSi (commented):
Thanks for this contribution! I could test it on my laptop and works fine when the new function is in the correct namespace

AlexandreSi:
You might need to run auto formatting in the GDJS file as well. DM me if you need any help for that.

tristanbob:
@AlexandreSi I think I addressed all your suggestions. Thanks for coaching me on this!  

Please let me know if there are any other things that need to be changed on this PR.  :)

tristanbob:
@AlexandreSi how hard would it be to make a build with this change? I have not been able to test the new function yet.  

If you can provide a build I will make a simple example game to test it.

AlexandreSi:
LGTM But there's a formatting issue. I'll use your code to push a version that passes the tests so that it's in the upcoming version

AlexandreSi:
> @AlexandreSi how hard would it be to make a build with this change? I have not been able to test the new function yet.
> 
> If you can provide a build I will make a simple example game to test it.

It would take some time. I already tested it and it seemed fine to me. You'll be able to test it when the new version is available and if anything weird happens, we'll fix or revert it.

PR #4585, opened 2022-11-25
Show the list of parameters in expression documentations

D8H:
image

It's a place where the unit could be displayed.

### DIFF
--- a/newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/ExpressionAutocompletionsDisplayer.js
+++ b/newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/ExpressionAutocompletionsDisplayer.js
@@ -262,7 +262,7 @@ const ExpressionDocumentation = ({
         expressionMetadata.getParameters(),
         (parameter, parameterIndex) =>
           isParameterVisible(expressionMetadata, parameterIndex) && (
-            <Text style={defaultTextStyle} size="body2">
+            <Text style={defaultTextStyle} size="body2" key={parameterIndex}>
               <i>
                 {i18n._(
                   parameterRenderingService.getUserFriendlyTypeName(
--- a/newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/ExpressionAutocompletionsDisplayer.js
+++ b/newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/ExpressionAutocompletionsDisplayer.js
@@ -226,6 +226,22 @@ const DisplayedBehaviorAutocompletion = React.forwardRef(
   }
 );
 
+const isParameterVisible = (
+  expressionMetadata: gdExpressionMetadata,
+  parameterIndex: number
+): boolean => {
+  const parameter = expressionMetadata.getParameter(parameterIndex);
+  return (
+    !parameter.isCodeOnly() &&
+    // This filters parameters that are implicit because of the context
+    // (MyObject.MyBehavior::).
+    // Free functions have an instanceContainer as first parameter,
+    // their first object parameter are kept.
+    (parameterIndex !== 0 || parameter.getType() !== 'object') &&
+    (parameterIndex !== 1 || parameter.getType() !== 'behavior')
+  );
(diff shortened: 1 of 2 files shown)


newIDE/app/src/EventsSheet/ParameterRenderingService.js
```diff
@@ -156,7 +156,7 @@ const userFriendlyTypeName: { [string]: MessageDescriptor } = {
   objectAnimationName: t`Object animation name`,
   functionParameterName: t`Parameter name`,
   externalLayoutName: t`Name of the external layout`,
-  identifier: t`Name of the identifier`,
+  identifier: t`Identifier`,
```
D8H, line 159: It could even be "String" as the fact that it is an generalized identifier is a technical aspect.

newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/ExpressionAutocompletionsDisplayer.js
```diff
@@ -225,6 +226,46 @@ const DisplayedBehaviorAutocompletion = React.forwardRef(
...
+      </Text>
+      {mapVector(
+        expressionMetadata.getParameters(),
+        (parameter, index) =>
+          !parameter.isCodeOnly() &&
+          // This filters parameters that are implicit because of the context.
+          // Free functions have an instanceContainer as first parameter,
+          // their first object parameter are kept.
+          (index !== 0 || parameter.getType() !== 'object') &&
+          (index !== 1 || parameter.getType() !== 'behavior') && (
```
AlexandreSi, line 253: Could these different checks be put in a dedicated function with a comment per line explaining why they are excluded?

4ian: Agree! We had this kind of stuff in some files related to the events based extensions editors.

4ian (approved):
lgtm once comments are addressed! Will be very useful.

newIDE/app/src/EventsSheet/ParameterFields/GenericExpressionField/ExpressionAutocompletionsDisplayer.js
```diff
@@ -225,6 +226,46 @@ const DisplayedBehaviorAutocompletion = React.forwardRef(
...
+            <Text style={defaultTextStyle} size="body2">
```
AlexandreSi, line 254: It might feel better organized visually if you remove the margin on the `Text` components and put all the components in a column.
Plus, you should have a warning because they don't have a key props. I think you can use the type + an index for the key?

D8H: I'm not sure to follow.
The screenshot of the description is not up to date. It looks like this:

image

D8H: @AlexandreSi do you think something need to be changed?

AlexandreSi: What I saw was that there is lot of space between each parameter (due to the `Text` component margins). I thought that maybe you could remove those margins (but maybe they will be too close to each other, I'll let you decide).

If, as a result, there's not enough margin between the description and the parameters, you can wrap the parameters in a `Column` component that brings its own margin.

Finally, the component that results from the map does not have a key, so React should be printing some errors/warnings in your JS console saying that they miss a key. So you could set the key such as:

```jsx
<Text style={defaultTextStyle} size="body2" key={`expression-metadata-parameter-${parameterIndex}`}>
```

D8H: Ok, I added the key.
I like the spacing, it's not far from what VSC does.

image

PR #4590, opened 2022-11-28
Improve initial dialog logic to avoid opening tab on subsequent profile openings.

ClementPasteau:
Fix edge case, where after opening a link to a game tab, all subsequent profile openings redirect to the game dashboard. (because state is not updated)


AlexandreSi (approved):
LGTM

PR #4592, opened 2022-11-28
Correctly deactivate cloud projects on desktop app in prod

AlexandreSi:
Feature flag misses a check on presence of electron in files that are used by both the Browser and the Local apps


ClementPasteau (approved):
🆗

PR #4605, opened 2022-11-30
Fix boolean property condition sentences

D8H:
It also factorize the declaration for behavior and object.

For boolean properties, I wanted to add quotes (escaped) around the property label because it is weird grammatically but it seems like the translation tool doesn't like it.

image

image

image


newIDE/app/src/EventsFunctionsExtensionsLoader/MetadataDeclarationHelpers.js
```diff
@@ -830,6 +830,85 @@ const getStringifiedExtraInfo = (property: gdPropertyDescriptor) => {
     : '';
 };
 
+const uncapitalizedFirstLetter = (string: string): string =>
+  string.length < 1
```
AlexandreSi, line 834: You could use the condition `!string` instead but I'm not sure it's easier to read.

AlexandreSi (approved):
LGTM

PR #4606, opened 2022-12-01
Display some properties in one row

D8H:
It also avoid empty property groups to show (if all their properties are private). This could allow extension creators to organize their private properties.

image

### DIFF
--- a/Extensions/TopDownMovementBehavior/TopDownMovementBehavior.cpp
+++ b/Extensions/TopDownMovementBehavior/TopDownMovementBehavior.cpp
@@ -46,47 +46,55 @@ TopDownMovementBehavior::GetProperties(
     const gd::SerializerElement& behaviorContent) const {
   std::map<gd::String, gd::PropertyDescriptor> properties;
 
-  properties[_("Allows diagonals")]
+  properties["AllowDiagonals"]
+      .SetLabel(_("Allows diagonals"))
       .SetGroup(_("Movement"))
       .SetValue(behaviorContent.GetBoolAttribute("allowDiagonals") ? "true"
                                                                    : "false")
       .SetType("Boolean");
-  properties[_("Acceleration")]
+  properties["Acceleration"]
+      .SetLabel(_("Acceleration"))
       .SetGroup(_("Movement"))
       .SetType("Number")
       .SetMeasurementUnit(gd::MeasurementUnit::GetPixelAcceleration())
       .SetValue(
           gd::String::From(behaviorContent.GetDoubleAttribute("acceleration")));
-  properties[_("Deceleration")]
+  properties["Deceleration"]
+      .SetLabel(_("Deceleration"))
       .SetGroup(_("Movement"))
       .SetType("Number")
       .SetMeasurementUnit(gd::MeasurementUnit::GetPixelAcceleration())
       .SetValue(
           gd::String::From(behaviorContent.GetDoubleAttribute("deceleration")));
-  properties[_("Max. speed")]
+  properties["MaxSpeed"]
+      .SetLabel(_("Max. speed"))
       .SetGroup(_("Movement"))
       .SetType("Number")
       .SetMeasurementUnit(gd::MeasurementUnit::GetPixelSpeed())
       .SetValue(
           gd::String::From(behaviorContent.GetDoubleAttribute("maxSpeed")));
-  properties[_("Rotation speed")]
+  properties["AngularMaxSpeed"]
+      .SetLabel(_("Rotation speed"))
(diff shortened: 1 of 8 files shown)


Extensions/TopDownMovementBehavior/TopDownMovementBehavior.cpp
```diff
@@ -46,47 +46,55 @@ TopDownMovementBehavior::GetProperties(
...
-  properties[_("Rotation speed")]
+  properties["AngularMaxSpeed"]
+      .SetLabel(_("Rotation speed"))
       .SetGroup(_("Rotation"))
       .SetType("Number")
       .SetMeasurementUnit(gd::MeasurementUnit::GetAngularSpeed())
       .SetValue(gd::String::From(
           behaviorContent.GetDoubleAttribute("angularMaxSpeed")));
-  properties[_("Rotate object")]
+  properties["rotateObject"]
```
ClementPasteau, line 83: capital `R`?

Extensions/TopDownMovementBehavior/TopDownMovementBehavior.cpp
```diff
@@ -135,19 +146,19 @@ bool TopDownMovementBehavior::UpdateProperty(
     gd::SerializerElement& behaviorContent,
     const gd::String& name,
     const gd::String& value) {
-  if (name == _("Default controls")) {
+  if (name == "ignoreDefaultControls") {
```
ClementPasteau, line 149: capital `I`?

newIDE/app/src/UI/ColorField/index.js
```diff
@@ -71,10 +72,14 @@ export default class ColorField extends React.Component<Props, State> {
   render() {
     return (
       <div
-        style={{
-          ...styles.container,
-          width: this.props.fullWidth ? '100%' : undefined,
-        }}
+        style={
+          this.props.style
```
ClementPasteau, line 76: what style is passed to this component?
I'm wondering if this could be simplified to just a set of props, to avoid making this component's style overridden that easily

4ian: Agree, 99% of the time we want to avoid these custom styles.

D8H: `PropertiesEditor` passes this:
```JS
  field: {
    flex: 1,
    width: 'auto',
  },
```

D8H: Maybe I should just add `<Column expand>` around the ColorField in the PropertiesEditor?

newIDE/app/src/PropertiesEditor/index.js
```diff
@@ -333,21 +333,23 @@ const PropertiesEditor = ({
...
-            fullWidth
-            color={getFieldValue(instances, field)}
-            onChange={color => {
-              const rgbString =
-                color.length === 0 ? '' : rgbOrHexToRGBString(color);
-              instances.forEach(i => setValue(i, rgbString));
-              _onInstancesModified(instances);
-            }}
-          />
+          <Column expand noMargin>
```
4ian, line 336: Looks much better 👍

Extensions/TopDownMovementBehavior/TopDownMovementBehavior.cpp
```diff
@@ -46,47 +46,55 @@ TopDownMovementBehavior::GetProperties(
     const gd::SerializerElement& behaviorContent) const {
   std::map<gd::String, gd::PropertyDescriptor> properties;
 
-  properties[_("Allows diagonals")]
+  properties["AllowDiagonals"]
```
4ian, line 49: Any thoughts on using camelCase rather than PascalCase? I think some behaviors already use camelCase for properties, because they are like a, well, property on an object (`myObject.myProperty` in code)

newIDE/app/src/UI/ColorField/index.js
4ian: Much better, thanks!

Extensions/TopDownMovementBehavior/TopDownMovementBehavior.cpp
D8H: This is the good practice for community extensions. It's better if every extensions use the same format. It avoids to care about the case in code handling property name most of the time.

4ian: Alright then. At some point casing should not be an issue anyway, it's the kind of stuff we wonder about as developers but that should be a non issue entirely.

4ian (approved):

PR #4623, opened 2022-12-05
Add object name in keys and ids to avoid duplicates

AlexandreSi:


newIDE/app/src/EventsSheet/InstructionEditor/InstructionEditorDialog.js
```diff
@@ -87,7 +87,7 @@ const getInitialTab = (
  * A responsive instruction editor in a dialog, showing InstructionParametersEditor
  * at the end.
  */
-export default function NewInstructionEditorDialog({
```
ClementPasteau, line 90: 🤝

newIDE/app/src/EventsSheet/InstructionEditor/InstructionOrExpressionSelector/index.js
```diff
@@ -171,12 +174,9 @@ export default class InstructionOrExpressionSelector<
                   }) =>
                     renderInstructionOrExpressionListItem({
                       instructionOrExpressionMetadata: enumeratedInstructionOrExpressionMetadata,
-                      id:
-                        'instruction-or-expression-' +
-                        enumeratedInstructionOrExpressionMetadata.type.replace(
-                          /:/g,
-                          '-'
-                        ),
+                      id: getInstructionOrExpressionIdentifier(
```
ClementPasteau, line 177: just to be sure, Isn't this ID used for the in-app tutorial?

newIDE/app/src/EventsSheet/InstructionEditor/SelectorListItems/SelectorInstructionsTreeListItem.js
```diff
@@ -57,6 +57,9 @@ export const renderInstructionOrExpressionTree = <
             primaryText={key}
             selected={selected}
             id={
+              // TODO: This id is used by in app tutorials. When in app tutorials
```
ClementPasteau, line 60: I feel that this todo will stay in the code for a while if we don't do it now.
Shall we update the tutorial now?

4ian: Also, in addition of an id, feel free to use a class name. After all, it's not that different and while an identifier is good, a class name will do the trick too. Just make sure, like identifiers, that is unique and well prefixed so it can't be confused for something else.

newIDE/app/src/EventsSheet/InstructionEditor/InstructionOrExpressionSelector/index.js
AlexandreSi: It is but not for instructions linked to object so it does not change their ids

ClementPasteau (approved):

PR #4624, opened 2022-12-05
Add link to asset store license + Add message when no projects in the list

AlexandreSi:
- Add a link to the asset store license

  image

- Add a message in list of recent projects when there is no projects (inspired by Figma)

  image

  image

### DIFF
--- a/newIDE/app/src/AssetStore/PrivateAssets/PrivateAssetPackInformationDialog.js
+++ b/newIDE/app/src/AssetStore/PrivateAssets/PrivateAssetPackInformationDialog.js
@@ -29,6 +29,7 @@ import RaisedButton from '../../UI/RaisedButton';
 import { sendAssetPackBuyClicked } from '../../Utils/Analytics/EventSender';
 import { MarkdownText } from '../../UI/MarkdownText';
 import Paper from '../../UI/Paper';
+import Window from '../../Utils/Window';
 
 const sortedContentType = [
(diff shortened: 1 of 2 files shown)


newIDE/app/src/AssetStore/PrivateAssets/PrivateAssetPackInformationDialog.js
```diff
@@ -278,6 +279,21 @@ const PrivateAssetPackDialog = ({
...
+                                <Trans>See details</Trans>{' '}
+                                <Link
+                                  onClick={() =>
+                                    Window.openExternalURL(
+                                      'https://gdevelop.io/page/asset-store-license-agreement'
+                                    )
+                                  }
+                                  href="https://gdevelop.io/page/asset-store-license-agreement"
+                                >
+                                  <Trans>here</Trans>
```
4ian, line 293: A bit hard to translate ("Voir détail" "ici" even in French is not super great), I think you can put the link on the whole text "See details here."?

4ian (approved):

newIDE/app/src/AssetStore/PrivateAssets/PrivateAssetPackInformationDialog.js
AlexandreSi: Good idea, I'll change it

PR #4625, opened 2022-12-06
Search variables

### LINKED ISSUE #4295: Search and replace does not find strings with Variable prefix thus no search and replace possible.
**Bug**
If I search for example Variable(VariableName) it gives no results even if there are dozens of Variable(VarableName) strings in multiple calculations.
BUT it does find just the (VariableName) strings without the Variable prefix just fine.

image
In this example there are multiple Variable(BeamRayCastX) strings inside those calculations that are not found so I cannot find and thus not being able to replace them either.

**To Reproduce**
Make something having calculations using Variables and then search the Variable(VariableName)

Tested with Win10 & 11
GDev version 5.0.140

AlexandreSi:
Fixes #4295 

Also:
- Make it possible to use search panel on small screens (mobile)

  image

### DIFF
--- a/newIDE/app/src/EventsSheet/SearchPanel.js
+++ b/newIDE/app/src/EventsSheet/SearchPanel.js
@@ -166,9 +166,9 @@ const SearchPanel = (
 
   return (
     <Background noFullHeight noExpand>
-      <Column>
+      <Column noMargin>
         <Line>
-          <Column expand>
+          <Column expand noOverflowParent>
             <Tabs
               value={currentTab}
               onChange={setCurrentTab}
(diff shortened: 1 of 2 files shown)


newIDE/app/src/EventsSheet/SearchPanel.js
```diff
@@ -185,7 +189,7 @@ const SearchPanel = (
         </Line>
         <Line noMargin>
           <ColumnStackLayout expand>
-            <Line alignItems="baseline" noMargin>
+            <LineStackLayout alignItems="baseline" noMargin>
```
ClementPasteau, line 192: ❤️

ClementPasteau (approved):
Nice!

PR #4627, opened 2022-12-06
Underline search results

### LINKED ISSUE #3725: Search results highlight UI degradation
The new "highlight search results" feature is a bit useless to me (sorryyy). 
And the problem is, the search word itself is difficult to read because of the lower contrast. And the whole text is really hard to read because of those squares in it. Please, please consider removing it or adding a switch in Preferences. (not just change from light gray to blue or so)

I wanted to use the old version because of that, but the autoupdate switch seems to be broken.
Others are having the same problem at the moment.

____________

The color and position of true/false and yes/now buttons is also worse. Not thaaat bad, but maybe we can use the old position and the transparent for not-selected and blue for selected. The white and purple(at light theme) doesn't really fit with the rest.

AlexandreSi:
Resolves #3725

Remove background color to highlight matching search results.
New highlight is more conform to web standards (underline)

image

### DIFF
--- a/newIDE/app/src/UI/Search/HighlightedText.js
+++ b/newIDE/app/src/UI/Search/HighlightedText.js
@@ -1,13 +1,15 @@
 // @flow
 import * as React from 'react';
-import GDevelopThemeContext from '../Theme/ThemeContext';
+
+const styles = {
+  highlightedText: { textDecoration: 'underline' },
+};
 
 const highlightText = (
   text: string,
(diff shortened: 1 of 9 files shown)


AlexandreSi:
Closing this as there is no clear decision on what's best between:
- Underlining the text
- Adding a flashy background color with contrasted text color
- Adding a light background color keeping the same text color

PR #4642, opened 2022-12-08
Improve the split buttons and the preview button to make it clearer

4ian:
image
image
image

### DIFF
--- a/newIDE/app/src/AssetStore/ExampleStore/ExampleListItem.js
+++ b/newIDE/app/src/AssetStore/ExampleStore/ExampleListItem.js
@@ -11,7 +11,7 @@ import ButtonBase from '@material-ui/core/ButtonBase';
 import Text from '../../UI/Text';
 import { Trans } from '@lingui/macro';
 import { Column, Line } from '../../UI/Grid';
-import RaisedButtonWithSplitMenu from '../../UI/RaisedButtonWithSplitMenu';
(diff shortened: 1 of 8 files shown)


newIDE/app/src/UI/RaisedButtonWithSplitMenu.js
```diff
@@ -11,7 +11,7 @@ import { Spacer } from './Grid';
 type Props = {|
   id?: string,
   label?: React.Node,
-  primary?: boolean,
+  primary: true, // Force making only primary raised split buttons.
```
ClementPasteau, line 14: maybe make the prop non mandatory, so we don't think about it.
And yet, we have a warning if we try to set false?

ClementPasteau: There are a few places in the code where this button is used, without setting this prop, you may have to change those

ClementPasteau (approved):

newIDE/app/src/UI/RaisedButtonWithSplitMenu.js
4ian: In theory Flow should error if there is any prop where primary is not set to true. I should have replaced these by FlatButtonWithSplitMenu - so that it's clear.
I'll keep this so that anyone trying to do a non primary RaisedButtonWithSplitMenu think twice about really doing it.

PR #4643, opened 2022-12-08
Event-based object layout

D8H:
### Demo
https://github.com/GDevelopApp/GDevelop-examples/pull/411


4ian (commented):
If I do a change to a custom object and then go back to the scene, is the instance renderer adapting properly? If not, do I need to re-open the scene (and what do we need to do so that it works seamlessly?)

4ian (approved):
lgtm.
There is a lot of logic which can be only "visually tested". I"m not sure we want a full test suite for this, but at the same time I have the feeling this will stay for a very long time in the app. 
I think it would be worth at some point having a unit test with a few programmatically created custom objects and check that the rendered child-instances are put at the right positions.

Last stuff: double check if we should say child-objects or child-instances in all the comments, to be sure we use the proper word :)

D8H:
> Last stuff: double check if we should say child-objects or child-instances in all the comments, to be sure we use the proper word :)

- The term "object" is used in comments about the layout declaration because the layout is done with one instance per object-child and the object name is used to reference these instances.
- The term "instance" is used for the layout calculus because it's actually instances that are in the scene editor.

I added this in a comment. As there is a 1 to 1 relation between objects and instances, I think it won't be too much confusing.

newIDE/app/src/ObjectsRendering/Renderers/CustomObjectLayoutingModel.spec.js
```diff
@@ -0,0 +1,462 @@
...
+
+  getWidth() {
+    return this.width;
+  }
+
+  getHeight() {
+    return this.height;
+  }
+
+  addChid(
```
4ian, line 438: Everywhere: addChid => addChild

4ian (approved):
lgtm after the fix for the typo.

PR #4648, opened 2022-12-08
Allow to set the text alignment for text objects by property.

D8H:
The text aligns on the custom size box (for x axis only).

image

The center of rotation is
- the center of the custom width (wrapping width)
- still the center of the actual text height

I kept the vertical center as it was because at runtime the height is dynamically updated according to the text to display. There is no actual custom height.

https://user-images.githubusercontent.com/2611977/206686337-77fe2316-649b-4392-b689-15fb7763a0fb.mp4

This means that a text with lines smaller than the wrapping will no revolve around the rendered text. It allows a consistent behavior when the whole wrapping width is never used because words are not break.

https://user-images.githubusercontent.com/2611977/206686349-e8730b13-66a1-4bca-b7bb-acd2c7f2d5aa.mp4

https://user-images.githubusercontent.com/2611977/206686362-65771bf6-123b-4315-ad29-75dc461db4f6.mp4

When the wrapping is enabled, it's no longer possible to know the actual size of the rendered text. It was useful to fix the text alignment, but, as it's now correctly handled by the objet, it's no longer useful.

The object editor:

image

### DIFF
--- a/newIDE/app/src/ObjectEditor/Editors/TextEditor.js
+++ b/newIDE/app/src/ObjectEditor/Editors/TextEditor.js
@@ -7,12 +7,17 @@ import Checkbox from '../../UI/Checkbox';
 import { Line, Column } from '../../UI/Grid';
 import ColorPicker from '../../UI/ColorField/ColorPicker';
 import MiniToolbar, { MiniToolbarText } from '../../UI/MiniToolbar';
+import Text from '../../UI/Text';
 import ResourceSelector from '../../ResourcesList/ResourceSelector';
 import ResourcesLoader from '../../ResourcesLoader';
 import { type EditorProps } from './EditorProps.flow';
 import SemiControlledTextField from '../../UI/SemiControlledTextField';
 import ButtonGroup from '@material-ui/core/ButtonGroup';
 import Button from '@material-ui/core/Button';
+import Tooltip from '@material-ui/core/Tooltip';
+import LeftTextAlignment from '../../UI/CustomSvgIcons/LeftTextAlignment';
+import CenterTextAlignment from '../../UI/CustomSvgIcons/CenterTextAlignment';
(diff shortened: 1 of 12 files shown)


4ian (approved):
lgtm!

4ian:
> I kept the vertical center as it was because at runtime the height is dynamically updated according to the text to display. There is no actual custom height.

This seems correct to me and the expected behavior. A text height can't be defined, it's computed dynamically. Only the wrapping width (or lack of) can be set and followed by the text.

newIDE/app/src/ObjectEditor/Editors/TextEditor.js
```diff
@@ -97,26 +107,64 @@ export default class TextEditor extends React.Component<EditorProps, void> {
...
-            initialResourceName={textObjectConfiguration.getFontName()}
-            onChange={resourceName => {
-              textObjectConfiguration.setFontName(resourceName);
-              this.forceUpdate();
-            }}
-            hintText={<Trans>Choose a font</Trans>}
-            style={styles.resourcesSelector}
-          />
+          <ButtonGroup>
+            <Tooltip title={<Trans>Align text on the left</Trans>}>
```
AlexandreSi, line 111: I think you can use the component IconButton with the `tooltip` prop

AlexandreSi: You might need to add some props to the component

D8H: I used the same way as `YesNoField`. It doesn't seem possible to do the color inversion with `IconButton`.
I added a screenshot of the UI at the end of the description.

4ian: I'm fine with this exceptional use of material ui components because this is the first time we're doing a "three part toggle". If we do it again, we'll factor it.

PR #4652, opened 2022-12-09
Improve the design of the project manager and buttons to add items in lists

4ian:
Desktop:
image

Web (with shortcuts)
image

Web (small screens):
image

### DIFF
--- a/newIDE/app/src/ProjectManager/ShortcutsReminder.js
+++ b/newIDE/app/src/ProjectManager/ShortcutsReminder.js
@@ -1,0 +1,90 @@
+// @flow
+import * as React from 'react';
+import { Trans } from '@lingui/macro';
+import { useResponsiveWindowWidth } from '../UI/Reponsive/ResponsiveWindowMeasurer';
+import optionalRequire from '../Utils/OptionalRequire';
+import AlertMessage from '../UI/AlertMessage';
+import { Line, Spacer } from '../UI/Grid';
+import Text from '../UI/Text';
+import { adaptAcceleratorString } from '../UI/AcceleratorString';
+import { getElectronAccelerator } from '../KeyboardShortcuts';
+import { type ShortcutMap } from '../KeyboardShortcuts/DefaultShortcuts';
+import { Column } from '../UI/Grid';
+const electron = optionalRequire('electron');
+
+const styles = {
+  shortcutReminders: {
+    opacity: 0.7,
+  },
+};
+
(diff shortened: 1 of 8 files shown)


newIDE/app/src/ProjectManager/ShortcutsReminder.js
```diff
@@ -0,0 +1,90 @@
...
+              )}
+            </Trans>
+          </Text>
+        </Line>
+        <Line noMargin justifyContent="space-between">
+          <Text size="body2" noMargin>
+            <Trans>Close</Trans>
+          </Text>
+          <Text size="body2" noMargin>
+            <Trans>
```
AlexandreSi, line 79: Unless there's something I don't see, I don't think this needs to be translated 😅

newIDE/app/src/UI/List.js
```diff
@@ -203,6 +204,8 @@ export const ListItem = React.forwardRef<ListItemProps, ListItemRefType>(
       return null;
     };
 
+    const noPaddingStyle = props.noPadding ? { paddingLeft: 0 } : undefined;
```
AlexandreSi, line 207: `{ paddingLeft: 0 }` to move in the `styles` object

AlexandreSi (approved):
LGTM
The change is real!

AlexandreSi:
Should the add buttons of the project manager be hidden when something is searched? They bring noise when looking at the filtered results

4ian:
> Should the add buttons of the project manager be hidden when something is searched? They bring noise when looking at the filtered results

Yes! Great remark. Doing that.

newIDE/app/src/ProjectManager/ShortcutsReminder.js
```diff
@@ -0,0 +1,90 @@
...
+              {adaptAcceleratorString(
```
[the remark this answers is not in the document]
4ian: Done

newIDE/app/src/ProjectManager/ShortcutsReminder.js
4ian: Good catch! Fixed.

newIDE/app/src/UI/List.js
4ian: Done

4ian (commented):
Should have addressed everything.

AlexandreSi:
LGTM

PR #4654, opened 2022-12-09
Bug when picking folder is not updating input field

danired:
Hi,
When user selects a folder, the input field next to the button is not updated properly

https://user-images.githubusercontent.com/21989259/206730542-022af2e4-4609-4863-b12b-41bc010924f2.mp4

I made minor changes to fix that. I added this line: `setTextValue(filePath);`


4ian (approved):
Looks good! Thanks for noticing and fixing this :) 
The file was updated to be "semi controlled" but this button was not updated.

PR #4655, opened 2022-12-09
Use same design for open from dialog as for save to dialog

AlexandreSi:
- Use the same design for opening as for saving
- Display message in opening dialog regarding cloud projects
  - Web
  
  image

  - Desktop
  
  image


4ian:
Good idea!

newIDE/app/src/ProjectsStorage/StorageProviderListItem.js
```diff
@@ -25,7 +25,7 @@ const useListItemStyles = makeStyles(theme => {
   return {
     root: {
       border: `1px solid ${theme.palette.divider}`,
-      borderRadius: '5px',
+      borderRadius: 8,
```
4ian, line 28: 👍

4ian (approved):

PR #4660, opened 2022-12-11
Objects List - Filter objects by type

danired:
I added a new filter by type icon in the Objects List. Example video:

https://user-images.githubusercontent.com/21989259/206910083-812c71f4-cc86-4c0b-b879-37997bc89f24.mp4

#### Briefing

- New icon "Filter by type" next to the "Filter by tag" option
  - I began copying the same files and code from the Tags functionality.
- Added tooltips in both icons.
- Both filters work together.
- Types Chips working when selecting a filter.

#### Details

- I got some problems when recovering the "type" of the object, but I think now is right. If you see anything wrong ask me.
- I had to use "gdProject" and "gdObjectsContainer" in some cases, so I didn't want to use it in `EnumerateObjects.filterObjectsList`, because this function is used in more places. Then, I created a new `filterObjectsListEnhaced` that is only used in the Objects List.

### DIFF
--- a/newIDE/app/src/EventsBasedObjectEditor/EventBasedObjectChildrenEditor.js
+++ b/newIDE/app/src/EventsBasedObjectEditor/EventBasedObjectChildrenEditor.js
@@ -240,6 +240,9 @@ export default class EventBasedObjectChildrenEditor extends React.Component<
                 selectedObjectTags={[]}
                 onChangeSelectedObjectTags={selectedObjectTags => {}}
                 getAllObjectTags={() => []}
+                selectedObjectTypes={[]}
+                onChangeSelectedObjectTypes={selectedObjectTypes => {}}
+                getAllObjectTypes={() => []}
                 ref={
                   // $FlowFixMe Make this component functional.
                   objectsList => (this._objectsList = objectsList)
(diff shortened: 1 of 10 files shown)


4ian:
Hi! Thanks for opening this PR. Can you describe the problem this is addressing - have you seen a lot of users complaining about this lack of filter? Is this used for large games, and if so is this faster than doing a search with the search box?
Basically, I would need a bit more explanation about the problem we're solving here for users, as I'm unsure is this is something that will be used a lot :)

danired:
Hi @4ian .
Yes, I thought about large games with lot of resources, so it could be easy to filter for a specific type in the scene.
I didnt' find anyone complaining about that.

That's all. Perhaps you are right and we can close this PR.

4ian:
There are certainly useful things to do to, for example allowing objects to be sorted in folders/subfolders :) But while I appreciate your work here, I think this would make the interface more complex and the codebase too for something that would not be used a lot (this could even make some beginners worried about yet another icon on the interface to understand).

Let's avoid this for now, and still, thanks for your work there!

danired:
You are welcome! If you have anything relatively easy I can do tell me!

PR #4663, opened 2022-12-12
Disable project settings in new project setup dialog if creating from example

AlexandreSi:
image


D8H:
Should the message be directly inside the field?

4ian (approved):

4ian:
If it's too complex, I would almost just hide entirely the fields that are disabled. They just don't apply and we can afford not displaying them at all.

AlexandreSi:
As recommended by LuniMoon, I hid the options.

PR #4665, opened 2022-12-12
Fix extension function missing code files when exporting a project

D8H:
It always includes every function of an extension when the extension is used.


4ian (approved):
Ok for the first part, but worried about removing applyFunctionIncludeFilesDependencyTransitivity

4ian (approved):
After checking with you, good to go as dependencies between extensions was not supported so things are not worse (no regressions)

PR #4672, opened 2022-12-12
Includes the free functions of every extension that is actually used

D8H:
Event-based free functions only need to declare the include for their own file because the exporter look for used extension and  includes every of their free functions.

### DIFF
--- a/GDJS/GDJS/IDE/Exporter.cpp
+++ b/GDJS/GDJS/IDE/Exporter.cpp
@@ -54,8 +54,8 @@ bool Exporter::ExportWholePixiProject(
 
   auto usedExtensions = gd::UsedExtensionsFinder::ScanProject(project);
 
-  auto exportProject = [this, &exportedProject, &exportOptions, &helper](
-                           gd::String exportDir) {
+  auto exportProject = [this, &exportedProject, &exportOptions, &helper,
+                        &usedExtensions](gd::String exportDir) {
     bool exportForCordova = exportOptions["exportForCordova"];
     bool exportForFacebookInstantGames =
         exportOptions["exportForFacebookInstantGames"];
@@ -89,6 +89,7 @@ bool Exporter::ExportWholePixiProject(
         includesFiles);
 
     // Export files for object and behaviors
+    helper.ExportFreeFunctionIncludes(exportedProject, includesFiles, usedExtensions);
     helper.ExportObjectAndBehaviorsIncludes(exportedProject, includesFiles);
     helper.ExportObjectAndBehaviorsRequiredFiles(exportedProject, resourcesFiles);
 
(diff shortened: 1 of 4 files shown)


GDJS/GDJS/IDE/ExporterHelper.cpp
```diff
@@ -797,6 +799,37 @@ bool ExporterHelper::ExportIncludesAndLibs(
...
+    }
+  };
+
+  auto usedExtensions = gd::UsedExtensionsFinder::ScanProject(project);
+  for (auto &&usedExtension : usedExtensions) {
+    if (project.HasEventsFunctionsExtensionNamed(usedExtension)) {
+      auto &extension = project.GetEventsFunctionsExtension(usedExtension);
+
+      for (size_t functionIdex = 0;
+           functionIdex < extension.GetEventsFunctionsCount(); functionIdex++) {
```
ClementPasteau, line 816: typo functionIndex

GDJS/GDJS/IDE/ExporterHelper.cpp
```diff
@@ -115,7 +116,8 @@ bool ExporterHelper::ExportProjectForPixiPreview(
                  immutableProject.GetLoadingScreen().GetGDevelopLogoStyle(),
                  includesFiles);
 
-  // Export files for object and behaviors
+  // Export files for free function, object and behaviors
+  ExportFreeFunctionIncludes(exportedProject, includesFiles);
```
4ian, line 120: You also need to add this in `ExportWholePixiProject` (yes, this should be factored at some point)

GDJS/GDJS/IDE/ExporterHelper.cpp
```diff
@@ -797,6 +799,37 @@ bool ExporterHelper::ExportIncludesAndLibs(
...
 
+void ExporterHelper::ExportFreeFunctionIncludes(
+    gd::Project &project, std::vector<gd::String> &includesFiles) {
+  auto addIncludeFiles = [&](const std::vector<gd::String> &newIncludeFiles) {
+    for (const auto &includeFile : newIncludeFiles) {
+      InsertUnique(includesFiles, includeFile);
+    }
+  };
+
+  auto usedExtensions = gd::UsedExtensionsFinder::ScanProject(project);
```
4ian, line 810: There is `auto usedExtensions = gd::UsedExtensionsFinder::ScanProject(project);` called already in `ExportWholePixiProject`. To avoid doing it twice (and to avoid encouraging traversing the project too much, which can become costly at the project becomes larger), I think we can pass these as parameter. You just need also put this in `ExportProjectForPixiPreview`, which is fine.

GDJS/GDJS/IDE/Exporter.cpp
```diff
@@ -89,6 +89,7 @@ bool Exporter::ExportWholePixiProject(
         includesFiles);
 
     // Export files for object and behaviors
```
4ian, line 91: // Export files for free function, object and behaviors

4ian (approved):

PR #4674, opened 2022-12-12
Use a constant zoom speed, regardless of zoom factor

AlexandreSi:
Use multiplier factors when setting zoom
Also:
- Fix image preview for animation preview overflowing the dialog
  before: 
  image
  after:
  image

### DIFF
--- a/newIDE/app/src/InstancesEditor/index.js
+++ b/newIDE/app/src/InstancesEditor/index.js
@@ -182,7 +182,7 @@ export default class InstancesEditor extends Component<Props> {
     this.pixiRenderer.view.onwheel = (event: any) => {
       const zoomFactor = this.getZoomFactor();
       if (this.keyboardShortcuts.shouldZoom()) {
-        this.zoomOnCursorBy(-event.deltaY / 5000);
+        this.zoomOnCursorBy((-event.deltaY / 2000) * this.getZoomFactor());
       } else if (this.keyboardShortcuts.shouldScrollHorizontally()) {
         const deltaX = event.deltaY / (5 * zoomFactor);
         this.scrollBy(-deltaX, 0);
--- a/newIDE/app/src/InstancesEditor/index.js
+++ b/newIDE/app/src/InstancesEditor/index.js
@@ -182,7 +182,7 @@ export default class InstancesEditor extends Component<Props> {
     this.pixiRenderer.view.onwheel = (event: any) => {
       const zoomFactor = this.getZoomFactor();
       if (this.keyboardShortcuts.shouldZoom()) {
-        this.zoomOnCursorBy((-event.deltaY / 2000) * this.getZoomFactor());
+        this.zoomOnCursorBy(1 - event.deltaY / 100);
       } else if (this.keyboardShortcuts.shouldScrollHorizontally()) {
         const deltaX = event.deltaY / (5 * zoomFactor);
         this.scrollBy(-deltaX, 0);
@@ -516,7 +516,7 @@ export default class InstancesEditor extends Component<Props> {
   }
 
   zoomBy(value: number) {
-    this.setZoomFactor(this.getZoomFactor() + value);
+    this.setZoomFactor(this.getZoomFactor() * value);
   }
 
   /**
@@ -524,7 +524,7 @@ export default class InstancesEditor extends Component<Props> {
    */
   zoomOnCursorBy(value: number) {
     const beforeZoomCursorPosition = this.getLastCursorSceneCoordinates();
-    this.setZoomFactor(this.getZoomFactor() + value);
+    this.setZoomFactor(this.getZoomFactor() * value);
     const afterZoomCursorPosition = this.getLastCursorSceneCoordinates();
     // Compensate for the cursor change in position
     this.scrollBy(
--- a/newIDE/app/src/InstancesEditor/index.js
+++ b/newIDE/app/src/InstancesEditor/index.js
@@ -36,7 +36,10 @@ import {
 } from './InstancesEditorSettings';
 import Rectangle from '../Utils/Rectangle';
 import { isNoDialogOpened } from '../UI/MaterialUISpecificUtil';
-import { getContinuousZoomFactor } from '../Utils/ZoomUtils';
+import {
+  clampInstancesEditorZoom,
+  getContinuousZoomFactor,
+} from '../Utils/ZoomUtils';
 const gd: libGDevelop = global.gd;
 
 const styles = {
@@ -541,7 +544,7 @@ export default class InstancesEditor extends Component<Props> {
   setZoomFactor = (zoomFactor: number) => {
     this.props.onChangeInstancesEditorSettings({
       ...this.props.instancesEditorSettings,
-      zoomFactor: Math.max(Math.min(zoomFactor, 100), 0.01),
+      zoomFactor: clampInstancesEditorZoom(zoomFactor),
     });
   };
 
--- a/newIDE/app/src/InstancesEditor/index.js
+++ b/newIDE/app/src/InstancesEditor/index.js
@@ -36,6 +36,7 @@ import {
 } from './InstancesEditorSettings';
 import Rectangle from '../Utils/Rectangle';
 import { isNoDialogOpened } from '../UI/MaterialUISpecificUtil';
+import { getContinuousZoomFactor } from '../Utils/ZoomUtils';
 const gd: libGDevelop = global.gd;
 
 const styles = {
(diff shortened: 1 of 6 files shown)


D8H:
I guess the issue is here:
```
  zoomBy(value: number) {
    this.setZoomFactor(this.getZoomFactor() + value);
  }
```
A zoom should not be summed but multiplied.

newIDE/app/src/SceneEditor/index.js
```diff
@@ -983,14 +983,14 @@ export default class SceneEditor extends React.Component<Props, State> {
...
-      editor.zoomBy(0.1 * editor.getZoomFactor());
+      editor.zoomBy(1.1);
     }
   };
 
   zoomOut = () => {
     const { editor } = this;
     if (editor) {
-      editor.zoomBy(-0.1 * editor.getZoomFactor());
+      editor.zoomBy(0.9);
```
D8H, line 993: `0.9 * 1.1 = 0.99` which is not `1`.
May I suggest `2**(1/8) = 1.09050773266526` and `2**(-1/8) = 0.917004043204671`?

newIDE/app/src/InstancesEditor/index.js
```diff
@@ -182,7 +182,7 @@ export default class InstancesEditor extends Component<Props> {
     this.pixiRenderer.view.onwheel = (event: any) => {
       const zoomFactor = this.getZoomFactor();
       if (this.keyboardShortcuts.shouldZoom()) {
-        this.zoomOnCursorBy((-event.deltaY / 2000) * this.getZoomFactor());
+        this.zoomOnCursorBy(1 - event.deltaY / 100);
```
D8H, line 185: What if `deltaY >= 100`?

AlexandreSi: I'm not sure about that: should we respect the sensitivity set by the user for their mouse or should it be a predefined step?

D8H: What is the variable for the "sensitivity set by the user"?

AlexandreSi: I think deltaY is the one: https://adom.as/javascript-mouse-wheel/#:~:text=You%20might%20wonder,e.g.%2030.

AlexandreSi: I think that's the quantity of scroll of a wheel event.
A user that tuned their mouse with high sensitivity will maybe have a deltaY of 3 and a user with lower sensitivity something like 0.5.

newIDE/app/src/SceneEditor/index.js
AlexandreSi: Good remark. I'll see what I can do

newIDE/app/src/InstancesEditor/index.js
D8H: > I think that's the quantity of scroll of a wheel event. A user that tuned their mouse with high sensitivity will maybe have a deltaY of 3 and a user with lower sensitivity something like 0.5.

For a scroll, does it mean applying 3 times the unitary scroll displacement? In this case it's the same as applying the function 3 time with `deltaY = 1`. As zoom are multiplications, applying the zoom (`initialZoom * myFactor`) 3 times would mean doing `initialZoom * myFactor * myFactor * myFactor` which is `initialZoom * myFactor ** deltaY`.

newIDE/app/src/InstancesEditor/index.js
```diff
@@ -540,7 +541,7 @@ export default class InstancesEditor extends Component<Props> {
   setZoomFactor = (zoomFactor: number) => {
     this.props.onChangeInstancesEditorSettings({
       ...this.props.instancesEditorSettings,
-      zoomFactor: Math.max(Math.min(zoomFactor, 10), 0.01),
+      zoomFactor: Math.max(Math.min(zoomFactor, 100), 0.01),
```
D8H, line 544: Nitpick: it could be 128 and 1/128 as powers of 2 are used.

AlexandreSi: I have not stored the zoom factor as a power of 2 as you suggested. it's still the result of 2**(x/8)

newIDE/app/src/Utils/ZoomUtils.js
```diff
@@ -0,0 +1,13 @@
...
+export const zoomInFactor = stepZoomFactor;
+export const zoomOutFactor = Math.pow(stepZoomFactor, -1);
+const continuousZoomFactor = Math.sqrt(stepZoomFactor);
+
+// TODO: Use absolute value of signal that should represent either:
+// - Mouse sensitivity
+// - MacOS scroll acceleration
+// Signal is usually WheelEvent.deltaY
+export const getContinuousZoomFactor = (signal: number): number =>
+  Math.pow(continuousZoomFactor, Math.sign(signal));
```
D8H, line 13: Does it feel right without the `Math.sign`?

newIDE/app/src/Utils/ZoomUtils.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+
+const stepZoomFactor = Math.pow(2, 1 / 8);
+export const zoomInFactor = stepZoomFactor;
+export const zoomOutFactor = Math.pow(stepZoomFactor, -1);
+const continuousZoomFactor = Math.sqrt(stepZoomFactor);
```
D8H, line 6: As it's a constant that will only be calculated once `2 ** (1/16)` will be easier to understand as an half step.

newIDE/app/src/Utils/ZoomUtils.js
AlexandreSi: Nope, it's too quick, and the macos scroll acceleration makes it hard to normalize the value between, let's say, 0.5 and 3. Moreover I don't know what are the values for deltaY usually.

newIDE/app/src/InstancesEditor/index.js
AlexandreSi: I don't think it's needed at the moment. It could be an improvement for the future

D8H: Yes, it's probably better the way you did it, but it will still do a x2 every 8 steps (with some rounding errors that probably don't matter) so it won't go to x100 but it will go about x128. So if we reach the limit, we can still find some integer zooms from x128 but probably not from x100.
It's not a very common use case though.

newIDE/app/src/InstancesEditor/index.js
```diff
@@ -182,7 +186,7 @@ export default class InstancesEditor extends Component<Props> {
     this.pixiRenderer.view.onwheel = (event: any) => {
       const zoomFactor = this.getZoomFactor();
       if (this.keyboardShortcuts.shouldZoom()) {
-        this.zoomOnCursorBy(-event.deltaY / 5000);
+        this.zoomOnCursorBy(getContinuousZoomFactor(-event.deltaY));
```
D8H, line 189: `getContinuousZoomFactor` is a missleading name because it's not continuous, it's discrete but with smaller steps.
What about "getWheelZoomStep" ?

newIDE/app/src/SceneEditor/index.js
```diff
@@ -981,11 +982,11 @@ export default class SceneEditor extends React.Component<Props, State> {
   };
 
   zoomIn = () => {
-    if (this.editor) this.editor.zoomBy(0.1);
+    if (this.editor) this.editor.zoomBy(zoomInFactor);
```
D8H, line 985: We could maybe round the zoom to the nearest `2 ** (i/16)` to avoid the steps to be shifted after a fit to content, but I guess it's not blocking.

newIDE/app/src/Utils/ZoomUtils.js
```diff
@@ -0,0 +1,27 @@
+// @flow
+
+export const zoomStepBasePower = 1 / 16;
+
+const stepZoomFactor = Math.pow(2, 2 * zoomStepBasePower);
```
D8H, line 5: Is there a reason for not using the power operator `**`?

D8H (approved):

newIDE/app/src/Utils/ZoomUtils.js
AlexandreSi: 4ian once told me about the risk of having a navigator/browser that does not support this operator so I'm not using it anymore 😅 But I think that is more important for the runtime.
Better safe than sorry as they say

newIDE/app/src/SceneEditor/index.js
AlexandreSi: We'll see if the users notice this, in which case we could do this rounding indeed.

PR #4676, opened 2022-12-13
Fix the wiki extension page generator and make it use headers instead of extension files

D8H:


newIDE/app/scripts/extract-extensions-document.js
```diff
@@ -58,31 +58,32 @@ const generateAuthorNamesWithLinks = authors => {
 /**
  * Return the list of all extensions and their associated short headers
  * (useful as containing author public profiles information).
+ * @returns {Promise<Array<{ tier: 'community' | 'reviewed', shortDescription: string, authorIds: Array<string>, authors?: Array<{id: string, username: string}>, extensionNamespace: string, fullName: string, name: string, version: string, gdevelopVersion?: string, url: string, headerUrl: string, tags: Array<string>, category: string, previewIconUrl: string, eventsBasedBehaviorsCount: number, eventsFunctionsCount: number, helpPath: string, description: string, iconUrl: string}>>} A promise to all extension headers
```
D8H, line 61: Maybe there is a way to declare types inside a comment to factorize them, but I can't find the syntax.

4ian: You can use `typedef` for this:
`/** @typedef {{inPath: string; outPath: string;}} InOutPath */`

(this can be even used to import types from a .d.ts file, though we don't do that in this repository, but in the backend:
/** @typedef {import('../path/to/your.d.ts').SomeType} SomeType */

this could be used when one day we generate .d.ts files for GDevelop.js for example
)

4ian (approved):
Great stuff to use the header instead of the extension itself

